PR #163, opened 2017-06-07
Update ConvertService for external usage 👼

vdemeester:

### DIFF
--- a/cli/compose/convert/service.go
+++ b/cli/compose/convert/service.go
@@ -41,7 +41,7 @@ func Services(
 			return nil, errors.Wrapf(err, "service %s", service.Name)
 		}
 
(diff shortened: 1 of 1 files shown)


cli/compose/convert/service.go
```diff
@@ -51,7 +51,7 @@ func Services(
 	return result, nil
 }
 
-func convertService(
+func ConvertService(
```
dnephin: I guess this might complain about a missing comment/godoc string?

vdemeester: ah dang it you're right 😅

thaJeztah:
Hm, looks like golint doesn't like it;

```
cli/compose/convert/service.go:55:6:warning: func name will be used as convert.ConvertService by other packages, and that stutters; consider calling this Service (golint)
Exited with code 1
```

thaJeztah:
This one isn't in 17.06 - moving milestone

PR #164, opened 2017-06-07
Clarify ability to attach multiple times

mstanleyjones:
Clarify that you can attach to the contained process multiple times from different sessions on the Docker host. cc/ @aaronlehmann @mlaventure


mstanleyjones:
How in the world did this fail a test?

PR #171, opened 2017-06-09
removed the output leaked from stack remove error test case

adshmh:

**- What I did**
The 'TestContinueAfterError' test for the stack/remove command leaks an error message from an error test case to the output. This PR is to fix the leak to remove noise from output when running the tests.

**- How I did it**
Redirected the output of the command for the specific test to ioutil/discard

**- How to verify it**
- Run the tests (make -f docker.Makefile test)
- Without the proposed fix, the output will show the following:

> === RUN   TestContinueAfterError
>Error: Failed to remove some resources from stack: foo
>Usage:
>  rm STACK [STACK...] [flags]
>
>Aliases:
>  rm, remove, down
>
>--- PASS: TestContinueAfterError (0.00s)

- With the proposed fix, the output will not contain the error message from the error test case:
>=== RUN   TestContinueAfterError
>--- PASS: TestContinueAfterError (0.00s)
>PASS

Removed the message leaked from error test case for stack remove 

### DIFF
--- a/cli/command/stack/remove_test.go
+++ b/cli/command/stack/remove_test.go
@@ -3,6 +3,7 @@ package stack
 import (
 	"bytes"
 	"errors"
+	"io/ioutil"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


adshmh:
Thanks for the review. Since this has been approved, I can follow up with another PR for a similar fix in cli/command/secret package.

PR #176, opened 2017-06-09
New linters

dnephin:
Some of these linters didn't work on the first pass. I realized it was because `CGO_ENABLED=0` wasn't set.

Adds:
* `unused` - finds unused variables
* `misspell` - finds common spelling mistakes
* `unconvert` - finds unnecessary type convertions
* `unparam` - finds unused parameters (`_` is automatically ignored so functions that have extra params for interfaces are not errors)
* `gosimple` finds code simplifications

Right now this is a bit slower than I'd like it to be. I might look at https://github.com/alecthomas/gometalinter/issues/167 to see if it improves things. 

I suspect running things in-process would be much faster (instead of having to re-parse the AST each time), but that would be a much larger change.

### DIFF
--- a/cli/command/formatter/disk_usage.go
+++ b/cli/command/formatter/disk_usage.go
@@ -234,7 +234,6 @@ func (c *diskUsageImagesContext) Reclaimable() string {
 
 type diskUsageContainersContext struct {
 	HeaderContext
-	verbose    bool
 	containers []*types.Container
 }
 
@@ -297,7 +296,6 @@ func (c *diskUsageContainersContext) Reclaimable() string {
 
 type diskUsageVolumesContext struct {
 	HeaderContext
-	verbose bool
 	volumes []*types.Volume
 }
 
--- a/cli/command/formatter/disk_usage.go
+++ b/cli/command/formatter/disk_usage.go
@@ -65,7 +65,7 @@ reclaimable: {{.Reclaimable}}
 }
 
 func (ctx *DiskUsageContext) Write() (err error) {
-	if ctx.Verbose == false {
+	if !ctx.Verbose {
 		ctx.buffer = bytes.NewBufferString("")
 		ctx.preFormat()
 
--- a/cli/command/formatter/history.go
+++ b/cli/command/formatter/history.go
@@ -79,21 +79,18 @@ func (c *historyContext) ID() string {
 }
 
 func (c *historyContext) CreatedAt() string {
-	var created string
-	created = units.HumanDuration(time.Now().UTC().Sub(time.Unix(int64(c.h.Created), 0)))
-	return created
+	return units.HumanDuration(time.Now().UTC().Sub(time.Unix(c.h.Created, 0)))
 }
 
 func (c *historyContext) CreatedSince() string {
-	var created string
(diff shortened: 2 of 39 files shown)


cli/command/formatter/disk_usage.go
```diff
@@ -65,7 +65,7 @@ reclaimable: {{.Reclaimable}}
 }
 
 func (ctx *DiskUsageContext) Write() (err error) {
-	if ctx.Verbose == false {
+	if !ctx.Verbose {
```
aaronlehmann: Is one of the linters objecting to `ctx.Verbose == false`? I would personally probably write `!ctx.Verbose`, but in my view either are valid, and I certainly don't see potential for bugs or maintainability problems by using one over the other.

cli/command/formatter/history.go
```diff
@@ -79,21 +79,18 @@ func (c *historyContext) ID() string {
...
-	created = units.HumanDuration(time.Now().UTC().Sub(time.Unix(int64(c.h.Created), 0)))
+	created := units.HumanDuration(time.Now().UTC().Sub(time.Unix(c.h.Created, 0)))
 	return created + " ago"
 }
 
 func (c *historyContext) CreatedBy() string {
 	createdBy := strings.Replace(c.h.CreatedBy, "\t", " ", -1)
 	if c.trunc {
-		createdBy = stringutils.Ellipsis(createdBy, 45)
+		return stringutils.Ellipsis(createdBy, 45)
```
aaronlehmann: I don't have strong feelings about one of these forms versus the other, but I see the original as simpler because the function only returns from one place. If a linter is forcing an extra return, that seems a bit odd.

aaronlehmann:
Most of these changes are good. As usual, I want to make sure the linters are not too coercive. Pointed out two things that I'm not sure about.

aaronlehmann:
I suspect the problematic one is `gosimple`. The others seem fine to add - with the possible exception of `unparam`, but I don't feel strongly about this one.

cli/command/formatter/disk_usage.go
dnephin: Ya, this is one of the `gosimple` checks. It's a minor thing, but in a longer expression it's probably easier to read without the `== true/false`.

cli/command/formatter/history.go
dnephin: This one wasn't a linter (I'm pretty sure). I think I just noticed it while I was making changes in a nearby function.

aaronlehmann:
BTW, I'd really like to have a linter that suggests simpler equivalents of these statements:

```go
fmt.Sprintf("%s%s", string1, string2)
fmt.Sprintf("%s-suffix", input)
fmt.Sprintf("%d", input)
```

I see these all the time in Go code.

dnephin:
We could probably add those to `gosimple`. It's just a list of functions that check the AST for different patterns.

aaronlehmann:
Yeah, I think we can move forward. I wasn't thrilled about https://github.com/docker/cli/pull/176#discussion_r122082962 but it's a relatively minor concern.

cpuguy83:
Great! Merging.

PR #177, opened 2017-06-11
Add `rm-replica` command to to scale down a replicated service by removing a specific replica.

sungwonh:
This PR adds a `rm-replica` subcommand to `service` command to scale down a replicated service by removing a specific replica.
Please refer to [moby/moby#33426](https://github.com/moby/moby/issues/33426) for more details.

This PR depends on [moby/moby#33628](https://github.com/moby/moby/pull/33628).


dnephin:
I don't think it makes sense to add another command that does almost the exact same thing as scale. If this feature is necessary we should add a flag to the scale command.

cpuguy83:
Let's close this since the mob/moby PR was also closed and we can discuss design on the referenced issue.

Thanks!

Issue #167, opened 2017-06-08
Service creation with duplicate network name on different scopes (one in local and one in swarm)


yongtang:
**Description**

In command line `docker service create`, the network name is now converted to network ID then passed to daemon (vs. done in the daemon previously):
https://github.com/docker/cli/blob/ab3ea637b6f6ad766a61117f3040b82e45a59dc0/cli/command/service/opts.go#L348-L360

However, as `convertNetworks` uses `inspect` to do the network name conversion, a duplication error would be returned in case there are multiple network IDs available for the same network name.

In certain situations, it might be possible that there are multiple networks with the same name but with different scope, e.g, one `foo` in local and one `foo` in swarm.

For that I think it makes sense to change the way `convertNetworks` is done so that only networks related to swarm scope are searched.

**Additional information you deem important (e.g. issue happens only occasionally):**

This issue may need to be resolved in order to eventually fix network duplication issue in moby/moby/pull/30897 , moby/moby/issues/33561, moby/moby/issues/30242

dnephin:
Are you proposing to add a parameter to service inspect that allows the user to specify which scope they are inspecting? I think that sounds good.

PR #184, opened 2017-06-13
Use scope=swarm for service related network inspect and revendor docker/docker

yongtang:
This fix updates docker/docker to 4310f7da7e6bcd8185bf05e032f9b7321cfa6ea2 and use `scope=swarm` for service related network inspect.

This fix use `scope=swarm` for service related network inspect. The purpose is that, in case multiple networks with the same name exist in different scopes, it is still possible to obtain the network for services.

This fix is related to  docker/cli#167, moby/moby#30897, moby/moby#33561, moby/moby#30242

### DIFF
--- a/cli/command/network/inspect.go
+++ b/cli/command/network/inspect.go
@@ -6,6 +6,7 @@ import (
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/inspect"
+	"github.com/docker/docker/api/types"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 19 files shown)


aaronlehmann:
Sorry for my ignorance about network scoping, but I see that the node-local networks PR (https://github.com/moby/moby/pull/32981) says it added the `scope=swarm` option to `network create`. Will the change here work correctly with networks that were created before that PR, and networks that are not node-local?

The only comment I have on the code itself is that there are many places that specify `Verbose: false`, but these are not needed because `Verbose` will default to `false`.

dnephin:
I believe network scopes have existed since swarm was integrated, we just started exposing them to new endpoints more recently. So every swarm network should alredy have `scope=swarm` even if it was created before that PR.

yongtang:
Thanks all for the review. The PR has been updated and now the unnecessary `Verbose: false` has been removed.

For moby/moby#32981, it adds allows a node-local network to be promoted to the `swarm` scope. Once a previous node-local network is promoted to `swarm` scope, it will be possible to be used by service created. For that the `scope=swarm` option passed in this PR is consistent with moby/moby#32981

aaronlehmann:
LGTM

PR #160, opened 2017-06-06
fixes #46 Always exit 0 when remove image with force option

e11137:
Always exit with 0 if the -f force option is used

PR #187, opened 2017-06-13
Fix some problems with image force remove

dnephin:
Sorry, I didn't notice these until #160 was merged.

* the warning should go to stderr
* the warning message should not say "NotFound". The err could be anything, "not found" is just one option (also camel case for user errors is not great)
* fix the code so that `opts.force` doesn't have to be checked twice.

@thaJeztah @vdemeester

### DIFF
--- a/cli/command/image/remove.go
+++ b/cli/command/image/remove.go
@@ -59,9 +59,6 @@ func runRemove(dockerCli command.Cli, opts removeOptions, images []string) error
 	for _, img := range images {
 		dels, err := client.ImageRemove(ctx, img, options)
 		if err != nil {
-			if opts.force {
-				fmt.Fprintf(dockerCli.Out(), "NotFound: %s\n", img)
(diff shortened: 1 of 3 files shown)


thaJeztah:
> the warning should go to stderr

I assumed this was part of the fix; e.g., we recently had a PR in engine that changed a warning from stderr to stdout https://github.com/moby/moby/issues/29209

dnephin:
I don't think that's right. Both this warning and the one you referenced should go to stderr. If the desire was to hide a warning we should add a flag or environment variable to disable it. I don't think moving a warning to stdout because powershell treats stderr as an error is the right solution. stdout should only be for "normal program output".

If I were scripting a `docker image rm -f | awk ... | xargs echo "Removed ids: "` then sending a warning to stdout would break my script.

aaronlehmann:
LGTM

PR #188, opened 2017-06-13
Do not call `waitExitOrRemoved` when detaching

cpuguy83:
Before this change, `waitExitOrRemoved` is always called which
needlessly hits the API. This is particularly an issue with classic
swarm which has some issues with the events API (used for older API
daemons for `waitExitOrRemoved`).

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -107,7 +107,7 @@ func runRun(dockerCli *command.DockerCli, flags *pflag.FlagSet, opts *runOptions
 }
 
 // nolint: gocyclo
-func runContainer(dockerCli *command.DockerCli, opts *runOptions, copts *containerOptions, containerConfig *containerConfig) error {
+func runContainer(dockerCli *command.DockerCli, opts *runOptions, copts *containerOptions, containerConfig *containerConfig) (retErr error) {
 	config := containerConfig.Config
 	hostConfig := containerConfig.HostConfig
 	stdout, stderr := dockerCli.Out(), dockerCli.Err()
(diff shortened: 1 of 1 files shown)


cpuguy83:
ping @jlhawn

cli/command/container/run.go
```diff
@@ -184,7 +184,19 @@ func runContainer(dockerCli *command.DockerCli, opts *runOptions, copts *contain
 		}
 	}
 
-	statusChan := waitExitOrRemoved(ctx, dockerCli, createResponse.ID, copts.autoRemove)
+	if attach {
```
dnephin: The block above this is already `if attach {...}` could we merge these two blocks?

cpuguy83: You got it.

cpuguy83:
Looks ok here. The `gofmt` check would also fail here.

ripcurld0:
@cpuguy83 oh yeah, you're right, you removed the `}`. I knew it was the hour. LGTM 👍

cli/command/container/run.go
```diff
@@ -182,9 +182,19 @@ func runContainer(dockerCli *command.DockerCli, opts *runOptions, copts *contain
 		if err != nil {
 			return err
 		}
-	}
 
-	statusChan := waitExitOrRemoved(ctx, dockerCli, createResponse.ID, copts.autoRemove)
+		statusChan := waitExitOrRemoved(ctx, dockerCli, createResponse.ID, copts.autoRemove)
+		defer func() {
```
aaronlehmann: The old code seems to skip this if the user enters the detach escape sequence, or if only stdin is attached. Is the change intentional?

aaronlehmann:
There also seems to be a change in behavior where a non-attached run with auto-remove turned on will not wait for a container which fails to start to be removed before having the client exit. Is this correct? I don't have a good sense of whether changing this behavior would cause problems. It seems like it's better to wait for the resources associated with the container to be cleaned up before returning, since it failed to start, which is a different scenario from failing after starting. But I don't know if we've ever made any guarantees along these lines.

cli/command/container/run.go
cpuguy83: This only runs if `attach` is true.
I'll have to check for stdin only.

cpuguy83:
@aaronlehmann I don't think the old code waited around in any case if `-d` was set.

cpuguy83:
Ok, definitely more tricky than first look, but I think it's good now... PTAL.

aaronlehmann:
It still looks like a container with only stdin attached will wait now, whereas it wouldn't wait before (the `if !config.AttachStdout && !config.AttachStderr` special case used to skip reading from `statusChan`). Not sure which behavior is correct.

I think the code would be clearer about the `defer`. It seems like there are two cases where we want to block on `statusChan`: `autoRemove` after an error, and before a successful return with a non-detached container. These cases are more or less disjoint, and I think handling them in-place is better than a defer with exceptions for error cases and user-initiated detach.

vdemeester:
@thaJeztah @cpuguy83 @mlaventure what is the status here ?

mlaventure:
@vdemeester well it needs a rebase :stuck_out_tongue_closed_eyes: 

Also, I think I agree with @aaronlehmann it would be easier to read/maintain without the defer.

cpuguy83:
Did we end up solving this problem in a different way?

PR #193, opened 2017-06-14
Speed up testing with coverage

dnephin:
By running a `go test -i` on all the packages first the overall run time is significantly
decreased.

On my desktop this reduces the run time from 200s to 34s.
On circleCI this reduces the run time (which includes building an image) from 9 minutes to under 2 minutes.


aaronlehmann:
LGTM

We also do this for swarmkit and found a similar speedup.

PR #194, opened 2017-06-14
Run docs and man generation in CI

dnephin:
This way if we make changes to the scripts, or a docs change is invalid somehow we might be able to catch that problem in CI. We could probably do a little more validation on the output, but this is better than nothing.

Also cleanup the scripts a bit to be more consistent, and fail on errors.

### DIFF
--- a/scripts/docs/generate-man.sh
+++ b/scripts/docs/generate-man.sh
@@ -1,25 +1,14 @@
 #!/usr/bin/env bash
-#
-# Generate man pages for docker/docker
-#
-
-set -eu
+# Generate man pages for docker/cli
+set -eu -o pipefail
 
 mkdir -p ./man/man1
 
-MD2MAN_REPO=github.com/cpuguy83/go-md2man
-MD2MAN_COMMIT=$(grep -F "$MD2MAN_REPO" vendor.conf | cut -d' ' -f2)
-
-(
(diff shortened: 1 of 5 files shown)


scripts/docs/generate-man.sh
```diff
@@ -1,14 +1,11 @@
...
-
-set -eu
+# Generate man pages for docker/cli
+set -eu -o pipefail
 
 mkdir -p ./man/man1
 
 MD2MAN_REPO=github.com/cpuguy83/go-md2man
-MD2MAN_COMMIT=$(grep -F "$MD2MAN_REPO" vendor.conf | cut -d' ' -f2)
+MD2MAN_COMMIT=v1.0.6
```
thaJeztah: Should we bump the version in vendor.conf to the same? Was there a specific reason to not keep them in sync?

dnephin: There is no version in vendor.conf anymore. If there was we could remove this `go get` entirely.

dnephin: Ok, apparently it is in the `vendor.conf` ... let me try this again.

dnephin:
@thaJeztah thanks! I've fixed this up to use the code we already have in vendor. This is also faster because we don't have to go downloading anything.

PR #195, opened 2017-06-14
fix the issue `--help` doesn't work if daemon is not available

### LINKED ISSUE #114: `--help` doesn't work if daemon is not available
```
C:\Users\friism                                                                          
λ  docker deploy --help                                                                  
docker deploy is only supported on a Docker daemon with experimental features enabled    
C:\Users\friism                                                                          
λ  docker version                                                                        
Client:                                                                                  
 Version:      17.05.0-ce                                                                
 API version:  1.29                                                                      
 Go version:   go1.7.5                                                                   
 Git commit:   89658be                                                                   
 Built:        Thu May  4 21:43:09 2017                                                  
 OS/Arch:      windows/amd64                                                             
Error response from daemon: i/o timeout 
```

Speaking with @dnephin, he thinks it's related to `--experimental`

Ideally I'd be able to get help from the CLI without having access to the daemon.

Zebrilee:
Hi,

fixes #114 

- What I did

Let `--help` be available even is the daemon is not running
Then I add a condition to show an error message in case the daemon is not running

- How I did it
1. For now I just added a comment to the return in the `setHelpFunc` function ligne 95 
2. In the `isSupported` function I added an if statement. Because hasExperimental is false it means the daemon is off. 
Then I suggested a customized error message (line 250)

- How to verify it
1. run the "make -f docker.Makefile cross" command
2. go to the folder build and build with the "deploy --help" (In my case for mac ./docker-darwin-amd64 deploy --help )
3. Verify the normal behaviour , help visible and message if experimental features not enabled
4. Stop Docker 
5 run ./docker-darwin-amd64 deploy --help again 
6. Check if the help is shown with the new error message : 
"Can't connect to Docker daemon.
Docker deploy is only supported on a Docker daemon running and with experimental features enabled"

Make `--help` available even if the daemon is not available, add a custom error message

simonscat

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -202,6 +202,9 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 		if versions.LessThan(ping.APIVersion, cli.client.ClientVersion()) {
 			cli.client.UpdateClientVersion(ping.APIVersion)
 		}
+	} else {
+		// Default to true if we fail to connect to daemon
+		cli.server = ServerInfo{HasExperimental: true}
 	}
 
 	return nil


dnephin:
Thanks for the PR!

Unfortunately I don't think this is the correct solution, it's making an incorrect assumption about the state at the time of the error.

As I mentioned in the issue we should set the value of `ServiceInfo.HasExperimental` to `true` in `cli/command/cli.go` `Initialize()`. Something like this:

```go
if ping, err := cli.client.Ping(context.Background()); err == nil {
    ...
} else {
    // Default to true if we fail to connect to daemon
    cli.server = ServiceInfo{HasExperimental: true}
}
```

We might also want to set a default version here. I'm not sure if that's necessary. It would depend on what the version check does with the default.

Zebrilee:
Hi @dnephin ,
I saw your comment but did not understood it well. Thanks for the explaination.
I tried and it's ok but wouldn't it be nicer for the user to know that the deamon is not running?

dnephin:
Yes, that would be great!

Zebrilee:
Of course, you're right !

Zebrilee:
Hi, 
do you want me to modifiy something or is it good?

ps: this is during the docker Hackergarten too 
thx

PR #197, opened 2017-06-15
expose config and credential functionalities to 3rd parties

AkihiroSuda:
**- What I did**

expose config and credential fuhnctionalities to 3rd parties

Example: contrib/client-with-cli-config (but without whole cli)
See also moby/moby#30817

**- How I did it**

https://github.com/moby/moby/pull/30817

**- How to verify it**

```
$ go run ./contrib/client-with-cli-config/main.go  push example.com/foo/bar
$ go run ./contrib/client-with-cli-config/main.go  pull example.com/foo/bar
```

expose config and credential fuhnctionalities to 3rd parties

penguin


AkihiroSuda:
@dnephin @vieux @vdemeester @jdgiotta

registry/registry_test.go
```diff
@@ -1,16 +1,12 @@
...
-	"bytes"
 	"testing"
 
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"golang.org/x/net/context"
 
-	// Prevents a circular import with "github.com/docker/cli/cli/internal/test"
-	. "github.com/docker/cli/cli/command"
-	"github.com/docker/cli/cli/internal/test"
```
dnephin: Nice!

registry/registry.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/docker/cli/config/configfile"
+	"github.com/docker/docker/api/types"
+	registrytypes "github.com/docker/docker/api/types/registry"
+	"github.com/docker/docker/client"
+	dockerregistry "github.com/docker/docker/registry"
+)
+
+// ElectAuthServer returns the default registry to use (by asking the daemon)
+func ElectAuthServer(ctx context.Context, c client.APIClient) (string, []error, error) {
```
dnephin: I noticed you added an `error` return value, but it seems to always be `nil`. Can we remove it?

contrib/client-with-cli-config/main.go
```diff
@@ -0,0 +1,85 @@
+// demonstration of using REST API client with cli config (push/pull with credential)
```
dnephin: Maybe this could be a in a gist or a separate example repo that we can link to from docs?

cli/command/container/create.go
```diff
@@ -77,13 +78,19 @@ func pullImage(ctx context.Context, dockerCli command.Cli, image string, out io.
...
-	authConfig := command.ResolveAuthConfig(ctx, dockerCli, repoInfo.Index)
-	encodedAuth, err := command.EncodeAuthToBase64(authConfig)
+	authConfig, warns, err := registry.ResolveAuthConfig(ctx, dockerCli.Client(), dockerCli.ConfigFile(), repoInfo.Index)
+	for _, w := range warns {
+		fmt.Fprintf(dockerCli.Err(), "Warning: %v\n", w)
+	}
+	if err != nil {
+		return err
+	}
+	encodedAuth, err := registry.EncodeAuthToBase64(authConfig)
```
dnephin: I notice this code seems to be duplicated a bunch of times. I looked through the use of `authConfigs` and it seems that in almost every single case we either immediately encode them, or in a few cases we pass it around a bit, then encode it later on.

I'd like to suggest a refactoring which will remove this duplication. Since we're making this change to expose an interface I think it's probably the right time to fix this, as changing it later will break the interface.

Instead of running these 3 returns values, I propose we return a new `AuthConfig` struct (as noted elsewhere I believe that `err` is unnecessary and can be removed).

In `registry/registry.go`:

```go
type AuthConfig struct {
   config types.AuthConfig
   warnings []string
}

func (a *AuthConfig) PrintWarnings(out io.Writer) {
    ... // the for loop and fmt.Printf()
}

func (s *AuthConfig) Encode() (string, error) {
    return EncodeAuthToBase64(s.config)
}

func (s *AuthConfig) Config() types.AuthConfig {
    return s.config
}
```

Then consuming code could call:

```go
authConfig := registry.ResolveAuthConfig(ctx, dockerCli.Client(), dockerCli.ConfigFile(), repoInfo.Index)
authConfig.PrintWarnings(dockerCli.Err())

encodedAuth, err := authConfig.Encode()
if err != nil {
    return err
}
```

Or it could pass around the `authConfig` and worry about encoding later.

WDYT?

registry/config.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"github.com/docker/cli/config/configfile"
+	"github.com/docker/cli/config/credentials"
+	"github.com/docker/docker/api/types"
+)
+
+// GetAllCredentials returns all of the credentials stored in all of the
+// configured credential stores.
+func GetAllCredentials(configFile *configfile.ConfigFile) (map[string]types.AuthConfig, error) {
```
dnephin: I don't see this used. Is the code still using the old function?

registry/config.go
```diff
@@ -0,0 +1,56 @@
...
+func addAll(to, from map[string]types.AuthConfig) {
+	for reg, ac := range from {
+		to[reg] = ac
+	}
+}
+
+// CredentialsStore returns a new credentials store based
+// on the settings provided in the configuration file. Empty string returns
+// the default credential store.
+func CredentialsStore(configFile *configfile.ConfigFile, serverAddress string) credentials.Store {
```
dnephin: same here, not used?

registry/config.go
```diff
@@ -0,0 +1,56 @@
+package registry
```
dnephin: I think this file belongs in `config/`. It does return registry types, but it's operating on the config file more than anything else.

dnephin:
I need this change to support unit testing of some build functions. I'm going to resolve the conflicts and address my comments so I can cherry-pick these commits.

dnephin:
I addressed the config part in #227 I haven't touched the registry functions yet

AkihiroSuda:
Thank you, seems I should close this?

PR #199, opened 2017-06-15
removed logging from command/formatter package tests

adshmh:

**- What I did**
Removed logging from the tests in command/formatter package (for container, network, node, service, and volume commands tests)

**- How I did it**
Removed the Logf calls when there are no errors.

**- How to verify it**
- Run the tests (make -f docker.Makefile test)
- Output on the master branch (only shown for volume command tests, others are similar):

>=== RUN   TestVolumeContextWriteJSON
>--- PASS: TestVolumeContextWriteJSON (0.00s)
>        volume_test.go:156: Output: line 0: {"Driver":"foo","Labels":"","Links":"N/A","Mountpoint":"","Name":"foobar_baz","Scope
":"","Size":"N/A"}
>        volume_test.go:156: Output: line 1: {"Driver":"bar","Labels":"","Links":"N/A","Mountpoint":"","Name":"foobar_bar","Scope
":"","Size":"N/A"}
>=== RUN   TestVolumeContextWriteJSONField
>--- PASS: TestVolumeContextWriteJSONField (0.00s)
>        volume_test.go:176: Output: line 0: "foobar_baz"
>        volume_test.go:176: Output: line 1: "foobar_bar"

- Output with the changes in this PR:
>=== RUN   TestVolumeContextWriteJSON
>--- PASS: TestVolumeContextWriteJSON (0.00s)
>=== RUN   TestVolumeContextWriteJSONField
>--- PASS: TestVolumeContextWriteJSONField (0.00s)

Removed logging from command/formatter package tests

### DIFF
--- a/cli/command/formatter/container_test.go
+++ b/cli/command/formatter/container_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/stringid"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestContainerPsContext(t *testing.T) {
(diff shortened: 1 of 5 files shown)


cli/command/formatter/container_test.go
```diff
@@ -357,7 +357,6 @@ func TestContainerContextWriteJSON(t *testing.T) {
 		t.Fatal(err)
 	}
 	for i, line := range strings.Split(strings.TrimSpace(out.String()), "\n") {
-		t.Logf("Output: line %d: %s", i, line)
```
dnephin: I think it might be valuable to use this as a message in the assertions.

All the assert/require functions accept additional arguments which are printed as an additional message if the test fails. This can help debugging tests where there are assertions in loops, because the line numbers and error messages don't always give you enough information.

It would like something like this:

```go
msg := fmt.Sprintf("Output: line %d: %s", i, line)
var m map[string]interface{}
err := json.Unmarshal([]byte(line), &m)
require.NoError(t, err, msg)
assert.Equal(t, expectedJSONs[i], m, msg)
```

The same applies to the other changes

adshmh: Thanks for the review. I will update the PR.

adshmh: I have updated the PR: removal of Logf calls and addition of output details to assert calls.

PR #200, opened 2017-06-15
Speed up CI

dnephin:
Branched from #196 

Runtime is now under 3min (from ~9min)


dnephin:
circleCI seems a bit confused. I think it might be because I had enabled it on my fork to test this. I'm going to try closing this and opening a new PR.

PR #202, opened 2017-06-18
Unmarshal a number as a Number in RawInspectFallback

ripcurld0:
closes https://github.com/moby/moby/issues/32235

**- What I did**

Set the decoder to unmarshal numbers into an interface correctly.

**- How I did it**

Use the [`UseNumber` function](https://golang.org/pkg/encoding/json/#Decoder.UseNumber)

**- How to verify it**

Run unit-tests

### DIFF
--- a/cli/command/inspect/inspector.go
+++ b/cli/command/inspect/inspector.go
@@ -110,6 +110,7 @@ func (i *TemplateInspector) tryRawInspectFallback(rawElement []byte) error {
 	buffer := new(bytes.Buffer)
 	rdr := bytes.NewReader(rawElement)
 	dec := json.NewDecoder(rdr)
+	dec.UseNumber()
 
 	if rawErr := dec.Decode(&raw); rawErr != nil {
 		return errors.Errorf("unable to read inspect data: %v", rawErr)
--- a/cli/command/inspect/inspector_test.go
+++ b/cli/command/inspect/inspector_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 
 	"github.com/docker/docker/pkg/templates"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type testElement struct {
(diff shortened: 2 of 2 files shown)


ripcurld0:
@thaJeztah @yosifkit can you please review?

ripcurld0:
tests failed due to:
```
 ---> Running in 454533c654b7
W: Failed to fetch http://deb.debian.org/debian/dists/jessie-updates/main/binary-amd64/Packages  Hash Sum mismatch

E: Some index files failed to download. They have been ignored, or old ones used instead.
```
which seems to be unrelated to this change.
@dnephin correct me if I'm wrong.

cli/command/inspect/inspector_test.go
```diff
@@ -219,3 +224,24 @@ func TestIndentedInspectorRawElements(t *testing.T) {
 		t.Fatalf("Expected `%s`, got `%s`", expected, b.String())
 	}
 }
+
+// moby issue #32235
+func TestTemplateInspectorRawFallbackNumber(t *testing.T) {
+	b := new(bytes.Buffer)
+	tmpl, err := templates.Parse("{{.Size}} {{.Id}}")
+	if err != nil {
```
dnephin:
```go
require.NoError(t, err)
```

same below, also, `assert.Equal(t, ...)`

ripcurld0: Done! Thanks a lot 💯

cli/command/inspect/inspector_test.go
```diff
@@ -219,3 +226,19 @@ func TestIndentedInspectorRawElements(t *testing.T) {
 		t.Fatalf("Expected `%s`, got `%s`", expected, b.String())
 	}
 }
+
+// moby issue #32235
```
dnephin: Minor thing, but It would be good to be more explicit here by using the short-form url moby/moby#32235.  "moby" is a github org and project, so issue #32235 isn't specific enough. It might make sense today, but I'm not sure it will be so clear in a few months.

ripcurld0: Done. 🥇

cli/command/inspect/inspector.go
```diff
@@ -110,6 +110,8 @@ func (i *TemplateInspector) tryRawInspectFallback(rawElement []byte) error {
 	buffer := new(bytes.Buffer)
 	rdr := bytes.NewReader(rawElement)
 	dec := json.NewDecoder(rdr)
+	// unmarshal a number into an interface{} as a Number instead of float64.
+	dec.UseNumber()
```
dnephin: One more nit:

This comment doesn't tell us much about why, it just repeats the docstring on `UseNumber()` (which should be easy for someone to look up).

I think we could either remove this comment, or make it specific to this usage:

> // Ensure numbers are always printed in the same format as they are received from the API

or something like that.

cli/command/inspect/inspector_test.go
```diff
@@ -219,3 +226,19 @@ func TestIndentedInspectorRawElements(t *testing.T) {
 		t.Fatalf("Expected `%s`, got `%s`", expected, b.String())
 	}
 }
+
+// moby/moby#32235
+func TestTemplateInspectorRawFallbackNumber(t *testing.T) {
```
dnephin: Does this test fail without the change?

I'm not sure I understand why the json decode decides to convert it to a float in some cases.

I'm also not sure this test actually tests the change.  Isn't `tryRawInspectFallback()` only called when inspect receives a type without those fields? It seems we're never creating a json object.

Could this test hit `tryRawInspectFallback()` directly instead. The smaller surface area makes it easier to tell what the test is exercising.

Maybe make this a table test and try it with a float as well, so that we can verify that a float is printed using the same literal?

thaJeztah:
Might be worth updating the commit message to explain what we're fixing here (you can copy parts, or the entire description from https://github.com/moby/moby/issues/32235 (with a bit of wrapping/formatting)

ripcurld0:
@thaJeztah @dnephin I have updated the commit. I hope it is more clear than before. If it's still not good enough, let me know.

Thanks.

cli/command/inspect/inspector_test.go
ripcurld0: We can't hit `tryRawInspectFallback()` directly because it is not exposed by `TemplateInspector`.

dnephin: But the test is in the same package, so it doesn't have to be exported, right?

PR #205, opened 2017-06-19
Add 'docker service rollback' subcommand

### LINKED ISSUE #142: [RFC] Add "docker service rollback" subcommand
docker 1.13 added a `--rollback` flag to `docker service update` (https://github.com/moby/moby/pull/26421) to allow manually rolling back a service to a previous version.

This flag cannot be combined with other flags on the `service update` subcommand, which is confusing (https://github.com/moby/moby/issues/33444)

Looking at this functionality, I think this flag is essentially a subcommand "disguised" as a flag, and I suggest to add a `rollback` subcommand;

    docker service rollback <service-name|service-id>

which would be the equivalent of

    docker service update --rollback <service-name|service-id>

Given that this functionality is already present in the API, this can be implemented as a cli-only change.

ping @aaronlehmann @dnephin @tiborvass

redpanda:
Signed-off-by: Jimmy Leger <jimmy.leger@gmail.com>

fixes #142
PS: This bug was fixed during a Docker Hackergarten. 

**- What I did**

I add the docker service rollback subcommand

**- How I did it**

I extracted the rollback process from **runUpdate**

**- How to verify it**

Create a service with a single replica:

```bash
$ docker service create --name my-service -p 8080:80 nginx:alpine
```

Confirm that the service is running with a single replica:

```bash
$ docker service ls

xbw728mf6q0d        my-service          replicated          1/1                 nginx:alpine        *:8080->80/tcp
```

Update the service to use three replicas:

```bash
$ docker service update --replicas=3 my-service

$ docker service ls

xbw728mf6q0d        my-service          replicated          3/3                 nginx:alpine        *:8080->80/tcp
```

Now roll back the service to its previous version, and confirm it is
running a single replica again:

```bash
$ docker service rollback my-service

$ docker service ls

xbw728mf6q0d        my-service          replicated          1/1                 nginx:alpine        *:8080->80/tcp
```

Add docker service rollback subcommand

redpanda picture


cli/command/service/rollback.go
```diff
@@ -0,0 +1,27 @@
...
+
+func newRollbackCommand(dockerCli *command.DockerCli) *cobra.Command {
+	options := newServiceOptions()
+
+	cmd := &cobra.Command{
+		Use:     "rollback [OPTIONS]",
+		Aliases: []string{"rollback"},
+		Short:   "Rollback a service",
+		Args:    cli.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
dnephin: The pattern we use in these packages is to keep the inline function separate from adding flag.

The flags can be added at the end of `newRollbackCommand()` before `return cmd`

cli/command/service/rollback.go
```diff
@@ -0,0 +1,27 @@
...
+			flags := cmd.Flags()
+			flags.Bool("rollback", true, "")
```
dnephin: Since the command is rollback, we don't need a rollback flag, right?

I guess the calling code expects this?  We should refactor `runUpdate()` so that we have a function we can call directly that doesn't require such a flag.

cli/command/service/rollback.go
```diff
@@ -0,0 +1,27 @@
...
+			flags.Bool("force", false, "Force update even if no changes require it")
+			addServiceFlags(flags, options, nil)
```
dnephin: I don't think we want to add all the service flags. We only want the force flag added.

redpanda:
Hello @dnephin!

First thank for your feedback.

I don't make the choice to refactor the `runUpdate` function but just implement `runRollback` which only call `ServiceUpdate` endpoint with the needed parameters.

BTW I have some code duplication between `runUpdate` and `runRollback` see: https://github.com/docker/cli/pull/205/files#diff-09b79b8a3ba1feabc2d859a294bb1ac5R43
Should I create a function to initialize `types.ServiceUpdateOptions` ?

Thanks by advance for your feedback.

cli/command/service/rollback.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func newRollbackCommand(dockerCli *command.DockerCli) *cobra.Command {
+	options := newServiceOptions()
+
+	cmd := &cobra.Command{
+		Use:     "rollback [OPTIONS]",
+		Aliases: []string{"rollback"},
```
dnephin: This alias is not necessary because it's the same as the first word in `Use:`, so this line can be removed.

cli/command/service/rollback.go
```diff
@@ -0,0 +1,67 @@
...
+
+	service, _, err := apiClient.ServiceInspectWithRaw(ctx, serviceID, types.ServiceInspectOptions{})
+	if err != nil {
+		return err
+	}
+
+	updateOpts := types.ServiceUpdateOptions{}
+	spec := &service.Spec
+
+	if versions.LessThan(dockerCli.Client().ClientVersion(), "1.28") {
```
dnephin: I think we should just hide this command on earlier versions, so we don't need to check version here. You can do that by setting `Tags` on the `cobra.Command` above

```go
Tags: map[string]string{"version": "1.31"}
```

That way we can always set `updateOpts.Rollback` to `"previous"`

redpanda:
Yeah I'm working on it! 👍

cli/command/service/helpers.go
```diff
@@ -9,35 +9,34 @@ import (
...
 		err = <-errChan
 	}
 	return err
 }
 
 // warnDetachDefault warns about the --detach flag future change if it's supported.
-func warnDetachDefault(err io.Writer, clientVersion string, flags *pflag.FlagSet, msg string) {
-	if !flags.Changed("detach") && versions.GreaterThanOrEqualTo(clientVersion, "1.29") {
+func warnDetachDefault(err io.Writer, clientVersion string, msg string) {
+	if versions.GreaterThanOrEqualTo(clientVersion, "1.29") {
```
dnephin: Why this change? This doesn't look right

redpanda: You right! I'm gonna fix that

redpanda:
ping @vdemeester

cli/command/service/client_test.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (f *fakeClient) NodeList(ctx context.Context, options types.NodeListOptions) ([]swarm.Node, error) {
+	nodes := []swarm.Node{
+		*Node(NodeName("node-1")),
+	}
+	return nodes, nil
+}
+
+func (f *fakeClient) TaskList(ctx context.Context, options types.TaskListOptions) ([]swarm.Task, error) {
```
vdemeester: @redpanda not needed anymore 😛 ?

cli/command/service/helpers.go
```diff
@@ -15,20 +15,20 @@ import (
 
 // waitOnService waits for the service to converge. It outputs a progress bar,
 // if appropriate based on the CLI flags.
-func waitOnService(ctx context.Context, dockerCli *command.DockerCli, serviceID string, quiet bool) error {
+func waitOnService(ctx context.Context, cli command.Cli, serviceID string, quiet bool) error {
```
vdemeester: I would keep using `dockerCli` as variable name (because we do it elsewhere and to not *shadow* the `cli` package)

cli/command/service/client_test.go
redpanda: Good catch!

redpanda: This receiver is used in `service/ps_test.go`

cli/command/service/rollback.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return cmd
+}
+
+func runRollback(dockerCli command.Cli, flags *pflag.FlagSet, options *serviceOptions, serviceID string) error {
+	apiClient := dockerCli.Client()
+	ctx := context.Background()
+
+	service, _, err := apiClient.ServiceInspectWithRaw(ctx, serviceID, types.ServiceInspectOptions{})
+
```
dnephin: nit: this extra space is odd, normally the `if err` check is flush with the previous line which returns the err

cli/command/service/rollback.go
```diff
@@ -0,0 +1,67 @@
...
+		return err
+	}
+
+	spec := &service.Spec
+	updateOpts := types.ServiceUpdateOptions{
+		Rollback: "previous",
+	}
+
+	response, err := apiClient.ServiceUpdate(ctx, service.ID, service.Version, *spec, updateOpts)
+
```
dnephin: same nit

cli/command/service/rollback.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	for _, warning := range response.Warnings {
+		fmt.Fprintln(dockerCli.Err(), warning)
+	}
+
+	fmt.Fprintf(dockerCli.Out(), "%s\n", serviceID)
+
+	if options.detach {
+		warnDetachDefault(dockerCli.Err(), apiClient.ClientVersion(), flags, "rollbacked")
```
dnephin: hmm, "rollbacked" or "reverted" ?

cli/command/service/rollback_test.go
```diff
@@ -0,0 +1,107 @@
...
+			expectedDockerCliErr: "- warning 1\n- warning 2",
+		},
+	}
+
+	for _, tc := range testCases {
+		err := new(bytes.Buffer)
+		cli := test.NewFakeCli(&fakeClient{
+			serviceUpdateFunc: tc.serviceUpdateFunc,
+		})
+		cli.SetErr(err)
```
dnephin: minor: With the recent change to `NewFakeCli()` you don't need to do this anymore. The default is to set stderr to a bytes.Buffer.

To get the buffer you call `cli.ErrBuffer()`

cli/command/service/rollback.go
dnephin: cc @thaJeztah

thaJeztah: "rolled back"? I'd have to check in what context it's used, what it looks like (will do when I'm at my computer)

dnephin: "rolled back" sounds correct. I checked the context:

"Since --detach=false was not specified, tasks will be rolled back in the background."

cli/command/service/rollback.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	for _, warning := range response.Warnings {
+		fmt.Fprintln(dockerCli.Err(), warning)
+	}
+
+	fmt.Fprintf(dockerCli.Out(), "%s\n", serviceID)
+
+	if options.detach {
+		warnDetachDefault(dockerCli.Err(), apiClient.ClientVersion(), flags, "roll backed")
```
dnephin: "rolled back"

redpanda:
@vdemeester everything is OK

cli/command/service/rollback.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/spf13/pflag"
+)
+
+func newRollbackCommand(dockerCli command.Cli) *cobra.Command {
+	options := newServiceOptions()
+
+	cmd := &cobra.Command{
+		Use:   "rollback [OPTIONS] SERVICE",
+		Short: "Rollback a service",
+		Args:  cli.ExactArgs(1),
```
thaJeztah: For consistency with other commands, it would be good to support multiple services but let's leave that as a follow-up as it may need further discussion

thaJeztah:
Can you;

- create a `service_rollback.md` page for the CLI reference, having the usage description, and a simple example (https://github.com/docker/cli/tree/master/docs/reference/commandline)
- update the "related commands" sections in the `docker_service_xyz` files (e.g. https://github.com/docker/cli/blob/master/docs/reference/commandline/service_update.md#related-commands)
- this needs an update to the completion scripts (https://github.com/docker/cli/tree/master/contrib/completion), but this can be done separately (@albers @sdurrheimer @jphuynh)

redpanda:
Sure 👍

docs/reference/commandline/service_rollback.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Description
+
+Rollback a specified service from the swarm. This command has to be run
+targeting a manager node.
+
+## Examples
+
+Rollback the `redis` service:
+
```
mstanleyjones: I think you should explain what each of these commands is doing, and that only the last one is actually rolling back the service. It might be worth explaining what the state is before and after the rollback.

docs/reference/commandline/service_rollback.md
```diff
@@ -0,0 +1,71 @@
...
+```bash
+$ docker service create --name redis redis:3.0.6
```
thaJeztah: I wonder if "updating" the image from `3.0.6` to `3.0.7`, then rolling back could make people think the "rollback" is rolling back an image update, not the service definition. Although it makes a nice example to update the image version, and roll back that change, would it be better to use something else for illustration? e.g.

    ### Roll back a service to its previous version

    This example creates a service with a single replica, updates the
    service to use three replicas, and then rolls back the service to the
    previous version, having one replica.

    Create a service with a single replica:

    ```bash
    $ docker service create --name my-service -p 8080:80 nginx:alpine
    ```

    Confirm that the service is running with a single replica:

    ```bash
    $ docker service ls

    xbw728mf6q0d        my-service          replicated          1/1                 nginx:alpine        *:8080->80/tcp
    ```

    Update the service to use three replicas:

    ```bash
    $ docker service update --replicas=3 my-service

    $ docker service ls

    xbw728mf6q0d        my-service          replicated          3/3                 nginx:alpine        *:8080->80/tcp
    ```

    Now rollback the service to its previous version, and confirm it is
    running a single replica again:

    ```bash
    $ docker service rollback my-service

    $ docker service ls

    xbw728mf6q0d        my-service          replicated          1/1                 nginx:alpine        *:8080->80/tcp
    ```

docs/reference/commandline/service_rollback.md
thaJeztah: Agreed, this could use some additional information (also strip down the output to show just the relevant information, as mentioned in my other comments), something like;

    ### Roll back a service to a previous version

    In this example, a service is created using the `redis:3.0.6` image,
    updated to use the `redis:3.0.7` image, and then rolled back to its
    previous version.

    Create a service using the `redis:3.0.6` image:

    ```bash
    $ docker service create --name my-service redis:3.0.6
    ```

    Confirm that the service is running:

    ```bash
    $ docker service ps my-service

    ID                  NAME                IMAGE               NODE                DESIRED STATE       CURRENT STATE           ERROR               PORTS
    s2xak85rftfc        my-service.1        redis:3.0.6         moby                Running             Running 4 seconds ago
    ```

    Update the service to use the `redis:3.0.7` image:

    ```bash
    $ docker service update --image=redis:3.0.7 my-service
    ```

    The service is now updated to use the `redis:3.0.7` image:

    ```bash
    $ docker service ps my-service

    ID                  NAME                IMAGE               NODE                DESIRED STATE       CURRENT STATE                    ERROR               PORTS
    7jtl9bxnxnx1        my-service.1        redis:3.0.7         moby                Running             Running less than a second ago
    s2xak85rftfc         \_ my-service.1    redis:3.0.6         moby                Shutdown            Shutdown 10 seconds ago
    ```

    Roll back the service to its previous version. Rolling back the service
    reverts all changes made to the service in the last update. In this
    example, rolling back the service changes the service's image back to
    `redis:3.0.6`.

    ```bash
    $ docker service rollback my-service

    $ docker service ps my-service

    ID                  NAME                IMAGE               NODE                DESIRED STATE       CURRENT STATE            ERROR               PORTS
    8gwl1xoi0xh3        my-service.1        redis:3.0.6         moby                Running             Running 3 seconds ago
    7jtl9bxnxnx1         \_ my-service.1    redis:3.0.7         moby                Shutdown            Shutdown 4 seconds ago
    s2xak85rftfc         \_ my-service.1    redis:3.0.6         moby                Shutdown            Shutdown 2 minutes ago
    ```

docs/reference/commandline/service_rollback.md
thaJeztah: @mstanleyjones wdyt? ^^

mstanleyjones: Yep, agreed.

docs/reference/commandline/service_update.md
```diff
@@ -171,9 +171,9 @@ $ docker service update --mount-rm /somewhere myservice
 myservice
 ```
 
-### Rolling back to the previous version of a service 
+### Rolling back to the previous version of a service
```
mstanleyjones: Don't use gerunds (-ing words) in titles. How about "Roll back to the previous version of a service"? By the way, when you are using the verb form, it is "roll back" rather than "rollback".

docs/reference/commandline/service_rollback.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+## Description
+
+Rollback a specified service from the swarm. This command has to be run
+targeting a manager node.
+
+## Examples
+
+Rollback the `redis` service:
```
redpanda: I agree.

docs/reference/commandline/service_rollback.md
redpanda: Thanks!  I am going to use this example instead of the other one.

redpanda:
Should I implement the flag `--update-delay`.
This flag can be used with the `--rolback` flag `docker service update` command.
See: https://docs.docker.com/engine/reference/commandline/service_update/#rolling-back-to-the-previous-version-of-a-service

thaJeztah:
@redpanda I think it's best to do so in a follow up. This pull request is already in docs review, and provides the core functionality needed; we can enhance the functionality in a follow up 👍

docs/reference/commandline/service_update.md
thaJeztah: @redpanda I realise this wasn't added in your pull request, but can you address this comment as well?

docs/reference/commandline/service_rollback.md
```diff
@@ -0,0 +1,91 @@
...
+## Description
+
+Roll back a specified service to its previous version from the swarm. This command has to be run
+targeting a manager node.
+
+## Examples
+
+### Roll back to the previous version of a service
+
+This will revert the service to the configuration that was in place before the most recent docker service update command.
```
thaJeztah: "This" is a bit unclear, because here it doesn't have the introduction sentence that is present in the section it's copied from.

How about;

```Markdown
Use the `docker service rollback` command to roll back to the previous version
of a service. After executing this command, the service is reverted to the
configuration that was in place before the most recent `docker service update`
command.
````

File: docs/reference/commandline/service_rollback.md [hunk 20]
```diff
@@ -0,0 +1,94 @@
+---
+title: "service rollback"
```
redpanda: Should I replaced that line by: `title: service rollback` ?

docs/reference/commandline/service_rollback.md
```diff
@@ -0,0 +1,94 @@
...
+     pull requests against that repo. If you see this file in
+     another repository, consider it read-only there, as it will
+     periodically be overwritten by the definitive file. Pull
+     requests which include edits to this file in other repositories
+     will be rejected.
+-->
+
+# service rollback
+
+```Markdown
```
redpanda: I'm gonna fix it in `service_update.md`

docs/reference/commandline/service_rollback.md
redpanda: I just read some information about the front-matter.
Should I changed it like that ?
```markdown
---
title: service rollback
description: The service rollback command description and usage
keywords:
 - service
 - rollback
---
```
PS: I just keep the same front-matter used in other command line documentation but I can change it. ^^

mstanleyjones: It doesn't seem to hurt anything. We can probably do a bulk fix later on.

mstanleyjones:
I feel like I confused you. Don't change the `Markdown` hint to `markdown` on files unrelated to the original intent of this commit. We can fix the other ones later.

docs/reference/commandline/service_rollback.md
redpanda: Ok

redpanda:
You right! I will revert these changes. ^^

mstanleyjones:
Ping @thaJeztah

PR #206, opened 2017-06-19
Faster CI

dnephin:
Runtime is now under 3min (from ~9min)


dnephin:
I cleaned up the `cross` script to remove the duplication of jobs (that was in previous versions of this PR)

dockerfiles/Dockerfile.cross
```diff
@@ -1,18 +1,2 @@
...
-            libltdl-dev \
-            gcc-mingw-w64 \
-            parallel \
-            ;
-
-COPY    dockerfiles/osx-cross.sh /tmp/
-RUN     /tmp/osx-cross.sh
-ENV     PATH /osxcross/target/bin:$PATH
-
+FROM    dnephin/golang-cross@sha256:3a23c1791a8f9bb4a79e2708b2a87b51502ef24315170ed0fc27baa056c1a862
```
thaJeztah: should we try to get this in the `dockercore` org? I think we used that for other images used during build/ci; https://hub.docker.com/u/dockercore/

thaJeztah: We can still decide to keep the dockerfile for building it in this repository

dnephin: I'd be happy to move the image to `dockercore`, what org should receive the github repo?

I don't think we should keep a copy here if it's not used by CI. Keeping around files that aren't actually used just causes confusion. If someone needs to make a change they can go to the source repo.

We pin an exact digest here, so upgrades are easy.

dnephin: @tiborvass do you know if we already have a general purpose git repo for base images used in CI? 

We could move the `golang-cross` Dockerfile to that repo, if one exists.

dnephin:
Ok, this is now using the `docker` org on github, and the `dockercore`  org on hub.

dnephin:
ping for another review

PR #48, opened 2017-05-08
Synchronous CLI command for root CA rotation

cyli:
Provide command line tool to view and rotate swarm's currently CA root certificate.

Signed-off-by: Ying Li <ying.li@docker.com>

This code was originally in https://github.com/moby/moby/pull/32993.

```
root@ubuntu:~#  docker swarm ca
-----BEGIN CERTIFICATE-----
MIIBazCCARCgAwIBAgIUJPzo67QC7g8Ebg2ansjkZ8CbmaswCgYIKoZIzj0EAwIw
EzERMA8GA1UEAxMIc3dhcm0tY2EwHhcNMTcwNTAzMTcxMDAwWhcNMzcwNDI4MTcx
MDAwWjATMREwDwYDVQQDEwhzd2FybS1jYTBZMBMGByqGSM49AgEGCCqGSM49AwEH
A0IABKL6/C0sihYEb935wVPRA8MqzPLn3jzou0OJRXHsCLcVExigrMdgmLCC+Va4
+sJ+SLVO1eQbvLHH8uuDdF/QOU6jQjBAMA4GA1UdDwEB/wQEAwIBBjAPBgNVHRMB
Af8EBTADAQH/MB0GA1UdDgQWBBSfUy5bjUnBAx/B0GkOBKp91XvxzjAKBggqhkjO
PQQDAgNJADBGAiEAnbvh0puOS5R/qvy1PMHY1iksYKh2acsGLtL/jAIvO4ACIQCi
lIwQqLkJ48SQqCjG1DBTSBsHmMSRT+6mE2My+Z3GKA==
-----END CERTIFICATE-----

root@ubuntu:~# docker swarm ca --help

Usage:	docker swarm ca [OPTIONS]

Manage root CA

Options:
      --ca-cert pem-file          Path to the PEM-formatted root CA certificate to use for the new cluster
      --ca-key pem-file           Path to the PEM-formatted root CA key to use for the new cluster
      --cert-expiry duration      Validity period for node certificates (ns|us|ms|s|m|h) (default 2160h0m0s)
  -d, --detach                    Exit immediately instead of waiting for the root rotation to converge
      --external-ca external-ca   Specifications of one or more certificate signing endpoints
      --help                      Print usage
  -q, --quiet                     Suppress progress output
      --rotate                    Rotate the swarm CA - if no certificate or key are provided, new ones will be generated
```

[...]

PR #207, opened 2017-06-19
If `docker swarm ca` is not called with the `--rotate` flag, warn if other flags are passed

cyli:
If `docker swarm ca` is not called with the `--rotate` flag, the other flags, including cert expiry, will be ignored, so warn if a user attempts to use `docker swarm ca --cert-expiry` or something.

See https://github.com/docker/cli/pull/48#issuecomment-309448046.

image

### DIFF
--- a/cli/command/swarm/ca.go
+++ b/cli/command/swarm/ca.go
@@ -61,6 +61,11 @@ func runRotateCA(dockerCli command.Cli, flags *pflag.FlagSet, opts caOptions) er
 	}
 
 	if !opts.rotate {
+		for _, f := range []string{flagCACert, flagCAKey, flagCACert, flagExternalCA} {
+			if flags.Changed(f) {
+				return fmt.Errorf("`--%s` flag requires the `--rotate` flag to update the CA", f)
+			}
+		}
 		if swarmInspect.ClusterInfo.TLSInfo.TrustRoot == "" {
 			fmt.Fprintln(dockerCli.Out(), "No CA information available")
 		} else {
@@ -71,7 +76,7 @@ func runRotateCA(dockerCli command.Cli, flags *pflag.FlagSet, opts caOptions) er
 
 	genRootCA := true
 	spec := &swarmInspect.Spec
-	opts.mergeSwarmSpec(spec, flags)
+	opts.mergeSwarmSpec(spec, flags) // updates the spec given the cert expiry or external CA flag
 	if flags.Changed(flagCACert) {
 		spec.CAConfig.SigningCACert = opts.rootCACert.Contents()
 		genRootCA = false


cli/command/swarm/ca.go
```diff
@@ -61,6 +61,10 @@ func runRotateCA(dockerCli command.Cli, flags *pflag.FlagSet, opts caOptions) er
 	}
 
 	if !opts.rotate {
+		if flags.Changed(flagCACert) || flags.Changed(flagCAKey) || flags.Changed(flagCertExpiry) {
+			fmt.Fprintln(dockerCli.Out(), "Warning: because the `--rotate` flag was not provided, all other flags will be ignored")
+		}
```
dnephin: Any reason not to make this an error instead of a warning? 

If this remains a warning it should be on `stderr` (`dockerCli.Err()`). 

If it remains a warning maybe this could be split out into a separate function and we can print a warning for each flag? "all other flags" isn't really accurate because `-d` and `-q` are allowed, and we could always add other flags later.

Something like 

"WARNING: Flag --%s ignored because, you must --rotate to update the CA"

dnephin: I notice `flagExternalCA` is not being checked. Is that right?

cyli: @dnephin It just didn't particularly seem like an error to me, although I don't feel very strongly.  

> I notice `flagExternalCA` is not being checked. Is that right?

Ah sorry, that is also being addressed by `mergeSwarmSpec`.  I'll add that to the comment.

dnephin: As a user, if I run a command like 

```
docker swarm ca --ca-cert foo
```

I think I'm expecting an update, but the command acts as basically a "print the current value". That feels like an error to me.

cyli: I see, sure, I'll make that update.

cli/command/swarm/ca.go
```diff
@@ -61,6 +61,11 @@ func runRotateCA(dockerCli command.Cli, flags *pflag.FlagSet, opts caOptions) er
 	}
 
 	if !opts.rotate {
+		for _, f := range []string{flagCACert, flagCAKey, flagCACert, flagExternalCA} {
+			if flags.Changed(f) {
+				return fmt.Errorf("`--%s` flag ignored because you must `--rotate` to update the CA", f)
+			}
```
dnephin: I guess now that it's an error it's not ignored anymore. So maybe "--%s flag requires `--rotate` to update the CA" ?

cli/command/swarm/ca.go
```diff
@@ -61,6 +61,11 @@ func runRotateCA(dockerCli command.Cli, flags *pflag.FlagSet, opts caOptions) er
 	}
 
 	if !opts.rotate {
+		for _, f := range []string{flagCACert, flagCAKey, flagCACert, flagExternalCA} {
```
albers: @cyli `flagCACert` appears twice here. Is that correct?

cyli: Ah you're right, it's duplicated.  Will update.

PR #210, opened 2017-06-20
Add push-as flag to build command. Allow to build, tag then push in a single command

guillaumerose:
**- What I did**

I added a flag to the docker build command that push just after the build

**- How I did it**

With a new flag ```--push-as``` and re-using push command.

**- How to verify it**

```
$ docker build --push-as guillaumerose/foo .
Step 1/2 : FROM alpine
 ---> a41a7446062d
Step 2/2 : CMD echo "Hello world"
 ---> Using cache
 ---> faa24c25dd9b
Successfully built faa24c25dd9b
Successfully tagged guillaumerose/foo:latest
The push refers to a repository [docker.io/guillaumerose/foo]
3fb66f713c9f: Layer already exists
latest: digest: sha256:42ca234f64d11fbd76d351141d5f3a4a72acec2f769e4720d0c5be76cec4fdff size: 528
3fb66f713c9f: Layer already exists
sample: digest: sha256:42ca234f64d11fbd76d351141d5f3a4a72acec2f769e4720d0c5be76cec4fdff size: 528
$ docker run guillaumerose/foo
Hello world
```

Add push-as flag in build command.

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -36,6 +36,7 @@ type buildOptions struct {
 	context        string
 	dockerfileName string
 	tags           opts.ListOpts
+	pushAs         string
 	labels         opts.ListOpts
 	buildArgs      opts.ListOpts
 	extraHosts     opts.ListOpts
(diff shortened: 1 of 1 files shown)


aaronlehmann:
I think having a second flag that can specify a tag to create clashes with `-t`. I'd be more open to something like `--push` that pushes all the tags, but I'm still not sure that's justified. It seems simpler to keep `build` and `push` as separate operations. Are there any situations where needing to run two commands as part of this flow is burdensome?

cpuguy83:
I agree that keeping them separate is more desirable.

cpuguy83:
Ok, it seems like we agree that we don't want to take this one.

Thanks for submitting @guillaumerose, sorry it didn't work out!
If you are having a specific issue you'd like to solve, it may be helpful to open an issue to discuss the problem.

Thanks again!

PR #211, opened 2017-06-20
adding network_create tests 

ksouf:
**- What I did**
 adding unit tests to network.go
**- How I did it**
adding class netowr_test.go
adding client_test.go to network package
**- How to verify it**
make -f docker.Makefile cross

angry-aladdin

### DIFF
--- a/cli/command/network/cmd.go
+++ b/cli/command/network/cmd.go
@@ -8,8 +8,7 @@ import (
 )
 
 // NewNetworkCommand returns a cobra command for `network` subcommands
-// nolint: interfacer
-func NewNetworkCommand(dockerCli *command.DockerCli) *cobra.Command {
+func NewNetworkCommand(dockerCli command.Cli) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "network",
 		Short: "Manage networks",
--- a/cli/command/network/create.go
+++ b/cli/command/network/create.go
@@ -36,7 +36,7 @@ type createOptions struct {
 	ipamOpt     opts.MapOpts
 }
 
-func newCreateCommand(dockerCli *command.DockerCli) *cobra.Command {
+func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	options := createOptions{
 		driverOpts: *opts.NewMapOpts(nil, nil),
 		labels:     opts.NewListOpts(opts.ValidateEnv),
@@ -82,7 +82,7 @@ func newCreateCommand(dockerCli *command.DockerCli) *cobra.Command {
 	return cmd
 }
 
-func runCreate(dockerCli *command.DockerCli, options createOptions) error {
+func runCreate(dockerCli command.Cli, options createOptions) error {
 	client := dockerCli.Client()
 
 	ipamCfg, err := consolidateIpam(options.ipamSubnet, options.ipamIPRange, options.ipamGateway, options.ipamAux.GetAll())
@@ -232,13 +232,13 @@ func subnetMatches(subnet, data string) (bool, error) {
 
 	_, s, err := net.ParseCIDR(subnet)
 	if err != nil {
-		return false, errors.Errorf("Invalid subnet %s : %v", s, err)
+		return false, errors.Wrap(err, "invalid subnet")
 	}
 
 	if strings.Contains(data, "/") {
 		ip, _, err = net.ParseCIDR(data)
 		if err != nil {
-			return false, errors.Errorf("Invalid cidr %s : %v", data, err)
+			return false, err
 		}
 	} else {
 		ip = net.ParseIP(data)
--- a/cli/command/network/create_test.go
+++ b/cli/command/network/create_test.go
@@ -1,0 +1,178 @@
+package network
+
+import (
+	"bytes"
+	"io/ioutil"
+	"strings"
+	"testing"
+
+	"github.com/docker/cli/cli/internal/test"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/network"
+	"github.com/docker/docker/pkg/testutil"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/net/context"
+)
+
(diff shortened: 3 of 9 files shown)


cli/command/network/create_test.go
```diff
@@ -0,0 +1,182 @@
...
+			expectedError: "error creating network",
+		},
+		{
+			args: []string{"toto"},
+			flags: map[string]string{
+				"ip-range": "255.255.0.0/24",
+				"gateway":  "255.0.255.0/24",
+				"subnet":   "10.1.2.0.30.50",
+			},
+			expectedError: "Invalid subnet <nil> : invalid CIDR address: 10.1.2.0.30.50",
```
vdemeester: That error is weird :o (cc @dnephin)

cli/command/network/cmd.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // NewNetworkCommand returns a cobra command for `network` subcommands
 // nolint: interfacer
-func NewNetworkCommand(dockerCli *command.DockerCli) *cobra.Command {
+func NewNetworkCommand(dockerCli command.Cli) *cobra.Command {
```
dnephin: Please also remove the `// nolint: interfacer` line here (and anywhere else you change to use `command.Cli`).

We don't need the `nolint` once we change to use the interface.

cli/command/network/create_test.go
dnephin: Hmm yes. This is a bug! The code is using `s` instead of `subnet` in the error message. It looks like `net.ParseCIDR()` already includes the address, so the expected value should be  

```
invalid subnet: invalid CIDR address: 10.1.2.0.30.50
```

(invalid should start with a lowercase)

Could you update the code so to remove the `%s` that prints the `s` ?

cli/command/network/create_test.go
```diff
@@ -0,0 +1,182 @@
...
+		},
+		{
+			args: []string{"toto"},
+			flags: map[string]string{
+				"ip-range": "255.255.0.0.30/24",
+				"gateway":  "255.0.255.0/24",
+				"subnet":   "255.0.0.0/24",
+			},
+			expectedError: "Invalid cidr 255.255.0.0.30/24 : invalid CIDR address: 255.255.0.0.30/24",
```
dnephin: This one is not great either. I think this should be:

```
invalid CIDR address: 255.255.0.0.30/24"
```

It can return the err directly without wrapping it.

cli/command/network/create_test.go
```diff
@@ -0,0 +1,182 @@
...
+	for _, tc := range testCases {
+		buf := new(bytes.Buffer)
+		cmd := newCreateCommand(
+			test.NewFakeCli(&fakeClient{
+				networkCreateFunc: tc.networkCreateFunc,
+			}, buf),
+		)
+		cmd.SetArgs(tc.args)
+		for key, value := range tc.flags {
+			cmd.Flags().Set(key, value)
```
dnephin: `Set()` can actually error, so a good safety here is to call this with `require.NoError(...)` so that we know the tests are being setup correctly:

```go
require.NoError(t, cmd.Flags().Set(key, value))
```

cli/command/network/create_test.go
```diff
@@ -0,0 +1,182 @@
...
+		testutil.ErrorContains(t, cmd.Execute(), tc.expectedError)
+
+	}
+}
+func TestNetworkCreateWithFlags(t *testing.T) {
+	expectedDriver := "foo"
+	expectedOpts := []network.IPAMConfig{
+		{"192.168.4.0/24",
+			"192.168.4.0/24",
+			"192.168.4.1/24",
```
dnephin: The formatting is a bit strange here. I think the first entry in the array should be on a new line so that they are all aligned.

cli/command/network/create_test.go
```diff
@@ -0,0 +1,182 @@
...
+			map[string]string{},
+		},
+	}
+	buf := new(bytes.Buffer)
+	cli := test.NewFakeCli(&fakeClient{
+		networkCreateFunc: func(ctx context.Context, name string, createBody types.NetworkCreate) (types.NetworkCreateResponse, error) {
+			if createBody.Driver != expectedDriver {
+				return types.NetworkCreateResponse{}, errors.Errorf("expected driver %q, got %q", expectedDriver, createBody.Driver)
```
dnephin: I think you can `assert.Equal(t, ...)` here. You don't have to return the error.

cli/command/network/create_test.go
```diff
@@ -0,0 +1,182 @@
...
+			}
+			if !reflect.DeepEqual(createBody.IPAM.Config, expectedOpts) {
```
dnephin: same here, `assert.Equal()`

ksouf: done

cli/command/network/create_test.go
ksouf: done

cli/command/network/create_test.go
ksouf: done

cli/command/network/create_test.go
ksouf: done

cli/command/network/create_test.go
ksouf: done

cli/command/network/cmd.go
ksouf: done

cli/command/network/create_test.go
ksouf: done

ksouf:
@dnephin could you check one more time 
thanks :)

cli/command/network/create.go
```diff
@@ -232,13 +232,13 @@ func subnetMatches(subnet, data string) (bool, error) {
 
 	_, s, err := net.ParseCIDR(subnet)
 	if err != nil {
-		return false, errors.Errorf("Invalid subnet %s : %v", s, err)
+		return false, errors.Errorf("invalid subnet : %v", err)
```
dnephin: One more minor thing, this would be best with 

```go
errors.Wrap(err, "invalid subnet)
```

PR #213, opened 2017-06-20
Refactor and UI changes to `swarm ca` command

dnephin:
Split out a `swarmCAOptions` struct for options that are shared between the ca and update commands. Previously only 2 fields from the `swarmOptions` struct were being used by `CACommand`. 

Updated the usage string to better reflect the functionality provided by the command.

Change the "no trust root" message to an error. With this change the message will be displayed on `stderr` and the exit code will be non-zero. I believe this case should be an error because the intent of the command was to print some information, but the result was the information was not available to print. Also moved those lines to a new function, since they were duplicated.

Add some unit tests. for the new functions.

cc @cyli @aaronlehmann

### DIFF
--- a/cli/command/swarm/ca.go
+++ b/cli/command/swarm/ca.go
@@ -3,23 +3,22 @@ package swarm
 import (
 	"fmt"
 	"io"
-	"strings"
-
-	"golang.org/x/net/context"
-
 	"io/ioutil"
+	"strings"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/swarm/progress"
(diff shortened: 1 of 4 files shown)


dnephin:
oops, this is missing a test case for a partial update. Will add that now

aaronlehmann:
LGTM

cyli:
Sorry for the delay @dnephin - thanks!  LGTM modulo conflicts

dnephin:
rebased

cyli:
👍

thaJeztah:
this is ok for 17.07, or needs to go into current stable (17.06)?

dnephin:
17.07 is fine

PR #218, opened 2017-06-20
login: warn about --password on command line

tych0:
This isn't safe, since every commands arguments are available via
/proc/<pid>/cmdline.

Let's print a nasty warning and then add a --password-file option, so
people can use a password file instead if they want automated access.

Signed-off-by: Tycho Andersen <tycho@tycho.ws>

cute dog

### DIFF
--- a/cli/command/registry/login.go
+++ b/cli/command/registry/login.go
@@ -2,6 +2,7 @@ package registry
 
 import (
 	"fmt"
+	"io/ioutil"
 
 	"golang.org/x/net/context"
 
@@ -16,6 +17,7 @@ type loginOptions struct {
 	serverAddress string
 	user          string
 	password      string
+	passwordFile  string
 }
 
 // NewLoginCommand creates a new `docker login` command
@@ -39,13 +41,63 @@ func NewLoginCommand(dockerCli command.Cli) *cobra.Command {
 
 	flags.StringVarP(&opts.user, "username", "u", "", "Username")
 	flags.StringVarP(&opts.password, "password", "p", "", "Password")
+	flags.StringVarP(&opts.passwordFile, "password-file", "", "", "Password file whose contents are the password itself")
 
 	return cmd
+}
+
+// getPasswordFromFile reads the contents of the file, or stdin if the file is
+// == "-".
+//
+// It also trims off the last \n in the file, if it exists. Most people don't
+// have \ns in their password, and this allows stuff like echo "password" >
+// foo, without having to remember to pass -n, or vi, which can be configured
+// to automatically append newlines, etc.
+//
+// For users that do have a \n as the last character of their password, they
+// need to store it as \n\n. I think this conforms to the principle of least
+// surprise, but I could be wrong :)
+func getPasswordFromFile(dockerCli command.Cli, name string) (string, error) {
+	var err error
+	var raw []byte
+
+	if name == "-" {
+		raw, err = ioutil.ReadAll(dockerCli.In())
+	} else {
+		raw, err = ioutil.ReadFile(name)
+	}
+
+	if err != nil {
+		return "", err
+	}
+
+	contents := string(raw)
+	if contents[len(contents)-1] == '\n' {
+		contents = contents[:len(contents)-1]
+	}
+
+	return contents, nil
 }
 
 func runLogin(dockerCli command.Cli, opts loginOptions) error {
 	ctx := context.Background()
 	clnt := dockerCli.Client()
+
+	if opts.password != "" {
+		fmt.Fprintf(dockerCli.Err(), "Using --password via the CLI is insecure. Please use --password-file.\n")
+		if opts.passwordFile != "" {
+			return errors.Errorf("--password and --password-file are mutually exclusive")
+		}
+	}
+
+	if opts.passwordFile != "" {
+		contents, err := getPasswordFromFile(dockerCli, opts.passwordFile)
+		if err != nil {
+			return err
+		}
+
+		opts.password = contents
+	}
 
 	var (
 		serverAddress string


cpuguy83:
Needs a rebase.

tych0:
Done, looks like the previous merge broke the build, so I fixed that too, let me know if my fix is wrong; I don't know anything about this code :)

tych0:
Hmm. And it seems the linter is complaining about this... but if we try it the way it was (the way the linter is suggesting), it doesn't compile.

cli/command/registry/login.go
```diff
@@ -47,6 +51,34 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
 	ctx := context.Background()
 	clnt := dockerCli.Client()
 
+	if opts.password != "" {
+		fmt.Fprintf(os.Stderr, "Using --password via the CLI is insecure. Please use --password-file.\n")
```
dnephin: `os.Stderr` should be `dockerCli.Err()`

dnephin:
> but if we try it the way it was (the way the linter is suggesting), it doesn't compile.

It does compile, CI was green. Are you building outside of a container with an older version of GO maybe?

cli/command/registry/login.go
```diff
@@ -47,6 +51,34 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
...
+	}
+
+	if opts.passwordFile != "" {
```
dnephin: We should also support reading from stdin if `passwordFile == "-"`. There is code to handle that from an arg in other commands. I'm not sure if we have a function for it yet, or if maybe we could extract one.

cli/command/registry/login.go
```diff
@@ -47,6 +51,34 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
...
+		 * echo "password" > foo, without having to remember to pass
+		 * -n, or vi, which can be configured to automatically append
+		 * newlines, etc.
+		 *
+		 * For users that do have a \n as the last character of their
+		 * password, they need to store it as \n\n. I think this
+		 * conforms to the principle of least surprise, but I could be
+		 * wrong :)
+		 */
+		if contents[len(contents)-1] == '\n' {
```
dnephin: Anytime I have a long comment like this I think it's a good idea to split it into a new function. That way the comment can serve as a docstring on the function instead of cluttering up the logic within a function. Even if the function is only 3 lines, that's fine.

Maybe something like `normalizePasswordFileContent()` ?

dnephin:
I'm looking into the other 2 linter errors. I don't see them failing on master.

dnephin:
Linter errors are fixed in #221 if you want to rebase on that commit (or just wait until it gets merged). There's some explanation of the problem as well.

tych0:
On Tue, Jun 20, 2017 at 9:40 PM, Daniel Nephin <notifications@github.com>
wrote:

> but if we try it the way it was (the way the linter is suggesting), it
> doesn't compile.
>
> It does compile, CI was green. Are you building outside of a container
> with an older version of GO maybe?
>

I'm just typing "make" :). My host go version is 1.7.4.

dnephin:
You can use `make -f docker.Makefile ...`  to run in a conatiner: https://github.com/docker/cli#development

tych0:
Thanks; seems that still didn't work for me (it didn't use the current version, where there was an obvious syntax error, but instead compiled successfully). Anyway, I dropped that commit and addressed all your feed back. I'm happy to rebase once #221 goes in, just ping me.

cli/command/registry/login.go
```diff
@@ -39,14 +41,62 @@ func NewLoginCommand(dockerCli command.Cli) *cobra.Command {
...
+ *
+ * It also trims off the last \n in the file, if it exists. Most people don't
+ * have \ns in their password, and this allows stuff like echo "password" >
+ * foo, without having to remember to pass -n, or vi, which can be configured
+ * to automatically append newlines, etc.
+ *
+ * For users that do have a \n as the last character of their password, they
+ * need to store it as \n\n. I think this conforms to the principle of least
+ * surprise, but I could be wrong :)
+ */
```
dnephin: I believe the go convention is to use `//` single line comments for everything instead of `/* ... */`

tych0:
Fixed, thanks!

tych0:
I've just rebased this now that #221 is merged, so this should be good to go as well.

n4ss:
@vdemeester big improvement on the discussion we had. I'll open an issue to make sure that users simply don't provide passwords on the cli

tych0:
The CLI library doesn't seem to like -pf:

```
panic: shorthand is more than one character

goroutine 1 [running]:
panic(0xd24a00, 0xc4203b1050)
	/usr/lib/go-1.7/src/runtime/panic.go:500 +0x1a1
github.com/docker/cli/vendor/github.com/spf13/pflag.(*FlagSet).AddFlag(0xc420400900, 0xc420410000)
	/home/tycho/packages/go/src/github.com/docker/cli/vendor/github.com/spf13/pflag/flag.go:738 +0x61f
github.com/docker/cli/vendor/github.com/spf13/pflag.(*FlagSet).VarPF(0xc420400900, 0x135e120, 0xc42035d1f0, 0xe85c29, 0xd, 0xe7974b, 0x2, 0xeb2e26, 0x34, 0xc420403f40)
	/home/tycho/packages/go/src/github.com/docker/cli/vendor/github.com/spf13/pflag/flag.go:706 +0x141
github.com/docker/cli/vendor/github.com/spf13/pflag.(*FlagSet).VarP(0xc420400900, 0x135e120, 0xc42035d1f0, 0xe85c29, 0xd, 0xe7974b, 0x2, 0xeb2e26, 0x34)
	/home/tycho/packages/go/src/github.com/docker/cli/vendor/github.com/spf13/pflag/flag.go:712 +0x8e
github.com/docker/cli/vendor/github.com/spf13/pflag.(*FlagSet).StringVarP(0xc420400900, 0xc42035d1f0, 0xe85c29, 0xd, 0xe7974b, 0x2, 0x0, 0x0, 0xeb2e26, 0x34)
	/home/tycho/packages/go/src/github.com/docker/cli/vendor/github.com/spf13/pflag/string.go:42 +0xab
github.com/docker/cli/cli/command/registry.NewLoginCommand(0x13620e0, 0xc4201cf740, 0xc4203f8480)
	/home/tycho/packages/go/src/github.com/docker/cli/cli/command/registry/login.go:44 +0x2e1
github.com/docker/cli/cli/command/commands.AddCommands(0xc4203ae480, 0xc4201cf740)
	/home/tycho/packages/go/src/github.com/docker/cli/cli/command/commands/commands.go:51 +0x175
main.newDockerCommand(0xc4201cf740, 0xc42002e008)
	/home/tycho/packages/go/src/github.com/docker/cli/cmd/docker/docker.go:68 +0x4ca
main.main()
	/home/tycho/packages/go/src/github.com/docker/cli/cmd/docker/docker.go:166 +0xbd
```

so I've just dropped the short flag for now, and made them mutually exclusive.

thaJeztah:
Looks like I'm a bit late to the party; I'm

:+1: on a big red warning when using `--password`
:-1: on adding a `password-file` option

Having a password file means: hey! you should store your password in a plain-text file on your machine!

I don't think we should promote this

What if, instead, we allow the password to be piped in? Looks like it's currently not supported;

```bash
# or whatever means to pipe the password
$ echo $SECRET | docker login myname
Error: Cannot perform an interactive login from a non TTY device
```

tych0:
Done, see #270 and #271 (note that 271 has 270's commit in it too, because I update the message there).

jikuma:
please dont just have --password-file options but also have environment variables like DOCKER_ServerURL, DOCKER_Username and DOCKER_Secret. This can help in atomating the docker login process.

tych0:
Putting secrets in environment variables is just as unsafe as putting them in the command line (the task's initial environment is available via /proc/pid/environ). There are some other ways this leaks too: https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/

So the patches will only support taking passwords via stdin.

PR #219, opened 2017-06-21
hide `--detach` for docker < 17.05

vieux:
1st I added `--detach` to `docker service scale`
then while testing I realized that using docker 17.05 with a 17.03 client doesn't work with `--detach=true` since it's using an unknown filter to 17.03. so I hid the flag (as it should have been)

### DIFF
--- a/cli/command/service/helpers_test.go
+++ b/cli/command/service/helpers_test.go
@@ -1,0 +1,40 @@
+package service
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWarnDetachDefault(t *testing.T) {
+	var detach bool
+	flags := pflag.NewFlagSet("test", pflag.ContinueOnError)
+	addDetachFlag(flags, &detach)
+
+	var tests = []struct {
+		detach  bool
+		version string
+
+		expectWarning bool
+	}{
+		{true, "1.28", false},
+		{true, "1.29", false},
+		{false, "1.28", false},
+		{false, "1.29", true},
+	}
+
+	for _, test := range tests {
+		out := new(bytes.Buffer)
+		flags.Lookup(flagDetach).Changed = test.detach
+
+		warnDetachDefault(out, test.version, flags, "")
+
+		if test.expectWarning {
+			assert.NotEmpty(t, out.String(), "expected warning")
+		} else {
+			assert.Empty(t, out.String(), "expected no warning")
+		}
+	}
+}
--- a/cli/command/service/opts.go
+++ b/cli/command/service/opts.go
@@ -690,6 +690,11 @@ func buildServiceDefaultFlagMapping() flagDefaults {
 	return defaultFlagValues
 }
 
+func addDetachFlag(flags *pflag.FlagSet, detach *bool) {
+	flags.BoolVarP(detach, flagDetach, "d", true, "Exit immediately instead of waiting for the service to converge")
+	flags.SetAnnotation(flagDetach, "version", []string{"1.29"})
+}
+
 // addServiceFlags adds all flags that are common to both `create` and `update`.
 // Any flags that are not common are added separately in the individual command
 func addServiceFlags(flags *pflag.FlagSet, opts *serviceOptions, defaultFlagValues flagDefaults) {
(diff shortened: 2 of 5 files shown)


vieux:
Here is the current behavior:

```
$docker service update --replicas=8 test
test
Since --detach=false was not specified, tasks will be updated in the background.
In a future release, --detach=false will become the default.
```

When adding the flag as suggested:

```
$docker service update --detach=false --replicas=8 test
test
Error response from daemon: Invalid filter '_up-to-date'
```

vieux:
ping @thaJeztah

cli/command/service/scale.go
```diff
@@ -59,6 +70,18 @@ func runScale(dockerCli *command.DockerCli, args []string) error {
 		if err := runServiceScale(dockerCli, serviceID, scale); err != nil {
 			errs = append(errs, fmt.Sprintf("%s: %v", serviceID, err))
 		}
+
+		if options.detach {
+			if !flags.Changed("detach") && versions.GreaterThanOrEqualTo(dockerCli.Client().ClientVersion(), "1.29") {
+				fmt.Fprintln(dockerCli.Err(), "Since --detach=false was not specified, tasks will be scaled in the background.\n"+
```
thaJeztah: Should we add a global definition of this error somewhere?

thaJeztah: Hm; also (this is gonna be tricky); the warning should no longer be printed if the client talks to docker 17.09 (or whichever version changes the default)

vieux: This sentence is slightly different from the create one: "scaled" vs "updated" :d 

I agree 17.09 is probably a good time to do the switch. I'll work on a config file key for it in the meantime.

cli/command/service/scale.go
```diff
@@ -10,19 +10,28 @@ import (
...
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/versions"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 func newScaleCommand(dockerCli *command.DockerCli) *cobra.Command {
-	return &cobra.Command{
+	options := newServiceOptions()
```
dnephin: This shouldn't use service options (which is a huge struct).

We should create a new `scaleOptions` struct which would include just `detach` for now.

cli/command/service/scale.go
dnephin: Instead of defining the error in a global we should make this a function and pass in the different string as an arg.

```go
func warnDetachDefault(dockerCli command.Cli, flags pflag.FlatSet, msg string) { ... }
```

We can use that function from all 3 locations, and it would be easy to test.

cli/command/service/scale.go
```diff
@@ -10,19 +10,28 @@ import (
...
 		Args:  scaleArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runScale(dockerCli, args)
+			return runScale(dockerCli, cmd.Flags(), options, args)
 		},
 	}
+
+	flags := cmd.Flags()
+	flags.BoolVarP(&options.detach, "detach", "d", true, "Exit immediately instead of waiting for the service to converge")
+	flags.SetAnnotation("detach", "version", []string{"1.29"})
```
dnephin: We could move these two lines to an `addDetachFlag()` function and call it from `addServiceFlags()` (and here).

We should also create a constant `flagDetach` like we have for a bunch of  the other flags in `service/opts.go`

docs/reference/commandline/service_scale.md
```diff
@@ -16,11 +16,12 @@ keywords: "service, scale"
 # service scale
 
 ```markdown
-Usage:  docker service scale SERVICE=REPLICAS [SERVICE=REPLICAS...]
+Usage:  docker service scale [OPTIONS] SERVICE=REPLICAS [SERVICE=REPLICAS...]
 
 Scale one or multiple replicated services
 
 Options:
+  -d, --detach                             Exit immediately instead of waiting for the service to converge (default true)
```
thaJeztah: Can you also add it to the completion scripts, or want someone to take care of that in a follow-up?

cli/command/service/scale.go
```diff
@@ -59,6 +68,15 @@ func runScale(dockerCli *command.DockerCli, args []string) error {
 		if err := runServiceScale(dockerCli, serviceID, scale); err != nil {
```
dnephin: `runServiceScale` should accept this new `ctx`, so we don't create another one

cli/command/service/opts.go
```diff
@@ -453,6 +453,10 @@ func convertExtraHostsToSwarmHosts(extraHosts []string) []string {
 	return hosts
 }
 
+type scaleOptions struct {
+	detach bool
+}
```
dnephin: minor, but this struct should go in `scale.go` since it's only used in that one file. `serviceOptions` is a rare case where there are a lot of options shared between multiple commands.

cli/command/service/helpers_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}{
+		{true, "1.28", false},
+		{false, "1.28", false},
+		{false, "1.29", true},
+	}
+
+	for _, test := range tests {
+		out := new(bytes.Buffer)
+		cli.SetErr(out)
+		detach = test.detach
```
dnephin: I don't think this is triggering `flags.Changed()`. You can use `flags.Lookup(flagDetach).Changed = ...`

There's one more test case here, isn't there? `{true, "1.29", false)`, which I think will fail until ^ is fixed

thaJeztah:
@vieux looks like this PR fixes https://github.com/moby/moby/issues/32368 ?

vieux:
split the PR in two, here is the bugfix/refactor, new feature in https://github.com/docker/cli/pull/243

tiborvass:
LGTM

PR #221, opened 2017-06-21
Fix lint errors merged while new lint branch was in PR

dnephin:
There was a panic in `gosimple` which is why the `newAttach` change is being reverted and whitelisted for now. It was fixed in https://github.com/dominikh/go-tools/commit/769824fffc0626f9febdcdeb1c80d3644c12e788 so we just have to pickup that fix in gometalinters vendored copy. I will do that tomorrow.

The other 2 were from PRs that were merged at the same time as the new linter PR, so the new linters had never run on them.

### DIFF
--- a/cli/command/service/opts.go
+++ b/cli/command/service/opts.go
@@ -354,7 +354,11 @@ func convertNetworks(ctx context.Context, apiClient client.NetworkAPIClient, net
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


aaronlehmann:
LGTM, but curious why this doesn't seem to be failing other PRs. Are the new linters not running on the whole codebase in each PR?

dnephin:
It's because of this https://github.com/alecthomas/gometalinter/issues/299. I'm going to look at getting a fix into gometalinter for that soon.

PR #222, opened 2017-06-21
Explain `stack deploy` with multiple Compose files

thaJeztah:
The Docker Compose docs suggest using a separate override
configuration file for production-specific settings, but
it is not obvious how to feed this to `docker stack deploy`,
which only supports a single Compose file as input. Thus,
we now describe how to do this by merging the configuration
files with `docker-compose config`.

carry of https://github.com/moby/moby/pull/33290 (thanks @denisw!)


thaJeztah:
ping @vdemeester PTAL

PR #227, opened 2017-06-21
expose config credentials without needing the Cli

dnephin:
Related to moby/moby#30817

Move `GetAllCredentials()`, and `CredentialsStore()` from the `DockerCli` to the `ConfigFile` struct, so that are easier to use externally.

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -68,48 +68,42 @@ func Load(configDir string) (*configfile.ConfigFile, error) {
 		configDir = Dir()
 	}
 
-	configFile := configfile.ConfigFile{
-		AuthConfigs: make(map[string]types.AuthConfig),
-		Filename:    filepath.Join(configDir, ConfigFileName),
-	}
+	filename := filepath.Join(configDir, ConfigFileName)
+	configFile := configfile.NewConfigFile(filename)
 
 	// Try happy path first - latest config file
-	if _, err := os.Stat(configFile.Filename); err == nil {
-		file, err := os.Open(configFile.Filename)
+	if _, err := os.Stat(filename); err == nil {
+		file, err := os.Open(filename)
 		if err != nil {
-			return &configFile, errors.Errorf("%s - %v", configFile.Filename, err)
+			return configFile, errors.Errorf("%s - %v", filename, err)
 		}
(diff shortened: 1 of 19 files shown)


cli/config/config.go
```diff
@@ -118,3 +111,16 @@ func Load(configDir string) (*configfile.ConfigFile, error) {
 	}
 	return &configFile, nil
 }
+
+// LoadDefaultConfigFile attempts to load the default config file and returns
+// an initialized ConfigFile struct if none is found.
+func LoadDefaultConfigFile(err io.Writer) *configfile.ConfigFile {
+	configFile, e := Load(Dir())
+	if e != nil {
+		fmt.Fprintf(err, "WARNING: Error loading config file:%v\n", e)
```
aaronlehmann: I realize this code is only moved, but it looks like a space is missing between `:` and `%v`.

dnephin:
Fixed the string format and added a couple more unit tests.

thaJeztah:
Given that @dnephin is on PTO this week, perhaps someone can carry those nits?

vdemeester:
I'll carry this one 😉

cli/config/credentials/native_store_test.go
```diff
@@ -147,102 +129,66 @@ func TestNativeStoreAddInvalidCredentials(t *testing.T) {
...
-	}
-	if a.IdentityToken != "" {
-		t.Fatalf("expected identity token to be empty, got %s", a.IdentityToken)
-	}
-	if a.Email != "foo@example.com" {
-		t.Fatalf("expected email `foo@example.com`, got %s", a.Email)
+	expected := types.AuthConfig{
+		Username: "foo",
+		Password: "bar",
+		Email:    "foo@example.com",
```
thaJeztah: Do we still have to test for the `Email` field here?

vdemeester: Maybe not but could be handled in a follow-up 👼

dnephin: If the credential helper stores email, I think we should still test for it.

dnephin:
@vdemeester thanks for making that `New()` change. Last commit LGTM

PR #230, opened 2017-06-22
Remove specific config info about userns-remap

mstanleyjones:
When https://github.com/docker/docker.github.io/pull/3678 is merged and live on docs.docker.com, we should merge this PR which removes the specific config info from the CLI reference.

cc/ @thaJeztah @estesp 

baby hummingbird

### DIFF
--- a/docs/reference/commandline/dockerd.md
+++ b/docs/reference/commandline/dockerd.md
@@ -969,14 +969,14 @@ $ sudo dockerd \
 
 The currently supported cluster store options are:
 
-| Option                | Description |
-|-----------------------|-------------|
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/dockerd.md
```diff
@@ -1005,152 +1005,18 @@ plugin](../../extend/plugins_authorization.md) section in the Docker extend sect
...
-> content is downloaded. This design preserves the same performance for `docker
-> pull`, `docker push`, and container startup as users expect with
-> user namespaces disabled.
-
-##### Start the daemon with user namespaces enabled
-
-To enable user namespace support, start the daemon with the
-`--userns-remap` flag, which accepts values in the following formats:
-
- - uid
```
thaJeztah: I think the new content does not yet;

- have a mention of the flag (`--userns-remap`); although we _recommend_ using `daemon.json`, I think we should _mention_ the flag
- also we should document the accepted formats (perhaps I overlooked in the new content)

PR #231, opened 2017-06-22
Incrementally sending build context

tonistiigi:
Adds support for incremental context updates to `docker build` https://github.com/moby/moby/pull/32677 . The feature is behind a `--stream=true` flag atm, mostly for a testing period and in case there is some reason to use the old method. When we feel confident enough we can just switch the default value and users shouldn't see any changes except things being faster.

I also noticed that it vendors in chrootarchive and reexec package. These are not really needed but require some refactoring. Either splitting the receiver and sender part to a separate package, or passing an archiver instance into the transfer function. I can do that in a follow-up with next update.

@dnephin @vieux


cli/command/cli.go
```diff
@@ -200,7 +200,7 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 
 		// if server version is lower than the current cli, downgrade
 		if versions.LessThan(ping.APIVersion, cli.client.ClientVersion()) {
-			cli.client.UpdateClientVersion(ping.APIVersion)
+			cli.client.NegotiateAPIVersionPing(ping)
```
dnephin: `NegotiateAPIVersionPing()` does the minimum version check as well, so the previous `if ping.APIVersion == ""` can be removed.

cli/command/image/build.go
```diff
@@ -133,6 +146,10 @@ func NewBuildCommand(dockerCli *command.DockerCli) *cobra.Command {
 	flags.SetAnnotation("squash", "experimental", nil)
 	flags.SetAnnotation("squash", "version", []string{"1.25"})
 
+	flags.BoolVar(&options.stream, "stream", false, "Stream attaches to server to negotiate build context")
+	flags.SetAnnotation("stream", "experimental", nil)
+	flags.SetAnnotation("stream", "version", []string{"1.31"})
```
dnephin: I think we normally don't set the version until it's out of experimental. I'm not sure how it works if we set both.  Maybe it just works?

cli/command/image/build.go
```diff
@@ -250,16 +272,34 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
 		}
-		// Wrap the tar archive to replace the Dockerfile entry with the rewritten
-		// Dockerfile which uses trusted pulls.
-		buildCtx = replaceDockerfileTarWrapper(ctx, buildCtx, relDockerfile, translator, &resolvedTags)
+		if buildCtx != nil {
+			// Wrap the tar archive to replace the Dockerfile entry with the rewritten
+			// Dockerfile which uses trusted pulls.
+			buildCtx = replaceDockerfileTarWrapper(ctx, buildCtx, relDockerfile, translator, &resolvedTags)
+		} else if dockerfileCtx != nil {
+			newDockerfile, _, err := rewriteDockerfileFrom(context.Background(), dockerfileCtx, translator)
```
dnephin: Shouldn't this use `ctx` for the context?

cli/command/image/build.go
```diff
@@ -250,16 +272,34 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
 		}
 	}
 
-	ctx := context.Background()
+	if options.stream && dockerfileCtx == nil {
```
dnephin: Is `dockerfileCtx == nil` the same as `!options.dockerfileFromStdin()` ?  Could we use the latter so it's more clear what this means?

cli/command/image/build.go
```diff
@@ -250,16 +272,34 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
+		f, err := os.Open(relDockerfile)
```
dnephin: minor nit: you can assign directly to `dockerfileCtx` here, no need to create a temporary variable.

cli/command/image/build.go
```diff
@@ -250,16 +272,34 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
 			return TrustedReference(ctx, dockerCli, ref, nil)
 		}
-		// Wrap the tar archive to replace the Dockerfile entry with the rewritten
-		// Dockerfile which uses trusted pulls.
-		buildCtx = replaceDockerfileTarWrapper(ctx, buildCtx, relDockerfile, translator, &resolvedTags)
+		if buildCtx != nil {
+			// Wrap the tar archive to replace the Dockerfile entry with the rewritten
+			// Dockerfile which uses trusted pulls.
+			buildCtx = replaceDockerfileTarWrapper(ctx, buildCtx, relDockerfile, translator, &resolvedTags)
+		} else if dockerfileCtx != nil {
```
dnephin: This will be `true` when Dockerfile is from stdin, but we're ignoring the work done by the "replace Dockerfile if added dynamically" branch in that case. I'm not sure if this is a bug, or just confusing, but I think something needs to change here.

cli/command/image/build.go
```diff
@@ -268,7 +308,59 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
 		progressOutput = &lastProgressOutput{output: progressOutput}
 	}
 
-	var body io.Reader = progress.NewProgressReader(buildCtx, progressOutput, 0, "", "Sending build context to Docker daemon")
+	if dockerfileCtx != nil && buildCtx == nil {
+		buildCtx = dockerfileCtx
+	}
```
dnephin: At this point we've rewritten and swapped out these two variables I'm not sure what this branch is really handling.

This will be true `if ("dockerfile is from stdin" OR "trust is enabled") AND (...?)` . From what I can tell `buildCtx` is always not-nil at this point, no?

cli/command/image/build.go
```diff
@@ -250,16 +272,34 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
 
 	var resolvedTags []*resolvedTag
 	if command.IsTrusted() {
 		translator := func(ctx context.Context, ref reference.NamedTagged) (reference.Canonical, error) {
 			return TrustedReference(ctx, dockerCli, ref, nil)
 		}
-		// Wrap the tar archive to replace the Dockerfile entry with the rewritten
-		// Dockerfile which uses trusted pulls.
-		buildCtx = replaceDockerfileTarWrapper(ctx, buildCtx, relDockerfile, translator, &resolvedTags)
+		if buildCtx != nil {
```
dnephin: Isn't `buildCtx` always non-nil at this point? In what case is it nil?

cli/command/image/build.go
```diff
@@ -268,7 +308,59 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
 		progressOutput = &lastProgressOutput{output: progressOutput}
 	}
 
-	var body io.Reader = progress.NewProgressReader(buildCtx, progressOutput, 0, "", "Sending build context to Docker daemon")
+	if dockerfileCtx != nil && buildCtx == nil {
+		buildCtx = dockerfileCtx
+	}
+	var s *session.Session
+	if isSessionSupported(dockerCli) {
```
dnephin: This function is already way too large. There's no way we can probably unit test most of it. We really should not be adding any significant code to it.

Could we make the rest of this (up to line 362, and the `if s != nil ` block) into a few new functions? Even if the arg list is long, at this point I think it's still an improvement. We can work on refactoring args into structs in a follow up.

Maybe even a new file along with `sizeProgress` and `bufferedWriter` ? That way all the session handling stuff is together and not mixed up with other build things.

cli/command/image/build.go
```diff
@@ -299,13 +391,26 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
+		}()
+		buildOptions.SessionID = s.UUID()
 	}
 
 	response, err := dockerCli.Client().ImageBuild(ctx, body, buildOptions)
 	if err != nil {
 		if options.quiet {
 			fmt.Fprintf(dockerCli.Err(), "%s", progBuff)
 		}
+		cancel()
```
dnephin: Isn't this already handled by the `defer cancel()` ?

cli/command/image/build.go
```diff
@@ -268,7 +308,59 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
+		buf := newBufferedWriter(syncDone, buildBuff)
+		defer func() {
+			select {
+			case <-buf.flushed:
+			case <-ctx.Done():
+			}
+		}()
+		buildBuff = buf
+
+		remote = "client-session"
```
dnephin: constant?

cli/command/image/build.go
```diff
@@ -373,6 +478,86 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
+	default:
+		bw.Writer.Write(bw.buf.Bytes())
+		close(bw.flushed)
+	}
+	bw.mu.Unlock()
+}
+
+func getBuildSharedKey(dir string) (string, error) {
+	dt := []byte(cliconfig.Dir()) // if no access to config dir then only use path
+	if err := os.MkdirAll(cliconfig.Dir(), 0700); err == nil {
```
dnephin: Shouldn't "no access" be an error? This seems like it could result in some pretty surprising errors, and is also likely to be rare, so not really a big inconvenience. 

Maybe there should be a flag to set a filepath or directory for this?

cli/command/image/build.go
```diff
@@ -373,6 +478,86 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
+				if _, err := rand.Read(b); err != nil {
+					return "", err
+				}
+				if err := ioutil.WriteFile(sessionFile, []byte(hex.EncodeToString(b)), 0600); err != nil {
+					return "", err
+				}
+			}
+		}
+
+		dt, err = ioutil.ReadFile(sessionFile)
```
dnephin: I think this function would be easier to read as a struct, something like 

```
type SharedKey struct {
    ...
}

func (...) Read() ([]byte, error) {...}
func (...) create() (error) {...}
```

Then `getBuildSharedKey()` could just call `Read()` and would only be concerned with hashing the value.

cli/command/image/build.go
tonistiigi: `squash` option above seems to set it

cli/command/cli.go
dnephin: Oh, it also does this LessThan check, so this if  can be removed as well.

cli/command/image/build.go
tonistiigi: > This will be true when Dockerfile is from stdin

Only if there is no archive context. If there is no context then there is nowhere to replace Dockerfile either. But line 268 is wrong as it doesn't check `options.stream`. I'll add more comments as well.

cli/command/image/build.go
tonistiigi: For example when `options.stream` was true

cli/command/image/build.go
tonistiigi: I'll add a comment. In this case they are equal as nothing has modified `dockerfileCtx` yet but this will change in other comparisons. `dockerfileCtx` doesn't really mean that it has to be from stdin, it means that custom `Dockerfile` source has been set.

cli/command/image/build.go
tonistiigi: There is a defer that makes sure we have always written out the progress before returning from this function. This cancel is for unblocking that. I'll add a comment.

cli/command/image/build.go
tonistiigi: Currently it would still use the `configdir+dir` combination as `sharedKey` what should work well for 99% of the cases. And even if there is a collision it just means that performance will not be ideal. Maybe print a warning?

cli/command/image/build.go
tonistiigi: There will never be multiple instances of sharedkey and create is only called if read fails. Maybe helps to split this to `getSessionSharedKey(dir)` and `getNodeIdentifier()` as some hashing is always needed. Getting node identifier can be skipped if it has already run once.

tonistiigi:
@dnephin Updated. Added more comments for the conditions and moved the helper functions to a separate file where possible.

vieux:
LGTM

cli/command/formatter/disk_usage_test.go
```diff
@@ -23,6 +23,7 @@ func TestDiskUsageContextFormatWrite(t *testing.T) {
 Images              0                   0                   0B                  0B
 Containers          0                   0                   0B                  0B
 Local Volumes       0                   0                   0B                  0B
+Build Cache                                                 0B                  0B
```
thaJeztah: Wonder if we should have something (`-`) in the first two columns, but we can tweak that later

cli/command/image/build.go
```diff
@@ -373,6 +414,36 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
+			if _, err := rand.Read(b); err != nil {
+				return "", err
+			}
+			if err := os.MkdirAll(cliconfig.Dir(), 0600); err != nil {
+				return "", err
+			}
+			if err := ioutil.WriteFile(sessionFile, []byte(hex.EncodeToString(b)), 0600); err != nil {
+				return "", err
+			}
+		} else {
```
tiborvass: small nit: not needed. Don't push a fix only for this small nit.

cli/command/image/build.go
```diff
@@ -373,6 +414,36 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
 	return nil
 }
 
+func getBuildSharedKey(dir string) (string, error) {
+	// build session is hash of build dir with node based randomness
+	sessionFile := filepath.Join(cliconfig.Dir(), ".buildsharedkey")
+	if _, err := os.Lstat(sessionFile); err != nil {
```
tiborvass: Curious: why `Lstat` and not simply `Stat` ?

cli/command/image/build.go
```diff
@@ -282,22 +298,41 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
+	}
+
+	var body io.Reader
+	if buildCtx != nil && !options.stream {
+		body = progress.NewProgressReader(buildCtx, progressOutput, 0, "", "Sending build context to Docker daemon")
 	}
 
-	var body io.Reader = progress.NewProgressReader(buildCtx, progressOutput, 0, "", "Sending build context to Docker daemon")
+	// add context stream to the session
+	if options.stream && s != nil {
```
thaJeztah: This function (`runBuild()`) is getting awfully long now; we should have a look at refactoring it a bit, and splitting parts to separate functions where possible

cli/command/image/build.go
tonistiigi: No particular reason. We are not creating symlinks to this file so I don't think we should handle them as well.

cli/command/image/build_session.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/docker/docker/client/session"
+	"github.com/docker/docker/client/session/filesync"
+	"github.com/docker/docker/pkg/progress"
+	"github.com/pkg/errors"
+	"golang.org/x/time/rate"
+)
+
+const clientSessionRemote = "client-session"
+
+func isSessionSupported(dockerCli *command.DockerCli) bool {
```
thaJeztah: commit 0c780c83cd2d886d24b17e1f797622a80aaff7d1 adds `isSessionSupported()` and `getBuildSharedKey()`, which are moved in 9776fb556e2cb08441e242cf484ffbe6e616dc2e, perhaps those commits should be squashed, or the change moved from one to the other.

cli/command/image/build_session.go
```diff
@@ -0,0 +1,151 @@
...
+func getBuildSharedKey(dir string) (string, error) {
+	// build session is hash of build dir with node based randomness
+	s := sha256.Sum256([]byte(fmt.Sprintf("%s:%s", tryNodeIdentifier(), dir)))
+	return hex.EncodeToString(s[:]), nil
+}
+
+func tryNodeIdentifier() (out string) {
+	out = cliconfig.Dir() // return config dir as default on permission error
+	if err := os.MkdirAll(cliconfig.Dir(), 0700); err == nil {
+		sessionFile := filepath.Join(cliconfig.Dir(), ".buildsharedkey")
```
thaJeztah: Should this be in a tmp-dir? What happens if multiple clients perform a build in parallel?

cli/command/image/build.go
```diff
@@ -373,6 +414,36 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
 	return nil
 }
 
+func getBuildSharedKey(dir string) (string, error) {
+	// build session is hash of build dir with node based randomness
+	sessionFile := filepath.Join(cliconfig.Dir(), ".buildsharedkey")
```
tonistiigi: No, it needs to persist. This is only written once and correctness of this value only affects performance.

cli/command/image/build_session.go
tonistiigi: I can squash but that merges session+filesync. The move was requested from review. I think splitting up would take too much time.

cli/command/image/build.go
tonistiigi: I already moved some of it to the session file (and actually think it became less readable). This is synchronized with the build runner so needs to be in the same function. We could, of course, split things up like opts setting of context preparing but this is already large enough.

cli/command/image/build.go
```diff
@@ -268,7 +298,41 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
+	}
+
+	var body io.Reader
+	if buildCtx != nil && !options.stream {
+		body = progress.NewProgressReader(buildCtx, progressOutput, 0, "", "Sending build context to Docker daemon")
+	}
+
+	// add context stream to the session
+	if options.stream && s != nil {
+		syncDone := make(chan error)
```
tiborvass: Please add comment on why this `chan error` is not used to retrieve errors but as a syncing mechanism.

tonistiigi:
@thaJeztah Updated the filename and squashed.

@tiborvass Added comment.

cli/command/image/build.go
thaJeztah: Definitely doesn't have to be part of this PR; I opened an issue for it; https://github.com/docker/cli/issues/242

tiborvass:
LGTM

cli/command/image/build.go
```diff
@@ -242,24 +251,45 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
-		// Dockerfile which uses trusted pulls.
-		buildCtx = replaceDockerfileTarWrapper(ctx, buildCtx, relDockerfile, translator, &resolvedTags)
+		// if there is a tar wrapper, the dockerfile needs to be replaced inside it
+		if buildCtx != nil {
+			// Wrap the tar archive to replace the Dockerfile entry with the rewritten
+			// Dockerfile which uses trusted pulls.
+			buildCtx = replaceDockerfileTarWrapper(ctx, buildCtx, relDockerfile, translator, &resolvedTags)
+		} else if dockerfileCtx != nil {
+			// if there was not archive context still do the possible replacements in Dockerfile
+			newDockerfile, _, err := rewriteDockerfileFrom(ctx, dockerfileCtx, translator)
```
dnephin: @tonistiigi I notice that `resolvedTags` is not being assigned here. Isn't this necessary for content trust?

cli/command/image/build.go
```diff
@@ -268,7 +298,43 @@ func runBuild(dockerCli *command.DockerCli, options buildOptions) error {
...
+		}
+
+		buf := newBufferedWriter(syncDone, buildBuff)
+		defer func() {
+			select {
+			case <-buf.flushed:
+			case <-ctx.Done():
+			}
+		}()
+		buildBuff = buf
```
dnephin: If `options.quiet` is true then `buildBuff` is a `bytes.Buffer`. Later on it is expected that this remains a `bytes.Buffer()` so that its contents can be printed. 

I think this breaks with `--quiet`

tonistiigi: agh, sprintf of `io.Writer` https://github.com/docker/cli/blob/30933b516b69fb7431ada7e1ce658d023f696c01/cli/command/image/build.go#L436 🤦‍♂️

dnephin: I have this fixed in #294, it only sets buf if `!quiet `

PR #237, opened 2017-06-24
progress: Show progress of replicated tasks before they are assigned

aaronlehmann:
This was only showing tasks that belong to nodes that are currently up,
so that tasks on down nodes don't appear to be stuck. But this
unintentionally excludes tasks that haven't been assigned yet, so if a
task is stuck before assignment, for example because no nodes meet its
constraints, a progress bar won't even be shown. The check should only
apply to tasks that have a node assignment.

Related to https://github.com/moby/moby/issues/33807

cc @thaJeztah

### DIFF
--- a/cli/command/service/progress/progress.go
+++ b/cli/command/service/progress/progress.go
@@ -275,7 +275,11 @@ func (u *replicatedProgressUpdater) update(service swarm.Service, tasks []swarm.
 				continue
 			}
 		}
-		if _, nodeActive := activeNodes[task.NodeID]; nodeActive {
+		if task.NodeID != "" {
+			if _, nodeActive := activeNodes[task.NodeID]; nodeActive {
+				tasksBySlot[task.Slot] = task
+			}
+		} else {
 			tasksBySlot[task.Slot] = task
 		}
 	}


thaJeztah:
Without this patch;

```bash
$ docker service create --name donotdeploy --constraint "node.labels.somelabel==foobar" --detach=false nginx:alpine
0wx8h7rp50tqg0xnykwbxy4n6
1/1:
```

With this patch applied; 

```bash
./docker service create --name donotdeploy --constraint "node.labels.somelabel==foobar" --detach=false nginx:alpine
ve5pnzf0swqtapeb0uqnla2cz
1/1: pending   [================>                                  ]
```

thaJeztah:
After applying the label;

```bash
$ docker service create --name donotdeploy --constraint "node.labels.somelabel==foobar" --detach=false nginx:alpine
ve5pnzf0swqtapeb0uqnla2cz
verify: Waiting 1 seconds to verify that tasks are stable...
```

guess I found another issue (`verify: Waiting 1 seconds to verify that tasks are stable...`  isn't remove after it finishes), let me open an issue for that :smile:

thaJeztah:
All green 👍

PR #243, opened 2017-06-27
add `--detach` to docker scale 

vieux:
depends on https://github.com/docker/cli/pull/219

fixes  https://github.com/moby/moby/issues/32368

### DIFF
--- a/cli/command/service/scale.go
+++ b/cli/command/service/scale.go
@@ -56,6 +56,7 @@ func scaleArgs(cmd *cobra.Command, args []string) error {
 
 func runScale(dockerCli *command.DockerCli, flags *pflag.FlagSet, options *scaleOptions, args []string) error {
 	var errs []string
+	var serviceIDs []string
 	ctx := context.Background()
 
 	for _, arg := range args {
@@ -71,19 +72,22 @@ func runScale(dockerCli *command.DockerCli, flags *pflag.FlagSet, options *scale
 
 		if err := runServiceScale(ctx, dockerCli, serviceID, scale); err != nil {
 			errs = append(errs, fmt.Sprintf("%s: %v", serviceID, err))
+		} else {
+			serviceIDs = append(serviceIDs, serviceID)
 		}
 
-		if options.detach {
-			continue
-		}
-
-		if err := waitOnService(ctx, dockerCli, serviceID, false); err != nil {
-			errs = append(errs, fmt.Sprintf("%s: %v", serviceID, err))
-		}
 	}
 
-	if options.detach {
-		warnDetachDefault(dockerCli.Err(), dockerCli.Client().ClientVersion(), flags, "scaled")
+	if len(serviceIDs) > 0 {
+		if options.detach {
+			warnDetachDefault(dockerCli.Err(), dockerCli.Client().ClientVersion(), flags, "scaled")
+		} else {
+			for _, serviceID := range serviceIDs {
+				if err := waitOnService(ctx, dockerCli, serviceID, false); err != nil {
+					errs = append(errs, fmt.Sprintf("%s: %v", serviceID, err))
+				}
+			}
+		}
 	}
 
 	if len(errs) == 0 {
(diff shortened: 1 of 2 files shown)


vieux:
@thaJeztah not sure how to add the completion scripts here, I don't see --detach for the other service commands

cli/command/service/scale.go
```diff
@@ -56,7 +69,16 @@ func runScale(dockerCli *command.DockerCli, args []string) error {
 			continue
 		}
 
-		if err := runServiceScale(dockerCli, serviceID, scale); err != nil {
+		if err := runServiceScale(ctx, dockerCli, serviceID, scale); err != nil {
+			errs = append(errs, fmt.Sprintf("%s: %v", serviceID, err))
+		}
+
+		if options.detach {
+			warnDetachDefault(dockerCli.Err(), dockerCli.Client().ClientVersion(), flags, "scaled")
```
thaJeztah: Should this be inside the `for` loop?

thaJeztah:
CI is failing;

```
cli/command/service/helpers_test.go:7: testing redeclared as imported package name
	previous declaration at cli/command/service/helpers_test.go:5
FAIL	github.com/docker/cli/cli/command/service [build failed]
```

cli/command/service/scale.go
vieux: oh good point

vieux:
@thaJeztah PTAL

aaronlehmann:
Thanks. It would also be great to add this flag to `docker stack deploy`. There has been a lot of interest in that: https://github.com/moby/moby/issues/32367. I see @vdemeester assigned this to himself.

cli/command/service/scale.go
```diff
@@ -56,20 +69,31 @@ func runScale(dockerCli *command.DockerCli, args []string) error {
...
-		if err := runServiceScale(dockerCli, serviceID, scale); err != nil {
+		if err := runServiceScale(ctx, dockerCli, serviceID, scale); err != nil {
+			errs = append(errs, fmt.Sprintf("%s: %v", serviceID, err))
+		}
+
+		if options.detach {
+			continue
+		}
+
+		if err := waitOnService(ctx, dockerCli, serviceID, false); err != nil {
```
aaronlehmann: This will change the behavior to wait for the first service to scale before scaling the next one. Previously, they would all be scaled at the same time. I wonder if it's better to continue scaling them simultaneously, and show the progress for all services at the same time. On the other hand, this is a lot more work, and I'm not sure enough people try to scale multiple services at once for this to matter. Any thoughts?

dnephin: Thanks, I missed this. I think we should scale them all at once.

vieux: yeah I agree it would be best to not change the existing behaviour

thaJeztah: @vieux you'll be working on this?

vieux: yes

vieux:
@dnephin @vdemeester how about this ? (I kept it simple)

vieux:
I understand it's not a perfect solution, but I think it's OK for now.

cli/command/service/scale.go
```diff
@@ -43,11 +54,15 @@ func scaleArgs(cmd *cobra.Command, args []string) error {
...
-func runScale(dockerCli *command.DockerCli, args []string) error {
+func runScale(dockerCli *command.DockerCli, flags *pflag.FlagSet, options *scaleOptions, args []string) error {
 	var errs []string
+	var serviceIDs []string
+	ctx := context.Background()
+
 	for _, arg := range args {
 		parts := strings.SplitN(arg, "=", 2)
 		serviceID, scaleStr := parts[0], parts[1]
+		serviceIDs = append(serviceIDs, serviceID)
```
dnephin: Do we still want to wait on a service if it failed to update?

Maybe we should only add the serviceID to this list if there is no error from `runServiceScale()` ?

vieux: I agree

vieux: I also removed the warning if there is only errors.

vieux:
@dnephin PTAL

PR #233, opened 2017-06-23
Build stdin with compress

dnephin:
Fixes #212
Branched from #227  (only the last commit is new)

Compress after the archive is rewritten.

PR #245, opened 2017-06-27
builder: add config var for setting context streaming default

tonistiigi:
@tiborvass @thaJeztah 

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -29,6 +29,7 @@ import (
 	units "github.com/docker/go-units"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"golang.org/x/net/context"
 )
 
@@ -77,6 +78,15 @@ func (o buildOptions) contextFromStdin() bool {
 	return o.context == "-"
 }
 
+func setDefaultsFromConfig(cli command.Cli, flags *pflag.FlagSet, options *buildOptions) {
+	if !flags.Changed("stream") && isSessionSupported(cli) {
+		// if stream is not set try the default from config file
+		if v := cli.ConfigFile().BuildStreamContext; v != nil {
+			options.stream = *v
+		}
+	}
+}
+
 func newBuildOptions() buildOptions {
 	ulimits := make(map[string]*units.Ulimit)
 	return buildOptions{
@@ -98,6 +108,9 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 		Args:  cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.context = args[0]
+
+			setDefaultsFromConfig(dockerCli, cmd.Flags(), &options)
+
 			return runBuild(dockerCli, options)
 		},
 	}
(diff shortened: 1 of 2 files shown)


cli/command/image/build.go
```diff
@@ -94,6 +94,15 @@ func NewBuildCommand(dockerCli *command.DockerCli) *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.context = args[0]
+
+			flags := cmd.Flags()
+			if !flags.Changed("stream") && isSessionSupported(dockerCli) {
+				// if stream is not set try the default from config file
+				if v := dockerCli.ConfigFile().BuildStreamContext; v != nil {
+					options.stream = *v
+				}
+			}
```
dnephin: Could we move this to a new function ?

```go
func setDefaultsFromConfig(dockerCli command.Cli, flags *pflag.FlatSet, options buildOptions) bool {
    ...
}
```

That will make it easy to unit test.

tonistiigi: You mean `setDefaultsFromConfig(dockerCli command.Cli, flags *pflag.FlatSet, options *buildOptions)` ? Need a pointer of the options to change it. What is the bool return supposed to mean? Also, this wouldn't work because this check needs `DockerCli` not `Cli` to check if session is supported.

cli/config/configfile/file.go
```diff
@@ -43,6 +43,7 @@ type ConfigFile struct {
 	NodesFormat          string                      `json:"nodesFormat,omitempty"`
 	PruneFilters         []string                    `json:"pruneFilters,omitempty"`
 	Proxies              map[string]ProxyConfig      `json:"proxies,omitempty"`
+	BuildStreamContext   *bool                       `json:"buildStreamContext,omitempty"`
```
tiborvass: @thaJeztah buildStreamContext or streamBuildContext?

cli/command/image/build.go
dnephin: Yes, I forgot to remove the `bool` return (fixed now), and it would need to be a pointer.

It shouldn't need `DockerCli`. I fixed it in #233

cli/config/configfile/file.go
dnephin: I guess most of these start with the object name (service, images, etc). but `buildStreamContext` doesn't read too well.

How about `buildAlwaysStreamContext` ? That way it's a little more clear what this is enabling.

thaJeztah: Argh. naming. hm

Do we have other contexts other than for build? For booleans, I always liked, e.g. `enableSomething` (`EnableContextStreaming` e.g.), but that may be a bit too verbose

Do we expect more build-related options? If so, possibly have a `build` key, containing `stream-context` (in the JSON)

dnephin: I'd be good with a `build` key.

I don't think `Enable` is right, it's not enabling a feature, it's just changing the default.

vieux:
what are the possible values ?

vdemeester:
@thaJeztah @tonistiigi @dnephin what is the status here (should we close for now) ?

tiborvass:
Nope

PR #246, opened 2017-06-27
Add bash completion for `label` filter of `prune` commands

albers:
This adds bash completion for `docker container|image|network|system|volume prune --filter label|label!`
Ref: https://github.com/moby/moby/pull/30740


albers:
ping @sdurrheimer for zsh completion

vieux:
LGTM

PR #118, opened 2017-05-22
Don't prune volumes on `docker system prune`

cpuguy83:
Volumes tend to carry important data and pruning them on `docker system
prune` can easily cause unwanted data loss.

Let's play it safe and not prune volumes on `system prune`.

Throwing this out there.

PR #255, opened 2017-06-27
system prune: only warn for volumes if --volumes is given

albers:
#118 changed `docker system prune` to only remove volumes if `--volumes` is given.
The warning message, however, still always warns about volume removal:

```bash
$ docker system prune
WARNING! This will remove:
        - all stopped containers
        - all volumes not used by at least one container
        - all dangling images
        - all build cache
```
The warning message in the above example is wrong: no volumes will be deleted.

With this PR, the volume removal warning is included selectively depending on whether `--volumes` is given or not.

### DIFF
--- a/cli/command/system/prune.go
+++ b/cli/command/system/prune.go
@@ -1,7 +1,9 @@
 package system
 
 import (
+	"bytes"
 	"fmt"
+	"text/template"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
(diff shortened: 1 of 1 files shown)


mlaventure:
I know it's only for 2 lines, but I think a text template would be cleaner.

Something like: https://play.golang.org/p/tOJCeyb_RV

albers:
@mlaventure Thanks for your advice and the code sample. I pulled out message generation into a function that uses a template. PTAL.

PR #259, opened 2017-06-27
progress: Show task error in place of progress bar

aaronlehmann:
If a task encounters an error, the interactive "service create" and "service update" commands should show that error instead of showing a stuck progress bar.

To validate:

```
docker service create --detach=false --name broken --restart-condition=none --replicas 3 busybox asdf
```
and
```
docker service create --detach=false --name broken --mode global --restart-condition=none busybox asdf
```

Related to https://github.com/moby/moby/issues/33807
Also related: https://github.com/docker/swarmkit/pull/2287

cc @thaJeztah

### DIFF
--- a/cli/command/service/progress/progress.go
+++ b/cli/command/service/progress/progress.go
@@ -48,7 +48,7 @@ const (
 )
 
 type progressUpdater interface {
-	update(service swarm.Service, tasks []swarm.Task, activeNodes map[string]swarm.Node, rollback bool) (bool, error)
+	update(service swarm.Service, tasks []swarm.Task, activeNodes map[string]struct{}, rollback bool) (bool, error)
 }
 
 func init() {
@@ -199,16 +199,16 @@ func ServiceProgress(ctx context.Context, client client.APIClient, serviceID str
 	}
 }
 
-func getActiveNodes(ctx context.Context, client client.APIClient) (map[string]swarm.Node, error) {
+func getActiveNodes(ctx context.Context, client client.APIClient) (map[string]struct{}, error) {
 	nodes, err := client.NodeList(ctx, types.NodeListOptions{})
 	if err != nil {
 		return nil, err
 	}
 
-	activeNodes := make(map[string]swarm.Node)
+	activeNodes := make(map[string]struct{})
 	for _, n := range nodes {
 		if n.Status.State != swarm.NodeStateDown {
-			activeNodes[n.ID] = n
+			activeNodes[n.ID] = struct{}{}
 		}
 	}
 	return activeNodes, nil
@@ -266,7 +266,7 @@ type replicatedProgressUpdater struct {
 }
 
 // nolint: gocyclo
-func (u *replicatedProgressUpdater) update(service swarm.Service, tasks []swarm.Task, activeNodes map[string]swarm.Node, rollback bool) (bool, error) {
+func (u *replicatedProgressUpdater) update(service swarm.Service, tasks []swarm.Task, activeNodes map[string]struct{}, rollback bool) (bool, error) {
 	if service.Spec.Mode.Replicated == nil || service.Spec.Mode.Replicated.Replicas == nil {
 		return false, errors.New("no replica count")
 	}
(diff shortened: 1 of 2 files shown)


cli/command/service/progress/progress.go
```diff
@@ -303,7 +327,23 @@ func (u *replicatedProgressUpdater) update(service swarm.Service, tasks []swarm.
 			u.slotMap[task.Slot] = mappedSlot
 		}
 
-		if !u.done && replicas <= maxProgressBars && uint64(mappedSlot) <= replicas {
+		if !terminalState(task.DesiredState) && task.Status.State == swarm.TaskStateRunning {
+			running++
+		}
+
+		if u.done || replicas > maxProgressBars || uint64(mappedSlot) > replicas {
```
dnephin: It seems like the evaluation of `replicas > maxProgressBars` will never change inside this for loop (one is a constant, the other is set one above). Is that right?

I guess the purpose of this check is to not display progress bars if there are too many replicas? Couldn't this be handled at the top of the function?

cli/command/service/progress/progress.go
```diff
@@ -303,7 +327,23 @@ func (u *replicatedProgressUpdater) update(service swarm.Service, tasks []swarm.
 			u.slotMap[task.Slot] = mappedSlot
 		}
 
-		if !u.done && replicas <= maxProgressBars && uint64(mappedSlot) <= replicas {
+		if !terminalState(task.DesiredState) && task.Status.State == swarm.TaskStateRunning {
+			running++
+		}
```
dnephin: Could the body of this for loop be a separate method which returns the running count?

I think that would improve the readability of this function, and would make it easier to test this new code.

cli/command/service/progress/progress.go
```diff
@@ -388,7 +426,23 @@ func (u *globalProgressUpdater) update(service swarm.Service, tasks []swarm.Task
 
 	for _, task := range tasksByNode {
 		if node, nodeActive := activeNodes[task.NodeID]; nodeActive {
-			if !u.done && nodeCount <= maxProgressBars {
+			if !terminalState(task.DesiredState) && task.Status.State == swarm.TaskStateRunning {
+				running++
+			}
+
+			if u.done || nodeCount > maxProgressBars {
```
dnephin: same question about `nodeCount > maxProgressBars` not changing value within this for loop.

And also splitting this out into a new function (which should remove the need for nolint, and help testing).

cli/command/service/progress/progress.go
aaronlehmann: > Could the body of this for loop be a separate method which returns the running count?

Wouldn't it have the side effect of outputting progress information?

cli/command/service/progress/progress.go
aaronlehmann: > I guess the purpose of this check is to not display progress bars if there are too many replicas? Couldn't this be handled at the top of the function?

We also have to count the number of running tasks, which this loop does as well.

aaronlehmann:
> Full integration tests of the progress bars is going to be difficult, but we should be able to unit test a bunch of the internal logic.

I am going to work on some tests for the internal logic. Any preferences on adding them to this PR or a separate one?

cli/command/service/progress/progress.go
dnephin: Yes it would, but the current function has that side effect plus a bunch of other logic, so it's not any worse off, right?  It actually moves almost all of the progress printing to that new function. With a couple other smaller changes we could probably remove all of the progress printing from `update()`

dnephin:
My preference would be this PR

aaronlehmann:
Added unit tests and did some light refactoring.

cli/command/service/progress/progress.go
```diff
@@ -328,30 +324,73 @@ func (u *replicatedProgressUpdater) update(service swarm.Service, tasks []swarm.
...
+			if numberedStates[existingTask.DesiredState] == numberedStates[task.DesiredState] &&
+				numberedStates[existingTask.Status.State] <= numberedStates[task.Status.State] {
 				continue
 			}
 		}
-		tasksByNode[task.NodeID] = task
+		if task.NodeID != "" {
+			if _, nodeActive := activeNodes[task.NodeID]; nodeActive {
+				tasksBySlot[task.Slot] = task
+			}
```
dnephin: If `task.NodeID` is not empty but the node is not active then we skip the task.

Minor nit, but this case seems like it would be more obvious like this:

```go

if _, nodeActive := activeNodes[task.NodeID]; !nodeActive {
    continue
}

tasksBySlot[task.Slot] = task
```

That way it matches the checks above and there is only one line which assigns to `tasksBySlot`

aaronlehmann: Updated

PR #262, opened 2017-06-28
on_failure -> on-failure in compose code

cdrage:
on_failure should be on-failure

Signed-off-by: Charlie Drage <charlie@charliedrage.com>


cdrage:
ping @dnephin @denverdino

PR #264, opened 2017-06-29
Update docker kill reference docs

thaJeztah:
- explain that a custom stop-signal can be set in an image
- explain the either "name" or "id" can be used to reference a container

continues the work started in https://github.com/moby/moby/pull/32803  by @scjane

### DIFF
--- a/docs/reference/commandline/kill.md
+++ b/docs/reference/commandline/kill.md
@@ -27,9 +27,45 @@ Options:
 
 ## Description
 
-The main process inside the container will be sent `SIGKILL`, or any
-signal specified with option `--signal`.
+The `docker kill` subcommand kills one or more containers. The main process
+inside the container is sent `SIGKILL` signal (default), or the signal that is
+specified with the `--signal` option. You can kill a container using the
+container's ID, ID-prefix, or name.
 
 > **Note**: `ENTRYPOINT` and `CMD` in the *shell* form run as a subcommand of
 > `/bin/sh -c`, which does not pass signals. This means that the executable is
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/kill.md
```diff
@@ -27,8 +27,11 @@ Options:
 
 ## Description
 
-The main process inside the container will be sent `SIGKILL`, or any
-signal specified with option `--signal`.
+The `docker kill` subcommand kills one or more containers. The main process
+inside the container will be sent `SIGKILL` (default), the stop-signal that is
+defined in the image using the [`STOPSIGNAL`](../../builder/#stopsignal)
+Dockerfile instruction, or any signal specified with option `--signal` or
+`STOPSIGNAL` . You can kill a container using the container's ID or name.
```
dnephin: I don't think this is correct, the `STOPSIGNAL` should only be used for `stop`, not `kill`. I just tested quickly on 17.05 and that's the case. Did this change in 17.06?

docs/reference/commandline/kill.md
```diff
@@ -27,8 +27,11 @@ Options:
 
 ## Description
 
-The main process inside the container will be sent `SIGKILL`, or any
-signal specified with option `--signal`.
+The `docker kill` subcommand kills one or more containers. The main process
```
mstanleyjones: Maybe not for this PR but do we talk anywhere about the nuances between "stop", "terminate", "kill" etc, as concepts?

thaJeztah:
@dnephin @vdemeester updated, PTAL

PR #266, opened 2017-06-29
Add shellcheck for bash completion

jphuynh:

This is the equivalent of the PR from the moby project before cli was relocated here (See: https://github.com/moby/moby/pull/33260/).

A new docker image has been added because of the lack of a precompiled version of shellcheck on Alpine. Installing the dependencies and compiling from source would take more build time.

Eventually it would be nice to merge the validate and shellcheck jobs to use only `Dockerfile.validate` instead of `Dockerfile.dev`


dockerfiles/Dockerfile.validate
```diff
@@ -0,0 +1,9 @@
+FROM debian:stretch-slim
```
dnephin: Minor: could we name this file `Dockerfile.shellcheck` or `Dockerfile.shellvalidate` so that it's clear that it's different from "go validation"

docker.Makefile
```diff
@@ -7,6 +7,7 @@
 DEV_DOCKER_IMAGE_NAME = docker-cli-dev
 LINTER_IMAGE_NAME = docker-cli-lint
 CROSS_IMAGE_NAME = docker-cli-cross
+VALIDATE_IMAGE_NAME = docker-cli-validate
```
dnephin: `s/docker-cli-validate/docker-cli-shell-validate`

scripts/validate/shellcheck
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -e
+
+# Maintain an array of files to shellcheck not the best solution but will do for the time being
+FILES=()
+FILES+=("contrib/completion/bash/docker")
+FILES+=("scripts/validate/shellcheck")
```
dnephin: Shouldn't we also validate everything under `scripts/` ?

jphuynh: We should but the intent of this initial PR is to introduce shellcheck validation. I don't want it to end up like my original PR on the Moby project where it started to be very difficult to review and merge.

The idea is once we have that integrated to the build, we start adding the rest of `scripts/`

dockerfiles/Dockerfile.validate
jphuynh: Yep works for me.

scripts/validate/shellcheck
dnephin: SGTM

dockerfiles/Dockerfile.shellcheck
```diff
@@ -0,0 +1,9 @@
+FROM debian:stretch-slim
```
dnephin: Looks like there are some warnings related to https://github.com/koalaman/shellcheck/issues/785, but probably not an issue.

Could we use `debian:jessie` here?

dockerfiles/Dockerfile.shellcheck
```diff
@@ -0,0 +1,9 @@
...
+
+RUN apt-get update && \
+    apt-get -y install make shellcheck && \
+    apt-get clean
+
+WORKDIR /tmp
```
dnephin: Minor: could we use `/go/src/github.com/docker/cli` here too? It just makes it easier to re-use `MOUNTS` in the `docker.Makefile` if the code is mounted at the same place in all conatiners.

dockerfiles/Dockerfile.shellcheck
jphuynh: I noticed that as well. The problem with `debian:jessie` is the version of Shellcheck. (See: https://github.com/moby/moby/pull/33260#issuecomment-302330346).

I would have to recompile from source which involves `cabal-install`. I'm not sure it's worth the effort but I understand those warnings annoy me as well :)

dnephin:
cc @albers

dnephin:
looks like this needs a rebase

dnephin:
I think we can merge this, but I'd like to give @albers a chance to take a look at it first, since he's the primary contributor to the bash completion.

albers:
@jphuynh How can I invoke shellcheck? I tried `make -f docker.Makefile shellcheck` but got error messages:
```bash
$ make -f docker.Makefile shellcheck
[...]
docker run -ti --rm -v "/home/user/go/src/github.com/docker/cli":/go/src/github.com/docker/cli docker-cli-shell-validate make shellcheck
scripts/validate/shellcheck
shellcheck: unable to decommit memory: Invalid argument
shellcheck: unable to decommit memory: Invalid argument
shellcheck: unable to decommit memory: Invalid argument
```
exit code was 0, though.

contrib/completion/bash/docker
```diff
@@ -142,13 +146,15 @@ __docker_images() {
 	esac
 
 	local repo_print_command
+	# shellcheck disable=SC2016
```
albers: I think [SC2016](https://github.com/koalaman/shellcheck/wiki/SC2016) is so basic that we should not warn about it. 
We have several computed awk expressions that all produce these warnings. Please add SC2016 to the general exclusions.

contrib/completion/bash/docker
```diff
@@ -1,4 +1,8 @@
 #!/usr/bin/env bash
+# shellcheck disable=SC2155
```
albers: Please add a comment where to find documentation on shellcheck warnings: `https://github.com/koalaman/shellcheck/wiki/SCXXXX`

dnephin:
Now that #206 is merged this can be updated to use the shared engine (see the new circle.yml config)

contrib/completion/bash/docker
```diff
@@ -1838,7 +1851,7 @@ _docker_container_start() {
 			COMPREPLY=( $( compgen -W "$options" -- "$cur" ) )
 			;;
 		*)
-			__docker_complete_containers_stopped
+			__docker_complete_containers_stopped "$@"
```
jphuynh: TBH I didn't even check. It's coming from my previous PR and are leftovers of an attempt to automate certain rules on my side. It didn't work really great.

PR #268, opened 2017-06-29
Add bash completion for `swarm ca`

albers:
Ref: [reference page](https://github.com/docker/cli/blob/master/docs/reference/commandline/swarm_ca.md)

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -3270,6 +3270,7 @@ _docker_service_update_and_create() {
 
 _docker_swarm() {
 	local subcommands="
+		ca
 		init
 		join
 		join-token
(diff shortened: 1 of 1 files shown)


dnephin:
@albers Thanks for all your work on bash completion!  Now that github has better support for looking at individual commits within a PR, what do you think about having a single PR with multiple commits? I think that might make it easier to batch review them all at once.

cc @thaJeztah

albers:
@dnephin I think it's better to keep them in individual PRs. This way, they can be more easily scheduled for different releases. If we batch them, we would get into trouble if the schedule for one of the associated feature changes, see #118 for an example.

Your proposal also does not fit well into my workflow. I usually have to do my work in very short time slots after work or in breaks. I'd like to make available the results as soon as possible. Having to sort these by proposed release dates of the associated features would force me to queue already finished work for an unknown amount of time.
(The current bunch of PRs is not typical. It was the result of taking one day off after being forced to pause my work on Docker for several weeks (repo reorg, completion not available in dev container until https://github.com/moby/moby/pull/33801, vacation, a conference, high workload, family).)

Third, I found it hard to find reviewers for bigger changes in bash completion in the past. Such PRs could be hanging around several weeks. It's simply that there are not so many mainainers familiar with bash scripting. If we combine several features from e.g. swarm, builder and daemon into one PR, more reviewers would be needed.

albers:
@dnephin After all, this is about coupling, which should be avoided.

dnephin:
Ok, if this is easier for you that's fine with me. I wouldn't want your change to sit around, but if you happen to work on something that could conflict you could push it to the same PR. As long as they are in separate commits it should be just as easy to cherry-pick for multiple releases.

For reviewing, I'm hoping that by adding more validation (and possibly even test coverage) reviews will be faster. Right now I think we have to manually test the change, which can take a while.

#266 should be a good start for validation.

PR #269, opened 2017-06-29
Fix bash completion for `swarm init|join --data-path-addr`

albers:
bash completion for `swarm init --data-path-addr` and `swarm join --data-path-addr` treated `--data-path-addr` as a boolean option.
This fix adds completion for interfaces and local addresses, similar to `--advertise-addr` (but without trailing port).
Also corrects sort order of options.


thaJeztah:
ping @vdemeester PTAL

PR #270, opened 2017-06-29
login: print a big warning when using --password

tych0:
Task command lines are world readable via /proc/pid/cmdline, so this isn't
safe.

### DIFF
--- a/cli/command/registry/login.go
+++ b/cli/command/registry/login.go
@@ -47,6 +47,10 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
 	ctx := context.Background()
 	clnt := dockerCli.Client()
 
+	if opts.password != "" {
+		fmt.Fprintln(dockerCli.Err(), "WARNING! Using --password via the CLI is insecure.")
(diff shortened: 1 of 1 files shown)


cli/command/registry/login.go
```diff
@@ -47,6 +47,10 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
 	ctx := context.Background()
 	clnt := dockerCli.Client()
 
+	if opts.password != "" {
+		fmt.Fprintf(dockerCli.Err(), "Using --password via the CLI is insecure.\n")
```
thaJeztah: Can you change this to `fmt.Fprintln()`?

Perhaps prefix with `WARNING! `;

```go
fmt.Fprintln(dockerCli.Err(), "WARNING! Using --password via the CLI is insecure.")
```

tych0: Done.

n4ss:
LGTM!

vieux:
LGTM ping @thaJeztah

PR #271, opened 2017-06-29
Password stdin

tych0:
Add a --password-stdin argument to the login command, and update the warning message from #270 to tell people about it.

### DIFF
--- a/cli/command/registry/login.go
+++ b/cli/command/registry/login.go
@@ -2,6 +2,8 @@ package registry
 
 import (
 	"fmt"
+	"io/ioutil"
+	"strings"
 
 	"golang.org/x/net/context"
 
@@ -16,6 +18,7 @@ type loginOptions struct {
 	serverAddress string
 	user          string
 	password      string
+	passwordStdin bool
 }
 
 // NewLoginCommand creates a new `docker login` command
@@ -39,6 +42,7 @@ func NewLoginCommand(dockerCli command.Cli) *cobra.Command {
 
 	flags.StringVarP(&opts.user, "username", "u", "", "Username")
 	flags.StringVarP(&opts.password, "password", "p", "", "Password")
+	flags.BoolVarP(&opts.passwordStdin, "password-stdin", "", false, "Take the password from stdin")
 
 	return cmd
 }
@@ -48,7 +52,24 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
 	clnt := dockerCli.Client()
 
 	if opts.password != "" {
-		fmt.Fprintln(dockerCli.Err(), "WARNING! Using --password via the CLI is insecure.")
+		fmt.Fprintln(dockerCli.Err(), "WARNING! Using --password via the CLI is insecure. Use --password-stdin.")
+		if opts.passwordStdin {
+			return errors.New("--password and --password-stdin are mutually exclusive")
+		}
+	}
+
+	if opts.passwordStdin {
+		if opts.user == "" {
+			return errors.New("Must provide --username with --password-stdin")
+		}
+
+		contents, err := ioutil.ReadAll(dockerCli.In())
+		if err != nil {
+			return err
+		}
+
+		opts.password = strings.TrimSuffix(string(contents), "\n")
+		opts.password = strings.TrimSuffix(opts.password, "\r")
 	}
 
 	var (


cli/command/registry/login.go
```diff
@@ -47,6 +50,28 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
 	ctx := context.Background()
 	clnt := dockerCli.Client()
 
+	if opts.password != "" {
+		fmt.Fprintf(dockerCli.Err(), "Using --password via the CLI is insecure. Use --password-stdin\n")
+		if opts.passwordStdin {
+			return errors.Errorf("--password and --password-stdin are mutually exclusive")
+		}
```
dnephin: I wonder if it would make sense to use `-p -` to mean `stdin`. Usually `-` is only for files, but maybe it would be ok to use that convention here as well.

cli/command/registry/login.go
```diff
@@ -47,6 +50,28 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
...
+	}
+
+	if opts.passwordStdin {
+		if opts.user == "" {
+			return errors.Errorf("Must provide --username with --password-stdin")
+		}
```
dnephin: This doesn't seem to be a requirement for `--password`, why is it necessary for `password-stdin` ?

cli/command/registry/login.go
tych0: It would break everyone whose password was "-", so it seems better not to.

cli/command/registry/login.go
tych0: If you don't supply a username via --username, it asks via stdin, which then may eat parts of your password, depending on what characters it has in it (e.g. `\n`). Seems better to just disallow it to prevent confusion to me.

cli/command/registry/login.go
dnephin: Hopefully `-` is not an acceptable password on any registries, and they all require at least a few more characters..

cli/command/registry/login.go
```diff
@@ -47,6 +50,28 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
...
+
+		contents, err := ioutil.ReadAll(dockerCli.In())
+		if err != nil {
+			return err
+		}
+
+		if contents[len(contents)-1] == '\n' {
```
n4ss: We should still check that `contents != ""` before doing that.

n4ss: Also on Windows we have CRLF so we should test "\r\n" when running on windows.

Do we have helper functions for this @thaJeztah @vdemeester ?

cli/command/registry/login.go
```diff
@@ -47,6 +50,28 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
...
+			opts.password = string(contents[:len(contents)-1])
```
n4ss: same here for the CRLF, make sure we don't include the '\r'

cli/command/registry/login.go
vdemeester: cc @johnstep

tych0: Good point. in fact, we should probably just use the stdlib to do all of this. Fixed, thanks!

cli/command/registry/login.go
```diff
@@ -47,6 +52,29 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
...
+			return errors.Errorf("Must provide --username with --password-stdin")
+		}
+
+		contents, err := ioutil.ReadAll(dockerCli.In())
+		if err != nil {
+			return err
+		}
+
+		opts.password = strings.TrimSuffix(string(contents), "\n")
+		if runtime.GOOS == "windows" {
```
aaronlehmann: Any reason to limit this to Windows? The password could end up with a CR if it's piped from a file with DOS line endings, or pasted from a weird source. I don't think it will matter much in practice, but applying the same trimming on both platforms would be a slight simplification.

tych0: Done, thanks.

tych0:
Ping. Any movement on this?

n4ss:
LGTM!

cli/command/registry/login.go
```diff
@@ -47,6 +51,27 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
 	ctx := context.Background()
 	clnt := dockerCli.Client()
 
+	if opts.password != "" {
+		fmt.Fprintln(dockerCli.Err(), "WARNING! Using --password via the CLI is insecure. Use --password-stdin")
```
aaronlehmann: Should there be a period at the end of this? If `Using --password via the CLI is insecure` is formatted as a sentence, I think `Use --password-stdin` should be as well.

cli/command/registry/login.go
```diff
@@ -47,6 +51,27 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
...
+		if opts.passwordStdin {
+			return errors.Errorf("--password and --password-stdin are mutually exclusive")
```
aaronlehmann: `errors.New`

And I'm wondering if there's a way to phrase this that's more friendly to people who don't speak english as a first language, but I don't have any good ideas right now.

cli/command/registry/login.go
```diff
@@ -47,6 +51,27 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
...
+		}
+	}
+
+	if opts.passwordStdin {
+		if opts.user == "" {
+			return errors.Errorf("Must provide --username with --password-stdin")
```
aaronlehmann: `errors.New`.

cli/command/registry/login.go
tych0: Sure, I've left it as is now and fixed everything else. If someone has a better idea, let me know and I'll change it.

aaronlehmann:
LGTM

tych0:
Sure, I can work up docs patches, feel free to merge.

PR #182, opened 2017-06-12
Do not call the config endpoint if API is lower than 1.30

vdemeester:
Another would be to handle `Page not found` error in those cases 👼
(and https://github.com/moby/moby/pull/33571 could help too 👼)

Fix #180 

/cc @mavenugo @friism 

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

PR #277, opened 2017-07-01
Don't attempt to remove unsupported resources on older daemon

thaJeztah:
When running `docker stack rm <some stack>` against an older daemon, a warning was printed for "configs" being ignored;

    WARNING: ignoring "configs" (requires API version 1.30, but the Docker daemon API version is 1.26)

Given that an old daemon cannot _have_ configs, there should not be a need to warn, or _attempt_ to remove these resources.

This patch removes the warning, and skips removing configs.

A check if _secrets_ are supported by the daemon is also added, given that this would result in an error when attempted against an older (pre 1.13) daemon.

There is one situation where this could lead to secrets or configs being left behind; if the client is connecting to a daemon that _does_ support secrets, configs, but the API version is overridden using `DOCKER_API_VERSION`, no warning is printed, and secrets and configs are not attempted to be removed.

Given that `DOCKER_API_VERSION` is regarded a feature for debugging / "power users", it should be ok to ignore this.

ping @dnephin @vdemeester @briantd

### DIFF
--- a/cli/command/stack/remove.go
+++ b/cli/command/stack/remove.go
@@ -51,20 +51,16 @@ func runRemove(dockerCli command.Cli, opts removeOptions) error {
 			return err
 		}
 
-		secrets, err := getStackSecrets(ctx, client, namespace)
-		if err != nil {
-			return err
+		var secrets []swarm.Secret
+		if versions.GreaterThanOrEqualTo(client.ClientVersion(), "1.25") {
+			secrets, err = getStackSecrets(ctx, client, namespace)
+			if err != nil {
+				return err
+			}
 		}
 
 		var configs []swarm.Config
-
-		version, err := client.ServerVersion(ctx)
-		if err != nil {
-			return err
-		}
-		if versions.LessThan(version.APIVersion, "1.30") {
-			fmt.Fprintf(dockerCli.Err(), "WARNING: ignoring \"configs\" (requires API version 1.30, but the Docker daemon API version is %s)\n", version.APIVersion)
-		} else {
+		if versions.GreaterThanOrEqualTo(client.ClientVersion(), "1.30") {
 			configs, err = getStackConfigs(ctx, client, namespace)
 			if err != nil {
 				return err
--- a/cli/command/stack/remove_test.go
+++ b/cli/command/stack/remove_test.go
@@ -40,13 +40,49 @@ func TestRemoveStack(t *testing.T) {
 	}
 	allConfigIDs := buildObjectIDs(allConfigs)
 
+	// Using API 1.24; removes services, networks, but doesn't remove configs and secrets
 	cli := &fakeClient{
+		version:  "1.24",
 		services: allServices,
 		networks: allNetworks,
 		secrets:  allSecrets,
 		configs:  allConfigs,
 	}
 	cmd := newRemoveCommand(test.NewFakeCli(cli, &bytes.Buffer{}))
+	cmd.SetArgs([]string{"foo", "bar"})
+
+	assert.NoError(t, cmd.Execute())
+	assert.Equal(t, allServiceIDs, cli.removedServices)
+	assert.Equal(t, allNetworkIDs, cli.removedNetworks)
+	assert.Nil(t, cli.removedSecrets)
+	assert.Nil(t, cli.removedConfigs)
+
+	// Using API 1.25; removes services, networks, but doesn't remove configs
+	cli = &fakeClient{
+		version:  "1.25",
+		services: allServices,
+		networks: allNetworks,
+		secrets:  allSecrets,
+		configs:  allConfigs,
+	}
+	cmd = newRemoveCommand(test.NewFakeCli(cli, &bytes.Buffer{}))
+	cmd.SetArgs([]string{"foo", "bar"})
+
+	assert.NoError(t, cmd.Execute())
+	assert.Equal(t, allServiceIDs, cli.removedServices)
+	assert.Equal(t, allNetworkIDs, cli.removedNetworks)
+	assert.Equal(t, allSecretIDs, cli.removedSecrets)
+	assert.Nil(t, cli.removedConfigs)
+
+	// Using API 1.30; removes services, networks, configs, and secrets
+	cli = &fakeClient{
+		version:  "1.30",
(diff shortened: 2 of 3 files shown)


thaJeztah:
Actually, I realised this is what we had originally in https://github.com/docker/cli/pull/182, but I failed to see that we only called these endpoints on _removal_, not in other cases 😓

cli/command/stack/remove.go
```diff
@@ -39,7 +39,17 @@ func runRemove(dockerCli command.Cli, opts removeOptions) error {
 	client := dockerCli.Client()
 	ctx := context.Background()
 
-	var errs []string
+	var (
+		errs    []string
+		secrets []swarm.Secret
+		configs []swarm.Config
+	)
```
dnephin: This is a minor nit, but I think it's generally better to keep variable declarations closer to the code where they are used. It makes it easier to see the scope of the variable and to refactor the code later.

So these could move to immediately before the `if versions...` check for each

cli/command/stack/remove.go
```diff
@@ -39,7 +39,17 @@ func runRemove(dockerCli command.Cli, opts removeOptions) error {
...
+
+	version, err := client.ServerVersion(ctx)
```
dnephin: I missed this in the earlier PR. I think we should already have version stored, we shouldn't need to make another API call.

It's available from `cli.Client().ClientVersion()`. The name is misleading, but this version was already set from the server version.

cli/command/stack/remove.go
```diff
@@ -77,8 +81,12 @@ func runRemove(dockerCli command.Cli, opts removeOptions) error {
 		}
 
 		hasError := removeServices(ctx, dockerCli, services)
-		hasError = removeSecrets(ctx, dockerCli, secrets) || hasError
-		hasError = removeConfigs(ctx, dockerCli, configs) || hasError
+		if versions.GreaterThanOrEqualTo(version.APIVersion, "1.25") {
+			hasError = removeSecrets(ctx, dockerCli, secrets) || hasError
+		}
+		if versions.GreaterThanOrEqualTo(version.APIVersion, "1.30") {
```
dnephin: These two conditions aren't necessary because `removeSecrets()` and `removeConfigs()` iterate over the array, so it's already a no-op if they are empty.

thaJeztah:
@dnephin @vdemeester updated PTAL

dnephin:
test failures

cli/command/stack/remove.go
thaJeztah: Updated; was thinking that declaring the variables inside the loop was bad, but reading up, Go is smart enough to handle that 👍

cli/command/stack/remove.go
thaJeztah: Ok, so I tried using this, but ran into issues. the `ClientVersion()` is [returning `cli.version`](https://github.com/docker/cli/blob/4adf701567e4e64226e2e3622375b78a612b2198/vendor/github.com/docker/docker/client/client.go#L236-L242)

`cli.version` is only set by [`NegotiateAPIVersionPing()`](https://github.com/docker/cli/blob/4adf701567e4e64226e2e3622375b78a612b2198/vendor/github.com/docker/docker/client/client.go#L263-L266), and [`NewClient()`](https://github.com/docker/cli/blob/4adf701567e4e64226e2e3622375b78a612b2198/vendor/github.com/docker/docker/client/client.go#L191-L200), both of which are not used for our fakeClient.

I decided to use the "old" approach for now, but we should look into this, as it's only an issue for testing

Perhaps you have a trick up your sleeve to make it work though :smile:

thaJeztah:
Tests are passing again, see my comments above 👍

cli/command/stack/remove.go
```diff
@@ -38,7 +38,10 @@ func runRemove(dockerCli command.Cli, opts removeOptions) error {
 	namespaces := opts.namespaces
 	client := dockerCli.Client()
 	ctx := context.Background()
-
+	version, err := client.ServerVersion(ctx)
```
dnephin: You can set `ClientVersion()` to the version you want like this:

```go
&fakeClient{
    Client: client.NewClient("", apiVersion, nil, nil),
    ...
}
```

thaJeztah: I think I tried adding `version` and `ClientVersion()` to `fakeClient{}`, but it didn't satisfy the Interface. Hm, let me check, perhaps I didn't

thaJeztah: Oh, I know what I did 🤦‍♂️ - yes this works, thanks! Adding some additional test-cases for different versions now

cli/command/stack/remove_test.go
```diff
@@ -72,6 +108,7 @@ func TestRemoveStackSkipEmpty(t *testing.T) {
 	allConfigIDs := buildObjectIDs(allConfigs)
 
 	fakeClient := &fakeClient{
+		version:  "1.30",
```
thaJeztah: We can probably remove the 1.30 versions once https://github.com/moby/moby/pull/33887 is merged and vendored

cli/command/stack/remove_test.go
```diff
@@ -52,6 +54,40 @@ func TestRemoveStack(t *testing.T) {
 	assert.NoError(t, cmd.Execute())
 	assert.Equal(t, allServiceIDs, cli.removedServices)
 	assert.Equal(t, allNetworkIDs, cli.removedNetworks)
+	assert.Nil(t, cli.removedSecrets)
+	assert.Nil(t, cli.removedConfigs)
+
+	// Using API 1.25; removes services, networks, but doesn't remove configs
```
dnephin: I generally prefer to make these separate test cases instead of inline comments. It's never really clear where an inline comment ends. With a separate test function the scope is obvious.

Any shared setup can be moved into a helper function to remove the duplication, which also helps make the test more readable.

This is a minor style thing, so not blocking.

thaJeztah:
yes; at first was doubting to make it a test-table, or separate tests, but kept it like this; I can update if you want, or do a follow up. If we want this into 17.06.1, perhaps a follow up

vdemeester:
@thaJeztah counting on you for the follow-up 👼 🤗

PR #279, opened 2017-07-03
Add bash completion for `build --iidfile`

albers:
This adds bash completion for https://github.com/moby/moby/pull/32406.


vieux:
LGTM

PR #282, opened 2017-07-03
Updates to bash completion for `dockerd`

albers:
A bunch of options were missing in bash completion of `dockerd`.
See individual commit messages for details and references.
ping @sdurrheimer for zsh completion

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -1985,6 +1985,7 @@ _docker_daemon() {
 		--data-root
 		--default-gateway
 		--default-gateway-v6
+		--default-runtime
 		--default-shm-size
 		--default-ulimit
 		--dns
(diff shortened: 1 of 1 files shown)


contrib/completion/bash/docker
```diff
@@ -2124,10 +2147,20 @@ _docker_daemon() {
 			__docker_complete_log_options
 			return
 			;;
+		--metrics-addr)
+			__docker_complete_local_interfaces --ip-only
```
thaJeztah: Wonder if it would be "cleaner" to have a separate function for ip-addresses

albers: @thaJeztah updated, thanks for the hint.

vieux:
LGTM

vieux:
LGTM

Issue #88, opened 2017-05-15
Add --pretty option to "secret inspect" and "config inspect"


aaronlehmann:
Other "inspect" subcommands have a `--pretty` option. `secret inspect` and `config inspect` appear to be missing it. This is important because otherwise there's no human-readable way to get information about a config or secret object. It's particularly necessary for `config inspect` so the config payload can be exposed in cleartext, instead of the base64-encoded form that gets spit out over the JSON API.

cc @dhiltgen

PR #45, opened 2017-05-08
Add support for configs

aaronlehmann:
This adds CLI support for the configs feature (https://github.com/moby/moby/pull/32336).

It should not be merged until the moby PR is merged.

PR #283, opened 2017-07-03
Add bash completion for `docker config` command family

albers:
This adds bash completion for #45.

Additional commits add related functionality/refactorings:
- bash completion for `secret inspect --pretty` (Ref: #88)
- a minor refactoring in bash completion for `secret rm|remove`

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -4151,7 +4151,7 @@ _docker_secret_inspect() {
 
 	case "$cur" in
 		-*)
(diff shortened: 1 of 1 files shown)


mlaventure:
@albers do you think it'd be possible to have path completion enabled for `config create` ?

i.e.:
```
$ docker config my-<tab>
```

correctly doesn't offer suggestion

but
```
$ docker config my-config /etc/<tab>
```

Doesn't either, and I would have expected it to do so.

albers:
@mlaventure Nice catch, thanks a lot. I added path completion to `docker config create`.

PR #236, opened 2017-06-23
Import DisplayablePorts from moby/moby source

vdemeester:
This has nothing to do in the `api` package for moby as it's just a
cli display function.

Related PR in moby/moby: https://github.com/moby/moby/issues/33798

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

PR #286, opened 2017-07-03
Bump moby to 87df0e533b619c088091fd1e2310e92bb9a24822

thaJeztah:
Includes changes from;

- Add a LastTagTime for images (https://github.com/moby/moby/pull/31497)
- Fix handling of remote "git@" notation (https://github.com/moby/moby/pull/33696)
- Move some `api` package functions away (https://github.com/moby/moby/pull/33798) (related to https://github.com/docker/cli/pull/236)
- Set ping version even on error (https://github.com/moby/moby/pull/33827)
- Do not add duplicate platform information to service spec (https://github.com/moby/moby/pull/33867)
- Refactor MountPoint Setup function in volume.go (https://github.com/moby/moby/pull/33890)

### DIFF
--- a/cli/command/image/testdata/inspect-command-success.simple-many.golden
+++ b/cli/command/image/testdata/inspect-command-success.simple-many.golden
@@ -21,6 +21,9 @@
         },
         "RootFS": {
             "Type": ""
+        },
+        "Metadata": {
+            "LastTagTime": "0001-01-01T00:00:00Z"
         }
     },
     {
@@ -45,6 +48,9 @@
         },
         "RootFS": {
             "Type": ""
+        },
+        "Metadata": {
+            "LastTagTime": "0001-01-01T00:00:00Z"
         }
     }
 ]
--- a/cli/command/image/testdata/inspect-command-success.simple.golden
+++ b/cli/command/image/testdata/inspect-command-success.simple.golden
@@ -21,6 +21,9 @@
         },
         "RootFS": {
             "Type": ""
+        },
+        "Metadata": {
+            "LastTagTime": "0001-01-01T00:00:00Z"
         }
     }
 ]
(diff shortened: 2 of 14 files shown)


thaJeztah:
Looks like a test needs updating, or we need to change upstream to omit the `Metadata.LastTagTime` if not set;

```
--- FAIL: TestNewInspectCommandSuccess (0.00s)
Error Trace:	helpers.go:23
		inspect_test.go:89
Error:      	Not equal: 
expected: "[{\"Id\":\"\",\"RepoTags\":null,\"RepoDigests\":null,\"Parent\":\"\",\"Comment\":\"\",\"Created\":\"\",\"Container\":\"\",\"ContainerConfig\":null,\"DockerVersion\":\"\",\"Author\":\"\",\"Config\":null,\"Architecture\":\"\",\"Os\":\"\",\"Size\":0,\"VirtualSize\":0,\"GraphDriver\":{\"Data\":null,\"Name\":\"\"},\"RootFS\":{\"Type\":\"\"}}]"
received: "[{\"Id\":\"\",\"RepoTags\":null,\"RepoDigests\":null,\"Parent\":\"\",\"Comment\":\"\",\"Created\":\"\",\"Container\":\"\",\"ContainerConfig\":null,\"DockerVersion\":\"\",\"Author\":\"\",\"Config\":null,\"Architecture\":\"\",\"Os\":\"\",\"Size\":0,\"VirtualSize\":0,\"GraphDriver\":{\"Data\":null,\"Name\":\"\"},\"RootFS\":{\"Type\":\"\"},\"Metadata\":{\"LastTagTime\":\"0001-01-01T00:00:00Z\"}}]"
```

@dnephin wdyt?

thaJeztah:
Actually, I wonder if we should change the way we test; quoting from the API reference;

> The API uses an open schema model, which means server may add extra properties to responses. Likewise, the server will ignore any extra query parameters and request body properties. When you write clients, you need to ignore additional properties in responses to ensure they do not break when talking to newer Docker daemons.

dnephin:
I think we just need to update the test to match the new expected result

thaJeztah:
@dnephin @vdemeester updated the test PTAL

PR #291, opened 2017-07-04
Bump swarmkit and dependencies to 79381d0840be27f8b3f5c667b348a4467d866eeb

thaJeztah:
this PR;

- Bumps SwarmKit to 79381d0840be27f8b3f5c667b348a4467d866eeb 
- Bumps gRPC to v1.3.0
  - adds genproto
  - updates protobuf
- Bumps golang.org/x/net to 7dcfb8076726a3fdd9353b6b8a1f1b6be6811bd6

PTAL if this is all okay :smile:


aaronlehmann:
LGTM

In the long term, I'd like to move away from importing swarmkit in the cli repo. I believe the only use is for the `github.com/docker/swarmkit/api/defaults` package, which was a stopgap to avoid declaring default values in the CLI code. It would be better to have an API that exposes these default values. Then the CLI will see the defaults that apply to the daemon, not the ones that were linked into its binary.

thaJeztah:
Yes, indeed, it looks like that's the only reason, and it would definitely be great if we could get rid of the dependency

PR #294, opened 2017-07-04
WIP: refactor runBuild

### LINKED ISSUE #242: Refactor `build/runBuild()`
Per the discussion in https://github.com/docker/cli/pull/231#discussion_r124122908, the `runBuild()` function in cli/command/image/build.go is getting quite long, and may need a refactor

dnephin:
Fixes #242


cli/command/image/build.go
```diff
@@ -162,183 +159,62 @@ func (out *lastProgressOutput) WriteProgress(prog progress.Progress) error {
...
-		return err
+	var contextSession *session.Session
+	if options.stream {
+		contextSession, err = newBuildContextSession(dockerCli, buildInput.contextDir)
+		if err != nil {
+			return err
+		}
 	}
 
+	progressOutput := newProgaressOutput(dockerCli.Out(), buildBuffer.progress)
```
tonistiigi: `Progress`

cli/command/image/build.go
```diff
@@ -528,57 +363,122 @@ func rewriteDockerfileFrom(ctx context.Context, dockerfile io.Reader, translator
...
+}
 
-			if _, err := io.Copy(tarWriter, content); err != nil {
-				pipeWriter.CloseWithError(err)
-				return
+// PrintImageIDIfQuiet to stdout. If quiet is not set then the image ID is
+// printed elsewhere, so do nothing.
+func (bb *buildBuffer) PrintImageIDIfQuiet() {
+	if bb.quiet {
+		fmt.Fprintf(bb.stdout, "%s", bb.output)
```
tonistiigi: Still printing `io.Writer` as a string, please let's not do this.

cli/command/image/build.go
```diff
@@ -528,57 +363,122 @@ func rewriteDockerfileFrom(ctx context.Context, dockerfile io.Reader, translator
...
+// the progress is printed elsewhere, so do nothing.
+func (bb *buildBuffer) PrintProgressOnError() {
+	if bb.quiet {
+		fmt.Fprintln(bb.stderr, bb.progress)
+	}
+}
+
+// SetOutput to a new Writer. If quiet is set then do nothing.
+func (bb *buildBuffer) SetOutput(out io.Writer) {
+	if !bb.quiet {
```
tonistiigi: I didn't check but would guess this breaks showing progress output on errors.

cli/command/image/build.go
dnephin: Fair, We should cast it to `bytes.Buffer` first, and panic if it's the wrong type.

Since this is now encapsulated, it's easy to verify that the writer is the correct type when `quiet`.

cli/command/image/build.go
dnephin: I'll look into it further. It seems `--stream` is using these buffers differently.

tonistiigi:
> They are far from arbitrary.

I didn't mean that they are not correctly picked, just that this is complex code and drawing boundaries with a ruler to complex code doesn't necessarily improve it. Example: the first function `setupContextAndDockerfile` returns a `buildInput` but if a reader thinks that this is the actual input then they are mistaken. There are several places after this call that modify the properties of that structure(trust, compress etc). Stream properties are set after this function so you would think that it should be just called `setupContextAndDockerfileIfNoStream` but then you see that there is nothing that sets up dockerfile for a stream. Then you look at `setupContextAndDockerfile` implementation more carefully and discover it does set up dockerfile for a stream but not the context.

> The name seems fairly obvious to me (it's a file which stores a build ID). 

The obvious question is what makes this different than any other file so it needs to be wrapped like this. The answer seems to be that it is to preserve one specific error being outputted in the `Save()` method. From the caller, I have no idea that this error exists.

> Please comment on these where you see them so the naming can be improved.

I would if I had better ideas. `setupContextStream->addDirToSessionWithProgressAndBlockingBuildBufferUntilCompleted` doesn't seem like a good solution.


I don't want to argue about this much as this is far from the worst things we have merged. After spending time on it I'm starting to understand what is going on, but I'm not sure if I want other people to have the same experience of repeatedly switching through a lot of files looking for the code that I know needs to be somewhere to understand it.

cli/command/image/build.go
tonistiigi: I don't think cast is a correct solution. For the case where it appeared it would have just meant a different error. Type safety is the main thing protecting against incorrect code changes.

dnephin: Sorry, by "cast", I meant "type check". We can check if the field is a buffer, and call `buffer.String()` instead.

cli/command/image/build.go
dnephin: What is the correct behaviour when `--stream --quiet` are both set? 

Will stream print the errors correctly to the buffer or do we need a special case like we have with just `--quiet`?

dnephin:
> From the caller, I have no idea that this error exists.

Exactly, that's kind of the point! The caller shouldn't be aware of any of the implementation details. It should only be concerned with the high level operation it needs to perform (save the file, or remove the file).

> I'm not sure if I want other people to have the same experience of repeatedly switching through a lot of files looking for the code that I know needs to be somewhere to understand it.

I don't see how this PR changes this. Before this change there was plenty of code that was not in `runBuild()`. You still had to go to a bunch of files to understand everything, but understanding the logic of `runBuild()` was a lot more difficult because many different concerns were mixed up together, and the size of the function meant there were many state changes happening to a single variable.

By splitting it up further we get a better sense of the high level work done as part of "build". You shouldn't need to understand the implementation of every function to understand what build does. Every line in  `runBuild` that isn't a high level operation is cognitive load that someone has to take on just to understand the basics of what happens on build. This changes gets us a bit closer to function that reads cleanly. It's still too large, but it's getting closer at least.

cli/command/image/build.go
dnephin: This is fixed. We are no longer printing `io.Writer`, and I renamed `buildBuffer` to `buildOutputBuffer` and made it more generic.

cli/command/image/build.go
dnephin: For now the behaviour is that `--stream` overrides the buffer, which means that progress/output buffers will not be printed by the `--quiet` special case. 

I'm not sure if this is correct or not, because I'm not sure how `--stream` handles  quiet.

vdemeester:
ping @dnephin @tonistiigi what is the status here ?

dnephin:
The build command still needs to be refactored.

PR #298, opened 2017-07-05
Fix stack output stream

dnephin:
Fixes https://github.com/moby/moby/issues/33954

"Nothing found in stack" should go to stderr

Also cleaned up some duplication in task format selection


thaJeztah:
ping @vdemeester PTAL

PR #299, opened 2017-07-05
Less verbose fake cli

dnephin:
Almost every use of `FakeCli` uses a `bytes.Buffer` for stdout/stderr. Instead of having the caller setup these buffers every time we can default to using a buffer.

I updated some tests to remove unnecessary calls to `SetErr()` and `SetConfigFile()`, and a few tests to use the new `NewFakeCli()`. I've deprecated the only one.

### DIFF
--- a/cli/command/image/load_test.go
+++ b/cli/command/image/load_test.go
@@ -43,7 +43,7 @@ func TestNewLoadCommandErrors(t *testing.T) {
 		},
 	}
 	for _, tc := range testCases {
-		cli := test.NewFakeCli(&fakeClient{imageLoadFunc: tc.imageLoadFunc}, new(bytes.Buffer))
+		cli := test.NewFakeCliWithOutput(&fakeClient{imageLoadFunc: tc.imageLoadFunc}, new(bytes.Buffer))
 		cli.In().SetIsTerminal(tc.isTerminalIn)
 		cmd := NewLoadCommand(cli)
 		cmd.SetOutput(ioutil.Discard)
@@ -54,7 +54,7 @@ func TestNewLoadCommandErrors(t *testing.T) {
 
 func TestNewLoadCommandInvalidInput(t *testing.T) {
 	expectedError := "open *"
-	cmd := NewLoadCommand(test.NewFakeCli(&fakeClient{}, new(bytes.Buffer)))
+	cmd := NewLoadCommand(test.NewFakeCliWithOutput(&fakeClient{}, new(bytes.Buffer)))
 	cmd.SetOutput(ioutil.Discard)
 	cmd.SetArgs([]string{"--input", "*"})
 	err := cmd.Execute()
@@ -93,7 +93,7 @@ func TestNewLoadCommandSuccess(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		buf := new(bytes.Buffer)
-		cmd := NewLoadCommand(test.NewFakeCli(&fakeClient{imageLoadFunc: tc.imageLoadFunc}, buf))
+		cmd := NewLoadCommand(test.NewFakeCliWithOutput(&fakeClient{imageLoadFunc: tc.imageLoadFunc}, buf))
 		cmd.SetOutput(ioutil.Discard)
 		cmd.SetArgs(tc.args)
 		err := cmd.Execute()
--- a/cli/command/image/load_test.go
+++ b/cli/command/image/load_test.go
@@ -43,7 +43,7 @@ func TestNewLoadCommandErrors(t *testing.T) {
 		},
 	}
 	for _, tc := range testCases {
-		cli := test.NewFakeCliWithOutput(&fakeClient{imageLoadFunc: tc.imageLoadFunc}, new(bytes.Buffer))
+		cli := test.NewFakeCli(&fakeClient{imageLoadFunc: tc.imageLoadFunc})
 		cli.In().SetIsTerminal(tc.isTerminalIn)
 		cmd := NewLoadCommand(cli)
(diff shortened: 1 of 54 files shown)


aaronlehmann:
What do you think about functional args for `NewFakeCli`? `WithOutput`, `WithConfigFile`, etc. I'm not sure if it's worth it if passing in a buffer is immediately deprecated, but thought I'd offer the suggestion.

It feels like this PR goes halfway, in that it renames the function but doesn't update most tests to use the new style. Would it be difficult to update the tests in the same PR, to avoid touching all these tests multiple times?

dnephin:
Difficult no, but time consuming, there are lots of instances. A regex wont work because they are almost always on multiple lines. I'll see about converting a few more.

What's bad about touching the files twice? 


> What do you think about functional args for NewFakeCli? WithOutput, WithConfigFile, etc.

Works for me, but I don't think it's related to this change. They would save a few lines when assigning new values to these fields. This change is about removing unnecessary lines, not replacing necessary ones with helpers.

aaronlehmann:
> What's bad about touching the files twice?

Touching them once would ease review, overall.

dnephin:
Rebased to fix conflicts, and I converted a bunch more tests to use the new function.

I was thinking that the rest would just be handled if/when tests are updated. It's not really a high enough priority that we need to convert them all, but making this change now lets us write all the new tests with the less verbose option. It shouldn't really incur any extra review time because they can be changed as part of updating tests.

cli/command/image/load_test.go
```diff
@@ -43,7 +43,7 @@ func TestNewLoadCommandErrors(t *testing.T) {
 		},
 	}
 	for _, tc := range testCases {
-		cli := test.NewFakeCli(&fakeClient{imageLoadFunc: tc.imageLoadFunc}, new(bytes.Buffer))
+		cli := test.NewFakeCliWithOutput(&fakeClient{imageLoadFunc: tc.imageLoadFunc}, new(bytes.Buffer))
```
aaronlehmann: Could just be `NewFakeCli`, but I know you're not trying to exhaustively convert these.

cli/command/image/load_test.go
```diff
@@ -54,7 +54,7 @@ func TestNewLoadCommandErrors(t *testing.T) {
 
 func TestNewLoadCommandInvalidInput(t *testing.T) {
 	expectedError := "open *"
-	cmd := NewLoadCommand(test.NewFakeCli(&fakeClient{}, new(bytes.Buffer)))
+	cmd := NewLoadCommand(test.NewFakeCliWithOutput(&fakeClient{}, new(bytes.Buffer)))
```
aaronlehmann: Could just be `NewFakeCli`, but I know you're not trying to exhaustively convert these.

cli/command/image/remove_test.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 func TestNewRemoveCommandAlias(t *testing.T) {
-	cmd := newRemoveCommand(test.NewFakeCli(&fakeClient{}, new(bytes.Buffer)))
+	cmd := newRemoveCommand(test.NewFakeCliWithOutput(&fakeClient{}, new(bytes.Buffer)))
```
aaronlehmann: Could just be `NewFakeCli`, but I know you're not trying to exhaustively convert these.

cli/command/image/remove_test.go
```diff
@@ -44,7 +44,7 @@ func TestNewRemoveCommandErrors(t *testing.T) {
 		},
 	}
 	for _, tc := range testCases {
-		cmd := NewRemoveCommand(test.NewFakeCli(&fakeClient{
+		cmd := NewRemoveCommand(test.NewFakeCliWithOutput(&fakeClient{
```
aaronlehmann: Could just be `NewFakeCli`, but I know you're not trying to exhaustively convert these.

cli/command/image/save_test.go
```diff
@@ -45,7 +45,7 @@ func TestNewSaveCommandErrors(t *testing.T) {
 		},
 	}
 	for _, tc := range testCases {
-		cli := test.NewFakeCli(&fakeClient{imageSaveFunc: tc.imageSaveFunc}, new(bytes.Buffer))
+		cli := test.NewFakeCliWithOutput(&fakeClient{imageSaveFunc: tc.imageSaveFunc}, new(bytes.Buffer))
```
aaronlehmann: Could just be `NewFakeCli`, but I know you're not trying to exhaustively convert these.

cli/command/stack/list_test.go
```diff
@@ -48,7 +48,7 @@ func TestListErrors(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		cmd := newListCommand(test.NewFakeCli(&fakeClient{
+		cmd := newListCommand(test.NewFakeCliWithOutput(&fakeClient{
```
aaronlehmann: Could just be `NewFakeCli`, but I know you're not trying to exhaustively convert these.

cli/command/stack/remove_test.go
```diff
@@ -46,7 +46,7 @@ func TestRemoveStack(t *testing.T) {
 		secrets:  allSecrets,
 		configs:  allConfigs,
 	}
-	cmd := newRemoveCommand(test.NewFakeCli(cli, &bytes.Buffer{}))
+	cmd := newRemoveCommand(test.NewFakeCliWithOutput(cli, &bytes.Buffer{}))
```
aaronlehmann: Could just be `NewFakeCli`, but I know you're not trying to exhaustively convert these.

PR #300, opened 2017-07-05
build: follow up fixes for build stream

tonistiigi:
Addresses the follow-up comments in https://github.com/docker/cli/pull/231#pullrequestreview-47914776

For the content-trust update, I have no idea why the `FROM` images should be retagged by the builder. I would argue that this is wrong but made it match the non-stream behavior.

@dnephin 

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -166,15 +166,16 @@ func (out *lastProgressOutput) WriteProgress(prog progress.Progress) error {
 // nolint: gocyclo
 func runBuild(dockerCli command.Cli, options buildOptions) error {
 	var (
-		buildCtx      io.ReadCloser
-		dockerfileCtx io.ReadCloser
-		err           error
-		contextDir    string
-		tempDir       string
-		relDockerfile string
-		progBuff      io.Writer
-		buildBuff     io.Writer
-		remote        string
+		buildCtx          io.ReadCloser
+		dockerfileCtx     io.ReadCloser
+		err               error
+		contextDir        string
+		tempDir           string
+		relDockerfile     string
+		progBuff          io.Writer
+		buildBuff         io.Writer
+		remote            string
+		quietOutputBuffer *bytes.Buffer
 	)
 
 	if options.dockerfileFromStdin() {
@@ -189,7 +190,8 @@ func runBuild(dockerCli command.Cli, options buildOptions) error {
 	buildBuff = dockerCli.Out()
 	if options.quiet {
 		progBuff = bytes.NewBuffer(nil)
-		buildBuff = bytes.NewBuffer(nil)
+		quietOutputBuffer = bytes.NewBuffer(nil)
+		buildBuff = quietOutputBuffer
 	}
 	if options.imageIDFile != "" {
 		// Avoid leaving a stale file if we eventually fail
@@ -284,7 +286,8 @@ func runBuild(dockerCli command.Cli, options buildOptions) error {
 			buildCtx = replaceDockerfileTarWrapper(ctx, buildCtx, relDockerfile, translator, &resolvedTags)
 		} else if dockerfileCtx != nil {
 			// if there was not archive context still do the possible replacements in Dockerfile
-			newDockerfile, _, err := rewriteDockerfileFrom(ctx, dockerfileCtx, translator)
+			var newDockerfile []byte
+			newDockerfile, resolvedTags, err = rewriteDockerfileFrom(ctx, dockerfileCtx, translator)
 			if err != nil {
 				return err
 			}
@@ -415,7 +418,7 @@ func runBuild(dockerCli command.Cli, options buildOptions) error {
 				jerr.Code = 1
 			}
 			if options.quiet {
-				fmt.Fprintf(dockerCli.Err(), "%s%s", progBuff, buildBuff)
+				fmt.Fprintf(dockerCli.Err(), "%s%s", progBuff, quietOutputBuffer)
 			}
 			return cli.StatusError{Status: jerr.Message, StatusCode: jerr.Code}
 		}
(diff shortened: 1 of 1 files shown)


dnephin:
There's another one or two places where this buffer is printed I believe. I have this fixed in #294

thaJeztah:
ping @tonistiigi @dnephin what's the status on this one?

dnephin:
I prefer the solution in #294

tonistiigi:
I didn't notice that this got stuck. This is not a refactor but fixes `build -q --stream` not working at all. I recommend merging and picking to relevant docker-ce releases.

andrewhsu:
@dnephin would it be possible to bring this PR in as a more isolated change and then refactored for #294 ?

dnephin:
Sure, that's possible. This PR has no tests, so I'm not sure how confident we can be that it still fixes the problem.

thaJeztah:
ping @tonistiigi is it possible to add a test for this PR?

thaJeztah:
(or go for https://github.com/docker/cli/pull/294 if you think that's better)

tonistiigi:
@thaJeztah This behavior should be tested with e2e/integration test. So not atm(or any more) afaik. I repeat that this is a fix for a bug still existing in `17.07+`

dnephin:
>  This behavior should be tested with e2e/integration test.

I disagree. The bug was not in the integration of components, it was in a single component, which means it can be unit tested, if the code is structured properly.

There is an e2e suite in docker/cli https://github.com/docker/cli/tree/master/e2e, but writing a e2e test for every minor bug is not a sustainable way of testing software.

tonistiigi:
The bug is in no component. The bug is in the cli wrapper so the test needs to use the cli. Implying that these cli unit-tests with mocks actually catch any bugs is something I can not agree with and to my knowledge has become evident in almost any time cli and engine are combined in `docker-ce` (eg #654 #577). A sustainable way of testing software is to write tests that catch bugs, not ones that add coverage.

Feel free to close, as it shouldn't take 3 months to merge a fix for not printing `io.Writer` as a string.

dnephin:
> The bug is in the cli wrapper so the test needs to use the cli

I'm not sure what you mean by "cli wrapper". The bug is in the cli binary, which is a component of a larger system. This bug wasn't with the integration of components, it was internal to a single module, so doesn't need a binary with an API to exercise the bug.

> cli unit-tests with mocks actually catch any bugs is something I can not agree 

Unit tests will not catch integration errors, that's true. That's not their purpose. Their purpose is to validate that a unit behaves correctly. In this case the unit is a module in the cli repo.

Integration and e2e test are also important, but when every test is written in that style is causes many problems. Integration and e2e tests should test integration of components, and end user features, not internal logic of a module.

> has become evident in almost any time cli and engine are combined in docker-ce

Yes, there are plenty of missing tests in `docker/cli` and `moby/moby`. Sometimes those holes in the test suite are discovered through integration tests, which is a good opportunity to write the missing unit tests. Many of the tests failures during integration are actually due to bad assumptions in the integration suite. Others failures are due to missing unit test coverage.

Considering that test coverage is barely 50% I'm actually surprised that we average fewer than 1 integration bug per release.

tonistiigi:
I've added another commit to open Dockerfile from correct location.

> Integration and e2e tests should test integration of components, and end user features, not internal logic of a module.

End user feature is broken. You can't build with certain flags.

>  I'm actually surprised that we average fewer than 1 integration bug per release.

The examples were from regressions in the old code that happened to have tests before 17.03. It is very hard to assume anything about the number of bugs in new features as like this one they are only discovered by manual testing after release is out.

dnephin:
> End user feature is broken. You can't build with certain flags.

Yes, every bug would impact some user feature, that's almost the definition of a bug. Are you arguing that there is no place for unit tests, and that every test must exercise the system end-to-end to be of any value? I don't really understand how anyone could make that argument.

> It is very hard to assume anything about the number of bugs in new features as like this one they are only discovered by manual testing after release is out.

Which is exactly why I've been asked for test coverage, so we don't have to assume.

tonistiigi:
> If this was as simple as translating command line flags into an API call I don't see why it would take 1200 lines. There are plenty of units in there to test. They may not be obvious right now, because of the code structure, but they are there.

Absolutely, but I'm not trying to rewrite these 1200 lines. I'm trying to fix a bug. In first commit, the bug is printing `io.Writer` as a string. In second it is using a wrong variable in one line.

> There is no duplicate code.

I didn't mean that the mock is a duplicate, but that the tests don't test a behavior but follow the implementation. This does not apply in cases where there is a testable unit, but cases, where there isn't one, but unit-test is used to cover a specific implementation line that changed.

For the data, I meant data on that our test coverage/quality has gone down. That is based recollection that every `*.0` release since `17.06` has had some builder related bug. Just today I was looking at https://github.com/moby/moby/issues/33974#issuecomment-340010900 . I don't know what is the issue there yet as more info is needed, but I do know that we don't have proper coverage about these cases (our previous exchange on that, that I lost https://github.com/moby/moby/pull/34063#pullrequestreview-49337533). Some of these issues have fixes in cli, some don't. The bottom line issue is the same - that we don't have tests for user features.

Both of the issues you mentioned had plenty of coverage in `moby` test suite that used to test this code before `17.06`. #577 especially, as it fixes a case that made `docker run/attach` completely unusable but nothing caught it in the review process of the `cli` repo.

> You could contribute an e2e feature test in e2e/image/build_test.go that tests the success path of using --stream. 

I'll do that.

thaJeztah:
@tonistiigi have you had time to look at adding a test in e2e/image/build_test.go ?

tonistiigi:
@thaJeztah I still plan to do it but don't know when. I didn't notice that there were no cases that cover building in `e2e` package yet, eg the file you referenced doesn't exist.

dnephin:
Opened #709 which adds an e2e test for `docker build` using a directory as context.

vdemeester:
ping @dnephin @tonistiigi what is the status here ? 👼

dnephin:
Just needs a test

PR #303, opened 2017-07-06
fixing SIGSEGV when running containers

yastij:
Signed-off-by: Yassine TIJANI <yasstij11@gmail.com>

**- What I did** fixed moby/moby#33975

**- How I did it** checking that close is not before deferring it.

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -190,10 +190,11 @@ func runContainer(dockerCli *command.DockerCli, opts *runOptions, copts *contain
 		}
 
 		close, err := attachContainer(ctx, dockerCli, &errCh, config, createResponse.ID)
-		defer close()
+
 		if err != nil {
 			return err
 		}
+		defer close()
 	}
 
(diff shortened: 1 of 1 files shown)


cli/command/container/run.go
```diff
@@ -190,7 +190,11 @@ func runContainer(dockerCli *command.DockerCli, opts *runOptions, copts *contain
 		}
 
 		close, err := attachContainer(ctx, dockerCli, &errCh, config, createResponse.ID)
-		defer close()
+
+		if close != nil {
+			defer close()
+		}
```
dnephin: I think the more common way of doing this is to just move the `defer close()` after the `if err != nil`. This is only `nil` if there is an error, so handling the error first will fix the problem.

yastij: @dnephin -  okay, i thought that we might have some error and still have non nil close.

Thanks for the hint.

yastij:
@dnephin - LGTY ?

yastij:
@dnephin  Done.

aaronlehmann:
LGTM

PR #304, opened 2017-07-06
Add .exe extensions to .gitignore for local build

jhowardmsft:

Same as for moby/moby, adds the .exe extensions to .gitignore for when doing local builds


dnephin:
Nothing in our build system writes to this path, we already ignore `./build`.

Maybe this should go into the local `.git/info/exclude` ?

Issue #310, opened 2017-07-08
system prune fails if "until" filter is used


thaJeztah:
Originally reported by @twooster in https://github.com/moby/moby/issues/31122#issuecomment-313872632

When running `docker system prune` with an `until` filter, pruning fails and does not remove images, volumes, and networks.

```
Error response from daemon: Invalid filter 'until'
```

The cause of this is that all resources (containers, images, networks) support filtering by `until` _except_ volumes. The `docker system prune` sequentially calls the various prune commands, and when it calls the `docker volume prune` endpoint, an error is returned by the daemon, and pruning stops;

```
# docker system prune --filter until=24h
WARNING! This will remove:
        - all stopped containers
        - all volumes not used by at least one container
        - all networks not used by at least one container
        - all dangling images
Are you sure you want to continue? [y/N] y
Error response from daemon: Invalid filter 'until'
```

Daemon logs;

```
DEBU[1185] Calling POST /v1.30/containers/prune?filters=%7B%22label%22%3A%7B%22label.foo%3D%3Dbar%22%3Atrue%7D%2C%22until%22%3A%7B%2224h%22%3Atrue%7D%7D
DEBU[1185] Calling POST /v1.30/volumes/prune?filters=%7B%22label%22%3A%7B%22label.foo%3D%3Dbar%22%3Atrue%7D%2C%22until%22%3A%7B%2224h%22%3Atrue%7D%7D
ERRO[1185] Handler for POST /v1.30/volumes/prune returned error: Invalid filter 'until'
Error response from daemon: Invalid filter 'until'
```

```
# docker version
Client:
 Version:      17.06.0-ce
 API version:  1.30
 Go version:   go1.8.3
 Git commit:   02c1d87
 Built:        Fri Jun 23 21:23:31 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.06.0-ce
 API version:  1.30 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   02c1d87
 Built:        Fri Jun 23 21:19:04 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

We need to remove the `until` filter for the "volume prune" endpoint (but only when called as part of `system prune`, not when directly calling `docker volume prune`).

Given that we don't support a client talking to a _newer_ API version, it should not be a problem to simply remove the filter, as long as we don't forget to remove that part of the code once volumes _do_ support the `until` filter.

[...]

PR #311, opened 2017-07-08
Error if "until" filter is combined with "--volumes" on system prune

thaJeztah:
fixes https://github.com/docker/cli/issues/310
relates to https://github.com/moby/moby/issues/31122#issuecomment-313872632

The "until" filter is supported by all object types, except for
volumes.

Before this patch, the "until" filter would attempted to be used for the volume
prune endpoint, resulting in an error being returned by the daemon, and
further prune endpoints (networks, images) to be skipped.

```bash
$ docker system prune --filter until=24h --filter label=label.foo=bar

WARNING! This will remove:
        - all stopped containers
        - all volumes not used by at least one container
        - all dangling images
Are you sure you want to continue? [y/N] y
Error response from daemon: Invalid filter 'until'
```

    Calling POST /v1.30/containers/prune?filters=%7B%22label%22%3A%7B%22label.foo%3D%3Dbar%22%3Atrue%7D%2C%22until%22%3A%7B%2224h%22%3Atrue%7D%7D
    Calling POST /v1.30/volumes/prune?filters=%7B%22label%22%3A%7B%22label.foo%3D%3Dbar%22%3Atrue%7D%2C%22until%22%3A%7B%2224h%22%3Atrue%7D%7D
    Handler for POST /v1.30/volumes/prune returned error: Invalid filter 'until'
    Error response from daemon: Invalid filter 'until'

With this patch, an error is produced instead, preventing "partial" prune.

```bash
$ docker system prune --filter until=24h --filter label=foo==bar --volumes
ERROR: The "until" filter is not supported with "--volumes"
```

Note that `docker volume prune` does not have this problem, and produces an
error if the `until` filter is used;

```bash
$ docker volume prune --filter until=24h

WARNING! This will remove all volumes not used by at least one container.
Are you sure you want to continue? [y/N] y
Error response from daemon: Invalid filter 'until'
```

### DIFF
--- a/cli/command/system/prune.go
+++ b/cli/command/system/prune.go
@@ -67,6 +67,10 @@ func runBuildCachePrune(dockerCli command.Cli, _ opts.FilterOpt) (uint64, string
 }
 
 func runPrune(dockerCli command.Cli, options pruneOptions) error {
+	// TODO version this once "until" filter is supported for volumes
+	if options.pruneVolumes && options.filter.Value().Include("until") {
+		return fmt.Errorf(`ERROR: The "until" filter is not supported with "--volumes"`)
+	}
 	if versions.LessThan(dockerCli.Client().ClientVersion(), "1.31") {
 		options.pruneBuildCache = false
 	}


thaJeztah:
ping @cpuguy83 ptal

cli/command/prune/prune.go
```diff
@@ -37,7 +37,25 @@ func RunContainerPrune(dockerCli command.Cli, filter opts.FilterOpt) (uint64, st
...
+	// To "remove" the "until" filter if set, we just copy the only supported
+	// filter for now ("label").
+	//
+	// Keep this hack in sync with supported filters for volume prune to prevent
+	// new filters from silently being ignored!
+	//
+	// TODO make `opts.FilterOpt` more flexible to prevent awkward hacks like this
+	// TODO version this hack once "until" filter is supported for volumes
+	vf := opts.NewFilterOpt()
+	if filter.Value().Include("label") {
```
thaJeztah: If anyone has suggestions for doing this in a cleaner way, let me know

I didn't want to update the package upstream (moby/moby) to keep the diff small

cpuguy83:
We could keep the list of supported filters in the API spec.

dnephin:
This seems strange to me. What would be the use case for removing other things older than x, but removing all volumes? I can't think of when that would ever be useful.

Generally I don't think that `system prune` should support filters. The only use case I see for `system prune` is to remove everything. If you want to remove things with more fine grained control use `docker x prune` with the correct filters. Applying the same filters to a bunch of different resources types makes very little sense to me. Resources have very different lifecycles in most cases.

I think the correct fix here is to remove filters from `system prune` entirely.

thaJeztah:
It's been there since 17.04 https://github.com/moby/moby/pull/29226, so if we want to remove it, it should go through a deprecation cycle

I _do_ see value when using filtering by label (i.e. being able to prune everything, and exclude resources labeled with x.y.z), but can understand your thinking on 'until' for these.

For this PR; it's not adding a feature but fixing a bug

thaJeztah:
Updated; I decided to simplify this PR, and instead error out if `--filter until=...` is combined with `--volumes`.

ping @dnephin @vdemeester PTAL

vieux:
LGTM

thaJeztah:
I'll merge, but happy to do a follow-up if we want this tested

PR #318, opened 2017-07-10
pin vndr to 9909bb2b8a0b7ea464527b376dc50389c90df587

thaJeztah:
This make updating vndr a deliberate action, and prevents updates to vndr from making the vendor validation fail in CI.

Pinning to 9909bb2b8a0b7ea464527b376dc50389c90df587, which is current "master", and the same version as is used in https://github.com/moby/moby/pull/34047

ping @tonistiigi @tiborvass @dnephin PTAL

### DIFF
--- a/dockerfiles/Dockerfile.dev
+++ b/dockerfiles/Dockerfile.dev
@@ -3,8 +3,11 @@ FROM    golang:1.8.3-alpine
 
 RUN     apk add -U git make bash coreutils
 
-RUN     go get github.com/LK4D4/vndr && \
-        cp /go/bin/vndr /usr/bin && \
+ARG     VNDR_COMMIT=9909bb2b8a0b7ea464527b376dc50389c90df587
+RUN     git clone https://github.com/LK4D4/vndr.git "/go/src/github.com/LK4D4/vndr" && \
+        cd "/go/src/github.com/LK4D4/vndr" && \
+        git checkout -q "$VNDR_COMMIT" && \
+        go build -v -o /usr/bin/vndr . && \
         rm -rf /go/src/* /go/pkg/* /go/bin/*
 
 RUN     go get github.com/jteeuwen/go-bindata/go-bindata && \


dockerfiles/Dockerfile.dev
```diff
@@ -3,8 +3,11 @@ FROM    golang:1.8.3-alpine
 
 RUN     apk add -U git make bash coreutils
 
-RUN     go get github.com/LK4D4/vndr && \
-        cp /go/bin/vndr /usr/bin && \
+ARG     VNDR_COMMIT=9909bb2b8a0b7ea464527b376dc50389c90df587
```
mlaventure: should go in `hack/dockerfile/binaries-commits`

`hack/dockerfile/install-binaries.sh` will override this version otherwise

thaJeztah: @mlaventure this is the docker/cli repo, there's no `hack/dockerfile/binaries-commits`, so I just put it in the dockerfile

mlaventure: :man_facepalming: 

Indeed :sweat_smile:

vieux:
LGTM

dockerfiles/Dockerfile.dev
```diff
@@ -3,8 +3,11 @@ FROM    golang:1.8.3-alpine
...
+RUN     git clone https://github.com/LK4D4/vndr.git "/go/src/github.com/LK4D4/vndr" && \
+        cd "/go/src/github.com/LK4D4/vndr" && \
+        git checkout -q "$VNDR_COMMIT" && \
+        go build -v -o /usr/bin/vndr . && \
```
dnephin: Isn't this going to break as soon as `vndr` adds a dependency, since we don't have a `go get` anywhere here?

`go get` will also get all the git files, so I think we can use `go get` to fetch, and still `git checkout <sha>`

thaJeztah: > Isn't this going to break as soon as vndr adds a dependency, since we don't have a go get anywhere here?

I assume it would vendor that dependency, but it's a ood point; I took the same approach as is used in moby/moby, so we may want to modify the approach there as well

PR #319, opened 2017-07-11
Bugfix docker cp

keloyang:
Vendor, update moby to d58ffa0364c04d03a8f25704d7f0489ee6cd9634 , this will fix the bug that docker cp dir with hard link. Refer to https://github.com/moby/moby/pull/33956 for detail.


scripts/vendor.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+# This file is just wrapper around vndr (github.com/LK4D4/vndr) tool.
+# For updating dependencies you should change `vendor.conf` file in root of the
+# project. Please refer to https://github.com/LK4D4/vndr/blob/master/README.md for
+# vndr usage.
```
dnephin: I don't think this file is necessary. We already have `make vendor` and the tool is already available in the container.

cyphar:
I was under the impression that the updated version of docker/docker#33956 doesn't require any CLI changes, because we only changed what links we generated (which doesn't affect the client) -- that was the main purpose of making that change.

dnephin:
It looks like we call `archive.RebaseArchiveEntries()` from `cli/command/container/cp.go`, so it's possible we need this fix in the cli.

keloyang:
@dnephin  @cyphar, thanks for your review,  I am sure cli call archive.RebaseArchiveEntries()  from  cli/command/container/cp.go ,  and archive.RebaseArchiveEntries is in vendor/github.com/docker/docker/pkg/archive/copy.go
```
root@ubuntu:~/workspace/go/src/github.com/docker/cli# grep RebaseArchiveEntries ./ -rsn
Binary file ./build/docker-linux-amd64 matches
./vendor/github.com/docker/docker/pkg/archive/copy.go:272:              return dstDir, RebaseArchiveEntries(srcContent, srcBase, dstBase), nil
./vendor/github.com/docker/docker/pkg/archive/copy.go:283:              return dstDir, RebaseArchiveEntries(srcContent, srcBase, dstBase), nil
./vendor/github.com/docker/docker/pkg/archive/copy.go:300:              return dstDir, RebaseArchiveEntries(srcContent, srcBase, dstBase), nil
./vendor/github.com/docker/docker/pkg/archive/copy.go:305:// RebaseArchiveEntries rewrites the given srcContent archive replacing
./vendor/github.com/docker/docker/pkg/archive/copy.go:307:func RebaseArchiveEntries(srcContent io.Reader, oldBase, newBase string) io.ReadCloser {
./cli/command/container/cp.go:172:              preArchive = archive.RebaseArchiveEntries(content, srcBase, srcInfo.RebaseName)
```

scripts/vendor.sh
keloyang: Ok, I will remove the first commit,thank you.

cyphar:
>  so it's possible we need this fix in the cli.

cyphar:
> so it's possible we need this fix in the cli.

If that's the case then doesn't that make docker/docker#33956 a breaking API change? Or am I misunderstanding and removing `archive.RebaseArchiveEntries` is purely a cleanliness issue and doesn't change the actual functionality of `docker/cli`. I'll test this.

keloyang:
ping @dnephin

keloyang:
Whether or not grpc has a problem, I think it's better that all registers are called before starting grpc service.

dnephin:
For some reason the circleCI status isn't reflected in this PR and it hasn't run on the latest commit.

Do you have circleCI setup for your fork by chance? I'm not sure what causes this problem.

dnephin:
It looks like closing the PR and re-opened it triggered the build, cool

keloyang:
Thank you.

thaJeztah:
ping @dnephin PTAL

PR #321, opened 2017-07-11
Remove bash completion for `service update --network`

albers:
The `--network` option is only valid for `docker service create`.
This PR removes it from bash completion for `docker service update` (which has `--network-add` and `--network-rm` instead).


vieux:
LGTM

PR #322, opened 2017-07-11
Pin all tools used in the Dockerfiles

dnephin:
Also update `gometalinter` to use the official version instead of my fork, now that the official version has accepted my PRs. The latest version includes some updates to linters, which found some new gosimple errors. This PR includes the fixes.

### DIFF
--- a/cli/command/formatter/disk_usage.go
+++ b/cli/command/formatter/disk_usage.go
@@ -243,7 +243,7 @@ func (c *diskUsageImagesContext) Reclaimable() string {
 	if c.totalSize > 0 {
 		return fmt.Sprintf("%s (%v%%)", units.HumanSize(float64(reclaimable)), (reclaimable*100)/c.totalSize)
 	}
-	return fmt.Sprintf("%s", units.HumanSize(float64(reclaimable)))
+	return units.HumanSize(float64(reclaimable))
 }
 
 type diskUsageContainersContext struct {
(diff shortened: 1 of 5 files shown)


cli/command/formatter/disk_usage.go
```diff
@@ -243,7 +243,7 @@ func (c *diskUsageImagesContext) Reclaimable() string {
 	if c.totalSize > 0 {
 		return fmt.Sprintf("%s (%v%%)", units.HumanSize(float64(reclaimable)), (reclaimable*100)/c.totalSize)
 	}
-	return fmt.Sprintf("%s", units.HumanSize(float64(reclaimable)))
+	return units.HumanSize(float64(reclaimable))
```
aaronlehmann: Awesome!

dockerfiles/Dockerfile.dev
```diff
@@ -3,19 +3,25 @@ FROM    golang:1.8.3-alpine
...
+        git checkout -q "$VNDR_SHA" && \
         go build -v -o /usr/bin/vndr . && \
         rm -rf /go/src/* /go/pkg/* /go/bin/*
 
+ARG     BINDATA_SHA=a0ff2567cfb70903282db057e799fd826784d41d
 RUN     go get github.com/jteeuwen/go-bindata/go-bindata && \
-        cp /go/bin/go-bindata /usr/bin && \
+        cd /go/src/github.com/jteeuwen/go-bindata/go-bindata && \
+        git checkout -q "$BINDATA_SHA" && \
+        go build -v -o /usr/bin/go-bindata . && \
```
aaronlehmann: I wonder if it would be a little more idiomatic to change `$PATH` and `go install` into `/go/bin`. I don't think it really matters though - installing the tools into `/usr/bin` of a container image is reasonable.

aaronlehmann:
LGTM

dnephin:
Ya, multi-stage build would be great, but circleCI is still on 17.03.0-ce, so we're waiting for them to upgrade to 17.06

PR #325, opened 2017-07-12
Give an example for --reserve-memory

mistyhacks:
Carrying https://github.com/moby/moby/pull/32631/

Incorporated feedback from that PR.

### DIFF
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -696,6 +696,74 @@ appends a new placement preference after all existing placement preferences.
 `--placement-pref-rm` removes an existing placement preference that matches the
 argument.
 
+### Specify memory requirements and constraints for a service (--reserve-memory and --limit-memory)
+
+If your service needs a minimum amount of memory in order to run correctly,
+you can use `--reserve-memory` to specify that the service should only be
+scheduled on a node with this much memory available to reserve. If no node is
+available that meets the criteria, the service is not scheduled, but remains in
+a pending state.
+
+The following example requires that 4GB of memory be available and reservable
+on a given node before scheduling the service to run on that node.
+
+```bash
+$ docker service create --reserve-memory=4GB --name=too-big nginx:alpine
+```
+
+The managers won't schedule a set of containers on a single node whose combined
+reservations exceed the memory available on that node.
+
+After a task is scheduled and running, `--reserve-memory` does not enforce a
+memory limit. Use `--limit-memory` to ensure that a task uses no more than a
+given amount of memory on a node. This example limits the amount of memory used
+by the task to 4GB. The task will be scheduled even if each of your nodes has
+only 2GB of memory, because `--limit-memory` is an upper limit.
+
+```bash
+$ docker service create --limit-memory=4GB --name=too-big nginx:alpine
+```
+
+Using `--reserve-memory` and `--limit-memory` does not guarantee that Docker
+will not use more memory on your host than you want. For instance, you could
+create many services, the sum of whose memory usage could exhaust the available
+memory.
+
+You can prevent this scenario from exhausting the available memory by taking
+into account other (non-containerized) software running on the host as well. If
+`--reserve-memory` is greater than or equal to `--limit-memory`, Docker won't
+schedule a service on a host that doesn't have enough memory. `--limit-memory`
+will limit the service's memory to stay within that limit, so if every service
+has a memory-reservation and limit set, Docker services will be less likely to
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/service_create.md
```diff
@@ -696,6 +696,72 @@ appends a new placement preference after all existing placement preferences.
...
+  "normal" requirement and the 500MB "peak" requirement". This assumes that when
+  this service is at "peak", other services or non-container workloads probably
+  won't be.
+
+The approach you take depends heavily on the memory-usage patterns of your
+workloads. You should test under normal and peak conditions before settling
+on an approach.
+
+On Linux, you can also limit a service's overall memory footprint on a given
+host, `cgroups` or other mechanisms at the level of the operating system.
```
aaronlehmann: I'm having trouble understanding this sentence; I think maybe a word is missing somewhere.

docs/reference/commandline/service_create.md
```diff
@@ -696,6 +696,72 @@ appends a new placement preference after all existing placement preferences.
...
+create many services, the sum of whose memory usage could exhaust the available
+memory.
+
+You can prevent this scenario from exhausting the available memory by taking
+into account other (non-containerized) software running on the host as well. If
+`--reserve-memory` is greater than or equal to `--limit-memory`, Docker won't
+schedule a service on a host that doesn't have enough memory. `--limit-memory`
+will limit the service's memory to stay within that limit, so if every service
+has a memory-reservation and limit set, Docker services  won't saturate the
+host.
```
aaronlehmann: I think this over-promises slightly - even if all services have memory reservations and limits, there could still be other non-containerized (or non-service container) software on the host using memory, as you point out earlier, which would cause Docker to exhaust the available memory.

aaronlehmann:
Overall this is very good. It's a difficult topic to explain.

I think the most important thing to get across is that the scheduler doesn't look at actual measurements of real-time memory use. It only pays attention to the reservations. So there's no magic that makes sure enough memory is available before scheduling a container on a certain host - it's entirely up to the operator to specify reservations, and make sure that the reserved memory will actually be available (for example, by limiting service memory use as well). The text does a good job of explaining this.

I would also add special emphasis on the danger of mixing services with reservations and services without reservations. Since a service without a memory reservation doesn't have any memory reserved for it, the scheduler would still be able to schedule that service to a node where all the memory is reserved. If you want to use memory reservations, it's important to use them for all services, otherwise the services without reservations can negate the careful work you did to keep memory usage under control.

docs/reference/commandline/service_create.md
mstanleyjones: Yes, "using" is missing.

mstanleyjones:
Addressed feedback. PTAL again.

docs/reference/commandline/service_create.md
```diff
@@ -696,6 +696,74 @@ appends a new placement preference after all existing placement preferences.
 `--placement-pref-rm` removes an existing placement preference that matches the
 argument.
 
+### Specify memory requirements and constraints for a service (--reserve-memory and --limit-memory)
+
+If your service needs a minimum amount of memory in order to run correctly,
+you can use `--reserve-memory` to specify that the service should only be
+scheduled on a node with this much memory available to reserve. If no node is
+available that meets the criteria, the service is not scheduled, but remains in
```
aaronlehmann: "the task is not scheduled"

docs/reference/commandline/service_create.md
```diff
@@ -696,6 +696,74 @@ appends a new placement preference after all existing placement preferences.
...
+  this service is at "peak", other services or non-container workloads probably
+  won't be.
+
+The approach you take depends heavily on the memory-usage patterns of your
+workloads. You should test under normal and peak conditions before settling
+on an approach.
+
+On Linux, you can also limit a service's overall memory footprint on a given
+host at the level of the host operating system, using `cgroups` or other
+relevant operating system tools.
```
aaronlehmann: Just want to confirm it's possible to do this at the host OS level for a swarm service. ping @crosbymichael @justincormack

thaJeztah: Probably by creating a cgroup, and setting it as  parent-cgroup on the daemon

nevmerzhitsky:
Finish this change guys, please. The community have a lot of issues with limiting resources for containers. In the swarm especially. IMO

thaJeztah:
carried in https://github.com/docker/cli/pull/2316

PR #326, opened 2017-07-12
Fix test and lint failures on master

dnephin:
Fix the conflict between the changes in #299 and #277, which just merged together.

Also split up the test cases from #277


thaJeztah:
ignoring codecov guess it's because master was broken

PR #328, opened 2017-07-12
Fix warning in docker CLI when `swarm ca` is called with flags other than `--rotate`

cyli:
As @albers pointed out in https://github.com/docker/cli/pull/207/files#r125589944, the `flagCACert` var was duplicated.  Fixed this to be `flagCertExpiry` instead, and added a test for the invalid flags.

image

cc @dnephin @thaJeztah

### DIFF
--- a/cli/command/swarm/ca_test.go
+++ b/cli/command/swarm/ca_test.go
@@ -2,10 +2,14 @@ package swarm
 
 import (
 	"bytes"
+	"io/ioutil"
+	"os"
 	"testing"
 	"time"
 
+	"github.com/docker/cli/cli/internal/test"
 	"github.com/docker/docker/api/types/swarm"
+	"github.com/docker/docker/pkg/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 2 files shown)


cli/command/swarm/ca_test.go
```diff
@@ -34,6 +38,70 @@ func TestDisplayTrustRootNoRoot(t *testing.T) {
...
+
+			"--ca-cert=" + tmpfile.Name(),
+			"--external-ca=protocol=cfssl,url=https://some.com/https/url",
+		},
+	}
+
+	for _, args := range errorTestCases {
+		buf := new(bytes.Buffer)
+		cmd := newCACommand(
+			test.NewFakeCliWithOutput(&fakeClient{
```
dnephin: Since `buf` is unused this should use `NewFakeCli()` and omit the second argument.

cyli: Good point, thanks.

dnephin:
I've got a fix for the lint error  in #329

cyli:
Thanks @dnephin - I'll rebase as soon as it's merged.

PR #329, opened 2017-07-12
Update gometalinter

dnephin:
Another fix for master.

There was a bug in `b4ebfc554d8f36bfef1f180ad0aaaaac99b430d` (the old pinned commit) where it was importing a package that it wasn't vendoring. That import was removed, which means that `go get` doesn't download it anymore. This is causing the lint to fail on master because it's no longer possible to build `b4ebfc554d8f36bfef1f180ad0aaaaac99b430d` now that `gometalinter` master has removed the dependency.


thaJeztah:
yolo, should be fine to merge

PR #332, opened 2017-07-13
Docs fixups for password stdin

tych0:
Add various documentation (--help output, man page, etc) notes about the new `--password-stdin`. Also add things to the shell completion profiles.

cc @thaJeztah

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -2684,7 +2684,7 @@ _docker_login() {
 
 	case "$cur" in
 		-*)
-			COMPREPLY=( $( compgen -W "--help --password -p --username -u" -- "$cur" ) )
+			COMPREPLY=( $( compgen -W "--help --password --password-stdin -p --username -u" -- "$cur" ) )
 			;;
 	esac
 }
--- a/contrib/completion/fish/docker.fish
+++ b/contrib/completion/fish/docker.fish
@@ -229,6 +229,7 @@ complete -c docker -A -f -n '__fish_seen_subcommand_from load' -s i -l input -d 
 complete -c docker -f -n '__fish_docker_no_subcommand' -a login -d 'Log in to a Docker registry server'
 complete -c docker -A -f -n '__fish_seen_subcommand_from login' -l help -d 'Print usage'
 complete -c docker -A -f -n '__fish_seen_subcommand_from login' -s p -l password -d 'Password'
+complete -c docker -A -f -n '__fish_seen_subcommand_from login' -l password-stdin -d 'Read password from stdin'
 complete -c docker -A -f -n '__fish_seen_subcommand_from login' -s u -l username -d 'Username'
 
 # logout
--- a/contrib/completion/zsh/_docker
+++ b/contrib/completion/zsh/_docker
@@ -2761,6 +2761,7 @@ __docker_subcommand() {
             _arguments $(__docker_arguments) -A '-*' \
                 $opts_help \
                 "($help -p --password)"{-p=,--password=}"[Password]:password: " \
+                "($help)--password-stdin[Read password from stdin]" \
                 "($help -u --user)"{-u=,--user=}"[Username]:username: " \
                 "($help -)1:server: " && ret=0
             ;;
--- a/docs/reference/commandline/login.md
+++ b/docs/reference/commandline/login.md
@@ -22,9 +22,10 @@ Log in to a Docker registry.
 If no server is specified, the default is defined by the daemon.
 
 Options:
-      --help              Print usage
-  -p, --password string   Password
-  -u, --username string   Username
+      --help                    Print usage
+  -p, --password       string   Password
+      --password-stdin          Read password from stdin
+  -u, --username       string   Username
 ```
 
 ## Description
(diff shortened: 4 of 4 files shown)


cli/command/registry/login.go
```diff
@@ -41,7 +41,7 @@ func NewLoginCommand(dockerCli command.Cli) *cobra.Command {
 	flags := cmd.Flags()
 
 	flags.StringVarP(&opts.user, "username", "u", "", "Username")
-	flags.StringVarP(&opts.password, "password", "p", "", "Password")
+	flags.StringVarP(&opts.password, "password", "p", "", "Password (deprecated, use --password-stdin)")
```
thaJeztah: Hm, is it officially deprecated? If so, it needs to go through the deprecation steps; https://github.com/docker/cli/blob/master/docs/deprecated.md

Given that this flag has been around for a _long_ time, I'm not sure we can deprecate it (at least, it would require a really long deprecation period)

FWI, the `pflags` package allows marking a flag as deprecated, which will hide it, and print a message if used

contrib/completion/bash/docker
```diff
@@ -2684,7 +2684,7 @@ _docker_login() {
 
 	case "$cur" in
 		-*)
-			COMPREPLY=( $( compgen -W "--help --password -p --username -u" -- "$cur" ) )
+			COMPREPLY=( $( compgen -W "--help --password -p --username -u --password-stdin" -- "$cur" ) )
```
thaJeztah: These should be kept in alphabetical order

contrib/completion/fish/docker.fish
```diff
@@ -230,6 +230,7 @@ complete -c docker -f -n '__fish_docker_no_subcommand' -a login -d 'Log in to a
 complete -c docker -A -f -n '__fish_seen_subcommand_from login' -l help -d 'Print usage'
 complete -c docker -A -f -n '__fish_seen_subcommand_from login' -s p -l password -d 'Password'
 complete -c docker -A -f -n '__fish_seen_subcommand_from login' -s u -l username -d 'Username'
+complete -c docker -A -f -n '__fish_seen_subcommand_from login' -l password-stdin -d 'Read password from stdin'
```
thaJeztah: Can you keep these alphabetical as well?

contrib/completion/zsh/_docker
```diff
@@ -2762,6 +2762,7 @@ __docker_subcommand() {
                 $opts_help \
                 "($help -p --password)"{-p=,--password=}"[Password]:password: " \
                 "($help -u --user)"{-u=,--user=}"[Username]:username: " \
+                "($help)--password-stdin[Read password from stdin]" \
```
thaJeztah: Can you keep these alphabetical as well?

docs/reference/commandline/login.md
```diff
@@ -23,8 +23,9 @@ If no server is specified, the default is defined by the daemon.
 
 Options:
       --help              Print usage
-  -p, --password string   Password
+  -p, --password string   Password (deprecated, use --password-stdin)
   -u, --username string   Username
+      --password-stdin    Read password from stdin
```
thaJeztah: same here

thaJeztah: Also, can you add an example on how to use this flag in the examples section below?

cli/command/registry/login.go
tych0: https://docs.docker.com/engine/#feature-deprecation-policy says we simply need to label it and let it sit for 3+ releases. I don't mind waiting a while, but it seems good to tell people it's bad.

In any case. I've dropped those bits from this patch.

thaJeztah: It really depends on what's being deprecated, e.g. "deprecate `docker run`" in favour of `docker container run`" won't be something that could be done in three releases; some features cannot be removed.

Removing the `--email` flag from `docker login` (even though it was non-functional for more than a year, and had been printing a deprecation message for all that time) was already problematic

docs/reference/commandline/login.md
```diff
@@ -40,6 +41,15 @@ adding the server name.
 $ docker login localhost:8080
 ```
 
+### Provide a password automatically
```
thaJeztah: Should we just use `stdin` in the title?

```markdown
### Provide a password using STDIN
```

docs/reference/commandline/login.md
```diff
@@ -40,6 +41,15 @@ adding the server name.
...
+
+To provide a password (e.g. to script a Docker login command), use
+--password-stdin:
```
thaJeztah: can you put this in back tics (`--password-stdin`)?

Perhaps something like;

    ### Provide a password using STDIN

    To run the `docker login` command non-interactively, you can set the
    `--password-stdin` flag to provide a password through `STDIN`. Using
    `STDIN` prevents the password from ending up in the Shell's history,
    or log-files.

    The following example reads a password from a file, and passes it to the
    `docker login` command using `STDIN`:

    ```bash
    $ cat ~/my_password.txt | docker login --username foo --password-stdin
    ```

thaJeztah: (better suggestions welcome 👍 )

tych0: Sure, I just used your paragraph with one minor change. Thanks!

PR #333, opened 2017-07-13
Improve docs about ARG in FROM

dnephin:
See https://github.com/docker/for-linux/issues/59 and https://github.com/moby/moby/pull/31352#issuecomment-315105319 and https://github.com/moby/moby/issues/34008

### DIFF
--- a/docs/reference/builder.md
+++ b/docs/reference/builder.md
@@ -530,8 +530,10 @@ FROM extras:${CODE_VERSION}
 CMD  /code/run-extras
 ```
 
-To use the default value of an `ARG` declared before the first `FROM` use an
-`ARG` instruction without a value:
+An `ARG` declared before a `FROM` is outside of a build stage, so it
+can't be used in any instruction after a `FROM`. To use the default value of
+an `ARG` declared before the first `FROM` use an `ARG` instruction without
+a value inside of a build stage:
 
 ```Dockerfile
 ARG  SETTINGS=default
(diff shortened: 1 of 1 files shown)


docs/reference/builder.md
```diff
@@ -530,8 +530,10 @@ FROM extras:${CODE_VERSION}
 CMD  /code/run-extras
 ```
 
-To use the default value of an `ARG` declared before the first `FROM` use an
-`ARG` instruction without a value:
+An `ARG` declared before a `FROM` is outside of a build stage, so it
```
tonistiigi: Maybe better to mention that every stage has individual args, then it is better to understand that "arg-in-from" is not an exception but just follows the rule.

How about this example:
```
ARG VERSION=latest
FROM busybox:$VERSION
ARG VERSION
RUN echo $VERSION > image_version
```

mstanleyjones: Maybe something like "An `ARG` instructions is only relevant for the next `FROM` instruction that comes after it in the Dockerfile, and then using this example to illustrate that.

dnephin: > Maybe something like "An ARG instructions is only relevant for the next FROM instruction that comes after it in the Dockerfile,

That's not exactly correct. There are two cases:
* any `ARG` before the first `FROM` can be used in any `FROM` line
* any `ARG` within a build stage (after a `FROM`) can be used in that build stage

I can add more details to the `ARG` reference section as well (this stuff is under `FROM`) , but I think this line needs to be updated as well something along the lines of what I have.

dnephin:
I've updated to include more details, and added some more structure to the `ARG` section, PTAL

dnephin:
Makes sense, I've swapped the examples.

tonistiigi:
LGTM

PR #324, opened 2017-07-12
docs: add documentation for dm.libdm_log_level

cyphar:
This is a new option added specifically to allow for debugging of bugs
in Docker's storage drivers or libdm itself.

See docker/docker#33845
Signed-off-by: Aleksa Sarai <asarai@suse.de>

PR #334, opened 2017-07-14
Add bash completion for `--storage-opt dm.libdm_log_level`

albers:
This adds bash completion for https://github.com/moby/moby/pull/33845.
See also documentation PR #324.


albers:
This should go into 17.07.0 because the feature is present in that release.

PR #340, opened 2017-07-17
Fix: skip build-cache prune on unsupported versions

thaJeztah:
**- What I did**

Disabled "build-cache prune" on older daemons

**- How I did it**

**- How to verify it**

Running system prune against an older daemon should not print "build cache" in the confirmation message, and should not print a "version" error when running

```bash
DOCKER_API_VERSION=1.30 docker system prune
```

### DIFF
--- a/cli/command/system/client_test.go
+++ b/cli/command/system/client_test.go
@@ -1,0 +1,15 @@
+package system
+
+import (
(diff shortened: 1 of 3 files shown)


tonistiigi:
LGTM

PR #107, opened 2017-05-17
Replace makefile with a bash script

dnephin:
Related to #99 

Replace a 70 line `.PHONY` Makefile with 40 lines of bash which actually does a lot more than the original Makefile. This bash script will allow the addition of new targets to the `Makefile` without having to touch `./tasks`.

`shell` prompts are provided to all images without any additional configuration.

I don't necessarily think this  is the best solution, but I think it's an improvement over the Makefile.

Issue #99, opened 2017-05-17
Improve makefile situation


tonistiigi:
A couple of issues with Makefiles in this repo:

The first issue is that they are full of `@` output suppression. A lot of the targets execute for a long time and I can't see what is happening or if anything is stuck. Also `docker build` uses `-q` for some reason.

The other problem is the duplication of content in multiple Makefiles and aggressive use of mounts. Bind mounts break all cases when building against a remote API and they are almost never needed. By not using mounts you have similar performance and better caching. Similarily I see no reason to define these local duplicate targets that can easily go out of sync and have no guarantees of correctness as they rely on the state of host system. By looking at issues in this repo there are already multiple reports of these problems showing up.

My proposal is:

- Replace `Makefile` with `docker.Makefile`
- Add a build target that copies artifacts back to the host to keep the current behavior. `GOOS` can be sent with a build arg.
- Same for vendor and go generate target
- Build step should run as part of a Dockerfile not in a separate container. Target stages can be used to build without tests or get devenv.
- Dev-env target can keep using the mounts(for now)
- Separate Dockerfile for CI should not be needed. At least after ci supports multi-stage builds if we want to use some better syntax for the local version.

@dnephin

dnephin:
I think you're looking at an older version? Some of these issues are already fixed.

* the `docker build -q` is fixed
* the Dockerfile.ci is gone, which means just about all the duplication is gone

I agree with removing the `@` from Makefile targets, the output is valuable.

> Bind mounts break all cases when building against a remote API

bind mounts are necesasry. See #87. There's no reason to support building against a remote API.

> and better caching

What is the better caching? I don't know of any operation that can be cached that isn't already.

> Similarly I see no reason to define these local duplicate targets that can easily go out of sync and have no guarantees of correctness as they rely on the state of host system.

Do you mean the targets in `Makefile`? They are not local targets, they are the "in-container" targets. We can definitely remove all the duplication with my `build.sh` proposed solution below.

> Replace Makefile with docker.Makefile

I think this is the wrong idea. We started some discussion about this here: https://github.com/docker/cli/pull/78#discussion_r116513360

Makefiles are a bad choice for this use case. We should replace `docker.Makefile` with a  `./build.sh` until we have a better tool.

> Add a build target that copies artifacts back to the host to keep the current behavior. GOOS can be sent with a build arg.

Not sure what you mean by this. What is the "current" behaviour? Where do we need to set `GOOS` ?

[...]

PR #343, opened 2017-07-17
Replace makefile with a bash script

dnephin:
Related to #99 
(replaces #107 which seemed to be bugged, github was not triggering any webhooks for that PR)

Replace the `.PHONY` Makefile with a bash script which does a lot more than the original Makefile. This bash script will allow the addition of new targets to the `Makefile` without having to touch `./tasks`. Allows us to use the same script for DEV and CI, and adds support for different workflows.

`shell` prompts are provided to all images without any additional configuration.

I don't necessarily think this  is the best solution, but I think it's an improvement over the Makefile.


dnephin:
ping again

vdemeester:
ping @dnephin should we close this one for now ?

dnephin:
I still think this is a better solution. If we close it we should have some reason for closing it. I'd like to see some arguments for why the Makefile is a better solution.

PR #344, opened 2017-07-18
Use version in VERSION file instead `unknown-version` when build cli

coolljt0725:
when build cli, the version of the `docker version` is `unknown-version`,
we should use the version in the `VERSION` file.

Signed-off-by: Lei Jitang <leijitang@huawei.com>

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/build/.variables
+++ b/scripts/build/.variables
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -eu
 
-VERSION=${VERSION:-"unknown-version"}
+VERSION=${VERSION:-$(cat VERSION)}
 GITCOMMIT=${GITCOMMIT:-$(git rev-parse --short HEAD 2> /dev/null || true)}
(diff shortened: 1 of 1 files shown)


scripts/build/.variables
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -eu
 
-VERSION=${VERSION:-"unknown-version"}
+VERSION=${VERSION:-$(cat VERSION)}
```
dnephin: This file could be called with a different working dir, so I think it needs to do something like this:

https://github.com/docker/cli/pull/343/files#diff-11460b30ebb8794f3c4f1061c483157dR4

dnephin:
ping. Do you want to made these changes or should we close this PR?

coolljt0725:
@dnephin  sorry for the delay, let's close this one.

PR #345, opened 2017-07-18
Remove useless prune package 👼

vdemeester:
The `cli/command/prune` just does some aliasing of `Prune*` methods,
doesn't have any use.


thaJeztah:
damn, you won, I'll have to rebase now, haha

cli/command/system/prune.go
```diff
@@ -52,18 +55,38 @@ const confirmationTemplate = `WARNING! This will remove:
 {{- end }}
 Are you sure you want to continue?`
 
+// runContainerPrune executes a prune command for containers
+func runContainerPrune(dockerCli command.Cli, filter opts.FilterOpt) (uint64, string, error) {
+	return container.RunPrune(dockerCli, filter)
+}
```
dnephin: Most of these aliases have the exact same signature, right?

I don't think we need them at all. We can call the underlying method directly.

thaJeztah: ah, yes, we could probably remove them altogether

PR #347, opened 2017-07-18
support --compose-file - as stdin

mmariani:
As discussed in https://github.com/moby/moby/issues/34036

### DIFF
--- a/cli/command/stack/deploy_composefile.go
+++ b/cli/command/stack/deploy_composefile.go
@@ -2,6 +2,7 @@ package stack
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -22,7 +23,7 @@ import (
 )
 
 func deployCompose(ctx context.Context, dockerCli command.Cli, opts deployOptions) error {
-	configDetails, err := getConfigDetails(opts.composefile)
+	configDetails, err := getConfigDetails(opts.composefile, dockerCli.In())
 	if err != nil {
 		return err
 	}
@@ -118,16 +119,24 @@ func propertyWarnings(properties map[string]string) string {
 	return strings.Join(msgs, "\n\n")
 }
 
-func getConfigDetails(composefile string) (composetypes.ConfigDetails, error) {
+func getConfigDetails(composefile string, stdin io.Reader) (composetypes.ConfigDetails, error) {
 	var details composetypes.ConfigDetails
 
-	absPath, err := filepath.Abs(composefile)
-	if err != nil {
-		return details, err
+	if composefile == "-" {
+		workingDir, err := os.Getwd()
+		if err != nil {
+			return details, err
+		}
+		details.WorkingDir = workingDir
+	} else {
+		absPath, err := filepath.Abs(composefile)
+		if err != nil {
+			return details, err
+		}
+		details.WorkingDir = filepath.Dir(absPath)
 	}
-	details.WorkingDir = filepath.Dir(absPath)
 
-	configFile, err := getConfigFile(composefile)
+	configFile, err := getConfigFile(composefile, stdin)
 	if err != nil {
 		return details, err
 	}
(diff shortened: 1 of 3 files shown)


mmariani:
I would have added a test but I'm pretty new at go and don't see a clean way to mock /dev/stdin.

docs/reference/commandline/stack_deploy.md
```diff
@@ -57,6 +57,8 @@ Creating service vossibility_ghollector
 Creating service vossibility_lookupd
 ```
 
+The Compose file can also be provided as standard input with `--compose-file -`.
```
dnephin: I think this would be better in the help text for the `--compose-file` flag. Something like

```
Path to a Compose file, use "-" to read from stdin
```

cli/command/stack/deploy_composefile.go
```diff
@@ -121,11 +122,20 @@ func propertyWarnings(properties map[string]string) string {
 func getConfigDetails(composefile string) (composetypes.ConfigDetails, error) {
 	var details composetypes.ConfigDetails
 
-	absPath, err := filepath.Abs(composefile)
-	if err != nil {
-		return details, err
+	if composefile == "-" && runtime.GOOS != "windows" {
+		composefile = "/dev/stdin"
```
dnephin: I think we can make this easier to test, and supported on windows.

Instead of setting `composefile = "/dev/stdin"` here, only set WorkingDir.

You'll also need to pass `dockerCli.In()` as an `io.Reader` from `deployCompose()` where this is called, so we have access to it in `getConfigFile()`.

In `getConfigFile` before calling `ioutil.ReadFile()` check if `filename == "-"` and use this instead to get the bytes:

```go
bytes, err = ioutil.ReadAll(stdin)
```

mmariani:
this one should do, thanks

cli/command/stack/deploy_composefile.go
```diff
@@ -150,15 +159,24 @@ func buildEnvironment(env []string) (map[string]string, error) {
 	return result, nil
 }
 
-func getConfigFile(filename string) (*composetypes.ConfigFile, error) {
-	bytes, err := ioutil.ReadFile(filename)
+func getConfigFile(filename string, stdin io.Reader) (*composetypes.ConfigFile, error) {
```
thaJeztah: Wondering if this should just take an `io.Reader` in both cases @dnephin wdyt?

docs/reference/commandline/stack_deploy.md
mstanleyjones: I agree, then it doesn't need to be specifically dealt with in the reference file at all, unless you wanted to add an example (which would probably be good to have).

cli/command/stack/deploy_composefile.go
dnephin: Ya, that sounds like a good idea

mmariani: I did not do that, because it needs the filename in the return.

thaJeztah: Ah, I see; yes, we may have to refactor some things. None of these are exported functions, so I guess we can do that in a follow up.

cli/command/stack/deploy_composefile_test.go
```diff
@@ -25,13 +26,29 @@ services:
...
 	assert.Len(t, details.ConfigFiles, 1)
 	assert.Len(t, details.Environment, len(os.Environ()))
 }
 
+func TestGetConfigDetailsStdin(t *testing.T) {
+	content := `
+version: "3.0"
+services:
+foo:
+image: alpine:3.5
```
thaJeztah: Looks like the indentation got lost here. We're not testing if the content is actually parsed, so won't make a difference (currently); I see the other test also doesn't do this, but wondering if we should, e.g.;

```patch
diff --git a/cli/command/stack/deploy_composefile_test.go b/cli/command/stack/deploy_composefile_test.go
index 90d866a2..6e2a1f2c 100644
--- a/cli/command/stack/deploy_composefile_test.go
+++ b/cli/command/stack/deploy_composefile_test.go
@@ -37,15 +37,16 @@ func TestGetConfigDetailsStdin(t *testing.T) {
        content := `
 version: "3.0"
 services:
-foo:
-image: alpine:3.5
+  foo:
+    image: alpine:3.5
 `
        details, err := getConfigDetails("-", strings.NewReader(content))
        require.NoError(t, err)
        cwd, err := os.Getwd()
        require.NoError(t, err)
        assert.Equal(t, cwd, details.WorkingDir)
-       assert.Len(t, details.ConfigFiles, 1)
+       require.Len(t, details.ConfigFiles, 1)
+       assert.Equal(t, "3.0", details.ConfigFiles[0].Config["version"])
        assert.Len(t, details.Environment, len(os.Environ()))
 }
```

mmariani:
done, thanks for reminding. also updated to master

PR #348, opened 2017-07-18
Plugable secret backend

liron-l:
This commit extends SwarmKit secret management with pluggable secret
backends support.

Following previous commits:
1. docker/swarmkit@eebac27434d34708fac993f9f5181d106c5c2fae
2. docker/docker@08f7cf05268782a0dd8e4c41a4cc65fdf78d09f2

Added driver parameter to `docker secret` command.
Specifically:
1. `docker secret create [secret_name] --driver [driver_name]`
2.  Displaying the driver in
```
    $ docker secret ls
    $ docker secret inspect [secret_name]
    $ docker secret inspect [secret_name] -pretty
```

There is a bug in serialization of the secret data. Handled in docker/docker#34157.

**- What I did**

**- How I did it**

**- How to verify it**


cpuguy83:
Validations and tests appear to have failed but I can't see them in CircleCI.

liron-l:
Thanks @cpuguy83, waiting for the fix in docker/docker#34157 (tests should work after updating the vendor folder).

cli/command/formatter/secret.go
```diff
@@ -23,6 +23,7 @@ Labels:
 {{- range $k, $v := .Labels }}
  - {{ $k }}{{if $v }}={{ $v }}{{ end }}
 {{- end }}{{ end }}
+Driver:              	{{.Driver}}
 Created at:            	{{.CreatedAt}}
 Updated at:            	{{.UpdatedAt}}`
```
dnephin: It looks like the existing lines have a weird mix of tabs and spaces...

I think we'll need to fix this to be all spaces so that the new line aligns properly.

cli/command/secret/create.go
```diff
@@ -29,15 +30,18 @@ func newSecretCreateCommand(dockerCli command.Cli) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create [OPTIONS] SECRET file|-",
 		Short: "Create a secret from a file or STDIN as content",
-		Args:  cli.ExactArgs(2),
+		Args:  cli.RequiresRangeArgs(1, 2),
```
dnephin: The `Use:` line needs to be updated to make the second arg optional:

```
create [OPTIONS] SECRET [file | -]
```

cli/command/secret/create.go
```diff
@@ -46,19 +50,26 @@ func runSecretCreate(dockerCli command.Cli, options createOptions) error {
...
-		defer file.Close()
+	if options.driver != "" && options.file != "" {
+		return errors.Errorf("When using secret driver secret data must be empty")
 	}
 
-	secretData, err := ioutil.ReadAll(in)
-	if err != nil {
-		return errors.Errorf("Error reading content from %q: %v", options.file, err)
+	var secretData []byte
+	if options.driver == "" {
```
dnephin: The intent here will be more obvious is this check is changed to

```go
if options.file != "" {
    ...
}
```

cli/command/secret/create.go
```diff
@@ -46,19 +50,26 @@ func runSecretCreate(dockerCli command.Cli, options createOptions) error {
...
+				return err
+			}
+			in = file
+			defer file.Close()
+		}
+		var err error
+		secretData, err = ioutil.ReadAll(in)
+		if err != nil {
+			return errors.Errorf("Error reading content from %q: %v", options.file, err)
+		}
```
dnephin: I know you've just adding another condition here, but now that this function is growing I think it would be appropriate to extract this as a new function.

```go
secretData, err := readSecretData(options.file, dockerCli.In())
...
```

Even the initial `options.file != ""` check could be inside the function.

cli/command/formatter/secret.go
liron-l: Thanks, modified :+1:

cli/command/secret/create.go
liron-l: Good catch, thanks :+1:

cli/command/secret/create.go
liron-l: Fixed as part of function refactoring :+1:

cli/command/secret/create.go
liron-l: thanks @dnephin, makes sense.

cli/command/formatter/secret.go
```diff
@@ -23,6 +23,7 @@ Labels:
 {{- range $k, $v := .Labels }}
  - {{ $k }}{{if $v }}={{ $v }}{{ end }}
 {{- end }}{{ end }}
+Driver:                 {{.Driver}}
 Created at:            	{{.CreatedAt}}
 Updated at:            	{{.UpdatedAt}}`
```
dnephin: I still see tabs here

cli/command/secret/create.go
```diff
@@ -77,3 +79,24 @@ func runSecretCreate(dockerCli command.Cli, options createOptions) error {
...
+	}
+	if file != "-" {
+		var err error
+		in, err = system.OpenSequential(file)
+		if err != nil {
+			return nil, err
+		}
+		defer in.Close()
+	}
+	var err error
```
dnephin: this line is unused I think

cli/command/secret/create.go
```diff
@@ -77,3 +79,24 @@ func runSecretCreate(dockerCli command.Cli, options createOptions) error {
...
+	data, err := ioutil.ReadAll(in)
+	if err != nil {
+		return nil, errors.Errorf("Error reading content from %q: %v", file, err)
```
dnephin:
```go
errors.Wrapf(err, "error reading content from %q", file)
```

Errors should start with lowercase so they look more appropriate when wrapped.

cli/command/secret/create.go
liron-l: right. Thanks

cli/command/secret/create.go
liron-l: thanks, but those lines are not wrapped with anything (they are printed directly in the cli). I returned the error and wrapped with upper case in the main method. Let me know if this makes sense.

cli/command/formatter/secret.go
liron-l: Got it, you mean I should remove all tabs from the formatter :+1:

liron-l:
Thanks for the comments @dnephin! do you know what `vndr` is complaining?
CC @diogomonica @aaronlehmann @cpuguy83

dnephin:
If I run `make -f docker.Makefile vendor` locally I see the same problem. I think you vendored using an older version of `vndr`. If you run that command and check in the changes it should fix it.

liron-l:
Thanks @dnephin, it worked!

cpuguy83:
Can we do the vendoring in a separate commit?

liron-l:
Sure @cpuguy83 makes sense. I've updated the commit.

thaJeztah:
Does this also update the `docker-compose` format to support this?

thaJeztah:
I wonder if it would make sense to do the version bump in two stages; one to https://github.com/moby/moby/commit/d0a8e73e7b60f61db0c3799643aaccbbf33f3601 (last commit before the upstream related change was merged), and one to https://github.com/moby/moby/commit/0304c98d85404fe75a1b4a35d3c111931e062f41 (the diff is quite large https://github.com/moby/moby/compare/87df0e533b619c088091fd1e2310e92bb9a24822...0304c98d85404fe75a1b4a35d3c111931e062f41)

I see there's also some dependencies bumped in moby's vendor.conf; do we have to bump those as well accordingly (if used here?)

liron-l:
Thanks @thaJeztah.
Re docker-compose, I checked that setting `external: true` for the secret reference works with the new plugin subsystem. 
Do we need to support `driver: name` parameter, which will create the secret value?

Re version bump, let me know which commit history makes sense or which additional packages should be updated.

liron-l:
@thaJeztah and @dnephin I've added a new parameter to secrets definitions:
```
driver: my-driver
```
So now it's possible to run:
```
version: '3.3'
services:
  web:
    image: nginx
    secrets:                    
     - hello-secret
secrets:                       
  hello-secret:
    driver: secrets:next
```
Can you please review?

dnephin:
Sorry, the change should not be made to v3.3. It should go into v3.4. You can cherry-pick #358 to get v3.4, or we can add the compose changes in a separate PR. I think I would prefer a separate PR

liron-l:
Thanks @dnephin makes sense. I will push those changes in the next commit. 
I removed the compose changes, how do I re-trigger a build?

dnephin:
hmm, we've been having problems with github not triggering builds. I'm going to email github support.

liron-l:
Thanks @dnephin! was the issue resolved? can I trigger A build without closing the issue?

dnephin:
github said they are looking into it, and that it's likely a bug on their end.

I've been able to trigger the build by closing and re-opening the PR in some cases, so I'm going to try that now.

dnephin:
and in some cases CI fails with "fatal: Could not parse object" which is what's happening this time...

PR #349, opened 2017-07-18
Add scripts folder to shellcheck

jphuynh:
Is `scripts/gen/windows-resources` still used? I didn't change it as it looks like it's not used anymore.

### DIFF
--- a/scripts/gen/windows-resources
+++ b/scripts/gen/windows-resources
@@ -6,6 +6,7 @@
 set -eu -o pipefail
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+# shellcheck source=/go/src/github.com/docker/cli/scripts/build/.variables
 source $SCRIPTDIR/../build/.variables
 
 RESOURCES=$SCRIPTDIR/../winresources
@@ -15,30 +16,28 @@ trap 'rm -rf $TEMPDIR' EXIT
 
 if [ "$(go env GOHOSTOS)" == "windows" ]; then
 	WINDRES=windres
-	WINDMC=windmc
 else
 	# Cross compiling
 	WINDRES=x86_64-w64-mingw32-windres
-	WINDMC=x86_64-w64-mingw32-windmc
 fi
 
 # Generate a Windows file version of the form major,minor,patch,build (with any part optional)
-VERSION_QUAD=$(echo -n $VERSION | sed -re 's/^([0-9.]*).*$/\1/' | tr . ,)
+VERSION_QUAD=$(echo -n "$VERSION" | sed -re 's/^([0-9.]*).*$/\1/' | tr . ,)
 
 # Pass version and commit information into the resource compiler
 defs=
-[ ! -z $VERSION ]      && defs="$defs -D DOCKER_VERSION=\"$VERSION\""
-[ ! -z $VERSION_QUAD ] && defs="$defs -D DOCKER_VERSION_QUAD=$VERSION_QUAD"
-[ ! -z $GITCOMMIT ]    && defs="$defs -D DOCKER_COMMIT=\"$GITCOMMIT\""
+[ ! -z "$VERSION" ]      && defs+=( "-D DOCKER_VERSION=\"$VERSION\"")
+[ ! -z "$VERSION_QUAD" ] && defs+=( "-D DOCKER_VERSION_QUAD=$VERSION_QUAD")
+[ ! -z "$GITCOMMIT" ]    && defs+=( "-D DOCKER_COMMIT=\"$GITCOMMIT\"")
 
 function makeres {
-	$WINDRES \
-		-i $RESOURCES/$1 \
-		-o $3 \
-		-F $2 \
+	"$WINDRES" \
+		-i "$RESOURCES/$1" \
+		-o "$3" \
+		-F "$2" \
 		--use-temp-file \
-		-I $TEMPDIR \
-		$defs
+		-I "$TEMPDIR" \
+		${defs[*]}
 }
 
 makeres docker.rc pe-x86-64 rsrc_amd64.syso
--- a/scripts/test/unit
+++ b/scripts/test/unit
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -eu -o pipefail
 
-go test -tags daemon -v $@
+go test -tags daemon -v "$@"
--- a/scripts/test/watch
+++ b/scripts/test/watch
@@ -1,2 +1,3 @@
 #!/bin/sh
+# shellcheck disable=SC2016
 exec filewatcher -L 6 -x build -x script go test -timeout 10s -v './${dir}'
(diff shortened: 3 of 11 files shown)


dnephin:
> Is `scripts/gen/windows-resources` still used?

Yes, it's called by `go generate` in `cli/winresources/res_windows.go`, which is called from `scripts/build/windows`.

jphuynh:
Isn't that commented out in the go file?

dnephin:
No, that's what `go:generate` lines should look like: https://blog.golang.org/generate

scripts/test/unit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -eu -o pipefail
 
-go test -tags daemon -v $@
+go test -tags daemon -v "$@"
```
dnephin: This surprised me because I thought we wanted the word splitting to happen here, but I guess this is an array? 

The docs confirm that "$@" is actually correct, so :+1:

scripts/test/watch
```diff
@@ -1,2 +1,3 @@
 #!/bin/sh
-exec filewatcher -L 6 -x build -x script go test -timeout 10s -v './${dir}'
+# shellcheck disable=SC2154
+exec filewatcher -L 6 -x build -x script go test -timeout 10s -v "./${dir}"
```
dnephin: This is wrong. It needs to be single quotes around `./${dir}` that literal is passed to `filewatcher` and interpolated by it. I realize now that choosing bash-like variables for this tool was not a great option, but I haven't had a chance to change it yet.

Does the shellcheck whitelist need to exist with single quotes?

jphuynh:
Ha! I learned something today :)

scripts/validate/shellcheck
```diff
@@ -1,11 +1,26 @@
...
+FILES+=("scripts/test/unit")
+FILES+=("scripts/test/unit-with-coverage")
+FILES+=("scripts/test/watch")
+FILES+=("scripts/validate/check-git-diff")
 FILES+=("scripts/validate/shellcheck")
 
+RETURN="0"
 for f in "${FILES[@]}"; do
     shellcheck "$f"
+    RETURN=$((RETURN+$?))
```
dnephin: I think this script could be a lot smaller:

```bash
 #!/usr/bin/env bash
set -eu -o pipefail

shellcheck contrib/completion/bash/docker
find scripts/ -type f | grep -v scripts/winresources | xargs shellcheck
```

I think this has the following advantages

* we should be able to remove a bunch of the `shellcheck disable=SC1091` above the `source` because the shellcheck run has access to all the files. This might actually improve shellcheck coverage, not sure.
* as new files are added to `scripts/` we don't have to remember to add them here, because I'm sure we wont remember
* the intent is more obvious (we're running shellcheck on two sets of files)
* we don't need to remove the `set -e`, it's fine if we fail after the first run
* we could potentially run each with different `--exclude` if we ever wanted to exclude a code from one type of script

scripts/gen/windows-resources
```diff
@@ -15,30 +16,28 @@ trap 'rm -rf $TEMPDIR' EXIT
 
 if [ "$(go env GOHOSTOS)" == "windows" ]; then
 	WINDRES=windres
-	WINDMC=windmc
```
jphuynh: Looks like this variable was unused so I removed it. Can someone confirm?

scripts/gen/windows-resources
```diff
@@ -15,30 +16,28 @@ trap 'rm -rf $TEMPDIR' EXIT
...
 else
 	# Cross compiling
 	WINDRES=x86_64-w64-mingw32-windres
-	WINDMC=x86_64-w64-mingw32-windmc
```
jphuynh: Looks like this variable was unused so I removed it. Can someone confirm?

scripts/gen/windows-resources
dnephin: Sounds right. I copied these from `moby/moby` which used it for building the daemon binary as well. I guess I missed this line while I was pruning the file.

scripts/gen/windows-resources
```diff
@@ -15,30 +16,28 @@ trap 'rm -rf $TEMPDIR' EXIT
...
-		-o $3 \
-		-F $2 \
+		-i "$RESOURCES/$1" \
+		-o "$3" \
+		-F "$2" \
 		--use-temp-file \
-		-I $TEMPDIR \
-		$defs
+		-I "$TEMPDIR" \
+		"$defs"
```
dnephin: I think the build failures are related to this line?

I guess `defs` should be an array instead of a string?

scripts/test/watch
jphuynh: I wasn't sure for that since dir wasn't declared. 

The shellcheck whitelist code is actually wrong here. It will need to be present as:

`#shellcheck disable=SC2016` when I roll back to single quotes.

scripts/validate/shellcheck
jphuynh: I agree with most points although I found it nicer to not fail at the first error and see all the shellcheck warnings in one go. I don't have a strong opinion on it though and happy to change it if it feels less spammy.

dnephin: I guess you could still support running them all and collecting the error codes, but I think the other changes would still be a good idea.

jphuynh: I ended up using your solution with fail on first warning.

scripts/validate/shellcheck
```diff
@@ -1,11 +1,5 @@
...
-# Maintain an array of files to shellcheck not the best solution but will do for the time being
-FILES=()
-FILES+=("contrib/completion/bash/docker")
-FILES+=("scripts/validate/shellcheck")
-
-for f in "${FILES[@]}"; do
-    shellcheck "$f"
-done
+shellcheck contrib/completion/bash/docker
+find scripts/ -type f | grep -v scripts/winresources | xargs shellcheck -x
```
dnephin: Is this `-x` necessary? I think I was able to get it to pass without it. Not a big deal either way, just curious.

scripts/gen/windows-resources
```diff
@@ -6,6 +6,7 @@
 set -eu -o pipefail
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+# shellcheck source=/go/src/github.com/docker/cli/scripts/build/.variables
```
dnephin: I believe this can be removed, as well as the other 4 instances of it in the `script/build` files

scripts/validate/shellcheck
jphuynh: Correct we can remove it.

scripts/gen/windows-resources
jphuynh: The others can go but this particular one is needed because of `$SCRIPTDIR`

```
SC1090: Can't follow non-constant source. Use a directive to specify location.
```

dnephin: Ah, right! sounds good

jphuynh:
Hi folks. Any update on this PR? 🐱

PR #351, opened 2017-07-18
Misc cleanup of dockerfiles and circleci

dnephin:
Plus remove unnecessary prune function aliases.

Add `-d` to `go get` so we don't waste time building a binary before we change to the correct git sha. This was an even bigger problem with `gometalinter` because we don't remove `/go/bin` so we are actually running the unpinned master version.

Remove an unnecessary `apk add` from circle.yml. This package is already in the base image now.

Upgrade base images since the `17.05` tag is not supported anymore according to hub.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,7 @@ jobs:
 
   lint:
     working_directory: /work
-    docker: [{image: 'docker:17.05-git'}]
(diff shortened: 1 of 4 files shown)


aaronlehmann:
LGTM

PR #338, opened 2017-07-15
Service privileges: CLI reference docs & completion

thaJeztah:
This documents the Service privileges API changes, that were added in: https://github.com/moby/moby/commit/091b5e68ea735bf4e8ece708bbc8c413a32eab73 (https://github.com/moby/moby/pull/32339)

Carries the CLI changes of https://github.com/moby/moby/pull/32483, with some additional information about the values

ping @johnstep PTAL

PR #352, opened 2017-07-18
Improve bash completion for `--credential-spec`

albers:
This option is Windows specific and should only be available if the daemon runs on Windows.
Ref: #338


albers:
@mlaventure Would you mind reviewing this PR? It's been hanging around for several weeks now.

PR #353, opened 2017-07-18
Use native formatting in bash completion of nodes

albers:
Completion of nodes now uses native Docker commands that were not available when this function was created.

Hint for reviewers: `__docker_nodes` can be called directly, which makes testing the effect of `$DOCKER_COMPLETION_SHOW_NODE_IDS` and its options easier, e.g.
```bash
DOCKER_COMPLETION_SHOW_NODE_IDS=yes __docker_nodes --name --add hugo
```


vieux:
LGTM

albers:
@mlaventure Would you mind reviewing this PR? It's been hanging around for several weeks now.

PR #354, opened 2017-07-18
Add a warning when make is run outside of a container

dnephin:
The warning can be disabled by setting the environment variable

### DIFF
--- a/scripts/warn-outside-container
+++ b/scripts/warn-outside-container
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+set -eu
+
+target="${1:-}"
+
+if [[ -z "${DISABLE_WARN_OUTSIDE_CONTAINER:-}" ]]; then
+    (
+        echo
+        echo
+        echo "WARNING: you are not in a container."
+        echo "Use \"make -f docker.Makefile $target\" or set"
+        echo "DISABLE_WARN_OUTSIDE_CONTAINER=1 to disable this warning."
+        echo
+        echo "Press Ctrl+C now to abort."
+        echo
+    ) >&2
+    sleep 10
+fi
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@
 all: binary
 
 
+_:=$(shell ./scripts/warn-outside-container $(MAKECMDGOALS))
+
 .PHONY: clean
 clean: ## remove build artifacts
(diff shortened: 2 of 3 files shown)


scripts/warn-outside-container
```diff
@@ -0,0 +1,17 @@
...
+    (
+        echo
+        echo
+        echo "WARNING: you are not in a container."
+        echo "Use \"make -f docker.Makefile $target\" or set"
+        echo "DISABLE_WARN_OUTSIDE_CONTAINER=1 to disable this warning."
+        echo
+        echo
+    ) >&2
+    sleep 5
```
thaJeztah: Should we just `exit` here? Currently looks like

```bash
$ make binary

WARNING: you are not in a container.
Use "make -f docker.Makefile binary" or set
DISABLE_WARN_OUTSIDE_CONTAINER=1 to disable this warning.

(5 second "hang")

WARNING: binary creates a Linux executable. Use cross for macOS or Windows.
./scripts/build/binary
make: *** [binary] Error 1
```

dnephin: > Should we just exit here?

`exit` won't actually stop the make target from running. There's no way (that I could find) to force a target before every other target in a Makefile. We'd have to add it manually as a dependency to everything, and I don't think that's a good solution.

So this is implemented as just a shell call `$(shell ...)` which means we get stdout/stderr, but can't actually end the execution. Maybe there's some way to have to exit, but I couldn't find a way to do it with make.

dnephin: > Does it work with parallel installed? If it's not a requirement, perhaps we should detect, and skip parallel

It could work without parallel, but I don't see why we need to support running it without. We have a container that provides parallel, and we shouldn't support running it outside of a container.

`parallel` is only required for cross compiling from linux, so if someone is interested in building just a specific target they can run `script/build/osx`, `script/build/window` or `script/build/binary` which don't require parallel.

thaJeztah: > Maybe there's some way to have to exit, but I couldn't find a way to do it with make.

Perhaps just print `Press Ctrl+C now to abort this script.`

> It could work without parallel, but I don't see why we need to support running it without. We have a container that provides parallel, and we shouldn't support running it outside of a container.

Yes, you're right; should not be needed

dnephin: Added "Press Ctrl+C now to abort"

dnephin:
ping again

PR #356, opened 2017-07-18
Fix .github/CODEOWNERS patterns

dnephin:
Apparently they use https://git-scm.com/docs/gitignore#_pattern_format

I'm not sure if these `**` are necessary, the doc suggests that just a `/` might work.

I tried testing these against my fork, but nothing works there. I also tried testing by modifying files in this PR, which also doesn't work. I'm guessing it has to be in master.


vieux:
LGTM

PR #315, opened 2017-07-10
Sort swarm stacks and nodes using natural sorting

ripcurld0:
follow up to https://github.com/moby/moby/pull/31638

**- What I did**

Use natural sorting to display swarm stacks and nodes.

**- How I did it**

Import vbom.ml/util/sortorder and update Less(i, j int) bool in both `cli/command/node/list.go` and `cli/command/stack/list.go`

**- How to verify it**
Run unit-tests

PR #357, opened 2017-07-19
Sort services names in a natural order

ripcurld0:
Follow-up to #315 

@aaronlehmann Let me know if there are other resources that need to be natural-sorted.

**- What I did**

Use natural sorting to display swarm services

**- How I did it**

Using the `sortorder.NaturalLess` function and the `Sort` interface

**- How to verify it**

```
$ TESTDIR='cli/command/service' go test -v
```

### DIFF
--- a/cli/command/service/list_test.go
+++ b/cli/command/service/list_test.go
@@ -1,0 +1,32 @@
+package service
+
+import (
+	"testing"
+
+	"golang.org/x/net/context"
+
+	"github.com/docker/cli/cli/internal/test"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/swarm"
+	"github.com/docker/docker/pkg/testutil"
+	"github.com/docker/docker/pkg/testutil/golden"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestServiceListOrder(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{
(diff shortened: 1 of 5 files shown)


cli/command/service/list_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"bytes"
+	"testing"
+
+	"golang.org/x/net/context"
+
+	"github.com/docker/cli/cli/internal/test"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/swarm"
+	"github.com/docker/docker/pkg/testutil"
+	"github.com/moby/moby/pkg/testutil/golden"
```
dnephin: `docker/docker/pkg/testutil/golden`

cli/command/service/list_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestServiceListOrder(t *testing.T) {
+	buf := new(bytes.Buffer)
+	cli := test.NewFakeCliWithOutput(&fakeClient{
```
dnephin: `NewFakeCliWithOutput` is deprecated. Please use `NewFakeCli()`. It defaults to an output buffer.

You can get the buffer from `cli.OutBuffer()`

aaronlehmann:
LGTM

PR #306, opened 2017-07-06
Make compose file allow to specify names for non-external volume

lipingxue:
**- What I did**
Fixes https://github.com/docker/cli/issues/274

**- How I did it**
I changed the config_schema.json to allow names for non-external volumes, and also made change in function ```convertVolumeToMount``` accordingly.
**- How to verify it**
1. Try ``` docker stack deploy ``` with the following yaml file(specify customized volume name for non-external volume), and it works as expected
```
root@sc-rdops-vm02-dhcp-52-237:~/tmp_with_fix# cat /root/postgres2.yml 
version: "3.3"
 
services:
 
  postgres:
    image: postgres
    ports:
      - "5432:5432"
    volumes:
      - postgres_vol:/var/lib/data
    environment:
      - "POSTGRES_PASSWORD=secretpass"
      - "PGDATA=/var/lib/data/db"
    deploy:
      restart_policy:
        condition: on-failure
      placement:
        constraints:
          - node.role == worker
volumes:
   postgres_vol:
      name: "postgres_vol@sharedVmfs-1"
      driver: "vsphere"
      driver_opts:
        size: "1GB"
root@sc-rdops-vm02-dhcp-52-237:~/tmp_with_fix# 


root@sc-rdops-vm02-dhcp-52-237:~/tmp_with_fix# ./docker-linux-amd64 stack deploy -c /root/postgres2.yml postgres
Creating network postgres_default
Creating service postgres_postgres
root@sc-rdops-vm02-dhcp-52-237:~/tmp_with_fix# docker volume ls
DRIVER              VOLUME NAME
vsphere:latest      postgres_vol@sharedVmfs-1

root@sc-rdops-vm02-dhcp-52-237:~/tmp_with_fix# docker service ps postgres_postgres 
ID            NAME                 IMAGE            NODE                       DESIRED STATE  CURRENT STATE           ERROR  PORTS
qx2nvqqwcqyu  postgres_postgres.1  postgres:latest  sc-rdops-vm02-dhcp-52-237  Running        Running 12 seconds ago        
```
2. Try ``` docker stack deploy ``` with the following yaml file(specify volume name for external volume), and it returns with proper error
```
root@sc-rdops-vm02-dhcp-52-237:~# cat postgres_external.yml 
version: "3.3"
 
services:
 
  postgres:
    image: postgres
    ports:
      - "5432:5432"
    volumes:
      - postgres_vol:/var/lib/data
    environment:
      - "POSTGRES_PASSWORD=secretpass"
[...]

PR #358, opened 2017-07-19
[WIP] Add Compose format 3.4

dnephin:
Don't merge this PR. Just putting up this commit so that contributors can use this as a base for new compose file changes. We shouldn't introduce v3.4 until 17.08.


dnephin:
This was merged as part of #306

PR #360, opened 2017-07-19
Add support for update order in compose deployments

### LINKED ISSUE #346: Include an "order" key in the "update_config"
When updating a Docker service, the following option exists:

```bash
      --update-order string                Update order ("start-first"|"stop-first")
```

This can be used to create rolling update deployments, even if a service has just one replica.

I propose adding the following key to the Docker Compose spec, which will allow for these updates to work with `docker deploy -c ...`.

```yml
    deploy:
      replicas: 1
      update_config:
        parallelism: 1
        order: start-first | stop-first
```

I would be happy to contribute the needed code if that's something we'd like to have in the spec.

akalipetis:
Fixes #346

**- What I did**

Added support for the "order" key in the "update_config":

```yml
    deploy:
      replicas: 1
      update_config:
        parallelism: 1
        order: start-first | stop-first
```

**- How I did it**

Added the new key in the Compose V3.4 schema and hooked it with the implementation that already existed for service create/update.

**- How to verify it**

```bash
docker deploy -c docker-compose.yml demo
```

using this stack:

```yml
version: '3.4'

services:
  nginx:
    image: nginx:1.13.1-alpine
    deploy:
      replicas: 1
      update_config:
        parallelism: 1
        order: start-first
    ports:
      - 80:80
```

This will create a rolling update, even though we're just using 1 replica.

Support for `update_config.order: start-first | stop-first` in Docker Compose version 3.4.

large_smart_squirrel


akalipetis:
/cc @dnephin I have cherry-picked https://github.com/docker/cli/pull/360/commits/da96115eafa36f25442440e5678dfbe4e7f2eba2 and also included https://github.com/docker/cli/pull/360/commits/b882ab3f4ff7f3d13ae9ed2efa5ded4654e034ae which (hopefully) rebuilds correctly the binary data.

This PR is based on #358

dnephin:
Looks like CI is broken on this PR. We've been having problems with that on a few PRs

akalipetis:
Thanks @dnephin - I'll definitely do that too. In regards to the CI, I thought it was because I had not included the `-s` flag in one commit and Gordon did not like me 😂

akalipetis:
@dnephin added a full-example test, but I still have issues in the following tests:

* `validate` - I have changed to schema to include the `order` key, should I rollback this? This will crash the full example, since it needs v3.4 with order included to work
* `lint` - no idea really, when I open the logs they're stuck in the `docker version` step loading without ever showing anything...

cli/compose/schema/bindata.go
```diff
@@ -303,4 +327,3 @@ func _filePath(dir, name string) string {
 	cannonicalName := strings.Replace(name, "\\", "/", -1)
 	return filepath.Join(append([]string{dir}, strings.Split(cannonicalName, "/")...)...)
 }
-
```
dnephin: I think this might be why `validate` is failing. Do you have precommit hooks that remove this line?

akalipetis: Oh, I see. I'll fix this and update the PR. Thanks 👍

akalipetis:
@dnephin seems like this was the issue after all 👍 - let me know if there's anything else you'd like me to do 😄

dnephin:
I think you can remove the WIP from the title now.

Needs a rebase since we merged another PR which added the 3.4 schema. You can ignore the conflicts in `bindata.go`, just regenerate the file and add that.

Then squash your commits and this should be ready

akalipetis:
Done 👌

thaJeztah:
thanks!

hairyhenderson:
@thaJeztah FYI this is missing in the docs for the `update_config` field: https://docs.docker.com/compose/compose-file/#update_config - I've logged an issue in the docs repo: https://github.com/docker/docker.github.io/issues/5088

thaJeztah:
@hairyhenderson oh, thanks for spotting that! Looks like I forgot checking for that 😅

PR #362, opened 2017-07-19
Skip inspect of built-in networks on stack deploy

alexmavr:
**- What I did**

The current flow of `docker stack deploy -c` performs a network inspect operation for every external network. However, built-in networks are always known to exist so these inspect calls are redundant. This PR removes them.

**- How I did it**
For every external network in a compose file, omit the network inspect operation if it's a built-in network ("none", "default", "container", "bridge" or "host")

**- How to verify it**
- Use the provided unit test

or 

- Build a docker API server that returns a 404 for the "host" and "bridge" networks (such as https://github.com/docker/swarm) and run the following compose file:

```
version: '3.2'
services:
  api:
    image: nginx
    networks:
      host : null
networks:
  host:
    external: true
```

Then, verify the deployed task is attached to the "host" network on the target node.

* Stack deploy for compose files no longer inspects built-in networks

important-stuff

Signed-off-by: Alex Mavrogiannis <alex.mavrogiannis@docker.com>

### DIFF
--- a/cli/command/stack/deploy_composefile_test.go
+++ b/cli/command/stack/deploy_composefile_test.go
@@ -42,6 +42,8 @@ func (n notFound) NotFound() bool {
 
 func TestValidateExternalNetworks(t *testing.T) {
 	var testcases = []struct {
+		inspected       bool
+		noInspect       bool
 		inspectResponse types.NetworkResource
 		inspectError    error
 		expectedMsg     string
@@ -56,7 +58,8 @@ func TestValidateExternalNetworks(t *testing.T) {
 			expectedMsg:  "Unexpected",
 		},
 		{
-			network: "host",
+			noInspect: true,
+			network:   "host",
 		},
 		{
 			network:     "user",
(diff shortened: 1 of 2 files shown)


alexmavr:
cc @mavenugo

mavenugo:
LGTM

Not sure why the CI is failing though...

alexmavr:
Amended my commit with no code diff - looks like CI just needed a little nudge

cli/command/stack/deploy_composefile_test.go
```diff
@@ -56,7 +58,8 @@ func TestValidateExternalNetworks(t *testing.T) {
 			expectedMsg:  "Unexpected",
 		},
 		{
-			network: "host",
+			noInspect: true,
+			network:   "host",
```
dnephin: I'm not rally a fan of this storing state in the testcase struct.

I think the same thing can be accomplished by setting `inspectError`, and asserting there was no error, right?

thaJeztah:
This didn't make the cut for 17.07, opened https://github.com/docker/docker-ce/issues/177 to back port it

PR #363, opened 2017-07-20
Update README and add help target to make

jphuynh:
Small update to include shellcheck in the README

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,61 +3,60 @@
 #
 all: binary
 
-# remove build artifacts
+
 .PHONY: clean
-clean:
+clean: ## remove build artifacts
 	rm -rf ./build/* cli/winresources/rsrc_* ./man/man[1-9] docs/yaml/gen
 
-# run go test
-# the "-tags daemon" part is temporary
 .PHONY: test
-test:
+test: ## run go test
 	./scripts/test/unit $(shell go list ./... | grep -v '/vendor/')
 
 .PHONY: test-coverage
-test-coverage:
+test-coverage: ## run test coverage
 	./scripts/test/unit-with-coverage $(shell go list ./... | grep -v '/vendor/')
 
 .PHONY: lint
-lint:
+lint: ## run all the lint tools
 	gometalinter --config gometalinter.json ./...
 
 .PHONY: binary
-binary:
+binary: ## build executable for Linux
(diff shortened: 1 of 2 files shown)


dnephin:
We don't include all the available tasks in the `README` these are just here as an example. 

Maybe shellcheck could be added to lint (`make -f docker.Makefile lint shellcheck`).

We could also add a `CONTRIBUTING` to document it further

thaJeztah:
Perhaps we can add a `make help` to make the Makefile self-documenting, like we have in https://github.com/moby/moby/blob/master/Makefile#L134

jphuynh:
I like the idea @thaJeztah 👍

jphuynh:
I gave it a go. 

@dnephin I'm not sure about the wording for the `watch` target. Let me know if it sounds good.

README.md
```diff
@@ -23,10 +23,10 @@ Build binaries for all supported platforms:
 $ make -f docker.Makefile cross
 ```
 
-Run all linting:
+Run all linting and shellcheck validation:
 
 ```
-$ make -f docker.Makefile lint
+$ make -f docker.Makefile lint shellcheck
```
thaJeztah: We can mention the `help` make target in the readme (to see a list of all make options, type make ......help)

jphuynh: Why didn't think about that? 😅  It makes more sense. Thanks.

README.md
```diff
@@ -23,12 +23,18 @@ Build binaries for all supported platforms:
 $ make -f docker.Makefile cross
 ```
 
-Run all linting:
+Run all linting and shellcheck validation:
```
dnephin: this comment doesn't match the code example anymore

Makefile
```diff
@@ -3,62 +3,61 @@
...
 .PHONY: clean
-clean:
+clean: ## remove build artifacts
 	rm -rf ./build/* cli/winresources/rsrc_* ./man/man[1-9] docs/yaml/gen
 
-# run go test
-# the "-tags daemon" part is temporary
 .PHONY: test
-test:
+test: ## run go test (the "-tags daemon" part is temporary)
```
dnephin: The note about `-tags daemon` is more for the reading of the file, not someone running tasks, can we remove that bit?

jphuynh: Ok

README.md
jphuynh: Oops :)

PR #365, opened 2017-07-20
Fix shellcheck on master and other cleanup

dnephin:
Also fix `shellcheck` failure on master
Also cleanup a test added in #362

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,7 @@
 # Github code owners
 # See https://github.com/blog/2392-introducing-code-owners
 
+cli/command/stack/**        @dnephin @vdemeester
 cli/compose/**              @dnephin @vdemeester
 contrib/completion/bash/**  @albers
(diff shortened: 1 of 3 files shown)


dnephin:
I could make it 3 PRs, but the changes are so small I thought I'd save some time. If there is something controversial in here that you'd like me to split out to discuss separately let me know what it is.

thaJeztah:
> I could make it 3 PRs, but the changes are so small I thought I'd save some time. If there is something controversial in here that you'd like me to split out to discuss separately let me know what it is.

It was mainly that the title of the PR is about the change to add yourself to the code owners, but the biggest part of the PR is fixing tests / completion. I usually try to keep them separate, also in case things need to be cherry-picked separate from each other

dnephin:
That's fair. The master fix is in a separate commit so it can be cherry-picked separately. I can split the other two if necessary.

I fixed the title

PR #366, opened 2017-07-20
Plugable secrets backend

liron-l:
This commit extends SwarmKit secret management with pluggable secret
backends support.
Following previous commits:
1. docker/swarmkit@eebac27434d34708fac993f9f5181d106c5c2fae
2. docker/docker@08f7cf05268782a0dd8e4c41a4cc65fdf78d09f2

Added driver parameter to `docker secret` command.
Specifically:

1. `docker secret create [secret_name] --driver [driver_name]`
2.  Displaying the driver in
```
    $ docker secret ls
    $ docker secret inspect [secret_name]
    $ docker secret inspect [secret_name] -pretty
```

Signed-off-by: Liron Levin <liron@twistlock.com>

### DIFF
--- a/cli/command/secret/create.go
+++ b/cli/command/secret/create.go
@@ -17,6 +17,7 @@ import (
 
 type createOptions struct {
 	name   string
+	driver string
 	file   string
 	labels opts.ListOpts
 }
@@ -27,17 +28,21 @@ func newSecretCreateCommand(dockerCli command.Cli) *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:   "create [OPTIONS] SECRET file|-",
+		Use:   "create [OPTIONS] SECRET [file|-]",
 		Short: "Create a secret from a file or STDIN as content",
-		Args:  cli.ExactArgs(2),
+		Args:  cli.RequiresRangeArgs(1, 2),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.name = args[0]
-			options.file = args[1]
+			if len(args) == 2 {
+				options.file = args[1]
+			}
 			return runSecretCreate(dockerCli, options)
 		},
 	}
 	flags := cmd.Flags()
 	flags.VarP(&options.labels, "label", "l", "Secret labels")
+	flags.StringVarP(&options.driver, "driver", "d", "", "Secret driver")
+	flags.SetAnnotation("driver", "version", []string{"1.31"})
 
 	return cmd
 }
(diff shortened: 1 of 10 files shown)


liron-l:
@dnephin @thaJeztah I closed and re-opend #348 since the tests did not get triggered on push.
I hope it's OK if we continue the discussion here.

liron-l:
@dnephin I think that the commit approval is what's causing the builds to stop.

dnephin:
Hmm, I don't know about that. We approve plenty of PRs that are still able to build afterward. 

I think it's something about the files/changes. On my PR which had this issue (#107) I also retried recreating it, and after the first build it also stopped building again.

liron-l:
Makes sense @dnephin. Let me know what further steps are required to complete this PR.

dnephin:
Once someone else gives their approval we can run the tests manually and merge. Let's not worry about the github bug for now

liron-l:
Great @dnephin thanks!

liron-l:
@thaJeztah PTAL, I separated `vendor.conf` bump to another commit. 
Let me know if I need to bump additional packages. Thanks!

liron-l:
@thaJeztah @cpuguy83 please take a look, I've updated the docker vendor dependencies again. Thanks!

liron-l:
@thaJeztah  PTAL

liron-l:
@vdemeester, @cpuguy83, @diogomonica, @thaJeztah, should I add additional changes to this commit?

cli/command/secret/create.go
```diff
@@ -27,17 +28,20 @@ func newSecretCreateCommand(dockerCli command.Cli) *cobra.Command {
...
-			options.file = args[1]
+			if len(args) == 2 {
+				options.file = args[1]
+			}
 			return runSecretCreate(dockerCli, options)
 		},
 	}
 	flags := cmd.Flags()
 	flags.VarP(&options.labels, "label", "l", "Secret labels")
+	flags.StringVar(&options.driver, "driver", "", "Secret driver")
```
thaJeztah: This flag needs to have a version annotation, so that it is hidden / produces an error on older API versions

thaJeztah: Also, to be consistent with `docker network create`, `docker volume create`, in this case I would be +1 to add a `-d` shorthand as well

cli/command/secret/create.go
```diff
@@ -46,28 +50,26 @@ func runSecretCreate(dockerCli command.Cli, options createOptions) error {
...
-	var in io.Reader = dockerCli.In()
-	if options.file != "-" {
-		file, err := system.OpenSequential(options.file)
-		if err != nil {
-			return err
-		}
-		in = file
-		defer file.Close()
+	if options.driver != "" && options.file != "" {
+		return errors.Errorf("When using secret driver secret data must be empty")
```
thaJeztah: Could this be handled by the external driver? IIUC, with this change the secret's value is taken from the external store, so

```bash
$ docker secret create -d external mysecret
```

Doesn't really "create" a secret, but "associates" the external secret with a secret in Docker.

Do we need to keep the option open to have external secret stores create new secrets from the Docker CLI? i.e.

```bash
$ echo "new secret" | docker secret create -d external mysecret -
```

If so, this validation should be done by the driver (just like currently checking for _empty_ values is done in the backend)

cli/command/secret/create.go
liron-l: Thanks @thaJeztah I set the version annotation to 1.31, hope this makes sense.

cli/command/secret/create.go
liron-l: @thaJeztah according to the secrets plugin subsystem [design](https://docs.google.com/document/d/1kNA6apIFAiaE9d96A2_rcEulVrglvq05zEXGdijSZac), the secrets plugin will be readonly (that is, swarm will not populate the secret values). 
I think it might be better to prevent incorrect usage as soon as possible (since `file` is a cli only option).

liron-l:
Thanks @thaJeztah, I've updated the review according to your comments.
1. Set minimum version for `--secret` parameter (`1.31`)
2. Kept the CLI validation, according to the secret plugin design, we will never populate the secrets from the cli. In the context of plugins, the creation means that we create a real secret in swarmkit (secret name and metadata) and only the value is fetched from the plugin.

liron-l:
@thaJeztah @diogomonica I've updated the code, please take a look if the new secrets driver API makes sense.

liron-l:
@diogomonica @thaJeztah PTAL.

liron-l:
@thaJeztah PTAL, let me know if additional changes are required.

liron-l:
@diogomonica @thaJeztah PTAL, all other dependencies related to this feature are completed. Thanks.

diogomonica:
This LGTM, but I don't have merging rights.

liron-l:
@vdemeester  @dnephin @thaJeztah can somebody please merge?

PR #367, opened 2017-07-20
Introduce/document new IPC modes

kolyshkin:
This builds (and depends) on https://github.com/moby/moby/pull/34087

### DIFF
--- a/docs/reference/run.md
+++ b/docs/reference/run.md
@@ -265,11 +265,21 @@ more advanced use case would be changing the host's hostname from a container.
 
 ## IPC settings (--ipc)
 
-    --ipc=""  : Set the IPC mode for the container,
-                 'container:<name|id>': reuses another container's IPC namespace
-                 'host': use the host's IPC namespace inside the container
+    --ipc="MODE"  : Set the IPC mode for the container
 
-By default, all containers have the IPC namespace enabled.
+The following values are accepted:
+
+| Value                      | Description                                                                       |
+|:---------------------------|:----------------------------------------------------------------------------------|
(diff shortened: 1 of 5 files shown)


kolyshkin:
I guess the tests will fail as it requires vendoring moby/moby

docs/reference/run.md
```diff
@@ -266,6 +266,8 @@ more advanced use case would be changing the host's hostname from a container.
 ## IPC settings (--ipc)
 
     --ipc=""  : Set the IPC mode for the container,
+                 'private': own IPC namespace
```
thaJeztah: This needs the new `none` mode as well

Also we should remove the client-side validation of acceptable options, per https://github.com/moby/moby/pull/34087#pullrequestreview-54514249

> One thing I noticed that has to be fixed is this;

```bash
$ docker run --ipc=none -dit --name bar busybox
docker: --ipc: invalid IPC mode.
```

> Looks like there's validation performed in the CLI; [cli/command/container/opts.go#L424-L427]( https://github.com/docker/cli/blob/edfc89f4de7e602bb06729164ae75ab7a90cbef8/cli/command/container/opts.go#L424-L427).

> We should remove that, as it's the daemon that should validate if a value is valid (important if the CLI version doesn't match the daemon version)

kolyshkin:
This is no longer WIP

kolyshkin:
Anything else I need to do in order to get it merged?

vdemeester:
@kolyshkin was waiting for the moby/moby PR to get merge 😛. It's merge so let's merge this one too 👍

PR #368, opened 2017-07-20
Add zsh completion for `system prune --volumes`

jphuynh:
This is the equivalent of #254 in zsh.


thaJeztah:
Thanks!

ping @sdurrheimer PTAL

PR #371, opened 2017-07-21
Remove duplicate publish option in docker service create/update

jphuynh:
The publish option was showing twice. The fix has been made to be compliant with [docker service create](https://docs.docker.com/engine/reference/commandline/service_create/) and [docker service update](https://docs.docker.com/engine/reference/commandline/service_update/)

screen shot 2017-07-21 at 09 49 08

Note: The whole `create_update` option list is really outdated. I'm planning on doing am update. Would you recommend everything in a single PR or multiple small ones?


thaJeztah:
ping @sdurrheimer PTAL

PR #381, opened 2017-07-26
Fix sort order of options in bash completion

albers:
This is a follow-up on https://github.com/docker/cli/pull/332/files#diff-754420541b12dbcbbea00825942a4635. No functional changes.


thaJeztah:
sorry for that; I thought `-p` after `--password....` was correct (guess it depends on which locale we assume sorting haha)

PR #386, opened 2017-07-26
Fix image resolve detection

thaJeztah:
When using `--resolve-image change`, the image that was resolved during deploy, as well as `Platform` information were overwritten by the image specified in the docker-compose.yaml.  As a result, services were always re-deployed.

fixes https://github.com/moby/moby/issues/34242

**- What I did**

If `QueryRegistry` is "false", preserve the `image` and `Platform` information from the existing service.

**UPDATE:**
removed preserving the platform information, as is discussed in https://github.com/moby/moby/issues/34268

**- How to verify it**

```bash
cat > docker-compose.yml <<"EOF"
version: "3.3"
services:
    redis:
        image: redis:3.2.8
EOF

docker stack deploy -c docker-compose.yml repro \
&& docker service inspect repro_redis -f '{{json .Spec}}' | jq . > before.json \
&& docker stack deploy -c docker-compose.yml --resolve-image changed repro \
&& docker service inspect repro_redis -f '{{json .Spec}}' | jq . > after.json \
&& docker stack rm repro
```

The diff should be empty:

```
git diff --no-index before.json after.json
```

Before this change, the diff would show:

```patch
diff --git a/before.json b/after.json
index f65ecfff..0eecbc51 100644
--- a/before.json
+++ b/after.json
@@ -23,14 +23,7 @@
       "Delay": 5000000000,
       "MaxAttempts": 0
     },
-    "Placement": {
-      "Platforms": [
-        {
-          "Architecture": "amd64",
-          "OS": "linux"
-        }
-      ]
-    },
+    "Placement": {},
     "Networks": [
       {
         "Target": "njp31bnesn79ojar7vj34saqc",
```

ping @aaronlehmann @nishanttotla @dnephin PTAL

### DIFF
--- a/cli/command/stack/deploy_composefile.go
+++ b/cli/command/stack/deploy_composefile.go
@@ -318,10 +318,17 @@ func deployServices(
 
 			updateOpts := types.ServiceUpdateOptions{EncodedRegistryAuth: encodedAuth}
 
-			if resolveImage == resolveImageAlways || (resolveImage == resolveImageChanged && image != service.Spec.Labels[convert.LabelImage]) {
+			switch {
+			case resolveImage == resolveImageAlways || (resolveImage == resolveImageChanged && image != service.Spec.Labels[convert.LabelImage]):
+				// image should be updated by the server using QueryRegistry
 				updateOpts.QueryRegistry = true
+			case image == service.Spec.Labels[convert.LabelImage]:
+				// image has not changed; update the serviceSpec with the
+				// existing information that was set by QueryRegistry on the
+				// previous deploy. Otherwise this will trigger an incorrect
+				// service update.
+				serviceSpec.TaskTemplate.ContainerSpec.Image = service.Spec.TaskTemplate.ContainerSpec.Image
 			}
-
 			response, err := apiClient.ServiceUpdate(
 				ctx,
 				service.ID,
--- a/cli/command/stack/deploy_test.go
+++ b/cli/command/stack/deploy_test.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/docker/cli/cli/compose/convert"
 	"github.com/docker/cli/cli/internal/test"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/swarm"
 	"github.com/stretchr/testify/assert"
 	"golang.org/x/net/context"
 )
@@ -21,4 +23,81 @@ func TestPruneServices(t *testing.T) {
 
 	pruneServices(ctx, dockerCli, namespace, services)
 	assert.Equal(t, buildObjectIDs([]string{objectName("foo", "remove")}), client.removedServices)
+}
+
+// TestServiceUpdateResolveImageChanged tests that the service's
+// image digest is preserved if the image did not change in the compose file
+func TestServiceUpdateResolveImageChanged(t *testing.T) {
+	namespace := convert.NewNamespace("mystack")
+
+	var (
+		receivedOptions types.ServiceUpdateOptions
+		receivedService swarm.ServiceSpec
+	)
+
+	client := test.NewFakeCli(&fakeClient{
+		serviceListFunc: func(options types.ServiceListOptions) ([]swarm.Service, error) {
+			return []swarm.Service{
+				{
+					Spec: swarm.ServiceSpec{
+						Annotations: swarm.Annotations{
+							Name:   namespace.Name() + "_myservice",
+							Labels: map[string]string{"com.docker.stack.image": "foobar:1.2.3"},
+						},
+						TaskTemplate: swarm.TaskSpec{
+							ContainerSpec: swarm.ContainerSpec{
+								Image: "foobar:1.2.3@sha256:deadbeef",
+							},
+						},
+					},
+				},
+			}, nil
+		},
+		serviceUpdateFunc: func(serviceID string, version swarm.Version, service swarm.ServiceSpec, options types.ServiceUpdateOptions) (types.ServiceUpdateResponse, error) {
+			receivedOptions = options
+			receivedService = service
+			return types.ServiceUpdateResponse{}, nil
+		},
+	})
+
+	var testcases = []struct {
+		image                 string
+		expectedQueryRegistry bool
+		expectedImage         string
+	}{
+		// Image not changed
+		{
+			image: "foobar:1.2.3",
+			expectedQueryRegistry: false,
+			expectedImage:         "foobar:1.2.3@sha256:deadbeef",
+		},
+		// Image changed
+		{
+			image: "foobar:1.2.4",
+			expectedQueryRegistry: true,
+			expectedImage:         "foobar:1.2.4",
+		},
+	}
+
+	ctx := context.Background()
+
+	for _, testcase := range testcases {
+		t.Logf("Testing image %q", testcase.image)
+		spec := map[string]swarm.ServiceSpec{
+			"myservice": {
+				TaskTemplate: swarm.TaskSpec{
+					ContainerSpec: swarm.ContainerSpec{
+						Image: testcase.image,
+					},
+				},
+			},
+		}
+		err := deployServices(ctx, client, spec, namespace, false, resolveImageChanged)
+		assert.NoError(t, err)
+		assert.Equal(t, testcase.expectedQueryRegistry, receivedOptions.QueryRegistry)
+		assert.Equal(t, testcase.expectedImage, receivedService.TaskTemplate.ContainerSpec.Image)
+
+		receivedService = swarm.ServiceSpec{}
+		receivedOptions = types.ServiceUpdateOptions{}
+	}
 }
--- a/cli/command/stack/client_test.go
+++ b/cli/command/stack/client_test.go
@@ -34,10 +34,13 @@ type fakeClient struct {
 	nodeListFunc       func(options types.NodeListOptions) ([]swarm.Node, error)
 	taskListFunc       func(options types.TaskListOptions) ([]swarm.Task, error)
 	nodeInspectWithRaw func(ref string) (swarm.Node, []byte, error)
-	serviceRemoveFunc  func(serviceID string) error
-	networkRemoveFunc  func(networkID string) error
-	secretRemoveFunc   func(secretID string) error
-	configRemoveFunc   func(configID string) error
+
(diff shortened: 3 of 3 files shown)


cli/command/stack/deploy_composefile.go
```diff
@@ -321,7 +321,16 @@ func deployServices(
 			if resolveImage == resolveImageAlways || (resolveImage == resolveImageChanged && image != service.Spec.Labels[convert.LabelImage]) {
 				updateOpts.QueryRegistry = true
 			}
-
+			if !updateOpts.QueryRegistry {
+				serviceSpec.TaskTemplate.ContainerSpec.Image = service.Spec.TaskTemplate.ContainerSpec.Image
```
aaronlehmann: What happens if you change the image in the compose file?

cli/command/stack/deploy_test.go
```diff
@@ -22,3 +25,84 @@ func TestPruneServices(t *testing.T) {
...
+		spec := map[string]swarm.ServiceSpec{
+			"myservice": {
+				TaskTemplate: swarm.TaskSpec{
+					ContainerSpec: swarm.ContainerSpec{
+						Image: image,
+					},
+				},
+			},
+		}
+		deployServices(ctx, client, spec, namespace, false, resolveImageChanged)
```
dnephin: This test would continue to pass if `serviceUpdateFunc()` was not called, even if the test was failing. I think we need to adjust the assertions.

`deployServices()` can return an error, so we should `assert.NoError(t, err)` on that error.

Maybe instead of asserting in the fake, the fake could store the arg values, and the assertions could happen here after the call to `deployServices()` ?

cli/command/stack/deploy_composefile.go
```diff
@@ -321,7 +321,16 @@ func deployServices(
...
+
+				if service.Spec.TaskTemplate.Placement != nil {
+					if serviceSpec.TaskTemplate.Placement == nil {
+						serviceSpec.TaskTemplate.Placement = &swarm.Placement{}
+					}
+					serviceSpec.TaskTemplate.Placement.Platforms = service.Spec.TaskTemplate.Placement.Platforms
```
dnephin: This seems like the wrong place to set defaults.

If these values are not set in the compose file where do they come from?

cli/command/stack/deploy_composefile.go
dnephin: I think the intent here would be more clear as an `else` instead of checking the flag set in the previous branch.

An inline comment about why this is necessary would also be great. I'm sure I'm going to forget. Something like this:

> If the image hasn't changed the serviceSpec needs to be updated to use the existing image that was set by `QueryRegistry` on the previous deploy. Otherwise this will trigger an incorrect service update.

dnephin: I believe in that case `image != service.Spec.Labels[convert.LabelImage]` would be false, so we wouldn't enter this block?  I think changing this to an `else` makes that more obvious ( as commented)

aaronlehmann: The label stuff only applies when `resolveImage` is set to `resolveImageChanged`. What if `resolveImage` is set to "never"?

dnephin: True, there needs to be another condition here on the `else if`

thaJeztah: I initially created some code that simply did a compare between the actual image in the spec (stripping digest if the image in the docker-compose file didn't have one), then noticed it's actually using the _label_ to compare

cli/command/stack/deploy_test.go
thaJeztah: Yes, I was indeed looking at that, but didn't have a clean solution, I'll do some more thinking to address that

cli/command/stack/deploy_composefile.go
thaJeztah: :+1:

cli/command/stack/deploy_composefile.go
thaJeztah: This is in the update case; these are the values taken from the current service spec (resolved during initial deploy)

dnephin: Why are they changed by the engine/swarmkit? Shouldn't it only change the image?

cli/command/stack/deploy_composefile.go
dnephin: I'm thinking these two blocks should look something like this (using a `switch/case` instead of an `if/else if`):

```go
switch {
// image should be updated by the server using QueryRegistry
case resolveImage == resolveImageAlways || (resolveImage == resolveImageChanged && image != service.Spec.Labels[convert.LabelImage]):
    updateOpts.QueryRegistry = true
// image hasn't change, so the serviceSpec needs to be updated to use the existing image that was set by QueryRegistry on the previous deploy. Otherwise this will result in an incorrect service update.
case image == service.Spec.Labels[convert.LabelImage]:
    serviceSpec.TaskTemplate.ContainerSpec.Image = service.Spec.TaskTemplate.ContainerSpec.Image
}
```

cli/command/stack/deploy_composefile.go
thaJeztah: These were added based on the image information; i.e. only deploy a Windows image on a Windows host, Arm on Arm etc

But there are already some issues around it, so things may change

dnephin: Does that mean we could leave these out of this PR? I'm not sure it's related to the original bug, but maybe I'm misunderstanding.

Wouldn't these be set in all cases, not just when `QueryRegistry` is enabled?

thaJeztah: They are not set in the docker compose file, and only when QueryRegistry is set; therefore the values that were set during deploy are removed on re-deploy

thaJeztah: I kept the commit separate but if you check the ServiceUpdate function (sorry, typing on phone so can't link easily); you'll see that it's closely related to the issue;

dnephin: I agree it is related, however I don't think this is the correct fix. It is not the responsibility of this code to set default values on update that it never set on create.

This looks like a bug in `ServiceUpdate()` we should not have merged that. That logic is being handled in the wrong place.

thaJeztah: I considered moving this to the ServiceUpdate function but this problem only occurs with stack deploy, because it doesn't "diff" the same way as `docker service update` does. `docker service update` takes the existing service as a starting point, and only makes changes set through the `-add` / `-rm` flags, whereas stack deploy takes the compose-file as source of truth

dnephin: I opened https://github.com/moby/moby/issues/34268. I think we're going to need to fix that first

Once we have a package that is responsible for handling this logic I think at least these lines will move to that package. Possible the switch/case as well

thaJeztah:
@dnephin updated PTAL

cli/command/stack/deploy_composefile.go
```diff
@@ -320,8 +320,20 @@ func deployServices(
 
 			if resolveImage == resolveImageAlways || (resolveImage == resolveImageChanged && image != service.Spec.Labels[convert.LabelImage]) {
 				updateOpts.QueryRegistry = true
+			} else {
```
dnephin: In this case: `resolveImage == resolveImageNever && image !=  service.Spec.Labels[convert.LabelImage]` we would set the image back to the previous image, which is wrong, because the user requested an update to the image.

I think this needs to be:

```go
switch {
// image should be updated by the server using QueryRegistry
case resolveImage == resolveImageAlways || (resolveImage == resolveImageChanged && image != service.Spec.Labels[convert.LabelImage]):
    updateOpts.QueryRegistry = true
// image hasn't changed, so the serviceSpec needs to be updated to use the existing image that was 
// set by QueryRegistry on the previous deploy. Otherwise this will result in an incorrect service update.
case image == service.Spec.Labels[convert.LabelImage]:
    serviceSpec.TaskTemplate.ContainerSpec.Image = service.Spec.TaskTemplate.ContainerSpec.Image
}
```

cli/command/stack/deploy_composefile.go
dnephin: This placement/platforms stuff is a bug in the `client/` package. Let's not hack around it here.

thaJeztah: If we want to fix this for 17.06 I'm not sure we have a good alternative at this moment (without the refactoring of the client package)

cli/command/stack/deploy_composefile.go
thaJeztah: Argh, you're right; changed it to the if/else based on your previous comment. I'm not entirely convinced using a "switch" is much clearer; wether or not to keep the existing information is based on `updateOpts.QueryRegistry ` being `true`, which was why I had it like that in the first implementation

dnephin: I don't think it should be based on `updateOpts.QueryRegistry`.  `updateOpts.QueryRegistry` is set by comparing the existing state to the desired state of the service.  Setting the Image should be based on that same comparison. `updateOpts.QueryRegistry` is an output, not the source of the state.

>  I'm not entirely convinced using a "switch" is much clearer

I find switch/case reads much better than any if/else, but it's true you could also implement this using if/else if

cli/command/stack/deploy_composefile.go
dnephin: If this is only a bug fix for 17.06 let's commit it to only the 17.06 branch in docker-ce and leave the hack out of master. We can open an issue for the correct fix.

thaJeztah:
@dnephin updated, now only preserving the image digest / information

cli/command/stack/deploy_test.go
```diff
@@ -22,3 +24,80 @@ func TestPruneServices(t *testing.T) {
...
+		expectedQueryRegistry bool
+		expectedImage         string
+	}{
+		// Image not changed
+		{
+			image: "foobar:1.2.3",
+			expectedQueryRegistry: false,
+			expectedImage:         "foobar:1.2.3@sha256:deadbeef",
+		},
+		// Image changed
```
dnephin: I think this test covers 2 of the 4 cases. The other 2 are different `resolveImage` values. 

We could add coverage for that in a follow up

PR #387, opened 2017-07-26
Fast fail when save directory does not exist

darrenstahlmsft:
Signed-off-by: Darren Stahl <darst@microsoft.com>

**- What I did**

When using `docker save -o foo/bar.tar`, if the directory foo does not exist, the client currently calls into the daemon to save the image, sends the entire tar stream to the client just to fail to create the file. This problem exists on both Windows and Linux.

This PR adds a fast fail check prior to calling the daemon to ensure the directory exists.

**- How I did it**

Check if the containing directory exists prior to starting the save.

**- How to verify it**

```
PS > .\docker.exe save microsoft/nanoserver -o c:\fakedir\out.tar
unable to save to c:\fakedir\out.tar: directory does not exist
```

Save now fast fails when the directory does not exist.

### DIFF
--- a/cli/command/image/save.go
+++ b/cli/command/image/save.go
@@ -2,6 +2,8 @@ package image
 
 import (
 	"io"
+	"os"
+	"path/filepath"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
@@ -41,6 +43,10 @@ func runSave(dockerCli command.Cli, opts saveOptions) error {
 		return errors.New("cowardly refusing to save to a terminal. Use the -o flag or redirect")
 	}
 
+	if err := validateOutputPath(opts.output); err != nil {
+		return errors.Wrap(err, "failed to save image")
+	}
+
 	responseBody, err := dockerCli.Client().ImageSave(context.Background(), opts.images)
 	if err != nil {
 		return err
@@ -53,4 +59,14 @@ func runSave(dockerCli command.Cli, opts saveOptions) error {
 	}
 
 	return command.CopyToFile(opts.output, responseBody)
+}
+
+func validateOutputPath(path string) error {
+	dir := filepath.Dir(path)
+	if dir != "" && dir != "." {
+		if _, err := os.Stat(dir); os.IsNotExist(err) {
+			return errors.Errorf("unable to validate output path: directory %q does not exist", dir)
+		}
+	}
+	return nil
 }
--- a/cli/command/image/save_test.go
+++ b/cli/command/image/save_test.go
@@ -43,6 +43,11 @@ func TestNewSaveCommandErrors(t *testing.T) {
 				return ioutil.NopCloser(strings.NewReader("")), errors.Errorf("error saving image")
 			},
 		},
+		{
+			name:          "output directory does not exist",
+			args:          []string{"-o", "fakedir/out.tar", "arg1"},
+			expectedError: "failed to save image: unable to validate output path: directory \"fakedir\" does not exist",
+		},
 	}
 	for _, tc := range testCases {
 		cli := test.NewFakeCli(&fakeClient{imageSaveFunc: tc.imageSaveFunc})


cli/command/image/save.go
```diff
@@ -41,6 +44,14 @@ func runSave(dockerCli command.Cli, opts saveOptions) error {
 		return errors.New("cowardly refusing to save to a terminal. Use the -o flag or redirect")
 	}
 
+	dir := filepath.Dir(opts.output)
+	if dir != "" && dir != "." {
+		_, err := os.Stat(dir)
+		if err != nil && os.IsNotExist(err) {
+			return fmt.Errorf("unable to save to %s: directory does not exist", opts.output)
+		}
+	}
```
dnephin: I think this should be a separate function  `func validateOutputPath(...) error`.

darrenstahlmsft:
Updated to add `validateOutputPath` and added a test.

cli/command/image/save.go
```diff
@@ -41,6 +43,11 @@ func runSave(dockerCli command.Cli, opts saveOptions) error {
 		return errors.New("cowardly refusing to save to a terminal. Use the -o flag or redirect")
 	}
 
+	err := validateOutputPath(opts.output)
+	if err != nil {
```
thaJeztah: nit; should probably be

```go
if err := validateOutputPath(opts.output); err != nil {
	return errors.Wrap(err, "failed to save image")
}

File: cli/command/image/save.go [hunk 3]
```diff
@@ -54,3 +61,14 @@ func runSave(dockerCli command.Cli, opts saveOptions) error {
 
 	return command.CopyToFile(opts.output, responseBody)
 }
+
+func validateOutputPath(path string) error {
+	dir := filepath.Dir(path)
+	if dir != "" && dir != "." {
+		_, err := os.Stat(dir)
+		if err != nil && os.IsNotExist(err) {
+			return errors.New("directory does not exist")
```
thaJeztah: Was there a reason for rewriting this error here? (not at my computer, but wondering if the original message contains the _path_ that was not found; which is hidden in this custom message)

cli/command/image/save.go
```diff
@@ -54,3 +61,14 @@ func runSave(dockerCli command.Cli, opts saveOptions) error {
 
 	return command.CopyToFile(opts.output, responseBody)
 }
+
+func validateOutputPath(path string) error {
+	dir := filepath.Dir(path)
+	if dir != "" && dir != "." {
+		_, err := os.Stat(dir)
+		if err != nil && os.IsNotExist(err) {
```
thaJeztah:
```go
if _,err := os.Stat(dir); os.IsNotExist(err) {
    ...
}

File: cli/command/image/save.go (hunk 3 above)
Comment (darrenstahlmsft) [inline cli/command/image/save.go]: A custom error was used because the returned error from `os.Stat` is:
```
GetFileAttributesEx fakedir: The system cannot find the file specified.
```
which is less useful imo (other than showing the stat'd file name).

```
fmt.Errorf("directory %q does not exist", dir)
```
should fix the issue that the dir name is not shown while still being a better error message. WDYT?

File: cli/command/image/save.go (hunk 4 above)
Comment (darrenstahlmsft) [inline cli/command/image/save.go]: Fixed locally, will push changes once error message is decided.

File: cli/command/image/save.go (hunk 2 above)
Comment (darrenstahlmsft) [inline cli/command/image/save.go]: Fixed locally, will push changes once error message is decided.

File: cli/command/image/save.go (hunk 3 above)
Comment (dnephin) [inline cli/command/image/save.go]: I think it should use `errors.Errorf()` (instead of `fmt`) otherwise sounds good to me.

File: cli/command/image/save_test.go [hunk 5]
```diff
@@ -43,6 +43,11 @@ func TestNewSaveCommandErrors(t *testing.T) {
 				return ioutil.NopCloser(strings.NewReader("")), errors.Errorf("error saving image")
 			},
 		},
+		{
+			name:          "output directory does not exist",
+			args:          []string{"-o", "fakedir/out.tar", "arg1"},
+			expectedError: "failed to save image: directory does not exist",
```
dnephin: I guess the error message should also include the word "output", otherwise it may not be clear.

darrenstahlmsft:
Fixed feedback. New error is:
```
failed to save image: unable to validate output path: directory "fakedir" does not exist
```

cli/command/image/save.go
thaJeztah: ah, thanks for explaining @darrenstahlmsft - yes that makes sense.

thaJeztah:
all green

PR #389, opened 2017-07-27
Add zsh completion for `docker service scale --detach=false`

jphuynh:
zsh completion equivalent of #382 for the flag option in #243.

### DIFF
--- a/contrib/completion/zsh/_docker
+++ b/contrib/completion/zsh/_docker
@@ -2040,6 +2040,7 @@ __docker_service_subcommand() {
         (scale)
             _arguments $(__docker_arguments) \
                 $opts_help \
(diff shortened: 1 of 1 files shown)


contrib/completion/zsh/_docker
```diff
@@ -2040,6 +2040,7 @@ __docker_service_subcommand() {
         (scale)
             _arguments $(__docker_arguments) \
                 $opts_help \
+                "($help -d --detach)"{-d=false,--detach=false}"[Disable detached mode]" \
```
sdurrheimer: Boolean flag doesn't need `=false`.

Sorry for the delay, I have not a lot of time these days.

jphuynh: Thanks I'll change that.

PR #390, opened 2017-07-27
Add test for github.com special handling

thaJeztah:
Found this code still lingering around on my computer :smile:

```
Add test for github.com special handling
```

pika-pixdaus

(image taken from https://sustainableutah.files.wordpress.com/2015/03/pika-pixdaus.jpg)


cli/command/image/build_test.go
```diff
@@ -68,3 +68,24 @@ func TestRunBuildDockerfileFromStdinWithCompress(t *testing.T) {
...
+	assert.Contains(t, err.Error(), "unable to prepare context: unable to 'git clone'")
+
+	tmpDir, err := ioutil.TempDir("", "docker-build-from-local-dir-")
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	os.Chdir(tmpDir)
+	os.MkdirAll(filepath.Join(tmpDir, "github.com/docker/no-such-repository"), 0777)
+	err = cmd.Execute()
+	assert.Contains(t, err.Error(), "unable to prepare context: unable to evaluate symlinks in Dockerfile path:")
```
dnephin: Since the expectation here is "does build from local directory" maybe this would be more clear by adding a simple `Dockerfile` to the directory and checking for no error?

cli/command/image/build_test.go
```diff
@@ -68,3 +68,24 @@ func TestRunBuildDockerfileFromStdinWithCompress(t *testing.T) {
 	sort.Strings(actual)
 	assert.Equal(t, []string{dockerfileName, ".dockerignore", "foo"}, actual)
 }
+
+// TestRunBuildFromLocalGitHubDirNonExistingRepo tests that build contexts
+// starting with `github.com/` are special-cased, and the build command attempts
+// to clone the remote repo, however a local directory should take precedence.
+func TestRunBuildFromLocalGitHubDirNonExistingRepo(t *testing.T) {
```
dnephin: I think this should be split into two test cases

thaJeztah: Was doubting; in a single test, it more "proves" that without local directory it does X, and with it does Y, but guess it doesn't matter if split; will update

cli/command/image/build_test.go
thaJeztah: Agreed, I'll look into that

thaJeztah:
ping @vdemeester PTAL 😅

thaJeztah:
@vdemeester good call; do we have a `requires(network)` or similar constraint for tests?

Issue #159, opened 2017-06-06
Missing documentation for `--config*` flags for `docker service ...`


jlhawn:
The current repo doesn't document how to use the `--config` flag when running `docker service create ...`   nor does it document how to use either the `--config-rm` or `--config-add` flags when running `docker service update ...`.

Links:

- [Documentation for `docker service create`](https://github.com/docker/cli/blob/master/docs/reference/commandline/service_create.md)
- [Documentation for `docker service update`](https://github.com/docker/cli/blob/master/docs/reference/commandline/service_update.md)

thaJeztah:
The flags were added to the reference docs in https://github.com/docker/cli/pull/403, and this section of the documentation handles using configs; https://docs.docker.com/engine/swarm/configs/

Closing this one, but some additional information in the command line reference docs could still be needed

PR #403, opened 2017-07-31
update service create and update options in commandline documentation

Zebrilee:
Fix issue #159 
Signed-off-by: zebrilee <zebrilee@gmail.com>

**- What I did**
`--config`was not up to date in service_create.md and service_update.md so I checked if it was the only one and updated all options' part for both of those files

PS : the actual PR #205 will also modify those files by adding a `[service rollback](service_rollback.md)`at the end

**- How I did it**
Copy paste `docker service update --help` and `docker service create --help` results in their own .md files 

**- How to verify it**
The documentation and the result of the commandline `docker service update` and `docker service create --help` must be the same

(did a `make -f docker.Makefile` too)

Update the docker service create and update commandline documentation 

### DIFF
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -21,9 +21,10 @@ Usage:  docker service create [OPTIONS] IMAGE [COMMAND] [ARG...]
 Create a new service
 
 Options:
+      --config config                      Specify configurations to expose to the service
       --constraint list                    Placement constraints
       --container-label list               Container labels
-      --credential-spec                    Credential spec for managed service account (Windows only)
+      --credential-spec credential-spec    Credential spec for managed service account (Windows only)
   -d, --detach                             Exit immediately instead of waiting for the service to converge (default true)
       --dns list                           Set custom DNS servers
       --dns-option list                    Set DNS options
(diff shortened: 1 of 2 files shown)


docs/reference/commandline/service_create.md
```diff
@@ -21,9 +21,10 @@ Usage:  docker service create [OPTIONS] IMAGE [COMMAND] [ARG...]
 Create a new service
 
 Options:
+      --config config                      Specify configurations to expose to the service
       --constraint list                    Placement constraints
       --container-label list               Container labels
-      --credential-spec                    Credential spec for managed service account (Windows only)
+      --credential-spec credential-spec    Credential spec for managed service account (Windows only)
```
mstanleyjones: Make sure these flags are updated in the Go code as well, since our docs get this straight from the code.

Zebrilee:
Hi @mstanleyjones ,

If I'm correct it is implemented [here](https://github.com/docker/cli/blob/eabdace6282d64ff9b999aa5746e138ff806a174/contrib/completion/bash/docker) 
and also in the [docker compose](https://docs.docker.com/compose/compose-file/#credential_spec) since 17.06

docs/reference/commandline/service_update.md
```diff
@@ -80,7 +83,7 @@ Options:
       --rollback-failure-action string     Action on rollback failure ("pause"|"continue")
       --rollback-max-failure-ratio float   Failure rate to tolerate during a rollback
       --rollback-monitor duration          Duration after each task rollback to monitor for failure (ns|us|ms|s|m|h)
-      --rollback-order string              Rollback order ("start-first"|"stop-first") (default "stop-first")
```
thaJeztah: I wondered why the default was removed, but for `service update` there's no default value, so 👍

docs/reference/commandline/service_create.md
thaJeztah: I just verified by comparing with the usage output on master, and this PR matches that output

PR #406, opened 2017-08-01
adding network list test

ksouf:
Signed-off-by: khaled souf <khaled.souf@gmail.com>

**- What I did**
adding  network list test 
**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/network/list_test.go
+++ b/cli/command/network/list_test.go
@@ -1,0 +1,69 @@
+package network
+
+import (
+	"testing"
+
+	"io/ioutil"
+
+	"strings"
+
+	"github.com/docker/cli/internal/test"
+	. "github.com/docker/cli/internal/test/builders"
+	"github.com/docker/cli/internal/test/testutil"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/gotestyourself/gotestyourself/golden"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/net/context"
+)
+
+func TestNetworkListErrors(t *testing.T) {
+	testCases := []struct {
+		networkListFunc func(ctx context.Context, options types.NetworkListOptions) ([]types.NetworkResource, error)
+		expectedError   string
+	}{
+		{
+			networkListFunc: func(ctx context.Context, options types.NetworkListOptions) ([]types.NetworkResource, error) {
+				return []types.NetworkResource{}, errors.Errorf("error creating network")
+			},
+			expectedError: "error creating network",
+		},
+	}
+
+	for _, tc := range testCases {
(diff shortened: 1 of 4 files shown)


cli/command/network/list_test.go
```diff
@@ -0,0 +1,61 @@
...
+	filterArgs.Add("image.name", "ubuntu")
+
+	expectedOpts := types.NetworkListOptions{
+		filterArgs,
+	}
+
+	cli := test.NewFakeCli(&fakeClient{
+		networkListFunc: func(ctx context.Context, options types.NetworkListOptions) ([]types.NetworkResource, error) {
+			assert.Equal(t, expectedOpts, options, "not expected options error")
+			return []types.NetworkResource{}, nil
```
dnephin: We also print out the returned ID, so we should include an ID here, and check that the ID is printed

cli/internal/test/builders/network.go
```diff
@@ -0,0 +1,72 @@
...
+		Scope:      "default scope",
+		Created:    time.Time{},
+		EnableIPv6: false,
+		Attachable: false,
+		Ingress:    false,
+		ConfigOnly: false,
+		Internal:   false,
+		IPAM:       network.IPAM{},
+		ConfigFrom: network.ConfigReference{},
+		Containers: map[string]types.EndpointResource{"toto": {
```
vdemeester: This one could be extract in another *builder* function 👼

dnephin: I believe many of these fields are already the default "zero" value, so you don't have to include them here either (false, network.IPAM{},etc)

cli/internal/test/builders/network.go
```diff
@@ -0,0 +1,45 @@
+package builders
```
dnephin: This package has moved to `internal/test/builders`, please `git mv` this file

cli/command/network/list_test.go
```diff
@@ -0,0 +1,69 @@
...
+				NetworkResourceDriver("09.7.01"),
+				NetworkResourceScope("global"))}, nil
+		},
+	})
+	cmd := newListCommand(cli)
+
+	cmd.Flags().Set("filter", "image.name=ubuntu")
+	assert.NoError(t, cmd.Execute())
+	assert.Equal(t, "NETWORK ID          NAME                DRIVER              SCOPE\n"+
+		"123454321           network_1           09.7.01             global", strings.TrimSpace(cli.OutBuffer().String()))
```
dnephin: Actually, you might want to use `golden.Assert()` here to make it easier to update in the future. There are a few examples in the repo

ksouf:
@dnephin @vdemeester there is something weird with jenkins build could you check it please :)

dnephin:
looks like jenkins failed to send the build status to github, but the tests passed, so it's all good

PR #413, opened 2017-08-02
Add bash completion for `docker build --target`

albers:
This adds bash completion for https://github.com/moby/moby/pull/32496.

`--target` is completed with targets that are extracted from the dockerfile.

This does not always work, though. According to the syntax of `docker build`,

    Usage:  docker build [OPTIONS] PATH | URL | -

`PATH` should be the last argument and therefore won't be available for calculation of the dockerfile location during completion of `--target`.

But `docker build . --target` also works, so `--file` and `PATH` are honored if they are specified before `--target`. As the default value for the dockerfile `./Dockerfile` is assumed.

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -2439,6 +2439,7 @@ _docker_image_build() {
 		--network
 		--shm-size
 		--tag -t
+		--target
 		--ulimit
 	"
 	__docker_daemon_os_is windows && options_with_args+="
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @tonistiigi @tianon PTAL

contrib/completion/bash/docker
```diff
@@ -2505,6 +2506,19 @@ _docker_image_build() {
...
+			local context_pos=$( __docker_pos_first_nonflag "$( __docker_to_alternatives "$options_with_args" )" )
+			local context="${words[$context_pos]}"
+			context="${context:-.}"
+
+			local file="$( __docker_value_of_option '--file|f' )"
+			local default_file="${context%/}/Dockerfile"
+			local dockerfile="${file:-$default_file}"
+
+			local targets="$( sed -n 's/^FROM .\+ AS \(.\+\)/\1/p' "$dockerfile" 2>/dev/null )"
+			COMPREPLY=( $( compgen -W "$targets" -- "$cur" ) )
```
tianon: `--target` would also accept stage numbers, right?  Should we also scan the `Dockerfile` for how many `FROM` instructions there are and suggest `0-(n-1)` as valid inputs?

tonistiigi: Stage name is preferred. I don't think we need to advertise for the index numbers.

tianon: Ok, fair enough! :+1:

PR #415, opened 2017-08-02
Enable TCP Keep-Alive in Docker client

jlhawn:
Some network environments may have NATs, proxies, or gateways which
kill idle connections. There are many Docker API operations which may
be idle for long periods of time (such as ContainerWait and ContainerAttach)
and may result in unexpected connection closures or hangs if TCP keepalives
are not used.

This patch updates the default HTTP transport used by the Docker client
package to enable TCP Keep-Alive with a keep-alive interval of 30 seconds.
It also sets a connect timeout of 30 seconds.

This PR complements the change to the moby client package here: https://github.com/moby/moby/pull/34359

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -2,9 +2,11 @@ package command
 
 import (
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"runtime"
+	"time"
 
 	"github.com/docker/cli/cli"
 	cliconfig "github.com/docker/cli/cli/config"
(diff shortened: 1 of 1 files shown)


thaJeztah:
ugh circleCI keeps having problems, now its a memory issue (possibly);

```
./scripts/build/cross
Exited with code 137

Hint: Exit code 137 typyically means the process is killed because it was running out of memory
Hint: Check if you can optimize the memory usage in your app
Hint: Max memory usage of this container is 58572800
 according to /sys/fs/cgroup/memory/memory.max_usage_in_bytes
```

cli/command/cli.go
```diff
@@ -214,6 +216,10 @@ func newHTTPClient(host string, tlsOptions *tlsconfig.Options) (*http.Client, er
 	}
 	tr := &http.Transport{
 		TLSClientConfig: config,
+		DialContext: (&net.Dialer{
+			KeepAlive: 30 * time.Second,
```
andrewhsu: For now this can help keep connections alive over the NAT timeout issue we've been facing, but would really like to see more knobs to turn in a configurable manner much like curl: https://curl.haxx.se/libcurl/c/CURLOPT_TCP_KEEPALIVE.html

PR #418, opened 2017-08-04
docker image rm -f should exit with 1 if the image was not removed

### LINKED ISSUE #394: `docker image rm -f` should exit with 1 if the image was not removed
This was changed in #160, but there are some problems with this new behaviour.

`rm -f /some/file` still exits with 1 if you don't have permission to remove it.
`docker image rm -f` should exit with 1 if the image was not removed.

adshmh:
`docker image rm -f` should exit with error code 1 if the image was not removed. Fixes #394 

**- What I did**
Added logic that makes 'docker image rm -f <image>' exit with 1 instead of 0 when an image cannot be removed (e.g. because running containers are using the image)

**- How I did it**
- Added a new type of error (imageRemoveConflict) to docker/docker/client/errors.go, and converted the 409 status code to the new error in docker/docker/client/image_remove.go.
- The image rm command uses the new type of error to exit.

**- How to verify it**
- Run 'docker image rm -f <image-id>' for an image used by running container(s).
- On the master branch, the exit code will be 0:
> $ docker image rm -f 7328f6f8b418
> Error response from daemon: conflict: unable to delete 7328f6f8b418 (cannot be forced) - image is being used by running container 36ea0cc00015
> $ echo $?
>0

- With this PR, the exit code will be 1:
> $ docker image rm -f 7328f6f8b418
Error response from daemon: conflict: unable to delete 7328f6f8b418 (cannot be forced) - image is being used by running container 36ea0cc00015
> $ echo $?
>1

`docker image rm -f` exits with 1 if the image is not removed.


adshmh:
I edited the files under vendor directory so all changes are seen together in the PR. If the changes are approved, I will submit a PR against docker/docker (with the addition of tests for new type of error and the API client).

cli/command/image/remove.go
```diff
@@ -56,9 +57,13 @@ func runRemove(dockerCli command.Cli, opts removeOptions, images []string) error
 	}
 
 	var errs []string
+	var removeConflict = false
 	for _, img := range images {
 		dels, err := client.ImageRemove(ctx, img, options)
 		if err != nil {
+			if apiclient.IsErrImageRemoveConflict(err) {
+				removeConflict = true
+			}
```
dnephin: I think we should invert this. Any `NotFound` can be skipped in the force case, but any other errors should still be reported as errors.

adshmh: Thank you for the review. I just updated the PR. Two of the unit tests also needed to be changed, as with the new logic, the -f option can only skip imageNotFound errors.

cli/command/image/remove_test.go
```diff
@@ -68,15 +77,6 @@ func TestNewRemoveCommandSuccess(t *testing.T) {
 			},
 		},
 		{
-			name: "Image Deleted with force option",
```
dnephin: I guess we should still have a case here "Image not found with force option"

cli/command/image/remove.go
```diff
@@ -56,9 +57,13 @@ func runRemove(dockerCli command.Cli, opts removeOptions, images []string) error
 	}
 
 	var errs []string
+	var unresolvableErr = false
 	for _, img := range images {
 		dels, err := client.ImageRemove(ctx, img, options)
 		if err != nil {
+			if !apiclient.IsErrImageNotFound(err) {
```
dnephin: minor, but these object specific error checkers should be deprecated. Please use `IsErrNotFound()` instead. `IsErrImageNotfound()` just calls that directly anyway.

cli/command/image/remove.go
```diff
@@ -56,9 +57,13 @@ func runRemove(dockerCli command.Cli, opts removeOptions, images []string) error
 	}
 
 	var errs []string
+	var unresolvableErr = false
```
dnephin: minor: I'm not sure about the term "resolvable" here. Maybe "unrecoverable"  or just "fatal" ?

cli/command/image/remove_test.go
adshmh: Thank you for the review. I added a test case. 
Would it be a good idea for the docker/docker/client package to export the notFound* errors to make testing easier? I think in addition to the new test, _stack/deploy_composefile_test.go_ also implements a type of notFound error for testing.

dnephin: Possibly. The problem with exporting them is that other packages could start to use them to report errors, which is not really the intent.

dnephin:
Thanks, looks good! Please ping me on the moby/moby PR to apply this fix to `client/`

dnephin:
@adshmh #424 has merged, so you should be able to rebase now and the vendor changes you need may already be in vendor/

adshmh:
@dnephin thank you for the reminder, it seems the changes are now in vendor/. I will rebase.

thaJeztah:
ping @adshmh could you rebase? I think docker 17.11-rc1 is scheduled to be cut off today; would be nice to get this in :+1:

dnephin:
I'll carry this

thaJeztah:
@dnephin noticed this one as well (probably related); https://github.com/docker/cli/pull/567

dnephin:
Yes, carrying this will also close #567

PR #419, opened 2017-08-05
Add "build" dir to PATH in dev-container

thaJeztah:
This makes running the client easier inside
the container; allowing to use just `docker`
instead of `build/docker`.

**- What I did**

**- How I did it**

**- How to verify it**

```bash
$ make -f docker.Makefile binary shell
$ docker version
```


thaJeztah:
ping @vdemeester 😄

PR #420, opened 2017-08-06
Added "scope" to docker-cli compose

glyif:
**- What I did**
I updated the docker-cli's compose to be able to take in scope as a network key with a value of swarm or local.

**- How I did it**
I updated the config_schema, added the scope to the Network Config struct, generated new bindata, edited tests to test for the scope.

**- How to verify it**
This currently only works with the --compose-file flag in the docker-cli, but doesn't work with docker-compose yet. I will be working on that next

I updated the config_schema, added the scope to the Network Config struct, generated new bindata, edited tests to test for the scope.

My dogs, Hex the Husky (left) and Linux the Aussie (right).


vdemeester (dismissed):
LGTM 🐸
/cc @vieux @dnephin same question as before, Is 17.08 freeze in place ? if yes, the schema modification should target `3.5` instead of 3.4.

dnephin:
afaik we're still good for 3.4

thaJeztah:
@glyif looks like this needs a rebase

@dnephin @vdemeester is this ready to go other than a rebase?

vdemeester:
@thaJeztah @dnephin I think we missed the `3.4` window :sweat_smile: but other than that we are good to go :angel:

dnephin:
I think so, this will need to be moved to the 3.5 schema since 3.4 has been released, but we can do that as a follow up if necessary

vdemeester:
@dnephin @thaJeztah Added a 3.7 schema version (as 3.6 is already on 18.02) and rebased this PR 👼

thaJeztah (dismissed):
LGTM

vdemeester:
cc @shin- for `docker/compose` 👼

dnephin:
Hmm, thinking about this again. Are we sure this is something we want to expose in the Compose file?

If someone uses `scope: local` will the services actually be able to connect to it? I guess only if the service is on the same node?

thaJeztah:
moving back to design review

ping @fcrisciani WDYT?

fcrisciani:
@thaJeztah host, none, bridge are all only local networks, overlay is only swarm scope and macvlan can be both, but in the case of swarm scope a config is going to be needed for each node. Not super clear to me how these possible scenarios are actually handled with a simple compose

dankolesnikov:
Will this be merged anytime soon? Thx!

unixfox:
Would love to see this PR merged!

nbx3:
Any chance of this being merged?

PR #421, opened 2017-08-06
Refactor/cleanup of docker info

thaJeztah:
Some mild refactoring of the docker info command;

- Use `fmt.Fprinln()` instead of `fmt.Fprintf()` where possible
- Rename `fprintfIfNotEmpty()` to `fprintNonEmpty()`, and removed
  return variables, because they were not used. `fprintNonEmpty()`
  now uses fmt.Fprintln()` instead of `fmt.Fprintf()`, because
  formatting was not used.
- Use `fprintNonEmpty()` to get rid of some `if` statements
- Extract printing Swarm-related information to a `printSwarmInfo()` function
- Change `Http Proxy` and `Https Proxy` to `HTTP Proxy` / `HTTPS Proxy`

ping @vdemeester @dnephin

### DIFF
--- a/cli/command/system/info.go
+++ b/cli/command/system/info.go
@@ -55,109 +55,50 @@ func runInfo(dockerCli *command.DockerCli, opts *infoOptions) error {
 
 // nolint: gocyclo
 func prettyPrintInfo(dockerCli command.Cli, info types.Info) error {
-	fmt.Fprintf(dockerCli.Out(), "Containers: %d\n", info.Containers)
-	fmt.Fprintf(dockerCli.Out(), " Running: %d\n", info.ContainersRunning)
-	fmt.Fprintf(dockerCli.Out(), " Paused: %d\n", info.ContainersPaused)
-	fmt.Fprintf(dockerCli.Out(), " Stopped: %d\n", info.ContainersStopped)
-	fmt.Fprintf(dockerCli.Out(), "Images: %d\n", info.Images)
-	fprintfIfNotEmpty(dockerCli.Out(), "Server Version: %s\n", info.ServerVersion)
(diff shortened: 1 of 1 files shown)


cli/command/system/info.go
```diff
@@ -373,9 +366,8 @@ func formatInfo(dockerCli *command.DockerCli, info types.Info, format string) er
 	return err
 }
 
-func fprintfIfNotEmpty(w io.Writer, format, value string) (int, error) {
+func fprintNonEmpty(w io.Writer, label, value string) {
```
stevvooe: This should be `fprintlnNonEmpty` to make it clear that this is backed by `fmt.Fprintln`, which has slightly different behavior for spacing from `fmt.Fprint`.

stevvooe:
LGTM on design/approach/intent.

This will impact output stability slightly, if people depend on it. Some `%s` and `%v` may have different output depending on the implementation of the type when passed to `fmt.Fprintln`.

thaJeztah:
> Some %s and %v may have different output depending on the implementation of the type when passed to fmt.Fprintln.

Good point; I looked at the information we're printing, and only `bool`, `string` and`int` values were printed, so I considered those safe for what we're currently printing

> is should be `fprintlnNonEmpty` to make it clear that this is backed by `fmt.Fprintln`, which has slightly different behavior for spacing from `fmt.Fprint`.

Yup let me rename, I was doubting between an even more generating `outputIfSet()` and the "old" name, but now it's neither of those, :joy:

thaJeztah:
updated 👍

thaJeztah:
rebased because there was a conflict with https://github.com/docker/cli/commit/3428b78e966954741325061800919984fb7257e9

PR #422, opened 2017-08-06
Output a warning when running in a two-manager setup

thaJeztah:
Running a cluster in a two-manager configuration effectively *doubles* the chance of loosing control over the cluster (compared to running in a single-manager setup). Users may have the assumption that having two managers provides fault tolerance, so it's best to warn them if they're using this configuration.

This patch adds a warning to `docker info` if Swarm is configured with two managers:

    WARNING: Running Swarm in a two-manager configuration. This configuration provides
             no fault tolerance, and poses a high risk to loose control over the cluster.
             Refer to https://docs.docker.com/engine/swarm/admin_guide/ to configure the
             Swarm for fault-tolerance.

**- How to verify it**

run swarm in a two manager setup, and run `docker info`

```
Output a warning when running in a two-manager setup
```

panda-twins-cute-animals-pictures-pics

### DIFF
--- a/cli/command/system/info.go
+++ b/cli/command/system/info.go
@@ -221,6 +221,15 @@ func prettyPrintInfo(dockerCli command.Cli, info types.Info) error {
 	fmt.Fprintln(dockerCli.Out(), "Live Restore Enabled:", info.LiveRestoreEnabled)
 	fmt.Fprint(dockerCli.Out(), "\n")
 
+	if getManagerCount(info.Swarm) == 2 {
+		msg := "WARNING: Running Swarm in a two-manager configuration. This configuration provides \n"
+		msg += "         no fault tolerance, and poses a high risk to loose control over the cluster.\n"
+		msg += "         Refer to https://docs.docker.com/engine/swarm/admin_guide/ to configure the\n"
+		msg += "         Swarm for fault-tolerance."
+
+		fmt.Fprintln(dockerCli.Err(), msg)
+	}
+
 	// Only output these warnings if the server does not support these features
 	if info.OSType != "windows" {
 		printStorageDriverWarnings(dockerCli, info)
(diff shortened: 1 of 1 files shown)


dnephin:
I wonder if we should print this warning on `swarm join`? I guess it's fine on `info` as well, but it seems like it would be less visible there.

thaJeztah:
Thought about `join` and `leave` / `demote`, but wondered if it would be misleading in cases where you'd go from `1 -> 2 -> 3` managers

thaJeztah:
@dnephin opened an alternative that outputs a warning on;

- `docker info`
- `docker swarm join`
- `docker node demote`
- `docker node promote`

But not sure we need to do for all of those (also, the extra `/info` call for the `promote/demote` case may cause overhead, as all we need is the swarm info); https://github.com/docker/cli/compare/master...thaJeztah:warn-manager-count-alternative

vdemeester:
ping @thaJeztah @dnephin @nishanttotla what is the status here ?

cli/command/system/info.go
```diff
@@ -221,6 +221,15 @@ func prettyPrintInfo(dockerCli command.Cli, info types.Info) error {
 	fmt.Fprintln(dockerCli.Out(), "Live Restore Enabled:", info.LiveRestoreEnabled)
 	fmt.Fprint(dockerCli.Out(), "\n")
 
+	if getManagerCount(info.Swarm) == 2 {
+		msg := "WARNING: Running Swarm in a two-manager configuration. This configuration provides \n"
+		msg += "         no fault tolerance, and poses a high risk to loose control over the cluster.\n"
+		msg += "         Refer to https://docs.docker.com/engine/swarm/admin_guide/ to configure the\n"
+		msg += "         Swarm for fault-tolerance."
+
```
dnephin: Nit, maybe use a raw string for this so it's easier to read?

thaJeztah:
daemon-side implementation; https://github.com/moby/moby/pull/38891

silvin-lubecki:
This PR needs to be rebased 🦁

thaJeztah:
@silvin-lubecki I should probably close this one if we agree that https://github.com/moby/moby/pull/38891 is a better approach

thaJeztah:
closing, as https://github.com/moby/moby/pull/38891 was merged

Issue #414, opened 2017-08-02
Add support for "isolaton" in Compose files


Techwolfy:
The Compose v3 format includes a parameter to specify the type of [container isolation](https://docs.docker.com/engine/reference/commandline/run/#specify-isolation-technology-for-container-isolation) that the host should use.  This is always `default` on Linux, but can be either `process` or `hyperv` on Windows. This setting was added in v2.1 and has remained in the documentation since.

However, it was left out of the v3+ schema definition, so it currently can't be used in newer Compose files despite working correctly if the file version is downgraded. I've submitted a patch in docker/compose#5075 to fix this for local deployments. Swarm has never had support for this feature though so it will likely be a bit more complicated to fix.

Currently, running `docker stack deploy -c docker-compose.yml swarm` results in the error message `isolation Additional property isolation is not allowed`. I'm working on a Windows-based Swarm application and not all of the servers share an OS version so this feature would be quite helpful to me.

thaJeztah:
For reference; this depends on https://github.com/moby/moby/issues/31616 to be decided on (and implemented) first

PR #424, opened 2017-08-07
updated vendoring

simonferquel:
**- What I did**
Updated dependencies so that we can use the new sirupsen import path
**- How I did it**
Had to fork notary
**- How to verify it**
See vendor.conf

PR #426, opened 2017-08-07
--isolation for setting swarm service isolation mode

simonferquel:
fixes https://github.com/docker/cli/issues/414

**- What I did**
Added --isolation to service create/update and to compose files, to bypass default isolation mode on the host.
**- How I did it**
Updated dependencies to moby/moby and swarmkit (also had to depend on the logrus mega PR), added the flag in the service update/create cmd, passing the value accordingly
Updated the compose service struct and conversion to ServiceSpec accordingly
**- How to verify it**
Tests incoming
- Service creation and update supports --isolation flag (default, process, hyperv) for bypassing default isolation mode on the host
- Isolation can also be set in compose files using the "isolation" property

~~Depends on https://github.com/docker/cli/pull/424~~


vdemeester:
@simonferquel can you link PRs in swarmkit and moby related to this feature ? 👼

simonferquel:
PR on Swarmkit: https://github.com/docker/swarmkit/pull/2342
PR on Moby: https://github.com/moby/moby/pull/34424

thaJeztah:
https://github.com/moby/moby/pull/34424 was merged, so this PR can be worked on again; @simonferquel can you do a rebase/revendor?

simonferquel:
Requires https://github.com/moby/moby/pull/35382. in the mean time i'll add tests to improve coverage (mainly on update)

cli/command/service/opts_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
```
dnephin: nit: This extra newline can be removed

cli/command/service/update.go
```diff
@@ -269,6 +269,19 @@ func updateService(ctx context.Context, apiClient client.NetworkAPIClient, flags
...
 	}
 
+	updateIsolation := func(flag string, field *container.Isolation) error {
+		if flags.Changed(flag) {
+			o := isolationOpts{}
+			val := flags.Lookup(flag).Value.String()
+			err := o.Set(val)
+			if err != nil {
+				return err
+			}
```
dnephin: You don't need to create a new `isolationOpts` struct here. `Value` is that struct, you just have to cast it (like the functions above).

cli/command/service/update_test.go
```diff
@@ -518,3 +518,23 @@ func TestUpdateStopSignal(t *testing.T) {
...
+		TaskTemplate: swarm.TaskSpec{
+			ContainerSpec: &swarm.ContainerSpec{},
+		},
+	}
+	err = updateService(context.Background(), nil, flags, &spec)
+	require.NoError(t, err)
+	assert.Equal(t, container.IsolationProcess, spec.TaskTemplate.ContainerSpec.Isolation)
+}
+
+func TestUpdateIsolationInvalid(t *testing.T) {
```
dnephin: I think this test needs to call `updateService()` otherwise it's only testing `isolationOpt`

cli/compose/convert/service.go
```diff
@@ -122,6 +122,11 @@ func Service(
 		}
 	}
 
+	isolation := container.Isolation(service.Isolation)
+	if !isolation.IsDefault() && !isolation.IsHyperV() && !isolation.IsProcess() {
+		return swarm.ServiceSpec{}, fmt.Errorf("Unsupported isolation %s. Valid values are %s, %s, %s", isolation, container.IsolationDefault, container.IsolationProcess, container.IsolationHyperV)
```
dnephin: This is already validated by the `isolationOpt` we shouldn't need to do it again here, right?

The validation should be handled by the flags/value types so that any value stored in the opt is valid to send to the API.

cli/compose/convert/service_test.go
```diff
@@ -372,3 +372,14 @@ func TestConvertUpdateConfigOrder(t *testing.T) {
 	})
 	assert.Equal(t, updateConfig.Order, "stop-first")
 }
+
+func TestIsolation(t *testing.T) {
```
dnephin: `s/TestIsolation/TestServiceConvertsIsolation/`

cli/compose/convert/service_test.go
```diff
@@ -372,3 +372,14 @@ func TestConvertUpdateConfigOrder(t *testing.T) {
...
+	src := composetypes.ServiceConfig{
+		Isolation: "hyperv",
+	}
+	result, err := Service("1.32", Namespace{name: "foo"}, src, nil, nil, nil, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
```
dnephin: `require.NoError(t, err)` ?

cli/command/service/update_test.go
simonferquel: Ok, makes sense, I'll remove the test as opts is already tested and updateService won't be called if flags parsing fails

cli/compose/convert/service.go
simonferquel: In this case, the source is not an isolationOpt flag, but a compose-file field (which is of type string)

simonferquel: Did not see that, thank! I changed validation by typing the isolation field in service type to container.Isolation, and introduced a transformValidation function at loading time (that does the validation). Also added a parsing test for invalid isolation.

thaJeztah:
This will also need;

- some changes to the command-line reference docs (documenting the new options for `docker service create` and `docker service update`)
- updates in the bash/zsh completion scripts
- updates to the docker-compose file reference (https://github.com/docker/docker.github.io/blob/master/compose/compose-file/index.md)

thaJeztah:
Also opened a moby bump in https://github.com/docker/cli/pull/679, which should bring in the required changes

simonferquel:
@thaJeztah sorry for the delay, it has been quite a busy time lately...
I will implement your feedback today.

cli/command/service/opts.go
```diff
@@ -457,6 +457,36 @@ func convertExtraHostsToSwarmHosts(extraHosts []string) []string {
...
+}
+
+func (o *isolationOpts) Set(value string) error {
+	o.source = value
+	tv := container.Isolation(strings.ToLower(value))
+	if tv.IsDefault() || tv.IsHyperV() || tv.IsProcess() {
+		o.value = &tv
+		return nil
+	}
+	return fmt.Errorf("Unknown isolation mode %s. Valid values are %v, %v, %v", value, container.IsolationDefault, container.IsolationProcess, container.IsolationHyperV)
```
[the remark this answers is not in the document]
simonferquel: Good point. Actually it is already validated daemon-side with different validation logic on Linux (which only supports empty or default isolation) and windows (that also supports process and hyperv). I'll remove that client-side validation alltogether, and I'll make sure daemon-side errors are surfaced in a good way to the user

thaJeztah:
Thanks!

docs/reference/commandline/service_create.md
```diff
@@ -875,6 +876,14 @@ $ docker inspect --format="{{.Config.Hostname}}" 2e7a8a9c4da2-wo41w8hg8qanxwjwsg
 x3ti0erg11rjpg64m75kej2mz-hosttempl
 ```
 
+### Specify isolation (Windows)
+
+By default, tasks scheduled on Windows nodes are run using the default isolation mode configured for this particular node. If you want to force a specific isolation, you can use the `--isolation` flag (valid values are `default`, `process` and `hyperv`): 
+```bash
```
mistyhacks: Nit: you'll need a newline before the  code block.

I'd rewrite this so that the values are listed out with their meanings, rather than just being within parentheses inline.

docs/reference/commandline/service_create.md
```diff
@@ -42,6 +42,7 @@ Options:
       --help                               Print usage
       --host list                          Set one or more custom host-to-IP mappings (host:ip)
       --hostname string                    Container hostname
+      --isolation isolation                defines the isolation mode (hyperv, process)
```
mistyhacks: Here and below, this help text could just b e "Isolation mde for the service". For most commands we don't list out potential values here. @thaJeztah WDYT?

cli/command/service/opts.go
```diff
@@ -784,6 +814,8 @@ func addServiceFlags(flags *pflag.FlagSet, opts *serviceOptions, defaultFlagValu
 
 	flags.StringVar(&opts.stopSignal, flagStopSignal, "", "Signal to stop the container")
 	flags.SetAnnotation(flagStopSignal, "version", []string{"1.28"})
+	flags.Var(&opts.isolation, flagIsolation, "isolation mode for the service")
```
thaJeztah: Can you use a capital `I` here for the flag description? We do so for all the other flags

cli/compose/schema/schema_test.go
```diff
@@ -84,3 +84,16 @@ func TestValidatePlacement(t *testing.T) {
...
 	assert.NoError(t, Validate(config, "3.3"))
 }
+
+func TestValidateIsolation(t *testing.T) {
+	config := dict{
+		"version": "3.5",
+		"services": dict{
+			"foo": dict{
+				"image":     "busybox",
+				"isolation": "process",
```
thaJeztah: The actual validation of the _value_ for isolation is now handled by the daemon, correct? Given that this test is around validation (but only validating that _an_ "isolation" key is allows), I'd prefer the value to be different to show it's not being validated here. Something like `"some-isolation-value"`

docs/reference/commandline/service_create.md
```diff
@@ -42,6 +42,7 @@ Options:
       --help                               Print usage
       --host list                          Set one or more custom host-to-IP mappings (host:ip)
       --hostname string                    Container hostname
+      --isolation isolation                isolation mode for the service
```
thaJeztah: Capital `I` here as well

contrib/completion/zsh/_docker
```diff
@@ -1955,6 +1955,7 @@ __docker_service_subcommand() {
         "($help)--health-retries=[Consecutive failures needed to report unhealthy]:retries:(1 2 3 4 5)"
         "($help)--health-timeout=[Maximum time to allow one check to run]:time: "
         "($help)--hostname=[Service container hostname]:hostname: " \
+        "($help)--isolation=[Container isolation mode]:isolation:(default process hyperv)" \
```
thaJeztah: Should this be the same description as the actual flag? ("Isolation mode for the service") ?

thaJeztah: Or "Service container isolation mode"

docs/reference/commandline/service_create.md
```diff
@@ -875,6 +876,20 @@ $ docker inspect --format="{{.Config.Hostname}}" 2e7a8a9c4da2-wo41w8hg8qanxwjwsg
 x3ti0erg11rjpg64m75kej2mz-hosttempl
 ```
 
+### Specify isolation (Windows)
```
thaJeztah: `s/isolation/isolation mode/` ?

docs/reference/commandline/service_update.md
```diff
@@ -258,6 +259,12 @@ $ docker service update \
 Some flags of `service update` support the use of templating.
 See [`service create`](./service_create.md#templating) for the reference.
 
+
+### Specify isolation (Windows)
```
thaJeztah: `s/isolation/isolation mode/` ?

docs/reference/commandline/service_create.md
```diff
@@ -875,6 +876,20 @@ $ docker inspect --format="{{.Config.Hostname}}" 2e7a8a9c4da2-wo41w8hg8qanxwjwsg
...
+By default, tasks scheduled on Windows nodes are run using the default isolation mode configured for this particular node. If you want to force a specific isolation, you can use the `--isolation` flag: 
+
+```bash
+$ docker service create --name processisolated --isolation=process microsoft/nanoserver
+```
+
+Supported isolation modes on Windows are:
+- `default`: use default settings specified on the node running the task
+- `process`: use process isolation (Windows server only)
+- `hyperv`: use hyperv isolation
```
thaJeztah: Can you change `use hyperv isolation` to `use Hyper-V isolation`? Looks like that's the preferred name for this; https://en.wikipedia.org/wiki/Hyper-V

docs/reference/commandline/service_create.md
```diff
@@ -875,6 +876,20 @@ $ docker inspect --format="{{.Config.Hostname}}" 2e7a8a9c4da2-wo41w8hg8qanxwjwsg
 x3ti0erg11rjpg64m75kej2mz-hosttempl
 ```
 
+### Specify isolation (Windows)
+
+By default, tasks scheduled on Windows nodes are run using the default isolation mode configured for this particular node. If you want to force a specific isolation, you can use the `--isolation` flag:
```
thaJeztah: `s/If you want to force a/To force a/`

Also, can you wrap this to 80 columns?

docs/reference/commandline/service_create.md
```diff
@@ -875,6 +876,20 @@ $ docker inspect --format="{{.Config.Hostname}}" 2e7a8a9c4da2-wo41w8hg8qanxwjwsg
 x3ti0erg11rjpg64m75kej2mz-hosttempl
 ```
 
+### Specify isolation (Windows)
+
+By default, tasks scheduled on Windows nodes are run using the default isolation mode configured for this particular node. If you want to force a specific isolation, you can use the `--isolation` flag: 
+
+```bash
+$ docker service create --name processisolated --isolation=process microsoft/nanoserver
```
thaJeztah: nit: `processisolated` looks funny, perhaps just `myservice` (also makes it clearer there's nothing "special" in this value)

vendor.conf
```diff
@@ -53,4 +53,4 @@ golang.org/x/time a4bde12657593d5e90d0533a3e4fd95e635124cb
 google.golang.org/genproto d80a6e20e776b0b17a324d0ba1ab50a39c8e8944
 google.golang.org/grpc v1.3.0
 gopkg.in/yaml.v2 4c78c975fe7c825c6d1466c42be594d1d6f3aba6
-vbom.ml/util 928aaa586d7718c70f4090ddf83f2b34c16fdc8d
+vbom.ml/util 928aaa586d7718c70f4090ddf83f2b34c16fdc8d
```
thaJeztah: Can you un-stage this file?

cli/command/service/opts.go
```diff
@@ -457,6 +457,33 @@ func convertExtraHostsToSwarmHosts(extraHosts []string) []string {
...
 }
 
+type isolationOpts struct {
+	value  *container.Isolation
+	source string
+}
+
+func (o *isolationOpts) Set(value string) error {
+	o.source = value
+	tv := container.Isolation(strings.ToLower(value))
```
thaJeztah: I think we should remove the `strings.ToLower()`, and only accept strict values; other options are case-sensitive as well;

```bash
$ docker service create --mode=gLoBaL nginx:alpine
Unknown mode: gLoBaL, only replicated and global supported
```

contrib/completion/bash/docker
```diff
@@ -3305,6 +3305,7 @@ _docker_service_update_and_create() {
 		--health-start-period
 		--health-timeout
 		--hostname
+		--isolation
```
albers: This option is Windows-specific. There is a way to selectively add the completion only when run against a daemon running on Windows, see https://github.com/docker/cli/blob/master/contrib/completion/bash/docker#L2456-L2458.

contrib/completion/bash/docker
```diff
@@ -3493,6 +3494,10 @@ _docker_service_update_and_create() {
 			__docker_nospace
 			return
 			;;
+		--isolation)
+			COMPREPLY=( $( compgen -W "default process hyperv" -- "$cur" ) )
+			return
+			;;
```
albers: There is existing code to handle this completion, see https://github.com/docker/cli/blob/master/contrib/completion/bash/docker#L2496-L2500.

vendor.conf
simonferquel: I'll do this when squashing

contrib/completion/bash/docker
simonferquel: It is not: Your swarm master can be running Linux, but your cluster might contain Windows nodes and thus accept services with isolation set.

simonferquel:
@albers completion for build and run is out of scope of this PR (this one is about isolation setting at the service level, completion scripts being a part of the whole lot). I'd prefer not to pollute this PR with that.

contrib/completion/bash/docker
albers: Ah yes, you're right.

thaJeztah: Ah, that keeps tripping me up as well 😅

cli/command/service/opts.go
```diff
@@ -457,6 +457,33 @@ func convertExtraHostsToSwarmHosts(extraHosts []string) []string {
...
+type isolationOpts struct {
+	value  *container.Isolation
+	source string
+}
+
+func (o *isolationOpts) Set(value string) error {
+	o.source = value
+	tv := container.Isolation(value)
+	o.value = &tv
+	return nil
```
thaJeztah: This never returns an error; is the error to satisfy an interface?

simonferquel: Yes, it is the flag.Value interface.

simonferquel:
Squashed & rebased

cli/command/service/opts.go
thaJeztah: Check ✅

cli/compose/loader/loader.go
```diff
@@ -725,6 +727,14 @@ func transformSize(value interface{}) (interface{}, error) {
 	panic(errors.Errorf("invalid type for size %T", value))
 }
 
+func transformIsolation(value interface{}) (interface{}, error) {
+	str, ok := value.(string)
+	if !ok {
+		return value, errors.Errorf("invalid type %T for isolation", value)
+	}
+	return container.Isolation(strings.ToLower(str)), nil
```
thaJeztah: Missed this in my previous review; can you remove the `strings.ToLower()` here as well?

thaJeztah:
ping @vdemeester @dnephin PTAL

cli/command/service/opts.go
```diff
@@ -457,6 +457,33 @@ func convertExtraHostsToSwarmHosts(extraHosts []string) []string {
...
+}
+
+func (o *isolationOpts) Type() string {
+	return "isolation"
```
dnephin: This type is what gets reported in the help text. I wonder if `string` would be more helpful to users.

cli/compose/types/types.go
```diff
@@ -125,7 +127,8 @@ type ServiceConfig struct {
 	Ulimits         map[string]*UlimitsConfig
 	User            string
 	Volumes         []ServiceVolumeConfig
-	WorkingDir      string `mapstructure:"working_dir"`
+	WorkingDir      string              `mapstructure:"working_dir"`
+	Isolation       container.Isolation `mapstructure:"isolation"`
```
dnephin: I think this should be `string`. Everywhere else we accept plain types and do the convert to swarm/api types in convert. This helps keep the compose config packages decoupled from the api.

I believe this change will remove the need for the custom transform in loader.go as well.

dnephin: I guess this was here previously because of validation. I should have looked closer when I was commenting the first time. Sorry about that.

simonferquel:
@dnephin I just reverted service options to use plain string type for isolation, and modified compose types to remove all traces of validation (removing transformIsolation). Should be ok now.

docs/reference/commandline/service_create.md
```diff
@@ -42,6 +42,7 @@ Options:
       --help                               Print usage
       --host list                          Set one or more custom host-to-IP mappings (host:ip)
       --hostname string                    Container hostname
+      --isolation isolation                Service container isolation mode
```
thaJeztah: this should be "string" now then? (same for the other usage output)

simonferquel:
Good catch @thaJeztah , fixed :)

thaJeztah:
Oh, actually; dang, test is failing:

```
--- FAIL: TestLoadV35 (0.00s)
	Error Trace:	loader_test.go:1473
	Error:      	Not equal: 
	            	expected: container.Isolation("process")
	            	received: string("process")
=== RUN   TestLoadV35InvalidIsolation
--- FAIL: TestLoadV35InvalidIsolation (0.00s)
	Error Trace:	loader_test.go:1490
	Error:      	Not equal: 
	            	expected: container.Isolation("invalid")
	            	received: string("invalid")
```

cli/compose/schema/data/config_schema_v3.5.json
```diff
@@ -0,0 +1,544 @@
+{
```
thaJeztah: Wondering what we did previous times when adding a new version; perhaps it's cleaner if there were two commits; one that adds a "pristine" compose 3.5 (exact copy of 3.4), and one commit that adds your changes

thaJeztah:
🍪

PR #433, opened 2017-08-11
Minor changes to the command Build

tmelnychenko:
To avoid getting the error
docker : unable to prepare context: unable to find 'git': exec: "git": executable file not found in %PATH%

### DIFF
--- a/docs/reference/commandline/build.md
+++ b/docs/reference/commandline/build.md
@@ -96,7 +96,7 @@ For example, run this command to use a directory called `docker` in the branch
 `container`:
 
 ```bash
-$ docker build https://github.com/docker/rootfs.git#container:docker
+$ docker build https://github.com/docker/rootfs#container:docker
 ```
 
 The following table represents all the valid suffixes with their build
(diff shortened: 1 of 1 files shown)


mstanleyjones:
Please sign the commit. 🎉

docs/reference/commandline/build.md
```diff
@@ -96,22 +96,22 @@ For example, run this command to use a directory called `docker` in the branch
 `container`:
 
 ```bash
-$ docker build https://github.com/docker/rootfs.git#container:docker
+$ docker build https://github.com/docker/rootfs#container:docker
```
thaJeztah: I don't think this change is correct; the error message indicates that `git` is not installed on your machine, which is needed to checkout the linked repository;

```
unable to find 'git': exec: "git": executable file not found in %PATH%
```

For example, this just works:

```bash
$ docker build -t repro-433 https://github.com/thaJeztah/pgadmin4-docker.git#1.4

Sending build context to Docker daemon  169.5kB
Step 1/8 : FROM python:2-alpine
 ---> c80455665c57
Step 2/8 : ENV PGADMIN_VERSION 1.4 PYTHONDONTWRITEBYTECODE 1
 ---> Running in cf5c0caf0ef4
 ---> affe9f50eeb0
Removing intermediate container cf5c0caf0ef4
....
```

dnephin: Since the server supports git urls as a remote, I wonder if the client should detect this and perform the API request with the git url instead of creating the context locally.

thaJeztah: There were some concerns about the "remote" functionality in general (separate discussion), e.g. giving a client permission to have the remote server use keys that happen to be present on the server's root home (that's for `git@` notation)

So not entirely sure we should automatically do that

/cc @tonistiigi

tmelnychenko: Sorry, it was my bad, the change is incorrect.
Git is installed on the machine, but I have tried to run commands in Windows PowerShell ISE (my work machine). On MAC (home machine) all works like a charm and on my work machine all works in cmd.

tmelnychenko:
Change is incorrect.
The issue occurs only when run commands on Windows PowerShell ISE without additional configuration for working with GIT.

PR #442, opened 2017-08-15
Docs for Windows daemon graphdriver options

jhowardmsft:

@thaJeztah @johnstep 

Follow up to https://github.com/moby/moby/pull/34508#issuecomment-322513603 and adds documentation for the graphdriver options on Windows (both WCOW and LCOW).

### DIFF
--- a/docs/reference/commandline/dockerd.md
+++ b/docs/reference/commandline/dockerd.md
@@ -248,7 +248,7 @@ $ docker -H tcp://127.0.0.1:2375 pull ubuntu
 
 ### Daemon storage-driver
 
-The Docker daemon has support for several different image layer storage
+On Linux, the Docker daemon has support for several different image layer storage
 drivers: `aufs`, `devicemapper`, `btrfs`, `zfs`, `overlay` and `overlay2`.
 
 The `aufs` driver is the oldest, but is based on a Linux kernel patch-set that
@@ -296,11 +296,16 @@ to use it.
 > **Note**: Both `overlay` and `overlay2` are currently unsupported on `btrfs`
 > or any Copy on Write filesystem and should only be used over `ext4` partitions.
 
+On Windows, the Docker daemon supports a single image layer storage driver
+depending on the image platform: `windowsfilter` for Windows images, and
+`lcow` for Linux containers on Windows.
+
 ### Options per storage driver
 
 Particular storage-driver can be configured with options specified with
 `--storage-opt` flags. Options for `devicemapper` are prefixed with `dm`,
-options for `zfs` start with `zfs` and options for `btrfs` start with `btrfs`.
+options for `zfs` start with `zfs`, options for `btrfs` start with `btrfs`
+and options for `lcow` start with `lcow`.
 
 #### Devicemapper options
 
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/dockerd.md
```diff
@@ -780,6 +785,113 @@ conditions the user can pass any size less then the backing fs size.
...
+###### Example
+
+```PowerShell
+C:\> dockerd --storage-opt lcow.timeout=240
+```
+
+##### `lcow.sandboxsize`
+
+Specifies the size in GB to use when creating the sandbox which is used for
+containers. Defaults to 20. Cannot be less than 20.
```
thaJeztah: should we have a `size` option as well for this?

docs/reference/commandline/dockerd.md
```diff
@@ -780,6 +785,113 @@ conditions the user can pass any size less then the backing fs size.
...
+
+###### Example
+
+```PowerShell
+C:\> dockerd --storage-opt size=40G
+```
+
+#### LCOW (Linux Containers on Windows) options
+
+##### `lcow.globalmode`
```
thaJeztah: Are the `lcow` options merged with the container's `--storage-opt` now (sorry if a silly question, wasn't sure if `runtime.GOOS == "windows"` is / is not true for LCOW)

docs/reference/commandline/dockerd.md
```diff
@@ -296,11 +296,16 @@ to use it.
...
+depending on the image platform: `windowsfilter` for Windows images, and
+`lcow` for Linux containers on Windows.
+
 ### Options per storage driver
 
 Particular storage-driver can be configured with options specified with
 `--storage-opt` flags. Options for `devicemapper` are prefixed with `dm`,
-options for `zfs` start with `zfs` and options for `btrfs` start with `btrfs`.
+options for `zfs` start with `zfs`, options for `btrfs` start with `btrfs`
+and options for `lcow` start with `lcow`.
```
thaJeztah: nit: should be a comma before `and` (no need to fix if there's no other things to fix) :smile:

docs/reference/commandline/dockerd.md
jhowardmsft: I'm documenting what's implemented. Plus you told me to prefix options with `lcow.` 😇

docs/reference/commandline/dockerd.md
jhowardmsft: Not sure I understand the question here..... Merging as per the PR which made this documentation followup? If so, yes. And on LCOW, `runtime.GOOS=="windows"`, yes. It's still a daemon running on Windows. Just the target containers (and images) are Linux ones.

docs/reference/commandline/dockerd.md
jhowardmsft: Oh the Oxford command debate. I would have put one in, but it wasn't there before....

jhowardmsft:
Commit string updated

docs/reference/commandline/dockerd.md
thaJeztah: In that case, the `docker run` and `docker create` reference docs also need an update, because they support all those options (per container)? https://github.com/docker/cli/blob/master/docs/reference/commandline/run.md

jhowardmsft: Let's do that as further follow-up PRs....

thaJeztah: Well, was wondering if all these options were intended to be configurable per container (also why I asked https://github.com/moby/moby/pull/34508#discussion_r133084507); I didn't know there were this many options; and if we want to have all these bells & whistles on `docker run`.

Are there use cases of having all these configurable?

jhowardmsft: These are all the supported daemon options for startup. For run, with the exception of globalmode, they would all be supported.  



> Are there use cases of having all these configurable?

Absolutely. A user may want to run different containers with different kernel revisions as one such example.

jhowardmsft:
@mstanleyjones To be fair a) I'm doing the update where @thaJeztah asked me to.... and b) the updates are consistent with the Linux updates. If you think they should move, that should probably be an entire-section-move in a follow up PR. The contents of this PR seem valid with the current organization of the content.

docs/reference/commandline/dockerd.md
thaJeztah: It feels odd that a _kernel version_ is a _storage driver_ option

jhowardmsft: Ultimately it's the storage driver that instantiates and manages the lifetime of all the utility VMs for the remote file system operations. Storage opts are the only thing that get this down to the driver (and also are present for `run`). I'm really not sure you'd want me changing graphdriver interfaces....

However, this is documentation, and documenting what's implemented and in master.... If you strongly think a code change is necessary (I'm far from convinced), then that should be a separate issue on moby/moby.

jhowardmsft:
ping @thaJeztah @mstanleyjones - Can you guys figure out between you where this should be reorg'd to in a future PR? This seems good with the current structure.

mstanleyjones:
I'm good with considering organization in a future PR.

jhowardmsft:
ping @thaJeztah

PR #446, opened 2017-08-16
Fixed grammar

krizalys:
**- What I did**

Minor grammar change in `service create` documentation: `read to` => `read from` and `write from` => `write to`.

**- How to verify it**

https://docs.docker.com/engine/reference/commandline/service_create/#create-a-service-with-specific-hostname-hostname

Minor grammar change in `service create` documentation.

### DIFF
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -243,14 +243,12 @@ $ docker service create \
 For more information about labels, refer to [apply custom
 metadata](https://docs.docker.com/engine/userguide/labels-custom-metadata/).
 
-### Add bind-mounts or volumes
+### Add bind-mounts, volumes or memory filesystems
 
-Docker supports two different kinds of mounts, which allow containers to read to
-or write from files or directories on other containers or the host operating
-system. These types are _data volumes_ (often referred to simply as volumes) and
-_bind-mounts_.
-
-Additionally, Docker supports `tmpfs` mounts.
+Docker supports three different kinds of mounts, which allow containers to read
+from or write to files or directories, either on the host operating system, or
+on memory filesystems. These types are _data volumes_ (often referred to simply
+as volumes), _bind-mounts_, and _tmpfs_.
 
 A **bind-mount** makes a file or directory on the host available to the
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/service_create.md
```diff
@@ -245,8 +245,8 @@ metadata](https://docs.docker.com/engine/userguide/labels-custom-metadata/).
 
 ### Add bind-mounts or volumes
 
-Docker supports two different kinds of mounts, which allow containers to read to
-or write from files or directories on other containers or the host operating
+Docker supports two different kinds of mounts, which allow containers to read
+from or write to files or directories on other containers or the host operating
```
thaJeztah: Thanks! 

The `on other containers` part is incorrect as well; mounts are always mounted from the host to the container; if two containers share the same host path for a mount, docker is mounting that path from the host, not the other container's filesystem.

Can you update your PR and remove the `on other containers`?

docs/reference/commandline/service_create.md
```diff
@@ -245,8 +245,8 @@ metadata](https://docs.docker.com/engine/userguide/labels-custom-metadata/).
 
 ### Add bind-mounts or volumes
 
-Docker supports two different kinds of mounts, which allow containers to read to
-or write from files or directories on other containers or the host operating
+Docker supports two different kinds of mounts, which allow containers to read
```
mstanleyjones: Technically, there are three different kinds of mounts. The other is `tmpfs`. See https://docs.docker.com/engine/admin/volumes/#choose-the-right-type-of-mount

thaJeztah: Ah, I think `tmpfs` wasn't functional at the time of writing, so likely wasn't updated when it became functional

krizalys:
Based on your comments, @thaJeztah @mstanleyjones I updated a bit the paragraph to match the current feature more closely. Let me know.

krizalys:
Thanks @mstanleyjones, also slightly update the section heading and squashed all commits together.

Consolidating `bind-mounts` / `bind mounts` usages can be taken care of in a separate PR.

mstanleyjones:
LGTM

PR #449, opened 2017-08-16
Remove dependency on (most of) docker/docker/pkg/testutil

dnephin:
Use `gotestyourself/golden` assertion, which is more strict, so golden files were updated to match.

Remove the deprecated `NewFakeCliWithOutput()`

Removed some superfluous node tests that are now covered by a single test using `golden.Assert()`

Use `gotestyourself/fs` to in a couple packages that used `pkg/testutil/tempfile`.

Use `gotestyourself/skip` for a couple tests that don't support windows.

The only remaining import of `pkg/testutil` is for `ErrorContains()`. I opened a PR against testify a couple weeks ago (https://github.com/stretchr/testify/pull/486) to see about adding it there.  

If it doesn't get added we could either add it to `gotestyourself`, or to `cli/internal/test/assert/assert.go` in a follow up PR.


tiborvass:
LGTM

PR #451, opened 2017-08-17
Use pass backend

tych0:
To the end of encrypting passwords by default, there is now a `pass` backend, which will be useable on headless linuxen. Let's prefer that instead of the secretservice backend when choosing a default password backend on linux.

### DIFF
--- a/docs/reference/commandline/login.md
+++ b/docs/reference/commandline/login.md
@@ -63,8 +63,9 @@ $ cat ~/my_password.txt | docker login --username foo --password-stdin
 2.  user is added to the `docker` group.  This will impact the security of your system; the `docker` group is `root` equivalent.  See [Docker Daemon Attack Surface](https://docs.docker.com/security/security/#docker-daemon-attack-surface) for details.
 
 You can log into any public or private repository for which you have
-credentials.  When you log in, the command stores encoded credentials in
-`$HOME/.docker/config.json` on Linux or `%USERPROFILE%/.docker/config.json` on Windows.
+credentials.  When you log in, the command stores credentials in
+`$HOME/.docker/config.json` on Linux or `%USERPROFILE%/.docker/config.json` on
+Windows, via the procedure described below.
 
 ### Credentials store
 
@@ -82,6 +83,7 @@ you can download them from:
 - D-Bus Secret Service: https://github.com/docker/docker-credential-helpers/releases
 - Apple macOS keychain: https://github.com/docker/docker-credential-helpers/releases
 - Microsoft Windows Credential Manager: https://github.com/docker/docker-credential-helpers/releases
+- [pass](https://www.passwordstore.org/): https://github.com/docker/docker-credential-helpers/releases
 
 You need to specify the credentials store in `$HOME/.docker/config.json`
 to tell the docker engine to use it. The value of the config property should be
@@ -96,6 +98,15 @@ For example, to use `docker-credential-osxkeychain`:
 
 If you are currently logged in, run `docker logout` to remove
 the credentials from the file and run `docker login` again.
+
+### Default behavior
+
+By default, Docker looks for the native binary on each of the platforms, i.e.
+"osxkeychain" on macOS, "wincred" on windows, and "pass" on Linux. A special
+case is that on Linux, Docker will fall back to the "secretservice" binary if
+it cannot find the "pass" binary. If none of these binaries are present, it
+stores the credentials (i.e. password) in base64 encoding in the config files
+described above.
 
 ### Credential helper protocol
 
--- a/vendor.conf
+++ b/vendor.conf
@@ -5,7 +5,7 @@ github.com/cpuguy83/go-md2man a65d4d2de4d5f7c74868dfa9b202a3c8be315aaa
 github.com/davecgh/go-spew 346938d642f2ec3594ed81d874461961cd0faa76
 github.com/docker/distribution edc3ab29cdff8694dd6feb85cfeb4b5f1b38ed9c
 github.com/docker/docker 84144a8c66c1bb2af8fa997288f51ef2719971b4
-github.com/docker/docker-credential-helpers v0.5.1
+github.com/docker/docker-credential-helpers 3c90bd29a46b943b2a9842987b58fb91a7c1819b
 
 # the docker/go package contains a customized version of canonical/json
 # and is used by Notary. The package is periodically rebased on current Go versions.
--- a/cli/config/credentials/default_store.go
+++ b/cli/config/credentials/default_store.go
@@ -7,13 +7,15 @@ import (
 // DetectDefaultStore return the default credentials store for the platform if
 // the store executable is available.
 func DetectDefaultStore(store string) string {
+	platformDefault := defaultCredentialsStore()
+
 	// user defined or no default for platform
-	if store != "" || defaultCredentialsStore == "" {
+	if store != "" || platformDefault == "" {
 		return store
 	}
 
-	if _, err := exec.LookPath(remoteCredentialsPrefix + defaultCredentialsStore); err == nil {
-		return defaultCredentialsStore
+	if _, err := exec.LookPath(remoteCredentialsPrefix + platformDefault); err == nil {
+		return platformDefault
 	}
 	return ""
 }
--- a/cli/config/credentials/default_store_darwin.go
+++ b/cli/config/credentials/default_store_darwin.go
@@ -1,3 +1,5 @@
 package credentials
 
-const defaultCredentialsStore = "osxkeychain"
+func defaultCredentialsStore() string {
+	return "osxkeychain"
+}
--- a/cli/config/credentials/default_store_linux.go
+++ b/cli/config/credentials/default_store_linux.go
@@ -1,3 +1,5 @@
 package credentials
 
-const defaultCredentialsStore = "secretservice"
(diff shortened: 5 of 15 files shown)


vendor.conf
```diff
@@ -8,7 +8,7 @@ github.com/cpuguy83/go-md2man a65d4d2de4d5f7c74868dfa9b202a3c8be315aaa
 github.com/davecgh/go-spew 346938d642f2ec3594ed81d874461961cd0faa76
 github.com/docker/distribution b38e5838b7b2f2ad48e06ec4b500011976080621
 github.com/docker/docker d58ffa0364c04d03a8f25704d7f0489ee6cd9634
-github.com/docker/docker-credential-helpers v0.5.1
+github.com/docker/docker-credential-helpers 3c90bd29a46b943b2a9842987b58fb91a7c1819b
```
dnephin: Should we tag a v0.5.3 release so we can keep using a tag?

What's the release policy on this repo?

tych0:
On Thu, Aug 17, 2017 at 04:08:48PM +0000, Daniel Nephin wrote:
> dnephin approved this pull request.
> 
> LGTM
> 
> > @@ -8,7 +8,7 @@ github.com/cpuguy83/go-md2man a65d4d2de4d5f7c74868dfa9b202a3c8be315aaa
>  github.com/davecgh/go-spew 346938d642f2ec3594ed81d874461961cd0faa76
>  github.com/docker/distribution b38e5838b7b2f2ad48e06ec4b500011976080621
>  github.com/docker/docker d58ffa0364c04d03a8f25704d7f0489ee6cd9634
> -github.com/docker/docker-credential-helpers v0.5.1
> +github.com/docker/docker-credential-helpers 3c90bd29a46b943b2a9842987b58fb91a7c1819b
> 
> Should we tag a v0.5.3 release so we can keep using a tag?
> 
> What's the release policy on this repo?

Happy to update if necessary. If we do tag, can we merge
https://github.com/docker/docker-credential-helpers/pull/84 first?

vdemeester:
@tych0 yes 👼 once it's merged in docker-credential-helpers I'll make a 0.6.0 release 😉

andrewhsu:
Does this handle migration of credentials from older format?

Also, any suitable tests we should have verify expected functionality of this PR?

tych0:
On Wed, Aug 30, 2017 at 01:13:35AM +0000, Andrew Hsu wrote:
> Does this handle migration of credentials from older format?

It doesn't migrate credentials from the older format, although we
could implement that. It simply uses them in whatever cred store
they're currently stored in (including the base64 encoded native one).
For new hosts with pass configured it will use that cred store.

> Also, any suitable tests we should have verify expected functionality of this PR?

Yes, if you test `docker login`, a `docker login` with `pass`
configured would be sufficient, and we could probably test that the
raw password file has the right contents at the end of a `docker
login` too, just to be sure.

tych0:
Hey guys, just a heads up, I've updated this version to a newer version than the 0.6.0 tag, because that includes a better initialization check. I can revert to the tag if we really want, but since a few other things aren't, and it offers a relevant feature, I figured I'd just use the hash.

I've also rebased on top of the latest master. Let me know if anything looks funny. Thanks!

tych0:
Ping, any movement on this?

vendor.conf
```diff
@@ -5,7 +5,7 @@ github.com/cpuguy83/go-md2man a65d4d2de4d5f7c74868dfa9b202a3c8be315aaa
 github.com/davecgh/go-spew 346938d642f2ec3594ed81d874461961cd0faa76
 github.com/docker/distribution edc3ab29cdff8694dd6feb85cfeb4b5f1b38ed9c
 github.com/docker/docker 84144a8c66c1bb2af8fa997288f51ef2719971b4
-github.com/docker/docker-credential-helpers v0.5.1
+github.com/docker/docker-credential-helpers 3c90bd29a46b943b2a9842987b58fb91a7c1819b
```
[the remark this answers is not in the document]
tych0: As I mentioned in the comment above, there is no tagged version with commits relevant to this patch, so unless we tag a 0.6.1, I don't think we should.

tych0:
On Thu, Sep 21, 2017 at 02:07:00AM -0700, Sebastiaan van Stijn wrote:
> thaJeztah requested changes on this pull request.
> 
> Generally LGTM, but a couple of things:
> 
> - Can we test/double check upgrading if a user currently uses the "secret service" credentials helper? I.e., if a user uses that currently, we should make docker keep using that (similar to the way we handle graph-drivers; if a previous directory of a graph-driver is detected, that one is used)

It already does this. If it uses the secret service helper, it sets
the credsStore key to "secretservice", and the defaults are no longer
used.

> - This needs an update to the documentation; https://github.com/docker/cli/blob/master/docs/reference/commandline/login.md#credentials-store. We need to mention the new helper, and describe what the defaults are / priority is

Sure, I can update those.

> - The docs also mention that `credsStore` must be present in the `config.json` to make docker use a credentials helper; is that still the case?

Well, it depends :). If you have a pre-configured set of credentials
then yes, but if you're doing a new login and have the binary, it will
auto-detect that and use it, and save the credsStore key for future
reference.

tych0:
Ok, I've updated the docs. I think this should be good to go now.

docs/reference/commandline/login.md
```diff
@@ -97,6 +98,16 @@ For example, to use `docker-credential-osxkeychain`:
 If you are currently logged in, run `docker logout` to remove
 the credentials from the file and run `docker login` again.
 
+### Default behavior
+
+By default, Docker will look for the native binary on each of the platforms,
```
mstanleyjones: s/will look/looks

Maybe a table for the default backend per OS?

docs/reference/commandline/login.md
```diff
@@ -97,6 +98,16 @@ For example, to use `docker-credential-osxkeychain`:
...
+i.e. "osxkeychain" on macOS, "wincred" on windows, and "pass" on Linux. A
+special case is that on Linux, Docker will fall back to the "secretservice"
+binary if it cannot find the "pass" binary.
+
+*NOTE:* If you do not supply a credential helper binary, Docker will store your
```
mstanleyjones: Just Note, not NOTE

s/will store/stores

tych0:
On Thu, Sep 21, 2017 at 09:33:47AM -0700, Misty Stanley-Jones wrote:
> mstanleyjones requested changes on this pull request.
> 
> 
> 
> > @@ -97,6 +98,16 @@ For example, to use `docker-credential-osxkeychain`:
>  If you are currently logged in, run `docker logout` to remove
>  the credentials from the file and run `docker login` again.
>  
> +### Default behavior
> +
> +By default, Docker will look for the native binary on each of the platforms,
> 
> s/will look/looks
> 
> Maybe a table for the default backend per OS?

I thought about that, but the point is mostly to state that linux is a
bit different and that it tries pass and falls back to secret service,
and that doesn't really fit nicely in the table :(

Anyway, I've updated the other bits, thanks!

thaJeztah:
> and save the credsStore key for future reference.

Oh! I forgot (or wasn't aware) the client _saves_ the `credStore` to the config.json after detection; I thought it checked each time if the user did not set this.

docs/reference/commandline/login.md
```diff
@@ -97,6 +98,16 @@ For example, to use `docker-credential-osxkeychain`:
...
 the credentials from the file and run `docker login` again.
 
+### Default behavior
+
+By default, Docker looks for the native binary on each of the platforms, i.e.
+"osxkeychain" on macOS, "wincred" on windows, and "pass" on Linux. A special
+case is that on Linux, Docker will fall back to the "secretservice" binary if
+it cannot find the "pass" binary.
+
+*Note:* If you do not supply a credential helper binary, Docker stores your
```
thaJeztah: This should  be formatted as a note;

```markdown
> **Note**: If you do not supply a credential helper binary, Docker stores your
> password in cleartext (base64 encoded) in your home directory!
```

Should this mention the actual file this is saved in as well (`~/.docker/config.json`)?

thaJeztah: I see this is being referred to a couple lines up as well;

> You can log into any public or private repository for which you have credentials. When you log in, the command stores encoded credentials in `$HOME/.docker/config.json` on Linux or `%USERPROFILE%/.docker/config.json` on Windows.

tych0:
Ok, I tried to address those (I'm not sure what you meant by formatted as a note, so I just moved the text into the paragraph, since it probably should be there anyway).

thaJeztah:
@tych0 @seemethere are there changes needed for packaging?

tych0:
Yep, we talked a few weeks ago about it and I think we're all on the same page.

tych0:
Oh, but we didn't actually talk about the OSX or windows packaging/install files. @seemethere do you know who does that?

vdemeester:
@tych0 for docker4desktop, there is already osx-keychain and wincreds support (cc @ebriney @simonferquel) 😉

tych0:
Yes, but are we distributing the credhelpers binaries in the packages?

vdemeester:
I'm pretty sure we do yes 👼

tych0:
Sweet, so that part is solved already then :)

ebriney:
Yes, on windows and on mac, it is enabled by default. Thanks for this PR!

PR #456, opened 2017-08-18
Add unit tests for initializing the client

dnephin:
Replaces tests removed in https://github.com/moby/moby/pull/34560


vieux:
SGTM

PR #463, opened 2017-08-21
Remove docker/docker/pkg/testutil

dnephin:
* first moves `cli/internal/test` to `internal/test` so that we can use these testing packages in packages not under `cli/`
* next replace the one function we used from `pkg/testutil` with `internal/test/testutil`


dnephin:
Oh, codecov is finally fixed, going to run a build for master

vieux:
LGTM !

PR #467, opened 2017-08-23
Add docs for Dockerfile ADD/COPY --chown flag

estesp:
Document the new `--chown` flag added to the ADD and COPY commands in
the Dockerfile format.

Signed-off-by: Phil Estes <estesp@linux.vnet.ibm.com>

The actual implementation is ready to merge in moby/moby#34263

### DIFF
--- a/docs/reference/builder.md
+++ b/docs/reference/builder.md
@@ -817,16 +817,23 @@ change them using `docker run --env <key>=<value>`.
 
 ADD has two forms:
 
-- `ADD <src>... <dest>`
-- `ADD ["<src>",... "<dest>"]` (this form is required for paths containing
+- `ADD [--chown=<user>:<group>] <src>... <dest>`
+- `ADD [--chown=<user>:<group>] ["<src>",... "<dest>"]` (this form is required for paths containing
 whitespace)
+
+> **Note**:
+> The `--chown` feature is only supported on Dockerfiles used to build Linux containers,
+> and will not work on Windows containers. Since user and group ownership concepts do
+> not translate between Linux and Windows, the use of `/etc/passwd` and `/etc/group` for
+> translating user and group names to IDs restricts this feature to only be viable for
+> for Linux OS-based containers.
 
 The `ADD` instruction copies new files, directories or remote file URLs from `<src>`
 and adds them to the filesystem of the image at the path `<dest>`.
 
-Multiple `<src>` resource may be specified but if they are files or
-directories then they must be relative to the source directory that is
-being built (the context of the build).
+Multiple `<src>` resources may be specified but if they are files or
+directories, their paths are interpreted as relative to the source of
+the context of the build.
 
 Each `<src>` may contain wildcards and matching will be done using Go's
 [filepath.Match](http://golang.org/pkg/path/filepath#Match) rules. For example:
@@ -848,7 +855,26 @@ named `arr[0].txt`, use the following;
     ADD arr[[]0].txt /mydir/    # copy a file named "arr[0].txt" to /mydir/
 
 
-All new files and directories are created with a UID and GID of 0.
+All new files and directories are created with a UID and GID of 0, unless the
+optional `--chown` flag specifies a given username, groupname, or UID/GID
+combination to request specific ownership of the content added. The
+format of the `--chown` flag allows for either username and groupname strings
+or direct integer UID and GID in any combination. Providing a username without
+groupname or a UID without GID will use the same numeric UID as the GID. If a
+username or groupname is provided, the container's root filesystem
+`/etc/passwd` and `/etc/group` files will be used to perform the translation
+from name to integer UID or GID respectively. The following examples show
+valid definitions for the `--chown` flag:
+
+    ADD --chown=55:mygroup files* /somedir/
+    ADD --chown=bin files* /somedir/
+    ADD --chown=1 files* /somedir/
+    ADD --chown=10:11 files* /somedir/
+
+If the container root filesystem does not contain either `/etc/passwd` or
+`/etc/group` files and either user or group names are used in the `--chown`
+flag, the build will fail on the `ADD` operation. Using numeric IDs requires
+no lookup and will not depend on container root filesystem content.
 
 In the case where `<src>` is a remote file URL, the destination will
 have permissions of 600. If the remote file being retrieved has an HTTP
@@ -938,15 +964,23 @@ guide](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practi
 
 COPY has two forms:
 
-- `COPY <src>... <dest>`
-- `COPY ["<src>",... "<dest>"]` (this form is required for paths containing
+- `COPY [--chown=<user>:<group>] <src>... <dest>`
+- `COPY [--chown=<user>:<group>] ["<src>",... "<dest>"]` (this form is required for paths containing
 whitespace)
+
+> **Note**:
+> The `--chown` feature is only supported on Dockerfiles used to build Linux containers,
+> and will not work on Windows containers. Since user and group ownership concepts do
+> not translate between Linux and Windows, the use of `/etc/passwd` and `/etc/group` for
+> translating user and group names to IDs restricts this feature to only be viable for
+> for Linux OS-based containers.
 
 The `COPY` instruction copies new files or directories from `<src>`
 and adds them to the filesystem of the container at the path `<dest>`.
 
-Multiple `<src>` resource may be specified but they must be relative
-to the source directory that is being built (the context of the build).
+Multiple `<src>` resources may be specified but the paths of files and
+directories will be interpreted as relative to the source of the context
+of the build.
 
 Each `<src>` may contain wildcards and matching will be done using Go's
 [filepath.Match](http://golang.org/pkg/path/filepath#Match) rules. For example:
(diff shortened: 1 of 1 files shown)


docs/reference/builder.md
```diff
@@ -817,14 +817,14 @@ change them using `docker run --env <key>=<value>`.
...
+- `ADD [--chown <user>:<group>] <src>... <dest>`
+- `ADD [--chown <user>:<group>] ["<src>",... "<dest>"]` (this form is required for paths containing
 whitespace)
 
 The `ADD` instruction copies new files, directories or remote file URLs from `<src>`
 and adds them to the filesystem of the image at the path `<dest>`.
 
-Multiple `<src>` resource may be specified but if they are files or
+Multiple `<src>` resources may be specified but if they are files or
 directories then they must be relative to the source directory that is
```
mstanleyjones: directories, they must be

Also, "relative" I think means that they must be subdirectories of the build context? Maybe you can reword this to make it clearer.

docs/reference/builder.md
```diff
@@ -848,7 +848,25 @@ named `arr[0].txt`, use the following;
     ADD arr[[]0].txt /mydir/    # copy a file named "arr[0].txt" to /mydir/
 
 
-All new files and directories are created with a UID and GID of 0.
+All new files and directories are created with a UID and GID of 0, unless the
+optional `--chown` flag is used to provide a user and group definition which
+will be used to change the ownership of the created files and directories. The
```
mstanleyjones: How about "optional `--chown` flag specifies a given username, groupname, or UID/GID combination."

(if you only want to give a GID, how does Docker know it's not a UID? I think if you only provide a single integer it's always assumed to be a UID. Maybe best to state this too.)

mstanleyjones:
cc/ @gbarr01 for FYI

docs/reference/builder.md
```diff
@@ -817,14 +817,14 @@ change them using `docker run --env <key>=<value>`.
 
 ADD has two forms:
 
-- `ADD <src>... <dest>`
-- `ADD ["<src>",... "<dest>"]` (this form is required for paths containing
+- `ADD [--chown <user>:<group>] <src>... <dest>`
```
shouze: Sure about that syntax? Not working for me & @presto9292 see https://github.com/moby/moby/pull/34263#issuecomment-328034492

estesp: Thanks @shouze ; you are correct that it requires an `=`; I will update this PR as it also has other edits required before being complete.

estesp:
@mstanleyjones thanks for the review; I have updated per your comments; PTAL!

@shouze examples and flag documentation updated with the *correct* argument style--thanks for the catch!

rcjsuen:
What is the behaviour of using this flag on Windows as the `/etc/passwd` and `/etc/group` files do not exist there? Should this be included in the documentation?

docs/reference/builder.md
```diff
@@ -848,7 +848,26 @@ named `arr[0].txt`, use the following;
     ADD arr[[]0].txt /mydir/    # copy a file named "arr[0].txt" to /mydir/
 
 
-All new files and directories are created with a UID and GID of 0.
+All new files and directories are created with a UID and GID of 0, unless the
```
mstanleyjones: How do the concepts of UID and GID 0 map to Windows?

docs/reference/builder.md
```diff
@@ -848,7 +848,26 @@ named `arr[0].txt`, use the following;
...
+optional `--chown` flag specifies a given username, groupname, or UID/GID
+combination to request specific ownership of the content added. The
+format of the `--chown` flag allows for either username and groupname strings
+or direct integer UID and GID in any combination. Providing a username without
+groupname or a UID without GID will use the name numeric UID as the GID. If a
+username or groupname is provided, the container's root filesystem
+`/etc/passwd` and `/etc/group` files will be used to perform the translation
```
mstanleyjones: How does this map to Windows?

docs/reference/builder.md
estesp: They do not apply/map at all.. I assume we need a note about that :)  A `Dockerfile` with the `--chown` flag on an `ADD` or `COPY` command will fail during `docker build` if it is built on a Windows Docker daemon.

This was discussed in the PR for the feature implementation itself--the other option was to "ignore" the flag on Windows but today this is **not** the way it is implemented.

docs/reference/builder.md
```diff
@@ -848,7 +848,26 @@ named `arr[0].txt`, use the following;
     ADD arr[[]0].txt /mydir/    # copy a file named "arr[0].txt" to /mydir/
 
 
-All new files and directories are created with a UID and GID of 0.
+All new files and directories are created with a UID and GID of 0, unless the
+optional `--chown` flag specifies a given username, groupname, or UID/GID
+combination to request specific ownership of the content added. The
+format of the `--chown` flag allows for either username and groupname strings
+or direct integer UID and GID in any combination. Providing a username without
+groupname or a UID without GID will use the name numeric UID as the GID. If a
```
thaJeztah: s/name/same/ (probably)

docs/reference/builder.md
thaJeztah: Perhaps we should add a header above this section; both for easier linking (anchor), and an introduction "This is a Linux-only feature". Also, the sections for `COPY` and `ADD` for this feature could then be combined if we think that's a good idea (_"All new files and directories that are added through `ADD` or `COPY` ...."_).  (I'm sure @mstanleyjones has better suggestions :smile:)

What is actually the owner of the files on Windows? (@johnstep?)

docs/reference/builder.md
```diff
@@ -968,7 +988,26 @@ named `arr[0].txt`, use the following;
 
     COPY arr[[]0].txt /mydir/    # copy a file named "arr[0].txt" to /mydir/
 
-All new files and directories are created with a UID and GID of 0.
+All new files and directories are created with a UID and GID of 0, unless the
+optional `--chown` flag specifies a given username, groupname, or UID/GID
+combination to request specific ownership of the copied content. The
+format of the `--chown` flag allows for either username and groupname strings
+or direct integer UID and GID in any combination. Providing a username without
+groupname or a UID without GID will use the name numeric UID as the GID. If a
```
thaJeztah: s/name/same/

mstanleyjones:
I like the idea of a note like "The `--chown` and `--chgrp` features are only supported on Dockerfiles for Linux containers, and will not work on Windows containers." Maybe you can elaborate on what "will not work" looks like.

thaJeztah:
ping @estesp 😇

estesp:
@mstanleyjones @thaJeztah updated; PTAL.

I did not combine the sections yet; I'm definitely willing to do so, but currently the anchor points are to each possible command in a `Dockerfile`; so I don't know what to name that anchor. Is it "ADD, COPY" or "ADD and COPY" or "ADD/COPY" ? :)

Given they are still distinct I added the same exact note about lack of capability on Windows to use the `--chown` feature.

docs/reference/builder.md
```diff
@@ -938,15 +964,23 @@ guide](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practi
 
 COPY has two forms:
 
-- `COPY <src>... <dest>`
-- `COPY ["<src>",... "<dest>"]` (this form is required for paths containing
+- `COPY [--chown=<user>:<group>] <src>... <dest>`
+- `COPY [--chown=<user>:<group>] ["<src>",... "<dest>"]` (this form is required for paths containing
```
thaJeztah: lol, this will be a bit confusing as the first square brackets are to indicate "it's optional" and the second are literal brackets for JSON.

Don't have a better suggestion at this point though (other than having an example perhaps) - I'm okay with adding that in a follow up

docs/reference/builder.md
johnstep: @estesp What I've seen is that passing `--chown` succeeds on Windows, if a _number_ is supplied, but fails for a name since the code attempts to resolve the name by looking in `\etc\passwd` on Windows, which does not make sense.

Failing outright probably makes more sense, or at least warning, but I would prefer a more deliberate check and message. We have an issue on adding a warning, but it needs some more details: https://github.com/moby/moby/issues/35026

@thaJeztah We'll look into how the owner is set on Windows. From a quick test, it seems to be set differently based on which directory a file is copied to, but we need to verify.

estesp:
This feature went into the 17.09 release, so it currently undocumented as far as online docs. Is there anything else not addressed here? Just don't want it to get lost for too long!

karser:
@estesp keep track changelog on docker-ce [releases](https://github.com/docker/docker-ce/releases) page to stay current

mstanleyjones:
@estesp well, this PR would have needed to be merged before `docker/cli` was merged into the appropriate `docker/docker-ce` branch.  The docs currently pull the `builder.md` from https://github.com/docker/docker-ce/blob/17.09/components/cli/docs/reference/builder.md.

rcjsuen:
Unfortunately, this didn't make 17.10 either as far as I can tell. That's too bad but hey, it happens.

What outstanding issues are there left that are preventing it from being merged? I would like to add support for this flag in my project once the documentation has been finalized and is made available online.

thaJeztah:
I'll go ahead and merge :+1: thanks!

PR #471, opened 2017-08-24
Include whether the managers in the swarm are autolocked as part of `docker info`

cyli:
Currently, the only way to tell if the cluster is autolocked is to attempt to unlock or get the unlock key.  The info is already available on the API for `docker info`, so also include it in the text output on the CLI.

cc @friism 

image

### DIFF
--- a/cli/command/system/info.go
+++ b/cli/command/system/info.go
@@ -54,7 +54,7 @@ func runInfo(dockerCli *command.DockerCli, opts *infoOptions) error {
 }
 
 // nolint: gocyclo
-func prettyPrintInfo(dockerCli *command.DockerCli, info types.Info) error {
+func prettyPrintInfo(dockerCli command.Cli, info types.Info) error {
 	fmt.Fprintf(dockerCli.Out(), "Containers: %d\n", info.Containers)
 	fmt.Fprintf(dockerCli.Out(), " Running: %d\n", info.ContainersRunning)
 	fmt.Fprintf(dockerCli.Out(), " Paused: %d\n", info.ContainersPaused)
@@ -76,7 +76,7 @@ func prettyPrintInfo(dockerCli *command.DockerCli, info types.Info) error {
 	fprintfIfNotEmpty(dockerCli.Out(), "Logging Driver: %s\n", info.LoggingDriver)
 	fprintfIfNotEmpty(dockerCli.Out(), "Cgroup Driver: %s\n", info.CgroupDriver)
 
-	fmt.Fprintf(dockerCli.Out(), "Plugins: \n")
+	fmt.Fprintf(dockerCli.Out(), "Plugins:\n")
 	fmt.Fprintf(dockerCli.Out(), " Volume:")
 	fmt.Fprintf(dockerCli.Out(), " %s", strings.Join(info.Plugins.Volume, " "))
 	fmt.Fprintf(dockerCli.Out(), "\n")
@@ -130,6 +130,7 @@ func prettyPrintInfo(dockerCli *command.DockerCli, info types.Info) error {
 					fmt.Fprintf(dockerCli.Out(), "    %s: %s\n", entry.Protocol, entry.URL)
 				}
 			}
+			fmt.Fprintf(dockerCli.Out(), " Autolock Managers: %v\n", info.Swarm.Cluster.Spec.EncryptionConfig.AutoLockManagers)
 			fmt.Fprintf(dockerCli.Out(), " Root Rotation In Progress: %v\n", info.Swarm.Cluster.RootRotationInProgress)
 		}
 		fmt.Fprintf(dockerCli.Out(), " Node Address: %s\n", info.Swarm.NodeAddr)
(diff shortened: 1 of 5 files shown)


cli/command/system/info_test.go
```diff
@@ -0,0 +1,302 @@
...
+ Autolock Managers: true
+ Root Rotation In Progress: false
+ Node Address: 165.227.107.89
+ Manager Addresses:
+  165.227.107.89:2377`
+
+func newDockerCLI() (*command.DockerCli, *bytes.Buffer, *bytes.Buffer) {
+	var in, out, err bytes.Buffer
+	return command.NewDockerCli(ioutil.NopCloser(&in), &out, &err), &out, &err
+}
```
dnephin: We have a fake that we use for testing at `internal/test/` `NewFakeCli()`

You can change `prettyPrintInfo()` to accept a `dockerCli command.Cli` interface instead of the struct

cli/command/system/info_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+func TestInfoNoWarnings(t *testing.T) {
+	cli, stdout, stderr := newDockerCLI()
+
+	// without swarm
+	require.NoError(t, prettyPrintInfo(cli, sampleInfoNoSwarm))
+	result, err := ioutil.ReadAll(stdout)
+	require.NoError(t, err)
+	require.Equal(t, []byte(sampleTextOutputNoSwarm), result)
```
dnephin: I would recommend moving these sample texts to "golden files" (files in `./testdata` relative to this package directory) and using `gotestyourself/golden.Assert()` to compare them.  There are a bunch of example in this repo and the godoc is here: https://godoc.org/github.com/gotestyourself/gotestyourself/golden

This will make it easier to update and maintain the expected result, and it also uses `difflib` to give you a nice multiline diff when there are failures as long as you use the string version (not bytes). Comparing bytes gives you a diff of an array of uint8 which is impossible to debug.

cli/command/system/info_test.go
```diff
@@ -0,0 +1,302 @@
...
+	result, err = ioutil.ReadAll(stderr)
+	require.NoError(t, err)
+	require.Empty(t, result)
```
dnephin: Using this fake this would be `assert.Equal(t, "", cli.ErrBuffer().String())`

cli/command/system/info_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// with swarm
+	cli, stdout, stderr = newDockerCLI()
```
dnephin: Please make this a separate test function `TestPrettyPrintInfoWithSwarm()` which removes the need for this comment

cli/command/system/info_test.go
```diff
@@ -0,0 +1,302 @@
...
+ Node Address: 165.227.107.89
+ Manager Addresses:
+  165.227.107.89:2377`
+
+func newDockerCLI() (*command.DockerCli, *bytes.Buffer, *bytes.Buffer) {
+	var in, out, err bytes.Buffer
+	return command.NewDockerCli(ioutil.NopCloser(&in), &out, &err), &out, &err
+}
+
+func TestInfoNoWarnings(t *testing.T) {
```
dnephin: Following our [testing guidelines](https://github.com/docker/cli/blob/master/TESTING.md#details) (which we just merged), this should be `TestPrettyPrintInfoNoSwarm()`

cli/command/system/info_test.go
```diff
@@ -0,0 +1,302 @@
...
+	result, err = ioutil.ReadAll(stderr)
+	require.NoError(t, err)
+	require.Empty(t, result)
+
+	// with swarm
+	cli, stdout, stderr = newDockerCLI()
+	infoWithSwarm := sampleInfoNoSwarm
+	infoWithSwarm.Swarm = sampleSwarmInfo
+	expected := []byte(
+		strings.Replace(sampleTextOutputNoSwarm, "Swarm: inactive", sampleTextOutputJustSwarm, 1))
```
dnephin: I would just write out the exact file and compare using `golden` instead of doing a replace

cli/command/system/info.go
```diff
@@ -222,9 +223,11 @@ func prettyPrintInfo(dockerCli *command.DockerCli, info types.Info) error {
 	fprintfIfNotEmpty(dockerCli.Out(), "No Proxy: %s\n", info.NoProxy)
 
 	if info.IndexServerAddress != "" {
-		u := dockerCli.ConfigFile().AuthConfigs[info.IndexServerAddress].Username
-		if len(u) > 0 {
-			fmt.Fprintf(dockerCli.Out(), "Username: %v\n", u)
+		if dockerCli.ConfigFile() != nil {
```
dnephin: If this is just because of the test, I think you can revert this. Using the `FakeCli` should give you a default config file I believe.

cyli:
Thanks for the review @dnephin!  I think I've addressed all the comments, and added a test for the warnings since the golden testdata files makes that way easier.

PR #138, opened 2017-05-30
Add manifest command

clnperez:
[Enable](url) inspection (aka "shallow pull") of images' manifest info, and also the creation of manifest lists (aka "fat manifests").

The workflow for creating a manifest list will be:

`docker manifest create new-list-ref-name image-ref [image-ref...]`
`docker manifest annotate new-list-ref-name image-ref --os linux --arch arm`
`docker manifest push new-list-ref-name`

There is also a `manifest inspect` command to allow for a "shallow pull"
of an image's manifest: `docker manifest inspect
manifest-or-manifest_list`.

To be more in line with the existing external manifest tool, there is
also a `-v` option for inspect that will show information depending on
what the reference maps to (list or single manifest).

Signed-off-by: Christy Norman Perez <christy@linux.vnet.ibm.com>

**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

See https://github.com/docker/docker/pull/27455 for history.

TODO:
- [x] move/refactor tests from original pr to this

PR #472, opened 2017-08-24
docker trust: view, revoke, sign subcommands (experimental)

riyazdf:
This PR introduces the `docker trust` subcommand.  The command has been fleshed out in detail in [this design document](https://docs.google.com/a/docker.com/document/d/1JOBAlCDuf5JnpVLW54voGuAdsnMmSR2LIbs8fjBDSaM/edit?usp=sharing).  `docker trust` provides a direct notary signing experience beyond what is provided with the `DOCKER_CONTENT_TRUST` environment variable.

In this PR, we introduce three subcommands:
- `docker trust view`: displays signatures on image tags, and who signed them
- `docker trust revoke`: revokes signatures from signed image tag(s)
- `docker trust sign`: adds a signature to an image tag, pushing the image if it doesn't already exist

cc @ashfall and @eiais who helped implement this feature

</img>


docs/reference/commandline/trust_inspect.md
```diff
@@ -0,0 +1,131 @@
...
+```markdown
+Usage:  docker trust inspect [OPTIONS] IMAGE[:TAG]
+
+Display detailed information about keys and signatures
+
+```
+
+## Description
+
+Docker trust inspect provides detailed information on signed repositories.
```
mstanleyjones: Mark this up: `docker trust inspect`. If you don't feel comfortable starting the sentence with lowercase mono text, then `The `docker trust inspect` command...."

docs/reference/commandline/trust_inspect.md
```diff
@@ -0,0 +1,131 @@
...
+This includes all image tags that are signed, who signed them, and who can sign
```
mstanleyjones: Does it show untagged images? This kind of implies not.

docs/reference/commandline/trust_inspect.md
```diff
@@ -0,0 +1,131 @@
...
+new tags.
+
+By default, `docker trust inspect` will render results in a table.
```
mstanleyjones: s/will render/renders

docs/reference/commandline/trust_inspect.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+## Examples
+
+### Get details about signatures for a single image tag
+
+
+```bash
+$ docker trust inspect alpine:latest
+SIGNED TAG          DIGEST                                                             SIGNERS
```
mstanleyjones: Please add a newline between input and output.

docs/reference/commandline/trust_inspect.md
```diff
@@ -0,0 +1,131 @@
...
+latest              1072e499f3f655a032e88542330cf75b02e7bdf673278f701d7ba61629ee3ebe   (Repo Admin)
+
+Administrative keys for alpine:latest:
+Repository Key:	5a46c9aaa82ff150bb7305a2d17d0c521c2d784246807b2dc611f436a69041fd
+Root Key:	a2489bcac7a79aa67b19b96c4a3bf0c675ffdf00c6d2fabe1a5df1115e80adce
+```
+
+Note that the `SIGNED TAG` maps to the image tag itself, and associates to given image `DIGEST`. `SIGNERS` lists all entities who have signed.
```
mstanleyjones: s/Note that the/The

"and associates to given image" this phrase isn't parsing right for me. Maybe a rewrite?

docs/reference/commandline/trust_inspect.md
```diff
@@ -0,0 +1,131 @@
...
+
+The administrative keys listed specify the root key of trust, as well as the administrative repository key.
```
mstanleyjones: Maybe a little more explanation -- the administrator can add new signers, right?

docs/reference/commandline/trust_inspect.md
```diff
@@ -0,0 +1,131 @@
...
+
+If signers are set up for the repository via other `docker trust` commands, `docker trust inspect` will display them appropriately as a `SIGNER` and specify their `KEYS`:
```
mstanleyjones: s/will display/displays

docs/reference/commandline/trust_inspect.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+If the image tag is unsigned or unavailable, `docker trust inspect` will not display any signed tags.
+```
+$ docker trust inspect unsigned-img
+No signatures or cannot access unsigned-img
+```
+
+However, if other tags are signed in the same image repository, `docker trust inspect` will report relevant key information.
+```
```
mstanleyjones: Need a newline on either side of this code block. Also make sure to hint `bash` after the three back-ticks if it's a command-line example.

docs/reference/commandline/trust_inspect.md
```diff
@@ -0,0 +1,131 @@
...
+edge                79d50d15bd7ea48ea00cf3dd343b0e740c1afaa8e899bee475236ef338e1b53b   (Repo Admin)
+latest              1072e499f3f655a032e88542330cf75b02e7bdf673278f701d7ba61629ee3ebe   (Repo Admin)
+
+Administrative keys for alpine:
+Repository Key:	5a46c9aaa82ff150bb7305a2d17d0c521c2d784246807b2dc611f436a69041fd
+Root Key:	a2489bcac7a79aa67b19b96c4a3bf0c675ffdf00c6d2fabe1a5df1115e80adce
+```
+
+Here's an example with signers that are set up by `docker trust` commands:
+```bash
```
mstanleyjones: Newlines

docs/reference/commandline/trust_revoke.md
```diff
@@ -0,0 +1,131 @@
...
+
+```markdown
+Usage:  docker trust revoke [OPTIONS] IMAGE[:TAG]
+
+Remove trust for an image
+
+Options:
+      --help   Print usage
+  -y, --yes    Answer yes to the removal question (no confirmation)
+
```
mstanleyjones: Remove extra newline here

docs/reference/commandline/trust_revoke.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+## Description
+
+Docker trust revoke removes signatures from tags in signed repositories.
```
mstanleyjones: Again, "The `docker trust revoke` command"

(there is no `Docker` command)

docs/reference/commandline/trust_revoke.md
```diff
@@ -0,0 +1,131 @@
...
+SIGNER              KEYS
+alice               05e87edcaecb
+bob                 5600f5ab76a2
+
+Administrative keys for example/trust-demo:
+Repository Key:	ecc457614c9fc399da523a5f4e24fe306a0a6ee1cc79a10e4555b3c6ab02f71e
+Root Key:	3cb2228f6561e58f46dbc4cda4fcaff9d5ef22e865a94636f82450d1d2234949
+```
+
+When alice runs `docker trust revoke`  like so:
```
mstanleyjones: When `alice`, one of the signers, runs...

docs/reference/commandline/trust_revoke.md
```diff
@@ -0,0 +1,131 @@
...
+
+```bash
+$ docker trust revoke example/trust-demo:red
+Enter passphrase for delegation key with ID 27d42a8:
+Successfully deleted signature for example/trust-demo:red
+```
+
+the said tag gets removed from the list of released tags:
```
mstanleyjones: After revocation, the tag is removed from...

docs/reference/commandline/trust_revoke.md
```diff
@@ -0,0 +1,131 @@
...
+SIGNER              KEYS
+alice               05e87edcaecb
+bob                 5600f5ab76a2
+
+Administrative keys for example/trust-demo:
+Repository Key:	ecc457614c9fc399da523a5f4e24fe306a0a6ee1cc79a10e4555b3c6ab02f71e
+Root Key:	3cb2228f6561e58f46dbc4cda4fcaff9d5ef22e865a94636f82450d1d2234949
+```
+
+When alice runs `docker trust revoke`  like so:
```
mstanleyjones: As the comment above

docs/reference/commandline/trust_revoke.md
```diff
@@ -0,0 +1,131 @@
...
+
+```bash
+$ docker trust revoke example/trust-demo
+Please confirm you would like to delete all signature data for example/trust-demo? (y/n) y
+Enter passphrase for delegation key with ID 27d42a8:
+Successfully deleted signature for example/trust-demo
+```
+
+the all tags that have alice's signature on them get removed from the list of released tags:
```
mstanleyjones: s/the all/All

docs/reference/commandline/trust_revoke.md
```diff
@@ -0,0 +1,131 @@
...
+
+```bash
+$ docker trust inspect example/trust-demo
+
+No signatures for example/trust-demo
+
+
+List of signers and their KeyIDs:
```
mstanleyjones: and the IDs of their `KEYS`:

docs/reference/commandline/trust_sign.md
```diff
@@ -0,0 +1,179 @@
...
+```markdown
+Usage:  docker trust sign [OPTIONS] IMAGE:TAG
+
+Sign an image
+
+```
+
+## Description
+
+Docker trust sign adds signatures to tags to create signed repositories.
```
mstanleyjones: As above

docs/reference/commandline/trust_sign.md
```diff
@@ -0,0 +1,179 @@
...
+$ docker trust inspect example/trust-demo
+SIGNED TAG          DIGEST                                                             SIGNERS
+v1                  c24134c079c35e698060beabe110bb83ab285d0d978de7d92fed2c8c83570a41   (Repo Admin)
+
+Administrative keys for example/trust-demo:
+Repository Key:	36d4c3601102fa7c5712a343c03b94469e5835fb27c191b529c06fd19c14a942
+Root Key:	246d360f7c53a9021ee7d4259e3c5692f3f1f7ad4737b1ea8c7b8da741ad980b
+```
+
+We can sign a new tag with `docker trust sign`:
```
mstanleyjones: s/we can sign/Sign

docs/reference/commandline/trust_sign.md
```diff
@@ -0,0 +1,179 @@
...
+
+```bash
+$ docker trust sign example/trust-demo:v2
```
mstanleyjones: Add a newline between input and output

docs/reference/commandline/trust_sign.md
```diff
@@ -0,0 +1,179 @@
...
+c69f806905c2: Layer already exists
+582f327616f1: Layer already exists
+a3fbb648f0bd: Layer already exists
+5eac2de68a97: Layer already exists
+8d4d1ab5ff74: Layer already exists
+v2: digest: sha256:8f6f460abf0436922df7eb06d28b3cdf733d2cac1a185456c26debbff0839c56 size: 1787
+Signing and pushing trust metadata
+Enter passphrase for repository key with ID 36d4c36:
+Successfully signed "docker.io/example/trust-demo":v2
+```
```
mstanleyjones: Newlines before and after

docs/reference/commandline/trust_sign.md
```diff
@@ -0,0 +1,179 @@
...
+SIGNER              KEYS
+alice               05e87edcaecb
+bob                 5600f5ab76a2
+
+Administrative keys for example/trust-demo:
+Repository Key:	ecc457614c9fc399da523a5f4e24fe306a0a6ee1cc79a10e4555b3c6ab02f71e
+Root Key:	3cb2228f6561e58f46dbc4cda4fcaff9d5ef22e865a94636f82450d1d2234949
+```
+
+We can sign a new tag with `docker trust sign`:
```
mstanleyjones: As above

docs/reference/commandline/trust_sign.md
```diff
@@ -0,0 +1,179 @@
...
+8a5132998025: Layer already exists
+aca233ed29c3: Layer already exists
+e5d2f035d7a4: Layer already exists
+v1: digest: sha256:74d4bfa917d55d53c7df3d2ab20a8d926874d61c3da5ef6de15dd2654fc467c4 size: 1357
+Signing and pushing trust metadata
+Enter passphrase for delegation key with ID 27d42a8:
+Successfully signed "docker.io/example/trust-demo":v1
+```
+
+`docker trust inspect` should now list the new signature:
```
mstanleyjones: s/should now list/lists

docs/reference/commandline/trust_sign.md
```diff
@@ -0,0 +1,179 @@
...
+bob                 5600f5ab76a2
+
+Administrative keys for example/trust-demo:
+Repository Key:	ecc457614c9fc399da523a5f4e24fe306a0a6ee1cc79a10e4555b3c6ab02f71e
+Root Key:	3cb2228f6561e58f46dbc4cda4fcaff9d5ef22e865a94636f82450d1d2234949
+```
+
+## Initialize a new repo and sign a tag
+
+When signing an image on a repo for the first time, `docker trust sign` sets up new keys and then signs the image.
```
mstanleyjones: s/and then signs/before signing

vendor/github.com/docker/notary/README.md
```diff
@@ -78,18 +78,21 @@ to use `notary` with Docker images.
 
 ## Building Notary
 
-Prerequisites:
+Note that our [latest stable release](https://github.com/docker/notary/releases) is at the head of the
```
mstanleyjones: s/Note that our/Notary's

vendor/github.com/docker/notary/README.md
```diff
@@ -78,18 +78,21 @@ to use `notary` with Docker images.
...
     - Ubuntu: `apt-get install libltdl-dev`
     - CentOS/RedHat: `yum install libtool-ltdl-devel`
     - Mac OS ([Homebrew](http://brew.sh/)): `brew install libtool`
 
-Run `make binaries`, which creates the Notary Client CLI binary at `bin/notary`.
-Note that `make binaries` assumes a standard Go directory structure, in which
+Run `make client`, which creates the Notary Client CLI binary at `bin/notary`.
+Note that `make client` assumes a standard Go directory structure, in which
 Notary is checked out to the `src` directory in your `GOPATH`. For example:
 ```
```
mstanleyjones: Newlines

vendor/github.com/docker/notary/README.md
```diff
@@ -98,3 +101,5 @@ $GOPATH/
             docker/
                 notary/
 ```
+
+To build the server and signer, please run `docker-compose build`.
```
mstanleyjones: s/please//

docs/reference/commandline/trust_inspect.md
eiais: This will not display unsigned images.

docs/reference/commandline/trust_sign.md
riyazdf: added!

docs/reference/commandline/trust_revoke.md
riyazdf: this is direct output of `docker trust inspect` - are you suggesting we change this?

vendor/github.com/docker/notary/README.md
riyazdf: vendored code - we can address this in docker/notary

vendor/github.com/docker/notary/README.md
riyazdf: vendored code - we can address this in docker/notary

vendor/github.com/docker/notary/README.md
riyazdf: vendored code - we can address this in docker/notary

riyazdf:
thank you @mstanleyjones for the review!  Just updated with the fixes, and left a comment one potential command output change.

docs/reference/commandline/trust_revoke.md
mstanleyjones: Well, I'm not sure, `KeyID` doesn't appear anywhere else ever that I can see. is it actually a thing?

cli/command/trust/revoke.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	defer clearChangeList(notaryRepo)
+	if err := revokeTestHelper(notaryRepo, tag); err != nil {
+		return fmt.Errorf("could not remove signature for %s: %s", remote, err)
+	}
+	fmt.Fprintf(cli.Out(), "Successfully deleted signature for %s\n", remote)
+	return nil
+}
+
+func revokeTestHelper(notaryRepo *client.NotaryRepository, tag string) error {
```
dnephin: What does `TestHelper` refer to here? Could this be `revokeSignature()` ?

cli/command/trust/revoke.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err = clearChangeList(notaryRepo); err != nil {
+		return err
+	}
+	defer clearChangeList(notaryRepo)
+	if err := revokeTestHelper(notaryRepo, tag); err != nil {
+		return fmt.Errorf("could not remove signature for %s: %s", remote, err)
```
dnephin:
```
errors.Wrapf(err, "could not remove...")
```

from `pkg/errors`

There's a couple other instances of these as well

cli/command/trust/inspect_test.go
```diff
@@ -0,0 +1,380 @@
...
+		cmd.SetOutput(ioutil.Discard)
+		testutil.ErrorContains(t, cmd.Execute(), tc.expectedError)
+	}
+}
+
+func TestTrustInfo(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{})
+	cmd := newInspectCommand(cli)
+	cmd.SetArgs([]string{"alpine"})
+	assert.NoError(t, cmd.Execute())
```
dnephin: Does this contact the "real" notary server?  Is that what you were referring to by fixture images in the description?

I think maybe `trust.GetNotaryRepository()` should be changed to `Cli.GetNotaryClient()` (or something like that) which would return an interface instead of a `client.NotaryRepostiory` struct . Once it's on the `command.Cli` interface it makes it really easy to fake it in tests.

dnephin:
I think the design looks good. It sounds like you've already done a lot of work on the design before opening this PR so I mostly focused on the code.

cli/command/image/trust.go
```diff
@@ -28,8 +28,8 @@ type target struct {
 	size   int64
 }
 
-// trustedPush handles content trust pushing of an image
-func trustedPush(ctx context.Context, cli command.Cli, repoInfo *registry.RepositoryInfo, ref reference.Named, authConfig types.AuthConfig, requestPrivilege types.RequestPrivilegeFunc) error {
+// TrustedPush handles content trust pushing of an image
+func TrustedPush(ctx context.Context, cli command.Cli, repoInfo *registry.RepositoryInfo, ref reference.Named, authConfig types.AuthConfig, requestPrivilege types.RequestPrivilegeFunc) error {
```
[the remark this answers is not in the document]
riyazdf: I think we certainly can for the `GetSignableRoles` function and so I've moved over.

This one is a little more tricky because it calls `imagePushPrivileged` - can we export that function?

cli/command/trust/inspect_test.go
riyazdf: it does contact a real notary server. Is the idea that `Cli.GetNotaryClient` would wrap a notary repo with a subset of the API?  It seems that could certainly work

dnephin: It doesn't necessary have to wrap it with new structs, it could just return an Interface type instead of a struct, so that we can use a fake during testing.

riyazdf:
@dnephin: thank you for the review!  I've addressed all comments except for the following:
- for the `notary` tag you suggest: there isn't a "canonical" notary server or a way to currently pre-configure one per-image (you could in theory contact a different one for different images) - so I'm inclined to add this later if you think that makes sense
- I've flattened the map for the admin keys.  Haven't done so for signers and the signatures because I want to make sure I don't mess up the table printing 😅 
- totally agree for mocking out the notary client behind the CLI so we can mock it.  Will get to this.

cli/command/trust/helpers_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+func TestGetTag(t *testing.T) {
+	ref, err := reference.ParseNormalizedNamed("ubuntu@sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2")
+	assert.NoError(t, err)
+	tag, err := getTag(ref)
+	assert.EqualError(t, err, "cannot use a digest reference for IMAGE:TAG")
+
+	ref, err = reference.ParseNormalizedNamed("alpine:latest")
+	assert.NoError(t, err)
+	tag, err = getTag(ref)
```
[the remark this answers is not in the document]
riyazdf: added!

dnephin:
> so I'm inclined to add [the notary tag] later if you think that makes sense

Sounds good. I think we can forget about the tag, I can't think of a good reason to hide the commands.

cli/command/image/trust.go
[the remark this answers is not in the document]
dnephin: Ya, I think it makes sense to export it. There's nothing in there that's really specific to the command routing. The only change I would make is to accept a `client.ImageAPIClient` instead of the full `command.Cli`

cli/command/trust/inspect_test.go
riyazdf: I've been trying to wrap my head around this but I'm not sure if it makes sense since a `NotaryRepository` is custom per-image.  Is that OK in the context of `command.Cli`, or are assuming that a `command.Cli` may persist for longer than a single image command?

dnephin: That does make it more difficult. We ran into this problem in #138 with the distribution client. In that case I wrapped the whole thing in an interface that was more appropriate for our use case in the cli (https://github.com/docker/cli/pull/138/files#diff-cb6e6f1ab385cb866626ee63bfa1a273R29).

It sounds like notary is using a similar pattern. 

I think something like this in `cli/command/cli.go` should work:

```go

func (cli *DockerCli) NotaryClient(ref trust.ImageRefAndAuth) trust.Repository {
    ...
}
```
where `trust.Repository` is the local interface we define (maybe in `cli/trust`).

We can still have multiple instances per `command.Cli`. In tests we can return a `fakeTrustRepository`.

riyazdf: awesome, that would work.  I'll implement this 👍

docs/reference/commandline/trust_revoke.md
riyazdf: it is a thing in notary, but we'd prefer not to expose it in this docker command - so I've updated this.  Thanks for spotting this!

riyazdf:
Some small updates: I've made a couple of tweaks to address small discrepancies in CLI output and we also have an interface for a trust repository incoming from Notary https://github.com/docker/notary/pull/1220 that will help provide testable mocks for this PR.  I'll update here once that is merged and vendored.

riyazdf:
@dnephin: I've rebased and vendored in the notary client interface and exposed it as `cli.NotaryClient`.  I'm rewriting the tests now to used mocks of that, but figured I'd push the update so you can see how the cli changes look

cli/command/cli.go
```diff
@@ -151,6 +158,89 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
...
+func (cli *DockerCli) NotaryClient(imgRefAndAuth ImageRefAndAuth, forWrite bool) (notaryclient.Repository, error) {
+	actions := []string{"pull"}
+	if forWrite {
+		actions = append(actions, "push")
+	}
+	return trust.GetNotaryRepository(cli.In(), cli.Out(), UserAgent(), imgRefAndAuth.RepoInfo(), imgRefAndAuth.AuthConfig(), actions...)
+}
+
+// ImageRefAndAuth contains all reference information and the auth config for an image request
+type ImageRefAndAuth struct {
```
dnephin: I think the `cli/trust` package might be a better place for this type

cli/command/cli.go
```diff
@@ -151,6 +158,89 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	return nil
 }
 
+// NotaryClient provides a Notary Repository to interact with signed metadata for an image
+func (cli *DockerCli) NotaryClient(imgRefAndAuth ImageRefAndAuth, forWrite bool) (notaryclient.Repository, error) {
```
dnephin: I always try to avoid boolean arguments because from the caller it's not really clear what `true` or `false` means. Maybe this could takes a `[]string` of actions, and there could be two predefined slices in `cli/trust` 

```go

var (
  ActionsPullOnly = []string{"pull"}
  ActionsPushAndPull = []string{"pull", "push"}
)
```
So from the caller it will be obvious what is being requested.

cli/command/trust/cli_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/docker/cli/cli/config/configfile"
+	"github.com/docker/cli/cli/trust"
+	"github.com/docker/docker/client"
+	notaryclient "github.com/docker/notary/client"
+)
+
+type notaryClientFuncType func(imgRefAndAuth trust.ImageRefAndAuth, actions []string) (notaryclient.Repository, error)
+
+// FakeCli emulates the default DockerCli
+type FakeCli struct {
```
dnephin: Isn't this a dupliate of `internal/test/cli.go:FakeCli`? I think all we need is `NotaryClient()` and a `SetNotaryClient(...)` added to the existing FakeCli

cli/command/trust/inspect_test.go
```diff
@@ -0,0 +1,661 @@
...
+
+	// Check for the signed tag headers
+	assert.Contains(t, cli.OutBuffer().String(), "SIGNED TAG")
+	assert.Contains(t, cli.OutBuffer().String(), "DIGEST")
+	assert.Contains(t, cli.OutBuffer().String(), "SIGNERS")
+	// Check for the signer headers
+	assert.Contains(t, cli.OutBuffer().String(), "Administrative keys for alpine:")
+	assert.Contains(t, cli.OutBuffer().String(), "(Repo Admin)")
+	// no delegations on this repo
+	assert.NotContains(t, cli.OutBuffer().String(), "List of signers and their keys:")
```
dnephin: Could this use `gotestyourself/golden.Assert()` to be more strict about the expected output?

cli/command/trust/inspect_test.go
```diff
@@ -0,0 +1,661 @@
...
+		},
+		Name: data.CanonicalTargetsRole,
+	}
+	targetsRoleWithSigs := client.RoleWithSignatures{Role: targetsRole, Signatures: nil}
+	assert.Equal(t, "Repository Key:\tabc, key11, key99\n", formatAdminRole(targetsRoleWithSigs))
+}
+
+// Sample mock CLI interfaces
+
+func getOfflineNotaryRepository(imgRefAndAuth trust.ImageRefAndAuth, actions []string) (client.Repository, error) {
```
dnephin: We have a convention of putting these fakes in a `client_test.go` file in the package

riyazdf: moved!

cli/command/trust/cli_test.go
riyazdf: sounds good, done!

cli/command/trust/inspect_test.go
riyazdf: I haven't done this yet but can do so - out of curiosity, what's more strict about `gotestyourself/golden.Assert()`?

cli/command/cli.go
```diff
@@ -151,6 +157,51 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
...
 }
 
+// NotaryClient provides a Notary Repository to interact with signed metadata for an image
+func (cli *DockerCli) NotaryClient(imgRefAndAuth trust.ImageRefAndAuth, actions []string) (notaryclient.Repository, error) {
+	return trust.GetNotaryRepository(cli.In(), cli.Out(), UserAgent(), imgRefAndAuth.RepoInfo(), imgRefAndAuth.AuthConfig(), actions...)
+}
+
+// GetImageReferencesAndAuth retrieves the necessary reference and auth information for an image name
+// as a ImageRefAndAuth struct
+func GetImageReferencesAndAuth(ctx context.Context, cli Cli, imgName string) (*trust.ImageRefAndAuth, error) {
```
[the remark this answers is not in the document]
riyazdf: ah ok, thanks for clarifying!  This will cause a package import cycle because of the call to `command.ResolveAuthConfig`, unless we're ok with shifting that code around or duplicating it?

cli/command/trust/inspect_test.go
dnephin: golden.Assert() will do an exact string match of the entire string instead of just checking that specific substrings exist somewhere. It will also give a nice diff output of the difference if the assertion fails, which makes it a bit easier to debug failures.

cli/command/cli.go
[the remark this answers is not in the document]
dnephin: Ah, right! I ran into that problem with the manifest command as well. I solved it like this (https://github.com/docker/cli/pull/138/files#diff-b4a7ebaa86bd3765289ff28c6934850fR122):

```go
func (cli *DockerCli) RegistryClient() registryclient.RegistryClient {
 	resolver := func(ctx context.Context, index *registrytypes.IndexInfo) types.AuthConfig {
 		return ResolveAuthConfig(ctx, cli, index)
 	}
 	return registryclient.NewRegistryClient(resolver, UserAgent())
 }
```

`GetImageReferencecsAndAuth()` could accept a `resolver func(ctx context.Context, index *registrytypes.IndexInfo)` as an arg instead of the `cli Cli` parameter.

cli/command/trust/inspect_test.go
riyazdf: awesome, added!

cli/command/cli.go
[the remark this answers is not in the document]
riyazdf: ok this worked great 👍

cli/command/image/trust_test.go
```diff
@@ -55,3 +60,25 @@ func TestNonOfficialTrustServer(t *testing.T) {
...
+	assert.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	notaryRepo, err := client.NewFileCachedRepository(tmpDir, "gun", "https://localhost", nil, passphrase.ConstantRetriever("password"), trustpinning.TrustPinConfig{})
+	target := client.Target{}
+	err = AddTargetToAllSignableRoles(notaryRepo, &target)
+	assert.EqualError(t, err, "client is offline")
+}
+
+func TestGetSignableRolesError(t *testing.T) {
```
dnephin: This is testing a function in `cli/trust` so should be moved to that package

riyazdf:
thank you @dnephin for your continuous review!  Just updated the tests and rebased.

cli/command/formatter/trust_test.go
```diff
@@ -0,0 +1,238 @@
...
+			},
+			"ashwini, kyle, riyaz",
+			ctx.Signers,
+		},
+		// alphabetic signing on Signers
+		{
+			trustTagContext{
+				s: SignedTagInfo{Name: trustedTag,
+					Digest:  digest,
+					Signers: []string{"riyaz", "kyle", "ashwini"},
```
mstanleyjones: I know this is just a test, but is it ok to use real people's names in the code like this?

cli/command/trust/inspect_test.go
```diff
@@ -0,0 +1,433 @@
...
+}
+
+func TestTrustInspectCommandErrors(t *testing.T) {
+	testCases := []struct {
+		name          string
+		args          []string
+		expectedError string
+	}{
+		{
+			name:          "not-enough-args",
```
mstanleyjones: Do you really need two tests for this??

riyazdf: one test is with 0 args and the other is with > 1 - if you'd like we can consolidate into a single != 1 arg test but I do like the two cases in the event that `ExactArgs` were to somehow change

vieux:
@johnstep can you please take a look at this ? to make sure it works well on windows.

johnstep:
Yes, will look today.

docs/reference/commandline/trust_sign.md
```diff
@@ -0,0 +1,180 @@
...
+
+Given an image:
+
+```bash
+$ docker trust inspect example/trust-demo
+
+No signatures for example/trust-demo
+
+
+List of signers and their keys:
```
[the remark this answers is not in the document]
riyazdf: This is per-repository output, so only information for `example/trust-demo` is displayed.

Signers are defined per-repo so it's not really possible or helpful to have a global list of signers.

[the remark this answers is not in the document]
thaJeztah: clear, thanks! perhaps we should include that in the output (signers for [repository])

riyazdf:
@thaJeztah: thank you for the very thorough review and thoughtful writeup!  I've replied inline.  Please let me know if our (cc @ashfall @eiais) reasoning doesn't make sense.

I think the most pressing issue that we should address in this PR is renaming `docker trust inspect` to `docker trust view` or similar.  There are a few other issues you've raised that definitely deserve followup PRs and we can help with those.

> docker trust inspect doesn't default to :latest

For Content Trust (and `docker trust`) most operations affect a full repository, as all signers and keys are assigned per-repo - so my gut is that we should have the default information display per-repo.  It's also likely that repos may have signed tags, but not a signed `latest` tag, which would make `docker trust inspect <repo>` confusing if no signed tags showed up.


> sha256: prefix missing for digests

We intentionally omitted this since it is repetitive information at present.  I'd be inclined to add the prefix once we support multi-hashing - WDYT?


> No REPOSITORY column in output of docker trust inspect

`docker trust inspect` operates over a single repository.  Operating over multiple repositories would be very confusing because the keys and signers are per-repo.  So for that reason I do not think we should have a `REPOSITORY` column.

> Is there a way to show unsigned tags in a repository?

We intentionally omitted unsigned tags because they are unpullable with `DOCKER_CONTENT_TRUST=1`. 

> Output format of docker trust inspect

We've discussed this at length with @dnephin: IIRC we agreed to potentially add JSON output controlled by a flag in the future - though as you mention this isn't consistent with other `inspect` commands.

We had initially named this command `docker trust info` - if you prefer we revert to that name I wouldn't be opposed; I don't think JSON output would be very helpful for this information by default, I'd like it to be an optional `--json` or similar flag if we deem it desirable.  `docker trust view` could work as well.

> docker trust sign and pushing

If the image doesn't exist: `docker trust sign` will push and sign the image because a signature without associated image data is not of much use because you won't be able to pull down or run the image.

As for the ordering of push and signing: this mimics the behavior users are accustomed to from `DOCKER_CONTENT_TRUST=1 docker push` and so we decided to keep this consistent.  Also, if there is a failure during the image push, we won't sign anything because there is no content to reference.

> docker trust sign alternative keys?

We're actually just about to follow up with `docker trust key-*` commands!  `docker trust sign` could also take on a `--key` flag in the future if we deem this useful.

> Unauthorized shows HTTP status code

This is also an issue in DCT:
```
$ DOCKER_CONTENT_TRUST=1 docker pull foo:latest
you are not authorized to perform this operation: server returned 401.
```

I'm happy to update this but I'd prefer this to be a separate PR to update both errors for consistency.

> Instructions could use some formatting

This is also from DCT and Notary, but I think this is necessary. It is a lot of text but it is interactive and for entering sensitive passphrases.

> Quotes between repo and tag
> ...
> In the last line, there's quotes around the repository, and the :latest tag is printed outside of the quotes:
> Successfully signed "docker.io/thajeztah/testing":latest

This is an issue across both DCT and `docker trust`: I'd be more than happy to update both in a followup.

> Also wondering if it should print the digest that was signed as well.

It's printed a couple of lines above.

cc @ruchitarr

docs/reference/commandline/trust_sign.md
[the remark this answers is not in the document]
riyazdf: :+1: I can update to `List of signers and their keys for example/trust-demo:`

riyazdf:
@thaJeztah: just pushed an update to rename to `docker trust view` and add the suggested repository output in the signer keys table.

Also, from our offline discussion (cc @friism and @mstanleyjones who also discussed) I've marked this feature as experimental in the CLI command description and made a note in the docs.

mstanleyjones:
Make sure you change the title before merge.

riyazdf:
@mstanleyjones: updated title!

dnephin:
BTW, I think the codecov report isn't being sent to github because of circleCI. I think you have it enabled on your fork, which apparently confuses codecov.

The report is here: https://codecov.io/gh/riyazdf/cli/tree/941f7cb8e28996804da756a2e54ff1cb14ea1945 but the diff doesn't work.

It looks like `cli/command/trust` has around 70% coverage, which SGTM

johnstep:
I verified that signing and viewing works from Windows.

PR #416, opened 2017-08-04
[WIP pending moby/moby] - LCOW: Add --platform to `docker build`

jhowardmsft:
Signed-off-by: John Howard jhoward@microsoft.com

This is the matching PR (and dependent on being merged first to resolve the currently hacked vendoring) to https://github.com/moby/moby/pull/34401, as described in https://github.com/moby/moby/issues/33850 to support Linux containers on Windows (aka `LCOW`)

PR #474, opened 2017-08-27
LCOW: CLI changes to add platform flag - pull, run, create and build

jhowardmsft:
Signed-off-by: John Howard jhoward@microsoft.com

Replacement for https://github.com/docker/cli/pull/416.  This adds the CLI flag changes to match those described in https://github.com/moby/moby/issues/34617, and PR'd to moby/moby in https://github.com/moby/moby/pull/34642.

It adds `--platform` to `pull`, `run`, `create`, and `build`. Note `import` requires further daemon changes so has not been done yet (will be in a further PR once the daemon has been updated).

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -29,6 +29,7 @@ type runOptions struct {
 	sigProxy   bool
 	name       string
 	detachKeys string
+	platform   string
 }
 
 // NewRunCommand create a new `docker run` command
@@ -62,6 +63,7 @@ func NewRunCommand(dockerCli command.Cli) *cobra.Command {
 	// with hostname
 	flags.Bool("help", false, "Print usage")
 
+	command.AddPlatformFlag(flags, &opts.platform)
 	command.AddTrustVerificationFlags(flags)
 	copts = addFlags(flags)
 	return cmd
@@ -160,7 +162,7 @@ func runContainer(dockerCli command.Cli, opts *runOptions, copts *containerOptio
 
 	ctx, cancelFun := context.WithCancel(context.Background())
 
-	createResponse, err := createContainer(ctx, dockerCli, containerConfig, opts.name)
+	createResponse, err := createContainer(ctx, dockerCli, containerConfig, opts.name, opts.platform)
 	if err != nil {
 		reportError(stderr, cmdPath, err.Error(), true)
 		return runStartContainerErr(err)
--- a/cli/command/image/trust.go
+++ b/cli/command/image/trust.go
@@ -180,7 +180,7 @@ func imagePushPrivileged(ctx context.Context, cli command.Cli, authConfig types.
 }
 
 // trustedPull handles content trust pulling of an image
-func trustedPull(ctx context.Context, cli command.Cli, imgRefAndAuth trust.ImageRefAndAuth) error {
+func trustedPull(ctx context.Context, cli command.Cli, imgRefAndAuth trust.ImageRefAndAuth, platform string) error {
 	refs, err := getTrustedPullTargets(cli, imgRefAndAuth)
 	if err != nil {
 		return err
@@ -202,7 +202,7 @@ func trustedPull(ctx context.Context, cli command.Cli, imgRefAndAuth trust.Image
 		if err != nil {
 			return err
 		}
-		if err := imagePullPrivileged(ctx, cli, updatedImgRefAndAuth, false); err != nil {
+		if err := imagePullPrivileged(ctx, cli, updatedImgRefAndAuth, false, platform); err != nil {
 			return err
 		}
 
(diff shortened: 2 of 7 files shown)


cli/command/commands/commands.go
```diff
@@ -87,6 +87,7 @@ func AddCommands(cmd *cobra.Command, dockerCli *command.DockerCli) {
 		hide(container.NewLogsCommand(dockerCli)),
 		hide(container.NewPauseCommand(dockerCli)),
 		hide(container.NewPortCommand(dockerCli)),
+		hide(image.NewPullCommand(dockerCli)),
```
dnephin: looks like this moved for no reason? now it's not sorted

cli/command/container/run.go
```diff
@@ -123,6 +128,7 @@ func runContainer(dockerCli *command.DockerCli, opts *runOptions, copts *contain
 	hostConfig := containerConfig.HostConfig
 	stdout, stderr := dockerCli.Out(), dockerCli.Err()
 	client := dockerCli.Client()
+	ociPlatform := system.ParsePlatform(opts.platform)
```
dnephin: I think this should be parsed by the flag instead after flag parsing. To accomplish this you can create a new `PlatformOpt` type in `opts/platform.go` with a `func Value() spec.Platform` to return the parsed type.

The field in the options struct will be a `PlatformOpt` instead of a `string`

cli/command/container/run.go
```diff
@@ -59,6 +61,9 @@ func NewRunCommand(dockerCli *command.DockerCli) *cobra.Command {
 	flags.StringVar(&opts.name, "name", "", "Assign a name to the container")
 	flags.StringVar(&opts.detachKeys, "detach-keys", "", "Override the key sequence for detaching a container")
 
+	flags.StringVar(&opts.platform, "platform", "", "Set platform if server is multi-platform capable")
+	flags.SetAnnotation("platform", "version", []string{"1.32"})
```
dnephin: Since this is going to be repeated in a bunch of commands how about we move this to 

```go
func addPlatformFlag(flags *pflag.FlagSet, opt *PlatformOpt)
```

in `cli/flags/common.go`


There will also need to be another annotation for hiding based on the header:

```go
flags.SetAnnotation("platform", "platform", nil)
```

The code to hide flags based on annotations is in `cmd/docker/docker.go`

cli/command/image/cmd.go
```diff
@@ -20,8 +20,8 @@ func NewImageCommand(dockerCli command.Cli) *cobra.Command {
 		NewHistoryCommand(dockerCli),
 		NewImportCommand(dockerCli),
 		NewLoadCommand(dockerCli),
-		NewPullCommand(dockerCli),
 		NewPushCommand(dockerCli),
+		NewPullCommand(dockerCli),
```
dnephin: unsorted

cli/command/image/trust.go
```diff
@@ -296,7 +298,10 @@ func trustedPull(ctx context.Context, cli command.Cli, repoInfo *registry.Reposi
 }
 
 // imagePullPrivileged pulls the image and displays it to the output
-func imagePullPrivileged(ctx context.Context, cli command.Cli, authConfig types.AuthConfig, ref string, requestPrivilege types.RequestPrivilegeFunc, all bool) error {
+func imagePullPrivileged(ctx context.Context, cli command.Cli, authConfig types.AuthConfig, ref string, requestPrivilege types.RequestPrivilegeFunc, all bool, platform specs.Platform) error {
+	if platform.OS == "" && os.Getenv("DOCKER_DEFAULT_PLATFORM") != "" {
+		platform.OS = os.Getenv("DOCKER_DEFAULT_PLATFORM")
```
dnephin: This should be handled somewhere else before it gets into the pull handling. Possibly as part of the `PlatformOpt` type. It can be used as a default value for `Value()`

jhowardmsft:
Rebased and reworked based on https://github.com/moby/moby/pull/34642 and addressed most of the comments. Still a work in progress until 34642 is merged in moby/moby as the vendoring will remain hacked.

jhowardmsft:
Rebased to address conflicts, and reworked to use two utility functions rather than replicate the same code in a few places.

Ready for review, removed WIP from the title, updated the commit comment and the description of the PR.  @vdemeester can this be moved to code review now?

jhowardmsft:
@johnstep PTAL too. Thanks.

cli/command/utils.go
```diff
@@ -117,3 +118,18 @@ func PruneFilters(dockerCli Cli, pruneFilters filters.Args) filters.Args {
 
 	return pruneFilters
 }
+
+// PlatformFromEnvironment sets the platform from an environment variable
+// if it isn't otherwise populated.
+func PlatformFromEnvironment(platform string) string {
```
dnephin: Since this is effectively setting a default value for platform I think the environment variable should be read as the default value in `AddPlatformFlag()` instead of exporting this function:

```go
flags.StringVar(target, "platform", os.Getenv("DOCKER_DEFAULT_PLATFORM"), ...)
```

jhowardmsft: That makes sense. Updated.

For hiding the flag, it is going to require getting consensus on https://github.com/moby/moby/pull/34628 first, which I don't think there is among other maintainers yet.

johnstep:
This should probably be hidden for a daemon without experimental.

jhowardmsft:
@johnstep - OK, added that annotation in utils.go::AddPlatformFlag() and verified it operates as expected.

jhowardmsft:
Is this OK to merge with 2 LGTM's?

PR #477, opened 2017-08-28
A few tests for container command

dnephin:
And cleanup some for `exec` and `attach`

### DIFF
--- a/cli/command/container/client_test.go
+++ b/cli/command/container/client_test.go
@@ -8,12 +8,32 @@ import (
 
 type fakeClient struct {
 	client.Client
-	containerInspectFunc func(string) (types.ContainerJSON, error)
+	inspectFunc     func(string) (types.ContainerJSON, error)
+	execInspectFunc func(execID string) (types.ContainerExecInspect, error)
+	execCreateFunc  func(container string, config types.ExecConfig) (types.IDResponse, error)
 }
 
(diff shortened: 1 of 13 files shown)


cli/command/container/client_test.go
```diff
@@ -1,19 +1,155 @@
...
 func (cli *fakeClient) ContainerInspect(_ context.Context, containerID string) (types.ContainerJSON, error) {
-	if cli.containerInspectFunc != nil {
-		return cli.containerInspectFunc(containerID)
+	if cli.inspectFunc != nil {
+		return cli.inspectFunc(containerID)
 	}
 	return types.ContainerJSON{}, nil
 }
+
+func (cli *fakeClient) ContainerAttach(ctx context.Context, container string, options types.ContainerAttachOptions) (types.HijackedResponse, error) {
```
vdemeester: @dnephin why redefining all of thoses ? `fakeClient` is already a `ContainerClient` as it embedded `client.Client` — and even better, if a call to an unexpected function is done, it will fail the test.. By redifining thoses, this is not the case anymore 😓

dnephin: I think eventually we will add all of these anyway because to test every command we will use every client method. I just added them optimistically instead of waiting. If you would prefer to wait until they are used I can remove them.

dnephin:
This is cleaned up now. Added more tests for exec and removed the unnecessary methods from the fake.

vieux:
LGTM

PR #481, opened 2017-08-29
Add 'build' to types.go

cdrage:
This adds 'build' to types.go in order for projects that use docker/cli
to parse Docker Compose files to correctly retrieve `build` keys.

### DIFF
--- a/cli/compose/loader/full-example.yml
+++ b/cli/compose/loader/full-example.yml
@@ -2,6 +2,21 @@ version: "3.4"
 
 services:
   foo:
+
+    build:
+      context: ./dir
+      dockerfile: Dockerfile
+      args:
+        foo: bar
+      target: foo
+      network: foo
+      cache_from: 
+        - foo
+        - bar
+      labels: [FOO=BAR]
+
+
+    
     cap_add:
       - ALL
 
--- a/cli/compose/types/types.go
+++ b/cli/compose/types/types.go
@@ -78,6 +78,7 @@ type Config struct {
 type ServiceConfig struct {
 	Name string
 
+	Build           BuildConfig
 	CapAdd          []string `mapstructure:"cap_add"`
 	CapDrop         []string `mapstructure:"cap_drop"`
 	CgroupParent    string   `mapstructure:"cgroup_parent"`
@@ -123,6 +124,18 @@ type ServiceConfig struct {
 	User            string
 	Volumes         []ServiceVolumeConfig
 	WorkingDir      string `mapstructure:"working_dir"`
+}
+
+// BuildConfig is a type for build
+// using the same format at libcompose: https://github.com/docker/libcompose/blob/master/yaml/build.go#L12
+type BuildConfig struct {
+	Context    string
+	Dockerfile string
+	Args       MappingWithEquals
+	Labels     Labels
+	CacheFrom  StringList `mapstructure:"cache_from"`
+	Network    string
+	Target     string
 }
 
 // ShellCommand is a string or list of string args
--- a/cli/compose/loader/loader_test.go
+++ b/cli/compose/loader/loader_test.go
@@ -517,12 +517,14 @@ version: "3"
 services:
   web:
     image: web
-    build: ./web
+    build: 
+     context: ./web
     links:
       - bar
   db:
     image: db
-    build: ./db
+    build: 
+     context: ./db
 `))
 	assert.NoError(t, err)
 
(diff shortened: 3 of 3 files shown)


cdrage:
ping @vdemeester @dnephin 

This unblocks us on https://github.com/kubernetes/kompose/issues/636

cli/compose/types/types.go
```diff
@@ -125,6 +125,14 @@ type ServiceConfig struct {
 	WorkingDir      string `mapstructure:"working_dir"`
 }
 
+// BuildConfig is a type for build
+// using the same format at libcompose: https://github.com/docker/libcompose/blob/master/yaml/build.go#L12
+type BuildConfig struct {
+	Context    string
+	Dockerfile string
+	Args       map[string]*string
```
dnephin: This looks incomplete. It should have all the fields from here: https://github.com/docker/cli/blob/e6006219abba35fe118d330b1e99b7cf2c9735f1/cli/compose/schema/data/config_schema_v3.4.json#L84-L90

cli/compose/types/types.go
```diff
@@ -6,7 +6,6 @@ import (
 
 // UnsupportedProperties not yet supported by this implementation of the compose file
 var UnsupportedProperties = []string{
-	"build",
```
dnephin: We need to keep `build` in the list of unsupported properties because it's still not used by the cli right now.

cli/compose/types/types.go
cdrage: @dnephin I'm following @vdemeester 's implementation, but sure, I'll add the other options.

cdrage:
Man you're fast, I'm just fixing the tests failing now @dnephin 

I've already added the `build` keys from types.go to the full-example.

Other than that, it's adding the rest of the types from the schema document.

cli/compose/types/types.go
cdrage: @dnephin Okay :+1:

cdrage:
Problem I have @dnephin in regards to adding `build` to unsupported is that it fails to build / be used... How is it possible to add `build` to the supported types but still have it "unsupported"? When we parse the YAML it still exits ( see the failing test in circleci)

dnephin:
The test failure is because the example config in that test is using a string for `build`. If you update the test to be something like this, it should work:

```yaml
build:
  context: ./web
```

Unsupported properties are only used as a warning, and that warning comes from the cli, not from the `loader` package, so it shouldn't affect you. We could probably move that list out of `types.go` into a different package, but I don't think its necessary right now.

vdemeester:
Design LGTM 🐸

cdrage:
Perfect, thank you @dnephin for the explanation. I wasn't too sure about removing `build` from the failing test case.

I've gone ahead and removed it, hopefully tests pass this time now!

Also greetings fellow Canadian :canada:

cdrage:
Adding WIP tag so I make sure to do two things:

- Have tests pass
- Add rest of the schema

cdrage:
Hey @vdemeester @dnephin is it possible to help out with the failing test?

cdrage:
@dnephin Finally passes! But looks like the tests flaked / need rebuilding on circleci. I'm going to update my commit message and see if it fixes it.

cdrage:
Ayyy, tests pass :+1:

cli/compose/types/types.go
```diff
@@ -78,9 +78,10 @@ type Config struct {
 type ServiceConfig struct {
 	Name string
 
-	CapAdd          []string `mapstructure:"cap_add"`
-	CapDrop         []string `mapstructure:"cap_drop"`
-	CgroupParent    string   `mapstructure:"cgroup_parent"`
+	Build           BuildConfig `mapstructure:"build"`
```
dnephin: not a problem, but technically you don't need the mapstructure annotation here, since it defaults to build

cli/compose/types/types.go
```diff
@@ -125,6 +126,18 @@ type ServiceConfig struct {
...
 }
 
+// BuildConfig is a type for build
+// using the same format at libcompose: https://github.com/docker/libcompose/blob/master/yaml/build.go#L12
+type BuildConfig struct {
+	Context    string
+	Dockerfile string
+	Args       MappingWithEquals
+	Labels     Labels
+	CacheFrom  StringList
```
dnephin: This one will need a `mapstructure:` annotation because it defaults to using a dash instead of an underscore to separate the words.

cli/compose/loader/full-example.yml
```diff
@@ -2,6 +2,13 @@ version: "3.4"
 
 services:
   foo:
+
+    build:
+      context: ./dir
+      dockerfile: Dockerfile
+      args:
+        foo: bar
```
dnephin: Please add the other fields (target, network, cache_from, labels) to this as well.

thaJeztah:
@cdrage can you address the review comments?

cdrage:
@thaJeztah @dnephin 

Done! Updated the full-example.yml as well as the `loader_test.go` file with the aforementioned updates.

PR #485, opened 2017-08-30
Re-adds test target to the Makefile

seemethere:
**- What I did**
The test target existed before, this is to provide a legacy interface to
allow easy testing for downstream Docker CE.

Without this we would need separate Makefiles/Jenkinsfiles for releases
past 17.07. Later on this target could also be used to test both unit
tests and integration tests at the same time.

**- How I did it**
Re-added the target, it's basically an alias now.

**- How to verify it**
* `make test`
* `make -f docker.Makefile test`

* Re-added `test` target to the Makefile and docker.Makefile

Are you not entertained

### DIFF
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -46,6 +46,9 @@ clean: build_docker_image
 test-unit: build_docker_image
 	docker run --rm $(ENVVARS) $(MOUNTS) $(DEV_DOCKER_IMAGE_NAME) make test-unit
 
+.PHONY: test
+test: test-unit test-e2e
+
 # build the CLI for multiple architectures using a container
 .PHONY: cross
 cross: build_cross_image
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,9 @@ clean: ## remove build artifacts
 test-unit: ## run unit test
 	./scripts/test/unit $(shell go list ./... | grep -vE '/vendor/|/e2e/')
 
+.PHONY: test
+test: test-unit ## run tests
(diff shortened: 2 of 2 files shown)


docker.Makefile
```diff
@@ -46,6 +46,10 @@ clean: build_docker_image
 test-unit: build_docker_image
 	docker run --rm $(ENVVARS) $(MOUNTS) $(DEV_DOCKER_IMAGE_NAME) make test-unit
 
+# run go test
+.PHONY: test
+test: test-unit
```
dnephin: This should be an alias for running both test suites I think

```
.PHONY:
test: test-unit test-e2e
```

The comment can be removed, I don't think it adds anything

seemethere: @dnephin Removed

dnephin: Thanks! How about the other part (aliasing both test suite) ?

docker.Makefile
```diff
@@ -46,6 +46,9 @@ clean: build_docker_image
 test-unit: build_docker_image
 	docker run --rm $(ENVVARS) $(MOUNTS) $(DEV_DOCKER_IMAGE_NAME) make test-unit
 
+.PHONY: test
+test: test-unit
```
dnephin: I think `test` should be an alias for running all the test suites, so please make this: `test: test-unit test-e2e`

seemethere: @dnephin Added! 👍

PR #486, opened 2017-08-31
refactor and export hijack.go

AntaresS:

**- What I did**
- Refactored and exported `hijack.go` code so that other packages can use
- Updated other container code that uses `hijack.go` accordingly

**- How I did it**
- Exported `Stream` function as main entry point

**- How to verify it**
- Run `make -f docker.MakeFile shell` and verified no behavior has been changed

Exported `Stream` function in hijack.go so that developers can use it to handle interactive tty.

image

### DIFF
--- a/cli/command/container/attach.go
+++ b/cli/command/container/attach.go
@@ -1,7 +1,6 @@
 package container
 
 import (
-	"io"
 	"net/http/httputil"
 
 	"github.com/docker/cli/cli"
@@ -13,6 +12,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"golang.org/x/net/context"
+	"github.com/docker/cli/cli/hijack"
 )
 
 type attachOptions struct {
@@ -87,11 +87,6 @@ func runAttach(dockerCli command.Cli, opts *attachOptions) error {
 		DetachKeys: dockerCli.ConfigFile().DetachKeys,
 	}
 
-	var in io.ReadCloser
-	if options.Stdin {
-		in = dockerCli.In()
-	}
-
 	if opts.proxy && !c.Config.Tty {
 		sigc := ForwardAllSignals(ctx, dockerCli, opts.container)
 		defer signal.StopCatch(sigc)
@@ -123,17 +118,16 @@ func runAttach(dockerCli command.Cli, opts *attachOptions) error {
 		resizeTTY(ctx, dockerCli, opts.container)
 	}
 
-	streamer := hijackedIOStreamer{
-		streams:      dockerCli,
-		inputStream:  in,
-		outputStream: dockerCli.Out(),
-		errorStream:  dockerCli.Err(),
-		resp:         resp,
-		tty:          c.Config.Tty,
-		detachKeys:   options.DetachKeys,
+	hijackOpts := hijack.StreamOptions{
+		Hijacked:     resp,
+		AttachStdin:  options.Stdin,
+		AttachStdout: true,
+		AttachStderr: true,
+		Tty:          c.Config.Tty,
+		DetachKeys:   options.DetachKeys,
 	}
 
-	if err := streamer.stream(ctx); err != nil {
+	if err := hijack.Stream(ctx, dockerCli, hijackOpts); err != nil {
 		return err
 	}
 
--- a/cli/command/container/exec.go
+++ b/cli/command/container/exec.go
@@ -2,7 +2,6 @@ package container
 
 import (
 	"fmt"
-	"io"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
@@ -15,6 +14,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"golang.org/x/net/context"
+	"github.com/docker/cli/cli/hijack"
 )
 
 type execOptions struct {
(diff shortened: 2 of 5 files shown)


AntaresS:
@tiborvass @dnephin Hey as we discussed earlier. Although I still feel like `hijack.go` is still kind of "heavy", at least we can use it freely for other packages now.

AntaresS:
@jlhawn cc you as well, since I've made many changes on your early refactoring.

cli/command/container/hijack.go
```diff
@@ -19,41 +19,41 @@ import (
...
 // connection.
-type hijackedIOStreamer struct {
-	streams      command.Streams
-	inputStream  io.ReadCloser
-	outputStream io.Writer
-	errorStream  io.Writer
-
-	resp types.HijackedResponse
+type HijackOptions struct {
+	//config       *container.Config
```
tiborvass: @AntaresS spurious comment

AntaresS: Oops.. deleted

AntaresS:
@tiborvass not sure why linter is failing...

dnephin:
I clicked rebuild on `lint` and it looks fine now. I'll take a look over the code soon

cli/command/container/hijack.go
```diff
@@ -19,41 +19,40 @@ import (
...
+type HijackOptions struct {
+	resp       types.HijackedResponse
+	ttyInput   io.ReadCloser
+	inputFlag  bool
+	outputFlag bool
+	errorFlag  bool
 
 	tty        bool
 	detachKeys string
 }
```
jlhawn: Exporting this type isn't going to be very helpful unless you also export all of the struct fields 😉

cli/command/container/hijack.go
```diff
@@ -19,41 +19,40 @@ import (
...
-type hijackedIOStreamer struct {
-	streams      command.Streams
-	inputStream  io.ReadCloser
-	outputStream io.Writer
-	errorStream  io.Writer
-
-	resp types.HijackedResponse
+type HijackOptions struct {
+	resp       types.HijackedResponse
+	ttyInput   io.ReadCloser
```
jlhawn: I don't think we necessarily need this field either. Check to see if we actually ever call `Close()` on this value. We might be able to get away removing it the same you did with `outputStream` and `errorStream` and just use the passed in `streams.In()`.

cli/command/container/hijack.go
```diff
@@ -19,41 +19,40 @@ import (
...
+	inputFlag  bool
+	outputFlag bool
+	errorFlag  bool
```
jlhawn: We should be consistent with these names from other struct types: i.e., use `AttachStdin`, `AttachStdout`, and `AttachStderr`.

cli/command/container/hijack.go
AntaresS: @jlhawn I saw this io.ReadCloser is tricker than the other two. On [line 104](https://github.com/AntaresS/cli/blob/64f436f08c10656745ce8c489d1f218b3a6f7690/cli/command/container/hijack.go#L104) it is being overwritten and `Close()` is called on [line 202](https://github.com/AntaresS/cli/blob/64f436f08c10656745ce8c489d1f218b3a6f7690/cli/command/container/hijack.go#L202). Seems like it is no longer an on/off switch any more like `outputStream` and `errorStream`

cli/command/container/hijack.go
AntaresS: I hesitated on renaming this. Thanks for helping me make the decision ;)

cli/command/container/hijack.go
AntaresS: Ah good catch. Forgot to capitalize the fields as well.

AntaresS:
Updated and rebased

cli/command/container/hijack.go
jlhawn: The Close() on line 202 is effectively calling `streams.In().Close()` though.

AntaresS:
@dnephin Made a couple of changes accordingly. Please let me know if we can merge it. 🙂

cli/command/container/hijack.go
```diff
@@ -19,41 +19,38 @@ import (
 // TODO: This could be moved to `pkg/term`.
 var defaultEscapeKeys = []byte{16, 17}
 
-// A hijackedIOStreamer handles copying input to and output from streams to the
```
dnephin: This docstring need to be updated to:

"HijackOptions used by Stream to configure streaming behaviour"

cli/command/container/hijack.go
```diff
@@ -71,45 +68,44 @@ func (h *hijackedIOStreamer) stream(ctx context.Context) error {
 	}
 }
 
-func (h *hijackedIOStreamer) setupInput() (restore func(), err error) {
-	if h.inputStream == nil || !h.tty {
+func setupInput(streams command.Streams, opts *HijackOptions) (restore func(), reader io.Reader, err error) {
```
dnephin: I think this should receive a non-pointer type for `opts`. It shouldn't be updating `opts`

cli/command/container/hijack.go
```diff
@@ -19,41 +19,38 @@ import (
...
-	inputStream  io.ReadCloser
-	outputStream io.Writer
-	errorStream  io.Writer
-
-	resp types.HijackedResponse
-
-	tty        bool
-	detachKeys string
+type HijackOptions struct {
+	Resp         types.HijackedResponse
```
dnephin: Maybe `Hijacked` instead of `Resp` for the field name?

AntaresS: I don't hold strong opinion on naming this guy. Changed to `Hijacked`

AntaresS:
> I think we should move hijack.go to a new package, so consumers don't have to include all the dependencies of cli/command/container.
> A new package at cli/hijack/ might be good.
> Another problem is we have basically no tests for this right now. Some test coverage would be great. It looks like the build is failing right now.
@dnephin @tiborvass 
I moved `hijack.go` to `cli/hijack` folder. Don't think this part of code should belong to container cli either. 
For the unit test, is it a must requirement for this PR? Otherwise I suggest we ship unit test in another PR. I hope we can get this one merged in order to unblock my other task 🙂

dnephin:
#495 should provide some indirect converage, but some a couple unit tests would really be good.

Looks like there is one lint error.

You can use `vndr` to point at your branch so that you can start developing against this before it gets merged.

thaJeztah:
ping @AntaresS are you still working on this? Looks like this needs a rebase now

thaJeztah:
ping @AntaresS 😇

AntaresS:
@thaJeztah Sorry I lost track on this PR. Let's close it since the original feature request in testkit is no longer active.

Issue #490, opened 2017-09-01
Docker stack rm sent to stderr


FrenchBen:
**Description**

When removing a stack docker output is sent to `stderr` 

**Steps to reproduce the issue:**
Given a docker-compose.yml file:
```
version: '3'

services:
    nginx:
        image: nginx
        deploy:
                replicas: 3
```
1. create a stack: `docker stack deploy -c docker-compose.yml nginx`
2. remove the stack and capture stdout and stderr: `docker stack rm nginx 1>stdout.log 2>stderr.log`

**Describe the results you received:**
Checking the `stdout` and `stderr` files, only the `stderr` had the "successful" removal calls:
```
Removing service nginx_nginx
Removing network nginx_default
```

**Describe the results you expected:**
I expected the `remove` calls to be sent to `stdout`. But instead they were sent to `stderr.log` and `stdout.log` is empty 😞 

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
$ docker version
Client:
 Version:      17.07.0-ce
 API version:  1.31
 Go version:   go1.8.3
 Git commit:   8784753
 Built:        Tue Aug 29 17:40:27 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.07.0-ce
 API version:  1.31 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   8784753
 Built:        Tue Aug 29 17:46:50 2017
 OS/Arch:      linux/amd64
 Experimental: true
```

PR #491, opened 2017-09-01
Move output of stack rm to stdout

FrenchBen:
**- What I did**
Updated the output of `docker stack rm` to be sent to stdout.

Fixes https://github.com/docker/cli/issues/490

**- How I did it**
`s/Err/Out`

**- How to verify it**
With no errors, doing `docker stack rm nginx 1>stdout.log 2>stderr.log` should show all output in `stdout`

`docker stack rm` now sends info messages to Stdout

doge

### DIFF
--- a/cli/command/stack/remove_test.go
+++ b/cli/command/stack/remove_test.go
@@ -101,7 +101,13 @@ func TestRemoveStackSkipEmpty(t *testing.T) {
 	cmd.SetArgs([]string{"foo", "bar"})
 
 	assert.NoError(t, cmd.Execute())
-	assert.Equal(t, "", fakeCli.OutBuffer().String())
+	expectedList := []string{"Removing service bar_service1",
+		"Removing service bar_service2",
+		"Removing secret bar_secret1",
+		"Removing config bar_config1",
+		"Removing network bar_network1\n",
+	}
+	assert.Equal(t, strings.Join(expectedList, "\n"), fakeCli.OutBuffer().String())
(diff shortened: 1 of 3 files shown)


dnephin:
Why should it go to stdout?

FrenchBen:
It's not an error?

dnephin:
Ok, makes sense to be consistent with the other remove commands.

The e2e test needs to be fixed as well

cli/command/stack/remove_test.go
```diff
@@ -101,7 +101,13 @@ func TestRemoveStackSkipEmpty(t *testing.T) {
...
 	assert.NoError(t, cmd.Execute())
-	assert.Equal(t, "", fakeCli.OutBuffer().String())
+	expectedList := []string{"Removing service bar_service1",
+		"Removing service bar_service2",
+		"Removing secret bar_secret1",
+		"Removing config bar_config1",
+		"Removing network bar_network1\n",
+	}
+	assert.Equal(t, strings.Join(expectedList, "\n"), fakeCli.OutBuffer().String())
 	assert.Contains(t, fakeCli.ErrBuffer().String(), "Nothing found in stack: foo\n")
```
dnephin: hmm, actually this seems weird. Why is it outputing both "nothing found in stack" and that it's removing things?

Not sure how this is passing. Something seems wrong

dnephin: Oh, nevermind, I see it now. There are two stacks, one is empty, the other is not.

FrenchBen:
Can you squash on merge?

PR #492, opened 2017-09-01
Add bash completion for `dockerd --network-control-plane-mtu`

albers:
This adds bash completion for https://github.com/moby/moby/pull/34103


albers:
self-LGTM

PR #493, opened 2017-09-01
Change sshfs by cloudstor on Installing and using a plugin

jmaitrehenry:
Signed-off-by: Julien Maitrehenry <julien.maitrehenry@me.com>

**- What I did**
Update the documentation page about plugins
See for the original PR: https://github.com/docker/docker.github.io/pull/4294

Replace sshfs plugin installation and usage example by cloudstor

### DIFF
--- a/docs/extend/index.md
+++ b/docs/extend/index.md
@@ -43,37 +43,49 @@ Plugins that start successfully are listed as enabled in the output.
 After a plugin is installed, you can use it as an option for another Docker
 operation, such as creating a volume.
 
-In the following example, you install the `sshfs` plugin, verify that it is
+In the following example, you install the `docker4x/cloudstor` plugin, verify that it is
 enabled, and use it to create a volume.
-
-> **Note**: This example is intended for instructional purposes only. Once the volume is created, your SSH password to the remote host will be exposed as plaintext when inspecting the volume. You should delete the volume as soon as you are done with the example.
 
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@ddebroy PTAL

ddebroy:
LGTM .. also adding @friism as FYI

friism:
I'm not sure I think cloudstor is a good example to use in the plugin sample docs:

 * It only works on AWS/Azure (which means that trying it out requires much more setup)
 * It's more complicated, which makes it less useful as an example because the user also has to grok a bunch of AWS-specifics that are unrelated to the samples purpose of teaching people about plugins

mstanleyjones:
:( I'll revert and we can continue discussion here or maybe @jmaitrehenry can do a follow-on PR.

jmaitrehenry:
Where could we add documentation about cloudstor?

First idea: put it with docker for Azure/AWS because it's only for those two but, as @ddebroy already mention, it comes pre-installed / pre-configured on those products

Second idea: add the information on the plugin page here: https://store.docker.com/community/images/docker4x/cloudstor

Third idea: create a tutorial on how create a docker cluster with swarm and cloudstor on Azure here: https://docs.docker.com/samples/ but why we want to create a cluster on Azure without using Docker for Azure?

Fourth: Open source the cloudstor plugin and add documentation inside the plugin repository but it's not for now I think.

PR #509, opened 2017-09-07
sort secrets and configs to ensure idempotence

xificurC:
`docker stack deploy` keeps restarting services it doesn't need to (no changes)
because the entries' order gets randomized at some previous (de)serialization.
Maybe it would be worth looking into this at a higher level and ensure
all (de)serialization happens in an ordered collection.

This quick fix sorts secrets and configs (in place, mutably) which ensures the
same order for each run.

Based on
https://github.com/moby/moby/pull/30506

Fixes
https://github.com/moby/moby/issues/34746

### DIFF
--- a/cli/compose/convert/service.go
+++ b/cli/compose/convert/service.go
@@ -295,7 +295,13 @@ func convertServiceSecrets(
 		})
 	}
 
-	return servicecli.ParseSecrets(client, refs)
+	secrs, err := servicecli.ParseSecrets(client, refs)
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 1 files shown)


xificurC:
refactored as requested

mghazizadeh:
How's this tested?

vieux:
@mghazizadeh the steps are given on the upstream PR: https://github.com/moby/moby/issues/34746

vieux:
@mghazizadeh test was added in https://github.com/docker/cli/pull/719

PR #510, opened 2017-09-07
Use a local copy of ParseLogDetails

dnephin:
It's being removed from client/ in https://github.com/moby/moby/pull/34767


vieux:
LGTM

PR #513, opened 2017-09-08
 Reset uid/gid to 0 in build context to fix cache busting issues on ADD/COPY

shouze:
**- What I did**

Fix of https://github.com/moby/moby/issues/32816

**- How I did it**

By resetting uid/gid to 0/0 in the build context. Prior to that, made possible by adding chownOpts in pkg/archive tarWithOption https://github.com/moby/moby/pull/34590

**- How to verify it**

`go test -v ./cli/command/image -run TestRunBuildResetsUidAndGidInContext`

Reset uid/gid to 0 in build context to fix cache busting issues on ADD/COPY

133-insolite-25

### DIFF
--- a/cli/command/image/build_test.go
+++ b/cli/command/image/build_test.go
@@ -6,18 +6,57 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
+	"syscall"
 	"testing"
 
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/internal/test"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/archive"
+	"github.com/gotestyourself/gotestyourself/fs"
+	"github.com/gotestyourself/gotestyourself/skip"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/net/context"
 )
 
+func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
+	skip.IfCondition(t, runtime.GOOS == "windows", "uid and gid not relevant on windows")
+	dest := fs.NewDir(t, "test-build-context-dest")
+	defer dest.Remove()
+
+	fakeImageBuild := func(_ context.Context, context io.Reader, options types.ImageBuildOptions) (types.ImageBuildResponse, error) {
+		assert.NoError(t, archive.Untar(context, dest.Path(), nil))
+
+		body := new(bytes.Buffer)
+		return types.ImageBuildResponse{Body: ioutil.NopCloser(body)}, nil
+	}
+	cli := test.NewFakeCli(&fakeClient{imageBuildFunc: fakeImageBuild})
+
+	dir := fs.NewDir(t, "test-build-context",
+		fs.WithFile("foo", "some content", fs.AsUser(65534, 65534)),
+		fs.WithFile("Dockerfile", `
+			FROM alpine:3.6
+			COPY foo bar /
+		`),
+	)
+	defer dir.Remove()
+
+	options := newBuildOptions()
+	options.context = dir.Path()
+
+	err := runBuild(cli, options)
+	require.NoError(t, err)
+
+	files, err := ioutil.ReadDir(dest.Path())
+	require.NoError(t, err)
+	for _, fileInfo := range files {
+		assert.Equal(t, uint32(0), fileInfo.Sys().(*syscall.Stat_t).Uid)
+		assert.Equal(t, uint32(0), fileInfo.Sys().(*syscall.Stat_t).Gid)
+	}
+}
 func TestRunBuildDockerfileFromStdinWithCompress(t *testing.T) {
 	dest, err := ioutil.TempDir("", "test-build-compress-dest")
 	require.NoError(t, err)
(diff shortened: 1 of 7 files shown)


shouze:
@tonistiigi I need to write the test but looks like not so trivial to me ATM as buildCtx is internal concept of runBuild(). Still need to read the code. I also want to address --stream, will I need to make changes inside buildkit?

tonistiigi:
@dnephin may have some tips for the test. For `--stream` I need to fix https://github.com/moby/moby/issues/34708 first.

dnephin:
@tonistiigi this would be really easy to test if we merged #294 , it would be a unit test on `createBuildContextFromLocalDir()`

shouze:
@tonistiigi @dnephin ok I've added the test, ready for a review.

@tonistiigi ok so about `--stream` maybe I can make a dedicated PR as it's still an experimental feature while this PR fixes a legacy issue. WDYT?

shouze:
@tonistiigi I've started https://github.com/moby/moby/issues/34708 in https://github.com/moby/moby/pull/34797 but looks like it's probably just more than an update, I'll take a look as a fsutil upgrade seems required too.

cli/command/image/build_test.go
```diff
@@ -18,6 +20,52 @@ import (
 	"golang.org/x/net/context"
 )
 
+func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Invalid test on Windows")
+	}
```
dnephin: We have some helpers for this, please use https://godoc.org/github.com/gotestyourself/gotestyourself/skip

```go
skip.IfCondition(t, runtime.GOOS == "windows", "uid and gid not relevant on windows")
```

cli/command/image/build_test.go
```diff
@@ -18,6 +20,52 @@ import (
...
+	dest, err := ioutil.TempDir("", "test-build-context-dest")
+	require.NoError(t, err)
+	defer os.RemoveAll(dest)
```
dnephin: We have helpers for this: https://godoc.org/github.com/gotestyourself/gotestyourself/fs

```go
dest := fs.NewDir(t, "test-build-context-dest")
defer dest.Remove()

// use dest.Path() to get the path
```

cli/command/image/build_test.go
```diff
@@ -18,6 +20,52 @@ import (
...
+	`)
+	cli.SetIn(command.NewInStream(ioutil.NopCloser(dockerfile)))
+
+	dir, err := ioutil.TempDir("", "test-build-context")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	ioutil.WriteFile(filepath.Join(dir, "foo"), []byte("some content"), 0644)
+	err = os.Lchown(filepath.Join(dir, "foo"), 65534, 65534)
+	require.NoError(t, err)
```
dnephin: This test can use a Dockerfile from the context, instead of stdin. So `dockerfile` can be just the string, instead of a bytes.Buffer. and you don't need to set `cli.SetIn()`. Below you won't need the `options.dockerfileName = "-"` line either.

You can replace this block (line 45-53) with:

```go
dir := fs.NewDir(t, "test-build-context", 
    fs.WithFile("foo", "some content", AsUser(65534, 65534)),
    fs.WithFile("Dockerfile", dockerfile),
)
defer dir.Remove()
```

You'll have to update `vendor.conf` to use gotestyourself `v1.2.0` as `WithFile()` was fixed in that version to accept `PathOp`

shouze:
@dnephin ready for a new review, thx for gotestyourself tips, it reduces a lot the noise when we read the test 👍

cli/command/image/build_test.go
```diff
@@ -6,18 +6,59 @@ import (
...
 )
 
+func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
+	skip.IfCondition(t, runtime.GOOS == "windows", "uid and gid not relevant on windows")
+	dest := fs.NewDir(t, "test-build-context-dest")
+	defer dest.Remove()
+
+	fakeImageBuild := func(_ context.Context, context io.Reader, options types.ImageBuildOptions) (types.ImageBuildResponse, error) {
+		buffer := new(bytes.Buffer)
+		tee := io.TeeReader(context, buffer)
```
dnephin: Is this necessary? It doesn't seem like `buffer` is used.

cli/command/image/build_test.go
```diff
@@ -6,18 +6,58 @@ import (
...
 )
 
+func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
+	skip.IfCondition(t, runtime.GOOS == "windows", "uid and gid not relevant on windows")
+	dest := fs.NewDir(t, "test-build-context-dest")
+	defer dest.Remove()
+
+	fakeImageBuild := func(_ context.Context, context io.Reader, options types.ImageBuildOptions) (types.ImageBuildResponse, error) {
+		tee := io.TeeReader(context, new(bytes.Buffer))
+		assert.NoError(t, archive.Untar(tee, dest.Path(), nil))
```
dnephin: I think you can pass `context` here and remove `tee` altogether, no?

The reason I'm using `io.TeeReader()` in the other test is because I want to assert the header is as expected, but in this case the test isn't concerned with the compression, it just wants the files, so I don't think you need the extra buffer.

shouze: You're right, done!

tonistiigi:
LGTM

@shouze Needs rebase

shouze:
@tonistiigi thx, rebased!

shouze:
@dnephin @tonistiigi I guess that after the merge a documentation upgrade would be required?

Issue #494, opened 2017-09-01
Capture Engine CLI experimental annotations on commands and flags


johndmulhausen:
@mstanleyjones commented on [Tue Feb 14 2017](https://github.com/docker/docker.github.io/issues/1663)

### Problem description

It would be good to capture, in the CLI YAML, when a command or a flag is experimental. Here are some examples in the code:

- [experimental command](https://github.com/docker/docker/blob/master/cli/command/checkpoint/cmd.go#L16)
- [experimental flag](https://github.com/docker/docker/blob/master/cli/command/image/build.go#L115)

If this info was surfaced in the YAML, we could template it to automatically mark commands and flags as experimental.

cc/ @johndmulhausen @thaJeztah @dnephin @vieux 

---

@thaJeztah commented on [Tue Feb 14 2017](https://github.com/docker/docker.github.io/issues/1663#issuecomment-279870497)

There's also an [API version annotation](https://github.com/docker/docker/blob/master/cli/command/image/build.go#L116), which could be used to document "new in version x.y.z" (if we translate API version to docker version

---

@FrenchBen commented on [Tue Feb 14 2017](https://github.com/docker/docker.github.io/issues/1663#issuecomment-279870707)

Experimental command example:
https://github.com/docker/docker/blob/master/cli/command/checkpoint/cmd.go

Experimental flag example:
https://github.com/docker/docker/blob/master/cli/command/container/start.go

johndmulhausen:
Ping @tiborvass

thaJeztah:
Oh, I think this will be resolved by https://github.com/docker/cli/pull/523 (still catching up on my post-vacation notifications, so didn't see this one)

@johndmulhausen can you have a look at that one, and if that would resolve things for you?

PR #523, opened 2017-09-12
Add more information to the generated YAML for documentation

thaJeztah:
Fixes https://github.com/docker/cli/issues/494

**- What I did**

While looking at https://github.com/docker/docker.github.io/issues/3886, I noticed there's some information not included in the generated YAML for the documentation that could be useful. This PR contains two patches that adds additional information about commands and flags;

The first patch adds aditional information about command flags to the YAML files that are generated for the reference documentation.

The following fields are added for each flag:

Property          | Type      | Description
------------------|-----------|---------------------------------------------------------------------------------------
value_type        | String    | The "type" of value to be passed to this flag (e.g., `uint64`, `list`)
deprecated        | Boolean   | Indicates if the flag is marked deprecated
min_api_version   | String    | The API version required to use this flag (e.g. "1.23")
experimental      | Boolean   | Indicates if the flag requires the daemon to run with experimental features enabled

For example (taken from the `docker image build` command):

```yaml
- option: security-opt
  value_type: stringSlice
  default_value: '[]'
  description: Security options
  deprecated: false
  experimental: false
- option: shm-size
  value_type: bytes
  default_value: "0"
  description: Size of /dev/shm
  deprecated: false
  experimental: false
- option: squash
  value_type: bool
  default_value: "false"
  description: Squash newly built layers into a single new layer
  deprecated: false
  min_api_version: "1.25"
  experimental: true
```

The second patch adds aditional information about _commands_. The following fields are added for each command:

Property          | Type      | Description
------------------|-----------|---------------------------------------------------------------------------------------
deprecated        | Boolean   | Indicates if the command is marked deprecated
min_api_version   | String    | The API version required to use this command (e.g. "1.23")
experimental      | Boolean   | Indicates if the command requires the daemon to run with experimental features enabled

For example (taken from the experimental `docker checkpoint create` command):

```yaml
command: docker checkpoint create
short: Create a checkpoint from a running container
long: Create a checkpoint from a running container
usage: docker checkpoint create [OPTIONS] CONTAINER CHECKPOINT
pname: docker checkpoint
plink: docker_checkpoint.yaml
options:
- option: checkpoint-dir
  value_type: string
  description: Use a custom checkpoint storage directory
  deprecated: false
  experimental: false
- option: leave-running
  value_type: bool
  default_value: "false"
  description: Leave the container running after checkpoint
  deprecated: false
  experimental: false
deprecated: false
min_api_version: "1.25"
experimental: true
```

**- How to verify it**

Run `make yamldocs` and check the files that are generated in the `docs/yaml/gen` directory

ping @vdemeester @tiborvass @mstanleyjones PTAL

### DIFF
--- a/docs/yaml/yaml.go
+++ b/docs/yaml/yaml.go
@@ -14,10 +14,14 @@ import (
 )
 
 type cmdOption struct {
-	Option       string
-	Shorthand    string `yaml:",omitempty"`
-	DefaultValue string `yaml:"default_value,omitempty"`
-	Description  string `yaml:",omitempty"`
+	Option        string
+	Shorthand     string `yaml:",omitempty"`
+	ValueType     string `yaml:"value_type,omitempty"`
+	DefaultValue  string `yaml:"default_value,omitempty"`
+	Description   string `yaml:",omitempty"`
+	Deprecated    bool
+	MinAPIVersion string `yaml:"min_api_version,omitempty"`
+	Experimental  bool
 }
 
 type cmdDoc struct {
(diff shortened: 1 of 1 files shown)


docs/yaml/yaml.go
```diff
@@ -93,6 +98,18 @@ func GenYamlCustom(cmd *cobra.Command, w io.Writer) error {
 	if len(cmd.Example) > 0 {
 		cliDoc.Example = cmd.Example
 	}
+	if len(cmd.Deprecated) > 0 {
+		cliDoc.Deprecated = true
+	}
+	// Check recursively so that, e.g., `docker stack ls` returns the same output as `docker stack`
+	for curr := cmd; curr != nil; curr = curr.Parent() {
+		if v, ok := curr.Tags["version"]; ok {
+			cliDoc.MinAPIVersion = v
```
dnephin: When the parent command has a minimum version greater then the child command I believe this will give you the wrong value. Ex: `docker foo` requires 1.33 and `docker foo bar` requires 1.34

I think this needs an "if less than" check before setting the version.

mstanleyjones:
In principle I'm really happy about this.  I asked for the `experimental` information a long time ago and nothing happened with it, so this is great. If this goes in, then I can modify the Liquid we are using to generate the CLI docs to expose these new values for the flags.

docs/yaml/yaml.go
thaJeztah: Ah, yes, good one; I went for the simple alternative, and stop at the first value found, both for "experimental" and "version"

thaJeztah:
@dnephin @vdemeester updated, ptal

thaJeztah:
@dnephin updated; it's green now 👍

dnephin:
For some reason the E2E suite didn't run on this PR, I've noticed that on a few, not sure why. 

In this case it's not actually providing any coverage of the change so I think it's fine to merge without it.

PR #525, opened 2017-09-12
Use non-detached mode as default for service commands

thaJeztah:
**- What I did**

Commit 330a0035334871d92207b583c1c36d52a244753f added a `--detach=false` option
to various service-related commands, with the intent to make this the default in
a future version (17.09).

This patch changes the default to use "interactive" (non-detached), allowing
users to override this by setting the `--detach` option.

To prevent problems when connecting to older daemon versions (17.05 and below,
see commit db60f255617fd90cb093813dcdfe7eec840eeff8), the detach option is
ignored for those versions, and detach is always true.

Before this change, a warning was printed to announce the upcoming default:

```bash
$ docker service create nginx:alpine
saxiyn3pe559d753730zr0xer
Since --detach=false was not specified, tasks will be created in the background.
In a future release, --detach=false will become the default.
```

After this change, no warning is printed, but `--detach` is disabled;

```bash
$ docker service create nginx:alpine
y9jujwzozi0hwgj5yaadzliq6
verify: Service converged
```

Setting the `--detach` flag makes the cli use the pre-17.06 behavior:

```bash
$ docker service create --detach nginx:alpine
280hjnzy0wzje5o56gr22a46n
```

Running against a 17.03 daemon, without specifying the `--detach` flag;

```bash
$ docker service create nginx:alpine
kqheg7ogj0kszoa34g4p73i8q
```

    Use interactive (non-detached) mode for `service create`, `service update`, `service rollback`, and `service scale` by default

maxresdefault

ping @vdemeester @aaronlehmann @dnephin


cli/command/service/rollback.go
```diff
@@ -56,8 +57,7 @@ func runRollback(dockerCli command.Cli, flags *pflag.FlagSet, options *serviceOp
 
 	fmt.Fprintf(dockerCli.Out(), "%s\n", serviceID)
 
-	if options.detach {
-		warnDetachDefault(dockerCli.Err(), apiClient.ClientVersion(), flags, "rolled back")
+	if options.detach || versions.LessThan(apiClient.ClientVersion(), "1.29") {
```
thaJeztah: One thing I was thinking is if we should put the version-check somewhere central, and just change `options.detach` to `true`, based on that version; this would prevent possible bugs if more commands get an interactive variant.

dnephin: I don't think we need the version check anymore. The behaviour is client-side only, right? So even if they are running against an older daemon we should keep the same default for detach. I think we can remove the version check.

thaJeztah: I wasn't sure if older daemons would return the right response for the `waitOnService()` to work correctly; if that's not an issue then I am +1 on removing

dnephin: Apparently it depends on this `_up-to-date=true` filter which was added in 17.05, so I guess it does depend on this version still.

thaJeztah:
Looks like there's a lint warning; I'll have a look at those;

```
cli/command/service/rollback.go:35:41:warning: parameter flags is unused (unparam)
cli/command/service/scale.go:58:38:warning: parameter flags is unused (unparam)
```

thaJeztah:
fixed linting error, PTAL;

```patch
diff --git a/cli/command/service/rollback.go b/cli/command/service/rollback.go
index da8c1fad..375e6d51 100644
--- a/cli/command/service/rollback.go
+++ b/cli/command/service/rollback.go
@@ -9,7 +9,6 @@ import (
        "github.com/docker/docker/api/types"
        "github.com/docker/docker/api/types/versions"
        "github.com/spf13/cobra"
-       "github.com/spf13/pflag"
 )

 func newRollbackCommand(dockerCli command.Cli) *cobra.Command {
@@ -20,7 +19,7 @@ func newRollbackCommand(dockerCli command.Cli) *cobra.Command {
                Short: "Revert changes to a service's configuration",
                Args:  cli.ExactArgs(1),
-                       return runRollback(dockerCli, cmd.Flags(), options, args[0])
+                       return runRollback(dockerCli, options, args[0])
                },
                Tags: map[string]string{"version": "1.31"},
        }
@@ -32,7 +31,7 @@ func newRollbackCommand(dockerCli command.Cli) *cobra.Command {
        return cmd
 }

-func runRollback(dockerCli command.Cli, flags *pflag.FlagSet, options *serviceOptions, serviceID string) error {
+func runRollback(dockerCli command.Cli, options *serviceOptions, serviceID string) error {
        apiClient := dockerCli.Client()
        ctx := context.Background()

diff --git a/cli/command/service/scale.go b/cli/command/service/scale.go
index 5ab083a1..d38b01b4 100644
--- a/cli/command/service/scale.go
+++ b/cli/command/service/scale.go
@@ -13,7 +13,6 @@ import (
        "github.com/docker/docker/api/types/versions"
        "github.com/pkg/errors"
        "github.com/spf13/cobra"
-       "github.com/spf13/pflag"
 )

 type scaleOptions struct {
@@ -28,7 +27,7 @@ func newScaleCommand(dockerCli command.Cli) *cobra.Command {
                Short: "Scale one or multiple replicated services",
                Args:  scaleArgs,
-                       return runScale(dockerCli, cmd.Flags(), options, args)
+                       return runScale(dockerCli, options, args)
                },
        }

@@ -55,7 +54,7 @@ func scaleArgs(cmd *cobra.Command, args []string) error {
        return nil
 }

-func runScale(dockerCli command.Cli, flags *pflag.FlagSet, options *scaleOptions, args []string) error {
+func runScale(dockerCli command.Cli, options *scaleOptions, args []string) error {
        var errs []string
        var serviceIDs []string
        ctx := context.Background()
```

PR #531, opened 2017-09-13
Add a way to just build windows and osx

seemethere:

**- What I did**
Added 2 targets for `binary-windows` and `binary-osx` to build them separately from the `cross` target.

This is useful for builds (like master.dockerproject.org) that don't necessarily care about
getting full binaries for multiarch but do care about Windows. This cuts time needed for
compiling just Windows binaries by more than half

With `cross`:
```
make -f docker.Makefile cross  0.19s user 0.14s system 0% cpu 1:50.96 total
```

With just `windows`:
```
make -f docker.Makefile binary-windows  0.21s user 0.15s system 1% cpu 27.269 total
```

**- How I did it**
Added make targets that correspond to their respective scripts in `scripts/build/$@`

**- How to verify it**
```shell
make -f docker.Makefile binary-windows
make -f docker.Makefile binary-osx
```

* Added targets to cross compile for just windows and osx

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,14 @@ binary: ## build executable for Linux
 cross: ## build executable for macOS and Windows
 	./scripts/build/cross
 
+.PHONY: binary-windows
+binary-windows: ## build executable for Windows
+	./scripts/build/windows
+
+.PHONY: binary-osx
+binary-osx: ## build executable for macOS
+	./scripts/build/osx
+
 .PHONY: dynbinary
 dynbinary: ## build dynamically linked binary
 	./scripts/build/dynbinary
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -34,6 +34,14 @@ binary: ## build executable for Linux
 cross: ## build executable for macOS and Windows
 	./scripts/build/cross
 
+.PHONY: windows
+windows: ## build executable for Windows
```
dnephin: small nit: how about `binary-windows` and `binary-osx` in both Makefiles?

seemethere: @dnephin changed

PR #535, opened 2017-09-15
Remove .swp file that was accidentally added

thaJeztah:
This file was added by accident in dd95731a21534040cc130348255fb160f08efab9 (https://github.com/docker/cli/pull/468) and not noticed during review.

ping @vdemeester PTAL

PR #536, opened 2017-09-15
Update gitignore

thaJeztah:
Adding some lines from the Moby gitignore

@dnephin @vdemeester for consideration; not sure if these would qualify as "should be present in your global ignore" as well

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,12 @@
+# if you want to ignore files created by your editor/tools,
+# please consider a global .gitignore https://help.github.com/articles/ignoring-files
+*.exe
+*.exe~
+*.orig
+.*.swp
 .DS_Store
+Thumbs.db
+.editorconfig
 /build/
 cli/winresources/rsrc_386.syso
 cli/winresources/rsrc_amd64.syso


.gitignore
```diff
@@ -1,4 +1,14 @@
+# if you want to ignore files created by your editor/tools,
+# please consider a global .gitignore https://help.github.com/articles/ignoring-files
+*.exe
+*.exe~
+*.orig
+.*.swp
 .DS_Store
+Thumbs.db
+# a .bashrc may be added to customize the build environment
+.bashrc
```
albers: The custom .bashrc only works in the context of the _moby/moby_ build scripts: https://github.com/moby/moby/blob/master/Dockerfile#L179. There's no corresponding feature in this repo.

thaJeztah: Ah, yes, makes sense, I can remove that, thanks 👍

thaJeztah:
Updated, PTAL 😅

thaJeztah:
To prevent things like https://github.com/docker/cli/pull/535

vdemeester:
@dnephin @thaJeztah what is the current status of this ? 👼

dnephin:
I think generally `.gitignore` should only include files that are created by a project, or are documented by the project, but should not be added to git. Other files should be ignored by the user.

I don't think adding these lines fixes the problem.

vdemeester:
@dnephin although I agree, part of them can be created by the project (`*.exe`, the file ending with `~` or `*.swp` when editing with vim or else, …) and I really think it doesn't really hurt to add them…

… and the PR is stuck since September, it's creeping me out. So I'll go ahead and merge it and if we feel something wrong with it we can remove some of them 👼

PR #544, opened 2017-09-20
Fix loader error with different build syntax

vdemeester:
`build: .` was not working anymore. Fixing this by adding a new
tranform function for BuildConfig.

/cc @cdrage 


thaJeztah:
Do we know where it broke, and if this needs to be back ported?

vdemeester:
@thaJeztah from https://github.com/docker/cli/pull/481. As we don't load it anyway in `docker/cli` we don't fails there. But using it as a `cli`, it fails 😓

cdrage:
@vdemeester kinda odd since `build` is yet to be supported in docker/cli compose, right? or did I miss a recent commit?

vdemeester:
@cdrage it's just that your PR didn't take into account `Load` and thus didn't take into account that both things below works 👼. I missed that when reviewing the first one 👼 

```
build: .
build:
  context: .
  dockerfile: foo.dockerfile
```

cdrage:
@vdemeester Doh. I forgot about that too.

Thanks for the fix :+1:

PR #545, opened 2017-09-20
Fix nocopy description

thaJeztah:
nocopy does not apply to bind mounts, only volumes.

fixes https://github.com/moby/moby/issues/27046


thaJeztah:
author LGTM; merging

PR #546, opened 2017-09-20
Set APIVersion on the client, even when Ping fails

### LINKED ISSUE #149: CLI should honor API version header when _ping returns unsuccessful status code
When `_ping` returns a non-200 HTTP status code, the CLI defaults to the latest API version. This can result in errors like:

> `your client is too new (API version 1.27). The newest supported API version is 1.26.0`

...when the CLI moves on to performing the actual operation.

The CLI should still honor the version header, so `_ping` returning a bad status code doesn't prevent the CLI from working at all. Right now the client's `Ping` method just returns an error, but it should still return the version information in this case. Maybe the HTTP status code should be moved to a field on `types.Ping` instead of triggering an error.

I ran into this issue running the CLI against Docker Datacenter. Its `_ping` endpoint was returning a 500 status code because the cluster was considered unhealthy. But this caused the CLI to use a newer API version than was supported, so CLI commands didn't work at all while the cluster was in this state.

dnephin:
Fixes #149

Refactor to support testing
Also add tests

### DIFF
--- a/cli/command/cli_test.go
+++ b/cli/command/cli_test.go
@@ -4,12 +4,18 @@ import (
 	"os"
 	"testing"
 
+	"crypto/x509"
+
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/cli/cli/flags"
+	"github.com/docker/cli/internal/test/testutil"
 	"github.com/docker/docker/api"
+	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/net/context"
 )
 
 func TestNewAPIClientFromFlags(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cli/command/cli_test.go
```diff
@@ -55,3 +61,137 @@ func patchEnvVariable(t *testing.T, key, value string) func() {
...
+			newClientErr: errors.New("failed to connect"),
+			expectedErr:  "failed to connect",
+		},
+	}
+
+	for _, testcase := range testcases {
+		t.Run(testcase.doc, func(t *testing.T) {
+			passRetriever := func(_, _ string, _ bool, attempts int) (passphrase string, giveup bool, err error) {
+				switch attempts {
+				case 0, 1, 2:
```
vdemeester: Not sure I understand why those are here — well, it's for `notary.passRetriever` I guess, but those magic number do not help me why we should fail the 3 first time :sweat_smile:

dnephin: There's really no significance, it could also just be just 0

vdemeester: :thinking:

dnephin: Fixed

PR #547, opened 2017-09-21
small reference change

rdxmb:
this makes the first sentence of this section less confusing.

### DIFF
--- a/docs/reference/run.md
+++ b/docs/reference/run.md
@@ -623,7 +623,7 @@ the container exits**, you can add the `--rm` flag:
 
     --rm=false: Automatically remove the container when it exits (incompatible with -d)
 
-> **Note**: When you set the `--rm` flag, Docker also removes the volumes
+> **Note**: When you set the `--rm` flag, Docker also removes the anonymous volumes
 associated with the container when the container is removed. This is similar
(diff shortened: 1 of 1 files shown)


docs/reference/run.md
```diff
@@ -623,7 +623,7 @@ the container exits**, you can add the `--rm` flag:
 
     --rm=false: Automatically remove the container when it exits (incompatible with -d)
 
-> **Note**: When you set the `--rm` flag, Docker also removes the volumes
+> **Note**: When you set the `--rm` flag, Docker also removes the anonymous volumes
```
mstanleyjones: Are you sure it's only anonymous volumes? Is this actually even right? @thaJeztah PTAL

Also you will need to sign the commit before it can be considered for merging.

rdxmb: 1) @mstanleyjones yes, of course. You can even mount this volume again when creating a new container. This is docker design ;) . If you read the following sentences in the docs you will see this fact is even explained there. That's why the first sentence is a bit confusing.
(There is also a great tutorial which explaines all that stuff with all the details: https://nschoe.com/articles/2017-01-28-Docker-Taming-the-Beast-Part-4.html  )

2) ok, signed with lines commented by @GordonTheTurtle

PR #549, opened 2017-09-22
Reset uid/gid to 0 in build context to fix cache busting issues on ADD/COPY in stream mode

shouze:
**- What I did**

Fix of moby/moby#32816 for stream mode. This is a follow up of #513. Also related to https://github.com/tonistiigi/fsutil/pull/10 & https://github.com/moby/buildkit/pull/126

**- How I did it**

By resetting uid/gid to 0/0 in the build context in stream mode.

**- How to verify it**

`go test -v ./cli/command/image -run TestRunBuildResetsUidAndGidInContextWithStreamMode`

Reset uid/gid to 0 in build context to fix cache busting issues on
ADD/COPY in stream mode

160511_sci_frustrated-squirrel jpg crop promo-xlarge2


shouze:
@tonistiigi I currently have this failure, any idea why? I admit I've still not looked for, just written the test as I was believing it should be implemented. Also I can merge `TestRunBuildResetsUidAndGidInContextWithStreamMode` & `TestRunBuildResetsUidAndGidInContext` in a next move and manage stream option as a test case, WDYT?
```
=== RUN   TestRunBuildResetsUidAndGidInContextWithStreamMode
--- FAIL: TestRunBuildResetsUidAndGidInContextWithStreamMode (0.00s)
        Error Trace:    build_test.go:52
	Error:      	Received unexpected error:
	            	open Dockerfile: no such file or directory
	            	failed to open Dockerfile
	            	github.com/docker/cli/cli/command/image.runBuild
	            		/go/src/github.com/docker/cli/cli/command/image/build.go:267
	            	github.com/docker/cli/cli/command/image.TestRunBuildResetsUidAndGidInContextWithStreamMode
	            		/go/src/github.com/docker/cli/cli/command/image/build_test.go:51
	            	testing.tRunner
	            		/usr/local/go/src/testing/testing.go:657
	            	runtime.goexit
	            		/usr/local/go/src/runtime/asm_amd64.s:2197
```

cli/command/image/build_session.go
```diff
@@ -53,7 +54,16 @@ func addDirToSession(session *session.Session, contextDir string, progressOutput
 
 	p := &sizeProgress{out: progressOutput, action: "Streaming build context to Docker daemon"}
 
-	workdirProvider := filesync.NewFSSyncProvider(contextDir, excludes)
+	workdirProvider := filesync.NewFSSyncProvider([]filesync.SyncedDir{
+		{Dir: contextDir},
+		{Excludes: excludes},
+		{Map: func(s *fsutil.Stat) bool {
```
dnephin: Shouldn't all of these be in a single `SyncedDir` instead of 3 separate ones?

I'm alsonot seeing a `Map` field in master,so we should wait for https://github.com/moby/buildkit/pull/126/ to merge and vendor in master instead of the PR branch.

shouze: Yes.. only one you're right 😊 . And yes `Map` is part of https://github.com/moby/buildkit/pull/126 this PR needs to use master of course before merge.

shouze:
@dnephin I've rebased with #294 and the bug remains. I still need to figure out why I have this Dockerfile error in stream mode. I guess I don't reproduce the way that stream mode currently works in the test, will investigate further.

shouze:
@tonistiigi @dnephin ok so I've tooked a look further and... the Dockerfile error is related to:
https://github.com/docker/cli/blob/bd6e1757f5f9097f8277ccadf0493bca94eb0a15/cli/command/image/build.go#L262-L270

BTW, If I comment this statement or if I use a Dockerfile from stdin, then I get this error from the fakeCli:
```
=== RUN   TestRunBuildResetsUidAndGidInContextWithStreamMode
--- FAIL: TestRunBuildResetsUidAndGidInContextWithStreamMode (0.00s)
        Error Trace:    build_test.go:31
	            	client_test.go:121
	            	build.go:390
	            	build_test.go:51
	Error:      	Received unexpected error:
	            	Empty archive
FAIL
```

@tonistiigi looks like it's ok as stream mode maybe don't produce a tar archive in fact?

if I dump some interesting fakeImage args in this test:
- context is nil
- options:
  - SessionID & RemoteContext remains undefined

Maybe because there's an async build process here with the session which is dialed, it there a way to wait against this async process?

dnephin:
I think it is expected that stream doesn't produce an archive, it would send files using the stream.

We need to fake the client lib parts for the session call as well I think, but I'm not sure if that's related to the error.

shouze:
Yes from what I've read stream mode uses diffcopy (was supporting a tarstream mode in the past but removed). Ok will look around to fake client lib parts for the session but yes for the original error about the Dockerfile I still don't know, will probably find on my way. Thank you.

shouze:
Not that easy BTW as I probably have to fake the build_session... but in the same time.... my patch is on `addDirToSession` in build_session! So... maybe it worth instead to write a unit test for this addDirToSession only in addDirToSession?

dnephin:
A unit test for `addDirToSession` seems appropriate.

cli/command/image/build_session_test.go
```diff
@@ -0,0 +1,37 @@
...
+		fs.WithFile("foo", "some content", fs.AsUser(65534, 65534)),
+	)
+	defer dest.Remove()
+
+	contextDir := dest.Path()
+	sharedKey, err := getBuildSharedKey(contextDir)
+	require.NoError(t, err)
+
+	var s *session.Session
+	s, err = session.NewSession(filepath.Base(contextDir), sharedKey)
```
shouze: @dnephin @tonistiigi ok I guess the best is to mock/fake the session here instead of using the true one. @tonistiigi does it worth that builkit provides a session faker at some point? I can put it in the cli ATM.

dnephin: Having the CLI depend on a `type Session interface{...}` instead of the struct would  be great, then it should be easy for us to fake.

A fake as part of the CLI testing code sounds fine to me.

thaJeztah:
this needs a rebase now, @shouze 😢

shouze:
@thaJestah, yes and even if green I need to introduce the session interface and to fake it prior to be able to write the good test.

I’ll be back on this soon.

thaJeztah:
No worries, thanks! Let us know if you need help :+1:

thaJeztah:
friendly nudge @shouze 🤗

vdemeester:
@shouze @dnephin @thaJeztah as I understand it's a bit tied to #294 and other build/build session enhancement. Given the activity, should we close this one and consolidate the work of all this in one PR ? ⚓️

dnephin:
I think this is a bug fix not a cleanup. 

Needs a rebase, I think we've updated the dependencies already, so maybe the PR will be smaller after the rebase.

dnephin:
But I guess there was no response from @shouze  since the ping in Oct, so we can close for now.

If you're still interested in contributing this fix please let us know and we can re-open the PR.

tonistiigi:
@thaJeztah @dnephin It is a big problem that we can't get these trivial fixes that are essentially a single line fix in anymore. The vendored component is already unit tested. With previous repository layout it would have been trivial to add an integration test that tests checks the actual feature (btw I think we still don't have a test for that although the original issue in moby is closed - we only have some tests that run internal functions with mock wrappers in a different repo). Atm we only have a caller that doesn't really have any functionality at all and we expect to mock all the functionality around that caller. It is not reasonable nor will improve quality.

dnephin:
I agree it's an issue. Which is why we need to merge #294 to fix it.

>  Atm we only have a caller that doesn't really have any functionality at all and we expect to mock all the functionality around that caller.

What is the caller?

tonistiigi:
> What is the caller?

Code that only takes configuration passed by user and calls to other packages with public go/http API, then takes the response from these calls and shows to the user.

dnephin:
That sounds like functionality to me. 

Build also does a lot more than that.

thaJeztah:
Is someone able to carry this?

shouze:
@dnephin @tonistiigi @thaJeztah @vdemeester sorry guys I was on holiday past 3 weeks, Yes this PR should be closed anyway, as I wasn't able to push a decent test, this was not trivial to test against buildkit the way runBuild was done.

koreno:
What's the latest on this issue - is there a workaround/solution to cache busting due to ownership change?
I'm currently on docker 19.3

PR #551, opened 2017-09-22
Use a local constant

dnephin:
This constant is going to be removed from jsonlog package in https://github.com/moby/moby/pull/34946


dnephin:
The moby/moby PR has merge so we should get this merged so it doesn't block any re-vendoring in the future.

PR #553, opened 2017-09-23
Added option for pulling quietly

pramodhkp:
**- What I did**
Added an option for quiet docker pulls `docker pull -q`. Based on the discussion here: https://github.com/moby/moby/issues/13588
**- How I did it**

**- How to verify it**

`docker pull -q alpine`

### DIFF
--- a/cli/command/image/pull.go
+++ b/cli/command/image/pull.go
@@ -16,6 +16,7 @@ import (
 type pullOptions struct {
 	remote string
 	all    bool
+	quiet  bool
 }
 
 // NewPullCommand creates a new `docker pull` command
@@ -35,6 +36,7 @@ func NewPullCommand(dockerCli command.Cli) *cobra.Command {
 	flags := cmd.Flags()
 
 	flags.BoolVarP(&opts.all, "all-tags", "a", false, "Download all tagged images in the repository")
+	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress verbose output")
 	command.AddTrustVerificationFlags(flags)
 
 	return cmd
--- a/cli/command/image/trust.go
+++ b/cli/command/image/trust.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/trust"
+	"github.com/docker/cli/opts"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
 	registrytypes "github.com/docker/docker/api/types/registry"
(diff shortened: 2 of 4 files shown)


cpuguy83:
Why is this controlled in the client library?

pramodhkp:
@cpuguy83 The issue is discussed here: https://github.com/moby/moby/issues/13588. Mostly because, piping to `dev/null` or equivalent would work, but would also suppress error information.

So, having a quiet flag is cleaner.

dnephin:
I think the bigger question (which I assume @cpuguy83 was getting at) is why is this even handled in the API? It could be an entirely client-side concern (ignore the bytes sent by pull and only print the relevant errors and final id).

Unfortunately this seems to be the way that `docker build -q` is implemented as well. Build output can be a lot more verbose so it might make more sense for build. I'm not sure we should do the same for pull. It seems like doing it client-side only would be fine here.

pramodhkp:
Maybe. My thoughts were that it would help even when you're making an API call directly to the daemon by skipping the client.

cpuguy83:
@dnephin exactly

Shouldn't error information go to stderr?
I'm not intimately familiar with how the pull stream is parsed.
I know build is using the jsonmessage thing which supports sending errors down the stream.

pramodhkp:
@thaJeztah Hm, I'll check that. Looks like it's probably because `pull_v1.go` defines imgID as a string, while `pull_v2.go` defines imgID as a digest. I'm probably making a mistake while converting. 

And yes, I do agree it's a lot of changes for a small feature. But, I'm not sure wheter that's a good thing or a bad thing.  :smile:

pramodhkp:
@dnephin @cpuguy83 What do you recommend I do with this PR? Modify this to work only on CLI? Or create a new PR?

dnephin:
I think implementing it entirely in the client is a good place to start. In the future if we wanted to add it to the API that would still be possible, but I think it's unlikely.

cpuguy83:
I think a CLI only implementation would work well here.

pramodhkp:
I need some advice implementing this. I'll have to change the `DisplayMessagesToStream` to ignore the progress messages, but the method is being used in multiple other places. Does it make sense to change everywhere?

Is it possible to send `status` outside of stream? That way, I can just ignore the stream, and only print status.

dnephin:
I would suggest not using `DisplayJSONMessagesToStream`, most of the functionality there is not relevant for `quiet`.  I think a local function for iterating over the messages, decoding them, and printing just the parts you want would be great.

thaJeztah:
ping @pramodhkp were you still working on this?

pramodhkp:
@thaJeztah I have implemented half of it. But I've been a bit busy. I will create a new PR when I'm done, if this needs to be closed.

vdemeester:
@pramodhkp carrying in #882 😉

PR #554, opened 2017-09-25
Add a Close method on DockerCli

lgarithm:
**- What I did**
Added a Close method on DockerCli.

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -199,6 +199,14 @@ func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.
 	return client.NewClient(host, verStr, httpClient, customHeaders)
 }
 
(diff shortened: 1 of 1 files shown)


dnephin:
Since this is unused by the cli I don't really like adding this. You can already call close with `DockerCli.Client().Client()`

lgarithm:
DockerCli.Client() returns an interface client.APIClient, which doesn't have a Close method.

* https://github.com/docker/cli/blob/master/cli/command/cli.go#L66
* https://github.com/docker/cli/blob/master/vendor/github.com/docker/docker/client/interface_stable.go#L4

dnephin:
That will be fixed in https://github.com/moby/moby/pull/34899

lgarithm:
That would be nice!
When will it be merged into docker/cli?

PR #558, opened 2017-09-26
Fixed import in loader.go

surajnarwade:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/compose/loader/loader.go
+++ b/cli/compose/loader/loader.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/Sirupsen/logrus"
 	"github.com/docker/cli/cli/compose/interpolation"
 	"github.com/docker/cli/cli/compose/schema"
 	"github.com/docker/cli/cli/compose/template"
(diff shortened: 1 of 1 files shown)


cli/compose/loader/loader.go
```diff
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/Sirupsen/logrus"
```
thaJeztah: This doesn't look correct; the repository was renamed from `Sirupsen/logrus` to `sirupsen/logrus` (see https://github.com/docker/cli/commit/a0113c3a448cb1ed875f2116eb8436ee8bd97bfc, and https://github.com/docker/cli/pull/424)

The correct import path is now with a lowercase `s`: https://github.com/sirupsen/logrus/issues/553

thaJeztah:
Closing per my comment above, but feel free to comment after that

surajnarwade:
@thaJeztah , URL looks like https://github.com/Sirupsen/logrus and it's breaking the code
(S is capital)

surajnarwade:
@thaJeztah my bad, I have not seen that issue before but somehow that is breaking my kompose code

vdemeester:
@surajnarwade the problem is that with the rename, once a dependency has update logrus to reference it with the lowercase `s`, all dependencies should update too. This mean `kubernetes/kompose`, `docker/libcompose`, etc 👼

surajnarwade:
@vdemeester thanks for clearing thing

PR #560, opened 2017-09-26
Support parsing of named pipes for compose volumes

dnephin:
Related to https://github.com/moby/moby/issues/34795

Fixes the client side parsing of volume specs to support windows named pipes.

### DIFF
--- a/cli/compose/loader/volume.go
+++ b/cli/compose/loader/volume.go
@@ -112,6 +112,11 @@ func isFilePath(source string) bool {
 		return true
 	}
 
+	// windows named pipes
+	if strings.HasPrefix(source, `\\`) {
+		return true
+	}
+
 	first, nextIndex := utf8.DecodeRuneInString(source)
 	return isWindowsDrive([]rune{first}, rune(source[nextIndex]))
 }
--- a/cli/compose/loader/volume_test.go
+++ b/cli/compose/loader/volume_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/docker/cli/cli/compose/types"
 	"github.com/docker/cli/internal/test/testutil"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestParseVolumeAnonymousVolume(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cli/compose/loader/volume.go
```diff
@@ -112,6 +112,11 @@ func isFilePath(source string) bool {
 		return true
 	}
 
+	// windows named pipes
+	if strings.HasPrefix(source, `\\`) {
```
StefanScherer: I wonder if we can make this slash-insensitive (like case-insensitive, but for slashes ;-) ).
From a Mac/Linux client remote controlling a Windows swarm node I guess many people still are used to unix slashes.

I also use eg. `docker run -v //./pipe/docker_engine://./pipe/docker_engine -it -p 8080:8080 stefanscherer/visualizer-windows:insider` right now and it works pretty good as Golang itself can handle backslashes and slashes in paths.

Slashes because bash needs extra escaped backslashes as you can see here.

´´´
$ docker run -v \\.\pipe\docker_engine:\\.\pipe\docker_engine -it -p 8080:8080 stefanscherer/visualizer-windows:insider
docker: Error response from daemon: invalid bind mount spec "\\.pipedocker_engine:\\.pipedocker_engine": invalid volume specification: '\.pipedocker_engine:\.pipedocker_engine'.
```


(oops as you can see even GitHub looses some backslashes although I'm using three backticks here)

Comment (dnephin) [reply cli/compose/loader/volume.go]: Normal slash is already handled in this function on the line above (111). A single `/` already returns true.

Comment (StefanScherer) [reply cli/compose/loader/volume.go]: Nice!

**Comment by StefanScherer** [comment]
It's so convenient to pick the Docker CLI from CircleCI build. So I gave it a try

```
$ ./docker-darwin-amd64 service create --name whoami --mount type=bind,source=//./pipe/docker_engine,destination=//./pipe/docker_engine stefanscherer/whoami:insider
vkwsqpgcgm3nso31nw5no8fcx
1/1: invalid mount target, must be an absolute path: //./pipe/docker_engine 
```

And with backslashes

```
$ ./docker-darwin-amd64 service create --name whoami --mount 'type=bind,source=\\.\pipe\docker_engine,destination=\\.\pipe\docker_engine' stefanscherer/whoami:insider
cruvatiet5yhuxta2teppqw8b
1/1: invalid mount target, must be an absolute path: \\.\pipe\docker_engine 
```

running against a Windows Insider with Docker engine 17.09.0-rc2 installed.

dnephin:
Yes, as I mentioned in https://github.com/moby/moby/issues/34795#issuecomment-332235854, this only fixes the client side. `docker service create` does not do any parsing on the client side. This still needs to be fixed in moby/moby (daemon/cluster/executor/container/validate.go).

Issue #559, opened 2017-09-26
Proposal: deprecate base64 credentials in config.json


n4ss:
We think w/ @tych0 that storing the credentials directly in a config file in home directory is not safe at all and we should remove this option as the credential management has improved a lot on every platform since then.

@tych0 has implemented a new (and lightweight) backend for credential-helpers on Linux based on `pass` so now users can easily store credentials on Linux, Windows and MacOS. His `docker/cli` PR also got merged: https://github.com/docker/cli/pull/451 so we should be good to go on removing plaintext password.

/cc @vdemeester @thaJeztah

tych0:
I guess we'll need to deprecate it and print a big warning and such, per the deprecation policy. I'm happy to send a patch that does this if we decide we want it.

thaJeztah:
Deprecating it would finally address https://github.com/moby/moby/issues/10318 :tada: - this would require a longer deprecation period though, because this has been supported for a _really_ long time. Perhaps we may even have to consider having an environment variable that allows keeping the old storage after it's been deprecated (but happy to hear if that's taking it too far) 😅

vdemeester:
> this would require a longer deprecation period though, because this has been supported for a really long time.

So let's try to deprecate it as soon as possible :stuck_out_tongue: :angel: :pray: 

cc @dnephin @vieux @tiborvass @cpuguy83

tiborvass:
@n4ss does this mean the docker-ce deb/rpm package would have a dependency on `pass` ? Do we care about behavior of static builds used on more esoteric platforms? 

I don't think we can completely remove the insecure option: where there is no `pass`, and when `docker pull` is called in scripts (not a tty). We just have to inform very explicitly that it is insecure, and do our best to have secure defaults on the most platforms.

tych0:
On Tue, Sep 26, 2017 at 05:31:25PM +0000, Tibor Vass wrote:
> @n4ss does this mean the docker-ce deb/rpm package would have a dependency on `pass` ? Do we care about behavior of static builds used on more esoteric platforms? 

It depends, but yes, I think it should. There's nothing that links
against `pass` binaries or anything, so you could install it, and the
pass helper would just try to exec the binary and fail, and default to
secretservice.

> I don't think we can completely remove the insecure option: where there is no `pass`, and when `docker pull` is called in scripts (not a tty). We just have to inform very explicitly that it is insecure, and do our best to have secure defaults on the most platforms.

What's the problem with docker pull? As far as I can tell, it just
[...]

PR #561, opened 2017-09-26
Deprecate unencrypted storage

tych0:
This is an implementation of #559.

cc @n4ss @thaJeztah @vdemeester @dnephin @vieux @tiborvass @cpuguy83

### DIFF
--- a/cli/command/registry/login.go
+++ b/cli/command/registry/login.go
@@ -17,6 +17,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const unencryptedWarning = `WARNING! Your password will be stored unencrypted in %s.
+Configure a credential helper to remove this warning. See
+https://docs.docker.com/engine/reference/commandline/login/#credentials-store
+`
+
 type loginOptions struct {
 	serverAddress string
 	user          string
@@ -50,6 +55,29 @@ func NewLoginCommand(dockerCli command.Cli) *cobra.Command {
 	return cmd
 }
 
+// unencryptedPrompt prompts the user to find out whether they want to continue
+// with insecure credential storage. If stdin is not a terminal, we assume they
+// want it (sadly), because people may have been scripting insecure logins and
+// we don't want to break them. Maybe they'll see the warning in their logs and
+// fix things.
+func unencryptedPrompt(dockerCli command.Streams, filename string) error {
+	fmt.Fprintln(dockerCli.Err(), fmt.Sprintf(unencryptedWarning, filename))
+
+	if dockerCli.In().IsTerminal() {
+		if command.PromptForConfirmation(dockerCli.In(), dockerCli.Out(), "") {
+			return nil
+		}
+		return errors.Errorf("User refused unencrypted credentials storage.")
+	}
+
+	return nil
+}
+
+type isFileStore interface {
+	IsFileStore() bool
+	GetFilename() string
+}
+
 func verifyloginOptions(dockerCli command.Cli, opts *loginOptions) error {
 	if opts.password != "" {
 		fmt.Fprintln(dockerCli.Err(), "WARNING! Using --password via the CLI is insecure. Use --password-stdin.")
@@ -74,7 +102,7 @@ func verifyloginOptions(dockerCli command.Cli, opts *loginOptions) error {
 	return nil
 }
 
-func runLogin(dockerCli command.Cli, opts loginOptions) error {
+func runLogin(dockerCli command.Cli, opts loginOptions) error { //nolint: gocyclo
 	ctx := context.Background()
 	clnt := dockerCli.Client()
 	if err := verifyloginOptions(dockerCli, &opts); err != nil {
@@ -113,7 +141,18 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
 		authConfig.Password = ""
 		authConfig.IdentityToken = response.IdentityToken
 	}
-	if err := dockerCli.ConfigFile().GetCredentialsStore(serverAddress).Store(*authConfig); err != nil {
+
+	creds := dockerCli.ConfigFile().GetCredentialsStore(serverAddress)
+
+	store, isDefault := creds.(isFileStore)
+	if isDefault {
+		err = unencryptedPrompt(dockerCli, store.GetFilename())
+		if err != nil {
+			return err
+		}
+	}
+
+	if err := creds.Store(*authConfig); err != nil {
 		return errors.Errorf("Error saving credentials: %v", err)
 	}
 
--- a/cli/config/credentials/file_store.go
+++ b/cli/config/credentials/file_store.go
@@ -8,6 +8,7 @@ import (
 type store interface {
 	Save() error
 	GetAuthConfigs() map[string]types.AuthConfig
+	GetFilename() string
 }
 
 // fileStore implements a credentials store using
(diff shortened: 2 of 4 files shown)


cli/config/credentials/file_store.go
```diff
@@ -8,27 +8,28 @@ import (
...
 	Save() error
 	GetAuthConfigs() map[string]types.AuthConfig
+	GetFilename() string
 }
 
-// fileStore implements a credentials store using
+// FileStore implements a credentials store using
 // the docker configuration file to keep the credentials in plain text.
-type fileStore struct {
+type FileStore struct {
```
dnephin: Instead of exporting this (something we want to deprecate) how about adding a new `IsNotSecure()` method to `fileStore`, and creating the interface in `cli/command/registry/login.go` to check it?

```go

type notSecure interface {
    IsNotSecure() bool
}

if _, isNotSecure := creds.(notSecure); isNotSecure {
 ...
}

```

cli/command/registry/login.go
```diff
@@ -96,7 +104,37 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
...
+	store, isDefault := creds.(*credentials.FileStore)
+	if isDefault {
+		fmt.Fprintln(dockerCli.Err(), fmt.Sprintf(unencryptedWarning, store.GetFilename()))
+
+		// If this is not a terminal, let's assume they want it
+		// (sadly), because people may have been scripting insecure
+		// logins and we don't want to break them. Maybe they'll see
+		// the warning in their logs and fix things. If it is
+		// interactive, let's ask them if they really want to do this.
+		if dockerCli.In().IsTerminal() {
```
dnephin: Please extract the entire confirmation block into a new function so the comment can be a godoc string instead of inline comment.

cli/command/registry/login.go
```diff
@@ -96,7 +104,37 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error {
...
+			for {
+
+				fmt.Fprintf(dockerCli.Err(), "Do you want to continue? (yes/no): ")
+				choice, _, err := bufio.NewReader(dockerCli.In()).ReadLine()
```
dnephin: Please use `PromptForConfirmation` https://github.com/docker/cli/blob/master/cli/command/utils.go#L67-L72

cli/config/credentials/file_store.go
tych0: I thought about this, but as it stands there are two backends: fileStore and nativeStore, which is what implements the docker credential helpers interface. You could imagine a credential helper that just wrote plain text to a file, which would not be secure, but writing "!isNotSecure" seems to imply that the backend is secure, when it's not: nativeStore can't vouch for what's on the other end.

dnephin: Ok, then how about `IsDeprecated()` and `type Deprecated interface{...}` ?

tiborvass:
FYI This is in design phase.

cli/command/registry/login.go
```diff
@@ -9,11 +10,18 @@ import (
...
+	"github.com/docker/cli/cli/config/credentials"
 	"github.com/docker/docker/registry"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
+const unencryptedWarning = `WARNING! Your password will be stored unencrypted in %s.
+Please configure a credential helper (see
+https://docs.docker.com/engine/reference/commandline/login/#credentials-store)
+to avoid this.
```
dnephin: How about:

```
https://docs.docker.com/engine/reference/commandline/login/#credentials-store
```

docs/deprecated.md
```diff
@@ -20,6 +20,17 @@ The following list of features are deprecated in Engine.
...
 
+### Unencrypted password storage via `docker login`
+
+**Deprecated In Release: v17.10.0**
+
+**Target For Removal In Release: v18.10**
+
+Docker before v17.10 stored user credentials in `~/.docker/config.json`
+unencrypted by default. As of v17.10 it prints a warning with a link to the
+credentials helper page and asks users if they want to continue. Eventually, we
```
thaJeztah: I'd skip the niceties and just, e.g.: "Support for unencrypted password storage will be removed in Docker 18.10."

cli/command/registry/login.go
thaJeztah: 👍 Better without the "please"

docs/deprecated.md
```diff
@@ -20,6 +20,17 @@ The following list of features are deprecated in Engine.
 To learn more about Docker Engine's deprecation policy,
 see [Feature Deprecation Policy](https://docs.docker.com/engine/#feature-deprecation-policy).
 
+### Unencrypted password storage via `docker login`
+
+**Deprecated In Release: v17.10.0**
+
+**Target For Removal In Release: v18.10**
```
tiborvass: I'm not sure this will be well received. As @cpuguy83 mentioned it, what about CI ? Same for any other situation where there is no secure alternative (`pass`, ...) and that it is run as a script (no tty)

cli/config/credentials/file_store.go
tych0: I thought about this too, but in the event that we don't deprecate this based on discussion, I think we should still merge this patch, but tagging it as deprecated would be incorrect. We could cast by GetFilename, since that is unique to the FileStore, but it doesn't really describe what we're trying to detect. So I dunno :)

dnephin: Feel free to pick some other interface, my concern is just that it doesn't make sense to export this entire struct if all we need is some way detect that it is of this type.

Another option is to export an `IsFileStore()`. That way we're only exporting one function and not  the entire struct.

docs/deprecated.md
dnephin: I agree. Warning about it being insecure is good, but I don't think we should actually set a target for removal.

tych0: Reasonable questions. I'm not sure what the threat model is here, I guess it must be people concerned about folks outside their organization reading their images off the docker (or other publicly accessible) hub? It seems that anyone who can commit code to the thing in CI can read out these passwords if they're stored in plaintext, so it can't be about internal threats.

If that's the model, perhaps we can do something a-la Google's app-specific passwords (which we could implement before deprecating this)? If it's not, then this is broken and we should in fact remove it.

tych0: Ok, so after some discussions with @n4ss, it seems that this is exactly what people are doing. So without some reasonable replacement for this, we can't deprecate this bit. So, I'll just drop the second patch from this PR, and leave it with the big warning patch.

thaJeztah:
@gbarr01 wondering if we should have special URLs that we can use when referring to the docs from the command-line. I know some URLs may move, and can become outdated; should we have (e.g.) docs URLs that redirect to where we want them? like `https://docs.docker.com/help/credentials-helpers` (and have it redirect to https://docs.docker.com/engine/reference/commandline/login/#credentials-store)?

gbarr01:
@thaJeztah, sure thing. To confirm, the benefit here that the developer can always use the same base URL (`https://docs.docker.com/help/`) even if we are pointing credentials-helper to credentials-store.

thaJeztah:
@gbarr01 almost, yes; we could use `/help/<some topic>` (in this case “credential-helpers”) and have those redirect

PR #562, opened 2017-09-26
Update gometalinter

dnephin:
Update gometalinter to pickup some bug fixes and updates to linters
Adjust the config to exclude a new category of errors added by unparam. Functions which always accept the same value for a parameter are just fine I think.
Refactor image pull and trust slightly to make use of the new struct added for the trust commands.

These new `// nolint: unparam` are for another new check added to unparam where it errors if no callers use the return value. I think this is ok in some cases, but a good check in others, so I just whitelisted a few.

### DIFF
--- a/gometalinter.json
+++ b/gometalinter.json
@@ -1,8 +1,12 @@
 {
   "Vendor": true,
   "Deadline": "2m",
-  "Sort": ["linter", "severity", "path"],
-  "Exclude": ["cli/compose/schema/bindata.go"],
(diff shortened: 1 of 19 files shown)


gometalinter.json
```diff
@@ -22,6 +26,6 @@
     "vet"
   ],
 
-  "Cyclo": 19,
+  "Cyclo": 16,
```
riyazdf: nice :+1:

riyazdf:
cc @eiais @ashfall

PR #564, opened 2017-09-27
Fix leading characters being stripped from example sections in YAML

thaJeztah:
`strings.Trim()` strips any character listed in the `cutset` argument,
so any example section having `E`, `x`, `a`, `m`, `p`, `l`, `e`, or `s`
in the first word, had these characters missing in the generated
YAML.

Also trim superfluent whitespace characters to consistently use `|-` ("strip")
as block chomping indicator (see http://www.yaml.org/spec/1.2/spec.html#id2794534)

Before:

screen shot 2017-09-27 at 10 30 03

After:

screen shot 2017-09-27 at 11 01 08

### DIFF
--- a/docs/yaml/yaml.go
+++ b/docs/yaml/yaml.go
@@ -220,10 +220,10 @@ func parseMDContent(mdString string) (description string, examples string) {
 	parsedContent := strings.Split(mdString, "\n## ")
 	for _, s := range parsedContent {
 		if strings.Index(s, "Description") == 0 {
-			description = strings.Trim(s, "Description\n")
+			description = strings.TrimSpace(strings.TrimPrefix(s, "Description"))
 		}
 		if strings.Index(s, "Examples") == 0 {
-			examples = strings.Trim(s, "Examples\n")
+			examples = strings.TrimSpace(strings.TrimPrefix(s, "Examples"))
 		}
 	}
 	return


docs/yaml/yaml.go
```diff
@@ -220,10 +220,10 @@ func parseMDContent(mdString string) (description string, examples string) {
 	parsedContent := strings.Split(mdString, "\n## ")
 	for _, s := range parsedContent {
 		if strings.Index(s, "Description") == 0 {
-			description = strings.Trim(s, "Description\n")
+			description = strings.TrimSpace(strings.TrimPrefix(s, "Description"))
 		}
 		if strings.Index(s, "Examples") == 0 {
-			examples = strings.Trim(s, "Examples\n")
+			examples = strings.TrimSpace(strings.TrimPrefix(s, "Examples"))
```
dnephin: actually, looking at this again. I think the `TrimSpace()` wont do what you want, because there are characters after this line.  I think it still needs to have the trailing `\n` in the TrimPrefix

The contents of `s` is a multiline string that includes the paragraphs after the header.

thaJeztah: It will first remove the `Examples` word; after that, the line is empty (or contains whitespace only); trimming the whitespace then cleans up all whitespace surrounding the section; https://play.golang.org/p/XrM2WQWCjlA

PR #566, opened 2017-09-27
docker trust: interact with signers and keys

riyazdf:
This PR introduces four more `docker trust` subcommands to interact with signers and keys:

- `docker trust signer add`: add a signer to one or more repos
- `docker trust signer remove`: remove a signer from one or more repos
- `docker trust key load`: load a private key into the local storage to use for signing
- `docker trust key generate`: generate a signing key-pair.  The private key is automatically loaded into the local storage, and the public key can be used for `signer add`

</img>

cc @ashfall and @eiais who also implemented these features

This PR includes code and documentation and extends the functionality in #472. As such, it can be included in a followup release.

[Original design doc here](https://docs.google.com/document/d/1JOBAlCDuf5JnpVLW54voGuAdsnMmSR2LIbs8fjBDSaM/edit )


thaJeztah:
Just from the description; should we think of making `signer` and `key` top-level objects, to make it fit better with the existing UX (`docker signer add`, `docker signer remove`, `docker key load`, `docker key generate`), or alternatively `docker trust-key load`, `docker trust-key generate`

riyazdf:
@thaJeztah: we kept `key` and `signer` under the `trust` command because the concepts probably won't be used elsewhere - that said I'm cool with factoring them out to `docker signer`/`docker key` or `docker trust-signer`/`docker trust-key` if you think that would be better.

For most users the flows will look something like:

```
# Add a signer and view to confirm
docker trust signer-add alice repo
docker trust view repo
```

```
# Load my signer key and sign with it
docker trust key-load alice.priv
docker trust sign repo
```

dnephin:
I like it either as it is now, or as another level under `docker trust` `docker trust signer add` , `docker trust key load`. 

Moving them to the top level would be confusing.

thaJeztah:
```
docker trust signer add
```

works for me; would that be more difficult to implement (as we currently don't have sub-sub commands)?

dnephin:
Shouldn't be. All we need is a couple additional `cobra.Command` objects for `signer` and `key` and add the new commands to them instead of the `Command` for trust.

riyazdf:
@dnephin @thaJeztah : sounds good!  Are we okay with these being one of the only (if not the only) commands in `docker/cli` with four levels of subcommands?

If there's consensus I'm happy to make the change.

thaJeztah:
I'm OK with it; it may be slightly awkward to have the commands nested this deep, but with the context given above, I think it's the best option (and most consistent with other commands as in `<something>  add`)

ping @docker/core-cli-maintainers PTAL

albers:
The completions are not equipped for another nesting level, so I'd prefer the less deeply nested forms.

vieux:
I'm ok with `docker trust signer add` and `docker trust key load`

We almost went for `docker swarm node update` instead of a top level`docker node` so I'm not shocked by this depth

riyazdf:
Ok!  From discussions with @dnephin, @thaJeztah, @vdemeester, @stevvooe, and @vieux : it sounds like `docker trust signer add`et al. is the best path forward.  I will update this PR shortly.

I'm not sure what the implications are for bash completions, is it possible to make it work with this level deep of commands?

albers:
For bash completion it will be 0.5 to 3 days of work.
The current implementation uses global variables for the command and subcommand positions. It might be hard to turn this into a proper generic solution.

I'm also not sure about the implications on zsh and fish completion.
@sdurrheimer can zsh completion deal with an additional level of subcommands?

riyazdf:
@albers: got it.

I've moved the commands around to be a level lower, now we have:
```
🐳 $ ./docker-darwin-amd64 trust

Usage:	docker trust COMMAND

Manage trust on Docker images (experimental)

Options:
      --help   Print usage

Management Commands:
  key         Manage keys for signing Docker images (experimental)
  signer      Manage entities who can sign Docker images (experimental)

Commands:
  revoke      Remove trust for an image
  sign        Sign an image
  view        Display detailed information about keys and signatures

Run 'docker trust COMMAND --help' for more information on a command.
```

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+
+	newSignerRoleName := data.RoleName(path.Join(data.CanonicalTargetsRole.String(), signerName))
+
+	signerPubKeys, err := ingestPublicKeys(keyPaths)
+	if err != nil {
+		return err
+	}
+	addStagedSigner(notaryRepo, newSignerRoleName, signerPubKeys)
```
dnephin: Not related to this PR, but I just took a look at `addStagedSigner()` and I noticed that it calls 4 interface methods that all return an error, but none of the errors are checked. Is that safe? Does the implementation not return important errors?

cli/command/trust/signer_remove.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func newSignerRemoveCommand(dockerCli command.Cli) *cobra.Command {
+	options := signerRemoveOptions{}
+	cmd := &cobra.Command{
+		Use:   "remove [OPTIONS] NAME IMAGE [IMAGE...]",
+		Short: "Remove a signer",
+		Args:  cli.RequiresMinArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return removeSigner(dockerCli, args[0], args[1:], &options)
```
dnephin: Most commands (including `newSignerAddCommand` above) pass args using the options struct instead of separate args. I think it would be good to do that here as well.

Also it shouldn't be necessary to use a pointer for options. The values should never be modified, so I believe we pass a struct (not a pointer) in all other commands.

cli/command/trust/signer_remove.go
```diff
@@ -0,0 +1,137 @@
...
+		},
+	}
+	flags := cmd.Flags()
+	flags.BoolVarP(&options.forceYes, "yes", "y", false, "Answer yes to removing most recent signer (no confirmation)")
```
dnephin: This should probably say something like "do not prompt for confirmation before ..." or something along those lines

cli/command/trust/signer_remove.go
```diff
@@ -0,0 +1,137 @@
...
+		for _, key := range roleWithSig.KeyIDs {
+			if signature.KeyID == key {
+				counter--
+			}
+		}
+	}
+	return counter < releasesRoleWithSigs.Threshold, nil
+}
+
+func removeSingleSigner(cli command.Cli, image, signerName string, forceYes bool) error {
```
dnephin: When `signerName` is part of options I think `signerName` and `forceYes` can be replaced with `options signerRemoveOptions`

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	flags := cmd.Flags()
+	options.keys = opts.NewListOpts(nil)
+	flags.VarP(&options.keys, "key", "k", "Path to the signer's public key(s)")
+	return cmd
+}
+
+var validSignerName = regexp.MustCompile(`^[a-z0-9]+[a-z0-9\_\-]*$`).MatchString
+
+func addSigner(cli command.Cli, options *signerAddOptions) error {
```
dnephin: should accept the struct, not pointer to struct

vendor.conf
```diff
@@ -1,5 +1,9 @@
 github.com/agl/ed25519 d2b94fd789ea21d12fac1a4443dd3a3f79cda72c
 github.com/Azure/go-ansiterm 19f72df4d05d31cbe1c56bfc8045c96babff6c7e
+github.com/bugsnag/bugsnag-go 13fd6b8acda029830ef9904df6b63be0a83369d0
+github.com/bugsnag/osext 0dd3f918b21bec95ace9dc86c7e70266cfc5c702
+github.com/bugsnag/panicwrap e2c28503fcd0675329da73bf48b33404db873782
+github.com/BurntSushi/toml a368813c5e648fee92e5f6c30e3944ff9d5e8895
```
dnephin: This is bringing in a whole lot of unnecessary dependencies.

I believe the problem is `github.com/docker/notary/util`. I think this illustrates one of the major problems with a `utils` package. 

from https://blog.golang.org/package-names

> Avoid meaningless package names. Packages named util ...

A generic package like `utils` tends to collect a bunch of unrelated functions/types that have completely different dependencies which means that using any of the functions results in the consumer having to depend on a bunch of unrelated things (sql libraryes, logging libraries, etc).

I think this needs to be fixed in notary and anything used in `docker/cli` needs to be moved out of `utils` so we can use it without bringing in a dozen unnecessary packages.

cli/command/trust/key_generate.go
```diff
@@ -0,0 +1,107 @@
...
+	if _, err := os.Stat(filepath.Join(cwdPath, pubKeyFileName)); err == nil {
+		return fmt.Errorf("public key file already exists: \"%s\"", pubKeyFileName)
+	}
+	return nil
+}
+
+func setupPassphraseAndGenerateKeys(streams command.Streams, keyName string) error {
+	// always use a fresh passphrase for each key generation
+	freshPassRetGetter := func() notary.PassRetriever { return trust.GetPassphraseRetriever(streams.In(), streams.Out()) }
+	cwd, err := os.Getwd()
```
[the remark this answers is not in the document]
riyazdf: agreed!  Added.

cli/command/trust/signer_remove.go
riyazdf: I left this function signature as is because otherwise there's a weird discrepancy between the unpacked image named and otherwise packed options.

vendor.conf
riyazdf: I'm looking into it - it's possible that this may not be so clean because the notary server-side services also use a key storage interface with connects with the DB (hence many of the imports).  I'll get back to you with an update.

riyazdf:
thanks @dnephin for the review!  I'm looking into package refactors on `docker/notary` and have applied your other suggestions :)

riyazdf:
@dnephin: great news!  I have a fix in flight for upstream notary that will remove all vendoring except for an updated notary commit hash.

We would love to move this forward before the next release - please let me know if there are other review comments I can address :)

cli/command/trust/signer_remove.go
dnephin: I don't really see how this is a problem. It's quite common for a function to only use some fields or methods on a struct.

riyazdf: @dnephin: I don't think that is a safe practice: this is an inner helper function that only needs to know about one repo.  Including the extra images in the slice is extraneous info, and I worry that a future contribution may misuse it.

That said, if you feel strongly about changing to the `signerRemoveOptions`, you are the maintainer and I will make the change - just let me know.

riyazdf:
updated: the only vendoring is now limited to an updated `docker/notary` commit sha.  Excited to move forward with this!

cli/command/trust/key_generate.go
```diff
@@ -0,0 +1,130 @@
...
+	// Automatically load the private key to local storage for use
+	privKeyFileStore, err := trustmanager.NewKeyFileStore(trust.GetTrustDirectory(), freshPassRetGetter())
+	if err != nil {
+		return err
+	}
+
+	pubPEM, err := generateKeyAndOutputPubPEM(keyName, privKeyFileStore)
+	if err != nil {
+		fmt.Fprintf(streams.Out(), err.Error())
+		return fmt.Errorf("Error generating key for: %s", keyName)
```
dnephin: I don't think it's correct to print an error on stdout. This should return an error without the `fmt.Fprintf()`

```go
return errors.Wrapf(err, "failed to generate key for %s", keyName)
```

cli/command/trust/key_generate.go
```diff
@@ -0,0 +1,130 @@
...
+		Bytes: pubKey.Public(),
+	}, nil
+}
+
+func writePubKeyPEMToDir(pubPEM pem.Block, keyName, workingDir string) (string, error) {
+	// Output the public key to a file in the CWD or specified dir
+	pubFileName := strings.Join([]string{keyName, "pub"}, ".")
+	pubFilePath := filepath.Join(workingDir, pubFileName)
+	if err := ioutil.WriteFile(pubFilePath, pem.EncodeToMemory(&pubPEM), notary.PrivNoExecPerms); err != nil {
+		return "", fmt.Errorf("Error writing public key to location: %s", pubFilePath)
```
dnephin: Maybe wrap the original error so it's easier to debug? Also errors are supposed to always start with lowercase so that they can be wrapped by other code.

```go
return "', errors.Wrapf(err, "failed to write public key to %s", pubFilePath)
```

cli/command/trust/key_load.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	privKeyImporters := []trustmanager.Importer{keyFileStore}
+
+	fmt.Fprintf(streams.Out(), "\nLoading key from \"%s\"...\n", keyPath)
+
+	// Always use a fresh passphrase retriever for each import
+	passRet := trust.GetPassphraseRetriever(streams.In(), streams.Out())
+	keyBytes, err := getPrivKeyBytesFromPath(keyPath)
+	if err != nil {
+		return fmt.Errorf("error reading key from %s: %s", keyPath, err)
```
dnephin:
```go
return errors.Wrapf(err, "error reading key from %s", keyPath)
```

cli/command/trust/key_load.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	if err := loadPrivKeyBytesToStore(keyBytes, privKeyImporters, keyPath, options.keyName, passRet); err != nil {
+		return fmt.Errorf("error importing key from %s: %s", keyPath, err)
```
dnephin: `errors.Wrapf()`

cli/command/trust/key_load.go
```diff
@@ -0,0 +1,94 @@
...
+	return nil
+}
+
+func getPrivKeyBytesFromPath(keyPath string) ([]byte, error) {
+	fileInfo, err := os.Stat(keyPath)
+	if err != nil {
+		return nil, err
+	}
+	if fileInfo.Mode() != ownerReadOnlyPerms && fileInfo.Mode() != ownerReadAndWritePerms {
+		return nil, fmt.Errorf("private key permission from %s should be set to 400 or 600", keyPath)
```
dnephin: What about 700, or 200? Maybe this could say "private key must not be readable by others" or something like that? 

I think the ssh client says something like this: "It is required that your private key files are NOT accessible by others."

cli/command/trust/key_load.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer from.Close()
+
+	keyBytes, err := ioutil.ReadAll(from)
+	if err != nil {
+		return nil, err
+	}
+	return keyBytes, nil
```
dnephin: hmm, not sure why gosimple isn't failing here. This should be:

```go
return ioutil.ReadAll(from)
```

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	if options.keys.Len() < 1 {
+		return fmt.Errorf("path to a valid public key must be provided using the `--key` flag")
+	}
+
+	var errImages []string
+	for _, imageName := range options.images {
+		if err := addSignerToImage(cli, signerName, imageName, options.keys.GetAll()); err != nil {
+			fmt.Fprintln(cli.Out(), err.Error())
```
dnephin: Should go to `cli.Err()` instead of stdout

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,138 @@
...
+func ingestPublicKeys(pubKeyPaths []string) ([]data.PublicKey, error) {
+	pubKeys := []data.PublicKey{}
+	for _, pubKeyPath := range pubKeyPaths {
+		// Read public key bytes from PEM file
+		pubKeyBytes, err := ioutil.ReadFile(pubKeyPath)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return nil, fmt.Errorf("file for public key does not exist: %s", pubKeyPath)
+			}
+			return nil, fmt.Errorf("unable to read public key from file: %s", pubKeyPath)
```
dnephin: Maybe `errors.Wrapf()` to make it easier to debug with the original error.

cli/command/trust/signer_remove.go
```diff
@@ -0,0 +1,137 @@
...
+	flags := cmd.Flags()
+	flags.BoolVarP(&options.forceYes, "yes", "y", false, "Do not prompt for confirmation before removing the most recent signer")
+	return cmd
+}
+
+func removeSigner(cli command.Cli, options signerRemoveOptions) error {
+	var errImages []string
+	for _, image := range options.images {
+		if err := removeSingleSigner(cli, image, options.signer, options.forceYes); err != nil {
+			fmt.Fprintln(cli.Out(), err.Error())
```
dnephin: `s/cli.Out()/cli.Err()/`

cli/command/trust/signer_remove.go
```diff
@@ -0,0 +1,137 @@
...
+	if signerDelegation == releasesRoleTUFName {
+		return fmt.Errorf("releases is a reserved keyword and cannot be removed")
+	}
+	notaryRepo, err := cli.NotaryClient(imgRefAndAuth, trust.ActionsPushAndPull)
+	if err != nil {
+		return trust.NotaryError(imgRefAndAuth.Reference().Name(), err)
+	}
+	delegationRoles, err := notaryRepo.GetDelegationRoles()
+	if err != nil {
+		return fmt.Errorf("Error retrieving signers for %s", imageName)
```
dnephin: Maybe `errors.Wrapf()`?

cli/command/trust/signer_remove.go
```diff
@@ -0,0 +1,137 @@
...
+			"Are you sure you want to continue?",
+			signerName, imageName, imageName,
+		))
+
+		if !removeSigner {
+			fmt.Fprintf(cli.Out(), "\nAborting action.\n")
+			return nil
+		}
+	} else if err != nil {
+		fmt.Fprintln(cli.Out(), err.Error())
```
dnephin: Shouldn't this return the error here to exit early?

If not I guess just `s/cli.Out/cli.Err/`

cli/command/trust/key_load.go
riyazdf: I don't think 200 or 700 make sense here, for 200 you'd have writable but not readable (I've never seen this done?), and with 700 the key would be executable (also don't see a use for this?).

riyazdf:
@dnephin: thanks for the quick turnaround!  Addressed all feedback and commented on file perms

cli/command/trust/key_load.go
dnephin: I agree those would be silly permissions. My point is that the error message is misleading. The requirement is that it can't be readable by others, not that it must be one of those two permissions. The error message should match the requirement. The ssh client is a good example of this.

If I'm not mistaken this error message will also be used on windows. "set to 400 or 600" is meaningless to windows users.

dnephin: Ah, I just noticed the code *does* actually check for specifically those permissions. I think this needs to be changed to just check that the specific bits are not set.

riyazdf: sounds good, updated!

cli/command/trust/key_generate.go
```diff
@@ -0,0 +1,131 @@
...
+
+type keyGenerateOptions struct {
+	name      string
+	directory string
+}
+
+func newKeyGenerateCommand(dockerCli command.Streams) *cobra.Command {
+	options := keyGenerateOptions{}
+	cmd := &cobra.Command{
+		Use:   "generate NAME [NAME...]",
```
thaJeztah: Usage is incorrect here, because only one `NAME` can be provided

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func ingestPublicKeys(pubKeyPaths []string) ([]data.PublicKey, error) {
+	pubKeys := []data.PublicKey{}
+	for _, pubKeyPath := range pubKeyPaths {
+		// Read public key bytes from PEM file
+		pubKeyBytes, err := ioutil.ReadFile(pubKeyPath)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return nil, fmt.Errorf("file for public key does not exist: %s", pubKeyPath)
```
thaJeztah: Looks like we don't have to special case `not exist` errors. Standard errors look to provide the right information. When removing the special case, it looks like:

```
unable to read public key from file: /no-such-key: open /no-such-key: no such file or directory

unable to read public key from file: /foo/foobar.pub: open /foo/foobar.pub: permission denied
```

So no `Wrapf()` needed as well, just:

```go
return nil, errors.Wrap(err, "unable to read public key from file")
```

cli/command/trust/key_generate.go
```diff
@@ -0,0 +1,131 @@
...
+	pubKeyFileName := keyName + ".pub"
+	if _, err := os.Stat(filepath.Join(cwdPath, pubKeyFileName)); err == nil {
+		return fmt.Errorf("public key file already exists: \"%s\"", pubKeyFileName)
+	}
+	return nil
+}
+
+func setupPassphraseAndGenerateKeys(streams command.Streams, opts keyGenerateOptions) error {
+	targetDir := opts.directory
+	// if the target dir is empty, default to CWD
```
thaJeztah: This comment is a bit redundant as it's just the code "in words", also leaning towards having the validation in this step, instead of further down in `validateAndGenerateKey()`, something like:

```patch
diff --git a/cli/command/trust/key_generate.go b/cli/command/trust/key_generate.go
index 913083be..2cb00401 100644
--- a/cli/command/trust/key_generate.go
+++ b/cli/command/trust/key_generate.go
@@ -59,25 +59,23 @@ func validateKeyArgs(keyName string, cwdPath string) error {
 }

 func setupPassphraseAndGenerateKeys(streams command.Streams, opts keyGenerateOptions) error {
-       targetDir := opts.directory
-       // if the target dir is empty, default to CWD
-       if targetDir == "" {
+       if opts.directory == "" {
                cwd, err := os.Getwd()
                if err != nil {
                        return err
                }
-               targetDir = cwd
+               opts.directory = cwd
        }
-       return validateAndGenerateKey(streams, opts.name, targetDir)
-}
-
-func validateAndGenerateKey(streams command.Streams, keyName string, workingDir string) error {
-       freshPassRetGetter := func() notary.PassRetriever { return trust.GetPassphraseRetriever(streams.In(), streams.Out()) }
-       if err := validateKeyArgs(keyName, workingDir); err != nil {
+       if err := validateKeyArgs(opts.name, opts.directory); err != nil {
                return err
        }
+       return generateKey(streams, opts.name, opts.directory)
+}
+
+func generateKey(streams command.Streams, keyName string, workingDir string) error {
        fmt.Fprintf(streams.Out(), "\nGenerating key for %s...\n", keyName)
        // Automatically load the private key to local storage for use
+       freshPassRetGetter := func() notary.PassRetriever { return trust.GetPassphraseRetriever(streams.In(), streams.Out()) }
        privKeyFileStore, err := trustmanager.NewKeyFileStore(trust.GetTrustDirectory(), freshPassRetGetter())
        if err != nil {
                return err
```

thaJeztah:
I also see there's still quite a few places that print "informational" information on `stdout`; that may be problematic if people want to script these things (perhaps not a concern, given that passwords have to be entered interactively)

cli/command/trust/helpers.go
```diff
@@ -31,3 +32,13 @@ func clearChangeList(notaryRepo client.Repository) error {
 	}
 	return cl.Clear("")
 }
+
+func getOrGenerateRootKeyAndInitRepo(notaryRepo client.Repository) error {
+	rootKey, err := getOrGenerateNotaryKey(notaryRepo, data.CanonicalRootRole)
+	if err != nil {
+		return err
+	}
+	// Initialize the notary repository with a remotely managed snapshot
```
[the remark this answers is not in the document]
riyazdf: added to the signature above

cli/command/trust/key_generate.go
```diff
@@ -0,0 +1,131 @@
...
+		fmt.Fprintf(streams.Out(), err.Error())
+		return errors.Wrapf(err, "failed to generate key for %s", keyName)
+	}
+
+	// Output the public key to a file in the CWD or specified dir
+	writtenPubFile, err := writePubKeyPEMToDir(pubPEM, keyName, workingDir)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(streams.Out(), "Successfully generated and loaded private key. Corresponding public key available: %s\n", writtenPubFile)
```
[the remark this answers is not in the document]
riyazdf: I don't want to expose the private key storage to users so I'd prefer not to (similar to comment below on the `TrustDirectory`)

riyazdf:
@thaJeztah: thank you for the thorough review!  I think I've addressed all comments.

As for the bug you found: I am not able to repro so maybe it is an odd tty issue in the container?

cli/command/trust/key_generate.go
```diff
@@ -0,0 +1,131 @@
...
+		Use:   "generate NAME [NAME...]",
+		Short: "Generate and load a signing key-pair",
+		Args:  cli.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			options.name = args[0]
+			return setupPassphraseAndGenerateKeys(dockerCli, options)
+		},
+	}
+	flags := cmd.Flags()
+	flags.StringVarP(&options.directory, "dir", "d", "", "Directory to generate key in, defaults to current directory")
```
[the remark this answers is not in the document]
riyazdf: I removed `-d` and left `--dir`, because I thought it'd good to be prescriptive with the filenames themselves (having `.pub` extension and the specified).

I'm holding off on `stdin` for now because the env var `DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE` can be used

cli/command/trust/key_generate.go
```diff
@@ -0,0 +1,131 @@
...
+	return cmd
+}
+
+// key names can use alphanumeric + _ + - characters
+var validKeyName = regexp.MustCompile(`^[a-zA-Z0-9\_]+[a-zA-Z0-9\_\-]*$`).MatchString
```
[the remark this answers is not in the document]
riyazdf: nice catch on the Mac/Windows and Linux!  I've changed to only lowercase

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,138 @@
...
+type signerAddOptions struct {
+	keys   opts.ListOpts
+	signer string
+	images []string
+}
+
+func newSignerAddCommand(dockerCli command.Cli) *cobra.Command {
+	var options signerAddOptions
+	cmd := &cobra.Command{
+		Use:   "add [OPTIONS] NAME IMAGE [IMAGE...] ",
```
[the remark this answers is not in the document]
riyazdf: We had the flag for `--name` originally but decided from user feedback (and @mbentley and @pdevine) that the positional argument made more sense

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,138 @@
...
+		Args:  cli.RequiresMinArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			options.signer = args[0]
+			options.images = args[1:]
+			return addSigner(dockerCli, options)
+		},
+	}
+	flags := cmd.Flags()
+	options.keys = opts.NewListOpts(nil)
+	flags.VarP(&options.keys, "key", "k", "Path to the signer's public key(s)")
```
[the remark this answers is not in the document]
riyazdf: removed shorthand `-k`

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	return notaryRepo.Publish()
+}
+
+func ingestPublicKeys(pubKeyPaths []string) ([]data.PublicKey, error) {
+	pubKeys := []data.PublicKey{}
+	for _, pubKeyPath := range pubKeyPaths {
+		// Read public key bytes from PEM file
+		pubKeyBytes, err := ioutil.ReadFile(pubKeyPath)
```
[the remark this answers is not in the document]
riyazdf: nice catch, limited!

cli/command/trust/key_generate.go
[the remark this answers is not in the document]
tych0: I see this env variable is from previous commits, so maybe it doesn't need fixing here, but it's worth noting that this is also very unsafe. Just as `/proc/self/cmdline` has the contents of the command line arguments, `/proc/self/environ` has the contents of the task's initial environment, and is world readable, exposing you to similar issues as `/proc/self/cmdline`.

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	signerPubKeys, err := ingestPublicKeys(options.keys.GetAll())
+	if err != nil {
+		return err
+	}
+	var errImages []string
+	for _, imageName := range options.images {
+		fmt.Fprintf(cli.Out(), "Adding signer \"%s\" to %s...\n", signerName, imageName)
+		if err := addSignerToImage(cli, signerName, imageName, signerPubKeys); err != nil {
+			fmt.Fprintln(cli.Err(), err.Error()+"\n")
```
dnephin: `Fprintln()` already adds a newline, but there's also a `\n` at the end of the message. Are you sure we want an extra blank line between these?

cli/command/trust/signer_remove.go
```diff
@@ -0,0 +1,138 @@
...
+	flags.BoolVarP(&options.forceYes, "yes", "y", false, "Do not prompt for confirmation before removing the most recent signer")
+	return cmd
+}
+
+func removeSigner(cli command.Cli, options signerRemoveOptions) error {
+	var errImages []string
+	for _, image := range options.images {
+		fmt.Fprintf(cli.Out(), "Removing signer \"%s\" from %s...\n", options.signer, image)
+		if err := removeSingleSigner(cli, image, options.signer, options.forceYes); err != nil {
+			fmt.Fprintln(cli.Err(), err.Error()+"\n")
```
dnephin: same here

cli/command/trust/signer_remove.go
```diff
@@ -0,0 +1,138 @@
...
+			errImages = append(errImages, image)
+		} else {
+			fmt.Fprintf(cli.Out(), "Successfully removed %s from %s\n\n", options.signer, image)
```
dnephin: same here, seems to be an extra blank line

cli/command/trust/signer_add.go
riyazdf: I added an extra blank line to separate the signer output if there are multiple ex:
```
$ docker trust signer add alice example/trust-demo example/trust-demo2 --key alice.crt
Adding signer "alice" to example/trust-demo...
Enter passphrase for repository key with ID 95b9e55: 
Successfully added signer: alice to example/trust-demo

Adding signer "alice" to example/trust-demo2...
Enter passphrase for repository key with ID ece554f: 
Successfully added signer: alice to example/trust-demo2
```

Let me know if you'd prefer for the blank line to be removed, but I think it helps improve readability

dnephin: Ok

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,141 @@
...
+		return fmt.Errorf("path to a public key must be provided using the `--key` flag")
+	}
+	signerPubKeys, err := ingestPublicKeys(options.keys.GetAll())
+	if err != nil {
+		return err
+	}
+	var errImages []string
+	for _, imageName := range options.images {
+		fmt.Fprintf(cli.Out(), "Adding signer \"%s\" to %s...\n", signerName, imageName)
+		if err := addSignerToImage(cli, signerName, imageName, signerPubKeys); err != nil {
```
thaJeztah: Not a blocker, but noticed that we're not consistent in requiring a `:tag` or not;

```bash
$ docker trust sign thajeztah/signingtest
No tag specified for thajeztah/signingtest
```

But

```bash
$ docker trust signer add bob thajeztah/signingtest --key=/foo/baz/bob.pub

Adding signer "bob" to thajeztah/signingtest...
Enter passphrase for repository key with ID 05cab95:
Successfully added signer: bob to thajeztah/signingtest
```

docs/reference/commandline/trust_signer_remove.md
```diff
@@ -0,0 +1,171 @@
...
+     another repository, consider it read-only there, as it will
+     periodically be overwritten by the definitive file. Pull
+     requests which include edits to this file in other repositories
+     will be rejected.
+-->
+
+# trust signer remove
+
+```markdown
+Usage:  docker trust signer remove [OPTIONS] NAME IMAGE [IMAGE...]
```
thaJeztah:
```
Usage:	docker trust signer remove [OPTIONS] NAME IMAGE [IMAGE...]

Remove a signer

Options:
      --help   Print usage
  -y, --yes    Do not prompt for confirmation before removing the most recent signer
```

docs/reference/commandline/trust_signer_remove.md
```diff
@@ -0,0 +1,171 @@
...
+`docker trust signer remove` is currently experimental.
+
+## Examples
+
+### Remove a signer from a repo
+
+To remove an existing signer, `alice`, from this repository: 
+
+```bash
+$ docker trust inspect example/trust-demo
```
thaJeztah: s/inspect/view/

docs/reference/commandline/trust_signer_remove.md
```diff
@@ -0,0 +1,171 @@
...
+Repository Key:	ecc457614c9fc399da523a5f4e24fe306a0a6ee1cc79a10e4555b3c6ab02f71e
+Root Key:	3cb2228f6561e58f46dbc4cda4fcaff9d5ef22e865a94636f82450d1d2234949
+```
+
+### Remove a signer from multiple repos
+
+To remove an existing signer, `alice`, from multiple repositories: 
+
+```bash
+$ docker trust inspect example/trust-demo
```
thaJeztah: s/inspect/view/

cli/command/trust/signer_remove.go
```diff
@@ -0,0 +1,138 @@
...
+		},
+	}
+	flags := cmd.Flags()
+	flags.BoolVarP(&options.forceYes, "yes", "y", false, "Do not prompt for confirmation before removing the most recent signer")
+	return cmd
+}
+
+func removeSigner(cli command.Cli, options signerRemoveOptions) error {
+	var errImages []string
+	for _, image := range options.images {
```
thaJeztah: Looks like we don't require a `:tag` for these images as well

docs/reference/commandline/trust_signer_add.md
```diff
@@ -0,0 +1,210 @@
...
+`docker trust signer add` is currently experimental.
+
+## Examples
+
+### Add a signer to a repo
+
+To add a new signer, `alice`, to this repository: 
+
+```bash
+$ docker trust inspect example/trust-demo
```
thaJeztah: s/inspect/view/

(some other places as well)

cli/command/trust/signer_remove.go
```diff
@@ -0,0 +1,138 @@
...
+		Short: "Remove a signer",
+		Args:  cli.RequiresMinArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			options.signer = args[0]
+			options.images = args[1:]
+			return removeSigner(dockerCli, options)
+		},
+	}
+	flags := cmd.Flags()
+	flags.BoolVarP(&options.forceYes, "yes", "y", false, "Do not prompt for confirmation before removing the most recent signer")
```
thaJeztah: Noticed that for `docker xx prune`, we're using `--force` for this;

```
Usage:	docker image prune [OPTIONS]

Remove unused images

Options:
  -a, --all             Remove all unused images, not just dangling ones
      --filter filter   Provide filter values (e.g. 'until=<timestamp>')
  -f, --force           Do not prompt for confirmation
      --help            Print usage
```

cli/command/trust/key_load.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return ioutil.ReadAll(from)
+}
+
+func loadPrivKeyBytesToStore(privKeyBytes []byte, privKeyImporters []trustmanager.Importer, keyPath, keyName string, passRet notary.PassRetriever) error {
+	if _, _, err := tufutils.ExtractPrivateKeyAttributes(privKeyBytes); err != nil {
+		return fmt.Errorf("provided file %s is not a supported private key - to add a signer's public key use docker trust signer add", keyPath)
+	}
+	// Make a reader, rewind the file pointer
+	return trustmanager.ImportKeys(bytes.NewReader(privKeyBytes), privKeyImporters, keyName, "", passRet)
```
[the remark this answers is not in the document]
riyazdf: I think this is fine - the name is simply extra metadata attached to the unique ID so it's ok to overwrite it if you have access to the same private key

cli/command/trust/signer_add.go
```diff
@@ -0,0 +1,141 @@
...
+		Args:  cli.RequiresMinArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			options.signer = args[0]
+			options.images = args[1:]
+			return addSigner(dockerCli, options)
+		},
+	}
+	flags := cmd.Flags()
+	options.keys = opts.NewListOpts(nil)
+	flags.Var(&options.keys, "key", "Path to the signer's public key file(s)")
```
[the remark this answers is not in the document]
riyazdf: I think we have a similar `(s)` for options like `volumes-from` so we had it here for consistency - is this okay or are we trying to remove this?

cli/command/trust/signer_add.go
riyazdf: for signing a tag is required to specify an image.  For `signer add` and `signer remove` we silently parse out any tags because they operate over the full image repo - we did this for convenience, but we could error out explicitly if you prefer?

cli/command/trust/key_load.go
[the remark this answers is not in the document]
thaJeztah: For a follow-up, we could add a message printing "updating existing key"

cli/command/trust/signer_add.go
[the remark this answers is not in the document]
thaJeztah: `volumes-from` is slightly different, because it copies all defined volumes from another container, _and_ can be specified multiple times (so a single `--volumes-from` can lead to one or more volumes being added)

cli/command/trust/signer_add.go
thaJeztah: Ah, right, so

```
docker trust signer add bob thajeztah/signingtest:one thajeztah/signingtest:two
```

is basically doing

```
docker trust signer add bob thajeztah/signingtest:*
```

So `docker trust signer add` adds a signer to a _repository_, not an _image_ ? I know "repository" is a bit overloaded, but perhaps we need to use that here (again: could use some input on that one)

cli/command/trust/signer_remove.go
riyazdf: I'll update to force :+1:

cli/command/trust/signer_add.go
riyazdf: yes it is, great catch! Notary operates a single trusted collection per repository. I'll update usage and argnames accordingly

cli/command/trust/key_load.go
```diff
@@ -82,9 +87,29 @@ func getPrivKeyBytesFromPath(keyPath string) ([]byte, error) {
...
 		return fmt.Errorf("provided file %s is not a supported private key - to add a signer's public key use docker trust signer add", keyPath)
 	}
+	if privKeyBytes, err = decodePrivKeyIfNecessary(privKeyBytes, passRet); err != nil {
+		return errors.Wrapf(err, "cannot load key from provided file %s", keyPath)
+	}
 	// Make a reader, rewind the file pointer
 	return trustmanager.ImportKeys(bytes.NewReader(privKeyBytes), privKeyImporters, keyName, "", passRet)
 }
+
+func decodePrivKeyIfNecessary(privPemBytes []byte, passRet notary.PassRetriever) ([]byte, error) {
```
thaJeztah: Would be good to have a unit-test for this one, but I'm okay adding that in a follow-up

riyazdf:
thanks @thaJeztah: I've cut down to ~half the commits!

thaJeztah:
Thanks! Discussed with @riyazdf and further squashing proved to be difficult, so I'm good with the current number of commits 👍

PR #567, opened 2017-09-27
test rmi -f returns success

andrewhsu:
**- What I did**

Added `e2e` test to verify `docker rmi -f` returns unix code success always because `-f`.

**- How I did it**

At my desk.

**- How to verify it**

```
$ make -f docker.Makefile test-e2e
...
=== RUN   TestForceRemoveImage
--- PASS: TestForceRemoveImage (0.16s)
PASS
```

N/A

🐛

### DIFF
--- a/e2e/container/run_test.go
+++ b/e2e/container/run_test.go
@@ -24,6 +24,11 @@ func TestRunAttachedFromRemoteImageAndRemove(t *testing.T) {
 	golden.Assert(t, result.Stderr(), "run-attached-from-remote-and-remove.golden")
 }
 
+func TestForceRemoveImage(t *testing.T) {
+	result := icmd.RunCmd(shell(t, "docker rmi -f badb33f99donteatme"))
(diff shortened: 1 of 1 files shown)


e2e/container/run_test.go
```diff
@@ -24,6 +24,11 @@ func TestRunAttachedFromRemoteImageAndRemove(t *testing.T) {
 	golden.Assert(t, result.Stderr(), "run-attached-from-remote-and-remove.golden")
 }
 
+func TestForceRemoveImage(t *testing.T) {
```
dnephin: We already have test coverage for that here in the unit test suite: https://github.com/docker/cli/blob/master/cli/command/image/remove_test.go#L71-L78

Following https://github.com/docker/cli/blob/master/TESTING.md#guidelines I don't think we need an e2e test for the force flag.

I unit test for a successful `docker rmi` would be good, but it would go in the  `e2e/image` package.

PR #573, opened 2017-09-29
Add composefile v1 / v2 parsing and validation support

vdemeester:
This adds validation and parsing support to v1 and v2 in `docker/cli/compose` package. There is still a lot to decide on that one 

- [x] Should it be supported or not by the `cli` (i.e. should `docker stack deploy` support other version than `>= 3.0` or not)
- [ ] What to do about `cpu_share`, `cpu_quota` and other removed field (either ignore them *or* update the struct)
- [ ] support `depends_on`
- [ ] support `extends`
- [ ] edge cases ?

As of now, it allows to parse and validate v1, v2 and v3 composefile version in the same struct when using `docker/cli/compose` as a library.

/cc @shin- 

boom


cli/compose/loader/loader_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/docker/cli/cli/compose/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"path/filepath"
```
silvin-lubecki: You should move this import up with other standard imports.

cli/compose/loader/loader_v1_test.go
```diff
@@ -0,0 +1,37 @@
...
+	_, err := loadYAML(`
+version: "0.1"
+services:
+  foo:
+    image: busybox
+`)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "version")
+
+	_, err = loadYAML(`
```
silvin-lubecki: Why testing the same error again?

vdemeester: you are right, I'll remove :stuck_out_tongue:

dnephin:
> Should it be supported or not by the cli (i.e. should docker stack deploy support other version than >= 3.0 or not)

It should not be supported. The config does not match what is supported by services, and it doesn't make sense to once again create client-side logic for services.

cli/compose/loader/testdata/v1-full.yml
```diff
@@ -0,0 +1,10 @@
+net:
+  image: busybox
+volume:
+  image: busybox
+  volumes:
```
dnephin: if this is a "full" test it should probably include all the fields supported by v1 and one service with build, one with image

cli/compose/loader/testdata/v2-full.yml
```diff
@@ -0,0 +1,23 @@
...
+  data:
+    driver: local
+
+networks:
+  front: {}
+
+services:
+  web:
+    build: .
+    networks:
```
dnephin: same here, should include all the fields

cli/compose/loader/testdata/v1-full.yml
vdemeester: @dnephin yeah I just copied the file from `docker/compose so far :stuck_out_tongue:

vdemeester:
I'm gonna close this one as it is stale.
cc @chris-crone @silvin-lubecki @simonferquel fyi 👼

mgoltzsche:
@vdemeester Why didn't you finish the implementation? The idea to provide a single code base that supports all compose versions was great. Now we still have to deal with various different parser implementations that could in theory differ from each other concerning the same compose version (besides the fact that the runtime implementations differ anyway). Also it creates more effort in other projects to support all compose versions.

olljanat:
@vdemeester ping. Look @mgoltzsche question above.

PR #575, opened 2017-09-29
trust sign: add --local flag

eiais:
The --local flag will force the signing of a local image.

cc @riyazdf @ashfall 

Signed-off-by: Kyle Spiers <kyle@spiers.me>

### DIFF
--- a/cli/command/trust/sign.go
+++ b/cli/command/trust/sign.go
@@ -18,19 +18,29 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type signOptions struct {
+	local     bool
+	imageName string
+}
+
 func newSignCommand(dockerCli command.Cli) *cobra.Command {
+	options := signOptions{}
 	cmd := &cobra.Command{
 		Use:   "sign IMAGE:TAG",
 		Short: "Sign an image",
 		Args:  cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runSignImage(dockerCli, args[0])
+			options.imageName = args[0]
+			return runSignImage(dockerCli, options)
 		},
 	}
+	flags := cmd.Flags()
+	flags.BoolVar(&options.local, "local", false, "Sign a locally tagged image")
 	return cmd
 }
 
-func runSignImage(cli command.Cli, imageName string) error {
+func runSignImage(cli command.Cli, options signOptions) error {
+	imageName := options.imageName
 	ctx := context.Background()
 	imgRefAndAuth, err := trust.GetImageReferencesAndAuth(ctx, image.AuthResolver(cli), imageName)
 	if err != nil {
@@ -71,13 +81,15 @@ func runSignImage(cli command.Cli, imageName string) error {
 	}
 	requestPrivilege := command.RegistryAuthenticationPrivilegedFunc(cli, imgRefAndAuth.RepoInfo().Index, "push")
 	target, err := createTarget(notaryRepo, imgRefAndAuth.Tag())
-	if err != nil {
+	if err != nil || options.local {
 		switch err := err.(type) {
-		case client.ErrNoSuchTarget, client.ErrRepositoryNotExist:
+		// If the error is nil then the local flag is set
+		case client.ErrNoSuchTarget, client.ErrRepositoryNotExist, nil:
 			// Fail fast if the image doesn't exist locally
 			if err := checkLocalImageExistence(ctx, cli, imageName); err != nil {
 				return err
 			}
+			fmt.Fprintf(cli.Err(), "Signing and pushing trust data for local image %s, may overwrite remote trust data\n", imageName)
 			return image.TrustedPush(ctx, cli, imgRefAndAuth.RepoInfo(), imgRefAndAuth.Reference(), *imgRefAndAuth.AuthConfig(), requestPrivilege)
 		default:
 			return err
--- a/cli/command/trust/sign_test.go
+++ b/cli/command/trust/sign_test.go
@@ -295,3 +295,13 @@ func TestSignCommandChangeListIsCleanedOnError(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, len(cl.List()), 0)
 }
+
+func TestSignCommandLocalFlag(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{})
+	cli.SetNotaryClient(getEmptyTargetsNotaryRepository)
+	cmd := newSignCommand(cli)
+	cmd.SetArgs([]string{"--local", "reg-name.io/image:red"})
+	cmd.SetOutput(ioutil.Discard)
+	testutil.ErrorContains(t, cmd.Execute(), "error during connect: Get /images/reg-name.io/image:red/json: unsupported protocol scheme")
+
+}
--- a/docs/reference/commandline/trust_sign.md
+++ b/docs/reference/commandline/trust_sign.md
@@ -16,9 +16,13 @@ keywords: "sign, notary, trust"
 # trust sign
 
 ```markdown
-Usage:  docker trust sign IMAGE:TAG
+Usage:  docker trust sign [OPTIONS] IMAGE:TAG
 
 Sign an image
+
+Options:
+      --help    print usage
+      --local   force the signing of a local image
 
 ```
 


docs/reference/commandline/trust_sign.md
```diff
@@ -16,10 +16,14 @@ keywords: "sign, notary, trust"
 # trust sign
 
 ```markdown
-Usage:  docker trust sign IMAGE:TAG
+Usage:  docker trust sign [OPTIONS] IMAGE:TAG
```
mstanleyjones: I think @riyazdf pulled `[OPTIONS]` out of a bunch of the other commands. Is it OK here?

docs/reference/commandline/trust_sign.md
```diff
@@ -16,10 +16,14 @@ keywords: "sign, notary, trust"
...
 
 Sign an image
 
+Options:
+      --help    print usage
+  -l, --local   force the signing of a local image
```
mstanleyjones: nit: alignment

docs/reference/commandline/trust_sign.md
eiais: Yes, we pulled it out because they didn't have options yet. This one now does.

cli/command/trust/sign_test.go
```diff
@@ -295,3 +295,13 @@ func TestSignCommandChangeListIsCleanedOnError(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, len(cl.List()), 0)
 }
+
+func TestLocalFlag(t *testing.T) {
```
vdemeester: @eiais Not sure I clearly understand that test :thinking: 
Shouldn't we also test *the happy path* (i.e. "Signing and pushing trust data for local image" […]).

thaJeztah:
I also left a similar proposal in the original PR: https://github.com/docker/cli/pull/472#pullrequestreview-63738876

> Wondering if it's always desirable to push the image, or if there should be an option to either make pushing optional, or to disable pushing.

An alternative could be to have a `--push` flag, but @riyazdf is more familiar with the common workflow / expectations than I am 😅

riyazdf:
@thaJeztah: yup - this is a followup PR that addresses your feedback, though in a slightly different fashion.

In the context of docker, it doesn't really make sense to have a signature without a pushed image - so we've deliberately made the workflows such that if you sign an image there should always be an associated image in the registry unless you manually delete images in the registry or the signatures with notary.

Taking this into account on `docker trust sign`: if the image doesn't exist in a signed repository then it will push and sign. However, if the image does exist with signatures `docker trust sign` will multi-sign over the same digest/tag without pushing anything to the registry.

This PR adds a `--local` flag to always attempt pushing and signing a locally tagged image, even if signatures already exist, potentially clobbering the digest for the tag if it was previously signed.

cli/command/trust/sign.go
```diff
@@ -18,19 +18,26 @@ import (
...
+
 func newSignCommand(dockerCli command.Cli) *cobra.Command {
+	options := signOptions{}
 	cmd := &cobra.Command{
 		Use:   "sign IMAGE:TAG",
 		Short: "Sign an image",
 		Args:  cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runSignImage(dockerCli, args[0])
+			return signImage(dockerCli, args[0], options)
```
dnephin: Small thing, but would you mind making the `args[0]` part of options struct by adding an `imageName` field? We do this for most commands.

cli/command/trust/sign_test.go
dnephin: The test name should reflect the function that is being tested, so the name should be something like `TestSignCommandLocalFlag`

cli/command/trust/sign.go
```diff
@@ -18,19 +18,26 @@ import (
...
 		},
 	}
+	flags := cmd.Flags()
+	flags.BoolVarP(&options.local, "local", "l", false, "Sign a locally tagged image")
 	return cmd
 }
 
-func runSignImage(cli command.Cli, imageName string) error {
+func signImage(cli command.Cli, imageName string, options signOptions) error {
```
dnephin: I think this should be `runSignImage()` (or `runSign` would be fine too). This is another convention we use in the cli. The `run` prefix refers to the fact this is the function called for `RunE` of the Command.

cli/command/trust/sign.go
```diff
@@ -18,19 +18,26 @@ import (
...
 		Use:   "sign IMAGE:TAG",
 		Short: "Sign an image",
 		Args:  cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runSignImage(dockerCli, args[0])
+			return runSignImage(dockerCli, args[0], options)
 		},
 	}
+	flags := cmd.Flags()
+	flags.BoolVarP(&options.local, "local", "l", false, "Sign a locally tagged image")
```
thaJeztah: Can you remove the shorthand `-l` for now? Trying to be a bit conservative adding shorthands, unless a) frequently used, and b) we know it's not going to conflict with a possibly more-frequently-used option.

It's easier to add them later then removing/deprecating

eiais: Removed

docs/reference/commandline/trust_sign.md
dnephin: `-l` should be removed here as well

cli/command/trust/sign_test.go
```diff
@@ -295,3 +295,13 @@ func TestSignCommandChangeListIsCleanedOnError(t *testing.T) {
...
 	assert.Equal(t, len(cl.List()), 0)
 }
+
+func TestSignCommandLocalFlag(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{})
+	cli.SetNotaryClient(getEmptyTargetsNotaryRepository)
+	cmd := newSignCommand(cli)
+	cmd.SetArgs([]string{"--local", "reg-name.io/image:red"})
+	cmd.SetOutput(ioutil.Discard)
+	testutil.ErrorContains(t, cmd.Execute(), "error during connect: Get /images/reg-name.io/image:red/json: unsupported protocol scheme")
```
dnephin: This seems like a strange expectation for a test case. Can't this use the notary fakes we have in `client_test.go` to make it a success case?

cli/command/trust/sign.go
```diff
@@ -71,13 +78,15 @@ func runSignImage(cli command.Cli, imageName string) error {
...
+	if err != nil || options.local {
 		switch err := err.(type) {
-		case client.ErrNoSuchTarget, client.ErrRepositoryNotExist:
+		// If the error is nil then the local flag is set
+		case client.ErrNoSuchTarget, client.ErrRepositoryNotExist, nil:
 			// Fail fast if the image doesn't exist locally
 			if err := checkLocalImageExistence(ctx, cli, imageName); err != nil {
 				return err
 			}
+			fmt.Fprintf(cli.Out(), "Signing and pushing trust data for local image %s, may overwrite remote trust data\n", imageName)
```
dnephin: This should use `cli.Err()` instead of `cli.Out()`. This message is informational (like logging), it's not part of  "normal program output" (what the user asked for by running the command).

cli/command/trust/sign.go
```diff
@@ -18,19 +18,26 @@ import (
...
+
 func newSignCommand(dockerCli command.Cli) *cobra.Command {
+	options := signOptions{}
 	cmd := &cobra.Command{
 		Use:   "sign IMAGE:TAG",
 		Short: "Sign an image",
 		Args:  cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runSignImage(dockerCli, args[0])
+			return runSignImage(dockerCli, args[0], options)
```
dnephin: Minor: now that we have an options struct, the `imageName` can be passed as part of that struct, instead of a separate parameter. Not super important at this point since there is only a single arg, but it would be more consistent.

cli/command/trust/sign.go
```diff
@@ -18,19 +18,26 @@ import (
...
 		},
 	}
+	flags := cmd.Flags()
+	flags.BoolVarP(&options.local, "local", "", false, "Sign a locally tagged image")
```
dnephin: I guess this should also use `flags.BoolVar()` now that it doesn't have a shorthand flag.

thaJeztah:
Linting failure, @eiais 

```
cli/command/trust/sign.go:1::warning: file is not gofmted with -s (gofmt)
cli/command/trust/sign.go:1::warning: file is not goimported (goimports)

PR #603, opened 2017-10-05
Fix markdown link in service-create reference docs

thaJeztah:
Markdown nested in a HTML table doesn't work, so changing the link to a plain HTML link.

fixes https://github.com/docker/cli/issues/602

ping @johndmulhausen @mstanleyjones

### DIFF
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -294,8 +294,9 @@ volumes in a service:
     <td>
       <p>The type of mount, can be either <tt>volume</tt>, <tt>bind</tt>, or <tt>tmpfs</tt>. Defaults to <tt>volume</tt> if no type is specified.
       <ul>
-        <li><tt>volume</tt>: mounts a [managed volume](volume_create.md) into the container.</li>
-        <li><tt>bind</tt>: bind-mounts a directory or file from the host into the container.</li>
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/service_create.md
```diff
@@ -294,8 +294,9 @@ volumes in a service:
     <td>
       <p>The type of mount, can be either <tt>volume</tt>, <tt>bind</tt>, or <tt>tmpfs</tt>. Defaults to <tt>volume</tt> if no type is specified.
       <ul>
-        <li><tt>volume</tt>: mounts a [managed volume](volume_create.md) into the container.</li>
-        <li><tt>bind</tt>: bind-mounts a directory or file from the host into the container.</li>
+        <li><tt>volume</tt>: mounts a <a href="https://docs.docker.com/engine/reference/commandline/volume_create/">managed volume</a>
```
mstanleyjones: This is an OK fix. Alternatively, you could add `markdown="span"` as an attribute to the `<li>` and leave the Markdown in there. Your choice. I think you could also avoid the `<tt>` tags too, in favor of backticks.

thaJeztah: Oh, didn't know about that one! I think I'll keep it like this for now (mixing HTML and Markdown in a table may be confusing), but we could change it in a single go if we think it's better

PR #606, opened 2017-10-06
Revert "Refactor image commands to make use of the new trust struct for trusted pull"

riyazdf:
This reverts commit 4203b4943154ad09917702dbae2f2b414f70dc5d.

I've reverted because the immutable nature of `ImageRefAndAuth` doesn't fit as well for image pull as it did for the trust commands.

In particular: updating an `ImageRefAndAuth` for adding `latest` as a default tag, and the `trustedRef` in `trustedPull` is not used.

cc @dnephin @andrewhsu @RobbKistler 

I will followup with regression tests in the existing `DockerTrustSuite`

### DIFF
--- a/cli/command/image/pull.go
+++ b/cli/command/image/pull.go
@@ -6,8 +6,8 @@ import (
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
-	"github.com/docker/cli/cli/trust"
 	"github.com/docker/distribution/reference"
+	"github.com/docker/docker/registry"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"golang.org/x/net/context"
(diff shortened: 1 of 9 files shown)


dnephin:
Can't we just add a method to return a new object with a new tag?

We need a test in this repo, otherwise it will break again. I would rather fix the small bug and add a test than revert the entire thing.

riyazdf:
@dnephin: we totally can, see https://github.com/docker/cli/pull/607.  I do want to keep the `ImgRefAndAuth` immutable so the helper returns a new one.

For both PRs: I would very much like e2e tests in this repo.  Are you ok with porting over nearly all of `trust-server.go` into this repo, including notary fixtures for mTLS? (integration-cli copies all of the notary GH repo, including these fixtures)

dnephin:
We should add it using the docker image instead, to https://github.com/docker/cli/blob/master/e2e/compose-env.yaml

dnephin:
Thanks, #607 looks great, I've marked this PR as closed by that one so it will be closed when it merges

PR #607, opened 2017-10-06
Mutate image references where needed for trusted pulls

riyazdf:
Closes #606 

This is the alternative to #606 - in both cases I would very much like to have proper e2e tests - ideally we can reuse almost all of `trust-server.go` in `moby/moby` though I haven't been able to start this work

Edit: was able to add e2e regression tests, thank you @dnephin for the help!

Test output before fix:
```
--- FAIL: TestPullWithContentTrust (2.65s)
	command.go:61: at pull_test.go:41 -
		Command:  docker pull registry:5000/trust:latest
		ExitCode: 1
		Error:    exit status 1
		Stdout:   Pull (1 of 1): registry:5000/trust:latest@sha256:641b95ddb2ea9dc2af1a0113b6b348ebc20872ba615204fbe12148e98fd6f23d
		latest: Pulling from trust
		Digest: sha256:030fcb92e1487b18c974784dcc110a93147c9fc402188370fbfd17efabffc6af
		Status: Downloaded newer image for registry:5000/trust:latest
		Tagging registry:5000/trust@sha256:641b95ddb2ea9dc2af1a0113b6b348ebc20872ba615204fbe12148e98fd6f23d as registry:5000/trust:latest

		Stderr:   Error response from daemon: No such image: registry:5000/trust@sha256:641b95ddb2ea9dc2af1a0113b6b348ebc20872ba615204fbe12148e98fd6f23d


		Failures:
		ExitCode was 1 expected 0
		Expected no error
FAIL
```

Signed-off-by: Riyaz Faizullabhoy <riyaz.faizullabhoy@docker.com>


e2e/container/dct_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+const notaryURL = "https://localhost:4443"
+const registryPrefix = "registry:5000"
+
+func TestPushImage(t *testing.T) {
+	image := createTrustedRemoteImage(t, "trust", "latest")
+
+	icmd.RunCmd(trustedCmdNoPassphrases(icmd.Command("docker", "pull", image))).Assert(t, icmd.Success)
+	result := icmd.RunCmd(shell(t,
+		"DOCKER_CONTENT_TRUST=1 DOCKER_CONTENT_TRUST_SERVER=%s docker run --rm %s echo this is output", notaryURL, image))
```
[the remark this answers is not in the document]
riyazdf: yes sorry - this is a very much WIP test.  Currently the framework isn't working with notary but I intend for this to be a trusted pull test

riyazdf:
@dnephin: thanks for the review!  In theory we won't need `trust_server_test.go`, but right now I haven't gotten the connection to the notary server working properly and I'm not sure why it isn't working

riyazdf:
thanks @dnephin for helping rig this up - both the fix and regression test are ready for review!

e2e/image/pull_test.go
```diff
@@ -0,0 +1,75 @@
...
+	icmd.RunCommand("docker", "rmi", image).Assert(t, icmd.Success)
+	return image
+}
+
+func trustedCmdWithPassphrases(cmd icmd.Cmd, rootPwd, repositoryPwd string) icmd.Cmd {
+	env := append(os.Environ(), []string{
+		"DOCKER_CONTENT_TRUST=1",
+		fmt.Sprintf("DOCKER_CONTENT_TRUST_SERVER=%s", notaryURL),
+		fmt.Sprintf("DOCKER_CONTENT_TRUST_ROOT_PASSPHRASE=%s", rootPwd),
+		fmt.Sprintf("DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE=%s", repositoryPwd),
```
thaJeztah: nit: no need to use `fmt.Sprintf` for these; just concatenating the strings should do

e2e/image/pull_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}...)
+	cmd.Env = append(cmd.Env, env...)
+	return cmd
+}
+
+func trustedCmdNoPassphrases(cmd icmd.Cmd) icmd.Cmd {
+	env := append(os.Environ(), []string{
+		"DOCKER_CONTENT_TRUST=1",
+		fmt.Sprintf("DOCKER_CONTENT_TRUST_SERVER=%s", notaryURL),
```
thaJeztah: Same here (but just a nit)

scripts/test/e2e/load-busybox
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+
+src=busybox@sha256:3e8fa85ddfef1af9ca85a5cfb714148956984e02f00bec3f7f49d3925a91e0e7
+dest=registry:5000/busybox:1.27.2
+docker pull $src
```
thaJeztah: wondering; is this pulling from hub? if so, should we consider having a "frozen image" baked in the base image (like we have in moby/moby https://github.com/moby/moby/blob/2bc9bff86707737ff5df5f23b613d16281f9d223/Dockerfile.e2e#L16-L21)?

riyazdf: I was following the same `load-alpine` script we have that pulls by digest.  Up to @dnephin on the best path forward here, but I think that makes sense in a separate PR.

e2e/image/pull_test.go
riyazdf: done!

e2e/image/pull_test.go
riyazdf: done!

scripts/test/e2e/load-busybox
dnephin: Ya, we could do that. I agree we should do it in a separate PR.

One thing I'd like to avoid is that in moby/moby we have about 4 or 5 different ways of loading fixtures. Right now everything is loaded from `setup()` in `scripts/test/e2e/run`. We can move some image freezing to the Dockerfile, but it will couple the run script to the Dockerfile.

e2e/image/pull_test.go
```diff
@@ -0,0 +1,75 @@
...
+	icmd.RunCommand("docker", "rmi", image).Assert(t, icmd.Success)
+
+	// push an unsigned image on the same reference name, but with different content (busybox)
+	icmd.RunCommand("docker", "pull", busyboxImage).Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "tag", busyboxImage, image).Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "push", image).Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "rmi", image).Assert(t, icmd.Success)
+
+	// now pull with content trust
+	result = icmd.RunCmd(trustedCmdNoPassphrases(icmd.Command("docker", "pull", image)))
```
dnephin: This `docker run` does not seem relevant to the test case. Can we remove it?

e2e/image/pull_test.go
```diff
@@ -0,0 +1,75 @@
...
+	result := icmd.RunCmd(trustedCmdNoPassphrases(icmd.Command("docker", "pull", image)))
+	result.Assert(t, icmd.Success)
+	firstPullOutput := result.String()
+	icmd.RunCommand("docker", "rmi", image).Assert(t, icmd.Success)
+
+	// push an unsigned image on the same reference name, but with different content (busybox)
+	icmd.RunCommand("docker", "pull", busyboxImage).Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "tag", busyboxImage, image).Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "push", image).Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "rmi", image).Assert(t, icmd.Success)
```
dnephin: This tag/push/rmi does not seem relevant to the test case. Can we remove it?

Pushing a trusted image can be a separate case.

e2e/image/pull_test.go
```diff
@@ -0,0 +1,75 @@
...
+const alpineImage = "registry:5000/alpine:3.6"
+const busyboxImage = "registry:5000/busybox:1.27.2"
+
+func TestPullWithContentTrust(t *testing.T) {
+	image := createTrustedRemoteImage(t, "trust", "latest")
+
+	// test that pulling without the tag defaults to latest
+	imageWithoutTag := strings.TrimSuffix(image, ":latest")
+	icmd.RunCmd(trustedCmdNoPassphrases(icmd.Command("docker", "pull", imageWithoutTag))).Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "rmi", image).Assert(t, icmd.Success)
```
dnephin: We can test "pulling without the tag defaults to latest" in a unit test, doesn't need to be a case in the e2e suite.

e2e/image/pull_test.go
riyazdf: we need the tag/push/rmi here to test that when you push a new image with the same name that is not signed, DCT will still use the signed version

scripts/test/e2e/load-busybox
thaJeztah: Yes - I agree ok looking at that in a separate PR; just want to prevent our test from relying on Docker Hub too much :+1:

e2e/image/pull_test.go
riyazdf: if you'd like we can remove it, though I think there's still value in testing this in the trust suite to ensure that there isn't any further manipulation on the default tag.

There's an existing unit tests for this in the image pull tests, output here: https://github.com/docker/cli/blob/master/cli/command/image/testdata/pull-command-success.simple-no-tag.golden

e2e/image/pull_test.go
riyazdf: I've pushed a new commit to factor out this setup logic into a helper, as per our offline conversation.  Please let me know if there are outstanding issues to fix

PR #609, opened 2017-10-07
 Add bash completion for `docker trust` commands

albers:
This adds bash completion for #472.

This should go into 17.10.0 because the feature is present in that release.

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -4505,7 +4505,7 @@ _docker_system() {
 		info
 		prune
 	"
-	__docker_subcommands "$subcommands $aliases" && return
+	__docker_subcommands "$subcommands" && return
 
 	case "$cur" in
 		-*)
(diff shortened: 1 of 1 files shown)


thaJeztah:
IIRC, we marked these commands "experimental" (which is a bit tricky, because it's a client-side only feature, so we don't currently hide them). 

I don't see a real harm in including them in the completion scripts (as 17.10 is an edge release, so no long-term consequences)

@riyazdf PTAL

albers:
@thaJeztah The commands are only considered in bash completion if the CLI runs against an experimental server.

riyazdf:
awesome, thanks @albers!  I am not familiar with how bash completion works so it'd be best if someone else can provide review/approval here.

thaJeztah:
ah, yes, I forgot for a minute we added the check for experimental here as well 👍

PR #618, opened 2017-10-11
Add bash completion for `secret create --driver`

albers:
Bash completion for #366


albers:
This should go into 17.10.0 as the feature is present in that release:

```bash
$ docker secret create --help

Usage:  docker secret create [OPTIONS] SECRET [file|-]

Create a secret from a file or STDIN as content

Options:
  -d, --driver string   Secret driver
      --help            Print usage
  -l, --label list      Secret labels
$ docker --version
Docker version 17.10.0-ce-rc1, build d866876
```

PR #621, opened 2017-10-13
Refactor stringutils and fix docker search output form

charrywanganthony:
when the description has CJK character

Signed-off-by: Chao Wang <wangchao.fnst@cn.fujitsu.com>

**- What I did**
Accoring to @thaJeztah 's suggestion, I move this work from moby to docker/cli [moby/pull35034](https://github.com/moby/moby/pull/35034)
**- How I did it**
East Asian character always takes two position of a English charactor. [Technical Reports about width of East Asian text](http://unicode.org/reports/tr11/) I use the unicode library  https://godoc.org/golang.org/x/text/width
**- How to verify it**
search result


cli/command/formatter/displayutils.go
```diff
@@ -0,0 +1,82 @@
...
+	if maxwidth == 1 {
+		return TruncateByWidth(s, 1)
+	}
+
+	displayWidth := 0
+	rs := []rune(s)
+	for _, r := range rs {
+		cw := charWidth(r)
+		displayWidth += cw
+	}
```
dnephin: It seems like we know at this point if the `displayWidth` is greater than or equal to the maxWidth. Could we actually return here instead of calling `TruncateByWidth()` ?

```go
for index, r := range rs {
    width := charWidth(r)
    switch {
    case displayWidth + width == maxWidth && index + 1 == len(rs):
        return s
    case displayWidth + width >= maxWidth:
        return string(rs[:index]) + "…"
    }
    displayWidth += width
}
return s
```

If this works, then I think we can simplify the `maxwidth == 1` case as well and we could remove `TruncateByWidth()` entirely.

cli/command/formatter/displayutils_test.go
```diff
@@ -0,0 +1,88 @@
+package formatter
+
+import "testing"
+
+func TestEllipsis(t *testing.T) {
+	str := "t🐳ststring"
+	newstr := Ellipsis(str, 0)
+	if newstr != "" {
+		t.Fatalf("Expected null, got %s", newstr)
+	}
```
dnephin: These tests would be great as a test table.

```go
var testcases = []struct{
    source string
    width int
    expected string
} {
    {source: "t🐳ststring", width: 0},
    ...
}

for _, testcase := range testcases {
    assert.Equal(t, testcase.expected, Ellipsis(testcase.source, testcase.width))
}
```

Also please use `assert.Equal()` instead of `if x != y { t.Fatalf(...) }`

dnephin:
For the lint failure (should range over string, not []rune(string) (S1029) (gosimple)) I think this is one of the very rare cases where `[]rune(string)` is actually necessary, since we (may) need to index into the string. You may need to add `// nolint: gosimple` above the line

cli/command/formatter/displayutils.go
```diff
@@ -0,0 +1,82 @@
...
+	switch width.LookupRune(r).Kind() {
+	case width.EastAsianWide, width.EastAsianFullwidth:
+		return 2
+	default:
+		return 1
+	}
+}
+
+// TruncateByWidth truncates a string to fit within maxwidth.
+// For maxwidth of 1, first char of string will return even if its width > 1.
```
thaJeztah: Perhaps we need some more information here what "width" is, but I don't have a good suggestion at this moment.

charrywanganthony: Modified. Sorry for the delay.

cli/command/formatter/displayutils_test.go
charrywanganthony: Modified. Sorry for the delay.

cli/command/formatter/history_test.go
```diff
@@ -191,7 +190,7 @@ imageID3            24 hours ago        /bin/bash ls
 imageID4            24 hours ago        /bin/bash grep                                                                                                                 183MB               Hi
 `
 	expectedTrunc := `IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
-imageID1            24 hours ago        /bin/bash ls && npm i && npm run test && k...   183MB               Hi
+imageID1            24 hours ago        /bin/bash ls && npm i && npm run test && kar…   183MB               Hi
```
thaJeztah: :heart: `…` nice!

cli/command/formatter/displayutils_test.go
charrywanganthony: Although the pr is merged, if u are free @dnephin  , I have a low level quetion why should we use `assert.Equal()` instead of `if x != y { t.Fatalf(...) }`? What's the disadvantages of `t.Fatalf`? (☉㊀☉)

dnephin: @charrywanganthony 

* `assert.Equal()` can't lie. It will always print the real value for the two values being compared. If you write `if x != y { t.FatalF(msg) }` the `msg` could be incomplete, or misleading, or even completely wrong, which makes debugging the failure more difficult. You can always provide extra `msgAndArgs` to `assert.Equal()` as the 4+ argument.
* `assert.Equal()` is less verbose, so it keeps test smaller and easier to read
* `assert.Equal()` reads in the affirmative, where as conditions for `if x != y ...` are always the inverse. I **want** `x == y`, but I have to write `x != y`.

These points may not be relevant in all cases, but it's easier to be consistent and just use it everywhere.

PR #628, opened 2017-10-19
Add support for building on OpenBSD

voutilad:
**- What I did**

I've made a first attempt at porting docker/cli to OpenBSD. See also a [PR](https://github.com/docker/machine/pull/4287) I submitted for docker/machine.

With these changes, and the similar `docker-machine` changes, a developer on OpenBSD can user docker from the command line successfully like a user on Linux or other platforms. (The assumption being they've used `docker-machine` to configure their target docker service.)

**- How I did it**
- Extended the `Makefile` to add a new target: `binary-openbsd`
- Updated `scripts/build/.variables` to handle differences in OpenBSD's `date` output
- Added a new build script that turns on `CGO`
- Tweaked some vendor packages to add OpenBSD support or disable certain unsupported features (e.g. credentials store)

**Note:** I'm not positive about the change I made in `cli/command/formatter/history.go`. The tests were failing due to the current logic on OpenBSD printing the Time adjusted for the local timezone. I needed to add `.UTC()` to force the time to stay UTC. 

**- How to verify it**

For OpenBSD, it should work on both `6.2` and `-current` using the following packages:
- `go` (go 1.9.1)
- `gmake` (GNU Make)

Running `gmake binary-openbsd` will build the binary.

Note that running `gmake test` will potentially appear to hang at the `github.com/docker/cli/cli/command/idresolver` test the username prompt might not visibly render (it doesn't in XTerm running tmux at least), so type your username and hit enter. You should then see the password prompt and can enter that as well. Tests will continue at that point.

Currently I haven't tested cross-compiling the OpenBSD binary as it will probably require a bit more work on the tweaks to the supporting vendor code (i.e. rewriting the inline C `ioctl()` calls to using Go's `syscall`).

I've tested building using `Makefile.docker` on macOS and it appears to build and behave as normal with these changes. `make test` also passes.

Adds support for building binary on OpenBSD.

Maple:
My dog

### DIFF
--- a/cli/command/formatter/history.go
+++ b/cli/command/formatter/history.go
@@ -79,7 +79,7 @@ func (c *historyContext) ID() string {
 }
 
 func (c *historyContext) CreatedAt() string {
-	return time.Unix(c.h.Created, 0).Format(time.RFC3339)
+	return time.Unix(c.h.Created, 0).UTC().Format(time.RFC3339)
 }
 
 func (c *historyContext) CreatedSince() string {
--- a/cli/command/image/build_test.go
+++ b/cli/command/image/build_test.go
@@ -24,6 +24,8 @@ import (
 
 func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
 	skip.IfCondition(t, runtime.GOOS == "windows", "uid and gid not relevant on windows")
+	skip.IfCondition(t, runtime.GOOS == "openbsd", "gotestyourself/fs not working right on openbsd")
+
 	dest := fs.NewDir(t, "test-build-context-dest")
 	defer dest.Remove()
 
@@ -58,6 +60,7 @@ func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
 	}
 }
 func TestRunBuildDockerfileFromStdinWithCompress(t *testing.T) {
+	skip.IfCondition(t, runtime.GOOS == "openbsd", "not working right on openbsd")
 	dest, err := ioutil.TempDir("", "test-build-compress-dest")
 	require.NoError(t, err)
 	defer os.RemoveAll(dest)
--- a/cli/command/image/build_test.go
+++ b/cli/command/image/build_test.go
@@ -24,7 +24,7 @@ import (
 
 func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
 	skip.IfCondition(t, runtime.GOOS == "windows", "uid and gid not relevant on windows")
-	skip.IfCondition(t, runtime.GOOS == "openbsd", "gotestyourself/fs not working right on openbsd")
+	skip.IfCondition(t, runtime.GOOS != "windows" && os.Getuid() != 0, "test requires elevated privileges")
 
 	dest := fs.NewDir(t, "test-build-context-dest")
 	defer dest.Remove()
@@ -60,7 +60,7 @@ func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
 	}
 }
 func TestRunBuildDockerfileFromStdinWithCompress(t *testing.T) {
-	skip.IfCondition(t, runtime.GOOS == "openbsd", "not working right on openbsd")
+	skip.IfCondition(t, runtime.GOOS != "windows" && os.Getuid() != 0, "test requires elevated privileges")
 	dest, err := ioutil.TempDir("", "test-build-compress-dest")
 	require.NoError(t, err)
 	defer os.RemoveAll(dest)
--- a/vendor/github.com/containerd/continuity/sysx/xattr_unsupported.go
+++ b/vendor/github.com/containerd/continuity/sysx/xattr_unsupported.go
@@ -1,4 +1,4 @@
-// +build freebsd solaris
+// +build openbsd freebsd solaris
 
 package sysx
 
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,11 @@ binary: ## build executable for Linux
 	@echo "WARNING: binary creates a Linux executable. Use cross for macOS or Windows."
 	./scripts/build/binary
 
+.PHONY: binary-openbsd
+openbsd: ## build executable for openbsd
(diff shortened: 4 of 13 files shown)


cli/command/image/build_test.go
```diff
@@ -24,6 +24,8 @@ import (
 
 func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
 	skip.IfCondition(t, runtime.GOOS == "windows", "uid and gid not relevant on windows")
+	skip.IfCondition(t, runtime.GOOS == "openbsd", "gotestyourself/fs not working right on openbsd")
```
dnephin: Do you have a stack trace for the error you hit here?

cli/command/image/build_test.go
```diff
@@ -58,6 +60,7 @@ func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
 	}
 }
 func TestRunBuildDockerfileFromStdinWithCompress(t *testing.T) {
+	skip.IfCondition(t, runtime.GOOS == "openbsd", "not working right on openbsd")
```
dnephin: Could you elaborate on what didn't work so that we have a better idea of when this skip can be removed?

cli/command/formatter/history.go
```diff
@@ -79,7 +79,7 @@ func (c *historyContext) ID() string {
 }
 
 func (c *historyContext) CreatedAt() string {
-	return time.Unix(c.h.Created, 0).Format(time.RFC3339)
+	return time.Unix(c.h.Created, 0).UTC().Format(time.RFC3339)
```
dnephin: I'm not sure about this change. What happens without it?

voutilad: This turns out to be not OpenBSD specific now that I dug into it a bit more.

The test works fine if using Docker containers since it seems they're using UTC as the timezone.

However, if run on your local system with a timezone other than UTC, you get a string formatted that doesn't match the expected pattern since `time.Format` uses the local timezone. (So on my machine, it's `2009-11-10T18:00:00-05:00`.)

dnephin: Ok, so I think the formatter is probably correct, but we could adjust the tests to expect the time with the local timezone set, instead of assuming UTC.

cli/command/image/build_test.go
voutilad: Good question. This looks to be related to needing elevated permissions. If run as root, the tests pass.

As an unprivileged user:
```
$ go test -tags daemon -v github.com/docker/cli/cli/command/image
=== RUN   TestRunBuildResetsUidAndGidInContext
--- FAIL: TestRunBuildResetsUidAndGidInContext (0.01s)
        Error Trace:    file.go:62
                        build_test.go:38
        Error:          Received unexpected error:
                        chown /tmp/test-build-context-503524866/foo: operation not permitted
=== RUN   TestRunBuildDockerfileFromStdinWithCompress
--- FAIL: TestRunBuildDockerfileFromStdinWithCompress (0.03s)
        Error Trace:    build_test.go:70
                        client_test.go:121
                        build.go:390
                        build_test.go:98
        Error:          Received unexpected error:
                        lchown /tmp/test-build-compress-dest607440761/foo: operation not permitted
        Error Trace:    build_test.go:108
        Error:          Not equal: 
                        expected: []string{".dockerfile.69701b88c18326ae964e", ".dockerignore", "foo"}
                        received: []string{"foo"}
                    
                        Diff:
                        --- Expected
                        +++ Actual
                        @@ -1,4 +1,2 @@
                        -([]string) (len=3) {
                        - (string) (len=32) ".dockerfile.69701b88c18326ae964e",
                        - (string) (len=13) ".dockerignore",
                        +([]string) (len=1) {
                          (string) (len=3) "foo"
...
```

cli/command/image/build_test.go
voutilad: See above. Seems to be a permissions issue running as an unprivileged user. This also happens if doing `make test` on macOS. 

I'd say these aren't specific to OpenBSD then and really an issue of being able to run the unit tests without elevated privileges.

dnephin: Yup, sounds right. Thanks for looking into it. 

Would you mind changing the skip condition to be `os.Getuid() != 0` that way it's clear why the test is being skipped.

voutilad:
Also, I've been looking into a CI vendor that works with OpenBSD. Haven't found one yet. If that's a showstopper, I understand. I think some of the changes might be beneficial and cherry picked.

Worst case scenario I can maintain some OpenBSD specific changes in an OpenBSD port project and try to upstream some changes to vendor dependencies.

vendor/github.com/containerd/continuity/sysx/xattr_unsupported.go
```diff
@@ -1,4 +1,4 @@
-// +build freebsd solaris
+// +build openbsd freebsd solaris
```
thaJeztah: Are there pull requests for these changes upstream? Otherwise vendor-ci will keep failing :smile:

```
 D vendor/github.com/containerd/continuity/sysx/xattr_openbsd.go
 M vendor/github.com/containerd/continuity/sysx/xattr_unsupported.go
 M vendor/github.com/docker/docker/pkg/term/winsize.go
 D vendor/github.com/docker/docker/pkg/term/winsize_openbsd_cgo.go
 D vendor/github.com/tonistiigi/fsutil/diskwriter_openbsd.go

```

voutilad:
Good point :-)

I'll get to work on submitting PR's upstream.

vdemeester:
ping @voutilad what's the status here ?

dnephin:
We have a tracking issue for this now: #873

nogoegst:
@voutilad, all the changes are now tracked upstream for build to work, see #873. This PR should be either reduced to adding a build target for OpenBSD or closed in favor of new PR which does so.

dnephin:
Closing this since there hasn't been any work on this for months, and we have a tracking issue for deps.

PR #630, opened 2017-10-20
Add StorageOS volume plugin to legacy_plugins.md

oicheryl:
**- What I did**
Add volume plugin documentation for StorageOS.

**- How I did it**
Via Github.

**- How to verify it**
View https://docs.docker.com/engine/extend/legacy_plugins/.

Add volume plugin documentation for StorageOS.

Signed-off-by: Cheryl Hung <cheryljhung@gmail.com>


mstanleyjones:
My impression was that we were not adding new legacy plugins. @thaJeztah @vdemeester thoughts?

thaJeztah:
Yes, I think we should definitely prefer managed plugins over legacy plugins. I see that StorageOS has a managed plugin available https://docs.storageos.com/docs/install/docker/plugin, and published on Docker Store; https://store.docker.com/plugins/storageos, so that approach should probably be preferred.

@oicheryl is there any reason you want to add the legacy plugins to the docs, instead of users using the managed plugin?

oicheryl:
I saw that the page listed a bunch of volume plugins and didn't realise it was a legacy page. Is there's a better place to put it in the docs?

thaJeztah:
No, there's no replacement for managed plugins in the documentation. Maintaining a list of plugins gets easy out-of-sync with reality (plugin projects come and go), and with the introduction of Docker Store, we decided it was better to use that instead.

Let me go ahead and close this PR, but thanks for contributing 👍

oicheryl:
Understood. Thanks for the info @thaJeztah.

thaJeztah:
You're welcome!

PR #632, opened 2017-10-20
For system prune, -f is not an alias for --filter

mstanleyjones:
Fixes https://github.com/docker/docker.github.io/issues/5059

It's also weird that for `system events`, `-f` does mean `--filter`. It seems like maybe someone needs to do a consistency pass on these, as a follow-up.


thaJeztah:
@mstanleyjones yes, it's a known problem that some shorthand flags can be a ambiguous; also see https://github.com/moby/moby/issues/24433#issuecomment-231526484, and https://github.com/moby/moby/pull/24591

PR #633, opened 2017-10-20
Switch to golang:1.8.4-alpine3.6

seemethere:

**- What I did**
golang:1.8.4-alpine does not have multi-arch images available in the
manifest. (s390x, ppc64le, etc.)

This makes it so that if you are trying to compile on different
arches you aren't forced to have to write your own Dockerfile and can
instead use the one bundled with the CLI repo.

**- How I did it**
Switch the images to `golang:1.8.4-alpine3.6` and switch the package management commands to
`apt-get`

**- How to verify it**
`make -f docker.Makefile binary`

> Switched build image to allow for multi-arch compilation

kayak

### DIFF
--- a/dockerfiles/Dockerfile.dev
+++ b/dockerfiles/Dockerfile.dev
@@ -1,5 +1,5 @@
 
-FROM    golang:1.8.4-alpine
+FROM    golang:1.8.4-alpine3.6
 
 RUN     apk add -U git make bash coreutils ca-certificates
 
--- a/dockerfiles/Dockerfile.lint
+++ b/dockerfiles/Dockerfile.lint
@@ -1,4 +1,4 @@
-FROM    golang:1.8.4-alpine
+FROM    golang:1.8.4-alpine3.6
 
 RUN     apk add -U git
 


seemethere:
I understand this takes up more space on our machines but I think the benefit of being able to do multi-arch things more than makes up for some `MBs`

thaJeztah:
These are the "dev" and "lint" images; given that we don't do any multi-arch in this repository, wondering if this change should be in the packaging repo

dnephin:
The problem with non-alpine images isn't so much the space, more that downloading the base image and installing packages is much slower, which makes CI much slower. (not sure why CI hasn't run yet on this PR) 

Can't we just wait for https://github.com/gliderlabs/docker-alpine/pull/337 which will give us alpine multi-arch? It sounds like it's almost ready.

dnephin:
Alternatively, I think we could build our own mutli-arch golang alpine image and use that until the official one is ready.

seemethere:
@thaJeztah We already use `golang:1.8.4` to build the images anyway when creating static binaries. This would make it easier so that we don't have to write mechanisms like:

```
	docker run --rm -i -e VERSION=$(VERSION) -e GITCOMMIT=$(GITCOMMIT) \
		-v $(CURDIR)/docker-ce/components/cli:/go/src/github.com/docker/cli \
		-w /go/src/github.com/docker/cli \
		golang:1.8.4 make binary
```

Instead we could just do a:

```
make -C docker-ce/components/cli -f docker.Makefile binary
```

thaJeztah:
@seemethere I agree we should eventually move to that; my concern was that that change would currently be a step back for the way the image is used in this repo; is there a real urgency to change it right now?

seemethere:
@thaJeztah Our workarounds seem to work for now but it'd be nice to eventually get multi-arch support in the CLI repo outside of the `cross` target which doesn't actually get used downstream.

dnephin:
Yes, once the multi-arch golang-alpine images are ready it should "just work" without us having to make any changes.

thaJeztah:
Alpine is multi arch now; https://github.com/gliderlabs/docker-alpine/pull/352

thaJeztah:
Golang is almost there as well; https://github.com/docker-library/official-images/pull/3624

tophj-ibm:
golang is there now too 🎉 Just tested with `golang:1.8.4-alpine3.6` and that works. 

```
docker manifest inspect golang:1.8.4-alpine3.6 | grep architecture
            "architecture": "amd64",
            "architecture": "arm",
            "architecture": "arm64",
            "architecture": "386",
            "architecture": "ppc64le",
            "architecture": "s390x",
```

dnephin:
Great, so I guess we just need to change the base image to ` golang:1.8.4-alpine3.6`. @seemethere would you like to do that in this PR?

seemethere:
I will update the PR later today!!

thaJeztah:
@seemethere is there any change needed to the current version in this repo? Those should now automatically pick the right architecture, correct?

thaJeztah:
oh, never mind; we wanted to bump to alpine 3.6

seemethere:
Updated with a bump to `golang:1.8.4-alpine3.6`

vieux:
@seemethere go1.8.5 was just released

Issue #634, opened 2017-10-21
Create config files from a directory


moncho:
This adds support to the docker config subcommand to allow creating **config** files for every file found on a directory, if a directory is given to the subcommand.

If a directory is given, config names will be taken from the file names.  Existing limitations for config names still apply (invalid characters, length). 

E.g.
``` 
.
|_ my-conf-dir
     |_ foo.txt
     |_ bar.yaml
```

To create configs: 
```
$ docker config create my-conf-dir

gsggkx171uuv9fm2de4bxhh18
ubltrg4eqy97pdqkhm8x7c8vz
```
Checking the result:
```
$ docker config ls

ID                          NAME          CREATED            UPDATED
gsggkx171uuv9fm2de4bxhh18   foo.txt       2 minutes ago      2 minutes ago
ubltrg4eqy97pdqkhm8x7c8vz   bar.yaml      2 minutes ago      2 minutes ago
```
PR will soon follow.

dnephin:
Please provide some additional information about why you believe this feature needs to be part of the Docker cli.

moncho:
Sure. I think it is not uncommon having to create a few configuration files as part of your Swarm setup. In my experience, these files are typically stored together at the origin, downloaded locally and then iterated somehow (e.g. bash loop) running `docker config create` for each file. 

This feature does the iteration and the `docker config create` invocation on the CLI, easing the config creation process. Not a big functional change, but I find it useful and it does not break existing behavior.

moncho:
Closing the issue, see #635.

PR #635, opened 2017-10-21
Configs from dir

moncho:
**- What I did**

Add support for directories to the `docker config create` subcommand as described in #634.

**- How I did it**

A check has been added to see if the *file* parameter is a directory. If it is, its content will be listed and every regular file will be used to create a config. Otherwise, current behaviour applies.

**- How to verify it**

Run `docker config create` specifying a directory. A config will be created for every regular file on the directory.

Support config creation from files in a directory

cute

### DIFF
--- a/docs/reference/commandline/config_create.md
+++ b/docs/reference/commandline/config_create.md
@@ -1,0 +1,117 @@
+---
+title: "config create"
+description: "The config create command description and usage"
+keywords: ["config, create"]
+---
+
+<!-- This file is maintained within the docker/cli GitHub
+     repository at https://github.com/docker/cli/. Make all
+     pull requests against that repo. If you see this file in
+     another repository, consider it read-only there, as it will
+     periodically be overwritten by the definitive file. Pull
+     requests which include edits to this file in other repositories
+     will be rejected.
+-->
+
+# config create
+
+```Markdown
+Usage:	docker config create [OPTIONS] config file|-
+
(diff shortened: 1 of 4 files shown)


docs/reference/commandline/config_create.md
```diff
@@ -0,0 +1,117 @@
...
+$ docker config ls
+
+ID                          NAME                CREATED             UPDATED
+dg426haahpi5ezmkkj5kyl3sn   my_config           7 seconds ago       7 seconds ago
+```
+
+### Create configurations from a directory
+
+```bash
+$ docker config create  ./my-conf-dirs
```
mstanleyjones: What's the syntax on Windows?

dnephin:
I agree with @thaJeztah. Creating configs from a directory may be something that works in specific cases, but it's not something that translates well to the general case. For the specific cases where it does work it should be pretty easy to accomplish with a script.

moncho:
Thanks for reviewing the PR and sharing your thoughts!

I fully agree that the change is just for ease of use and the same functionality can be achieved using a script, but I still think it is useful functionality to have in the CLI. All the valid points raised by @thaJeztah can be easily fixed (creation can be made atomic, naming logic can be improved, etc).

Anyway since there is no agreement and there are other ways to achieve the same I will close the PR.

Cheers.

thaJeztah:
Thanks @moncho!

PR #636, opened 2017-10-23
Fix stdout and errors in image/trust

### LINKED ISSUE #631: `docker run` with content trust enabled pollutes stdout
**Description**

Doing a `docker run` with `DOCKER_CONTENT_TRUST` outputs to `stdout` things other than the container's stdout. In particular this message https://github.com/docker/cli/blob/master/cli/command/image/trust.go#L358

This is breaking some output formats of Moby tool. As seen in https://github.com/linuxkit/linuxkit/pull/2612.

Using LinuxKit and Moby tool manually to simulate the effect of `moby build -format gcp linuxkit.yml`:

1. `moby build -format tar-kernel-initrd linuxkit.yml` produces `linuxkit-initrd.tar`
2. `docker image ls| grep linuxkit/mkimage-gcp | cut -c 114-125 | xargs docker rmi`, to ensure a pull will happen in the next step
3. Convert to gcp:
```
cat linuxkit-initrd.tar | DOCKER_CONTENT_TRUST=1 docker run --network=none --rm -i linuxkit/mkimage-gcp:d1883809d212ce048f60beb0308a4d2b14c256af > img.tar
```
4. `file img.tar` shows `data` and not `tar file`
5.
```
$ head -n1 img.tar 
Tagging linuxkit/mkimage-gcp@sha256:d9571a557e4b82a944f12082cd50987d3726385b5458846cbae89ea9bd694c85 as linuxkit/mkimage-gcp:d1883809d212ce048f60beb0308a4d2b14c256af
```

I have also tried using `-a stdin -a stdout` without success.

```
$ file img.tar 
img.tar: data
$ head -n1 img.tar 
Tagging linuxkit/mkimage-gcp@sha256:d9571a557e4b82a944f12082cd50987d3726385b5458846cbae89ea9bd694c85 as linuxkit/mkimage-gcp:d1883809d212ce048f60beb0308a4d2b14c256af
```

```
$ file img.tar 
img.tar: gzip compressed data, last modified: Fri Oct 20 14:49:07 2017, max compression, from Unix
```

```
Client:
 Version:      17.10.0-ce
 API version:  1.33
 Go version:   go1.8.3
 Git commit:   f4ffd25
 Built:        Tue Oct 17 19:02:43 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.10.0-ce
 API version:  1.33 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   f4ffd25
 Built:        Tue Oct 17 19:01:22 2017
 OS/Arch:      linux/amd64
 Experimental: true
```

I also reproduced with 17.06 and 17.09.

**Output of `docker info`:**

```
Containers: 2
 Running: 0
 Paused: 0
 Stopped: 2
Images: 1700
Server Version: 17.10.0-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 06b9cb35161009dcb7123345749fef02f7cea8e0
runc version: 0351df1c5a66838d0c392b4ac4cf9450de844e2d
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.12.0-1-amd64
Operating System: Debian GNU/Linux buster/sid
OSType: linux
Architecture: x86_64
CPUs: 4
Total Memory: 15.53GiB
Name: bokrug
ID: AU33:BO7D:7VGM:MOLB:RSDF:IBRV:GCWT:THHM:OEVM:TX3C:BNLM:WHAR
Debug Mode (client): false
Debug Mode (server): false
Username: ijc25
Experimental: true
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false
```

I'm running on Linux (Debian). I believe @rn is seeing it on MacOS and LinuxKit CI is tripping over it too.

dnephin:
Fixes #631 

Move some output to stderr.
Moved some error output into the error message instead of as a separate print.
Changed the behaviour in some cases where the conenction to notary failed to error instead of exit without an error.

### DIFF
--- a/cli/command/image/trust.go
+++ b/cli/command/image/trust.go
@@ -84,7 +84,7 @@ func PushTrustedReference(streams command.Streams, repoInfo *registry.Repository
 		if err := jsonmessage.DisplayJSONMessagesToStream(in, streams.Out(), nil); err != nil {
 			return err
 		}
-		fmt.Fprintln(streams.Out(), "No tag specified, skipping trust metadata push")
+		fmt.Fprintln(streams.Err(), "No tag specified, skipping trust metadata push")
 		return nil
 	}
 
@@ -97,16 +97,14 @@ func PushTrustedReference(streams command.Streams, repoInfo *registry.Repository
 	}
 
 	if target == nil {
-		fmt.Fprintln(streams.Out(), "No targets found, please provide a specific tag in order to sign it")
-		return nil
+		return errors.Errorf("no targets found, please provide a specific tag in order to sign it")
 	}
 
 	fmt.Fprintln(streams.Out(), "Signing and pushing trust metadata")
 
 	repo, err := trust.GetNotaryRepository(streams.In(), streams.Out(), command.UserAgent(), repoInfo, &authConfig, "push", "pull")
 	if err != nil {
-		fmt.Fprintf(streams.Out(), "Error establishing connection to notary repository: %s\n", err)
-		return err
+		return errors.Wrap(err, "error establishing connection to trust repository")
 	}
 
 	// get the latest repository metadata so we can figure out which roles to sign
@@ -146,7 +144,7 @@ func PushTrustedReference(streams command.Streams, repoInfo *registry.Repository
 	}
 
 	if err != nil {
-		fmt.Fprintf(streams.Out(), "Failed to sign %q:%s - %s\n", repoInfo.Name.Name(), tag, err.Error())
+		err = errors.Wrapf(err, "failed to sign %s:%s", repoInfo.Name.Name(), tag)
 		return trust.NotaryError(repoInfo.Name.Name(), err)
 	}
 
@@ -223,8 +221,7 @@ func trustedPull(ctx context.Context, cli command.Cli, imgRefAndAuth trust.Image
 func getTrustedPullTargets(cli command.Cli, imgRefAndAuth trust.ImageRefAndAuth) ([]target, error) {
 	notaryRepo, err := cli.NotaryClient(imgRefAndAuth, trust.ActionsPullOnly)
 	if err != nil {
-		fmt.Fprintf(cli.Out(), "Error establishing connection to trust repository: %s\n", err)
-		return nil, err
+		return nil, errors.Wrap(err, "error establishing connection to trust repository")
 	}
 
 	ref := imgRefAndAuth.Reference()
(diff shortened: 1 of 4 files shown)


cli/command/image/trust.go
```diff
@@ -97,16 +97,14 @@ func PushTrustedReference(streams command.Streams, repoInfo *registry.Repository
 	}
 
 	if target == nil {
-		fmt.Fprintln(streams.Out(), "No targets found, please provide a specific tag in order to sign it")
-		return nil
```
dnephin: Note this is now an error, so the cli will exit with non-zero. Since the push was not  actually completely it seems like this should be an error.

cli/command/image/trust.go
```diff
@@ -355,7 +350,7 @@ func TagTrusted(ctx context.Context, cli command.Cli, trustedRef reference.Canon
 	familiarRef := reference.FamiliarString(ref)
 	trustedFamiliarRef := reference.FamiliarString(trustedRef)
 
-	fmt.Fprintf(cli.Out(), "Tagging %s as %s\n", trustedFamiliarRef, familiarRef)
+	fmt.Fprintf(cli.Err(), "Tagging %s as %s\n", trustedFamiliarRef, familiarRef)
```
endophage: Note sure whether this should go to stderr. This is the success message.

dnephin: This is the fix for the issue raised in #631.

The convention we should follow for stdout vs stderr error is:
* `stdout` is for "normal program output"
* `stderr` is for everything else (logging, warnings, etc)

This function is called from 3 commands. For both `image build` and `container create` this is definitely not normal program output, it's logging, so stderr is definitely correct.

For the last (`image pull`) it is a little less clear, but I think it could still easily be considered logging, since the desired action was "pull", not "tag".

endophage: I guess that depends on whether one expects command "X" to be different to command "X with trust." In my mind, we should send some minimal amount of info to stdout such that if I pipe stderr to /dev/null I can still determine the operation behaved as desired.

dnephin: This is not how posix programs are supported to work. If we send **anything** to stdout we break anyone trying to use `docker build` and `docker run` in a script, when used with trust. The only output to stdout should be normal program output, which is the output from the command being run by `docker run`.

There is already a mechanism that allows you to determine if the operation behaved correctly or not, the exit status code. This PR also fixes some problems where the exit status was not being used correctly.

PR #641, opened 2017-10-25
Add note about reported memory stats on linux

cpuguy83:
Since the API and the CLI both have a "Usage" field for memory, clarify
that the CLI does additional calculations to avoid confusion of API
consumers.

Closes moby/moby#35295

### DIFF
--- a/docs/reference/commandline/stats.md
+++ b/docs/reference/commandline/stats.md
@@ -34,6 +34,8 @@ The `docker stats` command returns a live data stream for running containers. To
 
 If you want more detailed information about a container's resource usage, use the `/containers/(id)/stats` API endpoint.
 
+> **Note**: On Linux, the Docker CLI reports memory usage by subtracting page cache usage from the total memory usage. The API does not perform such a calculation but rather provides the total memory usage and the amount from the page cache so that clients can use the data as needed.
+
 ## Examples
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/stats.md
```diff
@@ -34,6 +34,8 @@ The `docker stats` command returns a live data stream for running containers. To
 
 If you want more detailed information about a container's resource usage, use the `/containers/(id)/stats` API endpoint.
 
+Note: On Linux, the Docker CLI reports memory usage by subtracting page cache usage from the total memory usage. The API does not perform such a calculation but rather provides the total memory usage and the amount from the page cache so that clients can use the data as needed.
```
mstanleyjones: 1. Please use correct syntax for notes:

```
> **Note**: <blah blah>
```

2.  It strikes me that the people who need to see this note are API consumers rather than CLI ones so I am not sure that it will get to its intended audience here...

cpuguy83: The confusion comes from the CLI and people comparing CLI results with API results, which is why I put it here.

Putting this in the API docs is a little weird since the API effectively does nothing other than pass along the low-level details. Having the moby API reference the docker CLI seems backwards.

PR #642, opened 2017-10-25
Use upstream spf13 cobra

dnephin:
`spf13/cobra` finally merged all my PRs! So we can use upstream again. 

I think the man generation check might fail because the new `UseLine` adds an unnecessary `[flags]` after the command. I was able to work around that for the yaml docs and help text, but I'll need to submit another patch to cobra to fix it for the man pages.


cli/command/swarm/testdata/update-noargs.golden
```diff
@@ -8,6 +8,7 @@ Flags:
       --cert-expiry duration            Validity period for node certificates (ns|us|ms|s|m|h) (default 2160h0m0s)
       --dispatcher-heartbeat duration   Dispatcher heartbeat period (ns|us|ms|s|m|h) (default 5s)
       --external-ca external-ca         Specifications of one or more certificate signing endpoints
+  -h, --help                            help for update
```
thaJeztah: ah, yes, think we previously special-cased `-h / `--help` and hid it, because using that flag will print the same information as is being printed here.

cli/cobra.go
```diff
@@ -97,9 +98,19 @@ func managementSubCommands(cmd *cobra.Command) []*cobra.Command {
 	return cmds
 }
 
+// UseLine returns the usage line for a command. This implementation is different
+// from the default Command.UseLine in that it does not add a `[flags]` to the
+// of the line.
```
thaJeztah: missed a word here "to the ..... of the line"

cli/command/swarm/testdata/update-noargs.golden
dnephin: Ya, I think we might have been relying on a side-effect of an issue that was fixed in cobra to hide this flag, after we migrated to cobra.  I think we can restore the old behaviour by hiding this flag. I'll try it out.

PR #644, opened 2017-10-26
Fix daemon.json runtime example

thaJeztah:
The example in the documentation used "runc", which is a
reserved runtime name (as it's the default).

This patch updates the example, and uses a different name.

fixes https://github.com/docker/for-linux/issues/144

thanks @mcastelino for reporting!


mcastelino:
@thaJeztah love the new example 👍

PR #645, opened 2017-10-26
Hide help flag from help output

dnephin:
Follow up to #642

### DIFF
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -26,6 +26,7 @@ func SetupRootCommand(rootCmd *cobra.Command) {
 
 	rootCmd.PersistentFlags().BoolP("help", "h", false, "Print usage")
 	rootCmd.PersistentFlags().MarkShorthandDeprecated("help", "please use --help")
+	rootCmd.PersistentFlags().Lookup("help").Hidden = true
 }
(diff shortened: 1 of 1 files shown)


cli/cobra.go
```diff
@@ -26,6 +26,7 @@ func SetupRootCommand(rootCmd *cobra.Command) {
 
 	rootCmd.PersistentFlags().BoolP("help", "h", false, "Print usage")
 	rootCmd.PersistentFlags().MarkShorthandDeprecated("help", "please use --help")
+	rootCmd.PersistentFlags().Lookup("help").Hidden = true
```
thaJeztah: hm, didn't the `.golden` file need to be updated? https://github.com/docker/cli/pull/642/files#r147009761

dnephin: I expected it would, but it doesn't because of how the flags are applied.

This help flag is applied as a persistent flag to the root command, so when commands are added to the root, they inherit this version of the flag, and don't add the default.

In the test we are testing directly against the `swarm update` command, so the root command is not there to provide this flag. Cobra adds its own default help flag that is not hidden, so the test output still contains the `--help` flag.

PR #654, opened 2017-10-30
Fix behaviour of `rmi -f` with unexpected errors

dnephin:
Carry #418
Closes #418
Closes #567


dnephin:
This PR should actually fix the problem you see in 17.09. It'll be in 17.12

Issue #667, opened 2017-11-06
Add "name" field to secrets and configs to allow interpolation in Compose files


ilyasotkov:
In Compose file version 3.4, volumes gained the "name" field which allowed the volumes' names to be interpolated for simple version bumps using environmental variables.

That feature is also needed for secrets and configs. Usage example:

```yaml
version: "3.5" # future version

secrets:
  my-secret:
    name: my-secret-${VERSION} # name of the Docker Secret with variable interpolation
    file: ./secret.html

configs:
  my-config:
    name: my-config-${VERSION} # same thing for configs
    file: ./default.conf

services:
  web:
    image: nginx:1.13.5-alpine
    ports:
      - "80:80"
    secrets:
      - source: my-secret # reference by the name of the key, consistent with how it is for volumes now in 3.4
        target: /usr/share/nginx/html/index.html
    configs:
      - source: my-config
        target: /etc/nginx/conf.d/default.conf

```

I'll add a pull request for the feature shortly.

ilyasotkov:
https://github.com/docker/cli/pull/668

Issue #608, opened 2017-10-06
Don't warn on volume.external.name for versions before 3.4


dnephin:
Original Issue: https://github.com/moby/moby/issues/35111

PR #668, opened 2017-11-06
Add "name" field to secrets and configs to allow interpolation in Compose files

ilyasotkov:
See issue https://github.com/docker/cli/issues/667

In Compose file version 3.4, volumes gained the "name" field which allowed the volumes' names to be interpolated for simple version bumps using environmental variables.

That feature is also needed for secrets and configs. Proposed usage:

```yaml
version: "3.5" # future Compose file version

secrets:
  my-secret:
    name: my-secret-${VERSION} # name of the Docker Secret with variable interpolation
    file: ./secret.html

configs:
  my-config:
    name: my-config-${VERSION} # same thing for configs
    file: ./default.conf

services:
  web:
    image: nginx:1.13.6-alpine
    ports:
      - "80:80"
    secrets:
      - source: my-secret # reference by the name of the key, consistent with how it is for volumes now in 3.4
        target: /usr/share/nginx/html/index.html
    configs:
      - source: my-config
        target: /etc/nginx/conf.d/default.conf

```

**What I did:** I just followed the way the feature is already implemented for volumes. I also bumped the Compose file schema version to 3.5.


cli/compose/convert/compose.go
```diff
@@ -97,6 +97,10 @@ func Networks(namespace Namespace, networks networkMap, servicesNetworks map[str
 func Secrets(namespace Namespace, secrets map[string]composetypes.SecretConfig) ([]swarm.SecretSpec, error) {
 	result := []swarm.SecretSpec{}
 	for name, secret := range secrets {
+		if secret.Name != "" {
+			name = secret.Name
+		}
```
dnephin: This is not used if `secret.External.External` evaluates to true, so I think this block should move after the next `if secret.External.External` block. 

Also, this name should not be scoped using `namespace.Scope(name)` (which is done below). So there needs to be an `else` for doing the `namespace.Scope()`


Same for configs below (I wish these two types shared more code...)

cli/compose/loader/loader.go
```diff
@@ -486,9 +486,17 @@ func LoadSecrets(source map[string]interface{}, workingDir string) (map[string]t
...
 	for name, secret := range secrets {
-		if secret.External.External && secret.External.Name == "" {
-			secret.External.Name = name
-			secrets[name] = secret
+		if secret.External.External {
+			if secret.External.Name == "" {
+				secret.External.Name = name
+				secrets[name] = secret
+			} else {
+				logrus.Warnf("secret %s: secret.external.name is deprecated in favor of secret.name", name)
```
dnephin: There's an issue open for this warning with volumes #608

This has the same problem, we'll need to pass in a version here so that we only warn `if version >= 3.5`

cli/compose/convert/service.go
```diff
@@ -271,6 +271,10 @@ func convertServiceSecrets(
 			source = secretSpec.External.Name
 		}
 
+		if secretSpec.Name != "" {
+			source = namespace.Scope(secretSpec.Name)
+		}
```
dnephin: If the name is specified it shouldn't be scoped.

dnephin: ya, I think it's by design that the `name` field is not scoped. Only the default names get scoped.

You can use a variable in the `name:` field to provide the scope to these explicitly named objects.

dnephin:
Opened #670 to fix the warnings on volume. For `LoadSecrets` and `LoadConfigObjs` you could pass in the `configDetails` since they already accept a field from that struct.

#670 also made a change so that `external.name` is never read

dnephin:
I opened #671 to remove a lot of the duplication between secrets and configs. This PR is going to conflict quite a bit. You might want to rebase on top of #671 so that you don't want to make all the changes twice (once for secrets, and once for configs).

ilyasotkov:
So I rewrote the PR on top of https://github.com/docker/cli/pull/671 to remove some of the duplication between configs / secrets and fixed some issues the original PR had as well.

I also added a couple of tests for the functionality, but I wasn't sure if duplicating secrets and configs there is the right way to go.

cli/compose/convert/compose.go
```diff
@@ -101,18 +101,17 @@ func Secrets(namespace Namespace, secrets map[string]composetypes.SecretConfig)
 			continue
 		}
 
-		data, err := ioutil.ReadFile(secret.File)
+		if secret.Name != "" {
+			name = secret.Name
+		} else {
+			name = namespace.Scope(name)
+		}
```
dnephin: This if/else block can be moved into `fileObjectConfig()` to remove the duplication.

cli/compose/loader/loader.go
```diff
@@ -484,42 +484,67 @@ func LoadVolumes(source map[string]interface{}, version string) (map[string]type
...
+			obj.External.Name = ""
+		} else {
+			if obj.Name == "" {
+				obj.Name = name
+			}
+		}
+		// if not "external: true"
+	} else {
+		if obj.File == "" {
+			return obj, errors.Errorf("%[1]s %[2]s: specify a file or \"external: true\"", objType, name)
```
dnephin: Is this a new error?

ilyasotkov: It is, although I definitely should've written a test on it 🙃 
It doesn't work,`secrets.<secret-name> must be a mapping` is being called before this one. I'll remove it for now, but it's probably a good idea to make all such errors a bit more specific in the future.

ilyasotkov:
Updated.

ilyasotkov:
So, what else needs to be done to get this merged? 👋

ilyasotkov:
Rebased ✅

YuryHrytsuk:
Question:

Why `name` field of secrets is not documented anywhere below ?
* https://docs.docker.com/reference/compose-file/services/#secrets
* https://docs.docker.com/reference/compose-file/secrets/

PR #678, opened 2017-11-08
Add bash completion for experimental `build --stream`

albers:
This adds bash completion for #231.
The new flag is only considered when the deamon runs in experimental mode.


thaJeztah:
ping @tonistiigi PTAL

PR #680, opened 2017-11-09
Add the ability to restart all running containers

TylerLubeck:
Signed-off-by: Tyler Lubeck <tyler@tylerlubeck.com>

Hi! We've been running in to some bumps at work lately, and we were kicking around how much we'd like to have `docker restart --all` in our toolbelt. Overall this might be something I'd like to see in a couple of commands, but I wanted to 1) see if it was possible to do and 2) see what everybody else thought of it.

**- What I did**
I added a `--all` flag to `docker restart`, as restarting all running containers is something we've been needing to do a lot recently. `docker restart $(docker ps -aq)` works fine, but isn't super discoverable to some of our newer team members

**- How I did it**
I make an API call to the docker engine to get the list of running containers, then add those container IDs to the list of containers specified

**- How to verify it**
Start some containers, run `docker restart --all`, make sure that they're restarted

Add the ability to restart all running containers

cute alpacas

### DIFF
--- a/cli/command/container/restart.go
+++ b/cli/command/container/restart.go
@@ -5,8 +5,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/docker/api/types"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"golang.org/x/net/context"
@@ -15,6 +15,8 @@ import (
 type restartOptions struct {
 	nSeconds        int
 	nSecondsChanged bool
+
+	nAll bool
 
 	containers []string
 }
@@ -26,7 +28,12 @@ func NewRestartCommand(dockerCli command.Cli) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "restart [OPTIONS] CONTAINER [CONTAINER...]",
 		Short: "Restart one or more containers",
-		Args:  cli.RequiresMinArgs(1),
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 && !cmd.Flags().Changed("all") {
+				return errors.New("\"docker restart\" requires at least 1 argument.")
+			}
+			return nil
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			opts.containers = args
 			opts.nSecondsChanged = cmd.Flags().Changed("time")
@@ -36,6 +43,7 @@ func NewRestartCommand(dockerCli command.Cli) *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.IntVarP(&opts.nSeconds, "time", "t", 10, "Seconds to wait for stop before killing the container")
+	flags.BoolVar(&opts.nAll, "all", false, "Restart all running containers")
 	return cmd
 }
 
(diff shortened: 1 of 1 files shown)


cli/command/container/restart.go
```diff
@@ -16,6 +16,8 @@ type restartOptions struct {
 	nSeconds        int
 	nSecondsChanged bool
 
+	nAll bool
```
TylerLubeck: I couldn't tell the naming convention here so I went with it

cli/command/container/restart.go
```diff
@@ -26,7 +28,12 @@ func NewRestartCommand(dockerCli command.Cli) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "restart [OPTIONS] CONTAINER [CONTAINER...]",
 		Short: "Restart one or more containers",
-		Args:  cli.RequiresMinArgs(1),
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 1 && !cmd.Flags().Changed("all") {
+				return errors.New("\"docker restart\" requires at least 1 argument.")
+			}
+			return nil
```
TylerLubeck: I wasn't totally sure how to handle this - IMO listing containers and specifying `--all` should be mutually exclusive. Alternatively, we could append the containers retrieved from the docker engine to the list of containers specified (which is currently what I'm doing) but this means that containers might be restarted twice unless we treat the list like a set and do processing on that. 

Open to suggestions, for the time being I went with the faster coding option for the sake of getting input

cli/command/container/restart.go
```diff
@@ -48,6 +56,17 @@ func runRestart(dockerCli command.Cli, opts *restartOptions) error {
 		timeout = &timeoutValue
 	}
 
+	if opts.nAll {
```
ripcurld0: I think it will be more readable like that:
```Go
if opts.nAll {
    var err error
    opts.containers, err = dockerCli.Client().ContainerList(context.Background(), types.ContainerListOptions{})
    if err != nil {
        errs = append(errs, err.Error())
    }
}
```

In general, if you are passing `--all` you would like to set `opts.containers` to all containers that are running. Thus, there is no need to save it in a different variable and then extend `opts.containers`.

cli/command/container/restart.go
ripcurld0: I think `all` is good enough. Don't know what the `n` prefix stands for, though.

cli/command/container/restart.go
ripcurld0: It looks ok just change the message to something like:
```
"docker restart" requires at least 1 argument unless "--all" was set
```

dnephin:
Thanks for the contribution!

Unfortunately I don't think we should add this flag:
1. (as you mentioned) this is already easy to accomplish with `docker restart $(docker ps -aq)`
2. having flags that change the accepted arguments is something that I think we should try to avoid, and as you mentioned `--all` with a list of contains should be an error.
3. I don't think this will be a common use case. Often an engine will be running multiple workloads, and only a subset of the containers should be restarted. There is also `docker-compose restart` that can be used to restart a bunch of containers at once (if they are defined in a Compose file).

> `docker restart $(docker ps -aq)` works fine, but isn't super discoverable to some of our newer team members

Perhaps this could be documented in a wiki, or put into a script to make it easier to run?

thaJeztah:
I agree with @dnephin here; I don't think it's a common scenario to restart all containers, and _if_ people want to restart multiple containers, it'll likely be based on certain criteria (i.e. filters are needed).

Combining commands on the command-line is common practice (think of simple examples, such as `<some command> | grep <something>`), and should be the recommended approach for such things.

I'm closing this PR, but thanks for contributing, and feel free to discuss

PR #682, opened 2017-11-10
Fix and simplify bash completion for service env, mounts and labels

albers:
`service create` and `service update` both used to have `--env`, `--label` and `--mount` options.
These options now are only valid for `service create`.
`service update` got corresponding `--xxx-add|rm` options instead.

Also removes some duplication between the completions for `--xxx` and the corresponding
`-xxx-add` and `-xxx-rm` options.


albers:
rebased because of a conflict with `--generic-resources-(add|rm)`

PR #685, opened 2017-11-10
docker trust sign: add e2e testing

eiais:

### DIFF
--- a/e2e/trust/sign_test.go
+++ b/e2e/trust/sign_test.go
@@ -1,0 +1,84 @@
+package trust
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/gotestyourself/gotestyourself/fs"
+	"github.com/gotestyourself/gotestyourself/icmd"
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	notaryURL    = "https://notary-server:4443"
+	alpineImage  = "registry:5000/alpine:3.6"
+	alpineSha    = "641b95ddb2ea9dc2af1a0113b6b348ebc20872ba615204fbe12148e98fd6f23d"
+	busyboxImage = "registry:5000/busybox:1.27.2"
+	busyboxSha   = "030fcb92e1487b18c974784dcc110a93147c9fc402188370fbfd17efabffc6af"
+	localImage   = "registry:5000/signlocal:v1"
+	signImage    = "registry:5000/sign:v1"
+)
+
+func TestSignLocalImage(t *testing.T) {
+	dir := setupConfigFile(t)
+	defer dir.Remove()
+	icmd.RunCmd(icmd.Command("docker", "pull", alpineImage)).Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "tag", alpineImage, signImage).Assert(t, icmd.Success)
+	result := icmd.RunCmd(
+		icmd.Command("docker", "trust", "sign", signImage),
+		withTrustAndPassphrase("root_password", "repo_password", dir))
+	result.Assert(t, icmd.Success)
+	assert.Contains(t, result.Stdout(), fmt.Sprintf("v1: digest: sha256:%s", alpineSha))
+
+}
+
+func TestSignWithLocalFlag(t *testing.T) {
+	dir := setupConfigFile(t)
+	defer dir.Remove()
+	setupTrustedImageForOverwrite(t, dir)
+	result := icmd.RunCmd(
+		icmd.Command("docker", "trust", "sign", "--local", localImage),
+		withTrustAndPassphrase("root_password", "repo_password", dir))
+	result.Assert(t, icmd.Success)
+	assert.Contains(t, result.Stdout(), fmt.Sprintf("v1: digest: sha256:%s", busyboxSha))
+}
+
+func withTrustAndPassphrase(rootPwd, repositoryPwd string, dir fs.Dir) func(cmd *icmd.Cmd) {
+	return func(cmd *icmd.Cmd) {
+		env := append(os.Environ(),
+			"DOCKER_CONTENT_TRUST_SERVER="+notaryURL,
+			"DOCKER_CONTENT_TRUST_ROOT_PASSPHRASE="+rootPwd,
+			"DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE="+repositoryPwd,
+			"DOCKER_CONFIG="+dir.Path(),
+		)
+		cmd.Env = append(cmd.Env, env...)
+	}
+}
+
+func setupConfigFile(t *testing.T) fs.Dir {
+	dir := fs.NewDir(t, "trust_test", fs.WithMode(0700), fs.WithFile("config.json", `
+	{
+		"auths": {
+			"registry:5000": {
+				"auth": "ZWlhaXM6cGFzc3dvcmQK"
+			},
+			"https://notary-server:4443": {
+				"auth": "ZWlhaXM6cGFzc3dvcmQK"
+			}
+		}
+	}
+	`))
+	return *dir
+}
+
+func setupTrustedImageForOverwrite(t *testing.T, dir fs.Dir) {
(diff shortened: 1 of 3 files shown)


e2e/trust/sign_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+func TestSignWithTrustSign(t *testing.T) {
+	setupConfigFile(t)
+	defer removeConfigFile(t)
+	icmd.RunCmd(icmd.Command("docker", "pull", alpineImage)).Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "tag", alpineImage, signImage).Assert(t, icmd.Success)
+	result := icmd.RunCmd(
+		icmd.Command("docker", "-D", "trust", "sign", signImage),
+		withTrustAndPassphrase("root_password", "repo_password"))
+	result.Assert(t, icmd.Success)
```
dnephin: Can we make any assertions about the stdout/stderr here? 

`golden.Assert()` makes it easy to store the expected output in a file and update it when it changes. 

If you expect nothing on stderr (or stdout) you can use `icmd.Expected{Out: icmd.None, Err: icmd.None}` to assert that it's empty. The default is to allow any string.

e2e/trust/sign_test.go
```diff
@@ -0,0 +1,80 @@
...
+			"DOCKER_CONTENT_TRUST_SERVER="+notaryURL,
+			"DOCKER_CONTENT_TRUST_ROOT_PASSPHRASE="+rootPwd,
+			"DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE="+repositoryPwd,
+		)
+		cmd.Env = append(cmd.Env, env...)
+	}
+}
+
+func setupConfigFile(t *testing.T) {
+	err := os.Mkdir("/root/.docker", os.FileMode(0700))
```
dnephin: Someone might run these tests outside of a container (even though they shouldn't) so we should use a temporary directory so that we don't clobber their config.

`fs.NewDir()` (https://godoc.org/github.com/gotestyourself/gotestyourself/fs) should let you create this easily and `fs.WithMode()` should let you set the mode.

This method can return the `fs.Dir` so you can `defer dir.Remove() to cleanup, and `dir.Path()` to get the path for the temporary directory.


The file in the directory can be created using `fs.WithFile("config.json", `....`)`

e2e/trust/sign_test.go
```diff
@@ -0,0 +1,80 @@
...
+func TestSignWithTrustSignlocal(t *testing.T) {
+	setupConfigFile(t)
+	defer removeConfigFile(t)
+	icmd.RunCmd(icmd.Command("docker", "pull", alpineImage)).Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "tag", alpineImage, localImage).Assert(t, icmd.Success)
+	result := icmd.RunCmd(
+		icmd.Command("docker", "-D", "trust", "sign", localImage),
+		withTrustAndPassphrase("root_password", "repo_password"))
+	result.Assert(t, icmd.Success)
+	icmd.RunCommand("docker", "tag", busyboxImage, localImage).Assert(t, icmd.Success)
```
dnephin: If everything above this is "setup" code then it would be great to extract it into a function that is named accordingly (`setupLocallyTrustedImage()` maybe?).  The idea here is to make it really easy for someone reading the test to be able to tell what is setup, and what is the actual command being tested.

e2e/trust/sign_test.go
```diff
@@ -0,0 +1,80 @@
...
+	result = icmd.RunCmd(
+		icmd.Command("docker", "trust", "sign", "--local", localImage),
+		withTrustAndPassphrase("root_password", "repo_password"))
+	result.Assert(t, icmd.Success)
```
dnephin: same question/comment about checking stdout/stderr here.

e2e/trust/sign_test.go
dnephin: I guess this failed on CI as well because some other test created this directory already

e2e/trust/sign_test.go
```diff
@@ -0,0 +1,90 @@
...
+	notaryURL    = "https://notary-server:4443"
+	alpineImage  = "registry:5000/alpine:3.6"
+	alpineSha    = "641b95ddb2ea9dc2af1a0113b6b348ebc20872ba615204fbe12148e98fd6f23d"
+	busyboxImage = "registry:5000/busybox:1.27.2"
+	busyboxSha   = "030fcb92e1487b18c974784dcc110a93147c9fc402188370fbfd17efabffc6af"
+	localImage   = "registry:5000/signlocal:v1"
+	signImage    = "registry:5000/sign:v1"
+)
+
+func TestSignWithTrustSign(t *testing.T) {
```
dnephin: What is "WithTrustSign" ? What is the case that is being tested here?  Is it just a local image? `TestSignLocalImage` ?

e2e/trust/sign_test.go
```diff
@@ -0,0 +1,90 @@
...
+	icmd.RunCommand("docker", "tag", alpineImage, signImage).Assert(t, icmd.Success)
+	result := icmd.RunCmd(
+		icmd.Command("docker", "trust", "sign", signImage),
+		withTrustAndPassphrase("root_password", "repo_password", dir))
+	result.Assert(t, icmd.Success)
+	assert.Contains(t, result.Stdout(), fmt.Sprintf("v1: digest: sha256:%s", alpineSha))
+
+}
+
+func TestSignWithTrustSignlocal(t *testing.T) {
```
dnephin: This case is not just `--local`, it's actually that it's writing over a previous signature, is that right? Should this be `TestSignWithLocalOverrides` or something like that?

e2e/trust/sign_test.go
```diff
@@ -0,0 +1,90 @@
...
+            "https://notary-server:4443": {
+            	"auth": "ZWlhaXM6cGFzc3dvcmQK"
+			}
+        }
+	}
+	`))
+	return *dir
+}
+
+func removeConfigFile(dir fs.Dir) {
```
dnephin: seems kind of unnecessary to have a function (with a longer name) to call a single other function. Maybe just `defer dir.Remove()` in the tests themselves?

e2e/trust/sign_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func setupConfigFile(t *testing.T) fs.Dir {
+	dir := fs.NewDir(t, "trust_test", fs.WithMode(0700), fs.WithFile("config.json", `
+	{
+		"auths": {
+
+			"registry:5000": {
+				"auth": "ZWlhaXM6cGFzc3dvcmQK"
+			},
+            "https://notary-server:4443": {
```
dnephin: formatting loosk off here, looks like mixed tabs and spaces

e2e/trust/sign_test.go
```diff
@@ -0,0 +1,90 @@
...
+	`))
+	return *dir
+}
+
+func removeConfigFile(dir fs.Dir) {
+	dir.Remove()
+}
+
+func setupTrustedImageForOverwrite(t *testing.T) fs.Dir {
+	dir := setupConfigFile(t)
```
dnephin: I think you'll want to do this in the calling test function and pass in `dir` so that you can defer immediately. Otherwise a failure in any of these setup calls will result in the temp directory not being cleaned up.

e2e/trust/sign_test.go
eiais: The only way of writing over the previous signature is with --local and that's really its only use

e2e/trust/sign_test.go
eiais: as opposed to DCT=1 push, I can change it.

e2e/trust/sign_test.go
dnephin: Ok, so maybe `TestSignWithLocalFlag` ?

riyazdf:
post-facto LGTM

PR #695, opened 2017-11-16
Add info about new port mapping syntax

mistyhacks:
New port-mapping syntax has been around since 1.13 but hasn't been documented until now. These docs need to go here first, then there needs to be a followup PR in the Docs repo to update all examples in the narrative docs to use the new syntax.

I need technical review on this and I also need to verify that you can't set the `mode` at all using the short syntax. Is that right?

cc/ @mbentley who will be super happy about this.

Partly-fixes https://github.com/docker/docker.github.io/issues/3913

panda cub

### DIFF
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -735,51 +735,75 @@ Containers on the same network can access each other using
 ### Publish service ports externally to the swarm (-p, --publish)
 
 You can publish service ports to make them available externally to the swarm
-using the `--publish` flag:
-
-```bash
-$ docker service create --publish <TARGET-PORT>:<SERVICE-PORT> nginx
-```
-
-For example:
+using the `--publish` flag. The `--publish` flag can take two different styles
+of arguments. The short version is positional, and allows you to specify the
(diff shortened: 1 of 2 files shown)


docs/reference/commandline/service_create.md
```diff
@@ -735,51 +735,75 @@ Containers on the same network can access each other using
...
 
-#### TCP and UDP
+The options you can specify are:
 
-```bash
-$ docker service create --name dns-cache -p 53:53/tcp -p 53:53/udp dns-cache
-```
-
-#### UDP only
+<table>
```
ripcurld0: I personally prefer to create tables using [markdown syntax](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables).

mistyhacks: Me too! However, `docker/docker.github.io` uses these docs in a partial way (we grab the `## Description` and `## Examples` sections of the files) and our Markdown processor runs into trouble with Markdown tables in some  cases. It's a hell of a thing to debug, so in these reference we tend to favor HTML tables except for the most simple cases.

PR #696, opened 2017-11-16
attach: Use ContainerWait instead of Inspect

mlaventure:
Signed-off-by: Kenfe-Mickael Laventure <mickael.laventure@gmail.com>

--

Addresses https://github.com/moby/moby/issues/35497

**- What I did**

Use `ContainerWait()` instead of `Inspect` to retrieve the final exit code in atach

**- How I did it**

Changed the code ;-)

**- How to verify it**
```
$ docker run -f --name exit alpine sh -c 'sleep 10; exit 21'
$ docker attach exit
$ echo $?
21
```

Ensure attach exit code matches container's

:dog:

### DIFF
--- a/cli/command/container/attach.go
+++ b/cli/command/container/attach.go
@@ -1,12 +1,14 @@
 package container
 
 import (
+	"fmt"
 	"io"
 	"net/http/httputil"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/signal"
 	"github.com/pkg/errors"
@@ -65,6 +67,9 @@ func NewAttachCommand(dockerCli command.Cli) *cobra.Command {
 func runAttach(dockerCli command.Cli, opts *attachOptions) error {
 	ctx := context.Background()
 	client := dockerCli.Client()
+
+	// request channel to wait for client
+	resultC, errC := client.ContainerWait(ctx, opts.container, "")
 
 	c, err := inspectContainerAndCheckState(ctx, client, opts.container)
 	if err != nil {
@@ -140,7 +145,24 @@ func runAttach(dockerCli command.Cli, opts *attachOptions) error {
 	if errAttach != nil {
 		return errAttach
 	}
-	return getExitStatus(ctx, dockerCli.Client(), opts.container)
+
+	return getExitStatus(errC, resultC)
+}
+
+func getExitStatus(errC <-chan error, resultC <-chan container.ContainerWaitOKBody) error {
+	select {
+	case result := <-resultC:
+		if result.Error != nil {
+			return fmt.Errorf(result.Error.Message)
+		}
+		if result.StatusCode != 0 {
+			return cli.StatusError{StatusCode: int(result.StatusCode)}
+		}
+	case err := <-errC:
+		return err
+	}
+
+	return nil
 }
 
 func resizeTTY(ctx context.Context, dockerCli command.Cli, containerID string) {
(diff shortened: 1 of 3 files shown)


mlaventure:
Sure, will give it a go

cli/command/container/attach.go
```diff
@@ -66,6 +66,9 @@ func runAttach(dockerCli command.Cli, opts *attachOptions) error {
 	ctx := context.Background()
 	client := dockerCli.Client()
 
+	// request channel to wait for client
+	resultC, errC := client.ContainerWait(ctx, opts.container, "")
```
dnephin: Oh, actually I think this should come after `inspectContainerAndCheckState()` which is a sanity check on the state of the container.

mlaventure: I wanted to avoid a race when by the time we return from `inspectContainerAndCheckState()` the container had exited, but I don't mind moving it afterwards.

dnephin: good point, maybe it is more correct before the inspect. The error from `ContainerWait()` wouldn't be reported until later, so the user would still see the helpful error message.

I think this is correct, ignore my comment.

mlaventure:
I'm not sure what to do about the gocyclo complain. I don't want to make a function just because of it. So if we don't want to raise the limit, I would need to think of a way of refactoring the would function.

Regarding the test failure, it is very strange, it looks like we do get the output, but don't notice that the process is dead? I haven't been able to reproduce it locally.

cli/command/container/attach.go
```diff
@@ -140,7 +143,18 @@ func runAttach(dockerCli command.Cli, opts *attachOptions) error {
 	if errAttach != nil {
 		return errAttach
 	}
-	return getExitStatus(ctx, dockerCli.Client(), opts.container)
+
+	// get status code
```
dnephin: This seems like a perfect place to create a new function. A function is always better than an inline comment as it provides a scope for the documentation.

I think the unit test you removed could be preserved and used to test this, with a few small changes to the test setup. The 3 cases in the table test are still relevant.

e2e/container/attach_test.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func TestAttachExitCode(t *testing.T) {
+	cName := "test-attach-exit-code"
+	icmd.RunCommand("docker", "run", "-d", "--rm", "--name", cName,
+		alpineImage, "sh", "-c", "sleep 5 ; exit 21").Assert(t, icmd.Success)
+	cmd := icmd.Command("docker", "wait", cName)
+	res := icmd.StartCmd(cmd)
+	icmd.RunCommand("docker", "attach", cName).Assert(t, icmd.Expected{ExitCode: 21})
+	icmd.WaitOnCmd(8, res).Assert(t, icmd.Expected{ExitCode: 0, Out: "21"})
```
dnephin: This is only waiting 8ms, I'm not sure if that's related to the failure.

I'll see if I can get this working

dnephin: This test fails for me locally as well so I'll prepare a patch to get it working. 

I think the `docker wait` command is unnecessary, all we need is a `poll.WaitOn()` for the container to be started.

mlaventure: Didn't realize it was in `ms` :sweat_smile: 

Thanks `poll.WaitOn()` seems to be what I was looking for. I'll try updating the code tonight.

cli/command/container/attach.go
mlaventure: It feels weird given that it's really just a select, the comment could be removed without affecting readability, but I'll try that.

dnephin:
I pushed a commit to this branch which I think will fix the e2e suite

e2e/container/attach_test.go
dnephin: I pushed a commit, it seems to be working now. I didn't use the WaitOn for container start. I'm not sure if this test can flake, maybe `runBackgroundContainsWithExitCode()` could use a WaitOn if we see it flake.

mlaventure:
All green :green_heart:

tonistiigi:
What caused this change? Was there a problem in all previous versions or did the API change? The linked issue seems like a regression.

mlaventure:
From my understanding of the code, the race was always here, is just that a bit more of the post processing has been moved to the event handler now and increased the latency.

tonistiigi:
LGTM

thaJeztah:
Looks like we have 2 LGTM's; merging

andrewhsu:
FYI this PR don't play nice with latest code from `moby/moby` in docker-ce 18.02.0 rc1. Getting [failures](https://jenkins.dockerproject.org/job/docker-ce-pr-s390x/333/execution/node/509/log/) like:
```
panic: DockerSuite.TestAttachAfterDetach test timed out after 5m0s [recovered]
	panic: DockerSuite.TestAttachAfterDetach test timed out after 5m0s
```
[Reverting](https://github.com/docker/docker-ce/pull/406) the changes in this PR and stuff goes green.

dnephin:
That test seems to have a 2s timeout for `attach` so I'm not sure how it's failing this way. I'll investigate

dnephin:
Seems to be an issue with the test, Opened moby/moby#36363

PR #711, opened 2017-11-22
Support network.name in the compose schema

dnephin:
Follow up to #668 now that volume, config, and secrets all support a custom name, apply the same change to networks for consistency.


dnephin:
cc @shin-

ilyasotkov:
It would also make sense to add this for services, so that all 5 Docker objects (configs, secrets, volumes, networks, and services) have the option to specify a custom and unscoped name.

dnephin:
cc @thaJeztah @vieux It would be good to get this in for 17.12 so that all the resources are consistent.

PR #712, opened 2017-11-23
Add raw input in docker-compose for secrets and configs

### LINKED ISSUE #320: Support for raw config in docker-compose file
Hi

I can't find a way to add a raw config in docker-compose file, so I think it's not possible
What I am trying to do is to deploy a stack with a config like :

```
version: "3.3"
services:
  caddy:
    image: caddy
    configs:
      - source: caddyfile
        target: /etc/Caddyfile
configs:
  caddyfile:
    raw: |
0.0.0.0
browse
log stdout
errors stdout
```

I think it impact the compose format itself

Cheers

WTFKr0:
Closes #320 

**- What I did**
I add a new raw field for secret and config creation in docker-compose file

**- How I did it**
I changed the schema v3.5 to add this field
I modify the way config and secret are loaded from file to add option to load from a string

**- How to verify it**
With this compose file for example : 
```
version: "3.5"
services:
  web:
    image: nginx
    configs:
      - source: rawconfig
        target: /rawconfig
    secrets:
      - onelineraw
      - multilineraw
      - filesecret
configs:
  rawconfig:
    raw: |
      Raw data
      of config
secrets:
  onelineraw:
    raw: "onelineraw data"
  multilineraw:
    raw: |
      a secret
      on

      multi lines
  filesecret:
    file: $PWD/mysecret
```
Once running, we got : 
```
docker exec -it 6f6ce919151b grep "" /rawconfig /run/secrets/filesecret /run/secrets/multilineraw /run/secrets/onelineraw
/rawconfig:Raw data
/rawconfig:of config
/run/secrets/filesecret:sqdfsfsdfsdok
/run/secrets/multilineraw:a secret
/run/secrets/multilineraw:on
/run/secrets/multilineraw:
/run/secrets/multilineraw:multi lines
/run/secrets/onelineraw:onelineraw data
```

Add raw input in docker-compose for secrets and configs

image


WTFKr0:
Yeah, I'll try to add that
If raw and file are present, no error for the moment, and raw is taken in priority

WTFKr0:
Should I force user to set `file` or `raw` ?
It's no sense to create a config without one of those fields I think. I tried with that : 
```
secrets:
  empty:
    name: plop
```
And `docker stack deploy` cli told me that my $PWD is a directory

dnephin:
yes, one should be required

WTFKr0:
After some tests, i think i got it with
```json
      "oneOf": [
        {
          "required": ["file"],
          "not": {"required": ["raw"]}
        },
        {
          "required": ["raw"],
          "not": {"required": ["file"]}
        }
      ],
```

But errors messages are like that : 
- When neither file or raw : 
`configs.configname file is required`
- When both file and raw : 
`configs.configname Must not validate the schema (not)`

WTFKr0:
Ahh need to allow no file and no raw for external config/secret...

cli/compose/schema/schema_test.go
```diff
@@ -65,6 +67,46 @@ func TestValidateSecretConfigNames(t *testing.T) {
...
+		},
+		"secrets": dict{
+			"baz": dict{
+				"name": "foobaz",
+			},
+		},
+	}
+
+	err := Validate(config, "3.5")
+	assert.Error(t, err)
```
dnephin: This should test for the error  with `EqualError()`, or `ErrorContains()`

cli/compose/schema/schema_test.go
```diff
@@ -65,6 +67,46 @@ func TestValidateSecretConfigNames(t *testing.T) {
...
+		"secrets": dict{
+			"baz": dict{
+				"file": "myfile",
+				"raw":  "mydata",
+			},
+		},
+	}
+
+	err := Validate(config, "3.5")
+	assert.Error(t, err)
```
dnephin: same here

cli/compose/schema/schema_test.go
```diff
@@ -65,6 +67,46 @@ func TestValidateSecretConfigNames(t *testing.T) {
...
+		"secrets": dict{
+			"baz": dict{
+				"file": "myfile",
+				"raw":  "mydata",
+			},
+		},
+	}
+
+	err := Validate(config, "3.5")
+	assert.EqualError(t, err, "configs.bar Must not validate the schema (not)")
```
dnephin: We need to find some way to provide a better error message here.

dnephin: It looks like these errors are handled in `cli/compose/schema/schema.go:getMostSpecificError()`. We probably need a special case for `not` that provides more detail.

WTFKr0: ok thanx, will check that soon

WTFKr0:
D'you want me to resolve the conflict by merging last master onto this branch and regenerate bindata ?

dnephin:
Ah yes, please do

thaJeztah:
> do you see any issues with accepting secrets and configs from the Compose file?

The bit I'm a bit worried about is that putting secrets in this file effectively nullifies all the security measures that went into the `docker secrets` design (i.e., keep the secrets secure, don't store them on disk). Putting a secret in the compose file is a big risk; i.e. compose files are usually committed to version control, which (if they contain secrets) would be a big no-no.

What are the benefits of putting the secrets in this file, instead of a separate file? (If you want to store the secret in a file)

thaJeztah:
For configs, we can be a bit more flexible, BUT, in many occasions those may contain secrets as well, so I'd like https://github.com/moby/moby/pull/33702 to be implemented so that secrets in config-files would no longer be needed (i.e. I can have a config-template that refers to a secret to use)

WTFKr0:
My use case is a docker-compose file commited with a demo secret or demo config, it's better for users who want to test an app to have a one command deployment (`docker stack deploy`) that do all things !
Also, we often used compose file for a one step deployment, even in production. I think we can modify compose to set config to wanted value, deploy, and remove compose-file
We used this with secret environment variable as well (for passwords for example)

thaJeztah:
But adding a separate demo-secret to the repository, and reference that from the compose file would bring the same, correct?

WTFKr0:
We often just provide a docker-compose file, not an entire git repo with all files
It's more like a catalog of compose-files, ready to deploy, and the user can override some values on the fly with what he wants
It's for a use case like a stacks self service

dnephin:
> The bit I'm worried about is that putting secrets in this file effectively nullifies all the security measures that went into the docker secrets design (i.e., keep the secrets secure, don't store them on disk). 

Supporting the `raw` field actually improves the situation. Right now the only way to provide secrets in a Compose file is by putting them into a file. With `raw` someone can use environment variables and the secret will be interpolated into the file, without ever having to store it on disk.

thaJeztah:
Sorry, I'm still -1 on this; as @WTFKr0's use case describes; this is gonna be used to _replace_ environment variables (separate from the problem that environment variables are really not suitable for storing secrets), and store the literal secrets in a compose file.

I would like the security team to review this before we go ahead

endophage:
@dnephin do you mean apart from the compose file in which the secret is written to disk and even worse, likely committed in to a VCS in some repo not specifically designed for protecting secrets?

If the compose/stack file is committed in to a repo, there is no reason not to simply commit your demo secrets in to files beside the compose file, or as we do in notary, in to a folder called something like `fixtures` next to the compose file, making it clear that they are not production secrets.

Beyond secrets, how much testing has been done for different file formats in raw? Does yaml work? How about toml, or ini?

From a security perspective, I strongly dislike this. From a testing/quality perspective, it's completely lacking in tests for various popular config file types that consider whitespace important.

dnephin:
>  do you mean apart from the compose file in which the secret is written to disk and even worse, likely committed in to a VCS in some repo not specifically designed for protecting secrets?

I don't understand what you're asking.

Currently there is only one way to use secrets/config with the Compose file. You must have them in a file on the local disk. By adding a `raw` field, we would support:

```yaml
secrets:
  foo:
    raw: "$FOO_SECRET"
```

Which removes the need to store something on disk.

> likely committed in to a VCS

I don't think this is a fair assumption. Even if it is committed to disk, isn't that choice (and the associated risk) obvious to the user? For most dev and testing use cases that is unlikely to be a problem.

> how much testing has been done for different file formats in raw? Does yaml work? How about toml, or ini?

This is not a concern. Yaml supports raw string value, the content is irrelevant, it is always interpreted as a string literal.

```
In [1]: src = """
   ...: field: |
   ...:   inline:
   ...:     yaml: [works, fine]
   ...:
   ...: """

In [2]: import yaml

In [3]: yaml.load(src)
Out[3]: {'field': 'inline:\n  yaml: [works, fine]\n'}
```

WTFKr0:
Hey all

Any information on status of this PR pls ?

Thanx for reading

thaJeztah:
> Any information on status of this PR pls ?

I'm still not convinced we should do this 😅 

> Currently there is only one way to use secrets/config with the Compose file. You must have them in a file on the local disk. By adding a raw field, we would support:
> ..
> `raw: "$FOO_SECRET"`

Secrets were explicitly designed to prevent using env-vars for secrets; that example (albeit not on the container itself) re-introduces the usage of environment variables to store them.

I also disagree that "there is only one way to use secrets/config"; secrets/configs can be created up-front (which allows creating them from stdin or whatever means), and can be referenced using `external:`.

dnephin:
> Secrets were explicitly designed to prevent using env-vars for secrets

Container/image environment variables are a problem because they are stored in the config and viewable from `inspect`. When the engine is a multi-user host, anyone with accept to the engine can see them. That's clearly a problem.

I believe regular process environment variables are only viewable from `/proc` by the user who created the process (or root). Why are those a problem?  I don't think we should assume the two are the same thing.

WTFKr0:
I need that for configs too
The process for reading configs and secret is the same for the moment, this is why i do it for both
Are you OK if we enable this feature only for configs, and not for secrets ?

Thanx

WTFKr0:
Hi all
Any thoughts on my last post ?

thaJeztah:
Also have a look at https://github.com/moby/moby/pull/33702, which may have some overlap

dnephin:
I don't think it overlaps.

It would be great to get this feature in for configs at least, if there are still concerns about using it for secrets.

WTFKr0:
Hey, just reviewing my old staled PR on github 😄 
What do we do for this one ?

Thanx

WTFKr0:
Hi

Any update ??

dazinator:
@thaJeztah 
My use case:

I am deploying a `stack` from an azure devops pipeline.
When the pipeline runs, azure devops exposes "secrets" e.g from KeyVault, as variables and environment variables to tasks that run during the CI/CD pipeline.

Right now, I have to either
- Write these sensitive values to disk on the build agent prior to doing `docker stack deploy` so that my stack (compose file) can own these secrets - e.g not external. This is simpler than the next approach, but less secure as I am forced to write them out to disk.
- Deploy them as secrets passing them as `std in` to dockr stack secret create. The in my compose file, have the secrets as `external` but also use a variable name for each secret such that when the value changes I can increment the variable name to rotate the secret.


The former is a security risk forced upon me, because I can't use `compose spec` yet (which as this which loooks promising https://docs.docker.com/compose/compose-file/09-secrets/#example-2) and the latter is an extra layer of headaches to content with - considering some services have 20 secrets, and only one value is changes, there has to be logic to work out which secret value changes, and a scheme to rotate these names appropriately such that updating a secret does not require a source code change to the docker compose file.

I would not be keen on using `raw` if it meant I still have to write the secret value to disk in the compose file as I may as well just write the values to secret files firts and reference them as files in compose, that doesn't solve anything.
However if `raw` suppoorts environment variable interpolation then this would solve the issue for me, or if the `compose spec` was implemented so I can use the new top level secret `environment` property - basically so I cann supply the value of the secret at deployment time without writing it to disk, and without it having to be external.

PR #715, opened 2017-11-29
Return error status when ps brings no containers

whoan:
Signed-off-by: Juan Eugenio Abadie <juaneabadie@gmail.com>

**- What I did**

I added a new validation to `ps` to know if the command matched some container. It can be very useful for scripting (avoiding greping the output).

**- How I did it**

I checked if `ps` is returning any container in the `containers` array.

**- How to verify it**

```
$ docker ps && echo "Some container found"
$ docker ps -f filter=value && echo "Some container found"
```

Return an error status when `ps [-f filter=value]` brings no containers.

### DIFF
--- a/cli/command/container/list.go
+++ b/cli/command/container/list.go
@@ -2,6 +2,7 @@ package container
 
 import (
 	"io/ioutil"
+	"errors"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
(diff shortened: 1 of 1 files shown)


vdemeester:
thanks @whoan, it's a bit tricky as it might break some people's script (that will now get an error code different than 0 when they previously got 0).

/cc @dnephin @thaJeztah for thoughts on this :ang

whoan:
In fact I'm trying to simulate what Unix-like `ps -C` does. Thanks for consider it.

dnephin:
```
cli/command/container/list.go:1::warning: file is not gofmted with -s (gofmt)
cli/command/container/list.go:1::warning: file is not goimported (goimports)
```

ripcurld0:
I would suggest setting a flag to determine whether return an error if no image was found.

whoan:
Thanks @thaJeztah for the information. I wll try with `docker inspect` when needed so.

thaJeztah:
Thanks @whoan !

PR #717, opened 2017-11-29
Improve and fix bash completion for images

albers:
# Shortcomings of the current solution
Bash completion of images uses the vanilla output of `docker images` and filters it with `awk`.
This is error prone because users may customize the output of `docker images` by setting a go template in _~/.docker/config.json_ with the `imagesFormat` key.
And it is ugly.

There are several valid combinations of repository, tag and ID in Docker CLI. This diversity is not sufficiently covered yet, leading to issues like https://github.com/moby/moby/issues/27791.

Furthermore, problems arise because completion of images is customizable through environment variables:
* The user may hide tags by setting `DOCKER_COMPLETION_SHOW_TAGS=no`. 
* The user may choose the amount of image IDs shown in completion: `DOCKER_COMPLETION_SHOW_IMAGE_IDS=none(default)|non-intermediate|all`

These settings are not fine-grained enough, though.
* In cases where an exact image specification is required, e.g. `docker history`, we have to show tags even if the user opted to hide them.
* IDs are not legal in all image specifications.

# Solution
* Use explicit formatting in `docker images` like in the completions of containers, networks, volumes... .
* Specify which of repository, tag and id are valid with arguments (`--repo`, `--tag`, `--id`) to one central completion function and let this function decide what to complete wrt environment settings.
* Create an override for cases where tags are required (`--force-tag`)

We now use the following combinations:
`__docker_complete_images --force-tag --id`
`__docker_complete_images --repo`
`__docker_complete_images --repo --tag`
`__docker_complete_images --repo --tag --id`

I prefer to leave the invocations as listed here instead of creating 4 helper functions because these would not increase readibility. The explicit list of legal choices  very clearly expresses differences between related completetions, see `images --filter (before|since)=` and `images --filter reference=` for an example.

# Fixes
* https://github.com/moby/moby/issues/27791 Fix bash completion of images when default tag does not exist
For these cases, `--repo` is not used and `--force-tags` is applied. Only valid tags are completed.
* completion of `image ls repo` and `image pull --all-tags repo` failed for repository names containing a `:` (e.g. localhost:5000/some-repo)

# Changes
Completion of `image import` now completes local files instead of nothing. The previous choice was made because URLs are also valid syntax. The result feels like a broken completion.

Some minor changes in style were applied in a separate commit in order to increase code consistency.

# Testing
If you have bash completion installed and working, just source the updated completion script.
The main helper function can also be called directly to study the influence of the customizion via environment variables, e.g.

```bash
__docker_images --repo --tag --id
DOCKER_COMPLETION_SHOW_IMAGE_IDS=none __docker_images --repo --tag --id
DOCKER_COMPLETION_SHOW_IMAGE_IDS=non-intermediate __docker_images --repo --tag --id
DOCKER_COMPLETION_SHOW_TAGS=no __docker_images --force-tag
```

ping @mlaventure Fixes https://github.com/moby/moby/issues/27791, PTAL

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -1372,10 +1372,7 @@ _docker_container_commit() {
 			if [ "$cword" -eq "$counter" ]; then
 				__docker_complete_containers_all
 				return
-			fi
-			(( counter++ ))
-
-			if [ "$cword" -eq "$counter" ]; then
+			elif [ "$cword" -eq "$((counter + 1))" ]; then
 				__docker_complete_images --repo --tag
 				return
 			fi
@@ -2593,7 +2590,7 @@ _docker_image_history() {
 			COMPREPLY=( $( compgen -W "--format --help --human=false -H=false --no-trunc --quiet -q" -- "$cur" ) )
 			;;
 		*)
-			local counter=$(__docker_pos_first_nonflag)
+			local counter=$(__docker_pos_first_nonflag '--format')
 			if [ "$cword" -eq "$counter" ]; then
 				__docker_complete_images --force-tag --id
 			fi
@@ -2619,11 +2616,9 @@ _docker_image_import() {
 		*)
 			local counter=$(__docker_pos_first_nonflag '--change|-c|--message|-m')
 			if [ "$cword" -eq "$counter" ]; then
+				_filedir
 				return
-			fi
-			(( counter++ ))
-
-			if [ "$cword" -eq "$counter" ]; then
+			elif [ "$cword" -eq "$((counter + 1))" ]; then
 				__docker_complete_images --repo --tag
 				return
 			fi
@@ -2798,10 +2793,7 @@ _docker_image_tag() {
 			if [ "$cword" -eq "$counter" ]; then
 				__docker_complete_images --force-tag --id
 				return
-			fi
-			(( counter++ ))
-
-			if [ "$cword" -eq "$counter" ]; then
+			elif [ "$cword" -eq "$((counter + 1))" ]; then
 				__docker_complete_images --repo --tag
 				return
 			fi
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -178,52 +178,78 @@ __docker_complete_container_ids() {
 	COMPREPLY=( $(compgen -W "${containers[*]}" -- "$cur") )
 }
 
+# __docker_images returns a list of images. For each image, up to three representations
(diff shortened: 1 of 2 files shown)


contrib/completion/bash/docker
```diff
@@ -178,52 +178,78 @@ __docker_complete_container_ids() {
...
+				fi
+				shift
+				;;
+			--id)
+				if [[ $DOCKER_COMPLETION_SHOW_IMAGE_IDS =~ ^(all|non-intermediate)$ ]] ; then
+					format+="$id_format\n"
+				fi
+				shift
+				;;
+			--force-tag)
```
mistyhacks: This seems like it should be additive with `--tag`, like `--tag --force` or something like that.

contrib/completion/bash/docker
```diff
@@ -178,52 +178,78 @@ __docker_complete_container_ids() {
...
+				break
+				;;
+		esac
+	done
 
-	__docker_q images $images_args | awk "$awk_script" | grep -v '<none>$'
+	__docker_q image ls --no-trunc --format "${format%\\n}" $all "$@" | grep -v '<none>$'
 }
 
+# __docker_complete_images appies completion of images based on the current value of `$cur` or
```
mistyhacks: nit: typo "appies" should be "applies"

contrib/completion/bash/docker
albers: This won't work because it is not clear what should be forced. My intent is to use `--repo`, `--tag` and `--id` to express that a representation _can_ be completed (if permitted by environment settings) and to use `--force-repo`, `--force-tag` and `--force-id` when a representation has to be added, regardless of environment settings. Of the latter, currently only `--force-tag` is used, so the other options from this set are not yet implemented.

contrib/completion/bash/docker
albers: Oh yes, I'll change that. Thanks.

contrib/completion/bash/docker
albers: I also want to keep this function very simple. That's why it doesn't care about the order of the arguments. Thus, a `--force is not deterministic`. And I want the invocations to fit on one line, so I opted for the short names like `--tag` instead of something more exact like `--try-tag` or `--allow-tag`.
I think this simplicity is appropriate for this function because it is only used internally in the bash completion script.

thaJeztah:
whoops forgot about this one LGTM, thanks!

albers:
@thaJeztah Would you mind looking at #682? It's been hanging around even longer.

PR #749, opened 2017-12-13
Add support for generic resources to bash completion

albers:
Adds bash completion for #429:
- `service create --generic-resource`
- `service update --generic-resource-(add|rm)`
- `dockerd --node-generic-resource`

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -2221,6 +2221,7 @@ _docker_daemon() {
 		--metrics-addr
 		--mtu
 		--network-control-plane-mtu
+		--node-generic-resource
 		--oom-score-adjust
 		--pidfile -p
 		--registry-mirror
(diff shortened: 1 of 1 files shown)


contrib/completion/bash/docker
```diff
@@ -3459,6 +3461,8 @@ _docker_service_update_and_create() {
 			--dns-search-rm
 			--group-add
 			--group-rm
+			--generic-resource-add
```
thaJeztah: Do you want to keep this alphabetical? If so, it should be above the `--group-...` lines 🤗

albers: The alphabet is really comlicated...
Thanks for catching. PR is updated.

thaJeztah:
😄 thanks!

merging, so that I can cherry-pick

PR #755, opened 2017-12-15
Show container wait error

kolyshkin:
If container wait has failed, show an error from the engine and return an appropriate exit code.

This relies on engine changes from https://github.com/moby/moby/pull/34999

### DIFF
--- a/cli/command/container/utils_test.go
+++ b/cli/command/container/utils_test.go
@@ -1,0 +1,69 @@
+package container
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/docker/cli/internal/test"
+	"github.com/docker/docker/api"
+	"github.com/docker/docker/api/types/container"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/net/context"
+)
+
+func waitFn(cid string) (<-chan container.ContainerWaitOKBody, <-chan error) {
+	resC := make(chan container.ContainerWaitOKBody)
+	errC := make(chan error, 1)
+	var res container.ContainerWaitOKBody
+
+	go func() {
+		switch {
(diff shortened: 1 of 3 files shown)


dnephin:
Looks like we are missing unit test coverage for this function. Would you mind writing a test for the error case and the success case? It should be easy to produce an error using the fake client in `cli/command/container/client_test.go` (just add the wait method), and `NewFakeCli()`.

kolyshkin:
@dnephin test cases added, PTAL

cli/command/container/utils_test.go
```diff
@@ -0,0 +1,69 @@
...
+			exitCode: 125,
+		},
+		{
+			cid:      "non-existent-container-id",
+			exitCode: 125,
+		},
+	}
+
+	client := test.NewFakeCli(&fakeClient{waitFunc: waitFn, Version: api.DefaultVersion})
+	//t.Logf("client version: %v\n", client.ClientVersion())
```
vdemeester: You can remove that one 👼

cli/command/container/utils_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"golang.org/x/net/context"
+)
+
+func waitFn(cid string) (<-chan container.ContainerWaitOKBody, <-chan error) {
+	resC := make(chan container.ContainerWaitOKBody)
+	errC := make(chan error, 1)
+	var res container.ContainerWaitOKBody
+
+	go func() {
+		if strings.Contains(cid, "exit-code-42") {
```
dnephin: nit: could use a `switch/case`

```go
switch {
case strings.Contains(cid, "exit-code-42"):
 ...
}
```

but not a big deal

kolyshkin:
Commits updated as per @vdemeester and @dnephin suggestions; I think it's ready to be merged.

PR #758, opened 2017-12-20
Add support for experimental Cli configuration

vdemeester:
Allow to mark some commands and flags experimental on cli (i.e. not
depending to the state of the daemon). This will allow more flexibility
on experimentation with the cli.

Marking `docker trust` as cli experimental as it is documented so.

```bash
$ docker trust --help
docker trust is only supported on a Docker cli with experimental features enabled
# edit docker client configuration file
$ cat ~/.docker/config.json
{
        "experimentalCli": "enabled"
}
$ docker trust --help
Usage:  docker trust COMMAND

Manage trust on Docker images (experimental)

Options:


Management Commands:
  key         Manage keys for signing Docker images (experimental)
  signer      Manage entities who can sign Docker images (experimental)

Commands:
  revoke      Remove trust for an image
  sign        Sign an image
  view        Display detailed information about keys and signatures

Run 'docker trust COMMAND --help' for more information on a command.
```

This is required for https://github.com/docker/cli/pull/721/ (as we will put the k8s support under experimental at first)

jpeg image 271 x 186 pixels

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -42,24 +42,25 @@ type Cli interface {
 	SetIn(in *InStream)
 	ConfigFile() *configfile.ConfigFile
 	ServerInfo() ServerInfo
+	ClientInfo() ClientInfo
 	NotaryClient(imgRefAndAuth trust.ImageRefAndAuth, actions []string) (notaryclient.Repository, error)
 }
 
 // DockerCli is an instance the docker command line client.
 // Instances of the client can be returned from NewDockerCli.
 type DockerCli struct {
-	configFile     *configfile.ConfigFile
-	in             *InStream
-	out            *OutStream
-	err            io.Writer
-	client         client.APIClient
-	defaultVersion string
-	server         ServerInfo
+	configFile *configfile.ConfigFile
+	in         *InStream
+	out        *OutStream
+	err        io.Writer
+	client     client.APIClient
+	serverInfo ServerInfo
+	clientInfo ClientInfo
 }
 
 // DefaultVersion returns api.defaultVersion or DOCKER_API_VERSION if specified.
 func (cli *DockerCli) DefaultVersion() string {
-	return cli.defaultVersion
+	return cli.clientInfo.DefaultVersion
 }
 
 // Client returns the APIClient
@@ -104,7 +105,12 @@ func (cli *DockerCli) ConfigFile() *configfile.ConfigFile {
 // ServerInfo returns the server version details for the host this client is
 // connected to
 func (cli *DockerCli) ServerInfo() ServerInfo {
-	return cli.server
+	return cli.serverInfo
+}
+
+// ClientInfo returns the client details for the cli
+func (cli *DockerCli) ClientInfo() ClientInfo {
+	return cli.clientInfo
 }
 
 // Initialize the dockerCli runs initialization that must happen after command
@@ -125,17 +131,34 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	if err != nil {
 		return err
 	}
+	hasExperimental, err := isEnabled(cli.configFile.Experimental)
+	if err != nil {
+		return errors.Wrap(err, "Experimental field")
+	}
+	cli.clientInfo = ClientInfo{
+		DefaultVersion:  cli.client.ClientVersion(),
+		HasExperimental: hasExperimental,
+	}
 	cli.initializeFromClient()
 	return nil
 }
 
-func (cli *DockerCli) initializeFromClient() {
-	cli.defaultVersion = cli.client.ClientVersion()
+func isEnabled(value string) (bool, error) {
+	switch value {
+	case "enabled":
+		return true, nil
+	case "", "disabled":
+		return false, nil
+	default:
+		return false, errors.Errorf("%q is not valid, should be either enabled or disabled", value)
+	}
+}
 
+func (cli *DockerCli) initializeFromClient() {
 	ping, err := cli.client.Ping(context.Background())
 	if err != nil {
 		// Default to true if we fail to connect to daemon
-		cli.server = ServerInfo{HasExperimental: true}
+		cli.serverInfo = ServerInfo{HasExperimental: true}
 
 		if ping.APIVersion != "" {
 			cli.client.NegotiateAPIVersionPing(ping)
@@ -143,7 +166,7 @@ func (cli *DockerCli) initializeFromClient() {
 		return
 	}
 
-	cli.server = ServerInfo{
+	cli.serverInfo = ServerInfo{
 		HasExperimental: ping.Experimental,
 		OSType:          ping.OSType,
 	}
@@ -174,6 +197,12 @@ func (cli *DockerCli) NotaryClient(imgRefAndAuth trust.ImageRefAndAuth, actions 
 type ServerInfo struct {
 	HasExperimental bool
 	OSType          string
+}
+
+// ClientInfo stores details about the supported features of the client
+type ClientInfo struct {
+	HasExperimental bool
+	DefaultVersion  string
 }
 
 // NewDockerCli returns a DockerCli instance with IO output and error streams set by in, out and err.
--- a/cli/command/cli_test.go
+++ b/cli/command/cli_test.go
@@ -1,17 +1,18 @@
 package command
 
 import (
+	"crypto/x509"
 	"os"
 	"testing"
 
-	"crypto/x509"
-
+	cliconfig "github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/cli/cli/flags"
 	"github.com/docker/cli/internal/test/testutil"
 	"github.com/docker/docker/api"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
+	"github.com/gotestyourself/gotestyourself/fs"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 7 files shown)


cli/command/cli.go
```diff
@@ -125,25 +133,36 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
...
+		HasExperimental: isExperimentalCli(cli.configFile),
+	}
 	cli.initializeFromClient()
 	return nil
 }
 
+func isExperimentalCli(configfile *configfile.ConfigFile) bool {
+	if e := os.Getenv("DOCKER_EXPERIMENTAL_CLI"); e != "" {
+		value := strings.ToLower(e)
+		return value == "1" || value == "y" || value == "yes"
```
AkihiroSuda: no support for "true"?

Probably strconv.ParseBool should be used

vdemeester: Thx @AkihiroSuda 🤗

cli/command/cli.go
```diff
@@ -125,25 +133,36 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
...
 	}
+	cli.clientInfo = ClientInfo{
+		HasExperimental: isExperimentalCli(cli.configFile),
+	}
 	cli.initializeFromClient()
 	return nil
 }
 
+func isExperimentalCli(configfile *configfile.ConfigFile) bool {
+	if e := os.Getenv("DOCKER_EXPERIMENTAL_CLI"); e != "" {
```
silvin-lubecki: I don't think this environment variable is useful. Contrary to the DOCKER_ORCHESTRATOR one which you can use to switch between orchestrators, I think switching to experimental features should be a "long term" choice from the user.

cli/command/cli.go
```diff
@@ -125,25 +133,36 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	if err != nil {
 		return err
 	}
+	cli.clientInfo = ClientInfo{
+		HasExperimental: isExperimentalCli(cli.configFile),
```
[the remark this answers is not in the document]
vdemeester: Ah right !

vdemeester:
@thaJeztah Interesting, I like that aprroach 👼

cli/command/cli.go
```diff
@@ -54,7 +55,8 @@ type DockerCli struct {
 	err            io.Writer
 	client         client.APIClient
 	defaultVersion string
-	server         ServerInfo
+	serverInfo     ServerInfo
+	clientInfo     ClientInfo
```
dnephin: We could move `defaultVersion` to `clientInfo` I think, but doesn't need to happen in this PR.

cmd/docker/docker.go
```diff
@@ -260,6 +276,9 @@ func isSupported(cmd *cobra.Command, details versionDetails) error {
 			if _, ok := f.Annotations["experimental"]; ok && !hasExperimental {
 				errs = append(errs, fmt.Sprintf("\"--%s\" is only supported on a Docker daemon with experimental features enabled", f.Name))
 			}
+			if _, ok := f.Annotations["experimentalCli"]; ok && !hasExperimentalCli {
+				errs = append(errs, fmt.Sprintf("\"--%s\" is only supported on a Docker cli with experimental features enabled", f.Name))
```
dnephin: "only supported when experimental cli features are enabled"

I just realized all of this was duplicated. We should be able to declare this visitor once, and call it for the root command case. Doesn't need to be part of this PR.

cmd/docker/docker.go
```diff
@@ -243,6 +256,9 @@ func isSupported(cmd *cobra.Command, details versionDetails) error {
 		if _, ok := curr.Annotations["experimental"]; ok && !hasExperimental {
 			return fmt.Errorf("%s is only supported on a Docker daemon with experimental features enabled", cmd.CommandPath())
 		}
+		if _, ok := curr.Annotations["experimentalCli"]; ok && !hasExperimentalCli {
+			return fmt.Errorf("%s is only supported on a Docker cli with experimental features enabled", cmd.CommandPath())
```
dnephin: "only supported when experimental cli features are enabled"

dnephin:
> From a user-perspective, I wonder if we should have some connection between "server experimental" and "client experimental".

I think this implies there is some relationship between the two, when in reality there is not. I think it would be better to keep them separate. Even using different names for them would be great (beta vs experimental).

cli/command/cli_test.go
```diff
@@ -125,12 +127,63 @@ func TestInitializeFromClient(t *testing.T) {
...
+	for _, testcase := range testcases {
+		t.Run(testcase.doc, func(t *testing.T) {
+			tmpDir, cleanup := createTemporaryConfigfile(t, testcase.doc, testcase.configfile)
+			defer cleanup()
+			apiclient := &fakeClient{
+				version: defaultVersion,
+			}
+
+			cli := &DockerCli{client: apiclient, err: os.Stderr}
+			t.Log("yo ", tmpDir)
```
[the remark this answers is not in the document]
vdemeester: oups 🤣

thaJeztah:
The tricky bit is that something being "client side" or "daemon side" should largely be an implementation detail; most users won't be aware where the functionality lives, and I'm not sure we should put that burden onto them if not needed for the 99% (but we can still consider having more fine-grained options in addition).

Some features may even move from cli to daemon (or vice versa)

dnephin:
I think the reality is that the user needs to be aware of where it's implemented because a single client can connect to more than one server. 

We could maybe avoid some of the confusing by using feature flags instead of a global "experimental". So users could enable trust, or k8s-stacks, instead of enabling everything.

cmd/docker/docker.go
vdemeester: @dnephin yep, we realized that too in #721 just didn't want to add more code to it 😛

vdemeester:
@thaJeztah @dnephin Updated with comments taken into account. Switch to `enabled` and `disabled` as value so we could, in the future, have more options for it 😉

thaJeztah:
`enabled|disabled` LGTM, but still would love to see `auto`, as it feels natural to me to (if no configuration is set) have the CLI follow the daemon (and not require people to set the "experimental" option in two locations)

> because a single client can connect to more than one server.

Different topic, but we could really use a way to set multiple configurations (per daemon/host), either by having a `nodes: [{}]` option in the configuration file, or (likely more flexible) having subdirectories with configuration files per-host (a-la docker-machine). We have `--config` already, but I can see there's configurations that are "global", and ones that are "per host"

vdemeester:
@thaJeztah could we add `auto` on a follow-up PR (to discuss it) ? 👼 

> Different topic, but we could really use a way to set multiple configurations (per daemon/host), either by having a nodes: [{}] option in the configuration file, or (likely more flexible) having subdirectories with configuration files per-host (a-la docker-machine).

Yes ! This is in my TODOs 👼

cli/command/cli.go
```diff
@@ -125,25 +131,42 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
...
-func (cli *DockerCli) initializeFromClient() {
-	cli.defaultVersion = cli.client.ClientVersion()
+func isEnabled(value string) (bool, error) {
+	switch value {
+	case "enabled":
+		return true, nil
+	case "", "disabled":
+		return false, nil
+	default:
+		return false, errors.Errorf("%q is not valid, should be either enabled or disabled", value)
```
thaJeztah: Think we should mention the configuration option here, currently it shows:

```
$ docker version
"Moby was here" is not valid, should be either enabled or disabled
```

Perhaps something like:

```
Invalid value for "experimentalCli": "Moby was here". Valid options are "enabled" and "disabled"
```

(better suggestions welcome)

cli/config/configfile/file.go
```diff
@@ -44,6 +44,7 @@ type ConfigFile struct {
 	NodesFormat          string                      `json:"nodesFormat,omitempty"`
 	PruneFilters         []string                    `json:"pruneFilters,omitempty"`
 	Proxies              map[string]ProxyConfig      `json:"proxies,omitempty"`
+	ExperimentalCli      string                      `json:"experimentalCli,omitempty"`
```
thaJeztah: `CLI` is an acronym, so this should be `ExperimentalCLI` (and `experimentalCLI` for JSON)

Given that this configuration file is for the CLI; is `CLI` a bit redundant? Perhaps just `Experimental`?

cli/command/cli.go
```diff
@@ -125,25 +131,42 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	if err != nil {
 		return err
 	}
+	hasExperimental, err := isEnabled(cli.configFile.ExperimentalCli)
+	if err != nil {
+		return err
```
dnephin: I guess this should `errors.Wrapf(err, "ExperimentalCLI field")` so that the error message mentions which field has a problem

vdemeester:
cc @AkihiroSuda @yongtang

cli/command/cli_test.go
```diff
@@ -124,13 +125,51 @@ func TestInitializeFromClient(t *testing.T) {
...
 			cli.initializeFromClient()
-			assert.Equal(t, defaultVersion, cli.defaultVersion)
-			assert.Equal(t, testcase.expectedServer, cli.server)
+			assert.Equal(t, testcase.expectedServer, cli.serverInfo)
 			assert.Equal(t, testcase.negotiated, apiclient.negotiated)
 		})
 	}
 }
 
+func TestExperimentalCli(t *testing.T) {
```
tiborvass: small nit: `TestExperimentalCLI` to stay consistent with casing

tiborvass:
LGTM

bitsofinfo:
thx

PR #769, opened 2017-12-28
Fix alignment of client built time in version output

leanderjanssen:
**- What I did**
Add one more tab to the `Built` time to align the output of the docker client version.

**- How to verify it**

Before:

```
$ docker version
Client:
 Version:	18.01.0-dev
 API version:	1.35
 Go version:	go1.9.2
 Git commit:	c97c6d6
 Built:	Wed Dec 27 20:11:19 2017
 OS/Arch:	linux/amd64
 Experimental:	false
```

After:

```
$ docker version
Client:
 Version:	18.01.0-dev
 API version:	1.35
 Go version:	go1.9.2
 Git commit:	0dd05ac1
 Built:		Thu Dec 28 09:19:35 2017
 OS/Arch:	linux/amd64
 Experimental:	false
```

Fix alignment of client built time in version output

Signed-off-by: Leander Janssen <leander@slaco.net>

### DIFF
--- a/cli/command/system/version.go
+++ b/cli/command/system/version.go
@@ -21,7 +21,7 @@ Client:{{if ne .Platform.Name ""}} {{.Platform.Name}}{{end}}
  API version:	{{.APIVersion}}{{if ne .APIVersion .DefaultAPIVersion}} (downgraded from {{.DefaultAPIVersion}}){{end}}
  Go version:	{{.GoVersion}}
  Git commit:	{{.GitCommit}}
- Built:	{{.BuildTime}}
+ Built:		{{.BuildTime}}
  OS/Arch:	{{.Os}}/{{.Arch}}
  Experimental:	{{.Experimental}}
 {{- end}}


leanderjanssen:
In GitHub this looks a bit weird, but in the actual code a single tab character is used.
Because the word `Built` is so short compared with the rest, two tabs are required to align it correctly.

thaJeztah:
The tabwriter is meant to convert the tabs to spaces during presentation, so that the columns are always aligned properly. There was a bug in the tabwriter though, that caused the alignment to not work properly, so it was temporarily disabled (IIRC).

I'm not exactly sure what the bug was, but @tiborvass mentioned that, so I guess he knows more. If that is resolved, alignment should work again (and this change would not be needed)

leanderjanssen:
Ah I see, I was actually quite surprised to see hard tabs in the code.

vdemeester:
ping @tiborvass

tiborvass:
Ugh, sorry for delay on this, thanks @leanderjanssen, but we need to use tabwriter here and I remember having some issues with it. If you want to tackle it let me know otherwise I'll try.

PR #771, opened 2017-12-28
Show a progress bar on system pruning

sterchelen:
Signed-off-by: Sterchele Nicolas <sterchele.nicolas@gmail.com>

**- What I did**
I've added a progress bar for system pruning.

**- How I did it**
-Using existing progress package to show the bar.
-Before pruning, I list all the object needed to delete
-While the pruning, I'm listening the events to show the progress.

**- How to verify it**
-Creating container, downloading images, network
-Run `docker system prune --all` and `docker system prune`.

Show a progress bar with the names of the deleted objects while system pruning.

### DIFF
--- a/cli/command/system/prune.go
+++ b/cli/command/system/prune.go
@@ -67,6 +67,12 @@ func runBuildCachePrune(dockerCli command.Cli, _ opts.FilterOpt) (uint64, string
 }
 
 func runPrune(dockerCli command.Cli, options pruneOptions) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	progressChan := pruneProgressChan{
+		errorChan: make(chan error),
+	}
+
 	// TODO version this once "until" filter is supported for volumes
 	if options.pruneVolumes && options.filter.Value().Include("until") {
 		return fmt.Errorf(`ERROR: The "until" filter is not supported with "--volumes"`)
@@ -77,6 +83,16 @@ func runPrune(dockerCli command.Cli, options pruneOptions) error {
 	if !options.force && !command.PromptForConfirmation(dockerCli.In(), dockerCli.Out(), confirmationMessage(options)) {
 		return nil
 	}
+
+	if err := startPruneProgress(dockerCli, &progressChan, options, ctx); err != nil {
+		return err
+	}
+
+	return runAllPrunes(dockerCli, options)
+}
+
+func runAllPrunes(dockerCli command.Cli, options pruneOptions) error {
+
 	imagePrune := func(dockerCli command.Cli, filter opts.FilterOpt) (uint64, string, error) {
 		return image.RunPrune(dockerCli, options.all, options.filter)
 	}
@@ -93,6 +109,7 @@ func runPrune(dockerCli command.Cli, options pruneOptions) error {
 	}
 
 	var spaceReclaimed uint64
+	var generalOutput bytes.Buffer
 	for _, pruneFn := range pruneFuncs {
 		spc, output, err := pruneFn(dockerCli, options.filter)
 		if err != nil {
@@ -100,11 +117,12 @@ func runPrune(dockerCli command.Cli, options pruneOptions) error {
 		}
 		spaceReclaimed += spc
 		if output != "" {
-			fmt.Fprintln(dockerCli.Out(), output)
+			generalOutput.WriteString(fmt.Sprintln(output))
 		}
 	}
 
-	fmt.Fprintln(dockerCli.Out(), "Total reclaimed space:", units.HumanSize(float64(spaceReclaimed)))
+	generalOutput.WriteString(fmt.Sprintln("Total reclaimed space:", units.HumanSize(float64(spaceReclaimed))))
+	fmt.Fprintln(dockerCli.Out(), generalOutput.String())
 
 	return nil
 }
--- a/cli/command/system/prune.go
+++ b/cli/command/system/prune.go
@@ -84,7 +84,7 @@ func runPrune(dockerCli command.Cli, options pruneOptions) error {
 		return nil
 	}
 
-	if err := startPruneProgress(dockerCli, &progressChan, options, ctx); err != nil {
+	if err := startPruneProgress(ctx, dockerCli, &progressChan, options); err != nil {
 		return err
 	}
 
--- a/cli/command/system/prune_progress.go
+++ b/cli/command/system/prune_progress.go
@@ -17,10 +17,10 @@ type pruneProgressChan struct {
 	errorChan chan error
 }
 
-func startPruneProgress(dockerCli command.Cli, progressChans *pruneProgressChan, options pruneOptions, ctx context.Context) error {
+func startPruneProgress(ctx context.Context, dockerCli command.Cli, progressChans *pruneProgressChan, options pruneOptions) error {
 	eventChan := make(chan string, 1)
-	monitorRemovingEvents(dockerCli, eventChan, progressChans, ctx)
-	objCount, err := countObjectsToRemove(dockerCli, options, ctx)
+	monitorRemovingEvents(ctx, dockerCli, eventChan, progressChans)
+	objCount, err := countObjectsToRemove(ctx, dockerCli, options)
 	if err != nil {
 		return err
 	}
@@ -61,7 +61,7 @@ func outputPruningProgress(dockerCli command.Cli, events chan string, numEvents 
 	}()
 }
 
-func countObjectsToRemove(dockerCli command.Cli, options pruneOptions, ctx context.Context) (int, error) {
+func countObjectsToRemove(ctx context.Context, dockerCli command.Cli, options pruneOptions) (int, error) {
 	volCount := 0
 	if options.pruneVolumes {
 		//Volumes listing call
@@ -140,7 +140,7 @@ func countObjectsToRemove(dockerCli command.Cli, options pruneOptions, ctx conte
 	return cntCount + volCount + imgCount + len(ntws), nil
 }
 
-func monitorRemovingEvents(dockerCli command.Cli, out chan<- string, progressChans *pruneProgressChan, ctx context.Context) {
+func monitorRemovingEvents(ctx context.Context, dockerCli command.Cli, out chan<- string, progressChans *pruneProgressChan) {
 	fArgs := filters.NewArgs(
 		filters.Arg("type", "container"),
 		filters.Arg("type", "volume"),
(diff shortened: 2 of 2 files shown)


cli/command/system/prune.go
```diff
@@ -67,6 +67,14 @@ func runBuildCachePrune(dockerCli command.Cli, _ opts.FilterOpt) (uint64, string
 }
 
 func runPrune(dockerCli command.Cli, options pruneOptions) error {
+
+	progressChan := pruneProgressChan{
+		errorChan:         make(chan error),
+		isEventsListening: make(chan bool),
+	}
+
+	go runPruneProgress(dockerCli, &progressChan, options)
```
dnephin: `runPruneProgress` starts a couple go routines, so it seems like it doesn't really need to be run in a goroutine as well. Would it be fine to run `countObjectsToRemove()` synchronously ?

cli/command/system/prune.go
```diff
@@ -92,19 +117,19 @@ func runPrune(dockerCli command.Cli, options pruneOptions) error {
 		pruneFuncs = append(pruneFuncs, runBuildCachePrune)
 	}
 
-	var spaceReclaimed uint64
```
dnephin: nit: I would rather keep the variable declaration immediately before the first use, instead of at the start of the function. Declaring everything at the top of a function makes the scope less obvious and makes it harder to refactor later.

cli/command/system/prune.go
```diff
@@ -92,19 +117,19 @@ func runPrune(dockerCli command.Cli, options pruneOptions) error {
...
 	for _, pruneFn := range pruneFuncs {
 		spc, output, err := pruneFn(dockerCli, options.filter)
 		if err != nil {
 			return err
 		}
 		spaceReclaimed += spc
 		if output != "" {
-			fmt.Fprintln(dockerCli.Out(), output)
+			generalOutput += fmt.Sprintln(output)
```
dnephin: Instead of string concatenation it is generally better (both faster, and more obvious) to use either a `bytes.Buffer{}` with `WriteString()` or a `[]string{}` with `strings.Join(outputs, "\n")`

cli/command/system/prune_progress.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func runPruneProgress(dockerCli command.Cli, progressChans *pruneProgressChan, options pruneOptions) {
+	eventChan := make(chan string, 1)
+	objCount, err := countObjectsToRemove(dockerCli, options)
+	if err != nil {
+		progressChans.errorChan <- err
+		return
+	}
+	go monitorRemovingEvents(dockerCli, eventChan, progressChans)
```
dnephin: I think monitoring needs to start before counting the objects. Otherwise anything that is created or removed after the count will mess with the progress.

cli/command/system/prune_progress.go
```diff
@@ -0,0 +1,173 @@
...
+	ntws, err := dockerCli.Client().NetworkList(ctx, ntwOpts)
+	if err != nil {
+		return 0, err
+	}
+
+	return cntCount + volCount + imgCount + len(ntws), nil
+}
+
+func monitorRemovingEvents(dockerCli command.Cli, out chan<- string, progressChans *pruneProgressChan) {
+	fArgs := filters.NewArgs()
```
dnephin: There's now a better way of creating these filters:

```go
filterArgs := filters.NewArgs(
  filters.Arg("type", "container"),
  filters.Arg("type", "volume"),
  filters.Arg("type", "network"),
  ...
```

cli/command/system/prune_progress.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}()
+
+	if err := jsonmessage.DisplayJSONMessagesToStream(pipeReader, dockerCli.Out(), nil); err != nil {
+		progressStruct.errorChan <- err
+		return
+	}
+}
+
+func countObjectsToRemove(dockerCli command.Cli, options pruneOptions) (int, error) {
```
dnephin: I need to look at this some more, but it seems unlikely that we can accurately count the objects that will be pruned. What happens if objects are created after this count runs, but before system prune starts pruning objects of that type?

cli/command/system/prune.go
sterchelen: Yep, corrected.

cli/command/system/prune.go
sterchelen: Corrected.

cli/command/system/prune.go
sterchelen: I used strings.Join as a correction.

cli/command/system/prune_progress.go
sterchelen: Well, we have a synchronization bit called `isEventsListening` that is used to authorize the pruning process when the monitoring is up and running.

cli/command/system/prune_progress.go
sterchelen: Unfortunately, we can't accurately predict the total number of objects to remove, in the case of an object created after the count runs.

sterchelen:
@dnephin Any news ?

cli/command/system/prune.go
```diff
@@ -67,6 +67,14 @@ func runBuildCachePrune(dockerCli command.Cli, _ opts.FilterOpt) (uint64, string
 }
 
 func runPrune(dockerCli command.Cli, options pruneOptions) error {
+
+	progressChan := pruneProgressChan{
+		errorChan:         make(chan error),
+		isEventsListening: make(chan bool),
+	}
+
+	runPruneProgress(dockerCli, &progressChan, options)
```
dnephin: I think this should be after the 3 checks below, so that 1) related code stays together, and 2) setting `options.pruneBuildCache = false` is read correctly (it isn't now).

Once it's moved to immediately before the select, I wonder if the `progressChan` could be moved into `runPruneProgress()` since it seems nothing else uses it:

```go
if err := startPruneProgress(dockerCli, options); err != nil {
    return err
}

return runAllPrunes(dockerCli, options)
```

cli/command/system/prune_progress.go
```diff
@@ -0,0 +1,175 @@
...
+
+func runPruneProgress(dockerCli command.Cli, progressChans *pruneProgressChan, options pruneOptions) {
+	eventChan := make(chan string, 1)
+	go monitorRemovingEvents(dockerCli, eventChan, progressChans)
+	objCount, err := countObjectsToRemove(dockerCli, options)
+	if err != nil {
+		progressChans.errorChan <- err
+		return
+	}
+	go outputPruningProgress(dockerCli, eventChan, objCount, progressChans)
```
dnephin: minor nit: maybe wrap the `jsonmessage.DisplayJSONMessagesToStream()` call in a `func()` and make that the goroutine instead of `outputPruningProgress` so that both are in the same place

cli/command/system/prune_progress.go
```diff
@@ -0,0 +1,175 @@
...
+				Current:    i,
+				Total:      int64(numEvents),
+				HideCounts: true,
+			}
+			if err := progressOut.WriteProgress(prog); err != nil {
+				progressStruct.errorChan <- err
+				return
+			}
+			i++
+		}
```
dnephin: doesn't this need to close the pipeWriter so that the reader goroutine will exit?

cli/command/system/prune_progress.go
```diff
@@ -0,0 +1,175 @@
...
+	progressOut := streamformatter.NewJSONProgressOutput(pipeWriter, false)
+
+	go func() {
+		var i int64 = 1
+		for s := range events {
+			prog := progress.Progress{
+				ID:         "Progress",
+				Action:     s,
+				Current:    i,
+				Total:      int64(numEvents),
```
dnephin: I need to check to see what happens if this receives more than `numEvents`,  which is possible in this case.

cli/command/system/prune_progress.go
```diff
@@ -0,0 +1,175 @@
...
+		//Delete only for image.
+		filters.Arg("event", "delete"),
+	)
+	eOpts := types.EventsOptions{
+		Since:   "",
+		Until:   "",
+		Filters: fArgs,
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
dnephin: This function should accept a `WithCancel` `ctx` as an argument so that the context can be shared with `runAllPrune` (also needs to be fixed in another PR). 

I think the `defer cancel()` should probably be in the `runPrune()` so that this goroutine and the output goroutines will exit if someone `ctrl-c` the main thread. I'm not sure that is working correctly right now, I'll need to try it out.

cli/command/system/prune_progress.go
```diff
@@ -0,0 +1,175 @@
...
+
+	for {
+		select {
+		case event := <-events:
+			out <- fmt.Sprintf("%s %s %s", event.Type, event.Action, event.Actor.ID)
+		case err := <-errors:
+			progressChans.errorChan <- err
+			return
+		}
+	}
```
dnephin: I think this needs to `close(out)` in both the error and non-error cases.

thaJeztah:
ping @sterchelen 🤗

sterchelen:
@thaJeztah yep... Sorry for the delay, I'm preparing the arrival of my baby :D.

I will work on it by the end of the week.

thaJeztah:
@sterchelen no worries, and _congratulations_!

sterchelen:
@dnephin @thaJeztah, please see my new commits :)

cpuguy83:
This would be nice, but I feel like this is inherently racey to do client side and also makes assumptions about what the daemon will actually be pruning.

At least to get rid of the assumptions a "dry run" prune mode to get the initial list could be added (this has been discussed before, just not implemented yet).
This still has race conditions.

thaJeztah:
If we're concerned about the race condition; one thing that would be a good improvement already is to print each object that's been pruned. I think currently, running `docker system prune` gives no feedback whatsoever, until it finishes, and then it prints the whole list of things that have been pruned.

sterchelen:
@vdemeester It's for the whole "pruning" command.
@thaJeztah Yep, agree ! Printing which objects that has been pruned. May I implement it ?

thaJeztah:
> May I implement it ?

Sure! Perhaps best as a new pull request, so that we still can discuss this one (with progress bars)

sterchelen:
@thaJeztah @vdemeester This PR is still relevant ?

tiborvass:
@sterchelen Thanks for this! I'd rather show what was actually deleted already and you may do that in this PR. Dry run is another story.

PR #772, opened 2017-12-30
Fix bash completion for "docker swarm"

muicoder:
"docker swarm" auto-completion is invalid.

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -4945,6 +4945,7 @@ _docker() {
 		secret
 		service
 		stack
+		swarm
 		system
 		volume
 	)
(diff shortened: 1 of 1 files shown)


albers:
Can you please explain why you want to remove support for `docker swarm` completion?

muicoder:
When I set DOCKER_HIDE_LEGACY_COMMANDS=true, swarm is not available, “swarm” is placed in legacy_commands, it should not be in management_commands?

albers:
Ah yes, you are right. `docker swarm` is a management command, not a legacy command.
Thanks very much for your fix.

albers:
@muicoder Please sign your commit.

PR #775, opened 2018-01-03
docker ps -N for verbose info: PID and namespace

Maksadbek:
**- What I did**
Listing init PID and namespaces easily by added -N flag to docker ps:
```
ubuntu@ubuntu-xenial:~/go/src/github.com/docker/cli/cmd/docker$ sudo ./docker ps -N
CONTAINER ID        NAME                PID                 PATH                CGROUP              IPC                 MNT                 NET                 USER                UTS
cf524fed5bbc        /sad_clarke         4026532138          nginx               4026531835          4026532137          4026532135          4026532140          4026531837          4026532136
```
I have no idea how to test it yet, as all info is gathered from symlinks.

Link to feature request in Moby project [32501](https://github.com/moby/moby/issues/32501)

**- How I did it**
I added a flag to ps command and created a formatter - namespaces.
All namespaces info is gathered from symlinks in the /proc/<pid>/ns/ directory by running readlink command(os.Readlink).
Accessing docker api to get PID of a container from its state property and caching this info to avoid accessing docker api multiple times.

**- How to verify it**
Verification can be done in Linux systems as information is gathered from filesystem: /proc/<pid>/ns. In addition, command must execution with root access:
1. Build the project.
2. Run: `sudo docker ps -N`

docker ps -N verbose info: PID and namespace

### DIFF
--- a/cli/command/formatter/namespaces.go
+++ b/cli/command/formatter/namespaces.go
@@ -1,0 +1,266 @@
+package formatter
+
+import (
+	"context"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+	"github.com/docker/docker/pkg/stringid"
+)
+
+const (
+	defaultNamespacesTableFormat = "table {{.ID}}\t{{.NAME}}\t{{.PID}}\t{{.PATH}}\t{{.CGROUP}}\t{{.IPC}}\t{{.MNT}}\t{{.NET}}\t{{.USER}}\t{{.UTS}}"
+	defaultNamespacesRawFormat   = "id: {{.ID}}\nname: {{.NAME}}\nPID: {{.PID}}\nPATH: {{.PATH}}\nCGROUP: {{.CGROUP}}\nIPC: {{.IPC}}\nMNT: {{.MNT}}\nNET: {{.NET}}\nUser: {{.USER}}\nUTS: {{.UTS}}"
+
+	pidHeader    = "PID"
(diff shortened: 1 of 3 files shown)


cli/command/formatter/namespaces.go
```diff
@@ -0,0 +1,266 @@
...
+	mnt, err := n.stat(c.State.Pid, "uts")
+	if err != nil {
+		return err.Error()
+	}
+
+	return mnt
+}
+
+func (n *namespacesContext) stat(pid int, ns string) (string, error) {
+	o, err := os.Readlink(n.procPath + "/" + strconv.Itoa(pid) + "/ns/" + ns)
```
dnephin: This will not work because it assumes the cli is running on the same host as the API, which is not always the case.

Maksadbek: Do you recommend to gather this info in API side ?

I assumed this feature is only for Linux platforms.
https://github.com/moby/moby/issues/32501#issuecomment-293359638:
_On non-Linux, it would report "This option only supported on Linux."_

cpuguy83:
I don't think this is the right place for such a feature (beyond that the actual implementation doesn't work here).
Gathering this data is very slow and providing it on a list view where there are many containers is undesirable.

We could discuss providing some platform specific details on the inspect API in moby/moby... and if we *really* wanted to we could make a format option to fetch this data on the list view here.

-1

cli/command/formatter/namespaces.go
thaJeztah: It's not the platform that's the problem, but it's the assumption that the CLI is running on the same host (what if the host is remote?)

Maksadbek:
Now I see the problem and I can change inspect API in moby/moby to provide this info.

PR #780, opened 2018-01-03
Fix --network-add adding duplicate networks

thaJeztah:
When adding a network using `docker service update --network-add`,
the new network was added by _name_.

Existing entries in a service spec are listed by network ID, which
resulted in the CLI not detecting duplicate entries for the same
network.

This patch changes the behavior to always use the network-ID,
so that duplicate entries are correctly caught.

Before this change;

```bash
$ docker network create -d overlay foo
$ docker service create --name=test --network=foo nginx:alpine
$ docker service update --network-add foo test
$ docker service inspect --format '{{ json .Spec.TaskTemplate.Networks}}' test
[
  {
    "Target": "9ot0ieagg5xv1gxd85m7y33eq"
  },
  {
    "Target": "9ot0ieagg5xv1gxd85m7y33eq"
  }
]
```

After this change:

```bash
$ docker network create -d overlay foo
$ docker service create --name=test --network=foo nginx:alpine
$ docker service update --network-add foo test
service is already attached to network 9ot0ieagg5xv1gxd85m7y33eq
```

```Markdown
* Fix duplicate networks being added with `docker service update --network-add` [docker/cli#780](https://github.com/docker/cli/pull/780)
```

### DIFF
--- a/cli/command/service/opts.go
+++ b/cli/command/service/opts.go
@@ -353,22 +353,21 @@ func (c *credentialSpecOpt) Value() *swarm.CredentialSpec {
 	return c.value
 }
 
-func convertNetworks(ctx context.Context, apiClient client.NetworkAPIClient, networks opts.NetworkOpt) ([]swarm.NetworkAttachmentConfig, error) {
+func resolveNetworkID(ctx context.Context, apiClient client.NetworkAPIClient, networkIDOrName string) (string, error) {
+	nw, err := apiClient.NetworkInspect(ctx, networkIDOrName, types.NetworkInspectOptions{Scope: "swarm"})
+	return nw.ID, err
+}
+
+func convertNetworks(networks opts.NetworkOpt) []swarm.NetworkAttachmentConfig {
 	var netAttach []swarm.NetworkAttachmentConfig
 	for _, net := range networks.Value() {
-		networkIDOrName := net.Target
-		_, err := apiClient.NetworkInspect(ctx, networkIDOrName, types.NetworkInspectOptions{Scope: "swarm"})
-		if err != nil {
-			return nil, err
-		}
 		netAttach = append(netAttach, swarm.NetworkAttachmentConfig{
 			Target:     net.Target,
 			Aliases:    net.Aliases,
 			DriverOpts: net.DriverOpts,
 		})
 	}
-	sort.Sort(byNetworkTarget(netAttach))
-	return netAttach, nil
+	return netAttach
 }
 
 type endpointOptions struct {
@@ -590,10 +589,15 @@ func (options *serviceOptions) ToService(ctx context.Context, apiClient client.N
 		return service, err
 	}
 
-	networks, err := convertNetworks(ctx, apiClient, options.networks)
-	if err != nil {
-		return service, err
+	networks := convertNetworks(options.networks)
+	for i, net := range networks {
+		nwID, err := resolveNetworkID(ctx, apiClient, net.Target)
+		if err != nil {
+			return service, err
+		}
+		networks[i].Target = nwID
 	}
+	sort.Sort(byNetworkTarget(networks))
 
 	resources, err := options.resources.ToResourceRequirements()
 	if err != nil {
--- a/cli/command/service/client_test.go
+++ b/cli/command/service/client_test.go
@@ -16,6 +16,7 @@ type fakeClient struct {
 	serviceListFunc           func(context.Context, types.ServiceListOptions) ([]swarm.Service, error)
 	taskListFunc              func(context.Context, types.TaskListOptions) ([]swarm.Task, error)
 	infoFunc                  func(ctx context.Context) (types.Info, error)
+	networkInspectFunc        func(ctx context.Context, networkID string, options types.NetworkInspectOptions) (types.NetworkResource, error)
 }
 
 func (f *fakeClient) NodeList(ctx context.Context, options types.NodeListOptions) ([]swarm.Node, error) {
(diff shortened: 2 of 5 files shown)


thaJeztah:
Didn't have a look yet at creating a unit-test for this (if we want one)

dnephin:
hmm, not sure how this works. Isn't the problem in `cli/command/service/update.go` `updateNetworks()`, where it compares `network.Target` to the id?

thaJeztah:
Yeah, basically `updateNetworks` takes;

- existing networks
- networks set on `--network-rm`
- networks set on `--network-add`

Each network set by `--network-rm` / `--network-add` is looked up, and network name is translated to network ID

However, that translation was not done for `--network-add` (also not on `docker service create`, but that's not problematic; the translation is done server-side in that case).

So on network create;

```json
{
  "EndpointSpec": {
    "Mode": "vip"
  },
  "Labels": {},
  "Mode": {
    "Replicated": {}
  },
  "Name": "bla",
  "TaskTemplate": {
    "ContainerSpec": {
      "DNSConfig": {},
      "Image": "nginx:alpine@sha256:40e95eef5082a5b26c5fd9441bd7ee6bcda1bb5f9fbf7a4a1ef9b2b0f88d5c43"
    },
    "ForceUpdate": 0,
    "Networks": [
      {
        "Target": "foo"
      }
    ],
    "Placement": {
      "Platforms": [
        {
          "Architecture": "amd64",
          "OS": "linux"
        }
      ]
    },
    "Resources": {
      "Limits": {},
      "Reservations": {}
    }
  }
}
```

Then, on _update_, the spec is fetched from the daemon (which returns network ID's), but the added network is added to the spec by _name_, so this is sent on update:

```json
{
  "EndpointSpec": {
    "Mode": "vip"
  },
  "Labels": {},
  "Mode": {
    "Replicated": {
      "Replicas": 1
    }
  },
  "Name": "bla",
  "TaskTemplate": {
    "ContainerSpec": {
      "DNSConfig": {},
      "Image": "nginx:alpine@sha256:40e95eef5082a5b26c5fd9441bd7ee6bcda1bb5f9fbf7a4a1ef9b2b0f88d5c43"
    },
    "ForceUpdate": 0,
    "Networks": [
      {
        "Target": "foo"
      },
      {
        "Target": "x8bnlf8np4o8gk8nnh0ilg6oo"
      }
    ],
    "Placement": {
      "Platforms": [
        {
          "Architecture": "amd64",
          "OS": "linux"
        }
      ]
    },
    "Resources": {
      "Limits": {},
      "Reservations": {}
    },
    "Runtime": "container"
  }
}
```

Where `foo` is the name for the network with ID `x8bnlf8np4o8gk8nnh0ilg6oo `

dnephin:
Ok, but the code that is changed here is for converting opts to the swarn spec, not specifically for update, right? Isn't the update logic  in cli/command/service/update.go updateNetworks() ?

thaJeztah:
Well, `updateNetworks()` calls this function (but _only_ for the `--network-add` networks);

For `--network-rm` this is called: https://github.com/docker/cli/blob/97b148b440f052679bd80cf2779801494c428afe/cli/command/service/update.go#L1101-L1107

Existing networks (in the spec) already have an ID (no name); https://github.com/docker/cli/blob/97b148b440f052679bd80cf2779801494c428afe/cli/command/service/update.go#L1111-L1118

For `--network-add`, all networks are looked up first, and converted, then added to the list; https://github.com/docker/cli/blob/97b148b440f052679bd80cf2779801494c428afe/cli/command/service/update.go#L1122-L1126

So this change will address both `docker service create` and `docker service update` to consistently use a network-ID

Alternatively, we move the `apiClient.NetworkInspect()` call out of `convertNetworks()` because it's not really related to the conversion itself, and currently only used to validate that a network exists

dnephin:
It would be great to remove `NetworkInspect` from `convertNetworks()` although I guess it's still needed by the other caller.

Thanks, I understand the change now. I missed that this was being called from `updateNetworks`.

thaJeztah:
> It would be great to remove NetworkInspect from convertNetworks() although I guess it's still needed by the other caller.

I agree; looking at this again, I think moving the `NetworkInspect` outside of that function may be the better solution (I want to look at the sorting though; not even sure if that should be in there, because it will sort on _name_ without my change, but after processing, it's sorted _again_ but this time on ID.

Given that this didn't make it for 18.01-rc1; let me change this to `WIP`, then;

- I'll move the `NetworkInspect`
- I'll write some unit tests (probably with a simple fakeClient)


I confirmed this bug already being in 17.09, so it's not new, so not super-urgent (and containers get only a single network assigned, so no duplicates there)

thaJeztah:
@dnephin updated with the changes discussed, PTAL

cli/command/service/opts.go
```diff
@@ -353,22 +353,24 @@ func (c *credentialSpecOpt) Value() *swarm.CredentialSpec {
...
-		if err != nil {
-			return nil, err
-		}
 		netAttach = append(netAttach, swarm.NetworkAttachmentConfig{
 			Target:     net.Target,
 			Aliases:    net.Aliases,
 			DriverOpts: net.DriverOpts,
 		})
 	}
-	sort.Sort(byNetworkTarget(netAttach))
```
thaJeztah: Moved the sorting out of here as well, otherwise service _create_ would sort by network name, and service _update_ sorted by network _ID_

cli/command/service/opts.go
```diff
@@ -353,22 +353,24 @@ func (c *credentialSpecOpt) Value() *swarm.CredentialSpec {
 	return c.value
 }
 
-func convertNetworks(ctx context.Context, apiClient client.NetworkAPIClient, networks opts.NetworkOpt) ([]swarm.NetworkAttachmentConfig, error) {
+func resolveNetworkID(ctx context.Context, apiClient client.NetworkAPIClient, networkIDOrName string) (string, error) {
+	nw, err := apiClient.NetworkInspect(ctx, networkIDOrName, types.NetworkInspectOptions{Scope: "swarm"})
+	if err != nil {
+		return "", err
+	}
+	return nw.ID, nil
```
dnephin: nit: the if is unnecessary since `NetworkIsnpect`  returns a struct not a pointer. It can be replaced by 

```go
return nw.ID, err
```

thaJeztah: Oh, nice one! I'll add that change

thaJeztah: done :)

thaJeztah:
ping @vdemeester PTAL

PR #784, opened 2018-01-04
Correct table html with close tag for </thead>

kizzie:
Currently shows as:

```
"The options you can specify are:
</table> "
```

on the docker website - missing a close tag for thead (to be honest https://docs.docker.com/engine/reference/commandline/service_create/#publish-service-ports-externally-to-the-swarm--p-publish is a bit of a mess at the moment)

(my user pic is one of my bunnies if that helps?)

### DIFF
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -764,6 +764,7 @@ The options you can specify are:
   <th>Long syntax</th>
   <th>Description</th>
 </tr>
+</thead>
 <tr>
   <td>published and target port </td>
   <td><tt></tt></td>


docs/reference/commandline/service_create.md
```diff
@@ -797,6 +798,7 @@ The options you can specify are:
     port for both protocols, specify the `-p` or `--publish` flag twice.
   </p></td>
 </tr>
+
```
thaJeztah: Can you remove this line change and amend your commit?

kizzie: Some swearing at text editors adding and removing spaces from ends of lines later - all done

thaJeztah: Oooh, sorry for that, thanks so much for fixing!

PR #785, opened 2018-01-04
use 'config' over 'configuration file'

friism:
**- What I did**

The help text for `config` commands is not consistent about whether the object being modified is called a "config" or a "configuration file". I changed it to be "config" for all commands since that's shorter.

Previously:

```
Commands:
  create      Create a configuration file from a file or STDIN as content
  inspect     Display detailed information on one or more configuration files
  ls          List configs
  rm          Remove one or more configuration files
```

**- How to verify it**

`docker config --help`

Descriptions for config commands now use the same name when describing configs.

### DIFF
--- a/cli/command/config/create.go
+++ b/cli/command/config/create.go
@@ -28,7 +28,7 @@ func newConfigCreateCommand(dockerCli command.Cli) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "create [OPTIONS] CONFIG file|-",
-		Short: "Create a configuration file from a file or STDIN as content",
+		Short: "Create a config from a file or STDIN",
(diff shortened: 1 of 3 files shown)


cli/command/config/create.go
```diff
@@ -28,7 +28,7 @@ func newConfigCreateCommand(dockerCli command.Cli) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "create [OPTIONS] CONFIG file|-",
-		Short: "Create a configuration file from a file or STDIN as content",
+		Short: "Create config using a file or STDIN as content",
```
dnephin: How about:

"Create a config from a file or STDIN"

friism:
squashed, thanks guys

PR #796, opened 2018-01-10
Added description about PIDS count.

ieugen:


docs/reference/commandline/stats.md
```diff
@@ -36,6 +36,8 @@ If you want more detailed information about a container's resource usage, use th
 
 > **Note**: On Linux, the Docker CLI reports memory usage by subtracting page cache usage from the total memory usage. The API does not perform such a calculation but rather provides the total memory usage and the amount from the page cache so that clients can use the data as needed.
 
+> **Note**: PIDS count contain the processes and kernel threads created by that process. Depending on OS threads are can be called light-weight process, kernel tasks, etc. A large number of PIDS and a small number processes (reported by `ps` or `top`) means that one process is creating lots of threads.
```
cpuguy83: `PIDS count contains`

`Depending on the OS, threads can be`

ieugen:
@cpuguy83 : done.

docs/reference/commandline/stats.md
```diff
@@ -36,6 +36,8 @@ If you want more detailed information about a container's resource usage, use th
 
 > **Note**: On Linux, the Docker CLI reports memory usage by subtracting page cache usage from the total memory usage. The API does not perform such a calculation but rather provides the total memory usage and the amount from the page cache so that clients can use the data as needed.
 
+> **Note**: PIDS count contains the processes and kernel threads created by that process. Depending on OS, threads can be called light-weight process, kernel tasks, etc. A large number of PIDS and a small number processes (reported by `ps` or `top`) means that one process is creating lots of threads.
```
cpuguy83: `... and kernel threads created by that container`

cpuguy83: `... may mean that something in the container is creating many threads`

cpuguy83: One more note... we may want to generalize this a bit more.

> PIDS count shows the number of processes running in the container. Depending on the platform, this number can mean different things. On Linux, the number refers to the number of "tasks" running in the container where a "task" can be a process or even a thread. A large PIDs count with a small number of processes may mean that are large number of threads are running.

ieugen: > may mean that are large number of threads are running.
I think a better wording would be : "may mean that there is a large number of threads running". 

I'm ok with your proposal. Make the change if you like. Anything is better than nothing and we can go around discussing a lot with very little extra value. I think both fragments convey the idea. +1 for merging.

mistyhacks: Is it a column? If so, it should be `PIDS count` or "the `PIDS count` column".

Maybe some rewording:

"Thread" is the term used by Linux. Other equivalent terms are "lightweight process" or "kernel task".
(this seems like trivia and I'm not sure it even needs to be stated)

A large number of PIDs combined with a small number of total processes may indicate that a single process is spawning a large number of threads.

thaJeztah:
ping @ieugen could you address the review comments?

PR #802, opened 2018-01-11
Fix docs YAML generation setting "kubernetes" for "swarm"

thaJeztah:
Due to a copy/paste error, commands annotated with "swarm" were incorrectly setting the "kubernetes" property.

This was introduced in https://github.com/docker/cli/pull/721/commits/f1b116179f2a95d0ea45780dfb8be51c2825b9c0 (https://github.com/docker/cli/pull/721)

```Markdown
* Fix wrong orchestrator attribute being set in docs YAML [docker/cli#802](https://github.com/docker/cli/pull/802)
```

ping @silvin-lubecki @mistyhacks @dnephin PTAL

### DIFF
--- a/docs/yaml/yaml.go
+++ b/docs/yaml/yaml.go
@@ -123,7 +123,7 @@ func GenYamlCustom(cmd *cobra.Command, w io.Writer) error {
 			cliDoc.Kubernetes = true
 		}
 		if _, ok := curr.Annotations["swarm"]; ok && !cliDoc.Swarm {
-			cliDoc.Kubernetes = true
+			cliDoc.Swarm = true
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


docs/yaml/yaml.go
```diff
@@ -123,7 +123,7 @@ func GenYamlCustom(cmd *cobra.Command, w io.Writer) error {
 			cliDoc.Kubernetes = true
 		}
 		if _, ok := curr.Annotations["swarm"]; ok && !cliDoc.Swarm {
-			cliDoc.Kubernetes = true
+			cliDoc.Swarm = true
```
dnephin: Something still looks weird here.

It shouldn't need to check `!cliDoc.Swarm` because if it's true, setting it to true again doesn't hurt. Is this just an unnecessary statement that can be removed?

I guess all the checks above are the same, so maybe it was just copied?

thaJeztah: Yes, was looking at that as well, but didn't want to change without going through the whole logic. I can look at it for a follow-up 👍

thaJeztah:
author LGTM, let's merge this 👍

mistyhacks:
@andrewhsu this is needed in 18.01 `docker/docker-ce` ASAP.

silvin-lubecki:
Sorry guys... 😕 
I was a little bit on "cargo cult" on this part.

PR #803, opened 2018-01-11
Remove aliases for orchestrator

thaJeztah:
Prefer "strict" values for orchestrator, as it's easier to add aliases (if we think it's needed) than to remove them later.

I commented on this in my review on https://github.com/docker/cli/pull/721#discussion_r158812964, but looks like we forgot to actually make that change.

### DIFF
--- a/cli/command/orchestrator.go
+++ b/cli/command/orchestrator.go
@@ -21,9 +21,9 @@ const (
 
 func normalize(flag string) Orchestrator {
 	switch flag {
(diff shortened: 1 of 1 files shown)


thaJeztah:
No worries; I thought that was the case; also was a bit in doubt between using `swarm` or `swarmkit` (but our annotations used `swarm` so I went for that - happy to hear if we think `swarmkit` is better)

PR #804, opened 2018-01-11
Annotate "stack" commands to be "swarm" and "kubernetes"

thaJeztah:
The `stack` commands require either "swarm" or "kubernetes"

ping @dnephin @vdemeester @mistyhacks

### DIFF
--- a/cli/command/stack/cmd.go
+++ b/cli/command/stack/cmd.go
@@ -9,11 +9,15 @@ import (
 // NewStackCommand returns a cobra command for `stack` subcommands
 func NewStackCommand(dockerCli command.Cli) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:         "stack",
-		Short:       "Manage Docker stacks",
-		Args:        cli.NoArgs,
-		RunE:        command.ShowHelp(dockerCli.Err()),
-		Annotations: map[string]string{"version": "1.25"},
+		Use:   "stack",
+		Short: "Manage Docker stacks",
+		Args:  cli.NoArgs,
+		RunE:  command.ShowHelp(dockerCli.Err()),
+		Annotations: map[string]string{
+			"kubernetes": "",
+			"swarm":      "",
+			"version":    "1.25",
+		},
 	}
 	cmd.AddCommand(
 		newDeployCommand(dockerCli),
@@ -37,6 +41,10 @@ func NewTopLevelDeployCommand(dockerCli command.Cli) *cobra.Command {
 	cmd := newDeployCommand(dockerCli)
 	// Remove the aliases at the top level
 	cmd.Aliases = []string{}
-	cmd.Annotations = map[string]string{"experimental": "", "version": "1.25"}
+	cmd.Annotations = map[string]string{
+		"experimental": "",
+		"swarm":        "",
+		"version":      "1.25",
+	}
 	return cmd
 }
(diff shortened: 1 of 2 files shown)


cli/command/stack/cmd.go
```diff
@@ -37,6 +41,10 @@ func NewTopLevelDeployCommand(dockerCli command.Cli) *cobra.Command {
 	cmd := newDeployCommand(dockerCli)
 	// Remove the aliases at the top level
 	cmd.Aliases = []string{}
-	cmd.Annotations = map[string]string{"experimental": "", "version": "1.25"}
+	cmd.Annotations = map[string]string{
+		"experimental": "",
+		"swarm":        "",
```
thaJeztah: I wasn't sure if I should add "kubernetes" for the top-level `docker deploy`; it's still marked "experimental" (and I think we should consider removing it)

vdemeester: Yes, I think we should remove the top level one (and thus we don't need to annotate it)

thaJeztah:
ping @dnephin okay to merge?

dnephin:
there is a test failure, let me try to rerun

dnephin:
@thaJeztah the e2e test seems to be failing consistently. Can you take a look? I think it's a real failure, doesn't look like a flake.

dnephin:
I think the logic for disabling kubernetes doesn't account for swarm also being set ?

thaJeztah:
oh, interesting one; I'd have to check where/how it's used there 😓

vdemeester:
@dnephin @thaJeztah right.. the code that checks for swarm/kubernetes is pretty naive and doesn't handle something that would support both (i.e. it make the assumption that it would have none of both…). That needs to change 😅

vdemeester:
```diff
diff --git a/cmd/docker/docker.go b/cmd/docker/docker.go
index 74d8447f..33f8765b 100644
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -277,10 +277,12 @@ func areFlagsSupported(cmd *cobra.Command, details versionDetails) error {
                        if _, ok := f.Annotations["experimentalCLI"]; ok && !hasExperimentalCLI {
                                errs = append(errs, fmt.Sprintf("\"--%s\" is only supported when experimental cli features are enabled", f.Name))
                        }
-                       if _, ok := f.Annotations["kubernetes"]; ok && !hasKubernetes {
+                       _, isKubernetesAnnotated := f.Annotations["kubernetes"]
+                       _, isSwarmAnnotated := f.Annotations["swarm"]
+                       if isKubernetesAnnotated && !isSwarmAnnotated && !hasKubernetes {
                                errs = append(errs, fmt.Sprintf("\"--%s\" is only supported on a Docker cli with kubernetes features enabled", f.Name))
                        }
-                       if _, ok := f.Annotations["swarm"]; ok && hasKubernetes {
+                       if isSwarmAnnotated && !isKubernetesAnnotated && hasKubernetes {
                                errs = append(errs, fmt.Sprintf("\"--%s\" is only supported on a Docker cli with swarm features enabled", f.Name))
                        }
                }
```

Something like the above should fix it 👼

thaJeztah:
Thanks! Added that change 👍

thaJeztah:
oh, we may need something similar for `areSubcommandsSupported()`, let me check if that's indeed the case

thaJeztah:
We should rewrite that logic a bit; it's quite confusing because `!hasKubernetes` means `no orchestrator set` means `use default` means `swarm`; we can do so in a follow up, as it may require some changes in other locations

vdemeester:
ping @dnephin 👼

Issue #698, opened 2017-11-16
Support `tmpfs-size` and `tmpfs-mode` in compose-file


luochen1990:
@thaJeztah suggests to use a `--mount` option as [an workaround of `--shm-size`](https://github.com/moby/moby/issues/26714) like that:

```
docker service create \
  --name tmpfstest \
  --mount type=tmpfs,dst=/dev/shm,tmpfs-size=1000000000 \
  --tty \
  debian:stretch-slim cat
```

But I can't find anything about how to set the `tmpfs-size` in the [doc about volume-long-syntax](https://docs.docker.com/compose/compose-file/#long-syntax-3). I think this feature should be supported.

thaJeztah:
Thanks for opening this!

/cc @vdemeester @dnephin

glyif:
What are you peeps thinking the docker-compose syntax should look like?

Something like this?

```
volumes:
    - type: tmpfs
       size: 10000
```

dnephin:
I think it should match the existing types (bind, volume):

```yaml
volumes:
 - type: tmpfs
   target: /tmp
   tmpfs:
     size: 10000

```

ypresto:
Looks `tmpfs-mode` one is not implemented yet. Only `tmpfs: size: ...` is implemented.

mterron:
Actually tmpfs supports all (any?) mount options, it's just that the documentation is incorrect.
This actually does what it's supposed to:
```yaml
version: "2.4"
services:
  my_app:
    image: my_app
    read_only: true
    restart: always
    tmpfs:
      - /run:mode=770,size=1k,uid=2000,gid=10000
`
It'll set /run mode to 770 with a 1k size limit and owned by uid 200 and gid 10000

alisonjr:
> I think it should match the existing types (bind, volume):
> 
> ```yaml
> volumes:
>  - type: tmpfs
>    target: /tmp
>    tmpfs:
>      size: 10000
> ```

Is this configuration valid?

PR #808, opened 2018-01-14
added support for tmpfs-mode in compose file

ethan-haynes:
Provides a fix for https://github.com/docker/cli/issues/698

**- What I did**
  Added the ability to use a tmpfs volume type in compose with the optional parameter of size. 
**- How I did it**
  Changed compose schemas 3.2-3.5 to allow for the type. Tweaked volumes parser to read it. 
  Removed it from unapproved params. 
**- How to verify it**

Test it with compose:

```yaml
version: "3.6"
services:
  tmpfs:
    image: nginx:latest
    volumes:
      - type: tmpfs
        target: /app
        tmpfs:
          size: 10000
```
```
$ docker stack deploy -c swarm-compose.yml nginx
```


thaJeztah:
design looks good to me as well; moving to code review

cli/compose/convert/volume.go
```diff
@@ -88,3 +67,77 @@ func convertVolumeToMount(
 	// Named volumes
 	return result, nil
 }
+
+func handleBindToMount(
+	volume composetypes.ServiceVolumeConfig,
+	result mount.Mount,
+) (mount.Mount, error) {
+	if volume.Volume != nil {
+		return result, errors.New("volume options are incompatible with type bind")
```
thaJeztah: On error-situations, we should not return a `result` with values; the result should be "empty" Mount, e.g.;

```go
return mount.Mount{}, errors.New("volume options are incompatible with type bind")
```

but also see my other comment on moving the `result` mount outside of these functions; it's likely better to construct the result at the end of each `handleXX` function

cli/compose/convert/volume.go
```diff
@@ -88,3 +67,77 @@ func convertVolumeToMount(
...
+	}
+
+	// Anonymous volumes
+	if volume.Source == "" && volume.Type != "tmpfs" {
+		return result, nil
+	}
+
+	switch volume.Type {
+	case "volume":
+		mount, err := handleVolumeToMount(volume, stackVolumes, namespace, result)
```
thaJeztah: no need for the intermediate variables here; this can just be;

```go
return handleVolumeToMount(volume, stackVolumes, namespace, result)
```

cli/compose/convert/volume.go
```diff
@@ -88,3 +67,77 @@ func convertVolumeToMount(
...
+		return mount, err
+	case "bind":
+		mount, err := handleBindToMount(volume, result)
```
thaJeztah: same here;

```go
return handleBindToMount(volume, result)
```

cli/compose/convert/volume.go
```diff
@@ -88,3 +67,77 @@ func convertVolumeToMount(
...
+		return mount, err
+	case "tmpfs":
+		mount, err := handleTmpfsToMount(volume, result)
```
thaJeztah: and here;

```go
return handleTmpfsToMount(volume, result)
```

cli/compose/convert/volume.go
```diff
@@ -88,3 +67,77 @@ func convertVolumeToMount(
...
+		return mount, err
+	}
+	return result, nil
```
thaJeztah: This doesn't look right: unknown `volume.Type` should produce an error, and return an empty result

cli/compose/convert/volume.go
```diff
@@ -88,3 +67,77 @@ func convertVolumeToMount(
...
+	// Tmpfs volumes
+	return result, nil
+}
+
+func convertVolumeToMount(
+	volume composetypes.ServiceVolumeConfig,
+	stackVolumes volumes,
+	namespace Namespace,
+) (mount.Mount, error) {
+	result := mount.Mount{
```
thaJeztah: I'm not a big fan of moving this outside of the `handleXX` functions; I assume you kept it out to reduce code duplication, but  now the conversion (and validation) is "split" between the `handleXX` functions and this function.

Given that not all mount-types use the same options in this type (e.g., `type=bind` uses `Source`, but `type=tmpfs` doesn't), I'd move this code to each `handleXX` function. If duplication of code is a concern (but I think it's minimal), you _could_ have a helper function for the shared code

cli/compose/convert/volume.go
```diff
@@ -88,3 +67,77 @@ func convertVolumeToMount(
...
+	}
+	if volume.Tmpfs != nil {
+		return result, errors.New("tmpfs options are incompatible with type bind")
+	}
+	if volume.Bind != nil {
+		result.BindOptions = &mount.BindOptions{
+			Propagation: mount.Propagation(volume.Bind.Propagation),
+		}
+	}
+	// Binds volumes
```
thaJeztah: This comment doesn't add much, so I'd remove it

cli/compose/convert/volume.go
```diff
@@ -88,3 +67,77 @@ func convertVolumeToMount(
...
+	}
+	if volume.Volume != nil {
+		return result, errors.New("volume options are incompatible with type tmpfs")
+	}
+	if volume.Tmpfs != nil {
+		result.TmpfsOptions = &mount.TmpfsOptions{
+			SizeBytes: volume.Tmpfs.Size,
+		}
+	}
+	// Tmpfs volumes
```
thaJeztah: This comment doesn't add much, so I'd remove it; if the function needs documentation, add a GoDoc for the function itself

cli/compose/schema/data/config_schema_v3.5.json
```diff
@@ -300,6 +300,12 @@
                     "properties": {
                       "nocopy": {"type": "boolean"}
                     }
+                  },
+                  "tmpfs": {
+                    "type": "object",
+                    "properties": {
+                      "size": {"type": "number"}
```
thaJeztah: should this be `"type": "integer"` ?

If only positive integers are allowed, likely:

```json
{
  "type": "integer",
  "exclusiveMinimum": 0
}
```

(not sure if `exclusiveMinimum` or `minimum`; likely depends on how empty values are handled)

cli/compose/types/types.go
```diff
@@ -296,6 +296,11 @@ type ServiceVolumeVolume struct {
 	NoCopy bool `mapstructure:"nocopy"`
 }
 
+// ServiceVolumeTmpfs are options for a service volume of type tmpfs
+type ServiceVolumeTmpfs struct {
+	Size int64
```
thaJeztah: This should probably be an `uint64`? Or is a signed integer used elsewhere?

ethan-haynes: Chose to use int64 over unit64 because the mount struct looked to expect TmpfsOptions to contain it. Source:
https://github.com/moby/moby/blob/6415f1dcf5c79da853c604733c56b43ae3d18ced/api/types/mount/mount.go#L108

ethan-haynes:
Thanks for the helpful feedback. I will make the necessary changes and update the PR.

cli/compose/types/types.go
thaJeztah: Thanks! Wonder if that was for a reason ("special `-1` value, e.g.) or just overlooked during review of the initial implementation. 🤔

cli/compose/schema/data/config_schema_v3.5.json
ethan-haynes: In the Docker [documentation](https://docs.docker.com/engine/admin/volumes/tmpfs/#specify-tmpfs-options) if tmpfs-size isn't provided it will be unlimited by default:

Option | Description
-- | --
tmpfs-size | Size of the tmpfs mount in bytes. Unlimited by default.

So if the user wants to specify a ceiling of how many bytes to allocate, then they should provide a positive integer above 0. The docker cli, as it behaves now, will not allow negative numbers:
```
invalid argument "type=tmpfs,destination=/app,tmpfs-size=-1" for --mount: invalid value for tmpfs-size: -1
```
It will allow zero, but when you inspect the container it just defaults TmpfsOptions to an empty object, which (since tmpfs-size won't actually be set) will allocate "unlimited" resources: 
```JSON
"Mounts": [
                        {
                            "Type": "tmpfs",
                            "Target": "/app",
                            "TmpfsOptions": {}
                        }
                    ]
```

So I think we should, as you said, use an exclusiveMinimum of zero. That way the user at least attaches the TmpfsOptions with a value for tmpfs-size.

thaJeztah: If the CLI uses zero as equivalent of "not set", then probably `minimum` should be used (zero is  a valid value, but indicates "no limit").

ethan-haynes: That makes sense. I will Git started on those changes :octocat:

ethan-haynes:
I believe that I addressed most of the concerns that @thaJeztah brought up in my latest PR, with the exception being the issue regarding the directory of the compose file being set when a source is not passed to a volume of type bind (using the long compose syntax). I added explicit checks for an empty source in the convert package, but by the time it is received there it is already set. I am happy to alter/fix the behavior in this PR if it is decided that the current functionality is not desired. Or we/I can make a separate issue for it and address it in a different PR. Thanks! 😄

dnephin:
Not sure why jenkins didn't run the integration suite.

cli/compose/convert/volume.go
```diff
@@ -85,6 +82,62 @@ func convertVolumeToMount(
...
+}
+
+func convertVolumeToMount(
+	volume composetypes.ServiceVolumeConfig,
+	stackVolumes volumes,
+	namespace Namespace,
+) (mount.Mount, error) {
+
+	switch volume.Type {
+	case "volume":
```
dnephin: I think the old default (when type is not set) would be volume, so I guess this case should also have `""`

cli/compose/convert/volume.go
```diff
@@ -22,43 +22,40 @@ func Volumes(serviceVolumes []composetypes.ServiceVolumeConfig, stackVolumes vol
...
-	if volume.Source == "" {
-		return result, nil
-	}
-	if volume.Type == "volume" && volume.Bind != nil {
-		return result, errors.New("bind options are incompatible with type volume")
-	}
-	if volume.Type == "bind" && volume.Volume != nil {
-		return result, errors.New("volume options are incompatible with type bind")
+	if volume.Type != "tmpfs" {
+		result.Source = volume.Source
```
dnephin: This isn't really necessary because `handleTmpfsToMount()` will error out in this case. I would remove this so we don't have to check it in two places.

thaJeztah:
ping @shin- for Docker Compose

thaJeztah:
Oh! before I forget; @ethan-haynes could you also take care of updating the compose-file documentation? Possibly needs to go in the `vnext-compose` branch; https://github.com/docker/docker.github.io/tree/vnext-compose/compose but @mistyhacks may be able to give guidance there 😊

mistyhacks:
Yes, that's correct if it applies to a not-yet-released feature of Compose.

ethan-haynes:
Sounds good. I'll go ahead and update the documentation with new examples for the 3.6v compose release. 📝

PR #809, opened 2018-01-14
Add info for events backlog and scope

### LINKED ISSUE #727: Add Docker Events notes about limited backlog of events and event scope.
@BretFisher commented on [Sat Dec 02 2017](https://github.com/docker/docker.github.io/issues/5432)

[engine/reference/commandline/events.md](https://docs.docker.com/engine/reference/commandline/events/), CC @mistyhacks

1) Great doc on the events command but I don't see anything to educate the user on the backlog limits of the events cli/api.  With all the nice date-based filter values, the reader might think it would be a file written to disk for long-term storage like the dockerd log.  I don't think that's the case, but I'm coming up with little info on how much is exactly stored.  I see a @cpuguy83 [mention here of 1000 event backlog maximum](https://github.com/moby/moby/issues/29548#issuecomment-268329529) (I assume per node). If that's true, it would be great to mention it here.
2) Second, only after using them a while did I realize that Swarm events (services/nodes/secrets/configs) are seen on managers only, and any other events are only from the local server. I know there's a filter shown for `local|swarm` but maybe it would improve visibility of this big difference to mention the two types in description, with something like "different event types have different scopes, local scoped events are only seen on the node they take place on, and swarm scoped events are seen on all managers".

If you think this is good additional info, I'm happy to submit PR but wanted to get confirmation and feedback first. Maybe these two should just be added to the [top description](https://github.com/docker/cli/blob/master/docs/reference/commandline/events.md#events)?

BretFisher:
Fixes #727 

**- What I did**

1. Add `docker events` description info on the two scope types of events.
2. Add `docker events` note in two places about backlog limit of event log.

**- How I did it**
After testing and using the events feature and discovering these undocumented things, I edited some markdown lines with my bare hands :)

**- How to verify it**
Like @joaofnfernandes mentioned in https://github.com/docker/docker.github.io/issues/5432#issuecomment-349084294 we may need @cpuguy83 and/or @thaJeztah to validate the 1000 message buffer limit.

Added documentation on docker events scope and buffer limits.

giphy

Signed-off-by: Bret Fisher <bret@bretfisher.com>

### DIFF
--- a/docs/reference/commandline/events.md
+++ b/docs/reference/commandline/events.md
@@ -31,7 +31,12 @@ Options:
 ## Description
 
 Use `docker events` to get real-time events from the server. These events differ
-per Docker object type.
+per Docker object type. Different event types have different scopes. The local 
+scoped events are only seen on the node they take place on, and swarm scoped 
+events are seen on all managers.
+
+Note that there is a limited backlog of events you can filter. `docker events` is 
+a rolling log, and has at most 1000 events at a time.
 
 ### Object types
 
@@ -158,6 +163,9 @@ timestamps enter seconds[.nanoseconds], where seconds is the number of seconds
 that have elapsed since January 1, 1970 (midnight UTC/GMT), not counting leap
 seconds (aka Unix epoch or Unix time), and the optional .nanoseconds field is a
 fraction of a second no more than nine digits long.
+
+Note that there is a limited backlog of events you can filter. `docker events` is 
+a rolling log, and has at most 1000 events at a time.
 
 #### Filtering
 
--- a/docs/reference/commandline/events.md
+++ b/docs/reference/commandline/events.md
@@ -31,12 +31,12 @@ Options:
 ## Description
 
 Use `docker events` to get real-time events from the server. These events differ
-per Docker object type. Different event types have different scopes. The local 
+per Docker object type. Different event types have different scopes. Local 
 scoped events are only seen on the node they take place on, and swarm scoped 
 events are seen on all managers.
 
-Note that there is a limited backlog of events you can filter. `docker events` is 
-a rolling log, and has at most 1000 events at a time.
+Only the last 1000 log events are returned. You can use filters to further limit 
+the number of events returned.
 
 ### Object types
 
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/events.md
```diff
@@ -31,7 +31,12 @@ Options:
 ## Description
 
 Use `docker events` to get real-time events from the server. These events differ
-per Docker object type.
+per Docker object type. Different event types have different scopes. The local 
+scoped events are only seen on the node they take place on, and swarm scoped 
+events are seen on all managers.
+
+Note that there is a limited backlog of events you can filter. `docker events` is
```
thaJeztah: If this is a "note", this should use our standard markup for notes:

```Markdown
> **Note**: There is a limited backlog of events you can filter. `docker events`
> is  a rolling log, and has at most 1000 events at a time.
```

However, trying to preserve "notes" for exceptional cases; perhaps this can be rephrased slightly to not be a "note". Something like:

```Markdown
Docker uses a rolling log to store the last 1000 events, which means that at
most 1000 events at a time are returned.
```

(not entirely happy with that example; perhaps you have a better suggestion)

docs/reference/commandline/events.md
```diff
@@ -159,6 +164,9 @@ that have elapsed since January 1, 1970 (midnight UTC/GMT), not counting leap
 seconds (aka Unix epoch or Unix time), and the optional .nanoseconds field is a
 fraction of a second no more than nine digits long.
 
+Note that there is a limited backlog of events you can filter. `docker events` is
```
thaJeztah: Same here 😅

docs/reference/commandline/events.md
```diff
@@ -31,7 +31,12 @@ Options:
 ## Description
 
 Use `docker events` to get real-time events from the server. These events differ
-per Docker object type.
+per Docker object type. Different event types have different scopes. The local
```
mistyhacks: s/The local/Local

docs/reference/commandline/events.md
mistyhacks: How about just "Only the last 1000 log events are returned. You can use filters to further limit the number of events returned."

BretFisher:
ping @thaJeztah and @mistyhacks

mistyhacks:
I updated the title to not include an issue number in it, since the title will become the commit message, and would ping issue 727 every time it hits a different branch or even gets put into a different repo (like `docker/docker-ce`). This kind of info is best left in the description like you already have it.

thaJeztah:
Ah, yes, bet to remove that from the commit description (GitHub is a pain in that respect) 😅

thaJeztah:
carried / squashed in https://github.com/docker/cli/pull/1999

PR #767, opened 2017-12-26
Add gcc and musl-dev to dev Dockerfile

seemethere:
Needed to be able to do `make -f docker.Makefile binary` on multiple
architectures.

Signed-off-by: Eli Uriegas <eli.uriegas@docker.com>

**- What I did**
Fixed `make -f docker.Makefile binary` on `aarch64`, would fail before with:

```
Step 4/11 : RUN     go get -d github.com/LK4D4/vndr &&         cd /go/src/github.com/LK4D4/vndr &&         git checkout -q "$VNDR_SHA" &&         go build -v -o /usr/bin/vndr . &&         rm -rf /go/src/* /go/pkg/* /go/bin/*
 ---> Running in bffec54b704f
github.com/LK4D4/vndr/godl/singleflight
github.com/LK4D4/vndr/build
github.com/LK4D4/vndr/godl
github.com/LK4D4/vndr
# github.com/LK4D4/vndr
/usr/local/go/pkg/tool/linux_arm64/link: running gcc failed: exec: "gcc": executable file not found in $PATH

The command '/bin/sh -c go get -d github.com/LK4D4/vndr &&         cd /go/src/github.com/LK4D4/vndr &&         git checkout -q "$VNDR_SHA" &&         go build -v -o /usr/bin/vndr . &&         rm -rf /go/src/* /go/pkg/* /go/bin/*' returned a non-zero code: 2
docker.Makefile:18: recipe for target 'build_docker_image' failed
make: *** [build_docker_image] Error 2
```

**- How I did it**

* Added `gcc` and `musl-dev` to the Dockerfile.dev

**- How to verify it**
* Running `make -f docker.Makefile binary` on `aarch64` 

**- Description for the changelog**

```
* Added `gcc` and `musl-dev` to the `Dockerfile.dev`
```

**- A picture of a cute animal (not mandatory but encouraged)**
darth bird

PR #816, opened 2018-01-18
Add dockerfile for building on non-amd64 platforms

dnephin:
Closes #767

Add a new `Dockerfile`, and `docker.Makefile` target for building on other platforms that require additional dependencies.

```
make -f docker.Makefile binary
```

cc @seemethere @arm64b


dnephin:
@vdemeester brought up a good point, the old `binary` target is actually not necessary anymore, so instead of adding a new target I've just replaced the old one.

arm64b:
Thanks for the support 😄

PR #821, opened 2018-01-19
Container: let rm -f exit with success

akimd:
`rm -f non-existing` exits with success, and so does `docker rmi -f non-existing`.  But not `docker rm -f`.

Arguably we should actually check whether these errors are really doesNotExist (rm -f exits with failure for lack of permissions for instance), but at least we should be consistent.

FWIW, in the code, we have Rm in cli/command/container/rm.go and Remove in cli/command/image/remove.go.  The latter has a test suite, not the former.

lamasticot

### DIFF
--- a/cli/command/container/rm.go
+++ b/cli/command/container/rm.go
@@ -67,7 +67,11 @@ func runRm(dockerCli command.Cli, opts *rmOptions) error {
 		fmt.Fprintln(dockerCli.Out(), name)
 	}
 	if len(errs) > 0 {
-		return errors.New(strings.Join(errs, "\n"))
+		msg := strings.Join(errs, "\n")
+		if !opts.force {
+			return errors.New(msg)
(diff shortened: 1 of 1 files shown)


dnephin:
This is the same change that was made to `rmi` initially and had to be reverted. This is not correct, it needs to check for a "Not Found" error and only ignore those.

akimd:
Thanks for the pointers!  I'll have a look, but I can't tell when (yet).

thaJeztah:
Looks like this was addressed in https://github.com/docker/cli/pull/2678. The error is still printed, but the exit-code is `0` ("success");

```bash
docker rm nosuchcontainer
Error: No such container: nosuchcontainer

echo $?
1

docker rm -f nosuchcontainer
Error: No such container: nosuchcontainer

echo $?
0
```

PR #828, opened 2018-01-24
Update authors

thaJeztah:
ping @vdemeester @dnephin


thaJeztah:
> We should create a bot that does that each and every month or sthg

It's updating the `.mailmap` that's usually the tricky bit; perhaps we need a check that the DCO sign-off doesn't use `<github-handle>` as "name"

PR #832, opened 2018-01-25
Document SELinux relabelling options and add warning

justincormack:
Make it clear what the `z` and `Z` options do, and that they
affect files on the host system.

See the discussion in https://github.com/moby/moby/issues/30934
for the reasons a warning is needed.

Signed-off-by: Justin Cormack <justin.cormack@docker.com>

animal rave

### DIFF
--- a/docs/reference/run.md
+++ b/docs/reference/run.md
@@ -1590,6 +1590,23 @@ For example, you can specify either `/foo` or `foo` for a `host-src` value.
 If you supply the `/foo` value, Docker creates a bind mount. If you supply
 the `foo` specification, Docker creates a named volume.
 
+The `z` and `Z` modifiers will change the SELinux labels on the mount points.
+The `z` option will make the files available to any container, using the `s0` label,
+while the `Z` option will label the files with the same label as the container, so
+that they are exclusive to that container.
+
+WARNING Use of :z or :Z can have impacts on your system if used too broadly.
+For example, the use of these options against a users HOME directory or other top
+level system directory would result in the relabeling of those directories for use
+by containers. There are some blacklisted paths, but this could cause issues such
+as being unable to SSH back into a system. When using the relabel options, it should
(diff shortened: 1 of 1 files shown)


docs/reference/run.md
```diff
@@ -1590,6 +1590,23 @@ For example, you can specify either `/foo` or `foo` for a `host-src` value.
...
+
+WARNING Use of :z or :Z can have impacts on your system if used too broadly.
+For example, the use of these options against a users HOME directory or other top
+level system directory would result in the relabeling of those directories for use
+by containers. There are some blacklisted paths, but this could cause issues such
+as being unable to SSH back into a system. When using the relabel options, it should
+only be used on directories that are intended for use by containers, and not the
+host system itself. In most cases, such as shared access,  it is better to label
+in advance. If your container does require broader access to system directories,
+then use of '--security-opt label:disable' with the 'docker run' command is a better
```
cpuguy83: Is this such a good thing to encourage?

mistyhacks: I swear we already have something about this in the docs, let me find it. Yes, here: https://docs.docker.com/engine/admin/volumes/bind-mounts/#configure-the-selinux-label Perhaps you can just link to that instead of putting these docs here.

mdlinville (changes requested):

docs/reference/run.md
thaJeztah: oh, forgot about that; good point. @justincormack perhaps you can look at that section in the docs and see if anything is missing there? Then we can link to there

AkihiroSuda:
What's current status?

@justincormack

AkihiroSuda:
What's current status?

cc @thaJeztah

Benehiko:
Since this document already exists somewhere else as pointed out [here](https://github.com/docker/cli/pull/832#discussion_r163920188) do we want to have this or do we just close this PR? /cc @dvdksn

PR #838, opened 2018-01-26
Fix `--label-file` weird behavior

vdemeester:
`--label-file` has the exact same behavior as `--env-file`, meaning any
placeholder (i.e. a simple key, no `=` sign, no value), it will get the
value from the environment variable.

For `--label-file` it should just add an empty label. 

(Same for `--label`, it should never get value from the environment variables).

Fix https://github.com/moby/moby/issues/31087.

🐅 🦁 

### DIFF
--- a/opts/parse.go
+++ b/opts/parse.go
@@ -2,6 +2,7 @@ package opts
 
 import (
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


opts/parse.go
```diff
@@ -10,10 +10,10 @@ import (
 
 // ReadKVStrings reads a file of line terminated key=value pairs, and overrides any keys
 // present in the file with additional pairs specified in the override parameter
-func ReadKVStrings(files []string, override []string) ([]string, error) {
+func ReadKVStrings(files []string, override []string, emptyFn func(string) string) ([]string, error) {
```
thaJeztah: A couple of lines below, this still refers to `environment variables`.

Perhaps we should (for now)

- un-export `ParseKeyValueFile`
- have two wrappers; one variation of `ReadKVString` for labels, and one for environment variables
- those wrappers don't need the `emptyFn` as an argument (while nice, we don't really need that functionality at this point)

vdemeester: 😝 I can take care of that 👼

vdemeester:
Ok this caused a regression, I'll fix it 😅

PR #841, opened 2018-01-27
Add e2e container kill test

vdemeester:

### DIFF
--- a/e2e/container/kill_test.go
+++ b/e2e/container/kill_test.go
@@ -1,0 +1,53 @@
+package container
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/docker/cli/e2e/internal/fixtures"
+	"github.com/gotestyourself/gotestyourself/icmd"
+	"github.com/gotestyourself/gotestyourself/poll"
+)
+
+func TestKillContainer(t *testing.T) {
+	containerID := runBackgroundTop(t)
+
+	// Kill with SIGTERM should kill the process
+	result := icmd.RunCmd(
+		icmd.Command("docker", "kill", "-s", "SIGTERM", containerID),
+	)
+
+	result.Assert(t, icmd.Success)
+	poll.WaitOn(t, containerStatus(t, containerID, "exited"), poll.WithDelay(100*time.Millisecond), poll.WithTimeout(5*time.Second))
+
+	// Kill on a stop container should return an error
+	result = icmd.RunCmd(
+		icmd.Command("docker", "kill", containerID),
+	)
+	result.Assert(t, icmd.Expected{
+		ExitCode: 1,
+		Err:      "is not running",
+	})
+}
+
+func runBackgroundTop(t *testing.T) string {
+	result := icmd.RunCmd(shell(t,
+		"docker run -d %s top", fixtures.AlpineImage))
+	result.Assert(t, icmd.Success)
+	return strings.TrimSpace(result.Stdout())
+}
+
+func containerStatus(t *testing.T, containerID, status string) func(poll.LogT) poll.Result {
+	return func(poll.LogT) poll.Result {
+		result := icmd.RunCmd(
+			shell(t, "docker inspect -f '{{ .State.Status }}' %s", containerID),
+		)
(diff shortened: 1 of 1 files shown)


e2e/container/kill_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Kill with a signal should send the signal to the process
+	// if the signal is not a deadly one, the container should still live on
+	result := icmd.RunCmd(
+		icmd.Command("docker", "kill", "-s", "SIGWINCH", containerID),
+	)
+	result.Assert(t, icmd.Success)
+
+	result = icmd.RunCmd(
+		shell(t, "docker inspect -f '{{ .State.Status }}' %s", containerID),
```
silvin-lubecki: Maybe poll the check of the status?

e2e/container/kill_test.go
```diff
@@ -0,0 +1,78 @@
...
+	// Kill on a stop container should return an error
+	result = icmd.RunCmd(
+		icmd.Command("docker", "kill", containerID),
+	)
+	result.Assert(t, icmd.Expected{
+		ExitCode: 1,
+	})
+}
+
+func TestKillRunningContainer(t *testing.T) {
```
silvin-lubecki: This test is a little redundant with the previous one, no?

vdemeester: arf this one was supposed to be gone yep 😛

e2e/container/kill_test.go
vdemeester: I shouldn't need to poll here (because the status is supposed to not changed). That said, I wonder if I should even test this signal (i.e. I should just test that the cli sends the correct signal to the daemon) — the handle of signals, etc.. should be on moby integration tests 🤔

dnephin: Yes, I think we can omit this case. "sends the correct signal" can be tested in unit tests, and the rest outside of this repo.

e2e/container/kill_test.go
```diff
@@ -0,0 +1,66 @@
...
+	)
+	result.Assert(t, icmd.Success)
+
+	result = icmd.RunCmd(
+		shell(t, "docker inspect -f '{{ .State.Status }}' %s", containerID),
+	)
+	result.Assert(t, icmd.Success)
+	require.Equal(t, "running", strings.TrimSpace(result.Stdout()))
+
+	// Kill with no signal should kill the process
```
dnephin: Defaults could be tested with unit tests. You could use a SIGTERM here to cover testing the CLI accepts the arg properly.

e2e/container/kill_test.go
```diff
@@ -0,0 +1,52 @@
...
+func TestKillContainer(t *testing.T) {
+	containerID := runBackgroundTop(t)
+
+	// Kill with SIGTERM should kill the process
+	result := icmd.RunCmd(
+		icmd.Command("docker", "kill", "-s", "SIGTERM", containerID),
+	)
+
+	result.Assert(t, icmd.Success)
+	poll.WaitOn(t, containerStatus(t, containerID, "exited"), poll.WithDelay(1*time.Second), poll.WithTimeout(10*time.Second))
```
dnephin: Won't the container be in the exited state before `kill` returns?

If not, could we use a delay of 100ms ?

e2e/container/kill_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Kill on a stop container should return an error
+	result = icmd.RunCmd(
+		icmd.Command("docker", "kill", containerID),
+	)
+	result.Assert(t, icmd.Expected{
+		ExitCode: 1,
```
dnephin: Can we expect any Stderr from this failure?

e2e/container/kill_test.go
dnephin: Thinking about it again, it definitely wont have that state before `kill` returns, so maybe just a shorter delay (like 100ms) ?

vdemeester: yep, `1s` might be too long for sure 😅

vdemeester:
@dnephin updated 👼

PR #845, opened 2018-01-29
[compose] Share the compose loading code between swarm and k8s stack deploy

vdemeester:
To ensure we are loading the composefile the same wether we are pointing
to swarm or kubernetes, we need to share the loading code between both.

This means though that `docker stack deploy` pointing to k8s doesn't handle `.env` anymore (same beahviour as `docker stack deploy` on swarm).

- [x] struct mashall to a valid composefile
- [x] testing this on a real kubernetes cluster with the compose kubernetes on.


cli/command/stack/deploy.go
```diff
@@ -47,3 +59,106 @@ func newDeployCommand(dockerCli command.Cli) *cobra.Command {
...
+			return details, err
+		}
+		details.WorkingDir = filepath.Dir(absPath)
+	}
+
+	configFile, err := getConfigFile(composefile, stdin)
+	if err != nil {
+		return details, err
+	}
+	// TODO: support multiple files
```
vdemeester: This will be fixed by https://github.com/docker/cli/pull/569 👼

cli/command/stack/kubernetes/loader.go
```diff
@@ -1,169 +1,27 @@
...
-	if err != nil {
-		return nil, nil, errors.Wrapf(err, "cannot load compose file")
-	}
-
-	result, err := processEnvFiles(processed, parsed, cfg)
-	if err != nil {
-		return nil, nil, errors.Wrapf(err, "cannot load compose file")
-	}
-
+	fmt.Println(string(res))
```
silvin-lubecki: println?

vdemeester: yeah it was for *debugging* purpose, I'll remove once not wip anymore

silvin-lubecki: Yeah right... 😏

cli/command/stack/deploy.go
```diff
@@ -19,14 +31,15 @@ func newDeployCommand(dockerCli command.Cli) *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
 			opts.Namespace = args[0]
+
 			if dockerCli.ClientInfo().HasKubernetes() {
 				kli, err := kubernetes.WrapCli(dockerCli, cmd)
 				if err != nil {
 					return err
 				}
-				return kubernetes.RunDeploy(kli, opts)
+				return kubernetes.RunDeploy(kli, opts, loadComposefile)
```
silvin-lubecki: I don't like this functor. I think the only reason is to avoid circular dependency between stack package and swarm/kubernetes packages. A more cleaner way could be moving the loader in its own package, like cli/command/stack/loader.

cli/command/stack/kubernetes/loader_test.go
```diff
@@ -1,34 +0,0 @@
-package kubernetes
-
-import (
-	"testing"
-
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-)
-
-func TestPlaceholders(t *testing.T) {
```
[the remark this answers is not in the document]
vdemeester: Yes 😏 interpolation is already tested in the `cli/compose/...` packages 😉

vdemeester:
So… yaml marshaling the `types.Config` struct from the `compose` package produce an invalid composefile 🤔 We may need to fix that 😅

dnephin:
Ya, I guess it's not necessarily designed to be round tripped. How is it invalid?

vdemeester:
Services are marshalled as a list, but they are unmarshalled as a mapping.

This PR is not mergeable until we can fix that 😅

vdemeester:
Updated: it should now generate a valid yaml composefile — also tried it out on docker-for-mac kubernetes and it works !

Next step should be to upgrade the [current test](https://github.com/docker/cli/pull/845/files#diff-3770c003fec9854400b167622146d41bR10) with everything, probably reuse `full-example` yaml file (and struct)

cli/compose/types/types.go
```diff
@@ -77,69 +78,96 @@ type Config struct {
...
+	if c.Services != nil {
+		services := map[string]ServiceConfig{}
+		for _, service := range c.Services {
+			s := service
+			s.Name = ""
+			services[service.Name] = s
+		}
+		m["services"] = services
+	}
+	if c.Networks != nil {
```
dnephin: Are these `nil` checks necessary (below)? If it is nil, it can assign nil without any problems, right? I think they can be removed.

dnephin:
With full test looks good!

cli/compose/types/types.go
vdemeester: Right 😛

cli/compose/loader/full-example.yml
```diff
@@ -193,7 +193,7 @@ services:
 
     ports:
       - 3000
-      - "3000-3005"
+      - "3001-3005"
```
vdemeester: Changed that to not have duplicate in the marshalled struct later.

vdemeester:
@dnephin Updated with the full-example 🎉

cli/command/stack/kubernetes/loader.go
```diff
@@ -1,170 +1,25 @@
...
-	result, err := processEnvFiles(processed, parsed, cfg)
+// LoadStack loads a stack from a Compose config, with a given name.
+func LoadStack(name, version string, cfg *composetypes.Config) (*apiv1beta1.Stack, error) {
+	res, err := yaml.Marshal(cfg)
 	if err != nil {
-		return nil, nil, errors.Wrapf(err, "cannot load compose file")
+		return nil, err
 	}
-
+	yaml := "version: \"" + version + "\"\n" + string(res)
```
dnephin: Could we do this by marshaling a local struct:

```go
type versionedConfig struct {
    composetypes.Config
    Version string
}
```

thaJeztah:
ping @dnephin still LGTY?

cli/compose/loader/types_test.go
```diff
@@ -0,0 +1,328 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	yaml "gopkg.in/yaml.v2"
+)
+
+func TestMarshallConfig(t *testing.T) {
+	cfg := fullExampleConfig("/foo", "/bar")
+	expected := `configs: {}
+networks:
```
dnephin: Hopefully `MarshalYAML` is stable and doesn't change order randomly when we change fields.

vdemeester: so far it did not 😇

PR #848, opened 2018-01-30
Fix the network option table

mistyhacks:
Fixes https://github.com/docker/docker.github.io/issues/5866

The table was formatted a little weird. I added a little more info and fixed a typo too.

bees on finger

### DIFF
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -766,18 +766,14 @@ The options you can specify are:
 </tr>
 </thead>
 <tr>
-  <td>published and target port </td>
-  <td><tt></tt></td>
-  <td><tt></tt></td>
-  <td></td>
-</tr>
-<tr>
-  <td>protocol</td>
+  <td>published and target port</td>
   <td><tt>--publish 8080:80</tt></td>
   <td><tt>--publish published=8080,target=80</tt></td>
   <td><p>
-    The port to publish the service to on the routing mesh or directly on
-    the node, and the target port on the container.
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/service_create.md
```diff
@@ -766,18 +766,14 @@ The options you can specify are:
...
-<tr>
-  <td>protocol</td>
+  <td>published and target port</td>
   <td><tt>--publish 8080:80</tt></td>
   <td><tt>--publish published=8080,target=80</tt></td>
   <td><p>
-    The port to publish the service to on the routing mesh or directly on
-    the node, and the target port on the container.
+    The target port within the container and the port to map it to on the
+    nodes, using the routing mesh (`ingress`) or host-level networking.
```
thaJeztah: Looks like (because a _html_ table is used), the backticks also have to be changed to `<tt>` / `</tt>`

screen shot 2018-01-30 at 13 44 45

thaJeztah: (Same for other occurrences in the table)

mistyhacks: I just forgot to set `markdown=1` on the table. I'll do that now.

mistyhacks: I just forgot to set `markdown=1` on the table. I'll do that now.

mistyhacks:
Decided to go with `<tt>` all the way down. The `markdown="1"` has to be set on each `<td>` element, which is too annoying for this.

PR #854, opened 2018-01-31
WIP Handle multiple compose versions

silvin-lubecki:
**- What I did**
A new compose component for Kubernetes version has been released (v1beta2), and I added support for both versions. I also added an `Orchestrator` section in the `docker version` command to log the compose component version and Kubernetes version we are working against.

```
$ docker version
Client:
 Version:       18.02.0-dev
 API version:   1.35 (downgraded from 1.36)
 Go version:    go1.9.3
 Git commit:    0d6a70514
 Built: Wed Jan 31 16:22:38 2018
 OS/Arch:       darwin/amd64
 Experimental:  true
 Orchestrator:  kubernetes

Server:
 Engine:
  Version:      18.01.0-ce
  API version:  1.35 (minimum version 1.12)
  Go version:   go1.9.2
  Git commit:   03596f5
  Built:        Wed Jan 10 20:13:12 2018
  OS/Arch:      linux/amd64
  Experimental: true
 Orchestrator:
   Kubernetes:  v1.8.2
   Compose:     v1beta2
```

**- How I did it**
I put an interface on the compose client used by docker stack subcommands and added an implementation for each compose version.

**- How to verify it**

* New Orchestrator section in docker version command and added support of kubernetes compose component v1beta2

image


cli/flags/common.go
```diff
@@ -69,6 +69,13 @@ func (commonOpts *CommonOptions) InstallFlags(flags *pflag.FlagSet) {
 	flags.Var(opts.NewQuotedString(&tlsOptions.CertFile), "tlscert", "Path to TLS certificate file")
 	flags.Var(opts.NewQuotedString(&tlsOptions.KeyFile), "tlskey", "Path to TLS key file")
 
+	flags.String("namespace", "default", "Kubernetes namespace to use")
+	flags.SetAnnotation("namespace", "kubernetes", nil)
+	flags.SetAnnotation("namespace", "experimentalCLI", nil)
+	flags.String("kubeconfig", "", "Kubernetes config file")
+	flags.SetAnnotation("kubeconfig", "kubernetes", nil)
+	flags.SetAnnotation("kubeconfig", "experimentalCLI", nil)
```
dnephin: Why do these need to move to common? They aren't common to both orchestrators

silvin-lubecki: Well I mostly need `kubeconfig` (not `namespace`) to be common, or at least available, for `docker version` command, as you may specify different kubernetes configs and check the version against the pointing kubernetes.

dnephin: I see, but this flag isn't added to the `version` command, so how is it going to be read?

cli/command/system/version.go
```diff
@@ -125,6 +141,11 @@ func runVersion(dockerCli command.Cli, opts *versionOptions) error {
 			Status: "Template parsing error: " + err.Error()}
 	}
 
+	kubeVersion, err := getKubernetesVersion(dockerCli, cmd)
+	if err != nil {
+		return err
```
dnephin: This could cause `version` to exit with an error if the stack API isn't available, which would result in the user not seeing any version info.

I don't think that's the correct behaviour.  I think the error case should set any any empty fields in `kubeVersion`  to "unknown", print the error as a warning, and continue with printing the version. For example maybe `getKubernetesVersion` is able to get a kube version, but not the stack API version

cli/command/system/version.go
```diff
@@ -75,12 +82,21 @@ type clientVersion struct {
 	Orchestrator      string `json:",omitempty"`
 }
 
+type kubernetesVersion struct {
+	Kubernetes string
+	ComposeAPI string
```
dnephin: `StackAPI` not `ComposeAPI`

cli/command/system/version.go
```diff
@@ -189,3 +211,32 @@ func getDetailsOrder(v types.ComponentVersion) []string {
...
+		return nil, nil
+	}
+
+	kli, err := kubernetes.WrapCli(dockerCli, cmd)
+	if err != nil {
+		return nil, err
+	}
+
+	apiVersion, err := kli.GetAPIVersion()
+	if err != nil {
```
dnephin: If this fails it should still attempt to get the kube version, so maybe this should be split into two functions.

cli/command/system/version.go
```diff
@@ -189,3 +211,32 @@ func getDetailsOrder(v types.ComponentVersion) []string {
 	sort.Strings(out)
 	return out
 }
+
+func getKubernetesVersion(dockerCli command.Cli, cmd *cobra.Command) (*kubernetesVersion, error) {
+	if !dockerCli.ClientInfo().HasKubernetes() {
+		return nil, nil
+	}
+
+	kli, err := kubernetes.WrapCli(dockerCli, cmd)
```
dnephin: Instead of `WrapCLI()` if this could call a method that returns `KubeConfig` then it can be used for both `GetAPIVersion` and getting a `kubeClient` without the need for `KubeCli`.

cli/command/system/version.go
```diff
@@ -109,7 +125,7 @@ func reformatDate(buildTime string) string {
 	return buildTime
 }
 
-func runVersion(dockerCli command.Cli, opts *versionOptions) error {
+func runVersion(dockerCli command.Cli, cmd *cobra.Command, opts *versionOptions) error {
```
dnephin: I think we should not be passing in `cobra.Command` to the run functions. With the other suggestions (refactoring to remove the need to use `KubeCli` to get version, and getting the kube config from a `path string` arg instead of flags this shouldn't be necessary any more and can be reverted.

cli/command/stack/kubernetes/cli.go
```diff
@@ -52,25 +51,27 @@ func WrapCli(dockerCli command.Cli, cmd *cobra.Command) (*KubeCli, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Failed to load kubernetes configuration file '%s'", kubeConfig)
 	}
-	cli.kubeConfig = config
+	cli.KubeConfig = config
```
dnephin: Related to other changes, this "get KubeConfig from flag or env" can be extracted into a function that just takes a single `path string` argument.

cli/command/stack/kubernetes/check.go
```diff
@@ -4,29 +4,57 @@ import (
...
-// APIPresent checks that an API is installed.
-func APIPresent(config *rest.Config) error {
-	log.Debugf("check API present at %s", config.Host)
-	clients, err := kubernetes.NewForConfig(config)
+// StackVersion represents the detected Compose Component on Kubernetes side.
+type StackVersion string
+
+const (
+	// StackNotFound is returned when no stack api at all is detected on kubernetes.
+	StackNotFound = "notFound"
```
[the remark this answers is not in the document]
silvin-lubecki: Ok! It was only to be consistent with the "enum" values. But "" is fine.

silvin-lubecki:
@dnephin Let me just fix your comments before I split it into 2 PR? Otherwise they may "disappear".

cli/command/stack/kubernetes/check.go
```diff
@@ -4,29 +4,57 @@ import (
...
+	// StackAPIV1Beta1 is returned if it's the most recent version available.
+	StackAPIV1Beta1 = "v1beta1"
+	// StackAPIV1Beta2 is returned if it's the most recent version available.
+	StackAPIV1Beta2 = "v1beta2"
+)
+
+// GetAPIVersion returns the most recent stack API installed.
+func (c *KubeCli) GetAPIVersion() (StackVersion, error) {
```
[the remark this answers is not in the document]
silvin-lubecki: Right, I'll change this.

kubernetes/README.md
```diff
@@ -1,4 +0,0 @@
-# Kubernetes client libraries
```
vdemeester: We should not remove that 👼

kubernetes/compose/v1beta1/scale.go
```diff
@@ -0,0 +1,77 @@
+package v1beta1 // import "github.com/docker/cli/kubernetes/compose/v1beta1"
+
+import (
+	"errors"
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+// Scale scale a service to the specified number of replicas
```
vdemeester: v1beta1 didn't support scale, it should still not support it (i.e. it comes in v1beta2).. so this should not be required

kubernetes/compose/v1beta1/stack_types.go
```diff
@@ -1,27 +1,45 @@
...
+	}
+	return result
+}
 
 // Stack defines a stack object to be register in the kubernetes API
-// +k8s:openapi-gen=true
-// +resource:path=stacks,strategy=StackStrategy
-// +subresource:request=Owner,path=owner,rest=OwnerStackREST
 type Stack struct {
+	StackImpl
```
vdemeester: I don't think this is required at all 🤔 (there shouldn't be any code change in v1beta1)

cli/command/stack/kubernetes/check.go
```diff
@@ -4,29 +4,57 @@ import (
...
-		if group.Name == apiv1beta1.SchemeGroupVersion.Group {
-			return nil
-		}
+	switch {
+	case findVersion(apiv1beta2.SchemeGroupVersion, groups.Groups):
+		return StackAPIV1Beta2, nil
+	case findVersion(apiv1beta1.SchemeGroupVersion, groups.Groups):
+		return StackAPIV1Beta1, nil
+	default:
+		return StackNotFound, fmt.Errorf("could not find %s api. Install it on your cluster first", apiv1beta1.SchemeGroupVersion.Group)
```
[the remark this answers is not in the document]
silvin-lubecki: Done

cli/command/stack/kubernetes/cli.go
```diff
@@ -52,25 +51,27 @@ func WrapCli(dockerCli command.Cli, cmd *cobra.Command) (*KubeCli, error) {
...
-	clientSet, err := composev1beta1.NewForConfig(c.kubeConfig)
-	if err != nil {
-		return nil, err
+	switch version {
+	case StackAPIV1Beta1:
+		return c.newStackV1Beta1()
+	case StackAPIV1Beta2:
+		return c.newStackV1Beta2()
+	default:
+		return nil, fmt.Errorf("could not find matching Stack API version")
```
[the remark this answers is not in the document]
silvin-lubecki: Done

cli/command/system/version.go
silvin-lubecki: Done

cli/command/system/version.go
silvin-lubecki: Done

cli/command/system/version.go
silvin-lubecki: Right, done

silvin-lubecki:
Split and reopened in #898 and #899

PR #858, opened 2018-01-31
Add an example for --with-registry-auth

mistyhacks:
Relates to https://github.com/docker/docker.github.io/issues/4107
Relates to https://github.com/docker/docker.github.io/issues/5697

beagle puppy

### DIFF
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -112,6 +112,26 @@ dmu1ept4cxcf  redis   replicated  1/1       redis:3.0.6
 a8q9dasaafud  redis2  global      1/1       redis:3.0.6
 ```
 
+#### Create a service using an image on a private registry
+
+If your image is available on a private registry which requires login, use the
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/service_create.md
```diff
@@ -112,6 +112,26 @@ dmu1ept4cxcf  redis   replicated  1/1       redis:3.0.6
...
+$ docker service  create \
+  --with-registry-auth \
+  --name my_service \
+  registry.example.com/acme/my_image:latest
+```
+
+This passes the login token from your local client to the swarm nodes where the
+service is deployed, using the encrypted WAL logs. With this information, the
+nodes are able to log into the registry and pull the image.
+
```
thaJeztah: As a follow up, perhaps we should explain that these credentials are _stored_ in the swarm (and are also used when scaling a service or if tasks are rescheduled). One thing people trip over is that some registries use a short-lived token for authentication, which means that if a task is rescheduled, the token may be expired, so pulling the image fails.

PR #862, opened 2018-02-02
Fix vendor validation (and vendor)

dnephin:
vendor/ must be removed first, otherwise files added to vendor/ but not vendor.conf
will not cause the validation to fail

It turns out we were missing a version for `x/sync` as well.


vendor.conf
```diff
@@ -68,6 +68,7 @@ github.com/xeipuuv/gojsonreference e02fc20de94c78484cd5ffb007f8af96be030a45
 github.com/xeipuuv/gojsonschema 93e72a773fade158921402d6a24c819b48aba29d
 golang.org/x/crypto 558b6879de74bc843225cde5686419267ff707ca
 golang.org/x/net a8b9294777976932365dabb6640cf1468d95c70f
+golang.org/x/sync f52d1811a62927559de87708c
```
thaJeztah: Oh, nice catch; looks like that was commented-out in https://github.com/docker/cli/commit/8417e4979261c7459f82b3dbd633d674a119e38e, and removed in https://github.com/docker/cli/commit/61713c42a4e02ff20cacce092ed651344af8bf92

PR #866, opened 2018-02-08
[WIP] manifest push using yaml file

clnperez:
Instead of:
manifest create list image1 image2
manifest annotate list image1
manifest push list

You can do:
manifiest push --filename myfile.yaml

Signed-off-by: Christy Norman Perez <christy@linux.vnet.ibm.com>

**- What I did**

**- How I did it**

**- How to verify it**


clnperez:
Also, so I don't forget, I need to update the doc to add this, and and to include how to set the *cli* experimental. It's confusing, because it's "true" for the engine, but "enabled" for the cli. :D

cli/command/manifest/push.go
```diff
@@ -42,32 +47,63 @@ type pushRequest struct {
...
+	flags.BoolVarP(&opts.purge, "purge", "p", false, "remove the local manifest list after push")
+	flags.BoolVar(&opts.insecure, "insecure", false, "allow push to an insecure registry")
+	flags.StringVar(&opts.file, "filename", "", "create, annotate, and push using a yaml file")
 	return cmd
 }
 
 func runPush(dockerCli command.Cli, opts pushOpts) error {
 
+	if opts.target == "" && opts.file == "" {
+		return errors.New("please specify either a file or local manifest name")
```
dnephin: I believe we generally avoid "please" in error messages. Maybe "either a filename or local manifest name is required" or something like that?

cli/command/manifest/push.go
```diff
@@ -42,32 +47,63 @@ type pushRequest struct {
...
+		yamlInput, err := getYamlManifestList(dockerCli, opts.file)
+		if err != nil {
+			return err
+		}
+		if err := createListFromYaml(dockerCli, yamlInput, opts.insecure); err != nil {
+			return err
+		}
+		if err := annotateListFromYaml(dockerCli, yamlInput); err != nil {
+			return err
+		}
```
dnephin: It seems a bit odd to me that `push` would create local files. could this push directly from yaml without creating the files locally?

If we want to support "load defaults from yaml" or something like that we could have `create` accept the same yaml format (either in this PR or a separate one).

cli/command/manifest/push.go
```diff
@@ -42,32 +47,63 @@ type pushRequest struct {
...
 			return runPush(dockerCli, opts)
 		},
 	}
 
 	flags := cmd.Flags()
-	flags.BoolVarP(&opts.purge, "purge", "p", false, "Remove the local manifest list after push")
-	flags.BoolVar(&opts.insecure, "insecure", false, "Allow push to an insecure registry")
+	flags.BoolVarP(&opts.purge, "purge", "p", false, "remove the local manifest list after push")
+	flags.BoolVar(&opts.insecure, "insecure", false, "allow push to an insecure registry")
+	flags.StringVar(&opts.file, "filename", "", "create, annotate, and push using a yaml file")
```
dnephin: I wonder if `file` should be a boolean flag that changes the positional argument from a "local manifest target" to "a filepath".  I'm not sure if that's more or less expected than having a conditional argument.

I think it would at least make the validation easier, because with a bool flag a positional argument would always be necessary.

cli/command/manifest/push.go
```diff
@@ -271,3 +307,47 @@ func mountBlobs(ctx context.Context, client registryclient.RegistryClient, ref r
...
+	}
+	return createManifestList(cli, args, createOpts{amend: false, insecure: insecureRegistry})
+}
+
+func getYamlManifestList(dockerCli command.Cli, yamlFile string) (yamlManifestList, error) {
+	var yamlInput yamlManifestList
+
+	if _, err := os.Stat(yamlFile); err != nil {
+		return yamlManifestList{}, err
+	}
```
dnephin: Is this Stat necessary? I think you'd get the same error from `ReadFile()`.

cli/command/manifest/push.go
```diff
@@ -271,3 +307,47 @@ func mountBlobs(ctx context.Context, client registryclient.RegistryClient, ref r
...
+	yamlBuf, err := ioutil.ReadFile(yamlFile)
+	if err != nil {
+		return yamlManifestList{}, err
+	}
+	if err := yaml.UnmarshalStrict(yamlBuf, &yamlInput); err != nil {
```
dnephin: This `if` is unnecessary:

```go
return yamlInput, yaml.UnmarshalStrict(yamlBuf, &yamlInput)
```

clnperez:
Redid this but haven't tested the original path yet, so I'll get to that on Monday, and also take the reset of @dnephin's comments into account. Sorry for abandoning this for so long!

cli/command/manifest/push.go
clnperez: nope, thanks!

clnperez:
tests added for yaml. removing `[WIP]`

@vdemeester @thaJeztah the week before dockercon, so, you've got time to take a look at this now right? :angel:

clnperez:
cc @tianon @StefanScherer @AceHack

StefanScherer:
Thanks @clnperez I'll look into it in the next days. Haven't looked into the details, but agree that pushing a yaml should be done with the `docker manifest push` command.

clnperez:
@vdemeester i wasn't a huge fan of the file change but it grew on me and did make validation easier. i'm open to changes for sure. but i'm not sure i like having someone specify any information on the command line that could be also in a yaml file (in this case, the target ref).

vdemeester:
@clnperez I really like the file, but in the current proposal there is two things that I feel could be better

- `cli` UX, a flag should not change the meaning of the positional args ; e.g. in the current design, having `--file=true` makes the positionnal arg being either a list (from the manifest store) or a file.
- the `list` is a image reference, and I think even with the file, we should specify that image reference as positional arguments. It would make the file not tied to the image reference pushed — it would be possible to do something like : `docker manifest push --file=foo.yml vdemeester/foo && docker manifest push --file=foo.yml grc.io/vdemeester/foo` — and the command would not be far from the current usage (we just add the possibility to pass a `file`)

So my proposal is to make the file something like the following : 

```yaml
image: test/hello-world:latest
manifests:
    - image: test/hello-world-ppc64le:latest
      platform:
          architecture: ppc64le
    - image: test/hello-world-amd64:latest
      platform:
          architecture: amd64
          os: linux
    - image: test/hello-world-s390x:latest
      platform:
          architecture: s390x 
          os: linux
          osversion: 1.1
          variant: xyz
          osfeatures: [a,b,c]
```

And have the cli being `docker manifest push --file=myfile.yaml list`.

@thaJeztah wdyt ?

dnephin:
I agree with @vdemeester

clnperez:
@vdemeester can you clarify: `the file not tied to the image reference pushed`...? The image reference pushed is the first line in the file the way the manifest tool works, and the way I intended this one to work as well. What's the meaning of `vdemeester/foo` in your example? Would it overwrite the target image on the first line of the file?

vdemeester:
@clnperez yes, my sugestion would be to not have the `image: …` first line in the file (so just listing in the manifest file the manifests). 

I updated my example above as I realized I did not remove that first line 😅.

clnperez:
Thanks @vdemeester. That's a simple enough code change. I'd like to hear from some others before we get attached. :D

cli/command/manifest/push.go
```diff
@@ -271,3 +291,68 @@ func mountBlobs(ctx context.Context, client registryclient.RegistryClient, ref r
...
+	if ym.Platform.Architecture != "" {
+		manifest.Platform.Architecture = ym.Platform.Architecture
+	}
+	if len(ym.Platform.OSFeatures) != 0 {
+		manifest.Platform.OSFeatures = ym.Platform.OSFeatures
+	}
+}
+
+func getYamlManifestList(yamlFile string) (yamlManifestList, error) {
+	var yamlInput yamlManifestList
```
silvin-lubecki: nit: you can move the var declaration closer to where it is used.

tianon:
I think it's relevant to bring up https://github.com/estesp/manifest-tool/issues/23 (multiple tags pushed at once; https://github.com/estesp/manifest-tool/pull/32) in the context of removing `image:` from the YAML itself -- being able to simply `docker manifest push --file xyz.yml abc:1.2.3 abc:1.2 abc:1 abc:latest` would be great (especially if it's efficient about skipping the expensive digest lookups and cross-blob mounts).

It probably won't be very useful for pushing to multiple registries though (hub+gcr.io for example), right? (since the manifest list image references still need to be on the same registry for cross-repository blob-mounting)

https://github.com/estesp/manifest-tool/issues/22 is also relevant for the official images use-case -- we have to wait to push the manifest lists until the full set of architectures have built (see https://github.com/docker-library/official-images/issues/3835), so having `--ignore-missing` helps us to be able to push a manifest list in the meantime, but it's only a workaround for a much larger problem in the way manifest list objects have to be created and updated (namely, the race conditions around "updating" a manifest list without API support for adding/updating entries directly to an existing tag's manifest list in an idempotent way at the registry level, which would allow us to build/update manifest lists more fluidly as architectures build and push like they were before manifest lists were implemented and we were single-arch).

It'd be neat to be able to specify this YAML content without needing a file on-disk, but that's more of a minor annoyance than a serious issue (since it's a matter of garbage collection after we build the appropriate YAML, push the bits, then clean up afterwards).

clnperez:
@silvin-lubecki nits taken into account, thanks :)

@vdemeester i'll have to redo this again if we go with https://github.com/docker/cli/pull/1156. Unless we want to merge this one first and make @dmcgowan rebase his. ;) 

for now though -- i can't get the vendor bit fixed
`make -f docker.Makefile vendor` ?

clnperez:
figured out vendor. i think i had switched a commit. PTAL

clnperez:
Rebased against pull/1156

clnperez:
ping @thaJeztah @vdemeester

clnperez:
Rebased. PTAL.

clnperez:
paging @thaJeztah & @vdemeester

cli/command/manifest/push.go
```diff
@@ -15,13 +16,16 @@ import (
 	"github.com/docker/distribution/manifest/schema2"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/registry"
+
```
vdemeester: shouldn't be there :angel:

silvin-lubecki (dismissed):
LGTM, thank you @clnperez !

silvin-lubecki:
@thaJeztah are you ok to merge this PR?

clnperez:
ping @thaJeztah . would love to close this :angel:

duglin:
@dnephin @vdemeester can chance of merging this one?

clnperez:
ping @thaJeztah. I saw that https://github.com/docker/cli/pull/1252 was merged so I have a sliver of hope that someone will take a look at this again.

olljanat:
Nice is very good idea. Would be nice to get this merged before #1355

grooverdan:
Nice work @clnperez, great way to create manifests by yaml definition.

PR #870, opened 2018-02-08
Improve reference docs for volumes

mdlinville:
Combine `--mount` and `-v` examples. Remove dangerous info about creating the Docker daemon from within the host as this seems like a super edge case and not something that should be in the mainline reference material.

Fixes https://github.com/docker/docker.github.io/issues/5941

newborn foal

### DIFF
--- a/docs/reference/commandline/run.md
+++ b/docs/reference/commandline/run.md
@@ -242,26 +242,50 @@ For the `overlay2` storage driver, the size option is only available if the
 backing fs is `xfs` and mounted with the `pquota` mount option.
 Under these conditions, user can pass any size less than the backing fs size.
 
-### Mount tmpfs (--tmpfs)
+### Mount tmpfs (--mount type=tmpfs)
+
+> **Note**: The `--tmpfs` flag still works, but `--mount type=tmpfs` is
+> preferred.
 
 ```bash
-$ docker run -d --tmpfs /run:rw,noexec,nosuid,size=65536k my_image
+$ docker run --rm -it \
+  --mount type=tmpfs,tmpfs-size=65536k,tmpfs-mode=600,dst=/run \
+  ubuntu:latest
 ```
 
-The `--tmpfs` flag mounts an empty tmpfs into the container with the `rw`,
-`noexec`, `nosuid`, `size=65536k` options.
+This example mounts a `tmpfs` filesystem 65536k in size into `/run` inside the
+container, with mode 600 (`rw-------`). Because the type of `tmpfs`, the file's
+contents of the file are not written to the container and are only written to a
+block device on the Docker host if memory is low enough to trigger swap usage.
 
-### Mount volume (-v, --read-only)
+### Mount volume (--mount, -v)
+
+The `--mount` flag allows you to mount volumes, host-directories and `tmpfs`
+mounts in a container.
+
+The `--mount` syntax is preferred. See
+[Differences between `-v` and `--mount` behavior](https://docs.docker.com/storage/volumes/#differences-between--v-and---mount-behavior)
+for more information. These examples show both syntax choices.
+
+#### Bind-mount the current working directory
+
+```bash
+$ docker run --mount src=`pwd`,dst=`pwd` -w `pwd` -i -t  ubuntu pwd
+```
 
 ```bash
 $ docker  run  -v `pwd`:`pwd` -w `pwd` -i -t  ubuntu pwd
 ```
 
-The `-v` flag mounts the current working directory into the container. The `-w`
+This example mounts the current working directory into the container. The `-w`
 lets the command being executed inside the current working directory, by
 changing into the directory to the value returned by `pwd`. So this
 combination executes the command using the container, but inside the
 current working directory.
+
+#### Populate the Docker host with the contents of a directory in the  container
+
+> **Note**: This only works with the `-v` syntax.
 
 ```bash
 $ docker run -v /doesnt/exist:/foo -w /foo -i -t ubuntu bash
@@ -272,6 +296,12 @@ will automatically create this directory on the host for you. In the
 example above, Docker will create the `/doesnt/exist`
 folder before starting your container.
 
+#### Use a read-only bind mount
+
+```bash
+$ docker run --mount src=/icanwrite,dst=/icanwrite,readonly busybox touch /icanwrite/here
+```
+
 ```bash
 $ docker run --read-only -v /icanwrite busybox touch /icanwrite/here
 ```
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/run.md
```diff
@@ -242,27 +242,50 @@ For the `overlay2` storage driver, the size option is only available if the
...
 ```bash
-$ docker run -d --tmpfs /run:rw,noexec,nosuid,size=65536k my_image
+$ docker run --rm -it \
+  --mount type=tmpfs,tmpfs-size=65536k,tmpfs-mode=600,dst=/run \
+  ubuntu:latest
 ```
 
-The `--tmpfs` flag mounts an empty tmpfs into the container with the `rw`,
-`noexec`, `nosuid`, `size=65536k` options.
+This example mounts an empty file 65536k in size into `/run` inside the
```
thaJeztah: It's mounting a tmpfs filesystem (not a file) at `/run`

docs/reference/commandline/run.md
```diff
@@ -242,27 +242,50 @@ For the `overlay2` storage driver, the size option is only available if the
...
+container, with mode 600 (`rw-------`). Because the type of `tmpfs`, the file's
+contents of the file are never written to a block device within the container
```
thaJeztah: IIRC, `tmpfs` is still being written to disk if there's insufficient memory, and memory is swapped to disk. Only `ramfs` is never written to disk (which is used for secrets and configs, but not supported by `--mount`)

docs/reference/commandline/run.md
```diff
@@ -242,27 +242,50 @@ For the `overlay2` storage driver, the size option is only available if the
 backing fs is `xfs` and mounted with the `pquota` mount option.
 Under these conditions, user can pass any size less than the backing fs size.
 
-### Mount tmpfs (--tmpfs)
+### Mount tmpfs (--mount type=tmpfs)
+
+> **Note**: The `--tmpfs` flag is deprecated.
```
thaJeztah: I think it's not officially deprecated yet (let me check)

mistyhacks:
@thaJeztah PTAL again

docs/reference/commandline/run.md
```diff
@@ -242,27 +242,51 @@ For the `overlay2` storage driver, the size option is only available if the
...
 ```
 
-The `-v` flag mounts the current working directory into the container. The `-w`
+This example mounts the current working directory into the container. The `-w`
 lets the command being executed inside the current working directory, by
 changing into the directory to the value returned by `pwd`. So this
 combination executes the command using the container, but inside the
 current working directory.
 
+#### Populate the Docker host with the contents of a directory in the  container
```
thaJeztah: The `-v <host-path>:<container-path>` syntax only creates the _directory_ on the host, but not copy the content from the container to the host. If the container has content at the given location, but the path on the host wasn't found, docker creates an empty directory on the host, and bind mounts that directory over the path in the container.

docs/reference/commandline/run.md
```diff
@@ -272,6 +296,12 @@ will automatically create this directory on the host for you. In the
 example above, Docker will create the `/doesnt/exist`
 folder before starting your container.
 
+#### Use a read-only bind mount
+
+```bash
+$ docker run --mount src=/icanwrite,dst=/icanwrite,readonly busybox touch /icanwrite/here
```
thaJeztah: Looks like we're mixing two things in the section; the existing example was about a read-only container, and using an (anonymous) volume to make parts of the container writable

docs/reference/commandline/run.md
```diff
@@ -315,31 +337,26 @@ docker run -v c:\foo:c: ...
...
-`--volume` flag, but uses a different syntax. For in-depth information on the
-`--mount` flag, and a comparison between `--volume` and `--mount`, refer to
-the [service create command reference](service_create.md#add-bind-mounts-or-volumes).
-
-Even though there is no plan to deprecate `--volume`, usage of `--mount` is recommended.
-
-Examples:
+This technique is useful in CI/CD set-ups where you need to test using a static
+Docker binary. This allows the container to create and manipulate the daemon on
+its host. This assumes that `/path/to/static-docker-binary` does not exist on
```
thaJeztah: This won't work; if `/usr/bin/docker` exists in the container, but `/path/to/static-docker-binary` doesn't (in the `-v` case), you'll end up with docker attempting to bind-mount a directory (`/path/to/static-docker-binary`) on top of a file/binary (`/usr/bin/docker`) inside the container.

docs/reference/commandline/run.md
```diff
@@ -315,31 +337,26 @@ docker run -v c:\foo:c: ...
...
+the Docker host but does exist within the container.
 
 ```bash
-$ docker run --read-only --mount type=volume,target=/icanwrite busybox touch /icanwrite/here
+$ docker run -t -i --mount src=/var/run/docker.sock,dst=/var/run/docker.sock --mount src=/path/to/static-docker-binary,dst=/usr/bin/docker busybox sh
```
thaJeztah: I don't think this example is very realistic; if I have both the _binary_ and access to the socket on the host, I could just as well run the binary on the host directly.

A more common scenario is an image that already contains a (static) binary, and is used to connect to the host daemon (http://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/)

docs/reference/commandline/run.md
```diff
@@ -315,31 +337,26 @@ docker run -v c:\foo:c: ...
...
 ```
 
 ```bash
-$ docker run -t -i --mount type=bind,src=/data,dst=/data busybox sh
+$ docker run -t -i -v /var/run/docker.sock:/var/run/docker.sock -v /path/to/static-docker-binary:/usr/bin/docker busybox sh
```
thaJeztah: For mounting a socket, we should recommend using `--mount`; using `-v` is especially troublesome when mounting files, because if the file is not found, docker creates a directory in its place, mounts that (see other comments above)

docs/reference/commandline/run.md
```diff
@@ -315,31 +337,26 @@ docker run -v c:\foo:c: ...
...
 
-### Add bind mounts or volumes using the --mount flag
-
-The `--mount` flag allows you to mount volumes, host-directories and `tmpfs`
-mounts in a container.
-
-The `--mount` flag supports most options that are supported by the `-v` or the
-`--volume` flag, but uses a different syntax. For in-depth information on the
-`--mount` flag, and a comparison between `--volume` and `--mount`, refer to
-the [service create command reference](service_create.md#add-bind-mounts-or-volumes).
```
thaJeztah: Should we keep a link to this section? (as it provides a handy overview of the options for each type)?

docs/reference/commandline/run.md
```diff
@@ -315,31 +337,26 @@ docker run -v c:\foo:c: ...
...
+$ docker run -t -i --mount src=/var/run/docker.sock,dst=/var/run/docker.sock --mount src=/path/to/static-docker-binary,dst=/usr/bin/docker busybox sh
 ```
 
 ```bash
-$ docker run -t -i --mount type=bind,src=/data,dst=/data busybox sh
+$ docker run -t -i -v /var/run/docker.sock:/var/run/docker.sock -v /path/to/static-docker-binary:/usr/bin/docker busybox sh
 ```
 
+See also
+[get the linux binary]( https://docs.docker.com/engine/installation/binaries/#/get-the-linux-binary)),
```
thaJeztah: Perhaps instead, we could show running the official `docker` image, and bind-mount the socket into that

vdemeester:
ping @thaJeztah

PR #874, opened 2018-02-12
Replace go-bindata with esc

### LINKED ISSUE #864: Remove go-bindata
The author of go-bindata has removed their github account which has been replaced by another author. It seems from this file that you do a go get and not vendor go-bindata directly: 

https://github.com/docker/cli/blob/819df0ebf600560698cd127a961f3d7753f59278/dockerfiles/Dockerfile.dev#L14

What it's generating: https://github.com/docker/cli/blob/6686ada6a4cf4e872af4d9bf3f61c387be934d87/cli/compose/schema/schema.go

Here are some alternatives that accomplish the same job: 
https://github.com/rakyll/statik
https://github.com/gobuffalo/packr

Related tweet and discussion: https://twitter.com/francesc/status/961249107020001280

dnephin:
Fixes #864

The new binary is 13.5k larger, so not too bad.
The CLI, and generated API are quite similar so it's easy to swap them.


dnephin:
Fixed the test failures. The new code expects absolute paths so I had to add a `/`

Issue #37, opened 2017-05-07
unit tests on commands


vdemeester:
Moving initial `docker/docker` issue https://github.com/moby/moby/issues/31217 here  

This issue is to keep track of unit test for the cli (i.e. `cli/commands` package and sub-packages).

- [x] bundlefile
- [x] checkpoint @adshmh 
- [x] container (fakeClient added in #61) @vdemeester
- [x] formatter
- [x] idresolver @adshmh
- [x] image @icapurro moby/moby#32248 
- [x] inspect
- [x] network (fakeClient added in #132) @ksouf
- [x] node
- [x] plugin @adshmh 
- [x] prune
- [x] registry
- [x] secret @adshmh 
- [x] service
- [x] stack @adshmh 
- [x] swarm
- [x] system
- [x] task
- [x] trust
- [x] volume (https://github.com/moby/moby/pull/31124)

The `swarm`, `node` and `volume` packages are example on how to write those tests (but they probably can be enhance, and enhancement are always welcome ).

- In order to write tests for these, you'll have to create more builder (`cli/internal/test/builders`) for objects that are not yet there.
- You can also use `golden` files (see examples)
- For each of these package, there is some integration tests (in `integration-cli`) that could be removed. I tried to mark them when I saw them, but it's not complete yet. I'll try to update this issue with a list of integration test that could be gone for each packages.

dnephin:
We may need to move some of the `fakeClient` into `cli/internal/test/` and export the struct and fields. There are some packages (stack, and service at least) that use parts of other clients.

jeremcs:
Some items on the list seem to be deprecated as they do not appear when setting the DOCKER_HIDE_LEGACY_COMMANDS variable:
- bundlefile
- inspect
- prune
- registry
- task

Should we remove them from the list ?

(cc @kelu27)

vdemeester:
@jeremcs I would say not (because some of them won't ever be removed).

clook:
Starting working on service (https://github.com/docker/cli/issues/201)

adshmh:
#dibs on checkpoint package.

Zebrilee:
hi @vdemeester ,
I was working on prune test in the last hackergarten but I saw the commit two days ago which modify the [prune.go file and add a prune_test.go too](https://github.com/docker/cli/tree/master/cli/command/system). 
I believe this part is ok or is there something else to do?

vdemeester:
@Zebrilee oh right yeah, I should mark it on the issue

[...]

PR #878, opened 2018-02-14
Add unit tests to plugin create/remove/enable/disable commands

adshmh:
add tests to plugin create/remove/enable/disable commands. Part of work on #37

**- What I did**
Added unit tests to several plugin commands (create/remove/enable/disable)

**- How I did it**
Added tests in create_test.go, remove_test.go, enable_test.go, disable_test.go

**- How to verify it**
Run the test suite:
make -f docker.Makefile test
Added unit tests to plugin create/remove/enable/disable commands.

### DIFF
--- a/cli/command/plugin/create_test.go
+++ b/cli/command/plugin/create_test.go
@@ -1,0 +1,114 @@
+package plugin
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"testing"
+
+	"github.com/docker/cli/internal/test"
+	"github.com/docker/cli/internal/test/testutil"
+	"github.com/docker/docker/api/types"
+	"github.com/gotestyourself/gotestyourself/fs"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCreateErrors(t *testing.T) {
+
+	testCases := []struct {
+		args          []string
+		expectedError string
+	}{
+		{
+			args:          []string{},
+			expectedError: "requires at least 2 arguments",
+		},
+		{
+			args:          []string{"INVALID_TAG", "context-dir"},
+			expectedError: "invalid",
+		},
+		{
+			args:          []string{"plugin-foo", "nonexistent_context_dir"},
+			expectedError: "no such file or directory",
+		},
+	}
+
+	for _, tc := range testCases {
+		cli := test.NewFakeCli(&fakeClient{})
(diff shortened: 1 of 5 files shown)


cli/command/plugin/create_test.go
```diff
@@ -0,0 +1,150 @@
...
+	tmpDir, err := ioutil.TempDir("", "plugin-create-test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	err = setupPluginDir(tmpDir, []byte("invalid-config-contents"))
+	if err != nil {
+		t.Fatal(err)
+	}
```
dnephin: This can be do with (`gotestyourself/fs`):

```go
tmpDir := fs.NewDir(t, "plugin-create-test",
    fs.WithDir("rootfs",
        fs.WithFile("config.json", "invalid-config-contents")))
defer tmpDir.Remove()
```

Just as an FYI, I think this works for now.

cli/command/plugin/create_test.go
```diff
@@ -0,0 +1,150 @@
...
+	cli := test.NewFakeCli(&fakeClient{
+		pluginCreateFunc: func(createContext io.Reader, createOptions types.PluginCreateOptions) error {
+			return nil
+		},
+	})
+
+	cmd := newCreateCommand(cli)
+	cmd.SetArgs([]string{"plugin-foo", tmpDir})
+	assert.NoError(t, cmd.Execute())
+	assert.Equal(t, "plugin-foo", strings.TrimSpace(cli.OutBuffer().String()))
```
dnephin: I think we should check the stdout without `TrimSpace()`. The trailing newline is important.

There's a few instances of this below as well.

cli/command/plugin/create_test.go
adshmh: Thank you for the review. The `gotestyourself` looks very interesting (just taking a look at the above example, it reads much better than the function I wrote for these tests). 

If you are OK with such a change, I can change to use `gotestyourself/fs` here.

dnephin: That would be great, thanks!

cli/command/plugin/create_test.go
adshmh: Thanks for the review. I have updated the PR accordingly.

PR #879, opened 2018-02-14
Replace testify assertions with gotestyourself/assert

dnephin:
We already use `gotestyourself` packages extensively in `docker/cli` and [`testify/assert` has many issues](https://github.com/test-go/testify/issues/13), so replace testify with `gotestyourself/assert` ([godoc](https://godoc.org/github.com/gotestyourself/gotestyourself/assert)).

The automated migration commit was done using [gty-migrate-from-testify](https://github.com/gotestyourself/gotestyourself/pull/36).


thaJeztah:
Are there any plans on picking up the effort to merge these projects? https://github.com/test-go/testify/issues/13#issuecomment-353411018

dnephin:
At this point, I don't think so. At the time `gotestyourself/assert` wasn't merged yet and I thought there might be some shared effort. Shortly after it seemed like very little was happening with `test-go`. Now I think there's enough in `gotestyourself/assert` that it would be a lot less work to add a few more comparisons than to try and cleanup what's in testify.

cli/command/config/ls_test.go
```diff
@@ -125,15 +126,15 @@ func TestConfigListWithFormat(t *testing.T) {
...
 	golden.Assert(t, cli.OutBuffer().String(), "config-list-with-format.golden")
 }
 
 func TestConfigListWithFilter(t *testing.T) {
 	cli := test.NewFakeCli(&fakeClient{
 		configListFunc: func(options types.ConfigListOptions) ([]swarm.Config, error) {
-			assert.Equal(t, "foo", options.Filters.Get("name")[0])
-			assert.Equal(t, "lbl1=Label-bar", options.Filters.Get("label")[0])
+			assert.Check(t, is.Equal("foo", options.Filters.Get("name")[0]))
+			assert.Check(t, is.Equal("lbl1=Label-bar", options.Filters.Get("label")[0]))
```
thaJeztah: Hm, I find `assert.NoError(...)` more easy on the eye than `assert.Check(....)`; same as `assert.Equal(..)`; for example:

```go
assert.Equal(t, "lbl1=Label-bar", options.Filters.Get("label")[0])
```

vs

```go
assert.Check(t, is.Equal("lbl1=Label-bar", options.Filters.Get("label")[0]))
```

The last one I have to read until position 20 (with indentation, position 32) to see what kind of check is done, and the `is.Equal()` add another level of nesting, which makes it quite a bit harder to read.

Basically `assert.Something(` enables me to "scan" the code, and at a glance grasp what steps are taken.

Was this a deliberate design choice in gotestyourself?

dnephin: > I find assert.NoError(...) more easy on the eye than assert.Check(....)

`assert.NoError()` (in testify) is almost always wrong because it only fails the test, but doesn't end execution immediately. If you don't end execution immediately on an error you will likely hit a panic on a following line. This is one of the many issues that should be fixed.

`gotestyourself` does have `assert.NilError()`. The automated migration tool preserves the existing behaviour so we ended up with a few `assert.Check(t, err)`, but these should likely be converted to `assert.NilError(t, err)` in a follow up.

> Was this a deliberate design choice in gotestyourself?

Yes, https://github.com/gotestyourself/gotestyourself/issues/49#issuecomment-362436026.

I think we should add more `assert.Something()`, but I want to make sure we add the right ones and don't clutter the interface with infrequently used, or weak assertions.

`assert.Check(t, is.Equal())` is an interesting case. We need a way to support assertions that use both `T.Fail` and `T.FailNow`. There are only so many ways this can be done. `testify` solved it with code generation and using a separate package. `gotestyourself` solves it by splitting the failure (`assert.Check` vs `assert.Assert`) from the comparison.

The convenience assertions added to the `assert` package all use `FailNow` because that should be the more common case. We could add a `assert.CheckEqual()` but I don't really see that as an improvement over `assert.Check(t, isEqual())`. Using `assert.Equal()` ([which already exists](https://godoc.org/github.com/gotestyourself/gotestyourself/assert#Equal)) should be correct in most cases.

silvin-lubecki: Maybe then you nay use `testify/require` instead? I think that require.NoError() terminates the execution.

dnephin: Yes, you could use `require.NoError()`, but this PR should make it clear that it's too easy to do the wrong thing with testify. It is just one of [the many problems with testify](https://github.com/test-go/testify/issues/13). require should be called assert, and assert should be something like "check".

`gotestyourself` makes it a lot easier to do the right thing by making it the most convenient and fixing the terminology.

dnephin: @thaJeztah @silvin-lubecki 

https://github.com/gotestyourself/gotestyourself/pull/71 adds some more top level asserts for handling errors.

thaJeztah: @dnephin looking good! I think we should move this forward (sorry for stalling this). Looks like this'll need a massive rebase (or was it all done automated?)

dnephin: It was mostly automated, so it should be easy to just re-run that automation.

dnephin:
rebased and green

cli/command/bundlefile/bundlefile_test.go
```diff
@@ -37,7 +38,7 @@ func TestLoadFileSyntaxError(t *testing.T) {
 	}`)
 
 	_, err := LoadFile(reader)
-	assert.EqualError(t, err, "JSON syntax error at byte 37: invalid character 'u' looking for beginning of value")
+	assert.Check(t, is.Error(err, "JSON syntax error at byte 37: invalid character 'u' looking for beginning of value"))
```
thaJeztah: Is it possible to automate changing these to `assert.Error()` (https://github.com/gotestyourself/gotestyourself/pull/71), or do you want to keep that for a follow-up ?

dnephin: It's possible to automate the change, but it's not the same as gotestyourself/gotestyourself#71.

The current automated migration preserves behaviour, so `testify/require` -> `Assert`, `testify/assert` -> `Check`. 

I can write another migration that converts specific checks to the canonical `Assert`, but that is a change in behaviour (Fail -> FailNow), so I'd like to do that in a separate PR so that we can more carefully review the change.

PR #881, opened 2018-02-16
Add container_* reference command to map root level ones

vdemeester:
I'm not sure how to do that, but it bother me each time to see `docker ps` reference documentation *complete* but not `docker container ls` (without link even).

This is probably the wrong solution, but opening the PR to discuss how we could fix it.

### DIFF
--- a/docs/reference/commandline/container_ls.md
+++ b/docs/reference/commandline/container_ls.md
@@ -1,0 +1,434 @@
+---
+title: "container ls"
+description: "The container ls command description and usage"
+keywords: "container, running, list"
+---
+
+<!-- This file is maintained within the docker/cli GitHub
+     repository at https://github.com/docker/cli/. Make all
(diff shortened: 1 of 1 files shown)


thaJeztah:
Oh! Was just thinking about that this morning again when working on that YAML bug.

Basically, what I _think_ we should try is;

- rename all the `.md` files to their canonical name (t.b.d. what we use as "canonical")
- either make the docs use redirects (non-canonical -> canonical), or have a "stub" page for non-canonical names, and refer/link people to the other one
- have the YAML generator look for "Examples" and "Description" in the canonical documents, and copy that to the other variations

Most importantly; we should not have two copies of the same content in the repository, but it's okay to have the YAML generator _duplicate_ the content on generation

vdemeester:
Closing this for now :angel:

PR #882, opened 2018-02-16
Carry #553 : Add option to pull images quietly

vdemeester:
fixes https://github.com/moby/moby/issues/13588

Add `--quiet` to the `docker image pull` subcommand that will pull
the image quietly.

```
$ docker pull -q golang:latest
$ docker pull -q golang
Using default tag: latest
```

Carrying #553 cc @pramodhkp

🐯 


cpuguy83:
Do we have a good way to add a test for this?

dnephin:
Yes, it should be possible to unit test this with the fake client

dnephin:
Design LGTM

thaJeztah:
@dnephin PTAL

dnephin:
It wouldn't hurt to have a unit test for the new flag since it seems our coverage is low in this area

thaJeztah:
looks like a test was added

vdemeester:
> looks like a test was added
Well, it doesn't test anything really.. Having a bit of trouble with `jsonmessage` :stuck_out_tongue_closed_eyes: :sweat_smile:

vdemeester:
oh dang it, forgot about that one.

vdemeester:
@thaJeztah updated :wink:

e2e/image/pull_test.go
```diff
@@ -28,6 +30,13 @@ func TestPullWithContentTrust(t *testing.T) {
 	golden.Assert(t, result.Stdout(), "pull-with-content-trust.golden")
 }
 
+func TestPullQuiet(t *testing.T) {
+	result := icmd.RunCommand("docker", "pull", "--quiet", fixtures.AlpineImage)
+	result.Assert(t, icmd.Success)
+	assert.Check(t, is.Equal(result.Stdout(), ""))
```
thaJeztah: Hm, I don't think it's correct that _nothing_ is output; it should print the image name and/or sha after pull completed. Perhaps we can combine, i.e.:

```
busybox:latest@sha256:58ac43b2cc92c687a32c8be6278e50a063579655fe3090125dcb2af0ff9e1a64
```

Sodki: Wouldn't that defeat the purpose of a `quiet` option?

vdemeester:
ping :angel:

thaJeztah:
Looks like it's missing a newline after the output;

```
/go/src/github.com/docker/cli # docker pull -q nginx:alpine
docker.io/library/nginx:alpine/go/src/github.com/docker/cli # 
```

thaJeztah:
ping @dnephin @silvin-lubecki PTAL

cli/command/image/pull.go
```diff
@@ -67,15 +69,18 @@ func runPull(cli command.Cli, opts pullOptions) error {
...
+		err = imagePullPrivileged(ctx, cli, imgRefAndAuth, opts)
 	}
 	if err != nil {
 		if strings.Contains(err.Error(), "when fetching 'plugin'") {
 			return errors.New(err.Error() + " - Use `docker plugin install`")
 		}
 		return err
 	}
+	if opts.quiet {
+		fmt.Fprintln(cli.Out(), imgRefAndAuth.Reference().String())
```
cpuguy83: We should just do this even without the quiet flag.
WDYT @thaJeztah @vdemeester?

vdemeester: Without quiet 
```
latest: Pulling from vdemeester/moby-dev-criu
c73ab1c6897b: Already exists 
1ab373b3deae: Already exists 
b542772b4177: Already exists 
57c8de432dbe: Already exists 
c81227e1ec90: Already exists 
c368dcc533d6: Already exists 
a0037d23043c: Already exists 
f64db6c96f10: Already exists 
6993efc377f4: Already exists 
963f9c7d7720: Pull complete 
Digest: sha256:9c5be10571181cd25ff13d9775c63017aeefc067df5eaa223375f9cf6b02d176
Status: Downloaded newer image for vdemeester/moby-dev-criu:latest
```

The reference already appears, in a different representation though so.. I wonder :stuck_out_tongue:

vdemeester:
ah right, I should update those

cpuguy83:
Indeed. With adding it, a script could even at least grab the last line.
Also just a little less to parse for a human as well :)

thaJeztah:
Yes, I'd be +1 to adding it in the non-quiet use as well (we should check what parts of the current output are printed to stderr, and what parts on stdout)

vdemeester:
@cpuguy83 @thaJeztah Updated, this add a new line to all `docker pull` like that.

docwhat:
The :tag@sha form would be helpful. It would have everything you could want in a compact reasonably computer parseable way. 

The docker.io should go away if it doesn’t match the image name.  That both `docker pull` and `docker run` should be able to use the returned string.

vdemeester:
> The docker.io should go away if it doesn’t match the image name. That both docker pull and docker run should be able to use the returned string.

@docwhat `docker run` and `docker pull` are able to run `docker.io/library/nginx:alpine` (it's just that docker has some default — hub — and special cases — library).

docwhat:
I grok that. But if I pull `nginx` then I can run `nginx`.

If I add the namespace and host then `docker run` will go pull it again (though it’ll be fast since all the layers are already here). 

Also `docker image ls` will show `nginx` without the namespace and host.

thaJeztah:
ping @vdemeester I forgot; what's the status on this one?

jregeimbal:
Bump, I've spent hours searching for a viable workaround to quiet down docker pull/push.  Nothing seems to work with Semaphore CI.  This would be a great feature to get in soon 👍

silvin-lubecki:
Please rebase @vdemeester 🦁

Gasol:
I'm on the same boat with @jregeimbal , It's very annoying to see the progress output in CI, We must press keycap `page down` twice or even more to ignore them over and over.

vdemeester:
@thaJeztah updated (and rebased) :wink:

cli/command/image/trust.go
```diff
@@ -202,7 +203,12 @@ func trustedPull(ctx context.Context, cli command.Cli, imgRefAndAuth trust.Image
 		if err != nil {
 			return err
 		}
-		if err := imagePullPrivileged(ctx, cli, updatedImgRefAndAuth, false, platform); err != nil {
+		if err := imagePullPrivileged(ctx, cli, updatedImgRefAndAuth, PullOptions{
+			all:      false,
```
thaJeztah: Not for this PR, but wondering if we need a validation step for `--all` somewhere (as it's being ignored here); i.e. is there a situation where I could've provided the `--all` flag, and no error is printed (but it's being ignored)

thaJeztah:
oh, dang; one test needs updating;

```
--- FAIL: TestNewPullCommandSuccess (0.00s)
    pull_test.go:87: assertion failed: 
        --- expected
        +++ actual
        @@ -1,2 +1,3 @@
        +Using default tag: latest
         docker.io/library/image:latest
         
```

steadysupply:
I can see `pull -q` made it into [19.03](https://docs.docker.com/engine/release-notes/#19030). Is there somewhere I can see when this will be released?

PR #891, opened 2018-02-20
Make sure we marshall version too…

vdemeester:
… otherwise the k8s controller might fail to parse the file as it will
think it's version 1.

### DIFF
--- a/cli/command/stack/kubernetes/loader.go
+++ b/cli/command/stack/kubernetes/loader.go
@@ -8,16 +8,30 @@ import (
 )
 
 type versionedConfig struct {
-	*composetypes.Config `yaml:",inline"`
-	Version              string
+	composetypes.Config
+	version string
+}
+
+func (c versionedConfig) MarshalYAML() (interface{}, error) {
+	services := map[string]composetypes.ServiceConfig{}
+	for _, service := range c.Services {
+		services[service.Name] = service
+	}
+	return map[string]interface{}{
+		"services": services,
+		"networks": c.Networks,
+		"volumes":  c.Volumes,
+		"secrets":  c.Secrets,
+		"configs":  c.Configs,
+		"version":  c.version,
+	}, nil
 }
 
 // LoadStack loads a stack from a Compose config, with a given name.
 func LoadStack(name, version string, cfg composetypes.Config) (*apiv1beta1.Stack, error) {
-	cfg.Filename = ""
 	res, err := yaml.Marshal(versionedConfig{
-		Version: version,
-		Config:  &cfg,
+		version: version,
+		Config:  cfg,
 	})
 	if err != nil {
 		return nil, err
--- a/cli/command/stack/kubernetes/loader_test.go
+++ b/cli/command/stack/kubernetes/loader_test.go
@@ -37,6 +37,7 @@ services:
     image: bar
   foo:
     image: foo
+version: "3.1"
 volumes: {}
 `),
 		},
--- a/cli/compose/types/types.go
+++ b/cli/compose/types/types.go
@@ -70,7 +70,7 @@ func (cd ConfigDetails) LookupEnv(key string) (string, bool) {
 
 // Config is a full compose file configuration
 type Config struct {
-	Filename string
+	Filename string `yaml:"-"`
 	Services []ServiceConfig
 	Networks map[string]NetworkConfig
 	Volumes  map[string]VolumeConfig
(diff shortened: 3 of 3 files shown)


cli/command/stack/kubernetes/loader.go
```diff
@@ -12,6 +12,15 @@ type versionedConfig struct {
 	Version              string
 }
 
+func (c versionedConfig) MarshalYAML() (interface{}, error) {
```
dnephin: I think the problem is that `composetypes.Config` is being used as an embeded struct, and it has a `MarshalYAML` so that is what is called when this `versionedConfig` struct is marshalled.

How about:
```diff
diff --git a/cli/command/stack/kubernetes/loader.go b/cli/command/stack/kubernetes/loader.go
index 14aec58914..66a4071792 100644
--- a/cli/command/stack/kubernetes/loader.go
+++ b/cli/command/stack/kubernetes/loader.go
@@ -8,17 +8,26 @@ import (
 )
 
 type versionedConfig struct {
-	*composetypes.Config `yaml:",inline"`
-	Version              string
+	composetypes.Config
+	Version string
 }
 
+// MarshalYAML makes Config implement yaml.Marshaller
 func (c versionedConfig) MarshalYAML() (interface{}, error) {
-	i, err := c.Config.MarshalYAML()
-	if err != nil {
-		return nil, err
+	m := map[string]interface{}{}
+	services := map[string]composetypes.ServiceConfig{}
+	for _, service := range c.Services {
+		s := service
+		s.Name = ""
+		services[service.Name] = s
 	}
-	i.(map[string]interface{})["version"] = c.Version
-	return i, nil
+	m["services"] = services
+	m["networks"] = c.Networks
+	m["volumes"] = c.Volumes
+	m["secrets"] = c.Secrets
+	m["configs"] = c.Configs
+	m["version"] = c.Version
+	return m, nil
 }
 
 // LoadStack loads a stack from a Compose config, with a given name.
@@ -26,7 +35,7 @@ func LoadStack(name, version string, cfg composetypes.Config) (*apiv1beta1.Stack
 	cfg.Filename = ""
 	res, err := yaml.Marshal(versionedConfig{
 		Version: version,
-		Config:  &cfg,
+		Config:  cfg,
 	})
 	if err != nil {
 		return nil, err
diff --git a/cli/compose/types/types.go b/cli/compose/types/types.go
index 2299871cfb..37ec1b5dd7 100644
--- a/cli/compose/types/types.go
+++ b/cli/compose/types/types.go
@@ -78,23 +78,6 @@ type Config struct {
 	Configs  map[string]ConfigObjConfig
 }
 
-// MarshalYAML makes Config implement yaml.Marshaller
-func (c *Config) MarshalYAML() (interface{}, error) {
-	m := map[string]interface{}{}
-	services := map[string]ServiceConfig{}
-	for _, service := range c.Services {
-		s := service
-		s.Name = ""
-		services[service.Name] = s
-	}
-	m["services"] = services
-	m["networks"] = c.Networks
-	m["volumes"] = c.Volumes
-	m["secrets"] = c.Secrets
-	m["configs"] = c.Configs
-	return m, nil
-}
-
 // ServiceConfig is the configuration of one service
 type ServiceConfig struct {
 	Name string `yaml:",omitempty"`
```

dnephin: Or we could make a type for `Config.Services` that has the `MarshalYAML` on it, add a `yaml:"-"` to `Config.Filename`, and then we should be able to just use `inline` here.

cli/command/stack/kubernetes/loader.go
```diff
@@ -8,16 +8,31 @@ import (
...
-	Version              string
+	composetypes.Config
+	Version string
+}
+
+func (c versionedConfig) MarshalYAML() (interface{}, error) {
+	m := map[string]interface{}{}
+	services := map[string]composetypes.ServiceConfig{}
+	for _, service := range c.Services {
+		s := service
```
silvin-lubecki: Why do you copy service here?

cli/command/stack/kubernetes/loader.go
```diff
@@ -8,16 +8,31 @@ import (
...
+		services[service.Name] = s
+	}
+	m["services"] = services
```
silvin-lubecki:
```go
return map[string]interface{}{
  "networks": c.Networks,
  "volumes": c.Volumes,
...
}, nil
```

cli/command/stack/kubernetes/loader.go
```diff
@@ -8,16 +8,30 @@ import (
 )
 
 type versionedConfig struct {
-	*composetypes.Config `yaml:",inline"`
-	Version              string
+	composetypes.Config
+	version string
+}
+
+func (c versionedConfig) MarshalYAML() (interface{}, error) {
```
dnephin: I think we'll be able to remove these `MarshalYAML` in the future by making `Config.Services` a type with a `MarshalYAML`.

PR #893, opened 2018-02-21
add support for mandatory variables to stack deploy

### LINKED ISSUE #805: Add support for ${VAR:?err} syntax for mandatory variables
Support for this notation was added in Docker Compose (https://github.com/docker/compose/pull/5531, and associated issue: https://github.com/docker/compose/issues/5505)

The same syntax should probably be supported for `docker stack deploy` to preserve compatibility with docker-compose files used by Docker Compose.

 @shin- @vdemeester @dnephin

adshmh:
This PR adds support for mandatory variables syntax to stack deploy. Fixes #805 

**- What I did**
Added support for mandatory variables syntax, i.e. ${VAR?error message} and ${VAR:?error message}, to stack deploy command.

**- How I did it**
Added the code required to handle the syntax in cli/compose/template/template.go, and the corresponding tests in cli/compose/template/template_test.go

**- How to verify it**
Run stack deploy command on a compose file that contains mandatory variable(s).

Added support for mandatory variables to stack deploy command.

### DIFF
--- a/cli/compose/template/template.go
+++ b/cli/compose/template/template.go
@@ -37,75 +37,72 @@ func Substitute(template string, mapping Mapping) (string, error) {
 	var err error
 	result := pattern.ReplaceAllStringFunc(template, func(substring string) string {
 		matches := pattern.FindStringSubmatch(substring)
-		groups := make(map[string]string)
-		for i, name := range pattern.SubexpNames() {
-			if i != 0 {
-				groups[name] = matches[i]
-			}
+		groups := matchGroups(matches)
+		if escaped := groups["escaped"]; escaped != "" {
+			return escaped
 		}
 
 		substitution := groups["named"]
 		if substitution == "" {
 			substitution = groups["braced"]
 		}
-		if substitution != "" {
-			// Soft default (fall back if unset or empty)
-			if strings.Contains(substitution, ":-") {
-				name, defaultValue := partition(substitution, ":-")
-				value, ok := mapping(name)
-				if !ok || value == "" {
-					return defaultValue
-				}
-				return value
-			}
 
-			// Hard default (fall back if-and-only-if empty)
-			if strings.Contains(substitution, "-") {
-				name, defaultValue := partition(substitution, "-")
-				value, ok := mapping(name)
-				if !ok {
-					return defaultValue
-				}
-				return value
-			}
+		switch {
 
-			if strings.Contains(substitution, ":?") {
-				name, errorMessage := partition(substitution, ":?")
-				value, ok := mapping(name)
-				if !ok || value == "" {
-					err = &InvalidTemplateError{Template: errorMessage}
-					return ""
-				}
-				return value
-			}
+		case substitution == "":
+			err = &InvalidTemplateError{Template: template}
+			return ""
 
-			if strings.Contains(substitution, "?") {
-				name, errorMessage := partition(substitution, "?")
-				value, ok := mapping(name)
-				if !ok {
-					err = &InvalidTemplateError{Template: errorMessage}
-					return ""
-				}
-				return value
+		// Soft default (fall back if unset or empty)
+		case strings.Contains(substitution, ":-"):
+			name, defaultValue := partition(substitution, ":-")
+			value, ok := mapping(name)
+			if !ok || value == "" {
+				return defaultValue
 			}
+			return value
 
-			// No default (fall back to empty string)
-			value, ok := mapping(substitution)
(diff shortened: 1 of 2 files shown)


thaJeztah:
Linter is complaining;

```
cli/compose/template/template.go:36::warning: cyclomatic complexity 17 of function Substitute() is high (> 16) (gocyclo)
```

thaJeztah:
Otherwise, change LGTM; tested this, and looks to work as expected

cli/compose/template/template.go
```diff
@@ -69,6 +69,26 @@ func Substitute(template string, mapping Mapping) (string, error) {
 				return value
 			}
 
+			if strings.Contains(substitution, ":?") {
+				name, errorMessage := partition(substitution, ":?")
+				value, ok := mapping(name)
+				if !ok || value == "" {
+					err = &InvalidTemplateError{Template: errorMessage}
```
dnephin: This message should include something about "missing value" or "value required"

cli/compose/template/template_test.go
```diff
@@ -81,3 +81,63 @@ func TestNonAlphanumericDefault(t *testing.T) {
...
+		{
+			template:      "not ok ${UNSET_VAR?}",
+			expectedError: "",
+		},
+	}
+
+	for _, tc := range testCases {
+		_, err := Substitute(tc.template, defaultMapping)
+		assert.Error(t, err)
+		assert.IsType(t, &InvalidTemplateError{tc.expectedError}, err)
```
dnephin: This should check the error message as well as the type.

dnephin:
The commit I pushed should fix the lint error

cli/compose/template/template_test.go
adshmh: Thank you for the review. I have updated the PR. Also removed the `tc.expectedError` from the `InvalidTemplateError` as it seemed unnecessary.

cli/compose/template/template_test.go
```diff
@@ -81,3 +89,64 @@ func TestNonAlphanumericDefault(t *testing.T) {
...
 }
+
+func TestMandatoryVariableErrors(t *testing.T) {
+	testCases := []struct {
+		template      string
+		expectedError string
+	}{
+		{
+			template:      "not ok ${UNSET_VAR:?Mandatory Variable Unset}",
+			expectedError: "Mandatory Variable Unset",
```
dnephin: I don't understand this test case. Why would anyone put a default if the value is required?

I don't think the error message should contain the unused default string, it should say something like "required variable UNSET_VAR is missing a value"

cli/compose/template/template_test.go
```diff
@@ -81,3 +89,64 @@ func TestNonAlphanumericDefault(t *testing.T) {
...
+		},
+		{
+			template:      "not ok ${BAR:?Mandatory Variable Empty}",
+			expectedError: "Mandatory Variable Empty",
+		},
+		{
+			template:      "not ok ${UNSET_VAR:?}",
+			expectedError: "",
```
dnephin: The error here should be something like "required variable UNSET_VAR is missing a value"

cli/compose/template/template_test.go
thaJeztah: For reference; this is what it looks like for `/bin/sh` and `Bash`:

```bash
echo ${UNSET_VAR:?Mandatory Variable Unset}
sh: UNSET_VAR: Mandatory Variable Unset

echo ${UNSET_VAR:?Mandatory Variable Unset}
bash: UNSET_VAR: Mandatory Variable Unset
```

cli/compose/template/template_test.go
adshmh: Sure. I am assuming you are referring to the error message in the test cases here, not changing the `Substitute` function, is that correct?

dnephin: No, I think the error from Substitute should be something like this:

```go
err = &InvalidTemplateError{
    Template: fmt.Sprintf("required variable %s is missing a value: %s", name, errorMessage),
}
```

For reference, this is the error in docker-compose: https://github.com/docker/compose/pull/5531/files#diff-b2372f2a81cc5911e04cc7df7d68675bR65

adshmh: Thanks for the clarification. I updated the PR (the reg. exp. also needed to be fixed to cover empty error messages)

thaJeztah:
Docs were added in https://github.com/docker/docker.github.io/pull/5826

PR #895, opened 2018-02-21
Simplify the marshaling of compose types.Config

vdemeester:
- Add `Version` to `types.Config`
- Add a new `Services` types (that is just `[]ServiceConfig`) and add
  `MarshalYAML` method on it.
- Clean other top-level custom marshaling as `Services` is the only one
  required.

Follow-up of https://github.com/docker/cli/pull/891 👼 

🦁 

### DIFF
--- a/cli/command/stack/loader/loader.go
+++ b/cli/command/stack/loader/loader.go
@@ -18,21 +18,21 @@ import (
 )
 
 // LoadComposefile parse the composefile specified in the cli and returns its Config and version.
-func LoadComposefile(dockerCli command.Cli, opts options.Deploy) (*composetypes.Config, string, error) {
+func LoadComposefile(dockerCli command.Cli, opts options.Deploy) (*composetypes.Config, error) {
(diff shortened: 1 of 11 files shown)


vdemeester:
The order of the marshaled `Config` changed because it's now using the order of the struct definition instead of natural order with the map.

thaJeztah:
Can we add some filler-code to fill up all the code that was removed? :joy:

cli/command/stack/loader/loader.go
```diff
@@ -46,7 +46,7 @@ func LoadComposefile(dockerCli command.Cli, opts options.Deploy) (*composetypes.
 		fmt.Fprintf(dockerCli.Err(), "Ignoring deprecated options:\n\n%s\n\n",
 			propertyWarnings(deprecatedProperties))
 	}
-	return config, configDetails.Version, nil
+	return config, nil
```
thaJeztah: :heart:

PR #896, opened 2018-02-21
Add --template-driver option for secrets/configs

thaJeztah:
relates to https://github.com/moby/moby/pull/33702

### DIFF
--- a/cli/command/config/create_test.go
+++ b/cli/command/config/create_test.go
@@ -82,14 +82,21 @@ func TestConfigCreateWithLabels(t *testing.T) {
 	}
 	name := "foo"
 
+	data, err := ioutil.ReadFile(filepath.Join("testdata", configDataFile))
+	assert.NoError(t, err)
+
+	expected := swarm.ConfigSpec{
+		Annotations: swarm.Annotations{
+			Name:   name,
+			Labels: expectedLabels,
+		},
+		Data: data,
+	}
+
 	cli := test.NewFakeCli(&fakeClient{
 		configCreateFunc: func(spec swarm.ConfigSpec) (types.ConfigCreateResponse, error) {
-			if spec.Name != name {
-				return types.ConfigCreateResponse{}, errors.Errorf("expected name %q, got %q", name, spec.Name)
-			}
-
-			if !reflect.DeepEqual(spec.Labels, expectedLabels) {
-				return types.ConfigCreateResponse{}, errors.Errorf("expected labels %v, got %v", expectedLabels, spec.Labels)
+			if !reflect.DeepEqual(spec, expected) {
+				return types.ConfigCreateResponse{}, errors.Errorf("expected %+v, got %+v", expected, spec)
 			}
 
 			return types.ConfigCreateResponse{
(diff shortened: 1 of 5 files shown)


cli/command/config/create_test.go
```diff
@@ -105,3 +105,32 @@ func TestConfigCreateWithLabels(t *testing.T) {
...
+	}
+	name := "foo"
+
+	cli := test.NewFakeCli(&fakeClient{
+		configCreateFunc: func(spec swarm.ConfigSpec) (types.ConfigCreateResponse, error) {
+			if spec.Name != name {
+				return types.ConfigCreateResponse{}, errors.Errorf("expected name %q, got %q", name, spec.Name)
+			}
+
+			if !reflect.DeepEqual(spec.Templating.Name, expectedDriver.Name) {
```
dnephin: isn't this just a comparing a string? so it could use `!=` ?

thaJeztah: hm, yes, copy-pasta'd from `TestSecretCreateWithDriver`. Let me update

thaJeztah:
Thinking if `templating-driver` or `template-driver` is the best name; let me know if `template-driver` is a better name

cli/command/config/create_test.go
```diff
@@ -105,3 +105,32 @@ func TestConfigCreateWithLabels(t *testing.T) {
 	assert.NoError(t, cmd.Execute())
 	assert.Equal(t, "ID-"+name, strings.TrimSpace(cli.OutBuffer().String()))
 }
+
+func TestConfigCreateWithTemplatingDriver(t *testing.T) {
```
cpuguy83: Test for config create without driver?

thaJeztah: There's a test above that doesn't have a driver, is that what you meant?

dnephin:
I like `template-driver`

thaJeztah:
Changed to `template-driver`

cli/command/config/create_test.go
cpuguy83: We should check and make sure Templating is nil in that case, I think?

dnephin: This would be done already if the test used `DeepEqual` on the entire spec instead of looking at specific fields. Maybe we should do that here as well.

thaJeztah:
hmf, secrets side is failing;

```
Error: expected {Annotations:{Name:foo Labels:map[]} Data:[115 101 99 114 101 116 95 102 111 111 95 98 97 114 10] Driver:<nil> Templating:<nil>}, got {Annotations:{Name:foo Labels:map[]} Data:[115 101 99 114 101 116 95 102 111 111 95 98 97 114 10] Driver:<nil> Templating:<nil>}
```

But I don't see a different:

```
{Annotations:{Name:foo Labels:map[]} Data:[115 101 99 114 101 116 95 102 111 111 95 98 97 114 10] Driver:<nil> Templating:<nil>}
{Annotations:{Name:foo Labels:map[]} Data:[115 101 99 114 101 116 95 102 111 111 95 98 97 114 10] Driver:<nil> Templating:<nil>}
```

PR #898, opened 2018-02-22
Print Stack API and Kubernetes versions in version command

silvin-lubecki:
**- What I did**
This pull request adds the Stack API resolution mechanism and print the selected version into the `version` command. It lays the foundation for a following one #899, which adds the v1beta2 version of the Stack API. It's extracted from the PR #854 . 

**- How I did it**

* Resolve Stack API using Kubernetes discovering API
* Refactor Kubernetes flags parsing

**- How to verify it**
Execute a `docker version` command against latest `docker for mac`, it should output at the end:
```
 Orchestrator:
   Kubernetes:  v1.9.2
   Stack:       v1beta1
```

*  Add Stack API and Kubernetes versions in version command

image


cli/command/system/version.go
```diff
@@ -48,16 +52,22 @@ Server:{{if ne .Platform.Name ""}} {{.Platform.Name}}{{end}}
    {{- end}}
   {{- end}}
  {{- end}}
-{{- end}}{{end}}`
+ {{- end}}{{- end}}
+ {{- if .KubernetesOK}}{{with .Kubernetes}}
+ Orchestrator:
+   Kubernetes:	{{.Kubernetes}}
+   Stack:	{{.StackAPI}}{{- end}}{{end}}`
```
vdemeester: What if there is multiple API version available ? Should we display sthg a bit like docker api versions (version negociated vs min version) or a list of version, of just the version used ?

kubernetes/check.go
```diff
@@ -0,0 +1,57 @@
...
+// StackVersion represents the detected Compose Component on Kubernetes side.
+type StackVersion string
+
+const (
+	// StackAPIV1Beta1 is returned if it's the most recent version available.
+	StackAPIV1Beta1 = StackVersion("v1beta1")
+)
+
+// GetAPIVersion returns the most recent stack API installed.
+func GetAPIVersion(kubeConfig *restclient.Config) (StackVersion, error) {
```
vdemeester: We may want to use `GetStackAPIVersion` here, as it's only the `Stack` api version we're getting here. We could have more api extension in the future, and `kubernetes.GetAPIVersion` kinda makes think it's the kubernetes api (server) version.

cli/command/system/version.go
silvin-lubecki: I don't see the benefits of listing all the versions? At least with just the version used, there is no ambiguity.

kubernetes/check.go
silvin-lubecki: ok

silvin-lubecki:
@vdemeester PTAL

kubernetes/check.go
```diff
@@ -0,0 +1,57 @@
...
+	StackAPIV1Beta1 = StackVersion("v1beta1")
+)
+
+// GetStackAPIVersion returns the most recent stack API installed.
+func GetStackAPIVersion(kubeConfig *restclient.Config) (StackVersion, error) {
+	log.Debugf("retrieve most recent stack API present at %s", kubeConfig.Host)
+	clients, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		return "", err
+	}
```
dnephin: This function could accept `clients` instead of a config. The client is being created twice.

cli/command/system/version.go
```diff
@@ -189,3 +211,46 @@ func getDetailsOrder(v types.ComponentVersion) []string {
...
+	if err != nil {
+		logrus.Debugf("failed to get Stack API version: %s", err)
+		return "Unknown"
+	}
+	return string(apiVersion)
+}
+
+func getKubernetesServerVersion(kubeConfig *restclient.Config) string {
+	version := "Unknown"
+	kubeClient, err := kubernetesClient.NewForConfig(kubeConfig)
```
dnephin: Could accept `kubeClient` instead of config, so the client can be passed to `GetStackAPIVersion` as well.

cli/command/system/version.go
```diff
@@ -48,16 +52,22 @@ Server:{{if ne .Platform.Name ""}} {{.Platform.Name}}{{end}}
    {{- end}}
   {{- end}}
  {{- end}}
-{{- end}}{{end}}`
+ {{- end}}{{- end}}
+ {{- if .KubernetesOK}}{{with .Kubernetes}}
+ Orchestrator:
+   Kubernetes:	{{.Kubernetes}}
```
dnephin: Minor naming nit: I assume in the future we will need to change "Server" to "Engine".

Maybe this should be formatted as:

```
Kubernetes:
  Version: {{.Kubernetes}}
  Stack API: {{.StackAPI}}
```

I think that would be more consistent with the other sections. What do you think?

cli/command/system/version.go
dnephin: Minor formatting nit: the `{{- end}}{{end}}` can go on a new line so that the template reads a bit better.

`-` in a template means "remove all whitespace before this token, so the trailing newline wont exist in the final rendered string.

cli/command/stack/deploy.go
```diff
@@ -20,7 +20,7 @@ func newDeployCommand(dockerCli command.Cli) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			opts.Namespace = args[0]
 			if dockerCli.ClientInfo().HasKubernetes() {
-				kli, err := kubernetes.WrapCli(dockerCli, cmd)
+				kli, err := kubernetes.WrapCli(dockerCli, kubernetes.NewOptions(cmd.Flags()))
```
dnephin: I hope that in the future when we add kube to more commands we can have this accept an `Options` struct instead of `cmd.Flags()`.

kubernetes/check.go
silvin-lubecki: Good catch, fixed.

cli/command/system/version.go
silvin-lubecki: Fixed

cli/command/system/version.go
silvin-lubecki: My first thought was to generalize this section to all orchestrators, but finally it seems that we don't have much things to print from swarm, and also this section seems very specific to Kubernetes. So I'm with renaming it.

cli/command/system/version.go
silvin-lubecki: Thanks, I wasn't aware of that, fixed.

cli/command/stack/deploy.go
silvin-lubecki: Ok I'll refactor that in a future PR.

silvin-lubecki:
PTAL @dnephin

PR #904, opened 2018-02-27
Fix `--label` behavior on run

vdemeester:
Commit 2b17f4c8a8caad552025edb05a73db683fb8a5c6 fixed the way empty labels
are taken into account (i.e. not interpolated from environment variable),
but it created a regression.

`ValidateLabel` functions doesn't allow empty label value, but it has
always been possible to pass an empty label via the cli (`docker run --label foo`).

This fixes that by not validating the label flag.

This needs to be backported to 18.03 👼 and this will fix one `docker-ce` CI failure.

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -145,7 +145,7 @@ func addFlags(flags *pflag.FlagSet) *containerOptions {
 		expose:            opts.NewListOpts(nil),
 		extraHosts:        opts.NewListOpts(opts.ValidateExtraHost),
 		groupAdd:          opts.NewListOpts(nil),
-		labels:            opts.NewListOpts(opts.ValidateLabel),
+		labels:            opts.NewListOpts(nil),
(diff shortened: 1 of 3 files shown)


cli/command/container/opts.go
```diff
@@ -145,7 +145,7 @@ func addFlags(flags *pflag.FlagSet) *containerOptions {
 		expose:            opts.NewListOpts(nil),
 		extraHosts:        opts.NewListOpts(opts.ValidateExtraHost),
 		groupAdd:          opts.NewListOpts(nil),
-		labels:            opts.NewListOpts(opts.ValidateLabel),
```
dnephin: I think this is the only place   where `ValidateLabel` is called, so I guess we can remove it as well?

vdemeester: As it's exported, it might be used elsewhere (for example, I'm pretty sure it's used in`libcompose`). That the only reason why I didn't remove it 😝

PR #905, opened 2018-02-27
Add appveyor setup to build and unit test

vdemeester:
The next step is to run e2e tests on windows too.

Adds a `make.ps1` powershell script to make it easy to compile and test.

```
.\scripts\make.ps1 -Binary
INFO: make.ps1 starting at 03/01/2018 14:37:28
INFO: Building...

 ________   ____  __.
 \_____  \ |    |/ _|
 /   |   \|      <
 /    |    \    |  \
 \_______  /____|__ \
         \/        \/

INFO: make.ps1 ended at 03/01/2018 14:37:30

.\scripts\make.ps1 -TestUnit
```

Related to #457

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

PR #906, opened 2018-02-27
Fixes some unit tests to be able to run them on windows

vdemeester:
Some of them are skipped for now (because the feature is not supported
or needs more work), some of them are fixed.

There is still some errors, but it's a start 👼 🙏 


cli/command/image/build_linux_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/docker/docker/pkg/archive"
+	"github.com/gotestyourself/gotestyourself/fs"
+	"github.com/gotestyourself/gotestyourself/skip"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/net/context"
+)
+
+func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
+	skip.IfCondition(t, runtime.GOOS == "windows", "uid and gid not relevant on windows")
```
vdemeester: This check is actually not relevant as `assert.Equal(t, uint32(0), fileInfo.Sys().(*syscall.Stat_t).Uid)` doesn't compile under windows… I should remove that `skip` completely.

cli/command/trust/key_load_test.go
```diff
@@ -6,8 +6,11 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 
+	"github.com/gotestyourself/gotestyourself/skip"
+
```
vdemeester: oh 😓

cli/command/stack/swarm/deploy_composefile_test.go
```diff
@@ -35,6 +35,7 @@ func TestValidateExternalNetworks(t *testing.T) {
 			expectedMsg:  "Unexpected",
 		},
 		{
+			// FIXME(vdemeester) that doesn't work under windows, the check needs to be smarter
```
vdemeester: This tests fails under Windows because of : https://github.com/docker/cli/commit/7f53c99dfe5241a985dd5ce4bed6c116a2908b41#r27815425

cli/command/formatter/history_test.go
```diff
@@ -50,6 +50,10 @@ func TestHistoryContext_ID(t *testing.T) {
 }
 
 func TestHistoryContext_CreatedSince(t *testing.T) {
+	dateStr := "2009-11-10T23:00:00Z"
+	if runtime.GOOS == "windows" {
+		dateStr = "2009-11-11T00:00:00+01:00"
```
thaJeztah: arf

Is this dependent on system configuration?

cli/command/plugin/create_test.go
```diff
@@ -29,7 +33,7 @@ func TestCreateErrors(t *testing.T) {
 		},
 		{
 			args:          []string{"plugin-foo", "nonexistent_context_dir"},
-			expectedError: "no such file or directory",
+			expectedError: noSuchFile,
```
thaJeztah: Ah, I recall some shenanigans like https://github.com/Microsoft/opengcs/blob/8cfbb6d8ae8203a80464ca6e2bd39102d6fd9deb/service/gcsutils/remotefs/utils.go#L93-L107 and https://github.com/Microsoft/opengcs/blob/8cfbb6d8ae8203a80464ca6e2bd39102d6fd9deb/service/gcsutils/remotefs/utils.go#L36-L50

cli/command/image/build_linux_test.go
thaJeztah: Isn't this file already skipped on Windows?

cli/command/formatter/history_test.go
vdemeester: I'm actually not sure 😓 cc @simonferquel @johnstep

cli/command/plugin/create_test.go
vdemeester: 🤔

cli/command/image/build_linux_test.go
```diff
@@ -0,0 +1,54 @@
+//+build linux
```
dnephin: Could we use `skip` instead of the build directive? Or does this not build at all on windows?

cli/command/trust/key_load_test.go
```diff
@@ -109,6 +115,7 @@ var testKeys = map[string][]byte{
 }
 
 func TestLoadKeyFromPath(t *testing.T) {
+	skip.If(t, func() bool { return runtime.GOOS == "windows" })
```
dnephin: Instead of  the `func() bool` here you can use `IfCondition`, or update `gotestyourself` to `1.3.0` so that `If` accepts the bool.

Maybe add a reason why as well?

cli/compose/loader/full-struct_test.go
```diff
@@ -313,10 +315,10 @@ func services(workingDir, homeDir string) []types.ServiceConfig {
 				{Target: "/var/lib/mysql", Type: "volume"},
 				{Source: "/opt/data", Target: "/var/lib/mysql", Type: "bind"},
 				{Source: workingDir, Target: "/code", Type: "bind"},
-				{Source: workingDir + "/static", Target: "/var/www/html", Type: "bind"},
+				{Source: filepath.Join(workingDir, "/static"), Target: "/var/www/html", Type: "bind"},
```
dnephin: remove the `/` from `/static` ? Also on opt below.

cli/command/image/build_linux_test.go
vdemeester: @dnephin it doesn't build at all on windows — there was a skip before 😓

cli/command/formatter/history_test.go
simonferquel: It seems. the good thing is that the timezone is in the string here, and that the format is in a standard parsable from the time package. Comparison should be done after date/time parsing instead that on strings would be better though (and would not require windows specific code).

cli/command/image/build_linux_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/pkg/archive"
+	"github.com/gotestyourself/gotestyourself/fs"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/net/context"
+)
+
+func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
+	dest := fs.NewDir(t, "test-build-context-dest")
```
simonferquel: Should'nt we run this test on Darwin as well ?

cli/command/plugin/create_test.go
simonferquel: I think there is an `os.IsNotFound()` or `os.IsFileNotFound` function that would be prettier

cli/command/stack/swarm/deploy_composefile_test.go
simonferquel: I think the info endpoint could be queried instead of making the assumption that daemon os/platform == client os/platform

cli/compose/loader/full-struct_test.go
simonferquel: yes the `/` is weird

cli/compose/loader/full-struct_test.go
```diff
@@ -313,10 +315,10 @@ func services(workingDir, homeDir string) []types.ServiceConfig {
...
 				{Source: homeDir + "/configs", Target: "/etc/configs/", Type: "bind", ReadOnly: true},
 				{Source: "datavolume", Target: "/var/lib/mysql", Type: "volume"},
-				{Source: workingDir + "/opt", Target: "/opt", Consistency: "cached", Type: "bind"},
+				{Source: filepath.Join(workingDir, "/opt"), Target: "/opt", Consistency: "cached", Type: "bind"},
```
simonferquel: same here

cli/compose/loader/full-struct_test.go
vdemeester: oups, I forgot to remove that one 🤦‍♂️

cli/command/plugin/create_test.go
thaJeztah: Oh! you're right, was thinking about Linux CLI calling a Windows daemon (in which errors don't match), but that's not the case here

cli/command/image/build_linux_test.go
vdemeester: @simonferquel yes, this is also on my TODO (but I should open an issue to track the work on that)

cli/command/stack/swarm/deploy_composefile_test.go
vdemeester: definitely, but probably need to be "skipped" for now, and follow-up on a fix PR for that problem only 👼

cli/compose/loader/full-struct_test.go
vdemeester: 🤦‍♂️

vdemeester:
Closing this one as I'm putting the commit in https://github.com/docker/cli/pull/905 (and I'll take comments into account there)

cli/command/formatter/history_test.go
dnephin: Since we're testing a formatter, which is responsible for formatting output that is seen by a user, I think we should be comparing the string, not a `time.Time`. The format is relevant here.

PR #907, opened 2018-02-27
Add more container cp tests

dnephin:
Migrated from moby/moby, where they will be removed shortly.

### DIFF
--- a/cli/command/container/cp_test.go
+++ b/cli/command/container/cp_test.go
@@ -3,6 +3,7 @@ package container
 import (
 	"io"
 	"io/ioutil"
+	"os"
 	"runtime"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


cli/command/container/cp_test.go
```diff
@@ -111,6 +112,33 @@ func TestRunCopyFromContainerToFilesystemMissingDestinationDirectory(t *testing.
...
+
+func TestRunCopyToContainerSourceDoesNotExist(t *testing.T) {
+	options := copyOptions{
+		source:      "/does/not/exist",
+		destination: "container:/path",
+	}
+	cli := test.NewFakeCli(&fakeClient{})
+	err := runCopy(cli, options)
+	expected := "no such file or directory"
+	if runtime.GOOS == "windows" {
```
thaJeztah: Should we consider having a utility for this (i.e.; check if it's a "not found" error)?

dnephin: Ya, sounds good

PR #912, opened 2018-02-28
Fix grammar typo in comment

kkirsche:
Overriding is the incorrect part of speech for this sentence. It is more common to state that the values are overridden instead.

Other options would include:

```golang
// ldflags is overriding these values
```

```golang
// These values will be overridden by ldflags
```

etc.

**- What I did**
Fixed a grammatical issue

**- How I did it**
By reviewing the phrasing in place and updating it accordingly

**- How to verify it**
Reading examples of the term's usage on websites such as http://www.dictionary.com/browse/overridden

Fix grammatical error in default ldflags build variable comment

cute puppy

### DIFF
--- a/cli/version.go
+++ b/cli/version.go
@@ -1,7 +1,7 @@
 package cli
 
 // Default build-time variable.
-// These values are overriding via ldflags
+// These values are overridden via ldflags
 var (
 	PlatformName = ""
 	Version      = "unknown-version"


kkirsche:
This was done via the web interface, is there a way to amend it with the signature? For a single word change I was hoping to avoid pulling down the repo and whatnot

PR #915, opened 2018-03-01
Update run.md fix broken links

LydellManganti:
Fix broken links to point to the working anchor:
https://docs.docker.com/engine/reference/run/#name-name
https://docs.docker.com/engine/reference/run/#restart-policies-restart
https://docs.docker.com/engine/reference/run/#ipc-settings-ipc
https://docs.docker.com/engine/reference/run/#clean-up-rm

**- What I did**
Apply same fix as line 566. The link works when displayed in https://docs.docker.com/engine/reference/run.
But does not work here in this preview.

**- How I did it**

**- How to verify it**


vdemeester:
Ok then, let's close it :angel: 
Thanks for the contribution @LydellManganti

PR #921, opened 2018-03-06
testing: Use canonical asserts

dnephin:
Branched from #920 

grep/sed/perl commands are included in the git commit message.


cli/command/image/build_test.go
```diff
@@ -161,12 +161,16 @@ COPY data /data
 // TestRunBuildFromLocalGitHubDirNonExistingRepo tests that build contexts
 // starting with `github.com/` are special-cased, and the build command attempts
 // to clone the remote repo.
+// TODO: test "context selection" logic directly when runBuild is refactored
+// to support testing (ex: docker/cli#294)
```
thaJeztah: Needs a rebase now?

dnephin: Rebase will only change the diff shown in github. It will merge clean because the commit is the same.

PR #924, opened 2018-03-07
Add some content trust tests

vdemeester:
Importing from moby's `DockerTrustSuite` tests.

This is still wip since there is some tests that are failing and some empty tests I did not yet migrate.
Also, there has been some slight changes between `17.06` and current `docker/cli`, so one some behavior, I'm not sure if it's supposed to fail or not (cc @n4ss @endophage)

See https://github.com/moby/moby/pull/36515 for original tests that are supposed to be migrated.

Probably require https://github.com/docker/cli/pull/929 to be able to write unit tests for content trust..


e2e/compose-env.yaml
```diff
@@ -17,5 +17,14 @@ services:
        - notary-fixtures:/fixtures
       command: ['notary-server', '-config=/fixtures/notary-config.json']
 
+  evil-notary-server:
```
dnephin: why is it evil?

vdemeester: See https://github.com/moby/moby/pull/36515/files#diff-4b1e56bb77ac16f2ccf956fc24cf0a82L350 :stuck_out_tongue: 
Definitely open to name it something else, but I'll need to re-generate the certs :sweat_smile:

e2e/container/create_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/docker/cli/e2e/internal/fixtures"
+	"github.com/gotestyourself/gotestyourself/icmd"
+)
+
+func TestCreateWithContentTrust(t *testing.T) {
```
dnephin: If both `create` and `run` use exactly the same code, do we need e2e tests for both?

e2e/image/push_trust_test.go
```diff
@@ -0,0 +1,160 @@
...
+		fixtures.WithConfig(dir.Path()),
+		fixtures.WithTrust,
+		fixtures.WithNotaryServer("https://invalidnotaryserver"),
+	)
+	result.Assert(t, icmd.Expected{
+		ExitCode: 1,
+		Err:      "error contacting notary server",
+	})
+
+	// FIXME(vdemeester) doesn't work
```
dnephin: ya, this is weird. I don't understand why it would hit notary if content trust is disabled.

e2e/image/push_trust_test.go
```diff
@@ -0,0 +1,160 @@
+package image
```
dnephin: `push_test.go` ?

e2e/image/push_trust_test.go
```diff
@@ -0,0 +1,160 @@
...
+		icmd.Command(notary, "-c", notaryDir.Path(), "delegation", "add", image, "targets/releases", pubkey, "--all-paths"),
+		withNotaryPassphrase("foo"),
+	)
+}
+
+func TestPushWithContentTrustSignsAllFirstLevelRolesWeHaveKeysFor(t *testing.T) {}
+
+func TestPushWithContentTrustSignsForRolesWithKeysAndValidPaths(t *testing.T) {}
+
+func TestPullWithContentTrustSignsForRolesWithKeysAndValidPaths(t *testing.T) {}
```
dnephin: There are way too many e2e tests here. It must be possible for some of these to be unit tests.

e2e/plugin/trust_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/docker/docker/api/types"
+	"github.com/gotestyourself/gotestyourself/assert"
+	"github.com/gotestyourself/gotestyourself/fs"
+	"github.com/gotestyourself/gotestyourself/icmd"
+	"github.com/pkg/errors"
+)
+
+const registryPrefix = "registry:5000"
+
+func TestPluginInstallWithContentTrust(t *testing.T) {
```
dnephin: `TestInstallWithContentTrust`

dockerfiles/Dockerfile.dev
```diff
@@ -24,6 +24,11 @@ RUN     go get -d github.com/dnephin/filewatcher && \
         go build -v -o /usr/bin/filewatcher . && \
         rm -rf /go/src/* /go/pkg/* /go/bin/*
 
+ARG     NOTARY_VERSION=v0.6.0
+RUN     export URL=https://github.com/theupdateframework/notary/releases/download; \
+        curl -Ls $URL/${NOTARY_VERSION}/notary-Linux-amd64 -o /usr/local/bin/notary && \
+        chmod +x /usr/local/bin/notary
```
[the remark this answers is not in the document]
vdemeester: For some tests (to come), we need to use the `notary` client to set up the server in some state.

e2e/container/create_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	result := icmd.RunCmd(
+		icmd.Command("docker", "create", fixtures.AlpineImage),
+		fixtures.WithConfig(dir.Path()),
+		fixtures.WithTrust,
+		fixtures.WithNotary,
+	)
+	result.Assert(t, icmd.Expected{
+		ExitCode: 1,
+		Err:      "does not have trust data for",
```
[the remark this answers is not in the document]
vdemeester: yes, and we probably should, that's a good point. I only worked moving from the moby's integration to cli's e2e tests but those could definitely be unit tests instead :angel: (with just one e2e tests that validates the behavior against a real server)

e2e/image/push_trust_test.go
vdemeester: Yep, but somehow, in moby, with 17.06, that's something it was doing :sweat_smile:

e2e/container/create_test.go
```diff
@@ -0,0 +1,80 @@
...
+		fixtures.WithNotaryServer(fixtures.EvilNotaryURL),
+		fixtures.WithPassphrase("foo", "bar"),
+	)
+	result.Assert(t, icmd.Expected{
+		Out: "Signing and pushing trust metadata",
+	})
+
+	// FIXME(vdemeester) doesn't work
+	/*
+		// Now, try running with the original client from this new trust server. This should fail because the new root is invalid.
```
vdemeester: @endophage @n4ss what's the expected behavior here ? (the commented code comes from initial moby test suite, but fails on current `docker/cli`)

e2e/image/push_test.go
```diff
@@ -0,0 +1,160 @@
...
+		fixtures.WithConfig(dir.Path()),
+		fixtures.WithTrust,
+		fixtures.WithNotaryServer("https://invalidnotaryserver"),
+	)
+	result.Assert(t, icmd.Expected{
+		ExitCode: 1,
+		Err:      "error contacting notary server",
+	})
+
+	// FIXME(vdemeester) doesn't work
```
vdemeester: @endophage @n4ss what's the expected behavior here ? pretty sure it works as expected and the commented code should be gone, but somehow, in 17.06 it did act that way (i.e. even with `--disable-content-trust` it would fail)

e2e/image/push_test.go
```diff
@@ -0,0 +1,160 @@
...
+		fixtures.WithPassphrase("foo", "bar"),
+	)
+	result.Assert(t, icmd.Expected{
+		Out: "Signing and pushing trust metadata",
+	})
+
+	// FIXME(vdemeester) doesn't work
+	/*
+		// With a wrong password
+		result = icmd.RunCmd(icmd.Command("docker", "push", image),
```
vdemeester: @endophage @n4ss what's the expected behavior here ? (the commented code comes from initial moby test suite, but fails on current `docker/cli`)

e2e/image/push_test.go
n4ss: I don't know.. If you disable content trust, it shouldn't try to reach the notary server, right?

@endophage @cyli if you have more guesses?

dockerfiles/Dockerfile.dev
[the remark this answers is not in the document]
dnephin: Maybe add a TODO to move to a separate stage when we make this multi-stage? and document that it's only used from the e2e runner?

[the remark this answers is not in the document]
vdemeester: :+1:

vdemeester:
Updated with removing commented code (and cleaning some stuff not required for now)

vdemeester:
The failure is weird :scream: 

```
17:18:22 --- FAIL: TestPushWithContentTrust (0.87s)
17:18:22 	push_test.go:43: assertion failed: 
17:18:22 		--- expected
17:18:22 		+++ actual
17:18:22 		@@ -1,6 +1,6 @@
17:18:22 		 The push refers to repository [registry:5000/trust-push]
17:18:22 		 5bef08742407: Preparing
17:18:22 		-5bef08742407: Mounted from trust-create
17:18:22 		+5bef08742407: Mounted from trust-pull-unreachable
17:18:22 		 latest: digest: sha256:641b95ddb2ea9dc2af1a0113b6b348ebc20872ba615204fbe12148e98fd6f23d size: 528
17:18:22 		 Signing and pushing trust metadata
17:18:22 		 Finished initializing "registry:5000/trust-push"
```

PS: I have another `Mounted from` on my laptop..

vdemeester:
We either need to clean the registry between each test or.. something else :sweat:

dnephin:
We could use the line matching system for image build (https://github.com/docker/cli/blob/master/e2e/image/build_test.go#L31). Just needs to be renamed to `assertLines` or something like that.

e2e/image/push_test.go
cyli: Apologies for not getting to this sooner - if it was this test: https://github.com/moby/moby/pull/36515/files#diff-cf31c649a9e77c869ff8c6cdcec05f60L328, it looks like the intention was to test the `--disable-content-trust` flag in combination with the content trust environment variable to say that even with the environment variable set, if `--disable-content-trust` is passed, it doesn't use content trust.  So the push should work, because it doesn't contact the notary server, hence it seem to be checking that there is no comment about the notary server being invalid (e.g. if it were going the path of content trust, the push should have failed with that error message).

e2e/image/push_test.go
cyli: Is that https://github.com/moby/moby/pull/36515/files#diff-cf31c649a9e77c869ff8c6cdcec05f60L368?  If so, it's possible that the error message has changed - not sure what the failure on `docker/cli` was.

cyli:
Thank you so much for porting these @vdemeester!  Apologies I didn't respond earlier.  This all LGTM - I can attempt to add the `TestTrustedPushWithoutServerAndUntrusted` test back in another PR - that seems useful to test that the flag and the environment variable interact in the correct way, or perhaps that should just be a unit test.  

`TestTrustedPushWithIncorrectPassphraseForNonRoot` possibly does not need to exist in `docker/cli` unless we translate the error messaging.  I feel like this should be part of the notary tests instead.  We have a test in notary for not being able to add data when the root passphrase is invalid (https://github.com/theupdateframework/notary/blob/master/client/client_test.go#L722) but not the repo passphrase.  I've added an issue here:  https://github.com/theupdateframework/notary/issues/1317

vdemeester:
@dnephin updated, using `assertBuildOuptut` (move the thing in a separate package)

e2e/service/trust_test.go
```diff
@@ -0,0 +1,9 @@
+package service
+
+import (
+	"testing"
+)
+
+func TestServiceCreateWithContentTrust(t *testing.T) {}
+
+func TestServiceUpdateWithContentTrust(t *testing.T) {}
```
dnephin: remove for now and open an issue for these?

vdemeester: huh :sweat:

cli/command/container/run_test.go
```diff
@@ -23,3 +26,48 @@ func TestRunLabel(t *testing.T) {
 	cmd.SetArgs([]string{"--label", "foo", "busybox"})
 	assert.NilError(t, cmd.Execute())
 }
+
+func TestRunCommandWithContentTrustErrors(t *testing.T) {
```
thaJeztah: This is basically the same test as for "create" (`TestNewCreateCommandWithContentTrustErrors`); wonder (for a follow-up) if we can share some of the code, so that we don't have to maintain it twice, but also so that we make sure they're kept in sync

PR #925, opened 2018-03-07
Add a build unit test for symlinked context

dnephin:
Replaced `TestBuildContextDirIsSymlink` removed in https://github.com/moby/moby/pull/36515/


dnephin:
cc @thaJeztah

PR #929, opened 2018-03-08
Refactor content_trust cli/flags handling

vdemeester:
Remove the global variable used. Allows easier unit testing.

This is linked to https://github.com/docker/cli/pull/924 as it is required to be able to write unit tests for content trust more easily.

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -52,6 +52,7 @@ type Cli interface {
 	DefaultVersion() string
 	ManifestStore() manifeststore.Store
 	RegistryClient(bool) registryclient.RegistryClient
+	IsTrusted() bool
 }
 
 // DockerCli is an instance the docker command line client.
@@ -64,6 +65,7 @@ type DockerCli struct {
 	client     client.APIClient
 	serverInfo ServerInfo
 	clientInfo ClientInfo
+	isTrusted  bool
 }
 
 // DefaultVersion returns api.defaultVersion or DOCKER_API_VERSION if specified.
@@ -119,6 +121,11 @@ func (cli *DockerCli) ServerInfo() ServerInfo {
 // ClientInfo returns the client details for the cli
 func (cli *DockerCli) ClientInfo() ClientInfo {
 	return cli.clientInfo
+}
+
+// IsTrusted returns if content trust is enabled for the cli
+func (cli *DockerCli) IsTrusted() bool {
+	return cli.isTrusted
 }
 
 // ManifestStore returns a store for local manifests
@@ -237,8 +244,8 @@ func (c ClientInfo) HasKubernetes() bool {
 }
 
 // NewDockerCli returns a DockerCli instance with IO output and error streams set by in, out and err.
-func NewDockerCli(in io.ReadCloser, out, err io.Writer) *DockerCli {
-	return &DockerCli{in: NewInStream(in), out: NewOutStream(out), err: err}
+func NewDockerCli(in io.ReadCloser, out, err io.Writer, isTrusted bool) *DockerCli {
+	return &DockerCli{in: NewInStream(in), out: NewOutStream(out), err: err, isTrusted: isTrusted}
 }
 
 // NewAPIClientFromFlags creates a new APIClient from command line flags
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -52,20 +52,20 @@ type Cli interface {
 	DefaultVersion() string
 	ManifestStore() manifeststore.Store
 	RegistryClient(bool) registryclient.RegistryClient
-	IsTrusted() bool
+	ContentTrustEnabled() bool
 }
 
 // DockerCli is an instance the docker command line client.
 // Instances of the client can be returned from NewDockerCli.
 type DockerCli struct {
-	configFile *configfile.ConfigFile
-	in         *InStream
-	out        *OutStream
-	err        io.Writer
-	client     client.APIClient
-	serverInfo ServerInfo
-	clientInfo ClientInfo
-	isTrusted  bool
+	configFile   *configfile.ConfigFile
+	in           *InStream
+	out          *OutStream
+	err          io.Writer
+	client       client.APIClient
+	serverInfo   ServerInfo
+	clientInfo   ClientInfo
+	contentTrust bool
 }
 
 // DefaultVersion returns api.defaultVersion or DOCKER_API_VERSION if specified.
@@ -123,9 +123,10 @@ func (cli *DockerCli) ClientInfo() ClientInfo {
 	return cli.clientInfo
 }
 
-// IsTrusted returns if content trust is enabled for the cli
-func (cli *DockerCli) IsTrusted() bool {
-	return cli.isTrusted
+// ContentTrustEnabled returns if content trust has been enabled by an
+// environment variable.
+func (cli *DockerCli) ContentTrustEnabled() bool {
+	return cli.contentTrust
 }
 
 // ManifestStore returns a store for local manifests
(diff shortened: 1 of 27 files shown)


internal/test/cli.go
```diff
@@ -151,3 +157,13 @@ func (c *FakeCli) SetManifestStore(store manifeststore.Store) {
...
 	c.registryClient = client
 }
+
+// IsTrusted on the fake cli
+func (c *FakeCli) IsTrusted() bool {
+	return c.isTrusted
+}
+
+// IsTrusted sets "enables" content trust on the fake cli
+func IsTrusted(c *FakeCli) {
```
mat007: SetTrusted instead maybe ?

cli/command/image/push.go
```diff
@@ -47,7 +55,7 @@ func runPush(dockerCli command.Cli, remote string) error {
 	authConfig := command.ResolveAuthConfig(ctx, dockerCli, repoInfo.Index)
 	requestPrivilege := command.RegistryAuthenticationPrivilegedFunc(dockerCli, repoInfo.Index, "push")
 
-	if command.IsTrusted() {
+	if !opts.untrusted && dockerCli.IsTrusted() {
```
dnephin: Does this need to check both values? It seems like either we can remove `dockerCli.IsTrusted()` from the call to `AddTrustSigningFlags` or we can remove `dockerCli.IsTrusted()` from this line and  check only `opts.untrusted` . `dockerCli.IsTrusted()` is being used as the default value to the trust flag, so I don't think we need to use it twice.

same thing in push.

cli/command/image/build.go
```diff
@@ -460,7 +462,7 @@ func runBuild(dockerCli command.Cli, options buildOptions) error {
 			return err
 		}
 	}
-	if command.IsTrusted() {
+	if !options.untrusted && dockerCli.IsTrusted() {
```
dnephin: Oh, also here , and the place above

dnephin:
Pushed 2 commits. The first removes reading of the global flag and just uses the `opt`. The second renames things to `ContentTrust` instead of `IsTrusted`. I left them as separate commits so that if you don't like the rename you can drop it easily.  Feel free to squash them if they look ok.

cli/command/cli.go
```diff
@@ -52,18 +52,20 @@ type Cli interface {
 	DefaultVersion() string
 	ManifestStore() manifeststore.Store
 	RegistryClient(bool) registryclient.RegistryClient
+	ContentTrustEnabled() bool
```
n4ss: way cleaner! 👍

cli/command/image/build.go
```diff
@@ -543,11 +544,11 @@ func rewriteDockerfileFrom(ctx context.Context, dockerfile io.Reader, translator
 	return buf.Bytes(), resolvedTags, scanner.Err()
 }
 
-// replaceDockerfileTarWrapper wraps the given input tar archive stream and
+// replaceDockerfileForContentTrust wraps the given input tar archive stream and
 // replaces the entry with the given Dockerfile name with the contents of the
 // new Dockerfile. Returns a new tar archive stream with the replaced
```
n4ss: Worth to add content trust in the function description?

cmd/docker/docker.go
```diff
@@ -155,7 +156,7 @@ func main() {
 	stdin, stdout, stderr := term.StdStreams()
 	logrus.SetOutput(stderr)
 
-	dockerCli := command.NewDockerCli(stdin, stdout, stderr)
+	dockerCli := command.NewDockerCli(stdin, stdout, stderr, isContentTrustEnabled())
```
n4ss: Do we want to have that in the initializer prototype? rather with a dependency injection or just done from within the initializer function without modifying the prototype?

dnephin: I'm ok with it either way

cli/command/image/build.go
dnephin: Updated the godoc

dnephin:
squashed the 2 commits from me

n4ss:
LGTM!

cli/command/cli.go
```diff
@@ -121,6 +123,12 @@ func (cli *DockerCli) ClientInfo() ClientInfo {
 	return cli.clientInfo
 }
 
+// ContentTrustEnabled returns if content trust has been enabled by an
```
thaJeztah: nit: reads a bit odd, perhaps could be `returns whether trust has been enabled..`

cli/command/image/build.go
```diff
@@ -503,7 +504,7 @@ type resolvedTag struct {
 // "FROM <image>" instructions to a digest reference. `translator` is a
 // function that takes a repository name and tag reference and returns a
 // trusted digest reference.
-func rewriteDockerfileFrom(ctx context.Context, dockerfile io.Reader, translator translatorFunc) (newDockerfile []byte, resolvedTags []*resolvedTag, err error) {
+func rewriteDockerfileFrom(ctx context.Context, dockerfile io.Reader, translator translatorFunc, istrusted bool) (newDockerfile []byte, resolvedTags []*resolvedTag, err error) {
```
thaJeztah: Looks like the whole reason this `rewriteDockerfileFrom()` file is there, is for docker content trust. Instead of passing this boolean, can we just skip the whole call to this function?

So, actually, looks like that's already what's done; it's only called in that case, so we can just remove the boolean (and check). This is the only location this function is called:
https://github.com/docker/cli/blob/82a80858857aeeb37008bd7d2f4afcda3d39c966/cli/command/image/build.go#L288-L305

cli/command/image/build.go
```diff
@@ -543,11 +544,10 @@ func rewriteDockerfileFrom(ctx context.Context, dockerfile io.Reader, translator
...
 
-// replaceDockerfileTarWrapper wraps the given input tar archive stream and
-// replaces the entry with the given Dockerfile name with the contents of the
-// new Dockerfile. Returns a new tar archive stream with the replaced
-// Dockerfile.
-func replaceDockerfileTarWrapper(ctx context.Context, inputTarStream io.ReadCloser, dockerfileName string, translator translatorFunc, resolvedTags *[]*resolvedTag) io.ReadCloser {
+// replaceDockerfileForContentTrust wraps the given input tar archive stream and
+// uses the translator to replace the Dockerfile which uses a trusted reference.
+// Returns a new tar archive stream with the replaced Dockerfile.
+func replaceDockerfileForContentTrust(ctx context.Context, inputTarStream io.ReadCloser, dockerfileName string, translator translatorFunc, resolvedTags *[]*resolvedTag) io.ReadCloser {
```
thaJeztah: We could rename these functions to, e.g. `resolveDigestsYadaYada` or something else (it's just that we use them for "content trust" as it requires all images to be referenced by digest)

cli/command/image/push.go
```diff
@@ -11,26 +11,34 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type pushOptions struct {
+	remote    string
+	untrusted bool
```
thaJeztah: Can we somehow flip this boolean? I'm a bit ... erm ... idk about having `untrusted`, and it looks like _everywhere_ we use it as `if !untrusted`

cmd/docker/docker.go
```diff
@@ -175,6 +176,16 @@ func main() {
 	}
 }
 
+func isContentTrustEnabled() bool {
```
thaJeztah: nit: the `is` prefix is a bit redundant because it already has an `Enabled` suffix; it reads equally well without (`if contentTrustEnabled()`)

cmd/docker/docker.go
thaJeztah: Was also looking at this a bit (not a fan of boolean args), but perhaps that's just a nit

vdemeester:
@thaJeztah you're the best :rose:

cli/command/image/push.go
dnephin: I don't think so. The flag is `disable-content-trust`, so it has to store true to disable it.

thaJeztah: Yeah, it's a weird flag;

```
docker run --help

--disable-content-trust          Skip image verification (default true)
```


```
DOCKER_CONTENT_TRUST=1 docker run --help

--disable-content-trust          Skip image verification
```

PR #930, opened 2018-03-08
Small unit tests fixes

vdemeester:
- skip some test when running on user (not root) as they require root permissions.
- fix history formating

### DIFF
--- a/cli/command/formatter/history.go
+++ b/cli/command/formatter/history.go
@@ -78,7 +78,7 @@ func (c *historyContext) ID() string {
 }
 
 func (c *historyContext) CreatedAt() string {
-	return time.Unix(c.h.Created, 0).Format(time.RFC3339)
+	return time.Unix(c.h.Created, 0).UTC().Format(time.RFC3339)
 }
 
 func (c *historyContext) CreatedSince() string {
--- a/cli/command/image/build_linux_test.go
+++ b/cli/command/image/build_linux_test.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"io"
 	"io/ioutil"
+	"os"
 	"syscall"
 	"testing"
 
@@ -15,10 +16,12 @@ import (
 	"github.com/gotestyourself/gotestyourself/assert"
 	is "github.com/gotestyourself/gotestyourself/assert/cmp"
 	"github.com/gotestyourself/gotestyourself/fs"
+	"github.com/gotestyourself/gotestyourself/skip"
 	"golang.org/x/net/context"
 )
 
 func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
+	skip.If(t, os.Getuid() != 0, "require root to run")
 	dest := fs.NewDir(t, "test-build-context-dest")
 	defer dest.Remove()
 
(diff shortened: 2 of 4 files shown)


cli/command/formatter/history.go
```diff
@@ -78,7 +78,7 @@ func (c *historyContext) ID() string {
 }
 
 func (c *historyContext) CreatedAt() string {
-	return time.Unix(c.h.Created, 0).Format(time.RFC3339)
+	return time.Unix(c.h.Created, 0).UTC().Format(time.RFC3339)
```
vdemeester: I'm actually not quite sure on this one. But I think we can make the assumption it's in `UTC` timezone.

dnephin: I don't know, it seems like this will change the value users are seeing from local time to UTC, which will be confusing if all the other times they see are local.

thaJeztah: hm, yes, good point

cli/command/image/build_linux_test.go
```diff
@@ -15,10 +16,12 @@ import (
 	"github.com/gotestyourself/gotestyourself/assert"
 	is "github.com/gotestyourself/gotestyourself/assert/cmp"
 	"github.com/gotestyourself/gotestyourself/fs"
+	"github.com/gotestyourself/gotestyourself/skip"
 	"golang.org/x/net/context"
 )
 
 func TestRunBuildResetsUidAndGidInContext(t *testing.T) {
+	skip.If(t, os.Getuid() != 0, "require root to run")
```
dnephin: It looks like this only needs root because of `archive.Untar`. I don't think this test actually needs to use that.

Maybe that will allow us to move it back to `build_test.go` ?

cli/command/manifest/push_test.go
```diff
@@ -51,6 +53,7 @@ func TestManifestPushErrors(t *testing.T) {
 
 // store a one-image manifest list and puah it
 func TestManifestPush(t *testing.T) {
+	skip.If(t, os.Getuid() != 0, "require root to run")
```
dnephin: I'm able to run this without root on linux. How does this fail for you?

vdemeester: `push_test.go:71: assertion failed: error is not nil: open /etc/docker/certs.d/example.com: permission denied`

dnephin: The client used by that test should be mocked by `newFakeRegistryClient`, so I'm not sure why it's looking for a cert. I can't reproduce the issue on linux.  Does it happen on both osx and linux for you?

dnephin:
Opened #931 which adds a new fake for dealing with context. I'm able to run most of the tests without root. Only the one that actually needs chown is skipped. Also skips the image history test if the timezone isn't UTC.

cli/command/manifest/push_test.go
dnephin: Ok, for me `/etc/docker` is `0600` and the test is ok with it.

vdemeester: Hum I'll check tomorrow but that ain't big deal 😛

vdemeester:
Closing in favor of #931

PR #931, opened 2018-03-08
Refactor build tests to re-use more code and not require root

dnephin:
Also removes the dependency on `docker/docker/pkg/archive`.

Related to #930


cli/command/image/history_test.go
```diff
@@ -42,7 +43,13 @@ func TestNewHistoryCommandErrors(t *testing.T) {
 	}
 }
 
+func notUTCTimezone() bool {
+	now := time.Now()
+	return now != now.UTC()
+}
+
 func TestNewHistoryCommandSuccess(t *testing.T) {
+	skip.If(t, notUTCTimezone, "expected output requires UTC timezone")
```
vdemeester: ❤️

PR #932, opened 2018-03-08
Migrate to golang/dep for vendor management

nogoegst:
**- What I did**
Converted vendor to golang/dep.

**- How I did it**
`dep init`

**- How to verify it**
`dep ensure`

- [x]  reduce resulting vendor size
- [x]  stick with reasonable SHAs/versions of dependencies from `vendor.conf`
- [x]  fix everything that depends on usage of `vndr`
- [x]  pin to latest release of `dep` in `dockerfiles/Dockerfile.dev`


dnephin:
Thanks for the PR!

The addition of 140k+ lines is not good. Could you try adding `non-go = true` to the prune config? Hopefully that still keeps license files. If not we might need to request this feature on the `dep` repo.

It seems a lot of versions have changed from `vendor.conf` and `Gopkg.lock`. We should try to pin more things to a specific sha, at least initially.

This will need a few more changes:
1. add `dep` and remove `vndr` from `dockerfiles/Dockerfile.dev`. Pinning to `v0.4.1` tag would be great.
2. update the vendor check to use `dep ensure`

nogoegst:
@dnephin, thanks for the notes! I've updated PR description to track progress on what should be done.

nogoegst:
@dnephin, I've enabled `non-go` and it didn't change the situation dramatically (+123,603 −33,711). There should be another reason for this.
btw `non-go` actually leaves `LICENSE` files in place. :wink:

nogoegst:
Oh okay, I found the reason: `k8s.io/kubernetes/Godeps/LICENSES` is 5 MB! They collect all the licenses of all the dependencies into this one file. This file is being preserved due to legal file preserving behavior of `dep`. We obviously have no need to have this file due to `dep` recreates vendor directory with all the licenses in it.
I've added it to `.gitignore` so it is not being commited after running `dep ensure`.
Now it's ( +27,554 −33,711), yay! :tada:

nogoegst:
@dnephin regarding pinning to SHAs of deps: dep has correctly converted all the SHAs from `vendor.conf` to `Gopkg.lock`. So they are actually pinned.
One should just run `dep ensure -update` only for specific packages thus leaving other packages intact.

nogoegst:
beware that because vendor got updated on master this PR should be rebased/reconverted. i'll drop a note when it's done.

vieux:
should we close this one and move to vgo ? :D

nogoegst:
@vieux, thanks for heads up!

I started a rebase and got a dependency diversion:
`Could not introduce github.com/moby/buildkit@aaff9d591ef128560018433fe61beb802e149de8, as it depends on github.com/docker/docker from https://github.com/tonistiigi/docker.git, but github.com/docker/docker is already marked as coming from github.com/docker/docker by github.com/docker/cli`

buildkit doesn't have this line anymore so it has to be updated. I've created #1078 for bumping buildkit vendor.

tiborvass:
Thank you for your contribution but if we are migrating to anything it will be Go modules at this point.

PR #936, opened 2018-03-09
Use multistage build in Dockerfile.dev; ignore apk cache when installing packages

nogoegst:
**- What I did**
 * Enabled mutlistage build in `dockerfiles/Dockerfile.dev` so that eliminated some `rm -rf` lines.
 * Enabled `--no-cache` option of `apk` so that there is no package cache in the image.

**- How I did it**
Edited `dockerfiles/Dockerfile.dev`.

**- How to verify it**
```docker build -t docker-cli-dev -f dockerfiles/Dockerfile.dev dockerfiles```

### DIFF
--- a/dockerfiles/Dockerfile.dev
+++ b/dockerfiles/Dockerfile.dev
@@ -1,28 +1,30 @@
+FROM    golang:1.9.4-alpine3.6 AS buildbase
+RUN     apk --no-cache add git make bash coreutils ca-certificates
 
-FROM    golang:1.9.4-alpine3.6
+FROM    buildbase AS tools
 
-RUN     apk add -U git make bash coreutils ca-certificates
-
-ARG     VNDR_SHA=a6e196d8b4b0cbbdc29aebdb20c59ac6926bb384
+ARG	VNDR_SHA=a6e196d8b4b0cbbdc29aebdb20c59ac6926bb384
 RUN     go get -d github.com/LK4D4/vndr && \
         cd /go/src/github.com/LK4D4/vndr && \
         git checkout -q "$VNDR_SHA" && \
-        go build -v -o /usr/bin/vndr . && \
-        rm -rf /go/src/* /go/pkg/* /go/bin/*
+        go build -v -o /usr/bin/vndr .
 
 ARG     ESC_SHA=58d9cde84f237ecdd89bd7f61c2de2853f4c5c6e
 RUN     go get -d github.com/mjibson/esc && \
         cd /go/src/github.com/mjibson/esc && \
         git checkout -q "$ESC_SHA" && \
-        go build -v -o /usr/bin/esc . && \
-        rm -rf /go/src/* /go/pkg/* /go/bin/*
+        go build -v -o /usr/bin/esc .
 
 ARG     FILEWATCHER_SHA=2e12ea42f6c8c089b19e992145bb94e8adaecedb
 RUN     go get -d github.com/dnephin/filewatcher && \
         cd /go/src/github.com/dnephin/filewatcher && \
         git checkout -q "$FILEWATCHER_SHA" && \
-        go build -v -o /usr/bin/filewatcher . && \
-        rm -rf /go/src/* /go/pkg/* /go/bin/*
+        go build -v -o /usr/bin/filewatcher .
+
+FROM    buildbase
+COPY    --from=tools /usr/bin/vndr /usr/bin/vndr
+COPY    --from=tools /usr/bin/esc /usr/bin/esc
+COPY    --from=tools /usr/bin/filewatcher /usr/bin/filewatcher
 
 ENV     CGO_ENABLED=0 \
         PATH=$PATH:/go/src/github.com/docker/cli/build \

### CODE REVIEW

File: dockerfiles/Dockerfile.dev [hunk 1]
```diff
@@ -1,28 +1,30 @@
+FROM    golang:1.9.4-alpine3.6 AS buildbase
+RUN     apk --no-cache add git make bash coreutils ca-certificates
 
-FROM    golang:1.9.4-alpine3.6
+FROM    buildbase AS tools
 
-RUN     apk add -U git make bash coreutils ca-certificates
-
-ARG     VNDR_SHA=a6e196d8b4b0cbbdc29aebdb20c59ac6926bb384
+ARG	VNDR_SHA=a6e196d8b4b0cbbdc29aebdb20c59ac6926bb384
```
thaJeztah: Looks like this changed from spaces to tabs 😅

thaJeztah:
I'm a bit on the fence for this one, as I wonder how much benefit this brings in practice. While multi-stage builds can address certain issues, and provide performance gains, using `COPY --from` can sometimes actually be slower (extra `COPY` step, instead of the files already being in the image without multi-stage).

Quick testing of this PR;

Before this change (so: non-multi-stage):

```
time docker build -t cli-before --no-cache -f dockerfiles/Dockerfile.dev .
28.12 real         0.37 user         0.48 sys
```

With this change (multi-stage):

```
time docker build -t cli-after --no-cache -f dockerfiles/Dockerfile.dev .
28.89 real         0.37 user         0.37 sys
```

Using "master", but changing to `api add --no-cache`, and removing the `rm -rf ... ` cleanup steps:

<details>

```patch
diff --git a/dockerfiles/Dockerfile.dev b/dockerfiles/Dockerfile.dev
index 5734b0cf..7275e812 100644
--- a/dockerfiles/Dockerfile.dev
+++ b/dockerfiles/Dockerfile.dev
@@ -1,28 +1,25 @@
 
 FROM    golang:1.9.4-alpine3.6
 
-RUN     apk add -U git make bash coreutils ca-certificates
+RUN     apk add --no-cache git make bash coreutils ca-certificates
 
 ARG     VNDR_SHA=a6e196d8b4b0cbbdc29aebdb20c59ac6926bb384
 RUN     go get -d github.com/LK4D4/vndr && \
         cd /go/src/github.com/LK4D4/vndr && \
         git checkout -q "$VNDR_SHA" && \
-        go build -v -o /usr/bin/vndr . && \
-        rm -rf /go/src/* /go/pkg/* /go/bin/*
+        go build -v -o /usr/bin/vndr .
 
 ARG     ESC_SHA=58d9cde84f237ecdd89bd7f61c2de2853f4c5c6e
 RUN     go get -d github.com/mjibson/esc && \
         cd /go/src/github.com/mjibson/esc && \
         git checkout -q "$ESC_SHA" && \
-        go build -v -o /usr/bin/esc . && \
-        rm -rf /go/src/* /go/pkg/* /go/bin/*
+        go build -v -o /usr/bin/esc .
 
 ARG     FILEWATCHER_SHA=2e12ea42f6c8c089b19e992145bb94e8adaecedb
 RUN     go get -d github.com/dnephin/filewatcher && \
         cd /go/src/github.com/dnephin/filewatcher && \
         git checkout -q "$FILEWATCHER_SHA" && \
-        go build -v -o /usr/bin/filewatcher . && \
-        rm -rf /go/src/* /go/pkg/* /go/bin/*
+        go build -v -o /usr/bin/filewatcher .
 
 ENV     CGO_ENABLED=0 \
         PATH=$PATH:/go/src/github.com/docker/cli/build \
```

</details>

```
time docker build -t cli-no-remove --no-cache -f dockerfiles/Dockerfile.dev .   
27.00 real         0.37 user         0.27 sys
```

When looking at the size difference for all of the above:

```
REPOSITORY                      TAG                                            IMAGE ID            CREATED                  SIZE
cli-no-remove                   latest                                         8f83767ba43a        Less than a second ago   342MB
cli-after                       latest                                         4827e97c9f6c        Less than a second ago   314MB
<none>                          <none>                                         c43ffd4abcb4        Less than a second ago   342MB
cli-before                      latest                                         f6ac084fbff0        Less than a second ago   315MB
```


- Current Dockerfile uses 315MB
- Multi-stage build actually uses _more_ disk space, because the intermediate image is also kept (which is 342 MB); there's some overlap between that image and the final image, but it's definitely using more disk space.
- If we're trying to address _speed_, then just removing the `rm -rf` cleanup steps is the fastest (and takes less disk space than the multi-stage builds)


The only situation when the multi-stage build would provide us a real advantage is if one of `github.com/LK4D4/vndr`, `github.com/mjibson/esc`, or `github.com/dnephin/filewatcher` has to be updated, which is infrequent, and total build time is not significant already (unlike the https://github.com/moby/moby dockerfile, which takes a long time in comparisson)


Is there a real issue we're trying to resolve?

dnephin:
Using a multi-stage Dockerfile may allow us to consolidate some of the other Dockerfiles into this one.

thaJeztah:
Time has passed, and we changed to using a multi-stage build for this one in https://github.com/docker/cli/pull/2696 (and various follow-ups after this one).


Let me close this PR, but thanks for contributing!!

PR #945, opened 2018-03-14
Small content trust enhancement

vdemeester:
- `replaceDockerfileForContentTrust` is only used when content trust is
  enabled, so remove the boolean.
- rename `isContentTrustEnabled` to `contentTrustEnabled`

follow-up from #929 

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -297,7 +297,7 @@ func runBuild(dockerCli command.Cli, options buildOptions) error {
 			buildCtx = replaceDockerfileForContentTrust(ctx, buildCtx, relDockerfile, translator, &resolvedTags)
 		} else if dockerfileCtx != nil {
 			// if there was not archive context still do the possible replacements in Dockerfile
-			newDockerfile, _, err := rewriteDockerfileFrom(ctx, dockerfileCtx, translator, !options.untrusted)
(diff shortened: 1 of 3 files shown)


cli/command/image/build.go
```diff
@@ -504,7 +504,7 @@ type resolvedTag struct {
 // "FROM <image>" instructions to a digest reference. `translator` is a
 // function that takes a repository name and tag reference and returns a
 // trusted digest reference.
-func rewriteDockerfileFrom(ctx context.Context, dockerfile io.Reader, translator translatorFunc, istrusted bool) (newDockerfile []byte, resolvedTags []*resolvedTag, err error) {
+func rewriteDockerfileFrom(ctx context.Context, dockerfile io.Reader, translator translatorFunc) (newDockerfile []byte, resolvedTags []*resolvedTag, err error) {
```
dnephin: This should document that it always sets trust enabled, so that it doesn't get accidentally used for something else.

thaJeztah: @dnephin basically it just resolves images to their digest inside a dockerfile, not change "trust" to enabled correct?

dnephin: Right, it doesn't change the global setting, but it does assume that the global setting is enabled.

PR #846, opened 2018-01-30
k8s: watch stack status in addition to pods scheduling

vdemeester:
If the controller updates the stack status to failure, we can stop
watching pods scheduling as it won't happen any time.

I'm not a huge fan of the `watcher` file but hopefully it will be easier to handle that with the next version 👼 

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

PR #948, opened 2018-03-15
K8s: more robust stack error detection on deploy

simonferquel:
This supersedes #846 

This takes advantage of k8s informers (which are much more robust than using only a watch but somewhat more complex as well). 
We look both at stack reconciliation errors, and POD scheduling. The logic is now:
When at least one POD of each service is Ready, we consider the stack to be ready.

On a terminal, service status reporting is now interactive (instead of being forward only), and is correctly interrupted in case of a reconciliation failure.

### DIFF
--- a/cli/command/stack/kubernetes/cli.go
+++ b/cli/command/stack/kubernetes/cli.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/kubernetes"
+	cliv1beta1 "github.com/docker/cli/kubernetes/client/clientset/typed/compose/v1beta1"
 	flag "github.com/spf13/pflag"
 	kubeclient "k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
@@ -112,4 +113,12 @@ func (c *KubeCli) checkHostsMatch() error {
 	fmt.Fprintf(c.Err(), "WARNING: Swarm and Kubernetes hosts do not match (docker host=%s, kubernetes host=%s).\n"+
 		"         Update $DOCKER_HOST (or pass -H), or use 'kubectl config use-context' to match.\n", daemonEndpoint.Hostname(), kubeEndpoint.Hostname())
 	return nil
+}
+
+func (c *KubeCli) stacksv1beta1() (cliv1beta1.StackInterface, error) {
+	raw, err := newStackV1Beta1(c.kubeConfig, c.kubeNamespace)
+	if err != nil {
+		return nil, err
+	}
+	return raw.stacks, nil
 }
--- a/cli/command/stack/kubernetes/deploy.go
+++ b/cli/command/stack/kubernetes/deploy.go
@@ -2,9 +2,12 @@ package kubernetes
 
 import (
 	"fmt"
+	"io"
 
+	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/stack/loader"
 	"github.com/docker/cli/cli/command/stack/options"
+	"github.com/morikuni/aec"
 	"github.com/pkg/errors"
 )
 
@@ -39,10 +42,6 @@ func RunDeploy(dockerCli *KubeCli, opts options.Deploy) error {
 	configMaps := composeClient.ConfigMaps()
 	secrets := composeClient.Secrets()
 	services := composeClient.Services()
-	pods := composeClient.Pods()
-	watcher := DeployWatcher{
-		Pods: pods,
-	}
 
 	if err := stacks.IsColliding(services, stack); err != nil {
 		return err
@@ -61,10 +60,109 @@ func RunDeploy(dockerCli *KubeCli, opts options.Deploy) error {
 	}
 
 	fmt.Fprintln(cmdOut, "Waiting for the stack to be stable and running...")
+	v1beta1Cli, err := dockerCli.stacksv1beta1()
+	if err != nil {
+		return err
+	}
 
-	<-watcher.Watch(stack.name, stack.getServices())
+	pods := composeClient.Pods()
+	watcher := &deployWatcher{
+		stacks: v1beta1Cli,
+		pods:   pods,
+	}
+	statusUpdates := make(chan serviceStatus)
+	displayDone := make(chan struct{})
+	go func() {
+		defer close(displayDone)
+		display := newStatusDisplay(dockerCli.Out())
+		for status := range statusUpdates {
+			display.OnStatus(status)
+		}
+	}()
 
-	fmt.Fprintf(cmdOut, "Stack %s is stable and running\n\n", stack.name)
-
+	err = watcher.Watch(stack.name, stack.getServices(), statusUpdates)
+	close(statusUpdates)
+	<-displayDone
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(cmdOut, "\nStack %s is stable and running\n\n", stack.name)
 	return nil
+
+}
+
+type statusDisplay interface {
+	OnStatus(serviceStatus)
+}
+type metaServiceState string
+
+const (
+	metaServiceStateReady   = metaServiceState("Ready")
+	metaServiceStatePending = metaServiceState("Pending")
+	metaServiceStateFailed  = metaServiceState("Failed")
+)
+
+func metaStateFromStatus(status serviceStatus) metaServiceState {
+	switch {
+	case status.podsReady > 0:
+		return metaServiceStateReady
+	case status.podsPending > 0:
+		return metaServiceStatePending
+	default:
+		return metaServiceStateFailed
+	}
+}
+
+type forwardOnlyStatusDisplay struct {
+	o      *command.OutStream
+	states map[string]metaServiceState
+}
+
+func (d *forwardOnlyStatusDisplay) OnStatus(status serviceStatus) {
+	state := metaStateFromStatus(status)
+	if d.states[status.name] != state {
+		d.states[status.name] = state
+		fmt.Fprintf(d.o, "%s: %s\n", status.name, state)
+	}
+}
+
+type interactiveStatusDisplay struct {
+	o        *command.OutStream
+	statuses []serviceStatus
+}
+
+func (d *interactiveStatusDisplay) OnStatus(status serviceStatus) {
+	b := aec.EmptyBuilder
+	for ix := 0; ix < len(d.statuses); ix++ {
+		b = b.Up(1).EraseLine(aec.EraseModes.All)
+	}
+	b = b.Column(0)
+	fmt.Fprint(d.o, b.ANSI)
+	updated := false
+	for ix, s := range d.statuses {
+		if s.name == status.name {
+			d.statuses[ix] = status
+			s = status
+			updated = true
+		}
+		displayInteractiveServiceStatus(s, d.o)
+	}
+	if !updated {
+		d.statuses = append(d.statuses, status)
+		displayInteractiveServiceStatus(status, d.o)
+	}
+}
+
+func displayInteractiveServiceStatus(status serviceStatus, o io.Writer) {
+	state := metaStateFromStatus(status)
+	totalFailed := status.podsFailed + status.podsSucceeded + status.podsUnknown
+	fmt.Fprintf(o, "%[1]s: %[2]s\t\t[pod status: %[3]d/%[6]d ready, %[4]d/%[6]d pending, %[5]d/%[6]d failed]\n", status.name, state,
+		status.podsReady, status.podsPending, totalFailed, status.podsTotal)
+}
+
+func newStatusDisplay(o *command.OutStream) statusDisplay {
+	if !o.IsTerminal() {
+		return &forwardOnlyStatusDisplay{o: o, states: map[string]metaServiceState{}}
+	}
+	return &interactiveStatusDisplay{o: o}
 }
--- a/cli/command/stack/kubernetes/watcher.go
+++ b/cli/command/stack/kubernetes/watcher.go
@@ -1,116 +1,255 @@
 package kubernetes
 
 import (
-	"fmt"
+	"context"
+	"sync"
 	"time"
 
+	apiv1beta1 "github.com/docker/cli/kubernetes/compose/v1beta1"
 	"github.com/docker/cli/kubernetes/labels"
+	"github.com/pkg/errors"
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	runtimeutil "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	cache "k8s.io/client-go/tools/cache"
+	podutils "k8s.io/kubernetes/pkg/api/v1/pod"
 )
 
+type stackListWatch interface {
+	List(opts metav1.ListOptions) (*apiv1beta1.StackList, error)
+	Watch(opts metav1.ListOptions) (watch.Interface, error)
+}
+
+type podListWatch interface {
+	List(opts metav1.ListOptions) (*apiv1.PodList, error)
+	Watch(opts metav1.ListOptions) (watch.Interface, error)
+}
+
 // DeployWatcher watches a stack deployement
-type DeployWatcher struct {
-	Pods corev1.PodInterface
+type deployWatcher struct {
+	pods   podListWatch
+	stacks stackListWatch
 }
 
 // Watch watches a stuck deployement and return a chan that will holds the state of the stack
-func (w DeployWatcher) Watch(name string, serviceNames []string) chan bool {
-	stop := make(chan bool)
+func (w *deployWatcher) Watch(name string, serviceNames []string, statusUpdates chan serviceStatus) error {
+	errC := make(chan error, 1)
+	defer close(errC)
 
-	go w.waitForPods(name, serviceNames, stop)
+	handlers := runtimeutil.ErrorHandlers
 
-	return stop
+	// informer errors are reported using global error handlers
+	runtimeutil.ErrorHandlers = append(handlers, func(err error) {
+		errC <- err
+	})
+	defer func() {
+		runtimeutil.ErrorHandlers = handlers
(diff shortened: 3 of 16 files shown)


cli/command/stack/kubernetes/deploy.go
```diff
@@ -79,12 +78,114 @@ func RunDeploy(dockerCli *KubeCli, opts options.Deploy) error {
...
+		displayInteractiveServiceStatus(status, d.o)
+	}
+}
+
+func displayInteractiveServiceStatus(status serviceStatus, o io.Writer) {
+	state := metaStateFromStatus(status)
+	color := aec.DefaultF
+	switch state {
+	case metaServiceStateFailed:
+		color = aec.RedF
```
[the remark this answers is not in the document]
simonferquel: This serves as way to identify status changes more clearly (manual testing that just modifying the text was not really clear.
Note that on non-terminal, output is sequencial and without coloring (and don't update quite as much, only when the metastate of a service changes)

cli/command/stack/kubernetes/watcher.go
```diff
@@ -1,117 +1,250 @@
...
+func (w *deployWatcher) Watch(stack *apiv1beta1.Stack, serviceNames []string, statusUpdates chan serviceStatus) error {
+	errC := make(chan error, 1)
+	defer close(errC)
+
+	// informers errors are reported using global error handlers
+	runtimeutil.ErrorHandlers = append(runtimeutil.ErrorHandlers, func(e error) {
+		errC <- e
+	})
+	defer func() {
+		runtimeutil.ErrorHandlers = runtimeutil.ErrorHandlers[:len(runtimeutil.ErrorHandlers)-1]
```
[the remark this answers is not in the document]
simonferquel: That is the main issue with this model, and this is not mutex protected. I changed the code a bit to at least avoid keeping an error handler that writes on a closed channel, but still, it is not the best piece of software coming from K8s

cli/command/stack/kubernetes/watcher.go
```diff
@@ -1,117 +1,250 @@
...
-		if !ready {
+	}
+	oldStatus := pw.services[serviceName]
+	if oldStatus != status {
+		pw.statusUpdates <- status
+	}
+	pw.services[serviceName] = status
+}
+func (pw *podWatcher) allReady() bool {
+	for _, status := range pw.services {
```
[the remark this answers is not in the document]
simonferquel: No, informers guarantee that OnAdd/OnUpdate/OnDelete are called sequencially for a given handler. So there is no concurrency there

cli/command/stack/kubernetes/watcher.go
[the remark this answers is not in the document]
silvin-lubecki: Ok...

cli/command/stack/kubernetes/watcher.go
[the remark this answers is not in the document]
silvin-lubecki: Ok!

simonferquel:
@vdemeester I aggree that splitting display changes would have been easier for reviewing, but the problem I found is that the nature of the service updates we presented to the user previously did not make sense in a K8s world. So it was difficult to map new status updates (with number of PODs per service in a running/failed state) with the old display code (and its weird container restarts thing).
Btw, working on this PR made me realize it would be nice if we would introduce Readyness probes on Compose/Kubernetes (but that is a completely out of scope concern)

simonferquel:
Squashing done

vdemeester:
@simonferquel lint failure — the e2e is getting fixed in https://github.com/docker/cli/pull/955
```
These files were changed:

 M vendor/k8s.io/client-go/tools/cache/store.go
```

simonferquel:
@vdemeester just fixed that. Still the strange perm flag on this file that breaks vendoring from a windows machine

simonferquel:
@vdemeester could you update your review ?

thaJeztah:
ping @simonferquel this needs a rebase 😢

cli/command/stack/kubernetes/deploy.go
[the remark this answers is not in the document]
thaJeztah: I think this should be in a separate PR; so far the CLI does not use any colouring, so this should be looked at in a wider scope, and doesn't seem critical

vendor.conf
```diff
@@ -81,5 +81,4 @@ k8s.io/apimachinery kubernetes-1.8.2
 k8s.io/client-go kubernetes-1.8.2
 k8s.io/kubernetes v1.8.2
 k8s.io/kube-openapi 61b46af70dfed79c6d24530cd23b41440a7f22a5
-vbom.ml/util 928aaa586d7718c70f4090ddf83f2b34c16fdc8d
-github.com/hashicorp/golang-lru 0a025b7e63adc15a622f29b0b2c4c3848243bbf6
\ No newline at end of file
+vbom.ml/util 928aaa586d7718c70f4090ddf83f2b34c16fdc8d
```
thaJeztah: can you squash this change with the other vendor commit?

vendor/github.com/hashicorp/golang-lru/simplelru/lru.go
```diff
@@ -47,7 +47,7 @@ func (c *LRU) Purge() {
 	c.evictList.Init()
 }
 
-// Add adds a value to the cache.  Returns true if an eviction occurred.
+// Add adds a value to the cache.  Returns true if an eviction occured.
```
thaJeztah: Wondering; if this is the only change, if we should bump at all

vendor/github.com/hashicorp/golang-lru/simplelru/lru.go
```diff
@@ -47,11 +47,7 @@ func (c *LRU) Purge() {
 	c.evictList.Init()
 }
 
-<<<<<<< HEAD
```
thaJeztah: Can you move this to the vendor commit?

cli/command/stack/kubernetes/cli.go
```diff
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/kubernetes"
+	cliv1beta1 "github.com/docker/cli/kubernetes/client/clientset/typed/compose/v1beta1"
```
thaJeztah: Can you combine these changes with the other commit?

cli/command/stack/kubernetes/deploy.go
```diff
@@ -178,24 +174,3 @@ func newStatusDisplay(o *command.OutStream) statusDisplay {
 	}
 	return &interactiveStatusDisplay{o: o}
 }
-
-// createFileBasedConfigMaps creates a Kubernetes ConfigMap for each Compose global file-based config.
-func createFileBasedConfigMaps(stackName string, globalConfigs map[string]composetypes.ConfigObjConfig, configMaps corev1.ConfigMapInterface) error {
```
thaJeztah: Same here, please squash this with the other commit

simonferquel:
Removed coloring, and resquashed everything into vendor+code PR style

cli/command/stack/kubernetes/deploy.go
```diff
@@ -61,10 +60,110 @@ func RunDeploy(dockerCli *KubeCli, opts options.Deploy) error {
...
+
+type interactiveStatusDisplay struct {
+	o        *command.OutStream
+	statuses []serviceStatus
+}
+
+func (d *interactiveStatusDisplay) OnStatus(status serviceStatus) {
+	b := aec.EmptyBuilder
+	for ix := 0; ix < len(d.statuses); ix++ {
+		b = b.Up(1).EraseLine(aec.EraseModes.All)
```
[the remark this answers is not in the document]
ijc: The original issue which lead (eventually) to `Gotty` use was #28111 (fixed in #28238, Gotty in followup #28304) solving a problem with `Up(0)` (and `Down(0)` etc) being undefined, but that isn't used here and this `aec` library looks to do the right thing but is just using plain ANSI codes rather than terminfo, so may theoretically not be as portable, although in practice by sticking to ANSI it's probably going to work most reasonable places.

`Gotty` is not maintained (my PRs made as part of the above still aren't merged) whereas `aec` has been more recently (and has no issue or prs open). Maybe it'd be worth moving the `jsonmessage` stuff over to that.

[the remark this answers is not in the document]
thaJeztah: Thanks for looking @ijc - let me open an issue in moby/moby to discuss replacing Gotty

cli/command/stack/kubernetes/deploy.go
```diff
@@ -61,10 +60,110 @@ func RunDeploy(dockerCli *KubeCli, opts options.Deploy) error {
...
+	}
+	if !updated {
+		d.statuses = append(d.statuses, status)
+		displayInteractiveServiceStatus(status, d.o)
+	}
+}
+
+func displayInteractiveServiceStatus(status serviceStatus, o io.Writer) {
+	state := metaStateFromStatus(status)
+	fmt.Fprintf(o, "%s: %s\t\t[pods details: %v ready, %v pending, %v failed]", status.name, state,
```
thaJeztah: - perhaps this should be `[pods status ..` (also, should it be plural (`pods status`) or singular (`pod status`))?
- can you use `%d` for the integers?
- would it be useful to show the total number of pods as well? (`1/5 ready, 2/5 pending, 2/5 failed`)

cli/command/stack/kubernetes/watcher.go
```diff
@@ -1,116 +1,253 @@
...
-func (w DeployWatcher) waitForPods(stackName string, serviceNames []string, stop chan bool) {
-	starts := map[string]int32{}
+// Watch watches a stuck deployement and return a chan that will holds the state of the stack
+func (w *deployWatcher) Watch(name string, serviceNames []string, statusUpdates chan serviceStatus) error {
+	errC := make(chan error, 1)
+	defer close(errC)
+
+	handlers := runtimeutil.ErrorHandlers
+
+	// informers errors are reported using global error handlers
```
thaJeztah: nit: s/informers/informer/`

cli/command/stack/kubernetes/watcher.go
```diff
@@ -1,116 +1,253 @@
...
-		for i := range list.Items {
-			pod := list.Items[i]
-			if pod.Status.Phase != apiv1.PodRunning {
-				continue
-			}
+func (sw *stackWatcher) OnAdd(obj interface{}) {
+	stack, ok := obj.(*apiv1beta1.Stack)
+	switch {
+	case !ok:
+		sw.resultChan <- errors.Errorf("stack %s has not the correct type", sw.stackName)
```
thaJeztah:
```
stack %s has incorrect type
```

or `invalid type` ?

or

```
incorrect type for stack %s 
```

Wondering; should we (can we?) show the actual type? (e.g. if it's a `apiv1beta2.Stack`)?

cli/command/stack/kubernetes/watcher.go
```diff
@@ -1,116 +1,253 @@
...
-	serviceUp := map[string]bool{}
+var _ cache.ResourceEventHandler = &podWatcher{}
 
-	for _, pod := range pods {
-		if time.Since(pod.GetCreationTimestamp().Time) < 10*time.Second {
-			return false
+func (pw *podWatcher) handlePod(obj interface{}) {
+	pod, ok := obj.(*apiv1.Pod)
+	if !ok {
+		pw.resultChan <- errors.New("unexpected type for a Pod")
```
thaJeztah: "for a Pod" seems a bit vague; can we print the Pod's name (or something?)

```
unexpected type for pod %s
``` 

Also, for consistency with stacks above; we should pick the same wording (`unexpected type`, `invalid type`, or `incorrect type`) - pick one (not sure which one's best)

cli/command/stack/kubernetes/watcher.go
```diff
@@ -1,116 +1,253 @@
...
+func (pw *podWatcher) OnDelete(obj interface{}) {
+	pw.handlePod(obj)
+}
+
+func (w *deployWatcher) waitForPods(ctx context.Context, stackName string, serviceNames []string, e chan error, statusUpdates chan serviceStatus) {
+	informer := newPodInformer(w.pods, stackName, cache.Indexers{
+		"byservice": func(obj interface{}) ([]string, error) {
+			pod, ok := obj.(*apiv1.Pod)
+			if !ok {
+				return nil, errors.New("pod has an unexpected type")
```
thaJeztah: Use same wording here as above (or vice-versa)

cli/command/stack/kubernetes/watcher.go
```diff
@@ -1,116 +1,253 @@
...
+		indexers,
+	)
+}
 
-	return "times"
+func newStackInformer(stacksClient stackListWatch, stackName string) cache.SharedInformer {
+	return cache.NewSharedInformer(
+		&cache.ListWatch{
+			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+				options.LabelSelector = "com.docker.stack.namespace=" + stackName
```
thaJeztah: Looks like we have defined the name for this label as a constant in two locations;

https://github.com/docker/cli/blob/236a84759af68d2f2a7d216f0708152a896229ac/cli/compose/convert/compose.go#L13-L16

and

https://github.com/docker/cli/blob/7ad30360c823c4f7748e1ac7ef98ec21414cb691/kubernetes/labels/labels.go#L8-L15

Can you use a constant here? (Also, we should discuss unifying those constants, but not for this PR)

cli/command/stack/kubernetes/watcher.go
```diff
@@ -1,116 +1,253 @@
...
+				return stacksClient.List(options)
+			},
+
+			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+				options.LabelSelector = "com.docker.stack.namespace=" + stackName
```
thaJeztah: Same here (constant)

cli/command/stack/kubernetes/watcher_test.go
```diff
@@ -0,0 +1,222 @@
...
+		if label.Matches(labels.Set(item.Labels)) {
+			list.Items = append(list.Items, item)
+		}
+	}
+	return list, err
+
+}
+func (p *testPodListWatch) Watch(opts metav1.ListOptions) (watch.Interface, error) {
+	return p.fake.
+		InvokesWatch(k8stesting.NewWatchAction(podsResource, p.ns, opts))
```
thaJeztah: nit: wondering why you split these (and other locations) over two lines, instead of

```go
return s.fake.InvokesWatch(k8stesting.NewWatchAction(stacksResource, s.ns, opts))
```

(don't think we do that in most places, but may be my personal preference, so just a nit)

cli/command/stack/kubernetes/deploy.go
```diff
@@ -61,10 +60,110 @@ func RunDeploy(dockerCli *KubeCli, opts options.Deploy) error {
...
+	o      *command.OutStream
+	states map[string]metaServiceState
+}
+
+func (d *forwardOnlyStatusDisplay) OnStatus(status serviceStatus) {
+	state := metaStateFromStatus(status)
+	if d.states[status.name] != state {
+		d.states[status.name] = state
+		fmt.Fprintf(d.o, "%s: %s", status.name, state)
+		fmt.Fprintln(d.o)
```
thaJeztah: Can you add a `\n` to the previous `fmt.Fprintf()` instead?

```go
fmt.Fprintf(d.o, "%s: %s\n", status.name, state)
```

cli/command/stack/kubernetes/deploy.go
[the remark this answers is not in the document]
simonferquel: Fyi, I chose this lib because of its simplicity, and because it is already used in buildkit (which has IMO a very modern and good looking UX)

cli/command/stack/kubernetes/deploy.go
simonferquel: This is for consistency with the platform EOL conventions (`\n` on *nix, `\r\n` on windows). But sure, I can simplify that.

cli/command/stack/kubernetes/watcher.go
simonferquel: Honestly, this error should really not happen (we are just covering for potential bugs in the way we initialize the informer). I agree we should use a consistent wording, but adding extra-work for adding more info in the message seem a bit overkill

simonferquel:
@thaJeztah Squashing done! :)

cli/command/stack/kubernetes/deploy.go
```diff
@@ -61,10 +60,109 @@ func RunDeploy(dockerCli *KubeCli, opts options.Deploy) error {
...
+	if !updated {
+		d.statuses = append(d.statuses, status)
+		displayInteractiveServiceStatus(status, d.o)
+	}
+}
+
+func displayInteractiveServiceStatus(status serviceStatus, o io.Writer) {
+	state := metaStateFromStatus(status)
+	totalFailed := status.podsFailed + status.podsSucceeded + status.podsUnknown
+	fmt.Fprintf(o, "%[1]s: %[2]s\t\t[pod status: %[3]d/%[6]d ready, %[4]d/%[6]d pending, %[5]d/%[6]d failed]\n", status.name, state,
```
silvin-lubecki: I got this:
```
backend: Failed         [pod status: 0/0 ready, 0/0 pending, 0/0 failed]
cpp: Failed             [pod status: 0/0 ready, 0/0 pending, 0/0 failed]
dockerfile: Failed              [pod status: 0/0 ready, 0/0 pending, 0/0 failed]
golang: Pending         [pod status: 0/1 ready, 1/1 pending, 0/1 failed]
python: Failed          [pod status: 0/0 ready, 0/0 pending, 0/0 failed]
```
Maybe it's missing a tab between the service name and the state?
```go
fmt.Fprintf(o, "%[1]s:\t%[2]s
```
Or maybe it's more complicated than that?

thaJeztah: Hm, looks like we need (the equivalent of) a tabwriter for that; guess the problem is that `dockerfile: Failed` would just reach a tab-stop, therefore the tab that _is_ being written bumps it to the next tab-stop.

silvin-lubecki: Ok that's what I had in mind...

thaJeztah: I think that would involve;

Instead of clearing + overwriting each line separately; buffer all lines first, then clear all lines and write them (using a tabwriter)

PR #955, opened 2018-03-21
Remove filewatcher

dnephin:
I'm not sure if anyone else uses this thing. If someone else is using it I can change this PR to just update it to v0.2.1.

Something bad happened to the gopkg.in url for fsnotify (https://gopkg.in/fsnotify.v1) , which is imported by filewatcher before version 0.2.1. So this will cause builds to fail until it's either updated or removed.

### DIFF
--- a/dockerfiles/Dockerfile.dev
+++ b/dockerfiles/Dockerfile.dev
@@ -17,13 +17,6 @@ RUN     go get -d github.com/mjibson/esc && \
         go build -v -o /usr/bin/esc . && \
         rm -rf /go/src/* /go/pkg/* /go/bin/*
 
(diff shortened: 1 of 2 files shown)


thaJeztah:
Looks like it was done deliberately;

> go-fsnotify/fsnotify was the original name of this project. The project organization was renamed from go-fsnotify to fsnotify some time ago, meaning Github created a silent redirect to the new organization name. This behavior is dangerous since it allows anyone to register the go-fsnotify account and create a fsnotify repository, which will eventually be pulled by projects that have dependencies to it. Dependency squatting can remain undetected for a long period of time because Github silently redirects git pulls, reducing breakage

PR #957, opened 2018-03-21
Docs touch-ups for "autoremove" and broken anchor

thaJeztah:
`--rm` moved to the daemon, so is now also supported
when combined with `-d`.

ping @gbarr01 PTAL


thaJeztah:
author LGTM

PR #964, opened 2018-03-26
Fix alignment issue in the version command output

kjagiello:
**- What I did**
This tiny PR fixes an alignment issue in the version command output.

Before:
```
$ docker version
Client:
 Version:       18.03.0-ce
 API version:   1.37
 Go version:    go1.9.4
 Git commit:    0520e24
 Built: Wed Mar 21 23:06:22 2018
 OS/Arch:       darwin/amd64
 Experimental:  false
 Orchestrator:  swarm

Server:
 Engine:
  Version:      18.03.0-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.4
  Git commit:   0520e24
  Built:        Wed Mar 21 23:14:32 2018
  OS/Arch:      linux/amd64
  Experimental: false
```

After:
```$ build/docker-darwin-amd64 version
Client:
 Version:       18.04.0-dev
 API version:   1.37
 Go version:    go1.9.4
 Git commit:    241c14d1
 Built:         Mon Mar 26 13:34:21 2018
 OS/Arch:       darwin/amd64
 Experimental:  false
 Orchestrator:  swarm

Server:
 Engine:
  Version:      18.03.0-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.4
  Git commit:   0520e24
  Built:        Wed Mar 21 23:14:32 2018
  OS/Arch:      linux/amd64
  Experimental: false
```

PR #913, opened 2018-03-01
Align `Built` in docker version output Client section

jhowardmsft:
Signed-off-by: John Howard <jhoward@microsoft.com>

Somewhat amazed this hadn't been noticed/fixed before.....

Before
image

image

After
image

PR #965, opened 2018-03-26
Fix docker version output alignment

thaJeztah:
closes https://github.com/docker/cli/pull/964
closes https://github.com/docker/cli/pull/769
closes https://github.com/docker/cli/pull/913

Use tabwriter to print the version output

@tiborvass I tried using a tab writer, and this seems to work correctly; do you know what bug you ran into?

Output with this change:

```
Client:
 Version:      18.04.0-dev
 API version:  1.37
 Go version:   go1.9.4
 Git commit:   76439457
 Built:        Mon Mar 26 14:10:50 2018
 OS/Arch:      linux/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.03.0-ce-rc4
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.4
  Git commit:   fbedb97
  Built:        Thu Mar 15 07:42:29 2018
  OS/Arch:      linux/amd64
  Experimental: true
```

### DIFF
--- a/cli/command/system/version.go
+++ b/cli/command/system/version.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"runtime"
 	"sort"
+	"text/tabwriter"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
Or do we want "client" and "server" columns to be aligned? (guess we can do so by making the "min" width bigger)

cli/command/system/version.go
```diff
@@ -195,11 +197,14 @@ func runVersion(dockerCli command.Cli, opts *versionOptions) error {
...
 	}
-
-	if err2 := tmpl.Execute(dockerCli.Out(), vd); err2 != nil && err == nil {
+	t := tabwriter.NewWriter(dockerCli.Out(), 15, 1, 1, ' ', 0)
+	buffer := bytes.NewBufferString("")
+	if err2 := tmpl.Execute(buffer, vd); err2 != nil && err == nil {
 		err = err2
 	}
-	dockerCli.Out().Write([]byte{'\n'})
+	buffer.WriteTo(t)
```
dnephin: Why do you need a buffer here? Can't you use `t` as the first arg to `tmpl.Execute` ?

thaJeztah: Oh! was debugging, looks like I don't need it here anymore no.

PR #974, opened 2018-03-27
[K8s] Do env-variable expansion on the uninterpreted Config files

simonferquel:
This fix an issue where composefiles like
```
version: '3'
services:
  front:
    image: nginx:1.12.1-alpine
      ports:
      - 80
```
get translated into
```
version: "3.0"
services:
  front:
    image: nginx:1.12.1-alpine
      ports:
      - mode: ingress
        target: 80
        protocol: tcp
```

The problem here is that the long format for describing ports has been
introduced in 3.2 and so, the generated compose file is invalid.

I thought first about just always use latest version information to
override the output composefile version, but that would just make the
problem worse once we introduce a new compose file schema in the CLI
that is not supported on the server side.

This PR makes it possible to interpolate the compose files without
converting them into composetypes.Config (but only work on the loaded
config file format: `map[string]interface{}`. This way, we don't expand
short representations into long ones, and don't risk breaking original
compose files.

### DIFF
--- a/cli/command/stack/kubernetes/stackclient.go
+++ b/cli/command/stack/kubernetes/stackclient.go
@@ -4,10 +4,14 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/docker/cli/cli/compose/loader"
+	"github.com/docker/cli/cli/compose/schema"
 	composetypes "github.com/docker/cli/cli/compose/types"
 	composev1beta1 "github.com/docker/cli/kubernetes/client/clientset/typed/compose/v1beta1"
 	composev1beta2 "github.com/docker/cli/kubernetes/client/clientset/typed/compose/v1beta2"
+	v1beta1types "github.com/docker/cli/kubernetes/compose/v1beta1"
 	"github.com/docker/cli/kubernetes/labels"
+	"github.com/pkg/errors"
 	yaml "gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -105,6 +109,7 @@ func verify(services corev1.ServiceInterface, stackName string, service string) 
 }
 
 func (s *stackV1Beta1) FromCompose(stderr io.Writer, name string, cfg *composetypes.Config) (stack, error) {
+	cfg.Version = v1beta1types.MaxComposeVersion
 	st, err := fromCompose(stderr, name, cfg)
 	if err != nil {
 		return stack{}, err
@@ -113,6 +118,15 @@ func (s *stackV1Beta1) FromCompose(stderr io.Writer, name string, cfg *composety
 	if err != nil {
 		return stack{}, err
 	}
+	// reload the result to check that it produced a valid 3.5 compose file
+	resparsedConfig, err := loader.ParseYAML(res)
+	if err != nil {
+		return stack{}, err
+	}
+	if err = schema.Validate(resparsedConfig, v1beta1types.MaxComposeVersion); err != nil {
+		return stack{}, errors.Wrapf(err, "the compose yaml file is invalid with v%s", v1beta1types.MaxComposeVersion)
+	}
+
 	st.composeFile = string(res)
 	return st, nil
 }
--- a/cli/command/stack/kubernetes/stackclient_test.go
+++ b/cli/command/stack/kubernetes/stackclient_test.go
@@ -26,7 +26,7 @@ func TestFromCompose(t *testing.T) {
 	})
 	assert.NilError(t, err)
 	assert.Equal(t, "foo", s.name)
-	assert.Equal(t, string(`version: "3.1"
+	assert.Equal(t, string(`version: "3.5"
 services:
   bar:
     image: bar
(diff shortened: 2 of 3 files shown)


simonferquel:
ping @mnottale @chris-crone @vdemeester

dnephin:
This is an excellent example of why the Compose file YAML should not be used as part of the API. The API should have its own types and the client should translate the compose file representation into the API types (just like with swarm mode). That way updates to the Compose file on the client don't unexpected break the API.

simonferquel:
@dnephin agreed. That is also why we did that for v1beta2 :)

cli/command/stack/kubernetes/deploy.go
```diff
@@ -21,12 +22,27 @@ func RunDeploy(dockerCli *KubeCli, opts options.Deploy) error {
 		return errors.Errorf("Please specify only one compose file (with --compose-file).")
 	}
 
+	// due to env var expansion logic (and the need to conserve the same compose version as original file)
```
vdemeester: > the need to conserve the same compose version as original file
Do we need that ?

simonferquel: At the beginning, I wanted to just override the original composefile version with the latest supported by the cli (in order to align the "long-formats" generating by marshalling composetypes.Config with the version it is supported in), but after a bit of thinking I found that it was a bad idea: if in a future version of the CLI we introduce a new compose schema version, we will generate composefiles that the server-side can't understand if not updated (and we don't want to break compatibility with Docker EE for example).

vdemeester: @simonferquel this is only for `v1beta1` api (compose controller) right ? we know which version we support on that release of the controller, so really, we could, for this case only, decide of a version and *never* update it (until we drop support for `v1beta1` :angel: ). It's a special case anyway so let's make the "fix" the simplest.

simonferquel: I don't think keeping an old config structure and implementing conversion from latest to this reference version is simpler than just doing the variable expansion on non-interpreted config files (this PR basically just expose exiting code to the kubernetes subpackage)

vdemeester:
cc @silvin-lubecki @thaJeztah

silvin-lubecki:
> That means we can/should know for sure in the CLI, which version of the composefile format we can and cannot send to the v1beta1 endpoint. As long format for describing ports has been introduced in 3.2, we know for a fact that v1beta1 supports that too. That means, we can/should send the compose file to the v1beta1 as version: "3.4" all the time if necessary (or at least 3.2 if it's lower than that).
👍 on this.
I think the CLI can error out if a Compose file with greater version than v1beta1 supports is used.

simonferquel:
Ok, so we would have to be very careful anytime the composetypes.Config (or any of its nested struct) is updated, that we don't replace constructs from v3.5 with a different representation (e.g. that we only add fields). (v3.5 is the latest format supported in ee)

simonferquel:
Done that. Also added a v3.5 schema validation of the resulting composefile after marshalling the composetypes.Config

simonferquel:
@vdemeester does it fit you ?

simonferquel:
0.1.x only shipped with a private beta of Docker for Mac, we don't want to support it any way (and docker for desktop ships its own version of the docker CLI aligned with the engine/kubernetes integration anyway)

vdemeester:
> 0.1.x only shipped with a private beta of Docker for Mac, we don't want to support it any way (and docker for desktop ships its own version of the docker CLI aligned with the engine/kubernetes integration anyway)

Right, but the image is there and public (I think ?) so it "could" still be used. I know we don't _support_ it (as of Docker inc.), but if the there is no need to use 3.5 over 3.4 for `v1beta1`, it's the safest choice 👼

cli/command/stack/kubernetes/loader.go
```diff
@@ -1,18 +1,34 @@
...
+	"github.com/docker/cli/cli/compose/loader"
+	"github.com/docker/cli/cli/compose/schema"
 	composetypes "github.com/docker/cli/cli/compose/types"
 	apiv1beta1 "github.com/docker/cli/kubernetes/compose/v1beta1"
+	"github.com/pkg/errors"
 	yaml "gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const v1beta1MaxComposeVersion = "3.5"
```
silvin-lubecki: Can we move this constant to the apiv1beta1 package? It should be the package responsibility to know which compose version it can handle.

silvin-lubecki: And maybe we should add one also for v1beta2, and generalize the process?

cli/command/stack/kubernetes/loader.go
```diff
@@ -1,18 +1,34 @@
...
 		return nil, err
 	}
+
+	// reload the result to check that it produced a valid 3.5 compose file
+	resparsedConfig, err := loader.ParseYAML(res)
+	if err != nil {
+		return nil, err
+	}
+	if err = schema.Validate(resparsedConfig, v1beta1MaxComposeVersion); err != nil {
+		return nil, errors.Wrapf(err, "the generated compose yaml file is invalid with v%s", v1beta1MaxComposeVersion)
```
silvin-lubecki: Remove "generated" in the message? I don't know if it will be a useful information for the user?

cli/command/stack/kubernetes/loader_test.go
```diff
@@ -31,7 +31,7 @@ func TestLoadStack(t *testing.T) {
 			Name: "foo",
 		},
 		Spec: apiv1beta1.StackSpec{
-			ComposeFile: `version: "3.1"
+			ComposeFile: `version: "3.5"
```
silvin-lubecki: Add a test with a compose file which fails with a 3.6 version?

cli/command/stack/kubernetes/loader.go
silvin-lubecki: Good point 😅

simonferquel:
@silvin-lubecki I've moved the const declaration and added a test for v3.6 files

thaJeztah:
ping @simonferquel this needs a rebase

PR #990, opened 2018-04-09
Import TestBuildIidFileSquash from moby to cli

vdemeester:
It's a cli only feature so the test belongs to the cli.

See https://github.com/moby/moby/pull/36746/files#diff-64d1c3a97fd8b1abf170680883e81aaaL6241

cc @dnephin 

It also make test-e2e run against experimental and non-experimental daemon.

### DIFF
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -119,5 +119,12 @@ shellcheck: build_shell_validate_image
 	docker run -ti --rm $(ENVVARS) $(MOUNTS) $(VALIDATE_IMAGE_NAME) make shellcheck
 
 .PHONY: test-e2e
-test-e2e: build_e2e_image
+test-e2e: test-e2e-non-experimental test-e2e-experimental
+
+.PHONY: test-e2e-experimental
+test-e2e-experimental: build_e2e_image
+	docker run -e DOCKERD_EXPERIMENTAL=1 --rm -v /var/run/docker.sock:/var/run/docker.sock $(E2E_IMAGE_NAME)
+
+.PHONY: test-e2e-non-experimental
+test-e2e-non-experimental: build_e2e_image
 	docker run --rm -v /var/run/docker.sock:/var/run/docker.sock $(E2E_IMAGE_NAME)
--- a/e2e/compose-env.experimental.yaml
+++ b/e2e/compose-env.experimental.yaml
@@ -1,0 +1,6 @@
+version: '2.1'
+
+services:
+  engine:
+      command: ["--insecure-registry=registry:5000", "--experimental"]
+
--- a/e2e/image/build_test.go
+++ b/e2e/image/build_test.go
@@ -2,11 +2,16 @@ package image
 
 import (
 	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/docker/cli/e2e/internal/fixtures"
 	"github.com/docker/cli/internal/test/environment"
 	"github.com/docker/cli/internal/test/output"
+	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
 	"gotest.tools/fs"
 	"gotest.tools/icmd"
 	"gotest.tools/skip"
@@ -101,6 +106,33 @@ func TestTrustedBuildUntrustedImage(t *testing.T) {
 		ExitCode: 1,
 		Err:      "does not have trust data for",
 	})
+}
+
+func TestBuildIidFileSquash(t *testing.T) {
+	environment.SkipIfNotExperimentalDaemon(t)
+	dir := fs.NewDir(t, "test-iidfile-squash")
+	defer dir.Remove()
+	iidfile := filepath.Join(dir.Path(), "idsquash")
+	buildDir := fs.NewDir(t, "test-iidfile-squash-build",
+		fs.WithFile("Dockerfile", fmt.Sprintf(`
+	FROM %s
+	ENV FOO FOO
+	ENV BAR BAR
+	RUN touch /foop`, fixtures.AlpineImage)),
+	)
+	defer buildDir.Remove()
+
+	imageTag := "testbuildiidfilesquash"
+	result := icmd.RunCmd(
+		icmd.Command("docker", "build", "--iidfile", iidfile, "--squash", "-t", imageTag, "."),
+		withWorkingDir(buildDir),
+	)
+	result.Assert(t, icmd.Success)
+	id, err := ioutil.ReadFile(iidfile)
+	assert.NilError(t, err)
+	result = icmd.RunCommand("docker", "image", "inspect", "-f", "{{.Id}}", imageTag)
+	result.Assert(t, icmd.Success)
+	assert.Check(t, is.Equal(string(id), strings.TrimSpace(result.Combined())))
 }
 
 func withWorkingDir(dir *fs.Dir) func(*icmd.Cmd) {
--- a/e2e/image/build_test.go
+++ b/e2e/image/build_test.go
@@ -118,6 +118,7 @@ func TestBuildIidFileSquash(t *testing.T) {
 	FROM %s
 	ENV FOO FOO
 	ENV BAR BAR
+	RUN touch /fiip
 	RUN touch /foop`, fixtures.AlpineImage)),
 	)
 	defer buildDir.Remove()
--- a/scripts/test/e2e/run
+++ b/scripts/test/e2e/run
@@ -15,7 +15,10 @@ function fetch_images {
 
 function setup {
     local project=$1
-    COMPOSE_PROJECT_NAME=$1 COMPOSE_FILE=$2 docker-compose up --build -d >&2
+    local file=$2
+
+    test "${DOCKERD_EXPERIMENTAL:-}" -eq "1" && file="${file}:./e2e/compose-env.experimental.yaml"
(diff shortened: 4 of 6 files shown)


e2e/compose-env.yaml
```diff
@@ -7,7 +7,7 @@ services:
   engine:
       image: 'docker:${TEST_ENGINE_VERSION:-edge-dind}'
       privileged: true
-      command: ['--insecure-registry=registry:5000']
+      command: ['--insecure-registry=registry:5000', '--experimental']
```
cpuguy83: Doesn't this change the whole test suite?

vdemeester: @cpuguy83 yes it does. I think it's fine right now, but we way want to run e2e test against experimental and non-experimental job, but then the changes would be heavier.

cpuguy83: The test would also need to ensure that experimental is enabled.
I think it may be prudent to support both modes in e2e and then skip tests that require experimental on non-experimental daemons rather than changing the suite for this test.

thaJeztah:
ping @cpuguy83 this good to go, or want changes w.r.t. the experimental option?

cpuguy83:
I'm more worried about changing to only testing experimental than I am about having this test.

thaJeztah:
ok, let's hold off merging to see if we can find a solution for that first

vdemeester:
@cpuguy83 @thaJeztah should I go for "skip tests that require experimental on non-experimental" and in a follow-up update the e2e run to support being run in `experimental` or not ?

thaJeztah:
SGTM for now (if we do the reverse as well, i.e. skip regular tests if experimental is enabled), we could just run them after each other?

vdemeester:
Updated

```bash
# […]
--- SKIP: TestBuildIidFileSquash (0.01s)
	testenv.go:40: !info.ExperimentalBuild: running against a non-experimental daemon
# […]
```

vdemeester:
And now with an additionnal commit : 

```
Make test-e2e run against experimental and non-experimental daemon

- `make test-e2e` runs the e2e tests twice : once against on
  non-experimental daemon (as before), once against an experimental
  daemon.
- adds `test-e2e-experimental` and `test-e2e-non-experimental` target
  to run tests for the specified cases
```

docker.Makefile
```diff
@@ -114,5 +114,13 @@ shellcheck: build_shell_validate_image
 	docker run -ti --rm $(ENVVARS) $(MOUNTS) $(VALIDATE_IMAGE_NAME) make shellcheck
 
 .PHONY: test-e2e
-test-e2e: binary
+test-e2e: test-e2e-non-experimental test-e2e-experimental
```
thaJeztah: Would be nice to have a `make help` in this makefile as well (definitely not for this PR though :smile:)

dockerfiles/Dockerfile.test-e2e-env
```diff
@@ -13,5 +13,6 @@ ENV     DISABLE_WARN_OUTSIDE_CONTAINER=1
 WORKDIR /work
 COPY    scripts/test/e2e scripts/test/e2e
 COPY    e2e/compose-env.yaml e2e/compose-env.yaml
+COPY    e2e/compose-env.yaml.experimental e2e/compose-env.yaml.experimental
```
thaJeztah:
```
compose-env-experimental.yaml
```

e2e/compose-env.yaml.experimental
```diff
@@ -0,0 +1,6 @@
+version: '2.1'
```
thaJeztah: See above:

```
s/compose-env.yaml.experimental/compose-env-experimental.yaml/
```

e2e/compose-env.yaml.experimental
```diff
@@ -0,0 +1,6 @@
...
+
+services:
+  engine:
+      command: ['--insecure-registry=registry:5000', '--experimental']
```
thaJeztah: Should this be with double-quotes (JSON notation)?

```json
 ["--insecure-registry=registry:5000", "--experimental"]
```

dockerfiles/Dockerfile.test-e2e-env
vdemeester: I used that to be able to do the `file="${file}:${file}.experimental"` hack :stuck_out_tongue:

vdemeester:
@thaJeztah this one should be ready, it does two e2e runs, one with experimental, the other without.

scripts/test/e2e/run
```diff
@@ -15,7 +15,11 @@ function fetch_images {
 
 function setup {
     local project=$1
-    COMPOSE_PROJECT_NAME=$1 COMPOSE_FILE=$2 docker-compose up --build -d >&2
+    local file=./e2e/compose-env.yaml
+
+    DOCKERD_EXPERIMENTAL="${DOCKERD_EXPERIMENTAL:-}"
+    test -n "${DOCKERD_EXPERIMENTAL}" && file="${file}:./e2e/compose-env.experimental.yaml"
```
ijc: The calling code uses `DOCKERD_EXPERIMENTAL=0`, which also passes the test above -- not what you wanted I think?

scripts/test/e2e/wrapper
```diff
@@ -2,10 +2,11 @@
 # Setup, run and teardown e2e test suite in containers.
 set -eu -o pipefail
 
+echo "exp $DOCKERD_EXPERIMENTAL"
```
ijc: Is this leftover debug? If not then due to `set -u` above if `DOCKERD_EXPERIMENTAL` is not set it'll error out I think.

scripts/test/e2e/run
vdemeester: :man_facepalming:

vdemeester:
@thaJeztah @ijc updated :sweat_smile:

scripts/test/e2e/run
```diff
@@ -15,7 +15,11 @@ function fetch_images {
 
 function setup {
     local project=$1
-    COMPOSE_PROJECT_NAME=$1 COMPOSE_FILE=$2 docker-compose up --build -d >&2
+    local file=./e2e/compose-env.yaml
+
+    DOCKERD_EXPERIMENTAL="${DOCKERD_EXPERIMENTAL:-}"
+    test "${DOCKERD_EXPERIMENTAL}" -eq "1" && file="${file}:./e2e/compose-env.experimental.yaml"
```
ijc: Unless you wanted to subsequently `export` it you could inline the `${DOCKERD_EXPERIMENTAL:-}` into the usage in the `test`, but that's a very minor nit.

scripts/test/e2e/run
```diff
@@ -15,7 +15,11 @@ function fetch_images {
 
 function setup {
     local project=$1
-    COMPOSE_PROJECT_NAME=$1 COMPOSE_FILE=$2 docker-compose up --build -d >&2
+    local file=./e2e/compose-env.yaml
```
ijc: Previously `$file` would (logically) have been `$2` but here you seem to have expanded/inlined it without changing the caller AFAICT.

e2e/image/build_test.go
```diff
@@ -103,6 +108,33 @@ func TestTrustedBuildUntrustedImage(t *testing.T) {
...
+	environment.SkipIfNotExperimentalDaemon(t)
+	dir := fs.NewDir(t, "test-iidfile-squash")
+	defer dir.Remove()
+	iidfile := filepath.Join(dir.Path(), "idsquash")
+	buildDir := fs.NewDir(t, "test-iidfile-squash-build",
+		fs.WithFile("Dockerfile", fmt.Sprintf(`
+	FROM %s
+	ENV FOO FOO
+	ENV BAR BAR
+	RUN touch /foop`, fixtures.AlpineImage)),
```
thaJeztah: Looks like this Dockerfile will only add a single layer; should we have something that actually has multiple layers before squashing?

(If there's some specific case we're testing for, we may need a comment to describe that case)

e2e/compose-env.experimental.yaml
```diff
@@ -0,0 +1,6 @@
+version: '2.1'
+
+services:
+  engine:
+      command: ['--insecure-registry=registry:5000', '--experimental']
```
thaJeztah: looks like my previous comment got collapsed; https://github.com/docker/cli/pull/990#discussion_r189855044

(nit); was wondering if we should prefer to use double-quotes here to be consistent with the Dockerfile format

e2e/image/build_test.go
vdemeester: Oh good point. I took the test case from `moby/moby` but yeah we could add one more `RUN` to it :wink:

ijc: The original issue was https://github.com/moby/moby/issues/33822, which lead to https://github.com/moby/moby/pull/33824.

Rereading that I think the number of layers was not the issue, just the wrong thing was being returned.

e2e/compose-env.experimental.yaml
vdemeester: Ah, I don't really care so I can switch to `"` :wink:

vdemeester:
@thaJeztah updated :wind_chime:

e2e/image/build_test.go
thaJeztah: Ah, thanks!

@vdemeester @ijc do think it would be good to add that as a comment/godoc for the test, so that the original intent doesn't get lost?

PR #973, opened 2018-03-27
Orchestrator column

silvin-lubecki:
**- What I did**
To perform administration tasks more easily, I added an ORCHESTRATOR column in the `docker stack ls` command.

⚠️ Depends on #903  ⚠️ 

**- How I did it**
I added a column in the Stack formatter, and hardcoded the orchestrator in each list command implementation.

**- How to verify it**
```
$ DOCKER_ORCHESTRATOR=kubernetes docker stack ls
NAME                SERVICES            ORCHESTRATOR
mystack             3                   Kubernetes

$ DOCKER_ORCHESTRATOR=swarm docker stack ls
NAME                SERVICES            ORCHESTRATOR
mystack             4                   Swarm
```

**- Description for the changelog**
Added ORCHESTRATOR column to docker stack ls command 

**- A picture of a cute animal (not mandatory but encouraged)**
image

PR #991, opened 2018-04-09
Better namespace experience with Kubernetes

silvin-lubecki:
**- What I did**
I redesigned a better UX for namespaces with Kubernetes:
- the default namespace defined in the `~/.kube/config` is now used with all stack commands
- a `NAMESPACE` column is added for `docker stack ls` command on Kubernetes orchestrator only
- a `--all-namespaces` flag is added for `docker stack ls` command on Kubernetes orchestrator only

**- How I did it**

1. For the default namespace, I modified the kubernetes/config.go:NewKubernetesConfig function to return a lower level config object, which can resolve an optional namespace defined in the kubernetes configuration file.
1. For the `NAMESPACE` column, I modified the formatter/stack.go:NewStackFormat function, which is now almost no-op. We now have two default format, a Kubernetes one with a NAMESPACE column, and a Swarm one without it, so I let the caller choosing the appropriate default table format. Buy the way now these two default format are exported. A closer look will be appreciated :smile: 
1. For the `--all-namespaces` flag, I patched the Stack client to handle the all-namespaces option on List command. All stack clients are namespaced, but with Kubernetes, setting the namespace to empty string means all namespaces.

**- How to verify it**
```bash
# Use Kubernetes by default for every following commands
$ export DOCKER_ORCHESTRATOR=kubernetes

# Deploy a stack on default namespace
$ docker stack deploy mystack -c docker-compose.yml
...
 
# List the stacks in the default namespace, it will show a new NAMESPACE column
$ docker stack ls
mystack             3                   Kubernetes          default

# Swarm hasn't any concept of namespace, so no NAMESPACE column
$ DOCKER_ORCHESTRATOR=swarm docker stack ls
NAME                SERVICES            ORCHESTRATOR

# Create a namespace on Kubernetes:
$ kubectl create namespace mynamespace
namespace "mynamespace" created

# Deploy a stack on that namespace
$ docker stack deploy mystack --namespace mynamespace -c docker-compose.yml
...

# List all stacks in all namespaces using --all-namespace flag
$ docker stack ls --all-namespaces
mystack             3                   Kubernetes          default
mystack             1                   Kubernetes          mynamespace

# Edit your ~/.kube/config file to add a different default namespace to your current context
$ cat ~/.kube/config
...
contexts:
- context:
    cluster: docker-for-desktop-cluster
    namespace: mynamespace
    user: docker-for-desktop
  name: docker-for-desktop
...

# List again the stacks in the default namespace
$ docker stack ls
mystack             3                   Kubernetes          mynamespace
```

- Fix using namespace defined in ~/.kube/config for stack commands
- Add a NAMESPACE column for docker stack ls command
- Add a --all-namespaces flag for docker stack ls command

image

⚠️ Depends on #973 ⚠️


silvin-lubecki:
Yes, I'm on it, thanks @thaJeztah!

silvin-lubecki:
PTAL

mat007:
Rebased!

vieux:
@silvin-lubecki what's the easiest way for me to test this ?

mat007:
@vieux silvin is on holiday, I'm carrying his PRs meanwhile.
The most simple means to install Kubernetes would be either through Docker for Desktop (Mac and Windows) or probably UCP (linux), after which following the «How to verify it» steps should work.
In Docker for Desktop there is only a checkbox to toggle in order to enable and install Kubernetes, I'm afraid I don't know how to proceed for UCP and linux, maybe @vdemeester has some experience with this?

vdemeester:
On non docker-for-desktop, you need to have `orchestrator: "kubernetes"` and `experimental: "enabled"` in the client configuration (`$HOME/.docker/config.json`) — and a kube configuration pointing to a live kubernetes cluster.

cli/command/stack/kubernetes/cli.go
```diff
@@ -57,7 +53,19 @@ func WrapCli(dockerCli command.Cli, opts Options) (*KubeCli, error) {
 			kubeConfig = filepath.Join(homedir.Get(), ".kube/config")
 		}
 	}
-	config, err := kubernetes.NewKubernetesConfig(kubeConfig)
+
+	clientConfig := kubernetes.NewKubernetesConfig(kubeConfig)
+
+	configNamespace, _, err := clientConfig.Namespace()
+	if err != nil {
+		return nil, err
```
thaJeztah: Looks like this code is executed (possibly showing an error) even if the namespace is overridden through `opts.Namespace`. Retrieving the namespace from the config should probably be done conditional (i.e., if there's no `--namespace` option set)?

cli/command/stack/kubernetes/cli.go
```diff
@@ -57,7 +53,19 @@ func WrapCli(dockerCli command.Cli, opts Options) (*KubeCli, error) {
...
+	}
+	cli.kubeNamespace = configNamespace
+	if opts.Namespace != "default" {
```
thaJeztah: - What if `opts.Namespace` is an empty string? (also; I'm not a big fan of magic words, such as `"default"` for default values) 
- What if I run `docker stack ls --namespace=default`; will the specified (`"default"`) namespace be ignored? (perhaps `kubernetes.NewOptions()` should use `flags.Changed()` to detect if the flag was set (even if it's with the default value https://github.com/docker/cli/blob/854aad8927e9e61c10be944ced7546f0b7142379/cli/command/stack/kubernetes/cli.go#L31-L41)

kubernetes/config.go
```diff
@@ -5,13 +5,12 @@ import (
...
 
 	"github.com/docker/docker/pkg/homedir"
-	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
 // NewKubernetesConfig resolves the path to the desired Kubernetes configuration file, depending
 // environment variable and command line flag.
-func NewKubernetesConfig(configFlag string) (*restclient.Config, error) {
+func NewKubernetesConfig(configFlag string) clientcmd.ClientConfig {
```
thaJeztah: not new, but the `configFlag` variable is a bit confusing (i.e., it's not the flag itself, but a custom path, which _happens_ to be passed through a flag). Perhaps renamed this to (e.g.) `configPath`

kubernetes/config.go
```diff
@@ -5,13 +5,12 @@ import (
...
 	kubeConfig := configFlag
 	if kubeConfig == "" {
```
thaJeztah: The exact same code is also in `cli.WrapCli()` (and looks to be redundant). I suggest extracting this to a function, e.g.

```go
func ConfigPath(kubeconfigPath string) string {
	if kubeconfigPath != "" {
		return kubeconfigPath
	}
	if config := os.Getenv("KUBECONFIG"); config != "" {
		return config
	}
	return filepath.Join(homedir.Get(), ".kube/config")
}
```

Is a non-existing path reason for an error?

cli/command/formatter/stack.go
```diff
@@ -17,16 +21,14 @@ type Stack struct {
...
+	// Orchestrator is the platform where the stack is deployed
 	Orchestrator string
+	// Namespace is the Kubernetes namespace assigned to the stack
+	Namespace string
 }
 
 // NewStackFormat returns a format for use with a stack Context
 func NewStackFormat(source string) Format {
-	switch source {
-	case TableFormatKey:
```
thaJeztah: This is a change in behaviour; before this change:

```bash
docker stack ls --format=table
NAME                SERVICES
foobar              1
```

After this change:

```bash
docker stack ls --format=table

# (no output)
```


Perhaps remove this function, and in swarm/kubernetes do;

```go
if len(format) == 0 || format == formatter.TableFormatKey {
	format = formatter.KubernetesStackTableFormat
}

stackCtx := formatter.Context{
	Output: dockerCli.Out(),
	Format: formatter.Format(format),
}
```

cli/command/stack/kubernetes/list.go
```diff
@@ -17,7 +17,7 @@ func RunList(dockerCli *KubeCli, opts options.List) error {
 	}
 	format := opts.Format
 	if len(format) == 0 {
```
thaJeztah: See my other comment above

cli/command/stack/swarm/list.go
```diff
@@ -25,7 +25,7 @@ func RunList(dockerCli command.Cli, opts options.List) error {
 	}
 	format := opts.Format
 	if len(format) == 0 {
```
thaJeztah: See my other comment above

cli/command/stack/list.go
```diff
@@ -31,5 +31,7 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.StringVar(&opts.Format, "format", "", "Pretty-print stacks using a Go template")
+	flags.BoolVarP(&opts.AllNamespaces, "all-namespaces", "", false, "List stacks among all Kubernetes namespaces")
+	flags.SetAnnotation("all-namespaces", "kubernetes", nil)
```
thaJeztah: Does this have to be gated by API-version as well? (possibly not because it was experimental so far)

cli/command/stack/list.go
```diff
@@ -31,5 +31,7 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.StringVar(&opts.Format, "format", "", "Pretty-print stacks using a Go template")
+	flags.BoolVarP(&opts.AllNamespaces, "all-namespaces", "", false, "List stacks among all Kubernetes namespaces")
```
thaJeztah: Wonder if we considered

- using `--namespace=all` (e.g.) to show all namespaces, also allowing listing stack in a single namespace `--namespace=<something>`
- showing all namespaces by default, but use a `--filter namespace=foo` to limit the output

thaJeztah:
Oh, this also needs a documentation update to document the `.Namespace` placeholder for `--format`

cli/command/stack/list.go
mat007: I believe what drives the current design is to mimic what ```kubectl``` does in order to not confuse Kubernetes users.

kubernetes/config.go
mat007: Yes, changing this.

cli/command/stack/kubernetes/cli.go
mat007: Good point, I'll invert the condition below, but the code in here is likely going to change when I address your next comment about ```--namespace```.

cli/command/stack/kubernetes/cli.go
mat007: Running a complete set of (manual) tests with both ```kubectl get stacks``` and ```docker stack ls``` reveals a few discrepancies. I'll sort that out.

kubernetes/config.go
mat007: Well actually ```WrapCli``` calls ```NewKubernetesConfig``` right after the duplicated code so this looks like the code was moved/factorized and not removed…
Fixed!

mat007:
For the record, here is the manual test suite I ran against kubectl, obviously this is highly dependent on the services deployed beforehand…
```
$ kubectl.exe config set-context docker-for-desktop --namespace=tata && kubectl.exe config view | grep namespace
Context "docker-for-desktop" modified.
    namespace: tata

$ kubectl get stacks --output=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace
NAME      NAMESPACE
dtc       tata

$ kubectl get stacks --output=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace --namespace=default
NAME          NAMESPACE
blaaaaaa      default
blaaaaaaaaa   default
dtcccccc      default

$ kubectl get stacks --output=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace --namespace=""
NAME      NAMESPACE
dtc       tata

$ kubectl get stacks --output=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace --namespace=tata
NAME      NAMESPACE
dtc       tata

$ kubectl get stacks --output=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace --namespace=bla
NAME      NAMESPACE

$ kubectl.exe config set-context docker-for-desktop --namespace= && kubectl.exe config view | grep namespace

$ kubectl get stacks --output=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace
NAME          NAMESPACE
blaaaaaa      default
blaaaaaaaaa   default
dtcccccc      default

$ kubectl get stacks --output=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace --namespace=default
NAME          NAMESPACE
blaaaaaa      default
blaaaaaaaaa   default
dtcccccc      default

$ kubectl get stacks --output=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace --namespace=""
NAME          NAMESPACE
blaaaaaa      default
blaaaaaaaaa   default
dtcccccc      default

$ kubectl get stacks --output=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace --namespace=tata
NAME      NAMESPACE
dtc       tata

$ kubectl get stacks --output=custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace --namespace=bla
NAME      NAMESPACE
```

mat007:
And here is the matching ```docker stack ls``` script output (after I fixed the issues with ```--namespace```):
```
+ kubectl.exe config set-context docker-for-desktop --namespace=tata
Context "docker-for-desktop" modified.
+ ./build/docker.exe stack ls
dtc                 5                   Kubernetes          tata
+ ./build/docker.exe stack ls --namespace=default
blaaaaaa            1                   Kubernetes          default
blaaaaaaaaa         1                   Kubernetes          default
dtcccccc            5                   Kubernetes          default
+ ./build/docker.exe stack ls --namespace=
dtc                 5                   Kubernetes          tata
+ ./build/docker.exe stack ls --namespace=tata
dtc                 5                   Kubernetes          tata
+ ./build/docker.exe stack ls --namespace=bla
+ kubectl.exe config set-context docker-for-desktop --namespace=
Context "docker-for-desktop" modified.
+ ./build/docker.exe stack ls
blaaaaaa            1                   Kubernetes          default
blaaaaaaaaa         1                   Kubernetes          default
dtcccccc            5                   Kubernetes          default
+ ./build/docker.exe stack ls --namespace=default
blaaaaaa            1                   Kubernetes          default
blaaaaaaaaa         1                   Kubernetes          default
dtcccccc            5                   Kubernetes          default
+ ./build/docker.exe stack ls --namespace=
blaaaaaa            1                   Kubernetes          default
blaaaaaaaaa         1                   Kubernetes          default
dtcccccc            5                   Kubernetes          default
+ ./build/docker.exe stack ls --namespace=tata
dtc                 5                   Kubernetes          tata
+ ./build/docker.exe stack ls --namespace=bla
```

cli/command/stack/list.go
mat007: Yes it was experimental and also targeting only Kubernetes and UCP. The behavior toward the Docker daemon is supposed to not have been changed.

mat007:
@thaJeztah I believe I have addressed all your comments. I added small commits so it remains easy for you to check the changes, but afterwards if you would rather have me collapse them or merge them with the bigger commits from @silvin-lubecki I'll do it.

mat007:
Done, thanks for your help!

PR #1011, opened 2018-04-18
Make testing helpers as such…

vdemeester:
That way, those lines won't be reported in the failure.

```
Helper marks the calling function as a test helper function. When printing file and line information, that function will be skipped. 
```

:lion: 


vdemeester:
ping :angel:

PR #1014, opened 2018-04-20
support SSH connection

AkihiroSuda:

**- What I did**

Added support for SSH connection. e.g. `docker -H ssh://me@server`

**- How I did it**

Followed @tonistiigi 's suggestion: https://github.com/docker/cli/pull/889#issuecomment-379077888

> The cli should accept `ssh://me@server` for `DOCKER_HOST` and `-H`. Using that would execute `ssh` with the passed config. 
> The ssh command would call a hidden command on the docker CLI binary on the remote side. For example, `docker dial-stdio`.
> This command will make a connection to the local `DOCKER_HOST` variable (almost always the default local socket) and forward that connection on the commands stdio. 
> Even though this command is supposed to run locally to the `dockerd` binary, we think that it is an invalid configuration for this feature to remove the local `docker` binary so we can rely on it always being present.

**- How to verify it**

`docker -H ssh://me@server run -it --rm busybox`

support SSH connection

penguin

https://commons.wikimedia.org/wiki/File:Manchot_Ad%C3%A9lie_juv%C3%A9nile.jpg

Vendor: https://github.com/moby/moby/pull/36630 (merged)

- - -

Tests (currently tested manually):
- [X] `echo hello | ./docker -H ssh://10.2.54.48 run -i --rm busybox cat; echo world`
- [X] `./docker -H ssh://10.2.54.48 run --rm busybox echo hello; echo world`
- [x] `./docker -H ssh://10.2.54.48 run -i --rm busybox echo hello; echo world`

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -14,6 +14,7 @@ import (
 	"github.com/docker/cli/cli/config"
 	cliconfig "github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/config/configfile"
+	"github.com/docker/cli/cli/connhelper"
 	cliflags "github.com/docker/cli/cli/flags"
 	manifeststore "github.com/docker/cli/cli/manifest/store"
 	registryclient "github.com/docker/cli/cli/registry/client"
@@ -248,9 +249,35 @@ func NewDockerCli(in io.ReadCloser, out, err io.Writer, isTrusted bool) *DockerC
 
 // NewAPIClientFromFlags creates a new APIClient from command line flags
 func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.ConfigFile) (client.APIClient, error) {
-	host, err := getServerHost(opts.Hosts, opts.TLSOptions)
+	unparsedHost, err := getUnparsedServerHost(opts.Hosts)
 	if err != nil {
 		return &client.Client{}, err
+	}
+	var clientOpts []func(*client.Client) error
+	helper, err := connhelper.GetConnectionHelper(unparsedHost)
+	if err != nil {
+		return &client.Client{}, err
+	}
+	if helper == nil {
+		clientOpts = append(clientOpts, withHTTPClient(opts.TLSOptions))
+		host, err := dopts.ParseHost(opts.TLSOptions != nil, unparsedHost)
+		if err != nil {
+			return &client.Client{}, err
+		}
+		clientOpts = append(clientOpts, client.WithHost(host))
+	} else {
+		clientOpts = append(clientOpts, func(c *client.Client) error {
+			httpClient := &http.Client{
+				// No tls
+				// No proxy
+				Transport: &http.Transport{
+					DialContext: helper.Dialer,
+				},
+			}
+			return client.WithHTTPClient(httpClient)(c)
+		})
+		clientOpts = append(clientOpts, client.WithHost(helper.Host))
+		clientOpts = append(clientOpts, client.WithDialContext(helper.Dialer))
 	}
 
 	customHeaders := configFile.HTTPHeaders
@@ -258,21 +285,18 @@ func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.
 		customHeaders = map[string]string{}
 	}
 	customHeaders["User-Agent"] = UserAgent()
+	clientOpts = append(clientOpts, client.WithHTTPHeaders(customHeaders))
 
 	verStr := api.DefaultVersion
 	if tmpStr := os.Getenv("DOCKER_API_VERSION"); tmpStr != "" {
 		verStr = tmpStr
 	}
+	clientOpts = append(clientOpts, client.WithVersion(verStr))
 
-	return client.NewClientWithOpts(
-		withHTTPClient(opts.TLSOptions),
-		client.WithHTTPHeaders(customHeaders),
-		client.WithVersion(verStr),
-		client.WithHost(host),
-	)
+	return client.NewClientWithOpts(clientOpts...)
 }
 
-func getServerHost(hosts []string, tlsOptions *tlsconfig.Options) (string, error) {
+func getUnparsedServerHost(hosts []string) (string, error) {
 	var host string
 	switch len(hosts) {
 	case 0:
@@ -282,8 +306,7 @@ func getServerHost(hosts []string, tlsOptions *tlsconfig.Options) (string, error
 	default:
 		return "", errors.New("Please specify only one -H")
 	}
-
-	return dopts.ParseHost(tlsOptions != nil, host)
+	return host, nil
 }
 
 func withHTTPClient(tlsOpts *tlsconfig.Options) func(*client.Client) error {
--- a/cli/command/system/dial_stdio.go
+++ b/cli/command/system/dial_stdio.go
@@ -1,0 +1,107 @@
+package system
+
+import (
+	"context"
+	"io"
+	"os"
+
+	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/command"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+// newDialStdioCommand creates a new cobra.Command for `docker system dial-stdio`
+func newDialStdioCommand(dockerCli command.Cli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "dial-stdio",
+		Short:  "Proxy the stdio stream to the daemon connection. Should not be invoked manually.",
+		Args:   cli.NoArgs,
+		Hidden: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runDialStdio(dockerCli)
+		},
+	}
+	return cmd
+}
+
+func runDialStdio(dockerCli command.Cli) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	dialer := dockerCli.Client().Dialer()
+	conn, err := dialer(ctx)
+	if err != nil {
+		return errors.Wrap(err, "failed to open the raw stream connection")
+	}
+	connHalfCloser, ok := conn.(halfCloser)
+	if !ok {
+		return errors.New("the raw stream connection does not implement halfCloser")
+	}
+	stdin2conn := make(chan error)
+	conn2stdout := make(chan error)
+	go func() {
+		stdin2conn <- copier(connHalfCloser, &halfReadCloserWrapper{os.Stdin}, "stdin to stream")
+	}()
+	go func() {
+		conn2stdout <- copier(&halfWriteCloserWrapper{os.Stdout}, connHalfCloser, "stream to stdout")
+	}()
+	select {
+	case err = <-stdin2conn:
+		if err != nil {
+			return err
+		}
+		// wait for stdout
+		err = <-conn2stdout
+	case err = <-conn2stdout:
+		// return immediately without waiting for stdin to be closed.
+		// (stdin is never closed when tty)
+	}
+	return err
+}
+
+func copier(to halfWriteCloser, from halfReadCloser, debugDescription string) error {
+	defer func() {
+		if err := from.CloseRead(); err != nil {
+			logrus.Errorf("error while CloseRead (%s): %v", debugDescription, err)
+		}
+		if err := to.CloseWrite(); err != nil {
+			logrus.Errorf("error while CloseWrite (%s): %v", debugDescription, err)
+		}
+	}()
+	if _, err := io.Copy(to, from); err != nil {
+		return errors.Wrapf(err, "error while Copy (%s)", debugDescription)
+	}
+	return nil
+}
+
+type halfReadCloser interface {
+	io.Reader
+	CloseRead() error
+}
+
+type halfWriteCloser interface {
+	io.Writer
+	CloseWrite() error
+}
+
+type halfCloser interface {
+	halfReadCloser
+	halfWriteCloser
+}
+
+type halfReadCloserWrapper struct {
+	io.ReadCloser
+}
+
+func (x *halfReadCloserWrapper) CloseRead() error {
+	return x.Close()
+}
+
+type halfWriteCloserWrapper struct {
+	io.WriteCloser
+}
+
+func (x *halfWriteCloserWrapper) CloseWrite() error {
+	return x.Close()
+}
--- a/cli/connhelper/connhelper.go
+++ b/cli/connhelper/connhelper.go
@@ -1,0 +1,239 @@
+// Package connhelper provides helpers for connecting to a remote daemon host with custom logic.
+package connhelper
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/url"
+	"os"
+	"os/exec"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/docker/cli/cli/connhelper/ssh"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// ConnectionHelper allows to connect to a remote host with custom stream provider binary.
+type ConnectionHelper struct {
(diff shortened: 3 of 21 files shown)


AkihiroSuda:
@cpuguy83 
Thank you for testing.

The issue can be mitigated by increasing this `timeout`, which is the equivalent of `socat -t` and `nc -q`: https://github.com/docker/cli/pull/1014/files#diff-065a55b1c9d14a6817c97ca733652920R17

However, increasing this value introduces extra sleep.

I'll try to find a more robust way

cli/command/cli.go
```diff
@@ -251,31 +252,54 @@ func NewDockerCli(in io.ReadCloser, out, err io.Writer, isTrusted bool) *DockerC
...
+	if helper == nil {
+		clientOpts = append(clientOpts, withHTTPClient(opts.TLSOptions))
+		host, err := dopts.ParseHost(opts.TLSOptions != nil, unparsedHost)
+		if err != nil {
+			return &client.Client{}, err
+		}
+		clientOpts = append(clientOpts, client.WithHost(host))
+	} else {
+		clientOpts = append(clientOpts, func(c *client.Client) error {
+			httpClient := &http.Client{
```
tonistiigi: I think we should just add `WithDialer` instead of overriding client and remove `WithHijackDialer`. Using it would create a new client with a transport pointing to the current dialer and set it to hijack as well.

cli/command/system/dial_stdio.go
```diff
@@ -0,0 +1,125 @@
...
+
+	flags := cmd.Flags()
+	flags.DurationVarP(&opts.timeout, "timeout", "t", 500*time.Millisecond, "After EOF from one of inputs, wait for this duration before exit")
+	return cmd
+}
+
+func runDialStdio(dockerCli command.Cli, opts dialStdioOptions) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	conn, err := dockerCli.Client().DialRaw(ctx)
```
tonistiigi: I think this would be slightly better as a `Dialer() func(ctx) (net.Conn, error)` . Otherwise, it seems like a method that calls to the remote side. We could also solve this by breaking up `NewClientWithOpts` so that we always get access to dialer in cli before creating the client and store the dialer so we can reuse it.

cli/command/system/dial_stdio.go
```diff
@@ -0,0 +1,125 @@
...
+		Short:  "Proxy the stdio stream to the daemon connection. Should not be invoked manually.",
+		Args:   cli.NoArgs,
+		Hidden: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runDialStdio(dockerCli, opts)
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.DurationVarP(&opts.timeout, "timeout", "t", 500*time.Millisecond, "After EOF from one of inputs, wait for this duration before exit")
```
tonistiigi: I'm not sure I understand this timeout. Why isn't the close of connection handled cleanly? This could be just a connection timeout but then it should be much bigger (20 sec) and increasing it shouldn't have any downsides when timeout isn't reached.

cli/command/system/dial_stdio.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	connHalfCloser, ok := conn.(halfCloser)
+	if !ok {
+		return errors.New("the raw stream connection does not implement halfCloser")
+	}
+	stdio := &cliHalfCloser{dockerCli}
+	stdin2conn := make(chan error)
+	conn2stdout := make(chan error)
+
+	go func() {
```
tonistiigi: These might be better handled with a wrapper around conn.  But I guess I don't understand the need for the current timeout atm.

cpuguy83:
Any update here?

AkihiroSuda:
Sorry still stuck at the stream connection issue https://github.com/docker/cli/pull/1014#discussion_r183923206

Will try to look into this again ASAP

cpuguy83:
No worries I just thought this was already merged and realized it wasn't!

AkihiroSuda:
updated, PTAL

tonistiigi:
@AkihiroSuda Thanks! This is starting look great and worked well when testing.

Can you offer a quick explanation of why these halfclosers, `CloseWrite/CloseRead` are needed. It wasn't obvious for me what would break without them. Maybe a case I could try out.

Probably nothing but when running buildkit integration through this and canceling I get `WARN[0031] commandConn.Close: CloseRead: wait: no child processe` in the output.

Everything else sgtm on first pass.

cpuguy83:
Looks good. There's a bunch of vendor changes are these needed?

cpuguy83:
Oh one thing I noticed, when the host is not valid we get a pretty bad EOF message without any details.
Wonder if we can fix that.

AkihiroSuda:
Updated PR with vendor for https://github.com/moby/moby/pull/36630 and UTs for the EOF case.

AkihiroSuda:
fixed lint issue

cpuguy83:
Can we detect a case where the CLI version on the server side is not supported?

```
./docker: error during connect: Post http://docker/v1.38/containers/create: command /usr/bin/ssh has exited with exit status 1, stderr="\nUsage:\tdocker system COMMAND\n\nManage Docker\n\nCommands:\n  df          Show docker disk usage\n  events      Get real time events from the server\n  info        Display system-wide information\n  prune       Remove unused data\n\nRun 'docker system COMMAND --help' for more information on a command.\n".
See './docker run --help'.
```

cli/command/system/dial_stdio.go
```diff
@@ -0,0 +1,107 @@
...
+		// return immediately without waiting for stdin to be closed.
+		// (stdin is never closed when tty)
+	}
+	return err
+}
+
+func copier(to halfWriteCloser, from halfReadCloser) error {
+	defer func() {
+		if err := from.CloseRead(); err != nil {
+			logrus.Error(errors.Wrapf(err, "error while CloseRead (from=%+v)", from))
```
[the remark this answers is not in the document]
AkihiroSuda: logrus seems already used https://github.com/docker/cli/search?q=%22logrus.Errorf%22&unscoped_q=%22logrus.Errorf%22

cli/connhelper/connhelper.go
```diff
@@ -0,0 +1,235 @@
...
+	if err == io.EOF {
+		err = c.onEOF(err)
+	}
+	return n, err
+}
+
+func (c *commandConn) CloseWrite() error {
+	// NOTE: maybe already closed here
+	if err := c.stdin.Close(); err != nil && !ignorableCloseError(err) {
+		logrus.Warnf("commandConn.CloseWrite: %v", err)
```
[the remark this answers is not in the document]
AkihiroSuda: Seems already used in several places https://github.com/docker/cli/search?q=%22logrus.Warnf%22&unscoped_q=%22logrus.Warnf%22

cli/connhelper/connhelper.go
```diff
@@ -0,0 +1,235 @@
...
+	c.stderrMu.Lock()
+	stderr := c.stderr.String()
+	c.stderrMu.Unlock()
+	return errors.Errorf("command %s has exited with %v, stderr=%q", c.cmd.Path, werr, stderr)
+}
+
+func ignorableCloseError(err error) bool {
+	errS := err.Error()
+	ss := []string{
+		os.ErrClosed.Error(),
```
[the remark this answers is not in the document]
AkihiroSuda: `errors.Cause` does not work for `error` that was not crafted with `github.com/pkg/errors`

cli/command/system/dial_stdio.go
```diff
@@ -0,0 +1,107 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	dialer := dockerCli.Client().Dialer()
+	conn, err := dialer(ctx)
+	if err != nil {
+		return errors.Wrap(err, "failed to open the raw stream connection")
+	}
+	connHalfCloser, ok := conn.(halfCloser)
+	if !ok {
+		return errors.New("the raw stream connection does not implement halfCloser")
```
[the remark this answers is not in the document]
AkihiroSuda: Actually this type assertion should never fail and the error won't be printed

cli/connhelper/connhelper.go
```diff
@@ -0,0 +1,235 @@
...
+	c.stdioClosedMu.Unlock()
+	if !stdioClosed {
+		return nil
+	}
+	var err error
+	// NOTE: maybe already killed here
+	if err = c.cmd.Process.Kill(); err == nil {
+		err = c.cmd.Wait()
+	}
+	if strings.Contains(err.Error(), "os: process already finished") {
```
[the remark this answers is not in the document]
AkihiroSuda: updated

cli/command/system/dial_stdio.go
```diff
@@ -0,0 +1,107 @@
...
+	defer func() {
+		if err := from.CloseRead(); err != nil {
+			logrus.Error(errors.Wrapf(err, "error while CloseRead (from=%+v)", from))
+		}
+		if err := to.CloseWrite(); err != nil {
+			logrus.Error(errors.Wrapf(err, "error while CloseWrite (to=%+v)", to))
+		}
+	}()
+	if _, err := io.Copy(to, from); err != nil {
+		return errors.Wrapf(err, "error while Copy (to=%+v, from=%+v)", to, from)
```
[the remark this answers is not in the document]
AkihiroSuda: updated

cli/connhelper/connhelper.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (c *commandConn) onEOF(eof error) error {
+	werr := c.cmd.Wait()
+	if werr == nil {
+		return eof
+	}
+	c.stderrMu.Lock()
+	stderr := c.stderr.String()
+	c.stderrMu.Unlock()
+	return errors.Errorf("command %s has exited with %v, stderr=%q", c.cmd.Path, werr, stderr)
```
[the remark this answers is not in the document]
AkihiroSuda: updated to print full args, so that user can notice whether the error is from SSH or Docker

```
$ ./docker -H ssh://dummy@github.com info
error during connect: Get http://docker/v1.38/info: command [ssh -l dummy github.com -- docker system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr="dummy@github.com: Permission denied (publickey).\r\n"
```

AkihiroSuda:
updated PR

cli/command/system/dial_stdio.go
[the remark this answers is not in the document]
cpuguy83: Maybe we should make the error a little more clear in this case?
"If you see this, this is a bad bug, please report to github.com/docker/cli"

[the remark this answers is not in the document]
AkihiroSuda: How about just panic here and `defer func(){panic(errors.Wrap(recover(), "If you see this, this is a bad bug, please report to github.com/docker/cli"))}` (in another PR)

AkihiroSuda:
@cpuguy83 PTAL?

cpuguy83:
Sorry, meant to link the comment ;)

https://github.com/docker/cli/pull/1014#issuecomment-409741628

AkihiroSuda:
I think the problem is substantially covered by this change, although it does not negotiate versions actually for avoiding potential performance overhead: https://github.com/docker/cli/pull/1014#discussion_r207093340

> updated to print full args, so that user can notice whether the error is from SSH or Docker

> ```console
> $ ./docker -H ssh://dummy@github.com info
> error during connect: Get http://docker/v1.38/info: command [ssh -l dummy github.com -- docker
> system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 
> 18.09 or later is installed on the remote host: stderr="dummy@github.com: Permission denied 
> (publickey).\r\n"
> ```

sandys:
How does one add options to SSH here ?
We enforce hardware tokens for all our SSH (using opensc). We use epass2003 tokens. This is how the command looks.

#Ubuntu - 
ssh  -o PKCS11Provider=/usr/lib/x86_64-linux-gnu/opensc-pkcs11.so  user@gcp.red.com

#OSX - 
ssh -o PKCS11Provider=/Library/OpenSC/lib/opensc-pkcs11.so user@gcp.red.com

AkihiroSuda:
@sandys ~/.ssh/config should work

```
Host foo
        Hostname          gcp.red.com
        PKCS11Provider    /usr/lib/x86_64-linux-gnu/opensc-pkcs11.so
```

sandys:
Thanks for your reply. Will it not accept parameters passed through command
line ? This is very convenient for portability (when using through
different laptops, etc).
Secondly we have more -o options that we pass.

AkihiroSuda:
Currently no, could you open a github issue?

> Secondly we have more -o options that we pass.

Any option that cannot be configured via ~/.ssh/config?

BretFisher:
From ops and sysadmins everywhere, we thank you for this fantastic and unexpected feature. I'm hoping this will seriously cut down the number of times I see people opening dockerd TCP w/o TLS and just opt for SSH endpoints for remote mgmt. 👍 😂

cpuguy83:
@overmike No, none of the client helpers will set this up for you. Nothing is stopping someone from doing this themselves with the client, though. `client.NewClient` accepts an http client, which you can setup on your own.

The CLI is using a helper subcommand (hidden from `--help`) to assist in wiring up std i/o with the remote.

overmike:
@cpuguy83 , thanks for your feedback. It clarifies I need to add connhelper code in our golang client. 

I think you meant "docker system dial-stdio" for the hidden subcommand. And seems like docker-py 3.6.0 is going to support ssh DOCKER_HOST.

bullno1:
How does it handle authentication? Does it support signed SSH certificate?

cpuguy83:
It just shells out to the ssh bin. You can add whatever options for the host to your ssh config. I think there is another pr to support passing ssh opts through an environment var as well.

cli/connhelper/connhelper.go
```diff
@@ -0,0 +1,239 @@
...
+type ConnectionHelper struct {
+	Dialer func(ctx context.Context, network, addr string) (net.Conn, error)
+	Host   string // dummy URL used for HTTP requests. e.g. "http://docker"
+}
+
+// GetConnectionHelper returns Docker-specific connection helper for the given URL.
+// GetConnectionHelper returns nil without error when no helper is registered for the scheme.
+// URL is like "ssh://me@server01".
+func GetConnectionHelper(daemonURL string) (*ConnectionHelper, error) {
+	u, err := url.Parse(daemonURL)
```
zmilonas: You can actually get a unparsedHost here which is not parseable - has no schema.
Like `127.0.0.1:2364`

AkihiroSuda: `GetConnectionHelper returns nil without error when no helper is registered for the scheme.`

zmilonas: I think, this is actually unrelated to what I'm talking about. Just a line below you check for `err` from `url.Parse()` and since there is a possibility the passed `daemonURL` is in `127.0.0.1:2345` format `url.Parse()` will rightfully return an error per its specification - no protocol scheme. I am not sure where should the scheme be added - here or higher in the call stack but this line in particular introduced a regression in the newest stable release of docker engine - 18.09.0

thaJeztah: @zmilonas looks like you mean the issue that was addressed by https://github.com/docker/cli/issues/1443 ?

zmilonas: Well it might have been addressed but is not fixed 😅 I will file a new issue shortly then.

thaJeztah: It's fixed on master, which is what this repository is tracking.

A backport PR is open to include it in the 18.09 codebase

thaJeztah: Yes, that's what the backport is for. In the meantime you could either modify your configuration (use `tcp://<IP|host>:<port>`), or download the static binaries for Docker 18.06 (or "nightly", which has the latest changes from "master") https://download.docker.com/linux/static/

Issue #1016, opened 2018-04-21
ports binding in violation of “secure by default”


richard-delorenzi:
Both ports mode = ingress, and binding to 0.0.0.0, and not giving an error when an IP address is specified but ignored, are in violation of “secure by default”.

There are 3 related issues:

 - Using ingress mode by default, is causing all nodes to listen when only one may be expected.  Is an inconvenience, and violates security by default.

 - Ignoring bind address, and not issuing an error, then binding to 0.0.0.0. Is  Is an inconvenience, and violates security by default.

 - Not supporting bind addresses, is an inconvenience.

The inconveniences, are making it impossible for me to use swarms. The security problems lead me to expose my data, but I now know how to avoid. However I should not need to think about this, and it could affect other people. (security should be by default). 

Usecases
--------------
 - Use cases that have effected me (definitely real)
   - 1 Bind to port of specific node, because port is in use on other nodes. You can use host mode for this, but the default is a surprise.
   - 2 Bind to port of specific node, because other nodes have public interfaces. You can use host mode for this, but **default is in violation of “secure by default”**.
   - 3 Bind locally because, you don't want it visible, to other hosts, **default is in violation of “secure by default”**
   - 4 Bind to 127.0.0.3, because your development machine has a lot of stuff on it, and 127.0.0.1 is in use, by this port. And use `/etc/hosts`, so that each domain-name is sent to a different container. This works with `docker run`, but not with compose.
 - Other usecases
   - Bind to specific interface e.g. 192.168.0.x, because this is an internal network. **default is in violation of “secure by default”**
   - Bind to specific specific node, but not constrain the service to run on this node. This is a similar use case to 1 or 2, but with out use of constraints. Traffic would be routed through the swarm.

So in summary
--------------------

 * **Ignoring IP address and binding to 0.0.0.0, and default ingress mode, are both in violation of “secure by default”.** Errors should be issued if IP-Address specified, and docs updated. Error should be issued if mode not specified (no default), and docs updated.
 * Support can then be added for IP addresses in host mode.
 * Support for IP addresses in ingress mode, that is limited to local addresses 127.x.x.x, could be added. (different local addresses e.g. 127.0.0.2 and 127.0.0.3 should be treated as different (just passed to OS)).

Allow bind to local address is useful for constrained nodes. Allow bind to particular address, would work for constrained node, or routed through the swarm to one of the addresses on one of the nodes (may be ingress mode only). This routing is done already

**Steps to reproduce the issue:**
[...]

PR #1017, opened 2018-04-21
Warn when using host-ip for published ports

thaJeztah:
Swarm Mode services don't support binding to a specific IP-address. Print a warning that the IP-address will be ignored if one is specified.

relates to https://github.com/docker/cli/issues/1016

```
Warn when using host-ip for published ports for services
```

### DIFF
--- a/opts/port.go
+++ b/opts/port.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/docker/docker/api/types/swarm"
 	"github.com/docker/go-connections/nat"
+	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


thaJeztah:
I was working on this a while back, and recalled I had this branch; should check if the validation/warning is printed in the right part of the code 😅 (may need a unit test as well), but thought; let me push this

opts/port.go
```diff
@@ -147,6 +148,9 @@ func ConvertPortToPortConfig(
 	ports := []swarm.PortConfig{}
 
 	for _, binding := range portBindings[port] {
+		if binding.HostIP != "" && binding.HostIP != "0.0.0.0" {
+			logrus.Warnf("ignoring IP-address (%s:%s:%s) service will listen on '0.0.0.0'", binding.HostIP, binding.HostPort, port)
```
cpuguy83: Should this be logrus? I think usually we just print directly to stderr? But also it's weird to write to stdio in this package.

thaJeztah: I think I used it based on other examples I found in the code-base (e.g. https://github.com/docker/cli/blob/cea4d37bcae4ae6f9d216ba1dda252227600c3b1/cli/compose/loader/loader.go#L475), but open to suggestions 👍

cpuguy83:
For backwards compatibility, I think it needs to just be a warning.
The default is not to expose ports, so you are already exposing things. It's also strictly a CLI bug, so it's not like some API client can hit moby and not realize they've made a mistake.

richard-delorenzi:
Having a warning is an improvement. Is any one looking at adding support for specifying addresses?

PR #1019, opened 2018-04-22
import environment variables that are present

ktomk:
fixes https://github.com/docker/for-linux/issues/284

previously docker did import environment variables if they were present
but created them if they were not when it was asked via a --env-file
cli option to import but not create them.

fix is to only import the variable into the environment if it is present.

additionally do not import variable names of zero-length (which are lines
w/ a potential variable definition w/o a variable name).

refs:

- https://github.com/docker/for-linux/issues/284

**- What I did**

1. Check for environment variable presence and only import if it is present.

2. Check the environment variable is not with a zero-length name

**- How I did it**

1. Swap os.Getenv w/ os.LookupEnv for emptyFn to gain presence information and handle it

2. String length check

**- How to verify it**

I have not run any automated test on these apart from the existing build. I'm new to go so I'm not yet comfortable adding tests.

Fix environment file parsing for imports of absent variables and those w/ no name.

### DIFF
--- a/opts/file.go
+++ b/opts/file.go
@@ -21,7 +21,7 @@ func (e ErrBadKey) Error() string {
 	return fmt.Sprintf("poorly formatted environment: %s", e.msg)
 }
 
-func parseKeyValueFile(filename string, emptyFn func(string) string) ([]string, error) {
+func parseKeyValueFile(filename string, emptyFn func(string) (string, bool)) ([]string, error) {
 	fh, err := os.Open(filename)
 	if err != nil {
 		return []string{}, err
@@ -53,17 +53,23 @@ func parseKeyValueFile(filename string, emptyFn func(string) string) ([]string, 
 			if strings.ContainsAny(variable, whiteSpaces) {
 				return []string{}, ErrBadKey{fmt.Sprintf("variable '%s' has white spaces", variable)}
 			}
+			if len(variable) == 0 {
+				return []string{}, ErrBadKey{fmt.Sprintf("no variable name on line '%s'", line)}
+			}
 
 			if len(data) > 1 {
 				// pass the value through, no trimming
 				lines = append(lines, fmt.Sprintf("%s=%s", variable, data[1]))
 			} else {
 				var value string
+				var present bool
 				if emptyFn != nil {
-					value = emptyFn(line)
+					value, present = emptyFn(line)
 				}
-				// if only a pass-through variable is given, clean it up.
-				lines = append(lines, fmt.Sprintf("%s=%s", strings.TrimSpace(line), value))
+				if present {
+					// if only a pass-through variable is given, clean it up.
+					lines = append(lines, fmt.Sprintf("%s=%s", strings.TrimSpace(line), value))
+				}
 			}
 		}
 	}
--- a/opts/envfile.go
+++ b/opts/envfile.go
@@ -18,5 +18,5 @@ import (
 // environment variables, that's why we just strip leading whitespace and
 // nothing more.
 func ParseEnvFile(filename string) ([]string, error) {
-	return parseKeyValueFile(filename, os.Getenv)
+	return parseKeyValueFile(filename, os.LookupEnv)
 }
--- a/opts/envfile_test.go
+++ b/opts/envfile_test.go
@@ -139,3 +139,26 @@ another invalid line`
 		t.Fatalf("Expected [%v], got [%v]", expectedMessage, err.Error())
 	}
 }
+
+// ParseEnvFile with environment variable import definitions
+func TestParseEnvVariableDefinitionsFile(t *testing.T) {
+	content := `# comment=
+UNDEFINED_VAR
+HOME
+`
+	tmpFile := tmpFileWithContent(content, t)
+	defer os.Remove(tmpFile)
+
+	variables, err := ParseEnvFile(tmpFile)
+	if nil != err {
+		t.Fatal("There must not be any error")
+	}
+
+	if "HOME="+os.Getenv("HOME") != variables[0] {
+		t.Fatal("the HOME variable is not properly imported as the first variable (but it is the only one to import)")
+	}
+
+	if 1 != len(variables) {
+		t.Fatal("exactly one variable is imported (as the other one is not set at all)")
+	}
+}
--- a/opts/envfile_test.go
+++ b/opts/envfile_test.go
@@ -162,3 +162,17 @@ HOME
 		t.Fatal("exactly one variable is imported (as the other one is not set at all)")
 	}
 }
+
+// ParseEnvFile with empty variable name
+func TestParseEnvVariableWithNoNameFile(t *testing.T) {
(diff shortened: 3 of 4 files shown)


ktomk:
I need some assistance w/ this PR: I reported the issue in the docker-linux repository and I don't know if that is the right location as I could identfiy the cli repository to create the PR later so was unsure whether or not it is necessary to re-create the issue or how to cross-link it in the description.

thaJeztah:
Thanks! I double checked the expected behavior as I know there were some oddities in this in the past 😅 

From the documentation: [set environment variables (-e, --env, --env-file)](https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file)

> When running the command, the Docker CLI client checks the value the variable has in your local environment and passes it to the container. If no `=` is provided and that variable is not exported in your local environment, the variable won’t be set in the container.

Using the `-e` / `--env` option seems to match that behavior (removed the default `HOME, PATH, and HOSTNAME` vars in the output):

```bash
export FROMENVIRONMENT=blabla

docker run --rm -e FROMENVIRONMENT -e NOSUCHVAR -e NOSUCHWITHEQUAL= busybox env

FOOBAR=blabla
NOSUCHWITHEQUAL=
```

But doing the same through an env-file indeed produces a different result:

```bash
cat <<EOF > myenv
FROMENVIRONMENT
NOSUCHVAR
NOSUCHWITHEQUAL=
EOF
```

```bash
docker run --rm --env-file ./myenv busybox env

FROMENVIRONMENT=
NOSUCHVAR=
NOSUCHWITHEQUAL=
```

Testing with this patch correctly produces

```bash
docker run --rm --env-file ./myenv busybox env

FROMENVIRONMENT=blabla
NOSUCHWITHEQUAL=
```

opts/file.go
```diff
@@ -53,17 +53,23 @@ func parseKeyValueFile(filename string, emptyFn func(string) string) ([]string,
...
 
 			if len(data) > 1 {
 				// pass the value through, no trimming
 				lines = append(lines, fmt.Sprintf("%s=%s", variable, data[1]))
 			} else {
 				var value string
+				var boolean bool
 				if emptyFn != nil {
-					value = emptyFn(line)
+					value, boolean = emptyFn(line)
```
thaJeztah: `boolean` is a bit odd for the variable name; perhaps change to something like `exists` or `isSet`

opts/file.go
```diff
@@ -53,17 +53,23 @@ func parseKeyValueFile(filename string, emptyFn func(string) string) ([]string,
 			if strings.ContainsAny(variable, whiteSpaces) {
 				return []string{}, ErrBadKey{fmt.Sprintf("variable '%s' has white spaces", variable)}
 			}
+			if len(variable) == 0 {
+				return []string{}, ErrBadKey{fmt.Sprintf("no variable name on line '%s'", line)}
```
thaJeztah: Also testing the validation, using this env-file;

```
FROMENVIRONMENT
# this is a comment
NOSUCHVAR
NOSUCHWITHEQUAL=
# empty line:

# whitespace-only line (spaces):
     
# whitespace-only line (tabs):
				
# whitespace-only line (tabs + spaces):
	 	 	
# starts with = (invalid)
=nothing
EOF
```

```bash
$ docker run --rm --env-file ./myenv busybox env
docker: poorly formatted environment: no variable name on line '=nothing'.
See 'docker run --help'.
```

opts/file.go
```diff
@@ -53,17 +53,23 @@ func parseKeyValueFile(filename string, emptyFn func(string) string) ([]string,
...
+			}
 
 			if len(data) > 1 {
 				// pass the value through, no trimming
 				lines = append(lines, fmt.Sprintf("%s=%s", variable, data[1]))
 			} else {
 				var value string
```
silvin-lubecki: nit:
```go
if emptyFn != nil{
    if value, isSet := emptyFn(line); isSet{
        // if only a pass-through variable is given, clean it up.
        lines = append(lines, fmt.Sprintf("%s=%s", strings.TrimSpace(line), value))
    }
}
```

ktomk: It's not suggested within POSIX that a well-formed variable name may contain a space character, however: "Other characters may be permitted by an implementation; applications shall tolerate the presence of such names." - so either refuse or accept but not alter (like trimming spaces). (quote from: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html)

opts/file.go
ktomk: Sure for the rename.

For the tests I need some assistance but I'm eager to add some / extend existing ones, I just need some guidance into.

opts/file.go
ktomk: At least needs tests against `emptyFn()` implementation how it deals with such thought-to-be-trim-able variable names as it is used to acquire the value. The behavior must then be same.

thaJeztah: Yes, we had some validation in the past, but there's no strict rule what's "allowed" as variable name, so we decided to remove validation and leave it to the container's process to decide what's valid and what not.

Having said the above; trimming trailing whitespace in variable names could be an exception (i.e., I don't think it's possible to define a var with leading/trailing whitespace, as the whitespace would be stripped by the shell), but I'm not sure we should; for example, the following would produce an error in your shell as well;

```bash
$ export foo = bar
bash: export: `=': not a valid identifier
```

Leading whitespace in the value leads to "no value being set";

```bash
$ export bar=    bar
$ env | grep bar
bar=
```

This looks to be something where we differ, because the shell requires quotes for that:

```bash
$ export baz="     bla"
$ env | grep baz
baz=     bla
```

silvin-lubecki: My nit was only a small refactor and should be equivalent with your version.

silvin-lubecki: Or maybe were you answering to @thaJeztah previous comment 😅

thaJeztah:
@ktomk for the tests; I think we need an integration test for this; there are some existing tests in https://github.com/docker/cli/blob/master/e2e/image/build_test.go, which could be a good starting point to look at.

Let me know if that's enough information, or if you need more 👍

ktomk:
@thaJeztah: Thanks for the follow up on tests, I'll take a look. From a very first glance, how to invoke (run) a test file then?

thaJeztah:
Just recalled we have some docs on that in https://github.com/docker/cli/blob/master/TESTING.md

opts/file.go
ktomk: Added the test for the fix in question before the fix. signed both commits. I rbased fresh on master, I guess this goes easily into the RC but I don't know of the Docker projects internal workflow regarding versioning, so just FYI. Let me know if you'd like to get TESTING.md cleaned up.

ktomk:
@thaJeztah Please review the tests I added and let me know about any ifs.

ktomk:
thank you for all your reviews, it's not often you get three reviewers, quite something,

PR #1020, opened 2018-04-23
 Handle some TODOs in tests 

vdemeester:
Use more gotestyourself for `env.Patch`, and `icmd.RunCommand`

Also remove some `FIXME` comments on vendor.conf as they don't really make sense.

:lion:

### DIFF
--- a/cli/command/plugin/install.go
+++ b/cli/command/plugin/install.go
@@ -124,9 +124,8 @@ func buildPullConfig(ctx context.Context, dockerCli command.Cli, opts pluginOpti
 		Disabled:              opts.disable,
 		AcceptAllPermissions:  opts.grantPerms,
 		AcceptPermissionsFunc: acceptPrivileges(dockerCli, opts.remote),
-		// TODO: Rename PrivilegeFunc, it has nothing to do with privileges
-		PrivilegeFunc: registryAuthFunc,
-		Args:          opts.args,
+		PrivilegeFunc:         registryAuthFunc,
+		Args:                  opts.args,
 	}
 	return options, nil
 }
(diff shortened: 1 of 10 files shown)


e2e/container/run_test.go
```diff
@@ -17,8 +16,8 @@ const registryPrefix = "registry:5000"
 func TestRunAttachedFromRemoteImageAndRemove(t *testing.T) {
 	image := createRemoteImage(t)
 
-	result := icmd.RunCmd(shell(t,
-		"docker run --rm %s echo this is output", image))
+	result := icmd.RunCommand("docker", "run", "--rm", image,
+		"echo", "this", "is", "output")
```
thaJeztah: Should "this is output" be a single string here?

vdemeester: Both work :stuck_out_tongue_winking_eye:

cli/command/plugin/install.go
```diff
@@ -124,9 +124,8 @@ func buildPullConfig(ctx context.Context, dockerCli command.Cli, opts pluginOpti
 		Disabled:              opts.disable,
 		AcceptAllPermissions:  opts.grantPerms,
 		AcceptPermissionsFunc: acceptPrivileges(dockerCli, opts.remote),
-		// TODO: Rename PrivilegeFunc, it has nothing to do with privileges
```
thaJeztah: This no longer relevant? 😇

vdemeester: It may be, but the comment should be on the API struct (thus in `moby/moby`)

cpuguy83: Why? This is referring to how to get auth credentials?

silvin-lubecki: So did you move this comment to moby/moby? 😄

cpuguy83: 🤦‍♂️

PR #458, opened 2017-08-19
push multiple docker images with `docker push`

glyif:
**- What I did**
I changed `docker push` to be able to push multiple images.

feature request for: https://github.com/docker/cli/issues/267

**- How I did it**
I created a new function that looped runPush with the variable arguments passed into `docker push`. Changed `ExactArgs` to `RequireMinArgs` to support multiple images.

**- How to verify it**
You can run something like:
```
docker push glyif/valid_image:latest glyif/invalid_image:latest glyif/another_valid_image:latest
```

It will push the 2 valid images and return the error that glyif/invlid_image was unable to be pushed.

**- Description for the changelog**
ability to push multiple docker images with `docker push`

**- A picture of a cute animal (not mandatory but encouraged)**
My dogs again :)

Issue #267, opened 2017-06-29
Allow `docker push' to push multiple tags


ebd2:
Allow the following usage:

```docker push org/image-a:some_tag org/image-b:other_tag```

This would make docker push simpler to use in systems that need to build and push multiple images to dockerhub by obviating the need to wrap `docker push` in a loop.

**Comment by glyif** [comment]
@dnephin, super cool tool! imma try it out. for my educational purposes, i'm going to implement this feature. :)

**Comment by iBobik** [comment]
Also it can save resources while building - the same image does not have to be uploaded twice.

**Comment by Vittel** [comment]
I wonder how that ticket is still open. it started in 2014.. was a quest to finally find the last instance here :)
Anyway. Im in for it.

**Comment by elboletaire** [comment]
> I wonder how that ticket is still open. it started in 2014.. was a quest to finally find the last instance here :) Anyway. Im in for it.

+1. It's curious/weird how you can tag multiple tags at once with `docker build -t`, but you can't publish them all at once...

**Comment by Hronom** [comment]
Guys what the progress on this issue? Still need `docker push` with multiple tags.

**Comment by Nowaker** [comment]
This, please!

**Comment by petrmalina** [comment]
+1

**Comment by smuryginim** [comment]
+1

**Comment by XBeg9** [comment]
+1

**Comment by pappnase99** [comment]
How is this still a thing?

**Comment by elboletaire** [comment]
There's currently the PR https://github.com/docker/cli/pull/458#pullrequestreview-57642310 which is in review. AFAIK for now it only adds a loop for the docker push instead of trying to do it in parallel, which is how it works in `docker build`; that's why it isn't still accepted.

**Comment by ggalihpp** [comment]
Oh its open. I though its already implemented :(

**Comment by AntonKrug** [comment]
Ignoring the performance, the options either to push image with all tags or single tag seems bit odd.

**Comment by javatechy** [comment]
You can install [dokr](https://pypi.org/project/dokr/) . and use `dokr dock --push image_name`. This will push your multiple images

**Comment by Hronom** [comment]
Please do this at least, so many time since this started, I was thinking it's already implemented.

**Comment by aramhakobyan** [comment]
+1

**Comment by achillesrasquinha** [comment]
This is more of a design problem. I'd blindly assumed that this would have worked for me

```
$ docker push $(docker images --format "{{.Repository}}" | grep "foobar")
```

[...]

PR #1021, opened 2018-04-24
Allow `docker push` to push multiple tags

ntindall:
**- What I did**

Changed `docker push` to push multiple image tags in parallel. Feature request for #267 , partially implemented and abandoned in #458. More discussion about this in moby/moby#7336, moby/moby#11682, moby/moby#16106, and related, moby/moby#23383. 

There seem to be further improvements that can be made here, especially with regard to caching. These are mostly performance improvements I think though... so IMO it is worth getting this or similar into the mainline, since it seems like it would get good use. We should decide what exact cli interface we want here. 
 
**- How I did it**

Changed docker push to allow more than one argument. 

**- How to verify it**

`docker push name_a:tag_1 name_b:tag_2`

`ability to push multiple docker images with docker push`

red_panda

### DIFF
--- a/cli/command/image/push.go
+++ b/cli/command/image/push.go
@@ -1,6 +1,9 @@
 package image
 
 import (
+	"fmt"
+	"strings"
+
 	"golang.org/x/net/context"
 
 	"github.com/docker/cli/cli"
@@ -12,7 +15,6 @@ import (
 )
 
 type pushOptions struct {
-	remote    string
 	untrusted bool
 }
 
@@ -21,12 +23,11 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
 	var opts pushOptions
 
 	cmd := &cobra.Command{
-		Use:   "push [OPTIONS] NAME[:TAG]",
-		Short: "Push an image or a repository to a registry",
-		Args:  cli.ExactArgs(1),
+		Use:   "push [OPTIONS] NAME[:TAG] [NAME[:TAG]...]",
+		Short: "Push images or a repository to a registry",
+		Args:  cli.RequiresMinArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			opts.remote = args[0]
-			return runPush(dockerCli, opts)
+			return pushImages(dockerCli, args, opts)
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


vdemeester:
@ntindall Thanks for the contribution. I'll paraphrase @stevvooe a bit (from https://github.com/moby/moby/issues/7336#issuecomment-255469607):

- Introducing this syntax will not make push faster. Remember that push is implemented in a daemon, not the cli, so this implementation will result in the same number of HTTP requests to the daemon as it would be in multiple commands.
- If we use this `docker push foo:v1 bar:v2` we lock the push command to that, and we may never be able to have something like `docker push remote image:tag` or `docker push image:tag repo/image:tag`. 

The latter is the fundamental *design* discussion we should have : 
- Will we ever want something like `docker push remote image:tag` (à-la-git) or `docker push image:tag repo/image:tag` (push a local image to another remote without needing to tag it locally) ? If we do, then we can't use this design (`docker push foo:v1 bar:v1`) to push multiple images — and thus we would need to find another way to do that… `docker push -- foo:v1 -- bar:v2` ?
- We really should have that on the API side (in `moby/moby`). The cli shouldn't be the one parallelizing — i.e. it's the daemon pushing the image(s), so we should send it the list of image(s) to push in one query. It would also make any docker API user benefit from this "multiple push" feature.

cpuguy83:
This PR has the same drawbacks as the other one and doesn't address any concerns.
This needs to be handled in the daemon rather than the CLI.

ntindall:
Thanks for the quick response @vdemeester @cpuguy83. So it seems like this is something we need to align on from an API perpspective... 

Regardless, I am getting some mixed signals on the best way forward... based upon what we have said here so far, it sounds like this requires some changes in the daemon as well (and a parallel change to the `cli` to hit the new API), is that right? This seems to conflict with https://github.com/moby/moby/issues/7336#issuecomment-314602615, which closed the issue as it is a "cli topic".

Additionally, it seems that `push` may have some more impactful improvements that should supercede this (albeit, the following quote is from 2015).

> we would appreciate help from the community in rooting out the actual performance problems in push

Do you know where I can find any reference materials on these problems, if they still exist?

Please advise how to move forward... it seems we think this needs to be done in the docker daemon, so maybe that needs to be done before we modify the cli.

herobix:
it is so close! 
Any update on fixing the conflicting files?

gangefors:
I would rather have `docker push` support the `docker push image:tag repo/image:tag` format since when building images on a shared resource you might want to build a new image for say `remote.repo/build-env:latest` but that same image name is used currently by other builds on the same CI server.
It would be awesome if my docker image build script could use a temporary name of my choosing and then be able to push that image to the remote registry as another name.

Pushing multiple images with the CLI is as easy as a for loop, but my example can't easily be done with the commands we currently have at our disposal.

acasademont:
Could this PR be adapted also for the `pull` command? Seems weird that the push could be parallelized but not the pull. Thanks :D

bhack:
Any news on this?

tiborvass:
Ping @dmcgowan

dmcgowan:
Go ahead and close this. This can be addressed in the API, but likely with a feature to support separating target repository from the source. This change would limit what the CLI could support and is only implemented as a loop across the API. The CLI impacting feature to keep any API changes in mind and wait until after the containerd integration

PR #72, opened 2017-05-11
Fix prefix-matching for service ps

thaJeztah:
The docker CLI matches objects either by ID _prefix_
or a full name match, but not partial name matches.

The correct order of resolution is;

- Full ID match (a name should not be able to mask an ID)
- Full name
- ID-prefix

This patch changes the way services are matched.

Also change to use the first matching service, if there's a
full match (by ID or Name) instead of continue looking for
other possible matches.

Error handling changed;

- Do not error early if multiple services were requested
  and one or more services were not found. Print the
  services that were not found after printing those that
  _were_ found instead
- Print an error if ID-prefix matching is ambiguous

This PR carries the CLI changes of https://github.com/moby/moby/pull/32800

PR #1023, opened 2018-04-24
[Kubernetes] stack services filters

simonferquel:
This PR implements stack services filtering, to align the `docker stack services` experience better.
 
This also fix an issue where global services where missing from the list in Kubernetes mode.

To do that, I leveraged Kubernetes LabelSelectors (so everything is done server side), and translated as closely as possible existing supported filters.

One occurrence of behavior consistency, is that k8s label selectors only support perfect matches while in Swarm mode, name filters behave as a StartsWith comparison


thaJeztah:
> One occurrence of behavior consistency, is that k8s label selectors only support perfect matches while in Swarm mode, name filters behave as a StartsWith comparison

@simonferquel could you elaborate a bit? I think name filters work on full matches (but there was a bug at some point that incorrectly did partial matches: https://github.com/docker/cli/pull/72)

```shell
$ docker stack ls

NAME                SERVICES
foobar              1


$ docker stack ps foo
nothing found in stack: foo

$ docker stack ps foobar
ID                  NAME                IMAGE               NODE                    DESIRED STATE       CURRENT STATE        ERROR               PORTS
u542iioaug1g        foobar_web.1        nginx:alpine        linuxkit-025000000001   Running             Running 3 days ago                       


$ docker service ps foobar
no such service: foobar


$ docker service ps foobar_web
ID                  NAME                IMAGE               NODE                    DESIRED STATE       CURRENT STATE        ERROR               PORTS
u542iioaug1g        foobar_web.1        nginx:alpine        linuxkit-025000000001   Running             Running 3 days ago                       
```

simonferquel:
@thaJeztah the problem is not on the stack name, but when you use the `-f name=<something>` with swarm, this name filter is a prefix filter.

thaJeztah:
looks like this needs a rebase as well

cli/command/stack/kubernetes/conversion.go
```diff
@@ -65,13 +68,44 @@ func toSwarmProtocol(protocol apiv1.Protocol) swarm.PortConfigProtocol {
...
+	if len(nodes) == 0 {
+		podsList, err := pods.List(metav1.ListOptions{LabelSelector: labelSelector})
+		if err != nil {
+			return nil, err
+		}
+		result = podsList.Items
+	} else {
+		result = []apiv1.Pod{}
+		for _, n := range nodes {
+			podsList, err := pods.List(metav1.ListOptions{LabelSelector: labelSelector, FieldSelector: "spec.nodeName=" + n})
```
wsong: This might result in tons of `pods.List()` calls. Can we not do like `spec.nodeName in (node1,node2,...)`? If not, we should probably just do one `pods.List()` call and post-filter the results (or just remove the `node` arg altogether).

cli/command/stack/kubernetes/services.go
```diff
@@ -2,43 +2,130 @@ package kubernetes
...
+			result = append(result, fmt.Sprintf("%s=%s", k, v[0]))
+		} else {
+			sort.Strings(v)
+			result = append(result, fmt.Sprintf("%s in (%s)", k, strings.Join(v, ",")))
+		}
+	}
+	return result
+}
+
+func parseLabelFilters(rawFilters []string) ([]string, map[string][]string) {
```
wsong: Is there any code in the Kubernetes CLI we could reuse here?

cli/command/stack/kubernetes/conversion.go
```diff
@@ -122,48 +156,71 @@ const (
...
+		return result[i].ID < result[j].ID
+	})
 	return result, infos, nil
 }
 
+func replicaToService(labls map[string]string, services *apiv1.ServiceList, containers []apiv1.Container) (*swarm.Service, string, error) {
+	serviceName := labls[labels.ForServiceName]
+	serviceHeadless, ok := findService(services, serviceName)
+	if !ok {
+		return nil, "", fmt.Errorf("could not find service '%s'", serviceName)
```
thaJeztah: `findService()` has a very odd signature; this error should be returned by `findService()`, i.e.

```go
serviceHeadless, err := findService(services, serviceName)
if err != nil {
    return nil, "", err
}

File: cli/command/stack/kubernetes/conversion.go [hunk 4]
```diff
@@ -122,48 +156,71 @@ const (
...
+			Replicas: fmt.Sprintf("%d/%d", d.Status.NumberReady, d.Status.DesiredNumberScheduled),
+		}
+	}
+	sort.Slice(result, func(i, j int) bool {
+		return result[i].ID < result[j].ID
+	})
 	return result, infos, nil
 }
 
+func replicaToService(labls map[string]string, services *apiv1.ServiceList, containers []apiv1.Container) (*swarm.Service, string, error) {
```
thaJeztah: ~~Wondering: Wwy pass labels and containers separately? Both are properties of the replica~~

Ahm, right, so this function is used for converting both `Daemonsets` and `Replicasets`, which are very similar, but different types

- Nit: rename this function to (e.g.) `convertToService()` (as it doesn't only convert replicas)
- We use `labls` only for the service name; may be better to just pass the name?
- Why return both the `ID` and the Service here? (i.e. can use `Service.ID` to get the `uid`)

Service signature could look like;

```go
func convertToService(serviceName string, services *apiv1.ServiceList, containers []apiv1.Container) (*swarm.Service, error) {
```

Then in the loop, this could do;

```go
infos[s.ID] = formatter.ServiceListInfo{
	Mode:     "replicated",
	Replicas: fmt.Sprintf("%d/%d", r.Status.AvailableReplicas, r.Status.Replicas),
}
```

cli/command/stack/kubernetes/ps.go
```diff
@@ -31,76 +37,73 @@ func RunPS(dockerCli *KubeCli, options options.PS) error {
...
+	if err = filters.Validate(supportedPSFilters); err != nil {
+		return err
 	}
 
-	pods, err := fetchPods(namespace, podsClient)
+	// Fetch pods
+	if _, err := stacks.Get(stackName); err != nil {
+		return fmt.Errorf("nothing found in stack: %s", stackName)
+	}
+	pods, err := fetchPods(stackName, podsClient, filters)
```
thaJeztah: Move `podsClient := client.Pods()` closer to this, or inline it here;

```go
pods, err := fetchPods(stackName, client.Pods(), filters)
```

cli/command/stack/kubernetes/ps.go
```diff
@@ -31,76 +37,73 @@ func RunPS(dockerCli *KubeCli, options options.PS) error {
...
 	if err != nil {
 		return err
 	}
 
 	if len(pods) == 0 {
-		return fmt.Errorf("nothing found in stack: %s", namespace)
+		return fmt.Errorf("nothing found in stack: %s", stackName)
```
thaJeztah: (just a suggestion); `fetchPods()` looks to be only used here, so this error could be returned by that function, then we can get rid of this block

cli/command/stack/kubernetes/ps.go
```diff
@@ -31,76 +37,73 @@ func RunPS(dockerCli *KubeCli, options options.PS) error {
...
 
-	tasksCtx := formatter.Context{
-		Output: dockerCli.Out(),
-		Format: formatter.NewTaskFormat(format, quiet),
-		Trunc:  trunc,
-	}
-
 	for i, task := range tasks {
-		nodeValue, err := nodeResolver(pods[i].Spec.NodeName)
+		nodeValue, err := resolveNode(options.NoResolve, client.Nodes(), pods[i].Spec.NodeName)
```
thaJeztah: should `client.Nodes()` be moved outside of the loop?

also, order of arguments is a bit odd; perhaps put `name` first

cli/command/stack/kubernetes/ps.go
```diff
@@ -31,76 +37,73 @@ func RunPS(dockerCli *KubeCli, options options.PS) error {
...
-				return "", fmt.Errorf("could not find node '%s'", name)
-			}
-			return string(n.Items[0].UID), nil
+		n, err := nodes.List(metav1.ListOptions{
+			FieldSelector: fields.OneTermEqualSelector(api.ObjectNameField, name).String(),
+		})
+		if err != nil {
+			return "", err
+		}
+		if len(n.Items) != 1 {
```
thaJeztah: Could more than 1 be returned? (i.e., can node names be ambiguous?)

cli/command/stack/kubernetes/services.go
```diff
@@ -2,43 +2,130 @@ package kubernetes
...
+	return presentLabels, valuedLabels
+}
+
+func generateLabelSelector(f filters.Args, stackName string) string {
+	names := f.Get("name")
+	sort.Strings(names)
+
+	for _, n := range names {
+		if strings.HasPrefix(n, stackName+"_") {
+			// also accepts with unprefixed service name (for compat with existing swarm scripts where service names are prefixed by stack names)
```
thaJeztah: Same question regarding backward-combat: if this is only for stacks created with the experimental feature, we may want to consider breaking backward compatibility (and just document)

cli/command/stack/kubernetes/services.go
```diff
@@ -2,43 +2,130 @@ package kubernetes
...
+	servicesList, err := client.Services().List(metav1.ListOptions{LabelSelector: labelSelector})
+	if err != nil {
+		return nil, nil, nil, err
 	}
+	return replicasList, daemonsList, servicesList, nil
+}
+
+// RunServices is the kubernetes implementation of docker stack services
+func RunServices(dockerCli *KubeCli, opts options.Services) error {
+	// Initialize clients
```
thaJeztah: Comment is a bit superfluous

cli/command/stack/kubernetes/services.go
```diff
@@ -2,43 +2,130 @@ package kubernetes
...
-	replicas := client.ReplicaSets()
 
 	if _, err := stacks.Get(opts.Namespace); err != nil {
 		fmt.Fprintf(dockerCli.Err(), "Nothing found in stack: %s\n", opts.Namespace)
 		return nil
 	}
-
-	replicasList, err := replicas.List(metav1.ListOptions{LabelSelector: labels.SelectorForStack(opts.Namespace)})
-	if err != nil {
+	filters := opts.Filter.Value()
```
thaJeztah: Best move this to the top; no need to initialise clients if filters are invalid

cli/command/stack/kubernetes/conversion.go
```diff
@@ -65,13 +68,44 @@ func toSwarmProtocol(protocol apiv1.Protocol) swarm.PortConfigProtocol {
...
 
-func fetchPods(namespace string, pods corev1.PodInterface) ([]apiv1.Pod, error) {
-	labelSelector := labels.SelectorForStack(namespace)
-	podsList, err := pods.List(metav1.ListOptions{LabelSelector: labelSelector})
-	if err != nil {
-		return nil, err
+func fetchPods(stackName string, pods corev1.PodInterface, f filters.Args) ([]apiv1.Pod, error) {
+	services := f.Get("service")
+	nodes := f.Get("node")
+	// for existing script compatibility, support either <servicename> or <stackname>_<servicename> format
```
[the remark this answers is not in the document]
chris-crone: "existing script functionality" here means existing user scripts that target Swarm

cli/command/stack/kubernetes/conversion.go
```diff
@@ -65,13 +68,44 @@ func toSwarmProtocol(protocol apiv1.Protocol) swarm.PortConfigProtocol {
...
+			}
+			result = append(result, podsList.Items...)
+		}
 	}
-	return podsList.Items, nil
+	// name filters is done client side for matching partials
+
+	for i := len(result) - 1; i >= 0; i-- {
+		fullName := fmt.Sprintf("%s_%s", stackName, result[i].Name)
+		if !f.Match("name", fullName) {
```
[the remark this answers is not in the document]
mat007: After further testing (I'll post my test suites with the outputs later), it looks if we want to mimic the ```docker --orchestrator=swarm stack ps``` behavior this needs to be `FuzzyMatch` indeed.

cli/command/stack/kubernetes/services.go
mat007: What do you mean by Kubernetes CLI? kubectl?

cli/command/stack/kubernetes/ps.go
mat007: In the [docs](https://kubernetes.io/docs/concepts/architecture/nodes/#management) there is `Kubernetes will create a node object internally (the representation), and validate the node by health checking based on the metadata.name field (we assume metadata.name can be resolved).`
and the example shows
 ```"metadata": {
    "name": "10.240.79.157",```
so I would assume if the name is not unique the duplicate node attempting to join the cluster will fail.

cli/command/stack/kubernetes/services.go
mat007: My understanding is that Kubernetes, unlike Swarm, does not prefix its service names with the name of the stack, therefore as a convenience to match the likeness of both outputs, the docker CLI decorates the names of the Kubernetes services to add the same style of prefix.

cli/command/stack/kubernetes/conversion.go
mat007: I have simplified the code and done the node filtering after querying for a full ```pods.List()```, this is much clearer in the end. Thanks for suggesting this!

mat007:
@thaJeztah @wsong I believe I have incorporated your suggestions, thank you both for your very helpful reviews. Would you care to take another look? Cheers!

mat007:
Oh by the way, here is the test suite I have been using as regression testing.
I know it is heavily depending on my environment, e2e tests are being worked on and should cover all this shortly.
```
+ ./build/docker.exe --orchestrator=kubernetes stack services dtcccccc
ID                  NAME                  MODE                REPLICAS            IMAGE                 PORTS
63662850-551        dtcccccc_backend      replicated          0/1                 docker-teaches-code   *:8081->8081/tcp
637c5e01-551        dtcccccc_cpp          replicated          0/1                 dtc-cpp               
638dd704-551        dtcccccc_dockerfile   replicated          0/1                 dtc-dockerfile        
63a3c53a-551        dtcccccc_golang       replicated          0/1                 dtc-golang            
63b2242a-551        dtcccccc_python       replicated          0/1                 dtc-python            
+ ./build/docker.exe --orchestrator=kubernetes stack services -f name=dtc dtcccccc
+ ./build/docker.exe --orchestrator=kubernetes stack services -f name=dtcccccc_py dtcccccc
+ ./build/docker.exe --orchestrator=kubernetes stack services -f name=dtcccccc_python dtcccccc
63b2242a-551        dtcccccc_python     replicated          0/1                 dtc-python          
+ ./build/docker.exe --orchestrator=kubernetes stack services -f name=python dtcccccc
63b2242a-551        dtcccccc_python     replicated          0/1                 dtc-python          
+ ./build/docker.exe --orchestrator=kubernetes stack services -f name=thon dtcccccc
+ ./build/docker.exe --orchestrator=kubernetes stack ps '--format=table {{.ID}}\t{{.Name}}\t{{.Node}}' dtcccccc
ID                  NAME                                   NODE
635e19bc-551        dtcccccc_backend-695899db65-tptbg      docker-for-desktop
6379fe84-551        dtcccccc_cpp-5b7bb8d649-sb8z4          docker-for-desktop
6389d833-551        dtcccccc_dockerfile-6f975bf945-bpxkl   docker-for-desktop
63a1c880-551        dtcccccc_golang-6d94cc9cd9-brx9r       docker-for-desktop
63b1ccdc-551        dtcccccc_python-57b458bb85-wqgzp       docker-for-desktop
+ ./build/docker.exe --orchestrator=kubernetes stack ps '--format=table {{.ID}}\t{{.Name}}\t{{.Node}}' -f name=dtc dtcccccc
ID                  NAME                                   NODE
635e19bc-551        dtcccccc_backend-695899db65-tptbg      docker-for-desktop
6379fe84-551        dtcccccc_cpp-5b7bb8d649-sb8z4          docker-for-desktop
6389d833-551        dtcccccc_dockerfile-6f975bf945-bpxkl   docker-for-desktop
63a1c880-551        dtcccccc_golang-6d94cc9cd9-brx9r       docker-for-desktop
63b1ccdc-551        dtcccccc_python-57b458bb85-wqgzp       docker-for-desktop
+ ./build/docker.exe --orchestrator=kubernetes stack ps '--format=table {{.ID}}\t{{.Name}}\t{{.Node}}' -f name=dtcccccc_cpp-5b7bb8d649-sb8z4 dtcccccc
ID                  NAME                            NODE
6379fe84-551        dtcccccc_cpp-5b7bb8d649-sb8z4   docker-for-desktop
+ ./build/docker.exe --orchestrator=kubernetes stack ps '--format=table {{.ID}}\t{{.Name}}\t{{.Node}}' -f name=python dtcccccc
nothing found in stack: dtcccccc
+ ./build/docker.exe --orchestrator=kubernetes stack ps '--format=table {{.ID}}\t{{.Name}}\t{{.Node}}' -f name=thon dtcccccc
nothing found in stack: dtcccccc
+ ./build/docker.exe --orchestrator=kubernetes stack ps '--format=table {{.ID}}\t{{.Name}}\t{{.Node}}' -f name=py -f name=cp dtcccccc
nothing found in stack: dtcccccc
+ ./build/docker.exe --orchestrator=kubernetes stack ps '--format=table {{.ID}}\t{{.Name}}\t{{.Node}}' -f node=bla dtcccccc
nothing found in stack: dtcccccc
+ ./build/docker.exe --orchestrator=kubernetes stack ps '--format=table {{.ID}}\t{{.Name}}\t{{.Node}}' -f node=docker dtcccccc
nothing found in stack: dtcccccc
+ ./build/docker.exe --orchestrator=kubernetes stack ps '--format=table {{.ID}}\t{{.Name}}\t{{.Node}}' -f node=desktop dtcccccc
nothing found in stack: dtcccccc
+ ./build/docker.exe --orchestrator=kubernetes stack ps '--format=table {{.ID}}\t{{.Name}}\t{{.Node}}' -f node=docker-for-desktop dtcccccc
ID                  NAME                                   NODE
635e19bc-551        dtcccccc_backend-695899db65-tptbg      docker-for-desktop
6379fe84-551        dtcccccc_cpp-5b7bb8d649-sb8z4          docker-for-desktop
6389d833-551        dtcccccc_dockerfile-6f975bf945-bpxkl   docker-for-desktop
63a1c880-551        dtcccccc_golang-6d94cc9cd9-brx9r       docker-for-desktop
63b1ccdc-551        dtcccccc_python-57b458bb85-wqgzp       docker-for-desktop
```

cli/command/stack/kubernetes/conversion.go
```diff
@@ -65,13 +68,43 @@ func toSwarmProtocol(protocol apiv1.Protocol) swarm.PortConfigProtocol {
...
+	podsList, err := pods.List(listOpts)
 	if err != nil {
 		return nil, err
 	}
-	return podsList.Items, nil
+	nodes := f.Get("node")
+	for _, pod := range podsList.Items {
+		if filterPod(pod, nodes) &&
+			// name filter is done client side for matching partials
+			f.FuzzyMatch("name", stackNamePrefix+pod.Name) {
```
wsong: So does this mean that we require the name field to be like `--filter name=<stack name prefix> + <pod name>`?

mat007: This is a `FuzzyMatch` meaning with a `stackNamePrefix` of `dtcccccc` and a `pod.Name` of `backend-695899db65-tptbg` all these `--filter` will match:
- `name=dtc`
- `name=dtcccccc`
- `name=dtcccccc-backend`
- `name=dtcccccc-backend-695899db65-tptbg`

However this one won't:
- `name=backend`

This seems to be consistent with the behavior with `--orchestrator=swarm`, however @simonferquel  initially put `Match` here thus allowing the following filters to match:
- `name=backend`
- `name=ccc`

etc…

I'm not sure which solution we want, the specification document was a bit vague and gave no example of expected outputs:
> * Best effort mapping of --filter flag
>   * Best effort as some filters will not map perfectly

thaJeztah:
This needs a rebase

cli/command/stack/kubernetes/conversion.go
wsong: Wait, aren't you just concatenating the `stackNamePrefix` and `pod.Name` with `stackNamePrefix+pod.Name`? You said that `dtcccccc-backend` should match, implying there should be a hyphen between `stackNamePrefix` and `pod.Name`, but that doesn't seem to be the case.

mat007: `stackNamePrefix` actually ends with an hyphen, it's defined a few lines before as
```
stackNamePrefix := stackName + "_"
```

wsong: Not sure if this is intentional, but that's an underscore, not a hyphen.

mat007: Sorry for the confusion, I made an error when writing the examples in https://github.com/docker/cli/pull/1023#discussion_r187767730: the hyphens should have been underscores i.e.
```
with a `stackNamePrefix` of `dtcccccc` and a `pod.Name` of `backend-695899db65-tptbg` all these `--filter` will match:
- `name=dtc`
- `name=dtcccccc`
- `name=dtcccccc_backend`
- `name=dtcccccc_backend-695899db65-tptbg`
```

cli/command/stack/kubernetes/services.go
wsong: Yeah, I'm wondering if there are any public functions in the `kubernetes/kubernetes` repo we could use here.

silvin-lubecki: I think this code transforms label filters (with swarm syntax) to their Kubernetes counterpart. I don't think there is any function in `kubernetes/kubernetes` for that, as the `kubectl` user will write its filters directly in Kubernetes syntax, which are resolved server side IIRC so `kubectl` doesn't have to parse/interpret it.

silvin-lubecki:
@vdemeester can you take another look on this PR? 👋

cli/command/stack/kubernetes/ps.go
```diff
@@ -29,78 +37,72 @@ func RunPS(dockerCli *KubeCli, options options.PS) error {
...
-				FieldSelector: fields.OneTermEqualSelector(api.ObjectNameField, name).String(),
-			})
-			if err != nil {
-				return "", err
-			}
-			if len(n.Items) != 1 {
-				return "", fmt.Errorf("could not find node '%s'", name)
-			}
-			return string(n.Items[0].UID), nil
+		n, err := nodes.List(metav1.ListOptions{
```
wsong: We're going to call this function for every single task, which is going to be super slow with large amounts of tasks. Is there any way we could just do a single `nodes.List` call?

mat007: Ah yes, well spotted… I've changed the code to request the nodes only once and loop on the tasks afterwards.

cli/command/stack/kubernetes/ps.go
thaJeztah: Sorry, missed your comment: so should we skip this validation (as k8s already handles this), or keep it?

cli/command/stack/kubernetes/ps.go
```diff
@@ -29,78 +37,72 @@ func RunPS(dockerCli *KubeCli, options options.PS) error {
...
+			FieldSelector: fields.OneTermEqualSelector(api.ObjectNameField, name).String(),
+		})
+		if err != nil {
+			return "", err
+		}
+		if len(n.Items) != 1 {
+			return "", fmt.Errorf("could not find node '%s'", name)
```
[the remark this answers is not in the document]
mat007: Kubernetes manages everything by name so there is no chance for two nodes to have the same name.
Also the code has changed now, the nodes are fetched once in a slice which is then used to resolve the name with a clearer plain `==` comparison.

mat007:
@thaJeztah sorry I just realized I had pushed the updated PR to my fork instead of Simon's…

cli/command/stack/kubernetes/conversion.go
```diff
@@ -121,56 +154,77 @@ const (
...
 			Mode:     "replicated",
 			Replicas: fmt.Sprintf("%d/%d", r.Status.AvailableReplicas, r.Status.Replicas),
 		}
 	}
+	for i, d := range daemons.Items {
+		s, err := convertToService(d.Labels[labels.ForServiceName], services, d.Spec.Template.Spec.Containers)
+		if err != nil {
+			return nil, nil, err
+		}
+		result[i+len(replicas.Items)] = *s
```
thaJeztah: Was there a reason for not doing;

```go
result = append(result, *s)
```

cli/command/stack/kubernetes/services.go
```diff
@@ -2,43 +2,136 @@ package kubernetes
...
+}
+
+func generateValuedLabelsSelector(valuedLabels map[string][]string) []string {
+	var result []string
+	for k, v := range valuedLabels {
+		if len(v) == 1 {
+			result = append(result, fmt.Sprintf("%s=%s", k, v[0]))
+		} else {
+			sort.Strings(v)
+			result = append(result, fmt.Sprintf("%s in (%s)", k, strings.Join(v, ",")))
```
thaJeztah: is the `in (x)` operator less efficient than `x=y`, or will the selector be smart enough to optimise if there's a single value?

If so, we could simplify to;

```go
for k, v := range valuedLabels {
    sort.Strings(v)
    result = append(result, fmt.Sprintf("%s in (%s)", k, strings.Join(v, ",")))
}
```

cli/command/stack/kubernetes/services.go
```diff
@@ -2,43 +2,136 @@ package kubernetes
...
+		}
+	}
+	return result
+}
+
+func parseLabelFilters(rawFilters []string) ([]string, map[string][]string) {
+	var presentLabels []string
+	valuedLabels := map[string][]string{}
+	for _, rawLabel := range rawFilters {
+		equalIndex := strings.Index(rawLabel, "=")
```
thaJeztah: This can be simplified by using `strings.SplitN()`;


```go
func parseLabelFilters(rawFilters []string) ([]string, map[string][]string) {
	var presentLabels []string
	valuedLabels := map[string][]string{}
	for _, rawLabel := range rawFilters {
		v := strings.SplitN(rawLabel, "=", 2)
		key := v[0]
		if len(v) > 1 {
			valuedLabels[key] = append(valuedLabels[key], v[1])
			continue
		}
		presentLabels = append(presentLabels, key)
	}
	return presentLabels, valuedLabels
}
```

However, filters can appear both as `valuedLabel` and as `presentLabel`, is that expected? see https://play.golang.org/p/EuJk1d7b5Ts

If that's not the intent, we could treat `presentLabels` and `valuedLabels` as identical; a `presentLabel` just being a `valuedLabel` without value, and change `generateValuedLabelsSelector()` to `generateSelector()`, something like;

```go
func generateSelector(labels map[string][]string) []string {
	var result []string
	for k, v := range labels {
		switch len(v) {
		case 0:
			result = append(result, k)
		case 1:
			result = append(result, fmt.Sprintf("%s=%s", k, v[0]))
		default:
			sort.Strings(v)
			result = append(result, fmt.Sprintf("%s in (%s)", k, strings.Join(v, ",")))
		}
	}
	return result
}
```

cli/command/stack/kubernetes/services.go
```diff
@@ -2,43 +2,136 @@ package kubernetes
...
+	daemons := client.DaemonSets()
+	modes := filters.Get("mode")
+
+	var replicasList *appsv1beta2.ReplicaSetList
+	if len(modes) == 0 || filters.ExactMatch("mode", "replicated") {
+		if replicasList, err = replicas.List(metav1.ListOptions{LabelSelector: labelSelector}); err != nil {
+			return nil, nil, nil, err
+		}
+	} else {
+		replicasList = &appsv1beta2.ReplicaSetList{}
```
thaJeztah: Is this needed (or done already by `var replicasList *appsv1beta2.ReplicaSetList` ?)

cli/command/stack/kubernetes/services.go
```diff
@@ -2,43 +2,136 @@ package kubernetes
...
+	}
+
+	var daemonsList *appsv1beta2.DaemonSetList
+	if len(modes) == 0 || filters.ExactMatch("mode", "global") {
+		if daemonsList, err = daemons.List(metav1.ListOptions{LabelSelector: labelSelector}); err != nil {
+			return nil, nil, nil, err
+		}
+	} else {
+		daemonsList = &appsv1beta2.DaemonSetList{}
```
thaJeztah: Is this needed (or done already by `var replicasList *appsv1beta2.DaemonSetList ` ?)

cli/command/stack/kubernetes/services.go
silvin-lubecki: I honestly don't know if it's really less efficient (we are talking about 1 equality vs a list of 1...)

thaJeztah:
Vendor check is somehow failing;

````
 D vendor/k8s.io/apimachinery/pkg/apimachinery/announced/announced.go
 D vendor/k8s.io/apimachinery/pkg/apimachinery/announced/group_factory.go
 D vendor/k8s.io/apimachinery/pkg/apimachinery/doc.go
 D vendor/k8s.io/apimachinery/pkg/apimachinery/registered/registered.go
 D vendor/k8s.io/apimachinery/pkg/apimachinery/types.go
 D vendor/k8s.io/kubernetes/pkg/api/annotation_key_constants.go
 D vendor/k8s.io/kubernetes/pkg/api/doc.go
 D vendor/k8s.io/kubernetes/pkg/api/field_constants.go
 D vendor/k8s.io/kubernetes/pkg/api/json.go
 D vendor/k8s.io/kubernetes/pkg/api/objectreference.go
 D vendor/k8s.io/kubernetes/pkg/api/register.go
 D vendor/k8s.io/kubernetes/pkg/api/resource.go
 D vendor/k8s.io/kubernetes/pkg/api/taint.go
 D vendor/k8s.io/kubernetes/pkg/api/toleration.go
 D vendor/k8s.io/kubernetes/pkg/api/types.go
 D vendor/k8s.io/kubernetes/pkg/api/zz_generated.deepcopy.go
```

mat007:
Not sure what goes wrong with the vendoring, the PR does not change any file in `vendor`…

cli/command/stack/kubernetes/ps.go
```diff
@@ -29,78 +35,71 @@ func RunPS(dockerCli *KubeCli, options options.PS) error {
...
-				FieldSelector: fields.OneTermEqualSelector(api.ObjectNameField, name).String(),
-			})
-			if err != nil {
-				return "", err
-			}
-			if len(n.Items) != 1 {
-				return "", fmt.Errorf("could not find node '%s'", name)
+		for _, node := range nodes.Items {
+			if node.Name == name {
+				return string(node.UID), nil
```
wsong: Is the `node.UID` really what we want? In Kube, there as you've already pointed out, names are the IDs, so there isn't really an analogue to the ID-vs-hostname split of Swarmkit (in Swarmkit, you can have two nodes with the same name, but that's not possible in Kube). As such, I think it'd make more sense to just ignore `noResolve` for Kube instead of using this UID which isn't used anywhere else in the system.

silvin-lubecki: It is more about having the same UX as Swarm than being really useful... Sure we can remove the noResolve flag and simplify the code, but in that case we may break some Swarm scripts which use this flag? From the start we choose to stick to Swarm UX, even if it is not very relevant sometimes. But maybe we are wrong and we should simplify where we can, WDYT?

wsong: Well, we can keep the flag, but I think when we translate Kube nodes to Swarm nodes, the `ID` field and the `name` field should contain the same value (i.e. the name of the node in Kube). There's no reason to expose the `node.UID` field in `docker node ls`; users will never need to know the value of `node.UID` for the purposes of `docker node ls`. It's not important that the `ID` field contains an opaque, alphanumeric string; what's important is that it's unique, and in Kube the name is unique.

The same goes for everywhere else we use `UID` in this PR.

silvin-lubecki: Thank you @wsong, but I think this a different issue then. I opened a new issue #1075 to take care of it in a follow-up so we can move forward on this PR. Does it sounds good to you?

cli/command/stack/kubernetes/services.go
vdemeester: > However, filters can appear both as `valuedLabel` and as `presentLabel`, is that expected? see https://play.golang.org/p/EuJk1d7b5Ts

Good point, on `docker`/`swarm` it's basically a `and`, i.e. `--filter=label=foo --filter=label=foo=bar` is the same as `--filter=label=foo=bar`. I think we should do/keep the same, and thus `presentLabels` and `valuedLabels` shouldn't have anything in common :angel: 

So yeah, we should tread them as identical and update/fix `generateSelector` instead :+1:

cli/command/stack/kubernetes/services_test.go
```diff
@@ -0,0 +1,115 @@
...
+		{
+			name:      "multi value label filter",
+			stackName: "test",
+			filters: filters.NewArgs(
+				filters.KeyValuePair{Key: "label", Value: "label1=test"},
+				filters.KeyValuePair{Key: "label", Value: "label1=test2"},
+			),
+			expectedSelectorParts: []string{
+				"com.docker.stack.namespace=test",
+				"label1 in (test,test2)",
```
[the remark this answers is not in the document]
silvin-lubecki: Ok good point. What do you suggest to fix this? As it seems the **AND** is impossible is Swarm, maybe we shouldn't be able to write it? My point is that the **OR** makes more sens, so maybe we should error out on swarm side and keep this valid for Kubernetes?

mat007:
Here is an update addressing the refactoring suggested (all comments for which I added a 👍 ).
For the record my local tests are as follows, with two services, `backend` has a `mylabel:mat` and `toto` a `mylabel:mot`:
```
+ ./build/docker.exe stack services --filter label=bla dtc
+ ./build/docker.exe stack services --filter label=mat dtc
+ ./build/docker.exe stack services --filter label=mylabel dtc
ID                  NAME                MODE                REPLICAS            IMAGE                 PORTS
0543f8fe-5e8        dtc_backend         replicated          0/1                 docker-teaches-code   *:8081->8081/tcp
5970841e-5e8        dtc_toto            replicated          0/1                 docker-teaches-code   *:8082->8082/tcp
+ ./build/docker.exe stack services --filter label=mylabel=bla dtc
+ ./build/docker.exe stack services --filter label=mylabel=mat dtc
ID                  NAME                MODE                REPLICAS            IMAGE                 PORTS
0543f8fe-5e8        dtc_backend         replicated          0/1                 docker-teaches-code   *:8081->8081/tcp
+ ./build/docker.exe stack services --filter label=mylabel=mat --filter label=mylabel=mot dtc
ID                  NAME                MODE                REPLICAS            IMAGE                 PORTS
0543f8fe-5e8        dtc_backend         replicated          0/1                 docker-teaches-code   *:8081->8081/tcp
5970841e-5e8        dtc_toto            replicated          0/1                 docker-teaches-code   *:8082->8082/tcp
+ ./build/docker.exe stack services --filter label=mylabel --filter label=mylabel=mat dtc
ID                  NAME                MODE                REPLICAS            IMAGE                 PORTS
0543f8fe-5e8        dtc_backend         replicated          0/1                 docker-teaches-code   *:8081->8081/tcp
+ ./build/docker.exe stack services --filter label=mylabel=mat --filter label=mylabel dtc
ID                  NAME                MODE                REPLICAS            IMAGE                 PORTS
0543f8fe-5e8        dtc_backend         replicated          0/1                 docker-teaches-code   *:8081->8081/tcp
```

Thanks!

cli/command/stack/kubernetes/services_test.go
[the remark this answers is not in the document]
thaJeztah: We should have a look (separate from this PR) at improving the filter design, and have syntax for "and", "or", but also (regex) matching, negative matches and so on. (Was discussing that with @vdemeester yesterday, as it keeps coming up)

> My point is that the OR makes more sense

Depends how you look at it; the "filter" was designed to reduce results until you have a match, so the more filters you add, the less results you get.

For _name_, yes, I agree, there 's likely not much use, but for `labels` (e.g.) definitely (show me all objects having both label-x and label-y). On the other hand; in what cases would I be looking at a stack named "foo" or "bar"?

mat007:
@thaJeztah @vdemeester as suggested I added a commit to match Swarm and Kubernetes behavior when combining filters so we can move forward here.

Issue #1010, opened 2018-04-18
Docker manifest fails for registries/users where JWT action "*" is restricted (e.g. gitlab)


jesusofsuburbia:
Hi everyone,

playing around with `docker manifest` on gitlab gives me some issues due to a bug(?) in the docker CLI.

When trying to do a `docker manifest push`, it fails because gitlab allows `*` actions only for site admins.

Now I wonder whether the action `*` for those commands is intentional -
shouldn't it be `push` and/or `pull` (at least that's how I understand [the docs](https://docs.docker.com/registry/spec/auth/scope/#resource-type))

I think (not really sure) I tracked this problem down to the line linked below:
https://github.com/docker/cli/blob/8618838bdb79d84c8395c2316b06af31b3cf6486/cli/registry/client/endpoint.go#L105
The `*` action hardcoded in the last argument of the function call.

If the `*` action is required/intentional, I will argue in the corresponding [gitlab bug](https://gitlab.com/gitlab-org/gitlab-ce/issues/40197) instead (since that means their permission model doesn't work).

Thanks for any input on that issue.

vdemeester:
cc @dmcgowan @clnperez

clnperez:
Thanks @vdemeester for the tag. I'll take a look.

PR #1024, opened 2018-04-24
specify specific permissions

clnperez:
When creating manifest lists, don't use "*" as the permission when
creating the token handler. This causes problems with gitlab's repos.

Fixes https://github.com/docker/cli/issues/1010

Signed-off-by: Christy Norman <christy@linux.vnet.ibm.com>

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/registry/client/endpoint.go
+++ b/cli/registry/client/endpoint.go
@@ -102,7 +102,7 @@ func getHTTPTransport(authConfig authtypes.AuthConfig, endpoint registry.APIEndp
 		modifiers = append(modifiers, auth.NewAuthorizer(challengeManager, passThruTokenHandler))
 	} else {
 		creds := registry.NewStaticCredentialStore(&authConfig)
-		tokenHandler := auth.NewTokenHandler(authTransport, creds, repoName, "*")
+		tokenHandler := auth.NewTokenHandler(authTransport, creds, repoName, "push", "pull")
 		basicHandler := auth.NewBasicHandler(creds)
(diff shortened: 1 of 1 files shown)


clnperez:
I hope this is the super simple solution here. I tried another way to do it, but you need push *and* pull perm for the push, so, it seems like this is easiest/best. 

@vdemeester is there a way to download the build artifacts for @jesusofsuburbia to test? I can't find it where I think it should be.

vdemeester:
@clnperez yeah, it's a bit tricky to get but it's possible to get them from the articfacts in the circleci build (somehow the tab doesn't show when not logged but is accessible) : https://circleci.com/gh/docker/cli/11565#artifacts/containers/0

clnperez:
@jesusofsuburbia can you try that one out? you can just chmod+x and run it instead of the docker installed on your system. If you expand the build section in the link from the previous comment, you can find binaries for all the platforms docker supports.

jesusofsuburbia:
@clnperez will do so today or tomorrow, thanks so far!

thaJeztah:
@jlhawn @dmcgowan PTAL

jesusofsuburbia:
@clnperez it works! I tested it multiple times...

`docker:18.04.0-ce` fails, but overwriting `/usr/local/bin/docker` with the binary you provided **works**

**thank you so much**

clnperez:
thank *you* for reporting :)

PR #1031, opened 2018-04-30
Better stack list experience with Kubernetes and Swarm

mat007:
**- What I did**

Made it easier to work with both swarm and kubernetes stacks combined, e.g.
* added support in ```stack ls``` for the value ```all``` in the ```--orchestrator``` flag which triggers the display of both swarm and kubernetes stacks
* allowed multiple ```--namespace``` for docker stack ls

**- How I did it**

1. The stack lists are now merged after having been queried for each orchestrator and before sorting them and applying the formatting
2. The namespace flag was moved from the stack command to all sub-commands thus allowing for the stack ls sub-command to accept several occurrences and loop over them

**- How to verify it**

```
# List both swarm and kubernetes with namespace
$ docker stack ls --orchestrator=all --namespace=tutu
dtc                 5                   Kubernetes          tutu
wall-e              2                   Swarm

# List stacks for both swarm and kubernetes in empty namespace
$ docker stack ls --orchestrator=all --namespace=pipo
wall-e              2                   Swarm

# List stacks for both swarm and kubernetes (all namespaces implied)
$ docker stack ls --orchestrator=all
abc                 2                   Kubernetes          molo
dtc                 5                   Kubernetes          tutu
wall-e              2                   Swarm

# List stacks for kubernetes in several namespaces
$ ./build/docker.exe stack ls --orchestrator=kubernetes --namespace=tutu,tutu,pipo,molo
abc                 2                   Kubernetes          molo
dtc                 5                   Kubernetes          tutu
```

* List stacks for both Swarm and Kubernetes with --orchestrator=all in docker stack ls
* Allow several occurrences of --namespace for Kubernetes with docker stack ls

image


mat007:
⛑ comes after https://github.com/docker/cli/pull/991, only the 5 last commits are new.

cli/command/stack/cmd.go
```diff
@@ -27,9 +27,6 @@ func NewStackCommand(dockerCli command.Cli) *cobra.Command {
 		newServicesCommand(dockerCli),
 	)
 	flags := cmd.PersistentFlags()
-	flags.String("namespace", "default", "Kubernetes namespace to use")
-	flags.SetAnnotation("namespace", "kubernetes", nil)
-	flags.SetAnnotation("namespace", "experimentalCLI", nil)
```
vdemeester: This should stay

cli/command/stack/deploy.go
```diff
@@ -45,5 +51,6 @@ func newDeployCommand(dockerCli command.Cli) *cobra.Command {
 		`Query the registry to resolve image digest and supported platforms ("`+swarm.ResolveImageAlways+`"|"`+swarm.ResolveImageChanged+`"|"`+swarm.ResolveImageNever+`")`)
 	flags.SetAnnotation("resolve-image", "version", []string{"1.30"})
 	flags.SetAnnotation("resolve-image", "swarm", nil)
+	kubernetes.InstallFlags(flags)
```
vdemeester: Why making these flags per stack subcommand instead of putting the on the `stack` command and make the subcommand inherit ?

cli/command/stack/list.go
```diff
@@ -18,18 +22,92 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
-				return kubernetes.RunList(kli, opts)
-			}
-			return swarm.RunList(dockerCli, opts)
+			return runList(cmd, dockerCli, opts)
 		},
 	}
 
 	flags := cmd.Flags()
 	flags.StringVar(&opts.Format, "format", "", "Pretty-print stacks using a Go template")
+	flags.StringSliceVar(&opts.Namespaces, "namespace", []string{"default"}, "Kubernetes namespaces to use")
```
vdemeester: Will setting a default here *will* override the default namespace defined in the context ?

cli/command/stack/list.go
```diff
@@ -18,18 +22,92 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
+	flags.SetAnnotation("namespace", "kubernetes", nil)
+	flags.SetAnnotation("namespace", "experimentalCLI", nil)
+	flags.BoolVarP(&opts.AllNamespaces, "all-namespaces", "", false, "List stacks among all Kubernetes namespaces")
+	flags.SetAnnotation("all-namespaces", "kubernetes", nil)
```
vdemeester: `all-namespaces` should probably be marked as `experimentalCLI` too

cli/command/stack/list.go
```diff
@@ -18,18 +22,92 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
+	case dockerCli.ClientInfo().HasKubernetes():
+		format = formatter.KubernetesStackTableFormat
+	default:
+		format = formatter.SwarmStackTableFormat
+	}
+	stackCtx := formatter.Context{
+		Output: dockerCli.Out(),
+		Format: formatter.NewStackFormat(format),
+	}
+	sort.Slice(stacks,
```
vdemeester: `sort.Slice(stacks, func(i, j int) bool { …`

cli/command/stack/list.go
```diff
@@ -18,18 +22,92 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
+		func(i, j int) bool {
+			return sortorder.NaturalLess(stacks[i].Name, stacks[j].Name)
+		})
+	return formatter.StackWrite(stackCtx, stacks)
+}
+
+func getStacks(stacks []*formatter.Stack, dockerCli command.Cli, opts options.List, kopts kubernetes.Options) ([]*formatter.Stack, error) {
```
vdemeester: `getStacks` should take `stacks` as argument — instead, the caller should do the `append(…`, this would be far more readable.

cli/command/stack/list_test.go
```diff
@@ -46,9 +47,11 @@ func TestListErrors(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		cmd := newListCommand(test.NewFakeCli(&fakeClient{
+		cli := test.NewFakeCli(&fakeClient{
 			serviceListFunc: tc.serviceListFunc,
-		}))
+		})
+		cli.SetClientInfo(func() command.ClientInfo { return command.ClientInfo{Orchestrator: "swarm"} })
```
vdemeester: Should probably go in the `NewFakeCli` builder

```go
cmd := newListCommand(test.NewFakeCli(&fakeClient{
  // …
}, test.OrchestratorSwarm))
```

and defining in `internal/test` : 

```go
func OrchestratorSwarm(c *FakeCli) {
  cli.SetClientInfo(func() command.ClientInfo { return command.ClientInfo{Orchestrator: "swarm"} })
}

func OrchestratorKubernetes(c *FakeCli) {
  cli.SetClientInfo(func() command.ClientInfo { return command.ClientInfo{Orchestrator: "swarm"} })
}

func OrchestratorAll(c *FakeCli) {
  cli.SetClientInfo(func() command.ClientInfo { return command.ClientInfo{Orchestrator: "swarm"} })
}
```

nit: `SetClientInfo` should be removed (with the `clientInfoFunc`) and make a builder of `ClientInfo` — to have sthg like `test.WithClientInfo(OrchestratorSwarm)` or sthg)

cli/command/stack/list_test.go
```diff
@@ -59,16 +62,18 @@ func TestListErrors(t *testing.T) {
...
+			serviceListFunc: func(options types.ServiceListOptions) ([]swarm.Service, error) {
+				return []swarm.Service{
+					*Service(
+						ServiceLabels(map[string]string{
+							"com.docker.stack.namespace": "service-name-foo",
+						}),
+					)}, nil
+			},
+		})
+	cli.SetClientInfo(func() command.ClientInfo { return command.ClientInfo{Orchestrator: "swarm"} })
```
vdemeester: Same here

cli/command/stack/list_test.go
```diff
@@ -86,6 +91,7 @@ func TestListWithoutFormat(t *testing.T) {
 				)}, nil
 		},
 	})
+	cli.SetClientInfo(func() command.ClientInfo { return command.ClientInfo{Orchestrator: "swarm"} })
```
vdemeester: and here …

cli/command/stack/list_test.go
```diff
@@ -139,6 +145,7 @@ func TestListOrder(t *testing.T) {
 				return uc.swarmServices, nil
 			},
 		})
+		cli.SetClientInfo(func() command.ClientInfo { return command.ClientInfo{Orchestrator: "swarm"} })
```
vdemeester: … and here

cli/command/stack/deploy.go
```diff
@@ -19,14 +21,18 @@ func newDeployCommand(dockerCli command.Cli) *cobra.Command {
 		Args:    cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			opts.Namespace = args[0]
-			if dockerCli.ClientInfo().HasKubernetes() {
+			switch {
+			case dockerCli.ClientInfo().HasAll():
+				return fmt.Errorf("Please specify an orchestrator (swarm|kubernetes)")
```
vdemeester: We may want to share this error somehow

cmd/docker/docker.go
```diff
@@ -54,7 +54,7 @@ func newDockerCommand(dockerCli *command.DockerCli) *cobra.Command {
 	flags = cmd.Flags()
 	flags.BoolVarP(&opts.Version, "version", "v", false, "Print version information and quit")
 	flags.StringVar(&opts.ConfigDir, "config", cliconfig.Dir(), "Location of client config files")
-	opts.Common.InstallFlags(flags)
+	opts.Common.InstallFlags(flags, cmd.PersistentFlags())
```
vdemeester: I don't get this change, this feels really weird to add the flag twice (on `cmd.Flags()` and `cmd.PersistentFlags()`)

cli/command/stack/cmd.go
mat007: This was moved into kubernetes.InstallFlags, so still there

cli/command/stack/deploy.go
mat007: Because stack ls now needs a different behaviour than other stack sub-commands for --namespace and we cannot set it once with type string and another time with []string (e.g. several namespaces).
So the drawback is a bit of code duplication setting the flags for all sub-commands, which was factorized into kubernetes.InstallFlags

cli/command/stack/list.go
mat007: OK

cmd/docker/docker.go
mat007: "orchestrator" is the only flag added to cmd.PersistentFlags() in InstallFlags, all the other flags are still in cmd.Flags()

cli/command/stack/list_test.go
mat007: OK

cli/command/stack/deploy.go
mat007: OK!

cli/command/stack/list.go
mat007: Actually there is code in kubernetes.WrapCli to deal with that and it seems to be working as expected
```
PS C:\dev\src\github.com\docker\cli> .\build\docker.exe stack ls  --orchestrator=kubernetes
stacks.compose.docker.com is forbidden: User "uu" cannot list stacks.compose.docker.com in the namespace "default": access denied
PS C:\dev\src\github.com\docker\cli> kubectl config set-context ucp_34.209.72.126:6443_uu --namespace=tutu
Context "ucp_34.209.72.126:6443_uu" modified.
PS C:\dev\src\github.com\docker\cli> .\build\docker.exe stack ls  --orchestrator=kubernetes
dtc                 5                   Kubernetes          tutu
```

mat007:
@vdemeester I fixed all comments except the ones under discussion about the flags e.g.
* persistence of --orchestrator
* multiple --namespace implementation

Would you kindly please take another look?

cli/command/stack/list.go
```diff
@@ -69,7 +82,33 @@ func runList(cmd *cobra.Command, dockerCli command.Cli, opts options.List) error
...
+		return nil, err
+	}
+	ss, err := kubernetes.GetStacks(kli, opts)
+	if err != nil {
+		return nil, err
+	}
+	return ss, nil
+}
+
+func getNamespaces(cmd *cobra.Command) (map[string]struct{}, error) {
```
thaJeztah: This could just take (`flags *flag.FlagSet`, or even `namespaces []string`, then call it with `getNamespaces(cmd.Flags())`, but see my other comments

cli/command/stack/kubernetes/cli.go
```diff
@@ -23,20 +23,28 @@ type Options struct {
 }
 
 // NewOptions returns an Options initialized with command line flags
-func NewOptions(flags *flag.FlagSet) Options {
+func NewOptions(flags *flag.FlagSet, namespace ...string) Options {
```
thaJeztah: Both `flags` and `namespace` basically come from `cmd.Flags()`; why pass both, if we have all information in `flags`?

cli/command/stack/kubernetes/cli.go
```diff
@@ -23,20 +23,28 @@ type Options struct {
...
 	var opts Options
-	if namespace, err := flags.GetString("namespace"); err == nil {
-		opts.Namespace = namespace
+	if len(namespace) > 0 {
+		opts.Namespace = namespace[0]
+	} else if nm, err := flags.GetString("namespace"); err == nil {
```
thaJeztah: in what situation would `flags.GetString("namespace")` give anything different than `namespace[0]`? If we only expect a single namespace to be passed, it should not take more than one

cmd/docker/docker.go
thaJeztah: Slightly confused as well

mat007: OK, I'll just remove the related commit.

cli/flags/common.go
```diff
@@ -55,7 +55,7 @@ func (commonOpts *CommonOptions) InstallFlags(flags *pflag.FlagSet) {
 	flags.StringVarP(&commonOpts.LogLevel, "log-level", "l", "info", `Set the logging level ("debug"|"info"|"warn"|"error"|"fatal")`)
 	flags.BoolVar(&commonOpts.TLS, "tls", dockerTLS, "Use TLS; implied by --tlsverify")
 	flags.BoolVar(&commonOpts.TLSVerify, FlagTLSVerify, dockerTLSVerify, "Use TLS and verify the remote")
-	flags.StringVar(&commonOpts.Orchestrator, "orchestrator", "", "Which orchestrator to use with the docker cli (swarm|kubernetes) (default swarm) (experimental)")
+	flags.StringVar(&commonOpts.Orchestrator, "orchestrator", "", "Which orchestrator to use with the docker cli (swarm|kubernetes|all) (default swarm) (experimental)")
```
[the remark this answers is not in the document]
mat007: That's a good point. I'll remove the ```(default swarm)```.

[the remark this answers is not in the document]
mat007: Also this was there before but I don't thing ```with the docker cli``` brings anything and the other flag descriptions don't have it. I'll remove it too.

cli/command/stack/list.go
```diff
@@ -28,6 +28,9 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.StringVar(&opts.Format, "format", "", "Pretty-print stacks using a Go template")
+	flags.StringSliceVar(&opts.Namespaces, "namespace", []string{}, "Kubernetes namespaces to use")
```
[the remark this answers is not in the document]
mat007: All ```docker stack``` sub-commands use ```AddNamespaceFlag``` and have a ```namespace``` flag defined as a ```StringVar``` except ```ls``` which has a ```namespace``` flag defined as a ```StringSliceVar```.
This is because for all sub-commands but ```ls``` only one ```--namespace``` can be set whereas ```ls``` can have several, e.g. ```docker stack ls --namespace=tutu,default,bla``` or ```docker stack ls --namespace=tutu --namespace=default --namespace=bla```.

cli/command/stack/kubernetes/cli.go
```diff
@@ -23,20 +23,28 @@ type Options struct {
...
+		opts.Namespace = nm
```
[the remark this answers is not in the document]
mat007: Yes, there first one is of type ```[]string``` and the latter ```string```.

cli/command/stack/list.go
```diff
@@ -18,20 +22,51 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
-			return swarm.RunList(dockerCli, opts)
+			return runList(cmd, dockerCli, opts)
 		},
 	}
 
 	flags := cmd.Flags()
 	flags.StringVar(&opts.Format, "format", "", "Pretty-print stacks using a Go template")
 	flags.BoolVarP(&opts.AllNamespaces, "all-namespaces", "", false, "List stacks among all Kubernetes namespaces")
 	flags.SetAnnotation("all-namespaces", "kubernetes", nil)
+	flags.SetAnnotation("all-namespaces", "experimentalCLI", nil)
```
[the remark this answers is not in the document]
mat007: You're right, too late though as #991 has been merged in the meanwhile…

cli/command/stack/list.go
```diff
@@ -69,7 +82,33 @@ func runList(cmd *cobra.Command, dockerCli command.Cli, opts options.List) error
...
+}
+
+func getNamespaces(cmd *cobra.Command) (map[string]struct{}, error) {
+	nms, err := cmd.Flags().GetStringSlice("namespace")
+	if err != nil {
+		return nil, err
+	}
+	mnms := map[string]struct{}{}
+	for _, nm := range nms {
+		mnms[nm] = struct{}{}
```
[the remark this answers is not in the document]
mat007: Yes this is to remove duplicates, again to match ```kubectl``` behavior. I'll change the return type to a ```[]string```.

cli/command/stack/list.go
```diff
@@ -44,18 +47,28 @@ func runList(cmd *cobra.Command, dockerCli command.Cli, opts options.List) error
...
-		}
-		kli, err := kubernetes.WrapCli(dockerCli, kubernetes.NewOptions(cmd.Flags()))
+		mnms, err := getNamespaces(cmd)
 		if err != nil {
 			return err
 		}
-		ss, err := kubernetes.GetStacks(kli, opts)
-		if err != nil {
-			return err
+		if opts.AllNamespaces || len(mnms) == 0 {
```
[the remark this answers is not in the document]
mat007: Oh yes, nice catch!

cli/command/stack/list.go
mat007: This has actually been changed to receive a ```flags *flag.FlagSet``` in a follow-up PR, but I might as well do it there.

mat007: Hmm sorry I meant ```[]string``` and passing it ```opts.Namespaces```…

cli/command/stack/list.go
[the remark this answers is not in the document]
mat007: …and refactor it to
```
func removeDuplicates(namespaces []string) []string {
	mnms := map[string]struct{}{}
	for _, nm := range namespaces {
		mnms[nm] = struct{}{}
	}
	namespaces = []string{}
	for nm := range mnms {
		namespaces = append(namespaces, nm)
	}
	return namespaces
}
```

cli/command/stack/kubernetes/cli.go
mat007: The way Kubernetes works implies that listing the stacks (or anything else) must be made within the context of a namespace, thus as we want to merge the lists of stacks from different namespaces we create a new ```kubernetes.Options``` for each namespace when iterating over them for ```docker stack ls```.
I'll refactor the code to separate the different behaviors in respect to namespaces.

mat007:
@thaJeztah I believe the code to be much clearer now, thank you for your comments!
I merged the changes into https://github.com/docker/cli/pull/1031/commits/539654c1a89f780f1cc8fdb2b5aa268f0f3ae042 as it was the commit from which they all originated from.
Would you please take another look?

cli/command/stack/list.go
```diff
@@ -18,20 +22,91 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
+		if err != nil {
+			return err
+		}
+		stacks = append(stacks, ss...)
+	}
+	if dockerCli.ClientInfo().HasKubernetes() {
+		kopts := kubernetes.NewOptions(cmd.Flags())
+		if opts.AllNamespaces || len(opts.Namespaces) == 0 {
+			if dockerCli.ClientInfo().HasAll() {
+				opts.AllNamespaces = true
```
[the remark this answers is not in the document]
mat007: This branch is only executed ```if opts.AllNamespaces || len(opts.Namespaces) == 0``` e.g. there is already ```--orchestrator=all``` or no ```--namespace```.
Here is what this gives me:
```
$ ./build/docker.exe --orchestrator=all stack ls
blaaaaaa            1                   Kubernetes          default
blaaaaaaaaa         1                   Kubernetes          default
dtc                 5                   Kubernetes          tutu
dtc                 5                   Kubernetes          tata
dtcccccc            5                   Kubernetes          default

$ ./build/docker.exe --orchestrator=all stack ls --namespace=tata
dtc                 5                   Kubernetes          tata
```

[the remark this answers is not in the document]
wsong: Ah, I see now.

cli/command/stack/list.go
```diff
@@ -18,20 +22,91 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
+}
+
+func removeDuplicates(namespaces []string) []string {
+	mnms := map[string]struct{}{}
+	for _, nm := range namespaces {
+		mnms[nm] = struct{}{}
+	}
+	namespaces = []string{}
+	for nm := range mnms {
+		namespaces = append(namespaces, nm)
```
thaJeztah: no need to loop twice here, could be something like;

```go
func removeDuplicates(namespaces []string) []string {
	found := make(map[string]bool)
	results := namespaces[:0]
	for _, n := range namespaces {
		if !found[n] {
			results = append(results, n)
			found[n] = true
		}
	}
	return results
}
```

cli/command/stack/list.go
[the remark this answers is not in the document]
thaJeztah: Wondering now if this should be a `StringSliceVar` or a `StringVar`; I think `StringVar` can be specified multiple times (`--namespace=bla --namespace=foo`), but does not support comma-separated values; we haven't documented using comma-separated values, and don't think we support it for other  options (possibly `--security-opt` only)

(note that if a command only accepts a single namespace, we could still produce an error if multiple are given)

cli/command/stack/list.go
```diff
@@ -18,20 +22,91 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
+		ss, err := swarm.GetStacks(dockerCli)
+		if err != nil {
+			return err
+		}
+		stacks = append(stacks, ss...)
+	}
+	if dockerCli.ClientInfo().HasKubernetes() {
+		kopts := kubernetes.NewOptions(cmd.Flags())
+		if opts.AllNamespaces || len(opts.Namespaces) == 0 {
+			if dockerCli.ClientInfo().HasAll() {
```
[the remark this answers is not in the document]
mat007: `--orchestrator=kubernetes` yields `HasKubernetes()` `true` and `HasAll()` `false`
`--orchestrator=all` yields both `HasKubernetes()` `true` and `HasAll()` `true`
So if `HasAll()` implies `HasKubernetes()`, the reverse is not necessary true.

cli/command/stack/list.go
mat007: Fair enough, I'll change that.

mat007:
This looks like what we want because
```
docker stack deploy --namespace=foo --namespace=bar -c docker-compose.yml --orchestrator=swarm foo
unknown flag: --orchestrator
```
The orchestrator flag must be right after `docker` (I removed the commit which made it persistent).

```
docker --orchestrator=swarm stack deploy --namespace=foo --namespace=bar -c docker-compose.yml foo
"--namespace" is only supported on a Docker cli with kubernetes features enabled
```
Passing `--orchestrator=swarm` tells `docker` to target that orchestrator for which the `--namespace` makes no sense, hence the error.

Or am I missing something?

cli/command/stack/list.go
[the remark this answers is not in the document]
mat007: I'll take another look but from what I remember there was no way to make it look nice, for both case, e.g.
```
$ docker stack ps  --help
…
Options:
…
      --namespace string    Kubernetes namespace to use
```
VS
```
$ ./build/docker.exe stack ls --help
…
Options:
…
      --namespace strings   Kubernetes namespaces to use
```
Also I don't see how a single `StringVar` can hold multiple `--namespace` values…

cli/command/stack/list.go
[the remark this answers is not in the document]
thaJeztah: But here `HasAll()` (ie `--orchestrator=all`) is used to enable `--all-namespaces`; I don't see how those relate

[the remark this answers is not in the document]
mat007: This is a UX requirement defined in the PRD

> * Provide intuitive mechanisms for switching Kubernetes namespace:
>   * When targeting all orchestrators, target all namespaces by default

cli/command/stack/list.go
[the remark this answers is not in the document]
mat007: > I think StringVar can be specified multiple times (--namespace=bla --namespace=foo), but does not support comma-separated values
In that case the latest value is silently selected (here `foo`), I suppose the flags are simply processed in order.
If we just let comma-separated values out of the question we could indeed emulate both behaviors with a `StringSliceVar` by using the last element when only one namespace is required.

(Incidentally passing `--namespace=bla,foo` to a `StringVar` flag yields a value of `bla,foo`…)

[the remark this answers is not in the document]
mat007: The best I could come up with by using `[]string` on the `stack` command level is
```
$ ./build/docker.exe stack --help

Usage:  docker stack COMMAND

Manage Docker stacks

Options:
      --kubeconfig string   Kubernetes config file
      --namespace strings   Kubernetes namespace to use
```
This is wrong for all sub-commands as for most of them it says `strings` plural whereas only one is supported (and differs for instance from `kubeconfig`), and for `docker stack ls` which supports several namespaces the description should have `namespaces`…

I don't see how to have types and descriptions accurate for all sub-commands without duplicating the flag.

thaJeztah:
> The orchestrator flag must be right after docker (I removed the commit which made it persistent).

hm, right; I'm in doubt if we should only accept it as a global flag. I know users can get very confused as to "where" to pass a flag; existing examples are, e.g. `docker container run -e foo mysql --host=bar` (where `--host=bar` is a mysql option, but `-e` is a docker option); adding global flags makes this even more confusing.

cli/command/stack/list.go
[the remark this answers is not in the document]
thaJeztah: I'm having trouble conflating these; this may become very confusing:

- configuration has `"orchestrator: all"`: `docker stack ls` shows all stacks (both swarm and k8s)
- I want to view only the `k8s` stacks, so I add `--orchestrator=kubernetes`. Now I no longer see all my stacks, only in the `default` namespace
- confused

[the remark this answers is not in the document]
chris-crone: The thinking behind this was as follows: When you list stacks across orchestrators your intention is to see all the stacks running on your cluster. When you specify `kubernetes` as your orchestrator; you are doing something more specific and want similar behaviour to what `kubectl` gives you.

I can see how this might be confusing for your flow. There will be a _NAMESPACE_ column which should help the user understand what has happened.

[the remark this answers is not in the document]
chris-crone: Ah, see you're referring to the case when a user has set `orchestrator` to `all` in their config file. I don't think many users will do this as they will only be able to list resources. In order to manipulate resources, they will need to specify an orchestrator.

mat007:
@thaJeztah I brought the refactoring over from https://github.com/docker/cli/pull/1034/files/e0dbcb3b88f615d358fc5bed93ca3d14aa094d87..691eeff5a5db87be3fe5e0f1ca852024edf08f1c#diff-0ade8f01def39989ced9bb51bab08a69
Tell me if you'd like me to change the `orchestrator` flag, maybe we could move it after `docker stack` for instance? This could be addressed in another PR if needed though as this has not been touched in this one?

mat007:
@vdemeester @thaJeztah I can remove the last commit if that helps moving forward with this PR (and either open a separate PR with it or just forget about it)?

cli/command/stack/kubernetes/list.go
```diff
@@ -1,44 +1,33 @@
...
-func (n byName) Less(i, j int) bool { return sortorder.NaturalLess(n[i].Name, n[j].Name) }
-
-func getStacks(kubeCli *KubeCli, allNamespaces bool) ([]*formatter.Stack, error) {
+func getStacks(kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error) {
 	composeClient, err := kubeCli.composeClient()
 	if err != nil {
 		return nil, err
 	}
-	stackSvc, err := composeClient.Stacks(allNamespaces)
+	stackSvc, err := composeClient.Stacks(opts.AllNamespaces)
```
thaJeztah: Not for this PR, but we should consider passing an `opts` here, instead of a boolean (`opts.AllNamespaces`) to prevent future bloat in the signature

thaJeztah:
I still have reservations on the magic relation between `--orchestrator=all` and `--all-namespaces` (copying my earlier comment https://github.com/docker/cli/pull/1031#discussion_r187891563 here, as it was now hidden);


I'm having trouble conflating these; this may become very confusing:

- configuration has `"orchestrator: all"`: `docker stack ls` shows all stacks (both swarm and k8s)
- I want to view only the `k8s` stacks, so I add `--orchestrator=kubernetes`. Now I no longer see all my stacks, only those in the `default` namespace
- confused

@chris-crone :

The thinking behind this was as follows: When you list stacks across orchestrators your intention is to see all the stacks running on your cluster. When you specify `kubernetes` as your orchestrator; you are doing something more specific and want similar behaviour to what `kubectl` gives you.

I can see how this might be confusing for your flow. There will be a _NAMESPACE_ column which should help the user understand what has happened.

@chris-crone : Ah, see you're referring to the case when a user has set `orchestrator` to `all` in their config file. I don't think many users will do this as they will only be able to list resources. In order to manipulate resources, they will need to specify an orchestrator.

> I don't think many users will do this as they will only be able to list resources.

I'm actually not sure: I can definitely see it being useful to just run any `docker <foo> list` or `docker <foo> ps` command, and see everything I have running, then use `--filter <something>` to limit the list to what I'm looking for, or (when deploying / interacting with objects), pass `--orchestrator=foo --namespace=bar` as part of the steps to deploy.

Which brings me to two things to consider:

- Add orchestrator and namespace as filter options (`--filter orchestrator=swarm`) (no high-priority, but for consistency)
- When interacting with existing objects, only require `--orchestrator` and/or `--namespace` if the object refered to is ambiguous, for example, if there's only one stack with a given name, there would be no need to ask for `--orchestrator`.

cli/command/stack/cmd.go
```diff
@@ -1,11 +1,15 @@
...
 
 import (
+	"fmt"
+
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/spf13/cobra"
 )
 
+var errUnsupportedAllOrchestrator = fmt.Errorf("Please specify an orchestrator (swarm|kubernetes)")
```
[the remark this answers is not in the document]
mat007: As an error message shouldn't it start with a lower case? I would probably avoid the dot too, e.g.
```
no orchestrator specified, use either "kubernetes" or "swarm"
```
or
```
no orchestrator specified: use either "kubernetes" or "swarm"
```
?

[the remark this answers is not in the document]
thaJeztah: I like this one:

```
no orchestrator specified: use either "kubernetes" or "swarm"
```

mat007:
I removed the commit with the tests refactoring and changed the error message when an orchestrator must be explicitly specified.

thaJeztah:
@mat007 could you move the `all orchestrators -> all namespaces` changes to a separate PR to unblock this one?

It's three lines, so easy to review from a code-perspective, so if we get agreement on that one, we should be able to merge that one fast;

```go
if dockerCli.ClientInfo().HasAll() {
	opts.AllNamespaces = true
}
```

mat007:
@thaJeztah done!

chris-crone:
Just to keep track, the all orchestrator all namespace work is in https://github.com/docker/cli/pull/1059.

PR #1034, opened 2018-05-02
Better user experience with Kubernetes namespaces for UCP

mat007:
**- What I did**

Made docker stack ls with the Kubernetes orchestrator list the namespaces a non-administrator user
has access to.

**- How I did it**

As this use case cannot be handled using the native Kubernetes API, I had to use an ad hoc workaround in the form of a call to the UCP API as a backup in the event of the call to the Kubernetes API failing.

Also as there is no simple means to retrieve the http client from the docker client, a new http client is being created…

**- How to verify it**

Create a Kubernetes user 'my-user' and grant it only access to the namespace of some installed services.

Without this PR docker stack ls fails when
```
> docker stack ls  --orchestrator=kubernetes --all-namespaces
stacks.compose.docker.com is forbidden: User "my-user" cannot list stacks.compose.docker.com at the cluster scope: access denied
```
whereas with the PR
```
> docker stack ls  --orchestrator=kubernetes --all-namespaces
dtc                 5                   Kubernetes          tutu

Also without ```--all-namespaces``` it falls back to the default ```--namespace=default``` and correctly fails
> docker stack ls  --orchestrator=kubernetes --all-namespaces
stacks.compose.docker.com is forbidden: User "my-user" cannot list stacks.compose.docker.com in the namespace "default": access denied
```

None as this is merely a tweak for what was added in https://github.com/docker/cli/pull/1031

image


mat007:
Comes after https://github.com/docker/cli/pull/1031, only the last commit is new.

cli/command/stack/list.go
```diff
@@ -18,18 +32,168 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
+		return nil, err
+	}
+	mnms := map[string]struct{}{}
+	for _, nm := range nms {
+		mnms[nm] = struct{}{}
+	}
+	return mnms, nil
+}
+
+func getUserVisibleNamespaces(dockerCli command.Cli, forbiddenErr error) (*core_v1.NamespaceList, error) {
```
vdemeester: This is *really* hacky… I feel that we should extract this somewhere in the `kubernetes` package and make and make the hack there (a func or list of func, …) … This would allow, in the future, to code different methods to get the list of namespace.

This also duplicates a bunch of what the `cli`/`client` does… We may want to Update `https://github.com/moby/moby/blob/master/client/request.go` to make it a bit more composable — like allowing to define new endpoint to query (with a function that would handle the http response).

cli/command/stack/list.go
```diff
@@ -18,18 +32,168 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
+	}
+	endpoint.Scheme = "https"
+	endpoint.Path = "/kubernetesNamespaces"
+	return listUserNamespaces(httpClient, *endpoint)
+}
+
+func listUserNamespaces(httpClient *http.Client, endpoint url.URL) (*core_v1.NamespaceList, error) {
+	resp, err := httpClient.Get(endpoint.String())
+	if err != nil {
+		return nil, fmt.Errorf("unable to list user namespaces: %s", err)
```
vdemeester: Here (and later on), prefer to use `github.com/pkg/errors` with `Wrap` or `Wrapf` : 

```go
return nil, errors.Wrap(err, "unable to list user namespaces")
```

cli/command/stack/list.go
```diff
@@ -18,18 +32,168 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
+	if err == nil || !errors.IsForbidden(err) {
+		return stacks, err
+	}
+	nms, err := getUserVisibleNamespaces(dockerCli, err)
+	if err != nil {
+		return nil, err
+	}
+	opts.AllNamespaces = false
+	for _, nm := range nms.Items {
+		stacks, err = getStacks(stacks, dockerCli, opts, kubernetes.NewOptions(flags, nm.Name))
```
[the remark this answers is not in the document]
mat007: ```getStacksWithNamespaces``` iterates on the namespaces given on the command line with ```--namespace```, here it iterates on the namespaces retrieved by getUserVisibleNamespaces, so a different set of namespaces.
We could factorize the loops but we would need to add code to convert one of the data structures to the other one, and in the end we wouldn't save much…

cli/command/stack/list.go
```diff
@@ -18,18 +32,168 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
+		tlsConfig, err := tlsconfig.Client(*tlsOptions)
+		if err != nil {
+			return nil, err
+		}
+		httpClient.Transport =
+			&http.Transport{
+				TLSClientConfig: tlsConfig,
+			}
+	}
+	endpoint.Scheme = "https"
```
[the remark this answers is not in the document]
mat007: Ah yes, good point, thanks!

cli/command/stack/list.go
```diff
@@ -18,18 +32,168 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
...
-			return swarm.RunList(dockerCli, opts)
+			return runList(cmd, dockerCli, opts)
 		},
 	}
 
 	flags := cmd.Flags()
 	flags.StringVar(&opts.Format, "format", "", "Pretty-print stacks using a Go template")
+	flags.StringSliceVar(&opts.Namespaces, "namespace", []string{"default"}, "Kubernetes namespaces to use")
+	flags.SetAnnotation("namespace", "kubernetes", nil)
+	flags.SetAnnotation("namespace", "experimentalCLI", nil)
```
dnephin: Can't this use `InstallFlags()` (which I would suggest should be renamed to `AddNamespaceFlag()`.

cli/command/stack/kubernetes/cli.go
```diff
@@ -28,26 +27,30 @@ type Options struct {
...
+		opts.Namespace = nm
 	}
 	if kubeConfig, err := flags.GetString("kubeconfig"); err == nil {
 		opts.Config = kubeConfig
 	}
 	return opts
 }
 
+// InstallFlags adds common Kubernetes related flags
+func InstallFlags(flags *flag.FlagSet) {
```
dnephin: Ok, 2 comments :)

`s/InstallFlags/AddNamespaceFlag/`

cli/command/stack/list.go
mat007: It can't because here it's
```
flags.StringSliceVar(&opts.Namespaces, "namespace", []string{"default"}, "Kubernetes namespaces to use")
```
whereas in InstallFlags soon to be AddNamespaceFlag it's
```
flags.String("namespace", "default", "Kubernetes namespace to use")
```
e.g. most ```docker stack``` sub-commands accept only one ```--namespace``` but ```docker stack ls``` can handle a list of them.

cli/command/stack/kubernetes/cli.go
mat007: Done, thanks!

mat007:
@vdemeester I added a commit to move the code into the kubernetes package and started to refactor it for toward using moby client directly, for which I have a prototype in https://github.com/mat007/cli/commit/fee52e12c225a4c9a04d480ef75ec06675b169bf
Does this look like what you had in mind?

cli/command/stack/kubernetes/list.go
```diff
@@ -57,3 +57,96 @@ func getStacks(kubeCli *KubeCli, allNamespaces bool) ([]*formatter.Stack, error)
...
+}
+
+func getUserVisibleNamespaces(dockerCli command.Cli) (*core_v1.NamespaceList, error) {
+	host := dockerCli.Client().DaemonHost()
+	endpoint, err := url.Parse(host)
+	if err != nil {
+		return nil, err
+	}
+	endpoint.Scheme = "https"
+	endpoint.Path = "/kubernetesNamespaces"
```
wsong: UCP does not have a versioned API at present. If we're going to rely on UCP endpoints, we need to be careful about making sure that this stays backwards-compatible.

Is this UX a requirement? Kubernetes does not have any other operations where you can view a list of only the resources that you have access to.

mat007: You're right, this is really fragile and ad hoc, however this is indeed a UX requirement and Kubernetes does not provide this feature…
If you have any suggestion on how to better manage this I'll be happy to update the code.

cli/command/container/create_test.go
```diff
@@ -156,7 +156,7 @@ func TestNewCreateCommandWithContentTrustErrors(t *testing.T) {
 			) (container.ContainerCreateCreatedBody, error) {
 				return container.ContainerCreateCreatedBody{}, fmt.Errorf("shouldn't try to pull image")
 			},
-		}, test.EnableContentTrust)
+		}).EnableContentTrust()
```
vdemeester: Why this change ? I tend to prefer functionnal argument that "chains" of call… If it's just to not have `test.`, we can *dot import*…

mat007: I believe @thaJeztah suggested it, I personally don't have any strong opinion on this.

cli/command/stack/list.go
```diff
@@ -86,27 +75,3 @@ func runList(cmd *cobra.Command, dockerCli command.Cli, opts options.List) error
...
-		return nil, err
-	}
-	ss, err := kubernetes.GetStacks(kli, opts)
-	if err != nil {
-		return nil, err
-	}
-	return ss, nil
-}
-
-func removeDuplicates(namespaces []string) []string {
```
thaJeztah: I just noticed that this commit https://github.com/docker/cli/pull/1034/commits/691eeff5a5db87be3fe5e0f1ca852024edf08f1c is refactoring code that wasn't even merged yet: https://github.com/docker/cli/pull/1031/commits/1f9c9767573c7b8876895d27a9d9b996216e8203 (https://github.com/docker/cli/pull/1031). To reduce code-churn, could this refactor be done in the commit that adds it?

mat007: Yes, good point.

cli/command/container/create_test.go
mat007: By the way this commit actually belongs to the PR before this one e.g. https://github.com/docker/cli/pull/1031

mat007:
@vdemeester I tweaked the custom request a bit and opened a PR upstream, can you please take another look?
I left out the functions as arguments as I don't see how to do that without adding a lot of code and/or types from the `http` package to `interface.go`…

cli/command/stack/kubernetes/list.go
wsong: All right, let's stick with this approach for now, but we'll have to be careful to test this with different combinations of CLI version + UCP version.

cli/command/stack/kubernetes/list.go
```diff
@@ -44,6 +54,55 @@ func getStacks(kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error)
...
+	resp, err := dockerCli.Client().HTTPClient().Get(endpoint.String())
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, errors.Wrapf(err, "received %d status and unable to read response", resp.StatusCode)
+	}
+	if resp.StatusCode != http.StatusOK {
```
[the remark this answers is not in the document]
wsong: Yes, but you just log a debug message there and return `err`.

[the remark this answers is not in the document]
mat007: It logs `err2`, the error returned by `getUserVisibleNamespaces`, and then returns `err` from the previous call, which is different and in practice due to the `apierrs.IsForbidden(err)` test right before is effectively as if `getUserVisibleNamespaces` had not been called at all:
```
$ docker --orchestrator=all stack ls
stacks.compose.docker.com is forbidden: User "user" cannot list stacks.compose.docker.com at the cluster scope: access denied
```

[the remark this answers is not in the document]
wsong: I see. I think a comment here would be helpful.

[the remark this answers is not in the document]
mat007: OK, I added
```
		// ignore this last error and return the previous one instead
```

cli/command/stack/kubernetes/list.go
```diff
@@ -44,6 +54,56 @@ func getStacks(kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error)
...
+	return stacks, nil
+}
+
+func getUserVisibleNamespaces(dockerCli command.Cli) (*core_v1.NamespaceList, error) {
+	host := dockerCli.Client().DaemonHost()
+	endpoint, err := url.Parse(host)
+	if err != nil {
+		return nil, err
+	}
+	endpoint.Scheme = "https"
```
vdemeester: This part is a bit sad as this is something we already do in the `moby/moby` `client` package (setting the scheme to `https` when tls is configured)..

cli/command/stack/kubernetes/list.go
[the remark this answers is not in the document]
thaJeztah: So, thinking:

- A 404 error is not an error, just means: we don't have UCP
- We're now hiding potential other errors (500 - UCP crashed, 403 - UCP is up, but credentials are incorrect (?), ??? - Other error)

Instead of returning expected errors, and ignoring them in `getStacksWithAllNamespaces()`, we should leave the logic/decision if it's an actual error or not to `getUserVisibleNamespaces()`. When running into a 404, just return an empty list.

```go
func getStacksWithAllNamespaces(dockerCli command.Cli, kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error) {
	stacks, err := getStacks(kubeCli, opts)
	if err == nil || !apierrs.IsForbidden(err) {
		return stacks, err
	}
	nms, err2 := getUserVisibleNamespaces(dockerCli)
	if err2 != nil {
		return nil, errors.Wrap(err2, "Failed to query user visible namespaces")
	}
	if err != nil {
		return nil, errors.Wrap(err, "Failed to retrieve stacks")
	}
	opts.AllNamespaces = false
	for _, namespace := range nms.Items {
		kubeCli.kubeNamespace = namespace.Name
		ss, err := getStacks(kubeCli, opts)
		if err != nil {
			return nil, err
		}
		stacks = append(stacks, ss...)
	}
	return stacks, nil
}

func getUserVisibleNamespaces(dockerCli command.Cli) (*core_v1.NamespaceList, error) {
	host := dockerCli.Client().DaemonHost()
	endpoint, err := url.Parse(host)
	if err != nil {
		return nil, err
	}
	endpoint.Scheme = "https"
	endpoint.Path = "/kubernetesNamespaces"
	resp, err := dockerCli.Client().HTTPClient().Get(endpoint.String())
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()
	body, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		return nil, errors.Wrapf(err, "received %d status and unable to read response", resp.StatusCode)
	}
	namespaces := &core_v1.NamespaceList{}
	switch resp.StatusCode {
	case http.StatusOK:
		if err := json.Unmarshal(body, namespaces); err != nil {
			return nil, errors.Wrapf(err, "unmarshal failed: %s", string(body))
		}
	case http.StatusNotFound:
		// UCP API not present
	default:
		return nil, fmt.Errorf("received %d status while retrieving namespaces: %s", resp.StatusCode, string(body))
	}
	return namespaces, nil
}
```


```patch
diff --git a/cli/command/stack/kubernetes/list.go b/cli/command/stack/kubernetes/list.go
index 8a7b1406..cb467f9b 100644
--- a/cli/command/stack/kubernetes/list.go
+++ b/cli/command/stack/kubernetes/list.go
@@ -11,7 +11,6 @@ import (
        "github.com/docker/cli/cli/command/formatter"
        "github.com/docker/cli/cli/command/stack/options"
        "github.com/pkg/errors"
-       "github.com/sirupsen/logrus"
        core_v1 "k8s.io/api/core/v1"
        apierrs "k8s.io/apimachinery/pkg/api/errors"
@@ -61,9 +60,10 @@ func getStacksWithAllNamespaces(dockerCli command.Cli, kubeCli *KubeCli, opts op
        }
        nms, err2 := getUserVisibleNamespaces(dockerCli)
        if err2 != nil {
-               logrus.Debugf("Failed to query user visible namespaces: %s", err2)
-               // ignore this last error and return the previous one instead
-               return nil, err
+               return nil, errors.Wrap(err2, "Failed to query user visible namespaces")
+       }
+       if err != nil {
+               return nil, errors.Wrap(err, "Failed to retrieve stacks")
        }
        opts.AllNamespaces = false
        for _, namespace := range nms.Items {
@@ -94,12 +94,16 @@ func getUserVisibleNamespaces(dockerCli command.Cli) (*core_v1.NamespaceList, er
        if err != nil {
                return nil, errors.Wrapf(err, "received %d status and unable to read response", resp.StatusCode)
        }
-       if resp.StatusCode != http.StatusOK {
-               return nil, fmt.Errorf(string(body))
-       }
        namespaces := &core_v1.NamespaceList{}
-       if err := json.Unmarshal(body, namespaces); err != nil {
-               return nil, errors.Wrapf(err, "unmarshal failed: %s", string(body))
+       switch resp.StatusCode {
+       case http.StatusOK:
+               if err := json.Unmarshal(body, namespaces); err != nil {
+                       return nil, errors.Wrapf(err, "unmarshal failed: %s", string(body))
+               }
+       case http.StatusNotFound:
+               // UCP API not present
+       default:
+               return nil, fmt.Errorf("received %d status while retrieving namespaces: %s", resp.StatusCode, string(body))
        }
        return namespaces, nil
 }
```

cli/command/stack/kubernetes/list.go
```diff
@@ -44,6 +54,56 @@ func getStacks(kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error)
...
+	endpoint.Path = "/kubernetesNamespaces"
+	resp, err := dockerCli.Client().HTTPClient().Get(endpoint.String())
```
thaJeztah: Wondering if there are other things that should be set on the request (content-type JSON, custom headers that are configured in the client-configuration, etc); https://github.com/docker/cli/blob/60930d309c1544cac67891030aac811f1aff82bf/vendor/github.com/docker/docker/client/request.go#L225-L241

mat007: I don't know, however this would not be an issue were the `Client` performing the request itself.
https://github.com/moby/moby/pull/37071 has been merged in moby, but maybe it's not to late to bring that up?

thaJeztah: Well, it's a bit of an odd situation; the `/kubernetesNamespaces` endpoint feels like an add-hoc/temporary solution. Some thoughts on that;

- It's undocumented, and not sure how well-defined the endpoint is
- the Engine remote API may perform validations, e.g.
  - correct request-type, respond different based on "accept" headers (plain-text/JSON)
  - version negotiation (not present in the UCP API) (this could be a problem if we used the `docker/client/request` functionality, as it may be done automatically - need to check)

So, should we export the `docker/client/request` `client.get()`, `client.head()` and so on functions? I don't know.
 
On the "plus" side; when using the `docker/client/request` functionality, we would get some things for free (improved/consistent errors if TLS fails, automatic handling of invalid JSON responses, etc.)

I think in an _ideal_ situation, given that UCP defines its own API (which is the Engine remote-API + Additional endpoints/extensions) there would be an official UCP client that we could use.

vdemeester: I do agree with @thaJeztah that it is really an odd situation : we end up hacking the `docker/docker` client around for something that isn't even docker api related.

As discussed a bit offline earlier, a cleaner way (*way cleaner*) would have been to have a kubernetes api extension (same as the current "compose on kubernetes" one) : 
- we wouldn't need to hack `docker/docker` client around, we would just consume the go code from the extension (again, same way as "compose on kubernetes")
- it would be "automatically" versionned (because of k8s way of doing things)
- it wouldn't be a direct dependency on UCP as it could be easily deployed on other cluster in the future if we wanted

mat007:
Obviously `ci/circleci: validate` fails, is there something else I'm supposed to do when updating a dependency?

cli/command/stack/kubernetes/list.go
```diff
@@ -43,6 +52,58 @@ func getStacks(kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error)
...
+		ss, err := getStacks(kubeCli, opts)
+		if err != nil {
+			return nil, err
+		}
+		stacks = append(stacks, ss...)
+	}
+	return stacks, nil
+}
+
+func getUserVisibleNamespaces(dockerCli command.Cli, previousErr error) (*core_v1.NamespaceList, error) {
```
wsong: @thaJeztah @mat007 I think it's a little weird to pass in `previousErr` to this function so that we can return it in certain cases.

@thaJeztah, what did you have in mind here?

mat007: I suppose I could either return also a boolean meaning «404», just ignore the call? Or create a custom error to be returned just for the sake a filtering right after the call?
The `previousErr` remains the solution with the less extra code, which is why I went for it.

wsong: I think your previous solution (return an error and then have `getStacksWithAllNamespaces` check for that err and return the previous error) was actually better, which is why I'm wondering if @thaJeztah had something different in mind.

thaJeztah: My suggestion (here https://github.com/docker/cli/pull/1034#discussion_r188932239) was because the previous solution was discarding all errors returned by UCP, which was definitely not the right thing to do.

I see I made one mistake in that example, which probably lead to @mat007 rewriting it (updated version below);

<details>

<summary>updated example:</summary>

```go
func getStacksWithAllNamespaces(dockerCli command.Cli, kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error) {
	stacks, err := getStacks(kubeCli, opts)
	if err == nil || !apierrs.IsForbidden(err) {
		return stacks, err
	}
	nms, err2 := getUserVisibleNamespaces(dockerCli)
	if err2 != nil {
		return nil, errors.Wrap(err2, "Failed to query user visible namespaces")
	}
	if nms == nil {
		return nil, errors.Wrap(err, "Failed to retrieve stacks")
	}
	opts.AllNamespaces = false
	for _, namespace := range nms.Items {
		kubeCli.kubeNamespace = namespace.Name
		ss, err := getStacks(kubeCli, opts)
		if err != nil {
			return nil, err
		}
		stacks = append(stacks, ss...)
	}
	return stacks, nil
}

func getUserVisibleNamespaces(dockerCli command.Cli) (*core_v1.NamespaceList, error) {
	host := dockerCli.Client().DaemonHost()
	endpoint, err := url.Parse(host)
	if err != nil {
		return nil, err
	}
	endpoint.Scheme = "https"
	endpoint.Path = "/kubernetesNamespaces"
	resp, err := dockerCli.Client().HTTPClient().Get(endpoint.String())
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()
	body, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		return nil, errors.Wrapf(err, "received %d status and unable to read response", resp.StatusCode)
	}
	switch resp.StatusCode {
	case http.StatusOK:
	  namespaces := &core_v1.NamespaceList{}
		if err := json.Unmarshal(body, namespaces); err != nil {
			return nil, errors.Wrapf(err, "unmarshal failed: %s", string(body))
		}
	  return namespaces, nil
	case http.StatusNotFound:
		// UCP API not present
		return nil, nil
	default:
		return nil, fmt.Errorf("received %d status while retrieving namespaces: %s", resp.StatusCode, string(body))
	}
}
```

</details>


<p><br /></p>
An alternative could be to flip the order (dummy code);

```go
if stacks, err := getUCPStacks(dockerCli, opts); err != nil || stacks != nil {
    return stacks, err
}
return getStacks(kubeCli, opts)
```

In the above;

- In case of a 404, `getUCPStacks()` returns no error, and `nil` for stacks, in that case, fall back to plain kubernetes
- Any other error is an issue with the UCP endpoint, so should be handled. In that case, don't fallback to plain kubernetes


- Advantage: There's no back and forth between trying UCP and k8s; currently, we try k8s, get an error, try UCP, discover it's not there, and head back to the original k8s error
- Disadvantage: It's odd for Docker CE clients to try an API that won't be there for 99% of the users

mat007: I believe we cannot swap the function calls as what we want is first Kubernetes, then UCP only if Kubernetes fails for lack of user rights.

@wsong we have 3 possible results when querying UCP : not there (404), there but error'ed and valid result, so we need to distinguish between the three
@thaJeztah ah yes, I suppose we could test for `nil` to distinguish a 404 from `[]string{}` an empty result, OK I'll do that instead

Thank you both!

cli/command/stack/kubernetes/list.go
```diff
@@ -43,6 +52,58 @@ func getStacks(kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error)
 	}
 	return formattedStacks, nil
 }
+func getStacksWithAllNamespaces(dockerCli command.Cli, kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error) {
+	stacks, err := getStacks(kubeCli, opts)
+	if err == nil || !apierrs.IsForbidden(err) {
```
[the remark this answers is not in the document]
mat007: You're right, I'll change it while I'm at it.

thaJeztah:
Merged!

@chris-crone can you handle opening tickets for the follow-up (better implementation of the all namespaces endpoint)?

Issue #1036, opened 2018-05-03
Fix mounts output style to match the rest of output


essamhassan:
**Description**
Provided an example output of pretty print
you can see that mounts output does not match the rest. In mounts its using a = instead of : in the rest of the output. Same for Labels

```
ID:		x2l4mdjijqncfeebbq7zmz5tr
Name:		testie
Service Mode:	Replicated
 Replicas:	1
Placement:
UpdateConfig:
 Parallelism:	1
 On failure:	pause
 Monitoring Period: 5s
 Max failure ratio: 0
 Update order:      stop-first
RollbackConfig:
 Parallelism:	1
 On failure:	pause
 Monitoring Period: 5s
 Max failure ratio: 0
 Rollback order:    stop-first
ContainerSpec:
 Image:		nginx:alpine@sha256:3a44395131c5a9704417d19ab4c8d6cb104013659f5babb2f1c632e789588196
Labels:
 bar=baz
 foo=bar
Mounts:
  Target = /bax
   Source = foobar
   ReadOnly = false
   Type = volume
  Target = /baz
   Source = foobar
   ReadOnly = false
   Type = volume
Resources:
Networks: testnet 
Endpoint Mode:	vip
```

**Steps to reproduce the issue:**
1. run 
```
$ docker service create \
  --label foo=bar \
  --label bar=baz \
  --mount type=volume,src=foobar,target=/bax \
  --mount type=volume,src=foobar,target=/baz \
  --network testnet \
  --name testie \
  nginx:alpine
```
2. run `docker service inspect testie --pretty`

**Describe the results you received:**
running `docker service inspect testie --pretty` outputs mounts using = 

**Describe the results you expected:**
mount output should use : instead of =

**Additional information you deem important (e.g. issue happens only occasionally):**

**Additional environment details (AWS, VirtualBox, physical, etc.):**

essamhassan:
@vdemeester this is related to our previous discussion about configs and secrets

thaJeztah:
fixed in https://github.com/docker/cli/issues/1037

PR #1037, opened 2018-05-03
use : instead of = in mounts service pretty inspect

essamhassan:
Signed-off-by: Essam A. Hassan <es.hassan187@gmail.com>

**- What I did**
Very trivial change for #1036 . I changed pretty print output from using equals to colons when printing mounts

**- How I did it**
Just modifying the pretty print template

**- How to verify it**
`docker inspect service [anyservicewithmounts] --pretty`

use colons instead of equals in mounts service pretty inspect

husky

### DIFF
--- a/cli/command/formatter/service.go
+++ b/cli/command/formatter/service.go
@@ -92,10 +92,10 @@ ContainerSpec:
 Mounts:
 {{- end }}
 {{- range $mount := .ContainerMounts }}
-  Target = {{ $mount.Target }}
-   Source = {{ $mount.Source }}
-   ReadOnly = {{ $mount.ReadOnly }}
-   Type = {{ $mount.Type }}
+ Target:	{{ $mount.Target }}
+  Source:	{{ $mount.Source }}
+  ReadOnly:	{{ $mount.ReadOnly }}
+  Type:		{{ $mount.Type }}
 {{- end -}}
 {{- if .Configs}}
 Configs:


cli/command/formatter/service.go
```diff
@@ -92,10 +92,10 @@ ContainerSpec:
 Mounts:
 {{- end }}
 {{- range $mount := .ContainerMounts }}
-  Target = {{ $mount.Target }}
-   Source = {{ $mount.Source }}
-   ReadOnly = {{ $mount.ReadOnly }}
-   Type = {{ $mount.Type }}
+ Target:	{{ $mount.Target }}
```
thaJeztah: I see the indentation changed; was this on purpose, or accidentally? (i.e. it should be indented under the `Mounts:` section)

cli/command/formatter/service.go
```diff
@@ -92,10 +92,10 @@ ContainerSpec:
...
 {{- end }}
 {{- range $mount := .ContainerMounts }}
-  Target = {{ $mount.Target }}
-   Source = {{ $mount.Source }}
-   ReadOnly = {{ $mount.ReadOnly }}
-   Type = {{ $mount.Type }}
+ Target: {{ $mount.Target }}
+  Source: {{ $mount.Source }}
+  ReadOnly: {{ $mount.ReadOnly }}
+  Type: {{ $mount.Type }}
```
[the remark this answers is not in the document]
thaJeztah: Yes, I'm +1 on cleaning up the overall output (let's keep that for a separate change :+1:)

cli/command/formatter/service.go
essamhassan: I changed the values identations to align values under each other. I guess that the keys are already matching the outer identation as other blocks (ie the keys identation under the Mounts: section)

thaJeztah: Ah, thanks, yes looks like Github showed it a bit odd

cli/command/formatter/service.go
```diff
@@ -92,10 +92,10 @@ ContainerSpec:
 Mounts:
 {{- end }}
 {{- range $mount := .ContainerMounts }}
-  Target = {{ $mount.Target }}
-   Source = {{ $mount.Source }}
-   ReadOnly = {{ $mount.ReadOnly }}
-   Type = {{ $mount.Type }}
+ Target:	{{ $mount.Target }}
+  Source:	{{ $mount.Source }}
+  ReadOnly: {{ $mount.ReadOnly }}
```
thaJeztah: Looks like this one has to be a tab as well;

```
Mounts:
 Target:	/bar
  Source:	foo
  ReadOnly: false
  Type:		volume
```

essamhassan: identation is acting weird since the last change I just pushed works well for me you can see the attached image

thaJeztah: Yeah, we should probably use a tabwriter for this, so that alignment always works independent of local settings of the shell

thaJeztah:
ping @vdemeester PTAL

PR #1038, opened 2018-05-04
Switch from x/net/context to context

kolyshkin:
Since go 1.7, "context" is a standard package. Since go 1.9, x/net/context merely provides some types aliased to those in the standard context package, so there is no functional change.

The changes were performed by the following script:
```sh
for f in $(git ls-files \*.go | grep -v ^vendor/); do
	sed -i 's|golang.org/x/net/context|context|' $f
       	goimports -w $f
	awk '/^$/ {e=1; next;}
		/\t"context"$/ {e=0;}
		{if (e) {print ""; e=0}; print;}' < $f > $f.new && \
			mv $f.new $f
	goimports -w $f
done
```

Similar PRs:
* https://github.com/docker/libnetwork/pull/2140
* https://github.com/containerd/containerd/pull/2305
* https://github.com/moby/moby/pull/36904

### DIFF
--- a/cli/command/service/list.go
+++ b/cli/command/service/list.go
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"context"
 	"fmt"
 	"sort"
 
@@ -14,7 +15,6 @@ import (
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/swarm"
 	"github.com/spf13/cobra"
-	"golang.org/x/net/context"
 )
 
 type listOptions struct {
--- a/cli/command/system/version_test.go
+++ b/cli/command/system/version_test.go
@@ -1,6 +1,7 @@
 package system
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"testing"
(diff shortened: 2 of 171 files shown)


vdemeester:
```
cli/command/container/run.go:159::error: the cancelFun function is not used on all paths (possible context leak) (vet)
cli/command/container/run.go:164::error: this return statement may be reached without using the cancelFun var defined on line 159 (vet)
cli/command/container/start.go:57::error: the cancelFun function is not used on all paths (possible context leak) (vet)
cli/command/container/start.go:63::error: this return statement may be reached without using the cancelFun var defined on line 57 (vet)
```

kolyshkin:
Those are the errors uncovered by the switch to the standard package (that's a deficiency in go vet I guess). I have added commits to fix those.

cli/command/system/version_test.go
```diff
@@ -5,14 +5,15 @@ import (
 	"strings"
 	"testing"
 
+	"context"
```
thaJeztah: nit: should be grouped with the other stdlib imports

cli/command/service/list.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"sort"
 
+	"context"
```
thaJeztah: nit: should be grouped with the above

kolyshkin:
Sure, let me update it. I was curious why my script was not working and it is because there are more than 2 groups of imports (separated by an empty line), and in case the "context" is not in the first two groups, one need to repeat the awk/goimports mantra twice.

Patch updated.

PR #1040, opened 2018-05-07
Document reserved namespaces deprecation

cyli:
Document that reserved namespaces org.docker.*, io.docker.*, and org.dockerproject.* in engine labels are now deprecated.

See https://github.com/moby/moby/pull/36921

cc @thaJeztah

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -19,6 +19,16 @@ The following list of features are deprecated in Engine.
 To learn more about Docker Engine's deprecation policy,
 see [Feature Deprecation Policy](https://docs.docker.com/engine/#feature-deprecation-policy).
 
+### Reserved namespaces in engine labels
+
+**Deprecated in Release: v18.06.0**
+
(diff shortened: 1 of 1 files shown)


docs/deprecated.md
```diff
@@ -19,6 +19,16 @@ The following list of features are deprecated in Engine.
 To learn more about Docker Engine's deprecation policy,
 see [Feature Deprecation Policy](https://docs.docker.com/engine/#feature-deprecation-policy).
 
+### Reserved namespaces in engine labels
+
+**Deprecated in Release: v18.05.0**
```
thaJeztah: Looks like it was merged after 18.05 was cut, so can you update this to 18.06?

The deprecation policy states we use 3 stable releases (https://docs.docker.com/engine/#installation-guides), so it should become an error in 18.12

cyli: Done, so sorry for the delay!  Thanks!

Issue #153, opened 2017-06-05
Memo per image


QwertyZW:
I wasn't sure where to open this feature request, it might be more appropriately placed in moby/*

I'd like the ability to add a comment/description to new image layers on every commit. Something like a commit message in a versioning system.

So I should be able to do something like
```
docker commit <running_container> <image_to_be_saved> -m "This new layer adds <thing1> and <thing2>
```

and when I do something like 
```
docker image <name> log
```

I would be able to see my commit messages.

dnephin:
That already exists. `docker commit -m "msg"`.

They can be viewed using `docker history <image>`

Going to close this issue, but let me know if that doesn't work for some reason.

PR #1041, opened 2018-05-07
Fix issue #153: Create network name if undefined

justyntemme:
As detailed in Issue #153 (https://github.com/docker/for-linux/issues/153)  , the docker create function requires a name for the network. This PR checks the length of the argument string, and generates a random number, then prepends this to "_network". I also changed the requires argument function to explicitly say the function can take at minimum 0 arguments and at most 1. The unit test, checking if the correct error message would show when the wrong number of arguments were given, also had to be updated.

**- What I did**
I have added the ability to create a network name if not provided in the `docker network create` command. I changed the unit test for this function.

**- How I did it**
I tested the length of the array of arguments to find if the name of the network was provided. If the length was greater than 0 the command functions as previously implemented. If the arguments provide no name, a random number is generated, than pre-pended to the text "_network". I changed the ExactArgs line to RequiresRangeArgs(0, 1), and tested this via the unit test file by sending the function two arguments, and comparing against the expected result. 

 **- How to verify it**
Run docker network create: without any arguments

Adds ability to create random network name when not provided.

See my profile image :)

### DIFF
--- a/cli/command/network/create.go
+++ b/cli/command/network/create.go
@@ -2,8 +2,11 @@ package network
 
 import (
 	"fmt"
+	"math/rand"
 	"net"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
@@ -47,9 +50,16 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create [OPTIONS] NETWORK",
 		Short: "Create a network",
-		Args:  cli.ExactArgs(1),
+		Args:  cli.RequiresRangeArgs(0, 1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			options.name = args[0]
+			if len(args) > 0 {
+				options.name = args[0]
+			} else {
+				//Generate a Random number to name network if not given
+				rand.Seed(time.Now().Unix())
+				options.name = (strconv.Itoa(rand.Intn(1000000)) + "_network")
+			}
+
 			return runCreate(dockerCli, options)
 		},
 	}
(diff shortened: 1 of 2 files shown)


thaJeztah:
Thanks for contributing; this change may need some discussion, as there may be deliberate reasons for not generating names.

Some initial thoughts:

- If we decide this feature should be added, this likely has to be implemented in the _daemon_, not the CLI (i.e.; the API should be changed to accept network-create requests without a name, and `dockerd` should generate a name)
- Network-names are also used in DNS registration in the embedded DNS server, so ideally the name should be a valid DNS component;

```
docker network create mynetwork
docker run --rm --network=mynetwork --name mycontainer alpine ping -c1 mycontainer.mynetwork
PING mycontainer.mynetwork (172.19.0.2): 56 data bytes
64 bytes from 172.19.0.2: seq=0 ttl=64 time=0.048 ms

--- mycontainer.mynetwork ping statistics ---
1 packets transmitted, 1 packets received, 0% packet loss
round-trip min/avg/max = 0.048/0.048/0.048 ms
```

thaJeztah:
ping @mavenugo @fcrisciani PTAL

justyntemme:
Thanks for your review, I will watch if the implementation needs to change according to the discussion and commit accordingly.

Benehiko:
@justyntemme Thank you for taking the time to create this PR!

I've been looking at this as well and agree that the name generation should not be in the CLI. I see for containers we generate a new name if not provided and the way we do so is by having an optional query parameter `name` e.g. `POST /containers/create?name=""` 
https://github.com/moby/moby/blob/master/client/container_create.go#L73-L75

The `daemon` then calls the `daemon.newContainer()` method which then uses the provided name or generates a new one here https://github.com/moby/moby/blob/master/daemon/container.go#L128

I suggest we close this PR here and rather continue the discussion on the linked issue https://github.com/docker/for-linux/issues/153.

Issue #850, opened 2018-01-30
docker system prune warn message should display filters if any


vdemeester:
**Description**

When running a system prune with filter, the warning message doesn't reflect that the filter will be used.
We should display that the elements pruned will be filtered.

**Describe the results you received:**

```bash
$ docker system prune --filter=label=foo                                                                                                      Tue 30 Jan 2018 03:16:42 PM PST
WARNING! This will remove:
        - all stopped containers
        - all networks not used by at least one container
        - all dangling images
        - all build cache
# […]
```

**Describe the results you expected:**

Same message but that acknowledge the filtering.

adshmh:
@vdemeester Does an output like the following look OK?

```
$ docker system prune --filter=label=foo
WARNING! This will remove:
        - all stopped containers
        - all networks not used by at least one container
        - all dangling images
        - all build cache

  Elements pruned will be filtered with:
      label=foo
```

adshmh:
Also it seems that volume, image, and container prune commands do not display the filter either.

olejorgenb:
The merged MR doesn't seem to be in effect anymore? https://github.com/docker/cli/blob/master/cli/command/system/prune.go

It is quite jarring that the warning shouts "ALL" when I've applied a filter.

thaJeztah:
Filters are printed as part of the template at https://github.com/docker/cli/blob/master/cli/command/system/prune.go#L57

But not all objects (volumes, networks, containers, images) accept the same set of filters

olejorgenb:
```
$ docker image prune --all --filter until=240h
WARNING! This will remove all images without at least one container associated to them.
Are you sure you want to continue? [y/N] n
$ docker --version
Docker version 20.10.7, build 20.10.7-0ubuntu5~21.04.2
```
Shouldn't the filter be displayed here? `until` is clearly supported by `image prune`

PR #1043, opened 2018-05-08
Fix issue #850, Warn user of filter when pruning

justyntemme:
Append a warning with which filter the elements will be pruned with if filter is set.
This is a fix for issue https://github.com/docker/cli/issues/850

**- What I did**
Added a message in prune warning if filter is applied.

**- How I did it**

I have checked the length of the string of the options.prune argument, and appended the warning along with the string value of that arguments supplied.

**- How to verify it**
docker system prune --filter=label=foo

Added warning to prune command when filter is applied.

My icon :)


thaJeztah:
ping @vdemeester PTAL

PR #1049, opened 2018-05-10
Default path added to the docker build command

WoodProgrammer:
I added the default path parameter to docker build command if client wants the build Dockerfile at the current directory .  @hasanemrebeyy 

I change 115.line of code, added only one if statement.

Link at below.
 https://github.com/WoodProgrammer/cli/blob/master/cli/command/image/build.go

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -112,7 +112,11 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 	flags.VarP(&options.tags, "tag", "t", "Name and optionally a tag in the 'name:tag' format")
 	flags.Var(&options.buildArgs, "build-arg", "Set build-time variables")
 	flags.Var(options.ulimits, "ulimit", "Ulimit options")
-	flags.StringVarP(&options.dockerfileName, "file", "f", "", "Name of the Dockerfile (Default is 'PATH/Dockerfile')")
+
+	if &options.dockerfileName == " " {
+		&options.dockerfileName = '.'
+	}
+
 	flags.VarP(&options.memory, "memory", "m", "Memory limit")
 	flags.Var(&options.memorySwap, "memory-swap", "Swap limit equal to memory plus swap: '-1' to enable unlimited swap")
 	flags.Var(&options.shmSize, "shm-size", "Size of /dev/shm")
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -109,14 +109,15 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 
 	flags := cmd.Flags()
 	
-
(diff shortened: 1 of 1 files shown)


cli/command/image/build.go
```diff
@@ -112,7 +112,11 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 	flags.VarP(&options.tags, "tag", "t", "Name and optionally a tag in the 'name:tag' format")
 	flags.Var(&options.buildArgs, "build-arg", "Set build-time variables")
 	flags.Var(options.ulimits, "ulimit", "Ulimit options")
-	flags.StringVarP(&options.dockerfileName, "file", "f", "", "Name of the Dockerfile (Default is 'PATH/Dockerfile')")
```
eiais: This line allows the path to the dockerfile to be passed on the command line. If you remove it you wouldn't be able to specify the path.

cli/command/image/build.go
```diff
@@ -112,7 +112,11 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
...
+
+	if &options.dockerfileName == " " {
+		&options.dockerfileName = '.'
```
eiais: I think the default path is set in https://github.com/docker/cli/blob/master/cli/command/image/build/context.go#L31

hasanemrebeyy: We think that there is no difference between in "docker build" with "docker build ." , if a path is not specified, it must be based on the same index. Context.go is still waiting for dockerfile.

WoodProgrammer: I think it is not default Dockerfile ' PATH ' , it is only default dockerfile name. I want to add the default ' Dockerfile Path ' .

WoodProgrammer:
I've added else statement for specify the dockerfile path.

eiais:
The `.` in `docker build .` is being set to options.context. Look at line 105.

thaJeztah:
@WoodProgrammer could you squash, and sign-off your commits, otherwise we won't be able to accept or review your contribution. Let me know if you need help doing so.

This change also needs a design review, because it's changing the default behavior of the `docker build` command.

cli/command/image/build.go
eiais: If you look at line 103, you can see that docker build currently requires at least one argument.

cpuguy83:
This to me is a risky change as it's changing the behavior of how the arguments are parsed and it's really not a big benefit.

eiais:
@cpuguy83 Do you think anyone is relying on `docker build` printing the help text or returning non-zero?

cpuguy83:
@eiais Most definitely. Especially automated systems.
For example, what if the path to the context was passed down through some variable that was mistakenly set to an empty string?

eiais:
@cpuguy83 Thanks for the example. Makes sense.

PR #1053, opened 2018-05-11
Change error message for unreadable files

justyntemme:
Fix for https://github.com/docker/for-linux/issues/44

**- What I did**
Change the error message for unreadable files to mention the possibility of a .Dockerignore entry for multi developer teams.
**- How I did it**
Simply changing the text in the error message
**- How to verify it**
Run docker and try to COPY a file with a .Dockerignore entry

Changed error message for unreadable files to clarify possibility of a .Dockerignore entry

### DIFF
--- a/cli/command/image/build/context.go
+++ b/cli/command/image/build/context.go
@@ -47,7 +47,7 @@ func ValidateContextDirectory(srcPath string, excludes []string) error {
 				return errors.Errorf("can't stat '%s'", filePath)
 			}
 			if os.IsNotExist(err) {
-				return nil
+				return errors.Errorf("file ('%s') not found or excluded by .dockerignore", filePath)
 			}
 			return err
 		}


cli/command/image/build/context.go
```diff
@@ -44,7 +44,7 @@ func ValidateContextDirectory(srcPath string, excludes []string) error {
 	return filepath.Walk(contextRoot, func(filePath string, f os.FileInfo, err error) error {
 		if err != nil {
 			if os.IsPermission(err) {
-				return errors.Errorf("can't stat '%s'", filePath)
+				return errors.Errorf("can't stat: file not found or excluded by .dockerignore '%s'", filePath)
```
cpuguy83: This doesn't seem right. This is a permission error, not a not found error.
In addition, we should just wrap the real error `errors.Wrap(err, "error reading file")`. Generally the file name should already be in the error.

justyntemme:
I have updated the PR to the correct error which should be `if os.IsNotExist`

PR #1054, opened 2018-05-11
Fix service update --host-rm not being granular enough

thaJeztah:
fixes https://github.com/moby/moby/issues/35325
fixes https://github.com/moby/moby/issues/37035

Removing a host by `<host>:<ip>` should only remove occurences of the host with
a matching IP-address, instead of removing all entries for that host.

In addition, combining `--host-rm` and `--host-add` for the same host should
result in the new host being added.

This patch fixes the way the diff is calculated to allow combining
removing/adding, and to support entries having both a canonical, and aliases.
Aliases cannot be added by the CLI, but are supported in the Service spec, thus
should be taken into account:

Entries can be removed by either a specific `<host-name>:<ip-address>`
mapping, or by `<host>` alone:

 - If both IP-address and host-name is provided, only remove the hostname
   from entries that match the given IP-address.
 - If only a host-name is provided, remove the hostname from any entry it
   is part of (either as _canonical_ host-name, or as _alias_).
 - If, after removing the host-name from an entry, no host-names remain in
   the entry, the entry itself should be removed.

For example, the list of host-entries before processing could look like this:

    hosts = &[]string{
        "127.0.0.2 host3 host1 host2 host4",
        "127.0.0.1 host1 host4",
        "127.0.0.3 host1",
        "127.0.0.1 host1",
    }

Removing `host1` removes every occurrence:

    hosts = &[]string{
        "127.0.0.2 host3 host2 host4",
        "127.0.0.1 host4",
    }

Whereas removing `host1:127.0.0.1` only remove the host if the IP-address matches:

    hosts = &[]string{
        "127.0.0.2 host3 host1 host2 host4",
        "127.0.0.1 host4",
        "127.0.0.3 host1",
    }

Before this patch:

    $ docker service create --name my-service --host foo:127.0.0.1  --host foo:127.0.0.2 --host foo:127.0.0.3 nginx:alpine
    $ docker service update --host-rm foo:127.0.0.1 --host-add foo:127.0.0.4 my-service
    $ docker service inspect --format '{{.Spec.TaskTemplate.ContainerSpec.Hosts}}' my-service
    []

After this patch is applied:

    $ docker service create --name my-service --host foo:127.0.0.1  --host foo:127.0.0.2 --host foo:127.0.0.3 nginx:alpine
    $ docker service update --host-rm foo:127.0.0.1 --host-add foo:127.0.0.5 my-service
    $ docker service inspect --format '{{.Spec.TaskTemplate.ContainerSpec.Hosts}}' my-service
    [127.0.0.2 foo 127.0.0.3 foo 127.0.0.4 foo]

```Markdown
* Fix `docker service update --host-add` does not update existing host entry [docker/cli#1054](https://github.com/docker/cli/pull/1054)
```

### DIFF
--- a/cli/command/service/update.go
+++ b/cli/command/service/update.go
@@ -963,45 +963,99 @@ func updateReplicas(flags *pflag.FlagSet, serviceMode *swarm.ServiceMode) error 
 	return nil
 }
 
+type hostMapping struct {
+	IPAddr string
+	Host   string
+}
+
 // updateHosts performs a diff between existing host entries, entries to be
 // removed, and entries to be added. Host entries preserve the order in which they
 // were added, as the specification mentions that in case multiple entries for a
 // host exist, the first entry should be used (by default).
+//
+// Note that, even though unsupported by the the CLI, the service specs format
+// allow entries with both a _canonical_ hostname, and one or more aliases
+// in an entry (IP-address canonical_hostname [alias ...])
+//
+// Entries can be removed by either a specific `<host-name>:<ip-address>` mapping,
+// or by `<host>` alone:
+//
+// - If both IP-address and host-name is provided, the hostname is removed only
+//   from entries that match the given IP-address.
+// - If only a host-name is provided, the hostname is removed from any entry it
+//   is part of (either as canonical host-name, or as alias).
+// - If, after removing the host-name from an entry, no host-names remain in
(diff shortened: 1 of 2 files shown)


thaJeztah:
ping @vdemeester @cpuguy83 @silvin-lubecki PTAL

PR #1059, opened 2018-05-15
Kubernetes orchestrator all namespaces

mat007:
**- What I did**

I made ```--orchestrator=all``` behave as if also ```--all-namespaces``` unless specific namespaces have been queried with ```--namespace```

**- How I did it**

Opts.AllNamespaces gets set to true if requesting for all orchestrators without `specifying` a namespace

Setting `--orchestrator=all` also sets `--all-namespaces` unless specific `--namespace` are set.

image

### DIFF
--- a/cli/command/stack/kubernetes/list.go
+++ b/cli/command/stack/kubernetes/list.go
@@ -10,6 +10,7 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/formatter"
 	"github.com/docker/cli/cli/command/stack/options"
+	"github.com/docker/cli/cli/config/configfile"
 	"github.com/pkg/errors"
 	core_v1 "k8s.io/api/core/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
@@ -19,9 +20,16 @@ import (
 // GetStacks lists the kubernetes stacks
 func GetStacks(kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error) {
 	if opts.AllNamespaces || len(opts.Namespaces) == 0 {
+		if isAllNamespacesDisabled(kubeCli.ConfigFile().Kubernetes) {
+			opts.AllNamespaces = true
+		}
 		return getStacksWithAllNamespaces(kubeCli, opts)
 	}
 	return getStacksWithNamespaces(kubeCli, opts, removeDuplicates(opts.Namespaces))
+}
+
+func isAllNamespacesDisabled(kubeCliConfig *configfile.KubernetesConfig) bool {
+	return kubeCliConfig == nil || kubeCliConfig != nil && kubeCliConfig.AllNamespaces != "disabled"
 }
 
 func getStacks(kubeCli *KubeCli, opts options.List) ([]*formatter.Stack, error) {
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -46,6 +46,7 @@ type ConfigFile struct {
 	Proxies              map[string]ProxyConfig      `json:"proxies,omitempty"`
 	Experimental         string                      `json:"experimental,omitempty"`
 	Orchestrator         string                      `json:"orchestrator,omitempty"`
+	Kubernetes           *KubernetesConfig           `json:"kubernetes,omitempty"`
 }
 
 // ProxyConfig contains proxy configuration settings
@@ -54,6 +55,11 @@ type ProxyConfig struct {
 	HTTPSProxy string `json:"httpsProxy,omitempty"`
 	NoProxy    string `json:"noProxy,omitempty"`
 	FTPProxy   string `json:"ftpProxy,omitempty"`
+}
+
+// KubernetesConfig contains Kubernetes orchestrator settings
+type KubernetesConfig struct {
+	AllNamespaces string `json:"allNamespaces,omitempty"`
 }
 
 // New initializes an empty configuration file for the given filename 'fn'
--- a/cli/flags/common.go
+++ b/cli/flags/common.go
@@ -55,8 +55,6 @@ func (commonOpts *CommonOptions) InstallFlags(flags *pflag.FlagSet) {
 	flags.StringVarP(&commonOpts.LogLevel, "log-level", "l", "info", `Set the logging level ("debug"|"info"|"warn"|"error"|"fatal")`)
 	flags.BoolVar(&commonOpts.TLS, "tls", dockerTLS, "Use TLS; implied by --tlsverify")
 	flags.BoolVar(&commonOpts.TLSVerify, FlagTLSVerify, dockerTLSVerify, "Use TLS and verify the remote")
-	flags.StringVar(&commonOpts.Orchestrator, "orchestrator", "", "Orchestrator to use (swarm|kubernetes|all) (experimental)")
-	flags.SetAnnotation("orchestrator", "experimentalCLI", nil)
 
 	// TODO use flag flags.String("identity"}, "i", "", "Path to libtrust key file")
 
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -51,6 +51,11 @@ func newDockerCommand(dockerCli *command.DockerCli) *cobra.Command {
 	flags.StringVar(&opts.ConfigDir, "config", cliconfig.Dir(), "Location of client config files")
 	opts.Common.InstallFlags(flags)
 
+	// Install persistent flags
+	persistentFlags := cmd.PersistentFlags()
+	persistentFlags.StringVar(&opts.Common.Orchestrator, "orchestrator", "", "Orchestrator to use (swarm|kubernetes|all) (experimental)")
+	persistentFlags.SetAnnotation("orchestrator", "experimentalCLI", nil)
+
 	setFlagErrorFunc(dockerCli, cmd, flags, opts)
 
 	setHelpFunc(dockerCli, cmd, flags, opts)


mat007:
Split from https://github.com/docker/cli/pull/1031, only the last commit is new.

mat007:
For reference, here is the current state of the discussion from #1031

@thaJeztah
> I still have reservations on the magic relation between --orchestrator=all and --all-namespaces (copying my earlier comment #1031 (comment) here, as it was now hidden);
>
> I'm having trouble conflating these; this may become very confusing:
>* configuration has "orchestrator: all": docker stack ls shows all stacks (both swarm and k8s)
>* I want to view only the k8s stacks, so I add --orchestrator=kubernetes. Now I no longer see all my stacks, only those in the default namespace
>* confused

@chris-crone
> The thinking behind this was as follows: When you list stacks across orchestrators your intention is to see all the stacks running on your cluster. When you specify kubernetes as your orchestrator; you are doing something more specific and want similar behaviour to what kubectl gives you.
>
> I can see how this might be confusing for your flow. There will be a NAMESPACE column which should help the user understand what has happened.

@chris-crone
> Ah, see you're referring to the case when a user has set orchestrator to all in their config file. I don't think many users will do this as they will only be able to list resources. In order to manipulate resources, they will need to specify an orchestrator.
>
> I don't think many users will do this as they will only be able to list resources.

@thaJeztah
> I'm actually not sure: I can definitely see it being useful to just run any docker <foo> list or docker <foo> ps command, and see everything I have running, then use --filter <something> and/or --orchestrator=x to limit the list to what I'm looking for, or (when deploying / interacting with objects), pass --orchestrator=foo --namespace=bar as part of the steps to deploy.
>
> Which brings me to two things to consider:
> *    Add orchestrator and namespace as filter options (--filter orchestrator=swarm) (no high-priority, but for consistency)
>*    When interacting with existing objects, only require --orchestrator and/or --namespace if the object refered to is ambiguous, for example, if there's only one stack with a given name, there would be no need to ask for --orchestrator.

thaJeztah:
So

- connecting `orchestrator=all` with `--all-namespaces` means that
  - `--orchestrator=all`: show all swarm objects and all kubernetes objects, in all namespaces
  - switching to `--orchestrator=swarm`: shows all swarm objects from the list above
  - switching to `--orchestrator=kubernetes`: show only kubernetes objects if they're in the `"default"` namespace (**this is the confusing bit**)

Thoughts: we could;

1. special-case the `"default"` namespace, and consider it (for list-views) equivalent of "no namespace set", in which case we always show "all namespaces", _unless_ user manually passes a namespace `--namespace=foo`
2. irregardless of the configured namespace, default to show all namespaces for kubernetes (in list views), _unless_ the user manually passes a namespace `--namespace=foo` (**this one is tricky, because we then _ignore_ the user's preference**)
3. add a configuration option to enable `--all-namespaces` by default (or, the reverse: allow opt-out and _dissable_ the default `--all-namespaces`)
4. add a configuration option to specify namespaces to use: (`"namespaces":["all"]` or `"namespaces":["myteam-stage", "myteam-prod"]`)
5. combination of 3. and 4. (to prevent the `"all"` magic value), in which case `"namespaces"` defaults to the "default" namespace (as configured by the user), and `--all-namespaces` either enabled by default or not (whatever we decide on: opt-in/opt-out) 

I'm leaning towards 4. (or 3.), as it keeps the user in control;

- Allows to see all namespaces by default, both for `--orchestrator=all` and `--orchestrator=kubernetes`, which may be a use-case if the user manages multiple namespaces (and orchestrators)
- Allows to _only_ see namespaces that the user is working on (which could be "some" of the ones they have access to), both when viewing kubernetes-only and when viewing "all" orchestrators
- No magic/unexpected behavior

silvin-lubecki:
cc @thaJeztah @vdemeester @mat007 I'm opened to any comment on the config file option 😄 
I took the third option among those proposed by @thaJeztah

silvin-lubecki:
Some details on the UX:
```
# Use kubernetes as default orchestrator
$ cat ~/.docker/config.json
{
  "experimental" : "enabled",
  "orchestrator" : "kubernetes"
}

$ docker version
...
 Orchestrator: kubernetes
...

# Check there is no stack
$ docker stack ls
NAME                SERVICES            ORCHESTRATOR

# Deploy a stack on the default namespace, optionally defined in the current context in your ~/.kube/config file 
$ docker stack deploy mystack -c docker-compose.yml
...

# Check we can list the stack in the default namespace
$ docker stack ls
mystack             5                   Kubernetes          default

# Add a new namespace
$ kubectl create namespace mynamespace
namespace "mynamespace" created

# And deploy another stack in it
$ docker stack deploy mystack -c docker-compose.yml --namespace mynamespace
...

# By default, list all stacks in all namespaces
$ docker stack ls
mystack             5                   Kubernetes          default
mystack             5                   Kubernetes          mynamespace

# List can be restricted to one or multiple namespaces
$ docker stack ls --namespace default
mystack             5                   Kubernetes          default

$ docker stack ls --namespace mynamespace --namespace default                                                                        NAME                SERVICES            ORCHESTRATOR        NAMESPACE
mystack             5                   Kubernetes          default
mystack             5                   Kubernetes          mynamespace

# Deploy a stack on Swarm
$ docker stack deploy mystack -c docker-compose.yml --orchestrator swarm
...

$ docker stack ls --orchestrator swarm
NAME                SERVICES            ORCHESTRATOR
mystack             5                   Swarm

# Listing on all orchestrators also lists all namespaces
$ docker stack ls --orchestrator all
mystack             5                   Kubernetes          default
mystack             5                   Kubernetes          mynamespace
mystack             5                   Swarm

# Listing all namespaces behavior by default can be disabled in the docker config file
$ cat ~/.docker/config.json
{
  "experimental" : "enabled",
  "orchestrator" : "kubernetes",
  "stackListAllNamespaces" : "disabled"
}

# Now it lists using the default configured namespace
$ docker stack ls
mystack             5                   Kubernetes          default

# List can be extended to one or multiple namespaces
$ docker stack ls --namespace mynamespace
mystack             5                   Kubernetes          mynamespace

$ docker stack ls --namespace mynamespace --namespace default                                                                        NAME                SERVICES            ORCHESTRATOR        NAMESPACE
mystack             5                   Kubernetes          default
mystack             5                   Kubernetes          mynamespace

# Listing all namespaces is now explicit
$ docker stack ls --all-namespaces
mystack             5                   Kubernetes          default
mystack             5                   Kubernetes          mynamespace
```

cli/config/configfile/file.go
```diff
@@ -24,28 +24,29 @@ const (
...
+	ServicesFormat         string                      `json:"servicesFormat,omitempty"`
+	TasksFormat            string                      `json:"tasksFormat,omitempty"`
+	SecretFormat           string                      `json:"secretFormat,omitempty"`
+	ConfigFormat           string                      `json:"configFormat,omitempty"`
+	NodesFormat            string                      `json:"nodesFormat,omitempty"`
+	PruneFilters           []string                    `json:"pruneFilters,omitempty"`
+	Proxies                map[string]ProxyConfig      `json:"proxies,omitempty"`
+	Experimental           string                      `json:"experimental,omitempty"`
+	Orchestrator           string                      `json:"orchestrator,omitempty"`
+	StackListAllNamespaces string                      `json:"stackListAllNamespaces,omitempty"`
```
vdemeester: I wonder if we should start to using some `nested` struct, like 
```
{
  "orchestrator": "kubernetes",
  "stack": {
    "list": {
      "allNamespaces": "disabled",
      "format": "…",
    }
  }
}
```

cc @thaJeztah

silvin-lubecki: Why not, but maybe in a follow up?

thaJeztah: Will we only use this configuration for _stacks_, or also for `docker service ls` ? (wondering about the `Stack` prefix.

Alternatively, if we want to use a boolean instead of the `disabled` string, we could use `{"singleNamespace": true}`

silvin-lubecki: Ok so what about
```
{
  "kubernetes": {
    "singleNamespace": true
  }
}
```
And by the way, singleNamespace makes you think you can only list one namespace, but you still can use multiple `--namespace` in the `docker stack ls` command, so it could be a little confusing?

vdemeester: @thaJeztah hum good point.
I tend to agree with @silvin-lubecki on the `singleNamespace` name… but nothing really pop in my head for a correct name.

vdemeester: SGTM

thaJeztah:
Looks like there's a linting failure;

```
cli/config/configfile/file.go:60:6:warning: exported type KubernetesConfig should have comment or be unexported (golint)
Exited with code 1 
```

silvin-lubecki:
Indeed 😔

silvin-lubecki:
PTAL 😄

cli/command/stack/kubernetes/list.go
```diff
@@ -19,11 +20,18 @@ import (
...
+		if isAllNamespacesDisabled(kubeCli.ConfigFile().Kubernetes) {
+			opts.AllNamespaces = true
+		}
 		return getStacksWithAllNamespaces(kubeCli, opts)
 	}
 	return getStacksWithNamespaces(kubeCli, opts, removeDuplicates(opts.Namespaces))
 }
 
+func isAllNamespacesDisabled(kubeCliConfig *configfile.KubernetesConfig) bool {
+	return kubeCliConfig == nil || kubeCliConfig != nil && kubeCliConfig.AllNamespaces != "disabled"
```
thaJeztah: Oh, one thing: do we want a strict check to only accept "disabled", "enabled" or `nil` ?

vdemeester: yes, @silvin-lubecki will follow up :angel:

silvin-lubecki: Sure I will!

silvin-lubecki:
Is this the only documentation we have about the CLI configuration? Nothing else on docs.docker.com?

thaJeztah:
We may need some more in general (about working with multiple orchestrators); not sure if "orchestrator" was already documented in the configuration; also some flags/options are currently experimental, so we'll have to do a "sweep" after they leave experimental, so see if everything is there.

Would be good to have a tracking issue in which we can add things that we think about (could be a checkbox-list style)

silvin-lubecki:
Not yet in the documentation as it was behind the experimental feature flag.

PR #1064, opened 2018-05-17
Create standalone e2e test image

chris-crone:
**- What I did**
- Build a standalone e2e test image that can be run against a cluster
- Add `docker stack` e2e tests for Kubernetes

**- How I did it**

Test modifications:
- Skip tests that require engine configuration modifications when running against a remote endpoint.
- Be less strict about CLI outputs to accommodate running the tests against a Swarm classic endpoint.
- Refactor stack tests so that we run the same test against Swarm and Kubernetes.
- Skip Kubernetes tests unless we have a `KUBECONFIG`.
- Add ability to skip tests that manipulate plugins so that this can be used against EE.

e2e image modifications:

- Do not require mounting volumes to run – image is now completely self-contained.

**- How to verify it**
- The existing flow for running e2e tests should work as before: `make -f docker.Makefile test-e2e`
- Run the image against a cluster running swarm: `docker run --rm -v /path/to/certs:/bundle -e DOCKER_HOST=tcp://host_ip:2376 -e DOCKER_CERT_PATH=/bundle -e DOCKER_TLS_VERIFY=1 -e REMOTE_DAEMON=1 -e SKIP_PLUGIN_TESTS=1 docker-cli-e2e`
- Run the image against a cluster running EE: `docker run --rm -v /path/to/certs:/bundle -e DOCKER_HOST=tcp://host_ip:443 -e DOCKER_CERT_PATH=/bundle -e DOCKER_TLS_VERIFY=1 -e REMOTE_DAEMON=1 -e SKIP_PLUGIN_TESTS=1 docker-cli-e2e`

N/A testing only.


e2e/stack/remove_test.go
```diff
@@ -13,45 +14,66 @@ import (
...
+	golden.Assert(t, result.Stdout(),
+		fmt.Sprintf("stack-remove-%s-success.golden", orchestrator))
 }
 
-func deployFullStack(t *testing.T, stackname string) {
+func deployFullStack(t *testing.T, orchestrator, stackname string) {
 	// TODO: this stack should have full options not minimal options
-	result := icmd.RunCommand("docker", "stack", "deploy",
-		"--compose-file=./testdata/full-stack.yml", stackname)
+
```
mat007: nit: extra line

e2e/stack/remove_test.go
```diff
@@ -13,45 +14,66 @@ import (
...
+	t.Run("Kubernetes", func(t *testing.T) {
+		environment.SkipIfKubernetesDisabled(t)
 
-	result := icmd.RunCommand("docker", "stack", "rm", stackname)
+		testRemove(t, "kubernetes")
+	})
+}
+
+func testRemove(t *testing.T, orchestrator string) {
+	stackname := fmt.Sprintf("test-stack-remove-%s", orchestrator)
```
mat007: nit: could be `"test-stack-remove-" + orchestrator`

e2e/stack/remove_test.go
```diff
@@ -13,45 +14,66 @@ import (
...
+	deployFullStack(t, orchestrator, stackname)
+	defer cleanupFullStack(t, orchestrator, stackname)
+
+	result := icmd.RunCommand("docker", "--orchestrator", orchestrator,
+		"stack", "rm", stackname)
```
mat007: nit: why the new line in the middle? (same a few lines down)

internal/test/output/output.go
```diff
@@ -7,15 +7,15 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Assert checks wether the output contains the specified lines
-func Assert(t *testing.T, actual string, expectedLines map[int]func(string) error) {
+// Assert checks output lines at specified locations
+func Assert(t *testing.T, actual string, expectedLines map[int]func(string, bool) error) {
```
mat007: What is that new `bool` for? Shouldn't we document it?

internal/test/output/output.go
```diff
@@ -24,22 +24,81 @@ func Assert(t *testing.T, actual string, expectedLines map[int]func(string) erro
...
+		found := false
+		var err error
+		for count, line := range toCheck {
+			if err = cmp(line, false); err == nil {
+				toCheck = toCheck[count:]
+				found = true
+				break
+			}
+		}
+		if found == false {
```
mat007: Did you mean `if !found`? :)

internal/test/output/output.go
```diff
@@ -24,22 +24,81 @@ func Assert(t *testing.T, actual string, expectedLines map[int]func(string) erro
...
+func Prefix(expected string) func(string, bool) error {
+	return func(actual string, exact bool) error {
 		if strings.HasPrefix(actual, expected) {
 			return nil
 		}
-		return errors.Errorf("expected %s to start with %s", actual, expected)
+		if exact {
+			return errors.Errorf("expected %q to start with %q", actual, expected)
+		} else {
+			return errors.Errorf("unable to find line that starts with %q", expected)
```
mat007: I believe the `else` should be removed:
```
		if exact {
			return errors.Errorf("expected %q to start with %q", actual, expected)
		}
		return errors.Errorf("unable to find line that starts with %q", expected)
```

dockerfiles/Dockerfile.e2e
```diff
@@ -0,0 +1,38 @@
+FROM golang:1.9.5-alpine3.6
```
vdemeester: 1.10.2 :angel: :pray:

e2e/container/run_test.go
```diff
@@ -25,6 +28,8 @@ func TestRunAttachedFromRemoteImageAndRemove(t *testing.T) {
 }
 
 func TestRunWithContentTrust(t *testing.T) {
+	environment.SkipIfRemoteDaemon(t)
```
vdemeester: nit: Just to look a bit more inline with what we do on other part (in moby or else) : 

```go
skip.If(t, environment.RemoteDaemon())
```

e2e/plugin/trust_test.go
```diff
@@ -19,6 +20,8 @@ import (
 const registryPrefix = "registry:5000"
 
 func TestInstallWithContentTrust(t *testing.T) {
+	environment.SkipIfPluginTestDisabled(t)
```
vdemeester: Same as `RemoteDaemon`

e2e/container/run_test.go
chris-crone: The reason I did it like this was to have the same skip message if we're skipping for the same reason.

I guess I could do `skip.If(t, environment.RemoteDaemon(), environment.RemoteDaemonSkipMessage)`

wdyt?

dockerfiles/Dockerfile.e2e
chris-crone: Should I bump the alpine version too?

e2e/compose-env.yaml
```diff
@@ -10,12 +10,9 @@ services:
       command: ['--insecure-registry=registry:5000']
 
   notary-server:
-      image: 'notary:server-0.4.2'
+      build:
+        context: ..
```
thaJeztah: Looks like we only need `testdata` from this directory;

```Dockerfile
COPY e2e/testdata/notary/ /fixtures
```

May want to change this to

```
context: ./testdata/
```

And the Dockerfile to;

```
COPY ./notary/ /fixtures/
```

(or a variation thereof)

internal/test/environment/testenv.go
```diff
@@ -14,7 +17,78 @@ func Setup() error {
...
+}
+
+// KubernetesEnabled returns if Kubernetes testing is enabled
+func KubernetesEnabled() bool {
+	return os.Getenv("KUBECONFIG") != ""
+}
+
+// SkipIfKubernetesDisabled skips the test if Kubernetes is not enabled
+func SkipIfKubernetesDisabled(t *testing.T) {
+	skip.If(t, !KubernetesEnabled(), "Kubernetes not enabled")
```
dnephin: These helpers aren't really necessary. It should already echo the second argument which reads basically the same way as the extra string.

If you keep them I would add `t.Helper()` here so at least it reports the correct line in the skip message.

internal/test/output/output.go
```diff
@@ -24,22 +25,78 @@ func Assert(t *testing.T, actual string, expectedLines map[int]func(string) erro
 	}
 }
 
-// Prefix returns whether if the line has the specified string as prefix
-func Prefix(expected string) func(string) error {
-	return func(actual string) error {
+// AssertContains checks whether the output contains specified lines in order
+func AssertContains(t *testing.T, actual string, expectedLines []func(string, bool) error) {
```
dnephin: Why is this necessary? Previously it was working with `Assert`.

chris-crone: I added this function so that the build test would pass against Docker EE which uses a 17.06 engine. With this engine, `COPY` steps have an extra `Removing intermediate container` line. This new function doesn't check exact line numbers but instead checks for desired outputs in order.

**17.06:**
```bash
$ docker build -t copy .
Sending build context to Docker daemon   2.56kB
Step 1/2 : FROM alpine
 ---> 3fd9065eaf02
Step 2/2 : COPY something /
 ---> 577b61f8471b
Removing intermediate container 86ec5fc54328
Successfully built 577b61f8471b
Successfully tagged copy:latest
```

**18.06:**
```bash
$ docker build -t copy .
Sending build context to Docker daemon   2.56kB
Step 1/2 : FROM alpine
 ---> 3fd9065eaf02
Step 2/2 : COPY something /
 ---> 93321d321367
Successfully built 93321d321367
Successfully tagged copy:latest
```

vdemeester: > Removing intermediate container 86ec5fc54328

This line disappears ? is it by design ? cc @thaJeztah @simonferquel

e2e/testdata/Dockerfile.notary-server
```diff
@@ -0,0 +1,4 @@
+ARG NOTARY_VERSION=0.5.0
```
vdemeester: Why `0.5.0` here ? `ARG     NOTARY_VERSION=v0.6.0`, it was `0.6.0` before :stuck_out_tongue_closed_eyes: 
We should even bump to `v0.6.1`

chris-crone: latest version on the hub is 0.5.0 https://hub.docker.com/_/notary/


cc @endophage

vdemeester: ohhh :sweat:

thaJeztah: Yes, see https://github.com/docker/cli/pull/1064#discussion_r189012761

internal/test/output/output.go
thaJeztah: Was wondering the same when reading that; I don't think that line should disappear, will have to check why it's not printed.

Can confirm it's not there though;

```bash
mkdir foo && cd foo
echo bla > something

docker build --no-cache . -f-<<EOF
FROM alpine
COPY something /
COPY something /else/
EOF
```

```
Sending build context to Docker daemon  2.607kB
Step 1/3 : FROM alpine
 ---> 3fd9065eaf02
Step 2/3 : COPY something /
 ---> Using cache
 ---> 97e7adea2403
Step 3/3 : COPY something /else/
 ---> b0cdf35d65d6
Successfully built b0cdf35d65d6
Successfully tagged copy:latest
```

thaJeztah: Same result when adding `--rm` and `--force-rm` (just in case);

```
docker build --no-cache --force-rm --rm . -f-<<EOF
FROM alpine
COPY something /
COPY something /else/
EOF
```

```
Sending build context to Docker daemon  2.607kB
Step 1/3 : FROM alpine
 ---> 3fd9065eaf02
Step 2/3 : COPY something /
 ---> cf910fa96aa7
Step 3/3 : COPY something /else/
 ---> 56bcd1f3113b
Successfully built 56bcd1f3113b
```

dnephin: Instead of making the assertion less strict (and less obvious) I would prefer to have different expected values based on the version.

chris-crone: @thaJeztah Thanks for the investigation. Chatting to @simonferquel offline; he believes the missing line is a bug.

@dnephin My thinking was that we should just be looking for lines of interest in order and that was enough of a test. I agree with you that it makes the check less clear. Since this appears to be an actual bug; I will wait for it to be fixed and then remove the check function.

chris-crone: Created a moby issue for this: https://github.com/moby/moby/issues/37164

Updated PR to remove the addition of `AssertContains`.

internal/test/environment/testenv.go
```diff
@@ -14,7 +15,63 @@ func Setup() error {
...
+		}
+	}
+
+	if val := boolFromString(os.Getenv("TEST_REMOTE_DAEMON")); val {
+		if err := os.Setenv("REMOTE_DAEMON", "1"); err != nil {
+			return err
+		}
+	}
+
+	if val := boolFromString(os.Getenv("TEST_SKIP_PLUGIN_TESTS")); val {
```
silvin-lubecki: There's like a pattern here 😇, I think it could be factorized, at least between `TEST_KUBECONFIG`, `TEST_REMOTE_DAEMON` and `TEST_SKIP_PLUGIN_TESTS`

e2e/stack/deploy_test.go
```diff
@@ -1,25 +1,43 @@
...
+	"github.com/gotestyourself/gotestyourself/skip"
 )
 
 func TestDeployWithNamedResources(t *testing.T) {
-	stackname := "test-stack-deploy-with-names"
+	t.Run("Swarm", func(t *testing.T) {
+		testDeployWithNamedResources(t, "swarm")
+	})
+	t.Run("Kubernetes", func(t *testing.T) {
+		// FIXME(chris-crone): currently does not work with compose for kubernetes.
```
silvin-lubecki: I'll try to see what's wrong here.

e2e/stack/remove_test.go
```diff
@@ -1,57 +1,77 @@
...
 	}
 }
 
-func taskCount(stackname string, expected int) func(t poll.LogT) poll.Result {
+func taskCount(orchestrator, stackname string, expected int) func(t poll.LogT) poll.Result {
 	return func(poll.LogT) poll.Result {
-		result := icmd.RunCommand(
-			"docker", "stack", "ps", "-f=desired-state=running", stackname)
+		args := []string{"--orchestrator", orchestrator, "stack", "ps", stackname}
+		// FIXME(chris-crone): remove when filtering is supported with kubernetes.
```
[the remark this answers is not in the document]
chris-crone: Unfortunately not for desired state

Issue #1050, opened 2018-05-10
platform support is missing in version 3.5 and 3.6


casz:
**Description**

Docker compose version 3.6 does not support platform attribute on a service object
I noticed it was added in 2.4 for docker compose https://github.com/docker/compose/pull/5767

which according to the version matrix should be supported by 3.5 and 3.6

**Steps to reproduce the issue:**
1. use version: '3.6' in docker-compose.yml file
2. add platform to service
3. run docker-compose config

My docker-compose.yml
```
version: '3.6'

services:
  jenkins:
    platform: linux
    env_file: prod.env
    container_name: jenkins
    image: jenkins-master
    build: .
    ports:
      - 8080:8080
      - 50000:50000
    volumes:
      - jenkins:/var/jenkins_home
    secrets:
      - id_rsa

volumes:
  jenkins:

secrets:
  id_rsa:
    file: .\secrets\jenkins_ssh_key
```

**Describe the results you received:**
```
ERROR: The Compose file '.\docker-compose.yml' is invalid because:
Unsupported config option for services.jenkins: 'platform'
```

**Describe the results you expected:**
That platform is supported for a service

**Additional information you deem important (e.g. issue happens only occasionally):**
https://github.com/docker/compose/issues/5953

**Output of `docker version`:**

```
Client:
 Version:      18.05.0-ce-rc1
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   33f00ce
 Built:        Thu Apr 26 00:57:53 2018
 OS/Arch:      windows/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.05.0-ce-rc1
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.10.1
  Git commit:   33f00ce
  Built:        Thu Apr 26 01:06:49 2018
  OS/Arch:      linux/amd64
  Experimental: true
```

**Output of `docker info`:**

```
Containers: 49
 Running: 0
 Paused: 0
 Stopped: 49
Images: 380
Server Version: 18.05.0-ce-rc1
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host ipvlan macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
[...]

PR #1081, opened 2018-05-24
Add composefile schema 3.7 with platform

vdemeester:
It's ignored for now (docker stack deploy) and will print a warning if present.

Fixes https://github.com/docker/cli/issues/1050

> That being said, I think one *easy fix* to that issue would be to add `platform` to v3 schema(s) — but it would start at schema `3.7` — **and** printing a warning when issuing a `docker stack deploy` that this field is not supported (same way we do it for `build`). It would allow to more easily updated from v2 to v3 for users and help us not make v2 and v3 schema diverge that much — at least not having elements in v2 that are not in v3…

Also linked to https://github.com/docker/compose/pull/5985


shin-:
Are we really okay with adding non-Swarm compatible options to v3? I'm asking because it's a significant depart from our strategy so far (with the exception of build sub-options) and I'm curious how it'll affect user experience.

I'm guessing the rationale is that Swarm (/kube?) will support the option eventually?

vdemeester:
> I'm guessing the rationale is that Swarm (/kube?) will support the option eventually?

Yes, that's part of the rationale — other part is I really wish we have as divergence between v2 and v3 composefile evolution 👼

casz:
How rigid do you want to be? Docker compose should perhaps support an experimental flag just like the docker for windows? So that we can get to try out the new stuff. Maybe I am asking for too much, I don't know. 🤔

thaJeztah:
> How rigid do you want to be?

Well, a spec, once published, cannot be changed, so that's why I want to be sure it's finalised and agreed on.

> Docker compose should perhaps support an experimental flag

Having this as an [extension field](https://docs.docker.com/compose/compose-file/#extension-fields) (`x-platform: foo`) would be an option for that

vdemeester:
Closing this as it staled :sweat:

casz:
no worries @vdemeester fixed by the auto-detection, so it's fine :)

PR #1088, opened 2018-05-28
Validate stack-names for empty values

thaJeztah:
Fixes https://github.com/moby/moby/issues/36776

Add validation for stack names to prevent an empty name resulting in _all_
stacks to be returned after filtering, which can result in removal of services
for all stacks if `--prune`, or `docker stack rm` is used.

Before this change;

    docker stack deploy -c docker-compose.yml one
    docker stack deploy -c docker-compose.yml two
    docker stack deploy -c docker-compose.yml three

    docker stack deploy -c docker-compose.yml --prune ''
    Removing service one_web
    Removing service two_web
    Removing service three_web

After this change:

    docker stack deploy -c docker-compose.yml one
    docker stack deploy -c docker-compose.yml two
    docker stack deploy -c docker-compose.yml three

    docker stack deploy -c docker-compose.yml --prune ''
    invalid stack name: ""

Other stack commands were updated as well:

Before this change;

    docker stack deploy -c docker-compose.yml ''
    Creating network _default
    failed to create network _default: Error response from daemon: rpc error: code = InvalidArgument desc = name must be valid as a DNS name component

    docker stack ps ''
    nothing found in stack:

    docker stack rm ''
    Removing service one_web
    Removing service three_web
    Removing service two_web

After this change:

    docker stack deploy -c docker-compose.yml ''
    invalid stack name: ""

    docker stack ps ''
    invalid stack name: ""

    docker stack rm ''
    invalid stack name: ""

```Markdown
* Fix `docker stack deploy --prune` with empty name removing all services [docker/cli#1088](https://github.com/docker/cli/pull/1088)
```

### DIFF
--- a/cli/command/stack/swarm/common.go
+++ b/cli/command/stack/swarm/common.go
@@ -2,6 +2,9 @@ package swarm
 
 import (
 	"context"
+	"fmt"
+	"strings"
+	"unicode"
 
 	"github.com/docker/cli/cli/compose/convert"
 	"github.com/docker/cli/opts"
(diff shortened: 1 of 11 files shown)


cli/command/stack/swarm/common.go
```diff
@@ -48,3 +51,35 @@ func getStackSecrets(ctx context.Context, apiclient client.APIClient, namespace
...
+
+func quotesOrWhitespace(r rune) bool {
+	if unicode.IsSpace(r) {
+		return true
+	}
+	switch r {
+	case '"', '\'':
+		return true
+	}
+	return false
```
mat007: nit: wouldn't this be more readable
```
    return unicode.IsSpace(r) || r == '"' || r == '\''
```
?

mat007: (which might optionally then be inlined into `validateStackName`)

thaJeztah: Good point; I initially had more stuff in the switch, but now it's not really needed anymore. I'll update

thaJeztah:
Updated 👍 

> which might optionally then be inlined into `validateStackName`

I decided not to inline, as I thought having a `quotesOrWhitespace()` was _slightly_ more readable, but let me know if you think I should 👍

PR #1091, opened 2018-05-29
Try to fix `--config` and shell expansion issues

vdemeester:
It makes `--config ~/.docker` and `--config=~/.docker` act the same.
If the specified `configDir` starts with `~`, then we use `homedir`
package (already used for default `configDir`) to get home.

Fixes https://github.com/moby/moby/issues/22526

cc @runcom 

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/cli/cli/config/credentials"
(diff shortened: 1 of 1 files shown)


cli/config/config.go
```diff
@@ -67,6 +68,9 @@ func Load(configDir string) (*configfile.ConfigFile, error) {
 	if configDir == "" {
 		configDir = Dir()
 	}
+	if strings.HasPrefix(configDir, homedir.GetShortcutString()) {
+		configDir = filepath.Join(homedir.Get(), configDir[2:])
```
silvin-lubecki: Not sure `configDir[2:]` is portable... See [Windows version of GetShortcutString](https://github.com/vdemeester/cli/blob/a9ceec72a84759c39fd7ff73cea80752d86e1e5b/vendor/github.com/docker/docker/pkg/homedir/homedir_windows.go#L23). Try with TrimPrefix instead?
```go
configDir = filepath.Join(homedir.Get(), strings.TrimPrefix(configDir, homedir.GetShortcutString()))
```

vdemeester: @silvin-lubecki very good point :+1:

vdemeester:
fair enough :angel: :stuck_out_tongue:

thaJeztah:
I agree it's nice to have, just that it may be a sliding slope 😓

PR #1097, opened 2018-05-31
Allow `x-*` extension on 3rd level objects

vdemeester:
As for top-level key, any 3rd-level key which starts with `x-` will be
ignored by compose. This allows for users to:
* include additional metadata in their compose files
* create YAML anchor objects that can be re-used in other parts of the config

This matches a similar feature in the swagger spec definition:
https://swagger.io/specification/#specificationExtensions

This means a composefile like the following is valid

```
verison: "3.7"
services:
  foo:
    image: foo/bar
    x-foo: bar
network:
  bar:
    x-bar: baz
x-top:
  foo: bar
```

It concerns services, volumes, networks, configs and secrets.


silvin-lubecki:
Does docker-compose need this feature too?

vdemeester:
@silvin-lubecki yes, there is already a PR in `docker/compose` for it https://github.com/docker/compose/pull/5995

thaJeztah:
Okay, let's merge!

PR #1098, opened 2018-05-31
Show long help message when defined

dhiltgen:
This fixes the help template so that if a command
includes a Long form help message that is displayed instead
of ignoring it and always showing the Short message.

For example:
https://github.com/docker/cli/blob/master/cli/command/container/cp.go#L49-L56

Before this fix:
```
% docker container cp --help

Usage:	docker container cp [OPTIONS] CONTAINER:SRC_PATH DEST_PATH|-
	docker cp [OPTIONS] SRC_PATH|- CONTAINER:DEST_PATH

Copy files/folders between a container and the local filesystem

Options:
  -a, --archive       Archive mode (copy all uid/gid information)
  -L, --follow-link   Always follow symbol link in SRC_PATH
```

After this fix
```
% docker container cp --help

Usage:	docker container cp [OPTIONS] CONTAINER:SRC_PATH DEST_PATH|-
	docker cp [OPTIONS] SRC_PATH|- CONTAINER:DEST_PATH

Copy files/folders between a container and the local filesystem

Use '-' as the source to read a tar archive from stdin
and extract it to a directory destination in a container.
Use '-' as the destination to stream a tar archive of a
container source to stdout.

Options:
  -a, --archive               Archive mode (copy all uid/gid information)
  -L, --follow-link           Always follow symbol link in SRC_PATH
```
(The orchestrator flag is a bug recently introduced - unrelated to this PR)

### DIFF
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -104,7 +104,7 @@ var usageTemplate = `Usage:
 {{- if not .HasSubCommands}}	{{.UseLine}}{{end}}
 {{- if .HasSubCommands}}	{{ .CommandPath}} COMMAND{{end}}
 
-{{ .Short | trim }}
+{{if ne .Long ""}}{{ .Long | trim }}{{ else }}{{ .Short | trim }}{{end}}
 
 {{- if gt .Aliases 0}}
 


dhiltgen:
I should note that the `--help` on the parent command still lists the `Short` message as it should
```
docker container --help

Usage:	docker container COMMAND

Manage containers

Options:

Commands:
  attach      Attach local standard input, output, and error streams to a running container
  commit      Create a new image from a container's changes
  cp          Copy files/folders between a container and the local filesystem
  create      Create a new container
...
```

cli/cobra.go
```diff
@@ -104,7 +104,7 @@ var usageTemplate = `Usage:
 {{- if not .HasSubCommands}}	{{.UseLine}}{{end}}
 {{- if .HasSubCommands}}	{{ .CommandPath}} COMMAND{{end}}
 
-{{ .Short | trim }}
+{{if ne .Long ""}}{{ .Long | trim }}{{ else }}{{ .Short | trim }}{{end}}
```
thaJeztah: Looks like this was changed in https://github.com/moby/moby/pull/23825/files#diff-4a2b6f090171fe18f7406cb69fb2e479R143

The original template was;

```
{{with or .Long .Short }}{{. | trim}}{{end}}{{if gt .Aliases 0}}
```

In a later PR, man-page generation was moved to external markdown files; https://github.com/moby/moby/pull/26830, but the template wasn't updated with that change, so I _think_ it's ok to use the long description again (if present)

ping @dnephin to double check

dnephin: Ya, seems fine to use Long now

PR #1100, opened 2018-05-31
Add support for `template_driver` in composefiles

vdemeester:
This maps the `--template-driver` flag on secret and config creation.

See https://github.com/docker/cli/pull/896 (and upstream https://github.com/moby/moby/pull/33702)

- [x] probably need some test
- [x] as for the initial flag, need some documentation too


thaJeztah:
Linting failure;

```
Successfully tagged cli-linter:13932
cli/compose/loader/loader_test.go:1::warning: file is not gofmted with -s (gofmt)
Exited with code 1

File: cli/compose/loader/loader_test.go [hunk 1]
```diff
@@ -1463,3 +1463,64 @@ services:
...
+		},
+		Secrets: map[string]types.SecretConfig{
+			"secret": {
+				Name:           "secret",
+				External:       types.External{External: true},
+				TemplateDriver: "secret-driver",
+			},
+		},
+	}
+	assert.DeepEqual(t, config, expected, cmpopts.EquateEmpty())
```
silvin-lubecki: You can simplify the test:
```go
assert.Assert(t, is.Equal(config.Configs["config"].TemplateDriver, "config-driver"))
assert.Assert(t, is.Equal(config.Secrets["secret"].TemplateDriver, "secret-driver"))
```

cpuguy83:
Is there another driver other than the built-in go one?
What's the reasoning for wanting this anytime soon?

asiplas:
> Good point; don't think there's another driver yet; otoh, isn't it required to set a driver in order to use templating? (will have to check)

It appears that it **is** required as the configs are copied into the container with the template intact (i.e. not substituted—`{{ secret "pw" }}` in final container) when using `docker-compose.yml` to pull in a file like this:
```
secrets:
  pw:
    file: password.txt
…
```
-----

Generating the templated file **first** via `docker secret create --template-driver golang pw ./password.txt` then creating the container with
```
secrets:
  pw:
    external: true
…
```
it is substituted in as expected.

thaJeztah:
To come back to my previous comment; 

> Templating controls whether and how to evaluate the config payload as a template. If no driver is set, no templating is used.

So, yes, we need this option to enable templating, otherwise no templating will be applied.


I see this needs a rebase though (and would now have to be moved to docker-compose schema 3.8

thaJeztah:
Carried in https://github.com/docker/cli/pull/1746

vdemeester:
Thanks @thaJeztah :kissing_cat:

PR #1102, opened 2018-05-31
Fix mapping a range of host ports to a single container port

### LINKED ISSUE #1074: Validation for mapping a range of host-ports to a single container-port fails
When running containers, it is both possible to publish a range of host/container ports using the shorthand `startport-endport` notation:

```bash
docker container run -d -p 8080-8081:80-81 --name multi nginx:alpine
fd9b13ed2a03270445812d9349831f936edc55342a9cc7209fc2ab57b0f37974
```

It's also possible to map multiple host-ports to a single container port: 

```bash
docker container run -d -p 8080-8081:80 --name multi nginx:alpine
8e289ce242e51934156cd6a2cf088bbf7d3ae645a42208feb5447fbb77526e0f
```

Mapping a range of host ports to a range of container ports works for services:

```bash
docker service create --name multi -p 8080-8081:80-81 nginx:alpine

9q6g4m30qqh5vw4ruffgn16px
verify: Service converged 
```

However, trying to map a range of hosts-ports to a single container port fails in validation:

```bash
docker service create --name multi -p 8080-8081:80 nginx:alpine

invalid argument "8080-8090:80" for "-p, --publish" flag: invalid hostport binding (8080-8090) for port (80)
See 'docker service create --help'.
```

The last example should work

Error looks to be generated in this function; https://github.com/docker/cli/blob/99bd7ed693c093b65f98e9f30a47faff9d7c059c/opts/port.go#L143-L167

sfluor:
**- What I did**
fixes #1074 

**- How I did it**
Splitting the host port range in a list of host ports

**- How to verify it**
run `docker service create --name multi -p 8080-8081:80 nginx:alpine`

fix #1074 by splitting host port range into a list of host ports

shiba


opts/port.go
```diff
@@ -148,20 +148,31 @@ func ConvertPortToPortConfig(
 	ports := []swarm.PortConfig{}
 
 	for _, binding := range portBindings[port] {
+		hostPorts := []uint64{}
```
silvin-lubecki: Thank you for your PR!
I think this code can be simplified a little (and avoid the double iteration):
```go
for _, binding := range portBindings[port] {
    if binding.HostIP != "" && binding.HostIP != "0.0.0.0" {
        logrus.Warnf("ignoring IP-address (%s:%s:%s) service will listen on '0.0.0.0'", binding.HostIP, binding.HostPort, port)
    }
    startHostPort, endHostPort, err := nat.ParsePortRange(binding.HostPort)
    if err != nil && binding.HostPort != "" {
        return nil, fmt.Errorf("invalid hostport binding (%s) for port (%s)", binding.HostPort, port.Port())
    }
    for i := startHostPort; i <= endHostPort; i++ {
        ports = append(ports, swarm.PortConfig{
            //TODO Name: ?
            Protocol:      swarm.PortConfigProtocol(strings.ToLower(port.Proto())),
            TargetPort:    uint32(port.Int()),
            PublishedPort: uint32(hostPort),
            PublishMode:   swarm.PortConfigPublishModeIngress,
        })
    }
}
```
Does it sounds good to you?
And is it possible to add some tests on this feature in opts/port_test.go?

sfluor:
Hello 😄  !

I just removed the for loop as you suggested and added test cases for this behavior.

sfluor:
I just removed the redundant test ^^

opts/port.go
```diff
@@ -148,20 +148,26 @@ func ConvertPortToPortConfig(
 	ports := []swarm.PortConfig{}
 
 	for _, binding := range portBindings[port] {
+
```
mat007: nit: extra line?

sfluor: I remove it 👍

thaJeztah:
Just thinking of this (haven't tried), but does this also work for `docker stack deploy`? (if not, we may have to add additional changes)

@sfluor could you squash your commits? I think its ok to have the tests and code-changes in the same commit.

thaJeztah:
Oh, and we should check if we can / want to support this notation for the long syntax; https://github.com/moby/moby/issues/32551

sfluor:
Hello !

I just checked and the change also work for `docker stack deploy`

I did the following:

`docker stack deploy -c docker-compose.yml multi`

With docker-compose.yml being:
```yaml
version: "3.5"
services:
  nginx:
    image: nginx:alpine
    ports: 
    - "8080-8081:80"
```

`docker service ls` then shows:

```
ra6q1sz8s1fq        multi_nginx         replicated          1/1                 nginx:alpine       *:8080-8081->80/tcp
```

And the redirection works as expected when doing a curl.

I also squashed both commits 👍.

I would be glad to look at the long syntax issue.

PR #1106, opened 2018-06-01
Add a "top-level" annotation to hide persistent flags

### LINKED ISSUE #1099: The new `--orchestrator` flag is showing up on too many commands
With the recent refinement of the `docker stack` command to include kubernetes support, it seems we've wired up the `--orchestrator` flag too broadly.  It is now showing up on top-level commands that it should not.

For example:
```
% docker container cp --help

Usage:	docker container cp [OPTIONS] CONTAINER:SRC_PATH DEST_PATH|-
	docker cp [OPTIONS] SRC_PATH|- CONTAINER:DEST_PATH

Copy files/folders between a container and the local filesystem

Use '-' as the source to read a tar archive from stdin
and extract it to a directory destination in a container.
Use '-' as the destination to stream a tar archive of a
container source to stdout.

Options:
  -a, --archive               Archive mode (copy all uid/gid information)
  -L, --follow-link           Always follow symbol link in SRC_PATH
```

and
```
% docker volume --help      

Usage:	docker volume COMMAND

Manage volumes

Options:

Commands:
  create      Create a volume
  ls          List volumes
  prune       Remove all unused local volumes

```

I believe this flag should only show up on the stack command for now.

silvin-lubecki:
**- What I did**
I added a "top-level" annotation to hide persistent flags in all sub-commands, excepting some specific commands, while printing help.
Fixes issue #1099.

**- How to verify it**

```sh
# All commands excepting stack and version hides orchestrator flag
$ docker container cp --help

Usage:  docker container cp [OPTIONS] CONTAINER:SRC_PATH DEST_PATH|-
        docker cp [OPTIONS] SRC_PATH|- CONTAINER:DEST_PATH

Copy files/folders between a container and the local filesystem

Options:
  -a, --archive       Archive mode (copy all uid/gid information)
  -L, --follow-link   Always follow symbol link in SRC_PATH

$ docker version --help
...
...

$ docker stack --help
...
...

# Works also with stack sub-commands
$ docker stack ls --help
...
...
```

* Only show orchestrator flag in root, stack and version commands in help (fixes #1099) 

image

### DIFF
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -54,6 +54,7 @@ func newDockerCommand(dockerCli *command.DockerCli) *cobra.Command {
 	// Install persistent flags
 	persistentFlags := cmd.PersistentFlags()
 	persistentFlags.StringVar(&opts.Common.Orchestrator, "orchestrator", "", "Orchestrator to use (swarm|kubernetes|all)")
+	persistentFlags.SetAnnotation("orchestrator", "top-level", []string{"version", "stack"})
 
 	setFlagErrorFunc(dockerCli, cmd, flags, opts)
 
@@ -245,6 +246,12 @@ func hideUnsupportedFeatures(cmd *cobra.Command, details versionDetails) {
 		if !isOSTypeSupported(f, osType) || !isVersionSupported(f, clientVersion) {
 			f.Hidden = true
 		}
+		// root command shows all top-level flags
+		if cmd.Parent() != nil {
+			if commands, ok := f.Annotations["top-level"]; ok {
+				f.Hidden = !findCommand(cmd, commands)
+			}
+		}
 	})
 
 	for _, subcmd := range cmd.Commands() {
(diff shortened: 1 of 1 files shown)


cmd/docker/docker.go
```diff
@@ -245,6 +246,12 @@ func hideUnsupportedFeatures(cmd *cobra.Command, details versionDetails) {
 		if !isOSTypeSupported(f, osType) || !isVersionSupported(f, clientVersion) {
 			f.Hidden = true
 		}
+		// root command shows all top-level flags
+		if cmd.Parent() != nil {
```
mat007: I take it the default is `false` for `f.Hidden`, therefore `--orchestrator` is displayed by `docker --help` (i.e. when `cmd.Parent()` is `nil`), is this what we want?

thaJeztah: Yes, I think that's the desired behavior, it's _possible_ to set it as global flag (i.e., `docker --orchestrator=foo stack deploy`), but when showing help for (e.g.) `docker stack deploy` we show it as a flag for that command, so: `docker stack deploy --orchestrator=foo`

Is that what you meant?

thaJeztah:
Merging, but @mat007 let me know if there's still something we should address

cmd/docker/docker.go
mat007: Yes, I find it a bit odd that it can be set as a global flag given that it then has no effect and is ignored for most of the sub-commands, however it's out of the scope of this PR anyway.

thaJeztah: Not a big fan of the top level options as well, although they can be handy for making shell aliases (`alias dockerk -> docker --orchestrator=foo --host=bla`)

silvin-lubecki: Yep that's what we want, so "top level" flags are shown in the root command help (`docker --help`), hidden in all sub commands help but still available, excepting the one declared in the annotations.

PR #1111, opened 2018-06-05
Support for experimental BuildKit

tiborvass:
This PR makes `docker build` use the experimental support in moby/moby#37151 when the `DOCKER_BUILDKIT` environment variable is set.

This is still a work in progress.

To test it, build the docker CLI from this PR, and use it with the engine built from moby/moby#37151.
Mainly looking for UX feedback and bad bugs.

The output of the builder is changing (using that of buildkit for interactive mode). Non-interactive output is currently JSON stream for now but it will likely evolve to a more human-readable output (the challenge being handling output of parallel RUN commands).

Known issues:
- no support for `--iidfile` (coming soon)
- no support for docker content trust (out of scope for this PR)

Ping @tonistiigi @thaJeztah @vdemeester @garethr

### DIFF
--- a/cli/command/formatter/disk_usage.go
+++ b/cli/command/formatter/disk_usage.go
@@ -414,7 +414,7 @@ func (c *diskUsageBuilderContext) Size() string {
 }
 
 func (c *diskUsageBuilderContext) Reclaimable() string {
-	inUseBytes := int64(0)
+	var inUseBytes int64
 	for _, bc := range c.buildCache {
 		if bc.InUse {
 			inUseBytes += bc.Size
--- a/cli/command/image/build/context.go
+++ b/cli/command/image/build/context.go
@@ -101,6 +101,7 @@ func DetectArchiveReader(input io.ReadCloser) (rc io.ReadCloser, isArchive bool,
 // name specified by DefaultDockerfileName and returns the path to the
 // temporary directory containing the Dockerfile.
 func WriteTempDockerfile(rc io.ReadCloser) (dockerfileDir string, err error) {
+	// err is a named return value, due to the defer call below.
 	dockerfileDir, err = ioutil.TempDir("", "docker-build-tempdockerfile-")
 	if err != nil {
 		return "", errors.Errorf("unable to create temporary context directory: %v", err)
--- a/cli/command/image/build_buildkit.go
+++ b/cli/command/image/build_buildkit.go
@@ -43,8 +43,6 @@ func runBuildBuildKit(dockerCli command.Cli, options buildOptions) error {
 		return errors.Errorf("buildkit not supported by daemon")
 	}
 
-	buildID := stringid.GenerateRandomID()
-
 	var (
 		remote           string
 		body             io.Reader
@@ -118,13 +116,6 @@ func runBuildBuildKit(dockerCli command.Cli, options buildOptions) error {
 		}))
 	}
 
-	// statusContext, cancelStatus := context.WithCancel(ctx)
-	// defer cancelStatus()
-
-	// if span := opentracing.SpanFromContext(ctx); span != nil {
-	// 	statusContext = opentracing.ContextWithSpan(statusContext, span)
-	// }
-
 	s.Allow(authprovider.NewDockerAuthProvider())
 
 	eg, ctx := errgroup.WithContext(ctx)
@@ -133,6 +124,7 @@ func runBuildBuildKit(dockerCli command.Cli, options buildOptions) error {
 		return s.Run(context.TODO(), dockerCli.Client().DialSession)
 	})
 
+	buildID := stringid.GenerateRandomID()
 	if body != nil {
 		eg.Go(func() error {
 			buildOptions := types.ImageBuildOptions{
@@ -233,20 +225,15 @@ func doBuild(ctx context.Context, eg *errgroup.Group, dockerCli command.Cli, opt
 			if jerr.Code == 0 {
 				jerr.Code = 1
 			}
-			// if options.quiet {
-			// 	fmt.Fprintf(dockerCli.Err(), "%s%s", progBuff, buildBuff)
-			// }
 			return cli.StatusError{Status: jerr.Message, StatusCode: jerr.Code}
 		}
-		return err
 	}
-
-	return nil
+	return err
 }
 
 func resetUIDAndGID(s *fsutil.Stat) bool {
-	s.Uid = uint32(0)
-	s.Gid = uint32(0)
+	s.Uid = 0
+	s.Gid = 0
 	return true
 }
 
@@ -268,6 +255,7 @@ func (t *tracer) write(msg jsonmessage.JSONMessage) {
 	}
 
 	var dt []byte
+	// ignoring all messages that are not understood
 	if err := json.Unmarshal(*msg.Aux, &dt); err != nil {
 		return
 	}


Makefile
```diff
@@ -52,7 +52,7 @@ watch: ## monitor file changes and run go test
 
 vendor: vendor.conf ## check that vendor matches vendor.conf
 	rm -rf vendor
-	bash -c 'vndr |& grep -v -i clone'
+	bash -c 'vndr -verbose'
```
simonferquel: Is this change related to the PR?

cli/command/image/build/context.go
```diff
@@ -81,59 +81,88 @@ func ValidateContextDirectory(srcPath string, excludes []string) error {
...
+// DetectArchiveReader detects whether the input stream is an archive or a
+// Dockerfile and returns a buffered version of input, safe to consume in lieu
+// of input. If an archive is detected, isArchive is set to true, and to false
+// otherwise, in which case it is safe to assume input represents the contents
+// of a Dockerfile.
+//
+// If an archive is detected in the input stream, rc stream is non-nil and ready
+// to be consumed in lieu of the input stream, and dockerfileDir is empty.
+//
+// If a Dockerfile is detected, rc is set to nil and dockerfileDir is set to a
```
simonferquel: Comment does not seem in sync

cli/command/image/build/context.go
```diff
@@ -81,59 +81,88 @@ func ValidateContextDirectory(srcPath string, excludes []string) error {
...
-		return nil, "", errors.New("build context is not an archive")
-	}
 
-	// Input should be read as a Dockerfile.
-	tmpDir, err := ioutil.TempDir("", "docker-build-context-")
+// WriteTempDockerfile writes a Dockerfile stream to a temporary file with a
+// name specified by DefaultDockerfileName and returns the path to the
+// temporary directory containing the Dockerfile.
+func WriteTempDockerfile(rc io.ReadCloser) (dockerfileDir string, err error) {
+	dockerfileDir, err = ioutil.TempDir("", "docker-build-tempdockerfile-")
```
simonferquel: error management in this function seem more complicated than necessary. As the err output parameter is named, why not use a single defer block to centralize error reporting ?

tiborvass:
@simonferquel updated. Appreciate your review!

cli/command/image/build/context.go
```diff
@@ -81,59 +81,81 @@ func ValidateContextDirectory(srcPath string, excludes []string) error {
...
-	f, err := os.Create(filepath.Join(tmpDir, DefaultDockerfileName))
+	f, err := os.Create(filepath.Join(dockerfileDir, DefaultDockerfileName))
 	if err != nil {
-		return nil, "", err
+		return "", err
 	}
-	_, err = io.Copy(f, buf)
+	_, err = io.Copy(f, rc)
 	if err != nil {
 		f.Close()
```
simonferquel:
```
defer func (){
  er := f.Close ()
  if err == nil {
    err = er
  }
}
```
Not sure about it, but I dislike having duplicated Close call. Seem error prone if/when new code paths are introduced in the future.

cli/command/image/build.go
```diff
@@ -149,6 +152,9 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation("stream", "experimental", nil)
 	flags.SetAnnotation("stream", "version", []string{"1.31"})
 
+	flags.BoolVar(&options.noConsole, "no-console", false, "Show non-console output")
```
AkihiroSuda: Maybe better to add "BuildKit mode only"?

cli/command/image/build.go
```diff
@@ -170,6 +176,10 @@ func (out *lastProgressOutput) WriteProgress(prog progress.Progress) error {
 
 // nolint: gocyclo
 func runBuild(dockerCli command.Cli, options buildOptions) error {
+	if os.Getenv("DOCKER_BUILDKIT") != "" {
+		return runBuildBuildKit(dockerCli, options)
```
AkihiroSuda: Any reason not to use CLI flag?

cli/command/image/build.go
tiborvass: Sure. BTW, the main reason we have this flag, is because the interactive output of buildkit doesn't show outputs of RUN commands. If it did, we might not need this flag. Which makes me wonder whether we could remove this flag in the future as it graduates from experimental. cc @tonistiigi @vdemeester

cli/command/image/build.go
tiborvass: @AkihiroSuda because in the future I'd like to remove it.

tiborvass:
Ping @simonferquel if you have any further comments.

simonferquel:
Lgtm :)

cli/command/image/build/context.go
```diff
@@ -81,59 +81,76 @@ func ValidateContextDirectory(srcPath string, excludes []string) error {
...
+	f, err := os.Create(filepath.Join(dockerfileDir, DefaultDockerfileName))
 	if err != nil {
-		return nil, "", errors.Errorf("unable to create temporary context directory: %v", err)
+		return "", err
+	}
+	defer f.Close()
+	if _, err := io.Copy(f, rc); err != nil {
+		return "", err
 	}
+	return dockerfileDir, rc.Close()
```
ijc: If `rc.Close()` returns an error then the callers will leak `tmpDir`, since they don't clean it up in the error case (and it would be a little odd to expect them to do so).

The dir is also leaked on errors in in the `os.Create` above too.

cli/command/image/build_buildkit.go
```diff
@@ -0,0 +1,311 @@
...
+		rc, isArchive, err := build.DetectArchiveReader(os.Stdin)
+		if err != nil {
+			return err
+		}
+		if isArchive {
+			body = rc
+			remote = uploadRequestRemote
+		} else {
+			if options.dockerfileName != "" {
+				// This is new behavior, but @tiborvass believes it's correct.
```
ijc: This looks like a discussion for the commit message rather than a code comment.

cli/command/image/build/context.go
tiborvass: @ijc Addressed.

cli/command/image/build_buildkit.go
tiborvass: @ijc addressed in commit message.

cli/command/image/build_buildkit.go
```diff
@@ -0,0 +1,312 @@
...
+			},
+		}))
+	}
+
+	// statusContext, cancelStatus := context.WithCancel(ctx)
+	// defer cancelStatus()
+
+	// if span := opentracing.SpanFromContext(ctx); span != nil {
+	// 	statusContext = opentracing.ContextWithSpan(statusContext, span)
+	// }
```
thaJeztah: Should this be removed before merging?

cli/command/image/build_buildkit.go
```diff
@@ -0,0 +1,312 @@
...
+
+	displayStatus := func(displayCh chan *client.SolveStatus) {
+		var c console.Console
+		out := os.Stderr
+		// TODO: Handle interactive output in non-interactive environment.
+		consoleOpt := options.console.Value()
+		if cons, err := console.ConsoleFromFile(out); err == nil && (consoleOpt == nil || *consoleOpt) {
+			c = cons
+		}
+		// not using shared context to not disrupt display but let is finish reporting errors
```
thaJeztah: typo: s/let is/let it/

cli/command/image/build_buildkit.go
```diff
@@ -0,0 +1,312 @@
...
+	err = jsonmessage.DisplayJSONMessagesStream(response.Body, os.Stdout, dockerCli.Out().FD(), dockerCli.Out().IsTerminal(), t.write)
+	if err != nil {
+		if jerr, ok := err.(*jsonmessage.JSONError); ok {
+			// If no error code is set, default to 1
+			if jerr.Code == 0 {
+				jerr.Code = 1
+			}
+			// if options.quiet {
+			// 	fmt.Fprintf(dockerCli.Err(), "%s%s", progBuff, buildBuff)
+			// }
```
thaJeztah: should this be removed before merging? (or a TODO added)

cli/command/formatter/disk_usage.go
```diff
@@ -377,17 +396,30 @@ func (c *diskUsageBuilderContext) Type() string {
...
+	return fmt.Sprintf("%d", numActive)
 }
 
 func (c *diskUsageBuilderContext) Size() string {
 	return units.HumanSize(float64(c.builderSize))
 }
 
 func (c *diskUsageBuilderContext) Reclaimable() string {
-	return c.Size()
+	inUseBytes := int64(0)
```
thaJeztah: `var inUseBytes int64` ?

cli/command/image/build/context.go
thaJeztah: nit: consider renaming the `err` output var, to prevent it from being confused with local vars (we ran into a couple of cases where this was overlooked in code changes)

cli/command/image/build_buildkit.go
```diff
@@ -0,0 +1,312 @@
...
+//nolint: gocyclo
+func runBuildBuildKit(dockerCli command.Cli, options buildOptions) error {
+	ctx := appcontext.Context()
+
+	s, err := trySession(dockerCli, options.context)
+	if err != nil {
+		return err
+	}
+	if s == nil {
+		return errors.Errorf("buildkit not supported by daemon")
```
thaJeztah: can we make `trysession` return this error?

cli/command/image/build_buildkit.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	buildID := stringid.GenerateRandomID()
```
thaJeztah: looks like this isn't used until quite some lines below; could you move this closer to where it's used?

cli/command/image/build_buildkit.go
```diff
@@ -0,0 +1,312 @@
...
+			// If no error code is set, default to 1
+			if jerr.Code == 0 {
+				jerr.Code = 1
+			}
+			// if options.quiet {
+			// 	fmt.Fprintf(dockerCli.Err(), "%s%s", progBuff, buildBuff)
+			// }
+			return cli.StatusError{Status: jerr.Message, StatusCode: jerr.Code}
+		}
+		return err
```
thaJeztah: this is redundant if you change the return below to `return err`.

perhaps changing the `if err != nil` to `if err == nil` above (and return early), could make the logic a bit less complicated

cli/command/image/build_buildkit.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	return nil
+}
+
+func resetUIDAndGID(s *fsutil.Stat) bool {
+	s.Uid = uint32(0)
+	s.Gid = uint32(0)
```
thaJeztah: is the type conversion needed here? (just wondering)

cli/command/image/build_buildkit.go
```diff
@@ -0,0 +1,312 @@
...
+	return true
```
thaJeztah: was wondering why the boolean return was here, but guess this is to satisfy an interface? (reading from my phone so couldn't check easily)

cli/command/image/build_buildkit.go
```diff
@@ -0,0 +1,312 @@
...
+
+	if msg.ID != "moby.buildkit.trace" {
+		return
+	}
+
+	var dt []byte
+	if err := json.Unmarshal(*msg.Aux, &dt); err != nil {
+		return
+	}
+	if err := (&resp).Unmarshal(dt); err != nil {
```
thaJeztah: it's ok to hide them actual errors? May be good to document that (as a comment here and/or godoc on the function)

andrewhsu:
Stuff still works with 5a103e1 with my manual build. I also ran a docker-ce integration test suite and stuff is [green](https://jenkins.dockerproject.org/job/docker-ce/372/console).

tiborvass:
@thaJeztah updated

cli/command/image/build/context.go
```diff
@@ -81,59 +81,82 @@ func ValidateContextDirectory(srcPath string, excludes []string) error {
...
+	return ioutils.NewReadCloserWrapper(buf, func() error { return input.Close() }), IsArchive(magic), nil
+}
 
-	if dockerfileName == "-" {
-		return nil, "", errors.New("build context is not an archive")
+// WriteTempDockerfile writes a Dockerfile stream to a temporary file with a
+// name specified by DefaultDockerfileName and returns the path to the
+// temporary directory containing the Dockerfile.
+func WriteTempDockerfile(rc io.ReadCloser) (dockerfileDir string, err error) {
+	// err is a named return value, due to the defer call below.
```
thaJeztah: was thinking naming it `retErr` (eg), but just a nit; no need to change

tonistiigi:
LGTM

PR #1112, opened 2018-06-06
Fix docker trust signer removal

n4ss:
Signed-off-by: Nassim 'Nass' Eddequiouaq <eddequiouaq.nassim@gmail.com>

**- What I did**

Only print `"Successfully removed "` if the signer is removed.

**- How I did it**

Make `trust.removeSingleSigner` return a second value besides the error; whether or not a removal actually happened. 

**- How to verify it**

``` shell
$> docker signer add foobar myrepo/isthebest
$> docker sign myrepo/isthebest:latest
$> docker signer remove foobar myrepo/isthebest
The signer "foobar" signed the last released version of myrepo/isthebest. Removing this signer will make myrepo/isthebest unpullable. Are you sure you want to continue? [y/N]
N

Aborting action.
```

### DIFF
--- a/cli/command/trust/signer_remove.go
+++ b/cli/command/trust/signer_remove.go
@@ -43,11 +43,13 @@ func removeSigner(cli command.Cli, options signerRemoveOptions) error {
 	var errRepos []string
 	for _, repo := range options.repos {
 		fmt.Fprintf(cli.Out(), "Removing signer \"%s\" from %s...\n", options.signer, repo)
-		if err := removeSingleSigner(cli, repo, options.signer, options.forceYes); err != nil {
+		if didRemove, err := removeSingleSigner(cli, repo, options.signer, options.forceYes); err != nil {
 			fmt.Fprintln(cli.Err(), err.Error()+"\n")
 			errRepos = append(errRepos, repo)
 		} else {
-			fmt.Fprintf(cli.Out(), "Successfully removed %s from %s\n\n", options.signer, repo)
+			if didRemove {
+				fmt.Fprintf(cli.Out(), "Successfully removed %s from %s\n\n", options.signer, repo)
+			}
 		}
 	}
 	if len(errRepos) > 0 {
@@ -78,24 +80,24 @@ func isLastSignerForReleases(roleWithSig data.Role, allRoles []client.RoleWithSi
 	return counter < releasesRoleWithSigs.Threshold, nil
 }
 
-func removeSingleSigner(cli command.Cli, repoName, signerName string, forceYes bool) error {
+func removeSingleSigner(cli command.Cli, repoName, signerName string, forceYes bool) (bool, error) {
 	ctx := context.Background()
 	imgRefAndAuth, err := trust.GetImageReferencesAndAuth(ctx, nil, image.AuthResolver(cli), repoName)
 	if err != nil {
-		return err
+		return false, err
 	}
 
 	signerDelegation := data.RoleName("targets/" + signerName)
 	if signerDelegation == releasesRoleTUFName {
-		return fmt.Errorf("releases is a reserved keyword and cannot be removed")
+		return false, fmt.Errorf("releases is a reserved keyword and cannot be removed")
 	}
 	notaryRepo, err := cli.NotaryClient(imgRefAndAuth, trust.ActionsPushAndPull)
 	if err != nil {
-		return trust.NotaryError(imgRefAndAuth.Reference().Name(), err)
+		return false, trust.NotaryError(imgRefAndAuth.Reference().Name(), err)
 	}
 	delegationRoles, err := notaryRepo.GetDelegationRoles()
 	if err != nil {
-		return errors.Wrapf(err, "error retrieving signers for %s", repoName)
+		return false, errors.Wrapf(err, "error retrieving signers for %s", repoName)
 	}
 	var role data.Role
 	for _, delRole := range delegationRoles {
@@ -105,11 +107,11 @@ func removeSingleSigner(cli command.Cli, repoName, signerName string, forceYes b
 		}
 	}
 	if role.Name == "" {
-		return fmt.Errorf("No signer %s for repository %s", signerName, repoName)
+		return false, fmt.Errorf("No signer %s for repository %s", signerName, repoName)
 	}
 	allRoles, err := notaryRepo.ListRoles()
 	if err != nil {
-		return err
+		return false, err
 	}
 	if ok, err := isLastSignerForReleases(role, allRoles); ok && !forceYes {
 		removeSigner := command.PromptForConfirmation(os.Stdin, cli.Out(), fmt.Sprintf("The signer \"%s\" signed the last released version of %s. "+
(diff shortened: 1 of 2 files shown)


cli/command/trust/signer_remove_test.go
```diff
@@ -71,9 +71,9 @@ func TestTrustSignerRemoveErrors(t *testing.T) {
 func TestRemoveSingleSigner(t *testing.T) {
 	cli := test.NewFakeCli(&fakeClient{})
 	cli.SetNotaryClient(notaryfake.GetLoadedNotaryRepository)
-	err := removeSingleSigner(cli, "signed-repo", "test", true)
+	_, err := removeSingleSigner(cli, "signed-repo", "test", true)
```
mat007: Don't we want to check the returned boolean?

cli/command/trust/signer_remove.go
```diff
@@ -43,11 +43,13 @@ func removeSigner(cli command.Cli, options signerRemoveOptions) error {
 	var errRepos []string
 	for _, repo := range options.repos {
 		fmt.Fprintf(cli.Out(), "Removing signer \"%s\" from %s...\n", options.signer, repo)
-		if err := removeSingleSigner(cli, repo, options.signer, options.forceYes); err != nil {
+		if didRemove, err := removeSingleSigner(cli, repo, options.signer, options.forceYes); err != nil {
 			fmt.Fprintln(cli.Err(), err.Error()+"\n")
 			errRepos = append(errRepos, repo)
 		} else {
-			fmt.Fprintf(cli.Out(), "Successfully removed %s from %s\n\n", options.signer, repo)
+			if didRemove {
```
mat007: nit: you can do
```
 		} else if didRemove {
```

cli/command/trust/signer_remove.go
```diff
@@ -120,16 +122,21 @@ func removeSingleSigner(cli command.Cli, repoName, signerName string, forceYes b
...
-		return err
+		return false, err
+	}
+
+	if err = notaryRepo.Publish(); err != nil {
+		return false, err
 	}
-	return notaryRepo.Publish()
+
+	return true, nil
```
mat007: Wouldn't it be simpler to move the logged success message before this return?
```
    fmt.Fprintf(cli.Out(), "Successfully removed %s from %s\n\n", signerName, repoName)
    return nil
```
?

cli/command/trust/signer_remove_test.go
n4ss: will fix.

cli/command/trust/signer_remove.go
n4ss: will fix.

cli/command/trust/signer_remove.go
n4ss: I'd say that I prefer to have it outside as it makes it more maintainable in the case where there are different codepaths that lead to a successful removal in this function and simply have this one return whether or not a removal happened.

cli/command/trust/signer_remove.go
```diff
@@ -43,11 +43,11 @@ func removeSigner(cli command.Cli, options signerRemoveOptions) error {
...
 	for _, repo := range options.repos {
 		fmt.Fprintf(cli.Out(), "Removing signer \"%s\" from %s...\n", options.signer, repo)
-		if err := removeSingleSigner(cli, repo, options.signer, options.forceYes); err != nil {
+		if didRemove, err := removeSingleSigner(cli, repo, options.signer, options.forceYes); err != nil {
 			fmt.Fprintln(cli.Err(), err.Error()+"\n")
 			errRepos = append(errRepos, repo)
-		} else {
-			fmt.Fprintf(cli.Out(), "Successfully removed %s from %s\n\n", options.signer, repo)
+		} else if didRemove {
+				fmt.Fprintf(cli.Out(), "Successfully removed %s from %s\n\n", options.signer, repo)
```
vdemeester: :thinking:

n4ss:
cc @andrewhsu as well

cli/command/trust/signer_remove.go
```diff
@@ -43,10 +43,10 @@ func removeSigner(cli command.Cli, options signerRemoveOptions) error {
 	var errRepos []string
 	for _, repo := range options.repos {
 		fmt.Fprintf(cli.Out(), "Removing signer \"%s\" from %s...\n", options.signer, repo)
-		if err := removeSingleSigner(cli, repo, options.signer, options.forceYes); err != nil {
+		if didRemove, err := removeSingleSigner(cli, repo, options.signer, options.forceYes); err != nil {
```
mat007: nit: `didRemove` -> `removed` maybe?

cli/command/trust/signer_remove.go
mat007: OK but for now there is only one code path which does that, this would minimize the changes required and there is already output logged in a similar way in the function e.g.
```
fmt.Fprintf(cli.Out(), "\nAborting action.\n")
```
Anyway it's just my personal view, obviously :)

cli/command/trust/signer_remove.go
mat007: nit: `didRemove` -> `removed` maybe?

cli/command/trust/signer_remove_test.go
```diff
@@ -71,10 +71,13 @@ func TestTrustSignerRemoveErrors(t *testing.T) {
...
 	cli.SetNotaryClient(notaryfake.GetLoadedNotaryRepository)
-	err := removeSingleSigner(cli, "signed-repo", "test", true)
+	didRemove, err := removeSingleSigner(cli, "signed-repo", "test", true)
 	assert.Error(t, err, "No signer test for repository signed-repo")
-	err = removeSingleSigner(cli, "signed-repo", "releases", true)
+	assert.Equal(t, didRemove, false, "No signer should be removed")
+
+	_, err = removeSingleSigner(cli, "signed-repo", "releases", true)
 	assert.Error(t, err, "releases is a reserved keyword and cannot be removed")
+	assert.Equal(t, didRemove, false, "No signer should be removed")
```
silvin-lubecki: I may miss something, but I think didRemove was not updated on the second call here (as the boolean result is dropped `_, err = removeSingleSigner`)

cli/command/trust/signer_remove.go
n4ss: fixed.

cli/command/trust/signer_remove_test.go
n4ss: good catch thanks, done.

cli/command/trust/signer_remove.go
n4ss: Makes sense! fixed.

cli/command/trust/signer_remove.go
```diff
@@ -43,11 +43,9 @@ func removeSigner(cli command.Cli, options signerRemoveOptions) error {
 	var errRepos []string
 	for _, repo := range options.repos {
 		fmt.Fprintf(cli.Out(), "Removing signer \"%s\" from %s...\n", options.signer, repo)
-		if err := removeSingleSigner(cli, repo, options.signer, options.forceYes); err != nil {
+		if _, err := removeSingleSigner(cli, repo, options.signer, options.forceYes); err != nil {
```
mat007: I would argue that there is no need to return a `bool` as all now that it's dropped here.

cli/command/trust/signer_remove.go
```diff
@@ -78,24 +76,26 @@ func isLastSignerForReleases(roleWithSig data.Role, allRoles []client.RoleWithSi
 	return counter < releasesRoleWithSigs.Threshold, nil
 }
 
-func removeSingleSigner(cli command.Cli, repoName, signerName string, forceYes bool) error {
+// removeSingleSigner returns whether the signer has been removed during this operation and an error
+// Note: the signer not being removed doesn't necessarily raise an error (eg. User saying "No" to the confirmation prompt)
```
mat007: Besides dropping the `bool` I would remove the `Note:` and add punctuation, something like:
```
removeSingleSigner attempts to remove a single signer.
The signer not being removed doesn't necessarily raise an error e.g. user choosing "No" when prompted for confirmation.
```

cli/command/trust/signer_remove.go
n4ss: I will implement more tests that need to know if the signer got actually removed, and I need the information through this boolean.

cli/command/trust/signer_remove.go
n4ss: will fix.

thaJeztah:
Arf; probably should've needed some squashing 😞

PR #1120, opened 2018-06-08
Fix moby/moby #37229 allowing to create networks even there are no services attached to it when deploying stacks

jaswdr:
**- What I did**
Make `docker stack deploy` create networks even if there are no services attached to it, so you can create stack files for networks only
**- How I did it**
Change `cli/compose/convert/compose.go` to consider the `networks` map when `servicesNetworks` lenght is equals 0
**- How to verify it**
The issue [#37229](https://github.com/moby/moby/issues/37229) in moby repo has more information about how to reproduce the error
Support deploy network only stacks

pug

### DIFF
--- a/cli/compose/convert/compose.go
+++ b/cli/compose/convert/compose.go
@@ -51,6 +51,32 @@ func AddStackLabel(namespace Namespace, labels map[string]string) map[string]str
 
 type networkMap map[string]composetypes.NetworkConfig
 
+func networkCreateOptions(namespace Namespace, network composetypes.NetworkConfig) types.NetworkCreate {
+	createOpts := types.NetworkCreate{
+		Labels:     AddStackLabel(namespace, network.Labels),
+		Driver:     network.Driver,
+		Options:    network.DriverOpts,
+		Internal:   network.Internal,
+		Attachable: network.Attachable,
+	}
+
+	if network.Ipam.Driver != "" || len(network.Ipam.Config) > 0 {
+		createOpts.IPAM = &networktypes.IPAM{}
+	}
+
+	if network.Ipam.Driver != "" {
+		createOpts.IPAM.Driver = network.Ipam.Driver
+	}
+	for _, ipamConfig := range network.Ipam.Config {
+		config := networktypes.IPAMConfig{
+			Subnet: ipamConfig.Subnet,
+		}
+		createOpts.IPAM.Config = append(createOpts.IPAM.Config, config)
+	}
+
+	return createOpts
+}
+
 // Networks from the compose-file type to the engine API type
 func Networks(namespace Namespace, networks networkMap, servicesNetworks map[string]struct{}) (map[string]types.NetworkCreate, []string) {
 	if networks == nil {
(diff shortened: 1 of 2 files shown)


cli/compose/convert/compose.go
```diff
@@ -59,40 +85,30 @@ func Networks(namespace Namespace, networks networkMap, servicesNetworks map[str
 
 	externalNetworks := []string{}
 	result := make(map[string]types.NetworkCreate)
+	if len(servicesNetworks) == 0 {
+		for networkName := range networks {
+			network := networks[networkName]
+			if network.External.External {
+				externalNetworks = append(externalNetworks, network.Name)
+				continue
```
silvin-lubecki: nit:
```go
if network.External.External {
    externalNetworks = append(externalNetworks, network.Name)
} else{
    result[networkName] = networkCreateOptions(namespace, network)
}
```

silvin-lubecki:
Thank you @jaswdr for this PR! The design is ok for me, what do you think of this feature @thaJeztah @vdemeester ?

thaJeztah:
Design works for me as well; I don't think we stated that a compose file _must_ contain services, so a file that only has networks (or other objects) should work.

I see the second commit doesn't have a `signed-off-by`, so DCO check is failing @jaswdr could you squash your commits to make the DCO check pass?

thaJeztah:
We should also have a test added for this 👍

jaswdr:
of course @thaJeztah, sorry for this, I will correct this now

jaswdr:
@thaJeztah done

jaswdr:
test created

cli/compose/convert/compose_test.go
```diff
@@ -108,6 +108,77 @@ func TestNetworks(t *testing.T) {
 	assert.DeepEqual(t, []string{"special"}, externals)
 }
 
+func TestNetworksCreatedWithoutServices(t *testing.T) {
+	namespace := Namespace{name: "foo"}
+	serviceNetworks := map[string]struct{}{}
+	source := networkMap{
+		"normal": composetypes.NetworkConfig{
```
silvin-lubecki: I think only one NetworkConfig, whatever it is, is enough to test the feature? The test would be then more compact and less redundant.

jaswdr: did you mean, add the content of the `TestNetworksCreatedWithoutServices` into the `TestNetworks`?

silvin-lubecki: No, I meant remove all network configs except one, like named, in this test 😄

jaswdr:
@silvin-lubecki simplified test :smile:

silvin-lubecki:
Thank you @jaswdr 👍 !
LGTM

vdemeester:
The problem I can think of with this PR is that it doesn't act the same as `docker-compose`.
Using a composefile without services (but network) with `docker-compose` has almost the same behavior as the current `docker stack deploy` : it won't create any network.

That said I would be in favor of this PR so let me ping @shin- @dnephin and @thaJeztah to have their thoughts on that :angel:

jaswdr:
@vdemeester if I make another PR to compose to handle the same behavior has this one, did you agree it is ok?

vdemeester:
@jaswdr yes :angel: I wanted to have @shin- @thaJeztah or @dnephin's opinion before suggesting you that though (so that you don't work on something if we don't reach consensus on whether we wanna do that or not).

thaJeztah:
Yes we should match the behavior in compose as well (also see my earlier comment https://github.com/docker/cli/pull/1120#issuecomment-396404467)

PR #1126, opened 2018-06-15
Add an `Extras` field on the compose config types.

vdemeester:
That field is automaticaly populated with any `x-*` field in the yaml.
And marshalling the compose config struct put them back into place.

This make it possible to get those extra fields without re-inventing
the wheel (i.e. reimplementing 80% of the `cli/compose/*` packages.

Will be very useful for experiments in `docker/app` :angel: 

### DIFF
--- a/cli/compose/loader/loader.go
+++ b/cli/compose/loader/loader.go
@@ -147,6 +147,7 @@ func loadSections(config map[string]interface{}, configDetails types.ConfigDetai
 			return nil, err
 		}
 	}
+	cfg.Extras = getExtras(config)
 	return &cfg, nil
 }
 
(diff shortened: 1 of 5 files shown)


simonferquel:
Should'nt we generalize that on every object (including portsconfig etc.) ?

vdemeester:
@simonferquel in this PR, it's only on object that are allowed to have `x-*` — i.e. top-level, a `service`, a `network`, a `volume`, a `secret` and a `config`. And I really think we don't need more `x-*` field.

cli/compose/loader/loader.go
```diff
@@ -364,9 +365,32 @@ func LoadService(name string, serviceDict map[string]interface{}, workingDir str
 	if err := resolveVolumePaths(serviceConfig.Volumes, workingDir, lookupEnv); err != nil {
 		return nil, err
 	}
+
+	serviceConfig.Extras = getExtras(serviceDict)
+
 	return serviceConfig, nil
 }
 
+func handleExtras(name string, source map[string]interface{}) map[string]interface{} {
```
dnephin: nit: `loadExtras` might be more consistent with existing naming

PR #1127, opened 2018-06-15
Allow `.` in variables used for interpolation

vdemeester:
> Environment variable names used by the utilities in the Shell and
> Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase
> letters, digits, and the '_' (underscore) from the characters
> defined in Portable Character Set and do not begin with a
> digit. **Other characters may be permitted by an implementation;
> applications shall tolerate the presence of such names**. Uppercase
> and lowercase letters shall retain their unique identities and shall
> not be folded together. The name space of environment variable names
> containing lowercase letters is reserved for applications.

As it is possible to use them in environment variable, we should also
allow them during interpolation.

It's also required if we want to allow smarter interpolation in the
future (and for `docker/app` too 👼)

:lion: 

### DIFF
--- a/cli/compose/template/template.go
+++ b/cli/compose/template/template.go
@@ -6,15 +6,17 @@ import (
 	"strings"
 )
 
-var delimiter = "\\$"
-var substitution = "[_a-z.][_a-z0-9.]*(?::?[-?][^}]*)?"
+var (
+	delimiter    = "\\$"
+	substitution = "[_a-z.][_a-z0-9.]*(?::?[-?][^}]*)?"
 
-var patternString = fmt.Sprintf(
-	"%s(?i:(?P<escaped>%s)|(?P<named>%s)|{(?P<braced>%s)}|(?P<invalid>))",
-	delimiter, delimiter, substitution, substitution,
+	patternString = fmt.Sprintf(
+		"%s(?i:(?P<escaped>%s)|(?P<named>%s)|{(?P<braced>%s)}|(?P<invalid>))",
+		delimiter, delimiter, substitution, substitution,
+	)
+
+	pattern = regexp.MustCompile(patternString)
 )
-
-var pattern = regexp.MustCompile(patternString)
 
 // InvalidTemplateError is returned when a variable template is not in a valid
 // format
--- a/cli/compose/template/template.go
+++ b/cli/compose/template/template.go
@@ -7,7 +7,7 @@ import (
 )
 
 var delimiter = "\\$"
-var substitution = "[_a-z][_a-z0-9]*(?::?[-?][^}]*)?"
+var substitution = "[_a-z.][_a-z0-9.]*(?::?[-?][^}]*)?"
 
 var patternString = fmt.Sprintf(
 	"%s(?i:(?P<escaped>%s)|(?P<named>%s)|{(?P<braced>%s)}|(?P<invalid>))",


cli/compose/template/template.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 var delimiter = "\\$"
-var substitution = "[_a-z][_a-z0-9]*(?::?[-?][^}]*)?"
```
silvin-lubecki: It is only said that `do not begin with a digit`, so starting with `.` is valid? 🤔 
So the regexp could be: `[_a-z.][_a-z0-9.]*(?::?[-?][^}]*)?` ?

And by the way a small nit:
```go
var(
    delimiter = 
    substitution = 
    patternString =
    pattern =
)
```

thaJeztah:
SGTM, but want to check if there's additional rules when using substitution (also wondering if we could somehow share code that's used in the builder (thinking of https://github.com/moby/moby/pull/37134)

tiborvass:
@vdemeester sorry if it's obvious, but I don't understand the usecase :S can you give an example of what this would enable?

thaJeztah:
So, on a POSIX shell, this is definitely not working. 

```shell
# export some.odd.var=foo
sh: 1: export: some.odd.var: bad variable name
# echo ${some.odd.var:-bla}
sh: 2: Bad substitution
```

So I'm a bit on the fence for this one; we removed validation for `ENV`, because we're delegating that to the container's process; however, in this case we don't delegate, but handle the variable-name ourselves.

Looking for things where this could (potentially) be troublesome, it looks like currently it won't, but did notice this pattern-matching example in the documentation; http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02_01, looks 

```
# ${parameter%word}
x=file.c
echo ${x%.c}.o
file.o
```

vdemeester:
@tiborvass the main _use case_ I can think of is for `docker/app`, see https://github.com/docker/app/blob/master/examples/voting-app/voting-app.dockerapp/docker-compose.yml#L28 — we have a yaml _struct_ that we use to interpolate.

```yaml
version: "3.2"
services:
  vote:
    image: ${vote.image.name}:${vote.image.tag}
    ports:
      - ${vote.port}:80
    networks:
      - frontend
    depends_on:
      - redis
    deploy:
      replicas: ${vote.replicas}
      update_config:
        parallelism: 2
      restart_policy:
condition: on-failure
```

As @thaJeztah said, I'm a bit on the fence too for that (at least for supporting it in `docker/cli`). I'll try to come up with a better solution, probably an extension of https://github.com/docker/cli/pull/1128

thaJeztah:
Hm, so wonder if we're conflating `env` substitution and Go templating; i.e. should `app` use go-templating (`{ .....}`), not `${.....}`)?

vdemeester:
@tiborvass not really except UX — i.e. the settings file behind a yaml *struct*, it feels normal to use it as such in the variables (i.e. with the `.`).

@thaJeztah we definitely are in a way :stuck_out_tongue_closed_eyes:. But the interpolation in the composefile *happens* to be `env` substitution like (and only support something close to that now). It doesn't mean it has to support only `env` substitution.

By doing that PR, I wish(ed) to make composefile used in `docker/app` also valid in `docker/cli` (with some smart sauce). It's not an hard requirement so I'll think of something else like https://github.com/docker/cli/pull/1128 :wink:.

thaJeztah:
> it feels normal to use it as such in the variables (i.e. with the .).

If I'm not mistaken, docker/distribution uses underscores as separator for overriding options in the config-file, so that could still be an option.

Could be problematic if the property itself already has an underscore...

vdemeester:
I'm going to close this one, as https://github.com/docker/cli/pull/1128/files#diff-0b0330d93ba7beed4c771a2bb1b79ce3R42 will allow advanced usage of the template/interpolation and allow to customize the pattern used.

PR #1128, opened 2018-06-15
Add options to the compose loader

vdemeester:
- Add the possibility to skip interpolation
- Add the possibility to skip schema validation
- Allow customizing the substitution function, to add special cases.

Example of use.

```go
loader.Load(composetypes.ConfigDetails{
	WorkingDir:  ".",
	ConfigFiles: files,
	Environment: env,
}, func(opts *loader.Options) {
	opts.Interpolate.Substitute = composetemplate.SubstituteWith(template, mapping, errorIfMissing)
})

// […]

func errorIfMissing(substitution string, mapping composetemplate.Mapping) (string, bool, error) {
	value, found := mapping(substitution)
	if !found {
		return "", true, &composetemplate.InvalidTemplateError{
			Template: fmt.Sprintf("required variable %s is missing a value", substitution),
		}
	}
	return value, true, nil
}
```

Again, this is gonna be *very useful* for `docker/app` :angel: 

### DIFF
--- a/cli/compose/loader/loader.go
+++ b/cli/compose/loader/loader.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strings"
 
+	interp "github.com/docker/cli/cli/compose/interpolation"
 	"github.com/docker/cli/cli/compose/schema"
 	"github.com/docker/cli/cli/compose/template"
 	"github.com/docker/cli/cli/compose/types"
@@ -21,6 +22,16 @@ import (
 	"github.com/sirupsen/logrus"
 	yaml "gopkg.in/yaml.v2"
 )
+
+// Options supported by Load
+type Options struct {
+	// Skip schema validation
+	SkipValidation bool
+	// Skip interpolation
+	SkipInterpolation bool
+	// Interpolation options
+	Interpolate *interp.Options
+}
 
 // ParseYAML reads the bytes from a file, parses the bytes into a mapping
 // structure, and returns it.
@@ -41,12 +52,25 @@ func ParseYAML(source []byte) (map[string]interface{}, error) {
 }
 
 // Load reads a ConfigDetails and returns a fully loaded configuration
-func Load(configDetails types.ConfigDetails) (*types.Config, error) {
+func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.Config, error) {
 	if len(configDetails.ConfigFiles) < 1 {
 		return nil, errors.Errorf("No files specified")
 	}
 
+	opts := &Options{
+		Interpolate: &interp.Options{
+			Substitute:      template.Substitute,
+			LookupValue:     configDetails.LookupEnv,
+			TypeCastMapping: interpolateTypeCastMapping,
+		},
+	}
+
+	for _, op := range options {
+		op(opts)
+	}
+
 	configs := []*types.Config{}
+	var err error
 
 	for _, file := range configDetails.ConfigFiles {
 		configDict := file.Config
@@ -62,14 +86,17 @@ func Load(configDetails types.ConfigDetails) (*types.Config, error) {
 			return nil, err
 		}
 
-		var err error
-		configDict, err = interpolateConfig(configDict, configDetails.LookupEnv)
-		if err != nil {
-			return nil, err
+		if !opts.SkipInterpolation {
+			configDict, err = interpolateConfig(configDict, *opts.Interpolate)
+			if err != nil {
+				return nil, err
+			}
 		}
 
-		if err := schema.Validate(configDict, configDetails.Version); err != nil {
-			return nil, err
+		if !opts.SkipValidation {
+			if err := schema.Validate(configDict, configDetails.Version); err != nil {
+				return nil, err
+			}
 		}
 
 		cfg, err := loadSections(configDict, configDetails)
--- a/cli/compose/template/template.go
+++ b/cli/compose/template/template.go
@@ -16,6 +16,14 @@ var patternString = fmt.Sprintf(
 
 var pattern = regexp.MustCompile(patternString)
 
+// DefaultSubstituteFuncs contains the default SubstitueFunc used by the docker cli
+var DefaultSubstituteFuncs = []SubstituteFunc{
+	softDefault,
+	hardDefault,
+	requiredNonEmpty,
+	required,
+}
+
 // InvalidTemplateError is returned when a variable template is not in a valid
 // format
 type InvalidTemplateError struct {
@@ -32,8 +40,14 @@ func (e InvalidTemplateError) Error() string {
 // and the absence of a value.
 type Mapping func(string) (string, bool)
 
-// Substitute variables in the string with their values
-func Substitute(template string, mapping Mapping) (string, error) {
+// SubstituteFunc is a user-supplied function that apply substitution.
+// Returns the value as a string, a bool indicating if the function could apply
+// the substitution and an error.
+type SubstituteFunc func(string, Mapping) (string, bool, error)
+
+// SubstituteWith subsitute variables in the string with their values.
+// It accepts additional substitute function.
+func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, subsFuncs ...SubstituteFunc) (string, error) {
 	var err error
 	result := pattern.ReplaceAllStringFunc(template, func(substring string) string {
 		matches := pattern.FindStringSubmatch(substring)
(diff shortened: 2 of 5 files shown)


cli/compose/template/template.go
```diff
@@ -47,49 +59,20 @@ func Substitute(template string, mapping Mapping) (string, error) {
...
-
-		case strings.Contains(substitution, ":?"):
-			name, errorMessage := partition(substitution, ":?")
-			value, ok := mapping(name)
-			if !ok || value == "" {
-				err = &InvalidTemplateError{
-					Template: fmt.Sprintf("required variable %s is missing a value: %s", name, errorMessage),
-				}
+		for _, f := range subsFuncs {
+			var value string
```
silvin-lubecki: nit:
```go
var(
    value string
    applied bool
)
```

cli/compose/template/template.go
```diff
@@ -101,6 +84,65 @@ func Substitute(template string, mapping Mapping) (string, error) {
...
 }
 
+// Substitute variables in the string with their values
+func Substitute(template string, mapping Mapping) (string, error) {
+	return SubstituteWith(template, mapping)
+}
+
+// Soft default (fall back if unset or empty)
+func softDefault(substitution string, mapping Mapping) (string, bool, error) {
+	if strings.Contains(substitution, ":-") {
```
silvin-lubecki: nit:
```go
if !strings.Contains(substitution, ":-"){
    return "", false, nil
}
...
```

cli/compose/template/template.go
```diff
@@ -101,6 +84,65 @@ func Substitute(template string, mapping Mapping) (string, error) {
...
+			return defaultValue, true, nil
+		}
+		return value, true, nil
+	}
+	return "", false, nil
+}
+
+// Hard default (fall back if-and-only-if empty)
+func hardDefault(substitution string, mapping Mapping) (string, bool, error) {
+	if strings.Contains(substitution, "-") {
```
silvin-lubecki: same here

cli/compose/template/template.go
```diff
@@ -101,6 +84,65 @@ func Substitute(template string, mapping Mapping) (string, error) {
...
+		if !ok {
+			return defaultValue, true, nil
+		}
+		return value, true, nil
+	}
+	return "", false, nil
+}
+
+func requiredNonEmpty(substitution string, mapping Mapping) (string, bool, error) {
+	if strings.Contains(substitution, ":?") {
```
silvin-lubecki: and here

cli/compose/template/template.go
```diff
@@ -101,6 +84,65 @@ func Substitute(template string, mapping Mapping) (string, error) {
...
+				Template: fmt.Sprintf("required variable %s is missing a value: %s", name, errorMessage),
+			}
+		}
+		return value, true, nil
+	}
+	return "", false, nil
+}
+
+func required(substitution string, mapping Mapping) (string, bool, error) {
+	if strings.Contains(substitution, "?") {
```
silvin-lubecki: and here

cli/compose/template/template.go
```diff
@@ -32,8 +32,20 @@ func (e InvalidTemplateError) Error() string {
 // and the absence of a value.
 type Mapping func(string) (string, bool)
 
-// Substitute variables in the string with their values
-func Substitute(template string, mapping Mapping) (string, error) {
+// SubstituteFunc is a user-supplied function that apply substition.
```
silvin-lubecki: s/substition/substitution/

cli/compose/template/template.go
```diff
@@ -32,8 +32,20 @@ func (e InvalidTemplateError) Error() string {
...
+// Returns the value as a string, a bool indicating if the function could apply
+// the substition and an error.
```
silvin-lubecki: s/substition/substitution/

cli/compose/loader/loader.go
```diff
@@ -41,11 +52,23 @@ func ParseYAML(source []byte) (map[string]interface{}, error) {
 }
 
 // Load reads a ConfigDetails and returns a fully loaded configuration
-func Load(configDetails types.ConfigDetails) (*types.Config, error) {
+func Load(configDetails types.ConfigDetails, ops ...func(*Options)) (*types.Config, error) {
```
thaJeztah: perhaps `s/ops/options/` but that's a nit

cli/compose/loader/loader.go
```diff
@@ -41,11 +52,23 @@ func ParseYAML(source []byte) (map[string]interface{}, error) {
...
+	opts := &Options{
+		Interpolate: &interp.Options{
+			Substitute:      template.Substitute,
+			LookupValue:     configDetails.LookupEnv,
+			TypeCastMapping: interpolateTypeCastMapping,
+		},
+	}
+
+	for _, op := range ops {
+		op(opts)
```
thaJeztah: This seems to be the standard pattern, but was just thinking why functional options cannot produce an error 🤔

cli/compose/loader/loader.go
```diff
@@ -62,14 +85,18 @@ func Load(configDetails types.ConfigDetails) (*types.Config, error) {
 			return nil, err
 		}
 
-		var err error
-		configDict, err = interpolateConfig(configDict, configDetails.LookupEnv)
-		if err != nil {
-			return nil, err
+		if !opts.SkipInterpolation {
+			var err error
```
thaJeztah: Can you remove this `var err ..` ? There's more `err` variables in this function, so having this here will be confusing.

cli/compose/loader/loader.go
```diff
@@ -62,14 +85,18 @@ func Load(configDetails types.ConfigDetails) (*types.Config, error) {
...
+			configDict, err = interpolateConfig(configDict, *opts.Interpolate)
+			if err != nil {
+				return nil, err
+			}
 		}
 
-		if err := schema.Validate(configDict, configDetails.Version); err != nil {
-			return nil, err
+		if !opts.SkipValidation {
```
thaJeztah: I guess the reason to allow skipping "validation" is that if "interpolation" is skipped, there's values that are invalid (because, e.g. `$port` is not a valid port-number)?

Is there a use-case to disable `validation` or `interpolation` separately?

Also, trying to think of situations where skipping validation could skip important checks (although, I guess in the end that's the daemon/API's responsibility)

cli/compose/template/template.go
```diff
@@ -32,8 +32,20 @@ func (e InvalidTemplateError) Error() string {
...
 
-// Substitute variables in the string with their values
-func Substitute(template string, mapping Mapping) (string, error) {
+// SubstituteFunc is a user-supplied function that apply substitution.
+// Returns the value as a string, a bool indicating if the function could apply
+// the substitution and an error.
+type SubstituteFunc func(string, Mapping) (string, bool, error)
+
+// SubstituteWith subsitute variables in the string with their values.
+// It accepts additionnal substitute function.
```
thaJeztah: `s/additionnal/additional/`

cli/compose/template/template.go
```diff
@@ -32,8 +32,20 @@ func (e InvalidTemplateError) Error() string {
...
+func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, subsFuncs ...SubstituteFunc) (string, error) {
+	subsFuncs = append([]SubstituteFunc{
```
thaJeztah: Should we always _append_ the default functions, or only do so if no custom functions are provided?

cli/compose/template/template.go
```diff
@@ -47,49 +59,22 @@ func Substitute(template string, mapping Mapping) (string, error) {
...
-				err = &InvalidTemplateError{
-					Template: fmt.Sprintf("required variable %s is missing a value: %s", name, errorMessage),
-				}
+		for _, f := range subsFuncs {
+			var (
+				value   string
+				applied bool
+			)
+			value, applied, err = f(substitution, mapping)
+			if err != nil {
```
thaJeztah: Is the error itself discarded?

cli/compose/loader/loader.go
vdemeester: @thaJeztah they could :wink: I just though they would not here :stuck_out_tongue_closed_eyes:

cli/compose/loader/loader.go
vdemeester: If we don't interpolate, for sure we don't want to validate (otherwise the composefile could be invalid). The other way could be *happens*, i.e. doing some interpolation but not validating (for whatever reason, like if it's later on used with something else).

> Also, trying to think of situations where skipping validation could skip important checks (although, I guess in the end that's the daemon/API's responsibility)

Yes. And also, this is *only* in the library part of `cli/compose` — i.e. with this PR it is not possible to not validate or interpolate (and we shouldn't allow that with the `docker` cli) :wink:

cli/compose/template/template.go
vdemeester: Good point. I did that to not *export* the default function, but looking back at it, this function should probably just take the list and use them (letting user decide which one to apply)

cli/compose/template/template.go
vdemeester: @thaJeztah it's not if it's the last one to be applied (i.e. if one function doesn't apply, or apply with error, it doesn't necessarly mean others wont).

PR #1129, opened 2018-06-18
Add `init` support in 3.7 schema

vdemeester:
> Run an init inside the container that forwards signals and reaps
  processes

This is supported on `run` and now on Swarm services too, so it's also
possible to have in on a composefile :).

**- What I did**

**- How I did it**

**- How to verify it**


cli/compose/loader/loader_test.go
```diff
@@ -1395,3 +1395,60 @@ networks:
...
+		},
+	}, cmpopts.EquateEmpty()))
+	config, err = loadYAML(`
+version: '3.7'
+services:
+  foo:
+    image: alpine
+    init: true`)
+	assert.NilError(t, err)
+	assert.Check(t, is.DeepEqual(config, &types.Config{
```
silvin-lubecki: I think you can just check the Init value, checking the equality on all the Config is a little bit extreme here:
```go
assert.Equal(t, isconfig.Services[0].Init, &booleanTrue)
```
You may argue that then you have to check Services length before, but I think we can skip it 😇

silvin-lubecki: Then you can refactor the test with these 3 cases using Table Tests.

PR #1138, opened 2018-06-21
Add env variable option for experimental

seemethere:

**- What I did**
Added an option to enable the experimental cli options with an environment variable

**- How to verify it**

<details>
<summary>Proof of work</summary>

```
~/go/src/github.com/docker/cli env_experimental eli@castlerock
❯ DOCKER_CLI_EXPERIMENTAL=enabled ./build/docker manifest

Usage:  docker manifest COMMAND

Manage Docker image manifests and manifest lists

Options:

Commands:
  annotate    Add additional information to a local image manifest
  create      Create a local manifest list for annotating and pushing to a registry
  inspect     Display an image manifest, or manifest list
  push        Push a manifest list to a repository

Run 'docker manifest COMMAND --help' for more information on a command.

~/go/src/github.com/docker/cli env_experimental eli@castlerock
❯ DOCKER_CLI_EXPERIMENTAL=disabled ./build/docker manifest
docker manifest is only supported on a Docker cli with experimental cli features enabled

~/go/src/github.com/docker/cli env_experimental eli@castlerock
❯ ./build/docker manifest
docker manifest is only supported on a Docker cli with experimental cli features enabled
```

</details>

```
* Added the option to enable experimental cli features through the `DOCKER_CLI_EXPERIMENTAL environment variable
```

pirate

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -162,7 +162,12 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	if err != nil {
 		return err
 	}
-	hasExperimental, err := isEnabled(cli.configFile.Experimental)
+	var experimentalValue string
+	// Environment variable always overrides configuration
+	if experimentalValue = os.Getenv("DOCKER_CLI_EXPERIMENTAL"); experimentalValue == "" {
+		experimentalValue = cli.configFile.Experimental
+	}
+	hasExperimental, err := isEnabled(experimentalValue)
 	if err != nil {
 		return errors.Wrap(err, "Experimental field")
 	}
--- a/docs/reference/commandline/cli.md
+++ b/docs/reference/commandline/cli.md
@@ -61,6 +61,7 @@ by the `docker` command line:
 * `DOCKER_API_VERSION` The API version to use (e.g. `1.19`)
 * `DOCKER_CONFIG` The location of your client configuration files.
 * `DOCKER_CERT_PATH` The location of your authentication keys.
+* `DOCKER_CLI_EXPERIMENTAL` Enable experimental features for the cli (e.g. `enabled` or `disabled`)
 * `DOCKER_DRIVER` The graph driver to use.
 * `DOCKER_HOST` Daemon socket to connect to.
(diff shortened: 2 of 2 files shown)


cli/command/cli.go
```diff
@@ -180,6 +180,10 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 }
 
 func isEnabled(value string) (bool, error) {
+	// Environment variable always overrides configuration
+	if os.Getenv("DOCKER_CLI_EXPERIMENTAL") != "" {
```
thaJeztah: Can you update the documentation, and add this env-var to the list in https://github.com/docker/cli/blob/master/docs/reference/commandline/cli.md#environment-variables ?

silvin-lubecki: So this means that `DOCKER_CLI_EXPERIMENTAL=false docker manifest` will be valid? Maybe we should match the configuration file values instead of just checking the existence of the environment variable?

mat007: This does not seem very consistent at the moment, for instance there is
```
func contentTrustEnabled() bool {
	if e := os.Getenv("DOCKER_CONTENT_TRUST"); e != "" {
		if t, err := strconv.ParseBool(e); t || err != nil {
			// treat any other value as true
			return true
		}
	}
	return false
}
```
but
```
func hide(cmd *cobra.Command) *cobra.Command {
	// If the environment variable with name "DOCKER_HIDE_LEGACY_COMMANDS" is not empty,
	// these legacy commands (such as `docker ps`, `docker exec`, etc)
	// will not be shown in output console.
	if os.Getenv("DOCKER_HIDE_LEGACY_COMMANDS") == "" {
		return cmd
	}
…
```

silvin-lubecki: Oh right...

seemethere: Yeah for it to be uniform with what the value in the config is, it would need to be either `enabled` or `disabled`. Should I change it to that?

seemethere: Here's what the diff would look like for that:

seemethere:
Updated to change to `enabled` or `disabled` per @silvin-lubecki's suggestion, and added a mention in the doc's per @thaJeztah's suggestion.

Issue #1110, opened 2018-06-04
Invalid DOCKER_CONFIG should warn using the current config pattern


silvin-lubecki:
**Description**
If the user specifies an invalid path in hers/his `DOCKER_CONFIG` environment variable, the CLI tries first reading `$DOCKER_CONFIG/config.json`, and then tries with old default config file but only inside `$HOME`  (`~/.dockercfg`), so not using the `DOCKER_CONFIG` path.

**Steps to reproduce the issue:**
1. 
```
$ export DOCKER_CONFIG=unknown
```
2.
```
$ docker ps
WARNING: Error loading config file: /Users/xxxx/.dockercfg: Invalid Auth config file
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
```

**Describe the results you expected:**

It should warn the user with the current config file pattern:
```
WARNING: Error loading config file: unknown/config.json: Invalid Auth config file
```
Bonus: the fallback should try `$DOCKER_CONFIG/.dockercfg` instead of `$HOME/.dockercfg`.

Vimal-Raghubir:
I am interested in fixing this issue. Therefore this search should never involved the `$HOME` variable based on my understanding? Only deal with the `$DOCKER_CONFIG` environment variable?

tonyzhoujingtao:
Has this been fixed?

I cannot reproduce the error
`
$ export DOCKER_CONFIG=unknown
$ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
$ docker --version
Docker version 18.03.1-ce, build 9ee9f40`

Vimal-Raghubir:
Not yet. Once my changes have been merged then it should be.

PR #1140, opened 2018-06-22
Fix: Warning Message and Fallback search

Vimal-Raghubir:
**- What I did**
fixes: #1110 . I removed the usage of the `$HOME` environment variable when the CLI searches for the `.dockercfg` file and changed it to use the `$DOCKER_CONFIG` variable as it should be doing in the first place. 

Next I changed the return statements for the error messages to display the new file path which is represented by the variable `filename`, instead of the old config file path indicated by `confFile`.
**- How I did it**

I scanned the function calls and noticed the usage of `homedir.Get()` for the `$HOME` variable. This was easy to realize that this has to be changed to the `$DOCKER_CONFIG` variable. After that I scanned the error messages that used this old file path and realized confFile renders the old file path. This had to be replaced with filename which represents the new file path.
**- How to verify it**

Verifying this can be done by rerunning the steps to reproduce the initial issue linked earlier. The warning message should be changed as well as the search should be done using the `$DOCKER_CONFIG` variable instead of the `$HOME` variable.

Fix warning message and ensure fallback search uses the DOCKER_CONFIG variable.

pug


cli/config/config.go
```diff
@@ -90,18 +90,18 @@ func Load(configDir string) (*configfile.ConfigFile, error) {
 	}
 
 	// Can't find latest config file so check for the old one
-	confFile := filepath.Join(homedir.Get(), oldConfigfile)
+	confFile := filepath.Join(configDir, oldConfigfile)
```
thaJeztah: Not sure this is actually correct; custom locations for the configuration file were added in https://github.com/moby/moby/pull/12856, but at that time, `.dockercfg` was already the _old_ file, and replaced with `~/.docker/config.json` (that change was made in https://github.com/moby/moby/pull/12009)


If we change this now, we're effectively adding new functionality (don't think setting `--config-dir=/foo` would read `/foo/.dockercfg` in older versions, correct? For a deprecated feature.

I think we should print a deprecation warning if we encounter an `~/.dockercfg` file, and remove the logic to look for this filename in the next release (or the one after); this needs to be added to the deprecated features document; https://github.com/docker/cli/blob/master/docs/deprecated.md

vdemeester: I do agree with @thaJeztah on this. This part of the code is looking for older/deprecated configuration file that was _hardcoded_ in `$HOME/.dockercfg`… So we shouldn't change that (otherwise we're effectively adding a new functionnality).

Vimal-Raghubir:
Thank you @thaJeztah and @vdemeester for your review and feedback! So based on my understanding to your comments, this file path search has to be deprecated and removed eventually? 

The reason why I am asking this is I would love to help with the deprecation and removal of this portion if you wish to explore it anytime soon.

Vimal-Raghubir:
Would I need to run the commands shown above by @GordonTheTurtle then run the squash commands? I small bit of help with this would be much appreciated. Thank you.

thaJeztah:
Whoops, sorry, missed your ping; so to _squash_ the commits, you do an "interactive rebase"; when doing an interactive rebase, you can mark commits to be "squashed" together. Here are some instructions that may help; https://docs.docker.com/v1.12/opensource/workflow/work-issue/#/pull-and-rebase-frequently.

Make sure that the commit after it's been squashed is signed-off as well; if you forgot to do so, you can amend your commit with `docker commit -s --amend`

If all is done, you need to _force push_ your branch (needed, because you did a rebase), which will update the pull request

thaJeztah:
rebased and squashed; @vdemeester PTAL

Vimal-Raghubir:
Hi @thaJeztah,
Any update on this PR? Thank you.

PR #1146, opened 2018-06-26
Extract RunDeploy to be re-used elsewhere 👼

vdemeester:
Mainly useful to re-use in `docker/app` :angel: 

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

PR #1147, opened 2018-06-26
Refactor `stack` command/package

vdemeester:
- Handle `bundlefile` directly in the `top-level`
  command. `bundlefile` is still experimental and will be deprecated
  in future version — this should make be easier to remove it.
- Validate the `stack` name in all cases (i.e. whatever the
  orchestrator is used)
- Load the composefile ahead of choosing the orchestrator. This
  removes some slight duplication.
- Makes `RunDeploy` easier to use from outside packages (like
  `docker/app`) with a preloaded configuration.

There is some *slight* benefits with this too, like `docker stack deploy` on k8s should support multiple files :wink: (which is not, *apparantly*, the case as of now)

This is a remake of #1146 :angel: 


cli/command/stack/deploy.go
```diff
@@ -19,20 +25,28 @@ func newDeployCommand(dockerCli command.Cli, common *commonOptions) *cobra.Comma
...
-				}
-				return kubernetes.RunDeploy(kli, opts)
-			default:
-				return swarm.RunDeploy(dockerCli, opts)
+			case opts.Bundlefile == "" && len(opts.Composefiles) == 0:
+				return errors.Errorf("Please specify either a bundle file (with --bundle-file) or a Compose file (with --compose-file).")
+			case opts.Bundlefile != "" && len(opts.Composefiles) != 0:
+				return errors.Errorf("You cannot specify both a bundle file and a Compose file.")
+			case opts.Bundlefile != "":
+				return swarm.DeployBundle(context.Background(), dockerCli, opts)
```
thaJeztah: we should produce an error if the current orchestrator is kubernetes (as we don't support bundle files on kube, correct?)

cli/command/stack/common.go
```diff
@@ -0,0 +1,32 @@
+package stack
+
+import (
+	"fmt"
+	"strings"
+	"unicode"
+)
+
+// validateStackName checks if the provided string is a valid stack name (namespace).
+//
```
mat007: nit: maybe remove this blank line?

cli/command/stack/common.go
```diff
@@ -0,0 +1,32 @@
...
+// It currently only does a rudimentary check if the string is empty, or consists
+// of only whitespace and quoting characters.
+func validateStackName(namespace string) error {
+	v := strings.TrimFunc(namespace, quotesOrWhitespace)
+	if len(v) == 0 {
```
mat007: nit: isn't `if v == ""` clearer?

cli/command/stack/common.go
vdemeester: copy/paste :stuck_out_tongue_closed_eyes:

cli/command/stack/common.go
vdemeester: yes it is (same, copy/pasta, but I'll update :wink: )

cli/command/stack/deploy.go
```diff
@@ -19,20 +25,32 @@ func newDeployCommand(dockerCli command.Cli, common *commonOptions) *cobra.Comma
...
-				kli, err := kubernetes.WrapCli(dockerCli, kubernetes.NewOptions(cmd.Flags(), common.orchestrator))
-				if err != nil {
-					return err
+			case opts.Bundlefile == "" && len(opts.Composefiles) == 0:
+				return errors.Errorf("Please specify either a bundle file (with --bundle-file) or a Compose file (with --compose-file).")
+			case opts.Bundlefile != "" && len(opts.Composefiles) != 0:
+				return errors.Errorf("You cannot specify both a bundle file and a Compose file.")
+			case opts.Bundlefile != "":
+				if commonOrchestrator != command.OrchestratorSwarm {
+					return errors.Errorf("bundle file are not supported on another orchestrator than swarm.")
```
thaJeztah: s/file/files/

vdemeester: :joy:

Issue #699, opened 2017-11-16
any "pass" command in $PATH is executed when running docker-cli


eipiminus1:
**Description**

Whenever I run any docker command from the terminal, the `pass` command is executed. In my case I had a bash script that opens keepass in a folder I added to my $PATH variable.

**Steps to reproduce the issue:**
I could not yet reproduce the behaviour on a debian or ubuntu server. So far I only noticed this on my machine.
1. Put a bash script named "pass" with `touch /tmp/IWASHERE` in a folder in you $PATH
2. make the file executable
3. run `docker ps` or any similar command
4. find a file `IWASHERE` in `/tmp/`

**Describe the results you received:**

any `pass` command in $PATH is executed

**Describe the results you expected:**

I did not expect this kind of side effect when running docker-cli

**Additional information you deem important (e.g. issue happens only occasionally):**

So far I could only reproduce this issue on my personal machine (archlinux) not on an ubuntu or debian server.

**Output of `docker version`:**

```
Client:
 Version:      17.10.0-ce
 API version:  1.33
 Go version:   go1.9.1
 Git commit:   f4ffd2511c
 Built:        Wed Oct 18 23:08:56 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.10.0-ce
 API version:  1.33 (minimum version 1.12)
 Go version:   go1.9.1
 Git commit:   f4ffd2511c
 Built:        Wed Oct 18 23:09:11 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

**Output of `docker info`:**

```
Containers: 48
 Running: 5
 Paused: 0
 Stopped: 43
Images: 1061
Server Version: 17.10.0-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: false
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 06b9cb35161009dcb7123345749fef02f7cea8e0
runc version: 0351df1c5a66838d0c392b4ac4cf9450de844e2d
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.13.8-1-ARCH
Operating System: Arch Linux
OSType: linux
[...]

PR #1131, opened 2018-06-18
[WIP] Do not call `pass` all the time…

vdemeester:
… even when it's not the active credential store.
The check (and calls) will only be when a command require some
credentials (like `login`, `logout`, `pull`, …)

Depends on https://github.com/docker/docker-credential-helpers/pull/106
Should fix https://github.com/docker/cli/issues/699

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

PR #1151, opened 2018-06-27
Update docker-credential-helpers dependency

vdemeester:
This is mainly for the `pass` helper ; `pass` shouldn't be called
every docker command anymore ;).

Linked to https://github.com/docker/docker-credential-helpers/issues/96

cc @euank

### DIFF
--- a/cli/config/credentials/default_store_linux.go
+++ b/cli/config/credentials/default_store_linux.go
@@ -5,7 +5,8 @@ import (
 )
 
 func defaultCredentialsStore() string {
-	if pass.PassInitialized {
+	passStore := pass.Pass{}
+	if passStore.CheckInitialized() {
 		return "pass"
 	}
 
--- a/vendor.conf
+++ b/vendor.conf
@@ -7,7 +7,7 @@ github.com/cpuguy83/go-md2man v1.0.8
 github.com/davecgh/go-spew 346938d642f2ec3594ed81d874461961cd0faa76
 github.com/docker/distribution 83389a148052d74ac602f5f1d62f86ff2f3c4aa5
 github.com/docker/docker c752b0991e31ba9869ab6a0661af57e9423874fb
-github.com/docker/docker-credential-helpers 3c90bd29a46b943b2a9842987b58fb91a7c1819b
+github.com/docker/docker-credential-helpers 5241b46610f2491efdf9d1c85f1ddf5b02f6d962
 # the docker/go package contains a customized version of canonical/json
 # and is used by Notary. The package is periodically rebased on current Go versions.
 github.com/docker/go d30aec9fd63c35133f8f79c3412ad91a3b08be06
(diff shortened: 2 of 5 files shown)


cli/config/credentials/default_store_linux.go
```diff
@@ -5,7 +5,8 @@ import (
 )
 
 func defaultCredentialsStore() string {
-	if pass.PassInitialized {
+	pass := pass.Pass{}
```
silvin-lubecki: nit:
Too much pass 😵 
And not fan of shadowing a package by a variable.

n4ss: yeah same, besides that, LGTM!

n4ss:
LGTM after a rename to passStore or something similar.

vdemeester:
@n4ss @silvin-lubecki Updated :wink:

thaJeztah:
@n4ss or @vdemeester should we tag a v0.6.1 release? https://github.com/docker/docker-credential-helpers/compare/v0.6.0...master

thaJeztah:
Also; this fixes https://github.com/docker/cli/issues/699, correct?

vdemeester:
@thaJeztah it fixes https://github.com/docker/docker-credential-helpers/issues/96 not #699 as it still execute *any* `pass` command in the `PATH`, just not for **all** docker commands.

euank:
> pass shouldn't be called every docker command anymore ;).

I don't think that's correct and I don't think this actually fixes #699.

Before, `docker-credential-helpers` called pass in init so it was impossible for anyone to use the library without pass being called.

It's now possible to use it without pass running, but I don't think `docker/cli` does.

Specifically, the cli call here will execute pass any time it loads a config file which doesn't contain any auth configuration:

https://github.com/docker/cli/blob/ea65e9043c6ad3480882fd573c96a7cff4e66b8e/cli/config/config.go#L109-L120

This will allow people to avoid the call now by setting their config file in such a way that's skipped, but for a default empty config file it will still execute.

I think either credential checks need to be lazy such that it only initializes that if it's needed, not as soon as it loads the config file, or it needs to be a simpler initialization check there of just `LookPath` on the well-known 'pass' name.

vdemeester:
@euank thanks for the comment :angel: I think you're right, somehow I though I also did sthg like https://github.com/docker/cli/pull/1131 in there :sweat_smile:

euank:
I'd be happy to PR lazy initialization (something like https://github.com/euank/cli/commit/ae945b7ac52a2120960386e4ab7ba8ec908d7fd3) if that seems like a preferable approach... but I don't know if invoking `pass` on `docker pull` when no auth config is present is a reasonable behaviour or not.

Perhaps there should be `IsPresent` in `credential-helpers` and I was wrong to not include that.

Since I've realized lazy initialization will still end up invoking it even on `pull` of public images, it seems to me like either adding `IsPresent` to `credential-helpers` or directly doing `LookPath` in this codebase will be the best bet. Do you have any preference @vdemeester?

vdemeester:
@euank I think you're right, either provide a `IsPresent` and using it in `docker/cli` or even simpler doing a `LookPath` in `docker/cli` seems fine by me
(cc @thaJeztah @n4ss )

euank:
PR filed (#1160) @vdemeester

PR #1152, opened 2018-06-27
Extract StackConverter from the StackClient

vdemeester:
It makes it easier to get the correct stack from a compose config
struct without requiring the client (and thus talking to k8s API)

### DIFF
--- a/cli/command/stack/kubernetes/convert.go
+++ b/cli/command/stack/kubernetes/convert.go
@@ -8,12 +8,71 @@ import (
 	"strings"
 
 	"github.com/docker/cli/cli/compose/loader"
+	"github.com/docker/cli/cli/compose/schema"
 	composeTypes "github.com/docker/cli/cli/compose/types"
 	composetypes "github.com/docker/cli/cli/compose/types"
 	"github.com/docker/cli/kubernetes/compose/v1beta1"
 	"github.com/docker/cli/kubernetes/compose/v1beta2"
+	"github.com/pkg/errors"
+	yaml "gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+// NewStackConverter returns a converter from types.Config (compose) to the specified
+// stack version or error out if the version is not supported or existent.
+func NewStackConverter(version string) (StackConverter, error) {
+	switch version {
+	case "v1beta1":
+		return stackV1Beta1Converter{}, nil
+	case "v1beta2":
+		return stackV1Beta2Converter{}, nil
+	default:
+		return nil, errors.Errorf("stack version %s unsupported", version)
+	}
+}
+
+// StackConverter converts a compose types.Config to a Stack
+type StackConverter interface {
+	FromCompose(stderr io.Writer, name string, cfg *composetypes.Config) (Stack, error)
+}
+
+type stackV1Beta1Converter struct{}
+
+func (s stackV1Beta1Converter) FromCompose(stderr io.Writer, name string, cfg *composetypes.Config) (Stack, error) {
+	cfg.Version = v1beta1.MaxComposeVersion
+	st, err := fromCompose(stderr, name, cfg)
+	if err != nil {
+		return Stack{}, err
+	}
+	res, err := yaml.Marshal(cfg)
+	if err != nil {
+		return Stack{}, err
+	}
+	// reload the result to check that it produced a valid 3.5 compose file
+	resparsedConfig, err := loader.ParseYAML(res)
+	if err != nil {
+		return Stack{}, err
+	}
+	if err = schema.Validate(resparsedConfig, v1beta1.MaxComposeVersion); err != nil {
+		return Stack{}, errors.Wrapf(err, "the compose yaml file is invalid with v%s", v1beta1.MaxComposeVersion)
+	}
+
+	st.ComposeFile = string(res)
+	return st, nil
+}
+
+type stackV1Beta2Converter struct{}
+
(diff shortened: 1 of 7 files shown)


cli/command/stack/kubernetes/convert_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
+)
+
+func TestNewStackConverter(t *testing.T) {
+	_, err := NewStackConverter("v1alpha1")
+	assert.Check(t, is.ErrorContains(err, "stack version v1alpha1 unsupported"))
+
+	_, err = NewStackConverter("v1beta1")
```
silvin-lubecki: nit: check the StackConverter returned is not nil?

cli/command/stack/kubernetes/stackclient.go
```diff
@@ -2,34 +2,28 @@ package kubernetes
...
-	yaml "gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 )
 
 // StackClient talks to a kubernetes compose component.
 type StackClient interface {
-	CreateOrUpdate(s stack) error
+	StackConverter
```
silvin-lubecki: Maybe we can really split the 2 interfaces and the 2 implementations? And manipulate 2 objects, one StackClient on one side, and one StackConverter on the other side? I know it will be a more complicated refactoring, so maybe in a follow-up?

vdemeester: Good point :) I did it that way to not do too much changes but I do agree it would be even better.. Doable in this PR too :wink:

cli/command/stack/kubernetes/convert.go
```diff
@@ -8,13 +8,58 @@ import (
...
+	case "v1beta2":
+		return v1beta2FromCompose(stderr, name, cfg)
+	default:
+		return Stack{}, errors.Errorf("stack version %s unsupported", version)
+	}
+}
+
+func v1beta1FromCompose(stderr io.Writer, name string, cfg *composetypes.Config) (Stack, error) {
+	cfg.Version = v1beta1.MaxComposeVersion
+	st, err := v1beta2FromCompose(stderr, name, cfg)
```
thaJeztah: this is confusing; what's beta2 doing in a beta1 conversion?

cli/command/stack/kubernetes/convert.go
```diff
@@ -8,13 +8,58 @@ import (
...
+	// reload the result to check that it produced a valid 3.5 compose file
+	resparsedConfig, err := loader.ParseYAML(res)
+	if err != nil {
+		return Stack{}, err
+	}
+	if err = schema.Validate(resparsedConfig, v1beta1.MaxComposeVersion); err != nil {
+		return Stack{}, errors.Wrapf(err, "the compose yaml file is invalid with v%s", v1beta1.MaxComposeVersion)
+	}
+
+	st.ComposeFile = string(res)
```
thaJeztah: wondering if there's consequences of not passing the original file content; is it to normalize the YAML? (if so, we may need some comments explaining)

cli/command/stack/kubernetes/convert.go
vdemeester: I just moved code, but mainly, `Stack` struct holds both the composefile (`v1beta1` and the `v1beta2` struct) :sweat_smile:

vdemeester: cc @silvin-lubecki because I kinda agree we shouldn't need to do that :angel:

silvin-lubecki:
Well, LGTM as it is 😉

thaJeztah:
(commenting here, to prevent the comment from being collapsed if code is updated). Discussing https://github.com/docker/cli/pull/1152#discussion_r198779254 with @vdemeester;

What it comes down to is that we try to replicate the swarm (client-side) handling of compose files as much as possible; this means that the actual _version_ (`version: 3.x`) of the compose file is "ignored", and we try to treat the file as "max version supported by server-side component"; https://github.com/docker/cli/blob/bac0d8f353ec2c9980db2b215881f762a282da8d/kubernetes/compose/v1beta1/parsing.go#L3-L4

For example, if a compose-file is used that has `version: 3.7`, but only uses features that were already available in version 3.5 of the compose file schema, it will be sent as a 3.5 compose-file if no errors are encountered).

The flow is as follows;

1. Read/parse the file, using the maximum schema that's supported by the CLI (3.7 currently), and convert it into the internal "compose" struct/representation.
    - The file is _parsed_ using the latest version so that _syntax_ features in the latest compose version can be supported
2. If step 1. was successful; re-encode the internal representation back to a YAML file
3. Re-parse the YAML file as a 3.5 version, and if that succeeds, we can know it's usable as a 3.5 compose file, and no 3.5+ features were used.

Do we need to put some information about that in the description of `FromCompose()`? Or should "transcoding" be a separate utility (taking "target version" as argument)?


Note that we were also discussing a limitation in the swarm handling of docker-compose files; unlike command-line options, the compose-file does not label features as being supported by a specific Docker API version, so even if the conversion to the internal format succeeded, actually _deploying_ may fail if the API version doesn't support a feature. Not really sure what the best approach for that would be; annotate options in the compose-file (likely doesn't make sense); perhaps validate the _internal_ representation, and make a swarm-equivalent of [`warnUnsupportedFeatures()`](https://github.com/docker/cli/blob/8963ab993e0c35d5e09a11cd165caeefbe88f5f8/cli/command/stack/kubernetes/warnings.go#L10-L19) that is "API version aware".  Both options may be too much.

PR #1144, opened 2018-06-25
Store the V2 manifest as an opaque blob

flx42:
Fixes: #1135
Tested-by: Christy Norman <christy@linux.vnet.ibm.com>
Signed-off-by: Felix Abecassis <fabecassis@nvidia.com>

PR #1156, opened 2018-06-29
Fix manifest lists to always use correct size

### LINKED ISSUE #1135: manifest subcommand creates invalid manifest list
Currently creating manifest lists from the manifest subcommand uses an incorrect for the manifest, creating invalid manifests. These manifests fail to be pullable with containerd since containerd validates the size.  This has lead to broken images being pushed to registries.

See conversation from https://github.com/containerd/containerd/issues/2401

dmcgowan:
Stores complete OCI descriptor instead of digest and platform fields. This includes the size which was getting lost by not storing the original manifest bytes.

Attempt to support existing cached files, if the original byte content cannot be retrieved output the filename with the incorrect content. This allows the user to remove the file and re-attempt to create their manifest list.

Note: I think before taking this out of experimental we should just update this tool to either use containerd or the content/image store from containerd.

Fixes #1135
Replaces #1144

ping @flx42


cli/command/manifest/push.go
```diff
@@ -195,7 +199,11 @@ func buildBlobRequestList(imageManifest types.ImageManifest, repoName reference.
 		if err != nil {
 			return nil, err
 		}
-		blobReqs = append(blobReqs, manifestBlob{canonical: canonical, os: imageManifest.Platform.OS})
+		var os string
+		if imageManifest.Descriptor.Platform != nil {
+			os = imageManifest.Descriptor.Platform.OS
```
tonistiigi: nit: this would be better as `requireMount bool`

cli/manifest/types/types.go
```diff
@@ -88,20 +113,11 @@ func (s *SerializableNamed) MarshalJSON() ([]byte, error) {
...
-	OSVersion    string   `json:"os.version,omitempty"`
-	OSFeatures   []string `json:"os.features,omitempty"`
-}
-
-// NewImageFromJSON creates an Image configuration from json.
-func NewImageFromJSON(src []byte) (*Image, error) {
-	img := &Image{}
-	if err := json.Unmarshal(src, img); err != nil {
+// PlatformFromJSON returns an OCI platform from formatted JSON bytes
+func PlatformFromJSON(src []byte) (*ocispec.Platform, error) {
```
tonistiigi: nit: no need for this function if called only once

cli/command/manifest/push.go
```diff
@@ -195,7 +199,11 @@ func buildBlobRequestList(imageManifest types.ImageManifest, repoName reference.
 		if err != nil {
 			return nil, err
 		}
-		blobReqs = append(blobReqs, manifestBlob{canonical: canonical, os: imageManifest.Platform.OS})
+		var os string
+		if imageManifest.Descriptor.Platform != nil {
```
clnperez: since this is called after buildManifestList, this should never be nil

cli/command/manifest/push.go
```diff
@@ -141,7 +142,9 @@ func buildManifestList(manifests []types.ImageManifest, targetRef reference.Name
 
 	descriptors := []manifestlist.ManifestDescriptor{}
 	for _, imageManifest := range manifests {
-		if imageManifest.Platform.Architecture == "" || imageManifest.Platform.OS == "" {
+		if imageManifest.Descriptor.Platform == nil ||
+			imageManifest.Descriptor.Platform.Architecture == "" ||
+			imageManifest.Descriptor.Platform.OS == "" {
```
thaJeztah: Wonder if this validation should be inside `buildManifestDescriptor()`, but I assume we can't move it there because that's also used by `manifest inspect`?

cli/manifest/store/store.go
```diff
@@ -50,8 +54,37 @@ func (s *fsStore) getFromFilename(ref reference.Reference, filename string) (typ
...
+		Digest   digest.Digest
+		Platform *manifestlist.PlatformSpec
+	}
+
+	if err := json.Unmarshal(bytes, &manifestInfo); err != nil {
+		return types.ImageManifest{}, err
+	}
+
+	// Compatibility with image manifests created before
+	// descriptor, newer versions omit Digest and Platform
```
thaJeztah: Compatibility here is only for the manifests created by earlier versions of `docker manifest` ("experimental") correct?

If so, we should probably add a TODO to remove this code at some point

cli/manifest/types/types.go
```diff
@@ -8,15 +8,46 @@ import (
...
+	return &ocispec.Platform{
+		Architecture: ps.Architecture,
+		OS:           ps.OS,
+		OSVersion:    ps.OSVersion,
+		OSFeatures:   ps.OSFeatures,
+		Variant:      ps.Variant,
+	}
+}
+
+// PlatformSpecFromOCI creates a platform spec from OCI platform
```
thaJeztah: Wondering if we somehow can get rid of having both types (given that they're effectively the same)

andrewhsu:
cc @corbin-coleman

PR #1162, opened 2018-06-29
Allow vendoring on compose file JSON schemas

silvin-lubecki:
**- What I did**
Compose JSON schemas are stored in directory, without any go code in it, so it cannot be vendored if the pruning is enabled.

**- How I did it**
I simply added a doc.go file...

**- How to verify it**

1. Vendor docker/cli in another project
2. Add somewhere in your project: 
```go
import _ github.com/docker/cli/cli/compose/schema/data
```
3. Prune the vendoring
=> Schemas are still there!

image

### DIFF
--- a/cli/compose/schema/data/doc.go
+++ b/cli/compose/schema/data/doc.go
@@ -1,0 +1,4 @@
+//
+// +domain=docker.com
+
+package data


cli/compose/schema/data/doc.go
```diff
@@ -0,0 +1,4 @@
+//
+// +domain=docker.com
+
+package data
```
vdemeester: As it is `doc.go` maybe we can add a comment of what's in there and how to use it with `docker/cli/cli/compotse` :stuck_out_tongue:

thaJeztah:
> Compose JSON schemas are stored in directory, without any go code in it, so it cannot be vendored if the pruning is enabled.

Note that the schemas are also in https://github.com/docker/cli/blob/master/cli/compose/schema/bindata.go#L191-L520

silvin-lubecki:
Oh you're right!!! Thank you @thaJeztah, it's far more cleaner!

PR #1176, opened 2018-07-02
build: --iidfile support with buildkit

tiborvass:

To be tested with https://github.com/moby/moby/pull/37368

This changes buildkit's output to Stdout instead of Stderr, to match the old builder behavior.

### DIFF
--- a/cli/command/image/build_buildkit.go
+++ b/cli/command/image/build_buildkit.go
@@ -1,12 +1,15 @@
 package image
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/containerd/console"
 	"github.com/docker/cli/cli"
@@ -41,6 +44,13 @@ func runBuildBuildKit(dockerCli command.Cli, options buildOptions) error {
 	}
 	if s == nil {
 		return errors.Errorf("buildkit not supported by daemon")
+	}
+
+	if options.imageIDFile != "" {
+		// Avoid leaving a stale file if we eventually fail
+		if err := os.Remove(options.imageIDFile); err != nil && !os.IsNotExist(err) {
+			return errors.Wrap(err, "removing image ID file")
+		}
 	}
 
 	var (
@@ -159,6 +169,7 @@ func runBuildBuildKit(dockerCli command.Cli, options buildOptions) error {
 	return eg.Wait()
 }
 
+//nolint: gocyclo
 func doBuild(ctx context.Context, eg *errgroup.Group, dockerCli command.Cli, options buildOptions, buildOptions types.ImageBuildOptions) (finalErr error) {
 	response, err := dockerCli.Client().ImageBuild(context.Background(), nil, buildOptions)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cli/command/image/build_buildkit.go
```diff
@@ -43,6 +46,13 @@ func runBuildBuildKit(dockerCli command.Cli, options buildOptions) error {
 		return errors.Errorf("buildkit not supported by daemon")
 	}
 
+	if options.imageIDFile != "" {
```
cpuguy83: I'm sure I'm missing some context for where this gets run, but this seems odd.
What if the user already has this file from a previous build and doesn't want it removed?

cli/command/image/build_buildkit.go
```diff
@@ -228,6 +262,26 @@ func doBuild(ctx context.Context, eg *errgroup.Group, dockerCli command.Cli, opt
...
+	// TODO: we may want to use Aux messages with ID "moby.image.id" regardless of options.quiet (i.e. don't send HTTP param q=1)
+	// instead of assuming that output is image ID if options.quiet.
+	if options.quiet {
+		imageID = buf.String()
+		fmt.Fprint(dockerCli.Out(), imageID)
+	}
+
+	if options.imageIDFile != "" {
+		if imageID == "" {
+			return errors.Errorf("Server did not provide an image ID. Cannot write %s", options.imageIDFile)
```
cpuguy83: "cannot write image ID file, server did not provide an image ID"

cli/command/image/build_buildkit.go
```diff
@@ -228,6 +262,26 @@ func doBuild(ctx context.Context, eg *errgroup.Group, dockerCli command.Cli, opt
...
+		}
+		imageID = strings.TrimSpace(imageID)
+		if err := ioutil.WriteFile(options.imageIDFile, []byte(imageID), 0666); err != nil {
+			return err
```
cpuguy83: `errors.Wrap(err, "error writing image ID file")

cli/command/image/build_buildkit.go
tiborvass: @cpuguy83 thanks! Should be fixed now!

tiborvass: @cpuguy83 as in you need to use it with https://github.com/moby/moby/pull/37368

cli/command/image/build_buildkit.go
tiborvass: @cpuguy83 this seems to be the behavior for old builder as well. The idea is to say, if we can't remove the file, we can't write to it, so fail early instead of running the whole successful build and then fail at writing the image id file.

cli/command/image/build_buildkit.go
```diff
@@ -210,15 +224,29 @@ func doBuild(ctx context.Context, eg *errgroup.Group, dockerCli command.Cli, opt
 					}
 					close(displayCh)
 				}()
-				displayStatus(displayCh)
+				displayStatus(out, displayCh)
```
tonistiigi: s/out/os.Stderr/

cli/command/image/build_buildkit.go
```diff
@@ -210,15 +224,29 @@ func doBuild(ctx context.Context, eg *errgroup.Group, dockerCli command.Cli, opt
...
 			}
 			return nil
 		})
 	} else {
-		displayStatus(t.displayCh)
+		displayStatus(out, t.displayCh)
```
tonistiigi: s/out/os.Stdout/

tonistiigi:
LGTM

PR #1178, opened 2018-07-03
Propagate the provided external CA certificate to the external CA object in swarm

cyli:
Also, fix some CLI command confusions:
1. If the `--external-ca` flag is provided, require a `--ca-cert` flag as well, otherwise
   the external CA is set but the CA certificate is actually rotated to an internal
   cert
2. If a `--ca-cert` flag is provided, require a `--ca-key` or `--external-ca` flag be
   provided as well, otherwise either the server will say that the request is
   invalid, or if there was previously an external CA corresponding to the cert, it
   will succeed.  While that works, it's better to require the user to explicitly
   set all the parameters of the new desired root CA.

This also changes the `swarm update` function to set the external CA's CACert field,
which while not strictly necessary, makes the CA list more explicit.

As pointed out in https://github.com/docker/swarmkit/issues/2680, the CA cert is not propagated correctly making it impossible to rotate the CA certificate to an external CA using the CLI.

This also fixes some other confusing ways the CLI worked previously.

### DIFF
--- a/cli/command/swarm/ca.go
+++ b/cli/command/swarm/ca.go
@@ -68,6 +68,17 @@ func runCA(dockerCli command.Cli, flags *pflag.FlagSet, opts caOptions) error {
 		return displayTrustRoot(dockerCli.Out(), swarmInspect)
 	}
 
+	if flags.Changed(flagExternalCA) && len(opts.externalCA.Value()) > 0 && !flags.Changed(flagCACert) {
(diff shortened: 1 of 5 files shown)


cyli:
Would be good to get this into 18.06 if we still have time.

thaJeztah:
ping @vdemeester @silvin-lubecki @justincormack PTAL!

PR #1179, opened 2018-07-03
Allow the external CAs to be removed entirely using the CLI

cyli:
Allow setting --external-ca to an empty string, to allow for removing
all external CAs entirely.  This will help for instance if rotating
from a fully external CA to an internal CA (if the CA's cert and
key are already in the swarm for instance).

Signed-off-by: Ying Li <ying.li@docker.com>

cc @BillMills

### DIFF
--- a/cli/command/swarm/opts.go
+++ b/cli/command/swarm/opts.go
@@ -4,6 +4,7 @@ import (
 	"encoding/csv"
 	"encoding/pem"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"strings"
 	"time"
@@ -96,7 +97,9 @@ func (m *ExternalCAOption) Set(value string) error {
 		return err
 	}
 
-	m.values = append(m.values, parsed)
+	if parsed != nil {
+		m.values = append(m.values, parsed)
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


thaJeztah:
Moving this back to design review for a bit; using an empty string to remove the external CA initially sounded dangerous to me (thinking of `--external-ca=$MY_VAR_THAT_WASNT_SET`); on the other hand, this is only on `docker swarm update`, correct? (so not a day-to-day command to be using)

thaJeztah:
Hm, right, so possibly `--external-ca-unset` / `--external-ca-rm` (a bit ugly because we don't have an  `-add` counterpart, but it _is_ more safe).

Open to suggestions 👍 

@vdemeester @silvin-lubecki @chris-crone ^^ happy to hear what you think

vdemeester:
I would definitely prefer `--external-ca-unset` or something like that instead of an empty string :sweat:

kolyshkin:
Perhaps `--no-external-ca` or `--remove-external-ca`?

thaJeztah:
rebased in https://github.com/docker/cli/pull/6979, but not sure we still need it

PR #1166, opened 2018-06-30
Sort plugin names in a natural order

adshmh:
**- What I did**
Sorted the output of `plugin list` command by plugin name. fixes #1164

**- How I did it**
Added sort functionality to `cli/command/plugin/list.go` and unit tests to `cli/command/plugin/list_test.go`

**- How to verify it**
1. Create multiple plugins
2. Run `docker plugin ls`
3. Output is sorted based on plugins' names

**- Description for the changelog**
docker plugin ls output is sorted based on plugin name

**- A picture of a cute animal (not mandatory but encouraged)**

PR #1194, opened 2018-07-06
use sort.Slice for sorting commands' output

adshmh:
**- What I did**
Refactored all commands under `cli/command` to use sort.Slice for sorting rather than declaring custom types (https://github.com/docker/cli/pull/1166#pullrequestreview-134849253)

**- How I did it**
modified all commands using `sort.Sort` under `cli/command` to use `sort.Slice`, and removed the custom types.

**- How to verify it**

commands use sort.Slice for sorting output.


adshmh:
A few other places use custom types for sort (under `cli/compose`). I kept those files out of this PR to limit the number of files modified. I can either submit a separate PR for those or add them to this one if that seems a better option.

PR #1215, opened 2018-07-16
Add bash completion for `service logs --details|--raw`

albers:
I found two missing options in docker service logs' bash completion.


albers:
@thaJeztah that was kinda fast, thanks!

thaJeztah:
I'll have a look at cherry picking as well when I get back to my computer; we may just as well include it if there may be a patch release

Issue #985, opened 2018-04-04
[k8s] docker could error out more clearly when there is no kubernetes configuration file


vdemeester:
When `kubernetes` is use as the `orchestrator` but there is not `$HOME/.kube/config`, the docker cli errors out without too much context.

```
$ docker stack deploy -c docker-compose.yml foo
Failed to load kubernetes configuration file '/home/vincent/.kube/config'
```

We may want to add more context on the error, like why we're looking at that file, etc..

cc @thaJeztah @silvin-lubecki

olljanat:
@vdemeester this one looks to be fixed?

thaJeztah:
Yes, looks like this was fixed through https://github.com/docker/cli/pull/1216

PR #1216, opened 2018-07-16
Update deploy.go

justyntemme:
Added proper error handling for no Kubernetes config file when it is used as the orchestrator.

**- What I did**

Added proper error handling for stack/deploy in the event Kubernetes is used as orchestrator, but no config file is present.

Fix for https://github.com/docker/cli/issues/985

**- How I did it**
Wrapped the err object with clearer error message.

**- How to verify it**
Run `docker stack deploy -c docker-compose.yml foo` with no `$HOME/.kube/config`

Added proper error handling for no Kubernetes config file when used as orchestrator.

cute pic

### DIFF
--- a/cli/command/stack/deploy.go
+++ b/cli/command/stack/deploy.go
@@ -81,7 +81,7 @@ func RunDeploy(dockerCli command.Cli, flags *pflag.FlagSet, config *composetypes
 	case commonOrchestrator.HasKubernetes():
 		kli, err := kubernetes.WrapCli(dockerCli, kubernetes.NewOptions(flags, commonOrchestrator))
 		if err != nil {
-			return err
+			return errors.Wrap(err, "unable to deploy to Kubernetes")
 		}
 		return kubernetes.RunDeploy(kli, opts, config)
 	default:
--- a/cli/command/stack/kubernetes/cli.go
+++ b/cli/command/stack/kubernetes/cli.go
@@ -4,10 +4,12 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os"
 
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/kubernetes"
 	cliv1beta1 "github.com/docker/cli/kubernetes/client/clientset/typed/compose/v1beta1"
+	"github.com/pkg/errors"
 	flag "github.com/spf13/pflag"
 	kubeclient "k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
@@ -58,7 +60,10 @@ func WrapCli(dockerCli command.Cli, opts Options) (*KubeCli, error) {
 	cli.kubeNamespace = opts.Namespace
 	if opts.Namespace == "" {
 		configNamespace, _, err := clientConfig.Namespace()
-		if err != nil {
+		switch {
+		case os.IsNotExist(err), os.IsPermission(err):
+			return nil, errors.Wrap(err, "unable to load configuration file")
+		case err != nil:
 			return nil, err
 		}
 		cli.kubeNamespace = configNamespace


thaJeztah:
Thanks for contributing @justyntemme !


Comparing "master" with this PR;

### Master:

```
DOCKER_STACK_ORCHESTRATOR=kubernetes docker stack deploy -c docker-compose.yml foo
stat /root/.kube/config: no such file or directory
```

### This PR:

```
DOCKER_STACK_ORCHESTRATOR=kubernetes docker stack deploy -c docker-compose.yml foo
A '$HOME/.kube/config' is needed to use Kubernetes as the orchestrator.: stat /root/.kube/config: no such file or directory
```


I think there's some more/other changes to be made to address the issue;


First of all; this change wraps / prepends any error that's returned by `kubernetes.WrapCli()` with the additional error message; we cannot make the assumption that the error is caused by a missing kube-config. I had a quick look at the `WrapCli()` function, and see there's four different conditions that return an error; https://github.com/docker/cli/blob/151990de626fac33763300ee55b702f542abffa7/cli/command/stack/kubernetes/cli.go#L52-L85

The actual error that's shown here comes from the first one; https://github.com/docker/cli/blob/151990de626fac33763300ee55b702f542abffa7/cli/command/stack/kubernetes/cli.go#L60-L63

Possible solutions would be to;

1. Make the prefix more generic (`"Unable to deploy to kubernetes: <error message returned from WrapCli>"`)
2. Move the improved error to `WrapCli()` (because there it's possible to better determin what the cause was of the error)
3. Combination of 1. and 2. 


Second; the error message now mentions `$HOME/.kube/config`, but the path to the configuration file is configurable (for example, through the `$KUBECONFIG` environment variable), so the error-message shouldn't refer to `$HOME`.

Last; we probably cannot really fix that, but `stat <somefile>` is a bit ugly (the example output in https://github.com/docker/cli/issues/985 "Failed to load kubernetes configuration file" looked better: was that the actual output at the time or a suggestion @vdemeester?). That error is likely the raw error generated by Go; `clientConfig.Namespace()` has multiple implementations (we use [`DeferredLoadingClientConfig.Namespace()`](https://github.com/docker/cli/blob/8417e4979261c7459f82b3dbd633d674a119e38e/vendor/k8s.io/client-go/tools/clientcmd/merged_client_builder.go#L131-L164)); but that's part of vendored Kubernetes code, so that would have to be fixed upstream.


I tried to make a quick implementation of some options mentioned;

```patch
diff --git a/cli/command/stack/deploy.go b/cli/command/stack/deploy.go
index 15296d42..d2d16b41 100644
--- a/cli/command/stack/deploy.go
+++ b/cli/command/stack/deploy.go
@@ -81,8 +81,7 @@ func RunDeploy(dockerCli command.Cli, flags *pflag.FlagSet, config *composetypes
        case commonOrchestrator.HasKubernetes():
                kli, err := kubernetes.WrapCli(dockerCli, kubernetes.NewOptions(flags, commonOrchestrator))
                if err != nil {
-
-                       return errors.Wrap(err, "A '$HOME/.kube/config' is needed to use Kubernetes as the orchestrator.")
+                       return errors.Wrap(err, "unable to deploy to Kubernetes")
                }
                return kubernetes.RunDeploy(kli, opts, config)
        default:
diff --git a/cli/command/stack/kubernetes/cli.go b/cli/command/stack/kubernetes/cli.go
index bcd0b01e..f69aaada 100644
--- a/cli/command/stack/kubernetes/cli.go
+++ b/cli/command/stack/kubernetes/cli.go
@@ -4,10 +4,12 @@ import (
        "fmt"
        "net"
        "net/url"
+       "os"
 
        "github.com/docker/cli/kubernetes"
        cliv1beta1 "github.com/docker/cli/kubernetes/client/clientset/typed/compose/v1beta1"
+       "github.com/pkg/errors"
        flag "github.com/spf13/pflag"
        kubeclient "k8s.io/client-go/kubernetes"
        restclient "k8s.io/client-go/rest"
@@ -58,7 +60,10 @@ func WrapCli(dockerCli command.Cli, opts Options) (*KubeCli, error) {
        cli.kubeNamespace = opts.Namespace
        if opts.Namespace == "" {
                configNamespace, _, err := clientConfig.Namespace()
-               if err != nil {
+               switch {
+               case os.IsNotExist(err), os.IsPermission(err):
+                       return nil, errors.Wrap(err, "unable to load Kubernetes configuration")
+               case err != nil:
                        return nil, err
                }
                cli.kubeNamespace = configNamespace

```

With those changes applied, it looks like this:

```
DOCKER_STACK_ORCHESTRATOR=kubernetes docker stack deploy -c docker-compose.yml foo
unable to deploy to Kubernetes: unable to load Kubernetes configuration: stat /root/.kube/config: no such file or directory
```

(Perhaps we should not repeat `Kubernetes` twice, and just change to `unable to load configuration file`)



@justyntemme @vdemeester WDYT?

justyntemme:
I like the suggestion to move the error to the specific location that causes it. Honestly, both error messages could be improved. If we are in agreement then I will upload a new commit with the proposed changes. Thank you for taking the time to explain that.

cli/command/stack/kubernetes/cli.go
```diff
@@ -58,7 +60,10 @@ func WrapCli(dockerCli command.Cli, opts Options) (*KubeCli, error) {
 	cli.kubeNamespace = opts.Namespace
 	if opts.Namespace == "" {
 		configNamespace, _, err := clientConfig.Namespace()
-		if err != nil {
+		switch {
+		case os.IsNotExist(err), os.IsPermission(err):
+			return nil, errors.Wrap(err, "unable to load Kubernetes configuration")
```
vdemeester: As @thaJeztah suggested, I think we can just say `unable to load configuration`, kubernetes is already present elsewhere :stuck_out_tongue:

cli/command/stack/deploy.go
```diff
@@ -81,7 +81,8 @@ func RunDeploy(dockerCli command.Cli, flags *pflag.FlagSet, config *composetypes
 	case commonOrchestrator.HasKubernetes():
 		kli, err := kubernetes.WrapCli(dockerCli, kubernetes.NewOptions(flags, commonOrchestrator))
 		if err != nil {
-			return err
+
```
thaJeztah: Oh, erm, can I be nit-picky? Could you remove this empty line?

justyntemme: Sure can! Will do that now.

PR #1230, opened 2018-07-25
Update k8s and other deps

vdemeester:
- Bump kubernetes dependencies to 1.11
- Bump some dependencies to more recent versions (and tagged if available)


vendor.conf
```diff
@@ -85,12 +76,12 @@ google.golang.org/genproto 694d95ba50e67b2e363f3483057db5d4910c18f9
...
+k8s.io/api kubernetes-1.11.0
+k8s.io/apimachinery kubernetes-1.11.0
+k8s.io/client-go kubernetes-1.11.0
+k8s.io/kube-openapi d8ea2fe547a448256204cfc68dfee7b26c720acb
+k8s.io/kubernetes v1.11.0
 vbom.ml/util 928aaa586d7718c70f4090ddf83f2b34c16fdc8d
 github.com/containerd/console cb7008ab3d8359b78c5f464cb7cf160107ad5925
 github.com/tonistiigi/units 29de085e9400559bd68aea2e7bc21566e7b8281d
-github.com/google/shlex 6f45313302b9c56850fc17f99e40caebce98c716
+github.com/google/shlex 6f45313302b9c56850fc17f99e40caebce98c716
```
thaJeztah: missing newline 😇

vendor.conf
```diff
@@ -1,11 +1,11 @@
-github.com/agl/ed25519 d2b94fd789ea21d12fac1a4443dd3a3f79cda72c
+github.com/agl/ed25519 5312a61534124124185d41f09206b9fef1d88403
 github.com/Azure/go-ansiterm d6e3b3328b783f23731bc4d058875b0371ff8109
-github.com/beorn7/perks 4c0e84591b9aa9e6dcfdf3e020114cd81f89d5f9
+github.com/beorn7/perks 3a771d992973f24aa725d07868b467d1ddfceafb
 github.com/containerd/containerd 08f7ee9828af1783dc98cc5cc1739e915697c667
 github.com/containerd/continuity d8fb8589b0e8e85b8c8bbaa8840226d0dfeb7371
-github.com/coreos/etcd v3.2.1
+github.com/coreos/etcd v3.3.8
```
thaJeztah: 3.3.9 was released

Also; I've been punting on this one (SwarmKit is also on a very old version, but I _think_ there were some complications there updating, but should check)

vendor.conf
```diff
@@ -15,32 +15,32 @@ github.com/docker/go d30aec9fd63c35133f8f79c3412ad91a3b08be06
...
-github.com/googleapis/gnostic e4f56557df6250e1945ee6854f181ce4e1c2c646
-github.com/gorilla/context v1.1
-github.com/gorilla/mux v1.1
+github.com/google/btree e89373fe6b4a7413d7acd6da1725b83ef713e6e4
+github.com/google/go-cmp v0.2.0
+github.com/google/gofuzz 24818f796faf91cd76ec7bddd72458fbced7a6c1
+github.com/googleapis/gnostic 7c663266750e7d82587642f65e60bc4083f1f84e # v0.2.0
+github.com/gorilla/context v1.1.1
+github.com/gorilla/mux v1.6.2
 gotest.tools v2.1.0
```
thaJeztah: we should make this list alphabetical again 😅

vendor.conf
vdemeester: oh :thinking:

vendor.conf
vdemeester: :sweat_smile:

vendor.conf
vdemeester: yeah, I think I'm gonna do a commit that reorders them :wink:

vendor.conf
```diff
@@ -1,11 +1,12 @@
...
-github.com/beorn7/perks 4c0e84591b9aa9e6dcfdf3e020114cd81f89d5f9
+github.com/beorn7/perks 3a771d992973f24aa725d07868b467d1ddfceafb
+github.com/containerd/console cb7008ab3d8359b78c5f464cb7cf160107ad5925
 github.com/containerd/containerd 08f7ee9828af1783dc98cc5cc1739e915697c667
 github.com/containerd/continuity d8fb8589b0e8e85b8c8bbaa8840226d0dfeb7371
-github.com/coreos/etcd v3.2.1
+github.com/coreos/etcd v3.3.9
 github.com/cpuguy83/go-md2man v1.0.8
-github.com/davecgh/go-spew 346938d642f2ec3594ed81d874461961cd0faa76
+github.com/davecgh/go-spew 346938d642f2ec3594ed81d874461961cd0faa76 # v1.1.0
```
silvin-lubecki: Why not using the tag directly?

thaJeztah: Ah! Yes, there's been many discussions about those. Ideally, there'd be a way to specify _both_ tag _and_ commit/sha;

reason to use a tag:

- it's human-readable
- it's a release done by the upstream dependency

reason to _not_ use a tag:

- we've had upstreams re-tagging (discovered when `vendor check` failed and no changes were made in the vendor.conf)
- we've had cases where it turned out not to be a _tag_ but a _branch_ named `v1.0` (but we _thought_ it was a tag), then found out about that (again) when `vendor check` failed.
- a commit-sha is unique, and can never change


So... how to get the best of both worlds?

- use sha
- leave comment what version it should match with?

It's tricky; because we also have to verify that the commit matches the tag 😞 possibly the `git describe` notation would be an option, e.g.;

`v0.2.0-7-g1817cd4`

- 7 commits since `v0.2.0`
- sha is `1817cd4`

not sure if that actually _verifies_ if they match though? 🤔

silvin-lubecki: Oh right, lot's of issues I wasn't aware of 😓 Then that's ok for me 😸

vendor.conf
```diff
@@ -15,82 +16,72 @@ github.com/docker/go d30aec9fd63c35133f8f79c3412ad91a3b08be06
...
-github.com/google/shlex 6f45313302b9c56850fc17f99e40caebce98c716
+gopkg.in/inf.v0 d2d2541c53f18d2a059457998ce2876cc8e67cbf # v0.9.1
+gopkg.in/yaml.v2 5420a8b6744d3b0345ab293f6fcba19c978f1183 # v2.2.1
+gotest.tools v2.1.0
+k8s.io/api kubernetes-1.11.0
+k8s.io/apimachinery kubernetes-1.11.0
+k8s.io/client-go kubernetes-1.11.0
+k8s.io/kube-openapi d8ea2fe547a448256204cfc68dfee7b26c720acb
+k8s.io/kubernetes v1.11.0
+vbom.ml/util 256737ac55c46798123f754ab7d2c784e2c71783
```
chris-crone: Missing a `\n`

thaJeztah: Was fixed

PR #1225, opened 2018-07-19
Use warnings provided by daemon

thaJeztah:
CLI changes for https://github.com/moby/moby/pull/37502

Warnings are now generated by the daemon, and returned as part of the `/info` API response.

If warnings are returned by the daemon; use those instead of generating them locally.

Engine bump: https://github.com/moby/moby/compare/1800883bd16664846db1572b8c8fbe8c85892cee...2629fe93266e82751af4f1c7568e21060f065b73

Relevant changes:
- https://github.com/moby/moby/pull/37502 Add "Warnings" to /info endpoint, and move detection to the daemon

PR #1233, opened 2018-07-25
Global Default Address Pool feature support

selansen:
This feature brings new attribute/option for swarm init command.
default-addr-pool will take string input which can be in below format.
"CIDR,CIDR,CIDR...:SUBNET-SIZE".

Signed-off-by: selansen <elango.siva@docker.com>

**- What I did**

**- How I did it**

**- How to verify it**


cpuguy83:
Do you have links to upstream PRs for this?

cli/command/swarm/init.go
```diff
@@ -44,6 +45,7 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
 	flags.BoolVar(&opts.forceNewCluster, "force-new-cluster", false, "Force create a new cluster from current state")
 	flags.BoolVar(&opts.autolock, flagAutolock, false, "Enable manager autolocking (requiring an unlock key to start a stopped manager)")
 	flags.StringVar(&opts.availability, flagAvailability, "active", `Availability of the node ("active"|"pause"|"drain")`)
+	flags.StringVar(&opts.defaultAddrPool, flagDefaultAddrPool, "", "List of default subnet addresses followed by subnet size (format: <subnet,subnet,..:subnet-size)")
```
thaJeztah: This flag likely needs an API version annotation, so that it's hidden on daemons not having this feature

cli/command/swarm/init.go
```diff
@@ -56,6 +58,7 @@ func runInit(dockerCli command.Cli, flags *pflag.FlagSet, opts initOptions) erro
 		ListenAddr:       opts.listenAddr.String(),
 		AdvertiseAddr:    opts.advertiseAddr,
 		DataPathAddr:     opts.dataPathAddr,
+		DefaultAddrPool:  opts.defaultAddrPool,
```
thaJeztah: Haven't seen the related dockerd PR yet, but from an API perspective, we should probably parse the flag and send the information more structured over the wire (i.e., not use the string format)

cli/command/swarm/init.go
thaJeztah: A bit unsure about the format used for this option as well (combination of comma-separated and colon-separated options); also wondering; must all address pools have the same subnet size? I can foresee situations where sizes can differ for each (could use input on that one)

selansen: We(our networking team) had gone through multiple rounds of discussion as part of design and decided to support single subnet size for all subnets.  Main reason is , currently there is no CLI option to explicitly specify  what kind of subnet (ex /24 or /20) user want when they create network.

selansen:
@cpuguy83 , I have just published Libnetwork, swarmkit and CLI PRs.  Once CI passed for all, then I will create PR for upstream linking all of of the PR.

cli/command/swarm/init.go
selansen: Validation is taken care in DockerD/swarmkit.

cli/command/swarm/init.go
selansen: @thaJeztah I didn't understand how I should address below comment. I will sync with you offline
"This flag likely needs an API version annotation, so that it's hidden on daemons not having this feature"

thaJeztah: for the annotation, see for example https://github.com/docker/cli/blob/48fbb12b7ca2b6c1536baa2a3cbedfb4ebf8eb53/cli/command/stack/deploy.go#L62

the API version in that annotation will enable the flag if the daemon supports that version of the API; if not (if the daemon is older), the flag is disabled and hidden

thaJeztah: For the format used to specify the address pools, I was thinking to use the same format as is used for the daemon option; 
https://github.com/moby/moby/pull/36396

The API can use the same format as is used in the daemon.json;

```json
{
  DefaultAddrPools:[{"base":"172.80.0.0/16","size":24},{"base":"172.90.0.0/16","size":24}]
}
```

docs/reference/commandline/swarm_init.md
```diff
@@ -26,6 +26,7 @@ Options:
       --availability string             Availability of the node ("active"|"pause"|"drain") (default "active")
       --cert-expiry duration            Validity period for node certificates (ns|us|ms|s|m|h) (default 2160h0m0s)
       --data-path-addr string           Address or interface to use for data path traffic (format: <ip|interface>)
+      --default-addr-pool string	List of default subnet addresses followed by subnet size (format: <subnet,subnet,..:subnet-size)
```
thaJeztah: Should this option also be on `swarm update`? (i.e., is it possible to change default address pools without having to tear down the cluster?)

cli/command/swarm/init.go
selansen: That was initial design. But after detailed discussion on supporting different subnet size for each subnet and going through the complexity of handling them in libnetwork, we decided to support only one subnet size. Hence removed previous format and made it simple.  If the user doesn't specify subnet size, default value 24 will be taken.  Its not mandatory field.

docs/reference/commandline/swarm_init.md
selansen: that is P1.  once I finish all P0 tasks, I will start working on P1.

cli/command/swarm/init.go
thaJeztah: So the daemon option we have for local subnets no longer works?

And design is all fixed, no review here, just press merge?

contrib/completion/bash/docker
```diff
@@ -3668,6 +3668,10 @@ _docker_swarm_init() {
 			__docker_complete_local_interfaces
 			return
 			;;
+		--default-addr-pool)
+			COMPREPLY=( $( compgen -W "list of subnet addresses followed by subnet size" -- "$cur" ) )
+			return
+			;;
```
albers: This is not correct. The argument to `-W` is the list of possible completions.
Furthermore, I cannot think of a useful completion, so it's best not to complete anything here, allowing any string to be entered.

You can acchieve this by removing this block and adding `--default-addr-pool` to the option list in https://github.com/selansen/cli/blob/14dc79f6d3604f474efc500f665574aaac9d3808/contrib/completion/bash/docker#L3664.

cli/command/swarm/init.go
selansen: local scope still works(Daemon option).  This is for global scope and work (swarm init option). So its kind of two different feature now.

We have design document which I can share it with you. Pls let me know.

selansen: I thought I responded to this question. May be in libnetwork PR.   Local scope configuration works in Daemon config and global scope configuration works in swarm Init . Now its kind of two separate feature.

selansen: @thaJeztah added API version info

selansen:
@albers addressed your comment.

contrib/completion/bash/docker
```diff
@@ -3661,7 +3661,7 @@ _docker_swarm_init() {
 			COMPREPLY=( $( compgen -W "active drain pause" -- "$cur" ) )
 			return
 			;;
-		--cert-expiry|--dispatcher-heartbeat|--external-ca|--max-snapshots|--snapshot-interval|--task-history-limit)
+		--cert-expiry|--dispatcher-heartbeat|--external-ca|--max-snapshots|--snapshot-interval|--task-history-limit | --default-addr-pool )
```
albers: Please add in alphabetical order.

contrib/completion/bash/docker
```diff
@@ -3661,7 +3661,7 @@ _docker_swarm_init() {
 			COMPREPLY=( $( compgen -W "active drain pause" -- "$cur" ) )
 			return
 			;;
-		--cert-expiry|--dispatcher-heartbeat|--external-ca|--max-snapshots|--snapshot-interval|--task-history-limit)
+		--cert-expiry | --default-addr-pool |--dispatcher-heartbeat|--external-ca|--max-snapshots|--snapshot-interval|--task-history-limit )
```
albers: Please remove the whitespace around the `|`.

selansen:
@albers @thaJeztah Pls take a look at it.  I addressed all review comments. I rewrote code such a way that we now pass IPNet struct and subnet size instead of input string as it is.

cli/command/swarm/init.go
```diff
@@ -44,21 +47,84 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
 }
 
+// getDefaultAddrPool extracts info from address pool string and fills PoolsOpt struct
+func getDefaultAddrPool(defaultAddrPool string) ([]*net.IPNet, int, error) {
+	var size int
+	var err error
+	var defaultPoolList []*net.IPNet
+	if defaultAddrPool == "" {
+		// defaultAddrPool is not defined
+		return nil, size, nil
```
ctelfer: very minor nit:  return 0 explicitly instead of size perhaps to make it clear what it's supposed to be?

cli/command/swarm/init.go
```diff
@@ -44,21 +47,84 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
+	}
+
+	result := strings.Split(defaultAddrPool, ":")
+	if len(result) > 2 {
+		return nil, size, fmt.Errorf("default address pool must be a non-zero IP address or addresses format CIDR,CIDR,...:SUBNET-SIZE")
+	}
```
ctelfer: Suggest that the error should say what went wrong.  Like:
`Invalid default address pool format.  Expected CIDR[,CIDR]*:MASKLEN`.

cli/command/swarm/init.go
```diff
@@ -44,21 +47,84 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
+
+	for i := range subnetlist {
+		// trim leading and trailing white spaces
+		subnetlist[i] = strings.TrimSpace(subnetlist[i])
+		_, b, err := net.ParseCIDR(subnetlist[i])
+		if err != nil {
+			return nil, size, fmt.Errorf("invalid base pool %q: %v", subnetlist[i], err)
+		}
+		ones, _ := b.Mask.Size()
+		if size <= 0 || size < ones {
```
ctelfer: Nit: this `size <= 0` test should probably go up right after one parses it from the parameter.  No need to recheck every time and having it here kind of gives the wrong impression. (i.e. that it could be changing)

cli/command/swarm/init.go
```diff
@@ -44,21 +47,84 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
+			return nil, size, fmt.Errorf("invalid pool size: %d", size)
+		}
```
ctelfer: Could then change the error to:  "subnet size too small for pool" or something like that.

cli/command/swarm/init.go
```diff
@@ -44,21 +47,84 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
+		defaultPoolList = append(defaultPoolList, b)
+	}
+
+	return defaultPoolList, size, nil
+}
+
 func runInit(dockerCli command.Cli, flags *pflag.FlagSet, opts initOptions) error {
+	var size int
```
ctelfer: should this default to 24 if nothing is specified?

cli/command/swarm/init_test.go
```diff
@@ -1,15 +1,18 @@
...
 
 import (
-	"fmt"
 	"io/ioutil"
+	"net"
 	"testing"
 
+	"fmt"
+
 	"github.com/docker/cli/internal/test"
```
ctelfer: nit:  placement of "fmt" seems to violate go idioms -- i.e. it should be up before "io/ioutil" so it is in alphabetical order.

cli/command/swarm/init_test.go
```diff
@@ -123,3 +126,31 @@ func TestSwarmInit(t *testing.T) {
...
+
+	// Pass CIDR and subnet with white space and make sure everything works
+	addrList, _, err = getDefaultAddrPool(" 20.20.0.0/20 , 30.30.0.0/20 , 40.40.0.0/16 : 24")
+	assert.NilError(t, err)
+	assert.Check(t, is.Equal(len(addrList), 3))
+
+	// Pass CIDR just with IP address and it should fail
+	_, _, err = getDefaultAddrPool(" 20.20.0.0, 30.30.0.0/20 : 24")
+	assert.Check(t, err != nil)
+
```
ctelfer: maybe add a check where the list should succeed but verify that the size parsed correctly and was not default?

ctelfer: Should there be a test where the subnet size == the pool allocation size?

docs/reference/commandline/swarm_init.md
ctelfer: nit:  no closing '>'  Actually, maybe the format should be `cidr[,cidr]*:masklen`. 

In general, a problem I'm having is that we refer to that last parameter as the "subnet size".  But that's actually misleading.  It's the length of the mask that accordingly defines the subnet size.  This applies to parts of the code as well.

Maybe clearer terminology would be
  * "address block" or "address range" -- contiguous address space that will be carved up into allocatable subnets
  * "subnet mask length" or just "mask length" -- the mask length for all subnets in the global address pool

These are a bit more verbose.  I dunno.  Maybe the existing terminology is more clear to others.  Would welcome comment.

docs/reference/commandline/swarm_init.md
```diff
@@ -128,6 +129,11 @@ management traffic of the cluster.
 If unspecified, Docker will use the same IP address or interface that is used for the
 advertise address.
 
+### `--default-addr-pool`
+This flag specifies default subnet pools for global scope networks. If subnet size is not
+specified then default value 24 will be used.
+If unspecified, Docker will use the predefined subnets as it works on older releases.
```
ctelfer: Probably worth repeating the format description here, but with more info.

docs/reference/commandline/swarm_init.md
selansen: I will remove "<"  and change the format . WRT subnet size, the same name has been used in existing code and changing it will again mislead having two different name which points to same attribute.

cli/command/swarm/init.go
selansen: done

cli/command/swarm/init.go
selansen: I was thinking otherway. if the input is wrong, we should tell more information about correct error format in the error message.   But I will make the changes

cli/command/swarm/init.go
selansen: make sense. I moved the check above in the logic , so it will check only one time.

cli/command/swarm/init.go
selansen: done

cli/command/swarm/init.go
selansen: getDefaultAddrPool() will set the value to default 24 if nothing is specified.

cli/command/swarm/init_test.go
selansen: done

docs/reference/commandline/swarm_init.md
selansen: added sample example

cli/command/swarm/init.go
```diff
@@ -44,21 +47,84 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
 	flags.StringVar(&opts.availability, flagAvailability, "active", `Availability of the node ("active"|"pause"|"drain")`)
+	flags.StringVar(&opts.defaultAddrPool, flagDefaultAddrPool, "", "List of default subnet addresses followed by subnet size (format: <subnet,subnet,..:subnet-size)")
+	flags.SetAnnotation(flagDefaultAddrPool, "version", []string{"1.39"})
 	addSwarmFlags(flags, &opts.swarmOptions)
 	return cmd
 }
 
+// getDefaultAddrPool extracts info from address pool string and fills PoolsOpt struct
+func getDefaultAddrPool(defaultAddrPool string) ([]*net.IPNet, int, error) {
+	var size int
```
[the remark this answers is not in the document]
selansen: done

cli/command/swarm/init_test.go
```diff
@@ -123,3 +126,31 @@ func TestSwarmInit(t *testing.T) {
 		golden.Assert(t, cli.OutBuffer().String(), fmt.Sprintf("init-%s.golden", tc.name))
 	}
 }
+
+func TestGetDefaultAddrPool(t *testing.T) {
+
```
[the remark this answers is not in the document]
selansen: done

cli/command/swarm/init.go
```diff
@@ -44,21 +47,84 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
+		defaultPoolList = append(defaultPoolList, b)
+	}
+
+	return defaultPoolList, size, nil
+}
+
 func runInit(dockerCli command.Cli, flags *pflag.FlagSet, opts initOptions) error {
+	var size int
+	var defaultAddrPool []*net.IPNet
+	var err error
```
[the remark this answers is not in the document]
selansen: done

cli/command/swarm/init_test.go
```diff
@@ -123,3 +126,31 @@ func TestSwarmInit(t *testing.T) {
...
+	var addrList []*net.IPNet
```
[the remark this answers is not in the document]
selansen: done

selansen:
@thaJeztah , Pls take a look at it and give your final blessing :)

cli/command/swarm/init.go
```diff
@@ -44,21 +47,87 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
 	flags.BoolVar(&opts.autolock, flagAutolock, false, "Enable manager autolocking (requiring an unlock key to start a stopped manager)")
 	flags.StringVar(&opts.availability, flagAvailability, "active", `Availability of the node ("active"|"pause"|"drain")`)
+	flags.StringVar(&opts.defaultAddrPool, flagDefaultAddrPool, "", "List of default subnet addresses followed by subnet size (format: <subnet,subnet,..:subnet-size)")
+	flags.SetAnnotation(flagDefaultAddrPool, "version", []string{"1.39"})
 	addSwarmFlags(flags, &opts.swarmOptions)
 	return cmd
 }
 
+// getDefaultAddrPool extracts info from address pool string and fills PoolsOpt struct
+func getDefaultAddrPool(defaultAddrPool string) ([]*net.IPNet, int, error) {
```
euanh: Minor, but it would be clearer to call this `parseDefaultAddrPool()`.   `get` suggests an accessor which fetches information from somewhere;  this function just parses the string which is passed to it.

cli/command/swarm/init.go
```diff
@@ -44,21 +47,87 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
+		defaultPoolList []*net.IPNet
+	)
+	if defaultAddrPool == "" {
+		// defaultAddrPool is not defined
+		return nil, 0, nil
+	}
+
+	result := strings.Split(defaultAddrPool, ":")
+	if len(result) > 2 {
+		return nil, size, fmt.Errorf("Invalid default address pool format. Expected format CIDR[,CIDR]*:SUBNET-SIZE")
```
euanh: This will break for IPv6 literals, including IPv4-mapped addresses such as `::ffff:192.0.2.0` which we could support even if actual IPv6 subnets are not supported.

Instead you could do something like this:
```split := strings.LastIndex(defaultAddrPool, ":")
        if split == -1 {
                return nil, size, fmt.Errorf(...)
        }
                ...
        size, err := strconv.Atoi(strings.TrimSpace(defaultAddrPool[split+1:]))
        ...
        subnetlist := strings.Split(defaultAddrPool[:split], ",")
        ...
```

cli/command/swarm/init_test.go
```diff
@@ -123,3 +124,28 @@ func TestSwarmInit(t *testing.T) {
 		golden.Assert(t, cli.OutBuffer().String(), fmt.Sprintf("init-%s.golden", tc.name))
 	}
 }
+
+func TestGetDefaultAddrPool(t *testing.T) {
+	// Test only CIRD input and make sure default subnet size is returned
```
euanh: CIRD -> CIDR

cli/command/system/info.go
```diff
@@ -261,7 +262,19 @@ func printSwarmInfo(dockerCli command.Cli, info types.Info) {
...
+		var strAddrPool strings.Builder
+		if info.Swarm.Cluster.DefaultAddrPool != nil {
+			for _, p := range info.Swarm.Cluster.DefaultAddrPool {
+				strAddrPool.WriteString(p.String() + "  ")
+			}
+			fmt.Fprintln(dockerCli.Out(), " Default Address Pool:", strAddrPool.String())
+			fmt.Fprintln(dockerCli.Out(), " SubnetSize:", info.Swarm.Cluster.SubnetSize)
+		} else {
+			fmt.Fprintln(dockerCli.Out(), " Default Address Pool:", "10.0.0.0/8")
+			fmt.Fprintln(dockerCli.Out(), " SunbetSize: 24")
```
euanh: `SunbetSize` -> `SubnetSize`

cli/command/swarm/init.go
euanh: It might be cleaner to have a separate `default-addr-pool-size` parameter, which could default to 24.   `:` is the group separator for IPv6 literals, so it is confusing also to use it to separate fields in a parameter.

selansen:
@vdemeester PTAL.

cli/command/swarm/init.go
selansen: actually , it extracts information from string. it returns IPNet , size and error if validation fails.

cli/command/swarm/init.go
selansen: I dont think we are going to use IPV6 or IPV4 mapped address format here.

cli/command/swarm/init_test.go
selansen: done

cli/command/system/info.go
selansen: done

andrewhsu:
@selansen looks like conflicts need to be resolved

selansen:
I have been keep doing rebase.  Will fix it now.

selansen:
@andrewhsu , @thaJeztah @vdemeester 
Rebased again. PTAL

andrewhsu:
@selansen conflicts need to be resolved

selansen:
I rebased yesterday and resolved all conflicts.  again :) will fix it now

mavenugo:
@bklau Yes, this one is specific to swarm-mode. We have to think about classic-swarm. But it is unrelated to this PR and we can discuss it in a new issue in docker/swarm if you prefer.

bklau:
@mavenugo @thaJeztah  I have created a new feature PR for a port to Swarm Classic:
https://github.com/docker/swarm/issues/2894
You can put down ideas and further comments on that PR

selansen:
Yes. Still getting my Moby PR to get merged. Then I will re vendor again.

cli/command/swarm/init.go
thaJeztah: As we were discussing a while back, we can use two separate flags; one to set the CIDR, and one to set the subnetmask/size. The flag for CIDR can be specified multiple times, that way we don't even have to do the manual splitting.

vendor/github.com/docker/docker/vendor.conf
```diff
@@ -26,7 +26,7 @@ github.com/imdario/mergo v0.3.6
 golang.org/x/sync 1d60e4601c6fd243af51cc01ddf169918a5407ca
 
 # buildkit
-github.com/moby/buildkit e57eed420c7573ae44875be98fa877175b4677a1
+github.com/moby/buildkit 46f9075ab68a07df2c40ae6e240ce4f9392b3a66 git://github.com/tiborvass/buildkit.git
```
thaJeztah: I ignored this bump, because it's a temporary fork, and this didn't bring in changes to vendored files.

vendor/github.com/docker/docker/vendor.conf
```diff
@@ -125,7 +125,7 @@ github.com/containerd/ttrpc 94dde388801693c54f88a6596f713b51a8b30b2d
 github.com/gogo/googleapis 08a7655d27152912db7aaf4f983275eaf8d128ef
 
 # cluster
-github.com/docker/swarmkit 8852e8840e30d69db0b39a4a3d6447362e17c64f
+github.com/docker/swarmkit cfa742c8abe6f8e922f6e4e920153c408e7d9c3b
```
thaJeztah: Not critical, but SwarmKit should also be bumped  in this PR, as the protobufs are used in the CLI (to show default values). I have another PR open that also brings in these changes, so if that's merged first, it's already taken care of (see https://github.com/docker/cli/pull/1225)

selansen: Then Ideally I should go back to Moby update swarmkit vndr there and then update moby vndr here. I cant directly update swarmkit vndr here, right ?

thaJeztah: Oh; sorry. perhaps my comment was confusing. When I update the (e.g.) github.com/docker/docker in `/vendor.conf`, I also check if the new version of github.com/docker/docker is expecting more recent versions of other dependencies. I do so by checking the nested `vendor.conf` of the `github.com/docker/docker` dependency (this file :smile:).

I then check any dependency that was changed there and that's also used by docker/cli, and check if it's more recent than the one currently used; if so, I consider updating the dependency in docker/cli as well.

selansen: will update now. In that way one PR will have all related information.

thaJeztah:
I still don't like the custom format for this flag for various reasons (some also mentioned above);

- We need to maintain custom parsing code to handle this format
- The format uses _positional arguments_ (1..n CIDR's, _optionally_ followed by a separator, and an _optional_ size). Because of this, the format is not easily extensible; adding another option (for whatever reason) would mean that new option is also positional, and with `size` being optional, this means including _empty_ values in the format (`10.10.0.0/16,20.20.0.0/16::someoption:::otheroption`; `::` meaning; an optional argument was left blank).
- Using `:` as a separator is troublesome; even though we currently don't support it, using `:` as a separator means we will never be able to specify IPv6 pools.
- This option will only be set _once_; when initializing a new swarm cluster. Because of that, there's no reason at all to define a shorthand syntax. We should limit shorthand syntaxes to options that are _frequently_ used (e.g. the shorthand `-p 1234:80` notation to publish ports). For options that are rarely used, we should prefer verbose over short.

## So, what options do we have?

### 1. (rejected) CSV notation; repeat flag for each pool.

This would've been my preferred option; use the same format as is used on the daemon;

```bash
docker swarm init \
  --advertise-addr 127.0.0.1:2377 \
  --default-addr-pool base=10.10.0.0/16,size=24 \
  --default-addr-pool base=20.20.0.0/16,size=24 \
```

We discussed that option, but having independent sizes per pool would complicate things, and (though this could be enforced in validation), preference was to have a single size for all pools, and to reflect this in the API and CLI


### 2. CSV notation; allow multiple pools to be in a single flag; 

As suggested by @mark-church https://github.com/docker/cli/pull/1233#discussion_r210976749


```bash
docker swarm init \
  --advertise-addr 127.0.0.1:2377 \
  --default-addr-pool pools=10.10.0.0/16,10.10.0.0/16,10.10.0.0/16,mask=20
```

Some initial thoughts;

- we may need special parsing for that format, due to a comma being both used as separator for different options in the CSV format, but also as separator between `pools`
- `pools` feels a bit redundant; should we use `base`? `cidr`? Something else?
- interestingly, `mask` is suggested; there was discussion around this on the SwarmKit PR, because `size` was not the best name for that option. In that PR it was decided to use `size` because that's what's already used on the daemon.
  - I'm still on the fence for that one; if the `daemon.json` option is the _only_place where we currently use this naming, we can still deprecate that option (keep it as an alias for backward compatibiity), and change to a better name. 


There is one more caveat with this format: if we (again, for whatever reason) need to add an additional configuation _per pool_, things will become complicated quickly. The flag now describes multiple pools, and a single `mask` (size); if we want to add an option per-pool, that option would (probably) need to be a comma-separated field, and options in that field will be positional.

Perhaps better described with an example;


```bash
docker swarm init \
  --default-addr-pool pools=<cidr-1>,<cidr-2>(......),<cidr-5>,mask=20,<option-for-cidr-1>,,,,<option-for-cidr-5>
```

To correlate "option-for-cidr-N" with "cidr-N", those options would have to be specifiedin the same order, and at the same position; easy to make mistakes there, and difficult to read.


### 3. CSV: one flag per pool, separate flags for CIDR and Size ("mask"? "mask-length"?)

Use CSV notation to allow adding options in future, but keep `size` separate (because it's shared between all pools);

```bash
docker swarm init \
  --advertise-addr 127.0.0.1:2377 \
  --default-addr-pool cidr=10.10.0.0/16 \
  --default-addr-pool cidr=20.20.0.0/16 \
  --default-addr-pool cidr=30.30.0.0/16 \
  --default-addr-pool-size 20
```

### 4. No CSV; one flag per pool, separate flags for CIDR and Size ("mask"? "mask-length"?)

The CSV format in 3. is currently "overkill", given that there's only one option (`cidr`).  We can use a plain `cidr` format;

```bash
docker swarm init \
  --advertise-addr 127.0.0.1:2377 \
  --default-addr-pool 10.10.0.0/16 \
  --default-addr-pool 20.20.0.0/16 \
  --default-addr-pool 30.30.0.0/16 \
  --default-addr-pool-size 20
```

If, in future, we _do_ need additional options per-pool, we can _add_ support for the CSV format; this can be done while remaining backward-compatible. Similar to `-p` accepting both a shorthand and CSV notation;



```
# no comma or "=": parse as plain CIDR
--default-addr-pool 10.10.0.0/16

# contains "=": parse as CSV
--default-addr-pool cidr=10.10.0.0/16
```



With the current status, my preference would be to use option 4;

- It's verbose, but not overly lengthy
- It uses well-defined, well-known formats for the values to specify
- It's easy to validate (Golang standard library should provide this)
- No custom format, so no custom parsing is needed

vdemeester:
Same as @thaJeztah, I would definitely prefer the option 4. :angel:

selansen:
@mark-church pls let me know if you are ok with @thaJeztah's option 4.  I will start working on the PR.

cli/command/system/info.go
```diff
@@ -261,7 +262,19 @@ func printSwarmInfo(dockerCli command.Cli, info types.Info) {
...
 		fmt.Fprintln(dockerCli.Out(), " Nodes:", info.Swarm.Nodes)
+		var strAddrPool strings.Builder
+		if info.Swarm.Cluster.DefaultAddrPool != nil {
+			for _, p := range info.Swarm.Cluster.DefaultAddrPool {
+				strAddrPool.WriteString(p + "  ")
+			}
+			fmt.Fprintln(dockerCli.Out(), " Default Address Pool:", strAddrPool.String())
+			fmt.Fprintln(dockerCli.Out(), " SubnetSize:", info.Swarm.Cluster.SubnetSize)
+		} else {
+			fmt.Fprintln(dockerCli.Out(), " Default Address Pool:", "10.0.0.0/8")
```
thaJeztah: Just noticed this, but wondering if this is correct; Is this the only pool that swarmkit uses?

Also; if this default is known daemon-side, it should be returned by the daemon (i.e., we should not hard-code the value here)

cli/command/swarm/init.go
```diff
@@ -45,21 +48,84 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
+	if defaultAddrPool == "" {
+		// defaultAddrPool is not defined
+		return nil, 0, nil
+	}
+
+	result := strings.Split(defaultAddrPool, ":")
+	if len(result) > 2 {
+		return nil, 0, fmt.Errorf("Invalid default address pool format. Expected format CIDR[,CIDR]*:SUBNET-SIZE")
+	}
+	// if size is not specified default size is 24
```
thaJeztah: We should make sure that a default is set daemon-side; at the API level, defaults should be useful, so sending `0` should make the daemon pick the default value.

cli/command/system/info.go
selansen: Yes. This is the only pool we have hardcoded in libnetwork. As this information is not available in swarmkit we dont store it in cluster object currently. I can set it in moby instead of CLI here. 
Either I can remove when I create new moby PR and update it or I can remove it here now and later update moby code with new PR. let me know how do  you prefer. I can fix it.

cli/command/swarm/init.go
selansen: This check will go away with new CLI syntax . But will add extra check and setting default value in moby along with above change you suggested

thaJeztah:
@selansen I just recalled we were discussing to use an IPNet-flag for this at the time; spf13 didn't have a flag for an IPNet _slice_, but there's an open PR for that; I did a quick try to update this PR with option 3, and using a version of spf13 that does have the IPNet slice flag (forked that repo and pulled in the changes from that PR); with that the code needed for this feature is really minimal; https://github.com/docker/cli/compare/master...thaJeztah:default_addr_pool_update

selansen:
Yes. that will save few validation code I have just finished now. 
is it ok to use temp commit id on cli vendor.conf ?

cli/command/system/info.go
thaJeztah: It's better to remove it here, and only print the information we actually get from the API.

> This is the only pool we have hardcoded in libnetwork. As this information is not available in swarmkit we dont store it in cluster object currently.

Perhaps it's worth considering to define a default either in moby, or in swarmkit, so that the pool is always initialised (also from a swarmkit perspective); we can discuss, and change that after this PR is merged though

cli/command/system/info.go
```diff
@@ -261,7 +262,19 @@ func printSwarmInfo(dockerCli command.Cli, info types.Info) {
 		fmt.Fprintln(dockerCli.Out(), " ClusterID:", info.Swarm.Cluster.ID)
 		fmt.Fprintln(dockerCli.Out(), " Managers:", info.Swarm.Managers)
 		fmt.Fprintln(dockerCli.Out(), " Nodes:", info.Swarm.Nodes)
+		var strAddrPool strings.Builder
+		if info.Swarm.Cluster.DefaultAddrPool != nil {
+			for _, p := range info.Swarm.Cluster.DefaultAddrPool {
+				strAddrPool.WriteString(p + "  ")
+			}
+			fmt.Fprintln(dockerCli.Out(), " Default Address Pool:", strAddrPool.String())
```
thaJeztah: oh, sorry, missed that: this should be `" Default Address Pools:"` (plural) here; because we're showing all address pools

cli/command/system/info.go
selansen: I initially thought of displaying only if the user configure it. But later I thought people will complain if they dont see output in in built case.  will remove from here and update it on Moby

cli/command/system/info.go
```diff
@@ -261,7 +262,19 @@ func printSwarmInfo(dockerCli command.Cli, info types.Info) {
 		fmt.Fprintln(dockerCli.Out(), " ClusterID:", info.Swarm.Cluster.ID)
 		fmt.Fprintln(dockerCli.Out(), " Managers:", info.Swarm.Managers)
 		fmt.Fprintln(dockerCli.Out(), " Nodes:", info.Swarm.Nodes)
+		var strAddrPool strings.Builder
+		if info.Swarm.Cluster.DefaultAddrPool != nil {
+			for _, p := range info.Swarm.Cluster.DefaultAddrPool {
+				strAddrPool.WriteString(p + "  ")
```
thaJeztah: Perhaps we should use the same formatting as is used for manager IP-addresses here (one pool per line); https://github.com/docker/cli/blob/fd060d2f337a2698797836ebfdae74ae2e604869/cli/command/system/info.go#L300-L304

cli/command/system/info.go
selansen: we had debate on that a while back. pool itself means group of address. so we decided to keep it like that.  I can change it if you want .

cli/command/system/info.go
selansen: there were other existing examples had similar format. Hence I followed the same. if you want I can change that too.
Docker info
 Network: bridge host macvlan null overlay

selansen:
@thaJeztah @vdemeester modified code with option 4.  PTAL

cli/command/swarm/init.go
```diff
@@ -45,21 +48,33 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
...
+	flags.IPNetSliceVar(&opts.defaultAddrPools, flagDefaultAddrPool, []net.IPNet{}, "default address pool in CIDR format")
+	flags.SetAnnotation(flagDefaultAddrPool, "version", []string{"1.39"})
+	flags.Uint32Var(&opts.DefaultAddrPoolMaskLength, flagDefaultAddrPoolMaskLength, 24, "default address pool subnet mask length")
+	flags.SetAnnotation(flagDefaultAddrPoolMaskLength, "version", []string{"1.39"})
 	addSwarmFlags(flags, &opts.swarmOptions)
 	return cmd
 }
 
 func runInit(dockerCli command.Cli, flags *pflag.FlagSet, opts initOptions) error {
+	var (
```
thaJeztah: nit: this is now a single variable declaration, so can be;

```go
var defaultAddrPool []string
```

cli/command/system/info.go
thaJeztah: It's not a blocker, we can discuss later, and do a follow up if we think the other format is better

docs/reference/commandline/swarm_init.md
```diff
@@ -26,6 +26,8 @@ Options:
       --availability string             Availability of the node ("active"|"pause"|"drain") (default "active")
       --cert-expiry duration            Validity period for node certificates (ns|us|ms|s|m|h) (default 2160h0m0s)
       --data-path-addr string           Address or interface to use for data path traffic (format: <ip|interface>)
+      --default-addr-pool IPnet     	List of default address pool (format: <cidr>)
```
thaJeztah: Indentation looks off here

docs/reference/commandline/swarm_init.md
```diff
@@ -128,6 +130,14 @@ management traffic of the cluster.
 If unspecified, Docker will use the same IP address or interface that is used for the
 advertise address.
 
+### `--default-addr-pool`
+This flag specifies default subnet pools for global scope networks.
+Format example is `--default-addr-pool 30.30.0.0/16 --default-addr-pool 40.40.0.0/16`
```
thaJeztah: I'll add a "docs-revisit" label; we'll probably want some more information about this feature (more details how it will be used by SwarmKit etc)

vendor.conf
```diff
@@ -72,7 +72,8 @@ github.com/satori/go.uuid d41af8bb6a7704f00bc3b7cba9355ae6a5a80048
 github.com/shurcooL/sanitized_anchor_name 10ef21a441db47d8b13ebcc5fd2310f636973c77
 github.com/sirupsen/logrus v1.0.6
 github.com/spf13/cobra v0.0.3
-github.com/spf13/pflag v1.0.1
+# temporary fork with https://github.com/spf13/pflag/pull/170 applied, which isn't merged yet upstream
+github.com/spf13/pflag 4cb166e4f25ac4e8016a3595bbf7ea2e9aa85a2c https://github.com/thaJeztah/pflag.git
```
thaJeztah: I'll open a tracking-issue for this; let's try to get the PR merged upstream 👍

cli/command/swarm/init.go
selansen: I will start email thread with you and @anshulpundir and cover all points we discussed.

crosbymichael:
LGTM

selansen:
Thanks @thaJeztah @crosbymichael

vendor.conf
thaJeztah: ^^ opened https://github.com/docker/cli/issues/1330

Issue #84, opened 2017-05-15
Trust pinning cannot be enabled by docker content trust 


cyc115:
**Description**
The current version of notary currently support two types of trust pinnings: 1) certificate pinning that pins to a specific certificate and 2) CA pinning that uses a provided certificate of a trusted CA to validate the leaf certificate in the metadata. 

Having the ability to pin to a trusted certificate/CA is extremely important as the current TUFUs model does not prevent MIMA for those who is pulling from the repository for the very first time. 

At the moment, this feature seems to be disabled as an empty trust pin config is being passed into `client.NewNotaryRepository()`.

thaJeztah:
ping @riyazdf @endophage PTAL

endophage:
It's in our TODO list for the next round of `docker trust` additions. Early thoughts on the CLI syntax can be found in the "Configuring Trust in Docker" section of this doc: https://docs.google.com/document/d/1JOBAlCDuf5JnpVLW54voGuAdsnMmSR2LIbs8fjBDSaM/

I talked to @cyc115 about contributing this a while ago. Are you planning to take this on or is this issue a feature request? If you're planning to implement it, let's bikeshed in this issue a little on the syntax, because I've been thinking maybe it could be shortened from `docker trust config pin ...` in that doc, to just `docker trust pin ...` 

Also, we should decide in `pin` is the term we want to use. Security people will understand it, but if there's a more intuitive term we can use we're not married to `pin`. Remember that one of the goals of the `docker trust` command is to be more intuitive than notary and to hide the TUF concepts behind something much more friendly to the typical user.

maltfield:
See also this issue, which would require user intervention to proceed with TOFU after warning them of the risks in doing so

 * https://github.com/docker/cli/issues/2752

thaJeztah:
- closing because we removed `docker trust` from the CLI; https://github.com/docker/cli/pull/6603

maltfield:
So, uhh, you guys just decided to give up on any chance of providing security in docker?

PR #1236, opened 2018-07-26
Enable notary trust pinning with docker content trust

kkroo:
This adds an optional trust_pinning configuration section that specifies how to bootstrap the root of docker's content trust.

As defined in https://github.com/theupdateframework/notary/blob/master/docs/reference/client-config.md, one can provide specific certificates to pin to, or a CA to pin to as a root of trust for a GUN.
Multiple sections can be specified, but the pinned certificates will take highest priority for validation, followed by the pinned CA, followed by TOFUS (TOFU over HTTPS).

Replaces empty trust pinning in `GetNotaryRepository` with  structure parsed from configuration.

Addresses #84

### DIFF
--- a/cli/command/image/trust.go
+++ b/cli/command/image/trust.go
@@ -102,7 +102,8 @@ func PushTrustedReference(streams command.Streams, repoInfo *registry.Repository
 
 	fmt.Fprintln(streams.Out(), "Signing and pushing trust metadata")
 
-	repo, err := trust.GetNotaryRepository(streams.In(), streams.Out(), command.UserAgent(), repoInfo, &authConfig, "push", "pull")
+	trustPinningConfig := streams.(command.Cli).ConfigFile().ParseTrustPinning()
+	repo, err := trust.GetNotaryRepository(streams.In(), streams.Out(), command.UserAgent(), repoInfo, &authConfig, trustPinningConfig, "push", "pull")
 	if err != nil {
 		return errors.Wrap(err, "error establishing connection to trust repository")
 	}
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -14,6 +14,7 @@ import (
 	"github.com/docker/cli/opts"
 	"github.com/docker/docker/api/types"
 	"github.com/pkg/errors"
+	"github.com/theupdateframework/notary/trustpinning"
 )
 
 const (
@@ -48,6 +49,14 @@ type ConfigFile struct {
 	Experimental         string                      `json:"experimental,omitempty"`
 	StackOrchestrator    string                      `json:"stackOrchestrator,omitempty"`
 	Kubernetes           *KubernetesConfig           `json:"kubernetes,omitempty"`
+	TrustPinning         *TrustPinningConfig         `json:"trustPinning,omitempty"`
+}
+
+// TrustPinningConfig contains notary trust pinning settings
+type TrustPinningConfig struct {
+	Certs       map[string][]string `json:certs,omitempty`
+	CA          map[string]string   `json:ca,omitempty`
+	DisableTOFU bool                `json:disableTofu,omitempty`
 }
 
 // ProxyConfig contains proxy configuration settings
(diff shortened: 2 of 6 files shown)


cli/config/configfile/file.go
```diff
@@ -186,6 +195,27 @@ func (configFile *ConfigFile) Save() error {
 	return configFile.SaveToWriter(f)
 }
 
+// ParseTrustPinning parses Notary trust pinning structure from configuration, and prepends CA certificate files with
+// the absolute path of the configuration directory
+func (configFile *ConfigFile) ParseTrustPinning() trustpinning.TrustPinConfig {
```
silvin-lubecki: Shouldn't this method replace in place the TrustPinning config instead of returning a new one? WDYT @vdemeester ?

silvin-lubecki (changes requested):

cli/command/plugin/push.go
```diff
@@ -69,7 +69,7 @@ func runPush(dockerCli command.Cli, opts pushOptions) error {
 
 	if !opts.untrusted {
 		repoInfo.Class = "plugin"
-		return image.PushTrustedReference(dockerCli, repoInfo, named, authConfig, responseBody)
+		return image.PushTrustedReference(dockerCli, repoInfo, named, authConfig, dockerCli.ConfigFile().ParseTrustPinning(), responseBody)
```
thaJeztah: Wondering if it would be better to pass the pinningconfig through `opts` (add new properties for this to the `pushOptions` type)

Also note that we already pass `dockerCli` as argument, so this information _could_ already be obtained inside `PushTrustedReference()` without changing it signature.

cli/command/image/trust.go
```diff
@@ -37,12 +38,14 @@ func TrustedPush(ctx context.Context, cli command.Cli, repoInfo *registry.Reposi
...
 
-	return PushTrustedReference(cli, repoInfo, ref, authConfig, responseBody)
+	return PushTrustedReference(cli, repoInfo, ref, authConfig, cli.ConfigFile().ParseTrustPinning(), responseBody)
 }
 
 // PushTrustedReference pushes a canonical reference to the trust server.
 // nolint: gocyclo
-func PushTrustedReference(streams command.Streams, repoInfo *registry.RepositoryInfo, ref reference.Named, authConfig types.AuthConfig, in io.Reader) error {
+func PushTrustedReference(streams command.Streams, repoInfo *registry.RepositoryInfo, ref reference.Named,
+	authConfig types.AuthConfig, trustPinningConfig trustpinning.TrustPinConfig, in io.Reader,
```
thaJeztah: The list of arguments here is becoming a bit lengthy; might be worth looking if we should refactor this (also see my other comment)

cli/config/config_test.go
```diff
@@ -151,7 +151,7 @@ func TestOldValidAuth(t *testing.T) {
 	// defaultIndexserver is https://index.docker.io/v1/
 	ac := config.AuthConfigs["https://index.docker.io/v1/"]
 	if ac.Username != "joejoe" || ac.Password != "hello" {
-		t.Fatalf("Missing data from parsing:\n%q", config)
+		t.Fatalf("Missing data from parsing:\n%q", &config)
```
thaJeztah: Why was this changed to a pointer?

thaJeztah (changes requested):
left some initial comments, but could use input from @cyli and @justincormack

kkroo:
PTAL

AkihiroSuda:
What's current status? @justincormack

amouat:
FWIW, we have some clients that would really benefit from this feature.

chaospuppy:
This is a massive blocker for tighter, more practical integration between Notary and Docker.  The ability to pin CAs (as well as specific certs) for GUNs as a feature of DCT is integral to self-hosting Notary in a secure manner.  Can anyone provide an update?

thaJeztah:
closing, because the `docker trust` subcommands were removed (code is still present currently to build it as a plugin; https://github.com/docker/cli/tree/v29.4.3/cmd/docker-trust)

PR #1237, opened 2018-07-27
Migrate to non-deprecated functions of `api/types/filters`

vdemeester:
- Use `Contains` instead of `Include`
- Use `ToJSON` instead of `ToParam`
- Remove usage of `ParseFlag` as it is deprecated too

See https://godoc.org/github.com/docker/docker/api/types/filters for `Deprecated` methods.


thaJeztah:
screen shot 2018-07-27 at 15 44 01

Needs 42.86% more coverage :trollface:

PR #1240, opened 2018-07-30
Add API-version anotation for --data-path-addr

thaJeztah:
This flag was added in Docker 17.06, API version 1.31 through https://github.com/moby/moby/pull/32717/commits/8dc8cd4719f165c01c98e7d3ce1d6cea6a8f60b 8(https://github.com/moby/moby/pull/32717), but didn't add API-version annotations.

This patch adds the missing annotations to hide this flag if the CLI is connected to an older version of the daemon that doesn't support that API.

Before this patch:

    DOCKER_API_VERSION=1.30 docker swarm init --help | grep data-path-addr

    DOCKER_API_VERSION=1.31 docker swarm init --help | grep data-path-addr

With this patch applied:

    DOCKER_API_VERSION=1.30 docker swarm init --help | grep data-path-addr
    # (no result)

    DOCKER_API_VERSION=1.31 docker swarm init --help | grep data-path-addr

**- How to verify it**

See above

```markdown
+ Hide `--data-path-addr` flags when connected to a daemon that doesn't support this option [docker/cli#1240](https://github.com/docker/cli/pull/1240)
```


thaJeztah:
ping @fcrisciani @selansen

PR #1247, opened 2018-07-31
Allow running `make help` without out-of-container warning

marcov:
Allow running `make help` without specifying the  extra `-f Makefile` option. Reasons:
* Currently `make help` warns about begin out-of-container and prompts using `make -f docker.Makefile help`
* `help` is not a valid target in `docker.Makefile`
* `README.md` instructs to use `make help` as is

A picture of a cute animal (my cat):
mycat

### DIFF
--- a/scripts/warn-outside-container
+++ b/scripts/warn-outside-container
@@ -3,7 +3,7 @@ set -eu
 
 target="${1:-}"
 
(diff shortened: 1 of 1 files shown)


marcov:
@thaJeztah sure!  I'll submit another PR with the `help` draft for docker.Makefile.

PR #1248, opened 2018-07-31
Add `help` and remove `watch` targets in docker.Makefile

marcov:
**- What I did**
* Add the `help` target to document make targets when building using a container
* Remove the `watch` target (filewatcher was removed with c0588a9c)

**- How I did it**
* I reused the smart way to generate an `help` already in place in the main Makefile;
* I moved most of the comments already available to have them printed with `help` and added new descriptions where needed;
* I removed the "using a container" from comments, as having `docker run` as target command is self documenting. Maybe we can add something like "All these targets are run inside a container" as header for `help`.

### DIFF
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -41,84 +41,70 @@ build_e2e_image:
 	docker build -t $(E2E_IMAGE_NAME) --build-arg VERSION=$(VERSION) --build-arg GITCOMMIT=$(GITCOMMIT) -f ./dockerfiles/Dockerfile.e2e .
 
 
-# build executable using a container
-binary: build_binary_native_image
+binary: build_binary_native_image ## build the CLI
 	docker run --rm $(ENVVARS) $(MOUNTS) $(BINARY_NATIVE_IMAGE_NAME)
 
-build: binary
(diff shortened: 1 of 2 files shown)


docker.Makefile
```diff
@@ -41,84 +41,70 @@ build_e2e_image:
...
-binary-windows: build_cross_image
+binary-windows: build_cross_image ## build the CLI for Windows
 	docker run --rm $(ENVVARS) $(MOUNTS) $(CROSS_IMAGE_NAME) make $@
 
 .PHONY: binary-osx
-binary-osx: build_cross_image
+binary-osx: build_cross_image ## build the CLI for macOS
 	docker run --rm $(ENVVARS) $(MOUNTS) $(CROSS_IMAGE_NAME) make $@
 
-.PHONY: watch
```
silvin-lubecki: Watch should also be removed in the Makefile no?

silvin-lubecki:
Follow up of #1247

docker.Makefile
marcov: Yea, thanks for pointing it out.

PR #1249, opened 2018-07-31
Add a new `ExtractVariables` function to `compose/template` package

vdemeester:
It allows to get easily all the variables defined in a
composefile (the `map[string]interface{}` representation that
`loader.ParseYAML` returns at least) and their default value too.

This commit also does some small function extract on substitution
funcs to reduce a tiny bit duplication.

Can be really useful for project that enhance the compose implementation like `docker/app`.

### DIFF
--- a/cli/compose/template/template.go
+++ b/cli/compose/template/template.go
@@ -93,25 +93,91 @@ func Substitute(template string, mapping Mapping) (string, error) {
 	return SubstituteWith(template, mapping, pattern, DefaultSubstituteFuncs...)
 }
 
+// ExtractVariables returns a map of all the variables defined in the specified
+// composefile (dict representation) and their default value if any.
+func ExtractVariables(configDict map[string]interface{}) map[string]string {
+	return recurseExtract(configDict)
+}
+
+func recurseExtract(value interface{}) map[string]string {
+	m := map[string]string{}
+
+	switch value := value.(type) {
+	case string:
+		if v, is := extractVariable(value); is {
+			m[v.name] = v.value
+		}
+	case map[string]interface{}:
+		for _, elem := range value {
+			submap := recurseExtract(elem)
+			for key, value := range submap {
+				m[key] = value
+			}
+		}
+
+	case []interface{}:
+		for _, elem := range value {
+			if v, is := extractVariable(elem); is {
+				m[v.name] = v.value
+			}
+		}
+	}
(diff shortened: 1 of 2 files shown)


cli/compose/template/template.go
```diff
@@ -93,25 +93,91 @@ func Substitute(template string, mapping Mapping) (string, error) {
...
+		if escaped := groups["escaped"]; escaped != "" {
+			return extractedValue{}, false
+		}
+		val := groups["named"]
+		if val == "" {
+			val = groups["braced"]
+		}
+		name := val
+		var defaultValue string
+		if strings.Contains(val, ":?") {
```
silvin-lubecki: nit
```go
switch{
case string.Contains(val, ":?"):
    name, _ = partition(val, ":?")
case strings.Contains(val, "?"):
    name, _ = partition(val, "?")
...
}
return extractedValue{name: name, value: defaultValue}, true
}
```

cli/compose/template/template_test.go
```diff
@@ -172,3 +172,91 @@ func TestSubstituteWithCustomFunc(t *testing.T) {
...
+			dict: map[string]interface{}{
+				"foo": "${bar?foo}",
+			},
+			expected: map[string]string{
+				"bar": "",
+			},
+		},
+		{
+			dict: map[string]interface{}{
+				"foo": "${bar:-foo}",
```
thaJeztah: Nested substitutions are not supported currently? (e.g. `${foo:-$bar}`)

cli/compose/template/template_test.go
```diff
@@ -172,3 +172,91 @@ func TestSubstituteWithCustomFunc(t *testing.T) {
...
+				},
+			},
+			expected: map[string]string{
+				"bar":   "foo",
+				"fruit": "banana",
+				"toto":  "",
+			},
+		},
+	}
+	for _, tc := range testCases {
```
thaJeztah: Could change this to subtests as a follow up?

vdemeester: Yes sir ! :stuck_out_tongue_closed_eyes:

cli/compose/template/template_test.go
vdemeester: I'm not actually sure abou that :stuck_out_tongue_closed_eyes:

PR #1137, opened 2018-06-21
Fix broken swarm commands with Kubernetes defined as orchestrator

silvin-lubecki:
**- What I did**

All swarm commands (node, service, etc...) were erroring out if the orchestrator was defined as Kubernetes. I moved the `orchestrator` flag, the environment variable and the config file value to more locally values regarding stack command.

**- How I did it**

* Renamed DOCKER_ORCHESTRATOR to DOCKER_STACK_ORCHESTRATOR
* Renamed config file option "orchestrator" to "stackOrchestrator"
* "--orchestrator" flag is no more global but local to stack command and subcommands
* Cleaned all global orchestrator code
* Replicated Hidden flags in help and Supported flags from root command to stack command

**- How to verify it**

```sh
$ DOCKER_STACK_ORCHESTRATOR=kubernetes docker node ls
ID                            HOSTNAME                STATUS              AVAILABILITY        MANAGER STATUS      ENGINE VERSION
1u26o5yll4mdqxfr1o86xphb5 *   linuxkit-025000000001   Ready               Active              Leader              18.06.0-ce-dev
```

**- Description for the changelog**
* Fixed broken swarm commands with Kubernetes defined as orchestrator
* Renamed DOCKER_ORCHESTRATOR to DOCKER_STACK_ORCHESTRATOR
* Renamed config file option "orchestrator" to "stackOrchestrator"
* "--orchestrator" flag is no more global but local to stack command and subcommands

**- A picture of a cute animal (not mandatory but encouraged)**
image

PR #1252, opened 2018-08-01
Add bash completion for experimental CLI commands (manifest)

albers:
This adds bash completion for `docker manifest` and its subcommands.

Completion for the `manifest` command is only activated if experimental CLI features are enabled (either by setting `"experimental": "enabled"` in _.docker/config.json_ or the env var `DOCKER_CLI_EXPERIMENTAL=enabled`). This configuration setting is harvested from the output of `docker version`.

Before adding the third configuration setting that is extracted from `docker version`'s output , I refactored the existing logic to a more consistent and clearer state. There is only one call to `docker version`. It is executed lazily and caches its result.

*Note: The initial concept also included gathering the orchestrator in this step, but this piece of information went bananas in #1137.*

In the context of `docker version` and its output, I changed the domain language to use _server_ instead of _daemon_ because this is the term used there (bounded context). _daemon_ is better suited for local calls whereas _server_ is more appropriate for (potentially) remote calls.

Edit: The list of valid os/arch values was taken from `validOSArches ` in [util.go](https://github.com/docker/cli/blob/master/cli/command/manifest/util.go). A comment was left as a reminder to keep both files in sync.

### DIFF
--- a/cli/command/manifest/util.go
+++ b/cli/command/manifest/util.go
@@ -18,6 +18,7 @@ type osArch struct {
 // list of valid os/arch values (see "Optional Environment Variables" section
 // of https://golang.org/doc/install/source
 // Added linux/s390x as we know System z support already exists
+// Keep in sync with _docker_manifest_annotate in contrib/completion/bash/docker
 var validOSArches = map[osArch]bool{
 	{os: "darwin", arch: "386"}:      true,
 	{os: "darwin", arch: "amd64"}:    true,
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -568,9 +568,17 @@ __docker_append_to_completions() {
 # The result is cached for the duration of one invocation of bash completion.
 __docker_fetch_info() {
 	if [ -z "$info_fetched" ] ; then
-		read -r server_experimental server_os < <(__docker_q version -f '{{.Server.Experimental}} {{.Server.Os}}')
+		read -r client_experimental server_experimental server_os < <(__docker_q version -f '{{.Client.Experimental}} {{.Server.Experimental}} {{.Server.Os}}')
 		info_fetched=true
 	fi
+}
+
+# __docker_client_is_experimental tests whether the Docker cli is configured to support
+# experimental features. If so, the function exits with 0 (true).
+# Otherwise, or if the result cannot be determined, the exit value is 1 (false).
+__docker_client_is_experimental() {
+	__docker_fetch_info
+	[ "$client_experimental" = "true" ]
 }
 
 # __docker_server_is_experimental tests whether the currently configured Docker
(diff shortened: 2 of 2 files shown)


albers:
@vdemeester It's still very basic with regard of the arguments. I don't know what are valid values for MANIFEST_LIST and MANIFEST:

    Usage:  docker manifest create MANIFEST_LIST MANIFEST [MANIFEST...]

So I do not complete anything right now, allowing any string.

albers:
@vdemeester Do you know who I might ask for help?

vdemeester:
cc @clnperez @estesp @thaJeztah :angel:

clnperez:
@albers is there a check done for image names anywhere? the same rules apply (though I admit I don't know much about how the bash-completion verification works).

albers:
@clnperez So I can use an image name as a _MANIFEST_?
I successfully tried `docker manifest inspect alpine:3.6`, but I got an error with
```bash
$ docker manifest create alpine-list alpine:3.6 alpine:3.7 alpine:3.8
docker.io/library/alpine:3.6 is a manifest list
```
Can you please give me a working example fo a `manifest create` command?

clnperez:
There are some examples in the doc: https://docs.docker.com/engine/reference/commandline/manifest/#working-with-insecure-registries

Your issue is that `alpine:3.6`, `alpine:3.7` and `alpine:3.8` are all already manifest lists, and you can't make a recursive manifest list.

albers:
@clnperez I found the examples not very helpful because they are not working code. There seem to be some prerequistes required for the commands to work apart from pulling down images and pushing them to a local registry.
I tried:
```bash
$ docker pull busybox:1.29.2-uclibc
$ docker tag busybox:1.29.2-uclibc localhost:5000/albers/busybox:1.29.2-uclibc
$ docker push localhost:5000/albers/busybox:1.29.2-uclibc
$ docker manifest create localhost:5000/albers/busybox:1.29.2 localhost:5000/albers/busybox:1.29.2-uclibc
no such manifest: localhost:5000/albers/busybox:1.29.2-uclibc
```

clnperez:
three things: 
- you shouldn't call a manifest list the same thing an your existing image. it will overwrite the name.
- can you try using the --insecure flag? 
- a manifest list is explained in the extended description in the docs. or there's this blog i wrote a while back: https://developer.ibm.com/linuxonpower/2017/07/27/create-multi-architecture-docker-image/

albers:
@clnperez Thanks very much for pointing me to your blog. This was exactly what I needed to understand the feature. I was able to create, annotate and push manifest lists.

albers:
I added completion for manifest lists.
The completion is not perfect because there is no way to 

- list unpushed manifest lists.
- get a list of all images with the manifest lists removed

As an approximation, I complete all images for both manifest lists and manifests.

albers:
Added support for completion of `--os` and `--arch` ([legal values](https://github.com/docker/cli/blob/master/cli/command/manifest/util.go#L21-L49])), rebased and sqashed some commits.

This is now ready for review.
ping @clnperez @thaJeztah

clnperez:
@albers This sounds reasonable to me. I might ask that we put a comment into the file you pulled those values from and in your PR that mentions that we should keep them in sync.

albers:
>... put a comment into the file you pulled those os & arch values from and in your PR that mentions that we should keep them in sync.

@clnperez Done.

clnperez:
@vdemeester @thaJeztah Thoughts on this one?

silvin-lubecki:
Do we wait for #1357 to be merged ?

PR #1254, opened 2018-08-01
Add bash completion for kubernetes orchestrator

albers:
**Depends on #1251** edit: #1251 was merged.

This
- adds completions for kubernetes-specific options if kubernetes orchestrator is configured
- hides completions for swarm-specific options unless swarm orchestrator is configured

The configured orchestrator is determined from the output of `docker stack deploy|ls --help`.
~As the required help output is broken in 18.06.0 (see #1243) this work depends on #1251.~

~To test before #1251 is merged, just use a binary built at #1251.~
I recommend using `export DOCKER_STACK_ORCHESTRATOR=kubernetes|swarm|all`, then invoking the completions of `docker stack (deploy|ls|ps|rm|services)? --` and `docker version --`.

Thanks @silvin-lubecki for fixing the help output so fast.

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -4435,8 +4435,9 @@ _docker_stack_deploy() {
 
 	case "$cur" in
 		-*)
-			local options="--compose-file -c --help --prune --resolve-image --with-registry-auth"
-			__docker_daemon_is_experimental && options+=" --bundle-file"
+			local options="--compose-file -c --help"
+			__docker_daemon_is_experimental && __docker_stack_orchestrator_is swarm && options+=" --bundle-file"
+			__docker_stack_orchestrator_is swarm && options+=" --prune --resolve-image --with-registry-auth"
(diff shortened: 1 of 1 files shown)


contrib/completion/bash/docker
```diff
@@ -4391,12 +4437,12 @@ _docker_stack() {
 }
 
 _docker_stack_deploy() {
+	__docker_complete_stack_orchestrator_options && return
+
 	case "$prev" in
 		--bundle-file)
```
albers: I removed the selective completion here although it is correct behavior.

The important part is that `--bundle-file` is not completed if it is not appropriate, which is already taken care for.
If someone insists on manually typing this option, he will now get it completed although it does not exist in this context.
This is an error, but given the growing complexity of hiding options based on operating system, client/server experimental state and orchestrator replicating the hiding logic in the individual option's completions will bloat the code without much value.

albers:
rebased in order to pull in #1251.

PR #1259, opened 2018-08-01
Add unit test to cover network list sort, refactor to table-driven tests

adshmh:
**- What I did**
Added unit test to cover the sort order for `network list` command
Refactored all the unit tests for `network list` command to table-driven

**- How I did it**
Added unit tests under `cli/command/network/list_test.go`, and test data under `cli/command/network/testdata`

**- How to verify it**

Added unit tests to cover network list sort order, refactored to table-driven


adshmh:
Currently the `network list` command uses direct string comparison, and not the `sortorder` library, which is used by other commands, e.g. `plugin list`. Did not modify that to avoid changing the existing behavior. If needed, I can update the PR to use the `sortorder` library for sorting.

silvin-lubecki:
Hi @adshmh , thank you for this PR! I think adding the `sortorder` is a good idea, your PR emphases that the `docker network list` behavior differs from the [other commands](https://github.com/docker/cli/blob/master/cli/command/service/testdata/service-list-sort.golden).

Could you modify it in a follow up? It will be simpler to track.

cli/command/network/list_test.go
```diff
@@ -41,23 +40,53 @@ func TestNetworkListErrors(t *testing.T) {
...
+					*NetworkResource(NetworkResourceName("network-10-foo"))}, nil
+			},
+		},
+	}
 
-	cmd.Flags().Set("filter", "image.name=ubuntu")
-	assert.NilError(t, cmd.Execute())
-	golden.Assert(t, strings.TrimSpace(cli.OutBuffer().String()), "network-list.golden")
+	for _, tc := range testCases {
+		cli := test.NewFakeCli(&fakeClient{networkListFunc: tc.networkListFunc})
```
thaJeztah: Looks like we have everything in place to use subtests here (only change two lines);

```patch
diff --git a/cli/command/network/list_test.go b/cli/command/network/list_test.go
index 55d0b109..2f76e517 100644
--- a/cli/command/network/list_test.go
+++ b/cli/command/network/list_test.go
@@ -81,12 +81,14 @@ func TestNetworkList(t *testing.T) {
        }
 
-               cli := test.NewFakeCli(&fakeClient{networkListFunc: tc.networkListFunc})
-               cmd := newListCommand(cli)
-               for key, value := range tc.flags {
-                       cmd.Flags().Set(key, value)
-               }
-               assert.NilError(t, cmd.Execute())
-               golden.Assert(t, cli.OutBuffer().String(), tc.golden)
+               t.Run(tc.doc, func(t *testing.T) {
+                       cli := test.NewFakeCli(&fakeClient{networkListFunc: tc.networkListFunc})
+                       cmd := newListCommand(cli)
+                       for key, value := range tc.flags {
+                               cmd.Flags().Set(key, value)
+                       }
+                       assert.NilError(t, cmd.Execute())
+                       golden.Assert(t, cli.OutBuffer().String(), tc.golden)
+               })
        }
 }
```

adshmh: Thank you for the review. Fixed.

adshmh:
Thank you for the reviews. I will follow up with a PR for the sort order of the network list command.

PR #1260, opened 2018-08-01
Add CLI support for running dockerd in a container on containerd

dhiltgen:
**- What I did**

This PR adds support to the Docker CLI to manage the lifecycle of a Docker Engine running as a privileged container on top of containerd.   New CLI UX is introduced to initialize a local engine, check for available updates, and update the local engine.  Additional UX is introduced to allow users to transition from a community based engine to an enterprise engine if they so desire, by downloading an existing enterprise license from the Docker Store or initiating a new free trial license.

Not included in this PR are packaging level changes to provide a streamlined user experience based on OS level packages, which allow users to continue to use OS level packages and related automation if they desire for installation and update operations.

**- How I did it**

The implementation leverages the containerd client API and interacts with a local containerd to download images from a registry (defaulting to Docker Hub) as well as a docker licensing library to interact with Docker Store for generating trial licenses and downloading existing licenses.

**- How to verify it**

Unit tests and a new e2e test are included in the PR.  The e2e test runs a nested containerd in a container upon which a nested dockerd is launched. 

Nightly builds for the engine will be published at https://hub.docker.com/r/docker/engine-community/tags/

Manual validation may be performed by installing containerd locally, then running this CLI, specifying a nightly version tag.  For example:

```
root@daniel-testkit-CE988A-0:~# docker engine init --engine-version 0.0.0-20180730163646-c19b8d9
Pulling docker.io/docker/engine-community:0.0.0-20180730163646-c19b8d9... done.
docker.io/docker/engine-community:0.0.0-20180730163646-c19b8d9 already present in containerd.
Waiting for engine to start... waiting for engine to be responsive... engine is online.
Success!  The docker engine is now running.
```

Add support for running dockerd on top of containerd

screen_2018-08-01-14-48-06


dhiltgen:
The recent lint failure looks like a CI hiccup, unrelated to the changes I just pushed
```
...
Step 10/10 : COPY . .
 ---> 3b3e3fa522cc
Successfully built 3b3e3fa522cc
Successfully tagged cli-linter:15395
docker: Cannot connect to the Docker daemon at tcp://35.185.65.192:2376. Is the docker daemon running?.
ERRO[0030] error waiting for container: context canceled 
Exited with code 125
```

Could someone whack CI to re-run that again?

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+			err2 := container.Delete(ctx, containerd.WithSnapshotCleanup)
+			return fmt.Errorf("Failed to create container task to create host dir %s: %s: %s", dirname, err, err2)
+		}
+		err = task.Start(ctx)
+		if err != nil {
+			return fmt.Errorf("Failed to start container task: %s", err)
+		}
+		status, err := task.Wait(ctx)
+		if err != nil {
+			logrus.Fatalf("Failed to wait for container task: %s", err)
```
vdemeester: Why isn't this returning an error and panics instead ?

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+			continue
+		}
+		if err != nil {
+			// Other failures are catastrophic
+			err2 := container.Delete(ctx, containerd.WithSnapshotCleanup)
+			return fmt.Errorf("Failed to create container task to create host dir %s: %s: %s", dirname, err, err2)
+		}
+		err = task.Start(ctx)
+		if err != nil {
+			return fmt.Errorf("Failed to start container task: %s", err)
```
vdemeester: use `errors.Wrap` here too

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+		status, err := task.Wait(ctx)
+		if err != nil {
+			logrus.Fatalf("Failed to wait for container task: %s", err)
+		}
+		<-status
+		err = task.Kill(ctx, SIGKILL)
+		if err != nil && !strings.Contains(err.Error(), "not found") {
+			return fmt.Errorf("failed to kill task: %s", err)
```
vdemeester: use `errors.Wrap` here too

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+		_, err = task.Delete(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to delete task: %s", err)
```
vdemeester: use `errors.Wrap` here too

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// VerifyDockerConfig ensures the specified docker config file exists on the host and returns the rootDir
+func VerifyDockerConfig(ctx context.Context, client WrappedContainerdClient, configFile string, image containerd.Image) (string, error) {
+
+	// First make sure the dir exists on the host
+	configDir := path.Dir(configFile)
+	err := VerifyHostDirExists(ctx, client, configDir, image)
+	if err != nil {
+		return "", fmt.Errorf("failed to verify host dir %s: %s", configDir, err)
```
vdemeester: use `errors.Wrapf` here too

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	// TODO - consider refining this to be a generic utility to create a file on the host
+	// with the file content passed in as a string
+	testerSpec := specs.Spec{
```
vdemeester: Could be extracted in a const or a var :angel:

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+				GID: 0,
+			},
+			Env: []string{
+				"PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin",
+			},
+			NoNewPrivileges: false,
+		},
+	}
+	name := stringid.GenerateRandomID()[:20]
+	cOpts := []containerd.NewContainerOpts{
```
vdemeester: Some of those (with `testerSpec.Mounts`) are duplicated with the function above, we should de-duplicate those :angel:

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+			Type:        "bind",
+		},
+	}
+	container, err := client.NewContainer(
+		ctx,
+		name,
+		cOpts...,
+	)
+	if err != nil {
+		return "", fmt.Errorf("Failed to verify config file %s: %s", configFile, err)
```
vdemeester: use `errors.Wrapf` here too

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	// Capture the output
+	buf := new(bytes.Buffer)
+	task, err := container.NewTask(ctx, defaultNewCreator(withByteBuffers(buf)))
+	if err != nil {
+		return "", fmt.Errorf("Failed to verify %s: %s", configFile, err)
```
vdemeester: use `errors.Wrapf` here too

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	err = task.Start(ctx)
+	if err != nil {
+		return "", fmt.Errorf("Failed to start container task: %s", err)
```
vdemeester: use `errors.Wrap` here too

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	status, err := task.Wait(ctx)
+	if err != nil {
+		return "", fmt.Errorf("Failed to wait for container task: %s", err)
```
vdemeester: use `errors.Wrap` here too

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	<-status
+	err = task.Kill(ctx, SIGKILL)
+	if err != nil {
+		logrus.Debugf("Failed to kill task: %s", err)
```
vdemeester: Same question as above, why panics instead of returning an error here ?

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	_, err = task.Delete(ctx)
+	if err != nil {
+		logrus.Debugf("Failed to delete task: %s", err)
```
vdemeester: almost silently ignoring these errors :thinking:  shouldn't we at least print some warnings ?

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	err = container.Delete(ctx, containerd.WithSnapshotCleanup)
+	if err != nil {
+		logrus.Debugf("Failed to delete container: %s", err)
```
vdemeester: Same here

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+		)
+		if err != nil {
+			return fmt.Errorf("Failed to create container to evaluate host dir %s: %s", target, err)
+		}
+		// We created the container, so run it and wait
+		task, err := container.NewTask(ctx, cio.NullIO)
+		if err != nil && strings.Contains(err.Error(), "no such file or directory") {
+			err = container.Delete(ctx, containerd.WithSnapshotCleanup)
+			if err != nil {
+				return fmt.Errorf("Failed to delete container to evaluate host dir %s: %s", target, err)
```
vdemeester: use `errors.Wrapf` here too

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+	err = container.Delete(ctx, containerd.WithSnapshotCleanup)
+	if err != nil {
+		logrus.Debugf("Failed to delete container: %s", err)
+	}
+	configFileData := buf.Bytes()
+
+	var config map[string]interface{}
+	err = json.Unmarshal(configFileData, &config)
+	if err != nil {
+		return "", fmt.Errorf("Malformed config file - %s: %s", configFile, err)
```
vdemeester: use `errors.Wrapf` here too

cli/command/engine/activate.go
```diff
@@ -0,0 +1,172 @@
...
+		},
+	}
+
+	flags := cmd.Flags()
+
+	flags.StringVarP(&options.licenseFile, "license", "l", "", "License File")
+	flags.StringVarP(&options.engineVersion, "version", "", "", "Specify engine version (default is to use existing version)")
+	flags.StringVarP(&options.registryPrefix, "registry-prefix", "", "docker.io/docker", "Override the default location where engine images are pulled")
+	flags.StringVarP(&options.format, "format", "", "", "Pretty-print licenses using a Go template")
+	flags.BoolVarP(&options.quiet, "queit", "q", false, "Only display available licenses by ID")
```
vdemeester: s/queit/quiet

cli/command/engine/init.go
```diff
@@ -0,0 +1,59 @@
...
+
+func runInit(dockerCli command.Cli, options initOptions) error {
+	dclient := dockerCli.Client()
+	ctx := namespaces.WithNamespace(context.Background(), containerdutils.EngineNamespace)
+	cclient, err := containerdutils.GetContainerdClient()
+	if err != nil {
+		return fmt.Errorf("Unable to access local containerd: %s", err)
+	}
+	defer cclient.Close()
+	return containerdutils.InitEngine(ctx, dockerCli, dclient, cclient, options.image, options.version, options.registryPrefix, options.configFile)
```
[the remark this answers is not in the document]
cpuguy83: Also wy are we passing both client and dockerCli? There's a few places this is done.

cli/command/registry/login.go
```diff
@@ -126,7 +130,13 @@ func runLogin(dockerCli command.Cli, opts loginOptions) error { //nolint: gocycl
 
 		response, err = clnt.RegistryLogin(ctx, *authConfig)
 		if err != nil {
-			return err
+			if strings.Contains(err.Error(), "annot connect to ") {
```
cpuguy83: Can we get a real error type?

cli/containerdutils/client_test.go
```diff
@@ -0,0 +1,370 @@
...
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/swarm"
+	prototypes "github.com/gogo/protobuf/types"
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+type (
+	fakeDockerClient struct {
```
cpuguy83: 🤕

cpuguy83: I'm not sure I see the purpose in these structs that do nothing except take functionality from other things.
Why not provide specific structs? If the interface you are trying to satisfy is too large, can't we narrow down the interface?

cli/containerdutils/containerd.go
```diff
@@ -0,0 +1,83 @@
...
+
+// IsContainerdPresent very simple routine to check to see if the containerd socket is present
+func IsContainerdPresent() bool {
+	// Check for presence first
+	_, err := os.Stat(ContainerdSockPath)
+	return err == nil
+}
+
+// GetContainerdClient verifies the containerd socket is present and writable by the current user
+func GetContainerdClient() (*containerd.Client, error) {
```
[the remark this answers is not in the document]
cpuguy83: Seems like we shouldn't bother with this.

cli/containerdutils/containerd.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Get a client to verify we can talk to it
+	return containerd.New(ContainerdSockPath)
+}
+
+// PullWithAuth will use the credential cache auth to pull via containerd
+func PullWithAuth(cli command.Cli, client WrappedContainerdClient, imageName string) error {
```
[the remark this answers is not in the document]
cpuguy83: This is all too much.
Let's consider what type should own this behavior.... e.g.  a wrapped client.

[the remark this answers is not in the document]
cpuguy83: Also seems like `cli` should be narrowed down to an interface for only what's used.

cli/containerdutils/hostpaths_test.go
```diff
@@ -0,0 +1,183 @@
...
+		},
+	}
+	client := &fakeContainerdClient{
+		newContainerFunc: func(ctx context.Context, id string, opts ...containerd.NewContainerOpts) (containerd.Container, error) {
+
+			return container, nil
+		},
+	}
+	err := VerifyHostDirExists(ctx, client, dirname, image)
+	assert.ErrorContains(t, err, "wait failure")
```
cpuguy83: Can we make sure to provide real typed errors for things instead of requiring checking error messages?

cli/containerdutils/types.go
```diff
@@ -0,0 +1,144 @@
...
+	"context"
+
+	"github.com/containerd/containerd"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/swarm"
+	ver "github.com/hashicorp/go-version"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+var (
```
cpuguy83: const

cli/containerdutils/types.go
```diff
@@ -0,0 +1,144 @@
...
+	EngineContainerName = "dockerd"
+
+	// CommunityEngineImage is the repo name for the community engine
+	CommunityEngineImage = "engine-community"
+
+	// EnterpriseEngineImage is the repo name for the enterprise engine
+	EnterpriseEngineImage = "engine-enterprise"
+
+	// EngineSpec is the default container spec for the docker engine
+	EngineSpec = specs.Spec{
```
cpuguy83: Seems like this should be a function to get the default spec (since that's the only way it could be immutable)...
This will probably work, though.

cli/containerdutils/types.go
```diff
@@ -0,0 +1,144 @@
...
+				Destination: "/usr/libexec",
+				Type:        "bind",
+			},
+		},
+	}
+)
+
+type (
+	// WrappedContainerdClient abstracts the containerd client to aid in testability
+	WrappedContainerdClient interface {
```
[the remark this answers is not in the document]
cpuguy83: Should also be scoped to the call site, not the package.

cli/licenseutils/client_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"context"
+
+	"github.com/docker/licensing"
+	"github.com/docker/licensing/model"
+)
+
+type (
+	fakeLicensingClient struct {
```
cpuguy83: Same comment as above for other of these `fake` clients.

cli/licenseutils/types.go
```diff
@@ -0,0 +1,41 @@
...
+	LicensingPublicKey = "LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0Ka2lkOiBKN0xEOjY3VlI6TDVIWjpVN0JBOjJPNEc6NEFMMzpPRjJOOkpIR0I6RUZUSDo1Q1ZROk1GRU86QUVJVAoKTUlJQ0lqQU5CZ2txaGtpRzl3MEJBUUVGQUFPQ0FnOEFNSUlDQ2dLQ0FnRUF5ZEl5K2xVN283UGNlWSs0K3MrQwpRNU9FZ0N5RjhDeEljUUlXdUs4NHBJaVpjaVk2NzMweUNZbndMU0tUbHcrVTZVQy9RUmVXUmlvTU5ORTVEczVUCllFWGJHRzZvbG0ycWRXYkJ3Y0NnKzJVVUgvT2NCOVd1UDZnUlBIcE1GTXN4RHpXd3ZheThKVXVIZ1lVTFVwbTEKSXYrbXE3bHA1blEvUnhyVDBLWlJBUVRZTEVNRWZHd20zaE1PL2dlTFBTK2hnS1B0SUhsa2c2L1djb3hUR29LUAo3OWQvd2FIWXhHTmw3V2hTbmVpQlN4YnBiUUFLazIxbGc3OThYYjd2WnlFQVRETXJSUjlNZUU2QWRqNUhKcFkzCkNveVJBUENtYUtHUkNLNHVvWlNvSXUwaEZWbEtVUHliYncwMDBHTyt3YTJLTjhVd2dJSW0waTVJMXVXOUdrcTQKempCeTV6aGdxdVVYYkc5YldQQU9ZcnE1UWE4MUR4R2NCbEp5SFlBcCtERFBFOVRHZzR6WW1YakpueFpxSEVkdQpHcWRldlo4WE1JMHVrZmtHSUkxNHdVT2lNSUlJclhsRWNCZi80Nkk4Z1FXRHp4eWNaZS9KR1grTEF1YXlYcnlyClVGZWhWTlVkWlVsOXdYTmFKQitrYUNxejVRd2FSOTNzR3crUVNmdEQwTnZMZTdDeU9IK0U2dmc2U3QvTmVUdmcKdjhZbmhDaVhJbFo4SE9mSXdOZTd0RUYvVWN6NU9iUHlrbTN0eWxyTlVqdDBWeUFtdHRhY1ZJMmlHaWhjVVBybQprNGxWSVo3VkQvTFNXK2k3eW9TdXJ0cHNQWGNlMnBLRElvMzBsSkdoTy8zS1VtbDJTVVpDcXpKMXlFbUtweXNICjVIRFc5Y3NJRkNBM2RlQWpmWlV2TjdVQ0F3RUFBUT09Ci0tLS0tRU5EIFBVQkxJQyBLRVktLS0tLQo="
+)
+
+type (
```
cpuguy83: I'm pretty sure I've never seen this done.... not that it shouldn't be done, just that it's not idiomatic in the sense that none of the other code does this.
Can we stick to `type Foo T`?

vendor.conf
```diff
@@ -85,4 +85,15 @@ k8s.io/client-go kubernetes-1.11.0
 k8s.io/kube-openapi d8ea2fe547a448256204cfc68dfee7b26c720acb
 k8s.io/kubernetes v1.11.0
 vbom.ml/util 256737ac55c46798123f754ab7d2c784e2c71783
-
+github.com/containerd/fifo 3d5202a
```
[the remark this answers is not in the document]
cpuguy83: Also, is this used? Don't see why we'd need fifo here.

cpuguy83:
It would be nice if there was an issue, or I suppose a comment in this PR, describing this change (what new commands are, what are they for, demo output, etc).

cli/command/engine/activate.go
```diff
@@ -0,0 +1,172 @@
...
+
+	return cmd
+}
+
+func runActivate(cli command.Cli, options activateOptions) error {
+	ctx := namespaces.WithNamespace(context.Background(), containerdutils.EngineNamespace)
```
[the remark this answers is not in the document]
crosbymichael: no, its not a client method

cli/command/engine/check.go
```diff
@@ -0,0 +1,123 @@
+package engine
+
+import (
+	"fmt"
+
+	"golang.org/x/net/context"
```
crosbymichael: it's just "context" now

cli/command/engine/check.go
```diff
@@ -0,0 +1,123 @@
...
+		return fmt.Errorf("Unable to access local containerd: %s", err)
+	}
+	ctx := namespaces.WithNamespace(context.Background(), containerdutils.EngineNamespace)
+	currentEngine, currentVersion, err := containerdutils.GetCurrentEngineVersion(ctx, cclient)
+	if err != nil {
+		return err
+	}
+
+	imageName := options.registryPrefix + "/" + currentEngine
+	versions, err := containerdutils.GetEngineVersions(context.Background(), dockerCli, currentVersion, imageName, options.all)
```
crosbymichael: pass ctx instead of context.Background() here

cli/containerdutils/containerd.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/namespaces"
+	"github.com/containerd/containerd/remotes/docker"
+	"github.com/docker/distribution/reference"
+)
+
+// IsContainerdPresent very simple routine to check to see if the containerd socket is present
+func IsContainerdPresent() bool {
```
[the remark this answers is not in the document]
crosbymichael: THere is a Health method if you already have a client

cli/containerdutils/containerd.go
```diff
@@ -0,0 +1,83 @@
...
+	// Check for presence first
+	_, err := os.Stat(ContainerdSockPath)
+	return err == nil
+}
+
+// GetContainerdClient verifies the containerd socket is present and writable by the current user
+func GetContainerdClient() (*containerd.Client, error) {
+	// Check for presence first
+	_, err := os.Stat(ContainerdSockPath)
```
crosbymichael: tisk tisk ;), connect and handle errors on connect os.IsNotExist, not check first.  These types of checks are racy

cli/containerdutils/containerd.go
```diff
@@ -0,0 +1,83 @@
...
+	authConfig := imgRefAndAuth.AuthConfig()
+	resolver := docker.NewResolver(docker.ResolverOptions{
+		Credentials: func(string) (string, string, error) {
+			return authConfig.Username, authConfig.Password, nil
+		},
+	})
+	fmt.Fprintf(cli.Out(), "Pulling %s... ", imageName)
+	ctx := namespaces.WithNamespace(context.Background(), EngineNamespace)
+
+	_, err = client.Pull(ctx, imageName, containerd.WithResolver(resolver), containerd.WithPullUnpack)
```
crosbymichael: If you want progress, you can use `content.Fetch` from containerd

cli/command/engine/activate.go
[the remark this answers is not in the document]
vdemeester: I meant the containers client in this repository, not in the containers project

dhiltgen:
> shouldn't we re-use the same "output formatting" when pulling images, etc… right now, it's telling me it's pulling something, I have no idea of the progress… We could use docker or containerd progress formatting.

Good suggestion.  Let me take a look at what it would take to wire that up.

> should /etc/docker path be configurable ? For example, my system's /etc/ is read-only (managed by nixos), and I want to test/run this PR, couldn't I be able to specify where it's gonna write the configuration ?

Yes, that was the intention but it seems you found a bug - I'll investigate this further and figure out what's going wrong and add a test case for this.


> it seems there is a need to login before having a docker daemon, and this is a problem as docker login needs a deamon — on this, we may want to ask for login(s) whenever the command the require them (listing license, …), and later on issuing a docker login on the newly bootstrapped daemon (programatically). The login client-side only seems hackish to me.

If you're just deploying a community engine from hub it works without a login.  If you're using an alternative registry (e.g., a DTR behind a firewall) then you might need to login to pull images.  If you're initializing an enterprise engine from hub you'll need to be logged in as well.  I tried to wire this up so it's not mandatory, but so we can support initialization from a private repo if necessary.  Does that help address your concern, or are you still concerned with this model?

> On update it's possible to change the version and the registry-prefix but not the engine-image, what is the rationale for that ?

Good point.  That wasn't intentional. I forgot to add the image flag to the update flows when I added it recently to the init flow.

> It would be really good to have unit tests on the engine package

I'll see what I can do. :-)

> I feel like containerdutils package should be some sort of "wrapper" client that get's created with our "view" of docker and containerd client. That way we wouldn't need to pass them for each and every exported functions.

Let me coordinate with @crosbymichael and @dmcgowan and see what we can come up with.

> In general, command.Cli should be only used in cmd/* packages.. having a dependency on it in the containerdutils package seems like a smell…

:+1:  I'll clean this up.

> This is a public facing repository, that is easily use as library… and in this context, any exported element (function, struct) should be considered stable.. So the less we exported at first, the better… This is not the case right now.

Will do  - I'll reduce the set of exported components in my next refactoring pass.

> github.com/docker/licensing needs some tests, and it would be very handy if it were using either vndr or golang/dep tool

\cc @mason-fish ^^

cli/command/engine/check.go
dhiltgen: Thanks - I'll clean this up - I started this work quite a while back and I think the CLI was still using that pattern when I started... just overlooked switching these as I rebased.

cli/command/engine/check.go
```diff
@@ -0,0 +1,123 @@
...
+		return err
+	}
+
+	availUpdates := []updates.Update{}
+
+	availUpdates = append(availUpdates, updates.Update{
+		Type:    "current",
+		Version: currentVersion,
+	})
+	if len(versions.Patches) == 0 {
```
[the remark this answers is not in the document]
dhiltgen: I was chatting with @thaJeztah and realized this pattern isn't going to work for a couple reasons.  I'll remove the logic to try to be smart about EOL reporting, and we can revisit that based on some other metadata source in the future.

cli/command/engine/cmd.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// NewEngineCommand returns a cobra command for `engine` subcommands
+func NewEngineCommand(dockerCli command.Cli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:         "engine COMMAND",
+		Short:       "Manage the docker engine",
+		Args:        cli.NoArgs,
+		RunE:        command.ShowHelp(dockerCli.Err()),
+		Annotations: map[string]string{"version": "1.36"},
```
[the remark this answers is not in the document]
dhiltgen: Hmm... I don't remember how this got in there - copy-pasta probably.  I'll remove it.

cli/command/formatter/licenses.go
```diff
@@ -0,0 +1,103 @@
...
+	case TableFormatKey:
+		if quiet {
+			return defaultSubscriptionsQuietFormat
+		}
+		return defaultSubscriptionsTableFormat
+	case RawFormatKey:
+		if quiet {
+			return `license_state: {{.State}}`
+		}
+		// TODO - wrong
```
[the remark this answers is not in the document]
dhiltgen: Final UX on the licenses is still WIP... hopefully I can get some more clarity for my next refactoring pass.

cli/command/registry/login.go
```diff
@@ -51,6 +51,10 @@ func NewLoginCommand(dockerCli command.Cli) *cobra.Command {
 	flags.StringVarP(&opts.password, "password", "p", "", "Password")
 	flags.BoolVarP(&opts.passwordStdin, "password-stdin", "", false, "Take the password from stdin")
 
+	// TODO - this doesn't work if you can't talk to a remote daemon, so it's broken for
```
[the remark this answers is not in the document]
dhiltgen: Oops, this comment is stale.  The code below solves the problem of login without a remote daemon but I forgot to remove the comment.

cli/command/registry_test.go
```diff
@@ -78,6 +79,7 @@ func TestElectAuthServer(t *testing.T) {
 			},
 		},
 	}
+	debug.Enable()
```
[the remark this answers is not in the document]
dhiltgen: I changed the warning messages to only be when debug was enabled so it wouldn't yell at you when you were going through the init flow by default, and the test case was built to look for those messages.

cli/containerdutils/client_test.go
```diff
@@ -0,0 +1,370 @@
+package containerdutils
```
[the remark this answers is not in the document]
dhiltgen: Hmm... this wasn't intended to emulate or wrap the existing containerd client, but rather be utilities for managing an engine running on top of containerd.  I'll work with @crosbymichael to try to come up with a better model for some of this, but I'm a little uneasy calling this "containerd" as it's most certainly not the same as `import "github.com/containerd/containerd"`  Maybe "containerizedengineutils" is more descriptive, but seems a bit long to me.

cli/containerdutils/containerd.go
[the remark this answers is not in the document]
dhiltgen: At one point I had this wired up in other places, but you're right, it's not used now.  I think it can go away.

cli/containerdutils/containerd.go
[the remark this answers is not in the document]
dhiltgen: Yup, realizing now this is ~pointless.  Nuking it.

cli/containerdutils/versions.go
```diff
@@ -0,0 +1,90 @@
...
+			testSegments[1] == currentSegments[1] {
+			patches = append(patches, testVersion)
+		} else {
+			upgrades = append(upgrades, testVersion)
+		}
+	}
+	sort.Sort(collection(downgrades))
+	sort.Sort(collection(patches))
+	sort.Sort(collection(upgrades))
+	for _, ver := range downgrades {
```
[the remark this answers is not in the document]
dhiltgen: I think there was a reason I didn't do that originally, but I can't remember why now.   I'll make the switch.

cli/licenseutils/types.go
```diff
@@ -0,0 +1,41 @@
...
+	HubUser struct {
+		client licensing.Client
+		token  string
+		User   model.User
+		Orgs   []model.Org
+	}
+
+	// LicenseDisplay stores license details in a human readable form
+	// for better display in the CLI
+	LicenseDisplay struct {
```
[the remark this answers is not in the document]
dhiltgen: I think we're probably going to look at creating an abstracted type in the underlying library so I think this will go away, but if not, I'll rename it.

cli/licenseutils/utils.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// ApplyLicense will store a license on the local system
+func ApplyLicense(ctx context.Context, cli command.Cli, license *model.IssuedLicense) error {
+	dclient := cli.Client()
+	info, err := dclient.Info(ctx)
+	if err != nil {
+		return err
+	}
+	return licensing.StoreLicense(ctx, dclient, license, info.DockerRootDir)
```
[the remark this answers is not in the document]
dhiltgen: :100: - https://github.com/docker/licensing/blob/master/storage.go#L105

cli/registry/client/client.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	manifesttypes "github.com/docker/cli/cli/manifest/types"
+	"github.com/docker/cli/cli/trust"
```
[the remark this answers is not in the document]
dhiltgen: It is on line https://github.com/docker/cli/pull/1260/files#diff-cb6e6f1ab385cb866626ee63bfa1a273R206

cli/updates/types.go
```diff
@@ -0,0 +1,8 @@
+package updates
```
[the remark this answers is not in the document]
dhiltgen: I'll try to find a better home for it - initial placement lead to circular imports so I plopped it here.

e2eengine/multi/main_test.go
```diff
@@ -0,0 +1,10 @@
+package check
+
+import (
+	"os"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
```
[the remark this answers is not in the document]
dhiltgen: Good point - it used to do more but got whittled down while I was refactoring.

vendor/github.com/docker/licensing/accounts.go
```diff
@@ -0,0 +1,34 @@
+package licensing
+
+import (
+	"context"
+
+	"github.com/docker/licensing/lib/go-clientlib"
+	"github.com/docker/licensing/model"
+)
+
+func (c *client) createAccount(ctx context.Context, dockerID string, request *model.AccountCreationRequest) (response *model.Account, err error) {
```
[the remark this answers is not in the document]
dhiltgen: \cc @mason-fish

vendor/github.com/docker/licensing/accounts.go
```diff
@@ -0,0 +1,34 @@
...
+	response = new(model.Account)
+	_, _, err = c.doReq(ctx, "PUT", &url, clientlib.SendJSON(request), clientlib.RecvJSON(response))
+	if err != nil {
+		return nil, err
+	}
+
+	return
+}
+
+func (c *client) getAccount(ctx context.Context, dockerID string) (response *model.Account, err error) {
```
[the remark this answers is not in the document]
dhiltgen: \cc @mason-fish

vendor/github.com/docker/licensing/client.go
```diff
@@ -0,0 +1,252 @@
...
+	GetHubUserByName(ctx context.Context, username string) (user *model.User, err error)
+	VerifyLicense(ctx context.Context, license model.IssuedLicense) (res *model.CheckResponse, err error)
+	GenerateNewTrialSubscription(ctx context.Context, authToken, dockerID, email string) (subscriptionID string, err error)
+	ListSubscriptions(ctx context.Context, authToken, dockerID string) (response []*model.SubscriptionDetail, err error)
+	DownloadLicenseFromHub(ctx context.Context, authToken, subscriptionID string) (license *model.IssuedLicense, err error)
+	ParseLicense(license []byte) (parsedLicense *model.IssuedLicense, err error)
+	StoreLicense(ctx context.Context, dclnt WrappedDockerClient, licenses *model.IssuedLicense, localRootDir string) error
+}
+
+func (c *client) LoginViaAuth(ctx context.Context, username, password string) (authToken string, err error) {
```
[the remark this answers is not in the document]
dhiltgen: \cc @mason-fish

vendor/github.com/docker/licensing/client.go
```diff
@@ -0,0 +1,252 @@
...
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/docker/libtrust"
```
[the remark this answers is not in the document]
dhiltgen: \cc @mason-fish

cli/containerdutils/engine.go
```diff
@@ -0,0 +1,282 @@
...
+			Options: []string{
+				"rbind",
+				"rw",
+				"rshared",
+			},
+			Source:      rootDir,
+			Destination: rootDir,
+			Type:        "bind",
+		},
+		/* Temporarily commented out as /etc is already mounted to workaround UCP install bugs
```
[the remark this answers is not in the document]
dhiltgen: https://github.com/docker/cli/pull/1260/files#diff-b275e7916c4ffc73030fb93fa5189b2eR62

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/cio"
+	"github.com/docker/docker/pkg/stringid"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// VerifyHostDirExists ensures that the desired directory exists on the host
+// nolint: gocyclo
+func VerifyHostDirExists(ctx context.Context, client WrappedContainerdClient, dirname string, image containerd.Image) error {
```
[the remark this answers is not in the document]
dhiltgen: Still noodling on exactly how, but I think this'll get refactored a bit more so it's more generally useful, and I think we will want it to be a public interface.  (Most likely following the WithXXX and Opts patterns used by other containerd client helpers)

cli/containerdutils/hostpaths.go
```diff
@@ -0,0 +1,294 @@
...
+		// Path exists, kill the task and proceed to clean up
+		err = task.Kill(ctx, SIGKILL)
+		if err != nil && !strings.Contains(err.Error(), "not found") {
+			return fmt.Errorf("failed to kill task: %s", err)
+		}
+		exitStatus, err := task.Wait(ctx)
+		if err != nil {
+			return fmt.Errorf("Failed to wait for container task: %s", err)
+		}
+		<-exitStatus
```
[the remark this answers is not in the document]
dhiltgen: Perhaps, but the only thing that matters is the task was created, since mount validation is done at task creation time.  If the path doesn't exist, the task create fails.

dhiltgen:
> Functionality needs ownership. Right now all functionality is scoped to a package when it should be scoped to a type.

:+1:

> Errors, especially in new functionality that's added here, should be well typed so they can be properly inspected (e.g. not by checking the message itself).

:+1:

> Interfaces are quite broad, both in the new interfaces being used and the interfaces being passed in to the new functionality. This should be scoped down so, for instance, a function that just needs an io.Writer doesn't need the whole docker CLI object.

:+1:

> The fakeClient approach where there is a struct that does nothing but store configured functions is not very clean and largely redundant. Might as well just make new objects with the needed functionality. Note you can have a base struct that provides "not implemented" sort of behavior and wrap that with specific functionality for what's being tested... but also scoping interfaces down should help a lot here.

I thought I was following existing patterns, but if there's a different preferred model I can adjust the implementation.  @cpuguy83 can you point to example's as reference in the code of how you'd like me to structure the test adapters?

cli/command/formatter/licenses.go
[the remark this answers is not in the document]
dhiltgen: \cc @mason-fish - keep me  posted on how you're progressing on the abstracted type for license information as that'll drive this UX cleanup.

cli/licenseutils/types.go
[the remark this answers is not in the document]
dhiltgen: \cc @mason-fish

cpuguy83:
>  can you point to example's as reference in the code of how you'd like me to structure the test adapters

If you need to implement many methods, then I'd use something like this:

```go
type FooBar interface {
    Foo() error
    Bar() error
}
type baseFoo struct {}

func (baseFoo) Foo() error { return errors.New("not implemented") }

func(baseFoo) Bar() error { return errors.New("not implemented") }

type withFoo struct {
    baseFoo
}

func(withFoo) Foo() error {
    return nil
}
```

But I would also try and reduce the things I need to mock out as much as possible, which means defining interfaces locally for the methods you are actually calling rather than using a concrete type or some (large) external interface.

vendor.conf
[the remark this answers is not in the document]
dhiltgen: I use the containerd/cio package which needs it.

cli/containerdutils/client_test.go
dhiltgen: It's purely a testing abstraction.  I looked around in this repo to find a pattern to emulate with the hope that I'd be following the preferred style and saw https://github.com/docker/cli/blob/master/cli/command/manifest/client_test.go which uses this model.

@cpuguy83 can you point me to a different/better reference you would prefer?

If this model is explicitly not what we want people to follow, could a maintainer put a comment in those files noting that those tests should be refactored so others don't fall into the same trap I did thinking this pattern is OK?

dhiltgen:
> It would be nice if there was an issue, or I suppose a comment in this PR, describing this change (what new commands are, what are they for, demo output, etc).

The following is a little dated and some changes have been made to the implementation based on UX feedback, but it shows the basic flow.

# Problem Abstract

The Docker EE platform is distributed and the software lifecycle
managed using two different patterns today.  This inconsistency leads
to customer confusion, operational inefficiencies, and in some cases
avoidable outages due to user mistakes.

This PR enables the docker engine to run as a privileged container running
on top of containerd.  This model also enables a smooth upgrade path from
CE to EE engines for users who choose to do so.  

# UX Example

The following examples outline how users would interact with the local
engine via the CLI.

### Before Initialization

This new version of the docker CLI would be capable of initializing
a local engine if containerd was detected via the socket or
named pipe.

```
% docker info
Cannot connect to the Docker daemon at unix:///var/run/docker.sock however containerd is running locally
- Set DOCKER_HOST or use the `-H` flag to point to a remote docker engine or cluster
- To initialize a local node, run `sudo docker engine init`
```


### Initializing a Local CE Engine

Provided containerd is present on the local system and the user has the
ability to write to the containerd socket (or named pipe), the CLI would
be capable of starting up the docker daemon.

```
% docker engine init --help

Usage:  docker engine init [OPTIONS]

Initialize a docker engine on the local containerd

Options:
      --config-file string       Specify the location of the daemon configuration file
                                 on the host (default "/etc/docker/daemon.json")
      --engine-version string    Specify engine version (default '18.09.0')
      --registry-prefix string   Override the default location where engine images are
                                 pulled (default "docker.io/docker")
      --root-dir string          Specify the location of the Docker Root Dir on the host
                                 (default "/var/lib/docker")
```
```
% sudo docker engine init
Pulling docker.io/dockereng/ce-engine:18.09.0... done.
Waiting for engine to start... waiting for engine to be responsive... engine 18.09.0 is online.
Success!  The docker engine is now running.
```

Notes:
* Tag conventions
  * CLI builds default to install their corresponding engine version
  * `YY.MM.P` - Immutable version specific tag
  * `nightly-YYMMDD` - maps to an nightly edge build
* helper packages would be generated to handle bootstrapping the
engine via OS level packages to mimic the current packaging behavior, but
those would leverage these CLI commands under the covers in post-install
packaging scripts.
* If the image is already present, it will not be pulled, so offline use-cases
can be supported by pre-loading the image via OS level packaging.

### Activating EE

If the user wants to start running an EE engine they would use the
following command to activate an EE engine with applicable license.
This can be performed after `init`ing a CE engine, or can be done directly
on top of containerd as the first step.  If they already have CE running
and workloads on top, those would be preserved.


```
% docker engine activate --help

Usage:  docker engine activate [OPTIONS]

With this command you may apply an existing Docker enterprise license, or
interactively download one from Docker.  In the interactive exchange, you can
sign up for a new trial, or download an existing license.  If you are currently
running a Community Edition engine, the daemon will be updated to the
Enterprise Edition Docker engine with additional capabilities and long term
support.

For more information about different Docker Enterprise license types visit
http://www.docker.com/licenses

For non-interactive scriptable deployments, download your license from
https://hub.docker.com/ then specify the file with the '--license' flag.

Options:
      --engine-version string    Specify engine version (default is to use existing version)
  -l, --license string           License File
      --registry-prefix string   Override the default location where engine images are pulled (default
                                 "docker.io/dockereng")
```

The following example shows the UX for a brand-new customer who has never logged into hub
```
% docker engine activate
Please run 'docker login' with your existing account then re-run 'docker engine activate'
```

```
% docker engine activate
You don't have any existing Docker licenses
0) Generate a new trial license for EE Basic (individual engines)
1) Generate a new trial license for EE Advanced (cluster support)
Please pick the license above by number (0): 0

Pulling docker.io/dockereng/ee-engine:18.09.0... done.
Starting the docker engine... waiting for engine to be responsive... engine 18.09.0 is online.
Success!  Your engine has been activated.
```

Notes:
* The users hub account would be granted access to the EE engines during their trial
* The license would be stored locally and may be used to activate specific license enabled features
* Tag conventions
  * Not visible unless you've paid, or have a trial license
  * `$VER` - Immutable version specific tag


The following variation shows the UX for an existing customer with some **existing licenses**.

```
% docker engine activate

Checking Docker Hub for licenses for dhiltgen... Licenses found.
0) Docker EE Basic- Sat Oct 14 2017
1) Docker EE Advanced, 10 nodes- Sat Oct 14 2017
Please pick the license above by number (0): 1

Pulling docker.io/dockereng/ee-engine:18.09.0... done.
Starting the docker engine... waiting for engine to be responsive... engine 18.09.0 is online.
Success!  Your engine has been activated.
See https://docs.docker.com/ee/ucp/admin/install/ for UCP installation instructions.
```

### Seeing current engine status

The info and version commands should be updated to expose if the engine
is a CE or EE engine, and the status of the license (expired, type of
license, etc.)

```
% docker info
Containers: 11
 Running: 0
 Paused: 0
 Stopped: 11
Images: 703
...
License: Community Edition Engine
```
```
License: Enterprise Standard trial - 20 days remaining.  To purchase go to http://hub.docker.com/renew
```
```
License: Enterprise Standard 10 nodes - expires 1/1/2019
```
```
License: EXPIRED!  You will no longer receive updates.  Please renew at http://hub.docker.com/renew
```

Notes:
* Once expired, the users access to EE patches would be disabled on the hub side.

### Detecting Updates are Available

Checking for updates requires scanning hub repos for new tags, and requires
hub credentials, so this should be done using an explicit command, not included
in existing commands like `info`.

```
% docker engine check --help

Usage:  docker engine check [OPTIONS]

Check for available updates

Options:
      --all                      Report all versions (older and pre-release versions)
      --patch-only               Only report available patch versions
      --registry-prefix string   Override the default location where engine images are
                                 pulled (default "docker.io/dockereng")
```
```
% docker engine check
TYPE        VERSION         RELEASE NOTES
current     18.03.0
patch       18.03.1         http://docs.docker.com/releases/18.03.1-ee.releasenotes
patch       18.03.2         http://docs.docker.com/releases/18.03.2-ee.releasenotes
upgrade     18.09.0         http://docs.docker.com/releases/18.09.0-ee.releasenotes
```

```
% docker engine check
(WARNING: your version is no longer supported, consider upgrading)
TYPE        VERSION         RELEASE NOTES
current     18.02.0
upgrade     18.09.0         http://docs.docker.com/releases/18.09.0-ee.releasenotes
```


Notes:
* If the EE license has expired, CE updates should be shown to allow "reverting" back to CE
if you decide to

### Updating a Local Engine

```
% docker engine update --help

Usage:  docker engine update [OPTIONS]

Update the local docker engine

Options:
      --engine-version string    Specify engine version - by default this command will update to the latest
                                 patch for the current engine version
      --registry-prefix string   Override the default location where engine images are
                                 pulled (default "docker.io/docker")
```
```
% docker engine update --engine-version 18.03.2
Using tag: 18.03.2
latest: Pulling from docker/ee-engine
57310166fe88: Pull complete
Digest: sha256:1669a6aa7350e1cdd28f972ddad5aceba2912f589f19a090ac75b7083da748db
Updating Local Docker Engine...
Docker Engine is ready.
```

dhiltgen:
> If you need to implement many methods, then I'd use something like this:

Hmm... maybe I'm just not seeing it, but isn't this pattern going to result in a massive explosion in the amount of code necessary to mock out various different scenarios?  Wouldn't I then have to create unique types for each scenario?  That seems pretty heavy weight.  The nice thing about the existing pattern I replicated is it only gets involved in the test code, not the mainline production code, and each test case can easily alter the behavior of any routine.

Do you have a good example in the tree that's following your pattern for test cases I can look more closely at to see how it should map?

cli/command/engine/activate.go
```diff
@@ -0,0 +1,176 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runActivate(dockerCli, options)
+		},
+	}
+
+	flags := cmd.Flags()
+
+	flags.StringVar(&options.licenseFile, "license", "", "License File")
+	flags.StringVar(&options.version, "version", "", "Specify engine version (default is to use existing version)")
+	flags.StringVar(&options.registryPrefix, "registry-prefix", "docker.io/docker", "Override the default location where engine images are pulled")
```
seemethere: Wouldn't this make more sense to have it as just `docker.io` and let the `engine-image` contain the prefix for the registry organization?

seemethere: This applies to all other uses of the `registry-prefix`

cli/command/engine/init.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+
+	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/containerizedengine"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var (
```
silvin-lubecki: const?

cli/command/engine/init.go
```diff
@@ -0,0 +1,64 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runInit(dockerCli, options)
+		},
+		Annotations: map[string]string{"experimentalCLI": ""},
+	}
+	flags := cmd.Flags()
+	flags.StringVar(&options.version, "version", defaultEngineVersion, "Specify engine version")
+	flags.StringVar(&options.image, "engine-image", containerizedengine.CommunityEngineImage, "Specify engine image")
+	flags.StringVar(&options.registryPrefix, "registry-prefix", "docker.io/docker", "Override the default location where engine images are pulled")
+	flags.StringVar(&options.configFile, "config-file", "/etc/docker/daemon.json", "Specify the location of the daemon configuration file on the host")
```
silvin-lubecki: Hmmm what about windows 🤔 ? Should this feature work with containerd on windows? If so should the default path be specified depending the platform? WDYT @thaJeztah @vdemeester ?

cli/command/engine/activate.go
```diff
@@ -0,0 +1,176 @@
...
+
+	return client.ActivateEngine(ctx, opts, cli.Out(), authResolver(cli))
+}
+
+func getLicenses(ctx context.Context, cli command.Cli, options activateOptions) (*model.IssuedLicense, error) {
+	user, err := licenseutils.Login(ctx, cli, options.registryPrefix)
+	if err != nil {
+		return nil, err
+	}
+	fmt.Fprintf(cli.Out(), "Looking for existing licenses for %s...", user.User.Username)
```
silvin-lubecki: Add a `\n` at the end?

cli/command/engine/init.go
```diff
@@ -0,0 +1,64 @@
...
+
+type initOptions struct {
+	version        string
+	image          string
+	registryPrefix string
+	configFile     string
+}
+
+func newInitCommand(dockerCli command.Cli) *cobra.Command {
+	options := initOptions{}
```
silvin-lubecki: You could use directly a `containerizedengine.EngineInitOptions` ?

cli/command/engine/update.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		options.image = currentEngine
+	}
+	opts := containerizedengine.EngineInitOptions{
+		RegistryPrefix: options.registryPrefix,
+		EngineImage:    options.image,
+		EngineVersion:  options.version,
+	}
+
+	err = client.DoUpdate(ctx, opts, dockerCli.Out(), authResolver(dockerCli))
```
silvin-lubecki: if err := ...; err != nil{

cli/command/engine/update.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+
+	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/containerizedengine"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+type updateOptions struct {
```
silvin-lubecki: Use a `containerizedengine.EngineInitOptions` directly?

cli/command/engine/update.go
```diff
@@ -0,0 +1,66 @@
...
+		RegistryPrefix: options.registryPrefix,
+		EngineImage:    options.image,
+		EngineVersion:  options.version,
+	}
+
+	err = client.DoUpdate(ctx, opts, dockerCli.Out(), authResolver(dockerCli))
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(dockerCli.Out(), "Success!  The docker engine is now running.\n")
```
silvin-lubecki: nit: fmt.Fprintln(dockerCli.Out(), "Success!  The docker engine is now running.")

cli/command/formatter/licenses.go
```diff
@@ -0,0 +1,103 @@
+package formatter
+
+import (
+	"time"
+
+	"github.com/docker/cli/cli/licenseutils"
+)
+
+// TODO - derived from plugin.go - replicate plugin_tests.go before PR
```
silvin-lubecki: Is this TODO comment still needed?

cli/command/formatter/licenses.go
```diff
@@ -0,0 +1,103 @@
...
+
+const (
+	defaultSubscriptionsTableFormat = "table {{.Num}}\t{{.Owner}}\t{{.ProductID}}\t{{.Expires}}\t{{.Components}}"
+	defaultSubscriptionsQuietFormat = "{{.Num}}:{{.ProductID}}"
+
+	numHeader                 = "NUM"
+	licenseSubscriptionHeader = "SUBSCRIPTION"
+	stateHeader               = "STATE"
+	productIDHeader           = "PRODUCTID"
```
silvin-lubecki: [Other headers](https://github.com/docker/cli/blob/master/cli/command/formatter/task.go#L20) seems to separate words with a space. So it should be `PRODUCT ID` and "RATE PLAN" ?

cli/command/formatter/licenses.go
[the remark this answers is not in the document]
silvin-lubecki: Can you maybe be more precise in your TODO comment? If it stays here, at least it will be more understandable.

cli/command/formatter/licenses.go
```diff
@@ -0,0 +1,103 @@
...
+func (c *licenseContext) Num() int {
+	return c.l.Num
+}
+func (c *licenseContext) Subscription() string {
+	return c.l.Subscription
+}
+func (c *licenseContext) ProductID() string {
+	return c.l.ProductID
+}
+
```
silvin-lubecki: nit: empy line between every function

cli/command/formatter/updates.go
```diff
@@ -0,0 +1,73 @@
...
+const (
+	defaultUpdatesTableFormat = "table {{.Type}}\t{{.Version}}\t{{.Notes}}"
+	defaultUpdatesQuietFormat = "{{.Version}}"
+
+	updatesTypeHeader = "TYPE"
+	versionHeader     = "VERSION"
+	notesHeader       = "NOTES"
+)
+
+// NewUpdatesFormat returns a Format for rendering using a updates Context
```
silvin-lubecki: `using an update context`?

cli/command/formatter/updates.go
```diff
@@ -0,0 +1,73 @@
...
+		if quiet {
+			return `update_version: {{.Version}}`
+		}
+		return `update_version: {{.Version}}\ntype: {{.Type}}\nnotes: {{.Notes}}\n`
+	}
+	return Format(source)
+}
+
+// UpdatesWrite writes the context
+func UpdatesWrite(ctx Context, availUpdates []containerizedengine.Update) error {
```
silvin-lubecki: nit: availUpdates -> availableUpdates

cli/command/formatter/updates_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
+)
+
+func TestUpdateContextWrite(t *testing.T) {
+	cases := []struct {
+		context  Context
+		expected string
+	}{
+
```
silvin-lubecki: nit: empty line

cli/command/formatter/updates_test.go
```diff
@@ -0,0 +1,144 @@
...
+`,
+		},
+	}
+
+	for _, testcase := range cases {
+		updates := []containerizedengine.Update{
+			{Type: "updateType1", Version: "version1", Notes: "description 1"},
+			{Type: "updateType2", Version: "version2", Notes: "description 2"},
+		}
+		out := bytes.NewBufferString("")
```
silvin-lubecki: nit: `out := &bytes.Buffer{}`

cli/command/formatter/updates_test.go
```diff
@@ -0,0 +1,144 @@
...
+	updates := []containerizedengine.Update{
+		{Type: "updateType1", Version: "version1", Notes: "note1"},
+		{Type: "updateType2", Version: "version2", Notes: "note2"},
+	}
+	expectedJSONs := []map[string]interface{}{
+		{"Version": "version1", "Notes": "note1", "Type": "updateType1"},
+		{"Version": "version2", "Notes": "note2", "Type": "updateType2"},
+	}
+
+	out := bytes.NewBufferString("")
```
silvin-lubecki: Same as above

cli/command/registry/login.go
```diff
@@ -167,3 +173,18 @@ func loginWithCredStoreCreds(ctx context.Context, dockerCli command.Cli, authCon
...
+		return registrytypes.AuthenticateOKBody{}, err
+	}
+
+	status, token, err := svc.Auth(ctx, &auth, command.UserAgent())
+
+	return registrytypes.AuthenticateOKBody{
+		Status:        status,
+		IdentityToken: token,
+	}, err
+
```
silvin-lubecki: nit: empty line

cli/command/engine/activate.go
dhiltgen: I think this boils down to a question of use-case and when these flags will be used.  I was trying to optimize the UX around end-user use-cases to support local mirrors, while enabling developer use-cases.  

Specifically, if you're a typical user, I don't think you'll ever use `--engine-image`  - the default behavior for the community engine and enterprise engine scenarios will do the right thing.  This flag is really only meant for developers who want to deviate from the standard naming scheme.

Many customers run an internal registry in their firewalled environments, and the intention with the `--registry-prefix` flag was to give them a flag to establish the prefix where they store all their docker product mirrored images, both community and enterprise engines, and in the future, additional products like UCP and DTR itself.

My concern with splitting it the way you suggest is I think it makes it harder for end-users using local registry mirrors to benefit from the automatic community/enterprise logic.  These mirroring customer would always have to pin the exact image name to use a mirror.  In this version of the engine, we're using a single image so it seems ~simple, but it's entirely possible as we work on the modularization work in moby (buildkit, etc.) we will start to split the engine into multiple images in the future.  When/if we do so, then we'd have to add new flags for developers to select the other component images, but one could see this getting out of control for a typical user.  By using a common prefix and letting the CLI select the right image names by default using the common prefix we hide that complexity of multiple component images from the typical end user.  It also will make it harder for us to support this mirroring pattern for other products down the road. For example, UCP today doesn't support installing from a mirror, and UCP itself is made up of a dozen plus images already.  I'd like to see us follow this pattern so someday you can `docker ucp install --registry-prefix mylocalregistry.acme.com/docker ...` (or something along those lines) where the registry-prefix is common for all the mirrored content.

cli/containerizedengine/client_test.go
```diff
@@ -0,0 +1,395 @@
+package containerizedengine
```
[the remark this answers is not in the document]
dhiltgen: There are a lot of types/apis we use from the "real" `github.com/containerd/containerd` so I think naming it this way is going to lead to poorer readability and some funky renaming.  It doesn't feel right to me, but I'll go ahead and add a discrete commit to make this change and lets see what folks think.

cli/containerizedengine/types.go
```diff
@@ -0,0 +1,175 @@
...
+		opts EngineInitOptions,
+		out OutStream,
+		authResolver func(
+			ctx context.Context,
+			index *registrytypes.IndexInfo,
+		) types.AuthConfig) error
+	GetEngineVersions(ctx context.Context, registryClient registryclient.RegistryClient, currentVersion, imageName string, includePrereleases bool) (AvailableVersions, error)
+
+	GetEngine(ctx context.Context) (containerd.Container, error)
+	RemoveEngine(ctx context.Context, engine containerd.Container) error
```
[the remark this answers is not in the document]
dhiltgen: I ran out of time to build out the CLI UX for removing the containerized engine, but I think it's something we're going to need.  Instead of hiding these, it's probably better to try to bite that off now.  I'll make it experimental to map to the init being experimental.

cli/containerizedengine/client_test.go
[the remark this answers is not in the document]
vdemeester: @dhiltgen if this package doesn't leak those types (i.e. doesn't return them), it's should not be a problem. And really, this package should not leak any `containerd` type *above*…

pkg/containerized/snapshot.go
```diff
@@ -0,0 +1,164 @@
...
+	mounts    []mount.Mount
+}
+
+/*
+func (r *revision) Mounts() []mount.Mount {
+	return r.mounts
+}
+*/
+
+// nolint: interfacer
```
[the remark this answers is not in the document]
dhiltgen: Yeah, it wanted me to change the types to be incompatible with the upstream containerd pattern which I think is wrong.

cli/containerizedengine/engine.go
```diff
@@ -0,0 +1,268 @@
...
+		),
+		containerized.WithHostDirMount(rootDir, []string{"rbind", "rw", "rshared"}),
+		// Temporarily commented out as /etc is already mounted to workaround UCP install bugs
+		//containerized.WithHostDirMount(path.Dir(configFile), []string{"rbind", "rw", "rshared"}),
+
+		// TODO XXX Workaround so the UCP install flow will work (bug in dockerd validation steps)
+		containerized.WithHostDirMount("/var", []string{"rbind", "rw", "rshared"}),
+		containerized.WithHostDirMount("/etc", []string{"rbind", "rw", "rshared"}),
+		containerized.WithHostDirMount("/opt", []string{"rbind", "rw", "rshared"}),
+		containerized.WithHostDirMount("/usr/libexec", []string{"rbind", "rw", "rshared"}),
```
silvin-lubecki: Will all these mounts work on a Windows host?

cli/containerizedengine/hostpaths.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/containerd/containerd/cio"
+	"github.com/docker/cli/pkg/containerized"
+	"github.com/docker/docker/pkg/stringid"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+func (c baseClient) verifyDockerConfig(ctx context.Context, configFile string, image containerd.Image) (string, error) {
+
```
silvin-lubecki: nit: empty line

cli/containerizedengine/hostpaths.go
```diff
@@ -0,0 +1,133 @@
...
+				"rbind",
+				"rw",
+				"rshared",
+			},
+			Source:      path.Dir(configFile),
+			Destination: path.Dir(configFile),
+			Type:        "bind",
+		},
+	}
+	container, err := c.cclient.NewContainer(
```
silvin-lubecki: Extract all the code above to a sub function, like `runTesterContainer`

cli/containerizedengine/hostpaths.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to verify config file %s", configFile)
+	}
+	// Capture the output
+	buf := new(bytes.Buffer)
+	task, err := container.NewTask(ctx, defaultNewCreator(withByteBuffers(buf)))
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to verify %s", configFile)
+	}
+	err = task.Start(ctx)
```
silvin-lubecki: if err := ...; err != nil{

cli/containerizedengine/hostpaths.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "failed to start container task")
+	}
+	status, err := task.Wait(ctx)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to wait for container task")
+	}
+	<-status
+	err = task.Kill(ctx, SIGKILL)
```
silvin-lubecki: if err := ...; err != nil{

cli/containerizedengine/hostpaths.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		logrus.Debugf("Failed to kill task: %s", err)
+	}
+	_, err = task.Delete(ctx)
+	if err != nil {
+		logrus.Debugf("Failed to delete task: %s", err)
+	}
+	err = container.Delete(ctx, containerd.WithSnapshotCleanup)
```
silvin-lubecki: Shouldn't the `task.Delete` and the `container.Delete` somehow defered? If an error occurs, the task and the container will remain?

cli/containerizedengine/hostpaths.go
```diff
@@ -0,0 +1,133 @@
...
+	status, err := task.Wait(ctx)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to wait for container task")
+	}
+	<-status
+	err = task.Kill(ctx, SIGKILL)
+	if err != nil {
+		logrus.Debugf("Failed to kill task: %s", err)
+	}
+	_, err = task.Delete(ctx)
```
silvin-lubecki: if _, err := ... ; err != nil{

cli/containerizedengine/hostpaths.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		logrus.Debugf("Failed to delete task: %s", err)
+	}
+	err = container.Delete(ctx, containerd.WithSnapshotCleanup)
+	if err != nil {
```
silvin-lubecki: if err := ... ; err != nil{

cli/containerizedengine/hostpaths.go
```diff
@@ -0,0 +1,133 @@
...
+		logrus.Debugf("Failed to delete container: %s", err)
+	}
+	configFileData := buf.Bytes()
+
+	var config map[string]interface{}
+	err = json.Unmarshal(configFileData, &config)
```
silvin-lubecki: if err := ... ; err != nil{

cli/containerizedengine/hostpaths.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return "", errors.Wrapf(err, "malformed config file - %s", configFile)
+	}
+	rootDir := defaultDockerRootDir
+	tmp, defined := config["data-root"]
```
silvin-lubecki:
```go
if data, defined := config["data-root"]; defined{
    rootDir = data.(string)
}
```

cli/containerizedengine/update.go
```diff
@@ -0,0 +1,107 @@
...
+		// current engine version and automatically apply it so users
+		// could stay in sync by simply having a scheduled
+		// `docker engine update`
+		return fmt.Errorf("please pick the version you want to update to")
+	}
+
+	imageName := fmt.Sprintf("%s/%s:%s", opts.RegistryPrefix, opts.EngineImage, opts.EngineVersion)
+
+	// Look for desired image
+	image, err := c.cclient.GetImage(ctx, imageName)
```
silvin-lubecki: This pattern seems to be the same as in `baseClient.InitEngine`, it could be extracted and factorized.

dhiltgen:
> Getting a weird error on my experimental enabled cli

Hmm...  I'm a bit confused by what's going on here.  I tried switching the `Annotations` from "experimentalCLI" to "experimental" but that seems to break it worse.  I couldn't find any docs covering this, so maybe I've got some other piece of wiring hooked up incorrectly.  Any pointers?

cli/containerizedengine/engine.go
```diff
@@ -0,0 +1,268 @@
...
+	authResolver func(ctx context.Context, index *registrytypes.IndexInfo) types.AuthConfig) error {
+
+	ctx = namespaces.WithNamespace(ctx, engineNamespace)
+	// Verify engine isn't already running
+	engine, _ := c.GetEngine(ctx)
+	if engine != nil {
+		return ErrEngineAlreadyPresent
+	}
+
+	imageName := fmt.Sprintf("%s/%s:%s", opts.RegistryPrefix, opts.EngineImage, opts.EngineVersion)
```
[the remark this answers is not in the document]
dhiltgen: I've refactored this to use and options struct instead - there are some use-cases where parts can be omitted (image name and/or version) and filled in based on the current state of the system so I think trying to force the caller to always have a valid registry reference would actually make the code more complex and the entrypoints less consistent.  Once I push the update (aiming for later today) take a look and see if you think that's acceptable.

cli/containerizedengine/progress.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/containerd/remotes"
+	"github.com/docker/docker/pkg/jsonmessage"
+	digest "github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/sirupsen/logrus"
+)
+
+// nolint: gocyclo
```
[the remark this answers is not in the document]
dhiltgen: this code is migrated from another repo with slight tweaks to use docker CLI progress reporting instead of ctr progress reporting - I was trying not to significantly restructure it, but I'll see if I can do some limited refactoring without messing things up.

cli/containerizedengine/types.go
```diff
@@ -0,0 +1,175 @@
...
+	GetCurrentEngineVersion(ctx context.Context) (string, string, error)
+}
+type baseClient struct {
+	dclient dockerClient
+	cclient containerdClient
+}
+
+// EngineInitOptions contains the configuration settings
+// use during initialization of a containerized docker engine
+type EngineInitOptions struct {
```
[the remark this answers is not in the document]
dhiltgen: These options are specific to initializing the engine, not other possible containerd based containers we might add in the future.  I suppose we could rename it at that point when we add something new that makes it now ambiguous, but my preference would be to try to name things more explicitly for their purpose.  Do you feel strongly this should be renamed?

cli/containerizedengine/update.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/containerd/containerd/namespaces"
+	"github.com/docker/cli/pkg/containerized"
+	"github.com/docker/distribution/reference"
+	"github.com/docker/docker/api/types"
+	registrytypes "github.com/docker/docker/api/types/registry"
+	"github.com/pkg/errors"
+)
+
+// GetCurrentEngineVersion determines the current type of engine (image) and version
+func (c baseClient) GetCurrentEngineVersion(ctx context.Context) (string, string, error) {
```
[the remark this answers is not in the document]
dhiltgen: I think it'll flow better in the code to return an init options struct instead so I'll proceed down that path.

cli/command/engine/check.go
```diff
@@ -0,0 +1,117 @@
...
+		Args:  cli.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runCheck(dockerCli, options)
+		},
+	}
+	flags := cmd.Flags()
+	// TODO - this should default to the existing engine image if not specified
+	flags.StringVar(&options.registryPrefix, "registry-prefix", "docker.io/docker", "Override the default location where engine images are pulled")
+	flags.BoolVar(&options.all, "all", false, "Report all versions (older and pre-release versions)")
+	flags.BoolVar(&options.patchOnly, "patch-only", false, "Only report available patch versions")
```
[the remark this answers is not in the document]
dhiltgen: This did get a bit messy, so let me clean it up, but I'd still like to be able to support only patches and pre-releases, so let me break "all" up into two flags.

cli/command/engine/init.go
dhiltgen: We're working towards supporting windows, but we're going to keep the `init` CLI UX experimental for now in the upcoming release since we don't quite have time to get all the details sorted out in time for FC.

That said, definitely open to suggestions on refinements to get this better aligned for windows support.

cli/command/formatter/licenses.go
dhiltgen: There's no unit coverage for this file yet, but it's likely going to be tweaked as the upstream license lib is being refined concurrently, so I was delaying that until the new licensing code lands (hopefully tomorrow)

dhiltgen:
@vdemeester @cpuguy83 I've added one commit that refactors one of the test mocks using the model that I think you are proposing. - https://github.com/docker/cli/pull/1260/commits/9d73826e6fd7327689ead6f64c62eebe95421ea3  Please take a look at that and let me know what you think.

My preference is the original model as I think its easier to reason about, but if this is how you'd like the tests structured, I'll apply it to the other mocks in a follow up commit.

pkg/containerized/snapshot.go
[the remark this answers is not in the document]
vdemeester: how is it incompatible with the upstream containerd pattern ? it just says there is a small interface you can "declare your variable with". The other "way around" would be to define a local interface that satisfy the contract you need 😝

cli/updates/types.go
```diff
@@ -0,0 +1,8 @@
...
+
+// Update stores available updates for rendering in a table
+type Update struct {
```
vdemeester: This can now be remove (https://github.com/docker/cli/pull/1260/files#diff-2349949258e8913f2e5141f6d989e2bdR144)

vdemeester:
In addition, not a huge fan of the proposed test refactoring, and as most of the code is under `internal` now, I feel we can easily do some refactoring (both on tests and part of the code) in follow-up PRs :wink:

dhiltgen:
> In addition, not a huge fan of the proposed test refactoring, and as most of the code is under internal now, I feel we can easily do some refactoring (both on tests and part of the code) in follow-up PRs

Sounds good.  Thanks!  

I'll drop the test refactoring commit, make the other requested changes, get a vendor bump of the licensing lib and add some unit test coverage on the licensing side, and get that pushed later today.  Hopefully that will get us in a state where this is mergeable with only minor comments, then I'll submit a follow up clean-up PR after FC to address those.

internal/containerizedengine/progress.go
```diff
@@ -0,0 +1,215 @@
...
+						StartedAt: active.StartedAt,
+						UpdatedAt: active.UpdatedAt,
+					}
+					activeSeen[active.Ref] = struct{}{}
+				}
+			}
+
+			err := updateNonActive(ctx, ongoing, cs, statuses, keys, activeSeen, &done, start)
+			if err != nil {
+				continue outer
```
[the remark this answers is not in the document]
dhiltgen: This algorithm has been structured this way for the past year (or more) in the containerd tree, so i'm a little hesitant to start changing it significantly without understanding all the partial failure modes.

dhiltgen:
Squashed to 2 commits (vendor bump, and the impl) and rebased on master so it should be ready to merge once CI goes green.

crosbymichael:
LGTM

PR #1262, opened 2018-08-02
Allow custom pattern when extracting variable…

vdemeester:
… as it is possible to do it when interpolating. It also fixes when
there is 2 variables on the same *value* (in the composefile, on the
same line)

Finaly, renaming the default, used in cli, pattern to `defaultPattern`
to not be shadowed unintentionally.

Useful for `docker/app` (and required to fix `docker/app` :sweat_smile:)

### DIFF
--- a/cli/compose/template/template.go
+++ b/cli/compose/template/template.go
@@ -14,7 +14,7 @@ var patternString = fmt.Sprintf(
 	delimiter, delimiter, substitution, substitution,
 )
 
-var pattern = regexp.MustCompile(patternString)
+var defaultPattern = regexp.MustCompile(patternString)
 
 // DefaultSubstituteFuncs contains the default SubstitueFunc used by the docker cli
 var DefaultSubstituteFuncs = []SubstituteFunc{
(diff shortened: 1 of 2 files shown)


cli/compose/template/template.go
```diff
@@ -90,35 +90,42 @@ func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, su
...
-	return SubstituteWith(template, mapping, pattern, DefaultSubstituteFuncs...)
+	return SubstituteWith(template, mapping, defaultPattern, DefaultSubstituteFuncs...)
 }
 
 // ExtractVariables returns a map of all the variables defined in the specified
 // composefile (dict representation) and their default value if any.
-func ExtractVariables(configDict map[string]interface{}) map[string]string {
-	return recurseExtract(configDict)
+func ExtractVariables(configDict map[string]interface{}, pattern *regexp.Regexp) map[string]string {
+	if pattern == nil {
```
vdemeester: Not sure if it's required :stuck_out_tongue:

PR #1271, opened 2018-08-03
Fix shellcheck warnings

albers:
Fixed some problems identified by shellcheck.

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -288,10 +288,9 @@ __docker_complete_networks() {
 	COMPREPLY=( $(compgen -W "$(__docker_networks "$@")" -- "$current") )
 }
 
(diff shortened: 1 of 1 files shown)


albers:
@silvin-lubecki We already have this check, it's the _ci/circleci: shellcheck_ check that's always executed in this repo. You can manually execute it with `make -f docker.Makefile shellcheck`.

Not all shellcheck recommendations make sense in the context of bash completion, e.g. we construct awk commands that contain a literal `$1`. Shellcheck warns us that in `'$1'` the $1 won't be expanded and we should use double quotes instead, which is not what we want. So we had to add some shellcheck directives in order to suppress the false positives.

When the checks were introduced in #266, quite a lot of exceptions were added in order to get it running. In this PR I remove some of them.

PR #1276, opened 2018-08-06
build: change --console=[auto,false,true] to --progress=[auto,plain,tty]

tiborvass:
This changes the experimental --console flag to --progress following
feedback indicating avoidable confusion.

In addition to naming changes, the help output now has an additional
clarification, specifically: container output during builds are only
shown when progress output is set to plain. Not mentioning this was also
a big cause of confusion.

### DIFF
--- a/opts/opts.go
+++ b/opts/opts.go
@@ -6,7 +6,6 @@ import (
 	"net"
 	"path"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/docker/docker/api/types/filters"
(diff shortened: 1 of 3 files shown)


opts/opts.go
```diff
@@ -308,6 +307,16 @@ func ValidateSysctl(val string) (string, error) {
 	return "", fmt.Errorf("sysctl '%s' is not whitelisted", val)
 }
 
+func ValidateProgressOutput(val string) (string, error) {
```
AntaresS: Do you really need to return a string here?

vdemeester: I think it's to respect the other validation `Validate*` signatures, that are able to mutate the value being validated. As we don't really use it into a type that require this signature (and as I'm not found of a validation function mutating the value it validates), I think we shouldn't return a `string` and just an error

tiborvass:
@vdemeester @AntaresS fixed!

PR #1292, opened 2018-08-15
[WIP] Add new `builder` subcommand and implement `builder prune` to prune build cache

tiborvass:
*WIP*: this modifies vendor
- [ ] Waiting on https://github.com/moby/moby/pull/37651

This patch adds a new builder subcommand to manage builds, allowing to add more
builder-related commands in the future. Unfortunately `build` expects an argument
so could not be used as a subcommand.

This also implements `docker builder prune`, which is needed to prune the builder
cache manually without having to call `docker system prune`.

Today when relying on the legacy builder, users are able to list a set of images
(used as build cache) based on time filters and delete them. This patch allows the
same usecase and more (specifying storage size) when relying on the new builder.

The behavior of the `--all` option on `system prune` is slightly modified with
respect to build cache: without `--all`, all but internal/frontend cache is removed
whereas with `--all`, it removes the entirety of buildkit's cache. This is similar
to keeping non-dangling images by default and removing them if `--all` is specified.

### DIFF
--- a/cli/command/system/prune.go
+++ b/cli/command/system/prune.go
@@ -13,6 +13,7 @@ import (
 	"github.com/docker/cli/cli/command/network"
 	"github.com/docker/cli/cli/command/volume"
 	"github.com/docker/cli/opts"
+	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/versions"
 	units "github.com/docker/go-units"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


cli/command/system/prune.go
```diff
@@ -42,7 +43,7 @@ func newPruneCommand(dockerCli command.Cli) *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.BoolVarP(&options.force, "force", "f", false, "Do not prompt for confirmation")
-	flags.BoolVarP(&options.all, "all", "a", false, "Remove all unused images not just dangling ones")
+	flags.BoolVarP(&options.all, "all", "a", false, "Remove all unused images not just dangling ones, as well as all internal build cache")
```
thaJeztah: if `--all` is not set, do we prune any build cache at all?

(thinking about slightly different phrasing)

andrewhsu:
@tiborvass needs conflict resolution

tiborvass:
This is superseded by #1295 and #1327

PR #1295, opened 2018-08-17
builder: Implement `builder prune` to prune build cache

tiborvass:
This is a version of #1292 without any API changes to make it easier. It would still be better than forcing users to run `docker system prune` which also removes stopped containers (undesirable side-effect).

This patch adds a new builder subcommand, allowing to add more builder-related commands in the future. Unfortunately `build` expects an argument so could not be used as a subcommand.

This also implements `docker builder prune`, which is needed to prune the builder cache manually without having to call `docker system prune`.

Today when relying on the legacy builder, users are able to prune dangling images (used as build cache) by running `docker image prune`. This patch allows the same usecase with buildkit.

### DIFF
--- a/cli/command/builder/cmd.go
+++ b/cli/command/builder/cmd.go
@@ -1,0 +1,22 @@
+package builder
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/command"
+)
+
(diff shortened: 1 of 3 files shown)


dmcgowan:
LGTM

I think it is best to get this simple version in first to enable cleanup while discussing what any filtering will look like in the API.

PR #1296, opened 2018-08-17
build: Remove API requirement for --progress as it is CLI only

tiborvass:


tiborvass:
@thaJeztah it's ignored. We could add a warning.

thaJeztah:
perhaps we need an annotation for buildkit-only options; I'm ok with this change for now, but possibly for a follow up

PR #1303, opened 2018-08-21
Remove "experimental" annotations for buildkit

thaJeztah:
BuildKit can now be enabled without the daemon having experimental features enabled.

related engine PR; https://github.com/moby/moby/pull/37686

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -155,10 +155,8 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation("stream", "version", []string{"1.31"})
 
 	flags.StringVar(&options.progress, "progress", "auto", "Set type of progress output (only if BuildKit enabled) (auto, plain, tty). Use plain to show container output")
-	flags.SetAnnotation("progress", "experimental", nil)
 
 	flags.StringArrayVar(&options.secrets, "secret", []string{}, "Secret file to expose to the build (only if BuildKit enabled): id=mysecret,src=/local/secret")
-	flags.SetAnnotation("secret", "experimental", nil)
 	flags.SetAnnotation("secret", "version", []string{"1.39"})
 	return cmd
 }
--- a/cli/command/image/build_session.go
+++ b/cli/command/image/build_session.go
@@ -28,6 +28,9 @@ import (
 const clientSessionRemote = "client-session"
 
 func isSessionSupported(dockerCli command.Cli) bool {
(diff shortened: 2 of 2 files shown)


thaJeztah:
ping @andrewhsu @tiborvass @AntaresS  PTAL

cli/command/image/build.go
```diff
@@ -155,10 +155,8 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation("stream", "version", []string{"1.31"})
 
 	flags.StringVar(&options.progress, "progress", "auto", "Set type of progress output (only if BuildKit enabled) (auto, plain, tty). Use plain to show container output")
-	flags.SetAnnotation("progress", "experimental", nil)
 
 	flags.StringArrayVar(&options.secrets, "secret", []string{}, "Secret file to expose to the build (only if BuildKit enabled): id=mysecret,src=/local/secret")
-	flags.SetAnnotation("secret", "experimental", nil)
```
thaJeztah: Wonder if we should have a new annotation to hide / disable these if the CLI (or daemon) is configured to use builder v1

AntaresS: you mean hide the `progress` and `secret` flags?

AntaresS: probably better when builder v1 is configured on the cli side

thaJeztah: Yes; not a blocker for now, but would be good to look into

thaJeztah:
Pushed a second commit to move "sessions" out of experimental as well (per https://github.com/moby/moby/pull/37686)

thaJeztah:
@tiborvass should the `--stream` option also be moved out of experimental, or is that separate? https://github.com/docker/cli/blob/2461cd618d49d63ccab18992fb146319fbd47f5f/cli/command/image/build.go#L153-L154

cli/command/image/build_session.go
```diff
@@ -28,6 +28,9 @@ import (
 const clientSessionRemote = "client-session"
 
 func isSessionSupported(dockerCli command.Cli) bool {
+	if versions.GreaterThanOrEqualTo(dockerCli.Client().ClientVersion(), "1.39") {
```
tiborvass: This should check daemon version, not CLI.

thaJeztah: Daemon could be version 1.40, but the client downgraded to 1.10; in that case the API version that's used is 1.10, and it's not supported

PR #1304, opened 2018-08-22
getEngineConfigFilePath is only used during test…

vdemeester:
… so moving it in test files for now.

### DIFF
--- a/internal/containerizedengine/engine.go
+++ b/internal/containerizedengine/engine.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/containerizedengine/engine.go
```diff
@@ -89,35 +88,6 @@ func (c baseClient) getEngineImage(engine containerd.Container) (string, error)
 	return image.Name(), nil
 }
 
-// getEngineConfigFilePath will extract the config file location from the engine flags
-func (c baseClient) getEngineConfigFilePath(ctx context.Context, engine containerd.Container) (string, error) {
```
dhiltgen: Originally this was in use, but in the refactorings it wasn't necessary.  My suspicion is we may need this again as things are fleshed out further in the future, but we can pull it back into the main package at that time.

vdemeester: yes exactly :wink:

PR #1306, opened 2018-08-23
man: obey SOURCE_DATE_EPOCH when generating man pages

cyphar:
Previously our man pages included the current time each time they were
generated. This causes an issue for reproducible builds, since each
re-build of a package that includes the man pages will have different
times listed in the man pages.

To fix this, add support for SOURCE_DATE_EPOCH ([which is a standardised
packaging environment variable, designed to be used specifically for
this purpose][1]). spf13/cobra doesn't support this natively yet (though
I will push a patch for that as well), but it's simpler to fix it
directly in docker/cli.
    
[1]: https://reproducible-builds.org/specs/source-date-epoch/
    
### DIFF
--- a/man/generate.go
+++ b/man/generate.go
@@ -6,6 +6,8 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
+	"time"
 
 	"github.com/docker/cli/cli/command"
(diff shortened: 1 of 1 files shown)


bmwiedemann:
Related: https://github.com/spf13/cobra/pull/667

PR #1275, opened 2018-08-06
[enhancement] enable buildkit from daemon side

AntaresS:
This PR needs to have latest of `github.com/docker/docker` vendored into the `docker/cli` codebase to make use of the feature flags.

**- What I did**
- added `--buildkit` option to enable buildkit in daemon https://github.com/moby/moby/pull/37593.
- Remove the extra need of `DOCKER_BUILDKIT` env var setup from client host.

**- How I did it**

**- How to verify it**
Now with "experimental" and "buildkit" enabled on the daemon side, `docker build` will use buildkit as builder

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #1307, opened 2018-08-23
update docs with the new features option in daemon.json

AntaresS:
**- What I did**
- updated the `dockerd.md` references with newly added option `features` in daemon.json
  - referring https://github.com/moby/moby/pull/37593 https://github.com/docker/cli/pull/1275
- added a new entry in "CONFIGURATION RELOAD BEHAVIOR" section
  - referring https://github.com/moby/moby/pull/37692

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/commandline/dockerd.md
+++ b/docs/reference/commandline/dockerd.md
@@ -1297,6 +1297,7 @@ This is a full example of the allowed configuration options on Linux:
 	"exec-opts": [],
 	"exec-root": "",
 	"experimental": false,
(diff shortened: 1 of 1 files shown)


AntaresS:
If we are eventually getting rid of `--experimental` and use only `features` to control enable/disable features. maybe at some point we should include a list of available features in the reference as well?

AntaresS:
ptal @thaJeztah @andrewhsu @brawong

PR #1316, opened 2018-08-28
[WIP] Bump Go to 1.11.0

thaJeztah:
Release notes: https://golang.org/doc/go1.11

this PR is created on top of https://github.com/docker/cli/pull/1315, which should go in first

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ clone_folder: c:\gopath\src\github.com\docker\cli
 
 environment:
   GOPATH: c:\gopath
-  GOVERSION: 1.10.4
+  GOVERSION: 1.11.0
   DEPVERSION: v0.4.1
 
 install:
--- a/dockerfiles/Dockerfile.binary-native
+++ b/dockerfiles/Dockerfile.binary-native
@@ -1,4 +1,4 @@
-FROM    golang:1.10.4-alpine
+FROM    golang:1.11.0-alpine
 
 RUN     apk add -U git bash coreutils gcc musl-dev
 
--- a/dockerfiles/Dockerfile.cross
+++ b/dockerfiles/Dockerfile.cross
@@ -1,3 +1,3 @@
-FROM    dockercore/golang-cross:1.10.4@sha256:55c7b933ac944f4922b673b4d4340d1a0404f3c324bd0b3f13a4326c427b1f2a
+FROM    dockercore/golang-cross:1.11.0@sha256:e968fd20bbd6bf3b65abaf2f6f5ee2420e1081ab840872a5b30bd2af4246b28c
 ENV     DISABLE_WARN_OUTSIDE_CONTAINER=1
 WORKDIR /go/src/github.com/docker/cli
--- a/dockerfiles/Dockerfile.dev
+++ b/dockerfiles/Dockerfile.dev
@@ -1,5 +1,5 @@
 
-FROM    golang:1.10.4-alpine
+FROM    golang:1.11.0-alpine
 
 RUN     apk add -U git make bash coreutils ca-certificates curl
 
--- a/dockerfiles/Dockerfile.e2e
+++ b/dockerfiles/Dockerfile.e2e
@@ -1,4 +1,4 @@
-ARG GO_VERSION=1.10.4
+ARG GO_VERSION=1.11.0
 
 FROM docker/containerd-shim-process:a4d1531 AS containerd-shim-process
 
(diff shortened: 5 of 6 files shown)


thaJeztah:
Interesting;

```
./scripts/build/cross
Building build/docker-darwin-amd64
# crypto/x509
osxcross: error: cannot find libc++ headers
osxcross: error: while detecting target
```

thaJeztah:
Alright; found some similar issues here; https://github.com/godotengine/godot/issues/17888, and here https://github.com/karalabe/xgo/issues/44.

The last one actually was using the SDK from dockerproject.org, but looks like they're patching it; https://github.com/karalabe/xgo/commit/647f256c447ee20f9bf13ebc42e612d55994a383 

Could use some help on this one 😓

kolyshkin:
I've checked the difference between these two images
> dockercore/golang-cross:1.10.4@sha256:55c7b933ac944f4922b673b4d4340d1a0404f3c324bd0b3f13a4326c427b1f2a
> dockercore/golang-cross:1.11.0@sha256:e968fd20bbd6bf3b65abaf2f6f5ee2420e1081ab840872a5b30bd2af4246b28c

and the only changes are in /usr/local/go...

kolyshkin:
Summary of my findings so far (to be continued)

OK the error message is

```
# crypto/x509
osxcross: error: cannot find libc++ headers
```

Meaning it happens while compiling crypto/x509, which is a standard go pkg.

Looking at the `git diff go1.10.4..go1.11 src/crypto/x509/`) output (in Golang sources) we can see:
```diff
--- a/src/crypto/x509/root_cgo_darwin.go
+++ b/src/crypto/x509/root_cgo_darwin.go
@@ -7,7 +7,7 @@
 package x509
 
 /*
-#cgo CFLAGS: -mmacosx-version-min=10.6 -D__MAC_OS_X_VERSION_MAX_ALLOWED=1080
+#cgo CFLAGS: -mmacosx-version-min=10.10 -D__MAC_OS_X_VERSION_MAX_ALLOWED=101300
 #cgo LDFLAGS: -framework CoreFoundation -framework Security
 
 #include <errno.h>
```

which probably means it requires at least 10.10 now. Same is mentioned in [Go 1.11 release notes](https://golang.org/doc/go1.11):
> As announced in the Go 1.10 release notes, Go 1.11 now requires OpenBSD 6.2 or later, macOS 10.10 Yosemite or later, or Windows 7 or later; support for previous versions of these operating systems has been removed.

** Maybe it makes sense to raise `OSX_VERSION_MIN=10.6` in https://github.com/docker/golang-cross/blob/master/osx-cross.sh to be `10.10`? **

In addition, https://github.com/tpoechtrager/osxcross/blob/master/README.md says: 

```
 >= 10.9 also defaults to libc++ instead of libstdc++, this behavior
can be overriden by explicitly passing -stdlib=libstdc++ to clang.
```

This is probably the root cause of the failure (as we don't supply libc++ headers in our image, and the version switch from 10.6 to 10.11 made it use libc++ rather than libstdc++).

kolyshkin:
The tarball used for building golang-cross, `MacOSX10.11.sdk.tar.xz`, lacks `/usr/include/c++/v1/*` files (that presumably comes from xcode).

I had luck using a similar tarball from the https://github.com/multiarch/crossbuild project, but that one lacks `libltdl-dev` includes and libs...

Guess I need to build my own kosher tarball that has it all

kolyshkin:
> Guess I need to build my own kosher tarball that has it all

The process is described at https://github.com/tpoechtrager/osxcross#packaging-the-sdk and seems more or less straightforward. In addition, one would need to pack ltdl library binaries for Mac -- and this is where I'm a bit stuck... Hmm, could it be that it's also part of xcode?

Well, I can't check that since downloading XCode requires apple developer ID.

Perhaps someone who has it can take a look? I'm more than happy to help

kolyshkin:
OK, I made a Frankenstein macosx tarball from the two that I had, and it works but gives a (legit) warning about libtool:

> ld: warning: object file (/osxcross/target/bin/../SDK/MacOSX10.10.sdk/usr/lib/libltdl.a(libltdl_libltdl_la-ltdl.o)) was built for newer OSX version (10.11) than being linked (10.10)

So, I need someone's help with that, as I don't have Mac, nor an Apple developer ID

kolyshkin:
Alternatively, we can build docker-cli for mac without PKCS11 support...

kolyshkin:
OK, I think I found a way. 

I believe I'll have a successful build at https://github.com/docker/cli/pull/1426. The changes in there need to be migrated back to `dockercore/golang-cross`, for now it's a shortcut for testing only.

kolyshkin:
This one should be closed in favor of https://github.com/docker/cli/pull/1426

thaJeztah:
Ah, yes :+1:

PR #1326, opened 2018-08-29
Allow marshalling of Compose config to JSON

shin-:
**- What I did**

Update Compose types to allow producing a JSON file from a compose `Config` type.
This would allow for a more elegant implementation of https://github.com/docker/app/pull/357

**- How I did it**

Peppered `json:` annotations throughout the structs and implemented a few `MarshalJSON` methods.
Other parts of the code (mainly around type conversion) were modified as necessary.

**- How to verify it**

`TestJSONMarshallConfig` should ensure JSON files can be produced and re-read (reading is performed using the existing YAML parser, as YAML is a superset of JSON)

**Note:** The current implementation does not allow for proper marshaling of extension fields (except for top-level extension fields). As a result, these fields are currently omitted from the output.

### DIFF
--- a/cli/compose/loader/full-struct_test.go
+++ b/cli/compose/loader/full-struct_test.go
@@ -65,17 +65,17 @@ func services(workingDir, homeDir string) []types.ServiceConfig {
 				Labels:   map[string]string{"FOO": "BAR"},
 				RollbackConfig: &types.UpdateConfig{
 					Parallelism:     uint64Ptr(3),
-					Delay:           time.Duration(10 * time.Second),
+					Delay:           types.Duration(10 * time.Second),
 					FailureAction:   "continue",
-					Monitor:         time.Duration(60 * time.Second),
+					Monitor:         types.Duration(60 * time.Second),
 					MaxFailureRatio: 0.3,
 					Order:           "start-first",
 				},
 				UpdateConfig: &types.UpdateConfig{
 					Parallelism:     uint64Ptr(3),
-					Delay:           time.Duration(10 * time.Second),
+					Delay:           types.Duration(10 * time.Second),
 					FailureAction:   "continue",
-					Monitor:         time.Duration(60 * time.Second),
+					Monitor:         types.Duration(60 * time.Second),
 					MaxFailureRatio: 0.3,
 					Order:           "start-first",
 				},
@@ -344,7 +344,7 @@ func services(workingDir, homeDir string) []types.ServiceConfig {
 			},
 			StdinOpen:       true,
 			StopSignal:      "SIGUSR1",
-			StopGracePeriod: durationPtr(time.Duration(20 * time.Second)),
+			StopGracePeriod: durationPtr(20 * time.Second),
 			Tmpfs:           []string{"/run", "/tmp"},
 			Tty:             true,
 			Ulimits: map[string]*types.UlimitsConfig{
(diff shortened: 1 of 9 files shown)


cli/compose/types/types.go
```diff
@@ -62,6 +65,74 @@ type ConfigDetails struct {
...
+type Duration struct {
+	time.Duration
+}
+
+// MarshalJSON makes Duration implement json.Marshaler
+func (d Duration) MarshalJSON() ([]byte, error) {
+	return json.Marshal(d.String())
+}
+
+// UnmarshalJSON bla bla
```
mat007: `bla bla`? :)

cli/compose/types/types.go
```diff
@@ -62,6 +65,74 @@ type ConfigDetails struct {
...
+}
+
+// UnmarshalYAML makes Duration implement yaml.Unmarshaler
+func (d *Duration) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var v interface{}
+	err := unmarshal(&v)
+	if err != nil {
+		return err
+	}
+	switch value := v.(type) {
```
silvin-lubecki: nit: The switch part can be extracted and factorized with UnmarshlJSON.

cli/compose/types/types.go
```diff
@@ -70,14 +141,67 @@ func (cd ConfigDetails) LookupEnv(key string) (string, bool) {
...
+	Volumes  map[string]VolumeConfig    `yaml:",omitempty" json:"volumes,omitempty"`
+	Secrets  map[string]SecretConfig    `yaml:",omitempty" json:"secrets,omitempty"`
+	Configs  map[string]ConfigObjConfig `yaml:",omitempty" json:"configs,omitempty"`
+	Extras   map[string]interface{}     `yaml:",inline", json:"-"`
+}
+
+// MarshalJSON makes Config implement json.Marshaler
+func (c Config) MarshalJSON() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	buf.WriteString(fmt.Sprintf("{\n  \"version\": %q,\n  \"services\": ", c.Version))
```
silvin-lubecki: Use `fmt.Fprintf` instead? and backquotes?
```go
fmt.Fprintf(buf, `{
    "version": %q,
    "services": `, c.Version)
```

cli/compose/types/types.go
```diff
@@ -409,19 +562,27 @@ func (e External) MarshalYAML() (interface{}, error) {
 	return External{Name: e.Name}, nil
 }
 
+// MarshalJSON makes External implement json.Marshaller
+func (e External) MarshalJSON() ([]byte, error) {
+	if e.Name == "" {
+		return []byte(fmt.Sprintf("%v", e.External)), nil
+	}
+	return []byte(fmt.Sprintf("{%q: %q}", "name", e.Name)), nil
```
silvin-lubecki:
```go
return []byte(fmt.Sprintf(`{"name": %q}`,  e.Name)), nil
```

cli/compose/types/types.go
```diff
@@ -359,46 +497,61 @@ func (u *UlimitsConfig) MarshalYAML() (interface{}, error) {
 	return u, nil
 }
 
+// MarshalJSON makes UlimitsConfig implement json.Marshaller
+func (u *UlimitsConfig) MarshalJSON() ([]byte, error) {
```
silvin-lubecki: Why don't you add also the `UnmarshalJSON` ? If read this code correctly, it either produce a simple int `18`, or a json map `{ "soft": 17, "hard": 12}`. I don't think the default JSON unmarshaller will parse these 2 results. It will expect a map with up to 3 keys.

cli/compose/types/types.go
```diff
@@ -409,19 +562,27 @@ func (e External) MarshalYAML() (interface{}, error) {
 	return External{Name: e.Name}, nil
 }
 
+// MarshalJSON makes External implement json.Marshaller
+func (e External) MarshalJSON() ([]byte, error) {
```
silvin-lubecki: Same comment as UlimitsConfig for the `UnmarshalJSON`.

cli/compose/types/types.go
```diff
@@ -268,74 +401,79 @@ func (u UnitBytes) MarshalYAML() (interface{}, error) {
 	return fmt.Sprintf("%d", u), nil
 }
 
+// MarshalJSON makes UnitBytes implement json.Marshaler
+func (u UnitBytes) MarshalJSON() ([]byte, error) {
```
silvin-lubecki: Is this really needed? The default JSON marshaller does not work on this type?

cli/compose/types/types.go
```diff
@@ -70,14 +141,67 @@ func (cd ConfigDetails) LookupEnv(key string) (string, bool) {
...
+	Services Services                   `json:"services"`
+	Networks map[string]NetworkConfig   `yaml:",omitempty" json:"networks,omitempty"`
+	Volumes  map[string]VolumeConfig    `yaml:",omitempty" json:"volumes,omitempty"`
+	Secrets  map[string]SecretConfig    `yaml:",omitempty" json:"secrets,omitempty"`
+	Configs  map[string]ConfigObjConfig `yaml:",omitempty" json:"configs,omitempty"`
+	Extras   map[string]interface{}     `yaml:",inline", json:"-"`
+}
+
+// MarshalJSON makes Config implement json.Marshaler
+func (c Config) MarshalJSON() ([]byte, error) {
```
silvin-lubecki: I don't understand why we need this function... I think the struct field tags you added should produce an equivalent output with the default marshaller.

cli/compose/types/types.go
shin-: For the scope of this PR, we're not really interested in parsing JSON using the `encoding/json` library, as we can use the existing code to parse the JSON file as if it were a YAML file. See the [YAML spec](http://yaml.org/spec/1.2/spec.html#id2759572) on the subject:

> YAML can therefore be viewed as a natural superset of JSON [...]. This is also the case in practice; every JSON file is also a valid YAML file.

cli/compose/types/types.go
shin-: This is necessary if we want to include top-level extension fields in the output, as there is no equivalent to the `yaml:inline` anotation for JSON. If we're okay with those being omitted from the JSON output, it can be removed.

cli/compose/types/types.go
shin-: This type would be unmarshaled into a number type by default, whereas the schema enforces the string type for these fields.

cli/compose/types/types.go
```diff
@@ -62,6 +65,74 @@ type ConfigDetails struct {
 	Environment map[string]string
 }
 
+// Duration is a thin wrapper around time.Duration with improved JSON marshalling
+type Duration struct {
```
vdemeester: `type Duration time.Duration`

cli/compose/types/types.go
```diff
@@ -70,14 +141,69 @@ func (cd ConfigDetails) LookupEnv(key string) (string, bool) {
...
+	fmt.Fprintf(buf, `{
+  "version": %q,
+  "services":`, c.Version)
+	js, err := json.MarshalIndent(c.Services, "  ", "  ")
+	if err != nil {
+		return nil, err
+	}
+	buf.Write(js)
+	if len(c.Networks) > 0 {
+		buf.WriteString(",\n  \"networks\": ")
```
vdemeester: I don't think we should do the `\n` or even that "string" manually. We should be able to build a struct or map that `encoding/json` can marshal in JSON (and thus the way it wants to be…). I feel this part of the code is way more complicated than it should…

cli/compose/types/types.go
vdemeester: Also, should be declared below `Config`, … (even better would be to split this file but it can be in a follow-up

cli/compose/types/types.go
vdemeester: The following code should do the trick nicely

```go

func (c Config) MarshalJSON() ([]byte, error) {
	m := map[string]interface{}{
		"version":  c.Version,
		"services": c.Services,
	}
	if len(c.Networks) > 0 {
		m["networks"] = c.Networks
	}
	if len(c.Volumes) > 0 {
		m["volumes"] = c.Volumes
	}
	if len(c.Secrets) > 0 {
		m["secrets"] = c.Secrets
	}
	if len(c.Configs) > 0 {
		m["configs"] = c.Configs
	}
	for k, v := range c.Extras {
		m[k] = v
	}
	return json.Marshal(m)
}
```

cli/compose/types/types.go
```diff
@@ -92,75 +218,84 @@ func (s Services) MarshalYAML() (interface{}, error) {
 	return services, nil
 }
 
+// MarshalJSON makes Services implement json.Marshaler
+func (s Services) MarshalJSON() ([]byte, error) {
```
vdemeester: Why is this required ?

cli/compose/types/types.go
vdemeester: Doesn't the following work ? (and would be closer to `MarshalYAML`)

```go
if u.Single != 0 {
    return json.Marshal(u.Single)
}
return json.Marshal(u)
```

… as we can make the assumption `Single` is `0`, this won't be marshalled with a `0` value.

cli/compose/types/types.go
vdemeester: Same here
```go

func (e External) MarshalJSON() ([]byte, error) {
	if e.Name == "" {
		return json.Marshal(e.External)
	}
	return json.Marshal(map[string]string{"name": e.Name})
}
```

shin-:
Thanks for the review guys! 

> Few comments:
>
> 1. on Duration / time.Duration, why is it required ? what was not working before that needed to be fixed there ? 

`encoding/json` marshals `time.Duration` values into their `Nanoseconds` integer, which
1. Isn't very human-readable
2. Doesn't comply with Compose's JSON schema that specifies durations must be strings.

You can check out my [compose-json-annotations_1 branch](https://github.com/shin-/cli/tree/compose-json-annotations_1) and run `go test github.com/docker/cli/cli/compose/loader` to see what happens if we keep the time.Duration part.

> If so, it would be really cool to have a test showing that (i.e. that would be red on master, green on the PR).

Marshaling to JSON doesn't produce a valid output on `master` for a myriad of reasons because it's not set up to do so, so I'm not sure how I would go about highlighting that specific issue. But you'll find that `TestJSONMarshallConfig` fails on `master`, in part because `time.Duration` isn't being marshaled into strings.

> This is something that induce a lot of change in the PR and I feel we can get away to just require duration to be strings (1m, …) like we do on yaml.

I'm not sure I understand what you're trying to say here, but strings for the input are already required - that doesn't matter after it gets converted to a `time.Duration` type.

> 2. There is inconsistency in types.go : all json tag define a name, even if it's not required, which is not the case for the yaml tag (where we only specify it if's required).

It's always required because `encoding/json` uses the literal field name without modification by default, which means all fields are capitalized in the output, which in turn wouldn't be a valid Compose file.

cli/compose/types/types.go
shin-: I tried that, but [maps get shuffled](https://nathanleclaire.com/blog/2014/04/27/a-surprising-feature-of-golang-that-colored-me-impressed/) when marshaled to JSON, so your solution doesn't preserve the order of the fields, which is an important feature of the `struct` type.

cli/compose/types/types.go
shin-: Because `Services` is a `ServiceConfig` slice and the expected output is a mapping. Same reason we have a `MarshalYAML` method for this type.

cli/compose/types/types.go
shin-: `json.Marshal(u)` will call `u.MarshalJSON()`, thus recursing infinitely, so no.

cli/compose/types/types.go
vdemeester: I don't think we should care about the order, it will be marshaled correctly anyway… the `json` format is just data…

cli/compose/types/types.go
shin-: My hunch is that that's probably slower (more function calls), but if you think that's better I'm happy to change it.

cli/compose/types/types.go
shin-: It's data that is likely to be read and modified by humans, who do care about semantic order.

cli/compose/types/types.go
dnephin: I don't think I understand why UnmarshalJSON/Yaml was added

cli/compose/types/types.go
```diff
@@ -62,6 +65,74 @@ type ConfigDetails struct {
...
+	time.Duration
+}
+
+// MarshalJSON makes Duration implement json.Marshaler
+func (d Duration) MarshalJSON() ([]byte, error) {
+	return json.Marshal(d.String())
+}
+
+// UnmarshalJSON makes Duration implement json.Unmarshaler
+func (d *Duration) UnmarshalJSON(b []byte) error {
```
dnephin: MarshalJSON/Yaml needs to exist in this package to serialize, but I'm not sure UnmarshalJSON/YAML need to exist. I see there was a comment about, but I think I'm missing some context. Wouldn't it be better to keep using `mapstructure` hooks to decode?

cli/compose/types/types.go
silvin-lubecki: What about the YAML marshaller then? I think that the YAML has more chances to be read by humans (because of the compose files) than the JSON version.

shin-: @silvin-lubecki What about it? The YAML output is always ordered the same way based on the ordering of the struct.

cli/compose/types/types.go
shin-: @dnephin Thanks! I needed those originally because I hadn't tinkered with the `mapstructure` hooks, but they're obsolete now. I've removed them as a result.

cli/compose/types/types.go
shin-: I've tried this, but `type Duration time.Duration` doesn't let me inherit the `String()` method, which I need to properly marshal the object.

dnephin: I think this would work:

```go
time.Duration(d).String()
```

cli/compose/types/types.go
garethr: JSON is very rarely used to author Compose files as far as I'm aware: GitHub shows [237](https://github.com/search?q=in%3Apath+docker-compose.json+extension%3Ajson&type=Code) vs [~300,000](https://github.com/search?q=in%3Apath+docker-compose.yml+extension%3Ayml&type=Code) for YAML.

I think given that it's reasonable to assume where JSON is used it's more of a machine to machine format, and ordering isn't overly important as long as it's deterministic. The usecase I've seen (and done myself) here is JSON is increasingly available in the standard library of languages, YAML (because it's crazy) isn't.

shin-: Thanks @garethr ! Updated accordingly.

cli/compose/types/types.go
vdemeester: As @dnephin mentionned, this works

```
package main

import (
	"fmt"
	"time"
)

type Duration time.Duration

func (d Duration) String() string {
	return "foo:" + time.Duration(d).String()
}

func main() {
	t := time.Duration(2000)
	fmt.Println("Hello, playground")
	fmt.Println(t)
	fmt.Println(Duration(t))
}
```

vdemeester: So in this case 

```go

// Duration is a thin wrapper around time.Duration with improved JSON marshalling
type Duration time.Duration

// MarshalJSON makes Duration implement json.Marshaler
func (d Duration) MarshalJSON() ([]byte, error) {
	return json.Marshal(d.String())
}

// MarshalYAML makes Duration implement yaml.Marshaler
func (d Duration) MarshalYAML() (interface{}, error) {
	return d.String(), nil
}

func (d Duration) String() string {
	return time.Duration(d).String()
}
```

… and thus no need for the `ConvertDuration` anymor

cli/compose/types/types.go
```diff
@@ -268,74 +326,79 @@ func (u UnitBytes) MarshalYAML() (interface{}, error) {
 	return fmt.Sprintf("%d", u), nil
 }
 
+// MarshalJSON makes UnitBytes implement json.Marshaler
+func (u UnitBytes) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("\"%d\"", u)), nil
```
vdemeester:
```go
return []byte(fmt.Sprintf(`"%d"`, u)), nil
```

cli/compose/types/types.go
vdemeester: Then

```go
if u.Single != 0 {
    return json.Marshal(u.Single)
}
return json.Marshal(&UlimitsConfig{Hard: u.Hard, Soft: u.Soft})
```

cli/compose/types/types.go
vdemeester: so I wouldn't worry about *slow* for now, but… it's not a blocker as of now

cli/compose/loader/full-struct_test.go
```diff
@@ -887,3 +887,551 @@ x-nested:
 		workingDir,
 		workingDir)
 }
+
+func fullExampleJSON(workingDir string) string {
+	return fmt.Sprintf(`{
```
silvin-lubecki: This should be in a golden file IMHO.

vdemeester: Well, there is some parameters (but we could "hard-code" them for the tests maybe)
The `yaml` one could be too indeed :stuck_out_tongue: 
(could be done in a followup for both)

cli/compose/types/types.go
vdemeester: @shin- *wrapping* works just fine.

```go	
td := time.Duration(10)
cd := Duration(10)
fmt.Println(td)
fmt.Println(cd)
fmt.Println(Duration(td))
fmt.Println(time.Duration(td))
```

cli/compose/types/types.go
vdemeester: @shin- it's gonna print the same as `"\"%d\""`, it's just to not having to escape the `"`

shin-: oh right sorry I blinked and didn't see the backquotes :man_facepalming:

cli/compose/types/types.go
shin-: Yep, but we have several structs that use pointers, and those need to be converted. See https://github.com/docker/cli/pull/1326/files#diff-b3114871e359bbcd13674fefe7b3e16eR288 for example.

cli/compose/types/types.go
```diff
@@ -62,6 +63,32 @@ type ConfigDetails struct {
...
 
+// Duration is a thin wrapper around time.Duration with improved JSON marshalling
+type Duration time.Duration
+
+func (d Duration) String() string {
+	return time.Duration(d).String()
+}
+
+// ConvertDuration converts a wrapped Duration pointer to a time.Duration pointer with the same value.
+func ConvertDuration(d *Duration) *time.Duration {
```
vdemeester: nit: `ConvertDurationPtr` maybe ?

cli/compose/types/types.go
```diff
@@ -92,75 +144,84 @@ func (s Services) MarshalYAML() (interface{}, error) {
...
-	Tmpfs           StringList                       `yaml:",omitempty"`
-	Tty             bool                             `mapstructure:"tty" yaml:"tty,omitempty"`
-	Ulimits         map[string]*UlimitsConfig        `yaml:",omitempty"`
-	User            string                           `yaml:",omitempty"`
-	UserNSMode      string                           `mapstructure:"userns_mode" yaml:"userns_mode,omitempty"`
-	Volumes         []ServiceVolumeConfig            `yaml:",omitempty"`
-	WorkingDir      string                           `mapstructure:"working_dir" yaml:"working_dir,omitempty"`
-
-	Extras map[string]interface{} `yaml:",inline"`
+	Name string ` yaml:"-" json:"-"`
```
thaJeztah: Looks like a space was added before `yaml` (looks accidental)

cli/compose/types/types.go
```diff
@@ -92,75 +144,84 @@ func (s Services) MarshalYAML() (interface{}, error) {
...
+	CapAdd          []string                         `mapstructure:"cap_add" yaml:"cap_add,omitempty" json:"cap_add,omitempty"`
+	CapDrop         []string                         `mapstructure:"cap_drop" yaml:"cap_drop,omitempty" json:"cap_drop,omitempty"`
+	CgroupParent    string                           `mapstructure:"cgroup_parent" yaml:"cgroup_parent,omitempty" json:"cgroup_parent,omitempty"`
+	Command         ShellCommand                     `yaml:",omitempty" json:"command,omitempty"`
+	Configs         []ServiceConfigObjConfig         `yaml:",omitempty" json:"configs,omitempty"`
+	ContainerName   string                           `mapstructure:"container_name" yaml:"container_name,omitempty" json:"container_name,omitempty"`
+	CredentialSpec  CredentialSpecConfig             `mapstructure:"credential_spec" yaml:"credential_spec,omitempty" json:"credential_spec,omitempty"`
+	DependsOn       []string                         `mapstructure:"depends_on"  yaml:"depends_on,omitempty" json:"depends_on,omitempty"`
+	Deploy          DeployConfig                     ` yaml:",omitempty" json:"deploy,omitempty"`
```
thaJeztah: Same here; and varioius lines below; extraneous space before `yaml`

cli/compose/types/types.go
```diff
@@ -92,75 +144,84 @@ func (s Services) MarshalYAML() (interface{}, error) {
...
 // using the same format at libcompose: https://github.com/docker/libcompose/blob/master/yaml/build.go#L12
 type BuildConfig struct {
-	Context    string            `yaml:",omitempty"`
-	Dockerfile string            `yaml:",omitempty"`
-	Args       MappingWithEquals `yaml:",omitempty"`
-	Labels     Labels            `yaml:",omitempty"`
-	CacheFrom  StringList        `mapstructure:"cache_from" yaml:"cache_from,omitempty"`
-	Network    string            `yaml:",omitempty"`
-	Target     string            `yaml:",omitempty"`
+	Context    string            ` yaml:",omitempty" json:"context,omitempty"`
```
thaJeztah: Here as well (probably copy/pasta)

cli/compose/types/types.go
thaJeztah: Would there be corner-cases where `e.Name` or `e.External` has to be encoded for JSON?

I agree that performance should not be a huge concern in this case; it's most likely gonna be used interactively, not in a performance-critical path.

cli/compose/types/types.go
shin-: https://play.golang.org/p/X1HpAqXsJHZ 🤷

cli/compose/types/types.go
shin-: @thaJeztah If I understand your question correctly, both are simple types (boolean/string), so there's no special casing necessary.

cli/compose/types/types.go
shin-: Black pointer magic >_>

😛

vdemeester: Oh it actually make sense :angel: (because `MarshalJSON` is not implemented on `UlimitsConfig` :angel:

shin-: e7f43f1d8f32aa49bff3ec84b06322fd823630e4 👍 

I get why it works, but it's probably going to be obscure to most people IMO especially if they're not familiar with Go

shin-:
@thaJeztah Rebased and squashed :+1:

PR #1327, opened 2018-08-30
build: add options to builder prune

tiborvass:

Related to https://github.com/moby/moby/pull/37651

This patch fixes the UX around build cache:
- Adds new options to builder prune: `--filter`, `--keep-storage`, `--all`, `--force`. This is needed to to avoid situations where user needs to prune all their cache, and they can thus prune only a subset of their cache.
- Shows table layout for build cache in `system df` consistent with the way other types are shown (containers, images, volumes...)

### DIFF
--- a/cli/command/formatter/disk_usage.go
+++ b/cli/command/formatter/disk_usage.go
@@ -12,22 +12,11 @@ import (
 )
 
 const (
-	defaultDiskUsageImageTableFormat     = "table {{.Repository}}\t{{.Tag}}\t{{.ID}}\t{{.CreatedSince}} ago\t{{.VirtualSize}}\t{{.SharedSize}}\t{{.UniqueSize}}\t{{.Containers}}"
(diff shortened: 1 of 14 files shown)


cli/command/formatter/disk_usage.go
```diff
@@ -196,11 +189,19 @@ func (ctx *DiskUsageContext) verboseWrite() error {
...
-	t := template.Must(template.New("buildcache").Parse(defaultBuildCacheVerboseFormat))
-
+	tmpl, err = ctx.startSubsection(defaultDiskUsageBuildCacheTableFormat)
+	if err != nil {
+		return err
+	}
+	buildCacheSort(ctx.BuildCache)
 	for _, v := range ctx.BuildCache {
-		t.Execute(ctx.Output, *v)
+		if !v.Shared {
```
tonistiigi: Shouldn't filter in here, but in the summary bytes.

PR #1329, opened 2018-08-31
update usage for 'docker build' with '--progress' and '--secret' options

AntaresS:
**- What I did**
- updated `docker build` usage with `--progress` option and `--secret` option

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/commandline/build.md
+++ b/docs/reference/commandline/build.md
@@ -48,8 +48,11 @@ Options:
                                 '<network-name>|<network-id>': connect to a user-defined network
       --no-cache                Do not use cache when building the image
       --pull                    Always attempt to pull a newer version of the image
+      --progress                Set type of progress output (only if BuildKit enabled) (auto, plain, tty). 
+                                Use plain to show container output
   -q, --quiet                   Suppress the build output and print image ID on success
       --rm                      Remove intermediate containers after a successful build (default true)
+      --secret                  Secret file to expose to the build (only if BuildKit enabled): id=mysecret,src=/local/secret"
       --security-opt value      Security Options (default [])
       --shm-size bytes          Size of /dev/shm
                                 The format is `<number><unit>`. `number` must be greater than `0`.


AntaresS:
cc @andrewhsu

docs/reference/commandline/build.md
```diff
@@ -48,8 +48,11 @@ Options:
                                 '<network-name>|<network-id>': connect to a user-defined network
       --no-cache                Do not use cache when building the image
       --pull                    Always attempt to pull a newer version of the image
+      --progress                Set type of progress output (only if BuildKit enabled) (auto, plain, tty).
```
andrewhsu: `(only takes effect if BuildKit is enabled)`

andrewhsu: should probably say what the default value is

AntaresS: The only thing is that's according to the cli `--help` manual 
https://github.com/docker/cli/blob/master/cli/command/image/build.go#L157
I think it's better to keep them consistent.

docs/reference/commandline/build.md
```diff
@@ -48,8 +48,11 @@ Options:
                                 '<network-name>|<network-id>': connect to a user-defined network
       --no-cache                Do not use cache when building the image
       --pull                    Always attempt to pull a newer version of the image
+      --progress                Set type of progress output (only if BuildKit enabled) (auto, plain, tty). 
+                                Use plain to show container output
   -q, --quiet                   Suppress the build output and print image ID on success
       --rm                      Remove intermediate containers after a successful build (default true)
+      --secret                  Secret file to expose to the build (only if BuildKit enabled): id=mysecret,src=/local/secret"
```
thaJeztah: Note that the usage section of these markdown files are only used on github; for docs.docker.com, this part is generated from code (and equivalent to the information that's shown for the `--help` output)

PR #1331, opened 2018-09-02
fixes: no name check in docker container command

lifubang:

**- What I did**
fix a bug when run docker container client command like this:
docker stop redis/../redisps

The container redisps will be stopped.
It's very strange.

I have fix rm, kill, and stop commands. If this solution is acceptable, I will fix other container commands.

**- How I did it**
add container check when run these container client cmd.
The regex in names.go is from https://github.com/moby/moby/blob/master/daemon/names/names.go
If we can refer it, it'll be better.

**- How to verify it**
**Before fix:**
```
root@dockerdemo:~/gocode/src/github.com/docker/cli# build/docker ps
CONTAINER ID        IMAGE                              COMMAND                  CREATED             STATUS              PORTS               NAMES
4ada0f1dc848        redis:testps                       "/st 6379"               6 hours ago         Up 20 minutes       6379/tcp            redisps
d5555192178f        docker.acmcoder.com/public/redis   "docker-entrypoint.s…"   22 hours ago        Up 2 seconds        6379/tcp            redis
root@dockerdemo:~/gocode/src/github.com/docker/cli# build/docker stop redis/../redisps
stopid:redis/../redisps
redis/../redisps
root@dockerdemo:~/gocode/src/github.com/docker/cli# build/docker ps
CONTAINER ID        IMAGE                              COMMAND                  CREATED             STATUS              PORTS               NAMES
d5555192178f        docker.acmcoder.com/public/redis   "docker-entrypoint.s…"   22 hours ago        Up 26 seconds       6379/tcp            redis
root@dockerdemo:~/gocode/src/github.com/docker/cli#
```

**After fix:**
```
root@dockerdemo:~/gocode/src/github.com/docker/cli# build/docker kill redis/../redisps
container name redis/../redisps is invalid
```

Add container name check when run docker container client command.

cat


cli/names.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/docker/cli/cli/names"
+)
+
+var (
+	validContainerNamePattern = names.RestrictedNamePattern
+)
+
+// CheckContainerName check container's name is valid or not
+func CheckContainerName(name string) bool {
+	if len(name) == 0 {
```
silvin-lubecki: Why checking the length? I think the regular expression won't match an empty string.

cli/names.go
```diff
@@ -0,0 +1,19 @@
...
+		return false
+	}
+	return validContainerNamePattern.MatchString(strings.TrimPrefix(name, "/"))
```
silvin-lubecki: Why trimming the prefix? Is a container "/aaaa" valid? @thaJeztah WDYT?

cli/names/names.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"regexp"
+)
+
+// RestrictedNameChars collects the characters allowed to represent a name, normally used to validate container and volume names.
+const RestrictedNameChars = `[a-zA-Z0-9][a-zA-Z0-9_.-]`
+
+// RestrictedNamePattern is a regular expression to validate names against the collection of restricted characters.
+var RestrictedNamePattern = regexp.MustCompile(`^` + RestrictedNameChars + `+$`)
```
silvin-lubecki: Why did you cut the regular expression? I think it would be better to put everything together
```go
const pattern = `^[a-zA-Z0-9][a-zA-Z0-9_.-]+$`
```

cli/names/names.go
```diff
@@ -0,0 +1,11 @@
+package names
+
+import (
+	"regexp"
+)
+
+// RestrictedNameChars collects the characters allowed to represent a name, normally used to validate container and volume names.
+const RestrictedNameChars = `[a-zA-Z0-9][a-zA-Z0-9_.-]`
```
silvin-lubecki: It shouldn't be exported.

cli/names.go
```diff
@@ -0,0 +1,19 @@
+package cli
+
+import (
+	"strings"
+
+	"github.com/docker/cli/cli/names"
+)
+
+var (
+	validContainerNamePattern = names.RestrictedNamePattern
```
silvin-lubecki: I think the package "names" you added is not very useful. Why not putting the regexp and the pattern definition directly here in this file? It will also remove the need to expose the pattern and the regexp.

silvin-lubecki (changes requested):
Thank you @lifubang for submitting this PR! 😄 
I have some comments and questions, but I think it mainly lacks some tests.

cli/names.go
lifubang: Yes, docker daemon said it's valid.
see https://github.com/moby/moby/blob/master/daemon/names.go#L59

cli/command/container/kill.go
```diff
@@ -26,6 +26,11 @@ func NewKillCommand(dockerCli command.Cli) *cobra.Command {
 		Short: "Kill one or more running containers",
 		Args:  cli.RequiresMinArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			for _, name := range args {
```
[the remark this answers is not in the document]
lifubang: It's a good idea, then the error messages can be in on file.

cli/names.go
lifubang: Yes, it's just from moby project. I'll remove it.

lifubang (commented):
Thank you for your wonderful review. I have some code optimizations. Thanks.

cli/names.go
lifubang: Yes, thanks.

cli/names/names.go
lifubang: Yes, thanks.

cli/names/names.go
lifubang: Yes, thanks.

cli/names.go
thaJeztah: Yes, the leading slash is automatically added, and for historic reasons; a container is named `/containera`, and a container _linked_ to that container (using the legacy `--link` option on the default network) gets an internal name `/containera/containerb` (or something along those lines)

thaJeztah: I don't think it's allowed to manually _create_ a container with a slash though; they should be added automatically by the daemon where needed

thaJeztah (changes requested):
Thanks! I'm not sure if this is the right thing to do though; the general design is to defer validation to the daemon; make the daemon / API server responsible for checking which values are valid, and what values are _invalid_.

The docker CLI can be used against a different (newer/older) version of the daemon, which may accept different characters for "object" names.

In addition, this patch would only handle _containers_, but (see https://github.com/moby/moby/issues/29126) the same could apply to any other object.

The original issue reported here is not actualy due to an invalid _name_, but because of the way this name is sent to the daemon; the cli sends `redis/../redisps` as part of the URL, and this gets expanded/rewritten to `redisps`.

Here's some older discussions/issues about this; https://github.com/moby/moby/issues/29126 (suggested fix with URLencode; https://github.com/moby/moby/pull/29131, and some related PR's; https://github.com/moby/moby/pull/31831, and https://github.com/moby/moby/pull/32127).

I _think_ the correct fix should be to URLencode values before sending, however (I can't find that discussions, but can dig some more) there were concerns about changing that on the client, as it could result in an incompatible change if the daemon version does not handle this.

thaJeztah:
Thanks! Main thing to look into if URLEncoding is used, is what happens daemon-side, i.e.; the concern is to break compatibility with older daemons; _perhaps_ it's handled automatically (haven't checked), but perhaps not. If it's not handled automatically, URLEncode could still be possible, but in that case it should be versioned (API 1.xx: don't url-encode, API > 1.xx use url-encode)

lifubang:
Please see https://github.com/moby/moby/pull/37857
I think we can just fix docker/docker/client library.
After 37857, then to fix docker/cli ?

PR #1333, opened 2018-09-04
builder: add df command

tiborvass:
Depends on #1327 

Adds `docker builder df` to show only build cache usage.


tiborvass:
@AkihiroSuda Because I'm working on another PR to open soon to show all detailed information in `system df` by removing the conflict between `-v` and `--format`. There were a couple of issues with `builder df`: it was calling `/system/df`, no way of filtering with options. We'd rather do it right then rush it in.

Issue #1344, opened 2018-09-06
warning on building through ssh dialer


andrewhsu:
using latest docker/cli master codebase https://github.com/docker/cli/tree/561c47f

```
$ /tmp/docker -H ssh://Warmachine68@192.168.0.2 build .
[+] Building 1.0s (5/5) FINISHED                                                
 => [internal] load .dockerignore                                          0.3s
 => => transferring context: 2B                                            0.0s
 => [internal] load Dockerfile                                             0.3s
 => => transferring dockerfile: 31B                                        0.0s
 => [internal] load metadata for docker.io/library/alpine:latest           0.7s
 => CACHED [1/1] FROM docker.io/library/alpine@sha256:7043076348bf5040220  0.0s
 => exporting to image                                                     0.0s
 => => exporting layers                                                    0.0s
 => => writing image sha256:0b4ed36840824a86314da6264a95047c664f3c2fe90d0  0.0s
WARN[0001] commandConn.CloseWrite: signal: killed
```

cc @AkihiroSuda @tonistiigi

PR #1345, opened 2018-09-06
connhelper: try sending SIGTERM before SIGKILL

AkihiroSuda:

Fix https://github.com/docker/cli/issues/1344

@andrewhsu @tonistiigi @cpuguy83

### DIFF
--- a/cli/connhelper/connhelper.go
+++ b/cli/connhelper/connhelper.go
@@ -10,8 +10,10 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"runtime"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/docker/cli/cli/connhelper/ssh"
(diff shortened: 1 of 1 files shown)


cli/connhelper/connhelper.go
```diff
@@ -102,21 +102,30 @@ func (c *commandConn) killIfStdioClosed() error {
...
 	}
-	var err error
 	c.cmdMutex.Lock()
-	// NOTE: maybe already killed here
-	if err = c.cmd.Process.Kill(); err == nil {
-		err = c.cmd.Wait()
+	killErr := c.cmd.Process.Kill()
+	waitErr := c.cmd.Wait()
+	// NOTE: checking ProcessState before Wait() result in nil panic
+	exited := c.cmd.ProcessState.Exited()
```
cpuguy83: Can we read process state from `waitErr` instead? (e.g. `waitErr.(*exec.ExitError)`)
I'm not a big fan of the error handling here, and seems pretty platform dependent.
It also seems like we'd run into  the `signal: killed` error in the normal case?

Should we also consider sending TERM instead of KILL and only send kill if TERM doesn't exit after some time?

AkihiroSuda: addressed the comment, PTAL

AkihiroSuda:
@cpuguy83 PTAL?

PR #1356, opened 2018-09-08
fix make -f docker.Makefile lint hang error

max8899:
**- What I did**
make -f docker.Makefile lint hangs, fix it

**- How I did it**
add make lint to docker.Makefile lint part
**- How to verify it**
make -f docker.Makefile lint

### DIFF
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -79,7 +79,7 @@ shell: dev ## alias for dev
 
 .PHONY: lint
 lint: build_linter_image ## run linters
-	docker run -ti $(ENVVARS) $(MOUNTS) $(LINTER_IMAGE_NAME)
+	docker run -ti --rm $(ENVVARS) $(MOUNTS) $(LINTER_IMAGE_NAME) make lint
 
 .PHONY: vendor
 vendor: build_docker_image vendor.conf ## download dependencies (vendor/) listed in vendor.conf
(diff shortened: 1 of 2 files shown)


max8899:
@vdemeester in fact, it's not some protential hanging of `make -f docker.Makefile lint`, it happens every time and I was using the code on master. Add `make lint` as a command for `lint` part in Makefile.

vdemeester:
@max8899 `docker run -ti --rm $(ENVVARS) $(MOUNTS) $(LINTER_IMAGE_NAME) make lint` should fail really quick as the `linter` image declare `ENTRYPOINT ["/usr/local/bin/gometalinter"]`… so it's gonna try to execute `/usr/local/bin/gometalinter make lint` which will definitely fails.

What system are you using ? 
Asking tihs, because it's a *known* fact that running `make -f docker.Makefile lint` on docker-for-desktop tends to be either really slow or timing out… it's mainly due to the fact that we *bind mound* the source in the container, which can cause some heavy performance problem on docker-for-desktop (as it's not a simple bind mount like on linux…). Removing the bind mount would help (but would mean rebuilding the image each and every `make -f docker.Makefile lint` run, which shouldn't be that big of a deal as only the last layer would change).

silvin-lubecki:
ping @max8899 ? Did @vdemeester comment helped you and maybe then you can close this PR?

PR #1359, opened 2018-09-10
cli/config/configfile: Atomically rewrite the config file when saving.

tsuna:
The config file was being truncated first, which created a window during
which it was empty, causing concurrent uses of the `docker` command to
potentially fail with:
```
WARNING: Error loading config file: /var/lib/jenkins/.docker/config.json: EOF
Error response from daemon: Get https://registry/v2/foo/manifests/latest: no basic auth credentials
```

Signed-off-by: Benoit Sigoure <tsunanet@gmail.com>

The use of commands that change the Docker CLI config (e.g. `docker login`) concurrently with other commands could lead to the other commands not being able to read the configuration.

🐹

### DIFF
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -175,15 +175,21 @@ func (configFile *ConfigFile) Save() error {
 		return errors.Errorf("Can't save config with empty filename")
 	}
 
-	if err := os.MkdirAll(filepath.Dir(configFile.Filename), 0700); err != nil {
+	dir := filepath.Dir(configFile.Filename)
+	if err := os.MkdirAll(dir, 0700); err != nil {
 		return err
 	}
-	f, err := os.OpenFile(configFile.Filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	temp, err := ioutil.TempFile(dir, filepath.Base(configFile.Filename))
 	if err != nil {
 		return err
 	}
-	defer f.Close()
-	return configFile.SaveToWriter(f)
+	err = configFile.SaveToWriter(temp)
+	temp.Close()
+	if err != nil {
+		os.Remove(temp.Name())
+		return err
+	}
+	return os.Rename(temp.Name(), configFile.Filename)
 }
 
 // ParseProxyConfig computes proxy configuration by retrieving the config for the provided host and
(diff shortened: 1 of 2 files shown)


tsuna:
Seems like others have run into this before me as well: https://issues.jenkins-ci.org/browse/JENKINS-44228

jwatte:
I for one would like to be the first to welcome our new atomic file writing overlords!

tsuna:
The build failed due to some unrelated failure in an early `apt-get update` step.

cli/config/configfile/file_test.go
```diff
@@ -413,3 +415,13 @@ func TestCheckKubernetesConfigurationRaiseAnErrorOnInvalidValue(t *testing.T) {
...
 }
+
+func TestSave(t *testing.T) {
+	configFile := New("test-save")
+	defer os.Remove("test-save")
+	err := configFile.Save()
+	assert.NilError(t, err)
+	cfg, err := ioutil.ReadFile("test-save")
+	assert.NilError(t, err)
+	assert.Check(t, is.Equal(string(cfg), "{\n	\"auths\": {}\n}"))
```
tiborvass: Feels like an omitempty tag is missing in the `Auths` field.

tiborvass: We can fix in another PR

cli/config/configfile/file.go
```diff
@@ -175,15 +175,21 @@ func (configFile *ConfigFile) Save() error {
...
 	}
-	defer f.Close()
-	return configFile.SaveToWriter(f)
+	err = configFile.SaveToWriter(temp)
+	temp.Close()
+	if err != nil {
+		os.Remove(temp.Name())
+		return err
+	}
+	return os.Rename(temp.Name(), configFile.Filename)
```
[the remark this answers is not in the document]
vdemeester: I feel it's not that big of a deal, so I'm ok keeping as-is and do a follow-up if we need :angel:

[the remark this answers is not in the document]
thaJeztah: alrighty; let's get this merged; thanks!

cli/config/configfile/file_test.go
silvin-lubecki: nit: Review after merge (so maybe for the just opened follow-up) 😅:
You can use backquotes instead:
```go
assert.Check(t, is.Equal(string(cfg), `{
	"auths": {}
}`))
```

PR #1360, opened 2018-09-10
Prefix engine repo with store

dhiltgen:
The official access point for the Q3 engine images will
be prefixed by store.

### DIFF
--- a/cli/command/engine/activate.go
+++ b/cli/command/engine/activate.go
@@ -56,7 +56,7 @@ https://hub.docker.com/ then specify the file with the '--license' flag.
 
 	flags.StringVar(&options.licenseFile, "license", "", "License File")
 	flags.StringVar(&options.version, "version", "", "Specify engine version (default is to use currently running version)")
-	flags.StringVar(&options.registryPrefix, "registry-prefix", "docker.io/docker", "Override the default location where engine images are pulled")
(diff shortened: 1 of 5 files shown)


tescherm:
will do

thaJeztah:
> docker.io/docker will still be valid right ? (so that 18.03 cli will, in mid-term, still be able to init, …)

Ah, had the same question, but notice I didn't post it; is this an alias, or is the repository being moved?

dhiltgen:
> `docker.io/docker` will still be valid right ? (so that 18.03 cli will, in mid-term, still be able to `init`, …)

This prefix wont work through the license entitlement flow on UCP so Enterprise customers wont get access to the repo correctly.  The `/store/` prefix is needed to trigger the proper entitlement checks on the back-end.

dhiltgen:
> Ah, had the same question, but notice I didn't post it; is this an alias, or is the repository being moved?

It is an alias on the hub/store side to trigger the entitlement checking flow.

thaJeztah:
oh, needs a rebase; let me carry so that we can move forward

thaJeztah:
carried in https://github.com/docker/cli/pull/1368

mavenugo:
Closing this as https://github.com/docker/cli/pull/1368 is merged

PR #1361, opened 2018-09-11
Remove containerizedengine package dependency from docker/cli/command…

vdemeester:
… this removes a whole lot of dependencies from people depending on docker/cli…

We should not leak `internal/containerizedcontainer` (and its dependencies) for users/projects that depend on `docker/cli`.

PS: I really think it should be backported to 18.03 too :angel: 

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -19,8 +19,8 @@ import (
 	manifeststore "github.com/docker/cli/cli/manifest/store"
 	registryclient "github.com/docker/cli/cli/registry/client"
 	"github.com/docker/cli/cli/trust"
-	"github.com/docker/cli/internal/containerizedengine"
 	dopts "github.com/docker/cli/opts"
+	clitypes "github.com/docker/cli/types"
 	"github.com/docker/docker/api"
 	"github.com/docker/docker/api/types"
 	registrytypes "github.com/docker/docker/api/types/registry"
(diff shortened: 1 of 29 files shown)


cli/command/cli.go
```diff
@@ -252,8 +252,8 @@ type ClientInfo struct {
 }
 
 // NewDockerCli returns a DockerCli instance with IO output and error streams set by in, out and err.
-func NewDockerCli(in io.ReadCloser, out, err io.Writer, isTrusted bool) *DockerCli {
-	return &DockerCli{in: NewInStream(in), out: NewOutStream(out), err: err, contentTrust: isTrusted}
+func NewDockerCli(in io.ReadCloser, out, err io.Writer, isTrusted bool, containerizedFn func(string) (ContainerizedClient, error)) *DockerCli {
```
vdemeester: This should probably be refactor with a config and some functionnal argument too

cli/command/cli.go
```diff
@@ -55,20 +55,21 @@ type Cli interface {
...
-	contentTrust bool
+	configFile            *configfile.ConfigFile
+	in                    *InStream
+	out                   *OutStream
+	err                   io.Writer
+	client                client.APIClient
+	serverInfo            ServerInfo
+	clientInfo            ClientInfo
+	contentTrust          bool
+	newContainerizeClient func(string) (clitypes.ContainerizedClient, error)
```
silvin-lubecki: Not a fan of this trick...

PR #1366, opened 2018-09-12
cli/config/configfile: Make sure to not leak a temp config behind and improve unit tests

tsuna:
This is a small followup change to #1359.  It addresses the astute observation from @thaJeztah that we can potentially leak temp files in the config directory.  While I couldn't easily test this particular failure from the unit test, I took the opportunity to expand the unit test to cover a few failure modes.

cc @vdemeester — thanks guys.

🦋

### DIFF
--- a/cli/config/configfile/file_test.go
+++ b/cli/config/configfile/file_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/docker/cli/cli/config/credentials"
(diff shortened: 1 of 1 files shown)


cli/config/configfile/file.go
```diff
@@ -183,10 +183,10 @@ func (configFile *ConfigFile) Save() error {
 	if err != nil {
 		return err
 	}
+	defer os.Remove(temp.Name())
```
thaJeztah: this would also attempt to remove `temp.Name()` in the happy case (i.e., it being succesfully renamed); guess that's not a problem.

thaJeztah (dismissed):
LGTM

vdemeester (commented):
Hum, the new test is failing though :sweat:

thaJeztah:
oh! hm...

tsuna:
I haven't yet had time to figure out why this is failing on CirleCI, it's as if the test's attempt to make the test directory read-only wasn't actually making it read-only... maybe they do something funky with the filesystem there. 🤔

I'll follow up as soon as I have a few free cycles, sorry guys.

thaJeztah:
Could it be because the tests run docker-in-docker, and tests are running as `root`?

PR #1368, opened 2018-09-12
[18.09] Prefix engine repo with store

thaJeztah:
carry of https://github.com/docker/cli/pull/1360
closes https://github.com/docker/cli/pull/1360

The official access point for the Q3 engine images will be prefixed by store.

### DIFF
--- a/cli/command/engine/activate.go
+++ b/cli/command/engine/activate.go
@@ -56,7 +56,7 @@ https://hub.docker.com/ then specify the file with the '--license' flag.
 
 	flags.StringVar(&options.licenseFile, "license", "", "License File")
 	flags.StringVar(&options.version, "version", "", "Specify engine version (default is to use currently running version)")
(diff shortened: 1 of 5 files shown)


cli/command/engine/activate.go
```diff
@@ -56,7 +56,7 @@ https://hub.docker.com/ then specify the file with the '--license' flag.
 
 	flags.StringVar(&options.licenseFile, "license", "", "License File")
 	flags.StringVar(&options.version, "version", "", "Specify engine version (default is to use currently running version)")
-	flags.StringVar(&options.registryPrefix, "registry-prefix", "docker.io/docker", "Override the default location where engine images are pulled")
+	flags.StringVar(&options.registryPrefix, "registry-prefix", "docker.io/store/docker", "Override the default location where engine images are pulled")
```
silvin-lubecki: What about extracting this string as a const? It seems to be copied everywhere. It can be in a follow-up, extracting the other strings like it.

PR #1369, opened 2018-09-13
formatter package heavy refactoring

vdemeester:
What this PR does is essentially moving the "formatter" implementation of each type in their respective packages. The *main* reason to do that, is to be able to depend on `cli/command/formatter` without depending of the implementation detail of the formatter. As of now, depending on `cli/command/formatter` means we depend on `docker/docker/api/types`, `docker/licensing`, … — that should not be the case. `formatter` should hold the common code (or helpers) to easily create formatter, not all formatter implementations.

- make it possible to extract the formatter implementation from the "common" code, that way, the formatter package stays small
- extract some formatter into their own packages

There is still a few formatter to migrate (`container`, `volume`, `image` and `buildcache`) — I didn't do in the PR because 1) it's a big PR already, and 2) they are interconnected with each other (`disk_usage` uses them all)


tiborvass:
Thanks @vdemeester ! If you don't mind I'd love to wait for https://github.com/docker/cli/pull/1370 to be merged first ^^

internal/test/strings.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"strings"
+	"testing"
+
+	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
+)
+
+// CompareMultipleValues compares comma-separated values, whatever the order is
+func CompareMultipleValues(t *testing.T, value, expected string) {
```
cpuguy83: Not sure about this function name, or quite what this is for.

vdemeester: Yeah… I moved code around, kept the name

PR #1371, opened 2018-09-13
LCOW: --platform on import (already in API)

jhowardmsft:

A tester here found import doesn't work for LCOW. No idea how I missed this - it's in the API, just not in the CLI. As the tar generated in export doesn't have any indication of platform, it needs to be explicit. See how in the before-fix case, it's going down the WCOW route (windowsfilter),

Before:
```
PS E:\> docker import test.tar
Error response from daemon: re-exec error: exit status 1: output: ProcessBaseLayer \\?\C:\control\windowsfilter\db15e89fc696ce2b7e8a9c7ee328553690cec3558fd01cb843c837b230e2e1f2: The system cannot find the path specified.
```

After:
```
PS E:\> docker import --platform=linux test.tar
sha256:f68c6c11dcd5f4b4bbdc71d07fb5712325b00ffaa546ec868a0775bd8d62d20a
```

### DIFF
--- a/cli/command/image/import.go
+++ b/cli/command/image/import.go
@@ -19,6 +19,7 @@ type importOptions struct {
 	reference string
 	changes   dockeropts.ListOpts
 	message   string
+	platform  string
 }
 
 // NewImportCommand creates a new `docker import` command
@@ -43,6 +44,7 @@ func NewImportCommand(dockerCli command.Cli) *cobra.Command {
 	options.changes = dockeropts.NewListOpts(nil)
 	flags.VarP(&options.changes, "change", "c", "Apply Dockerfile instruction to the created image")
 	flags.StringVarP(&options.message, "message", "m", "", "Set commit message for imported image")
+	command.AddPlatformFlag(flags, &options.platform)
 
 	return cmd
 }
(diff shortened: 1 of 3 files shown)


silvin-lubecki:
@thaJeztah do you think https://github.com/docker/cli/blob/master/man/src/image/import.md and https://github.com/docker/cli/blob/master/docs/reference/commandline/import.md should be updated accordingly?

jhowardmsft:
Which repo are the docs in again?

jhowardmsft:
Updated docs.

cli/command/image/import.go
```diff
@@ -43,6 +44,7 @@ func NewImportCommand(dockerCli command.Cli) *cobra.Command {
 	options.changes = dockeropts.NewListOpts(nil)
 	flags.VarP(&options.changes, "change", "c", "Apply Dockerfile instruction to the created image")
 	flags.StringVarP(&options.message, "message", "m", "", "Set commit message for imported image")
+	command.AddPlatformFlag(flags, &options.platform)
```
thaJeztah: For future reference (and save some digging); `command.AddPlatformFlag()` also sets the annotations for API-version and "experimental" being enabled.

PR #1381, opened 2018-09-20
[18.09] `docker engine` updates

dhiltgen:
- Updates the containerd client code
- Removes the `docker engine init` subcommand
- Removes the `docker engine rm` subcommand
- Disabled e2e tests for the `docker engine` features
- Changes the `docker engine activate` and `docker engine update` commands to be only accessible by `root` (instead of _trying_, and printing an error if permissions were insufficient)
- Changes the containerd namespace for the engine from `docker` to `com.docker`
- Introduces an `/var/lib/docker/distribution_based_engine.json` file in which we persist information about the engine
- Does some other refactors and changes

Will need a forward port to master.


cli/command/engine/cmd.go
```diff
@@ -15,11 +15,9 @@ func NewEngineCommand(dockerCli command.Cli) *cobra.Command {
 		RunE:  command.ShowHelp(dockerCli.Err()),
 	}
 	cmd.AddCommand(
```
vdemeester: Why are `init` and `rm` command removed ? :thinking:

cli/command/engine/activate.go
```diff
@@ -104,12 +107,17 @@ func runActivate(cli command.Cli, options activateOptions) error {
 		EngineVersion:  options.version,
 	}
 
-	return client.ActivateEngine(ctx, opts, cli.Out(), authConfig,
+	err = client.ActivateEngine(ctx, opts, cli.Out(), authConfig,
```
silvin-lubecki: nit: if err := ...; err != nil{
    return err
}

internal/containerizedengine/update.go
```diff
@@ -102,30 +72,136 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
...
+	err = c.cclient.Install(ctx, image, containerd.WithInstallReplace, containerd.WithInstallPath("/usr"))
+	if err != nil {
 		return err
 	}
 
-	// TODO verify the image has changed and don't update if nothing has changed
+	return c.WriteRuntimeMetadata(ctx, "", newMetadata)
+}
+
+var defaultDockerRoot = "/var/lib/docker"
```
thaJeztah: This should really be a `const`

internal/containerizedengine/update_test.go
```diff
@@ -292,30 +156,112 @@ func TestDoUpdatePullFail(t *testing.T) {
...
-			},
 		},
 	}
-	err := client.DoUpdate(ctx, opts, command.NewOutStream(&bytes.Buffer{}), &types.AuthConfig{}, healthfnHappy)
-	assert.ErrorContains(t, err, "unable to find existing engine")
+	opts := clitypes.EngineInitOptions{
+		EngineVersion:  "engineversiongoeshere",
+		RegistryPrefix: "registryprefixgoeshere",
+		ConfigFile:     "/tmp/configfilegoeshere",
+		//EngineImage:    clitypes.EnterpriseEngineImage,
```
thaJeztah: Should be removed here

internal/containerizedengine/update_test.go
```diff
@@ -292,30 +156,112 @@ func TestDoUpdatePullFail(t *testing.T) {
...
+	}
+
+	tmpdir, err := ioutil.TempDir("", "docker-root")
+	assert.NilError(t, err)
+	defer os.RemoveAll(tmpdir)
+	defaultDockerRoot = tmpdir
```
thaJeztah: This is probably the reason it isn't a `const`. I'm _really_ not a fan of this; also, looks like we don't change this back to its original value, so anything after this test could fail.

internal/containerizedengine/types.go
```diff
@@ -13,10 +13,7 @@ import (
 const (
 	containerdSockPath  = "/run/containerd/containerd.sock"
 	engineContainerName = "dockerd"
-	engineNamespace     = "docker"
-
-	// Used to signal the containerd-proxy if it should manage
-	proxyLabel = "com.docker/containerd-proxy.scope"
+	engineNamespace     = "com.docker"
```
thaJeztah: What's the context for this change?

internal/containerizedengine/types.go
```diff
@@ -67,6 +64,12 @@ var (
 			NoNewPrivileges: false,
 		},
 	}
+
+	// RuntimeMetadataName is the name of the runtime metadata file
+	RuntimeMetadataName = "distribution_based_engine"
+
+	// ReleaseNotePrefix is where to point users to for release notes
+	ReleaseNotePrefix = "https://docs.docker.com/releasenotes"
```
thaJeztah: This should be a `const`.  It's also not used outside this package, so should not be exported

internal/containerizedengine/update.go
```diff
@@ -102,30 +72,136 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
...
-		ctx, cancel := context.WithTimeout(ctx, engineWaitTimeout)
-		defer cancel()
-		return c.waitForEngine(ctx, out, healthfn)
-	})
-	if err == nil && opts.Scope != "" {
-		var labels map[string]string
-		labels, err = engine.Labels(ctx)
+// GetCurrentRuntimeMetadata loads the current daemon runtime metadata information from the local host
+func (c *baseClient) GetCurrentRuntimeMetadata(_ context.Context, dockerRoot string) (*RuntimeMetadata, error) {
+	if dockerRoot == "" {
```
thaJeztah: Looks like the `dockerRoot` argument is only used in tests, but otherwise `/var/lib/docker` is hard-coded. 

- Should this argument be there if it's only used for testing?
- Is hardcoding `/var/lib/docker` a concern here? Using a different data-dir is a very common use-case, and we can obtain the correct data-dir from the daemon

internal/containerizedengine/update.go
```diff
@@ -102,30 +72,136 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
...
+
+		if err := json.Unmarshal(p, &ociimage); err != nil {
+			return nil, err
+		}
+		config = ociimage.Config
+	default:
+		return nil, fmt.Errorf("unknown image %s config media type %s", image.Name(), ic.MediaType)
+	}
+
+	metadataString, ok := config.Labels[RuntimeMetadataName]
```
thaJeztah: If we're storing this information as labels on the image, these labels should be namespaced (e.g., `com.docker.<something>`)

internal/containerizedengine/types.go
```diff
@@ -67,6 +64,12 @@ var (
 			NoNewPrivileges: false,
 		},
 	}
+
+	// RuntimeMetadataName is the name of the runtime metadata file
+	RuntimeMetadataName = "distribution_based_engine"
```
thaJeztah: This should be a `const`. It's also not used outside this package, so should not be exported

internal/containerizedengine/update.go
```diff
@@ -102,30 +72,136 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
...
+		lv, err := ver.NewVersion(localVersion.Version)
+		if err != nil {
+			return nil, err
+		}
+		mv, err := ver.NewVersion(metadata.ContainerdMinVersion)
+		if err != nil {
+			return nil, err
+		}
+		if lv.LessThan(mv) {
+			return nil, fmt.Errorf("local containerd is too old: %s - this engine version requires %s or newer.\nPlease consult the release notes at %s for upgrade instructions",
```
thaJeztah: Remove "Please" here

internal/containerizedengine/update.go
```diff
@@ -102,30 +72,136 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
...
+		}
+	} // If omitted on metadata, no hard dependency on containerd version beyond 18.09 baseline
+
+	// All checks look OK, proceed with update
+	return &metadata, nil
+}
+
+// GetReleaseNotesURL returns a release notes url
+// If the image name does not contain a version tag, the base release notes URL is returned
+func (c *baseClient) GetReleaseNotesURL(imageName string) string {
```
thaJeztah: Why was this put as a method on `baseClient`? I see this utility is only used here locally, so could just be an un-exported function (`func releaseNotesUrl(imageName string) string`)

internal/containerizedengine/update.go
```diff
@@ -102,30 +72,136 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
...
+	if err == nil {
+		taggedRef, ok := distributionRef.(reference.NamedTagged)
+		if ok {
+			versionTag = taggedRef.Tag()
 		}
-		labels[proxyLabel] = opts.Scope
-		_, err = engine.SetLabels(ctx, labels)
 	}
-	return err
+	return fmt.Sprintf("%s/%s", ReleaseNotePrefix, versionTag)
```
thaJeztah: Actually, instead of attempting to reconstruct an URL here; should we just add a `release-notes` URL as metadata on the image? https://github.com/docker/docker-ce-packaging/blob/v18.09.0-ce-beta1/image/Dockerfile.engine#L82-L90

cli/command/engine/check.go
```diff
@@ -50,54 +51,47 @@ func newCheckForUpdatesCommand(dockerCli command.Cli) *cobra.Command {
 }
 
 func runCheck(dockerCli command.Cli, options checkOptions) error {
-	ctx := context.Background()
-	client, err := dockerCli.NewContainerizedEngineClient(options.sockPath)
-	if err != nil {
-		return errors.Wrap(err, "unable to access local containerd")
+	if !isRoot() {
+		return errors.New("must be privileged to activate engine")
```
silvin-lubecki: Shouldn't the message be something like "must be privileged to check the engine" ?

internal/containerizedengine/update.go
```diff
@@ -102,30 +72,136 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
...
+		return nil, fmt.Errorf("image %s does not contain runtime metadata label %s", image.Name(), RuntimeMetadataName)
+	}
+	err = json.Unmarshal([]byte(metadataString), &metadata)
+	if err != nil {
+		return nil, errors.Wrapf(err, "malformed runtime metadata file in %s", image.Name())
+	}
+
+	// Current CLI only supports host install runtime
+	if metadata.Runtime != "host_install" {
+		return nil, fmt.Errorf("unsupported runtime: %s\nPlease consult the release notes at %s for upgrade instructions", metadata.Runtime, c.GetReleaseNotesURL(image.Name()))
```
thaJeztah: Remove "Please" here.

thaJeztah: "unsupported runtime" is also a bit overloaded term (I'd interpret  "runtime" as `containerd` or `runc`, or `cri-o` or...)

internal/containerizedengine/update.go
```diff
@@ -102,30 +72,136 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
...
+	newMetadata, err := c.PreflightCheck(ctx, image)
 	if err != nil {
-		if err == ErrEngineNotPresent {
-			return errors.Wrap(err, "unable to find existing engine - please use init")
+		return err
+	}
+	// Grab current metadata for comparison purposes
+	if localMetadata != nil {
+		if localMetadata.Platform != newMetadata.Platform {
+			fmt.Fprintf(out, "\nNotice: you have switched to \"%s\".  Please refer to %s for update instructions.\n\n", newMetadata.Platform, c.GetReleaseNotesURL(imageName))
```
thaJeztah: Remove "Please" here

Also "Notice:" gives a feel of "something went wrong!" not sure if that's the intent here 😅

thaJeztah:
**warning** Github doesn't like many review comments, so some of my comments are hidden "out of the box" 😞 

screen shot 2018-09-20 at 15 38 14

cli/command/engine/activate_unix.go
```diff
@@ -0,0 +1,13 @@
+// +build !windows
+
+package engine
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+var (
+	isRoot = func() bool {
```
silvin-lubecki: I don't think putting a function as global variable to swap it in tests is a clean way to mock things. It does not scale, it's not future proof as any piece of code can erase it without notice. You should use an interface instead, or refactor your code so it is testable without this kind of "trick".

cli/command/engine/check_test.go
```diff
@@ -18,126 +20,87 @@ var (
...
+func TestCheckForUpdatesNoCurrentVersion(t *testing.T) {
+	isRoot = func() bool { return true }
+	c := test.NewFakeCli(&verClient{client.Client{}, types.Version{}, nil})
+	c.SetRegistryClient(testRegistryClient{})
+	cmd := newCheckForUpdatesCommand(c)
 	cmd.SilenceUsage = true
 	cmd.SilenceErrors = true
 	err := cmd.Execute()
-	assert.Assert(t, err == retErr)
+	assert.ErrorContains(t, err, "alformed version")
```
silvin-lubecki: typo: malformed

cli/command/engine/check_test.go
```diff
@@ -18,126 +20,87 @@ var (
...
 	cmd.Flags().Set("downgrades", "true")
+	cmd.SilenceUsage = true
+	cmd.SilenceErrors = true
 	err := cmd.Execute()
 	assert.NilError(t, err)
-	golden.Assert(t, testCli.OutBuffer().String(), "check-all.golden")
+	golden.Assert(t, c.OutBuffer().String(), "check-all.golden")
 
-	testCli.OutBuffer().Reset()
+	c.OutBuffer().Reset()
```
silvin-lubecki: All these 4 blocks could be run as sub tests, using a table test case pattern.

cli/command/engine/init.go
```diff
@@ -1,62 +1,10 @@
...
-	"context"
-
-	"github.com/docker/cli/cli"
-	"github.com/docker/cli/cli/command"
 	clitypes "github.com/docker/cli/types"
-	"github.com/pkg/errors"
-	"github.com/spf13/cobra"
 )
 
 type extendedEngineInitOptions struct {
```
silvin-lubecki: I guess this type could be renamed as engineUpdateOptions, and moved to the update command, so you can remove this file?

internal/containerizedengine/update.go
```diff
@@ -87,6 +43,20 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
 		return fmt.Errorf("please pick the version you want to update to")
 	}
 
+	localMetadata, err := c.GetCurrentRuntimeMetadata(ctx, "")
+	if err == nil {
+		if opts.EngineImage == "" {
+			if strings.Contains(strings.ToLower(localMetadata.Platform), "enterprise") {
+				opts.EngineImage = "engine-enterprise"
+			} else {
+				opts.EngineImage = "engine-community"
```
silvin-lubecki: These should be extracted as constants and documented.

internal/containerizedengine/update.go
```diff
@@ -102,30 +72,136 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
...
+		return err
+	}
+	// Grab current metadata for comparison purposes
+	if localMetadata != nil {
+		if localMetadata.Platform != newMetadata.Platform {
+			fmt.Fprintf(out, "\nNotice: you have switched to \"%s\".  Please refer to %s for update instructions.\n\n", newMetadata.Platform, c.GetReleaseNotesURL(imageName))
 		}
+	}
+
+	err = c.cclient.Install(ctx, image, containerd.WithInstallReplace, containerd.WithInstallPath("/usr"))
```
silvin-lubecki: nit: if err := ... ; err != nil

internal/versions/versions.go
```diff
@@ -24,7 +28,23 @@ func (c *baseClient) GetEngineVersions(ctx context.Context, registryClient regis
 		return clitypes.AvailableVersions{}, err
 	}
 
-	return parseTags(tags, currentVersion)
+	return parseTags(tags, serverVersion.Version)
+}
+
+func getEngineImage(registryPrefix string, serverVersion types.Version) string {
+	communityImage := "engine-community"
```
silvin-lubecki: Extract these as constants

vendor.conf
```diff
@@ -3,7 +3,7 @@ github.com/asaskevich/govalidator f9ffefc3facfbe0caee3fea233cbb6e8208f4541
 github.com/Azure/go-ansiterm d6e3b3328b783f23731bc4d058875b0371ff8109
 github.com/beorn7/perks 3a771d992973f24aa725d07868b467d1ddfceafb
 github.com/containerd/console c12b1e7919c14469339a5d38f2f8ed9b64a9de23
-github.com/containerd/containerd v1.2.0-beta.2
+github.com/containerd/containerd bb0f83ab6eec47c3316bb763d5c20a82c7750c31
```
silvin-lubecki: Do we really want to rely on an unstable version of containerd?

cli/command/engine/activate.go
```diff
@@ -67,6 +67,9 @@ https://hub.docker.com/ then specify the file with the '--license' flag.
 }
 
 func runActivate(cli command.Cli, options activateOptions) error {
+	if !isRoot() {
```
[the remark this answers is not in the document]
dhiltgen: containerd socket is owned by root.

cli/command/engine/activate_windows.go
```diff
@@ -0,0 +1,9 @@
+// +build windows
+
+package engine
+
+var (
+	isRoot = func() bool {
```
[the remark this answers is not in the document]
dhiltgen: We'd like to support this on windows in a future release.

cli/command/engine/update.go
```diff
@@ -32,22 +32,18 @@ func newUpdateCommand(dockerCli command.Cli) *cobra.Command {
 }
 
 func runUpdate(dockerCli command.Cli, options extendedEngineInitOptions) error {
+	if !isRoot() {
```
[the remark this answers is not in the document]
dhiltgen: needs root for now.

internal/containerizedengine/update.go
```diff
@@ -102,30 +72,136 @@ func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptio
...
-	err = containerized.AtomicImageUpdate(ctx, engine, image, func() error {
-		ctx, cancel := context.WithTimeout(ctx, engineWaitTimeout)
-		defer cancel()
-		return c.waitForEngine(ctx, out, healthfn)
-	})
-	if err == nil && opts.Scope != "" {
-		var labels map[string]string
-		labels, err = engine.Labels(ctx)
+// GetCurrentRuntimeMetadata loads the current daemon runtime metadata information from the local host
+func (c *baseClient) GetCurrentRuntimeMetadata(_ context.Context, dockerRoot string) (*RuntimeMetadata, error) {
```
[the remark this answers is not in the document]
dhiltgen: Was trying to be consistent, but it does seem a bit silly now, so I'll remove it.

docker.Makefile
```diff
@@ -105,7 +105,7 @@ shellcheck: build_shell_validate_image ## run shellcheck validation
 	docker run -ti --rm $(ENVVARS) $(MOUNTS) $(VALIDATE_IMAGE_NAME) make shellcheck
 
 .PHONY: test-e2e ## run e2e tests
-test-e2e: test-e2e-non-experimental test-e2e-experimental test-e2e-containerized
```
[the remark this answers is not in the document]
dhiltgen: They depended on the old init flow which we didn't port over to the updated model.  It's possible we might be able to come up with a new e2e test flow for this, but given other bugs we've been hitting on the packaging side, I don't think those tests belong in this tree, since it's not actually helping us find the interesting bugs.

internal/containerizedengine/update.go
dhiltgen: no, I needed to change it for testing purposes.

internal/containerizedengine/update_test.go
dhiltgen: I'll flip it back on the testcase.

internal/containerizedengine/types.go
dhiltgen: Hmm... not sure - I didn't write that part - let me poke around a bit.

internal/containerizedengine/types.go
dhiltgen: I wanted to use this elsewhere, but realized that's going to mess up the recent changes to abstract the dependency and didn't have time to wire it through.  I really think that refactoring was the wrong thing to do, but I'm trying not to break it if I can avoid it.

internal/containerizedengine/update.go
dhiltgen: Lets revisit this in a follow up - agreed the dockerroot is going to be something we need to handle.

internal/containerizedengine/update.go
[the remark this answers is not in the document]
dhiltgen: Good catch.  I'll flip it around so the failure mode is consistent.

internal/containerizedengine/update.go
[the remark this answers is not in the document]
dhiltgen: Actually, once we try to wire up dockerroot lookup, it'll be needed, so I'm going to leave it for now and as it'll be needed in a follow up.

internal/versions/versions.go
```diff
@@ -24,7 +28,23 @@ func (c *baseClient) GetEngineVersions(ctx context.Context, registryClient regis
...
+	enterpriseImage := "engine-enterprise"
+	platform := strings.ToLower(serverVersion.Platform.Name)
+	if platform != "" {
+		if strings.Contains(platform, "enterprise") {
+			return path.Join(registryPrefix, enterpriseImage)
+		}
+		return path.Join(registryPrefix, communityImage)
+	}
+	if strings.Contains(serverVersion.Version, "ee") {
```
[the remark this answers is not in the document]
dhiltgen: The "ee" string lookup could theoretically be removed from this flow too, but since we're in a transitional state on the packaging some older tp builds still have ee version strings and didn't have the platform populated, so I was trying to harden this specific flow.

I'll put a comment in the code that this can be removed once we no longer have to worry about the earlier 18.09 builds that had the packaging bugs.

cli/command/engine/update.go
```diff
@@ -32,22 +32,18 @@ func newUpdateCommand(dockerCli command.Cli) *cobra.Command {
...
-		currentOpts, err := client.GetCurrentEngineVersion(ctx)
-		if err != nil {
-			return err
-		}
-		if options.EngineImage == "" {
-			options.EngineImage = currentOpts.EngineImage
-		}
 		if options.RegistryPrefix == "" {
-			options.RegistryPrefix = currentOpts.RegistryPrefix
+			options.RegistryPrefix = "docker.io/store/docker"
```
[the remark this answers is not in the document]
dhiltgen: Good catch - the logic on this one was supposed to default to the prefix you're actively using, but with the new install model, that information is lost, so we should just default it clearly in the CLI args on both sides.  In the future we can refine this to default to the locally running image.

internal/containerizedengine/update_test.go
dhiltgen: this'll most likely disappear in a follow up to get dockerroot wired up to determine it dynamically.

types/types.go
```diff
@@ -36,9 +35,6 @@ type ContainerizedClient interface {
 		out OutStream,
 		authConfig *types.AuthConfig,
 		healthfn func(context.Context) error) error
-	GetEngineVersions(ctx context.Context, registryClient registryclient.RegistryClient, currentVersion, imageName string) (AvailableVersions, error)
-	GetCurrentEngineVersion(ctx context.Context) (EngineInitOptions, error)
-	RemoveEngine(ctx context.Context) error
```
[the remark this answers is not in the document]
dhiltgen: Turns out there's a lot more dead code now so I've purged all I could find.

internal/containerizedengine/types.go
dhiltgen: It's still used, and puts the image into the com.docker namespace on containerd.

internal/containerizedengine/types.go
dhiltgen: found a better way to move this up without breaking the abstraction.

internal/containerizedengine/update.go
dhiltgen: I wanted to use it in other places, but then ran into the abstraction layer thing and stopped, so I'll move this back to an unexported function.

internal/containerizedengine/update.go
dhiltgen: We don't want to have to download all the images on `check` to see their metadata.  We also don't want customers to have to bounce around in different places to find release notes, but rather want a common place that always works.

dhiltgen:
> I attempted to review some of the changes here, but there's a lot of things being changed in this PR, without much context, which makes this difficult to review, and check if the code matches the _intent_.

Sorry we didn't loop you in earlier @thaJeztah - this was a refactoring pass based on proposals from various folks internally on how to address a number of P0 defects we've got that are holding up beta2.  In short, we've run into more corner cases than we anticipated with the original design of running using the host process runtime, and have switched to using the containerd install model to de-risk the release.  The goal is to keep the update/activate UX, but we had to back off the other less critical UX goals for this release, and we'll come back to those capabilities in a future version.

thaJeztah:
Some linting failures;

```
internal/containerizedengine/types.go:13:1:warning: engineContainerName is unused (deadcode)
internal/containerizedengine/types.go:23:1:warning: engineSpec is unused (deadcode)
cli/command/engine/activate.go:118:26:warning: "Succesfully" is a misspelling of "Successfully" (misspell)
cli/command/engine/update.go:58:32:warning: "Succesfully" is a misspelling of "Successfully" (misspell)
internal/containerizedengine/client_test.go:53:2:warning: type fakeTask is unused (U1000) (unused)
internal/containerizedengine/types.go:15:2:warning: const engineContainerName is unused (U1000) (unused)
internal/containerizedengine/types.go:39:2:warning: var engineSpec is unused (U1000) (unused)
```

cli/command/engine/activate_windows.go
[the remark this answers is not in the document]
thaJeztah: Yeah, but it's not there. Coding "because someday it may be needed" is resulting can lead to many issues. We don't know _when_ it arrives, and if the implementation will be as we anticipate it to be right now. Not a blocker for me, but definitely not a fan.

internal/containerizedengine/update.go
thaJeztah: I saw `PreflightCheck` was getting metadata from the image, so thought it would get the labels from the image anyway, so if it had a label containing the release-notes for that image, we could get it from there (but perhaps not in all cases)

internal/containerizedengine/types.go
thaJeztah: Not a big concern, but will this affect people that upgrade from beta1 to beta2 (if beta1 started the container in the old namespace)?

crosbymichael:
LGTM

vendor.conf
crosbymichael: for a more detailed answer, this is for the containerd client.  We have always pined to commits for the client.  The engine pins to a daemon version that is usually a tagged release

dhiltgen:
I'm going to try to land a follow up PR today to address some of the other review comments, but I'm blocked on some of the packaging changes that are in flight, so it might not land until Monday.  (Shouldn't gate a beta1.5 build, but I'm hoping to get it through before we cut a beta2)

PR #1387, opened 2018-09-22
[18.09] Expose licensing details before loading

dhiltgen:
Expose licensing details before loading

Help the user understand which license they're about
to load in case they have multiple licenses they need to
figure out.
    
### DIFF
--- a/cli/command/engine/activate_test.go
+++ b/cli/command/engine/activate_test.go
@@ -2,9 +2,6 @@ package engine
 
 import (
 	"fmt"
-	"io/ioutil"
-	"os"
-	"path/filepath"
 	"testing"
 
 	"github.com/docker/cli/internal/test"
@@ -12,6 +9,7 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
 	"gotest.tools/assert"
+	"gotest.tools/fs"
 	"gotest.tools/golden"
 )
 
@@ -51,12 +49,9 @@ func TestActivateBadLicense(t *testing.T) {
 }
 
 func TestActivateExpiredLicenseDryRun(t *testing.T) {
-	tmpdir, err := ioutil.TempDir("", "license")
-	assert.NilError(t, err)
-	defer os.RemoveAll(tmpdir)
-	filename := filepath.Join(tmpdir, "docker.lic")
-	err = ioutil.WriteFile(filename, []byte(expiredLicense), 0644)
-	assert.NilError(t, err)
+	dir := fs.NewDir(t, "license", fs.WithFile("docker.lic", expiredLicense, fs.WithMode(0644)))
+	defer dir.Remove()
+	filename := dir.Join("docker.lic")
 	isRoot = func() bool { return true }
 	c := test.NewFakeCli(&verClient{client.Client{}, types.Version{}, nil})
 	c.SetContainerizedEngineClient(
@@ -70,7 +65,7 @@ func TestActivateExpiredLicenseDryRun(t *testing.T) {
 	cmd.Flags().Set("license", filename)
 	cmd.Flags().Set("display-only", "true")
 	c.OutBuffer().Reset()
-	err = cmd.Execute()
+	err := cmd.Execute()
 	assert.NilError(t, err)
 	golden.Assert(t, c.OutBuffer().String(), "expired-license-display-only.golden")
 }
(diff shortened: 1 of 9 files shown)


internal/licenseutils/utils.go
```diff
@@ -159,12 +160,38 @@ func LoadLocalIssuedLicense(ctx context.Context, filename string) (*model.Issued
 	return doLoadLocalIssuedLicense(ctx, filename, lclient)
 }
 
+// GetLicenseSummary summarizes the license for the user
+func GetLicenseSummary(ctx context.Context, license model.IssuedLicense) (string, error) {
```
silvin-lubecki: I see a lot of code duplication between `GetLicenseSummary` and `LoadLocalIssuedLicense/doLoadLocalIssuedLicense `. I guess a function which creates a new `licensing.Config` and returns a verified license could be extracted here.

dhiltgen:
@silvin-lubecki unit test coverage added and addressed most of the comments - mind another quick peek when you have a minute?

cli/command/engine/activate_test.go
```diff
@@ -37,3 +49,28 @@ func TestActivateBadLicense(t *testing.T) {
...
 	assert.Error(t, err, "open invalidpath: no such file or directory")
 }
+
+func TestActivateExpiredLicenseDryRun(t *testing.T) {
+	tmpdir, err := ioutil.TempDir("", "license")
+	assert.NilError(t, err)
+	defer os.RemoveAll(tmpdir)
+	filename := filepath.Join(tmpdir, "docker.lic")
+	err = ioutil.WriteFile(filename, []byte(expiredLicense), 0644)
+	assert.NilError(t, err)
```
vdemeester:
```go
dir := fs.NewDir(t, "license", fs.WithFile("docker.lic", expiredLicense, fs.WithMode(0644)))
defer dir.Remove()
```

thaJeztah:
I'll have a look at this one, but can you squash your "review comments" commit into the commit that made the actual changes? No need to keep the history of the review steps in git 😅

vendor/github.com/docker/licensing/client.go
```diff
@@ -185,6 +187,10 @@ func (c *client) DownloadLicenseFromHub(ctx context.Context, authToken, subscrip
 
 func (c *client) ParseLicense(license []byte) (*model.IssuedLicense, error) {
 	parsedLicense := &model.IssuedLicense{}
+	// The file may contain a leading BOM, which will choke the
```
thaJeztah: Do we know where this BOM comes from? Have we found this in licenses we sent, or is this because a user manually edited the file?

dhiltgen: The old license server stuffed this header on there, and customers with old licenses lying around will have them.  The new license server doesn't do this so that's why we didn't notice the glitch until we started testing with some old offline licenses.

thaJeztah: Ah, that's unfortunate. I was hoping it was something we could fix on the server before release, but looks like that's not the case.

vendor/github.com/docker/licensing/storage.go
```diff
@@ -87,18 +87,23 @@ func (c *client) LoadLocalLicense(ctx context.Context, clnt WrappedDockerClient)
 		licenseData, err = readLicenseFromHost(ctx, info.DockerRootDir)
 	} else {
 		// Load the latest license index
-		latestVersion, err := getLatestNamedConfig(clnt, licenseNamePrefix)
+		var latestVersion int
+		latestVersion, err = getLatestNamedConfig(clnt, licenseNamePrefix)
 		if err != nil {
 			if strings.Contains(err.Error(), "not a swarm manager.") {
```
thaJeztah: Please don't do string matching here; check if the node is a manger before calling this,  e.g., as is done here; https://github.com/docker/cli/blob/ae1618713f83e7da07317d579d0675f578de22fa/cli/command/stack/swarm/deploy.go#L49-L62

Not a blocker for this PR, but we should address this

dhiltgen: @mason-fish do you have any updates in the queue where you could absorb this or should I get a PR posted?

andrewhsu:
@thaJeztah @vdemeester i created a PR to address comments https://github.com/dhiltgen/cli-1/pull/1

i can rebase my PR to docker/cli to get the changes in after this one is merged

vendor/github.com/docker/licensing/storage.go
mason-fish: Sure I can make the change once I'm through refactoring the error lib

PR #1388, opened 2018-09-22
[WIP] Refine how docker root dir is handled

dhiltgen:
This is a follow up PR to #1381 to address some of the review comments
we didn't get to.

Marking this WIP for now since I can't test it e2e until some other packaging changes land.


dhiltgen:
Note: this depends on the packaging change PR(s)

cli/command/engine/activate.go
```diff
@@ -97,10 +98,23 @@ func runActivate(cli command.Cli, options activateOptions) error {
...
 	}
 
+	// Short circuit if the user didn't specify a version and we're already running enterprise
+	if options.version == "" {
+		serverVersion, err := dclient.ServerVersion(ctx)
+		if err != nil {
+			return err
+		}
+		if strings.Contains(strings.ToLower(serverVersion.Platform.Name), "enterprise") {
+			return nil
```
thaJeztah: We should either treat this as an error-situation, or inform the user that no changes were made. In the "happy path", a message will be printed "Successfully activated engine". If we're already on EE, I'd expect some form of feedback as well.

internal/containerizedengine/update.go
```diff
@@ -24,15 +22,34 @@ import (
 
 // ActivateEngine will switch the image from the CE to EE image
 func (c *baseClient) ActivateEngine(ctx context.Context, opts clitypes.EngineInitOptions, out clitypes.OutStream,
-	authConfig *types.AuthConfig, healthfn func(context.Context) error) error {
+	authConfig *types.AuthConfig, healthfn func(context.Context) error, runtimeMetadataDir string) error {
```
thaJeztah: Instead of adding another argument, can we add `runtimeMetadataDir` as a field on `clitypes.EngineInitOptions` ?

cli/command/engine/client_test.go
```diff
@@ -69,9 +69,9 @@ func (w *fakeContainerizedEngineClient) DoUpdate(ctx context.Context,
 	opts clitypes.EngineInitOptions,
 	out clitypes.OutStream,
 	authConfig *types.AuthConfig,
-	healthfn func(context.Context) error) error {
+	healthfn func(context.Context) error, runtimeMetadataDir string) error {
```
thaJeztah: Same here; Instead of adding another argument, can we add `runtimeMetadataDir` as a field on `clitypes.EngineInitOptions` ?

cli/command/engine/activate.go
```diff
@@ -112,7 +126,7 @@ func runActivate(cli command.Cli, options activateOptions) error {
 			client := cli.Client()
 			_, err := client.Ping(ctx)
 			return err
-		}); err != nil {
+		}, clitypes.RuntimeMetadataDir); err != nil {
```
thaJeztah: Related to my other comments about adding this option to the `clitypes.EngineInitOptions`; it looks like currently we only pass the default value.

We could get that default in `client.ActivateEngine()` /  `client.DoUpdate()`, i.e., something like

```go
func (c *baseClient) DoUpdate(ctx context.Context, opts clitypes.EngineInitOptions, out clitypes.OutStream,
	authConfig *types.AuthConfig, healthfn func(context.Context) error) error {
	if opts.MetaDataDir == "" {
		opts.MetaDataDir = clitypes.RuntimeMetadataDir
	}
	...
}
```

Doing so, makes the default (empty string) a useful value

cli/command/engine/client_test.go
thaJeztah: Wait: what's that `healthfn` argument actually for? Looks like it's only used for testing? That really shouldn't be in there; I'm sure that can be set on the mock during testing?

thaJeztah:
ping @silvin-lubecki @vdemeester PTAL

cli/command/engine/client_test.go
dhiltgen: Hmm.. hold over from prior implementations... we really need to get this in ASAP so let me excise that part in a follow up if you don't mind.

dhiltgen:
@thaJeztah if you're still online, could you take a quick peek?  We'd like to get this in today if we can so we can cut a beta1.8 build to wrap up a number of remaining P0's.

crosbymichael:
LGTM

cli/command/engine/activate.go
crosbymichael: sound good to make this change

PR #1389, opened 2018-09-24
[18.09] fix subscription filter

mason-fish:
Fixes the subscription filter to only include "active" subscriptions. As is, we are letting cancelled subscriptions through which misleads the user.

### DIFF
--- a/internal/licenseutils/utils.go
+++ b/internal/licenseutils/utils.go
@@ -97,7 +97,7 @@ func (u HubUser) GetAvailableLicenses(ctx context.Context) ([]LicenseDisplay, er
 	// Filter out expired licenses
 	i := 0
 	for _, s := range subs {
-		if s.State != "expired" && s.Expires != nil {
+		if s.State == "active" && s.Expires != nil {
 			owner := ""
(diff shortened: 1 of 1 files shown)


internal/licenseutils/utils.go
```diff
@@ -97,7 +97,7 @@ func (u HubUser) GetAvailableLicenses(ctx context.Context) ([]LicenseDisplay, er
 	// Filter out expired licenses
 	i := 0
 	for _, s := range subs {
-		if s.State != "expired" && s.Expires != nil {
+		if s.State == "active" && s.Expires != nil {
```
silvin-lubecki: What are the different states? I think these constants should be extracted and exposed and documented as a type in the licensing library, instead of plain string like this.

mason-fish: Good call, I opened an issue in docker/licensing to address this. Can swing back to this after that change and vendor bump happen.

thaJeztah:
I notice that the PR description is not in the commit message; @mason-fish for follow up PR's, could you make sure that the commit message also has the information 🙏 🤗

Issue #1391, opened 2018-09-25
[Regression] Environment variables are ignored during stack deployment


anokun7:
**Description**
Environment variables used inside a docker-compose.yml are not getting evaluated. Environment variables allows for dynamic properties to be added to docker-compose.yml files to use them in different environments for example.

**Steps to reproduce the issue:**
1. Example compose file
```
ᐅ cat test.yml -p
version: "3.3"

services:
  app:
    image: anoop/counter-demo:${version:-latest}
```
2. Run `docker stack deploy` using an environment variable
```
ᐅ version=onbuild docker stack deploy -c test.yml c1
Creating network c1_default
Creating service c1_app
```
3. Check the image used.
```
ᐅ docker stack ps c1
ID                  NAME                IMAGE                       NODE                DESIRED STATE       CURRENT STATE           ERROR               PORTS
yj1d99b0z2ty        c1_app.1            anoop/counter-demo:latest   ip-10-1-1-122.ec2   Running             Running 6 seconds ago
```
4. The image is `anoop/counter-demo:latest`

**Describe the results you received:**
The image is `anoop/counter-demo:latest`

**Describe the results you expected:**
The image should have been `anoop/counter-demo:onbuild` as that was the value of the env variable used at the time the stack was deployed

**Additional information you deem important (e.g. issue happens only occasionally):**
Seems to happen only on the latest docker client. I am using `18.09.0-ce-beta1`.
The same issue does not occur on `17.06.2-ee-16`

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Version:           18.09.0-ce-beta1
 API version:       1.30 (downgraded from 1.39)
 Go version:        go1.10.4
 Git commit:        78a6bdb
 Built:             Thu Sep  6 22:41:53 2018
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Enterprise Edition (EE) 2.0
 Engine:
  Version:          17.06.2-ee-16
  API version:      1.30 (minimum version 1.12)
  Go version:       go1.8.7
  Git commit:       9ef4f0a
  Built:            Thu Jul 26 16:42:19 2018
  OS/Arch:          linux/amd64
  Experimental:     false
 Universal Control Plane:
  Version:          3.0.5
  ApiVersion:       1.30
  Arch:             amd64
  BuildTime:        Thu Aug 30 17:47:03 UTC 2018
  GitCommit:        f588f8a
  GoVersion:        go1.9.4
  MinApiVersion:    1.20
[...]

PR #1392, opened 2018-09-25
Fix substitution with non-empty env-var

thaJeztah:
fixes https://github.com/docker/cli/issues/1391

Due to a typo introduced in https://github.com/docker/cli/pull/1249, substitution would not work if the given environment-variable was set.

Given the following docker compose file;

```yaml
version: "3.7"

services:
  app:
    image: nginx:${version:-latest}
```

Deploying a stack with `$version` set would ignore the `$version`
environment variable, and use the default value instead;

```bash
version=alpine docker stack deploy -c docker-compose.yml foobar

Creating network foobar_default
Creating service foobar_app

docker service ls

rskkjxe6sm0w        foobar_app          replicated          1/1                 nginx:latest
```

### DIFF
--- a/cli/compose/template/template.go
+++ b/cli/compose/template/template.go
@@ -176,15 +176,21 @@ func extractVariable(value interface{}, pattern *regexp.Regexp) ([]extractedValu
 
 // Soft default (fall back if unset or empty)
 func softDefault(substitution string, mapping Mapping) (string, bool, error) {
-	return withDefault(substitution, mapping, "-:")
+	sep := ":-"
+	if !strings.Contains(substitution, sep) {
+		return "", false, nil
+	}
+	name, defaultValue := partition(substitution, sep)
+	value, ok := mapping(name)
+	if !ok || value == "" {
+		return defaultValue, true, nil
+	}
+	return value, true, nil
 }
 
 // Hard default (fall back if-and-only-if empty)
 func hardDefault(substitution string, mapping Mapping) (string, bool, error) {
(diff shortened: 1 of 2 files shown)


thaJeztah:
This should be cherry-picked in the 18.09 branch as well


ping @anokun7 @vdemeester @silvin-lubecki PTAL

thaJeztah:
Oh, interesting; I tried my new test, but the existing ones broke 🤔 let me check why

thaJeztah:
Looks like the "soft default" doesn't actually check for empty values, so that's an existing bug as well (which went unnoticed because we never reached that part).

I'll have a look after dinner 😅

thaJeztah:
Fixed; should be good to go now

cli/compose/template/template.go
```diff
@@ -176,15 +176,21 @@ func extractVariable(value interface{}, pattern *regexp.Regexp) ([]extractedValu
 
 // Soft default (fall back if unset or empty)
 func softDefault(substitution string, mapping Mapping) (string, bool, error) {
-	return withDefault(substitution, mapping, "-:")
+	sep := ":-"
+	if !strings.Contains(substitution, sep) {
```
silvin-lubecki: I guess it can be factorized with `hardDefault` but I don't see an obvious and/or clean way to do it.

thaJeztah: Yes, I was looking at that, but adding an (e.g.) `checkIfSet func(val) bool` argument felt like over engineering just to prevent just a few lines of duplicated code.

PR #1401, opened 2018-09-29
fixes clean utf8bom in combined env file

lifubang:

**- What I did**
When combined two utf8(with bom) env files, one env variable will be lost.

**- How I did it**
Clean utf8bom for each line

**- How to verify it**
```
root@dockerdemo:~/deploytest# cat test2.env
﻿def=456
foo
bar
root@dockerdemo:~/deploytest#
```
```
root@dockerdemo:~/deploytest# cat test1.env 
﻿abc=123

root@dockerdemo:~/deploytest# 
```
```
root@dockerdemo:~/deploytest# cat test2.env > test3.env
root@dockerdemo:~/deploytest# cat test1.env >> test3.env
root@dockerdemo:~/deploytest# cat test3.env
﻿def=456
foo
bar
﻿abc=123

root@dockerdemo:~/deploytest#
```
```
root@dockerdemo:~/deploytest# cat docker-compose.yml 
version: "3.7"
services:
  app:
    environment:
      foo: "bar"
    env_file:
      - ./test3.env
    image: docker.acmcoder.com/public/ubuntu:ssh
    entrypoint: /bin/bash
    tty: true
root@dockerdemo:~/deploytest#
```
```
root@dockerdemo:~/deploytest# ~/gocode/src/github.com/docker/cli/build/docker stack deploy -c docker-compose.yml foobar
Creating network foobar_default
Creating service foobar_app
root@dockerdemo:~/deploytest# docker exec -it foobar_app.1.lv4pvvw5siuypvin53dkgbzl5 /bin/bash
root@ccd9a92c49f1:/# echo $def
456
root@ccd9a92c49f1:/# echo $foo
bar
root@ccd9a92c49f1:/# echo $abc

root@ccd9a92c49f1:/#
```
abc is lost.

Each line may contain utf8bom, so we need to clean them not just for the first line.

eo394r 4eip3y ecjns2cc

### DIFF
--- a/opts/file.go
+++ b/opts/file.go
@@ -38,9 +38,9 @@ func parseKeyValueFile(filename string, emptyFn func(string) (string, bool)) ([]
 			return []string{}, fmt.Errorf("env file %s contains invalid utf8 bytes at line %d: %v", filename, currentLine+1, scannedBytes)
 		}
 		// We trim UTF8 BOM
-		if currentLine == 0 {
-			scannedBytes = bytes.TrimPrefix(scannedBytes, utf8bom)
-		}
+		// For combined file, each line may contain utf8bom,
+		// not just the first line
+		scannedBytes = bytes.TrimPrefix(scannedBytes, utf8bom)
 		// trim the line from all leading whitespace first
 		line := strings.TrimLeftFunc(string(scannedBytes), unicode.IsSpace)
 		currentLine++


thaJeztah:
Thanks for contributing, @lifubang 

Looking at this change, this doesn't look correct to me. A BOM header is only a BOM header if it appears at the start of a file; Docker cannot make assumptions about the same characters appearing on other locations in the file.

The _actual_ issue here is that you're appending the contents of a file to an existing file, without stripping the BOM header first. This results in an invalid file, and causes the failure.

The correct approach here would be to;

- Don't use a BOM header (really, try avoid those, they cause issues in many situations)
- If you _have_ to use a BOM header; strip the header before concatenating the files; e.g. see https://unix.stackexchange.com/questions/381230/how-can-i-remove-the-bom-from-a-utf-8-file
 

So, based on the above, I'm -1 on this change, but perhaps @silvin-lubecki @vdemeester think otherwise 😅

lifubang:
I quite agree with you. If anyone have no more ideas, it can be closed. I feel so sorry to waste your precious time.

thaJeztah:
> I feel so sorry to waste your precious time.

No need to apologise (at all!!) 🤗 I understand the problem you were facing, and this _does_ fix that issue; it's just that accepting this change could become a "sliding slope", and could (potentially) cause issues for other situations.

Let me close this issue, but keep up the good work; I really appreciate people that see a problem, and try to fix it 👍 🙌 . Don't hesitate to open other pull requests! 🤗

PR #1406, opened 2018-10-01
Remove unused helath check func

dhiltgen:
Follow up from #1388 

During the refactoring for 18.09 the activate/update flows no longer
restart the engine explicitly but let the user do that when they're ready,
so the health check logic is no longer required.


thaJeztah:
Thanks!

PR #1407, opened 2018-10-02
Allow to transform any source type into the target struct

mindaugasrukas:
This is a small change to allow passing any type to `loader.Tranform` func instead of just a map. This change allows to reuse the docker library and parse docker compose file parts - for example ports.


mindaugasrukas:
> Thanks! Changes LGTM, but could you squash the "fix linter" commit with the first commit? That keeps the git history more clean.

done

PR #1408, opened 2018-10-02
[18.09] Fix panic in display only case for license

dhiltgen:
Prior refactoring passes missed a corner case.

Signed-off-by: Daniel Hiltgen <daniel.hiltgen@docker.com>

PR #1415, opened 2018-10-03
Fix panic in display only case for license - forward port #1408 to master

dhiltgen:
Forward port #1408 to master


cli/command/engine/activate.go
```diff
@@ -90,6 +92,9 @@ func runActivate(cli command.Cli, options activateOptions) error {
 		if license, err = getLicenses(ctx, authConfig, cli, options); err != nil {
 			return err
 		}
+		if options.displayOnly {
```
silvin-lubecki: I guess the fix is here, but isn't it safer and simpler to just check if license is nil?
```go
if license, err = getLicenses(ctx, authConfig, cli, options); err != nil || license == nil {
    return err
}
```

thaJeztah: @dhiltgen ^

thaJeztah: @silvin-lubecki discussing with @dhiltgen out of band; let's merge this one as-is, and do a follow up with improvements

PR #1416, opened 2018-10-04
fixes when no equal sign in label file

lifubang:

**- What I did**
When use --label:
`./build/docker run -d --name=redis5 --label foo redis`
```
root@dockerdemo:~/gocode/src/github.com/docker/cli# ./build/docker inspect --format '{{.Config.Labels}}' redis5
map[foo:]
```
When use --label-file:
```
root@dockerdemo:~/gocode/src/github.com/docker/cli# echo $foo
bar
root@dockerdemo:~/gocode/src/github.com/docker/cli# cat ~/labels.txt 
foo
bar=
root@dockerdemo:~/gocode/src/github.com/docker/cli# build/docker run -d --name=redis7 --label-file ~/labels.txt redis
root@dockerdemo:~/gocode/src/github.com/docker/cli# ./build/docker inspect --format '{{.Config.Labels}}' redis7
map[bar:]
```

label foo is lost.

The behavior of --label and --label-file is different when there is no equal sign after label variable.
**- How I did it**
When there is no equal sign after label variable, and no emptyFn, add it to output.

**- How to verify it**
After fix:
```
root@dockerdemo:~/gocode/src/github.com/docker/cli# build/docker run -d --name=redis8 --label-file ~/labels.txt redis
root@dockerdemo:~/gocode/src/github.com/docker/cli# ./build/docker inspect --format '{{.Config.Labels}}' redis8
map[bar: foo:]
```

fixes when no equal sign in label file

508c439d40d41d0a9c2e3256ee1ca5a

### DIFF
--- a/cli/command/container/opts_test.go
+++ b/cli/command/container/opts_test.go
@@ -965,6 +965,25 @@ func TestParseEnvfileVariables(t *testing.T) {
 	}
 }
 
+func TestParseEnvfileVariablesWithoutEqualSign(t *testing.T) {
+	// env without equal sign
+	config, _, _, err := parseRun([]string{"--env-file=testdata/noequalsign.env", "img", "cmd"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(config.Env) != 2 || config.Env[0] != "bar=" || config.Env[1] != "demo=true" {
+		t.Fatalf("Expected a config with [bar= demo=true], got %v", config.Env)
+	}
+	t.Setenv("foo", "bar")
+	config, _, _, err = parseRun([]string{"--env-file=testdata/noequalsign.env", "img", "cmd"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(config.Env) != 3 || config.Env[0] != "foo=bar" || config.Env[1] != "bar=" || config.Env[2] != "demo=true" {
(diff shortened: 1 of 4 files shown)


vdemeester (dismissed):
LGTM :frog:

thaJeztah (changes requested):
Haven't checked yet, but is this same function also used for env-files? If so, should it behave the same?

We should at least have a test for this, as the behavior of empty values vs not-set is already tricky, so easy to introduce regressions / change in behavior

vdemeester:
> Haven't checked yet, but is this same function also used for env-files? If so, should it behave the same?

Yeah, but for `env-files` the `emptyfn` is not nil so it should still act the same :wink:  But I agree on a test

lifubang:
Thank you for all your review, I'll add some test cases.

lifubang (commented):
I have add two test cases for label/env file without equal sign.

lifubang:
@thaJeztah I don't know if I have understood your points. Do you mean add some test cases, or change code to check the lines value?

tonistiigi:
ping @thaJeztah

thaJeztah:
Cleaning up old PRs; I thought this was fixed, but indeed looks like the "no value set" omits the label; still need to look if that was or was not by design, but did a quick rebase, and updated the patch to make the new tests added in this PR pass.

PR #1419, opened 2018-10-05
build: add SSH agent socket forwarder (`docker build --ssh $SSHMOUNTID=$SSH_AUTH_SOCK`)

AkihiroSuda:

**- What I did**

This commit adds SSH agent socket forwarder (`docker build --ssh $SSHMOUNTID=$SSH_AUTH_SOCK`) .

Unlike `docker build --secret`, `docker build --ssh` allows the build container to use SSH keys with passphrases.

```console
$ eval $(ssh-agent)
$ ssh-add ~/.ssh/id_rsa
(Input your passphrase here)
$ docker build --ssh default=$SSH_AUTH_SOCK ...
```

This feature requires the daemon with `CapExecMountSSH` build capability (moby/moby#37973) .

Currently, the official Dockerfile frontend does not provide the syntax for using the SSH forwarder.

However, the experimental `RUN --mount=type=ssh` syntax can be enabled by using
the Dockerfile frontend image built with the `BUILDTAGS="dfrunmount dfssh"`, via the `# syntax =` "shebang".

The Dockerfile for the Dockerfile frontend is available at  https://github.com/moby/buildkit/tree/master/frontend/dockerfile/cmd/dockerfile-frontend .
A pre-built image is also available as `tonistiigi/dockerfile:ssh20181002` .

An example Dockerfile with `RUN --mount=type=ssh`:

```dockerfile
# syntax = tonistiigi/dockerfile:ssh20181002
FROM alpine
RUN apk add --no-cache openssh-client
RUN mkdir -p -m 0700 ~/.ssh && ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
RUN --mount=type=ssh ssh git@gitlab.com | tee /hello
# "Welcome to GitLab, @GITLAB_USERNAME_ASSOCIATED_WITH_SSHKEY" should be printed here
```

More info available at moby/buildkit#608, moby/buildkit#655

**- How I did it**

See above

**- How to verify it**

See above

build: add SSH agent socket forwarder (`docker build --ssh $SSHMOUNTID=$SSH_AUTH_SOCK`)

penguin


AkihiroSuda:
(probably this is too late for v18.09)

AkihiroSuda:
@tonistiigi @tiborvass @ijc

tonistiigi:
@AkihiroSuda @tiborvass @vdemeester Are we sure about this syntax. I don't have better ideas but didn't give it that much thought when I added this in `buildctl`. The `default` constant seems weird but don't really have a way around that.

Btw. `docker build --ssh default=$SSH_AUTH_SOCK` is the same as `docker build --ssh default`.

tiborvass:
For the record, the easiest way is to do `docker build --ssh default` without even specifying the auth socket. Tested it with `tonistiigi/dockerfile:ssh20181005` which has support for `required` as so: `RUN --mount=type=ssh,required ...`

LGTM

tiborvass:
@tonistiigi I think the word `default` is fine. For those wondering why we need a word at all: the rationale is that we should enable security-minded people who want to specify different keys for different projects/endpoints. Those can be differentiated via IDs for `RUN --mount`.

```
RUN --mount=type=ssh,id=myrepo1 ssh ...`
RUN --mount=type=ssh,id=myrepo2 ssh ...`
```

A CI could do `docker build --ssh myrepo1=/path/to/myrepo1-readonly.pub --ssh myrepo2=/path/to/myrepo2-readonly.pub ...` while a user using an SSH agent would simply have to do: `docker build --ssh myrepo1 --ssh myrepo2 ...`

In order to allow that granularity, we need a default ID... and I'm fine with `default`.

cpuguy83:
I would call this something like "forward-ssh" or "ssh-agent".
Other than that, I wonder if there's some magic we can do with the flag parser to have the flag value be optional and rather treat it like a book flag that can potentially have a string value.

thaJeztah:
Had the same thought as Brian, but only was able to have a very brief look

tonistiigi:
Note that this syntax allows direct keys as well, and that should be the preferred way to use in CI (with special deploy keys). So this is not only for forwarding the pre-created ssh agent socket.

tiborvass:
@cpuguy83 @thaJeztah so are you okay with this?

tiborvass:
Thanks @cpuguy83 ! Just to clarify, the magic referred to is to be able to parse `--ssh` alone without argument to mean the same thing as `--ssh default`. It would be a bit too magical and could be prone to error cases. Agree that for now we can keep it simple.

AkihiroSuda:
Is this ready for 18.09?

seemethere:
FC date for 18.09 was back in September so this most likely will not make the GA for 18.09.0

thaJeztah:
Added `docs/revisit`; we could use some documentation about this feature

thaJeztah:
@rmoriz could you open an issue for that in the buildkit repository? https://github.com/moby/buildkit

PR #1420, opened 2018-10-05
Clarify the sparse "docker port" output + doc, what maps to what

thomas-klambauer:
**- What I did**

**- How I did it**

**- How to verify it**


thomas-klambauer:
See https://docs.docker.com/engine/reference/commandline/port/#examples 
Most people rate this doc page badly.(192 negative, 70 positive), assuming the same reason why I was annoyed, no explanation in what direction the mapping happens (contains first, host first?!)

PR #1421, opened 2018-10-08
[18.09] Update release note link to final location

dhiltgen:
We'll be using a redirect from this URL to the back-end docs system for
hosting release notes. Final location confirmed with Docs team and PM.

### DIFF
--- a/cli/command/engine/check.go
+++ b/cli/command/engine/check.go
@@ -117,7 +117,7 @@ func processVersions(currentVersion, verType string,
 			availUpdates = append(availUpdates, clitypes.Update{
 				Type:    verType,
 				Version: ver.Tag,
-				Notes:   fmt.Sprintf("%s/%s", clitypes.ReleaseNotePrefix, ver.Tag),
+				Notes:   fmt.Sprintf("%s?%s", clitypes.ReleaseNotePrefix, ver.Tag),
 			})
 		}
 	}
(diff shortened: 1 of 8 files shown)


cli/command/engine/check.go
```diff
@@ -117,7 +117,7 @@ func processVersions(currentVersion, verType string,
 			availUpdates = append(availUpdates, clitypes.Update{
 				Type:    verType,
 				Version: ver.Tag,
-				Notes:   fmt.Sprintf("%s/%s", clitypes.ReleaseNotePrefix, ver.Tag),
+				Notes:   fmt.Sprintf("%s?%s", clitypes.ReleaseNotePrefix, ver.Tag),
```
vdemeester: Shouldn't it be an `anchor` ? (`#`)

types/types.go
```diff
@@ -19,7 +19,7 @@ const (
 	RegistryPrefix = "docker.io/store/docker"
 
 	// ReleaseNotePrefix is where to point users to for release notes
-	ReleaseNotePrefix = "https://docs.docker.com/releasenotes"
+	ReleaseNotePrefix = "https://docker.com/engine/releasenotes"
```
vdemeester: this is 404 for now, normal ?

andrewhsu:
cc @JustinINevill

andrewhsu:
cc @mataneja

mataneja:
I see that `releasenotes` is redirected to `release-notes` and that's fine. We can continue to use the new URL as is.

mataneja:
LGTM

types/types.go
thaJeztah: Works now; redirects to the right page on docs.docker.com

PR #1424, opened 2018-10-09
Deprecate "devicemapper" storage driver.

thaJeztah:
The `devicemapper` storage driver is deprecated in favor of `overlay2`, and will
be removed in a future release. Users of the `devicemapper` storage driver are
recommended to migrate to a different storage driver, such as `overlay2`, which
is now the default storage driver.

The `devicemapper` storage driver facilitates running Docker on older (3.x) kernels
that have no support for other storage drivers (such as overlay2, or AUFS).

Now that support for `overlay2` is added to all supported distros (as they are
either on kernel 4.x, or have support for multiple lowerdirs backported), there
is no reason to continue maintenance of the `devicemapper` storage driver.

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -34,6 +34,21 @@ on pre 4.x kernels. Now that all supported distributions are able to run `overla
 (as they are either on kernel 4.x, or have support for multiple lowerdirs
 backported), there is no reason to keep maintaining the `overlay` storage driver.
 
+### device mapper storage driver
+
+**Deprecated in Release: v18.09.0**
+
+The `devicemapper` storage driver is deprecated in favor of `overlay2`, and will
+be removed in a future release. Users of the `devicemapper` storage driver are
+recommended to migrate to a different storage driver, such as `overlay2`, which
+is now the default storage driver.
+
+The `devicemapper` storage driver facilitates running Docker on older (3.x) kernels
+that have no support for other storage drivers (such as overlay2, or AUFS).
+
+Now that support for `overlay2` is added to all supported distros (as they are
+either on kernel 4.x, or have support for multiple lowerdirs backported), there
+is no reason to continue maintenance of the `devicemapper` storage driver.
 
 ### Reserved namespaces in engine labels
 


AkihiroSuda:
Can we have `WARNING: devicemapper is deprecated` in `docker info`?

AkihiroSuda:
Maybe we should have the warning message in daemon log?

thaJeztah:
Yes, having warnings both in logs and in `docker info` output is a good idea; I'll open PR's for those

thaJeztah:
ping @silvin-lubecki PTAL

vdemeester:
@thaJeztah needs a rebase :pray:

joshsleeper:
Sorry to necropost a bit, bit I haven't found any good documentation explaining how this doesn't negatively affect CentOS. At least according to the [current docs](https://docs.docker.com/storage/storagedriver/select-storage-driver/#docker-ce), `devicemapper` and `vfs` are the only supported storage drivers for CentOS, unless that's out of date and `overlay2` actually is supported?

image

thaJeztah:
@joshsleeper hm, looks like that table is quite out of date. Also not sure why `vfs` is listed under _recommended_ storage drivers; it definitively isn't 

overlay2 is supported on current versions of CentOS, Ubuntu, Debian, Fedora, and RHEL (but only Docker Enterprise ships for RHEL)

I'll open a pull request to amend that page

egernst:
@thaJeztah - Obviously late to the party here, but as a developer on a VM-isolated container runtime (kata-containers), there is *a lot* of utility to having a block based storage solution available.  AFAIU devicemapper is the only storage driver which provides this.  Are there any functional / support issues with device-mapper today that's leading to the deprecation?

thaJeztah:
@egernst I commented with some details on the roadmap in https://github.com/docker/for-linux/issues/452#issuecomment-440427796, and also was discussing options with containerd maintainers. They informed me that they are open for contributions to implement and add a containerd  snapshotter that provides block based storage. Wether or not this would be _devicemapper_ would have to be looked at (there may be more performant choices). I think they actually recently had a conversation with contributors that were interested to work on this, but I don't have the details.

Perhaps it would be good to open a tracking issue in the containerd issue tracker (if none exists yet)

egernst:
Thanks for the quick reply @thaJeztah  - that thread covers it nicely.  The need is for block-based, whether it be devicemapper or not isn't particularly important to us.  I'll open an issue in containerd if none exists yet and continue the conversation there. Thanks!

gerroon:
Can you tell us how to migrate at least? Failed Docker after the upgrade is no use at all.

cpuguy83:
Argh... I can't believe this didn't get fixed in 18.09.1...

sarahn:
Can there be an approximate date for when devicemapper will be removed entirely?

PR #1435, opened 2018-10-10
Add bash completion for `service --no-resolve-image`

trapier:
**- What I did**
add bash completion for `service {create,update} --no-resolve-image`

**- How I did it**
update `contrib/completion/bash/docker:_docker_service_update_and_create()`

**- How to verify it**
```
bash
source contrib/completion/bash/docker
docker service update <TAB><TAB>
docker service create <TAB><TAB>
```

NA

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -3440,6 +3440,7 @@ _docker_service_update_and_create() {
 		--help
 		--init
 		--no-healthcheck
+		--no-resolve-image
 		--read-only
 		--tty -t
 		--with-registry-auth


albers:
Sorry, I messed up my review comment, so here it is again:

In your implementation, the option requires an argument. Thus, the next option will be treated as an argument to `--no-resolve-image` and not be completed.
The new option has to be added to the `boolean_options` variable below.

trapier:
:man_facepalming:

good catch @albers!

So moved.

albers:
Thanks @trapier.

PR #1451, opened 2018-10-17
builder/prune: rename max-age filter to unused-for in help output

tiborvass:


thaJeztah:
Looks like there's no documentation yet; https://github.com/docker/cli/tree/master/docs/reference/commandline, so probably we need something as well (can be a follow up)

PR #1458, opened 2018-10-18
[18.09] Bump licensing lib

dhiltgen:
Removes the billing profile flow which is now handled on the back-end.


silvin-lubecki:
LGTM

dhiltgen:
@andrewhsu I tried to forward port this to master, but it looks like we're a bit behind and it had a merge conflict.  Not sure how you want to track that so it doesn't fall through the cracks.

Issue #1437, opened 2018-10-10
Require confirmation when removing (deleting) volumes


rcdailey:
Right now, it's a bit too easy to remove volumes in Docker. I can remove a volume via the command below and it requires no further interaction from the user:

```
docker volume rm myvolume
```

I propose that the default behavior be altered in such a way that requires the user to explicitly indicate that they want removal to be done silently without prompt:

```
docker volume rm -f myvolume
```

And the original behavior is changed to prompt the user:

```
$ docker volume rm myvolume
This will permanently delete volume 'myvolume'. Are you sure? [y/n] y
Volume deleted
```

AkihiroSuda:
This breaks compatibility

rcdailey:
You mean backward compatibility? I'm aware of that. To be honest, I don't really care. The benefits of this change far outweigh the inconvenience of having people to update their scripts (on a related note, I would question why people are automating volume deletions...).

PR #1465, opened 2018-10-21
Added confirmation when removing volumes #1437

grallc:
**- What I did**
Confirmation is now required for deleting a volume with `docker volume rm`. (#1437)
It will ask a confirmation for each volumes.

**- How I did it**
Simply added a condition in `/cli/command/volume/remove.go`.

**- How to verify it**
Create some volumes with `docker volume create [name]` (eg. `docker volume create dog-volume`, `docker volume create cat-volume`, `docker volume create rabbit-volume`)  then delete them with `docker volume rm` (`docker volume rm cat-volume dog-volume`).

image

### DIFF
--- a/cli/command/volume/remove.go
+++ b/cli/command/volume/remove.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/completion"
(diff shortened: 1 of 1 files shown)


thaJeztah:
We can't change the default behavior, because it will likely break many people. I do see the value of having some kind of "interactive" version of `docker volume rm`, but if we want that, we should think of making this configurable ( in the cli configuration file `~/.docker/config.json`).

So in its current form, I don't think we can make this change, but we can discuss options here.

PR #1468, opened 2018-10-22
fixes #1452 set path separator to slash in interpolation and add a test case

### LINKED ISSUE #1452: Variable type conversion in Compose file breaks if service name contains dot
**Description**

When using environment variables for integer or boolean types in a service definition, conversion will fail if the service name contains a dot (`.`) character.

1. Create the following `docker-compose.yml` file:
```yaml
version: "3"
services:
  app.0:
    image: redis
    ulimits:
      nofile:
        soft: ${APP_NOFILE}
        hard: ${APP_NOFILE}
```
2. Run `docker stack deploy -c docker-compose.yml stk`

```
$ docker stack deploy -c docker-compose.yml stk
services.app.0.ulimits.nofile.hard must be a integer
```

Stack is deployed

Originally reported as a Compose issue: docker/compose#6279 ; realized while fixing that it applies to docker stack as well

```
$ docker version
Client:
 Version:           18.06.1-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        e68fc7a
 Built:             Tue Aug 21 17:23:18 2018
 Experimental:      false

Server:
 Engine:
  Version:          18.06.1-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       e68fc7a
  Built:            Tue Aug 21 17:22:21 2018
  OS/Arch:          linux/amd64
  Experimental:     false
```

**Output of `docker info`:**

```
(irrelevant)
```

lifubang:

**- What I did**
Please see issue #1452 , Variable type conversion in Compose file breaks if service name contains dot.

**- How I did it**
set path separator from dot to slash in interpolation and add a test case

**- How to verify it**
After this patch:
```
root@dockerdemo:~/testd# cat file2.yaml 
version: "3.0"
services:
  app.test_1:
    image: redis
    ulimits:
      nofile:
        soft: ${APP_NOFILE}
        hard: ${APP_NOFILE}
root@dockerdemo:~/testd#
```
```
root@dockerdemo:~/testd# APP_NOFILE=1 ~/gocode/src/github.com/docker/cli/build/docker stack deploy -c file2.yaml stk2
Creating service stk2_app.test_1
```

set path separator from dot to slash in interpolation

**- Addition info**
But, but, the daemon said:
failed to create service stk2_app.test_1: Error response from daemon: rpc error: code = InvalidArgument desc = name must be valid as a DNS name component
The daemon's check regexp is `^[a-zA-Z0-9](?:[-_]*[A-Za-z0-9]+)*$`, don't contain dot.
But I think this is another issue( see #1470 ).
At least, by this patch, we can get the right response.

### DIFF
--- a/cli/compose/interpolation/interpolation.go
+++ b/cli/compose/interpolation/interpolation.go
@@ -108,7 +108,7 @@ func newPathError(path Path, err error) error {
 	}
 }
 
-const pathSeparator = "."
+const pathSeparator = "/"
 
 // PathMatchAll is a token used as part of a Path to match any key at that level
 // in the nested structure
--- a/cli/compose/interpolation/interpolation_test.go
+++ b/cli/compose/interpolation/interpolation_test.go
@@ -58,7 +58,7 @@ func TestInvalidInterpolation(t *testing.T) {
 		},
 	}
 	_, err := Interpolate(services, Options{LookupValue: defaultMapping})
-	assert.Error(t, err, `invalid interpolation format for servicea.image: "${". You may need to escape any $ with another $.`)
+	assert.Error(t, err, `invalid interpolation format for servicea/image: "${". You may need to escape any $ with another $.`)
 }
 
 func TestInterpolateWithDefaults(t *testing.T) {
--- a/cli/compose/loader/loader_test.go
+++ b/cli/compose/loader/loader_test.go
@@ -566,6 +566,26 @@ volumes:
 	assert.Check(t, is.Equal(home, config.Volumes["test"].Driver))
 }
 
+func TestLoadWithDotInServiceName(t *testing.T) {
+	dict, err := ParseYAML([]byte(`
+version: "3.4"
+services:
+  app.0:
+    ulimits:
+      nofile:
+        soft: ${APP_NOFILE}
+        hard: ${APP_NOFILE}`))
+	assert.NilError(t, err)
+	env := map[string]string{
+		"APP_NOFILE": "555",
+	}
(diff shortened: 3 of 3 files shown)


thaJeztah:
Right, if I understandd correctly, this only fixes that when parsing the YAML, it doesn't see

`foo.bar.baz: test` as 

```yaml
foo
  bar
    baz: test
```

I don't think we should change this, as it would probably break people using the existing syntax

Servicenames are not allowed to contain a `.`, as they are also used in DNS, which may result in service-names conflicting.

The correct way to pass such a name would be to add quotes;

```
docker stack deploy -c- foo <<EOF
version: "3.0"
services:
  "app.test_1":
    image: redis
EOF
```

But of course, that would still return the daemon error (as expected);

```
Creating network foo_default
Creating service foo_app.test_1
failed to create service foo_app.test_1: Error response from daemon: rpc error: code = InvalidArgument desc = name must be valid as a DNS name component
```

lifubang:
> The correct way to pass such a name would be to add quotes;
> 

@thaJeztah if there were no ${APP_NOFILE}, everything will be OK. We don't need to add quotes.

But if there is ${APP_NOFILE}, docker cli needs to check the type of ulimits.nofile.soft, it will raise an incorrect error.
```
APP_NOFILE=1 build/docker stack deploy -c ~/t1.yaml foo
services.app.test_1.ulimits.nofile.soft must be a integer
(but APP_NOFILE is already a integer).
```
Even we add quotes, it raise the same incorrect error.

Issue #1476, opened 2018-10-23
docker run -it fail when use ssh host with password


lifubang:
**Description**
docker cli use ssh is great, especially when use ssh logon free.
There is a small issue, when use ssh with password, docker run -it will fail.

**Steps to reproduce the issue:**
1. docker -H ssh://localhost run --name=testssh -it docker.acmcoder.com/public/ubuntu:ssh

**Describe the results you received:**
root@localhost's password: 
root@localhost's password: 
root@localhost's password: 
                           root@localhost's password: 
                                                      root@localhost's password: 
                                                                                 build/docker: error during connect: Post http://docker/v1.39/containers/6a7570cba6aac7b812df5665f077490c614917933c39c450954079e4d5fe6649/start: command [ssh localhost -- docker system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=Permission denied, please try again.
Permission denied, please try again.
Permission denied (publickey,password).

ERRO[0015] error waiting for container: context canceled 
WARN[0015] commandConn.CloseWrite: connhelper: failed to wait: signal: killed

**Describe the results you expected:**
root@localhost's password: 
root@localhost's password: 
root@localhost's password: 
root@25eaf3b846dc:/#

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**
Client:
 Version:           unknown-version(this code is from master)
 API version:       1.39
 Go version:        devel +30c1ed7 Sat Oct 6 00:02:35 2018 +0000
 Git commit:        
 Built:             Tue Oct 23 03:38:45 2018
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          dev
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.11.1
  Git commit:       5d82d77
  Built:            Wed Oct 17 03:39:36 2018
  OS/Arch:          linux/amd64
  Experimental:     true

**Output of `docker info`:**

**Additional environment details (AWS, VirtualBox, physical, etc.):**
Aliyun cloud

lfmunoz:
Also is there a way to specify the ssh key , equivalent of ssh -i ${ACCESS_KEY}?

thaJeztah:
Cleaning up old issues and PRs; I have not seen any recent requests for this; using SSH with a password probably be much less common now, so I'm not sure theres a strong case to add support for this.

[...]

PR #1477, opened 2018-10-23
fix 1476 let ssh use stdin to type password when docker run with -it

lifubang:

**- What I did**
As the issue #1476, when use ssh with password, docker run -it will fail.

**- How I did it**
Because after attachContainer, ssh can't use stdin to type password.
So, in func attachContainer, use a call back func beforeHijack to run ContainerStart before hijackedIOStreamer.

**- How to verify it**
After fix:
```
docker -H ssh://localhost  run --name=testssh -it --entrypoint=/bin/bash docker.acmcoder.com/public/ubuntu:ssh
root@localhost's password: 
root@localhost's password: 
root@localhost's password: 
root@25eaf3b846dc:/# ls -lh
total 64K
drwxr-xr-x   2 root root 4.0K Oct  1  2015 bin
drwxr-xr-x   2 root root 4.0K Apr 10  2014 boot
```

modify attachContainer add a param "beforeHijack func() error"

h _zaman0 0 al 3 _luj

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -182,37 +182,42 @@ func runContainer(dockerCli command.Cli, opts *runOptions, copts *containerOptio
 		}()
 	}
 	attach := config.AttachStdin || config.AttachStdout || config.AttachStderr
+	statusChan := waitExitOrRemoved(ctx, dockerCli, createResponse.ID, copts.autoRemove)
+	startc := func() error {
+		if err := client.ContainerStart(ctx, createResponse.ID, types.ContainerStartOptions{}); err != nil {
+			// If we have hijackedIOStreamer, we should notify
+			// hijackedIOStreamer we are going to exit and wait
+			// to avoid the terminal are not restored.
+			if attach {
+				cancelFun()
+				<-errCh
+			}
+
+			reportError(stderr, "run", err.Error(), false)
+			if copts.autoRemove {
+				// wait container to be removed
+				<-statusChan
+			}
+			return runStartContainerErr(err)
+		}
+		return nil
+	}
 	if attach {
 		if opts.detachKeys != "" {
 			dockerCli.ConfigFile().DetachKeys = opts.detachKeys
 		}
 
-		close, err := attachContainer(ctx, dockerCli, &errCh, config, createResponse.ID)
+		// attach and start the container
+		// use callback to let user type ssh password when needed
+		close, err := attachContainer(ctx, dockerCli, &errCh, config, createResponse.ID, startc)
 
 		if err != nil {
 			return err
 		}
 		defer close()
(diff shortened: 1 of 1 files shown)


lifubang:
ping @AkihiroSuda PTAL

AkihiroSuda:
Thx, can we have e2e?

lifubang:
> Thx, can we have e2e?
@AkihiroSuda It's very hard to add e2e case, because:
1. We need `expect` to auto input user's password;
2. But `expect` have no stderr;
3. There is at least one "root@localhost's password:" in stdout, it's very hard to remove it.

So, I think it's very hard to add e2e case.
You can see my commit: https://github.com/lifubang/cli/commit/b121f2484944605cefdefd068215349518437e49
I think we can add e2e cases in a new PR if we need.

cli/command/container/run.go
```diff
@@ -289,6 +294,14 @@ func attachContainer(
 		return nil, errAttach
 	}
 
+	if beforeHijack != nil {
+		err := beforeHijack()
+		if err != nil {
```
AkihiroSuda: nit: `if err := beforeHijack(); err != nil {`

AkihiroSuda (dismissed):

AkihiroSuda:
Yes, E2E can be added separately, but I think password auth should not be considered to be supported officially for now then
https://github.com/docker/cli/pull/1482

cli/command/container/run.go
```diff
@@ -182,37 +182,42 @@ func runContainer(dockerCli command.Cli, opts *runOptions, copts *containerOptio
...
-			<-errCh
-		}
-
-		reportError(stderr, "run", err.Error(), false)
-		if copts.autoRemove {
-			// wait container to be removed
-			<-statusChan
-		}
-		return runStartContainerErr(err)
+	if !attach {
```
silvin-lubecki: I guess this `if !attach` can be moved as an `else` to the previous condition?
```go
if attach {
...
} else {
// start the container
...
}

```

lifubang: Thanks, you are right.

AkihiroSuda:
What's current status?

lifubang:
This patch can input password correctly. But I can't find a way to add a test case, I have no idea now. If someone have a time, please take a look #1487 . Thanks.

makitsune:
Will this be fixed soon? I cant use my development docker cli with my home server. I really need this!

thaJeztah:
Cleaning up old PRs; I have not seen any recent requests for this; using SSH with a password probably be much less common now, so I'm not sure theres a strong case to add support for this.

I'm closing this PR, but if there's a strong motivation for this, we can revisit.

PR #1489, opened 2018-10-29
Request each service's tasks separately

dperny:
Signed-off-by: Drew Erny <drew.erny@docker.com>

**- What I did**

When doing a `docker service ls`, get each service's tasks in separate requests, instead of all in one request.

**- How I did it**

There is currently a bug in swarmkit and the docker engine (moby/moby#37997) where large numbers of tasks can cause the message size to exceed the gRPC limits. A more appropriate fix involves changing the swarmkit gRPC api internally in some way or another to avoid this case occurring, but that fix requires more engineering effort and will take longer.

As a short-term band-aid, when using the CLI, split task list requests by service, which will cause the message size to not grow beyond the gRPC limits.

This fix leaks a bit of the abstraction presented by the engine API, but the resulting code is functionally equivalent. The alternative band-aid fix, contained solely to the engine, would involve attempting to deconstruct filters in the engine to make multiple requests, instead of merely converting. It is therefore cleaner to make this change here, where the request ultimately originates.

This change can and will be reverted when the engine fix goes in.

**- How to verify it**

As the code is functionally equivalent, the extant tests should cover it.

Use a different request to get every service's tasks when doing a service ls to avoid a current engine bug.

### DIFF
--- a/cli/command/service/list.go
+++ b/cli/command/service/list.go
@@ -60,14 +60,22 @@ func runList(dockerCli command.Cli, options listOptions) error {
 	info := map[string]formatter.ServiceListInfo{}
 	if len(services) > 0 && !options.quiet {
 		// only non-empty services and not quiet, should we call TaskList and NodeList api
-		taskFilter := filters.NewArgs()
+		// we should get tasks individually, one-by-one. This takes longer (a
+		// new request every time) but works around an issue where attempting
+		// to list all tasks causes a gRPC message that's too big in the engine.
+		//
+		// NOTE(dperny): this can be reverted once a proper engine fix goes in,
+		// but for now this band-aid should be fine.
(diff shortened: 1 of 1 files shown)


andrewhsu:
Just ran a test with this change and it definitely slows stuff down.

With 17.06, on a system with about 70 services and about 20 tasks per service, takes about 16 seconds. With this PR, takes about 38 seconds.

Thinking of alternatives...

andrewhsu:
something's coming in the air tonight...

andrewhsu:
@dperny i elect to close this PR in favor of the work in moby to address the same issue: https://github.com/moby/moby/pull/38103

thaJeztah:
Agreed, I don't think think this is the solution we're looking for 🤗

PR #1498, opened 2018-11-05
Adds flag modifying pull behavior for running and creating containers

Zanadar:
Should close issue https://github.com/moby/moby/issues/34394

**- What I did**
 - This PR adds a new `--pull` flag to `docker run` and `docker create`, following the proposal in (https://github.com/moby/moby/issues/34394)
- Per this proposal, the flag is tristate: 
  1. `--pull=missing` (this is the current behaviour and will be the default.)
  1. `--pull=never`
  1. `--pull=always`

**- How I did it**
 - Adds a new config field,  `createOptions.pull` which forks the execution of `container.createContainer` to either pull the image if it does not exist at all locally `--pull=missing`, always try and update the image `--pull=always`, or never try and update the image, only using images that already exist on the local machine `--pull=never`

Both `docker run` and  `docker create` can consume this flag.

**- How to verify it**
- Run `docker run` and `docker create` with the flag. Notice that pulls images or does not, depending on the behavior specified in the flag. Running those commands without the flag should maintain the current behavior.

Adds flag modifying pull behavior for running and creating containers (Pull if Missing, Pull Always, Pull Never)

honey-badger

### DIFF
--- a/cli/command/container/create.go
+++ b/cli/command/container/create.go
@@ -223,34 +223,30 @@ func createContainer(ctx context.Context, dockerCli command.Cli, containerConfig
 		}
 	}
 
-	//create the container, pulling the image (or not) based on opts.pull
-	var response container.ContainerCreateCreatedBody
-
-	if opts.pull == PullImageAlways {
+	pullAndTagImage := func() error {
 		if err := pullImage(ctx, dockerCli, config.Image, opts.platform, stderr); err != nil {
-			return nil, err
+			return err
 		}
 		if taggedRef, ok := namedRef.(reference.NamedTagged); ok && trustedRef != nil {
-			if err := image.TagTrusted(ctx, dockerCli, trustedRef, taggedRef); err != nil {
-				return nil, err
-			}
+			return image.TagTrusted(ctx, dockerCli, trustedRef, taggedRef)
+		}
+		return nil
+	}
+
+	if opts.pull == PullImageAlways {
+		if err := pullAndTagImage(); err != nil {
+			return nil, err
 		}
 	}
 
-	response, err = dockerCli.Client().ContainerCreate(ctx, config, hostConfig, networkingConfig, opts.name)
-
+	response, err := dockerCli.Client().ContainerCreate(ctx, config, hostConfig, networkingConfig, opts.name)
 	if err != nil {
 		// Pull image if it does not exist locally and we have the PullImageMissing option. Default behavior.
 		if apiclient.IsErrNotFound(err) && namedRef != nil && opts.pull == PullImageMissing {
-			fmt.Fprintf(stderr, "Unable to find image '%s' locally\n", reference.FamiliarString(namedRef))
 			// we don't want to write to stdout anything apart from container.ID
-			if err := pullImage(ctx, dockerCli, config.Image, opts.platform, stderr); err != nil {
+			fmt.Fprintf(stderr, "Unable to find image '%s' locally\n", reference.FamiliarString(namedRef))
+			if err := pullAndTagImage(); err != nil {
 				return nil, err
-			}
-			if taggedRef, ok := namedRef.(reference.NamedTagged); ok && trustedRef != nil {
-				if err := image.TagTrusted(ctx, dockerCli, trustedRef, taggedRef); err != nil {
-					return nil, err
-				}
 			}
 
 			var retryErr error
--- a/cli/command/container/create.go
+++ b/cli/command/container/create.go
@@ -223,37 +223,10 @@ func createContainer(ctx context.Context, dockerCli command.Cli, containerConfig
 		}
 	}
 
-	//create the container
+	//create the container, pulling the image (or not) based on opts.pull
 	var response container.ContainerCreateCreatedBody
-	if opts.pull == PullImageMissing { // Pull image only if it does not exist locally. Default.
-		response, err = dockerCli.Client().ContainerCreate(ctx, config, hostConfig, networkingConfig, opts.name)
 
-		//if image not found try to pull it
-		if err != nil {
-			if apiclient.IsErrNotFound(err) && namedRef != nil {
-				fmt.Fprintf(stderr, "Unable to find image '%s' locally\n", reference.FamiliarString(namedRef))
-
-				// we don't want to write to stdout anything apart from container.ID
-				if err := pullImage(ctx, dockerCli, config.Image, opts.platform, stderr); err != nil {
-					return nil, err
-				}
-				if taggedRef, ok := namedRef.(reference.NamedTagged); ok && trustedRef != nil {
-					if err := image.TagTrusted(ctx, dockerCli, trustedRef, taggedRef); err != nil {
-						return nil, err
-					}
-				}
-				// Retry
-				var retryErr error
-				response, retryErr = dockerCli.Client().ContainerCreate(ctx, config, hostConfig, networkingConfig, opts.name)
-				if retryErr != nil {
-					return nil, retryErr
-				}
-			} else {
-				return nil, err
-			}
-		}
-
-	} else if opts.pull == PullImageAlways { // Always try and pull the image.
+	if opts.pull == PullImageAlways {
 		if err := pullImage(ctx, dockerCli, config.Image, opts.platform, stderr); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


Zanadar:
Hey docker folks. This is my first PR to the project (👋 ). It's likely deficcient on tests  (I especailly wasn't sure about any kind of integration or e2e testing that needed to happen. Let me know what ya'll think.

lanoxx:
Nice work. I hope the docker folks will merge this.

cli/command/container/create.go
```diff
@@ -20,10 +20,18 @@ import (
 	"github.com/spf13/pflag"
 )
 
+// Pull constants
+const (
+	PullImageAlways  = "always"
+	PullImageMissing = "missing" // Default (matches previous bahevior)
```
rsjethani:
```suggestion
	PullImageMissing = "missing" // Default (matches previous behavior)
```

svenstaro:
What's the work remaining in order to get this merged?

Zanadar:
AFAIK nothing (but it’s been a while since I opened this so I may have forgotten). Maybe project folks have a different opinion?

svenstaro:
@GordonTheTurtle could you take a look?

Morl99:
@Zanadar can you try to increase the test coverage, you are slightly falling short, as per the above Check. I would love to see this feature merged, it is really useful!

cpuguy83:
Overall this looks good, but I think we can clean up the implementation a bit.

cpuguy83:
Also keep in mind pull-policy=always only really works for moving tags (such as `latest`).

cli/command/container/create.go
```diff
@@ -194,31 +205,59 @@ func createContainer(ctx context.Context, dockerCli command.Cli, containerConfig
 	}
 
 	//create the container
-	response, err := dockerCli.Client().ContainerCreate(ctx, config, hostConfig, networkingConfig, opts.name)
+	var response container.ContainerCreateCreatedBody
+	if opts.pull == PullImageMissing { // Pull image only if it does not exist locally. Default.
```
cpuguy83: I think we can do this flow and it will clean up quite a bit:

warning, psuedo-code:

```
if PullImageAlways {
  pull()
}

err := createContainer()
if IsNotfound(err) && PullImageMissing {
  pull()
}

createContainer()
```

Zanadar:
> Also keep in mind pull-policy=always only really works for moving tags (such as `latest`).

That's probably something that would have to be left to users, though, right? We can't necessarily know whether a tag is moving or not.

cpuguy83:
Right.

cli/command/container/create.go
Zanadar: Latest commit improves the flow per suggestion.

Zanadar: One thing I'm not certain on is whether we should error in the case that the image is not found locally and the "never pull" option is selected? In which case the flow becomes more like so:
```
if PullImageAlways {
  pull()
}

err := createContainer()
if err {
  if IsNotfound(err) && PullImageMissing {
    pull()
    createContainer()
  }
  return err // we won't be pulling so just return whatever err we got
}
```
This also saves trying to create the container twice in the "never pull" case (ie only try to create again in the case where we might pull).

Zanadar:
> Should we error out early if pull=always and the pull failed?

This is now handled in the first block: https://github.com/docker/cli/pull/1498/files#diff-477f60a1b609e77fdca573fd8bd7b0c9R210

Zanadar:
@cpuguy83 Okay I cleaned things up a fair bit. I'd love a pass on the tests and your thoughts on how we handle the error case when "never pull" is selected (see above).

cli/command/container/create.go
```diff
@@ -193,14 +204,26 @@ func createContainer(ctx context.Context, dockerCli command.Cli, containerConfig
 		}
 	}
 
-	//create the container
-	response, err := dockerCli.Client().ContainerCreate(ctx, config, hostConfig, networkingConfig, opts.name)
+	//create the container, pulling the image (or not) based on opts.pull
+	var response container.ContainerCreateCreatedBody
```
cpuguy83: Is there a reason we need this up here?

cli/command/container/create.go
```diff
@@ -193,14 +204,26 @@ func createContainer(ctx context.Context, dockerCli command.Cli, containerConfig
...
+
+	if opts.pull == PullImageAlways {
+		if err := pullImage(ctx, dockerCli, config.Image, opts.platform, stderr); err != nil {
```
cpuguy83: Would it be helpful to put all this behavior into a closure that can be reused in the error handling for `ContainerCreate`?

Zanadar: Yup that's cleaner.

cli/command/container/create.go
Zanadar: Nope! Cleaned that up.

Zanadar:
@cpuguy83 Addressed those comments.

cli/command/container/create_test.go
```diff
@@ -121,6 +122,93 @@ func TestCreateContainerPullsImageIfMissing(t *testing.T) {
...
+	_, err := createContainer(context.Background(), cli, config, &createOptions{
+		name:      "name",
+		platform:  runtime.GOOS,
+		untrusted: true,
+		pull:      PullImageNever,
+	})
+	assert.ErrorContains(t, err, "fake not found")
+}
+
+func TestCreateContainerAlwaysPullsImage(t *testing.T) {
```
cpuguy83: I'm not sure this test is testing much.
We should perhaps create multiple times and check the number of pulls == number of create calls?

cli/command/container/create_test.go
```diff
@@ -121,6 +122,93 @@ func TestCreateContainerPullsImageIfMissing(t *testing.T) {
...
+		) (container.ContainerCreateCreatedBody, error) {
+			defer func() { responseCounter++ }()
+			switch responseCounter {
+			case 0:
+				return container.ContainerCreateCreatedBody{}, fakeNotFound{}
+			default:
+				return container.ContainerCreateCreatedBody{}, errors.New("unexpected")
+			}
+		},
+		imageCreateFunc: func(parentReference string, options types.ImageCreateOptions) (io.ReadCloser, error) {
```
cpuguy83: Perhaps a counter here to ensure that that pull is never called?

Zanadar:
@cpuguy83 Made some improvements to the test per your suggestions.

cli/command/container/create_test.go
```diff
@@ -121,6 +122,105 @@ func TestCreateContainerPullsImageIfMissing(t *testing.T) {
 	assert.Check(t, is.Contains(stderr, "Unable to find image 'does-not-exist-locally:latest' locally"))
 }
 
+func TestCreateContainerNeverPullsImage(t *testing.T) {
+	imageName := "does-not-exist-locally"
+	responseCounter := 0
+	pullCounter := 0
+
+	client := &fakeClient{
```
cpuguy83: I'm still not sure about this test.
Let's keep the handler functions here simple and do what's expected (e.g. image create -> create fake image response) and have them increment the counters we are interested in.
Then when we actually call `createContainer` we check the pull counter to make sure it stays at zero (or whatever we expect based on the input).

I also think we can pull all this into one test like:

```go
func TestCreateContainerImagePullPolicy(t *testing.T) {
   // setup client stuff

  type testCase struct {
    Policy string
    ExpectedPulls int
    ExpectedErr err
  }
  cases := []testCase {...}
  for _, c := range cases {
    t.Run(c.Policy, func(t *testing.T) {
      // createContainer
      // check actual vs expected
    }
  }
}

Zanadar: Made a single table test for all the logic.

Zanadar:
@cpuguy83 Combined testing into a single table-driven test and cleaned up the handlers. It definitely makes for a nicer read.

Zanadar:
🐿

cpuguy83:
ping @vdemeester

mattc41190:
-- getting excited for this to be in --

AndrewFarley:
Any time soon?  :P

jeffb4:
Was this PR's latest build failure due to a build environment failure?

Zanadar:
@jeffb4 Not sure if that question is directed at me or not, but if there's anything I can do, let me know. It'd be *really* nice to get this PR merged. I think *many* people would be happy!

Zanadar:
(I looked at the jenkins logs and I couldn't make heads or tails of there being anything wrong)

cpuguy83:
I just pushed a rebase to see if that somehow magically fixes this.

Zanadar:
*woosh* ✨ *magic*

Zanadar:
🎆 🍾

cpuguy83:
Thanks for your patience all.

bsamuels453:
What's the difference between using --pull=always and --no-cache?

cpuguy83:
`--no-cache` is a `docker build` command and refers to build cache not image pulls.

eduardobaitello:
Now that this is merged, how to know in which stable CE version this flag/feature will become available?

_Sorry if this question doesn't fit here._

cpuguy83:
@eduardobaitello This did not make it into 19.03 from what I can tell, so it will be the next major release after 19.03.

thaJeztah:
Yes, it missed the 19.03 train, but you can download a nightly build of the CLI to try it; 

- linux: https://download.docker.com/linux/static/nightly/x86_64/
- mac: https://download.docker.com/mac/static/nightly/x86_64/

srinathck:
Are these changes not going into docker for windows?

thaJeztah:
@srinathck they are, but download.docker.com does not have downloads for Windows binaries (you can install Docker EE on Windows server, or Docker Desktop on Windows client once there's a new release with this change)

robbaman:
Is there a reason why the latest version of Docker Desktop on Windows still has version 19.03 of the CLI? According to this thread it seems 19.09 should've been worked on/ready for about 6 months already.

jcchavezs:
Could the default behaviour be overriden by an env var? I could come up with a PR.

gitnik:
This is still not available in engine version 19.03.08 in Docker for Mac. Any estimate when this will be available?

thaJeztah:
Next release (20.xx)

nickgrealy:
Two (stupid?) questions:

1. Is it possible to have this work with `docker pull <someimage>` commands too?

e.g. `docker pull --pull=missing nginx'

2. Is it possible to have `--pull=update` - to pull the latest version of an image, **only** if it's newer in the repo?

FernandoMiguel:
> Two (stupid?) questions:
> 
> 1. Is it possible to have this work with `docker pull <someimage>` commands too?
> 
> e.g. `docker pull --pull=always nginx`
> 
> 2. Is it possible to have `--pull=newer` - to pull the latest version of an image, **only** if it's newer in the repo?
> 

I don't understand your first one.
The second one is the behaviour of pull.

PR #1500, opened 2018-11-07
[do not merge] Fast context switch: foundation

simonferquel:
**- What I did**

This is foundational work for enabling support for fast context switch. ~This is related to https://github.com/moby/moby/pull/38148~. -> actually, the code from moby/moby has been pulled out into this PR. Keeping the link for discussion history

What it does is that it modifies how the CLI behaves to configure clients for both Docker engine and Kubernetes. It also change the way the default value for orchestrator in stacks related command is defined, making it possible to set a different default value on each context

**- How I did it**

This leverages the moby side PR, and add similar facilities to create/parse endpoint information for Kubernetes

**- How to verify it**

Unit tests + upcoming commands PR related to context switch

Fast context switch: you can now store, import and export credentials for different clusters and switch between them quickly using the command line

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -3,26 +3,25 @@ package command
 import (
 	"context"
 	"io"
-	"net"
-	"net/http"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strconv"
-	"time"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/config"
 	cliconfig "github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/config/configfile"
-	"github.com/docker/cli/cli/connhelper"
+	dcontext "github.com/docker/cli/cli/context"
+	"github.com/docker/cli/cli/context/docker"
+	kubcontext "github.com/docker/cli/cli/context/kubernetes"
+	"github.com/docker/cli/cli/context/store"
 	cliflags "github.com/docker/cli/cli/flags"
 	manifeststore "github.com/docker/cli/cli/manifest/store"
 	registryclient "github.com/docker/cli/cli/registry/client"
 	"github.com/docker/cli/cli/trust"
 	dopts "github.com/docker/cli/opts"
 	clitypes "github.com/docker/cli/types"
-	"github.com/docker/docker/api"
 	"github.com/docker/docker/api/types"
 	registrytypes "github.com/docker/docker/api/types/registry"
 	"github.com/docker/docker/client"
@@ -33,6 +32,9 @@ import (
 	notaryclient "github.com/theupdateframework/notary/client"
 	"github.com/theupdateframework/notary/passphrase"
 )
+
+// ContextDockerHost is the reported context when DOCKER_HOST env var or -H flag is set
+const ContextDockerHost = "<DOCKER_HOST>"
 
 // Streams is an interface which exposes the standard input and output streams
 type Streams interface {
@@ -57,6 +59,9 @@ type Cli interface {
 	RegistryClient(bool) registryclient.RegistryClient
 	ContentTrustEnabled() bool
 	NewContainerizedEngineClient(sockPath string) (clitypes.ContainerizedClient, error)
+	ContextStore() store.Store
+	CurrentContext() string
+	StackOrchestrator(flagValue string) (Orchestrator, error)
 }
 
 // DockerCli is an instance the docker command line client.
@@ -71,7 +76,15 @@ type DockerCli struct {
 	clientInfo            ClientInfo
 	contentTrust          bool
 	newContainerizeClient func(string) (clitypes.ContainerizedClient, error)
+	contextStore          store.Store
+	currentContext        string
 }
+
+var storeConfig = store.NewConfig(
+	func() interface{} { return &DockerContext{} },
+	store.EndpointTypeGetter(docker.DockerEndpoint, func() interface{} { return &docker.EndpointMeta{} }),
+	store.EndpointTypeGetter(kubcontext.KubernetesEndpoint, func() interface{} { return &kubcontext.EndpointMeta{} }),
+)
 
 // DefaultVersion returns api.defaultVersion or DOCKER_API_VERSION if specified.
 func (cli *DockerCli) DefaultVersion() string {
@@ -167,14 +180,23 @@ func (cli *DockerCli) RegistryClient(allowInsecure bool) registryclient.Registry
 // line flags are parsed.
 func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
-
 	var err error
-	cli.client, err = NewAPIClientFromFlags(opts.Common, cli.configFile)
+	cli.contextStore = store.New(cliconfig.ContextStoreDir(), storeConfig)
+	cli.currentContext, err = resolveContextName(opts.Common, cli.configFile)
+	if err != nil {
+		return err
+	}
+	endpoint, err := resolveDockerEndpoint(cli.contextStore, cli.currentContext, opts.Common)
+	if err != nil {
+		return errors.Wrap(err, "unable to resolve docker endpoint")
+	}
+
+	cli.client, err = newAPIClientFromEndpoint(endpoint, cli.configFile)
 	if tlsconfig.IsErrEncryptedKey(err) {
 		passRetriever := passphrase.PromptRetrieverWithInOut(cli.In(), cli.Out(), nil)
 		newClient := func(password string) (client.APIClient, error) {
-			opts.Common.TLSOptions.Passphrase = password
-			return NewAPIClientFromFlags(opts.Common, cli.configFile)
+			endpoint.TLSPassword = password
+			return newAPIClientFromEndpoint(endpoint, cli.configFile)
 		}
 		cli.client, err = getClientWithPassword(passRetriever, newClient)
 	}
@@ -196,6 +218,75 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	}
 	cli.initializeFromClient()
 	return nil
+}
+
+// NewAPIClientFromFlags creates a new APIClient from command line flags
+func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.ConfigFile) (client.APIClient, error) {
+	store := store.New(cliconfig.ContextStoreDir(), storeConfig)
+	contextName, err := resolveContextName(opts, configFile)
+	if err != nil {
+		return nil, err
+	}
+	endpoint, err := resolveDockerEndpoint(store, contextName, opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to resolve docker endpoint")
+	}
+	return newAPIClientFromEndpoint(endpoint, configFile)
+}
+
+func newAPIClientFromEndpoint(ep docker.Endpoint, configFile *configfile.ConfigFile) (client.APIClient, error) {
+	clientOpts, err := ep.ClientOpts()
+	if err != nil {
+		return nil, err
+	}
+	customHeaders := configFile.HTTPHeaders
+	if customHeaders == nil {
+		customHeaders = map[string]string{}
+	}
+	customHeaders["User-Agent"] = UserAgent()
+	clientOpts = append(clientOpts, client.WithHTTPHeaders(customHeaders))
+	return client.NewClientWithOpts(clientOpts...)
+}
+
+func resolveDockerEndpoint(s store.Store, contextName string, opts *cliflags.CommonOptions) (docker.Endpoint, error) {
+	if contextName != ContextDockerHost {
+		ctxMeta, err := s.GetContextMetadata(contextName)
+		if err != nil {
+			return docker.Endpoint{}, err
+		}
+		epMeta, err := docker.EndpointFromContext(ctxMeta)
+		if err != nil {
+			return docker.Endpoint{}, err
+		}
+		return epMeta.WithTLSData(s, contextName)
+	}
+	host, err := getServerHost(opts.Hosts, opts.TLSOptions)
+	if err != nil {
+		return docker.Endpoint{}, err
+	}
+
+	var (
+		skipTLSVerify bool
+		tlsData       *dcontext.TLSData
+	)
+
+	if opts.TLSOptions != nil {
+		skipTLSVerify = opts.TLSOptions.InsecureSkipVerify
+		tlsData, err = dcontext.TLSDataFromFiles(opts.TLSOptions.CAFile, opts.TLSOptions.CertFile, opts.TLSOptions.KeyFile)
+		if err != nil {
+			return docker.Endpoint{}, err
+		}
+	}
+
+	return docker.Endpoint{
+		EndpointMeta: docker.EndpointMeta{
+			EndpointMetaBase: dcontext.EndpointMetaBase{
+				Host:          host,
+				SkipTLSVerify: skipTLSVerify,
+			},
+		},
+		TLSData: tlsData,
+	}, nil
 }
 
 func isEnabled(value string) (bool, error) {
@@ -253,6 +344,51 @@ func (cli *DockerCli) NewContainerizedEngineClient(sockPath string) (clitypes.Co
 	return cli.newContainerizeClient(sockPath)
 }
 
+// ContextStore returns the ContextStore
+func (cli *DockerCli) ContextStore() store.Store {
+	return cli.contextStore
+}
+
+// CurrentContext returns the current context name
+func (cli *DockerCli) CurrentContext() string {
+	return cli.currentContext
+}
+
+// StackOrchestrator resolves which stack orchestrator is in use
+func (cli *DockerCli) StackOrchestrator(flagValue string) (Orchestrator, error) {
+	var ctxOrchestrator string
+
+	configFile := cli.configFile
+	if configFile == nil {
+		configFile = cliconfig.LoadDefaultConfigFile(cli.Err())
+	}
+
+	currentContext := cli.CurrentContext()
+	if currentContext == "" {
+		currentContext = configFile.CurrentContext
+	}
+	if currentContext == "" {
+		currentContext = ContextDockerHost
+	}
+	if currentContext != ContextDockerHost {
+		contextstore := cli.contextStore
+		if contextstore == nil {
+			contextstore = store.New(cliconfig.ContextStoreDir(), storeConfig)
+		}
+		ctxRaw, err := contextstore.GetContextMetadata(currentContext)
+		if err != nil {
+			return "", err
+		}
+		ctxMeta, err := GetDockerContext(ctxRaw)
+		if err != nil {
+			return "", err
+		}
+		ctxOrchestrator = string(ctxMeta.StackOrchestrator)
+	}
+
+	return GetStackOrchestrator(flagValue, ctxOrchestrator, configFile.StackOrchestrator, cli.Err())
+}
+
 // ServerInfo stores details about the supported features and platform of the
 // server
 type ServerInfo struct {
@@ -272,51 +408,6 @@ func NewDockerCli(in io.ReadCloser, out, err io.Writer, isTrusted bool, containe
 	return &DockerCli{in: NewInStream(in), out: NewOutStream(out), err: err, contentTrust: isTrusted, newContainerizeClient: containerizedFn}
 }
 
-// NewAPIClientFromFlags creates a new APIClient from command line flags
-func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.ConfigFile) (client.APIClient, error) {
-	host, err := getServerHost(opts.Hosts, opts.TLSOptions)
-	if err != nil {
-		return &client.Client{}, err
-	}
-	var clientOpts []func(*client.Client) error
-	helper, err := connhelper.GetConnectionHelper(host)
-	if err != nil {
-		return &client.Client{}, err
-	}
-	if helper == nil {
-		clientOpts = append(clientOpts, withHTTPClient(opts.TLSOptions))
-		clientOpts = append(clientOpts, client.WithHost(host))
-	} else {
-		clientOpts = append(clientOpts, func(c *client.Client) error {
-			httpClient := &http.Client{
-				// No tls
-				// No proxy
-				Transport: &http.Transport{
-					DialContext: helper.Dialer,
-				},
-			}
-			return client.WithHTTPClient(httpClient)(c)
-		})
-		clientOpts = append(clientOpts, client.WithHost(helper.Host))
-		clientOpts = append(clientOpts, client.WithDialContext(helper.Dialer))
-	}
-
-	customHeaders := configFile.HTTPHeaders
-	if customHeaders == nil {
-		customHeaders = map[string]string{}
-	}
-	customHeaders["User-Agent"] = UserAgent()
-	clientOpts = append(clientOpts, client.WithHTTPHeaders(customHeaders))
-
-	verStr := api.DefaultVersion
-	if tmpStr := os.Getenv("DOCKER_API_VERSION"); tmpStr != "" {
-		verStr = tmpStr
-	}
-	clientOpts = append(clientOpts, client.WithVersion(verStr))
-
-	return client.NewClientWithOpts(clientOpts...)
-}
-
 func getServerHost(hosts []string, tlsOptions *tlsconfig.Options) (string, error) {
 	var host string
 	switch len(hosts) {
@@ -331,35 +422,37 @@ func getServerHost(hosts []string, tlsOptions *tlsconfig.Options) (string, error
 	return dopts.ParseHost(tlsOptions != nil, host)
 }
 
-func withHTTPClient(tlsOpts *tlsconfig.Options) func(*client.Client) error {
-	return func(c *client.Client) error {
-		if tlsOpts == nil {
-			// Use the default HTTPClient
-			return nil
-		}
-
-		opts := *tlsOpts
-		opts.ExclusiveRootPools = true
-		tlsConfig, err := tlsconfig.Client(opts)
-		if err != nil {
-			return err
-		}
-
-		httpClient := &http.Client{
-			Transport: &http.Transport{
-				TLSClientConfig: tlsConfig,
-				DialContext: (&net.Dialer{
-					KeepAlive: 30 * time.Second,
-					Timeout:   30 * time.Second,
-				}).DialContext,
-			},
-			CheckRedirect: client.CheckRedirect,
-		}
-		return client.WithHTTPClient(httpClient)(c)
-	}
-}
-
 // UserAgent returns the user agent string used for making API requests
 func UserAgent() string {
 	return "Docker-Client/" + cli.Version + " (" + runtime.GOOS + ")"
+}
+
+// resolveContextName resolves the current context name with the following rules:
+// - setting both --context and --host flags is ambiguous
+// - if --context is set, use this value
+// - if --host flag or DOCKER_HOST is set, fallbacks to use the same logic as before context-store was added
+// for backward compatibility with existing scripts
+// - if DOCKER_CONTEXT is set, use this value
+// - if Config file has a globally set "CurrentContext", use this value
+// - fallbacks to default HOST, uses TLS config from flags/env vars
+func resolveContextName(opts *cliflags.CommonOptions, config *configfile.ConfigFile) (string, error) {
+	if opts.Context != "" && len(opts.Hosts) > 0 {
+		return "", errors.New("Conflicting options: either specify --host or --context, not bot")
+	}
+	if opts.Context != "" {
+		return opts.Context, nil
+	}
+	if len(opts.Hosts) > 0 {
+		return ContextDockerHost, nil
+	}
+	if _, present := os.LookupEnv("DOCKER_HOST"); present {
+		return ContextDockerHost, nil
+	}
+	if ctxName, ok := os.LookupEnv("DOCKER_CONTEXT"); ok {
+		return ctxName, nil
+	}
+	if config != nil && config.CurrentContext != "" {
+		return config.CurrentContext, nil
+	}
+	return ContextDockerHost, nil
 }
--- a/cli/command/cli_test.go
+++ b/cli/command/cli_test.go
@@ -66,6 +66,7 @@ func TestNewAPIClientFromFlagsForDefaultSchema(t *testing.T) {
 func TestNewAPIClientFromFlagsWithAPIVersionFromEnv(t *testing.T) {
 	customVersion := "v3.3.3"
 	defer env.Patch(t, "DOCKER_API_VERSION", customVersion)()
+	defer env.Patch(t, "DOCKER_HOST", ":2375")()
 
 	opts := &flags.CommonOptions{}
 	configFile := &configfile.ConfigFile{}
--- a/cli/command/context.go
+++ b/cli/command/context.go
@@ -1,0 +1,27 @@
+package command
+
+import (
+	"errors"
+
+	"github.com/docker/cli/cli/context/store"
+)
+
+// DockerContext is a typed representation of what we put in Context metadata
+type DockerContext struct {
+	Description       string       `json:"description,omitempty"`
+	StackOrchestrator Orchestrator `json:"stack_orchestrator,omitempty"`
+}
+
+// GetDockerContext extracts metadata from stored context metadata
+func GetDockerContext(storeMetadata store.ContextMetadata) (DockerContext, error) {
+	if storeMetadata.Metadata == nil {
+		// can happen if we save endpoints before assigning a context metadata
+		// it is totally valid, and we should return a default initialized value
+		return DockerContext{}, nil
+	}
+	res, ok := storeMetadata.Metadata.(DockerContext)
+	if !ok {
+		return DockerContext{}, errors.New("context metadata is not a valid DockerContext")
+	}
+	return res, nil
+}
--- a/cli/command/orchestrator.go
+++ b/cli/command/orchestrator.go
@@ -44,7 +44,7 @@ func normalize(value string) (Orchestrator, error) {
 		return OrchestratorKubernetes, nil
 	case "swarm":
 		return OrchestratorSwarm, nil
-	case "":
+	case "", "unset":
 		return orchestratorUnset, nil
 	case "all":
 		return OrchestratorAll, nil
@@ -53,9 +53,14 @@ func normalize(value string) (Orchestrator, error) {
 	}
 }
 
+// NormalizeOrchestrator parses an orchestrator value and checks if it is valid
+func NormalizeOrchestrator(value string) (Orchestrator, error) {
+	return normalize(value)
+}
+
 // GetStackOrchestrator checks DOCKER_STACK_ORCHESTRATOR environment variable and configuration file
 // orchestrator value and returns user defined Orchestrator.
-func GetStackOrchestrator(flagValue, value string, stderr io.Writer) (Orchestrator, error) {
+func GetStackOrchestrator(flagValue, contextValue, globalDefault string, stderr io.Writer) (Orchestrator, error) {
 	// Check flag
 	if o, err := normalize(flagValue); o != orchestratorUnset {
 		return o, err
@@ -68,8 +73,10 @@ func GetStackOrchestrator(flagValue, value string, stderr io.Writer) (Orchestrat
 	if o, err := normalize(env); o != orchestratorUnset {
 		return o, err
 	}
-	// Check specified orchestrator
-	if o, err := normalize(value); o != orchestratorUnset {
+	if o, err := normalize(contextValue); o != orchestratorUnset {
+		return o, err
+	}
+	if o, err := normalize(globalDefault); o != orchestratorUnset {
 		return o, err
 	}
 	// Nothing set, use default orchestrator
--- a/cli/command/stack/kubernetes/cli.go
+++ b/cli/command/stack/kubernetes/cli.go
@@ -7,12 +7,14 @@ import (
 	"os"
 
 	"github.com/docker/cli/cli/command"
+	kubecontext "github.com/docker/cli/cli/context/kubernetes"
 	"github.com/docker/cli/kubernetes"
 	cliv1beta1 "github.com/docker/cli/kubernetes/client/clientset/typed/compose/v1beta1"
 	"github.com/pkg/errors"
 	flag "github.com/spf13/pflag"
 	kubeclient "k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 // KubeCli holds kubernetes specifics (client, namespace) with the command.Cli
@@ -55,7 +57,18 @@ func WrapCli(dockerCli command.Cli, opts Options) (*KubeCli, error) {
 	cli := &KubeCli{
 		Cli: dockerCli,
 	}
-	clientConfig := kubernetes.NewKubernetesConfig(opts.Config)
+	var (
+		clientConfig clientcmd.ClientConfig
+		err          error
+	)
+	if dockerCli.CurrentContext() == command.ContextDockerHost {
+		clientConfig = kubernetes.NewKubernetesConfig(opts.Config)
+	} else {
+		clientConfig, err = kubecontext.ConfigFromContext(dockerCli.CurrentContext(), dockerCli.ContextStore())
+	}
+	if err != nil {
+		return nil, err
+	}
 
 	cli.kubeNamespace = opts.Namespace
 	if opts.Namespace == "" {
--- a/cli/command/system/version.go
+++ b/cli/command/system/version.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	kubecontext "github.com/docker/cli/cli/context/kubernetes"
 	"github.com/docker/cli/kubernetes"
 	"github.com/docker/cli/templates"
 	"github.com/docker/docker/api/types"
@@ -18,6 +19,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	kubernetesClient "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 var versionTemplate = `{{with .Client -}}
@@ -126,7 +128,7 @@ func runVersion(dockerCli command.Cli, opts *versionOptions) error {
 		return cli.StatusError{StatusCode: 64, Status: err.Error()}
 	}
 
-	orchestrator, err := command.GetStackOrchestrator("", dockerCli.ConfigFile().StackOrchestrator, dockerCli.Err())
+	orchestrator, err := dockerCli.StackOrchestrator("")
 	if err != nil {
 		return cli.StatusError{StatusCode: 64, Status: err.Error()}
 	}
@@ -151,7 +153,7 @@ func runVersion(dockerCli command.Cli, opts *versionOptions) error {
 		vd.Server = &sv
 		var kubeVersion *kubernetesVersion
 		if orchestrator.HasKubernetes() {
-			kubeVersion = getKubernetesVersion(opts.kubeConfig)
+			kubeVersion = getKubernetesVersion(dockerCli, opts.kubeConfig)
 		}
 		foundEngine := false
 		foundKubernetes := false
@@ -230,15 +232,27 @@ func getDetailsOrder(v types.ComponentVersion) []string {
 	return out
 }
 
-func getKubernetesVersion(kubeConfig string) *kubernetesVersion {
+func getKubernetesVersion(dockerCli command.Cli, kubeConfig string) *kubernetesVersion {
 	version := kubernetesVersion{
 		Kubernetes: "Unknown",
 		StackAPI:   "Unknown",
 	}
-	clientConfig := kubernetes.NewKubernetesConfig(kubeConfig)
-	config, err := clientConfig.ClientConfig()
+	var (
+		clientConfig clientcmd.ClientConfig
+		err          error
+	)
+	if dockerCli.CurrentContext() == command.ContextDockerHost {
+		clientConfig = kubernetes.NewKubernetesConfig(kubeConfig)
+	} else {
+		clientConfig, err = kubecontext.ConfigFromContext(dockerCli.CurrentContext(), dockerCli.ContextStore())
+	}
 	if err != nil {
 		logrus.Debugf("failed to get Kubernetes configuration: %s", err)
+		return &version
+	}
+	config, err := clientConfig.ClientConfig()
+	if err != nil {
+		logrus.Debugf("failed to get Kubernetes client config: %s", err)
 		return &version
 	}
 	kubeClient, err := kubernetesClient.NewForConfig(config)
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -18,6 +18,7 @@ const (
 	ConfigFileName = "config.json"
 	configFileDir  = ".docker"
 	oldConfigfile  = ".dockercfg"
+	contextsDir    = "contexts"
 )
 
 var (
@@ -33,6 +34,11 @@ func init() {
 // Dir returns the directory the configuration file is stored in
 func Dir() string {
 	return configDir
+}
+
+// ContextStoreDir returns the directory the docker contexts are stored in
+func ContextStoreDir() string {
+	return filepath.Join(Dir(), contextsDir)
 }
 
 // SetDir sets the directory the configuration file is stored in
--- a/cli/context/docker/load.go
+++ b/cli/context/docker/load.go
@@ -1,0 +1,172 @@
+package docker
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/docker/cli/cli/connhelper"
+	"github.com/docker/cli/cli/context"
+	"github.com/docker/cli/cli/context/store"
+	"github.com/docker/docker/client"
+	"github.com/docker/go-connections/tlsconfig"
+	"github.com/pkg/errors"
+)
+
+// EndpointMeta is a typed wrapper around a context-store generic endpoint describing
+// a Docker Engine endpoint, without its tls config
+type EndpointMeta struct {
+	context.EndpointMetaBase
+	APIVersion string `json:"api_version,omitempty"`
+}
+
+// Endpoint is a typed wrapper around a context-store generic endpoint describing
+// a Docker Engine endpoint, with its tls data
+type Endpoint struct {
+	EndpointMeta
+	TLSData     *context.TLSData
+	TLSPassword string
+}
+
+// WithTLSData loads TLS materials for the endpoint
+func (c *EndpointMeta) WithTLSData(s store.Store, contextName string) (Endpoint, error) {
+	tlsData, err := context.LoadTLSData(s, contextName, DockerEndpoint)
+	if err != nil {
+		return Endpoint{}, err
+	}
+	return Endpoint{
+		EndpointMeta: *c,
+		TLSData:      tlsData,
+	}, nil
+}
+
+// tlsConfig extracts a context docker endpoint TLS config
+func (c *Endpoint) tlsConfig() (*tls.Config, error) {
+	if c.TLSData == nil && !c.SkipTLSVerify {
+		// there is no specific tls config
+		return nil, nil
+	}
+	var tlsOpts []func(*tls.Config)
+	if c.TLSData != nil && c.TLSData.CA != nil {
+		certPool := x509.NewCertPool()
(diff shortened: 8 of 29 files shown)


silvin-lubecki:
Please rebase 🦁

cli/command/cli.go
```diff
@@ -163,20 +174,35 @@ func (cli *DockerCli) RegistryClient(allowInsecure bool) registryclient.Registry
...
+
+// CurrentContext returns the current context name
+func (cli *DockerCli) CurrentContext() string {
+	return cli.currentContext
+}
+
 // Initialize the dockerCli runs initialization that must happen after command
 // line flags are parsed.
 func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
+
```
silvin-lubecki: nit: empty line

ijc:
Vast majority of the commits here seem to be unrelated test case fixes? Maybe they should be split out?

cli/command/context.go
```diff
@@ -0,0 +1,44 @@
+package command
+
+import (
+	"github.com/docker/docker/pkg/contextstore"
+)
+
+// ContextMetadata is a typed representation of what we put in Context metadata
```
ijc: I'm a little confused about the relationship between what is here and what is in the moby/moby/client stuff e.g. the `Context` defined there but also the `consts` subpackage. They both seem to cover various aspects of the method for a client to connect to an engine.

What is the division of responsibility here? Do we need to "namespace" the metadata keys somehow?

cli/command/context.go
```diff
@@ -0,0 +1,44 @@
...
+
+// GetContextMetadata extract metadata from stored context metadata
+func GetContextMetadata(ctx contextstore.ContextMetadata) (ContextMetadata, error) {
+	if ctx.Metadata == nil {
+		return ContextMetadata{}, nil
+	}
+	result := ContextMetadata{}
+	var err error
+	if val, ok := ctx.Metadata["description"]; ok {
+		result.Description, _ = val.(string)
```
ijc: Didn't I see a string-metadata accessor function in the moby/moby PR?

cli/flags/common.go
```diff
@@ -27,6 +27,7 @@ var (
 	dockerCertPath  = os.Getenv("DOCKER_CERT_PATH")
 	dockerTLSVerify = os.Getenv("DOCKER_TLS_VERIFY") != ""
 	dockerTLS       = os.Getenv("DOCKER_TLS") != ""
+	dockerContext   = os.Getenv("DOCKER_CONTEXT")
```
ijc: There's a definition for `"DOCKER_CONTEXT"` this in moby/moby/client/client.go, use it?

cli/command/cli.go
```diff
@@ -34,6 +36,11 @@ import (
 	"github.com/theupdateframework/notary/passphrase"
 )
 
+const (
+	// ContextDockerHost is the reported context when DOCKER_HOST is used
+	ContextDockerHost = "<DOCKER_HOST>"
```
ijc: The moby/moby/client stuff deals with `<DOCKER_HOST>` in some ways too. The separations of concerns is a bit unclear here.

kubernetes/context.go
```diff
@@ -0,0 +1,187 @@
...
+// Context is a typed wrapper around a context-store context
+type Context struct {
+	Name             string
+	Server           string
+	SkipTLSVerify    bool
+	DefaultNamespace string
+}
+
+// LoadKubernetesConfig loads the kubernetes client config from the context
+func (c *Context) LoadKubernetesConfig(s contextstore.Store) (clientcmd.ClientConfig, error) {
```
ijc: There's no (non-test) caller in this PR, even if I consider #1501 I only see a use in `runExport`. But nowhere seems to use it "for real" (so to speak). Did I miss it or is there a 3rd part somewhere?

kubernetes/context.go
```diff
@@ -0,0 +1,187 @@
...
+		return err
+	}
+	clientcfg, err := cfg.ClientConfig()
+	if err != nil {
+		return err
+	}
+	var ca, key, cert []byte
+	if clientcfg.CAFile != "" {
+		ca, err = ioutil.ReadFile(clientcfg.CAFile)
+		if err != nil {
```
ijc: Can use `if ca, err = ... ; err != nil {...` here.

kubernetes/context.go
```diff
@@ -0,0 +1,187 @@
...
+			return err
+		}
+	} else {
+		ca = clientcfg.CAData
+	}
+	if clientcfg.KeyFile != "" {
+		key, err = ioutil.ReadFile(clientcfg.KeyFile)
+		if err != nil {
```
ijc: and here and a bunch of other places, I won't comment on them all.

cli/command/context.go
simonferquel: The contextstore is completely typeless. It only stores key-value pairs.
ContextMetadata in this package is simply a typed wrapper.

cli/command/context.go
simonferquel: Now that everything lives in the CLI, I will change that to avoid duplication

cli/flags/common.go
simonferquel: Not anymore

cli/command/cli.go
simonferquel: It should now be clearer. ContextDockerHost is now a special value used when resolving docker and kubernetes endpoints

kubernetes/context.go
simonferquel: Was a bug, thanks

cli/command/cli.go
```diff
@@ -71,6 +75,9 @@ type DockerCli struct {
 	clientInfo            ClientInfo
 	contentTrust          bool
 	newContainerizeClient func(string) (clitypes.ContainerizedClient, error)
+	contextStore          store.Store
+	currentContext        string
+	dockerEndpoint        docker.Endpoint
```
silvin-lubecki: Is this really needed? I don't see where it is used?

cli/command/context.go
ijc: In this case though their appeared to be two such typed wrappers covering the same "connect to Docker Engine" configuration case, this one and another one which was `moby/moby/client.Context` in that PR. 

That's what I don't get.

cli/command/context_test.go
```diff
@@ -0,0 +1,42 @@
...
+			},
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			storeMeta := store.ContextMetadata{}
+			SetContextMetadata(&storeMeta, c.data)
+			result, err := GetContextMetadata(storeMeta)
+			assert.Check(t, err)
```
silvin-lubecki: I don't understand this check. Use `assert.NilError` or `assert.Error` instead?

cli/command/orchestrator.go
```diff
@@ -53,9 +55,14 @@ func normalize(value string) (Orchestrator, error) {
 	}
 }
 
+// NormalizeOrchestrator parses an orchestrator value and check if it is valid
+func NormalizeOrchestrator(value string) (Orchestrator, error) {
```
silvin-lubecki: Looks like it's unused? 🤔

cli/command/orchestrator.go
```diff
@@ -53,9 +55,14 @@ func normalize(value string) (Orchestrator, error) {
...
+	return normalize(value)
+}
+
 // GetStackOrchestrator checks DOCKER_STACK_ORCHESTRATOR environment variable and configuration file
 // orchestrator value and returns user defined Orchestrator.
-func GetStackOrchestrator(flagValue, value string, stderr io.Writer) (Orchestrator, error) {
+func GetStackOrchestrator(flagValue string, dockerCli Cli, stderr io.Writer) (Orchestrator, error) {
```
silvin-lubecki: You can pass the context store instead.

silvin-lubecki: Or even better the current context.

cli/command/orchestrator.go
```diff
@@ -68,8 +75,27 @@ func GetStackOrchestrator(flagValue, value string, stderr io.Writer) (Orchestrat
...
+		meta, err := GetContextMetadata(rawMeta)
+		if err != nil {
+			return defaultOrchestrator, err
+		}
+		if meta.StackOrchestrator != orchestratorUnset {
+			return meta.StackOrchestrator, nil
+		}
+	}
+	// Check specified orchestrator in global config file
+	configFile := dockerCli.ConfigFile()
```
silvin-lubecki: I think this part should be reverted to the previous state. I don't understand why you put this part inside?

cli/command/context.go
simonferquel: Nope, there is a wrapper around context-scope metadata (with only description and defaultStackOrchestrator), and 2 at the endpoint level (1 in cli/context/docker and 1 in cli/context/kubernetes). To better differentiate them, endpoints wrappers are now correctly named as "Endpoint" :)

cli/command/cli.go
simonferquel: for now, it is not. But at some point, we want to have a subcommand `docker context create --from-env` which uses whatever host/tls env variables have been set to resolve this endpoint, and put them in a context.

cli/command/context_test.go
simonferquel: seem to be equivalent to assert.NilError. but will do.

cli/command/orchestrator.go
simonferquel: Used by context subcommands (`docker context create`)

cli/command/cli.go
silvin-lubecki: Yep but that can be added in this follow up I guess, for now it's like dead code 😉

cli/command/orchestrator.go
simonferquel: 👍 will do

cli/command/context.go
ijc: Great, hopefully it'll all become clear when I go through it all again then!

cli/command/orchestrator.go
silvin-lubecki: Ok it's used in the follow up #1501

cli/command/system/version.go
```diff
@@ -126,7 +126,7 @@ func runVersion(dockerCli command.Cli, opts *versionOptions) error {
 		return cli.StatusError{StatusCode: 64, Status: err.Error()}
 	}
 
-	orchestrator, err := command.GetStackOrchestrator("", dockerCli.ConfigFile().StackOrchestrator, dockerCli.Err())
+	orchestrator, err := command.GetStackOrchestrator("", dockerCli, dockerCli.Err())
```
silvin-lubecki: Should be reverted here too

cli/config/config.go
```diff
@@ -35,6 +35,11 @@ func Dir() string {
 	return configDir
 }
 
+// ContextStoreDir returns the directory the docker contexts are stored in
+func ContextStoreDir() string {
+	return filepath.Join(Dir(), "contexts")
```
silvin-lubecki: extract and document this constant?

cli/context/docker/load.go
```diff
@@ -0,0 +1,192 @@
...
+	EndpointMeta
+	TLSData     *TLSData
+	TLSPassword string
+}
+
+// WithTLSData loads TLS materials for the endpoint
+func (c *EndpointMeta) WithTLSData(s store.Store) (Endpoint, error) {
+	tlsFiles, err := s.ListContextTLSFiles(c.ContextName)
+	if err != nil {
+		return Endpoint{}, errors.Wrapf(err, "failed to retrieve context tls files for context %s", c.ContextName)
```
silvin-lubecki: nit: use %q?

cli/context/docker/load.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	if epTLSFiles, ok := tlsFiles[dockerEndpointKey]; ok {
+		var tlsData TLSData
+		for _, f := range epTLSFiles {
+			data, err := s.GetContextTLSData(c.ContextName, dockerEndpointKey, f)
+			if err != nil {
+				return Endpoint{}, errors.Wrapf(err, "failed to retrieve context tls data for file %s of context %s", f, c.ContextName)
```
silvin-lubecki: nit: use %q?

cli/command/cli.go
```diff
@@ -198,6 +214,83 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
...
+	if tmpStr := os.Getenv("DOCKER_API_VERSION"); tmpStr != "" {
+		verStr = tmpStr
+	}
+
+	var (
+		skipTLSVerify bool
+		tlsData       *docker.TLSData
+	)
+
+	if opts.TLSOptions != nil {
```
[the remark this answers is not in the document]
simonferquel: TLSData is the same in both docker and kubernetes endpoints. I made a common package and put a TLSDataFromFiles function there.

(This thing will also be used by the `docker context create` command)

cli/context/docker/load.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+	}
+	tlsConfig, err := c.tlsConfig()
+	if err != nil {
+		return err
+	}
+	if tlsConfig != nil {
+		httpClient := cli.HTTPClient()
+		if transport, ok := httpClient.Transport.(*http.Transport); ok {
+			transport.TLSClientConfig = tlsConfig
```
silvin-lubecki: Even if I don't know how to refactor that, it feels really smelly...

cli/context/docker/load.go
```diff
@@ -0,0 +1,192 @@
...
+	return nil
+}
+
+// Parse parses a context docker endpoint metadata into a typed Endpoint structure
+func Parse(name string, metadata store.ContextMetadata) (EndpointMeta, error) {
+	ep, ok := metadata.Endpoints[dockerEndpointKey]
+	if !ok {
+		return EndpointMeta{}, errors.New("cannot find docker endpoint in context")
+	}
+	host, _ := ep.GetString(hostKey)
```
silvin-lubecki: Check the second arguments for the following lines? Something can be wrong I guess?

cli/context/docker/save.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+	case err != nil:
+		return err
+	}
+	epMeta := make(store.Metadata)
+	epMeta[hostKey] = endpoint.Host
+	if endpoint.APIVersion != "" {
+		epMeta[apiVersionKey] = endpoint.APIVersion
+	}
+	if endpoint.SkipTLSVerify {
```
silvin-lubecki: nit: epMeta[skipTLSVerifyKey] = endpoint.SkipTLSVerify

cli/context/docker/load.go
simonferquel: yes, I know :(. impossible without modifying moby/moby

cli/context/docker/load.go
simonferquel: no, in that case, default string value is good enough

cli/context/kubernetes/endpoint_test.go
```diff
@@ -0,0 +1,127 @@
...
+			DefaultNamespace: defaultNamespace,
+			Server:           server,
+			SkipTLSVerify:    skipTLSVerify,
+		},
+		TLSData: tlsData,
+	}
+}
+
+func TestSaveLoadContexts(t *testing.T) {
+	storeDir, err := ioutil.TempDir("", "test-load-save-k8-context")
```
silvin-lubecki: nit: use t.Name()

cli/context/kubernetes/endpoint_test.go
```diff
@@ -0,0 +1,127 @@
...
+	assert.Check(t, err)
```
silvin-lubecki: Use `assert.NilError(t, err)`

cli/context/kubernetes/load.go
```diff
@@ -0,0 +1,95 @@
...
+// Endpoint is a typed wrapper around a context-store generic endpoint describing
+// a Kubernetes endpoint, with TLS data
+type Endpoint struct {
+	EndpointMeta
+	TLSData *common.TLSData
+}
+
+// WithTLSData loads TLS materials for the endpoint
+func (c *EndpointMeta) WithTLSData(s store.Store) (Endpoint, error) {
+	tlsFiles, err := s.ListContextTLSFiles(c.ContextName)
```
silvin-lubecki: Try to factorize it with the docker equivalent? Maybe in a follow-up?

simonferquel: I'll do that now

cli/context/kubernetes/save.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	case err != nil:
+		return err
+	}
+	epMeta := make(store.Metadata)
+	epMeta[serverKey] = endpoint.Server
+	epMeta[defaultNamespaceKey] = endpoint.DefaultNamespace
+	epMeta[skipTLSVerifyKey] = endpoint.SkipTLSVerify
+	ctxMeta.Endpoints[kubernetesEndpointKey] = epMeta
+	err = s.CreateOrUpdateContext(endpoint.ContextName, ctxMeta)
```
silvin-lubecki: nit: if err := ...

cli/context/kubernetes/save.go
```diff
@@ -0,0 +1,111 @@
...
+	if data.Cert != nil {
+		result.Files[certKey] = data.Cert
+	}
+	if data.Key != nil {
+		result.Files[keyKey] = data.Key
+	}
+	return &result
+}
+
+// FromKubeConfig creates a Kuberneted endpoint from a Kubeconfig file
```
silvin-lubecki: typo: KuberneteS

cli/context/kubernetes/save.go
```diff
@@ -0,0 +1,111 @@
...
+	ns, _, err := cfg.Namespace()
+	if err != nil {
+		return Endpoint{}, err
+	}
+	clientcfg, err := cfg.ClientConfig()
+	if err != nil {
+		return Endpoint{}, err
+	}
+	var ca, key, cert []byte
+	if clientcfg.CAFile != "" {
```
silvin-lubecki: We have 3 times the same code. I guess it can be factorized 😄

cli/context/store/doc.go
```diff
@@ -0,0 +1,22 @@
+// Package store provides a generic way to store credentials to connect to virtually any kind of remote system
```
silvin-lubecki: Missing a final point.

cli/context/store/doc.go
```diff
@@ -0,0 +1,22 @@
...
+// the `context` term comes from the similar feature in Kubernetes kubectl config files.
```
silvin-lubecki: The

cli/context/store/doc.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// conceptually, a context is a set of metadata and TLS data, that can be used to connect to various endpoints
```
silvin-lubecki: Conceptually

cli/context/store/doc.go
```diff
@@ -0,0 +1,22 @@
...
+// of the remote system. TLS data and metadata are stored separately, so that in the future, we will be able to store sensitive
+// information in a more secure way, depending on the os we are running on (e.g.: on Windows we could use the user Certificate Store, on Mac OS the user Keychain...)
```
silvin-lubecki: Missing a final point.

cli/context/store/doc.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// current implementation is purely file based with the following structure:
```
silvin-lubecki: Current

cli/context/store/doc.go
```diff
@@ -0,0 +1,22 @@
...
+// ${CONTEXT_ROOT}
+//   - config.json: contains the current "default" context
+//   - meta/
+//     - context1/meta.json: contains context medata (key/value pairs) as well as a list of endpoints (themselves containing key/value pair metadata)
+//     - contexts/can/also/be/folded/like/this/meta.json: same as context1, but for a context named `contexts/can/also/be/folded/like/this`
+//   - tls/
+//     - context1/endpoint1/: directory containing TLS data for the endpoint1 in context1
+//
+// the context store itself has abolutely no knowledge about a docker or a kubernetes endpoint should contain in term of metadata or TLS config
```
silvin-lubecki: typo: absolutely

silvin-lubecki: about what a docker... ?

cli/context/store/doc.go
```diff
@@ -0,0 +1,22 @@
...
+// client code is responsible for generating and parsing endpoint metadata and TLS files
```
silvin-lubecki: Client + missing final point.

cli/context/store/doc.go
```diff
@@ -0,0 +1,22 @@
...
+// The multi-endpoint approach of this package allows to combine many different endpoints in the same "context" (e.g., the Docker CLI
```
silvin-lubecki: multi-endpointS ?

silvin-lubecki:
The linter is complaining:
```
cli/context/store/store_test.go:1::warning: file is not gofmted with -s (gofmt)
cli/context/store/store.go:34:13:warning: "retreive" is a misspelling of "retrieve" (misspell)
cli/context/store/store.go:44:14:warning: "retreive" is a misspelling of "retrieve" (misspell)
cli/command/stack/cmd.go:78:65:warning: parameter stderr is unused (unparam)
```

simonferquel:
ping @thaJeztah @dhiltgen

cli/context/store/tlsstore.go
```diff
@@ -0,0 +1,84 @@
...
+	return filepath.Join(s.root, contextName, name)
+}
+
+func (s *tlsStore) filePath(contextName, endpointName, filename string) string {
+	return filepath.Join(s.root, contextName, endpointName, filename)
+}
+
+func (s *tlsStore) createOrUpdate(contextName, endpointName, filename string, data []byte) error {
+	epdir := s.endpointDir(contextName, endpointName)
+	if err := os.MkdirAll(epdir, 0700); err != nil {
```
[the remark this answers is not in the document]
simonferquel: We can't do an MKDir there, because endpointName is a subdir of $context_root/tls/$context_name.
However, tlsStore is not exposed and can only be created trough store.New, which creates the directory structure with correct permissions.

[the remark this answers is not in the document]
simonferquel: see https://github.com/docker/cli/pull/1500/files#diff-52f67b7303ba5e1a6c9cd3ec76a84e0cR79

cli/context/store/store.go
```diff
@@ -0,0 +1,329 @@
...
+
+func (s *store) ListContextTLSFiles(name string) (map[string]EndpointFiles, error) {
+	return s.tls.listContextData(name)
+}
+
+func (s *store) GetContextTLSData(contextName, endpointName, fileName string) ([]byte, error) {
+	return s.tls.getData(contextName, endpointName, fileName)
+}
+
+// Export exports an existing namespace into a stream
```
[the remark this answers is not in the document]
simonferquel: IMO, we should consider it opaque as in "undocumented", but backward compatible. In the future, if we choose to have a tlsStore that interacts with the Windows cert store, or MacOS keychain, we'll keep the same tar based format for import/export, but the way tlsStore will persist them at rest will be quite different.

cli/context/store/store.go
```diff
@@ -0,0 +1,329 @@
...
+	if err := os.MkdirAll(metaRoot, 0755); err != nil {
+		return nil, err
+	}
+	if err := os.MkdirAll(tlsRoot, 0700); err != nil {
+		return nil, err
+	}
+	_, err := os.Stat(configFile)
+
+	switch {
+	case os.IsNotExist(err):
```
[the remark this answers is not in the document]
simonferquel: 👍 file is now written only on SetCurrentContext

cli/context/store/store.go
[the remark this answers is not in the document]
simonferquel: Actually the tarball has not the exact same layout as the context store implementation: it does not have the contextName subdir in meta/ and tls/ roots.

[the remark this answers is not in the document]
ijc: What do you think of prepending some sort of non-tar header so that tools like `file(1)` or file managers etc do not think it is a tarball and people aren't tempted to pick it apart? As a bonus you could include a version field there to give some wriggle room if this ever needs to change format (you could also use a custom PAXHeader to achieve that though).

OTOH for debugging and things being able to `tar xf` is probably quite useful.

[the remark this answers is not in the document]
simonferquel: I am not sure yet about that. We might want at some point to document the format, so that cloud provider can provide managed Docker Hosting and emit Docker Contexts in this format instead of scripts based bundles. @thaJeztah wdyt ?

cli/command/cli.go
```diff
@@ -272,51 +382,6 @@ func NewDockerCli(in io.ReadCloser, out, err io.Writer, isTrusted bool, containe
 	return &DockerCli{in: NewInStream(in), out: NewOutStream(out), err: err, contentTrust: isTrusted, newContainerizeClient: containerizedFn}
 }
 
-// NewAPIClientFromFlags creates a new APIClient from command line flags
-func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.ConfigFile) (client.APIClient, error) {
```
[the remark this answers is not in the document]
simonferquel: Yes, I will re-add that, leveraging the same logic as the new Initialize.

cli/command/cli.go
```diff
@@ -167,14 +174,23 @@ func (cli *DockerCli) RegistryClient(allowInsecure bool) registryclient.Registry
 // line flags are parsed.
 func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
-
 	var err error
-	cli.client, err = NewAPIClientFromFlags(opts.Common, cli.configFile)
+	cli.contextStore, err = store.New(cliconfig.ContextStoreDir())
+	if err != nil {
```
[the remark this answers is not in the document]
simonferquel: I have very mixed feelings about that: context store failing to initiate could make the CLI connect to a different API endpoint than expected quite silently

cli/command/cli_test.go
```diff
@@ -31,7 +32,9 @@ func TestNewAPIClientFromFlags(t *testing.T) {
 			"My-Header": "Custom-Value",
 		},
 	}
-	apiclient, err := NewAPIClientFromFlags(opts, configFile)
+	ep, err := resolveDockerEndpoint(emptyContextStore{}, ContextDockerHost, opts)
```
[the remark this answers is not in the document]
simonferquel: 👍

cli/command/cli_test.go
```diff
@@ -51,7 +54,9 @@ func TestNewAPIClientFromFlagsForDefaultSchema(t *testing.T) {
...
```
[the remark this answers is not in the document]
simonferquel: 👍

cli/command/cli_test.go
```diff
@@ -69,7 +74,9 @@ func TestNewAPIClientFromFlagsWithAPIVersionFromEnv(t *testing.T) {
 
 	opts := &flags.CommonOptions{}
 	configFile := &configfile.ConfigFile{}
-	apiclient, err := NewAPIClientFromFlags(opts, configFile)
+	ep, err := resolveDockerEndpoint(emptyContextStore{}, ContextDockerHost, opts)
+	assert.NilError(t, err)
```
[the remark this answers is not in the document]
simonferquel: 👍

cli/command/cli.go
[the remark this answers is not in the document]
simonferquel: (and actually, to put make sure it is tested, I just rolled-back the tests code changes)

cli/command/cli.go
```diff
@@ -253,6 +329,40 @@ func (cli *DockerCli) NewContainerizedEngineClient(sockPath string) (clitypes.Co
...
+	return cli.contextStore
+}
+
+// CurrentContext returns the current context name
+func (cli *DockerCli) CurrentContext() string {
+	return cli.currentContext
+}
+
+// StackOrchestrator resolves which stack orchestrator is in use
+func (cli *DockerCli) StackOrchestrator(flagValue string) (Orchestrator, error) {
```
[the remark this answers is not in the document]
simonferquel: Like NormalizeOrchestrator, it is used in the followup PR (but it makes more sense to put it in the foundations, as it is "foundational work")

cli/command/stack/kubernetes/cli.go
```diff
@@ -55,7 +55,10 @@ func WrapCli(dockerCli command.Cli, opts Options) (*KubeCli, error) {
 	cli := &KubeCli{
 		Cli: dockerCli,
 	}
-	clientConfig := kubernetes.NewKubernetesConfig(opts.Config)
+	clientConfig, err := kubernetes.NewKubernetesConfig(dockerCli.ContextStore(), dockerCli.CurrentContext(), opts.Config)
```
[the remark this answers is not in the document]
simonferquel: Which kubernetes package ? `cli/command/stack/kubernetes` or `kubernetes` ?
If it is the first, is it ok to only pass the whole dockerCli pointer ? (which introduce a dependency to cli/command in `kubernetes` package).
If it is the second, I don't really see the problem here: this package responsibility is to resolve kubernetes client config for the command. It needs to know if we are leveraging contexts, or if we reference a specific kubeconfig file (or if we should fallback to default kubeconfig file, for backward compat)

cli/context/common/endpoint.go
```diff
@@ -0,0 +1,43 @@
+package common
```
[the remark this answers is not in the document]
simonferquel: 👍

cli/command/cli.go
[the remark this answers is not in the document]
vdemeester: Well, with a warning, it would not :angel: 
My main concern is here, if on a setup, there is an existing docker config but the folder (like `.docker`) is read-only (for whatever reason), before that change, it used to work, after it will fail… which doesn't seem right.

cli/context/common/tlsdata.go
```diff
@@ -0,0 +1,92 @@
...
+		result.Files[certKey] = data.Cert
+	}
+	if data.Key != nil {
+		result.Files[keyKey] = data.Key
+	}
+	return &result
+}
+
+// LoadTLSData loads TLS data from the store
+func LoadTLSData(s store.Store, contextName, endpointName string) (*TLSData, error) {
```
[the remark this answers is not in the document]
simonferquel: It could, but I would like to keep the "store" package completely generic. I will move `common` files in the parent package though.

cli/command/cli.go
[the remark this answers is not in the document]
ijc: If Simon was to move the `MkdirAll` calls from `store.New` into the place which actuall writes files (similarly to how he previously changed `New` to not write an empty `{}` file if it doesn't already exist) then this call would be read only as well (AFAICS). Would that satisfy your concern?

cli/command/stack/kubernetes/cli.go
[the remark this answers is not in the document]
vdemeester: I mean the 2nd, aka https://github.com/docker/cli/pull/1500/files#diff-e9b256a7b6baeee90cd95e40eb13e218R16.
Given that `github.com/docker/cli/cli/context/kubernetes` already know how to generate a k8s config, maybe the signature should only be sthg like the following.

```go
func NewKubernetesConfig(configPath string, contextConfig *clientcmd.ClientConfig) (clientcmd.ClientConfig, error) {
// […]
```

That way, the `github.com/docker/cli/kubernetes` package doesn't have to know anything about context(s) packages — and thus can be used without.

cli/context/store/doc.go
```diff
@@ -0,0 +1,22 @@
+// Package store provides a generic way to store credentials to connect to virtually any kind of remote system.
+// The `context` term comes from the similar feature in Kubernetes kubectl config files.
+//
+// Conceptually, a context is a set of metadata and TLS data, that can be used to connect to various endpoints
+// of the remote system. TLS data and metadata are stored separately, so that in the future, we will be able to store sensitive
+// information in a more secure way, depending on the os we are running on (e.g.: on Windows we could use the user Certificate Store, on Mac OS the user Keychain...).
+//
+// Current implementation is purely file based with the following structure:
+// ${CONTEXT_ROOT}
+//   - config.json: contains the current "default" context
```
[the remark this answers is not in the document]
simonferquel: it is in $HOME/.docker/contexts/config.json

cli/context/store/doc.go
```diff
@@ -0,0 +1,22 @@
...
+//   - meta/
+//     - context1/meta.json: contains context medata (key/value pairs) as well as a list of endpoints (themselves containing key/value pair metadata)
+//     - contexts/can/also/be/folded/like/this/meta.json: same as context1, but for a context named `contexts/can/also/be/folded/like/this`
```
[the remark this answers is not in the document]
simonferquel: there is no inheritance at all (so far at least)

cli/command/stack/kubernetes/cli.go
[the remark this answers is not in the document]
simonferquel: I just tried and actually it feels worse: at every place, you have first to check if the current context is ContextDockerHost, then find if the context has actually a kubernetes endpoint, then extract the client config from it. Won't do that :/

cli/context/docker/load.go
```diff
@@ -0,0 +1,166 @@
...
+		helper, err := connhelper.GetConnectionHelper(c.Host)
+		if err != nil {
+			return err
+		}
+		if helper == nil {
+			if err := client.WithHost(c.Host)(cli); err != nil {
+				return err
+			}
+		} else {
+			httpClient := &http.Client{
```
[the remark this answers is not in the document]
simonferquel: there was something similar done at the cli/command package I think: https://github.com/docker/cli/blob/master/cli/command/cli.go#L334 (and we can't use the moby/moby WithTLS function, as it requires filepaths and not directly TLS data).

cli/command/stack/kubernetes/cli.go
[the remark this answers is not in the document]
vdemeester: :thinking: It's just about extracting the part in `if configPath == "" && contextName != command.ContextDockerHost` elsewhere (in `github.com/docker/cli/cli/context/kubernetes` for ex)…
There is 2 places where it's done ([here](https://github.com/docker/cli/pull/1500/files#diff-8d7a16ab24f6e96e9673178204f32156R58) and [here](https://github.com/docker/cli/pull/1500/files#diff-08f5ffbd1b84c8ee25fca4a4ed2647a1R238)).

What I am suggesting is something like the following.

```go
import (
    // […]
    contextKubernetes "github.com/docker/cli/cli/context/kubernetes"
    // […]
)

// […]
contextConfig, err := contextKubernetes.ConfigFromContext(dockerCli.ContextStore(), dockerCli.CurrentContext())
// […]
clientConfig, err := kubernetes.NewKubernetesConfig(kubeConfig, contextConfig)
// […]
```

cli/command/cli.go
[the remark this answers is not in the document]
simonferquel: Hmm, depending on the ordering of calls, root of the context store would be created either with 0700 or 0755... Is it ok with you if I make all the resources inside the context dir as 0700 ?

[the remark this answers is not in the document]
simonferquel: Oh and there is still
```
configBytes, err := ioutil.ReadFile(configFile)
	switch {
	case os.IsNotExist(err):
		// don't do anything, currentContext is not set
	case err != nil:
		return nil, err
	default:
		var cfg config
		if err := json.Unmarshal(configBytes, &cfg); err != nil {
			return nil, err
		}
		currentContext = cfg.CurrentContext
	}
``` 
should we treat all read errors as "IsNotExist" ?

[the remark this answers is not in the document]
vdemeester: @simonferquel with a warning (log), yeah I think so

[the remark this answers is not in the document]
ijc: > Hmm, depending on the ordering of calls, root of the context store would be created either with 0700 or 0755... Is it ok with you if I make all the resources inside the context dir as 0700 ?

Either that or everywhere which makes the restricted subdir needs to first make the root of the context store with 0755, perhaps via a helper? But if the code structure (e.g. availability of the context root at various points) makes that hard then having the whole context store be 0700 doesn't seem that bad to me.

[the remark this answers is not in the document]
simonferquel: What about removing the config.json file from the context store (which is here only to set the current context), and putting a currentContext field in the CLI config file ?

cli/command/stack/kubernetes/cli.go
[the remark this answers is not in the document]
simonferquel: what if I kept kubernetes.NewKubernetesConfig unmodified (with no reference to contexts), and have `contextKubernetes.ConfigFromContext(dockerCli.ContextStore(), dockerCli.CurrentContext(), kubeconfigOverride` which would fallback on kubernetes.NewKubernetesConfig instead ?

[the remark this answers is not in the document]
vdemeester: @simonferquel that would work too :angel: as long as there is no `cli/cli/context/…` import in the root `kubernetes` package :angel:

cli/command/cli.go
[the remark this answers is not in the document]
vdemeester: Works for me
/cc @thaJeztah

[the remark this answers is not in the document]
ijc: WFM too (and is probably less surprising for users to have it all in one place)

simonferquel:
@silvin-lubecki 
good catch. was related to the non-erroring init.
Added a test and fixed

simonferquel:
@vdemeester aside from documentation completion, could you have another look at this? I think I fixed most of your raised issues.
@thaJeztah PTAL as well :)

cli/context/store/store.go
```diff
@@ -0,0 +1,304 @@
...
+		return false, false
+	}
+	b, ok := v.(bool)
+	return b, ok
+}
+
+// ContextMetadata contains metadata about a context and its endpoints
+type ContextMetadata struct {
+	Metadata  Metadata            `json:"metadata,omitempty"`
+	Endpoints map[string]Metadata `json:"endpoints,omitempty"`
```
thaJeztah: I was wondering why we didn't make this a `map[string]EndpointMeta` but I see there's two different versions of that (kube/docker), which is a pity (would've made things a bit more simple I guess 😞 

```go
type EndpointMeta struct {
	context.EndpointMetaBase
	APIVersion string
}

type EndpointMeta struct {
	context.EndpointMetaBase
	DefaultNamespace string
}
```


Wondering if we made this;

```go
Endpoints map[string]interface{}
```

Then we could do the conversion when _storing_, and store either `KubeEndpoint` or `DockerEndpoint` here 🤔

cli/flags/common.go
```diff
@@ -70,6 +72,8 @@ func (commonOpts *CommonOptions) InstallFlags(flags *pflag.FlagSet) {
 	// opts.ValidateHost is not used here, so as to allow connection helpers
 	hostOpt := opts.NewNamedListOptsRef("hosts", &commonOpts.Hosts, nil)
 	flags.VarP(hostOpt, "host", "H", "Daemon socket(s) to connect to")
+	flags.StringVarP(&commonOpts.Context, "context", "c", dockerContext,
```
thaJeztah: Looks like we have to update the docs for this, or is that done in the other PR?

cli/flags/common.go
```diff
@@ -70,6 +72,8 @@ func (commonOpts *CommonOptions) InstallFlags(flags *pflag.FlagSet) {
...
+		`context to use to connect to the daemon (overrides host flag, DOCKER_HOST env var and default context set with "docker context use")`)
```
thaJeztah: "name of the context to use ..." perhaps?

cli/command/orchestrator.go
```diff
@@ -44,7 +44,7 @@ func normalize(value string) (Orchestrator, error) {
 		return OrchestratorKubernetes, nil
 	case "swarm":
 		return OrchestratorSwarm, nil
-	case "":
+	case "", "unset":
```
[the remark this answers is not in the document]
simonferquel: I agree. I had to handle the "unset" value case, as I normalize the orchestrator string both on context creation (to validate the input) and when I load the context. When creating the context, an empty orchestrator value gets normalized into "unset", so loading it must handle the special value correctly.

I could change the `orchestratorUnset` const, but I am worried about changing an existing behavior.

cli/command/context.go
```diff
@@ -0,0 +1,38 @@
...
+
+// ContextMetadata is a typed representation of what we put in Context metadata
+type ContextMetadata struct {
+	Description       string
+	StackOrchestrator Orchestrator
+}
+
+// SetContextMetadata sets the metadata inside a stored context
+func SetContextMetadata(ctx *store.ContextMetadata, metadata ContextMetadata) {
+	ctx.Metadata = map[string]interface{}{
```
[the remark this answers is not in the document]
simonferquel: Yeah, the second one being a strongly-typed representation of a Docker CLI context, I suppose DockerContext could make sense. WDYT?

cli/command/context.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// ContextMetadata is a typed representation of what we put in Context metadata
+type ContextMetadata struct {
+	Description       string
+	StackOrchestrator Orchestrator
+}
+
+// SetContextMetadata sets the metadata inside a stored context
+func SetContextMetadata(ctx *store.ContextMetadata, metadata ContextMetadata) {
```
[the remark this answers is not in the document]
simonferquel: `contextMetadata` ? `storeMetadata` ?

cli/context/store/store.go
simonferquel: :+1 It would greatly simplify Endpoint Marshalling/Unmarshalling

simonferquel: Hmm, actually no, it wouldn't. But it would make it much cleaner (ContextMetadata would need to implement the json Marshal interface)

cli/flags/common.go
simonferquel: Should do in this one

cli/command/cli.go
```diff
@@ -331,35 +405,27 @@ func getServerHost(hosts []string, tlsOptions *tlsconfig.Options) (string, error
...
-	}
-}
-
 // UserAgent returns the user agent string used for making API requests
 func UserAgent() string {
 	return "Docker-Client/" + cli.Version + " (" + runtime.GOOS + ")"
 }
+
+func resolveContextName(opts *cliflags.CommonOptions, config *configfile.ConfigFile) string {
+	if opts.Context != "" {
```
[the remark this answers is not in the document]
simonferquel: It is in sync with the `--context` flag description. But my guess is that having both opts.Context and opts.Hosts should be considered as ambiguous (and maybe also DOCKER_HOST vs DOCKER_CONTEXT, but I am a bit less sure for existing scripts compatibility)

cli/flags/common.go
simonferquel: Actually, makes more sense to put all the doc in the following PR so that documentation for commands impacting the default context etc. are documented in sync.

I think once I get 2 LGTM on this one, I will squash, rebase #1501 and close this one so that we merge the full context switch feature at once. Was very usefull for review to split though.

simonferquel:
@thaJeztah PTAL at the last commit I added. It brings a way not to have to do manual marshaling at the cost of some complexity on the context store side. Not sure if I like it.

cli/command/orchestrator.go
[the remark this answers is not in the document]
thaJeztah: > I could change the orchestratorUnset const, but I am worried about changing an existing behavior.

Ahm. yes, I see it was already there. At least `orchestratorUnset` is not exported, but we should be careful indeed.

@vdemeester @silvin-lubecki suggestions? (we can do that in a follow up if we want to)

cli/command/orchestrator.go
thaJeztah: Ah, yes; also noticed it wasn't used 😂

cli/command/orchestrator.go
```diff
@@ -53,9 +53,14 @@ func normalize(value string) (Orchestrator, error) {
...
 
+// NormalizeOrchestrator parses an orchestrator value and checks if it is valid
+func NormalizeOrchestrator(value string) (Orchestrator, error) {
+	return normalize(value)
+}
+
 // GetStackOrchestrator checks DOCKER_STACK_ORCHESTRATOR environment variable and configuration file
 // orchestrator value and returns user defined Orchestrator.
-func GetStackOrchestrator(flagValue, value string, stderr io.Writer) (Orchestrator, error) {
+func GetStackOrchestrator(flagValue, contextValue, globalDefault string, stderr io.Writer) (Orchestrator, error) {
```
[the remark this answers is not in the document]
simonferquel: it was exported before. I changed the signature to put the contextValue in the middle. I suspect 3rd party code (like docker-app) leverages this function.

cli/command/cli.go
```diff
@@ -331,35 +411,39 @@ func getServerHost(hosts []string, tlsOptions *tlsconfig.Options) (string, error
...
+// - if --context is set, use this value
+// - if --host flag or DOCKER_HOST is set, fallbacks to use the same logic as before context-store was added
+// for backward compatibility with existing scripts
+// - if DOCKER_CONTEXT is set, use this value
+// - if Config file has a globally set "CurrentContext", use this value
+// - fallbacks to default HOST, uses TLS config from flags/env vars
+func resolveContextName(opts *cliflags.CommonOptions, config *configfile.ConfigFile) (string, error) {
+	// spcifying both --host and --context is considered ambiguous
+	if opts.Context != "" && len(opts.Hosts) > 0 {
+		return "", errors.New("both --host and --context can't be set at the same time")
```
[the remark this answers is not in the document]
simonferquel: in golang, errors should not begin with a capital letter :) but otherwise I like the wording

cli/command/orchestrator.go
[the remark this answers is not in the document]
thaJeztah: Ah, yes, possible

cli/command/cli.go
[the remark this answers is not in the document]
thaJeztah: Yes, I was in doubt; I see that in the CLI we're not consistent, likely because this message is directly printed in the terminal (so won't be in a log with an `Error: ` prefix (e.g.))

cli/command/orchestrator.go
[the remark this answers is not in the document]
simonferquel: I confirm, docker-app uses it :)

cli/command/cli.go
```diff
@@ -331,35 +411,38 @@ func getServerHost(hosts []string, tlsOptions *tlsconfig.Options) (string, error
...
+	if len(opts.Hosts) > 0 {
+		return ContextDockerHost, nil
+	}
+	if _, present := os.LookupEnv("DOCKER_HOST"); present {
+		return ContextDockerHost, nil
+	}
+	if ctxName, ok := os.LookupEnv("DOCKER_CONTEXT"); ok {
+		return ctxName, nil
+	}
+	ctxName := ContextDockerHost
```
silvin-lubecki: nit:
```go
if config != nil && config.CurrentContext != ""{
    return config.CurrentContext, nil
}
return ContextDockerHost, nil
```

cli/command/cli.go
```diff
@@ -198,6 +220,75 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
...
+	customHeaders := configFile.HTTPHeaders
+	if customHeaders == nil {
+		customHeaders = map[string]string{}
+	}
+	customHeaders["User-Agent"] = UserAgent()
+	clientOpts = append(clientOpts, client.WithHTTPHeaders(customHeaders))
+	return client.NewClientWithOpts(clientOpts...)
+}
+
+func resolveDockerEndpoint(s store.Store, contextName string, opts *cliflags.CommonOptions) (docker.Endpoint, error) {
```
silvin-lubecki: This file becomes quite complicated (~= 450 lines). Do you think you can split it, like moving these helpers (`resolveDockerEndpoint`, `resolveContextName` or `newAPIClientFromEndpoint`) to another file? It can be in a follow-up.

cli/command/context.go
```diff
@@ -0,0 +1,27 @@
...
+// GetDockerContext extracts metadata from stored context metadata
+func GetDockerContext(storeMetadata store.ContextMetadata) (DockerContext, error) {
+	if storeMetadata.Metadata == nil {
+		// can happen if we save endpoints before assigning a context metadata
+		// it is totally valid, and we should return a default initialized value
+		return DockerContext{}, nil
+	}
+	res, ok := storeMetadata.Metadata.(DockerContext)
+	if !ok {
+		return DockerContext{}, errors.New("context metadata is not of type DockerContext")
```
silvin-lubecki: "context metadata is not of type DockerContext" -> "context metadata is not a valid DockerContext"

cli/command/orchestrator.go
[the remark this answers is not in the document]
silvin-lubecki: Not big fan either. I think it's safe to change the constant to `orchestratorUnset = Orchestrator("")` and remove `unset`. No-one should rely on this constant.

cli/command/system/version.go
```diff
@@ -230,17 +232,29 @@ func getDetailsOrder(v types.ComponentVersion) []string {
...
 
-func getKubernetesVersion(kubeConfig string) *kubernetesVersion {
+func getKubernetesVersion(dockerCli command.Cli, kubeConfig string) *kubernetesVersion {
 	version := kubernetesVersion{
 		Kubernetes: "Unknown",
 		StackAPI:   "Unknown",
 	}
-	clientConfig := kubernetes.NewKubernetesConfig(kubeConfig)
-	config, err := clientConfig.ClientConfig()
+	var (
```
silvin-lubecki: This code can be factorized with https://github.com/docker/cli/pull/1500/files#diff-8d7a16ab24f6e96e9673178204f32156R60

kubernetes/config.go
```diff
@@ -10,7 +10,8 @@ import (
 
 // NewKubernetesConfig resolves the path to the desired Kubernetes configuration file based on
 // the KUBECONFIG environment variable and command line flags.
-func NewKubernetesConfig(configPath string) clientcmd.ClientConfig {
+func NewKubernetesConfig(configPath string) (clientcmd.ClientConfig, error) {
+
```
silvin-lubecki: nit: empty line

cli/command/cli.go
simonferquel: I vote for a followup, to make it easier to track modification when I squash

cli/command/system/version.go
simonferquel: In a followup maybe ? the tricky part here is to avoid cyclic dependencies. We could always introduce a new package...

simonferquel:
@silvin-lubecki @vdemeester @thaJeztah, I fixed all your feedback, and squashed everything. PTAL

cli/command/cli.go
silvin-lubecki: ok then!

cli/context/kubernetes/endpoint_test.go
```diff
@@ -0,0 +1,183 @@
...
+	assert.NilError(t, testEndpoint("https://test", "test", nil, nil, nil, true).Save(store, "raw-notls-skip"))
+	assert.NilError(t, testEndpoint("https://test", "test", []byte("ca"), []byte("cert"), []byte("key"), true).Save(store, "raw-tls"))
+
+	kcFile, err := ioutil.TempFile(os.TempDir(), "test-load-save-k8-context")
+	assert.NilError(t, err)
+	defer os.Remove(kcFile.Name())
+	defer kcFile.Close()
+	cfg := clientcmdapi.NewConfig()
+	cfg.AuthInfos["user"] = clientcmdapi.NewAuthInfo()
+	cfg.Contexts["context1"] = clientcmdapi.NewContext()
```
thaJeztah: It's a pity that their constructors don't allow passing options for the new things 😞 they're currently not that useful

cli/context/store/doc.go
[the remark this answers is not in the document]
thaJeztah: Inheritance would be good indeed (being able to set global options and override them per context); guess we can do so in a follow up?

cli/flags/common.go
```diff
@@ -70,6 +71,8 @@ func (commonOpts *CommonOptions) InstallFlags(flags *pflag.FlagSet) {
 	// opts.ValidateHost is not used here, so as to allow connection helpers
 	hostOpt := opts.NewNamedListOptsRef("hosts", &commonOpts.Hosts, nil)
 	flags.VarP(hostOpt, "host", "H", "Daemon socket(s) to connect to")
+	flags.StringVarP(&commonOpts.Context, "context", "c", "",
```
thaJeztah: Needs a follow-up to update the completion scripts (and the documentation / man pages)

/cc @albers

simonferquel: I plan to add documentation as part of #1501. I might need help for completion scripts though.

thaJeztah:
Discussing this PR with @simonferquel @silvin-lubecki @andrewhsu @tiborvass, and we think it might be good to merge this _together_ with https://github.com/docker/cli/pull/1501, so we either want to create a feature branch (to collect the other changes that makes the feature complete), or just merge them to master together.

(That, to take into account possible issues that follow review / testing of #1501)

andrewhsu:
talked with @simonferquel @silvin-lubecki @thaJeztah @tiborvass and will close this PR in favor of #1501 but keep the commit bd396cd in #1501 because this one is already LGTM'ed in order to finish up the cli ui work that builds on top of this.

Issue #522, opened 2017-09-11
Proposal: Consistent inspect/output from CLI


FrenchBen:
**Description**

When going through the different cli commands, the output tends to differ based on the command being used. 
Although this has been there for ever, it'd be great to standardize it and allow more flexibility. 

**Steps to reproduce the issue:**
1. Get docker up and running
2. Check `docker info` - notice the 'pretty print'. How do you get that in a JSON format?
3. Check `docker node inspect self` - notice the jSON output. But you do have the option to --pretty pring.

**Describe the results you received:**
All outputs are different based on the command that you're executing

**Describe the results you expected:**
I expect all of my outputs to be in JSON or pretty-print format. I also expect all commands to support the `--pretty` flag in order to get a cleaner output.
For example, inspecting a container, gives a JSON output, but there is no `--pretty` support:
```
$ docker container inspect 1c0 --pretty
unknown flag: --pretty
```

**Additional information you deem important (e.g. issue happens only occasionally):**
This has been in the CLI for ever and isn't a huge deal, more of an annoyance.

**Output of `docker version`:**
```
$ docker version
Client:
 Version:      17.09.0-ce-rc1
 API version:  1.31
 Go version:   go1.8.3
 Git commit:   ae21824
 Built:        Wed Sep  6 22:25:36 2017
 OS/Arch:      darwin/amd64

Server:
 Version:      17.09.0-ce-rc1
 API version:  1.32 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   ae21824
 Built:        Wed Sep  6 22:31:19 2017
 OS/Arch:      linux/amd64
 Experimental: true
```

thaJeztah:
I was discussing this recently with @riyazdf when he was working on his `docker trust` proposal; historically, the `inspect` commands have always showed the "low level" information of objects (in JSON format). There's definitely a need for commands that just pretty print the information in a more presentable way; this was attempted for `docker service inspect` (i.e. pretty-print by default), but that was not consistent with other inspect commands.

My suggestion is to add `view` subcommands for objects; `docker network view`, `docker container view` etc, which present the information in a human readable  format.

PR #1501, opened 2018-11-07
Fast context switch: commands

simonferquel:
Depends on #1500 
This brings a context subcommands to manage contexts stored in the context store, as well as modifying docker info / docker version to show current context

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -33,9 +33,6 @@ import (
 	"github.com/theupdateframework/notary/passphrase"
 )
 
-// ContextDockerHost is the reported context when DOCKER_HOST env var or -H flag is set
-const ContextDockerHost = "<DOCKER_HOST>"
-
 // Streams is an interface which exposes the standard input and output streams
 type Streams interface {
 	In() *InStream
@@ -62,6 +59,7 @@ type Cli interface {
 	ContextStore() store.Store
 	CurrentContext() string
 	StackOrchestrator(flagValue string) (Orchestrator, error)
+	DockerEndpoint() docker.Endpoint
 }
 
 // DockerCli is an instance the docker command line client.
@@ -78,6 +76,7 @@ type DockerCli struct {
 	newContainerizeClient func(string) (clitypes.ContainerizedClient, error)
 	contextStore          store.Store
 	currentContext        string
+	dockerEndpoint        docker.Endpoint
 }
 
 var storeConfig = store.NewConfig(
@@ -182,7 +181,7 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
 	var err error
 	cli.contextStore = store.New(cliconfig.ContextStoreDir(), storeConfig)
-	cli.currentContext, err = resolveContextName(opts.Common, cli.configFile)
+	cli.currentContext, err = resolveContextName(opts.Common, cli.configFile, cli.contextStore)
 	if err != nil {
 		return err
 	}
@@ -190,6 +189,7 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	if err != nil {
 		return errors.Wrap(err, "unable to resolve docker endpoint")
 	}
+	cli.dockerEndpoint = endpoint
 
 	cli.client, err = newAPIClientFromEndpoint(endpoint, cli.configFile)
 	if tlsconfig.IsErrEncryptedKey(err) {
@@ -223,7 +223,7 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 // NewAPIClientFromFlags creates a new APIClient from command line flags
 func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.ConfigFile) (client.APIClient, error) {
 	store := store.New(cliconfig.ContextStoreDir(), storeConfig)
-	contextName, err := resolveContextName(opts, configFile)
+	contextName, err := resolveContextName(opts, configFile, store)
 	if err != nil {
 		return nil, err
 	}
@@ -249,7 +249,7 @@ func newAPIClientFromEndpoint(ep docker.Endpoint, configFile *configfile.ConfigF
 }
 
 func resolveDockerEndpoint(s store.Store, contextName string, opts *cliflags.CommonOptions) (docker.Endpoint, error) {
-	if contextName != ContextDockerHost {
+	if contextName != "" {
 		ctxMeta, err := s.GetContextMetadata(contextName)
 		if err != nil {
 			return docker.Endpoint{}, err
@@ -258,7 +258,7 @@ func resolveDockerEndpoint(s store.Store, contextName string, opts *cliflags.Com
 		if err != nil {
 			return docker.Endpoint{}, err
 		}
-		return epMeta.WithTLSData(s, contextName)
+		return docker.WithTLSData(s, contextName, epMeta)
 	}
 	host, err := getServerHost(opts.Hosts, opts.TLSOptions)
 	if err != nil {
@@ -280,10 +280,8 @@ func resolveDockerEndpoint(s store.Store, contextName string, opts *cliflags.Com
 
 	return docker.Endpoint{
 		EndpointMeta: docker.EndpointMeta{
-			EndpointMetaBase: dcontext.EndpointMetaBase{
-				Host:          host,
-				SkipTLSVerify: skipTLSVerify,
-			},
+			Host:          host,
+			SkipTLSVerify: skipTLSVerify,
 		},
 		TLSData: tlsData,
 	}, nil
@@ -367,15 +365,16 @@ func (cli *DockerCli) StackOrchestrator(flagValue string) (Orchestrator, error) 
 	if currentContext == "" {
 		currentContext = configFile.CurrentContext
 	}
-	if currentContext == "" {
-		currentContext = ContextDockerHost
-	}
-	if currentContext != ContextDockerHost {
+	if currentContext != "" {
 		contextstore := cli.contextStore
 		if contextstore == nil {
 			contextstore = store.New(cliconfig.ContextStoreDir(), storeConfig)
 		}
 		ctxRaw, err := contextstore.GetContextMetadata(currentContext)
+		if store.IsErrContextDoesNotExist(err) {
+			// case where the currentContext has been removed (CLI behavior is to fallback to using DOCKER_HOST based resolution)
+			return GetStackOrchestrator(flagValue, "", configFile.StackOrchestrator, cli.Err())
+		}
 		if err != nil {
 			return "", err
 		}
@@ -387,6 +386,11 @@ func (cli *DockerCli) StackOrchestrator(flagValue string) (Orchestrator, error) 
 	}
 
 	return GetStackOrchestrator(flagValue, ctxOrchestrator, configFile.StackOrchestrator, cli.Err())
+}
+
+// DockerEndpoint returns the current docker endpoint
+func (cli *DockerCli) DockerEndpoint() docker.Endpoint {
+	return cli.dockerEndpoint
 }
 
 // ServerInfo stores details about the supported features and platform of the
@@ -435,24 +439,28 @@ func UserAgent() string {
 // - if DOCKER_CONTEXT is set, use this value
 // - if Config file has a globally set "CurrentContext", use this value
 // - fallbacks to default HOST, uses TLS config from flags/env vars
-func resolveContextName(opts *cliflags.CommonOptions, config *configfile.ConfigFile) (string, error) {
+func resolveContextName(opts *cliflags.CommonOptions, config *configfile.ConfigFile, contextstore store.Store) (string, error) {
 	if opts.Context != "" && len(opts.Hosts) > 0 {
-		return "", errors.New("Conflicting options: either specify --host or --context, not bot")
+		return "", errors.New("Conflicting options: either specify --host or --context, not both")
 	}
 	if opts.Context != "" {
 		return opts.Context, nil
 	}
 	if len(opts.Hosts) > 0 {
-		return ContextDockerHost, nil
+		return "", nil
 	}
 	if _, present := os.LookupEnv("DOCKER_HOST"); present {
-		return ContextDockerHost, nil
+		return "", nil
 	}
 	if ctxName, ok := os.LookupEnv("DOCKER_CONTEXT"); ok {
 		return ctxName, nil
 	}
 	if config != nil && config.CurrentContext != "" {
-		return config.CurrentContext, nil
+		_, err := contextstore.GetContextMetadata(config.CurrentContext)
+		if store.IsErrContextDoesNotExist(err) {
+			return "", errors.Errorf("Current context %q is not found on the file system, please check your config file at %s", config.CurrentContext, config.Filename)
+		}
+		return config.CurrentContext, err
 	}
-	return ContextDockerHost, nil
+	return "", nil
 }
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -3,26 +3,25 @@ package command
 import (
 	"context"
 	"io"
-	"net"
-	"net/http"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strconv"
-	"time"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/config"
 	cliconfig "github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/config/configfile"
-	"github.com/docker/cli/cli/connhelper"
+	dcontext "github.com/docker/cli/cli/context"
+	"github.com/docker/cli/cli/context/docker"
+	kubcontext "github.com/docker/cli/cli/context/kubernetes"
+	"github.com/docker/cli/cli/context/store"
 	cliflags "github.com/docker/cli/cli/flags"
 	manifeststore "github.com/docker/cli/cli/manifest/store"
 	registryclient "github.com/docker/cli/cli/registry/client"
 	"github.com/docker/cli/cli/trust"
 	dopts "github.com/docker/cli/opts"
 	clitypes "github.com/docker/cli/types"
-	"github.com/docker/docker/api"
 	"github.com/docker/docker/api/types"
 	registrytypes "github.com/docker/docker/api/types/registry"
 	"github.com/docker/docker/client"
@@ -33,6 +32,9 @@ import (
 	notaryclient "github.com/theupdateframework/notary/client"
 	"github.com/theupdateframework/notary/passphrase"
 )
+
+// ContextDockerHost is the reported context when DOCKER_HOST env var or -H flag is set
+const ContextDockerHost = "<DOCKER_HOST>"
 
 // Streams is an interface which exposes the standard input and output streams
 type Streams interface {
@@ -57,6 +59,9 @@ type Cli interface {
 	RegistryClient(bool) registryclient.RegistryClient
 	ContentTrustEnabled() bool
 	NewContainerizedEngineClient(sockPath string) (clitypes.ContainerizedClient, error)
+	ContextStore() store.Store
+	CurrentContext() string
+	StackOrchestrator(flagValue string) (Orchestrator, error)
 }
 
 // DockerCli is an instance the docker command line client.
@@ -71,7 +76,15 @@ type DockerCli struct {
 	clientInfo            ClientInfo
 	contentTrust          bool
 	newContainerizeClient func(string) (clitypes.ContainerizedClient, error)
+	contextStore          store.Store
+	currentContext        string
 }
+
+var storeConfig = store.NewConfig(
+	func() interface{} { return &DockerContext{} },
+	store.EndpointTypeGetter(docker.DockerEndpoint, func() interface{} { return &docker.EndpointMeta{} }),
+	store.EndpointTypeGetter(kubcontext.KubernetesEndpoint, func() interface{} { return &kubcontext.EndpointMeta{} }),
+)
 
 // DefaultVersion returns api.defaultVersion or DOCKER_API_VERSION if specified.
 func (cli *DockerCli) DefaultVersion() string {
@@ -167,14 +180,23 @@ func (cli *DockerCli) RegistryClient(allowInsecure bool) registryclient.Registry
 // line flags are parsed.
 func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
-
 	var err error
-	cli.client, err = NewAPIClientFromFlags(opts.Common, cli.configFile)
+	cli.contextStore = store.New(cliconfig.ContextStoreDir(), storeConfig)
+	cli.currentContext, err = resolveContextName(opts.Common, cli.configFile)
+	if err != nil {
+		return err
+	}
+	endpoint, err := resolveDockerEndpoint(cli.contextStore, cli.currentContext, opts.Common)
+	if err != nil {
+		return errors.Wrap(err, "unable to resolve docker endpoint")
+	}
+
+	cli.client, err = newAPIClientFromEndpoint(endpoint, cli.configFile)
 	if tlsconfig.IsErrEncryptedKey(err) {
 		passRetriever := passphrase.PromptRetrieverWithInOut(cli.In(), cli.Out(), nil)
 		newClient := func(password string) (client.APIClient, error) {
-			opts.Common.TLSOptions.Passphrase = password
-			return NewAPIClientFromFlags(opts.Common, cli.configFile)
+			endpoint.TLSPassword = password
+			return newAPIClientFromEndpoint(endpoint, cli.configFile)
 		}
 		cli.client, err = getClientWithPassword(passRetriever, newClient)
 	}
@@ -196,6 +218,75 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	}
 	cli.initializeFromClient()
 	return nil
+}
+
+// NewAPIClientFromFlags creates a new APIClient from command line flags
+func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.ConfigFile) (client.APIClient, error) {
+	store := store.New(cliconfig.ContextStoreDir(), storeConfig)
+	contextName, err := resolveContextName(opts, configFile)
+	if err != nil {
+		return nil, err
+	}
+	endpoint, err := resolveDockerEndpoint(store, contextName, opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to resolve docker endpoint")
+	}
+	return newAPIClientFromEndpoint(endpoint, configFile)
+}
+
+func newAPIClientFromEndpoint(ep docker.Endpoint, configFile *configfile.ConfigFile) (client.APIClient, error) {
+	clientOpts, err := ep.ClientOpts()
+	if err != nil {
+		return nil, err
+	}
+	customHeaders := configFile.HTTPHeaders
+	if customHeaders == nil {
+		customHeaders = map[string]string{}
+	}
+	customHeaders["User-Agent"] = UserAgent()
+	clientOpts = append(clientOpts, client.WithHTTPHeaders(customHeaders))
+	return client.NewClientWithOpts(clientOpts...)
+}
+
+func resolveDockerEndpoint(s store.Store, contextName string, opts *cliflags.CommonOptions) (docker.Endpoint, error) {
+	if contextName != ContextDockerHost {
+		ctxMeta, err := s.GetContextMetadata(contextName)
+		if err != nil {
+			return docker.Endpoint{}, err
+		}
+		epMeta, err := docker.EndpointFromContext(ctxMeta)
+		if err != nil {
+			return docker.Endpoint{}, err
+		}
+		return epMeta.WithTLSData(s, contextName)
+	}
+	host, err := getServerHost(opts.Hosts, opts.TLSOptions)
+	if err != nil {
+		return docker.Endpoint{}, err
+	}
+
+	var (
+		skipTLSVerify bool
+		tlsData       *dcontext.TLSData
+	)
+
+	if opts.TLSOptions != nil {
+		skipTLSVerify = opts.TLSOptions.InsecureSkipVerify
+		tlsData, err = dcontext.TLSDataFromFiles(opts.TLSOptions.CAFile, opts.TLSOptions.CertFile, opts.TLSOptions.KeyFile)
+		if err != nil {
+			return docker.Endpoint{}, err
+		}
+	}
+
+	return docker.Endpoint{
+		EndpointMeta: docker.EndpointMeta{
+			EndpointMetaBase: dcontext.EndpointMetaBase{
+				Host:          host,
+				SkipTLSVerify: skipTLSVerify,
+			},
+		},
+		TLSData: tlsData,
+	}, nil
 }
 
 func isEnabled(value string) (bool, error) {
@@ -253,6 +344,51 @@ func (cli *DockerCli) NewContainerizedEngineClient(sockPath string) (clitypes.Co
 	return cli.newContainerizeClient(sockPath)
 }
 
+// ContextStore returns the ContextStore
+func (cli *DockerCli) ContextStore() store.Store {
+	return cli.contextStore
+}
+
+// CurrentContext returns the current context name
+func (cli *DockerCli) CurrentContext() string {
+	return cli.currentContext
+}
+
+// StackOrchestrator resolves which stack orchestrator is in use
+func (cli *DockerCli) StackOrchestrator(flagValue string) (Orchestrator, error) {
+	var ctxOrchestrator string
+
+	configFile := cli.configFile
+	if configFile == nil {
+		configFile = cliconfig.LoadDefaultConfigFile(cli.Err())
+	}
+
+	currentContext := cli.CurrentContext()
+	if currentContext == "" {
+		currentContext = configFile.CurrentContext
+	}
+	if currentContext == "" {
+		currentContext = ContextDockerHost
+	}
+	if currentContext != ContextDockerHost {
+		contextstore := cli.contextStore
+		if contextstore == nil {
+			contextstore = store.New(cliconfig.ContextStoreDir(), storeConfig)
+		}
+		ctxRaw, err := contextstore.GetContextMetadata(currentContext)
+		if err != nil {
+			return "", err
+		}
+		ctxMeta, err := GetDockerContext(ctxRaw)
+		if err != nil {
+			return "", err
+		}
+		ctxOrchestrator = string(ctxMeta.StackOrchestrator)
+	}
+
+	return GetStackOrchestrator(flagValue, ctxOrchestrator, configFile.StackOrchestrator, cli.Err())
+}
+
 // ServerInfo stores details about the supported features and platform of the
 // server
 type ServerInfo struct {
@@ -272,51 +408,6 @@ func NewDockerCli(in io.ReadCloser, out, err io.Writer, isTrusted bool, containe
 	return &DockerCli{in: NewInStream(in), out: NewOutStream(out), err: err, contentTrust: isTrusted, newContainerizeClient: containerizedFn}
 }
 
-// NewAPIClientFromFlags creates a new APIClient from command line flags
-func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.ConfigFile) (client.APIClient, error) {
-	host, err := getServerHost(opts.Hosts, opts.TLSOptions)
-	if err != nil {
-		return &client.Client{}, err
-	}
-	var clientOpts []func(*client.Client) error
-	helper, err := connhelper.GetConnectionHelper(host)
-	if err != nil {
-		return &client.Client{}, err
-	}
-	if helper == nil {
-		clientOpts = append(clientOpts, withHTTPClient(opts.TLSOptions))
-		clientOpts = append(clientOpts, client.WithHost(host))
-	} else {
-		clientOpts = append(clientOpts, func(c *client.Client) error {
-			httpClient := &http.Client{
-				// No tls
-				// No proxy
-				Transport: &http.Transport{
-					DialContext: helper.Dialer,
-				},
-			}
-			return client.WithHTTPClient(httpClient)(c)
-		})
-		clientOpts = append(clientOpts, client.WithHost(helper.Host))
-		clientOpts = append(clientOpts, client.WithDialContext(helper.Dialer))
-	}
-
-	customHeaders := configFile.HTTPHeaders
-	if customHeaders == nil {
-		customHeaders = map[string]string{}
-	}
-	customHeaders["User-Agent"] = UserAgent()
-	clientOpts = append(clientOpts, client.WithHTTPHeaders(customHeaders))
-
-	verStr := api.DefaultVersion
-	if tmpStr := os.Getenv("DOCKER_API_VERSION"); tmpStr != "" {
-		verStr = tmpStr
-	}
-	clientOpts = append(clientOpts, client.WithVersion(verStr))
-
-	return client.NewClientWithOpts(clientOpts...)
-}
-
 func getServerHost(hosts []string, tlsOptions *tlsconfig.Options) (string, error) {
 	var host string
 	switch len(hosts) {
@@ -331,35 +422,37 @@ func getServerHost(hosts []string, tlsOptions *tlsconfig.Options) (string, error
 	return dopts.ParseHost(tlsOptions != nil, host)
 }
 
-func withHTTPClient(tlsOpts *tlsconfig.Options) func(*client.Client) error {
-	return func(c *client.Client) error {
-		if tlsOpts == nil {
-			// Use the default HTTPClient
-			return nil
-		}
-
-		opts := *tlsOpts
-		opts.ExclusiveRootPools = true
-		tlsConfig, err := tlsconfig.Client(opts)
-		if err != nil {
-			return err
-		}
-
-		httpClient := &http.Client{
-			Transport: &http.Transport{
-				TLSClientConfig: tlsConfig,
-				DialContext: (&net.Dialer{
-					KeepAlive: 30 * time.Second,
-					Timeout:   30 * time.Second,
-				}).DialContext,
-			},
-			CheckRedirect: client.CheckRedirect,
-		}
-		return client.WithHTTPClient(httpClient)(c)
-	}
-}
-
 // UserAgent returns the user agent string used for making API requests
 func UserAgent() string {
 	return "Docker-Client/" + cli.Version + " (" + runtime.GOOS + ")"
+}
+
+// resolveContextName resolves the current context name with the following rules:
+// - setting both --context and --host flags is ambiguous
+// - if --context is set, use this value
+// - if --host flag or DOCKER_HOST is set, fallbacks to use the same logic as before context-store was added
+// for backward compatibility with existing scripts
+// - if DOCKER_CONTEXT is set, use this value
+// - if Config file has a globally set "CurrentContext", use this value
+// - fallbacks to default HOST, uses TLS config from flags/env vars
+func resolveContextName(opts *cliflags.CommonOptions, config *configfile.ConfigFile) (string, error) {
+	if opts.Context != "" && len(opts.Hosts) > 0 {
+		return "", errors.New("Conflicting options: either specify --host or --context, not bot")
+	}
+	if opts.Context != "" {
+		return opts.Context, nil
+	}
+	if len(opts.Hosts) > 0 {
+		return ContextDockerHost, nil
+	}
+	if _, present := os.LookupEnv("DOCKER_HOST"); present {
+		return ContextDockerHost, nil
+	}
+	if ctxName, ok := os.LookupEnv("DOCKER_CONTEXT"); ok {
+		return ctxName, nil
+	}
+	if config != nil && config.CurrentContext != "" {
+		return config.CurrentContext, nil
+	}
+	return ContextDockerHost, nil
 }
--- a/cli/command/cli_test.go
+++ b/cli/command/cli_test.go
@@ -66,6 +66,7 @@ func TestNewAPIClientFromFlagsForDefaultSchema(t *testing.T) {
 func TestNewAPIClientFromFlagsWithAPIVersionFromEnv(t *testing.T) {
 	customVersion := "v3.3.3"
 	defer env.Patch(t, "DOCKER_API_VERSION", customVersion)()
+	defer env.Patch(t, "DOCKER_HOST", ":2375")()
 
 	opts := &flags.CommonOptions{}
 	configFile := &configfile.ConfigFile{}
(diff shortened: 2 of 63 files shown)


cli/command/context/create.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/docker/cli/cli/context/kubernetes"
+	"github.com/docker/cli/cli/context/store"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+type createOptions struct {
+	name                     string
+	description              string // --description string: set the description
```
silvin-lubecki: Did you forget to remove this comment ? 😅

cli/command/context/create.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	command.SetContextMetadata(&ctx, command.ContextMetadata{
+		Description:       o.description,
+		StackOrchestrator: stackOrchestrator,
+	})
+
+	return s.CreateOrUpdateContext(o.name, ctx)
+}
+
+func newCreateCommand(dockerCli command.Cli) *cobra.Command {
```
silvin-lubecki: Can you move this bloc up in the file?

cli/command/context/create.go
```diff
@@ -0,0 +1,222 @@
...
+	host          string //--moby-host string: set moby endpoint host (same format as DOCKER_HOST)
+	apiVersion    string
+	ca            string //--moby-ca string: path to a CA file
+	cert          string //--moby-cert string: path to a client cert file
+	key           string //--moby-key string: path to a client key file
+	skipTLSVerify bool
+}
+
+func (o *dockerEndpointOptions) addFlags(flags *pflag.FlagSet) {
+	flags.StringVar(&o.host, "docker-host", "", "required: specify the docker endpoint on wich to connect")
```
silvin-lubecki: I don't like required flags, I think it should be an argument if it's mandatory. cc @thaJeztah @vdemeester

cli/command/context/create.go
```diff
@@ -0,0 +1,222 @@
...
+	cert              string //--kubernetes-cert string: path to a client cert file
+	key               string //--kubernetes-key string: path to a client key file
+	skipTLSVerify     bool
+	defaultNamespace  string
+	kubeconfigFile    string //--kubernetes-kubeconfig-file string: path to a kubernetes cli config file
+	kubeconfigContext string //--kubernetes-kubeconfig-context string: name of the kubernetes cli config file context to use
+}
+
+func (o *kubernetedEndpointOptions) addFlags(flags *pflag.FlagSet) {
+	flags.StringVar(&o.server, "kubernetes-host", "", "required: specify the docker endpoint on wich to connect")
```
silvin-lubecki: Why is it required? 🤔

cli/command/context/create.go
simonferquel: I don't like commands with more than one unnamed argument (except when it is multiple arguments of the same nature like `docker image rm`)

cli/command/context/setkubernetes.go
```diff
@@ -0,0 +1,41 @@
...
+		Short: "Reset the kubernetes endpoint of a context",
+		Args:  cli.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			name := args[0]
+			endpoint, err := opts.toEndpoint(name)
+			if err != nil {
+				return errors.Wrap(err, "unable to create kubernetes endpoint config")
+			}
+			if endpoint == nil {
+				// remove
```
vdemeester: :thinking: ?

cli/command/context/setoptions.go
```diff
@@ -0,0 +1,46 @@
...
+
+type contextScopeOptions struct {
+	description              string
+	defaultStackOrchestrator string
+}
+
+func newSetOptionsCommand(dockerCli command.Cli) *cobra.Command {
+	o := &contextScopeOptions{}
+	cmd := &cobra.Command{
+		Use:   "set-options <name> [options]",
```
vdemeester: `set-options [OPTIONS] NAME`

vdemeester: Also, why not *only* `set` ? (as the "options" really are just flags)

cli/command/context/setoptions.go
```diff
@@ -0,0 +1,46 @@
...
+		Short: "set common options of a context",
+		Args:  cli.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			rawMeta, err := dockerCli.ContextStore().GetContextMetadata(args[0])
```
vdemeester: Same as above, maybe `runSet…` ?

cli/command/context/use.go
```diff
@@ -0,0 +1,23 @@
+package context
+
+import (
+	"github.com/docker/cli/cli/command"
+	"github.com/spf13/cobra"
+)
+
+func newUseCommand(dockerCli command.Cli) *cobra.Command {
+	cmd := &cobra.Command{
```
vdemeester: `use [OPTIONS] CONTEXT` (we may skip `OPTIONS` though)

silvin-lubecki:
Linter complains
```
cli/command/context/setoptions.go:44::warning: line is 213 characters (lll)
```

silvin-lubecki:
One test is failing:
```
--- FAIL: TestVersionAlign (0.00s)
    version_test.go:112: assertion failed: 
        --- expected
        +++ actual
        @@ -1,4 +1,4 @@
         Client:
        -·Context:···········test-context
        +·Context:·test-context
         ·Version:···········18.99.5-ce
         ·API·version:·······1.38
```

cli/command/context/create.go
simonferquel: Actually the flag is not required. If not set, we use the default host value (just like when running `docker ...` without -H.

andrewhsu:
i talked with @simonferquel @silvin-lubecki @thaJeztah about this and opened up an issue to describe the bigger context and track outstanding issues https://github.com/docker/cli/issues/1586

andrewhsu:
after syncing with @simonferquel and he will squash some of these commits after bd396cd in this PR to make it easier to review so that reviewers can see (by commit) the diff on top of the #1500 (commit bd396cd) that looks to already be LGTM'ed

will close #1500 to have this PR contain the unified foundation changes and ui changes so they are easier to evaluate together and merged together

simonferquel:
@vdemeester PTAL (I fixed your suggestions, and introduced tests to the commands)

cli/command/context/create.go
```diff
@@ -0,0 +1,105 @@
...
+			return runCreate(dockerCli, opts)
+		},
+	}
+	flags := cmd.Flags()
+	flags.StringVar(&opts.description, "description", "", "Description of the context")
+	flags.StringVar(
+		&opts.defaultStackOrchestrator,
+		"default-stack-orchestrator", "",
+		"Default orchestrator for stack operations to use with this context (swarm|kubernetes|all)")
+	flags.BoolVar(&opts.use, "use", false, "Use this context by default")
```
silvin-lubecki: What about `--use-as-default` ? Just `--use` does not feel really self explanatory...

cli/command/context/create.go
```diff
@@ -0,0 +1,105 @@
...
+	name                     string
+	description              string
+	defaultStackOrchestrator string
+	docker                   dockerEndpointOptions
+	kubernetes               kubernetesEndpointOptions
+	use                      bool
+}
+
+func runCreate(cli command.Cli, o *createOptions) error {
+	if o.name == "" {
```
silvin-lubecki: Can `name` be anything? No restriction on it? I see in this current implementation it ends as a directory name, but you mentioned maybe other implementations. I'm just wondering if it's shooting ourselves in the foot or not 🤔

cli/command/context/create_test.go
```diff
@@ -0,0 +1,230 @@
...
+	return m.in()
+}
+
+func (m *cliMock) Out() *command.OutStream {
+	return m.out()
+}
+
+func makeCliMock(t *testing.T, opts ...func(*cliMock)) (*cliMock, func(), error) {
+	dir, err := ioutil.TempDir("", t.Name())
+	if err != nil {
```
silvin-lubecki: Why taking `t` as an argument and return this err? You can assert directly on this err no? 🤔

cli/command/context/create_test.go
```diff
@@ -0,0 +1,230 @@
...
+func TestCreateExitingContext(t *testing.T) {
+	cli, cleanup, err := makeCliMock(t)
+	assert.NilError(t, err)
+	defer cleanup()
+	assert.NilError(t, cli.ContextStore().CreateOrUpdateContext("test", store.ContextMetadata{}))
+
+	err = runCreate(cli, &createOptions{
+		name: "test",
+	})
+	assert.ErrorContains(t, err, "context \"test\" already exists")
```
silvin-lubecki: nit: \`context "test" already exists`

cli/command/context/create_test.go
```diff
@@ -0,0 +1,230 @@
...
+func TestCreateInvalidOrchestrator(t *testing.T) {
+	cli, cleanup, err := makeCliMock(t)
+	assert.NilError(t, err)
+	defer cleanup()
+
+	err = runCreate(cli, &createOptions{
+		name:                     "test",
+		defaultStackOrchestrator: "invalid",
+	})
+	assert.ErrorContains(t, err, "specified orchestrator \"invalid\" is invalid, please use either kubernetes, swarm or all")
```
silvin-lubecki: nit: use backquote notations

cli/command/context/create_test.go
```diff
@@ -0,0 +1,230 @@
...
+func TestCreateOrchestratorKubernetesNoEndpoint(t *testing.T) {
+	cli, cleanup, err := makeCliMock(t)
+	assert.NilError(t, err)
+	defer cleanup()
+
+	err = runCreate(cli, &createOptions{
+		name:                     "test",
+		defaultStackOrchestrator: "kubernetes",
+	})
+	assert.ErrorContains(t, err, "cannot specify orchestrator \"kubernetes\" without configuring a Kubernetes endpoint")
```
silvin-lubecki: nit: use backquote notation

cli/command/context/create_test.go
```diff
@@ -0,0 +1,230 @@
...
+}
+func TestCreateOrchestratorAllNoKubernetesEndpoint(t *testing.T) {
```
silvin-lubecki: nit: empty line

cli/command/context/create_test.go
```diff
@@ -0,0 +1,230 @@
...
+	cli, cleanup, err := makeCliMock(t)
+	assert.NilError(t, err)
+	defer cleanup()
+
+	err = runCreate(cli, &createOptions{
+		name:                     "test",
+		defaultStackOrchestrator: "all",
+	})
+	assert.ErrorContains(t, err, "cannot specify orchestrator \"all\" without configuring a Kubernetes endpoint")
```
silvin-lubecki: nit: use backquote notation

cli/command/context/create_test.go
silvin-lubecki: I guess this test is a little too much, as it only tests that `all` value `HasKubernetes`.

cli/command/context/export-import_test.go
```diff
@@ -0,0 +1,99 @@
...
+	contextFile := filepath.Join(contextDir, "exported")
+	cli, cleanup, err := makeCliMock(t)
+	assert.NilError(t, err)
+	defer cleanup()
+	createTestContextWithKube(t, cli)
+	assert.NilError(t, runExport(cli, &exportOptions{
+		contextName: "test",
+		dest:        contextFile,
+	}))
+	assert.NilError(t, runImport(cli, importOptions{
```
silvin-lubecki: You check here that import does not fail... But you don't check the result. It may succeed but without importing anything. Same for the export, maybe you can golden the export?

cli/command/context/export.go
```diff
@@ -0,0 +1,92 @@
...
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.BoolVarP(&opts.kubeconfig, "kubeconfig", "k", false, "Export as a kubeconfig file")
+	return cmd
+}
+
+func writeTo(dockerCli command.Cli, reader io.Reader, dest string) error {
+	if dest == "-" {
```
silvin-lubecki: nit:
```go
var writer io.Writer
if dest == - {
    writer = dockerCli.Out()
} else{
   f, err := os.OpenFile(dest, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
   if err != nil {
       return err
   }
   defer f.Close()
   writer = f    
}
_, err = io.Copy(f, reader)
return err
```

cli/command/context/create.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/docker/cli/cli/context/kubernetes"
+	"github.com/docker/cli/cli/context/store"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func newCreateCommand(dockerCli command.Cli) *cobra.Command {
+	opts := &createOptions{}
+	cmd := &cobra.Command{
+		Use:   "create [OPTIONS] NAME",
```
silvin-lubecki: `CONTEXT` instead of `NAME`

cli/command/context/import.go
```diff
@@ -0,0 +1,69 @@
...
+type importOptions struct {
+	use   bool
+	force bool
+	name  string
+}
+
+func newImportCommand(dockerCli command.Cli) *cobra.Command {
+	opts := importOptions{}
+	cmd := &cobra.Command{
+		Use:   "import [OPTIONS] FILE",
```
silvin-lubecki: This should be the mirror of the export command `import [OPTIONS] FILE [CONTEXT]`

cli/command/context/import.go
```diff
@@ -0,0 +1,69 @@
...
+		Args:  cli.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runImport(dockerCli, opts, args[0])
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.BoolVar(&opts.use, "use", false, "Use this context by default")
+	flags.BoolVar(&opts.force, "force", false, "Overwrite any existing context with the same name")
+	flags.StringVar(&opts.name, "name", "", "Name of the context")
```
silvin-lubecki: This should be a 2nd argument instead

cli/command/context/list.go
```diff
@@ -0,0 +1,83 @@
+package context
+
+import (
+	"fmt"
+	"sort"
+
+	"vbom.ml/util/sortorder"
+
```
silvin-lubecki: empty line

cli/command/context/list_test.go
```diff
@@ -0,0 +1,59 @@
...
+	assert.NilError(t, err)
+	defer cleanup()
+	createTestContextWithKubeAndSwarm(t, cli, "current")
+	createTestContextWithKubeAndSwarm(t, cli, "other")
+	cli.currentContext = "current"
+	assert.NilError(t, runList(cli, &listOptions{
+		format: formatter.TableFormatKey,
+		quiet:  true,
+	}))
+	assert.Equal(t, "current\nother\n", buf.String())
```
silvin-lubecki: You can also use a golden here 😄

cli/command/context/remove.go
```diff
@@ -0,0 +1,62 @@
...
+			return runRemove(dockerCli, opts, args)
+		},
+	}
+	cmd.Flags().BoolVarP(&opts.force, "force", "f", false, "Force the removal of a context in use")
+	return cmd
+}
+
+func runRemove(dockerCli command.Cli, opts removeOptions, names []string) error {
+	var errs []string
+	currentCtx := strings.ToLower(dockerCli.CurrentContext())
```
silvin-lubecki: Why this ToLower? Do you really want to be able to remove without case sensitive? Note that the `create` command is case sensitive. Well to be fair, it will depend the os (case insensitive on windows). But my point is that it should be homogeneous on every command.

silvin-lubecki: Ok my bad it's only for current context... But well the comment is still valid I guess?

cli/command/context/remove_test.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func TestRemoveNotAContext(t *testing.T) {
+	cli, cleanup, err := makeCliMock(t)
+	assert.NilError(t, err)
+	defer cleanup()
+	createTestContextWithKubeAndSwarm(t, cli, "current")
+	createTestContextWithKubeAndSwarm(t, cli, "other")
+	err = runRemove(cli, removeOptions{}, []string{"not-a-context"})
+	assert.ErrorContains(t, err, "context \"not-a-context\" does not exist")
```
silvin-lubecki: use backquote notation

cli/command/context/use.go
```diff
@@ -0,0 +1,25 @@
...
+		Use:     "use CONTEXT",
+		Aliases: []string{"select", "switch"},
```
silvin-lubecki: I think these aliases are a little bit too much, why don't you just provide `--use` for this version?

cli/command/context/create_test.go
simonferquel: I wanted to make sure we use HasKubernetes in the callee and not just an equality comparison with orchestratorKubernetes

cli/command/context/create.go
simonferquel: It will must be a valid directory name. I let the OS raise an error on directory creation instead of trying to enforce that with some difficult to maintain OS-specific code.

cli/command/context/export-import_test.go
simonferquel: No need for a golden: I deepEqual the exported and imported context (in following commit)

cli/command/context/import.go
simonferquel: I really hate commands with 2 unnamed args of same type (string) but completely different meaning. I can never remember the order of args (e.g. I never ever write `ln -s` in the correct way on the first try), that is why I originally wrote it like this. But I understand the need for mirroring.
My preferred alternate would be `docker context export CONTEXT --to FILE` and `docker context import CONTEXT --from FILE` but that does not feel aligned with the rest of the cli design.

cli/command/context/import.go
simonferquel: I made it first, for having it in the same order as export

cli/command/context/list_test.go
simonferquel: really, for 2 lines ?

cli/command/context/remove.go
simonferquel: Good catch, there is inconsistency here. I'll introduce a `normalizeName` func and use it on all subcommands.

cli/command/context/list_test.go
silvin-lubecki: https://github.com/docker/cli/blob/master/cli/command/service/testdata/service-list-sort.golden
Ok there are 3 lines, but still... 😄

cli/command/cli_test.go
```diff
@@ -66,6 +66,7 @@ func TestNewAPIClientFromFlagsForDefaultSchema(t *testing.T) {
 func TestNewAPIClientFromFlagsWithAPIVersionFromEnv(t *testing.T) {
 	customVersion := "v3.3.3"
 	defer env.Patch(t, "DOCKER_API_VERSION", customVersion)()
+	defer env.Patch(t, "DOCKER_HOST", ":2375")()
```
vdemeester: nit: shouldn't this be in another test (subtest ? :upside_down_face:)

cli/command/context/create_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/cli/cli/config/configfile"
+	"github.com/docker/cli/cli/context/docker"
+	"github.com/docker/cli/cli/context/kubernetes"
+	"github.com/docker/cli/cli/context/store"
+	"gotest.tools/assert"
+	"gotest.tools/env"
+)
+
+type cliMock struct {
```
vdemeester: :scream: 
s/Mock/Fake/

vdemeester: On that note, we *do* have `FakeCli` in `internal/test`, why not enhance it ?

cli/command/context/list_test.go
vdemeester: @simonferquel we have some golden file with 2 lines :angel: :trollface:

cli/command/cli_test.go
simonferquel: This is required to make sure the initializer goes trough the "don't use context store" code path. (I consider those tests as non-regression tests for existing scripts)

simonferquel:
Thanks @thaJeztah for spotting all this.
I have a few comments to do, and an updated todo-list:
## Comments:
- About the single `docker context update` command:
  - this has the same drawback as the `docker context create` one: so many flags to set.
  - it has an ambiguity with what should happen if I don't set one of a particular endpoint flag (e.g.: docker-ca)
  - but if we move as you suggest to a csv format for endpoint configurations, it should work: 
    - `docker context update --description="something`: no endpoint flag has been explicitly set, we don't touch them
    - `docker context update --docker host=https://whatever,ca=/path/to/ca --kubernetes=""`: docker endpoint flag value should be complete (we don't try to merge special endpoint config), kubernetes endpoint is explicitly removed
- About `docker-from-current/kubernetes-from-current` in import
  - I don't like it as I see the import/export function to be kind of the counterpart of `image save/load` as of saving/loading to/from files
  - It brings a bit of confusion about which command I should use to create a context
  - what if I want to create a context with the current kubernetes configuration but a different docker-host ?
- Empty context:
  - I see at least one useful case: Docker for Windows & invocation image based workflows (like docker/app): 
  - an empty context means "whatever default host endpoint is on the current client platform" which leads to having it work from both the host (resolving to npipe:////./pipe/docker_engine) and a Linux container (resolving to /var/run/docker.sock that can be mounted). For scenarios like docker/app where we want to share contexts from the host to an invocation container, that is really useful

## My current todolist:
- [ ] Remove the <DOCKER_HOST> constant
- [ ] When not on a context, make DOCKER_HOST based config appear in docker context ls
- [ ] Remove Context from `docker version`, add it in its own section in `docker info` (which I think makes more sense)
- [ ] Introduce `docker context show` showing the current context details (--quiet for just name)
- [ ] Introduce `docker context inspect`
- [ ] Introduce internal id (sha256 of the name) in the store to avoid problems with strange names. 
- [ ] Add flag for quoting on `docker context ls -q --quote-names`
- [ ] Hide "unset" stack orchestrator in ls
- [ ] Remove --docker-api-version (but honor the DOCKER_API_VERSION env var for backward compat with existing scripts)
- [ ] Remove --use-as-default in import and create
- [ ] `docker context use` prints the context name on stdout, confirmation on stderr
- [ ] `docker context rm` prints the context names on stdout
- [ ] Specify endpoints configuration as csv (and include a long description of the flags in create/update)
- [ ] Merge various set commands into a single `update` subcommand

thaJeztah:
> It brings a bit of confusion about which command I should use to create a context
> what if I want to create a context with the current kubernetes configuration but a different docker-host ?

Let me give those some thought; my train of thought was that `import` means "load/import from something existing", wheras `create` means "create something from scratch".


> For scenarios like docker/app where we want to share contexts from the host to an invocation container, that is really useful

Isn't that the default behavior though (i.e., if no context is set)? i.e., what's the added value of the "empty" context in that case? (Trying to understand :thinking_face:)


> Introduce `docker context show` showing the current context details (--quiet for just name)

If we add `docker context inspect`, I don't think we need this at the start. We could add a `--pretty`  (and/or `--format=pretty`) flag to `docker context inspect`. If we have `docker context inspect`, the name could be obtained using `docker context inspect --format '{{.Name}}'`, which would address the use-case.


Having a `show` command then would be a "nice to have", and may be something to do in a bigger change for all object-types; see https://github.com/docker/cli/issues/522, which discusses adding "human friendly" view/show/print commands for all object types.


> Introduce internal id (sha256 of the name)

If we want to address the future possibility of having a `docker context rename`, we should use a GUID / random value for this, otherwise renaming would also change the sha256, thus the ID.

> Add flag for quoting on `docker context ls -q --quote-names`

Could you explain this one? Is this so that names are still "unrestricted" (as in: any name and/or character/emoji is accepted?)

I think we should apply the same naming rules as we have on other objects (`[a-zA-Z0-9][a-zA-Z0-9_.-]`);

    docker create --name "hello world" busybox
    Error response from daemon: Invalid container name (hello world), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed

Or if we anticipate these ever gonna be used as part of an URL; a valid [path-component](https://github.com/docker/distribution/blob/3da015f8aaa8653690917031108bd45a2bed5cfa/reference/reference.go#L11) ([docker/distribution](https://github.com/docker/distribution/blob/3da015f8aaa8653690917031108bd45a2bed5cfa/reference/regexp.go#L5-L90)).

Having stricter names also makes sure that names are portable between platforms, and should provide enough options to express names; and there's a "Description" field that would allow describing other information about a context.


> Hide "unset" stack orchestrator in ls

I think in https://github.com/docker/cli/pull/1500#discussion_r242193859 we came to the conclusion that we could remove the `"unset"` value (@silvin-lubecki ?), so Ideally it would not be just a change in the "presentation", but the actual use of `"unset"` to be removed.


> Remove `--docker-api-version` (but honor the `DOCKER_API_VERSION` env var for backward compat with existing scripts)

Yes, that sounds correct; in general we should discourage the use of `DOCKER_API_VERSION`, because it was intended as a debugging tool, and no validation is performed whatsoever on its value.

Which allows fun things, such as inspecting images through `docker container inspect` :joy:

```bash
$ docker tag hello-world containers/hello-world
$ DOCKER_API_VERSION='blabla/../images/' docker container inspect --format '{{ .RepoTags }}' hello-world
[containers/hello-world:latest hello-world:latest]
```

> Specify endpoints configuration as csv (and include a long description of the flags in create/update)

Thanks; I can imagine some of that needing more eyes (e.g. I can see being able to update just "host" without changing other parts being a possible use-case)

simonferquel:
The problem I see with Guids, is that we would need somehow maintain a map of names to ids for fast lookups (whereas sha256 of the name does not require that). Also, renaming of a context can be done by renaming 2 directories.

cli/command/context/cmd.go
```diff
@@ -28,6 +29,13 @@ func NewContextCommand(dockerCli command.Cli) *cobra.Command {
 	return cmd
 }
 
-func normalizeContextName(name string) string {
-	return strings.ToLower(name)
+const restrictedNamePattern = "^[a-zA-Z0-9][a-zA-Z0-9_.+-]+$"
```
thaJeztah: Where did the `+` come from?

Note that this pattern still allows names such as `foo......bar` or `foo.....`; this is currently the case for containers as well, and may not be a concern; just calling it out here in case we want to be more conservative (e.g., only accept `_`, `.`, -` as a separator between alphanumeric characters, and only a single consecutive separator

simonferquel: I added the +, because as context are the aggregation of swarm and kubernetes endpoint configurations, I'd personnally like to name my context like `local-swarm+aks-west-eu`. Also, with the dir names being now the sha256 of the context names, I have no problem if a user might think that `fo.....` is a valuable context name

cli/context/store/metadata_test.go
```diff
@@ -77,16 +82,14 @@ func TestMetadataList(t *testing.T) {
 	assert.NilError(t, err)
 	defer os.RemoveAll(testDir)
 	testee := metadataStore{root: testDir, config: testCfg}
-	wholeData := map[string]ContextMetadata{
-		"simple":                    testMetadata,
-		"simple2":                   testMetadata,
-		"nested/context":            testMetadata,
```
silvin-lubecki: Is the nested feature dropped ?

cli/context/store/store.go
```diff
@@ -2,9 +2,11 @@ package store
 
 import (
 	"archive/tar"
+	_ "crypto/sha256" // ensure ids can be computed
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/opencontainers/go-digest"
```
silvin-lubecki: nit: re-order imports

cli/command/context/list.go
```diff
@@ -60,13 +60,13 @@ func runList(dockerCli command.Cli, opts *listOptions) error {
 			Name:               rawMeta.Name,
 			Current:            rawMeta.Name == curContext,
 			Description:        meta.Description,
-			StackOrchestrator:  string(meta.StackOrchestrator),
+			StackOrchestrator:  hideUnset(meta.StackOrchestrator),
```
silvin-lubecki: As @thaJeztah said, I think we should remove this "unset" value, instead of trying to hide it.

cli/command/context/update_test.go
```diff
@@ -0,0 +1,78 @@
...
+	assert.NilError(t, err)
+	dc, err := command.GetDockerContext(c)
+	assert.NilError(t, err)
+	assert.Equal(t, dc.StackOrchestrator, command.OrchestratorSwarm)
+	assert.Equal(t, dc.Description, "description of test")
+	assert.Check(t, cmp.Contains(c.Endpoints, kubernetes.KubernetesEndpoint))
+	assert.Check(t, cmp.Contains(c.Endpoints, docker.DockerEndpoint))
+	assert.Equal(t, c.Endpoints[docker.DockerEndpoint].(docker.EndpointMeta).Host, "tcp://some-host")
+}
+func TestUpdateStackOrchestratorStrategy(t *testing.T) {
```
silvin-lubecki: nit: empty line

cli/command/context/update_test.go
```diff
@@ -0,0 +1,78 @@
+package context
+
+import (
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/context/docker"
+	"github.com/docker/cli/cli/context/kubernetes"
+	"gotest.tools/assert"
+	"gotest.tools/assert/cmp"
+	"testing"
```
silvin-lubecki: nit: re-order imports

cli/command/context/create.go
```diff
@@ -128,3 +122,16 @@ func runCreate(cli command.Cli, o *createOptions) error {
 	}
 	return nil
 }
+
+func ensureNameIsValidAndContextDoesNotExist(s store.Store, name string) error {
```
silvin-lubecki: nit: Rename to `checkContextName` ?

cli/context/store/metadata_test.go
simonferquel: With the id derived from the name, nested contexts are actually stored flat.

cli/command/context/list.go
simonferquel: I think we should do that in a followup

simonferquel:
@thaJeztah I think I now have a design much closer to where you want it to go. PTAL

silvin-lubecki:
Still LGTM 😄

cli/command/context/import.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/context/store"
+	"github.com/spf13/cobra"
+)
+
+func newImportCommand(dockerCli command.Cli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "import [OPTIONS] CONTEXT FILE|-",
```
thaJeztah: Looks like it has no options;

```suggestion
		Use:   "import CONTEXT FILE|-",
```

thaJeztah:
Summary;

- [ ] not a blocker, but needs to be discussed; the "DOCKER_HOST" context is a bit confusing (no name, and disappears once a context is created)
- [ ] validate `host` value to prevent a bad UX if it's not valid
  - [ ] figure out if we can prevent the CLI from attempting to connect to the daemon on `--help` and such (it may be used for hiding options?)
- [ ] provide user-feedback on succesful `create`, `update`, `import`, `export`
- [ ] decide on the `ID` / "name hashed used as ID"
- [ ] decide on the UX for export
  - print where it's saved
  - or; export to `stdout` by default (consistent with `docker export` / `docker save`)
  - or; make `FILE` a required option (clear what's happening: downside: no "auto-naming")
  - I'm a bit on the fence on `-` for `stdout` (looking a bit, there doesn't seem to be a clear standard; `tar` uses `-O` / `--to-stdout`, but `zip` also uses `-` :thinking_face:)
- [ ] produce error when attempting to overwrite an existing export
- [ ] detect terminal when exporting to `stdout`
- [ ] (question) do we want `import` to be symmetrical with `export`, and add `--kubeconfig` to import from a kubeconfig?


FWIW; being able to switch between hosts this easily was pretty cool;

```
docker context ls
NAME                DESCRIPTION                          DOCKER ENDPOINT               KUBERNETES ENDPOINT   ORCHESTRATOR
local               Docker Desktop for Mac (local)       unix:///var/run/docker.sock                         swarm
swarm-test-01 *     A remote Swarm cluster for testing   ssh://swarm-test-01                                 
swarm-test-02       A remote Swarm cluster for testing   ssh://swarm-test-02                                 swarm
swarm-test-03       A remote Swarm cluster for testing   ssh://swarm-test-03                                 swarm
swarm-test-04       A remote Swarm cluster for testing   ssh://swarm-test-04                                 swarm
```

I wonder if we could somehow integrate with `docker node` :thinking_face: (i.e.; import all nodes as contexts)

ijc:
> I'm a bit on the fence on - for stdout (looking a bit, there doesn't seem to be a clear standard; tar uses -O / --to-stdout, but zip also uses - :thinking_face:)

`-` meaning `stdout` is very common in POSIX/Unixy tools.

`tar -f -` reads/writes the tarball from/to stdin/stdout. The `-O` option is for sending the contents of an extracted file to stdout so e.g. `tar xOf - ./foo/bar` will read a tarball from stdin, and print the content of the tarfile member `./foo/bar` to stdout.

thaJeztah:
Thanks @ijc It's probably me then (more familiar with it being used for `stdin`) I knew there's been discussions in the past about what to use, but those were "after the fact" (adding it to an existing command), so were a bit hairy 😓

simonferquel:
For the `not-a-context` context name, I suggest that for printing, we could re-introduce the `<DOCKER_HOST>` name (which is now invalid as a regular context name).

For the import/export to stdin/out by default, the idea was to have the default behavior creating a file with a known extension, so that Docker Desktop could associate them and provide a nice import GUI. I understand it brings inconsistency with the current docker image save/load experience though. With this context (huhu!), what do you thing about it ?

For the import from kubeconfig, the problem is that you can't create a context without configuring a docker endpoint. So the counterpart is a `docker context create <name> --docker <...> --kubernetes config-file=./my-kubeconfig`. It lacks a bit of symetry though. Maybe should I remove the --kubeconfig flag and make that feature into a separate subcommand?

For the ID, I really don't want to introduce an index of names to id, and this SHA based stuff seems good to me. However, I think that I should not call this an ID, because it is actually a SHA of the real identifier which is the name (and thus, it is confusing). What about naming it context_dir, and in the inspect function, I would report a metadata_dir and tls_dir with the path to the related storage ?

thaJeztah:
> For the not-a-context context name, I suggest that for printing, we could re-introduce the `<DOCKER_HOST>` name (which is now invalid as a regular context name).

Also thinking that we need a way to switch _back_ to the default/docker_host/unset situation (not a big fan of `<DOCKER_HOST>` as a name; it's very "shouty"); I'd be ok with a reserved name though, perhaps `"default"`, `"none"` or `"unset"` (😅 I know) ? That way, I could switch between `context X` and back to the default situation.

> For the import from kubeconfig, the problem is that you can't create a context without configuring a docker endpoint

Ah, right, gotcha

> Maybe should I remove the --kubeconfig flag and make that feature into a separate subcommand?

So, if we can't import the exported `--kubeconfig`, what is the exact use-case? If the "regular" export contains the full configuration, is there a need to export just a subset of that information in a format that cannot be imported by the `docker cli`?


> For the ID, I really don't want to introduce an index of names to id, and this SHA based stuff seems good to me. However, I think that I should not call this an ID, because it is actually a SHA of the real identifier which is the name (and thus, it is confusing)

Right, so I'm wondering what we gained by introducing it (also see https://github.com/docker/cli/pull/1501#issuecomment-449728980), as the original reason for it was to get a) consistency (have an `ID` as unique identifier and b) possibly allow name to be optional and/or allow renaming.

The only thing it brought is a "safe" name for the path (no chance on path-traversing and such), although we still use the name as default for `export`

>  What about naming it context_dir, and in the inspect function, I would report a metadata_dir and tls_dir with the path to the related storage ?

Perhaps `xx_path` (or CamelCase `..Path`), as that's used elsewhere (see `docker container inspect`)

Are `metadata` and `tls` always relative to that path? If so, perhaps there's no need to include them separately

simonferquel:
> > For the not-a-context context name, I suggest that for printing, we could re-introduce the `<DOCKER_HOST>` name (which is now invalid as a regular context name).
> 
> Also thinking that we need a way to switch _back_ to the default/docker_host/unset situation (not a big fan of `<DOCKER_HOST>` as a name; it's very "shouty"); I'd be ok with a reserved name though, perhaps `"default"`, `"none"` or `"unset"` (😅 I know) ? That way, I could switch between `context X` and back to the default situation.

I will go for `default`, as I don't like the idea of `docker context ls` printing a context named `none`.

> 
> > For the import from kubeconfig, the problem is that you can't create a context without configuring a docker endpoint
> 
> Ah, right, gotcha
> 
> > Maybe should I remove the --kubeconfig flag and make that feature into a separate subcommand?
> 
> So, if we can't import the exported `--kubeconfig`, what is the exact use-case? If the "regular" export contains the full configuration, is there a need to export just a subset of that information in a format that cannot be imported by the `docker cli`?
The idea is to be able to use it with other tools like kubectl etc.
> 
> > For the ID, I really don't want to introduce an index of names to id, and this SHA based stuff seems good to me. However, I think that I should not call this an ID, because it is actually a SHA of the real identifier which is the name (and thus, it is confusing)
> 
> Right, so I'm wondering what we gained by introducing it (also see [#1501 (comment)](https://github.com/docker/cli/pull/1501#issuecomment-449728980)), as the original reason for it was to get a) consistency (have an `ID` as unique identifier and b) possibly allow name to be optional and/or allow renaming.
> 
> The only thing it brought is a "safe" name for the path (no chance on path-traversing and such), although we still use the name as default for `export`
> 
> > What about naming it context_dir, and in the inspect function, I would report a metadata_dir and tls_dir with the path to the related storage ?
> 
> Perhaps `xx_path` (or CamelCase `..Path`), as that's used elsewhere (see `docker container inspect`)
> 
> Are `metadata` and `tls` always relative to that path? If so, perhaps there's no need to include them separately

ok for _path suffixed fields

cli/command/cli.go
```diff
@@ -331,35 +431,42 @@ func getServerHost(hosts []string, tlsOptions *tlsconfig.Options) (string, error
...
+	if _, present := os.LookupEnv("DOCKER_HOST"); present {
+		return "", nil
+	}
+	if ctxName, ok := os.LookupEnv("DOCKER_CONTEXT"); ok {
+		return ctxName, nil
+	}
+	if config != nil && config.CurrentContext != "" {
+		_, err := contextstore.GetContextMetadata(config.CurrentContext)
+		if store.IsErrContextDoesNotExist(err) {
+			warnCurrentContextNotFound(config.CurrentContext, stderr)
```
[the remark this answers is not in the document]
simonferquel: Erf, you are right, but this one will be tricky: docker context use implicitly requires an initialized dockerCli. 
As this error can occur only if the user messes directly with the store dir, is it ok if I do a hard fail, indicating that he should fix his configfile manually?

cli/command/context/list_test.go
```diff
@@ -0,0 +1,98 @@
...
+	cli.SetCurrentContext("current")
+	cli.OutBuffer().Reset()
+	assert.NilError(t, runList(cli, &listOptions{
+		format: formatter.TableFormatKey,
+		quiet:  true,
+	}))
+	golden.Assert(t, cli.OutBuffer().String(), "quiet-list.golden")
+}
+
+func TestInspect(t *testing.T) {
```
thaJeztah: Looks like this one is in the wrong file (should probably be in `inspect_test.go`)

cli/command/context/list.go
```diff
@@ -0,0 +1,100 @@
...
+		Aliases: []string{"list"},
+		Short:   "List contexts",
+		Args:    cli.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runList(dockerCli, opts)
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringVar(&opts.format, "format", formatter.TableFormatKey, "Pretty-print contexts using a Go template")
```
thaJeztah: nit (for a follow-up); for other commands we use `""` as default, and in the code use something like;

```
	format := opts.format
	if len(format) == 0 {
		format = formatter.TableFormatKey
	}
```

e.g. https://github.com/docker/cli/blob/a900ba8aeff764a5fb5bb7e18eee25d75a220732/cli/command/system/df.go#L45-L48


We can do a follow-up to decide which approach to take for all (either way makes some sense)

cli/command/context/list_test.go
```diff
@@ -0,0 +1,98 @@
...
+	createTestContextWithKubeAndSwarm(t, cli, "other", "all")
+	cli.SetCurrentContext("current")
+	cli.OutBuffer().Reset()
+	assert.NilError(t, runList(cli, &listOptions{
+		format: formatter.TableFormatKey,
+	}))
+	golden.Assert(t, cli.OutBuffer().String(), "list.unset.golden")
+}
+
+func TestListNoContext(t *testing.T) {
```
thaJeztah: I noticed the `default` context still disappears after selecting a context. Can we make the `default` context always visible in `docker context ls`?

```bash
docker context ls

NAME                DESCRIPTION                               DOCKER ENDPOINT               KUBERNETES ENDPOINT   ORCHESTRATOR
default *           Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                         swarm
foobar                                                        unix:///var/run/docker.sock                         


docker context use foobar
foobar
Current context is now "foobar"

docker context ls
NAME                DESCRIPTION         DOCKER ENDPOINT               KUBERNETES ENDPOINT   ORCHESTRATOR
foobar *                                unix:///var/run/docker.sock                         


docker context use default
default

NAME                DESCRIPTION                               DOCKER ENDPOINT               KUBERNETES ENDPOINT   ORCHESTRATOR
default *           Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                         swarm
foobar                                                        unix:///var/run/docker.sock                         
```

cli/command/context/cmd.go
```diff
@@ -0,0 +1,49 @@
...
+
+var restrictedNameRegEx = regexp.MustCompile(restrictedNamePattern)
+
+func validateContextName(name string) error {
+	if name == "" {
+		return errors.New("context name cannot be empty")
+	}
+	if name == "default" {
+		return errors.New(`"default" is a reserved context name`)
```
thaJeztah: nit (ok for a follow-up): we should probably handle this with for other situations as well. Perhaps when using inspect; return the information that we obtained from the default (DOCKER_HOST). Currently;

```bash
docker context inspect default
[]
context "default" does not exist

docker context rm default
default: context "default" does not exist
```

cli/context/store/store.go
```diff
@@ -0,0 +1,334 @@
...
+func (e *tlsDataDoesNotExistError) Error() string {
+	return fmt.Sprintf("tls data for %s/%s/%s does not exist", e.context, e.endpoint, e.file)
+}
+
+func (e *tlsDataDoesNotExistError) setContext(name string) {
+	e.context = name
+}
+
+// IsErrContextDoesNotExist checks if the given error is a "context does not exist" condition
+func IsErrContextDoesNotExist(err error) bool {
```
thaJeztah: Also for a follow-up; use an interface for these (like the `errdefs` package (or use that package); https://github.com/moby/moby/blob/d48392a35b157612d28c48eb8e4fad1272fa1442/errdefs/defs.go#L3-L6

```go
// ErrNotFound signals that the requested object doesn't exist
type ErrNotFound interface {
	NotFound()
}
``` 

https://github.com/moby/moby/blob/d48392a35b157612d28c48eb8e4fad1272fa1442/errdefs/is.go#L32-L36

```go
// IsNotFound returns if the passed in error is an ErrNotFound
func IsNotFound(err error) bool {
	_, ok := getImplementer(err).(ErrNotFound)
	return ok
}
```

https://github.com/moby/moby/blob/d48392a35b157612d28c48eb8e4fad1272fa1442/errdefs/helpers.go#L13-L19

```go
// NotFound is a helper to create an error of the class with the same name from any error type
func NotFound(err error) error {
	if err == nil {
		return nil
	}
	return errNotFound{err}
}
```

It would make these errors consistent with other handling, and also preserve the original error in some cases where it may be relevant.

simonferquel: > All the docker context commands are local operations, so (technically) should not need to care about the current context. i.e, if the "current" context uses unix:///no-such-thing), it should not be an error until I try to connect to dockerd or the k8s API, so docker context use <a valid context> should work (as it doesn't require a context to perform that action).

We don't need to do any request, but the initialization of the CLI still occur. And thus it parses the Host to create the WithHost option (without issuing request). so `unix:///some-not-existing-file` work (it is valid syntactically), but `unix///malformed` (note the missing `:`) fails with a parsing error

cli/command/context/export-import_test.go
```diff
@@ -34,9 +34,10 @@ func TestExportImportWithFile(t *testing.T) {
 
 func TestExportImportPipe(t *testing.T) {
 	closeChan := make(chan struct{})
-	cli, cleanup := makeFakeCli(t, withPipeInOut(closeChan))
+	cli, cleanup := makeFakeCli(t)
 	defer cleanup()
 	createTestContextWithKube(t, cli)
+	withPipeInOut(closeChan)(cli)
```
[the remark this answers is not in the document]
simonferquel: yep it is not needed anymore (and removed)

cli/context/store/store.go
thaJeztah: Arf, yes.

We'll have to think of a solution for that (e.g., if we can delay evaluating that option until it's actually needed 🤔 )

cli/command/context/cmd.go
simonferquel: Not sure about that: for now, when a context is set and I list the contexts, the `default` context appears with no docker/kubernetes endpoint. Not sure if Inspect makes sense.

Maybe indicate that the `default` context cannot be inspected ?

cli/command/context/list.go
simonferquel: ok for a followup

cli/context/store/store.go
simonferquel: Actually, I now validate the Host on create/update so it should not be a problem anymore. Only case this error can happen, is if the user delete its store content manually but keep the currentContext set in its config file

simonferquel:
ping @thaJeztah, can you PTAL the last commit ?

cli/command/context/cmd.go
thaJeztah: Yes, I think it's ok to produce a better /helpful error for now (then we can revisit in future)

simonferquel: It is in fact part of the latest commit :)

simonferquel:
The full export does not contain a kubeconfig file, no.
I'll squash the commits then, thanks a lot!

thaJeztah:
> The full export does not contain a kubeconfig file, no. 

That's something we should think about doing, so that we can export / backup

thaJeztah:
@silvin-lubecki @vdemeester still LGTY?

simonferquel:
Actually the full export contains everything we need to construct a kubeconfig file, but the layout is different (partly because we thought it would be nice to make it mandatory to put TLS material in a separate dir)

PR #1502, opened 2018-11-07
Add example get API version

Kilo59:
Getting the client API version is non-intuitive. 
Other keys follow the json example or the version output, however `'{{.Client.ApiVersion}}'` does not work, nor does `'{{.Client.APIversion}}'`

**- What I did**

Added an example to the `docker version` docs

**- How I did it**

**- How to verify it**

Added an example for getting Client API version

### DIFF
--- a/docs/reference/commandline/version.md
+++ b/docs/reference/commandline/version.md
@@ -69,6 +69,14 @@ $ docker version --format '{{.Server.Version}}'
 19.03.8
 ```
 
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
Thanks for the loooong wait; looks like I forgot about this one. I did a quick rebase, and changed the version to be a bit more current.

LGTM

PR #1505, opened 2018-11-08
ssh options env var

andrevtg:
Signed-off-by: Andre Fernandes <andre@vertigo.com.br>

Proposal: DOCKER_HOST_SSH_OPTIONS environment variable

- What I did
The new "ssh://" protocol for DOCKER_HOST is great, but sometimes we need to use different ssh options than the default. A common scenario if the need to point to a project-specific ssh config file instead of the default one at "~/.ssh/config".

- How I did it
An environment variable DOCKER_HOST_SSH_OPTIONS that can hold options for the ssh command docker uses when DOCKER_HOST contains the "ssh://" prefix.

- How to verify it
export DOCKER_HOST_SSH_OPTIONS="-F /my/own/ssh.cfg"
export DOCKER_HOST="ssh://somehost"
docker version

DOCKER_HOST_SSH_OPTIONS to change ssh options when using "ssh://" protocol.

### DIFF
--- a/cli/connhelper/ssh/ssh.go
+++ b/cli/connhelper/ssh/ssh.go
@@ -4,8 +4,15 @@ package ssh
 
 import (
 	"net/url"
+	"os"
+	"strings"
 
 	"github.com/pkg/errors"
+)
+
+// ssh options env variable
+var (
+	dockerSSHOptions = os.Getenv("DOCKER_HOST_SSH_OPTIONS")
 )
 
 // New returns cmd and its args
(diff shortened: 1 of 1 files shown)


andrevtg:
Resubmitted after major screwup on original PR.

cli/connhelper/ssh/ssh.go
```diff
@@ -65,6 +74,9 @@ func (sp *sshSpec) Args() []string {
 	if sp.port != "" {
 		args = append(args, "-p", sp.port)
 	}
+	if sp.options != "" {
+		args = append(args, strings.Split(sp.options, " ")...)
```
AkihiroSuda: maybe better to use https://godoc.org/github.com/google/shlex ?

PR #1509, opened 2018-11-12
 Data Path Port configuration support

selansen:
This PR chnages allow user to configure data path
port number. By default we use 4789 port number. But this commit
will allow user to configure port number during swarm init.
Data path port can't be modified after swarm init.

Signed-off-by: selansen <elango.siva@docker.com>

**- What I did**

**- How I did it**

**- How to verify it**


selansen:
@albers @thaJeztah @vdemeester PTAL

cli/command/swarm/opts.go
```diff
@@ -35,6 +35,7 @@ const (
 	flagAvailability              = "availability"
 	flagCACert                    = "ca-cert"
 	flagCAKey                     = "ca-key"
+	flagDataPathPort              = "data-path-port"
```
thaJeztah: Can you move this one up, to be below `flagDataPathAddr` ?

cli/command/swarm/init.go
```diff
@@ -52,6 +53,8 @@ func newInitCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation(flagDefaultAddrPool, "version", []string{"1.39"})
 	flags.Uint32Var(&opts.DefaultAddrPoolMaskLength, flagDefaultAddrPoolMaskLength, 24, "default address pool subnet mask length")
 	flags.SetAnnotation(flagDefaultAddrPoolMaskLength, "version", []string{"1.39"})
+	flags.Uint32Var(&opts.dataPathPort, flagDataPathPort, 4789, "Data Path port number")
```
thaJeztah: I _think_ we should remove the default here, as the default is set by the daemon.

Can you also move the new lines up (below the `data-path-addr` flag?

The flag description could be a bit more consistent with the existing flag, and (I think) should mention the valid port ranges;

```suggestion
	flags.Uint32Var(&opts.dataPathPort, flagDataPathPort, 0, "Port number to use for data path traffic (1024 - 49151). If no value is set or is set to 0, the default port (4789) is used.")
```

contrib/completion/bash/docker
```diff
@@ -3698,7 +3698,7 @@ _docker_swarm_init() {
 			COMPREPLY=( $( compgen -W "active drain pause" -- "$cur" ) )
 			return
 			;;
-		--cert-expiry|--default-addr-pool|--default-addr-pool-mask-length|--dispatcher-heartbeat|--external-ca|--max-snapshots|--snapshot-interval|--task-history-limit )
+		--cert-expiry|--data-path-port|--default-addr-pool|--default-addr-pool-mask-length|--dispatcher-heartbeat|--external-ca|--max-snapshots|--snapshot-interval|--task-history-limit )
```
thaJeztah: Are we missing `--data-path-addr` here? (if so, we should have a separate PR for that first)

@albers could you check if it has to be added here?

contrib/completion/bash/docker
```diff
@@ -3718,7 +3718,7 @@ _docker_swarm_init() {
 
 	case "$cur" in
 		-*)
-			COMPREPLY=( $( compgen -W "--advertise-addr --autolock --availability --cert-expiry --data-path-addr --default-addr-pool --default-addr-pool-mask-length --dispatcher-heartbeat --external-ca --force-new-cluster --help --listen-addr --max-snapshots --snapshot-interval --task-history-limit" -- "$cur" ) )
+			COMPREPLY=( $( compgen -W "--advertise-addr --autolock --availability --cert-expiry --data-path-addr --data-path    -port --default-addr-pool --default-addr-pool-mask-length --dispatcher-heartbeat --external-ca --force-new-cluster --help --listen-addr --max-snapshots --snapshot-interval --task-history-limit " -- "$cur" ) )
```
thaJeztah: Looks like there's some spaces that slipped in here;

```suggestion
			COMPREPLY=( $( compgen -W "--advertise-addr --autolock --availability --cert-expiry --data-path-addr --data-path-port --default-addr-pool --default-addr-pool-mask-length --dispatcher-heartbeat --external-ca --force-new-cluster --help --listen-addr --max-snapshots --snapshot-interval --task-history-limit " -- "$cur" ) )
```

contrib/completion/zsh/_docker
```diff
@@ -2296,7 +2296,8 @@ __docker_swarm_subcommand() {
                 "($help)--listen-addr=[Listen address]:ip\:port: " \
                 "($help)--max-snapshots[Number of additional Raft snapshots to retain]" \
                 "($help)--snapshot-interval[Number of log entries between Raft snapshots]" \
-                "($help)--task-history-limit=[Task history retention limit]:limit: " && ret=0
+                "($help)--task-history-limit=[Task history retention limit]:limit: " && ret=0 \
+                "($help)--data-path-port=[Data Path Port] "
```
thaJeztah: I think this may not work, because the `&& ret=0` should be last.

Could you also move this up, to below `--data-path-addr` ? (if that works)?

docs/reference/commandline/swarm_init.md
```diff
@@ -36,6 +36,7 @@ Options:
       --max-snapshots uint              Number of additional Raft snapshots to retain
       --snapshot-interval uint          Number of log entries between Raft snapshots (default 10000)
       --task-history-limit int          Task history retention limit (default 5)
+      --data-path-port uint             Data path port number
```
thaJeztah: Could you move this up; below `--data-path-addr`? The cli sorts this alphabetically in the output.

This will also have to be updated to match the actual output (if you make my other suggested changes for the flag description)

cli/command/system/info.go
```diff
@@ -235,6 +235,7 @@ func printSwarmInfo(dockerCli command.Cli, info types.Info) {
 			fmt.Fprintln(dockerCli.Out(), " Default Address Pool:", strAddrPool.String())
 			fmt.Fprintln(dockerCli.Out(), " SubnetSize:", info.Swarm.Cluster.SubnetSize)
 		}
+		fmt.Fprintln(dockerCli.Out(), " Data Path Port:", info.Swarm.Cluster.DataPathPort)
```
thaJeztah: We should only print this if `info.Swarm.Cluster.DataPathPort > 0`, otherwise it's printing incorrect information on older daemons (which don't return the port number)

cli/command/system/testdata/docker-info-with-swarm.golden
```diff
@@ -21,6 +21,7 @@ Swarm: active
  ClusterID: 9vs5ygs0gguyyec4iqf2314c0
  Managers: 1
  Nodes: 1
+ Data Path Port: 0
```
thaJeztah: With a daemon that supports this new feature, this would show `4789` (correct?) if so, can you change this to `4789`?

docs/reference/commandline/swarm_init.md
```diff
@@ -142,6 +143,11 @@ Format example is `--default-addr-pool-mask-length 24`
 
 This flag sets up task history retention limit.
 
+### `--data-path-port`
+
+This flag specifies data path port number.
```
thaJeztah: This doesn't provide a lot of information beyond the flag name itself. Perhaps you can write a bit more about what the purpose is of this flag, and in which case you'd want to set this option. I guess we also want to mention;

- (as mentioned above) why set this option?
- The default (if this option is not set)
- That this is used for the whole swarm (i.e., it's not configurable _per node_)
- That it's an UDP port
- Valid port range
- That it can only be set when initializing the swam. Updating is not (yet?) supported

If you need help phrasing; can you reach out to @ahh-docker ?

docs/reference/commandline/swarm_init.md
```diff
@@ -142,6 +143,11 @@ Format example is `--default-addr-pool-mask-length 24`
...
+Format example is `--data-path-port 4789`
```
thaJeztah: This example isn't very useful, because this is the default port; also, we may want to show a full example, and show how to check the port that is used;

```
docker swarm init --advertise-addr=xxx --data-path-addr=yyy --data-path-port=zzz

# output of the command
```

Then check the output of `docker info` to verify;

```
docker info --format='{{.Something.Something}}'
4789
```

cli/command/swarm/init.go
```diff
@@ -24,6 +24,7 @@ type initOptions struct {
 	availability              string
 	defaultAddrPools          []net.IPNet
 	DefaultAddrPoolMaskLength uint32
+	dataPathPort              uint32
```
thaJeztah: Can you move this under `dataPathAddr`?

vendor.conf
```diff
@@ -12,7 +12,7 @@ github.com/cpuguy83/go-md2man v1.0.8
 github.com/davecgh/go-spew 346938d642f2ec3594ed81d874461961cd0faa76 # v1.1.0
 github.com/dgrijalva/jwt-go a2c85815a77d0f951e33ba4db5ae93629a1530af
 github.com/docker/distribution 83389a148052d74ac602f5f1d62f86ff2f3c4aa5
-github.com/docker/docker bd224b5fe564f343e22fa76ae8569ec949f35802
+github.com/docker/docker 0b7cb16dde4a20d024c7be59801d63bcfd18611b
```
thaJeztah: Can you move the actual vendoring update, and changes to `vendor.conf` to a separate commit?

thaJeztah:
Left some comments; be careful, because GitHub is hiding some of those;

screen shot 2018-11-26 at 12 32 50

contrib/completion/bash/docker
albers: This is the correct location for an option with an argument for which we cannot not provide a specific completion.

thaJeztah: Thanks! I can open a PR to add it 👍

thaJeztah: oh, actually, I see it's a few lines below

cli/command/system/testdata/docker-info-with-swarm.golden
selansen: if I change to 4789 then CI fails now. If that is ok with you , I can go ahead and change it to 4789. To keep CI green, I made it to 0 as of now.  Let me know how do you like to proceed?

selansen:
> Left some comments; be careful, because GitHub is hiding some of those;

I checked all your review comments multiple times and not sure if I still not seeing any of your comment. Pls let me know if I miss anything. 

PATL @thaJeztah @albers

cli/command/system/testdata/docker-info-with-swarm.golden
thaJeztah: It's failing in that case, because the `DataPathPort` isn't set in that test; the test is using the info that's set in https://github.com/docker/cli/blob/ca782599fbf5bcea70b12fc05fc1b35cb8bebc64/cli/command/system/info_test.go#L125-L191

See https://github.com/docker/cli/blob/ca782599fbf5bcea70b12fc05fc1b35cb8bebc64/cli/command/system/info_test.go#L194-L195


If we want to test the output with and without `DataPathPort` set, we could add two test-cases (and corresponding `.golden` files) in https://github.com/docker/cli/blob/ca782599fbf5bcea70b12fc05fc1b35cb8bebc64/cli/command/system/info_test.go#L231-L251

IMO that's a "nice to have", but not super-critical (the `docker info` part is quite trivial)

selansen:
@thaJeztah Do I still need to make any change? Pls let me know

docs/reference/commandline/swarm_init.md
```diff
@@ -130,6 +131,21 @@ management traffic of the cluster.
 If unspecified, Docker will use the same IP address or interface that is used for the
 advertise address.
 
+### `--data-path-port`
+
+This flag specifies data path port number to use for data path traffic (1024 - 49151).
```
thaJeztah: Looking at this again, perhaps it's better to leave out the `--data-path-addr` option, otherwise we'll have to explain here that `--data-path-addr` only applies to the node that you're connected with, and not for the whole cluster (wheras `--data-path-port` is a cluster-wide configuration).

I tried rewriting it a bit; let me know if you like this suggestion (**note** I had to indent the suggestion below, due to the nested ` ``` ` );

```suggestion
	This flag allows you to configure the UDP port number to use for data path
	traffic. The provided port nummer must be within the 1024 - 49151 range. If
	this flag is not set or is set to 0, the default port (4789) is used.

	The data path port can only be configured when initializing the swarm, and
	applies to all nodes that join the swarm.

	The following example initializes a new Swarm, and configures the data path
	port to UDP port 7777;

	```bash
	docker swarm init --data-path-addr=20.20.20.1 --data-path-port=7777
	```

	After the swarm is initialized, use the `docker info` command to verify that
	the port is configured:

	```bash
	docker info

	...
	ClusterID: 9vs5ygs0gguyyec4iqf2314c0
	Managers: 1
	Nodes: 1
	Data Path Port: 7777
	...
	```
```

contrib/completion/zsh/_docker
```diff
@@ -2285,6 +2285,7 @@ __docker_swarm_subcommand() {
                 $opts_help \
                 "($help)--advertise-addr=[Advertised address]:ip\:port: " \
                 "($help)--data-path-addr=[Data path IP or interface]:ip " \
+                "($help)--data-path-port=[Data Path Port] " \
```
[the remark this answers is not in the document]
selansen: some lines has and some lines doesn't have. anyways corrected it.

docs/reference/commandline/swarm_init.md
selansen: did small corrections and update it. Pls check it out.

selansen:
@albers @vdemeester PTAL

docs/reference/commandline/swarm_init.md
```diff
@@ -26,6 +26,7 @@ Options:
       --availability string             Availability of the node ("active"|"pause"|"drain") (default "active")
       --cert-expiry duration            Validity period for node certificates (ns|us|ms|s|m|h) (default 2160h0m0s)
       --data-path-addr string           Address or interface to use for data path traffic (format: <ip|interface>)
+      --data-path-port uint32           Data path port number used for data path traffic
```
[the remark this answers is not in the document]
thaJeztah: Oh! Thanks I missed that one, thought it was updated 😅

selansen:
@thaJeztah PATL

PR #1511, opened 2018-11-12
Make it possible to override the volume mounts and shell for the dev container

wk8:
**- What I did**

This patch provides some way for non-Linux devs to work on this repo without
suffering too much from the notorious terrible I/O perf of mounted volumes (eg
on OS X).

**- How I did it**

We now allow to override the volume mounts and shell for the dev container
through the `DOCKER_CLI_MOUNTS` and `DOCKER_CLI_SHELL` env variables.
We also allow setting the dev container name through the
`DOCKER_CLI_CONTAINER_NAME` env var.

The motivation for allowing overriding the volume mounts is the same as for
https://github.com/moby/moby/pull/37845, namely that I/O perf on native mounted
disks on non-Linux (notably Mac OS) is just terrible, thus making it a real
pain to develop: one has to choose between re-building the image after every
single change (eg to run a test) or just work directly inside the same
container (eg with vim, but even then one would have to re-configure their dev
container every time it gets destroyed - containers, after all, are not
supposed to be long-lived).

Allowing to override DOCKER_CLI_MOUNTS makes it easy for everyone
to decide what their volume/syncing strategy is; for example
one can choose to use [docker-sync](https://github.com/EugenMayer/docker-sync).

As for the shell, it's nice to be able to use bash instead of the more
bare-bones `ash` if preferred.

Finally, being able to name the container can come in handy for easier
scripting on the host.

This patch won't change anything for anyone who doesn't set these env variables
in their environment.

**- How to verify it**

The output of:
```
make -f docker.Makefile -n shell
```
should be unchanged compared to what it was before this patch; but:
```
DOCKER_CLI_MOUNTS='-v test:/test' make -f docker.Makefile -n shell
```
should replace the `-v` flag from the command above with just
`-v test:/test`, and finally:
```
DOCKER_CLI_MOUNTS=' ' make -f docker.Makefile -n shell
```
should be devoid of any `-v` flags (besides the socket one).

All the make targets should be unchanged when not setting any of the
new env vars.

bunny

### DIFF
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -4,17 +4,21 @@
 # Makefile for developing using Docker
 #
 
+# Overridable env vars
+DOCKER_CLI_MOUNTS ?= -v "$(CURDIR)":/go/src/github.com/docker/cli
+DOCKER_CLI_SHELL ?= ash
+DOCKER_CLI_CONTAINER_NAME ?=
+DOCKER_CLI_GO_BUILD_CACHE ?= y
+
 DEV_DOCKER_IMAGE_NAME = docker-cli-dev$(IMAGE_TAG)
 BINARY_NATIVE_IMAGE_NAME = docker-cli-native$(IMAGE_TAG)
 LINTER_IMAGE_NAME = docker-cli-lint$(IMAGE_TAG)
 CROSS_IMAGE_NAME = docker-cli-cross$(IMAGE_TAG)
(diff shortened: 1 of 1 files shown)


wk8:
Small review please? :)

andrewhsu:
SGTM as well!

wk8:
Thanks @andrewhsu and @silvin-lubecki for the review :) What's the process to get it merged at this point?

wk8:
Thanks!

PR #1515, opened 2018-11-13
Prevent overwriting irregular files (cp, save, export commands)

### LINKED ISSUE #1514: cp, save and export allow replacing irregular files
**Description**

The docker commands `cp`, `save` and `export` allow replacing irregular files - e.g. devices under `/dev`. Please see the following screenshot to demonstrate this misbehaviour:

docker-dev-random

As you can see, the device `/dev/random` was replaced with a file containing the saved docker image. This was originally discovered when executing `docker save <image> -o /dev/null` when performing tests, only to find out that applications relying on the `/dev/null` device are misbehaving.

While I think there's no valid use-case for this behaviour the best solution would be to disallow irregular files for the commands mentioned above. This prevents people from accidentally damaging parts of their system.

A PR regarding this is on its way.

1. Pull an arbitrary docker image: `docker pull ubuntu:latest`
2. Save the pulled image to a file using the `-o` option: `docker save ubuntu:latest -o /dev/random`
3. Confirm that the device itself has been replaced with a regular file: `stat /dev/random` shows `regular file` instead of `character special file`.

The device/irregular file has been *replaced* with the output of `cp`/`export`/`save`

Replacing devices/irregular files should be prevented by the CLI to prevent system damage.

```
Docker version 19.03.0-dev, build 374b480e
```

**Output of `docker info`:**

```
Containers: 1
 Running: 0
 Paused: 0
 Stopped: 1
Images: 187
Server Version: 17.12.1-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 9b55aab90508bd389d7654c4baf173a981477d55
runc version: 9f9c96235cc97674e935002fc3d78361b696a69e
init version: v0.13.0 (expected: 949e6facb77383876aeff8a6944dde66b3089574)
Security Options:
 apparmor
 seccomp
  Profile: default
Kernel Version: [...]
Operating System: Ubuntu 18.04.1 LTS
OSType: linux
Architecture: x86_64
CPUs: 4
Total Memory: 15.55GiB
Name: [...]
ID: CR7L:OCP4:IZ76:VHOD:ZCDV:PLTN:QKYD:D7TV:OYKN:3QB5:TQTN:QCAK
Debug Mode (client): false
Debug Mode (server): false
Labels:
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false

WARNING: No swap limit support
```

sw-pschmied:
Signed-off-by: Philipp Schmied <pschmied@schutzwerk.com>

**- What I did**

This fixes #1514.

**- How I did it**

The function `command.ValidateOutputPath` was already present in `cli/command/image/save.go`. I've moved this to `cli/command/utils.go` in order to use this function to fix the issue for all three commands.

After that, I've added code to this existing function to use a new helper method called `ValidateOutputPathFileMode` which also resides in `cli/command/utils.go`. This is being used to check for irregular files/devices and returns an error in case an invalid target is being passed. This is now being called at the start of the functions implementing the `cp`, `export` and `save` commands. For `cp` operations *to* a container, the fix only uses the helper method mentioned above, so it's possible to unify this check for all three operations. The helper method also determines whether the target path points to a device or any other invalid/irregular file to display an error message accordingly.

Also, I've added test cases for all commands.

**- How to verify it**

- Clone the `cli` master branch and compile a static `docker` binary
- Pull an arbitrary docker image: `docker pull ubuntu:latest`
- Invoke `docker save ubuntu:latest -o /dev/random`
- Check the output of `stat /dev/random` indicating `/dev/random` is now a regular file
- Perform the steps above using my PR and see that this operation is prevented, yielding `failed to save image: invalid output path: "/dev/random" must be a directory or a regular file: Got a device`.

cp, save, export: Prevent overwriting irregular files

20foup2

### DIFF
--- a/cli/command/container/export_test.go
+++ b/cli/command/container/export_test.go
@@ -31,3 +31,19 @@ func TestContainerExportOutputToFile(t *testing.T) {
 
 	assert.Assert(t, fs.Equal(dir.Path(), expected))
 }
+
+func TestContainerExportOutputToIrregularFile(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{
+		containerExportFunc: func(container string) (io.ReadCloser, error) {
+			return ioutil.NopCloser(strings.NewReader("foo")), nil
+		},
+	})
+	cmd := NewExportCommand(cli)
+	cmd.SetOutput(ioutil.Discard)
+	cmd.SetArgs([]string{"-o", "/dev/random", "container"})
+
+	err := cmd.Execute()
+	assert.Assert(t, err != nil)
+	expected := `"/dev/random" must be a directory or a regular file`
+	assert.ErrorContains(t, err, expected)
+}
--- a/cli/command/utils.go
+++ b/cli/command/utils.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/pkg/system"
+	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
 )
 
@@ -124,4 +125,38 @@ func AddPlatformFlag(flags *pflag.FlagSet, target *string) {
 	flags.StringVar(target, "platform", os.Getenv("DOCKER_DEFAULT_PLATFORM"), "Set platform if server is multi-platform capable")
 	flags.SetAnnotation("platform", "version", []string{"1.32"})
 	flags.SetAnnotation("platform", "experimental", nil)
+}
+
+// ValidateOutputPath validates the output paths of the `export` and `save` commands.
+func ValidateOutputPath(path string) error {
+	dir := filepath.Dir(path)
+	if dir != "" && dir != "." {
+		if _, err := os.Stat(dir); os.IsNotExist(err) {
+			return errors.Errorf("invalid output path: directory %q does not exist", dir)
+		}
+	}
+	// check whether `path` points to a regular file
+	// (if the path exists and doesn't point to a directory)
+	if fileInfo, err := os.Stat(path); !os.IsNotExist(err) {
+		if fileInfo.Mode().IsDir() || fileInfo.Mode().IsRegular() {
+			return nil
+		}
+
+		if err := ValidateOutputPathFileMode(fileInfo.Mode()); err != nil {
+			return errors.Wrapf(err, fmt.Sprintf("invalid output path: %q must be a directory or a regular file", path))
+		}
+	}
+	return nil
+}
+
+// ValidateOutputPathFileMode validates the output paths of the `cp` command and serves as a
+// helper to `ValidateOutputPath`
+func ValidateOutputPathFileMode(fileMode os.FileMode) error {
+	switch {
+	case fileMode&os.ModeDevice != 0:
+		return errors.New("got a device")
+	case fileMode&os.ModeIrregular != 0:
+		return errors.New("got an irregular file")
+	}
+	return nil
 }
--- a/cli/command/container/cp.go
+++ b/cli/command/container/cp.go
@@ -128,6 +128,10 @@ func copyFromContainer(ctx context.Context, dockerCli command.Cli, copyConfig cp
 		}
 	}
 
+	if err := command.ValidateOutputPath(dstPath); err != nil {
+		return err
+	}
+
 	client := dockerCli.Client()
 	// if client requests to follow symbol link, then must decide target file to be copied
 	var rebaseName string
@@ -207,6 +211,11 @@ func copyToContainer(ctx context.Context, dockerCli command.Cli, copyConfig cpCo
 
 		dstInfo.Path = linkTarget
 		dstStat, err = client.ContainerStatPath(ctx, copyConfig.container, linkTarget)
+	}
+
+	// Validate the destination path
+	if err := command.ValidateOutputPathFileMode(dstStat.Mode); err != nil {
+		return errors.Wrapf(err, `destination "%s:%s" must be a directory or a regular file`, copyConfig.container, dstPath)
 	}
 
 	// Ignore any error and assume that the parent directory of the destination
--- a/cli/command/container/cp_test.go
+++ b/cli/command/container/cp_test.go
@@ -190,3 +190,12 @@ func TestSplitCpArg(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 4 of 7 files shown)


cli/command/utils.go
```diff
@@ -125,3 +126,21 @@ func AddPlatformFlag(flags *pflag.FlagSet, target *string) {
 	flags.SetAnnotation("platform", "version", []string{"1.32"})
 	flags.SetAnnotation("platform", "experimental", nil)
 }
+
+// ValidateOutputPath validates the output paths of the `export` and `save` commands.
+func ValidateOutputPath(path string) error {
+	dir := filepath.Dir(path)
+	if dir != "" && dir != "." {
+		if _, err := os.Stat(dir); os.IsNotExist(err) {
+			return errors.Errorf("unable to validate output path: directory %q does not exist", dir)
```
thaJeztah: perhaps this should be "invalid output path" instead of "unable to validate output path" (because we were able to validate it, but it turned out to be an invalid path)

cli/command/utils.go
```diff
@@ -125,3 +126,21 @@ func AddPlatformFlag(flags *pflag.FlagSet, target *string) {
...
+		}
+	}
+	// check whether `path` points to a regular file
+	// (if the path exists and doesn't point to a directory)
+	if fileInfo, err := os.Stat(path); !os.IsNotExist(err) {
+		if !fileInfo.Mode().IsDir() && !fileInfo.Mode().IsRegular() {
+			return errors.Errorf("unable to validate output path: %q must be a directory or a regular file", path)
```
thaJeztah: Same here for the "unable to validate.."

Also thinking if we should mention that the target is a "special" file (not sure)

sw-pschmied:
Hi and thanks for your feedback! Applying changes shortly.

sw-pschmied:
I've pushed the changes, made some additional refactoring and updated my original post. The major changes to the post are:

- Mention and describe the additional helper method `ValidateOutputPathFileMode`
- The error message now contains information on the type of invalid target being passed (device / misc irregular file)

sw-pschmied:
ping @thaJeztah - is there else to change?

cli/command/container/export_test.go
```diff
@@ -31,3 +31,19 @@ func TestContainerExportOutputToFile(t *testing.T) {
...
+			return ioutil.NopCloser(strings.NewReader("foo")), nil
+		},
+	})
+	cmd := NewExportCommand(cli)
+	cmd.SetOutput(ioutil.Discard)
+	cmd.SetArgs([]string{"-o", "/dev/random", "container"})
+
+	err := cmd.Execute()
+	assert.Assert(t, err != nil)
+	expected := "/dev/random\" must be a directory or a regular file"
```
thaJeztah: Same here (just a nit though 😅)

```suggestion
	expected := `"/dev/random" must be a directory or a regular file`
```

cli/command/utils.go
```diff
@@ -125,3 +126,34 @@ func AddPlatformFlag(flags *pflag.FlagSet, target *string) {
...
+	dir := filepath.Dir(path)
+	if dir != "" && dir != "." {
+		if _, err := os.Stat(dir); os.IsNotExist(err) {
+			return errors.Errorf("invalid output path: directory %q does not exist", dir)
+		}
+	}
+	// check whether `path` points to a regular file
+	// (if the path exists and doesn't point to a directory)
+	if fileInfo, err := os.Stat(path); !os.IsNotExist(err) {
+		if !fileInfo.Mode().IsDir() && !fileInfo.Mode().IsRegular() {
```
thaJeztah: You can reduce nesting here;

```suggestion
		if fileInfo.Mode().IsDir() || fileInfo.Mode().IsRegular() {
			return nil
		}
		if err := ValidateOutputPathFileMode(fileInfo.Mode()); err != nil {
			return errors.Wrap(err, fmt.Sprintf("invalid output path: %q must be a directory or a regular file", path))
		}		
```

cli/command/utils.go
```diff
@@ -125,3 +126,34 @@ func AddPlatformFlag(flags *pflag.FlagSet, target *string) {
...
+		}
+	}
+	return nil
+}
+
+// ValidateOutputPathFileMode validates the output paths of the `cp` command and serves as a
+// helper to `ValidateOutputPath`
+func ValidateOutputPathFileMode(fileMode os.FileMode) error {
+	if fileMode&os.ModeDevice != 0 {
+		return errors.New("Got a device")
```
thaJeztah: Should probably be lowercase for an error;

```suggestion
		return errors.New("got a device")
```

cli/command/utils.go
```diff
@@ -125,3 +126,34 @@ func AddPlatformFlag(flags *pflag.FlagSet, target *string) {
...
+	} else if fileMode&os.ModeIrregular != 0 {
+		return errors.New("Got an irregular file")
```
thaJeztah: Same here

```suggestion
		return errors.New("got an irregular file")
```

sw-pschmied:
ping @thaJeztah :)

cli/command/utils.go
```diff
@@ -125,3 +126,36 @@ func AddPlatformFlag(flags *pflag.FlagSet, target *string) {
...
+	}
+	// check whether `path` points to a regular file
+	// (if the path exists and doesn't point to a directory)
+	if fileInfo, err := os.Stat(path); !os.IsNotExist(err) {
+		if fileInfo.Mode().IsDir() || fileInfo.Mode().IsRegular() {
+			return nil
+		}
+
+		if err := ValidateOutputPathFileMode(fileInfo.Mode()); err != nil {
+			return errors.Wrap(err, fmt.Sprintf("invalid output path: %q must be a directory or a regular file", path))
```
silvin-lubecki: small nit: use [errors.Wrapf](https://godoc.org/github.com/pkg/errors#Wrapf) instead.

sw-pschmied:
Thanks, I just pushed my changes @silvin-lubecki

silvin-lubecki:
Thanks a lot @sw-pschmied for detecting this issue AND fixing it ! 👍👍

PR #1517, opened 2018-11-13
Migrate TestTrustedCreateFromBadTrustServer from moby into cli/e2e

### LINKED ISSUE #1218: Migrate TestTrustedCreateFromBadTrustServer to cli/e2e
Noticed this TODO, so let's create a ticket in case someone wants to work on this; https://github.com/docker/cli/blob/4cb3c70f36baeade76879694a587358be2a74854/e2e/container/create_test.go#L35

The `TestTrustedCreateFromBadTrustServer ` test was removed from the moby/moby integration tests (because it's testing client-side functionality), but didn't get re-implemented as an e2e test in this repository; the existing test can be found here; https://github.com/moby/moby/pull/36515/files#diff-4b1e56bb77ac16f2ccf956fc24cf0a82L331, but needs a rewrite to work in the e2e suite in this repo

/cc @vdemeester

glefloch:
Signed-off-by: glefloch <glfloch@gmail.com>

**- What I did**
I have migrate `TestTrustedCreateFromBadTrustServer` from `moby/moby` repository into `docker/cli` repository. This aims to closes #1218 

**- How I did it**
I have added a new notary-server into `e2e` test environment called `evil-notary-server`. Then duplicate the behavior of the test describe in moby repository.

**- How to verify it**
You can verify it by running e2e tests.

Add test on container creation using a wrong notary server


e2e/container/create_test.go
```diff
@@ -6,10 +6,16 @@ import (
...
 	"github.com/docker/cli/e2e/internal/fixtures"
 	"github.com/docker/cli/internal/test/environment"
+	"gotest.tools/fs"
 	"gotest.tools/icmd"
 	"gotest.tools/skip"
 )
 
+const (
+	evilNotaryURL = "https://evil-notary-server:4444"
+	evilImageName = "registry:5000/evil-alpine:latest"
```
eiais: Should these be put in e2e/internal/fixtures/fixtures.go?

e2e/container/create_test.go
```diff
@@ -32,4 +38,69 @@ func TestCreateWithContentTrust(t *testing.T) {
...
+
+func setupEvilNotaryServer(t *testing.T) fs.Dir {
+	t.Helper()
+	dir := fs.NewDir(t, "evil_notary_test", fs.WithMode(0700), fs.WithFile("config.json", `
+	{
+		"auths": {
+			"registry:5000": {
+				"auth": "ZWlhaXM6cGFzc3dvcmQK"
+			},
+			"https://evil-notary-server:4444": {
```
eiais: This seems to be the only difference from fixtures.SetupConfigFile, maybe pass notartURL as an argument there?

e2e/container/create_test.go
glefloch: Yes, you are right, this should be in `fixtures.go`. I am going to move them.

e2e/container/create_test.go
glefloch: yes this is the only difference. I am going to update the code.

e2e/container/create_test.go
```diff
@@ -32,4 +32,50 @@ func TestCreateWithContentTrust(t *testing.T) {
...
+
+	icmd.RunCmd(icmd.Command("docker", "image", "rm", evilImageName)).Assert(t, icmd.Success)
+
+	icmd.RunCmd(icmd.Command("docker", "create", evilImageName),
+		fixtures.WithConfig(dir.Path()),
+		fixtures.WithTrust,
+		fixtures.WithNotary,
+	).Assert(t, icmd.Expected{
+		ExitCode: 1,
+		Err:      "remote trust data does not exist for registry:5000/evil-alpine",
```
silvin-lubecki: When I look at the [removed test](https://github.com/moby/moby/pull/36515/files#diff-4b1e56bb77ac16f2ccf956fc24cf0a82L360) on `moby/moby` I can't see the same result as yours.
Can you add some context in the PR comment to explain why you expect this result, and maybe in the test too?

glefloch: Well, updating the WithNotary ends up with the command exitCode equals to 0. There may be another problem with the configuration.

glefloch: Finally, after looking to the previous test, I was actually missing the first part. I updated the test and report comments from the previous test.

silvin-lubecki: I can't see the comments from `moby/moby` test, did you push them?

glefloch: Sorry for this, I just pushed the update.

e2e/container/create_test.go
```diff
@@ -32,4 +32,72 @@ func TestCreateWithContentTrust(t *testing.T) {
...
+func TestTrustedCreateFromUnreachableTrustServer(t *testing.T) {
+	dir := fixtures.SetupConfigFile(t)
+	defer dir.Remove()
+	image := fixtures.CreateMaskedTrustedRemoteImage(t, registryPrefix, "trust-create", "latest")
+
+	result := icmd.RunCmd(
+		icmd.Command("docker", "create", image),
+		fixtures.WithConfig(dir.Path()),
+		fixtures.WithTrust,
+		fixtures.WithNotaryServer("https://invalidnotaryserver"),
```
justincormack: There is an IETF recommendation for invalid domains: https://tools.ietf.org/html/rfc2606#page-2 - so better to use `notary.invalid`.

justincormack:
One minor comment but otherwise LGTM

e2e/container/create_test.go
glefloch: You right, this is better. I pushed the update.

PR #1519, opened 2018-11-14
Export stack commands

silvin-lubecki:
**- What I did**
Export stack commands for vendoring and extract/factorize common code.

image

### DIFF
--- a/cli/command/stack/common.go
+++ b/cli/command/stack/common.go
@@ -4,6 +4,10 @@ import (
 	"fmt"
 	"strings"
 	"unicode"
+
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/command/stack/kubernetes"
+	"github.com/spf13/pflag"
 )
 
 // validateStackName checks if the provided string is a valid stack name (namespace).
@@ -28,4 +32,19 @@ func validateStackNames(namespaces []string) error {
 
 func quotesOrWhitespace(r rune) bool {
 	return unicode.IsSpace(r) || r == '"' || r == '\''
+}
+
+func runOrchestratedCommand(dockerCli command.Cli, flags *pflag.FlagSet, commonOrchestrator command.Orchestrator, swarmCmd func() error, kubernetesCmd func(*kubernetes.KubeCli) error) error {
+	switch {
+	case commonOrchestrator.HasAll():
+		return errUnsupportedAllOrchestrator
+	case commonOrchestrator.HasKubernetes():
+		kli, err := kubernetes.WrapCli(dockerCli, kubernetes.NewOptions(flags, commonOrchestrator))
+		if err != nil {
+			return err
+		}
+		return kubernetesCmd(kli)
+	default:
+		return swarmCmd()
+	}
 }
--- a/cli/command/stack/deploy.go
+++ b/cli/command/stack/deploy.go
@@ -50,7 +50,7 @@ func newDeployCommand(dockerCli command.Cli, common *commonOptions) *cobra.Comma
 			if err != nil {
 				return err
 			}
-			return RunDeploy(dockerCli, cmd.Flags(), config, commonOrchestrator, opts)
+			return RunDeploy(dockerCli, cmd.Flags(), config, common.Orchestrator(), opts)
 		},
 	}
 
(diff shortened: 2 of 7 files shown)


cli/command/stack/common.go
```diff
@@ -29,3 +33,23 @@ func validateStackNames(namespaces []string) error {
 func quotesOrWhitespace(r rune) bool {
 	return unicode.IsSpace(r) || r == '"' || r == '\''
 }
+
+func runOrchestratedCommand(dockerCli command.Cli, flags *pflag.FlagSet, common *commonOptions, swarmCmd func() error, kubernetesCmd func(*kubernetes.KubeCli) error) error {
+	commonOrchestrator := command.OrchestratorSwarm // default for top-level deploy command
+	if common != nil {
+		commonOrchestrator = common.orchestrator
+	}
+	//return RunPs(dockerCli, cmd.Flags(), commonOrchestrator, opts)
```
vdemeester: dead code ?

silvin-lubecki: What are you talking about... 😇

silvin-lubecki:
Yep good guess @vdemeester 😄

cli/command/stack/deploy.go
```diff
@@ -74,17 +74,8 @@ func newDeployCommand(dockerCli command.Cli, common *commonOptions) *cobra.Comma
...
-	case commonOrchestrator.HasKubernetes():
-		kli, err := kubernetes.WrapCli(dockerCli, kubernetes.NewOptions(flags, commonOrchestrator))
-		if err != nil {
-			return errors.Wrap(err, "unable to deploy to Kubernetes")
-		}
-		return kubernetes.RunDeploy(kli, opts, config)
-	default:
-		return swarm.RunDeploy(dockerCli, opts, config)
-	}
+func RunDeploy(dockerCli command.Cli, flags *pflag.FlagSet, config *composetypes.Config, common *commonOptions, opts options.Deploy) error {
```
simonferquel: this is exposed, but commonOptions is not :(

cli/command/stack/ps.go
```diff
@@ -46,3 +35,10 @@ func newPsCommand(dockerCli command.Cli, common *commonOptions) *cobra.Command {
 	kubernetes.AddNamespaceFlag(flags)
 	return cmd
 }
+
+// RunPs performs a stack ps against the specified orchestrator
+func RunPs(dockerCli command.Cli, flags *pflag.FlagSet, common *commonOptions, opts options.PS) error {
```
simonferquel: same here

cli/command/stack/remove.go
```diff
@@ -22,22 +23,17 @@ func newRemoveCommand(dockerCli command.Cli, common *commonOptions) *cobra.Comma
...
+			return RunRemove(dockerCli, cmd.Flags(), common, opts)
 		},
 	}
 	flags := cmd.Flags()
 	kubernetes.AddNamespaceFlag(flags)
 	return cmd
 }
+
+// RunRemove performs a stack remove against the specified orchestrator
+func RunRemove(dockerCli command.Cli, flags *pflag.FlagSet, common *commonOptions, opts options.Remove) error {
```
simonferquel: And here

cli/command/stack/services.go
```diff
@@ -44,3 +33,10 @@ func newServicesCommand(dockerCli command.Cli, common *commonOptions) *cobra.Com
 	kubernetes.AddNamespaceFlag(flags)
 	return cmd
 }
+
+// RunServices performs a stack services against the specified orchestrator
+func RunServices(dockerCli command.Cli, flags *pflag.FlagSet, common *commonOptions, opts options.Services) error {
```
simonferquel: and here as well

silvin-lubecki:
@simonferquel @vdemeester PTAL, I rebased and cleaned this PR 🦁

cli/command/stack/list.go
```diff
@@ -35,7 +35,8 @@ func newListCommand(dockerCli command.Cli, common *commonOptions) *cobra.Command
 	return cmd
 }
 
-func runList(cmd *cobra.Command, dockerCli command.Cli, opts options.List, orchestrator command.Orchestrator) error {
+// RunList performs a stack list against the specified orchestrator
+func RunList(cmd *cobra.Command, dockerCli command.Cli, opts options.List, orchestrator command.Orchestrator) error {
```
thaJeztah: Only concern I have with exporting all of these is that;

- These functions now become a public "API", with no interface defined for them
  - this can complicate making changes in the CLI (they were kept non-exported, as their signature/interface was not "stable")
- Some of these functions actually have a "UX" component built-in (e.g. they print confirmation messages on stdout/stderr, have an option to print formatted output); is that really suitable for external consumption?
  - calling these commands would become almost equivalent to `exec.Cmd("docker stack ls")`


For the _underlying_ code on the other hand, we have various interfaces defined; if those are lacking functionality that's needed, should we review those interfaces, and make that the API to use?

thaJeztah: was discussing with @chris-crone to add a note somewhere (perhaps the readme) to indicate that these functions, even though they're exported, should not be considered a stable API (as in; their signature can change, and shouldn't be relied on)

We could also add notes to the GoDoc of these, similar to https://golang.org/pkg/testing/#RunTests

> An internal function but exported because it is cross-package; part of the implementation of the "go test" command.

But that would be a bit more work to do

silvin-lubecki: > was discussing with @chris-crone to add a note somewhere (perhaps the readme) to indicate that these functions, even though they're exported, should not be considered a stable API (as in; their signature can change, and shouldn't be relied on)

Will do that in a follow-up 👍

PR #1524, opened 2018-11-17
Fix "warning" output on docker system prune

thaJeztah:
This PR has two commits;

### 1. Fix system prune warning missing filters from config-file

The warning, printed before runing docker system prune was missing any filter
that was set in the configuration file. In addition, the warning prefixes the
filters with `label=`, which is no longer accurate, now that the prune command
also supports "until" as a filter.

Before this change, only the filters set on the command-line were shown,
and any filter set in the configuration file was missing;

```
mkdir -p ./test-config
echo '{"pruneFilters": ["label!=never=remove-me", "label=remove=me"]}' > test-config/config.json
docker --config=./test-config system prune --filter until=24h --filter label=hello-world --filter label!=foo=bar --filter label=bar=baz

WARNING! This will remove:
        - all stopped containers
        - all dangling images
        - all dangling build cache
        - Elements to be pruned will be filtered with:
        - label={"label":{"bar=baz":true,"hello-world":true},"label!":{"foo=bar":true},"until":{"24h":true}}
```

With this patch applied, both options from the commandline and options set
in the configuration file are shown;

```
mkdir -p ./test-config
echo '{"pruneFilters": ["label!=never=remove-me", "label=remove=me"]}' > test-config/config.json
docker --config=./test-config system prune --filter until=24h --filter label=hello-world --filter label!=foo=bar --filter label=bar=baz

WARNING! This will remove:
        - all stopped containers
        - all dangling images
        - all dangling build cache
        - Elements to be pruned will be filtered with:
        - filter={"label":{"bar=baz":true,"hello-world":true,"remove=me":true},"label!":{"foo=bar":true,"never=remove-me":true},"until":{"24h":true}}
```

### 2. Make system prune warning filters human-readable

The warning, printed before running `docker system prune` was printing the
filters in JSON format.

This patch attempts to make the output human readable;

- updating the code, and template to print filters individually
- reducing the indentation (which was quite deep)

Before this patch was applied;

```
docker system prune --filter until=24h --filter label=hello-world --filter label!=foo=bar --filter label=bar=baz

WARNING! This will remove:
        - all stopped containers
        - all dangling images
        - all dangling build cache
        - Elements to be pruned will be filtered with:
        - label={"label":{"bar=baz":true,"hello-world":true},"label!":{"foo=bar":true},"until":{"24h":true}}
```

With this patch applied;

```
WARNING! This will remove:
  - all stopped containers
  - all dangling images
  - all dangling build cache

  Elements to be pruned will be filtered with:
  - label!=foo=bar
  - label!=never=remove-me
  - label=bar=baz
  - label=hello-world
  - label=remove=me
  - until=24h

```

### DIFF
--- a/cli/command/system/prune.go
+++ b/cli/command/system/prune.go
@@ -3,6 +3,7 @@ package system
 import (
 	"bytes"
 	"fmt"
+	"sort"
 	"text/template"
 
 	"github.com/docker/cli/cli"
@@ -13,10 +14,10 @@ import (
 	"github.com/docker/cli/cli/command/network"
 	"github.com/docker/cli/cli/command/volume"
 	"github.com/docker/cli/opts"
-	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/versions"
-	units "github.com/docker/go-units"
+	"github.com/docker/go-units"
 	"github.com/spf13/cobra"
+	"vbom.ml/util/sortorder"
 )
 
 type pruneOptions struct {
@@ -54,9 +55,15 @@ func newPruneCommand(dockerCli command.Cli) *cobra.Command {
 }
 
 const confirmationTemplate = `WARNING! This will remove:
-{{- range $_, $warning := . }}
-        - {{ $warning }}
+{{- range $_, $warning := .warnings }}
+  - {{ $warning }}
 {{- end }}
+{{if .filters}}
+  Items to be pruned will be filtered with:
+{{- range $_, $filters := .filters }}
+  - {{ $filters }}
+{{- end }}
+{{end}}
 Are you sure you want to continue?`
 
 func runPrune(dockerCli command.Cli, options pruneOptions) error {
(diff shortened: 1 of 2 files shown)


cli/command/system/prune_test.go
```diff
@@ -11,12 +12,40 @@ import (
 func TestPrunePromptPre131DoesNotIncludeBuildCache(t *testing.T) {
 	cli := test.NewFakeCli(&fakeClient{version: "1.30"})
 	cmd := newPruneCommand(cli)
+	cmd.SetArgs([]string{})
```
thaJeztah: Test was failing on my machine, because `Command.Execute()` "uses the args (`os.Args[1:]` by default)", which in 
my case was `-test.v -test.run ^TestPrunePromptPre131DoesNotIncludeBuildCache$`, and thus produces;

```
prune_test.go:15: assertion failed: error is not nil: "prune" accepts no arguments.
```

cli/command/system/prune.go
```diff
@@ -119,12 +127,23 @@ func confirmationMessage(options pruneOptions) string {
...
+	pruneFilters := command.PruneFilters(dockerCli, options.filter.Value())
+	if pruneFilters.Len() > 0 {
+		// TODO remove fixed list of filters, and print all filters instead,
+		// because the list of filters that is supported by the engine may evolve over time.
+		for _, name := range []string{"label", "label!", "until"} {
+			for _, v := range pruneFilters.Get(name) {
+				filters = append(filters, name+"="+v)
+			}
+		}
+		sort.Slice(filters, func(i, j int) bool {
```
thaJeztah: Prevent the test from being flaky, as the filters are stored as a map, so order is randomised

cli/command/system/prune.go
```diff
@@ -53,17 +55,23 @@ func newPruneCommand(dockerCli command.Cli) *cobra.Command {
 }
 
 const confirmationTemplate = `WARNING! This will remove:
-{{- range $_, $warning := . }}
-        - {{ $warning }}
+{{- range $_, $warning := .warnings }}
+  - {{ $warning }}
 {{- end }}
+{{if .filters}}
+  Elements to be pruned will be filtered with:
```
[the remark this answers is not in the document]
vdemeester: Items seems the best to me :angel:

cli/command/system/prune_test.go
vdemeester: interesting :thinking:

silvin-lubecki:
Not part of this PR, but do you think we could change the message `Elements to be pruned will be filtered with:` with something more explicit like this: `Elements matching the following contraints won't be pruned` ?
I think `filter` is an error prone term, you never know what it means, does it prevent the matching ones from deletion or does it remove them?

thaJeztah:
rebased, and changed `Elements` to `Items` in the last commit 👍

thaJeztah:
Rebased to trigger CI again 👍

PR #1525, opened 2018-11-20
Revert "set default schema to tcp for docker host"

zmilonas:
This partially reverts commit 2431dd144809812e8d3bb03d22af01703a5d36cd.

Signed-off-by: Zachary Milonas <zachary@milonas.pl>

**- What I did**
Reverted the commited change that broke using `DOCKER_HOST` env variable with a port like `127.0.0.1:2345` 

**- How I did it**
Git 

**- How to verify it**
`go test`?

Reverted a change that introduced a regression for setting `DOCKER_HOST` environmental variable with a port


zmilonas:
Well. The reason I'm bringing this up is because we're running into an issue where setting a `DOCKER_HOST` on 18.09.0 of Engine results in the `url.Parse()` error.

I don't see any relevant fixes for that.

lifubang:
I don't know why 18.09 didn't release my PR. If yes, you will get the right answer.

PR #1536, opened 2018-11-28
Replace environmental with environment

nirs:
Signed-off-by: Nir Soffer <nsoffer@redhat.com>


thaJeztah:
> Should I drop this change and make it in k8s.io?

Ah, yes; that's a vendored file, so those should not be modified; we pull those from external locations (k8s.io in this case), and don't modify them.

PR #1537, opened 2018-11-29
do not patch Dockerfiles in CI

thaJeztah:
When building the Dockerfiles for development, those images are mainly used to create a reproducible build-environment. The source code is bind-mounted into the image at runtime; there is no need to create an image with the actual source code, and copying the source code into the image would lead to a new image being created for each code-change (possibly leading up to many "dangling" images for previous code-changes).

However, when building (and using) the development images in CI, bind-mounting is not an option, because the daemon is running remotely.

To make this work, the circle-ci script patched the Dockerfiles when CI is run; adding a `COPY` to the respective Dockerfiles.

Patching Dockerfiles is not really a "best practice" and, even though the source code does not and up in the image, the source would still be _sent_ to the daemon for each build (unless BuildKit is used).

This patch updates the makefiles, circle-ci script, and Dockerfiles;

- When building the Dockerfiles locally, pipe the Dockerfile through stdin.
  Doing so, prevents the build-context from being sent to the daemon. This speeds
  up the build, and doesn't fill up the Docker "temp" directory with content that's
  not used
- Now that no content is sent, add the COPY instructions to the Dockerfiles, and
  remove the code in the circle-ci script to "live patch" the Dockerfiles.

Before this patch is applied (with cache):

```
$ time make -f docker.Makefile build_shell_validate_image
docker build -t docker-cli-shell-validate -f ./dockerfiles/Dockerfile.shellcheck .
Sending build context to Docker daemon     41MB
Step 1/2 : FROM    debian:stretch-slim
...
Successfully built 81e14e8ad856
Successfully tagged docker-cli-shell-validate:latest

2.75 real         0.45 user         0.56 sys
```

After this patch is applied (with cache)::

```
$ time make -f docker.Makefile build_shell_validate_image
cat ./dockerfiles/Dockerfile.shellcheck | docker build -t docker-cli-shell-validate -
Step 1/2 : FROM    debian:stretch-slim
...
Successfully built 81e14e8ad856
Successfully tagged docker-cli-shell-validate:latest

0.33 real         0.07 user         0.08 sys
```

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -16,9 +16,7 @@ jobs:
       - run:
           name: "Lint"
           command: |
-            dockerfile=dockerfiles/Dockerfile.lint
-            echo "COPY . ." >> $dockerfile
-            docker build -f $dockerfile --tag cli-linter:$CIRCLE_BUILD_NUM .
+            docker build -f dockerfiles/Dockerfile.lint --tag cli-linter:$CIRCLE_BUILD_NUM .
             docker run --rm cli-linter:$CIRCLE_BUILD_NUM
 
   cross:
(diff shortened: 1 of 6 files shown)


thaJeztah:
ping @ijc @vdemeester @dnephin PTAL

thaJeztah:
Let me update my commit message, as I swapped the duration. when copy/pasting the result time 😂

thaJeztah:
ping @silvin-lubecki @dnephin PTAL 🤗

thaJeztah:
Yes, I considered that solution; the downside of that solution is that the context is still sent, even though it's not consumed.

In this case, no context was needed at all, so I went for this option; the "FROM dev ..." approach could be used once we need it

thaJeztah:
ready for merge? 🤗

PR #1538, opened 2018-11-29
Use official shellcheck image

thaJeztah:
This patch switches the shellcheck image to use the official image from Docker Hub.

Note that this does not yet update shellcheck to the latest version (v0.5.x); Shellcheck v0.4.7 added some new checks, which makes CI currently fail, so will be done in a follow-up PR. Instead, the v0.4.6 version is used in this PR, which is closest to the same version as was installed in the image before this change;

```
docker run --rm docker-cli-shell-validate shellcheck --version
ShellCheck - shell script analysis tool
version: 0.4.4
license: GNU General Public License, version 3
website: http://www.shellcheck.net
```

### DIFF
--- a/dockerfiles/Dockerfile.shellcheck
+++ b/dockerfiles/Dockerfile.shellcheck
@@ -1,10 +1,7 @@
-FROM    debian:stretch-slim
-
-RUN     apt-get update && \
-        apt-get -y install make shellcheck && \
-        apt-get clean
-
+FROM    koalaman/shellcheck-alpine:v0.4.6
+RUN     apk add --no-cache bash make
 WORKDIR /go/src/github.com/docker/cli
 ENV     DISABLE_WARN_OUTSIDE_CONTAINER=1
-CMD     bash
+ENTRYPOINT [""]
+CMD     ["/bin/sh"]
 COPY    . .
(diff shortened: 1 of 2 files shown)


thaJeztah:
Ah; must've tried with `v0.4.6`, as that works; so difference is between shellcheck `v0.4.6` and `v0.4.7`; https://github.com/koalaman/shellcheck/compare/v0.4.6...v0.4.7

thaJeztah:
Right, and I see all the failures are due to three new checks that were added in shellcheck v0.4.7;

- SC2206/SC2207: Suggest alternatives when using word splitting in arrays
- SC1117: Warn about double quoted, undefined backslash sequences

```
In contrib/completion/bash/docker line 347:
		plugins=(${plugins[@]/$del/})
                         ^-- SC2206: Quote to prevent word splitting, or split robustly with mapfile or read -a.


In contrib/completion/bash/docker line 107:
	COMPREPLY=( $(compgen -W "$(__docker_configs "$@")" -- "$current") )
                    ^-- SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting).


In contrib/completion/bash/docker line 213:
				format+="$repo_format\n"
                                                     ^-- SC1117: Backslash is literal in "\n". Prefer explicit escaping: "\\n".
```
I'll update the PR to use `v0.4.6`, and we can look at fixing the new checks in a separate PR (and bump to v0.5.x)

PR #1069, opened 2018-05-18
Update Cobra and pflag, and use built-in --version feature

thaJeztah:
This updates cobra and pflag to tagged releases, and uses the new built-in `--version` feature

### Bump spf13/cobra to v0.0.3

Use a tagged release of Cobra

https://github.com/spf13/pflag/commit/97afa5e7ca8a08a383cb259e06636b5e2cc7897f...v0.0.3

Relevant changes:

- spf13/cobra#567 Add `CalledAs` method to cobra.Command
- spf13/cobra#580 Update error message for missing required flags
- spf13/cobra#584 Add support for --version flag
- spf13/cobra#614 If user has a project in symlink, just use its destination folder and work there
- spf13/cobra#649 terminates the flags when -- is found in commandline
- spf13/cobra#662 Add support for ignoring parse errors
- spf13/cobra#686 doc: hide hidden parent flags

Also various improvements were added for generating Bash
completion scripts (currently not used by us)

### Bump spf13/pflag to v1.0.1

Use a tagged release of this package

https://github.com/spf13/pflag/compare/97afa5e7ca8a08a383cb259e06636b5e2cc7897f...v1.0.1

Relevant changes:

- spf13/pflag#122 DurationSlice: implementation and tests
- spf13/pflag#115 Implement BytesHex type of argument 
- spf13/pflag#150 Add uintSlice and boolSlice to name prettifier
- spf13/pflag#155 Add multiline wrapping support
- spf13/pflag#158 doc: clarify difference between string slice vs. array
- spf13/pflag#160 add ability to ignore unknown flags
- spf13/pflag#163 Allow Users To Show Deprecated Flags

PR #1539, opened 2018-11-29
Fix yamldocs outputing `[flags]` in usage output

thaJeztah:
A similar change was made in the CLI itself in https://github.com/docker/cli/pull/1069, but is not inherited by the code that generates the YAML docs.

Before this patch is applied;

```
usage: docker container exec [OPTIONS] CONTAINER COMMAND [ARG...] [flags]
```

With this patch applied:

```
usage: docker container exec [OPTIONS] CONTAINER COMMAND [ARG...]
```

**- How to verify it**

run `make -f docker.Makefile yamldocs` and check the generated YAML files in `docs/yaml/gen/`


thaJeztah:
ping @tiborvass @vdemeester PTAL

/cc @seemethere @ahh-docker

PR #1558, opened 2018-12-10
Add a `fmt` build target and make use of a go build cache.

ijc:
A couple of (unrelated) tweaks to the build.

Firstly add a `make -f dockerfile.Makefile fmt` target which runs `gofmt` on all the non-vendored Go code as a developer convenience.

Secondly add (optional, but defaulted on) support for using a go build cache when using the `dockerfile.Makefile` targets. This speeds up repeated builds for me by 6x (over 1m to less than 10s), according to `time make build -f docker.Makefile DOCKER_BUILDKIT=1 GO_BUILD_CACHE=y` (or =n).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,10 @@ test: test-unit ## run tests
 test-coverage: ## run test coverage
 	./scripts/test/unit-with-coverage $(shell go list ./... | grep -vE '/vendor/|/e2e/')
 
+.PHONY: fmt
+fmt:
+	go list -f {{.Dir}} ./... | xargs gofmt -w -s -d
+
 .PHONY: lint
 lint: ## run all the lint tools
 	gometalinter --config gometalinter.json ./...
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -81,6 +81,10 @@ shell: dev ## alias for dev
 lint: build_linter_image ## run linters
 	docker run -ti $(ENVVARS) $(MOUNTS) $(LINTER_IMAGE_NAME)
 
+.PHONY: fmt
+fmt:
+	docker run --rm $(ENVVARS) $(MOUNTS) $(DEV_DOCKER_IMAGE_NAME) make fmt
+
 .PHONY: vendor
 vendor: build_docker_image vendor.conf ## download dependencies (vendor/) listed in vendor.conf
 	docker run -ti --rm $(ENVVARS) $(MOUNTS) $(DEV_DOCKER_IMAGE_NAME) make vendor
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -10,7 +10,12 @@ LINTER_IMAGE_NAME = docker-cli-lint$(IMAGE_TAG)
 CROSS_IMAGE_NAME = docker-cli-cross$(IMAGE_TAG)
 VALIDATE_IMAGE_NAME = docker-cli-shell-validate$(IMAGE_TAG)
 E2E_IMAGE_NAME = docker-cli-e2e$(IMAGE_TAG)
+GO_BUILD_CACHE ?= y
 MOUNTS = -v "$(CURDIR)":/go/src/github.com/docker/cli
+CACHE_VOLUME_NAME := docker-cli-dev-cache
+ifeq ($(GO_BUILD_CACHE),y)
+MOUNTS += -v "$(CACHE_VOLUME_NAME):/root/.cache/go-build"
+endif
 VERSION = $(shell cat VERSION)
 ENVVARS = -e VERSION=$(VERSION) -e GITCOMMIT -e PLATFORM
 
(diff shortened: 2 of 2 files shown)


silvin-lubecki:
I'm not sure about the `fmt` target, which I think should be an IDE configuration for all golangs projects.
But 👍 for the go build cache.

ijc:
The main benefit of the `fmt` target is it always uses the right version of `gofmt` from within the container (which matches the lint target etc), whereas an IDE might be picking up a different one from the host system.

docker.Makefile
```diff
@@ -10,7 +10,11 @@ LINTER_IMAGE_NAME = docker-cli-lint$(IMAGE_TAG)
 CROSS_IMAGE_NAME = docker-cli-cross$(IMAGE_TAG)
 VALIDATE_IMAGE_NAME = docker-cli-shell-validate$(IMAGE_TAG)
 E2E_IMAGE_NAME = docker-cli-e2e$(IMAGE_TAG)
+GO_BUILD_CACHE ?= y
 MOUNTS = -v "$(CURDIR)":/go/src/github.com/docker/cli
+ifeq ($(GO_BUILD_CACHE),y)
+MOUNTS += -v "$(CURDIR)/.go-build-cache:/root/.cache/go-build"
```
thaJeztah: Why not use an (anonymous) volume instead of a bind-mount?

Also wondering if we should start considering _not_ bind-mounting the source, but rebuild the image instead, with BuildKit being optimized for rebuilding. In that case, we can use the `RUN --mount` option for the cache;

```Dockerfile
# syntax=docker/dockerfile:1.0-experimental
RUN --mount=target=. --mount=type=cache,target=/go/src/github.com/docker/cli/.cache/go-build ....
```

ijc: > Why not use an (anonymous) volume instead of a bind-mount?

Won't an anonymous volume be freshly created for each build, rendering the cache useless? A named volume would work but do we want to be creating those here? A local directory seems simpler and easier to manage (e.g. the `clean` target can easily deal with it without the potential for getting confused by a users other volumes etc)

re buildkit, we should probably consider it but it would be a more major revamping of the build so should be a separate PR IMHO.

thaJeztah: > Won't an anonymous volume be freshly created for each build

Yes, it would; otoh (I think) the usecase here is to rebuild while inside the container, so the penalty for a "fresh" cache would only be once.

> A local directory seems simpler and easier to manage

Yes, and no; it would pollute my local machine

> without the potential for getting confused by a users other volumes etc

I guess if we carefully pick a name for the volume, that shouldn't be a big issue; here's what we do in moby/moby https://github.com/moby/moby/blob/master/Makefile#L75

```Makefile
DOCKER_MOUNT_CACHE := -v docker-dev-cache:/root/.cache
```

ijc: > Yes, it would; otoh (I think) the usecase here is to rebuild while inside the container, so the penalty for a "fresh" cache would only be once.

As I said in the intro the use case is running `make build -f docker.Makefile` repeatedly. It's good that it also helps with `make shell -f docker.Makefile` though.

I'll switch to use a named `docker-cli-dev-cache` volume.

docker.Makefile
```diff
@@ -10,7 +10,11 @@ LINTER_IMAGE_NAME = docker-cli-lint$(IMAGE_TAG)
 CROSS_IMAGE_NAME = docker-cli-cross$(IMAGE_TAG)
 VALIDATE_IMAGE_NAME = docker-cli-shell-validate$(IMAGE_TAG)
 E2E_IMAGE_NAME = docker-cli-e2e$(IMAGE_TAG)
+GO_BUILD_CACHE ?= y
 MOUNTS = -v "$(CURDIR)":/go/src/github.com/docker/cli
+ifeq ($(GO_BUILD_CACHE),y)
+MOUNTS += -v "docker-cli-dev-cache:/root/.cache/go-build"
```
thaJeztah: If you still want `make -f docker.Makefile clean`, that target could do a `docker volume rm docker-cli-dev-cache`

ijc:
Is this mergeable @thaJeztah @silvin-lubecki or are there other changes you'd like to see?

Makefile
```diff
@@ -21,6 +21,10 @@ test: test-unit ## run tests
 test-coverage: ## run test coverage
 	./scripts/test/unit-with-coverage $(shell go list ./... | grep -vE '/vendor/|/e2e/')
 
+.PHONY: fmt
+fmt:
+	go list -f {{.Dir}} ./... | xargs gofmt -w -s -d
```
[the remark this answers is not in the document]
thaJeztah: Interesting; wonder why it's needed in the other parts of the Makefile then 🤔. Maybe if they're running from inside a different directory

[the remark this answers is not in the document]
ijc: I believe it wasn't always the case with older versions of the go tooling. I suspect the two places which `grep` the `vendor` dir out are just out of date, but they both also want to exclude `e2e` so there isn't a whole lot of worthwhile cleanup to be done.

PR #1571, opened 2018-12-13
Fix warnings not being printed on "create", only on "run"

thaJeztah:
Previously, these errors were only printed when using `docker run`, but were omitted when using `docker container create` and `docker container start` separately.

Given that these warnings apply to both situations, this patch moves generation of these warnings to `docker container create` (which is also called by `docker run`)

### DIFF
--- a/cli/command/container/create.go
+++ b/cli/command/container/create.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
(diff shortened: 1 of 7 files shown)


cli/command/container/create.go
```diff
@@ -165,6 +166,9 @@ func createContainer(ctx context.Context, dockerCli command.Cli, containerConfig
 	networkingConfig := containerConfig.NetworkingConfig
 	stderr := dockerCli.Err()
 
+	warnOnOomKillDisable(*hostConfig, stderr)
+	warnOnLocalhostDNS(*hostConfig, stderr)
```
thaJeztah: FWIW, I actually want to move these warnings to the daemon (similar as to what I did in https://github.com/docker/cli/pull/1225 / https://github.com/moby/moby/pull/37502, that way the _daemon_ can decide if the container configuration has possible issues, and we can provide warnings for situations where (e.g.) memory-limit is not supported by the daemon, so will be ignored

thaJeztah:
Ah, boo! I disabled the automatic gofmt, because it keeps messing up my files when switching between repos haha

Issue #1576, opened 2018-12-14
feat: improves ValidateContextDirectory performance


orisano:
improves build.ValidateContextDirectory performance for dockerignore.
reduces regexp.Compile.

thaJeztah:
closing, as https://github.com/docker/cli/pull/1577 was merged

PR #1577, opened 2018-12-14
feat: improves ValidateContextDirectory performance

orisano:
for #1576 
reduces PatternMatcher construction.

### DIFF
--- a/cli/command/image/build/context.go
+++ b/cli/command/image/build/context.go
@@ -46,14 +46,6 @@ func ValidateContextDirectory(srcPath string, excludes []string) error {
 	if err != nil {
 		return err
 	}
-	matches := func(file string) (bool, error) {
-		file = filepath.Clean(file)
-		if file == "." {
-			// Don't let them exclude everything, kind of silly.
-			return false, nil
-		}
-		return pm.Matches(file)
-	}
 
 	return filepath.Walk(contextRoot, func(filePath string, f os.FileInfo, err error) error {
 		if err != nil {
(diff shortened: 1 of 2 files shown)


cli/command/image/build/context.go
```diff
@@ -41,6 +41,20 @@ func ValidateContextDirectory(srcPath string, excludes []string) error {
...
 	}
+
+	pm, err := fileutils.NewPatternMatcher(excludes)
+	if err != nil {
+		return err
+	}
+	matches := func(file string) (bool, error) {
+		file = filepath.Clean(file)
+		if file == "." {
+			// Don't let them exclude everything, kind of silly.
```
thaJeztah: I'm not sure we can ignore this case; I know there's repositories using this (i.e. have `.dockerignore` with `.` in it, to not send anything)

orisano: https://github.com/docker/cli/blob/master/vendor/github.com/docker/docker/pkg/fileutils/fileutils.go#L217-L232
it is original implementation.

thaJeztah: Hm... interesting. I definitely have seen people using this in their `.dockerignore`. Wondering if it worked at all for them (haven't tried it)

silvin-lubecki: I think we should add a test case here, or at least copy the tests from https://github.com/moby/moby/blob/master/pkg/fileutils/fileutils_test.go#L221 (as the code was copied too), and maybe it will need to extract this inlined function.

orisano: i add test cases

silvin-lubecki:
@thaJeztah PTAL 🦁

orisano:
sorry, what is the current status of this pr?

PR #1581, opened 2018-12-17
Don't use TLS for socket connections

thaJeztah:
fixes https://github.com/moby/moby/issues/36535

### DIFF
--- a/opts/hosts.go
+++ b/opts/hosts.go
@@ -84,6 +84,19 @@ func parseDockerDaemonHost(addr string) (string, error) {
 	}
 }
 
+// IsSocket checks if the given address is a Unix-socket (linux) or named pipe (Windows)
+func IsSocket(addr string) bool {
+	addrParts := strings.SplitN(addr, "://", 2)
+	if len(addrParts) == 1 && addrParts[0] != "" {
+		addrParts = []string{"tcp", addrParts[0]}
+	}
+	switch addrParts[0] {
+	case "unix", "npipe", "fd":
+		return true
+	}
+	return false
+}
+
 // parseSimpleProtoAddr parses and validates that the specified address is a valid
 // socket address for simple protocols like unix and npipe. It returns a formatted
 // socket address, either using the address parsed from addr, or the contents of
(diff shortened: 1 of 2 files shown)


thaJeztah:
@ijc @justincormack

opts/hosts.go
```diff
@@ -84,6 +84,19 @@ func parseDockerDaemonHost(addr string) (string, error) {
 	}
 }
 
+// IsSocket checks if the given address is a Unix-socket (linux) or named pipe (Windows)
+func IsSocket(addr string) bool {
+	addrParts := strings.SplitN(addr, "://", 2)
```
ijc: Is this not a poor man's reimplementation of `net.Parse()`?

It also reimplements the head of `parseDockerDaemonHost` (immediately above) which also then switches on the scheme. Perhaps `parseDockerDaemonHost` function should return some sort of `tls` indicator, which is plumbed up through the `getServerHost`→`ParseHost`→`parseDockerDaemonHost` call chain (which seems fairly linear, so not too much collateral change required) for use in `NewAPIClientFromFlags`?

Perhaps that plumbing could take the form of returning the appropriate `[]clientOpts` to be used?

thaJeztah: > Is this not a poor man's reimplementation of `net.Parse()`?

Yes, it's similar to what `url.Parse()` provides; not sure if we need all the functionality of `url.Parse()` for this though.

> It also reimplements the head of `parseDockerDaemonHost` (immediately above)

Yes, I was in doubt if I would extract it to a function (e.g. `getScheme(host)`), but that felt a bit over the top for two lines of code; happy to change though.


> Perhaps that plumbing could take the form of returning the appropriate []clientOpts to be used?

Looking if we should integrate this into `client.WithHost(host)`. i.e., if I set a `unix`, `npipe`, or `fd` host, then disable TLS. wdyt?

ijc: I agree with `client.WithHost(host)` doing the right thing, whether that means doing it there directly or further down the callchain I have no string opinion on.

opts/hosts.go
```diff
@@ -84,6 +84,19 @@ func parseDockerDaemonHost(addr string) (string, error) {
 	}
 }
 
+// IsSocket checks if the given address is a Unix-socket (linux) or named pipe (Windows)
+func IsSocket(addr string) bool {
```
silvin-lubecki: Can we add a test on this function?

vvoland (approved):

PR #1588, opened 2018-12-20
Add quiet flag to bash completion for docker pull

mdsouza2:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -2803,7 +2803,7 @@ _docker_image_pull() {
 
 	case "$cur" in
 		-*)
-			local options="--all-tags -a --disable-content-trust=false --help"
+			local options="--all-tags -a --disable-content-trust=false --help --quiet -q"
 			__docker_server_is_experimental && options+=" --platform"
 
 			COMPREPLY=( $( compgen -W "$options" -- "$cur" ) )


tiborvass:
@mdsouza2 Thank you for your contribution!

I believe the title of the PR can be confusing: it does not add the functionality, simply adds it to the bash autocompletion script.

Also it is good practice to add a description, to help reviewers. For instance you could reference the PR that added the functionality.

contrib/completion/bash/docker
```diff
@@ -2803,7 +2803,7 @@ _docker_image_pull() {
 
 	case "$cur" in
 		-*)
-			local options="--all-tags -a --disable-content-trust=false --help"
+			local options="--all-tags -a --quiet -q --disable-content-trust=false --help"
```
tiborvass: nitpicking (not super important): I could be wrong, but as far as I can tell, most of the flags in this script are ordered alphabetically, so I think it would be good to keep the list in alphabetical order.

thaJeztah: Yes, I believe they are kept in alphabetical order 👍

albers: Please update the argument order.

albers:
Note: The corresponding feature was added in #882

thaJeztah:
@albers I'll go ahead and merge, because I think your comment was addressed 👍

PR #1589, opened 2018-12-21
docs(metrics-addr): Use port 9323, allocated for Docker in prometheus

fhemberger:
According to https://github.com/prometheus/prometheus/wiki/Default-port-allocations, `9323` is the default port associated with Docker metrics. It should be used in the example as well to avoid assigning a random port which may interfere with other prometheus exporters on the same machine.

### DIFF
--- a/docs/reference/commandline/dockerd.md
+++ b/docs/reference/commandline/dockerd.md
@@ -1231,9 +1231,13 @@ The `--metrics-addr` option takes a tcp address to serve the metrics API.
 This feature is still experimental, therefore, the daemon must be running in experimental
 mode for this feature to work.
 
-To serve the metrics API on localhost:1337 you would specify `--metrics-addr 127.0.0.1:1337`
-allowing you to make requests on the API at `127.0.0.1:1337/metrics` to receive metrics in the
+To serve the metrics API on `localhost:9323` you would specify `--metrics-addr 127.0.0.1:9323`,
+allowing you to make requests on the API at `127.0.0.1:9323/metrics` to receive metrics in the
 [prometheus](https://prometheus.io/docs/instrumenting/exposition_formats/) format.
+
+Port `9323` is the [default port associated with Docker
+metrics](https://github.com/prometheus/prometheus/wiki/Default-port-allocations)
+to avoid collisions with other prometheus exporters and services.
 
 If you are running a prometheus server you can add this address to your scrape configs
 to have prometheus collect metrics on Docker.  For more information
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/dockerd.md
```diff
@@ -1231,10 +1231,12 @@ The `--metrics-addr` option takes a tcp address to serve the metrics API.
 This feature is still experimental, therefore, the daemon must be running in experimental
 mode for this feature to work.
 
-To serve the metrics API on localhost:1337 you would specify `--metrics-addr 127.0.0.1:1337`
-allowing you to make requests on the API at `127.0.0.1:1337/metrics` to receive metrics in the
+To serve the metrics API on localhost:9323 you would specify `--metrics-addr 127.0.0.1:9323`
+allowing you to make requests on the API at `127.0.0.1:9323/metrics` to receive metrics in the
 [prometheus](https://prometheus.io/docs/instrumenting/exposition_formats/) format.
 
+Note that port 9323 is the [default port associated](https://github.com/prometheus/prometheus/wiki/Default-port-allocations) with Docker metrics to avoid collisions with other prometheus exporters and services.
```
thaJeztah: Can you wrap this to ~80 chars? Also, I think we can drop the "Note" (we generally try to reserve "Note" for things to call out specifically). Something like;

```suggestion
Port `9323` is the [default port associated with Docker
metrics](https://github.com/prometheus/prometheus/wiki/Default-port-allocations)
to avoid collisions with other prometheus exporters and services.
```

docs/reference/commandline/dockerd.md
```diff
@@ -1231,10 +1231,12 @@ The `--metrics-addr` option takes a tcp address to serve the metrics API.
 This feature is still experimental, therefore, the daemon must be running in experimental
 mode for this feature to work.
 
-To serve the metrics API on localhost:1337 you would specify `--metrics-addr 127.0.0.1:1337`
-allowing you to make requests on the API at `127.0.0.1:1337/metrics` to receive metrics in the
+To serve the metrics API on localhost:9323 you would specify `--metrics-addr 127.0.0.1:9323`
```
thaJeztah: Looks like this was missed when this was added originally; can you put back-ticks around `localhost:9323` ?

```suggestion
To serve the metrics API on `localhost:9323` you would specify `--metrics-addr 127.0.0.1:9323`
```

fhemberger:
@thaJeztah Thanks for your feedback, updated the PR. PTAL.

fhemberger:
You're welcome!

PR #1578, opened 2018-12-17
Fixed wording per https://github.com/docker/docker.github.io/issues/7281

ahh-docker:
Signed-off-by: Anne Henmi <anne.henmi@docker.com>

**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #1590, opened 2018-12-23
Docs: add note about management commands targeting a manager

thaJeztah:
Alternative for https://github.com/docker/cli/pull/1578

~~closes https://github.com/docker/cli/pull/1578~~ already merged

### DIFF
--- a/docs/reference/commandline/stack_deploy.md
+++ b/docs/reference/commandline/stack_deploy.md
@@ -37,8 +37,12 @@ Options:
 
 ## Description
 
-Create and update a stack from a `compose` file on the swarm. This command has to
-be run targeting a manager node.
+Create and update a stack from a `compose` file on the swarm.
+
+> **Note**: This is a cluster management command. When using swarm as an orchestrator,
+> this command must be executed on a swarm manager node. To learn about managers
+> and workers, refer to the [Swarm mode section](https://docs.docker.com/engine/swarm/)
+> in the documentation.
 
 ## Examples
 
--- a/docs/reference/commandline/node_demote.md
+++ b/docs/reference/commandline/node_demote.md
@@ -27,9 +27,11 @@ Options:
 
 ## Description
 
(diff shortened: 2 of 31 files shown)


docs/reference/commandline/stack_deploy.md
```diff
@@ -38,8 +38,11 @@ Options:
 
 ## Description
 
-Create and update a stack from a `compose` or a `dab` file on the swarm. This command
-has to be run targeting a manager node.
+Create and update a stack from a `compose` or a `dab` file on the swarm.
+
+> **Note**: This is a cluster management command, and must be executed on a swarm
```
thaJeztah: Wasn't sure how to deal with k8s as an orchestrator here; this command can be run both on Swarm or Kubernetes, so we may need a different description to either make it generic enough to fit both use-cases, or extend the description to explicitly call out both use-cases

thaJeztah:
This is a quick approach for this, but I think we can add annotations for this, so that;

- it can be used to automatically show/hide commands if the node is not a manager
- it can be included in the generated YAML docs, so that we can add this information automatically when generating the documentation; similar to the API-version and Orchestrator annotations;

screen shot 2018-12-23 at 12 33 00

thaJeztah:
@ahh-docker @vdemeester

ahh-docker:
LGTM

docs/reference/commandline/stack_deploy.md
silvin-lubecki: Maybe add a something like: ```**Note**:If you use swarm as an orchestrator, this is a cluster management command...``` ?

SvenDowideit:
mmm, LGTM - though i'm saddened that I didn't build something to let you transclude the note from one place :D

thaJeztah:
@silvin-lubecki slightly changed the wording for `docker stack` commands, to explicitly mention that a manager node must be targeted when using "swarm" as orchestrator.

PR #1599, opened 2019-01-04
Update CircleCI Docker version to 18.09.3

thaJeztah:
18.03 has reached EOL; let's use a current version in CI as well

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -8,9 +8,9 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker:
-            version: 18.03.1-ce
-            reusable: true
-            exclusive: false
+          version: 18.03.1-ce
+          reusable: true
+          exclusive: false
       - run:
           command: docker version
       - run:
@@ -26,9 +26,9 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker:
-            version: 18.03.1-ce
-            reusable: true
-            exclusive: false
+          version: 18.03.1-ce
+          reusable: true
+          exclusive: false
       - run:
           name: "Cross"
           command: |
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -4,13 +4,13 @@ jobs:
...
     working_directory: /work
-    docker: [{image: 'docker:18.03-git'}]
+    docker: [{image: 'docker:18.09.0-git'}]
     steps:
       - checkout
       - setup_remote_docker:
-            version: 18.03.1-ce
-            reusable: true
-            exclusive: false
+          version: 18.09.0
```
thaJeztah: Looks like the Circle-CI images did not yet update the version-scheme; the Docker versions no longer have a `-ce` / `-ee` suffix 

```
Allocating a remote Docker Engine
failed to create host: rpc error: code = Unknown desc = image docker-18.09.0 is not supported
```

/cc @dnephin perhaps they should have an alias?

thaJeztah:
oh... no 18.09.0-ce either 🤔 perhaps without `.0` (

```
Allocating a remote Docker Engine
failed to create host: rpc error: code = Unknown desc = image docker-18.09.0-ce is not supported
```

Interesting; the documentation (https://circleci.com/docs/2.0/building-docker-images/#docker-version) refers to the static linux packages;

screen shot 2019-01-04 at 22 14 44


Which then "should" be able to use `docker-18.09.0` as name

screen shot 2019-01-04 at 22 13 45


Perhaps it's because the latest version dropped the `docker-` prefix from some binaries? (i.e., it's now `runc`, `containerd` instead of `docker-runc`, `docker-containerd`)

dnephin:
There is no 18.09 image yet for `setup_remote_docker`, `docker-18.06.0-ce` is the latest.

thaJeztah:
Ah, that's a pity. No 18.06.1 either?

thaJeztah:
related discussion on the CircleCI forum; https://discuss.circleci.com/t/docker-18-09-support-for-setup-remote-docker/26731

thaJeztah:
18.09.3 is now available (no more current versions yet, but at least we should now also be able to use BuildKit where needed; I'll have a look at that after

thaJeztah:
I'll go ahead and merge

PR #1604, opened 2019-01-07
Makefile use --rm ("autoremove") for all targets

thaJeztah:
These containers are stateless, so can be removed after exiting

### DIFF
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -82,7 +82,7 @@ binary-osx: build_cross_image ## build the CLI for macOS
 
 .PHONY: dev
 dev: build_docker_image ## start a build container in interactive mode for in-container development
-	docker run -ti $(ENVVARS) $(MOUNTS) \
+	docker run -ti --rm $(ENVVARS) $(MOUNTS) \
 		-v /var/run/docker.sock:/var/run/docker.sock \
 		$(DEV_DOCKER_IMAGE_NAME) ash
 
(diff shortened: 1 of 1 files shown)


docker.Makefile
```diff
@@ -124,15 +124,15 @@ test-e2e: test-e2e-non-experimental test-e2e-experimental test-e2e-connhelper-ss
...
+	docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -e DOCKERD_EXPERIMENTAL=1 $(E2E_IMAGE_NAME)
 
 .PHONY: test-e2e-non-experimental
 test-e2e-non-experimental: build_e2e_image
 	docker run --rm -v /var/run/docker.sock:/var/run/docker.sock $(E2E_IMAGE_NAME)
 
 .PHONY: test-e2e-connhelper-ssh
 test-e2e-connhelper-ssh: build_e2e_image
-	docker run -e TEST_CONNHELPER=ssh -e DOCKERD_EXPERIMENTAL=1 --rm -v /var/run/docker.sock:/var/run/docker.sock $(E2E_IMAGE_NAME)
+	docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -e DOCKERD_EXPERIMENTAL=1 -e TEST_CONNHELPER=ssh $(E2E_IMAGE_NAME)
```
thaJeztah: no change here; other than change the order of arguments to be the same as the targets above

docker.Makefile
```diff
@@ -124,15 +124,15 @@ test-e2e: test-e2e-non-experimental test-e2e-experimental test-e2e-connhelper-ss
 
 .PHONY: test-e2e-experimental
 test-e2e-experimental: build_e2e_image
-	docker run -e DOCKERD_EXPERIMENTAL=1 --rm -v /var/run/docker.sock:/var/run/docker.sock $(E2E_IMAGE_NAME)
+	docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -e DOCKERD_EXPERIMENTAL=1 $(E2E_IMAGE_NAME)
```
thaJeztah: no change here; other than change the order of arguments to be the same as the targets above

PR #1606, opened 2019-01-07
Add --device support for Windows

jhowardmsft:

Adds support for --device in Windows. This must take the form of: --device='class/clsid'. See https://blogs.technet.microsoft.com/virtualization/2018/08/13/bringing-device-support-to-windows-server-containers/ for more info. The server-side fix was merged in https://github.com/moby/moby/pull/37638 back in November 2018.

@thaJeztah As discussed offline. Annoyingly, due to the code pattern having flag validation happening before the _ping is sent to the daemon, I had to move the validation in the Linux case to during parsing. i.e. after the server OS is known. Fortunately, that avoided the need for a global as was the option on the table before. If (at a later time), the device can be validated entirely daemon side for Linux, it will greatly simplify things.

This replaces https://github.com/docker/cli/pull/1290. @jterry75 FYI.  You should be able to close the previous PR.

### DIFF
--- a/cli/command/container/create.go
+++ b/cli/command/container/create.go
@@ -72,7 +72,7 @@ func runCreate(dockerCli command.Cli, flags *pflag.FlagSet, options *createOptio
 		}
 	}
 	copts.env = *opts.NewListOptsRef(&newEnv, nil)
-	containerConfig, err := parse(flags, copts)
+	containerConfig, err := parse(flags, copts, dockerCli.ServerInfo().OSType)
 	if err != nil {
 		reportError(dockerCli.Err(), "create", err.Error(), true)
 		return cli.StatusError{StatusCode: 125}
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -141,7 +141,7 @@ func addFlags(flags *pflag.FlagSet) *containerOptions {
 		deviceReadIOps:    opts.NewThrottledeviceOpt(opts.ValidateThrottleIOpsDevice),
 		deviceWriteBps:    opts.NewThrottledeviceOpt(opts.ValidateThrottleBpsDevice),
 		deviceWriteIOps:   opts.NewThrottledeviceOpt(opts.ValidateThrottleIOpsDevice),
-		devices:           opts.NewListOpts(validateDevice),
+		devices:           opts.NewListOpts(nil), // devices can only be validated after we know the server OS
 		env:               opts.NewListOpts(opts.ValidateEnv),
 		envFile:           opts.NewListOpts(nil),
 		expose:            opts.NewListOpts(nil),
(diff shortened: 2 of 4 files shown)


jcoliz:
I have validated this change against https://github.com/Microsoft/Windows-iotcore-samples/blob/develop/Samples/EdgeModules/SerialWin32/CS/ and https://github.com/jcoliz/Windows-iotcore-samples/tree/jcoliz/squeezenet/Samples/EdgeModules/SqueezeNetObjectDetection/cs . This uses [COMPORT](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/guid-devinterface-comport), [DISPLAY_ADAPTER](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/guid-devinterface-display-adapter), and [VIDEO_CAMERA](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/kscategory-video-camera).

jhowardmsft:
@thaJeztah Could you ping some relevant folks to look at this?

cli/command/container/opts.go
```diff
@@ -812,14 +842,23 @@ func validDeviceMode(mode string) bool {
 }
 
 // validateDevice validates a path for devices
+func validateDevice(val string, serverOS string) (string, error) {
+	switch serverOS {
+	case "linux":
+		return validateLinuxPath(val, validDeviceMode)
+	case "windows":
+		// Windows does validation entirely server-side
```
thaJeztah: 👍 We should remove more of these validations from the client, and defer as much as possible to the daemon/api. Doing so is much more maintainable, and makes sure that the _daemon_ is in charge as to "what's valid" and "what not" (which may differ between daemon versions). I'll start working on an epic for that

cli/command/container/create.go
```diff
@@ -60,7 +60,7 @@ func NewCreateCommand(dockerCli command.Cli) *cobra.Command {
 }
 
 func runCreate(dockerCli command.Cli, flags *pflag.FlagSet, opts *createOptions, copts *containerOptions) error {
-	containerConfig, err := parse(flags, copts)
+	containerConfig, err := parse(flags, copts, dockerCli.ServerInfo().OSType)
```
thaJeztah: For other reviewers; I was discussing some options with @jhowardmsft - it's a bit ugly, but we should (see my other comment) move more of these platform-specific validations out of the CLI. The cli should limit validation to "well-formedness" of values (e.g. if we expect an integer; validate it's an integer), but (if possible) nothing further than that.

thaJeztah:
@jhowardmsft as a follow-up, could you help write a short example for the reference documentation? https://github.com/docker/cli/blob/master/docs/reference/commandline/run.md#add-host-device-to-container---device

I'm not exactly sure what the format looks like for adding a device on Windows (or when using LCOW), so if you could show an example, that would be great! (I see there's some examples in the blog post at https://blogs.technet.microsoft.com/virtualization/2018/08/13/bringing-device-support-to-windows-server-containers/), perhaps you can take some of that

jterry75:
LGTM (not a maintainer)

thaJeztah:
ping @docker/core-cli-maintainers ptal

thaJeztah:
@jhowardmsft needs a rebase


@silvin-lubecki PTAL

jhowardmsft:
@thaJeztah Yes, that's rebased now.

jhowardmsft:
@silvin-lubecki PTAL

srinathck:
Any update on When Docker CLI v19.03.0 will be out?

thaJeztah:
dates are mentioned here; https://github.com/docker/docker-ce/milestone/32

PR #1607, opened 2019-01-08
Engine CLI 404

paigehargrave:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/commandline/dockerd.md
+++ b/docs/reference/commandline/dockerd.md
@@ -3,6 +3,8 @@ title: "dockerd"
 aliases: ["/engine/reference/commandline/daemon/"]
 description: "The daemon command description and usage"
 keywords: "container, daemon, runtime"
+redirect_from:
+- /v1.11/engine/reference/commandline/daemon/
 ---
 
 <!-- This file is maintained within the docker/cli GitHub
--- a/docs/reference/commandline/run.md
+++ b/docs/reference/commandline/run.md
@@ -2,6 +2,8 @@
 title: "run"
 description: "The run command description and usage"
 keywords: "run, command, container"
+redirect_from: 
+- /v1.11/engine/reference/commandline/run/
 ---
 
 <!-- This file is maintained within the docker/cli GitHub
--- a/docs/reference/builder.md
+++ b/docs/reference/builder.md
@@ -3,6 +3,7 @@ description: "Dockerfiles use a simple DSL which allows you to automate the step
 keywords: "builder, docker, Dockerfile, automation, image creation"
 redirect_from:
 - /reference/builder/
+- /engine/reference/commandline/builder/
 ---
 
 <!-- This file is maintained within the docker/cli GitHub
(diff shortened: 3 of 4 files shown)


docs/reference/commandline/dockerd.md
```diff
@@ -3,6 +3,8 @@ title: "dockerd"
 aliases: ["/engine/reference/commandline/daemon/"]
 description: "The daemon command description and usage"
 keywords: "container, daemon, runtime"
+redirect_from:
+- /v1.11/engine/reference/commandline/daemon/
```
thaJeztah: Could you explain this redirect? Because the archive still exists at this location; https://docs.docker.com/v1.11/engine/reference/commandline/daemon/

docs/reference/commandline/run.md
```diff
@@ -2,6 +2,8 @@
 title: "run"
 description: "The run command description and usage"
 keywords: "run, command, container"
+redirect_from: 
+- /v1.11/engine/reference/commandline/run/
```
thaJeztah: Same for this one

https://docs.docker.com/v1.11/engine/reference/commandline/run/

docs/reference/commandline/dockerd.md
JustinINevill: Our policy is to document all currently supported versions, plus the most recently EOL'd version (to support upgrades from that version), therefore we are removing all docs content for versions old than 17.03 to reduce the associated maintenance burden and customer confusion. These pages account for < 0.75% of current site traffic and are already not indexed by search engine. However out of an abundance of caution we are implementing a handful of redirects for the most popular pages within this <0.75%.  This PR implements those redirect for those pages.

docs/reference/commandline/run.md
JustinINevill: Same answer as above.

docs/reference/commandline/dockerd.md
thaJeztah: Have we tested if this actually works? Because I think it won't;

- These markdown files are not used for docs.docker.com (they have been replaced by stubs; https://github.com/docker/docker.github.io/tree/master/engine/reference/commandline that are propagated with the generated YAML files (see https://github.com/docker/docker.github.io/pull/7753)
- _If_ the redirect is created, we should check if it is not overwritten by the archive itself (https://github.com/docker/docker.github.io/blob/master/Dockerfile#L69-L83)


Also, once the "current" docs will be archived, possibly this redirect will now become;

```
/v18.09/v1.11/engine/reference/commandline/daemon/
```

Which will become a maintenance nightmare (also; what about version `/v1.12/`, `/v1.13/`)?

Would it be an alternative to, once we decide to no longer publish archives for a specific version, to replace an archive `/vYY.MM/*`  to redirect to `/` ? (or show a landing-page, informing the user that that specific archive is no longer published?)

JustinINevill:
We believe we can accomplish the same outcome in a more simple way. @paigehargrave can you go ahead and close this?

paigehargrave:
Closing this based on previous conversations.

PR #1610, opened 2019-01-11
Remove myself from codeowners 😅

vdemeester:
Still maintaining but most likely shouldn't be see as code owner :sweat_smile: 


olljanat:
Side comment but it really sounds that some new recruitment to Docker side would be good idea.
There is lot of open PRs on here which are not going forward because no one have time to look them.

vdemeester:
cc @thaJeztah

PR #1611, opened 2019-01-11
Make child resource creation atomic when creating a k8s stack

simonferquel:
**- What I did**

I made sure that configs and secrets created with the CLI are
- correctly removed if the stack itself can't be created (validation error,...)
- have their owner reference set to the stack when the stack is correctly created

**- How I did it**

- added a childResource interface, with setOwner and delete functions
- keep track of all created childResources

**- How to verify it**

The PR comes with unit tests


simonferquel:
@silvin-lubecki @vdemeester PTAL :)

cli/command/stack/kubernetes/deploy.go
```diff
@@ -81,6 +74,29 @@ func RunDeploy(dockerCli *KubeCli, opts options.Deploy, cfg *composetypes.Config
...
+	}
+
+	cr, err = stack.createFileBasedSecrets(secrets)
+	childResources = append(childResources, cr...) // make sure we collect childresources already created in case of failure
+	if err != nil {
+		deleteChildResources(childResources)
+		return err
+	}
+
+	if err = stacks.CreateOrUpdate(stack, childResources); err != nil {
```
silvin-lubecki: nit: `return stacks.CreateOrUpdate(stack, childResources)` ? 😸

cli/command/stack/kubernetes/deploy_test.go
```diff
@@ -0,0 +1,219 @@
...
+		secrets))
+	c, err := configs.Get("test", metav1.GetOptions{})
+	assert.NilError(t, err)
+	assert.Equal(t, len(c.OwnerReferences), 1)
+	assert.Equal(t, c.OwnerReferences[0].Name, "test")
+	assert.Equal(t, c.OwnerReferences[0].Kind, "Stack")
+	assert.Equal(t, c.OwnerReferences[0].APIVersion, v1beta1.SchemeGroupVersion.String())
+	s, err := secrets.Get("test", metav1.GetOptions{})
+	assert.NilError(t, err)
+	assert.Equal(t, len(s.OwnerReferences), 1)
```
silvin-lubecki: I think you can factorize the check on OwnerReferences with the next test.

cli/command/stack/kubernetes/deploy.go
```diff
@@ -81,6 +74,29 @@ func RunDeploy(dockerCli *KubeCli, opts options.Deploy, cfg *composetypes.Config
 
 }
 
+func createResources(stack Stack, stacks StackClient, configMaps corev1.ConfigMapInterface, secrets corev1.SecretInterface) error {
+	var childResources []childResource
+
+	cr, err := stack.createFileBasedConfigMaps(configMaps)
+	childResources = append(childResources, cr...) // make sure we collect childresources already created in case of failure
+	if err != nil {
+		deleteChildResources(childResources)
```
silvin-lubecki: I'm not a fan of this deleter, I wonder if `createFileBasedConfigMaps` and `createFileBasedSecrets` could return the deleter itself instead?

simonferquel:
Fakes are a VERY useful way to test integration with the Kubernetes API, which let us use the existing client-go abstractions without having to write our own fakes. There is a tradeoff there: vendoring the whole thing makes test setup far easier (and cleaner). We could also partially vendor fakes for supporting only the Core API (configs and secrets are in the Core API), but that would require some code change and a more complex test setup. We could also don't vendor the fakes and write our own (but I strongly discourage it).

cli/command/stack/kubernetes/deploy.go
simonferquel: tried, but we still need childResources for setting owners, and that made the code uglier :(

PR #1619, opened 2019-01-17
Bash completion for `docker context` command family

albers:
Ref: #1501


albers:
@simonferquel Done.
* Added completion for the global `--context` option.
* If `--context` is present, its value will be passed on to all `docker` invocations done by the completion script (like `--host`).

simonferquel:
That seems good, I can't test it right now though :/ ping @vdemeester or @thaJeztah

PR #1623, opened 2019-01-19
[Fish completion] Enrich description for container/image candidates

tmshn:
**- What I did**

For container/image candidates of fish completion, enriched description to include more info rather than just showing "Container" or "Image"

before
↓
after

**- How I did it**

On fish completion, description can be added not only by `-d` option of `complete` command, but also by appending tab-separeted field for each candidate line.

> each argument may optionally have a tab character followed by the argument description

ref: `man complete` https://fishshell.com/docs/current/commands.html#complete

So I fixed the functions `__fish_print_docker_containers` and `__fish_print_docker_images` to include information of candidates, and remove `-d` option.

**- How to verify it**

1. Install fish
2. Install this completion
    - Just put this `docker.fish` file under `~/.config/fish/completions/`
3. Try completion 
    - E.g: `docker run `<kbd>TAB</kbd> or `docker stop `<kbd>TAB</kbd> etc

cli fish completion: enriched containers/images description

tiger

Photo by [@africansafari1](https://twitter.com/africansafari1/status/1070521494541717505), a zoo in Japan

### DIFF
--- a/contrib/completion/fish/docker.fish
+++ b/contrib/completion/fish/docker.fish
@@ -69,7 +69,7 @@ function __fish_docker_subcommand_path_without --description 'Test if command ha
 end
 
 function __fish_print_docker_images --description 'Print a list of docker images'
-    docker images --format "{{.Repository}}:{{.Tag}}" | command grep -v '<none>'
+    docker images --format "{{.Repository}}:{{.Tag}}\tImage created {{.CreatedSince}}, {{.Size}}" | command grep -v '<none>'
 end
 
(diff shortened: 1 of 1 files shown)


contrib/completion/fish/docker.fish
```diff
@@ -26,12 +26,14 @@ end
 function __fish_print_docker_containers --description 'Print a list of docker containers' -a select
     switch $select
```
krobelus, line 27: This looks great; just one nitpick: there should be a `set -l filter` here to declare `filter` as local variable, otherwise we might use a global variable with the same name.

PR #1628, opened 2019-01-21
Context switch - Plugin and Docker Desktop integration

simonferquel:
**- What I did**

- for plugins:
  - Expose the Context Store config
  - Make an existing config modifiable (to add endpoint types)
- for Docker Desktop:
  - Expose all context operations (except listing, which will be done by directly using context store functions)

**- How I did it**

- Just exposed a few existing symbols.
- Added a SetEndpoint function on store.Config

**- How to verify it**

- Modified code covered by unit tests
- Added a test for SetEndpoint, making sure it applied to already created context stores

ping @ijc for the plugin related part
ping @gtardif @pgayvallet @guillaumerose @ebriney for the Docker Desktop part

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -81,13 +81,8 @@ type DockerCli struct {
 	contextStore          store.Store
 	currentContext        string
 	dockerEndpoint        docker.Endpoint
+	contextStoreConfig    store.Config
 }
-
-var storeConfig = store.NewConfig(
-	func() interface{} { return &DockerContext{} },
-	store.EndpointTypeGetter(docker.DockerEndpoint, func() interface{} { return &docker.EndpointMeta{} }),
-	store.EndpointTypeGetter(kubcontext.KubernetesEndpoint, func() interface{} { return &kubcontext.EndpointMeta{} }),
-)
 
 // DefaultVersion returns api.defaultVersion or DOCKER_API_VERSION if specified.
 func (cli *DockerCli) DefaultVersion() string {
@@ -184,7 +179,7 @@ func (cli *DockerCli) RegistryClient(allowInsecure bool) registryclient.Registry
 func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
 	var err error
-	cli.contextStore = store.New(cliconfig.ContextStoreDir(), storeConfig)
+	cli.contextStore = store.New(cliconfig.ContextStoreDir(), cli.contextStoreConfig)
 	cli.currentContext, err = resolveContextName(opts.Common, cli.configFile, cli.contextStore)
 	if err != nil {
 		return err
@@ -226,7 +221,7 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 
 // NewAPIClientFromFlags creates a new APIClient from command line flags
 func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.ConfigFile) (client.APIClient, error) {
-	store := store.New(cliconfig.ContextStoreDir(), storeConfig)
+	store := store.New(cliconfig.ContextStoreDir(), defaultContextStoreConfig())
 	contextName, err := resolveContextName(opts, configFile, store)
 	if err != nil {
 		return nil, err
@@ -372,7 +367,7 @@ func (cli *DockerCli) StackOrchestrator(flagValue string) (Orchestrator, error) 
 	if currentContext != "" {
 		contextstore := cli.contextStore
 		if contextstore == nil {
-			contextstore = store.New(cliconfig.ContextStoreDir(), storeConfig)
+			contextstore = store.New(cliconfig.ContextStoreDir(), cli.contextStoreConfig)
 		}
 		ctxRaw, err := contextstore.GetContextMetadata(currentContext)
 		if store.IsErrContextDoesNotExist(err) {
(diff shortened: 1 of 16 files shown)


cli/command/cli.go
```diff
@@ -79,7 +79,9 @@ type DockerCli struct {
 	dockerEndpoint        docker.Endpoint
 }
 
-var storeConfig = store.NewConfig(
+// ContextStoreConfig contains the config used by the context store
+// Plugins init functions could modify it to define other typed endpoint metadata
```
ijc: Nit: "A plugin's init functions may modify..."

ijc: Less of a nit, but I think the sentence should end with something like "...by calling `ContextStoreConfig.SetEndpoint()`".

cli/command/context/create.go
```diff
@@ -65,7 +65,8 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	return cmd
 }
 
-func runCreate(cli command.Cli, o *createOptions) error {
+// RunCreate creates a Docker context
+func RunCreate(cli command.Cli, o *createOptions) error {
```
ijc: I don't see a way for an external caller of `RunCreate` to produce a `createOptions`. OTOH the same seems to be true of e.g. `RunPush` and `RunSave` so maybe thesre's something I'm missing...

cli/command/context/export.go
```diff
@@ -74,7 +74,8 @@ func writeTo(dockerCli command.Cli, reader io.Reader, dest string) error {
 	return nil
 }
 
-func runExport(dockerCli command.Cli, opts *exportOptions) error {
+// RunExport exports a Docker context
+func RunExport(dockerCli command.Cli, opts *exportOptions) error {
```
guillaumerose: exportOptions is not exported

cli/command/context/update.go
```diff
@@ -65,7 +65,8 @@ func newUpdateCommand(dockerCli command.Cli) *cobra.Command {
 	return cmd
 }
 
-func runUpdate(cli command.Cli, o *updateOptions) error {
+// RunUpdate updates a Docker context
+func RunUpdate(cli command.Cli, o *updateOptions) error {
```
guillaumerose: updateOptions not exported

simonferquel:
Thanks for spotting the missing exported symbols, fixed

thaJeztah:
Copying this from ;

The concern I have with exporting all of these is that;

- These functions now become a public "API", with no interface defined for them
  - this can complicate making changes in the CLI (they were kept non-exported, as their signature/interface was not "stable")
- Some of these functions actually have a "UX" component built-in (e.g. they print confirmation messages on stdout/stderr, have an option to print formatted output); is that really suitable for external consumption?
  - calling these functions would become almost equivalent to `exec.Cmd("docker stack ls")`


For the _underlying_ code on the other hand, we have various interfaces defined; if those are lacking functionality that's needed, should we review those interfaces, and make that the API to use?

simonferquel:
Ok, I will try to move name validation and default context handling lower in the stack then

ijc:
> These functions now become a public "API", with no interface defined for them
> * this can complicate making changes in the CLI (they were kept non-exported, as their signature/interface was not "stable")

I don't think that dichotomy really exists as strongly as you have stated it, it is entirely valid for software to export an API which is explicitly unstable if the maintainers choose to do so and for them to then feel no qualms about changing it as necessary despite it being exported. I think that would be an entirely valid approach to consider for something like `docker/cli` (at least the bits under discussion here and in #1519). It just means that external consumers should be aware and be prepared to adjust as the unstable interfaces change (e.g. while they are revendoring).

thaJeztah:
Ah, sorry, I saw the comment on the other PR first, so let me copy my reply from there (https://github.com/docker/cli/pull/1519#discussion_r250189248);

> I was discussing with @chris-crone to add a note somewhere (perhaps the readme) to indicate that these functions, even though they're exported, should not be considered a stable API (as in; their signature can change, and shouldn't be relied on)
> 
> We could also add notes to the GoDoc of these, similar to https://golang.org/pkg/testing/#RunTests
> 
> > An internal function but exported because it is cross-package; part of the implementation of the "go test" command.
> 
> But that would be a bit more work to do

I'm ok with this change, but we should add that note somewhere (separately from this PR), and start thinking of what changes would be needed to have a slightly higher-level interface than the API-client (to make it better consumable), but "lower" than the current functions (which, I think, are too tightly coupled to the CLI itself and the UX it has).

cli/command/cli.go
```diff
@@ -79,7 +79,10 @@ type DockerCli struct {
 	dockerEndpoint        docker.Endpoint
 }
 
-var storeConfig = store.NewConfig(
+// ContextStoreConfig contains the config used by the context store
+// A plugin's init function may modify it to define other typed endpoint metadata by calling
+// ContextStoreConfig.SetEndpoint()
+var ContextStoreConfig = store.NewConfig(
```
vdemeester: Wouldn't it better to have a function here ?

vdemeester: (to be used from outside)

simonferquel: something like `command.SetContextStoreEndpointType()` ?

thaJeztah: Or `store.DefaultConfig()` ? 🤔

simonferquel:
As discussed privately with @vdemeester, I'll wait for Silvin's pr to land and introduce a `WithAdditionalEndpoint` option

thaJeztah:
> As discussed privately with @vdemeester, I'll wait for Silvin's pr to land and introduce a WithAdditionalEndpoint option

❤️ even better!

simonferquel:
@thaJeztah done! PTAL :)

simonferquel:
@vdemeester WDYT?

simonferquel:
@thaJeztah rebased + squashed, thanks

PR #1633, opened 2019-01-23
Introduce functional arguments to NewDockerCli for a more stable API.

silvin-lubecki:
**- What I did**
Introduce functional arguments to NewDockerCli for a more stable API.
Add an Apply function to the Cli interface to modify it after construction.

image


cli/command/cli.go
```diff
@@ -62,6 +63,9 @@ type Cli interface {
 	DockerEndpoint() docker.Endpoint
 }
 
+// DockerCliOperator applies a modification on a DockerCli.
+type DockerCliOperator func(cli *DockerCli)
```
thaJeztah: perhaps `s/DockerCliOperator/DockerCliOption/` ?

note that strictly, we should've used CLI (all caps), but that ship has likely sailed 😓 


One thing I'm thinking; should we change the signature to return an `error`? This would allow us to capture possible errors in the configuration (which could be if these functional options take (e.g.) configuration from something else, and there's an issue with the configuration, or (e.g.) if there's conflicting options set


@vdemeester @ijc wdyt?

cli/command/cli.go
```diff
@@ -407,9 +418,55 @@ type ClientInfo struct {
...
-func NewDockerCli(in io.ReadCloser, out, err io.Writer, isTrusted bool, containerizedFn func(string) (clitypes.ContainerizedClient, error)) *DockerCli {
-	return &DockerCli{in: NewInStream(in), out: NewOutStream(out), err: err, contentTrust: isTrusted, newContainerizeClient: containerizedFn}
+// NewDockerCli returns a DockerCli instance with all operators applied on it.
+func NewDockerCli(ops ...DockerCliOperator) *DockerCli {
+	cli := &DockerCli{}
+	cli.Apply(ops...)
+	return cli
+}
+
+// WithStandardStreams sets a cli in, out and err streams with the standard streams.
```
thaJeztah: Perhaps we should move all these options to a separate file (e.g. `cli/command/cli_options.go` or `cli/command/options.go`)

cli/command/cli.go
vdemeester: I do agree on the "return `error`" :angel: :wink:. I do like `Operator` somehow :upside_down_face:

silvin-lubecki: 👍 for the renaming.
About the error, I thought about that too, I just wondered it may modify too much code for the caller, as suddenly it has to return and propagate an error, if it didn't already. But ok I'll add that 👍

cli/command/cli.go
silvin-lubecki: I'll do that 👍

cli/command/cli.go
```diff
@@ -407,9 +418,55 @@ type ClientInfo struct {
 	DefaultVersion  string
 }
 
-// NewDockerCli returns a DockerCli instance with IO output and error streams set by in, out and err.
-func NewDockerCli(in io.ReadCloser, out, err io.Writer, isTrusted bool, containerizedFn func(string) (clitypes.ContainerizedClient, error)) *DockerCli {
-	return &DockerCli{in: NewInStream(in), out: NewOutStream(out), err: err, contentTrust: isTrusted, newContainerizeClient: containerizedFn}
+// NewDockerCli returns a DockerCli instance with all operators applied on it.
+func NewDockerCli(ops ...DockerCliOperator) *DockerCli {
+	cli := &DockerCli{}
```
ijc: Does this mean that some `ops` are always needed (e.g. to provide a stream). Would we want to stub those out to some default (either `os.Std*` or `io.Discard` I suppose)?

cli/command/cli.go
```diff
@@ -407,9 +418,55 @@ type ClientInfo struct {
...
+
+// WithErrorStream sets a cli error stream.
+func WithErrorStream(err io.Writer) DockerCliOperator {
+	return func(cli *DockerCli) {
+		cli.err = err
+	}
+}
+
+// WithContentTrust enables content trust on a cli.
+func WithContentTrust() DockerCliOperator {
```
ijc: FWIW I think the default should be to obey the `DOCKER_CONTENT_TRUST` variable (encapsulated via `contentTrustEnabled()` today) and there should be one (or two) of these to force it on or off explicitly.

cmd/docker/docker_test.go
```diff
@@ -26,24 +25,25 @@ func TestClientDebugEnabled(t *testing.T) {
 }
 
 func TestExitStatusForInvalidSubcommandWithHelpFlag(t *testing.T) {
-	discard := ioutil.Discard
-	cmd := newDockerCommand(command.NewDockerCli(os.Stdin, discard, discard, false, nil))
+	cmd := newDockerCommand(command.NewDockerCli(command.WithInputStream(os.Stdin)))
```
ijc: cf what I wrote above this appears to rely on the streams defaulting to discard.

cli/command/cli.go
thaJeztah: > I think that this func typedef should certainly return an error, but I don't think it necessarily has to follow that NewDockerCli needs to propagate it for the user to handle

I've seen two patterns; one that never errors, and one that can error. I think having an error is more flexible (errors can be ignored if we don't need them, or a functional option can just return `nil`)

> but make a more dynamic decision about whether to use (and this Initialize) it further down the line.

Right, so I was thinking about that a bit; there's some options that may depend on runtime options (e.g. if a specific flag or environment variable is set) and those have to be evaluated when a command is triggered.

Functions are first-class citizens in Go; should we have something to pass functions that can be executed at runtime (e.g. for the context/config store to be evaluated when the command is run, instead of having to manually call `Initialize()` before running a command?

simonferquel: Having the process exit on NewDockerCli() seem a bit dangerous to me, for 3rd party consumers. I'd vote for NewDockerCli() returning an error.

simonferquel: For Post-Initialize functions, we can still use Apply()

cli/command/cli.go
```diff
@@ -407,9 +418,55 @@ type ClientInfo struct {
...
-	return &DockerCli{in: NewInStream(in), out: NewOutStream(out), err: err, contentTrust: isTrusted, newContainerizeClient: containerizedFn}
+// NewDockerCli returns a DockerCli instance with all operators applied on it.
+func NewDockerCli(ops ...DockerCliOperator) *DockerCli {
+	cli := &DockerCli{}
+	cli.Apply(ops...)
+	return cli
+}
+
+// WithStandardStreams sets a cli in, out and err streams with the standard streams.
+func WithStandardStreams() DockerCliOperator {
```
simonferquel: I feel like it should not be an option, and be set by default (the user would use `With*Stream` to override defaults). Ideally NewDockerCli without parameters should just work.

cli/command/cli.go
simonferquel: 👍 for default contentTrust value initialized using the env variable, plus to enable/disable explicitly, I would have the signature be `func WithContentTrust(enabled bool) DockerCliOperator`

cmd/docker/docker.go
```diff
@@ -172,9 +172,16 @@ func noArgs(cmd *cobra.Command, args []string) error {
 func main() {
 	// Set terminal emulation based on platform as required.
 	stdin, stdout, stderr := term.StdStreams()
-	logrus.SetOutput(stderr)
-
-	dockerCli := command.NewDockerCli(stdin, stdout, stderr, contentTrustEnabled(), containerizedengine.NewClient)
+	logrus.SetOutput(os.Stderr)
+
+	dockerCli := command.NewDockerCli(
+		command.WithInputStream(stdin),
```
simonferquel: within the CLI code, no option should be required. (good defaults for streams, containerizedClient and content trust in the constructor). This is to make it easier for 3rd party to have (by default) the exact same behavior as the CLI itself, and use the options only to customize this behavior.

cli/command/cli.go
silvin-lubecki: Fixed 👍

cli/command/cli.go
silvin-lubecki: I added default operations:
```go
func NewDockerCli(ops ...DockerCliOperator) (*DockerCli, error) {
	cli := &DockerCli{}
	defaultOps := []DockerCliOperator{
		WithStandardStreams(),
		WithContentTrustFromEnv(),
		WithContainerizedClient(containerizedengine.NewClient),
	}
	ops = append(defaultOps, ops...)
	if err := cli.Apply(ops...); err != nil {
		return nil, err
	}
	return cli, nil
}
```

silvin-lubecki: One can override them as users operations are applied after the default ones.

simonferquel:
lgtm

cli/command/cli.go
```diff
@@ -66,8 +69,8 @@ type Cli interface {
 // Instances of the client can be returned from NewDockerCli.
 type DockerCli struct {
 	configFile            *configfile.ConfigFile
-	in                    *InStream
-	out                   *OutStream
+	in                    *streams.InStream
```
vdemeester: -> `streams.In`

cli/command/cli.go
```diff
@@ -66,8 +69,8 @@ type Cli interface {
...
+	out                   *streams.OutStream
```
vdemeester: `streams.Out`

cli/command/cli_options.go
```diff
@@ -0,0 +1,91 @@
...
+
+// DockerCliOption applies a modification on a DockerCli.
+type DockerCliOption func(cli *DockerCli) error
+
+// WithStandardStreams sets a cli in, out and err streams with the standard streams.
+func WithStandardStreams() DockerCliOption {
+	return func(cli *DockerCli) error {
+		// Set terminal emulation based on platform as required.
+		stdin, stdout, stderr := term.StdStreams()
+		cli.in = streams.NewInStream(stdin)
```
ijc: If you accept @vdemeester 's suggestion to make the type `streams.In` you'd want to `s/Stream//g` here too.

cli/command/cli_options.go
```diff
@@ -0,0 +1,91 @@
...
+	return func(cli *DockerCli) error {
+		cli.err = err
+		return nil
+	}
+}
+
+// WithContentTrustFromEnv enables content trust on a cli from environment variable DOCKER_CONTENT_TRUST value.
+func WithContentTrustFromEnv() DockerCliOption {
+	return func(cli *DockerCli) error {
+		cli.contentTrust = func() bool {
```
ijc: What is the inner `func() bool` needed for here?

cli/command/cli_options.go
simonferquel: in order not to break (too much) existing code (and I know, this PR WILL break some code), we can alias all the moved Stream related symbols so that command.NewInStream, command.InStream etc still work

simonferquel: (on the + side, it will make for less code changes in this PR as well)

cli/command/cli_test.go
```diff
@@ -247,3 +252,31 @@ func TestGetClientWithPassword(t *testing.T) {
...
+
+	assert.Check(t, cli.In() != nil)
+	assert.Check(t, cli.Out() != nil)
+	assert.Check(t, cli.Err() != nil)
+	assert.Equal(t, cli.ContentTrustEnabled(), true)
+	client, err := cli.NewContainerizedEngineClient("")
+	assert.NilError(t, err)
+	assert.Equal(t, client, nil)
+
+	// Apply can modify a dockerCli after construction
```
ijc: I'd be happier that this test was really checking that if it was asserting that the in/out streams had `FD() != 0` before this apply.

Oh, except it can be 0 for `In` anyway, since that is the `stdin` fd too. Not sure how best to get around that -- perhaps by calling something (like `runVersion`?) to actually output something into the buf and check it went where you hoped it would?

cmd/docker/docker.go
```diff
@@ -170,17 +167,19 @@ func noArgs(cmd *cobra.Command, args []string) error {
 }
 
 func main() {
-	// Set terminal emulation based on platform as required.
-	stdin, stdout, stderr := term.StdStreams()
-	logrus.SetOutput(stderr)
+	logrus.SetOutput(os.Stderr)
```
ijc: Are we happy that this is still the same thing even on Windows? I know that `term.StdStreams` will still be called implicitly by the following `NewDockerCli`. I think it's the same fd as you'd get here so maybe^Wprobably that's fine.

One other subtle change here I just realised is that `NewDockerCli` will now _always_ call `term.StdStreams()`, which unconditionally does some ioctls on stdio, even if the caller uses their own `WithStandardStreams` to set something else. I'm not going to pretend I know what those ioctls are doing -- are we sure they are safe and universally desirable?

cli/command/cli_options.go
ijc: TBH I'm inclined towards just biting the bullet and making/propagating the change. The necessary updates are pretty mechanical and we should try and avoid building up compatibility debt (which can never be removed) when we don't really need to.

Up to the maintainers though.

cmd/docker/docker.go
simonferquel: Good call. Can you change the default stdstreams handling in NewDockerCLI to be done at the end, only if there are missing streams (and if only stderr is missng, don't call term.StdStreams()).

cli/command/cli_options.go
silvin-lubecki: Well I just moved `func contentTrustEnabled() bool` from `cmd/docker/docker.go` to here, but Ok I can simplify it 👍

cli/command/cli_test.go
silvin-lubecki: If fixed that directly reading or writing to the streams and check with their underlying buffers. 👍

cli/command/cli_options.go
thaJeztah: If it's a one-time change, I agree with grouping breaking changes together (separate commit though)

cli/command/cli.go
thaJeztah: I think the current approach looks valid; there may other caveats (what if want _without_ containerised client, or without content-trust)"

possible alternatives are;

- only apply defaults if `ops` is empty
- add a `WithDefaults()` option (that's applying these), which can be used if someone wants the defaults _plus_ some custom configurations.

But we can discuss those solutions (not sure which approach makes the most sense)

cmd/docker/docker.go
silvin-lubecki: 👍

cli/command/cli.go
```diff
@@ -407,9 +421,26 @@ type ClientInfo struct {
...
+	}
+	ops = append(defaultOps, ops...)
+	if err := cli.Apply(ops...); err != nil {
+		return nil, err
+	}
+	if cli.out == nil || cli.in == nil {
+		WithStandardStreams()(cli)
+	}
+	if cli.err == nil {
+		cli.err = ioutil.Discard
```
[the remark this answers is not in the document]
silvin-lubecki: I'm wondering about that... 🤔 But why not.

silvin-lubecki:
@simonferquel @thaJeztah @vdemeester @ijc Please take another look 😺

silvin-lubecki:
I'll squash the commits once they are LGTMed

cli/command/cli.go
```diff
@@ -407,9 +421,32 @@ type ClientInfo struct {
...
+	cli := &DockerCli{}
+	defaultOps := []DockerCliOption{
+		WithContentTrustFromEnv(),
+		WithContainerizedClient(containerizedengine.NewClient),
+	}
+	ops = append(defaultOps, ops...)
+	if err := cli.Apply(ops...); err != nil {
+		return nil, err
+	}
+	if cli.out == nil || cli.in == nil {
```
simonferquel:
```suggestion
	if cli.out == nil || cli.in == nil || cli.err == nil {
```

cli/command/cli_test.go
ijc: Looks good, thanks.

cli/command/streams.go
```diff
@@ -0,0 +1,23 @@
+package command
+
+import (
+	"github.com/docker/cli/cli/streams"
+)
+
+// InStream is an input stream used by the DockerCli to read user input
+// Deprecated: Use github.com/docker/cli/cli/streams.In instead
```
ijc: When do we expect we would be able to get rid of these if not right now?

silvin-lubecki: I don't think we have an official depreciation plan, but I think that 2 releases is a good start, so let's say we will remove them after the 19.09 release? Do you want me to add a comment about it in the code?

silvin-lubecki:
@thaJeztah just squashed and re-organized the commits. PTAL 🐱

vdemeester:
yay :tada:

cmd/docker/docker.go
```diff
@@ -170,17 +167,19 @@ func noArgs(cmd *cobra.Command, args []string) error {
 }
 
 func main() {
-	// Set terminal emulation based on platform as required.
-	stdin, stdout, stderr := term.StdStreams()
-	logrus.SetOutput(stderr)
+	dockerCli, err := command.NewDockerCli()
+	if err != nil {
+		fmt.Fprintln(dockerCli.Err(), err)
```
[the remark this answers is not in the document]
silvin-lubecki: Right, will fix that in a follow-up! Thank you for spotting that @ijc ! 🤗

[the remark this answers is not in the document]
thaJeztah: nice catch 👍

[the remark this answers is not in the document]
silvin-lubecki: Fixed with https://github.com/docker/cli/pull/1645

PR #1564, opened 2018-12-12
Basic framework for writing and running CLI plugins

ijc:
**- What I did**

This is a small part of #1534, I wanted to get the basic framework out so that potential consumers could start to play with it and give feedback on how it works for them as well as getting some early feedback on the code.

NB this also includes the changes for #1558 since I was working on top of that.

There is still lots to do:
- [x] ~~I've left off the `docker cli-plugin *` stuff since it is still being discussed (although some bits of the internal scaffolding here for listing and finding plugins are more related to that than the actual functionality added here, but I didn't tease those apart).~~ Not doing this any more, at least not in this round (https://github.com/docker/cli/issues/1534#issuecomment-446921207).
- [x] List broken plugins as a new section in `docker help`.
- [ ] There are no docs;
- [ ] Although there are some here there are insufficient tests (both unit & e2e);
- [x] `docker someplugin --help` produces the top-level help instead of calling the plugin for some reason (this means one of the new e2e tests currently fails), looks like `docker someplugin --version` has the same problem, I'm investigating.
- [x] Expose plugins in `docker info`
- [x] Plugin search path configurable via `~/.docker/config.json`
- [ ] Apply the package naming the maintainers decide upon.
- [ ] All the stuff around exposing configuration files, credential handling, table formatting, using `dial-stdio` etc isn't here (Some of those are nops in practice and just need checking that they are usable externally, otherwise I plan to do these bits in one or more followups, somewhat based on/prioritise by the experiences of folks trying to use this). _edit_: to clarify, I don't intend on doing this bit here but rather in a followup.

**- How to verify it**

The unit and e2e tests cover some of it.

Plugin path can be configured with the "cliPluginsExtraDirs" key in `~/.docker/config.json` so for having a play (using a local config dir for isolation) you can also `make -f docker.Makefile build plugins` then things like:
```console
$ mkdir .docker-plugin-test
$ echo "{\"cliPluginsExtraDirs\": [\"`pwd`/build/plugins-linux-amd64/\"]}" > .docker-plugin-test/config.json
$ $ jq . .docker-plugin-test/config.json
{
  "cliPluginsExtraDirs": [
    ".../gocode/src/github.com/docker/cli/build/plugins-linux-amd64"
  ]
}
$ ./build/docker --config=.docker-plugin-test helloworld
```
to have a play.

**- Description for the changelog**

- Basic support for writing and running external CLI commands

PR #1638, opened 2019-01-24
Separate client infomation in `docker system info`

ijc:
**- What I did**

Split the output of `docker info` (aka `docker system info`) into client and server sections.

Currently there is only one client setting (debug), but this arose from the discussion at https://github.com/docker/cli/pull/1564/#discussion_r249102770 which wants to add info about cli plugins.

The new arrangements are also to print as much as possible of both the client and server information even in the presence of errors (e.g. gathering the info or rendering some part of it) and to list those errors at the end (and return an overall error).

In the non-pretty case (`--format='{{json .}}'` etc) the daemon info remains at the top-level (for API backwards compatibility) while the client info is in a new `ClientInfo` key. Ideally the daemon info would be in a `DaemonInfo` key for consistency but that would break the format which I presume is considered stable and which people rely on (if not I should definitely make it cleaner). Errors are given in new `ClientErrors` and `DaemonErrors` keys. Not that things would go a bit strange if fields with the same names were added to the daemon `types.Info`. Given the names in question I think/hope that's unlikely to happen.

Note that the client debug setting was not previously in the JSON (which was purely daemon info) despite being in the pretty version.

**- How I did it**

Hard graft.

**- How to verify it**

I added and extended the units tests (including pulling some useful/relevant bits from #1564). You could also just try `docker info` with and without a valid daemon.

The `docker system info` output now segregates information relevant to the client and daemon.


cli/command/system/info_test.go
```diff
@@ -21,8 +21,9 @@ func base64Decode(val string) []byte {
 	return decoded
 }
 
+var sampleID = "EKHL:QDUU:QZ7U:MKGD:VDXK:S27Q:GIPU:24B7:R7VT:DGN6:QCSF:2UBX"
```
thaJeztah: nit: `const`? (no blocker)

cli/command/system/testdata/docker-info-no-swarm.golden
```diff
@@ -1,3 +1,9 @@
+Client
+------
+Debug Mode: true
+
+Daemon
+------
```
thaJeztah: Instead of a separator-line, should we use the same format as on `docker version`?

```
  Debug Mode: true

  Containers: 0
    Running: 0
    Paused: 0

....
```

docs/reference/commandline/info.md
```diff
@@ -55,7 +55,12 @@ information about the `devicemapper` storage driver is shown:
 
 ```bash
 $ docker info
+Client
+------
+Debug mode: false
```
thaJeztah: Looks like the code uses capital `M` now

docs/reference/commandline/info.md
```diff
@@ -96,8 +101,7 @@ Total Memory: 991.7 MiB
 Name: ip-172-30-0-91.ec2.internal
 ID: I54V:OLXT:HVMM:TPKO:JPHQ:CQCD:JNLC:O3BZ:4ZVJ:43XJ:PFHZ:6N2S
 Docker Root Dir: /var/lib/docker
-Debug mode (client): false
-Debug mode (server): false
+Debug mode: false
```
thaJeztah: Looks like the code uses capital `M` now

docs/reference/commandline/info.md
```diff
@@ -112,7 +116,12 @@ storage driver and a node that is part of a 2-node swarm:
 
 ```bash
 $ docker -D info
+Client
+------
+Debug Mode (client): true
```
thaJeztah: remove `(client)`

man/src/system/info.md
```diff
@@ -24,6 +24,12 @@ Here is a sample output for a daemon running on Ubuntu, using the overlay2
 storage driver:
 
     $ docker -D info
+    Client
+    ------ 
+    Debug Mode (client): true
```
thaJeztah: remove `(client)`

cli/command/system/info_test.go
```diff
@@ -279,6 +282,16 @@ func TestPrettyPrintInfo(t *testing.T) {
 			jsonGolden:    "docker-info-errors",
 			expectedError: "errors pretty printing info",
 		},
+		{
+			doc: "bad security info",
+			dockerInfo: info{
+				Info:       &sampleInfoBadSecurity,
```
thaJeztah: Perhaps the test should have something to return multiple errors (and a warning?)

cli/command/system/info_test.go
ijc: This came from #1564 and I had the same feedback there, thought I'd fixed it there already! Will do it now (again?)

cli/command/system/testdata/docker-info-no-swarm.golden
ijc: I was just trying to avoid touching every single line ;-) (there's also a small possibility of breaking scripts, but they should be using the machine readable form anyway).

I'll make the indentation change and look into adding the platform version -- not sure how tricky that is to get hold of here.

docs/reference/commandline/info.md
ijc: Indeed, that seems to be a longstanding difference rather than something I've done. Will fix it up since I'm touch it.

cli/command/system/info_test.go
ijc: There's only one unit testable error path out of `prettyPrintDaemonInfo` (the security parsing stuff), the other error path is a failure to talk to the daemon which can't be unit tested. I'll add a hardcoded error in the dummy info, that should at least test the appending bit.

Warnings are already covered by pre-existing tests.

cli/command/system/info.go
```diff
@@ -91,8 +90,7 @@ func prettyPrintInfo(dockerCli command.Cli, info info) error {
 	}
 
 	fmt.Fprintln(dockerCli.Out())
-	fmt.Fprintln(dockerCli.Out(), "Daemon")
-	fmt.Fprintln(dockerCli.Out(), "------")
+	fmt.Fprintln(dockerCli.Out(), "Daemon:")
```
thaJeztah: 😅 erm; should this be `Server:` (to be consistent with `docker version`)?

```suggestion
	fmt.Fprintln(dockerCli.Out(), "Server:")
```

ijc: Will do. Luckily they are the same length so it's easy(ish) to push through the series without worrying about alignment issues etc ;-)

ijc:
@thaJeztah made that `s/Daemon/Server/g` change, thanks!

thaJeztah:
Lets Get This Merged

PR #1640, opened 2019-01-25
improve documentation

ChaosGramer:
The remove all stopped containers was wrong. It was trying to remove also all running containers
The issue was opened here: [docker.github.io/issues/8065](https://github.com/docker/docker.github.io/issues/8065)

**- What I did**
I ran the command `docker rm $(docker ps -a -q) `as stated in the doc. 
This tried to delete all runnign containers.
**- How I did it**

**- How to verify it**
Have 1 running container and atleast 1 stopped container. 
Run `docker rm $(docker ps -a -q) `. The command will try to delete both containers.
Adopt documentation, because the command was wrong. 

img_0692

### DIFF
--- a/docs/reference/commandline/rm.md
+++ b/docs/reference/commandline/rm.md
@@ -68,14 +68,7 @@ The main process inside the container referenced under the link `redis` will rec
 
 ### Remove all stopped containers
 
-```bash
-$ docker rm (docker ps -a -f "status=exited" -q)
-```
-
-This command will delete all stopped containers. The command
-`docker ps -a -q` will return all existing container IDs and pass them to
-the `rm` command which will delete them. Any running containers will not be
-deleted.
+To remove stopped container, please refer to [docker prune container documentation](/docs/reference/commandline/container_prune.md) 
 
 ### Remove a container and its volumes
 
--- a/docs/reference/commandline/rm.md
+++ b/docs/reference/commandline/rm.md
@@ -69,7 +69,7 @@ The main process inside the container referenced under the link `redis` will rec
 ### Remove all stopped containers
 
 ```bash
-$ docker rm $(docker ps -a -q)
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/rm.md
```diff
@@ -69,7 +69,7 @@ The main process inside the container referenced under the link `redis` will rec
 ### Remove all stopped containers
 
 ```bash
-$ docker rm $(docker ps -a -q)
+$ docker rm (docker ps -a -f "status=exited" -q)
```
thaJeztah: The `$` should stay here, otherwise it won't work on Bash (when using Fish, the `$` should not be used though.

docs/reference/commandline/rm.md
```diff
@@ -69,7 +69,7 @@ The main process inside the container referenced under the link `redis` will rec
...
 ```
 
 This command will delete all stopped containers. The command
```
thaJeztah: It's actually explained below that the `docker ps -a -q` returns all containers, but the `docker rm` will not remove running containers in the list.

ChaosGramer: But still it is trying to rm them. Do you think we should leave it like it is?

ChaosGramer: Well good point is it documented, that the running container won't be deleted. But still those messages are like an unexpected behaviour. For people, which are new to docker this is confusing.

ChaosGramer: Well good point it is documented, that the running container won't be deleted. But still those messages are like an unexpected behaviour. For people, which are new to docker this is confusing.

ChaosGramer:
@thaJeztah I just changed the section to refer to the docker container prune section

ChaosGramer:
@thaJeztah I'm waiting for your input, what do you think?

PR #1645, opened 2019-01-29
Fix potential panic at DockerCli creation

silvin-lubecki:
**- What I did**

Fix using a nil dockerCli if an error occurred during cli creation, using the standard error stream instead.
The code itself wasn't really panicking as `NewDockerCli` never returns an error... yet.

image


silvin-lubecki:
Come on `codecov/patch`...

Issue #1622, opened 2019-01-19
Feature: add "docker context edit" function


thaJeztah:
This is just a quick thought, but thought I'd write it down.

The next version of the docker cli adds a new `docker context` subcommand (https://github.com/docker/cli/pull/1501). With this subcommand, I can import, create, update, and export contexts.

The current set of options per context  is limited, but may be extended in future (see 
https://github.com/docker/cli/issues/1621 for a proposal).

Updating, or creating a context works for the current set of options, but likely doesn't scale well if there's many new options are to be added; at some point, people will have to resort to scripts to update, or create a context (`docker context update --docker foo=bar,bar=baz --kubernetes foo=bar,bar=baz`).

As an alternative, users can edit the files in the context storage directly, but this may not be desirable;

```bash
cd $(docker context inspect --format='{{.Storage.MetadataPath}}' mycontext)
vi meta.json
```

In addition, `json` works great for automation, and to serialize as a string, but does not allow comments, nor may it be very user-friendly to work with.

## Proposal

Introduce a `docker context edit` subcommand.

The `docker context edit` command, using a similar approach as `git commit`, which (I think) works something like this;

- export the context to edit to a temporary file (e.g. `~/.docker/context/CONTEXT_EDIT`)
  - this file could be in JSON format (but pretty-printed), or
  - or use a user-friendly format (TOML?)
  - if we pick TOML or YAML, we can present an _anotated_ version of the context
- open the file in the default editor
- when _saving_ the file, the context is validated and updated; the temporary file is removed
- when _exiting_ the editor without saving, the update is aborted (`Aborting update.`)
- if validation fails, the update is aborted as well.

```bash
docker context edit mycontext
```

Opens an editor with the context in pretty-print;

```json
{
  "Name": "mycontext",
  "Metadata": {
    "Description": "this is my context"
  },
  "Endpoints": {
    "docker": {
      "Host": "unix:///var/run/docker.sock",
      "SkipTLSVerify": false
    }
  }
}
```

Or, in an annotated TOML format;

```toml
# Editing context "mycontext". Lines starting with '#' will be ignored,
# and exiting your editor without saving aborts the update.



# This is the name of the context. Changing the name will create a 
# copy of the existing context under the new name.
Name = "mycontext"

[Metadata]

# Description of this context.
Description = "this is my context"

[Endpoints]

  # Configuration for the Docker endpoint of this context
  [docker]
  # Address of the host of this endpoint.
  # 
  # Valid schemes are:
[...]

PR #1652, opened 2019-02-01
Add a field to the config file for plugin use.

ijc:
This is a bit manual (as the unit test attests) so we may find we want to add
some helpers/accessors, but this is enough to let plugins use it and to
preserve the information through round-trips.

**- What I did**

Added a map to the `config.json` top-level which can be used by plugins.

**- How I did it**

Added a map of string → json.RawMessage. Using RawMessage allows plugins to late bind on the structure without patching `docker/cli`.

**- How to verify it**

I added a unit test, but you could also port your plugin you use it ;-)

N/A -- covered by #1564's entry.

### DIFF
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -49,6 +49,7 @@ type ConfigFile struct {
 	Kubernetes           *KubernetesConfig           `json:"kubernetes,omitempty"`
 	CurrentContext       string                      `json:"currentContext,omitempty"`
 	CLIPluginsExtraDirs  []string                    `json:"cliPluginsExtraDirs,omitempty"`
+	Plugins              map[string]json.RawMessage  `json:"plugins,omitempty"`
 }
 
 // ProxyConfig contains proxy configuration settings
@@ -70,6 +71,7 @@ func New(fn string) *ConfigFile {
 		AuthConfigs: make(map[string]types.AuthConfig),
 		HTTPHeaders: make(map[string]string),
 		Filename:    fn,
+		Plugins:     make(map[string]json.RawMessage),
 	}
 }
 
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -24,32 +24,32 @@ const (
 
 // ConfigFile ~/.docker/config.json file info
 type ConfigFile struct {
-	AuthConfigs          map[string]types.AuthConfig `json:"auths"`
-	HTTPHeaders          map[string]string           `json:"HttpHeaders,omitempty"`
-	PsFormat             string                      `json:"psFormat,omitempty"`
-	ImagesFormat         string                      `json:"imagesFormat,omitempty"`
-	NetworksFormat       string                      `json:"networksFormat,omitempty"`
-	PluginsFormat        string                      `json:"pluginsFormat,omitempty"`
-	VolumesFormat        string                      `json:"volumesFormat,omitempty"`
-	StatsFormat          string                      `json:"statsFormat,omitempty"`
-	DetachKeys           string                      `json:"detachKeys,omitempty"`
-	CredentialsStore     string                      `json:"credsStore,omitempty"`
-	CredentialHelpers    map[string]string           `json:"credHelpers,omitempty"`
-	Filename             string                      `json:"-"` // Note: for internal use only
-	ServiceInspectFormat string                      `json:"serviceInspectFormat,omitempty"`
-	ServicesFormat       string                      `json:"servicesFormat,omitempty"`
-	TasksFormat          string                      `json:"tasksFormat,omitempty"`
-	SecretFormat         string                      `json:"secretFormat,omitempty"`
-	ConfigFormat         string                      `json:"configFormat,omitempty"`
-	NodesFormat          string                      `json:"nodesFormat,omitempty"`
-	PruneFilters         []string                    `json:"pruneFilters,omitempty"`
-	Proxies              map[string]ProxyConfig      `json:"proxies,omitempty"`
-	Experimental         string                      `json:"experimental,omitempty"`
-	StackOrchestrator    string                      `json:"stackOrchestrator,omitempty"`
-	Kubernetes           *KubernetesConfig           `json:"kubernetes,omitempty"`
-	CurrentContext       string                      `json:"currentContext,omitempty"`
-	CLIPluginsExtraDirs  []string                    `json:"cliPluginsExtraDirs,omitempty"`
-	Plugins              map[string]json.RawMessage  `json:"plugins,omitempty"`
+	AuthConfigs          map[string]types.AuthConfig  `json:"auths"`
+	HTTPHeaders          map[string]string            `json:"HttpHeaders,omitempty"`
+	PsFormat             string                       `json:"psFormat,omitempty"`
+	ImagesFormat         string                       `json:"imagesFormat,omitempty"`
+	NetworksFormat       string                       `json:"networksFormat,omitempty"`
+	PluginsFormat        string                       `json:"pluginsFormat,omitempty"`
+	VolumesFormat        string                       `json:"volumesFormat,omitempty"`
+	StatsFormat          string                       `json:"statsFormat,omitempty"`
+	DetachKeys           string                       `json:"detachKeys,omitempty"`
+	CredentialsStore     string                       `json:"credsStore,omitempty"`
+	CredentialHelpers    map[string]string            `json:"credHelpers,omitempty"`
+	Filename             string                       `json:"-"` // Note: for internal use only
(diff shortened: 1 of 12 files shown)


thaJeztah:
@ijc so, this would be configuration options for a plugin, right?

Some quick thoughts;

`json.RawMessage` would means, that "anything goes", right? So plugins can go crazy if they want to;

```json
{
  Plugins: {
    "hello": {
      "foobar": {
        "nestedfoobar" : {
          "centre-of-earth": true
        }
      },
      "bla": ["one", "two", "three"]
    }
  }
}
```

I know at the API level, we have options for plugins, and limit those to
a `map[string]string`. This can be somewhat limited (plugins may have to use workarounds such as "comma-separated values" and such), but keeps the format simple.


Another thing I'm wondering is;

Do we want plugins to use separate files, so that it's easier to;

- ship them with a configuration file
- keep it slightly more "separated" from the CLI

One note to add; we should rethink our configuration file format(s) (in general). JSON is great, but (e.g.) doesn't allow for annotated versions,
which sometimes limits its use (i.e., we cannot ship a "sample" configuration file), and is not always _that_ user-friendly

Related: https://github.com/docker/cli/issues/1622, https://github.com/moby/moby/issues/20151

ijc:
> json.RawMessage would means, that "anything goes", right? So plugins can go crazy if they want to;

Yes, that's the trade off between this or plugin authors having to patch docker/cli to add their custom fields. I think we should assume that plugin authors are generally sensible.

> Do we want plugins to use separate files, so that it's easier to;

I think having a single file is nicer for users and the integration is a feature. In particular plugins are encouraged to use global config from `config.json` wherever possible and only use the reserved key where that is not possible or sensible. Having all plugin config (both generic and specific) in a single place make more sense to me (otherwise the split is pretty arbitrary).

There some more stuff on this in the design (https://github.com/docker/cli/issues/1534) which said explicitly that the config should be in a single file.

thaJeztah:
> Yes, that's the trade off between this or plugin authors having to patch docker/cli to add their custom fields. I think we should assume that plugin authors are generally sensible.

Would making it a `map[string]string` (could be either by name spacing configuration options in the key, or providing such a map per plugin) satisfy the requirement? so either;

```json
{
  "Plugins": {
    "myplugin.option1": "value",
    "myplugin.option2": "value",
  }
}
```

or

```json
{
  "Plugins": {
    "myplugin": {
      "option1": "value",
      "option2": "value"
    }
  }
}
```

thaJeztah:
> Having all plugin config (both generic and specific) in a single place make more sense to me (otherwise the split is pretty arbitrary).

Yes, I think a single file could work for now

> That seems way out of scope for the particular change here.

The change itself is definitely out of scope, but meant to say; try to keep it format agnostic, as  JSON may become something else at some point (IOW, we should try and store the configuration itself in a known format, and not literal JSON)

ijc:
> Would making it a `map[string]string` (could be either by name spacing configuration options in the key, or providing such a map per plugin) satisfy the requirement

It'd work, but would mean plugins having to invent syntax for various structured stuff. So not ideal (but better than nothing). Maybe it's not a huge deal for the sorts of settings we imagine there being. Lists are annoying though since you need a separator, which means you need escaping etc. I suppose helpers might be the answer.

> try to keep it format agnostic, as JSON may become something else at some point (IOW, we should try and store the configuration itself in a known format, and not literal JSON)

Perhaps this calls for the helpers/accessors I hypothesised in the commit message getting pull forward, so there would be
```
func (ConfigFile *c) PluginConfig(pluginname string, config interface{})
func (ConfigFile *c) SetPluginConfig(pluginname string, config interface{})
```
which abstracts the use of `json.Unmarshal` (or similar for other formats) away?

The thing passed as config would still potentially need to have the `json:...` or `yaml:...` (or `toml:...`, (or,...)) tags though, so maybe that doesn't help?

thaJeztah:
Argh, thought I replied;

> Lists are annoying though since you need a separator, which means you need escaping etc. I suppose helpers might be the answer.

Not sure I grasp this fully; my main thought was to limit the struct for per-plugin configuration to a `map[string]string`; equivalent to the configurations that are currently used in the tests, but not allowing plugins to expand beyond that for now; so basically, a plugin gets a key/value pairs for configuration (values being strings).

> Perhaps this calls for the helpers/accessors I hypothesised in the commit message getting pull forward, so there would be
> 
> ```
> func (ConfigFile *c) PluginConfig(pluginname string, config interface{})
> func (ConfigFile *c) SetPluginConfig(pluginname string, config interface{})
> ```

Should the first example _return_ the `config`?

This could potentially be extended to;

```go
func (ConfigFile *c) PluginConfig(pluginname string, option string)
```

So, e.g. `configfile.PluginConfig("myplugin", "some-option")` would return the value for option `some-option` ("foo" with the configuration file below);

```json
{
  "Plugins": {
    "myplugin": {
      "some-option": "foo"
    }
}
```

Regarding `SetPluginConfig` - wondering if the configuration should be _updatable_ by plugins; I guess this is a bit of a grey area that never was fully fledged out (should the `config.json` be managed by the user, and only _read_ by the CLI, or should the CLI be in charge of updating it? It currently is a bit of a mix, but not well-defined).

ijc:
> > Lists are annoying though since you need a separator, which means you need escaping etc. I suppose helpers might be the answer.

> Not sure I grasp this fully; my main thought was to limit the struct for per-plugin configuration to a map[string]string; equivalent to the configurations that are currently used in the tests, but not allowing plugins to expand beyond that for now; so basically, a plugin gets a key/value pairs for configuration (values being strings).

What I meant was that if a plugin has a list of things to store (`["thinga", "thingb", "thingc"]`) then with a `map[string]string` they have to encode that list as a string, perhaps as `thinga;thingb;thingc`, but then `thing*` cannot contain a bare `;` (or whichever character is chosen), so then you need to support `\;` to escape or some other way to allow `;` in the values.

> Should the first example _return_ the `config`?

Not in what I was prpoposing, since you need to pass in a concrete thing (a `struct` in my mind) as the `interface` for the various reflection based marshalling things to have an actual thing to work to. I guess you could do
```go
func (ConfigFile *c) PluginConfig(pluginname string, config reflect.Type) interface{}
```
but that's pretty awful.

Your proposal is slightly different since it takes a single named option for which it returns the value. That would work too. It's a bit flat and I would expect to see plugins working around this with `foo.bar` keys or whatever.

Perhaps I'm over thinking this and we should go with a simple option (e.g. `map[string]string`) until we have more data on what plugins really want to do -- at which point we can evolve things (which should be easier from a simpler starting point).

silvin-lubecki:
I think we need more feedback from the plugin developers themselves and see if they really need complex configuration struct or juste few feature flags.

My 2cts is they don't really need something complex and just a `map[string]string` will fit most of their needs, but I may be wrong.

My other 2cts is that changing afterwards from a `map[string]string` to a `map[string]json.RawMessage` is possible, the opposite isn't.

ijc:
> My other 2cts is that changing afterwards from a `map[string]string` to a `map[string]json.RawMessage` is possible, the opposite isn't.

Yep, that was what I was trying to say with "is possible, the opposite isn't".

I will rework this PR to use the `map[string]string` approach ASAP (which might still be a few days since I've got a few other things on)

thaJeztah:
Thanks! Yes, that was basically my train of thought as well; for daemon drivers (logging, storage-drivers) we went for a simple `map[string]string` (or could be a `[]string{"key=val","key=val"}`) and (so far) that worked. So, yes, if it satisfies the immediate needs, I'd prefer to keep it simple, and extend in future (if there's a need).

ijc:
OK, I switched to a simple `map[string]string` scheme. I added a couple of simple accessors since there's a bit of logic to cleanup dead entries.

ijc:
Thanks. I added a few words to that doc

thaJeztah:
thanks @ijc

ping @silvin-lubecki still LGTY?

cli/config/configfile/file.go
```diff
@@ -330,6 +332,36 @@ func (configFile *ConfigFile) GetFilename() string {
 	return configFile.Filename
 }
 
+// PluginConfig retrieves the requested option for the given plugin.
+func (configFile *ConfigFile) PluginConfig(pluginname, option string) (string, bool) {
```
silvin-lubecki: Can we update the helloworld example to use these primitives? Appart from reading the unit tests, there is no obvious  way to be aware of this feature.

ijc: Sure, I'll taker a look ASAP

ijc: I did that, and made a unit test to cover these paths.

ijc:
Rebased to resolve the conflict with #1654's edits to `docs/extend/cli_plugins.md`.

PR #1654, opened 2019-02-01
cli-plugins: use system dial-stdio to contact the engine.

ijc:
**- What I did**

Arranged so that cli-plugins will use `docker system dial-stdio` to connect to the engine as proposed by @tonistiigi in https://github.com/docker/cli/issues/1534#issuecomment-446206041.

**- How I did it**

The main commit message goes into some detail:
```
    cli-plugins: use `docker system dial-stdio` to call the daemon
    
    This means that plugins can use whatever methods the monolithic CLI supports,
    which is good for consistency.
    
    This relies on `os.Args[0]` being something which can be executed again to
    reach the same binary, since it is propagated (via an envvar) to the plugin for
    this purpose. This essentially requires that the current working directory and
    path are not modified by the monolithic CLI before it launches the plugin nor
    by the plugin before it initializes the client. This should be the case.
    
    Previously the fake apiclient used by `TestExperimentalCLI` was not being used,
    since `cli.Initialize` was unconditionally overwriting it with a real one
    (talking to a real daemon during unit testing, it seems). This wasn't expected
    nor desirable and no longer happens with the new arrangements, exposing the
    fact that no `pingFunc` is provided, leading to a panic. Add a `pingFunc` to
    the fake client to avoid this.
    
```

It's a bit tricky, so could use some careful thought around the possible pitfalls. 

Things I've considered doing differently:
* Passing the original CLI via a new global argument instead of an envvar.
* Passing the entire `dial-stdio` command line to use explicitly to the plugin. Not sure how best to do that without getting into confusion around quoting all the way through.

**- How to verify it**

There is an e2e test which confirms correct argument parsing. Also the existing `TestCliInitialized` happens to use the mechanism.

N/A should be covered by #1564's entry.

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -175,9 +175,28 @@ func (cli *DockerCli) RegistryClient(allowInsecure bool) registryclient.Registry
 	return registryclient.NewRegistryClient(resolver, UserAgent(), allowInsecure)
 }
 
+// InitializeOpt is the type of the functional options passed to DockerCli.Initialize
+type InitializeOpt func(dockerCli *DockerCli) error
+
+// WithInitializeClient is passed to DockerCli.Initialize by callers who wish to set a particular API Client for use by the CLI.
+func WithInitializeClient(makeClient func(dockerCli *DockerCli) (client.APIClient, error)) InitializeOpt {
+	return func(dockerCli *DockerCli) error {
+		var err error
+		dockerCli.client, err = makeClient(dockerCli)
+		return err
+	}
+}
+
 // Initialize the dockerCli runs initialization that must happen after command
 // line flags are parsed.
-func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
+func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...InitializeOpt) error {
+	var err error
+
+	for _, o := range ops {
+		if err := o(cli); err != nil {
+			return err
+		}
+	}
 	cliflags.SetLogLevel(opts.Common.LogLevel)
 
 	if opts.ConfigDir != "" {
@@ -189,29 +208,31 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
 	}
 
 	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
-	var err error
-	cli.contextStore = store.New(cliconfig.ContextStoreDir(), cli.contextStoreConfig)
-	cli.currentContext, err = resolveContextName(opts.Common, cli.configFile, cli.contextStore)
-	if err != nil {
-		return err
-	}
-	endpoint, err := resolveDockerEndpoint(cli.contextStore, cli.currentContext, opts.Common)
-	if err != nil {
-		return errors.Wrap(err, "unable to resolve docker endpoint")
-	}
-	cli.dockerEndpoint = endpoint
 
-	cli.client, err = newAPIClientFromEndpoint(endpoint, cli.configFile)
-	if tlsconfig.IsErrEncryptedKey(err) {
-		passRetriever := passphrase.PromptRetrieverWithInOut(cli.In(), cli.Out(), nil)
-		newClient := func(password string) (client.APIClient, error) {
-			endpoint.TLSPassword = password
-			return newAPIClientFromEndpoint(endpoint, cli.configFile)
+	if cli.client == nil {
+		cli.contextStore = store.New(cliconfig.ContextStoreDir(), cli.contextStoreConfig)
+		cli.currentContext, err = resolveContextName(opts.Common, cli.configFile, cli.contextStore)
+		if err != nil {
+			return err
 		}
-		cli.client, err = getClientWithPassword(passRetriever, newClient)
-	}
-	if err != nil {
-		return err
+		endpoint, err := resolveDockerEndpoint(cli.contextStore, cli.currentContext, opts.Common)
+		if err != nil {
+			return errors.Wrap(err, "unable to resolve docker endpoint")
+		}
+		cli.dockerEndpoint = endpoint
+
+		cli.client, err = newAPIClientFromEndpoint(endpoint, cli.configFile)
+		if tlsconfig.IsErrEncryptedKey(err) {
+			passRetriever := passphrase.PromptRetrieverWithInOut(cli.In(), cli.Out(), nil)
+			newClient := func(password string) (client.APIClient, error) {
+				endpoint.TLSPassword = password
+				return newAPIClientFromEndpoint(endpoint, cli.configFile)
+			}
+			cli.client, err = getClientWithPassword(passRetriever, newClient)
+		}
+		if err != nil {
+			return err
+		}
 	}
 	var experimentalValue string
 	// Environment variable always overrides configuration
--- a/cli/connhelper/connhelper.go
+++ b/cli/connhelper/connhelper.go
@@ -53,6 +53,16 @@ func GetConnectionHelper(daemonURL string) (*ConnectionHelper, error) {
 	return nil, err
 }
 
+// GetCommandConnectionHelper returns a ConnectionHelp constructed from an arbitrary command.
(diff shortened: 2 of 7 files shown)


cli/command/cli.go
```diff
@@ -175,9 +175,26 @@ func (cli *DockerCli) RegistryClient(allowInsecure bool) registryclient.Registry
 	return registryclient.NewRegistryClient(resolver, UserAgent(), allowInsecure)
 }
 
+// WithInitializeOpt is the type of the functional options passed to DockerCli.Initialize
+type WithInitializeOpt func(dockerCli *DockerCli) error
```
thaJeztah: perhaps just `InitializeOpt` ?

cli/command/cli.go
```diff
@@ -175,9 +175,26 @@ func (cli *DockerCli) RegistryClient(allowInsecure bool) registryclient.Registry
...
+
+// WithInitilizeClient is passed to DockerCli.Initialize by callers who wish to set a particular API Client for use by the CLI.
+func WithInitilizeClient(makeClient func(dockerCli *DockerCli) (client.APIClient, error)) WithInitializeOpt {
```
thaJeztah: typo 😅 

```suggestion
func WithInitializeClient(makeClient func(dockerCli *DockerCli) (client.APIClient, error)) WithInitializeOpt {
```

cli/connhelper/connhelper.go
```diff
@@ -53,6 +53,16 @@ func GetConnectionHelper(daemonURL string) (*ConnectionHelper, error) {
 	return nil, err
 }
 
+// GetCommandConnectionHelper returns a ConnectionHelp constructed from an arbitrary command.
+func GetCommandConnectionHelper(cmd string, flags ...string) (*ConnectionHelper, error) {
```
thaJeztah: Wondering why we didn't call the other one `NewConnectionHelper` (and this one `NewCommandConnectionHelper`)

(just thinking out loud 🤔)

cli/connhelper/connhelper.go
```diff
@@ -53,6 +53,16 @@ func GetConnectionHelper(daemonURL string) (*ConnectionHelper, error) {
...
+	return &ConnectionHelper{
+		Dialer: func(ctx context.Context, network, addr string) (net.Conn, error) {
+			return newCommandConn(ctx, cmd, flags...)
+		},
+		Host: "http://docker",
```
thaJeztah: Also thinking out loud while trying to grasp this PR; we should add comments with these (future readers may be confused and think "what's this magic `docker` host?"

docs/extend/cli_plugins.md
```diff
@@ -75,6 +75,19 @@ A plugin is required to support all of the global options of the
 top-level CLI, i.e. those listed by `man docker 1` with the exception
 of `-v`.
 
+## Connecting to the docker engine
+
+For consistency plugins should prefer to dial the engine by using the
+`system dial-stdio` subcommand of the main Docker CLI binary.
+
+To facilitate this plugins will be executed with the
+`$DOCKER_CLI_PLUGIN_ORIGINAL_CLI_COMMAND` environment variable
```
thaJeztah: Given that plugins (when called through the `docker` cli) are executed as child-process of `docker` Would there be an approach to get the parent ID, and get the information from that? 🤔 (something like [`os.Getppid()`](https://golang.org/pkg/os/#Getppid), or do we need more information than the executable itself?

https://www.socketloop.com/tutorials/golang-get-executable-name-behind-process-id-example

cli/command/cli.go
ijc: Oops! The perils of enabling autocomplete means these things get propagated!

cli/connhelper/connhelper.go
ijc: This came pretty much verbatim from the existing example, I don't know where it came from now really where it shows up...

docs/extend/cli_plugins.md
ijc: That seems very platform specific (I don't know about it's portability to MacoS, let alone Windows) and probably quite fragile even on Linux compared with this approach which is pretty simple.

ijc:
@thaJeztah ping -- I think your remaining comments were just thinking out loud rather than things that should be changed.

ijc:
> LGTM, even if I think the design seems fragile: trying to call the CLI back with what we hope to be the same context (all the global flags we retrieve), but I don't see any cleaner way to do it.

This is certainly true. The other choice is the status quo which is to do the dial in the plugin itself, with whatever capabilities the vendored cli libraries happen to have, which may differ from what the calling top-level CLI supports (which might be confusing for users).

cli/command/cli.go
```diff
@@ -189,29 +206,32 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions) error {
...
-	cli.dockerEndpoint = endpoint
 
-	cli.client, err = newAPIClientFromEndpoint(endpoint, cli.configFile)
-	if tlsconfig.IsErrEncryptedKey(err) {
-		passRetriever := passphrase.PromptRetrieverWithInOut(cli.In(), cli.Out(), nil)
-		newClient := func(password string) (client.APIClient, error) {
-			endpoint.TLSPassword = password
-			return newAPIClientFromEndpoint(endpoint, cli.configFile)
+	if cli.client == nil {
+		var err error
```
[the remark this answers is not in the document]
ijc: This was preexisting code which was just moved in a level by the addition of the `if` block. It's only used in the following assignment and the one error check which follows, before it gets shadowed by a subsequent `:=` type assignment.

Not sure what sort of masking you foresee which would be a problem in this context, but I will move anyway.

thaJeztah:
Let's get this merged; thanks!

ijc:
Thanks, I ticked the box in #1661.

ijc:
Hrm, I hadn't appreciated that this would require 18.09 on the daemon side;
```
error during connect: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/json?all=1&«...»&limit=0: command [docker system dial-stdio] has exited with exit status 1, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=
```

This seems like a problem to me, @thaJeztah @tonistiigi WDYT?

Not sure how to address, since figuring out the daemon version requires calling out to the dameon, which currently required dial-stdio, rinse and repeat...

ijc:
The above was due to me misreading the logs -- in fact it was telling me that my test had accidentally found `/usr/bin/docker` which was 18.06 and this didn't support `docker system dial-stdio`. Now that I fixed the test to find the right cli this aspect is ok.

However, it appears that dial-stdio does not work on Windows:
```
docker@WIN-NUC0 C:\Users\docker>.\docker-windows-amd64.exe version
Client:
 Version:           19.03.0-dev
 API version:       1.39 (downgraded from 1.40)
 Go version:        go1.11.5
 Git commit:        8ddde26a
 Built:             Mon Mar  4 16:34:00 2019
 OS/Arch:           windows/amd64
 Experimental:      false

 Engine:
  Version:          18.09.0
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.4
  Git commit:       4d60db4
  Built:            Wed Nov  7 00:55:00 2018
  OS/Arch:          linux/amd64
  Experimental:     false

docker@WIN-NUC0 C:\Users\docker>.\docker-windows-amd64.exe system dial-stdio
```
Basically it seems that `net.Conn` doesn't have `ReadClose` on Windows (from my tests it does have `WriteClose`):
```
docker@WIN-NUC0 C:\Users\docker>.\docker-windows-amd64.exe -D system dial-stdio
DEBU[0000] conn is a *winio.win32MessageBytePipe
DEBU[0000] conn is a halfReadCloser: false
DEBU[0000] conn is a halfWriteCloser: true
```

(this seems to be independent of plugins using this, it's just the base functionality is not there in windows)

tonistiigi:
@AkihiroSuda If `ReadClose` is missing in win, could we add a fallback that doesn't need that.

ijc:
I raised https://github.com/docker/cli/pull/1710 in the meantime.

PR #1659, opened 2019-02-05
Added builder info

kwojcicki:
Signed-off-by: Krystian Wojcicki <kwojcicki@sympatico.ca>

Client part of https://github.com/moby/moby/issues/38518

**- What I did**
Added the builder version in the docker info command

**- How I did it**
reading the response from the server and printing out the corresponding version
also updated the version of moby/moby being used to get the new struct information.

**- How to verify it**
typing docker info/the tests

**- Description for the changelog**
Docker info prints the default builder 

**- A picture of a cute animal (not mandatory but encouraged)**
image

PR #1660, opened 2019-02-06
Feature/38518 Add builder info

kwojcicki:
PR Re-opened from #1659 

Signed-off-by: Krystian Wojcicki <kwojcicki@sympatico.ca>

Client part of https://github.com/moby/moby/issues/38518

**- What I did**
Added the builder version in the docker info command

**- How I did it**
reading the response from the server and printing out the corresponding version
also updated the version of moby/moby being used to get the new struct information.

**- How to verify it**
typing docker info/the tests

Docker info prints the default builder 

image


thaJeztah:
Looks like this needs a rebase now, because another PR brought in the "moby" bump.

There's still some debate about the API changes; if we want to include just the version or if we need to include more information about the builder in future (in which case, we may want to change the API response to use a "struct" for builder).

I was also thinking what this info would be used for; including (e.g.) `BuildKit` as a component in the `/version` output? (similar to https://github.com/moby/moby/pull/37974)

kwojcicki:
@thaJeztah would you happen to know the correct way to update the golden files? I had to hack at the makefile to set the flag correctly? And also will try rebasing.

thaJeztah:
hm, good one, I think I ran it manually last time 🤔

but, might want to join the discussion on https://github.com/moby/moby/pull/38688

thaJeztah:
closing per the discussion in https://github.com/moby/moby/pull/38688#issuecomment-461549262

but thanks so much!

PR #1663, opened 2019-02-07
Add docs for --device option in Windows

cwilhit:
Signed-off-by: Craig Wilhite <crwilhit@microsoft.com>

Adds documentation portion for #1606. Includes a reference for how to run with --device on Windows & explain limitations of the feature (no Hyper-V isolation, no LCOW).

@jhowardmsft FYI

### DIFF
--- a/docs/reference/commandline/run.md
+++ b/docs/reference/commandline/run.md
@@ -583,6 +583,28 @@ fdisk: unable to open /dev/xvdc: Operation not permitted
 > that may be removed should not be added to untrusted containers with
 > `--device`.
 
+For Windows, the format of the string passed to the `--device` option is in
+the form of `--device=<IdType>/<Id>`. Beginning with Windows Server 2019
+and Windows 10 October 2018 Update, Windows only supports an IdType of
+`class` and the Id as a [device interface class
+GUID](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/overview-of-device-interface-classes).
+Refer to the table defined in the [Windows container
+docs](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/hardware-devices-in-containers)
+for a list of container-supported device interface class GUIDs.
+
+If this option is specified for a process-isolated Windows container, _all_
+devices that implement the requested device interface class GUID are made
+available in the container. For example, the command below makes all COM
+ports on the host visible in the container.
+
+```powershell
+PS C:\> docker run --device=class/86E0D1E0-8089-11D0-9CE4-08003E301F73 mcr.microsoft.com/windows/servercore:ltsc2019
+```
+
+> **Note**: the `--device` option is only supported on process-isolated
+> Windows containers. This option fails if the container isolation is `hyperv`
+> or when running Linux Containers on Windows (LCOW).
+
 ### Restart policies (--restart)
 
 Use Docker's `--restart` to specify a container's *restart policy*. A restart


docs/reference/commandline/run.md
```diff
@@ -583,6 +583,17 @@ fdisk: unable to open /dev/xvdc: Operation not permitted
 > that may be removed should not be added to untrusted containers with
 > `--device`.
 
+For Windows, the format of the string passed to the `--device` option is in the form of `--device=<IdType>/<Id>`. Today, Windows only supports an IdType of `class` and the Id as a [device interface class GUID](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/overview-of-device-interface-classes). For a list of container-supported device interface class GUIDs, please refer to the table defined in the [Windows container docs](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/hardware-devices-in-containers).
```
jhowardmsft: Rather than `Today`, perhaps state what versions of Windows. (AFAIK, this is RS5/1809 on only).

cwilhit: Fixed

docs/reference/commandline/run.md
```diff
@@ -583,6 +583,17 @@ fdisk: unable to open /dev/xvdc: Operation not permitted
 > that may be removed should not be added to untrusted containers with
 > `--device`.
 
+For Windows, the format of the string passed to the `--device` option is in the form of `--device=<IdType>/<Id>`. Beginning with Windows Server 2019 and Windows 10 October 2018 Update, Windows only supports an IdType of `class` and the Id as a [device interface class GUID](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/overview-of-device-interface-classes). For a list of container-supported device interface class GUIDs, please refer to the table defined in the [Windows container docs](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/hardware-devices-in-containers).
```
thaJeztah: Trying to avoid "please" in the documentation; perhaps 

```Markdown
Refer to the table defined in the [Windows container
documentation](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/hardware-devices-in-containers)
for a list of container-supported device interface class GUIDs.
```

Also, could you wrap these to ~80 chars?

```suggestion
For Windows, the format of the string passed to the `--device` option is in
the form of `--device=<IdType>/<Id>`. Beginning with Windows Server 2019
and Windows 10 October 2018 Update, Windows only supports an IdType of
`class` and the Id as a [device interface class
GUID](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/overview-of-device-interface-classes).
Refer to the table defined in the [Windows container
documentation](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/hardware-devices-in-containers)
for a list of container-supported device interface class GUIDs.
```

docs/reference/commandline/run.md
```diff
@@ -583,6 +583,17 @@ fdisk: unable to open /dev/xvdc: Operation not permitted
...
+
+When this option is specified for a process-isolated Windows container, _all_ devices which implement the requested device interface class GUID will be made available in the container. For example, the command below will make all COM ports on the host visibile in the container.
```
thaJeztah: Should it be "devices _that_ implement" (instead of "which"?)

Also, try to avoid future tense in technical documentation; here's a quick go at changing it to remove the future tense, and wrap the paragraph to 80 chars;

```suggestion
If this option is specified for a process-isolated Windows container, _all_
devices that implement the requested device interface class GUID are made
available in the container. For example, the command below makes all COM
ports on the host visible in the container.
```

docs/reference/commandline/run.md
```diff
@@ -583,6 +583,17 @@ fdisk: unable to open /dev/xvdc: Operation not permitted
...
+
+```powershell
+PS C:\> docker run --device=class/86E0D1E0-8089-11D0-9CE4-08003E301F73 mcr.microsoft.com/windows/servercore:ltsc2019 
+```
+
+> **Note**: `--device` may only be used on Windows with process-isolated Windows containers. This option will fail if the container
```
thaJeztah: Slightly changed to wrap to 80-chars, and remove future tense;

```suggestion
> **Note**: the `--device` option is only supported on process-isolated
> Windows containers. This option fails if the container isolation is `hyperv`
> or when running Linux Containers on Windows (LCOW).
```

docs/reference/commandline/run.md
cwilhit: Thanks, added these fixes.

docs/reference/commandline/run.md
cwilhit: You are correct--I've now learned when to use 'which' vs 'that'. It always confounded me in the past.

docs/reference/commandline/run.md
cwilhit: Added

cwilhit:
@thaJeztah Ok, incorporated the changes. This is the first time I've squashed commits before, but I believe I got it correctly. Per instructions online, I had to force push to my branch on Github.

docs/reference/commandline/run.md
```diff
@@ -583,6 +583,30 @@ fdisk: unable to open /dev/xvdc: Operation not permitted
 > that may be removed should not be added to untrusted containers with
 > `--device`.
 
+For Windows, the format of the string passed to the `--device` option is in
+the form of `--device=<IdType>/<Id>`. Beginning with Windows Server 2019
+and Windows 10 October 2018 Update, Windows only supports an IdType of
+`class` and the Id as a [device interface class
+GUID](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/overview-of-device-
```
thaJeztah: oh; the URL itself cannot be wrapped (it's ok for the caption, but not for the URL)

docs/reference/commandline/run.md
```diff
@@ -583,6 +583,30 @@ fdisk: unable to open /dev/xvdc: Operation not permitted
...
+interface-classes).
+Refer to the table defined in the [Windows container
+docs](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy
```
thaJeztah: same here

PR #1665, opened 2019-02-08
Note caveat with detaching using key sequence

jcrben:
This has come up a few times, for example:
* https://stackoverflow.com/questions/19688314/how-do-you-attach-and-detach-from-dockers-process#comment40262294_19689048
* https://github.com/moby/moby/issues/20864 
* https://github.com/moby/moby/issues/35491
* https://github.com/moby/moby/issues/20227#issuecomment-221229241
* https://forums.docker.com/t/ctrl-p-ctrl-q-does-not-detach/40273

**- What I did**
N/A
**- How I did it**
N/A
**- How to verify it**
N/A
Update documentation of `docker attach` to note caveat with detaching using key sequence.

image
(from https://commons.wikimedia.org/wiki/File:Beagle_puppy_Cadet_2.jpg)

### DIFF
--- a/docs/reference/commandline/attach.md
+++ b/docs/reference/commandline/attach.md
@@ -44,8 +44,8 @@ from different sessions on the Docker host.
 
 To stop a container, use `CTRL-c`. This key sequence sends `SIGKILL` to the
 container. If `--sig-proxy` is true (the default),`CTRL-c` sends a `SIGINT` to
-the container. You can detach from a container and leave it running using the
- `CTRL-p CTRL-q` key sequence.
+the container. If the container was run with `-i` and `-t`, you can detach from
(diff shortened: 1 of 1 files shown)


jcrben:
My commit is Github "Verified"; kinda surprised that isn't enough. I'll see about pulling it down.

I tested using `-t` and `-i` independently (with an image I had on hand, redis) and it didn't work - needed both and then it emitted `read escape sequence` and closed.

thaJeztah:
> My commit is Github "Verified"; kinda surprised that isn't enough. I'll see about pulling it down.

Yes, sorry about that; it's checking for a "DCO" sign-off (which is different than the commit being GPG-signed).

PR #1670, opened 2019-02-13
Corrected max-file option - was incorrectly spelt as max-files

stevejr:
Signed-off-by: stevejr <>

**- What I did**

**- How I did it**

**- How to verify it**
Used the option spelt as max-files and confirmed the docker daemon would not restart. Amended to max-file and it started

Corrected the spelling of the max-file sub-option for log-opts - was spelt as max-files but should be max-file as per https://docs.docker.com/config/containers/logging/json-file/

### DIFF
--- a/docs/reference/commandline/dockerd.md
+++ b/docs/reference/commandline/dockerd.md
@@ -1312,7 +1312,7 @@ This is a full example of the allowed configuration options on Linux:
 	"log-driver": "json-file",
 	"log-opts": {
 		"max-size": "10m",
-		"max-files":"5",
+		"max-file":"5",
 		"labels": "somelabel",
 		"env": "os,customer"
 	},


thaJeztah:
Nice catch, thanks!

Confirmed that this is the correct name for this option;

```bash
mkdir -p /etc/docker
echo '{"log-opts": {"max-files":"5"}}' > /etc/docker/daemon.json
dockerd

...
Error starting daemon: failed to set log opts: unknown log opt 'max-files' for json-file log driver

echo '{"log-opts": {"max-file":"5"}}' > /etc/docker/daemon.json
...
INFO[2019-02-13T12:04:36.215638800Z] Daemon has completed initialization          
INFO[2019-02-13T12:04:36.226260100Z] API listen on /var/run/docker.sock        
```

stevejr:
@thaJeztah - Signed off text changed

PR #1671, opened 2019-02-13
Fix labels copying value from environment variables

thaJeztah:
This patch fixes a bug where labels use the same behavior as `--env`, resulting
in a value to be copied from environment variables with the same name as the
label if no value is set (i.e. a simple key, no `=` sign, no value).

An earlier pull request addressed similar cases for `docker run`;
2b17f4c8a8caad552025edb05a73db683fb8a5c6, but this did not address the
same situation for (e.g.) `docker service create`.

Digging in history for this bug, I found that use of the `ValidateEnv`
function for labels was added in the original implementation of the labels feature in
https://github.com/docker/docker/commit/abb5e9a0777469e64fe2c7ecfa66ea01083d2071#diff-ae476143d40e21ac0918630f7365ed3cR34

However, the design never intended it to expand environment variables,
and use of this function was either due to either a "copy/paste" of the
equivalent `--env` flags, or a misunderstanding (the name `ValidateEnv` does
not communicate that it also expands environment variables), and the existing
`ValidateLabel` was designed for _engine_ labels (which required a value to
be set).

Following the initial implementation, other parts of the code followed
the same (incorrect) approach, therefore leading the bug to be introduced
in services as well.

This patch:

- updates the `ValidateLabel` to match the expected validation
rules (this function is no longer used since 31dc5c0a9a8bdc11c7ad335aebb753ed527caa5a),
and the daemon has its own implementation)
- corrects various locations in the code where `ValidateEnv` was used instead of `ValidateLabel`.

Before this patch:

```bash
export SOME_ENV_VAR=I_AM_SOME_ENV_VAR
docker service create --label SOME_ENV_VAR --tty --name test busybox

docker service inspect --format '{{json .Spec.Labels}}' test
{"SOME_ENV_VAR":"I_AM_SOME_ENV_VAR"}
```

After this patch:

```bash
export SOME_ENV_VAR=I_AM_SOME_ENV_VAR
docker service create --label SOME_ENV_VAR --tty --name test busybox

docker container inspect --format '{{json .Config.Labels}}' test
{"SOME_ENV_VAR":""}
```

### Add back validation for invalid label values on containers

This adds validation to `docker container run` / `docker container create`;

Validation of labels provided through flags was removed in 31dc5c0a9a8bdc11c7ad335aebb753ed527caa5a,
after the validation was changed to fix labels without values, and to prevent
labels from being expanded with environment variables in 2b17f4c8a8caad552025edb05a73db683fb8a5c6

However, now empty label names from _files_ (`--label-file`) followed different
validation rules than labels passed through `--label`.

This patch adds back minimal validation for labels passed through the command-line

Before this patch:

```bash
docker container create \
--name label \
--label==with-leading-equal-sign \
--label=without-value \
--label=somelabel=somevalue \
--label " = " \
--label=with-quotes-in-value='{"foo"}' \
--label='with"quotes"in-key=test' \
busybox

docker container inspect --format '{{json .Config.Labels}}' label
```

```json
{
"": "with-leading-equal-sign",
" ": " ",
"somelabel": "somevalue",
"with\"quotes\"in-key": "test",
"with-quotes-in-value": "{\"foo\"}",
"without-value": ""
}
```

After this patch:

```bash
docker container create \
--name label \
--label==with-leading-equal-sign \
--label=without-value \
--label=somelabel=somevalue \
--label " = " \
--label=with-quotes-in-value='{"foo"}' \
--label='with"quotes"in-key=test' \
busybox

invalid argument "=with-leading-equal-sign" for "-l, --label" flag: invalid label format: "=with-leading-equal-sign"
```

### DIFF
--- a/opts/opts.go
+++ b/opts/opts.go
@@ -281,10 +281,10 @@ func ValidateLabel(val string) (string, error) {
 	arr := strings.SplitN(val, "=", 2)
 	key := strings.TrimLeft(arr[0], whiteSpaces)
 	if key == "" {
-		return "", fmt.Errorf("empty label name: '%s'", val)
+		return "", fmt.Errorf("invalid label '%s': empty name", val)
 	}
 	if strings.ContainsAny(key, whiteSpaces) {
-		return "", fmt.Errorf("label '%s' has white spaces", key)
+		return "", fmt.Errorf("label '%s' contains whitespaces", key)
 	}
 	return val, nil
 }
--- a/opts/opts.go
+++ b/opts/opts.go
@@ -267,10 +267,24 @@ func validateDomain(val string) (string, error) {
 }
 
 // ValidateLabel validates that the specified string is a valid label, and returns it.
-// Labels are in the form on key=value.
+//
+// Labels are in the form of key=value; key must be a non-empty string, and not
+// contain whitespaces. A value is optional (defaults to an empty string if omitted).
+//
+// Leading whitespace is removed during validation but values are kept as-is
+// otherwise, so any string value is accepted for both, which includes whitespace
+// (for values) and quotes (surrounding, or embedded in key or value).
+//
+// TODO discuss if quotes (and other special characters) should be valid or invalid for keys
+// TODO discuss if leading/trailing whitespace in keys should be preserved (and valid)
 func ValidateLabel(val string) (string, error) {
-	if strings.Count(val, "=") < 1 {
-		return "", fmt.Errorf("bad attribute format: %s", val)
+	arr := strings.SplitN(val, "=", 2)
+	key := strings.TrimLeft(arr[0], whiteSpaces)
+	if key == "" {
+		return "", fmt.Errorf("empty label name: '%s'", val)
+	}
+	if strings.ContainsAny(key, whiteSpaces) {
+		return "", fmt.Errorf("label '%s' has white spaces", key)
 	}
 	return val, nil
 }
--- a/opts/opts_test.go
+++ b/opts/opts_test.go
@@ -184,17 +184,17 @@ func TestValidateLabel(t *testing.T) {
 	}{
 		{
 			name:        "empty",
-			expectedErr: `empty label name: ''`,
+			expectedErr: `invalid label '': empty name`,
 		},
 		{
 			name:        "whitespace only ",
 			value:       " ",
-			expectedErr: `empty label name: ' '`,
+			expectedErr: `invalid label ' ': empty name`,
 		},
 		{
 			name:        "whitespace around equal-sign",
 			value:       " = ",
-			expectedErr: `empty label name: ' = '`,
+			expectedErr: `invalid label ' = ': empty name`,
 		},
 		{
 			name:  "leading whitespace",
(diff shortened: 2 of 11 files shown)


thaJeztah:
@silvin-lubecki @vdemeester ptal

partially thinking if we should also move more of this validation to the daemon 🤔

silvin-lubecki:
Were you an archeologist in a previous life @thaJeztah 😆 ? The description is impressive!

opts/opts.go
```diff
@@ -267,10 +267,24 @@ func validateDomain(val string) (string, error) {
...
+// otherwise, so any string value is accepted for both, which includes whitespace
+// (for values) and quotes (surrounding, or embedded in key or value).
+//
+// TODO discuss if quotes (and other special characters) should be valid or invalid for keys
+// TODO discuss if leading/trailing whitespace in keys should be preserved (and valid)
 func ValidateLabel(val string) (string, error) {
-	if strings.Count(val, "=") < 1 {
-		return "", fmt.Errorf("bad attribute format: %s", val)
+	arr := strings.SplitN(val, "=", 2)
+	key := strings.TrimLeft(arr[0], whiteSpaces)
```
silvin-lubecki: Why trimming these whitespaces in the key?
It contradicts your comment:
> key must be a non-empty string, and not contain whitespaces.

opts/opts.go
```diff
@@ -267,10 +267,24 @@ func validateDomain(val string) (string, error) {
...
+	if key == "" {
+		return "", fmt.Errorf("empty label name: '%s'", val)
```
silvin-lubecki: Isn't the label `name` the `key`? If so you should replace `val` by `key`.

silvin-lubecki: Well ok as the key is empty, there's no point printing it in the message... Then maybe rephrase it with something like: `"invalid label '%s': empty name", val` ?

opts/opts_test.go
```diff
@@ -175,19 +177,104 @@ func TestValidateDNSSearch(t *testing.T) {
...
+			expectedErr: `empty label name: ' '`,
+		},
+		{
+			name:        "whitespace around equal-sign",
+			value:       " = ",
+			expectedErr: `empty label name: ' = '`,
+		},
+		{
+			name:  "leading whitespace",
+			value: "    label=value",
```
silvin-lubecki: Do we really want to validate this? Or maybe that's already something valid and trimmed on the engine side?

opts/opts_test.go
```diff
@@ -175,19 +177,104 @@ func TestValidateDNSSearch(t *testing.T) {
...
+		},
+		{
+			name:        "whitespaces in key without value",
+			value:       "this is a label without value",
+			expectedErr: `label 'this is a label without value' has white spaces`,
+		},
+		{
+			name:        "whitespaces in key",
+			value:       "this is a label=value",
+			expectedErr: `label 'this is a label' has white spaces`,
```
silvin-lubecki: `has white spaces` or `has whitespaces` ? 🤔

silvin-lubecki:
> partially thinking if we should also move more of this validation to the daemon 🤔

What do you mean? Which part do you want to move? I think the validation is simple enough on the client side 🤔

opts/opts.go
thaJeztah: Yes, it's not proper english; I copied these from https://github.com/docker/cli/blob/9e712073274021ef1e552ecbb54595142b49edbb/opts/file.go#L51-L54

but perhaps I should update both

opts/opts.go
thaJeztah: Hm, yeah, it may not _contain_ whitespace, but leading whitespace is removed. Basically, I copied the logic from https://github.com/docker/cli/blob/9e712073274021ef1e552ecbb54595142b49edbb/opts/file.go#L51-L54 to be consistent

thaJeztah:
updated PTAL

PR #1674, opened 2019-02-15
Add exit status to docker exec manpage

ericcurtin:
There's little way of knowing what each exit status means at present
because it's not documented. I'm assuming they are the same as
docker run.

### DIFF
--- a/man/src/container/exec.md
+++ b/man/src/container/exec.md
@@ -23,3 +23,29 @@ the same capabilities as the container, which may be limited. Set
    --user [user | user:group | uid | uid:gid | user:gid | uid:group ]
 
    Without this argument the command will be run as root in the container.
+
+# Exit Status
+
+The exit code from `docker exec` gives information about why the container
+failed to exec or why it exited.  When `docker exec` exits with a non-zero code,
+the exit codes follow the `chroot` standard, see below:
+
+**_126_** if the **_contained command_** cannot be invoked
+
+    $ docker exec busybox /etc; echo $?
+    # exec: "/etc": permission denied
+      docker: Error response from daemon: Contained command could not be invoked
+      126
+
+**_127_** if the **_contained command_** cannot be found
+
+    $ docker exec busybox foo; echo $?
+    # exec: "foo": executable file not found in $PATH
+      docker: Error response from daemon: Contained command not found or does not exist
+      127
+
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @vdemeester @silvin-lubecki ptal

man/src/container/exec.md
```diff
@@ -23,3 +23,36 @@ the same capabilities as the container, which may be limited. Set
...
+The exit code from `docker exec` gives information about why the container
+failed to exec or why it exited.  When `docker exec` exits with a non-zero code,
+the exit codes follow the `chroot` standard, see below:
+
+**_125_** if the error is with Docker daemon **_itself_** 
+
+    $ docker exec --foo busybox; echo $?
+    # flag provided but not defined: --foo
+      See 'docker exec --help'.
+      125
```
ijc: Isn't this example of an error within the Docker *CLI*, rather than the daemon, it's the CLI which isn't recognising the `--foo` here isn't it?

thaJeztah: Good question; will have to dig into the history of that. Perhaps this was done on purpose to being able to distinguish a `exit 1` (from the container's process) from an `exit 1` (from the docker cli (or daemon))?

ericcurtin: I can a different example if you guys have a common one off the top of your heads! I'm away for the weekend now, can try and recreate a 125 when I get back.

thaJeztah: not sure if another example is possible (as in; I think the `125` is generated in the CLI, as it won't send the flag to the daemon if it's invalid 🤔

ijc: @thaJeztah IOW it is the description not the example which is wrong both here and in https://github.com/docker/cli/blob/561f6e399c3a0863c7b78ab797553b878cc4e73a/docs/reference/run.md#exit-status?

ericcurtin: Will I remove it for the moment?

ericcurtin:
Removed 125 exit status from man page

ericcurtin:
Poke poke @ijc @thaJeztah @vdemeester @silvin-lubecki 

I removed 125. Is this good to go?

ericcurtin:
@thaJeztah @vdemeester @silvin-lubecki any of you guys want to grab?

Issue #1661, opened 2019-02-07
Followup to #1564 (cli plugins)


ijc:
From [#1564 (comment)](https://github.com/docker/cli/pull/1564/#issuecomment-458964609):

- [ ] https://github.com/docker/cli/pull/1564#discussion_r251818285 handling of `docker someplugin --version`. The `--version` / `-v` flag is handled by the plugin root-command, and therefore shows the `docker` version information (`Docker version 19.03.0-dev, build c8f1f1ee`). We should fix this to be handled by the `plugin`.
- [ ] Related to the above; consider making `--version` / `-v` part of the pluging "API", so that a plugin can advertise its version (both when calling the plugin directly (`docker-myplugin --version`), and when calling through the `docker` cli (`docker myplugin --version`). Perhaps we should add a standard function and/or define a standard format (`docker-myplugin version X, commit deadbeef (ACME, Inc.)`)
- [ ] https://github.com/docker/cli/pull/1564#discussion_r251819147 install location for plugins; choose between `~/.docker/plugins` or `~/.docker/cli-plugins`
- [ ] https://github.com/docker/cli/pull/1564#discussion_r251816130 decide on the output format in `docker --help`: make plugin-command stand out from built-in commands, or make them appear as regular commands with a non-obtrusive indicator (`*`)
- [ ] https://github.com/docker/cli/pull/1564#discussion_r251814998 decide on the location to show version and vendor for plugins in in `docker info`
- [ ] https://github.com/docker/cli/pull/1564#discussion_r251890532 small nit: use `displayutils.Ellipsis()` instead of a custom implementation.
- [ ] https://github.com/docker/cli/pull/1564#discussion_r251822916 small nit: hardening `config.Path()` to disallow directory traversal
- [ ] https://github.com/docker/cli/pull/1564#discussion_r251825582 small nit: fix usage output of the sample plugin

Others:
* [ ] plugins shuld be able to indicate if they are "management" vs not commands in their metadata and be listed accordingly in e.g. `docker help`.
* [ ] plugins should have a field in `config.json` for their use, see *Configuration* in [design](https://github.com/docker/cli/issues/1534) (see #1652)
* [ ] plugins should call back to main cli using `dial-stdio` to connect to engine. From https://github.com/docker/cli/issues/1534#issuecomment-446206041 (see #1654).

thaJeztah:
noticed another thing; open a separate issue for that https://github.com/docker/cli/issues/1682

PR #1675, opened 2019-02-15
Reformat plugin's vendor position and add version on --help

ulyssessouza:
Checks the fourth box on: https://github.com/docker/cli/issues/1661

**- What I did**
- Refactor plugins' vendor location on --help by relocating it at the end of the short description
-  Add ", <version>" to command vendor

**- How I did it**
By changing the `usageTemplate` in cli/cobra.go to replace the information and decorating the command with a '*' when it's a plugin's first level command

**- How to verify it**
- Install a plugin (or make your configuration point to one)
- Execute `docker help` and check that the plugins' first level commands have now a '*' and it's vendor as short description suffix

Reformated plugins' first level command description on `--help` by replacing and vendor and adding version

salamander

### DIFF
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -22,6 +22,7 @@ func setupCommonRootCommand(rootCmd *cobra.Command) (*cliflags.ClientOptions, *p
 	flags.StringVar(&opts.ConfigDir, "config", cliconfig.Dir(), "Location of client config files")
 	opts.Common.InstallFlags(flags)
 
+	cobra.AddTemplateFunc("add", func(a, b int) int { return a + b })
 	cobra.AddTemplateFunc("hasSubCommands", hasSubCommands)
 	cobra.AddTemplateFunc("hasManagementSubCommands", hasManagementSubCommands)
 	cobra.AddTemplateFunc("hasInvalidPlugins", hasInvalidPlugins)
@@ -29,9 +30,10 @@ func setupCommonRootCommand(rootCmd *cobra.Command) (*cliflags.ClientOptions, *p
 	cobra.AddTemplateFunc("managementSubCommands", managementSubCommands)
 	cobra.AddTemplateFunc("invalidPlugins", invalidPlugins)
 	cobra.AddTemplateFunc("wrappedFlagUsages", wrappedFlagUsages)
-	cobra.AddTemplateFunc("commandVendor", commandVendor)
-	cobra.AddTemplateFunc("isFirstLevelCommand", isFirstLevelCommand) // is it an immediate sub-command of the root
+	cobra.AddTemplateFunc("vendorAndVersion", vendorAndVersion)
 	cobra.AddTemplateFunc("invalidPluginReason", invalidPluginReason)
+	cobra.AddTemplateFunc("isPlugin", isPlugin)
+	cobra.AddTemplateFunc("decoratedName", decoratedName)
 
 	rootCmd.SetUsageTemplate(usageTemplate)
 	rootCmd.SetHelpTemplate(helpTemplate)
@@ -155,19 +157,23 @@ func wrappedFlagUsages(cmd *cobra.Command) string {
 	return cmd.Flags().FlagUsagesWrapped(width - 1)
 }
 
-func isFirstLevelCommand(cmd *cobra.Command) bool {
-	return cmd.Parent() == cmd.Root()
+func decoratedName(cmd *cobra.Command) string {
+	decoration := " "
+	if isPlugin(cmd) {
+		decoration = "*"
+	}
+	return cmd.Name() + decoration
 }
 
-func commandVendor(cmd *cobra.Command) string {
-	width := 13
-	if v, ok := cmd.Annotations[pluginmanager.CommandAnnotationPluginVendor]; ok {
-		if len(v) > width-2 {
-			v = v[:width-3] + "…"
+func vendorAndVersion(cmd *cobra.Command) string {
+	if vendor, ok := cmd.Annotations[pluginmanager.CommandAnnotationPluginVendor]; ok && isPlugin(cmd) {
+		version := ""
+		if v, ok := cmd.Annotations[pluginmanager.CommandAnnotationPluginVersion]; ok && v != "" {
+			version = ", " + v
 		}
-		return fmt.Sprintf("%-*s", width, "("+v+")")
+		return fmt.Sprintf("(%s%s)", vendor, version)
 	}
-	return strings.Repeat(" ", width)
+	return ""
 }
 
 func managementSubCommands(cmd *cobra.Command) []*cobra.Command {
@@ -230,7 +236,7 @@ Options:
 Management Commands:
 
 {{- range managementSubCommands . }}
-  {{rpad .Name .NamePadding }} {{ if isFirstLevelCommand .}}{{commandVendor .}} {{ end}}{{.Short}}
+  {{rpad (decoratedName .) (add .NamePadding 1)}}{{.Short}}{{ if isPlugin .}} {{vendorAndVersion .}}{{ end}}
 {{- end}}
 
 {{- end}}
@@ -239,7 +245,7 @@ Management Commands:
 Commands:
 
 {{- range operationSubCommands . }}
-  {{rpad .Name .NamePadding }} {{ if isFirstLevelCommand .}}{{commandVendor .}} {{ end}}{{.Short}}
+  {{rpad (decoratedName .) (add .NamePadding 1)}}{{.Short}}{{ if isPlugin .}} {{vendorAndVersion .}}{{ end}}
 {{- end}}
 {{- end}}
 
--- a/cli/cobra_test.go
+++ b/cli/cobra_test.go
@@ -32,28 +32,29 @@ func TestVisitAll(t *testing.T) {
 	assert.DeepEqual(t, expected, visited)
 }
 
-func TestCommandVendor(t *testing.T) {
+func TestVendorAndVersion(t *testing.T) {
 	// Non plugin.
-	assert.Equal(t, commandVendor(&cobra.Command{Use: "test"}), "             ")
+	assert.Equal(t, vendorAndVersion(&cobra.Command{Use: "test"}), "")
 
 	// Plugins with various lengths of vendor.
 	for _, tc := range []struct {
 		vendor   string
+		version  string
 		expected string
 	}{
-		{vendor: "vendor", expected: "(vendor)     "},
-		{vendor: "vendor12345", expected: "(vendor12345)"},
-		{vendor: "vendor123456", expected: "(vendor1234…)"},
-		{vendor: "vendor1234567", expected: "(vendor1234…)"},
+		{vendor: "vendor", expected: "(vendor)"},
+		{vendor: "vendor", version: "testing", expected: "(vendor, testing)"},
 	} {
 		t.Run(tc.vendor, func(t *testing.T) {
 			cmd := &cobra.Command{
 				Use: "test",
 				Annotations: map[string]string{
-					pluginmanager.CommandAnnotationPluginVendor: tc.vendor,
+					pluginmanager.CommandAnnotationPlugin:        "true",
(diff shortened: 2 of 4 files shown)


thaJeztah:
Giving this a quick spin;

```bash
DOCKER_HIDE_LEGACY_COMMANDS=1 docker --config=.docker-plugin-test --help

Usage:	docker [OPTIONS] COMMAND

A self-sufficient runtime for containers

Options:
      --config string      Location of client config files (default "/root/.docker")
  -D, --debug              Enable debug mode
  -l, --log-level string   Set the logging level ("debug"|"info"|"warn"|"error"|"fatal") (default "info")
      --tls                Use TLS; implied by --tlsverify
      --tlscacert string   Trust certs signed only by this CA (default "/root/.docker/ca.pem")
      --tlscert string     Path to TLS certificate file (default "/root/.docker/cert.pem")
      --tlskey string      Path to TLS key file (default "/root/.docker/key.pem")
      --tlsverify          Use TLS and verify the remote
  -v, --version            Print version information and quit

Management Commands:
  builder     Manage builds
  checkpoint  Manage checkpoints
  config      Manage Docker configs
  container   Manage containers
  context     Manage contexts
  engine      Manage the docker engine
  image       Manage images
  network     Manage networks
  node        Manage Swarm nodes
  plugin      Manage plugins
  secret      Manage Docker secrets
  service     Manage services
  stack       Manage Docker stacks
  swarm       Manage Swarm
  system      Manage Docker
  trust       Manage trust on Docker images
  volume      Manage volumes

Commands:
  build       Build an image from a Dockerfile
  helloworld* A basic Hello World plugin for tests (Docker Inc., testing)
  login       Log in to a Docker registry
  logout      Log out from a Docker registry
  run         Run a command in a new container
  search      Search the Docker Hub for images
  version     Show the Docker version information

Run 'docker COMMAND --help' for more information on a command.
```


The new presentation of the plugin looks good to me 🤗

thaJeztah:
I just realise that I may not have mentioned that on the other PR (and probably something to address separately);

Plugins (most likely) will have subcommands; if so, we should show plugins under `Management Command` (Alternatively `Plugin Commands`)

cli/cobra.go
```diff
@@ -159,15 +161,31 @@ func isFirstLevelCommand(cmd *cobra.Command) bool {
...
+func isPluginFirstLevelCommand(cmd *cobra.Command) bool {
+	return isPlugin(cmd) && isFirstLevelCommand(cmd)
+}
+
+func decoratedCommandName(cmd *cobra.Command) string {
+	decoration := " "
+	if isPluginFirstLevelCommand(cmd) {
+		decoration = "*"
+	}
+	return cmd.Name() + decoration + strings.Repeat(" ", cmd.NamePadding()-len(cmd.Name()))
```
ijc: Can you use `.rpad decoratedCOmmandName .NamePadding` in the template in preference to padding here? If so that would be better I think.

ijc:
> Plugins (most likely) will have subcommands; if so, we should show plugins under Management Command (Alternatively Plugin Commands)

Ack, I think that's separate from this change though, #1661 has a separate entry for it:

>  plugins shuld be able to indicate if they are "management" vs not commands in their metadata and be listed accordingly in e.g. docker help.

thaJeztah:
> Ack, I think that's separate from this change though, #1661 has a separate entry for it:

Ah, yes. I started looking why it wouldn't see `.HasSubCommands()` but the plugin "proxy" commands don't know if a plugin command has subcommands (yet)

cli/cobra.go
```diff
@@ -159,15 +161,31 @@ func isFirstLevelCommand(cmd *cobra.Command) bool {
...
+
+func commandVendorAndVersion(cmd *cobra.Command) string {
 	width := 13
-	if v, ok := cmd.Annotations[pluginmanager.CommandAnnotationPluginVendor]; ok {
-		if len(v) > width-2 {
-			v = v[:width-3] + "…"
+	if vendor, ok := cmd.Annotations[pluginmanager.CommandAnnotationPluginVendor]; ok {
+		if len(vendor) > width-2 {
+			vendor = vendor[:width-3] + "…"
+		}
```
ijc: I forgot this earlier -- now that the vendor is at the end of the line I don't think it needs to be trimmed down any more, that was just to keep the columns aligned in the old way of doing things.

ijc:
@ulyssessouza I think a Unit terst for `decoratedCommandName` would increase the score here. Don't know if it would hit the 50% target but nothing else looks to be sanely unit testable to me.

cli/cobra.go
ulyssessouza: Done

cli/cobra.go
```diff
@@ -159,15 +161,27 @@ func isFirstLevelCommand(cmd *cobra.Command) bool {
...
-func commandVendor(cmd *cobra.Command) string {
-	width := 13
-	if v, ok := cmd.Annotations[pluginmanager.CommandAnnotationPluginVendor]; ok {
-		if len(v) > width-2 {
-			v = v[:width-3] + "…"
+func isPluginFirstLevelCommand(cmd *cobra.Command) bool {
+	return isPlugin(cmd) && isFirstLevelCommand(cmd)
+}
+
+func decoratedCommandName(cmd *cobra.Command) string {
```
thaJeztah: bike-shedding, but these are all commands, so perhaps just

```suggestion
func decoratedName(cmd *cobra.Command) string {
```

cli/cobra.go
```diff
@@ -159,15 +161,27 @@ func isFirstLevelCommand(cmd *cobra.Command) bool {
...
+	decoration := " "
+	if isPluginFirstLevelCommand(cmd) {
```
thaJeztah: Was just looking at this again, but I think we're only interested if a command is provided by a plugin (irregardless if it's a subcommand?). If this code is called from a plugin itself, it would not see itself as a plugin, whereas if it's called from the main CLI, if would just add the decoration. If correct, this could be changed to:

```suggestion
	if isPlugin(cmd) {
```

cli/cobra.go
```diff
@@ -159,15 +161,27 @@ func isFirstLevelCommand(cmd *cobra.Command) bool {
...
+		decoration = "*"
+	}
+	return cmd.Name() + decoration + strings.Repeat(" ", cmd.NamePadding()-len(cmd.Name()))
+}
+
+func commandVendorAndVersion(cmd *cobra.Command) string {
```
thaJeztah: Same as my other comment; I think it's implicit already that it's a command, so could be just

```suggestion
func vendorAndVersion(cmd *cobra.Command) string {
```

cli/cobra.go
```diff
@@ -159,15 +161,27 @@ func isFirstLevelCommand(cmd *cobra.Command) bool {
...
+	if vendor, ok := cmd.Annotations[pluginmanager.CommandAnnotationPluginVendor]; ok {
```
thaJeztah: Should we perhaps make this function return nothing if it's not a plugin? Then we won't have to check if it's a plugin in the templates;

```suggestion
	if !isPlugin(cmd) {
		return ""
	}
	if vendor, ok := cmd.Annotations[pluginmanager.CommandAnnotationPluginVendor]; ok {
```

cli/cobra.go
```diff
@@ -230,7 +244,7 @@ Options:
 Management Commands:
 
 {{- range managementSubCommands . }}
-  {{rpad .Name .NamePadding }} {{ if isFirstLevelCommand .}}{{commandVendor .}} {{ end}}{{.Short}}
+  {{decoratedCommandName .}}{{.Short}}{{ if isPluginFirstLevelCommand .}} {{commandVendorAndVersion .}}{{ end}}
```
thaJeztah: With my suggestions above, this could probably be 

```suggestion
  {{decoratedName .}}{{.Short}}{{ if isPlugin .}} {{vendorAndVersion .}}{{ end}}
```

Taking @ijc's suggestion into account;

```suggestion
  {{.rpad decoratedName .NamePadding}}{{.Short}}{{ if isPlugin .}} {{vendorAndVersion .}}{{ end}}
```

Or (if we make `vendorAndVersion` return nothing in the non-plugin case);

```suggestion
  {{.rpad decoratedName .NamePadding}}{{.Short}} {{vendorAndVersion .}}
```

cli/cobra.go
```diff
@@ -230,7 +245,7 @@ Options:
 Management Commands:
 
 {{- range managementSubCommands . }}
-  {{rpad .Name .NamePadding }} {{ if isFirstLevelCommand .}}{{commandVendor .}} {{ end}}{{.Short}}
+  {{rpad (decoratedCommandName .) (add .NamePadding 1)}}{{.Short}}{{ if isPluginFirstLevelCommand .}} {{commandVendorAndVersion .}}{{ end}}
```
thaJeztah: With my suggestions above, this could probably be 

```suggestion
  {{rpad (decoratedCommandName .) (add .NamePadding 1)}}{{.Short}}{{ if isPlugin .}} {{vendorAndVersion .}}{{ end}}
```

Or (if we make `vendorAndVersion` return nothing in the non-plugin case);

```suggestion
  {{rpad (decoratedCommandName .) (add .NamePadding 1)}}{{.Short}} {{vendorAndVersion .}}
```

cli/cobra.go
ulyssessouza: Done... Note that I had to add a function to have the separation (command->description) space

cli/cobra.go
```diff
@@ -239,7 +254,7 @@ Management Commands:
 Commands:
 
 {{- range operationSubCommands . }}
-  {{rpad .Name .NamePadding }} {{ if isFirstLevelCommand .}}{{commandVendor .}} {{ end}}{{.Short}}
+  {{rpad (decoratedCommandName .) (add .NamePadding 1)}}{{.Short}}{{ if isPluginFirstLevelCommand .}} {{commandVendorAndVersion .}}{{ end}}
```
thaJeztah: And this could be the same as what we decide on from my previous comment above

cli/cobra_test.go
```diff
@@ -32,28 +33,30 @@ func TestVisitAll(t *testing.T) {
...
 		vendor   string
+		version  string
 		expected string
 	}{
-		{vendor: "vendor", expected: "(vendor)     "},
-		{vendor: "vendor12345", expected: "(vendor12345)"},
-		{vendor: "vendor123456", expected: "(vendor1234…)"},
-		{vendor: "vendor1234567", expected: "(vendor1234…)"},
+		{vendor: "vendor", expected: "(vendor)"},
+		{vendor: "VeryLongVendorName", expected: "(VeryLongVendorName)"},
```
thaJeztah: Are the `veryLong` .. tests still relevant now that we no longer truncate?

ulyssessouza: No. Removing.

cli/cobra.go
ulyssessouza: Done

cli/cobra.go
ulyssessouza: Done

cli/cobra.go
ulyssessouza: Done

cli/cobra.go
ulyssessouza: Done

cli/cobra.go
ulyssessouza: Done

cli/cobra.go
ulyssessouza: Done

cli/cobra_test.go
```diff
@@ -32,28 +32,29 @@ func TestVisitAll(t *testing.T) {
 	assert.DeepEqual(t, expected, visited)
 }
 
-func TestCommandVendor(t *testing.T) {
+func TestCommandVendorAndVersion(t *testing.T) {
```
thaJeztah: nit (not a show-stopper); 
```suggestion
func TestVendorAndVersion(t *testing.T) {
```

ulyssessouza: Done

PR #1680, opened 2019-02-20
Fix issue where plugin command error exit code is printed out

cquon:
**- What I did**
When a plugin is run, it is ran as an `exec.Cmd.Run()`.  (https://golang.org/pkg/os/exec/#Cmd.Run) This is returning an error type of `*ExitError` (https://golang.org/pkg/os/exec/#ExitError) which string error message is `exit status xxx` which is output on the command line.  This change handles that and the command exits with the same error code as the plugin.

**- How I did it**
Modified the code to check for an ExitError, capture that error code and return it.  If it is unable to, return exit code 1 which will be the fallback error code.

**- How to verify it**
Have a plugin run with the following behavior:
1) Run successfully and verify that no behavior change.
2) Have plugin run and error (with arbitrary exit code) and verify when getting the error code of previously ran command 
```
docker <plugin_cmd> ...
echo $?
```
Verify that the exit code matches what the plugin exited with.

### DIFF
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -4,7 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
+	"syscall"
 
 	"github.com/docker/cli/cli"
 	pluginmanager "github.com/docker/cli/cli-plugins/manager"
@@ -55,7 +57,21 @@ func newDockerCommand(dockerCli *command.DockerCli) *cobra.Command {
 				return err
 			}
 
-			return plugincmd.Run()
+			err = plugincmd.Run()
+			if err != nil {
+				statusCode := 1
+				exitErr, ok := err.(*exec.ExitError)
+				if !ok {
+					return err
+				}
+				if ws, ok := exitErr.Sys().(syscall.WaitStatus); ok {
+					statusCode = ws.ExitStatus()
+				}
+				return cli.StatusError{
+					StatusCode: statusCode,
+				}
+			}
+			return nil
 		},
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			// flags must be the top-level command flags, not cmd.Flags()
--- a/e2e/cli-plugins/run_test.go
+++ b/e2e/cli-plugins/run_test.go
@@ -194,3 +194,16 @@ func TestCliInitialized(t *testing.T) {
 	assert.Assert(t, res.Stdout() != "")
 	assert.Assert(t, is.Equal(res.Stderr(), ""))
 }
+
+// TestPluginErrorCode tests when the plugin return with a given exit status.
+// We want to verify that the exit status does not get output to stdout and also that we return the exit code.
+func TestPluginErrorCode(t *testing.T) {
+	run, _, cleanup := prepare(t)
+	defer cleanup()
+	res := icmd.RunCmd(run("helloworld", "exitstatus2"))
+	res.Assert(t, icmd.Expected{
+		ExitCode: 2,
+		Err:      "Exiting with error status 2",
+	})
+	assert.Assert(t, is.Equal(res.Stdout(), ""))
+}
--- a/cli-plugins/examples/helloworld/main.go
+++ b/cli-plugins/examples/helloworld/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/docker/cli/cli-plugins/manager"
 	"github.com/docker/cli/cli-plugins/plugin"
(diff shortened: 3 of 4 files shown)


cmd/docker/docker.go
```diff
@@ -228,6 +231,34 @@ func initializeDockerCli(dockerCli *command.DockerCli, flags *pflag.FlagSet, opt
...
+				}
+				os.Exit(err.StatusCode)
+			}
+		case *exec.ExitError:
+			{
+				if ws, ok := err.Sys().(syscall.WaitStatus); ok {
+					os.Exit(ws.ExitStatus())
+				}
+				os.Exit(1)
+			}
```
ijc: I think it would be cleaner to have the code in `newDockerCommand.RunE` which calls `plugincmd.Run()` and returns the result to capture the `exec.ExitError` and convert it into a `cli.StatusError`. That way the knowledge that some commands are external does not leak any further that it already has.

cmd/docker/docker_test.go
```diff
@@ -55,3 +58,51 @@ func TestVersion(t *testing.T) {
...
+		RunE: func(_ *cobra.Command, _ []string) error {
+			os.Exit(2)
+			return nil
+		},
+	}
+
+	if os.Getenv("CHECK_RETURN_CODE") == "1" {
+		executeCmd(cli, cmd)
+		return
+	}
```
ijc: This reexecing with an envvar is a bit convoluted, I'd suggest to make it an e2e test instead. Putting it in `e2e/cli-plugins/run_test.go` would result it in being alongside a bunch of similar tests.

To enable that you could extend `cli-plugins/examples/helloworld/` to have a subcommand which exits with an error, or add a new plugin for e2e testing under `e2e/cli-plugins/plugins` alongside the `badmeta` one.

cquon: Perfect I was looking for something like this I didn't know it was in place

cmd/docker/docker.go
cquon: Agreed

cli/compose/convert/service.go
```diff
@@ -158,6 +158,7 @@ func Service(
 			Placement: &swarm.Placement{
 				Constraints: service.Deploy.Placement.Constraints,
 				Preferences: getPlacementPreference(service.Deploy.Placement.Preferences),
+				MaxReplicas: service.Deploy.Placement.MaxReplicas,
```
ijc: This and a bunch of other changes to `cli/compose/*` seem like they are from some other work?

cmd/docker/docker.go
```diff
@@ -55,7 +58,22 @@ func newDockerCommand(dockerCli *command.DockerCli) *cobra.Command {
...
+				exitErr, ok := err.(*exec.ExitError)
+				if !ok {
+					return err
+				}
+				if ws, ok := exitErr.Sys().(syscall.WaitStatus); ok {
+					statusCode = ws.ExitStatus()
+				}
+				return cli.StatusError{
+					StatusCode: statusCode,
+					Status: fmt.Sprintf("Running plugin command `%s` errored.", strings.Join(args, " ")),
```
ijc: Not sure if we shouldn't leave this blank instead. We could/should assume that the plugin printed some sort of error, I don't think a builtin command would print anything similarly generic on fail, would it?

cmd/docker/docker_test.go
```diff
@@ -54,4 +54,4 @@ func TestVersion(t *testing.T) {
 	err = cmd.Execute()
 	assert.NilError(t, err)
 	assert.Check(t, is.Contains(b.String(), "Docker version"))
-}
+}
```
ijc: nit: spurious change?

cli/compose/convert/service.go
cquon: Sorry I need to rebase my branch

cmd/docker/docker.go
cquon: Agreed

cmd/docker/docker.go
```diff
@@ -6,6 +6,9 @@ import (
 	"os"
 	"strings"
 
+	"os/exec"
```
silvin-lubecki: nit: can you group the standards imports together?

e2e/cli-plugins/run_test.go
```diff
@@ -194,3 +194,15 @@ func TestCliInitialized(t *testing.T) {
 	assert.Assert(t, res.Stdout() != "")
 	assert.Assert(t, is.Equal(res.Stderr(), ""))
 }
+
+// TestPluginErrorCode tests when the plugin return with a given exit status.
+// We want to verify that the exit status does not get output to stdout and also that we return the exit code.
```
ijc: Given that first bit you might want to add
```go
assert.Assert(t, is.Equal(res.Stdout(), ""))
```
since `icmd.Expected` doesn't check anything when `Out == ""`?

ijc:
FWIW I don't believe this change is at all unit testable, so the `codecov/patch` check can/should be ignored.

thaJeztah:
Looks like there's a failure in the e2e test;

```
18:06:19 === Errors
18:06:19 e2e/cli-plugins/run_test.go:201:15: assignment mismatch: 2 variables but 3 values
18:06:19 
```

cquon:
@thaJeztah addressed the test issue

PR #1688, opened 2019-02-25
use char to check blockIO type

luoyunpeng:
**- What I did**
when check the block IO type(read or write), (https://github.com/docker/cli/blob/master/cli/command/container/stats_helpers.go#L205), This method will switch the type string to lower using `strings.ToLower` , then check the type, This change will use char to check the type, no need to switch first, so the performance is better, and 0 alloc on heap.
**
BenchmarkChar-4 2000000000	0.79 ns/op	0 B/op	0 allocs/op
BenchmarkStringsToLower-4 20000000	66.8 ns/op	8 B/op	2 allocs/op
**
**- How I did it**
check the first char in type string, to see if it is 'r' ,"R" or 'w','W'
**- How to verify it**

### DIFF
--- a/cli/command/container/stats_helpers.go
+++ b/cli/command/container/stats_helpers.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"io"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


luoyunpeng:
Hi @silvin-lubecki , Yes, I found the unit test(https://github.com/docker/cli/blob/master/cli/command/container/stats_unit_test.go#L9)

silvin-lubecki:
Thank you @luoyunpeng ! By the way can you squash your 2 last commits?

thaJeztah:
looks like there's a merge-commit in there indeed; @luoyunpeng if you can rebase and squash the commits in this PR, the merge commit should disappear; you can find some instructions here; https://docs.docker.com/v17.06/opensource/workflow/work-issue/#pull-and-rebase-frequently

but feel free to ask for help if needed 👍

luoyunpeng:
sorry,  I did wrong way to to squash

luoyunpeng:
Thanks for checking @thaJeztah, please help review again,

luoyunpeng:
i fixed the sign, @thaJeztah, but the last one check occured error, seems like not caused by this pr change

PR #1690, opened 2019-02-25
Always initialize context store

jcsirot:
**- What I did**
Always initialize context store even when the CLI has already a client. It happens when the CLI is initialized from a plugin. 

image

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -209,12 +209,13 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...Initialize
 
 	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
 
+	cli.contextStore = store.New(cliconfig.ContextStoreDir(), cli.contextStoreConfig)
+	cli.currentContext, err = resolveContextName(opts.Common, cli.configFile, cli.contextStore)
+	if err != nil {
+		return err
+	}
+
 	if cli.client == nil {
-		cli.contextStore = store.New(cliconfig.ContextStoreDir(), cli.contextStoreConfig)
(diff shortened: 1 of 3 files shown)


cli/command/cli.go
```diff
@@ -209,12 +209,13 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...Initialize
...
+		return err
+	}
+
 	if cli.client == nil {
-		cli.contextStore = store.New(cliconfig.ContextStoreDir(), cli.contextStoreConfig)
-		cli.currentContext, err = resolveContextName(opts.Common, cli.configFile, cli.contextStore)
-		if err != nil {
-			return err
-		}
 		endpoint, err := resolveDockerEndpoint(cli.contextStore, cli.currentContext, opts.Common)
```
ijc: This is, below the context here, used as `cli.dockerEndpoint = endpoint`. Should that be pulled out too?

silvin-lubecki: I wasn't sure but yep you are right, let's move it up too 👍

simonferquel:
Please rebase this one :)

silvin-lubecki:
@thaJeztah PTAL

cli/command/cli.go
```diff
@@ -209,12 +209,13 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...Initialize
 
 	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
 
+	cli.contextStore = store.New(cliconfig.ContextStoreDir(), cli.contextStoreConfig)
```
thaJeztah: Not for this PR, but wondering if we should consider having a `WithContextStore()` ops

PR #1692, opened 2019-02-26
Remove some bashisms

thaJeztah:
ping @tianon @albers PTAL 🤗

PR #1693, opened 2019-02-26
Upgrade shellcheck 0.6.0

thaJeztah:
Follow-up to https://github.com/docker/cli/pull/1538 (decided to keep it separate in case we needed to discuss the new exclusions)

As a bonus; changing to the official, alpine-based image makes this image substantially smaller;

current master:

```
docker-cli-shell-validate            latest     91.2MB
```

with official shellcheck v0.4.6 image

```
docker-cli-shell-validate            latest     22.4MB
```

with official shellcheck v0.6.0 image

```
docker-cli-shell-validate            latest     12MB
```

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -1,10 +1,12 @@
 #!/usr/bin/env bash
-# shellcheck disable=SC2016,SC2119,SC2155
+# shellcheck disable=SC2016,SC2119,SC2155,SC2206,SC2207
 #
 # Shellcheck ignore list:
 #  - SC2016: Expressions don't expand in single quotes, use double quotes for that.
 #  - SC2119: Use foo "$@" if function's $1 should mean script's $1.
 #  - SC2155: Declare and assign separately to avoid masking return values.
+#  - SC2206: Quote to prevent word splitting, or split robustly with mapfile or read -a.
+#  - SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting).
 #
 # You can find more details for each warning at the following page:
 #    https://github.com/koalaman/shellcheck/wiki/<SCXXXX>
--- a/dockerfiles/Dockerfile.shellcheck
+++ b/dockerfiles/Dockerfile.shellcheck
@@ -1,10 +1,7 @@
-FROM    debian:stretch-slim
-
-RUN     apt-get update && \
-        apt-get -y install make shellcheck && \
-        apt-get clean
-
+FROM    koalaman/shellcheck-alpine:v0.4.6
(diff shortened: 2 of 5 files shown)


thaJeztah:
ping @albers @tianon PTAL

contrib/completion/bash/docker
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env bash
-# shellcheck disable=SC2016,SC2119,SC2155
+# shellcheck disable=SC2016,SC2119,SC2155,SC2206,SC2207
 #
 # Shellcheck ignore list:
 #  - SC2016: Expressions don't expand in single quotes, use double quotes for that.
 #  - SC2119: Use foo "$@" if function's $1 should mean script's $1.
 #  - SC2155: Declare and assign separately to avoid masking return values.
+#  - SC2206: Quote to prevent word splitting, or split robustly with mapfile or read -a.
+#  - SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting).
```
thaJeztah: Disabled these for this file to ignore errors, such as;

```
In contrib/completion/bash/docker line 314:
	COMPREPLY=( $(compgen -W "$(__docker_volumes "$@")" -- "$current") )
                    ^-- SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting).

In contrib/completion/bash/docker line 347:
		plugins=(${plugins[@]/$del/})
                         ^-- SC2206: Quote to prevent word splitting, or split robustly with mapfile or read -a.
```

scripts/validate/shellcheck
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 set -eo pipefail
```
thaJeztah: Looks like we may need bash after all if we want to keep the `set -eo pipefail`, see https://github.com/docker/cli/pull/1692

```
/go/src/github.com/docker/cli/scripts/build/windows: 6: set: Illegal option -o pipefail
/go/src/github.com/docker/cli/scripts/build/osx: 6: set: Illegal option -o pipefail
```

(although it seems to work fine for some reason, so slightly confused)

thaJeztah: Looks like busybox sh supports it (but `dash` will fail)

```
dash: 1: set: Illegal option -o pipefail
```

thaJeztah:
ping @albers PTAL 🤗

contrib/completion/bash/docker
albers: I verified: Yes, these warnings really don't make sense here.

PR #1698, opened 2019-02-28
Fix unknown $(MOUNTS) variable in makefile plugins target

silvin-lubecki:
**- What I did**
Mounts were broken for plugins as #1511 replaced `$(MOUNTS)` by `$(DOCKER_CLI_MOUNTS)`.

image

### DIFF
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -63,7 +63,7 @@ binary: build_binary_native_image ## build the CLI
 build: binary ## alias for binary
 
 plugins: build_binary_native_image ## build the CLI plugin examples
-	docker run --rm $(ENVVARS) $(MOUNTS) $(BINARY_NATIVE_IMAGE_NAME) ./scripts/build/plugins
+	docker run --rm $(ENVVARS) $(DOCKER_CLI_MOUNTS) $(BINARY_NATIVE_IMAGE_NAME) ./scripts/build/plugins
 
 .PHONY: clean
 clean: build_docker_image ## clean build artifacts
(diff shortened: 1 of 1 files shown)


docker.Makefile
```diff
@@ -63,7 +63,7 @@ binary: build_binary_native_image ## build the CLI
 build: binary ## alias for binary
 
 plugins: build_binary_native_image ## build the CLI plugin examples
-	docker run --rm $(ENVVARS) $(MOUNTS) $(BINARY_NATIVE_IMAGE_NAME) ./scripts/build/plugins
+	docker run --rm $(ENVVARS) $(DOCKER_CLI_MOUNTS) $(BINARY_NATIVE_IMAGE_NAME) ./scripts/build/plugins
```
thaJeztah: could we do the reverse, and remove the `DOCKER_CLI_` prefix from all of these? This is the makefile for the CLI repository, so not sure why we need the prefix

thaJeztah: I also noticed the `$DOCKER_CLI_SHELL` variable was added; alpine images now have `/bin/sh` set by default (previously they didn't, which was perhaps the reason this argument was added), but we install `bash` as well @wk8 was that variable added to switch to `bash`? If so, perhaps we should update the Dockerfile, and make it the default?

wk8:
Ooops. Sorry. In my defense the `plugin*` targets were added some time after I made https://github.com/docker/cli/pull/1511, and that wasn't caught as no conflicts. Apologies!

docker.Makefile
wk8: @thaJeztah the `DOCKER_CLI` prefix is for the variables which can be overriden with env variables. Allows setting them in eg a shell profile (and that's also why they are nicely set apart - https://github.com/wk8/cli/blob/2178fea84dbea0b278d412b5f9cf16e3542d2650/docker.Makefile#L7-L11)

thaJeztah: shouldn't those be set _after_ `make` though? 

```
make FOO=bar BAR=baz <target>
```

wk8: That's one way to do it but
```
export FOO=bar
make <target>
```

or
```
FOO=bar make <target>
```

work just the same.

wk8: As to `$DOCKER_CLI_SHELL`, yup I've set mine to `bash --login` so that it also loads my shell profile - just made it `ash` in https://github.com/docker/cli/pull/1511 to make the change seamless, but I don't any reason why we couldn't switch to bash.

thaJeztah: I'd be +1 to change it to bash by default, but we could do that separate

PR #1701, opened 2019-03-01
Fix cli-plugins build.

ijc:
Seems I rebased over b039db985a4b ("Make it possible to override the volume
mounts and shell for the dev container") at some point and failed to notice
that some of the variable names had changed.

### DIFF
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -63,7 +63,7 @@ binary: build_binary_native_image ## build the CLI
 build: binary ## alias for binary
 
 plugins: build_binary_native_image ## build the CLI plugin examples
-	docker run --rm $(ENVVARS) $(DOCKER_CLI_MOUNTS) $(BINARY_NATIVE_IMAGE_NAME) ./scripts/build/plugins
+	$(DOCKER_RUN) $(BINARY_NATIVE_IMAGE_NAME) ./scripts/build/plugins
 
 .PHONY: clean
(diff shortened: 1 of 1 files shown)


ijc:
I had a look to see if I should be tweaking the CI but it seems none of these targets (build, cross etc) are tested there -- the test-e2e which is run builds things internally to its own image and all the circle stuff is open coded in `circle.yml` rather than using the makefile targets. I won't look at this further here.

ijc:
Looks like @silvin-lubecki also fixed in 2c6b2ccbddae4a33fc2471e3e00263a36c51ba4e (#1698).

I rebased the switch to use `$(DOCKER_RUN)` and rewrote the commit message:
```
    Use `$(DOCKER_RUN)` for cli plugins build.
    
    Seems I rebased over b039db985a4b ("Make it possible to override the volume
    mounts and shell for the dev container") at some point and failed to notice
    that some of the variable names had changed.
    
    In the meantime the underlying issue was fixed in #1698 but here we switch to
    using `$(DOCKER_RUN)`. This means that these rules now use
    `$(DOCKER_RUN_NAME_OPTION)` and thus obey the `$(DOCKER_CLI_CONTAINER_NAME)`
    variable.
    
```

Not sure if this is still valuable or not. Feel free to close or merge as you see fit.

thaJeztah:
ping @silvin-lubecki @vdemeester @tiborvass PTAL

PR #1704, opened 2019-03-01
Allow flags to be passed to e2e tests

ijc:
This allows e.g.

    $ make -f docker.Makefile fmt test-e2e-non-experimental E2E_TESTFLAGS="-test.run TestRunGoodArgument|TestRunGood"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ clean: ## remove build artifacts
 
 .PHONY: test-unit
 test-unit: ## run unit tests, to change the output format use: GOTESTSUM_FORMAT=(dots|short|standard-quiet|short-verbose|standard-verbose) make test-unit 
-	gotestsum -- $(shell go list ./... | grep -vE '/vendor/|/e2e/')
+	gotestsum $(TESTFLAGS) -- $(shell go list ./... | grep -vE '/vendor/|/e2e/')
 
 .PHONY: test
 test: test-unit ## run tests
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ clean: ## remove build artifacts
 
 .PHONY: test-unit
 test-unit: ## run unit tests, to change the output format use: GOTESTSUM_FORMAT=(dots|short|standard-quiet|short-verbose|standard-verbose) make test-unit 
-	gotestsum $(TESTFLAGS) -- $(shell go list ./... | grep -vE '/vendor/|/e2e/')
+	gotestsum $(TESTFLAGS) -- $${TESTDIRS:-$(shell go list ./... | grep -vE '/vendor/|/e2e/')}
 
 .PHONY: test
 test: test-unit ## run tests
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -21,7 +21,7 @@ ifeq ($(DOCKER_CLI_GO_BUILD_CACHE),y)
 DOCKER_CLI_MOUNTS += -v "$(CACHE_VOLUME_NAME):/root/.cache/go-build"
 endif
 VERSION = $(shell cat VERSION)
-ENVVARS = -e VERSION=$(VERSION) -e GITCOMMIT -e PLATFORM -e TESTFLAGS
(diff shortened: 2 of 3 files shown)


docker.Makefile
```diff
@@ -141,15 +141,15 @@ test-e2e: test-e2e-non-experimental test-e2e-experimental test-e2e-connhelper-ss
 
 .PHONY: test-e2e-experimental
 test-e2e-experimental: build_e2e_image # run experimental e2e tests
-	docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -e DOCKERD_EXPERIMENTAL=1 $(E2E_IMAGE_NAME)
+	docker run --rm -e TESTFLAGS="$(E2E_TESTFLAGS)" -v /var/run/docker.sock:/var/run/docker.sock -e DOCKERD_EXPERIMENTAL=1 $(E2E_IMAGE_NAME)
```
thaJeztah: Should we make the e2e image itself use `E2E_TESTFLAGS` instead of `TESTFLAGS` ? Then we could unconditionally do

```
export E2E_TESTFLAGS

and change the `docker run` for all targets to include

docker run \
-e TESTFLAGS \
-e E2E_TESTFLAGS \
```

(as part of https://github.com/docker/cli/pull/1701 ?)

ijc: I can do that if you like, why both `-e TESTFLAGS and -e E2E_TESTFLAGS` though, are you suggesting they be incremental? If we do this shall I add "UNIT_TESTFLAGS" too?

Please confirm what you want me to do.

Lets not conflate it with #1701 though.

thaJeztah: If they're not expected to be incremental, it could be just a single env-var.

ijc: I can't tell from the above what you would like me to do.

thaJeztah: Oh, sorry, I was thinking why the `E2E_TESTFLAGS` variable was needed, if it could look like;

```
docker run -e TESTFLAGS
```

Which could be added to https://github.com/docker/cli/blob/2c6b2ccbddae4a33fc2471e3e00263a36c51ba4e/docker.Makefile#L24

(hope my thinking is clear, otherwise hook up with me on Slack 😅 )

ijc: I called it `E2E_*` because it only applies to the `test-e2e-*` targets and not the `test-unit` target, that was deliberate because just using `TESTFLAGS` everywhere wouldn't let you pass different args to unit vs e2e (e.g. `-test.run` doesn't make sense to set globally).

thaJeztah: Right, so I was thinking you'd only set those flags if (e.g.) you want to run a single integration test, so likely not when combining unit-tests and integration tests, and was looking for a bit of consistency with [the Moby Makefile](https://github.com/moby/moby/blob/master/Makefile#L20-L68)

ijc:
Rebased and switched to `TESTFLAGS` as @thaJeztah suggested. Plus I ensured it was wired up to unit tests too.

I also added `TESTDIR` (following the [moby/moby](https://github.com/moby/moby/blob/master/Makefile#L20-L68) precedent) to control the dirs to run tests on for both types of test.

Makefile
```diff
@@ -12,7 +12,7 @@ clean: ## remove build artifacts
 
 .PHONY: test-unit
 test-unit: ## run unit tests, to change the output format use: GOTESTSUM_FORMAT=(dots|short|standard-quiet|short-verbose|standard-verbose) make test-unit 
-	gotestsum -- $(shell go list ./... | grep -vE '/vendor/|/e2e/')
+	gotestsum $(TESTFLAGS) -- $${TESTDIRS:-$(shell go list ./... | grep -vE '/vendor/|/e2e/')}
```
thaJeztah: Should this be

```suggestion
	gotestsum $(TESTFLAGS) -- ${$TESTDIRS:-$(shell go list ./... | grep -vE '/vendor/|/e2e/')}
```

(perhaps not; but the nesting is always confusing me :joy:)

ijc: I don't think so, the `$$` is escaping a single `$` because this is a `Makefile`.

thaJeztah: Ah 😊 👍

PR #1705, opened 2019-03-02
Fix lack of completion after docker exec -it

domq:
I realize that in theory we should have a huge list of flag combinations here, but  (in my own experience) -it seems the most common.

Signed-off-by: Dominique Quatravaux <dominique@quatravaux.org>


vdemeester:
@domq thanks for contributing :hugs: but I do agree with @thaJeztah, I don't think we should go that road, people won't really gain anything from `-it` completion anyway I'm guessing (2 keystrokes for 2 characters)… I think completion is really useful for long flags

thaJeztah:
I'll go ahead and close this one, but thanks again for contributing, and feel free to continue the conversation 🤗

Issue #1376, opened 2018-09-14
docker ssh with OpenSSH-Win32 does not work with Powershell and CMD login shell


nunix:
**Description**
When trying to connect with `docker -H ssh://<host>` to a Windows host running OpenSSH-Win32, if the login shell is PowerShell or CMD, the client cannot connect to the daemon.
However, if the login shell is bash (WSL), then the client can connect to the daemon.

Complete examples with the 2 scenarios
![image](https://user-images.githubusercontent.com/905874/45569539-13998700-b860-11e8-9496-4c534e1a2775.png)

**Steps to reproduce the issue:**
1. Set registry key `HKLM:\SOFTWARE\OpenSSH\DefaultShell` to `C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe`
![image](https://user-images.githubusercontent.com/905874/45570395-dda9d200-b862-11e8-87bf-a26550f458d4.png)

2. Try to connect > we get an error
```
docker -H ssh://localhost version
nunix@localhost's password:
nunix@localhost's password:
Client: Docker Engine - Community
 Version:           18.09.0-ce-beta1
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        78a6bdb
 Built:             Thu Sep  6 22:42:13 2018
 OS/Arch:           windows/amd64
 Experimental:      false
error during connect: Get http://docker/v1.39/version: command [ssh localhost -- docker system dial-stdio] has exited with exit status 1, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr="the raw stream connection does not implement halfCloser\n"
```
![image](https://user-images.githubusercontent.com/905874/45570617-a4be2d00-b863-11e8-8928-5c06298be2e9.png)

3. Set registry key `HKLM:\SOFTWARE\OpenSSH\DefaultShell` to `C:\WINDOWS\System32\bash.exe`
![image](https://user-images.githubusercontent.com/905874/45570954-f87d4600-b864-11e8-9e54-de566bf3b1a4.png)

4. [Optional] if not running, start the `npiperelay.exe` from the WSL shell
```
sudo docker-relay &
```

5. Try to connect > we succeed
```
docker -H ssh://localhost version
nunix@localhost's password:
Client: Docker Engine - Community
 Version:           18.09.0-ce-beta1
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        78a6bdb
 Built:             Thu Sep  6 22:42:13 2018
[...]

PR #1710, opened 2019-03-04
Disable `docker system dial-stdio` on Windows

ijc:
The `conn` here is `*winio.win32MessageBytePipe` which does not have a
`CloseRead` method (it does have `CloseWrite`) resulting in:

    docker@WIN-NUC0 C:\Users\docker>.\docker-windows-amd64.exe system dial-stdio

Also disable the path which uses this for cli-plugins on Windows.

/cc @tonistiigi @AkihiroSuda if Windows can't be fixes soon perhaps we should do this and revert when Windows works?

### DIFF
--- a/cli-plugins/plugin/plugin.go
+++ b/cli-plugins/plugin/plugin.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"runtime"
 	"sync"
 
 	"github.com/docker/cli/cli"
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
According to https://github.com/docker/cli/issues/1376 it works on Windows with certain configuration?

AkihiroSuda:
I think `CloseRead` can be added as NOP

ijc:
> According to #1376 it works on Windows with certain configuration?

Ah, I guess there it is a proper tcp socket (which I would guess does indeed support half close in both directions since it is tcp) since it is doing `ssh://`.

For me (using dial-stdio in a plugin to connect back to the local dameon) it is (I presume) the named pipe, which I suppose doesn't.

ijc:
@thaJeztah I think this is only tangentially related to plugins (cf the label)

ijc:
#1718 is my attempt to fix this on Windows, PTAL (my testing on Windows has been rather limited)

silvin-lubecki:
Merged, to be reverted when #1718 will land.

PR #1714, opened 2019-03-05
container: --gpus support

### LINKED ISSUE #1200: [RFC] GPU support in CLI
The goal of this RFC is to discuss how we can expose the GPU support that was recently added to containerd by @crosbymichael in this PR: https://github.com/containerd/containerd/pull/2330

Since there are multiple ways of implementing this, I think it makes sense to start with the user experience, i.e. the CLI. If we can agree on the format, the next step will be that I write a PR for API changes in the `github.com/docker/docker/api/types` package. My rationale being that if we decide to not allow some configuration options in the CLI, it would be useless to have `api/types` expose it. As a concrete example, if we don't allow passing options on a per-GPU basis, it shouldn't be part of the API.

After discussing with @crosbymichael, he advised to limit the support to GPUs. We are excluding other types of specialized hardware such as FPGAs, InfiniBand, ASICs... from this RFC. 

This RFC introduces some vocabulary, then presents the approach I find is the most compelling.

## GPU Vendors
* While the containerd PR is only for NVIDIA GPUs, the CLI should be able to support multiple vendors.
* For simplicity, we should enforce the GPU vendor to be a lowercase string with no special characters.
e.g. `nvidia`, `intel` or `amd`.
* Reverse-DNS notation (e.g. `com.nvidia`) is too verbose and should probably be avoided for a CLI.
* The list of support GPU vendors **could** be hardcoded in the CLI and checked, since it's unlikely to change quickly, but it is not a requirement.

## GPU identifiers
* GPU support should not be a boolean flag like `--with-gpus`, users should be able to select a subset of all GPUs, similarly to `--cpuset-cpus`: ```nvidia=0,1,2,3```
* We should not assume a numerical identifier for GPUs (unlike `--cpuset-cpus`), GPUs can have a string representation, for instance a UUID: ```nvidia=GPU-fef8089b-4820-abfc-e83e-94318197576e```
* The CLI can only verify the basic format of GPU identifiers, it can't detect if they are ultimately valid (e.g. if the GPU exists or not), this will be done later and will be vendor-specific.

## GPU options
* In addition to selecting GPUs, the user should be to add options to modify the runtime behavior. For the containerd PR it would mean having a way to modify this struct from the CLI: https://github.com/containerd/containerd/blob/master/contrib/nvidia/nvidia.go#L88-L96
* The baseline would be to have options that apply to all GPUs/to the system: ```nvidia.capabilities = compute```.
* An extension would be to have options per-GPU, for instance to enforce that GPU 0 can only use 2 GB of memory, whereas GPU 1 can use 6 GB: ```limit_mem[0] = 2GB limit_mem[1] = 6GB```.
* Same as for identifiers, the CLI can't perform actual validation of the options passed by the user, it's vendor-specific.

## Suggested approach
Add two CLI options: 
* `--gpus` for specifying a list of GPUs (like `--cpuset-cpus`)
* `--gpus-opt` for specifying a list of GPU options (like `--storage-opt`)
```
docker run --gpus nvidia=0,1,2,3 --gpus nvidia=GPU-abcd --gpus-opt nvidia.capabilities=compute --gpus-opt nvidia.kmods=true
```
The important element being the vendor name, which serves as a key for specifying options and identifiers.
I think this is easy to read since you don't have to pack everything in a single argument.

If we want per-GPU options, we can introduce some sort of indexing in `--gpus-opt`:
```
--gpus-opt nvidia.0.limit_mem=2GB
--gpus-opt nvidia[0].limit_mem=2GB
```

## Discarded approaches
I think the approach above is the best, but here are some thoughts on other possibilities.

### Single CSV argument
Similar to `--mount`, add a new argument `--gpus` which expects a CSV value.
```
docker run --gpus vendor=nvidia,options=capabilities=compute,options=kmods=true,0,1,2,3,GPU-abcd
```
This is explicit and everything is contained in one option, but it's also bizarre since we have nested equal signs, and we have to expand the list of GPU identifiers as multiple arguments, to avoid needing complicated quoting from the shell.

A follow-up question would then be: can you specify multiple times the same vendor? If yes, what if the listed options don't match?

### Similar to `--device`
Add a new argument `--gpus` which expects 2 or 3 strings separated by `:`
```
docker run --gpus nvidia:0,1,2,3,GPU-abcd:capabilities=compute,kmods=true
```
Less problematic than approach above, but hard to read.

### NVIDIA specific
Same than the suggested approach, but without being generic:
```
docker run --nvidia-gpus 0,1,2,3 --nvidia-gpus GPU-abcd --nvidia-opt capabilities=compute --nvidia-opt kmods=true
```
It removes the need to use the vendor name as the key, and is fine for now, but is likely to be removed in the future if we add more vendors. Also, this doesn't match the approach used by the containerd PR.

tiborvass:

Closes #1200

Dependent on https://github.com/moby/moby/pull/38828

Most common usecase: `docker run --gpus all`
Otherwise can be further specified: `docker run --gpus 2,driver=nvidia,capabilities=compute`

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -46,6 +46,7 @@ type containerOptions struct {
 	labels             opts.ListOpts
 	deviceCgroupRules  opts.ListOpts
 	devices            opts.ListOpts
+	gpus               opts.GpuOpts
 	ulimits            *opts.UlimitOpt
 	sysctls            *opts.MapOpts
 	publish            opts.ListOpts
@@ -166,6 +167,8 @@ func addFlags(flags *pflag.FlagSet) *containerOptions {
 	flags.VarP(&copts.attach, "attach", "a", "Attach to STDIN, STDOUT or STDERR")
 	flags.Var(&copts.deviceCgroupRules, "device-cgroup-rule", "Add a rule to the cgroup allowed devices list")
 	flags.Var(&copts.devices, "device", "Add a host device to the container")
+	flags.Var(&copts.gpus, "gpus", "GPU devices to add to the container ('all' to pass all GPUs)")
+	flags.SetAnnotation("gpus", "version", []string{"1.40"})
 	flags.VarP(&copts.env, "env", "e", "Set environment variables")
 	flags.Var(&copts.envFile, "env-file", "Read in a file of environment variables")
 	flags.StringVar(&copts.entrypoint, "entrypoint", "", "Overwrite the default ENTRYPOINT of the image")
@@ -557,6 +560,7 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		Ulimits:              copts.ulimits.GetList(),
 		DeviceCgroupRules:    copts.deviceCgroupRules.GetAll(),
 		Devices:              deviceMappings,
+		DeviceRequests:       copts.gpus.Value(),
 	}
 
 	config := &container.Config{
--- a/opts/gpus.go
+++ b/opts/gpus.go
@@ -1,0 +1,112 @@
+package opts
+
+import (
+	"encoding/csv"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/docker/docker/api/types/container"
+	"github.com/pkg/errors"
+)
+
+// GpuOpts is a Value type for parsing mounts
+type GpuOpts struct {
+	values []container.DeviceRequest
+}
+
+func parseCount(s string) (int, error) {
+	if s == "all" {
+		return -1, nil
+	}
+	i, err := strconv.Atoi(s)
+	return i, errors.Wrap(err, "count must be an integer")
+}
+
+// Set a new mount value
+// nolint: gocyclo
+func (o *GpuOpts) Set(value string) error {
+	csvReader := csv.NewReader(strings.NewReader(value))
+	fields, err := csvReader.Read()
+	if err != nil {
+		return err
+	}
+
+	req := container.DeviceRequest{}
+
+	seen := map[string]struct{}{}
+	// Set writable as the default
+	for _, field := range fields {
+		parts := strings.SplitN(field, "=", 2)
+		key := parts[0]
+		if _, ok := seen[key]; ok {
+			return fmt.Errorf("gpu request key '%s' can be specified only once", key)
+		}
+		seen[key] = struct{}{}
+
+		if len(parts) == 1 {
+			seen["count"] = struct{}{}
+			req.Count, err = parseCount(key)
+			if err != nil {
+				return err
+			}
+			continue
+		}
+
+		value := parts[1]
+		switch key {
+		case "driver":
+			req.Driver = value
+		case "count":
+			req.Count, err = parseCount(value)
+			if err != nil {
+				return err
+			}
+		case "device":
+			req.DeviceIDs = strings.Split(value, ",")
+		case "capabilities":
+			req.Capabilities = [][]string{append(strings.Split(value, ","), "gpu")}
+		case "options":
+			r := csv.NewReader(strings.NewReader(value))
+			optFields, err := r.Read()
+			if err != nil {
+				return errors.Wrap(err, "failed to read gpu options")
+			}
+			req.Options = ConvertKVStringsToMap(optFields)
+		default:
+			return fmt.Errorf("unexpected key '%s' in '%s'", key, field)
+		}
+	}
+
+	if _, ok := seen["count"]; !ok && req.DeviceIDs == nil {
+		req.Count = 1
+	}
+	if req.Options == nil {
+		req.Options = make(map[string]string)
+	}
+	if req.Capabilities == nil {
+		req.Capabilities = [][]string{{"gpu"}}
+	}
+
+	o.values = append(o.values, req)
+	return nil
+}
+
+// Type returns the type of this option
+func (o *GpuOpts) Type() string {
+	return "gpu-request"
+}
+
+// String returns a string repr of this option
+func (o *GpuOpts) String() string {
+	gpus := []string{}
+	for _, gpu := range o.values {
+		gpus = append(gpus, fmt.Sprintf("%v", gpu))
+	}
+	return strings.Join(gpus, ", ")
+}
+
+// Value returns the mounts
+func (o *GpuOpts) Value() []container.DeviceRequest {
+	return o.values
+}
--- a/opts/gpus_test.go
+++ b/opts/gpus_test.go
@@ -1,0 +1,48 @@
+package opts
+
+import (
+	"testing"
+
+	"github.com/docker/docker/api/types/container"
+	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
+)
+
+func TestGpusOptAll(t *testing.T) {
+	for _, testcase := range []string{
+		"all",
+		"-1",
+		"count=all",
+		"count=-1",
+	} {
+		var gpus GpuOpts
+		gpus.Set(testcase)
+		gpuReqs := gpus.Value()
+		assert.Assert(t, is.Len(gpuReqs, 1))
+		assert.Check(t, is.DeepEqual(gpuReqs[0], container.DeviceRequest{
+			Count:        -1,
+			Capabilities: [][]string{{"gpu"}},
+			Options:      map[string]string{},
+		}))
+	}
+}
+
+func TestGpusOpts(t *testing.T) {
+	for _, testcase := range []string{
+		"driver=nvidia,\"capabilities=compute,utility\",\"options=foo=bar,baz=qux\"",
+		"1,driver=nvidia,\"capabilities=compute,utility\",\"options=foo=bar,baz=qux\"",
+		"count=1,driver=nvidia,\"capabilities=compute,utility\",\"options=foo=bar,baz=qux\"",
+		"driver=nvidia,\"capabilities=compute,utility\",\"options=foo=bar,baz=qux\",count=1",
+	} {
+		var gpus GpuOpts
+		gpus.Set(testcase)
+		gpuReqs := gpus.Value()
+		assert.Assert(t, is.Len(gpuReqs, 1))
+		assert.Check(t, is.DeepEqual(gpuReqs[0], container.DeviceRequest{
+			Driver:       "nvidia",
+			Count:        1,
+			Capabilities: [][]string{{"compute", "utility", "gpu"}},
+			Options:      map[string]string{"foo": "bar", "baz": "qux"},
+		}))
+	}
+}


opts/gpus.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// Type returns the type of this option
+func (o *GpuOpts) Type() string {
+	return "gpuRequest"
+}
+
+// String returns a string repr of this option
+func (o *GpuOpts) String() string {
+	mounts := []string{}
```
tiborvass: This entire function needs to be updated.

opts/gpus.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		value := parts[1]
+		switch key {
+		case "driver":
+			req.Driver = value
+		case "count":
+			req.Count = value
+		case "device":
+			req.DeviceIDs = strings.Split(value, ",")
```
tiborvass: @RenaudWasTaken for device IDs, I don't want to use `id` because I want to be consistent with other future plans (binding `id` to a "gpu request" instead, providing a more secure API which is needed for swarm and buildkit), so I went with `device`.

RenaudWasTaken: I'm ok with this :)

opts/gpus.go
```diff
@@ -0,0 +1,112 @@
...
+		case "driver":
+			req.Driver = value
+		case "count":
+			req.Count, err = parseCount(value)
+			if err != nil {
+				return err
+			}
+		case "device":
+			req.DeviceIDs = strings.Split(value, ",")
+		case "caps":
```
tiborvass: Maybe this should be `capabilities` spelled out.

tiborvass:
Ping @thaJeztah @vdemeester I revendored and it's ready for review.

opts/gpus.go
thaJeztah: yes, `capabilities` may be clearer 🤔

cli/command/container/opts.go
```diff
@@ -166,6 +167,7 @@ func addFlags(flags *pflag.FlagSet) *containerOptions {
 	flags.VarP(&copts.attach, "attach", "a", "Attach to STDIN, STDOUT or STDERR")
 	flags.Var(&copts.deviceCgroupRules, "device-cgroup-rule", "Add a rule to the cgroup allowed devices list")
 	flags.Var(&copts.devices, "device", "Add a host device to the container")
+	flags.Var(&copts.gpus, "gpus", "Request GPU devices for the container ('all' to pass all GPUs)")
```
thaJeztah: This needs an API version annotation

RenaudWasTaken:
Looks good to me!

thaJeztah:
This can be rebased now (moby vendor bump was merged)

cli/command/container/opts.go
```diff
@@ -557,6 +561,11 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		Ulimits:              copts.ulimits.GetList(),
 		DeviceCgroupRules:    copts.deviceCgroupRules.GetAll(),
 		Devices:              deviceMappings,
+		DeviceRequests:       deviceRequests,
+	}
+
+	if copts.pidsLimit != 0 {
+		resources.PidsLimit = &copts.pidsLimit
```
thaJeztah: Hm; did we not yet have a PR for this 🤔

tiborvass:
Ping @thaJeztah @silvin-lubecki @vdemeester updated

opts/gpus.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		return err
+	}
+
+	req := container.DeviceRequest{Options: make(map[string]string), Capabilities: [][]string{{"gpu"}}}
+
+	// Set writable as the default
+	for _, field := range fields {
+		parts := strings.SplitN(field, "=", 2)
+		key := strings.ToLower(parts[0])
```
thaJeztah: I'd prefer to keep this strict, and not do the `ToLower`, just to prevent (e.g.) `cApaBILitiEs` being "ok" to use; better to produce an error

opts/gpus.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+		case "device":
+			req.DeviceIDs = strings.Split(value, ",")
+		case "capabilities":
+			req.Capabilities = [][]string{append(strings.Split(value, ","), "gpu")}
+		case "options":
+			r := csv.NewReader(strings.NewReader(value))
+			optFields, err := r.Read()
+			if err != nil {
+				return errors.Wrap(err, "error reading gpu options")
```
thaJeztah: Perhaps `failed to` (we use that elsewhere, but possibly in different contexts)

```suggestion
				return errors.Wrap(err, "failed to read gpu options")
```

opts/gpus.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+			req.Options = make(map[string]string)
+			for _, optField := range optFields {
+				optParts := strings.SplitN(optField, "=", 2)
+				key := strings.ToLower(optParts[0])
```
thaJeztah: same here (`ToLower`); we should pass these as-is as we don't know what options the driver expects.

I'd be ok with trimming whitespace though if needed (assuming that options/values don't allow leading/trailing whitespace)

opts/gpus.go
```diff
@@ -0,0 +1,112 @@
...
+	values []container.DeviceRequest
+}
+
+func parseCount(s string) (int, error) {
+	i := -1
+	var err error
+	if s != "all" {
+		i, err = strconv.Atoi(s)
+		if err != nil {
+			err = errors.Wrap(err, "count must be an integer")
```
silvin-lubecki: `errors.Wrap` already checks if err is nil or not.

opts/gpus.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/pkg/errors"
+)
+
+// GpuOpts is a Value type for parsing mounts
+type GpuOpts struct {
+	values []container.DeviceRequest
+}
+
+func parseCount(s string) (int, error) {
+	i := -1
```
silvin-lubecki: What about
```go
func parseCount(s string) (int, error) {
    if s == "all" {
        return -1, nil
    }
    i, err := strconv.Atoi(s)
    return i, errors.Wrap(err, "count must be an integer")
}
```

opts/gpus.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+
+	o.values = append(o.values, req)
+	return nil
+}
+
+// Type returns the type of this option
+func (o *GpuOpts) Type() string {
+	return "gpuRequest"
```
thaJeztah: perhaps `gpu-options` this is what's shown in `--help` output; also looks like we don't use camelBack there

cli/command/container/opts.go
```diff
@@ -166,6 +167,8 @@ func addFlags(flags *pflag.FlagSet) *containerOptions {
 	flags.VarP(&copts.attach, "attach", "a", "Attach to STDIN, STDOUT or STDERR")
 	flags.Var(&copts.deviceCgroupRules, "device-cgroup-rule", "Add a rule to the cgroup allowed devices list")
 	flags.Var(&copts.devices, "device", "Add a host device to the container")
+	flags.Var(&copts.gpus, "gpus", "Request GPU devices for the container ('all' to pass all GPUs)")
```
thaJeztah: Perhaps `Add GPU devices to the container` or `GPU devices to add to the container`

opts/gpus.go
```diff
@@ -0,0 +1,112 @@
...
+
+		if len(parts) == 1 {
+			req.Count, err = parseCount(key)
+			if err != nil {
+				return err
+			}
+			continue
+		}
+
+		if len(parts) != 2 {
```
silvin-lubecki: I don't think it can happen, as you use `strings.SplitN(field, "=", 2)`  🤔
It can just be len(parts) == 1 if there is no `=`, or 2 (and the second part remains unsplit), right?

opts/gpus.go
thaJeztah: 👍 I like that

opts/gpus.go
```diff
@@ -0,0 +1,112 @@
...
+			req.Driver = value
+		case "count":
+			req.Count, err = parseCount(value)
+			if err != nil {
+				return err
+			}
+		case "device":
+			req.DeviceIDs = strings.Split(value, ",")
+		case "capabilities":
+			req.Capabilities = [][]string{append(strings.Split(value, ","), "gpu")}
```
silvin-lubecki: Do you really want to reset req.Capabilities? Shouldn't it be append?

opts/gpus.go
silvin-lubecki: `errors.Wrap` already checks if err is nil.

opts/gpus.go
tiborvass: At this time yes. I don't expect to support `capabilities=foo,capabilities=bar`

opts/gpus.go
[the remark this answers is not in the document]
tiborvass: Doesn't hurt to have it both places.

opts/gpus.go
tiborvass: I'm putting `gpu-request`

opts/gpus.go
tiborvass: Yes but here we need to return early.

tiborvass:
@thaJeztah @silvin-lubecki thank you for the review, i updated the PR accordingly.

cli/command/container/opts.go
```diff
@@ -557,6 +562,7 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		Ulimits:              copts.ulimits.GetList(),
 		DeviceCgroupRules:    copts.deviceCgroupRules.GetAll(),
 		Devices:              deviceMappings,
+		DeviceRequests:       deviceRequests,
```
thaJeztah: nit: looks like it could even be inlined here

```suggestion
		DeviceRequests:       copts.gpus.Value(),
```

opts/gpus.go
```diff
@@ -0,0 +1,104 @@
...
+		case "options":
+			r := csv.NewReader(strings.NewReader(value))
+			optFields, err := r.Read()
+			if err != nil {
+				return errors.Wrap(err, "failed to read gpu options")
+			}
+			req.Options = make(map[string]string)
```
thaJeztah: oh, could actually use https://github.com/docker/cli/blob/9e712073274021ef1e552ecbb54595142b49edbb/opts/parse.go#L40-L53

opts/gpus.go
silvin-lubecki: Yeah right 😅

opts/gpus.go
silvin-lubecki: Shouldn't you warn the user someway then?

tiborvass:
@thaJeztah updated

opts/gpus.go
tiborvass: Updated

tiborvass:
@vdemeester added a couple. The most important one is `--gpus all`

mviereck:
- It seems that the `--gpus` option is tied to NVIDIA hardware only and fails with other GPU vendors.
 - It seems that the `--gpus` option is tied to NVIDIA docker images only and fails with other images.

Am I right? If yes, this new option is a shame.
Please add GPU support for all GPU vendors. Seriously.

RenaudWasTaken:
Hello @mviereck !

The `--gpus` option is implemented more as a plugin than as a "vendor-lock-in". For exemple it doesn't work until you install the nvidia-container-toolkit (in other words the nvidia plugin).

We have plans to make it more like other plugins (`docker plugin install`), so that it's easier for everyone to build their own plugins. So far we have lacked time to do so. If you want to help drive this effort we'd be happy to help you (e.g: through review)!
The way the interface for GPUs has been designed should easily map to a REST API.

mviereck:
@RenaudWasTaken Thank you for your friendly response!
> If you want to help drive this effort we'd be happy to help you (e.g: through review)!

I am not experienced in Go, so I cannot contribute code directly, sorry. 
I have opened a new ticket on this topic with some proposals, maybe you want to look at it, too: #2063

PR #1716, opened 2019-03-05
Ensure context store initialization for plugins

simonferquel:
**- What I did**

Fixed a bug in cli-plugin initialization where the context store was not correctly initialized

Plugins have their cli initialized with a docker client pre-configured
(for dial-stdio). The problem here is that the cli initialization logic
skipped initializing the context store on this condition.

**- How I did it**

This PR adds a non-regression test for that alongside a fix (context
store and current context name are now resolved unconditionally)

**- How to verify it**

Run the unit tests

cc @ijc


simonferquel:
Closing in favor of #1690 
@jcsirot can you port the unit test in your PR though ?

PR #1719, opened 2019-03-06
Makes the default docker context behavior consistent

jcsirot:
**- What I did**
Make default context behaves like a real context
- when using "--context default" parameter
- when printing the list of contexts
- when exporting the default context to a tarball

d0jf2ttx4aefudv


cli/command/context/testdata/list.golden
```diff
@@ -1,5 +1,4 @@
-NAME                DESCRIPTION                               DOCKER ENDPOINT           KUBERNETES ENDPOINT            ORCHESTRATOR
-current *           description of current                    https://someswarmserver   https://someserver (default)   all
-default             Current DOCKER_HOST based configuration                                                            
-other               description of other                      https://someswarmserver   https://someserver (default)   all
-unset               description of unset                      https://someswarmserver   https://someserver (default)   
+NAME                DESCRIPTION              DOCKER ENDPOINT           KUBERNETES ENDPOINT            ORCHESTRATOR
+current *           description of current   https://someswarmserver   https://someserver (default)   all
+other               description of other     https://someswarmserver   https://someserver (default)   all
+unset               description of unset     https://someswarmserver   https://someserver (default)
```
simonferquel: why is there no "default" context in the golden list ?

cli/context/store/store.go
```diff
@@ -68,13 +84,44 @@ func New(dir string, cfg Config) Store {
 	}
 }
 
+func (s *store) WithDefaultContextInitializer(initializer DefaultContextInitializer) Store {
```
simonferquel: package store is all about storing contexts (and retrieving them). It should remain pure and don't know anything about a "default" context

jcsirot:
@simonferquel understood. Actually I was more considering the metadatastore and tlsstore as the real storage backend and the store as a proxy in front of these implementations. I also considered adding another implementation of Store for the default context and create a sort of multiplexer but I was wondering if it wasn't over-engineering for little gains. But ok, I'm going to refactor that.

cli/command/cli.go
```diff
@@ -280,7 +286,7 @@ func newAPIClientFromEndpoint(ep docker.Endpoint, configFile *configfile.ConfigF
 }
 
 func resolveDockerEndpoint(s store.Store, contextName string, opts *cliflags.CommonOptions) (docker.Endpoint, error) {
-	if contextName != "" {
+	if contextName != "" || contextName == DefaultContextName {
```
thaJeztah: Is this logic flipped? i.e. should this be

```suggestion
	if contextName != "" && contextName != DefaultContextName {
```


But perhaps better to return early for the special case;

```suggestion
	if contextName == "" || contextName == DefaultContextName {
		return resolveDefaultDockerEndpoint(opts)
	}
```

cli/command/context/list.go
```diff
@@ -61,39 +60,19 @@ func runList(dockerCli command.Cli, opts *listOptions) error {
 		if kubernetesEndpoint != nil {
 			kubEndpointText = fmt.Sprintf("%s (%s)", kubernetesEndpoint.Host, kubernetesEndpoint.DefaultNamespace)
 		}
+		if rawMeta.Name == command.DefaultContextName {
+			meta.Description = "Current DOCKER_HOST based configuration"
+		}
 		desc := formatter.ClientContext{
 			Name:               rawMeta.Name,
-			Current:            rawMeta.Name == curContext,
+			Current:            rawMeta.Name == curContext || (curContext == "" && rawMeta.Name == command.DefaultContextName),
```
thaJeztah: Thinking if it would be clearer to move this to an `if`, and use an intermediate variable (e.g. `isCurrent`) for this

cli/command/cli.go
simonferquel: feels like the comparison here is wrong. should be:
```suggestion
	if contextName != "" && contextName != DefaultContextName {
```

cli/command/context/list.go
simonferquel: Could be simplified by putting a simpler
```
if curContext == ""{
  curContext = command.DefaultContextName
}
```
on top of the loop

cli/command/context/list_test.go
```diff
@@ -36,20 +35,6 @@ func TestList(t *testing.T) {
 	golden.Assert(t, cli.OutBuffer().String(), "list.golden")
 }
 
-func TestListNoContext(t *testing.T) {
```
simonferquel: Why is this test removed ?

cli/command/cli.go
```diff
@@ -396,7 +406,7 @@ func (cli *DockerCli) StackOrchestrator(flagValue string) (Orchestrator, error)
 	if currentContext == "" {
 		currentContext = configFile.CurrentContext
 	}
-	if currentContext != "" {
+	if currentContext != "" && currentContext != DefaultContextName {
```
thaJeztah: doesn't `ContextStoreWithDefault` now also implement `GetContextMetadata()`? IOW; do we need to still have this special case?

cli/command/defaultcontextstore.go
```diff
@@ -0,0 +1,182 @@
...
+		contextTLSData.Endpoints[docker.DockerEndpoint] = *dockerEP.TLSData.ToStoreTLSData()
+	}
+
+	// Default context uses env-based kubeconfig for Kubernetes endpoint configuration
+	kubeconfig := os.Getenv("KUBECONFIG")
+	if kubeconfig == "" {
+		kubeconfig = filepath.Join(homedir.Get(), ".kube/config")
+	}
+	kubeEP, err := kubernetes.FromKubeConfig(kubeconfig, "", "")
+	if stackOrchestrator == OrchestratorKubernetes && err != nil {
```
simonferquel: Should we also handle OrchestratorAll ?

cli/command/defaultcontextstore.go
```diff
@@ -0,0 +1,182 @@
...
+
+// GetContextTLSData implements store.Store's GetContextTLSData
+func (s *ContextStoreWithDefault) GetContextTLSData(contextName, endpointName, fileName string) ([]byte, error) {
+	if contextName == DefaultContextName {
+		defaultContext, err := s.Resolver()
+		if err != nil {
+			return nil, err
+		}
+		if defaultContext.TLS.Endpoints[endpointName].Files[fileName] == nil {
+			return nil, errors.Errorf("tls data for %s/%s/%s does not exist", DefaultContextName, endpointName, fileName)
```
simonferquel: I think there is a typed error interface for not-found cases in the store package. It should return an error satisfying this interface

cli/command/context/export.go
```diff
@@ -77,7 +77,7 @@ func writeTo(dockerCli command.Cli, reader io.Reader, dest string) error {
 
 // RunExport exports a Docker context
 func RunExport(dockerCli command.Cli, opts *ExportOptions) error {
-	if err := validateContextName(opts.ContextName); err != nil {
+	if err := validateContextName(opts.ContextName); err != nil && opts.ContextName != "default" {
```
thaJeztah: Should probably use the `const` here

```suggestion
	if err := validateContextName(opts.ContextName); err != nil && opts.ContextName != command.DefaultContextName {
```

Also wondering if this exception is still needed, if `ContextStoreWithDefault ` implements the interface (so `GetContextMetadata` would work?)

cli/command/defaultcontextstore.go
```diff
@@ -0,0 +1,182 @@
...
+func (s *ContextStoreWithDefault) RemoveContext(name string) error {
+	if name == DefaultContextName {
+		return errors.New("default context and cannot be removed")
+	}
+	return s.Store.RemoveContext(name)
+}
+
+// GetContextMetadata implements store.Store's GetContextMetadata
+func (s *ContextStoreWithDefault) GetContextMetadata(name string) (store.ContextMetadata, error) {
+	if name == DefaultContextName {
```
simonferquel: Not sure about this, but should we also accept an empty name here ?

cli/command/defaultcontextstore.go
```diff
@@ -0,0 +1,182 @@
...
+func (s *ContextStoreWithDefault) ResetContextEndpointTLSMaterial(contextName string, endpointName string, data *store.EndpointTLSData) error {
+	if contextName == DefaultContextName {
+		return errors.New("The default context store does not support ResetContextEndpointTLSMaterial")
+	}
+	return s.Store.ResetContextEndpointTLSMaterial(contextName, endpointName, data)
+}
+
+// ListContextTLSFiles implements store.Store's ListContextTLSFiles
+func (s *ContextStoreWithDefault) ListContextTLSFiles(name string) (map[string]store.EndpointFiles, error) {
+	if name == DefaultContextName {
```
simonferquel: Same here

cli/command/defaultcontextstore.go
```diff
@@ -0,0 +1,182 @@
...
+			tlsfiles[epName] = files
+		}
+		return tlsfiles, nil
+	}
+	return s.Store.ListContextTLSFiles(name)
+}
+
+// GetContextTLSData implements store.Store's GetContextTLSData
+func (s *ContextStoreWithDefault) GetContextTLSData(contextName, endpointName, fileName string) ([]byte, error) {
+	if contextName == DefaultContextName {
```
simonferquel: Same here

cli/command/defaultcontextstore.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+		return defaultContext.TLS.Endpoints[endpointName].Files[fileName], nil
+
+	}
+	return s.Store.GetContextTLSData(contextName, endpointName, fileName)
+}
+
+// GetContextStorageInfo implements store.Store's GetContextStorageInfo
+func (s *ContextStoreWithDefault) GetContextStorageInfo(contextName string) store.ContextStorageInfo {
+	if contextName == DefaultContextName {
```
simonferquel: Same here

cli/command/cli.go
```diff
@@ -210,7 +210,13 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...Initialize
 	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
 
 	if cli.client == nil {
-		cli.contextStore = store.New(cliconfig.ContextStoreDir(), cli.contextStoreConfig)
+		baseContextSore := store.New(cliconfig.ContextStoreDir(), cli.contextStoreConfig)
+		cli.contextStore = &ContextStoreWithDefault{
```
thaJeztah: Should `ContextStoreWithDefault` be in its own package? So that we could do something like;

```go
storewithdefault.New(...)
```


Perhaps even integrate with the default `store` (but not sure if there's situations where we want that to be used separate)

cli/command/context/list.go
```diff
@@ -61,39 +60,19 @@ func runList(dockerCli command.Cli, opts *listOptions) error {
 		if kubernetesEndpoint != nil {
 			kubEndpointText = fmt.Sprintf("%s (%s)", kubernetesEndpoint.Host, kubernetesEndpoint.DefaultNamespace)
 		}
+		if rawMeta.Name == command.DefaultContextName {
```
thaJeztah: If `dockerCli.ContextStore().ListContexts()` is handled by `ContextStoreWithDefault.ListContexts ()`, is this special case still needed, or would it already return the right information?

cli/command/context/list_test.go
jcsirot: Since the code handling the default context in the list command code has been removed this test did not provide additional coverage.

simonferquel:
After rebasing https://github.com/docker/cli/pull/1690, can you rebase on this PR (we have pending work on docker-app requiring both)

cli/command/cli.go
simonferquel: `ContextStoreWithDefault` needs things defined in package `command` (related to resolving the docker context based on "legacy" cli flags and env var). That would:
- either introduce a cycling dependency (dockerCli.Initialize needs this ContextStoreWithDefault)
- or move much of the command package code in a separate package
- implementing that in `store` package might introduce many new dependencies to the `store` package (and I don't feel like it is a good idea)

cli/command/context/list_test.go
simonferquel: :+1

cli/command/cli.go
thaJeztah: Could we pass those options in the constructor?

cli/command/context/list.go
jcsirot: @thaJeztah I didn't want to add specific code for import/export command to avoid, when importing an exported default context, having a second context with the "Current DOCKER_HOST based configuration" description. Therefore I decided to leave the description empty for the default context and add handle that case here.

cli/command/context/remove_test.go
```diff
@@ -62,3 +62,12 @@ func TestRemoveCurrentForce(t *testing.T) {
...
 	assert.Equal(t, "", reloadedConfig.CurrentContext)
 }
+
+func TestRemoveDefault(t *testing.T) {
+	cli, cleanup := makeFakeCli(t)
+	defer cleanup()
+	createTestContextWithKubeAndSwarm(t, cli, "other", "all")
+	cli.SetCurrentContext("current")
+	err := RunRemove(cli, RemoveOptions{}, []string{"default"})
+	assert.ErrorContains(t, err, "default: context \"default\" cannot be removed")
```
silvin-lubecki: nit: use backslash notation `

cli/context/store/store.go
```diff
@@ -12,7 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/opencontainers/go-digest"
+	digest "github.com/opencontainers/go-digest"
```
silvin-lubecki: I think this change is not needed 🤔

cli/command/defaultcontextstore.go
```diff
@@ -0,0 +1,182 @@
...
+	DefaultContextName = "default"
+)
+
+// DefaultContext contains the default context data for all enpoints
+type DefaultContext struct {
+	Meta store.ContextMetadata
+	TLS  store.ContextTLSData
+}
+
+// DefaultContextResolver is a function which resolve the default context base on the configuration the env variables
```
silvin-lubecki: nit: I think something is missing at the end of the comment + resolve is missing an s?

cli/command/defaultcontextstore.go
silvin-lubecki: You can use `stackOrchestrator.HasKubernetes` then (see [code here](https://github.com/docker/cli/blob/48bd4c6deb59784b78b860bca14b9995acef0cb3/cli/command/orchestrator.go#L27))

cli/command/defaultcontextstore.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return append(contextList, defaultContext.Meta), nil
+}
+
+// CreateOrUpdateContext is not allowed for the default context and fails
+func (s *ContextStoreWithDefault) CreateOrUpdateContext(meta store.ContextMetadata) error {
+	if meta.Name == DefaultContextName {
+		return errors.New("default context and cannot be created nor updated")
```
silvin-lubecki: nit: `and` ?

cli/command/defaultcontextstore.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return s.Store.CreateOrUpdateContext(meta)
+}
+
+// RemoveContext is not allowed for the default context and fails
+func (s *ContextStoreWithDefault) RemoveContext(name string) error {
+	if name == DefaultContextName {
+		return errors.New("default context and cannot be removed")
```
silvin-lubecki: `and` ?

cli/command/defaultcontextstore.go
silvin-lubecki: Then maybe add a function `isDefaultContext` which resolves on `default` and on empty string? And use it in the other methods `RemoveContext` and `CreateOrUpdateContext` ?

cli/command/defaultcontextstore_test.go
```diff
@@ -0,0 +1,223 @@
...
+	srcMeta, err := s.GetContextMetadata("default")
+	assert.NilError(t, err)
+	destMeta, err := s.GetContextMetadata("dest")
+	assert.NilError(t, err)
+	assert.DeepEqual(t, destMeta.Metadata, srcMeta.Metadata)
+	assert.DeepEqual(t, destMeta.Endpoints, srcMeta.Endpoints)
+	srcFileList, err := s.ListContextTLSFiles("default")
+	assert.NilError(t, err)
+	destFileList, err := s.ListContextTLSFiles("dest")
+	assert.NilError(t, err)
```
silvin-lubecki: nit: can you add some space in this block? I mean divide into logical sub-blocks, add an empty line between them and/or a comment explaining what you assert?

cli/context/store/store_test.go
```diff
@@ -55,13 +61,22 @@ func TestExportImport(t *testing.T) {
 	assert.NilError(t, err)
 	destFileList, err := s.ListContextTLSFiles("dest")
 	assert.NilError(t, err)
-	assert.DeepEqual(t, srcFileList, destFileList)
-	srcData, err := s.GetContextTLSData("source", "ep1", "file1")
+	assert.Equal(t, 1, len(destFileList))
```
silvin-lubecki: nit: same here for adding more "space"

cli/command/defaultcontextstore_test.go
```diff
@@ -0,0 +1,223 @@
...
+	destData1, err := s.GetContextTLSData("dest", "ep2", "file1")
+	assert.NilError(t, err)
+	assert.DeepEqual(t, file1, destData1)
+	destData2, err := s.GetContextTLSData("dest", "ep2", "file2")
+	assert.NilError(t, err)
+	assert.DeepEqual(t, file2, destData2)
+}
+
+func TestListDefaultContext(t *testing.T) {
+	meta := testDefaultMetadata()
```
silvin-lubecki: All the following tests have the same initialization, can you factorize them using a table test pattern?

silvin-lubecki:
I like this default context store proxy on top of the context store 👍

jcsirot:
PR #1690 have been rebased on this branch

cli/command/context/list.go
jcsirot: fixed

cli/command/defaultcontextstore.go
jcsirot: Empty string is not a valid context name anymore. When not defined, the context name falls back to `default`

cli/command/defaultcontextstore_test.go
jcsirot: Indeed. Done

cli/command/defaultcontextstore.go
simonferquel: So just checked, there is no error interface but a specific error type (see here: https://github.com/docker/cli/blob/master/cli/context/store/store.go#L324). I suggest you introduce a tlsDataNotExistError interface in store package, and modify IsErrTLSDataDoesNotExist to check this interface, and makes this code return an error satisfying this interface.

cli/command/cli.go
```diff
@@ -384,38 +398,21 @@ func (cli *DockerCli) CurrentContext() string {
...
-	if configFile == nil {
-		configFile = cliconfig.LoadDefaultConfigFile(cli.Err())
-	}
-
 	currentContext := cli.CurrentContext()
-	if currentContext == "" {
-		currentContext = configFile.CurrentContext
+	ctxRaw, err := cli.ContextStore().GetContextMetadata(currentContext)
+	if store.IsErrContextDoesNotExist(err) {
+		// case where the currentContext has been removed (CLI behavior is to fallback to using DOCKER_HOST based resolution)
```
thaJeztah: Not for this PR, but something similar came up in another discussion with @simonferquel - I think this should be an error; we should not silently fall back to `DOCKER_HOST` (because then I may be running commands against the wrong environment, which is likely not what you want)

cli/command/context/list.go
thaJeztah: Still a bit on the fence on this one (i.e., the same applies to other contexts, which also by default copy the description), but not a blocker

jcsirot: Indeed. I think for other contexts it is less confusing. Maybe it would be nice to add a `--description` flag to import command in order to override the imported context description

PR #1720, opened 2019-03-06
hardening config.Path() to disallow directory traversal

zappy-shu:
**- What I did**
Resolving issue raised in [comment](https://github.com/docker/cli/pull/1564#discussion_r251822916), removing the ability to get paths from above the config directory's root path

**- How I did it**
Checks whether the resolved path is underneath the root config directory using a strings.HasPrefix on the cleaned paths. Returns an error if it's not.

**- How to verify it**
Added unit test for attempting to traverse outside of the config directory

- hardening config.Path() to disallow directory traversal

### DIFF
--- a/cli-plugins/manager/manager_test.go
+++ b/cli-plugins/manager/manager_test.go
@@ -92,10 +92,14 @@ func TestErrPluginNotFound(t *testing.T) {
 func TestGetPluginDirs(t *testing.T) {
 	cli := test.NewFakeCli(nil)
 
-	expected := []string{config.Path("cli-plugins")}
-	expected = append(expected, defaultSystemPluginDirs...)
+	pluginDir, err := config.Path("cli-plugins")
+	assert.NilError(t, err)
+	expected := append([]string{pluginDir}, defaultSystemPluginDirs...)
 
-	assert.Equal(t, strings.Join(expected, ":"), strings.Join(getPluginDirs(cli), ":"))
+	var pluginDirs []string
+	pluginDirs, err = getPluginDirs(cli)
+	assert.Equal(t, strings.Join(expected, ":"), strings.Join(pluginDirs, ":"))
+	assert.NilError(t, err)
 
 	extras := []string{
 		"foo", "bar", "baz",
@@ -104,5 +108,7 @@ func TestGetPluginDirs(t *testing.T) {
 	cli.SetConfigFile(&configfile.ConfigFile{
 		CLIPluginsExtraDirs: extras,
 	})
-	assert.DeepEqual(t, expected, getPluginDirs(cli))
+	pluginDirs, err = getPluginDirs(cli)
+	assert.DeepEqual(t, expected, pluginDirs)
+	assert.NilError(t, err)
 }
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/cli/cli/config/credentials"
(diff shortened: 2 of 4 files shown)


cli/config/config.go
```diff
@@ -47,8 +48,12 @@ func SetDir(dir string) {
 }
 
 // Path returns the path to a file relative to the config dir
-func Path(p ...string) string {
-	return filepath.Join(append([]string{Dir()}, p...)...)
+func Path(p ...string) (string, error) {
+	path := filepath.Join(append([]string{Dir()}, p...)...)
+	if !strings.HasPrefix(path, filepath.Clean(Dir())+string(filepath.Separator)) {
```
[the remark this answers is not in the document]
zappy-shu: The problem with just `filepath.Join(p...)` is when `p` is an absolute path then the join result will start with a slash or drive: [https://play.golang.org/p/Nm0ZY_C1cO5](https://play.golang.org/p/Nm0ZY_C1cO5).

Sorry, I should have (and will) add a test for that.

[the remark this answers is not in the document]
thaJeztah: Ah, good one; didn't think about that case 👍

cli/config/config_test.go
```diff
@@ -553,12 +553,19 @@ func TestConfigPath(t *testing.T) {
 	oldDir := Dir()
 
 	SetDir("dummy1")
```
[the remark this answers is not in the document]
zappy-shu: I've added tests that will also check functionality when the Dir() is set to an absolute path so I've kept this part

cli/config/config_test.go
```diff
@@ -553,12 +553,19 @@ func TestConfigPath(t *testing.T) {
...
-	f1 := Path("a", "b")
+	f1, err1 := Path("a", "b")
 	assert.Equal(t, f1, filepath.Join("dummy1", "a", "b"))
+	assert.NilError(t, err1)
 
 	SetDir("dummy2")
-	f2 := Path("c", "d")
+	f2, err2 := Path("c", "d")
```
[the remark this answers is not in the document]
zappy-shu: Good point. Thanks for the example. I've altered the tests to match this and added a few extra cases for testing absolute paths

cli-plugins/manager/manager_test.go
```diff
@@ -92,10 +92,14 @@ func TestErrPluginNotFound(t *testing.T) {
...
 	cli := test.NewFakeCli(nil)
 
-	expected := []string{config.Path("cli-plugins")}
+	pluginDir, err := config.Path("cli-plugins")
+	assert.NilError(t, err)
+	expected := []string{pluginDir}
 	expected = append(expected, defaultSystemPluginDirs...)
 
-	assert.Equal(t, strings.Join(expected, ":"), strings.Join(getPluginDirs(cli), ":"))
+	pluginDirs1, err1 := getPluginDirs(cli)
```
ijc: nit: you can (and should) reuse `pluginDir` and `err` here and the same for the `2`s below.

cli/config/config_test.go
```diff
@@ -552,13 +552,73 @@ func TestLoadDefaultConfigFile(t *testing.T) {
...
+		},
+		{
+			name:        "invalid_absolute_path_unix",
+			path:        []string{"/a", "..", ".."},
+			dir:         "root_dir",
+			expectedErr: `is outside of root config directory "root_dir"`,
+		},
+		{
+			name:        "invalid_absolute_path_windows",
+			path:        []string{"C:/", "..", ".."},
```
ijc: Does this test DTRT when `GOOS=linux` (or `!=windows` generally)?

thaJeztah: TIL [DTRT](https://www.urbandictionary.com/define.php?term=DTRT) - had to look that one up 😅

cli-plugins/manager/manager_test.go
```diff
@@ -92,10 +92,14 @@ func TestErrPluginNotFound(t *testing.T) {
 func TestGetPluginDirs(t *testing.T) {
 	cli := test.NewFakeCli(nil)
 
-	expected := []string{config.Path("cli-plugins")}
+	pluginDir, err := config.Path("cli-plugins")
+	assert.NilError(t, err)
+	expected := []string{pluginDir}
```
thaJeztah: nit: could be combined to a single line


```suggestion
	expected := append([]string{pluginDir}, defaultSystemPluginDirs...)
```

cli-plugins/manager/manager_test.go
thaJeztah: Yes, agreed; I see this PR still needs some squashing, so if you could make that change as part of that 👍

cli/config/config_test.go
thaJeztah: If we're testing for these situations, we also might need a test for 

- `C:\` (backslash)
- Embedded travering (for better name);  `[]string{"a", "../../..", "b"}` (although we might enter the area of testing Go stdlib itself)

ijc: The reason I ask is that `path/filepath` is (by definition) platform specific, so on Unix-y platforms it doesn't treat e.g .`C:/` as special (I expect it thinks it is a directory in the root called `C:`).

zappy-shu: Yeah, on second look this won't work as the build is on unix. Probably easiest to simply remove the "windows" tests for now

PR #1722, opened 2019-03-07
Allow plugins to have --flags which are the same as the top-level 

### LINKED ISSUE #1699: Plugins can't re-use the same flags as cli global flags
**Description**
A plugin can't use the same name as one of the cli global flags (`debug`, `log-level`, `context`, ...), or worse can't use the same shortcut (ex: `-c`).

If I add a new flag to the `hello-world` plugin like this following:

```go
flags.StringVarP(&compose, "compose", "c", "", "compose file")
```

and try to use it:
```console
$ docker helloworld -c my-compose-file.yml
unable to resolve docker endpoint: context "my-compose-file.yml" does not exist

$ docker --help
...
```

This context flag is a root command flag, not a PersistentFlag and so shouldn't be used by the docker cli and forwarded directly to the plugin.

ijc:
This was complicated. The final commit message explains it best:
```
    allow plugins to have argument which match a top-level flag.
    
    The issue with plugin options clashing with globals is that when cobra is
    parsing the command line and it comes across an argument which doesn't start
    with a `-` it (in the absence of plugins) distinguishes between "argument to
    current command" and "new subcommand" based on the list of registered sub
    commands.
    
    Plugins breaks that model. When presented with `docker -D plugin -c foo` cobra
    parses up to the `plugin`, sees it isn't a registered sub-command of the
    top-level docker (because it isn't, it's a plugin) so it accumulates it as an
    argument to the top-level `docker` command. Then it sees the `-c`, and thinks
    it is the global `-c` (for AKA `--context`) option and tries to treat it as
    that, which fails.
    
    In the specific case of the top-level `docker` subcommand we know that it has
    no arguments which aren't `--flags` (or `-f` short flags) and so anything which
    doesn't start with a `-` must either be a (known) subcommand or an attempt to
    execute a plugin.
    
    We could simply scan for and register all installed plugins at start of day, so
    that cobra can do the right thing, but we want to avoid that since it would
    involve executing each plugin to fetch the metadata, even if the command wasn't
    going to end up hitting a plugin.
    
    Instead we can parse the initial set of global arguments separately before
    hitting the main cobra `Execute` path, which works here exactly because we know
    that the top-level has no non-flag arguments.
    
    One slight wrinkle is that the top-level `PersistentPreRunE` is no longer
    called on the plugins path (since it no longer goes via `Execute`), so we
    arrange for the initialisation done there (which has to be done after global
    flags are parsed to handle e.g. `--config`) to happen explictly after the
    global flags are parsed. Rather than make `newDockerCommand` return the
    complicated set of results needed to make this happen, instead return a closure
    which achieves this.
    
    The new functionality is introduced via a common `TopLevelCommand` abstraction
    which lets us adjust the plugin entrypoint to use the same strategy for parsing
    the global arguments. This isn't strictly required (in this case the stuff in
    cobra's `Execute` works fine) but doing it this way avoids the possibility of
    subtle differences in behaviour.
    
    Fixes #1699, and also, as a side-effect, the first item in #1661.
    
```

### DIFF
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -124,8 +124,10 @@ func (tcmd *TopLevelCommand) HandleGlobalFlags() (*cobra.Command, []string, erro
 
 	// We manually parse the global arguments and find the
 	// subcommand in order to properly deal with plugins. We rely
-	// on the root command never having any non-flag arguments.
-	flags := cmd.Flags()
+	// on the root command never having any non-flag arguments. We
+	// create our own FlagSet so that we can configure it
+	// (e.g. `SetInterspersed` below) in an idempotent way.
+	flags := pflag.NewFlagSet(cmd.Name(), pflag.ContinueOnError)
 
 	// We need !interspersed to ensure we stop at the first
 	// potential command instead of accumulating it into
@@ -133,9 +135,9 @@ func (tcmd *TopLevelCommand) HandleGlobalFlags() (*cobra.Command, []string, erro
 	// arguments which we try and treat as globals (when they are
 	// actually arguments to the subcommand).
 	flags.SetInterspersed(false)
-	defer flags.SetInterspersed(true) // Undo, any subsequent cmd.Execute() in the caller expects this.
 
 	// We need the single parse to see both sets of flags.
+	flags.AddFlagSet(cmd.Flags())
 	flags.AddFlagSet(cmd.PersistentFlags())
 	// Now parse the global flags, up to (but not including) the
 	// first command. The result will be that all the remaining
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -2,9 +2,11 @@ package cli
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	pluginmanager "github.com/docker/cli/cli-plugins/manager"
+	"github.com/docker/cli/cli/command"
 	cliconfig "github.com/docker/cli/cli/config"
 	cliflags "github.com/docker/cli/cli/flags"
 	"github.com/docker/docker/pkg/term"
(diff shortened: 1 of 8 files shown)


thaJeztah:
One small linting error;

```
cmd/docker/docker.go:82:23:warning: dockerCli can be github.com/docker/cli/cli/command.Cli (interfacer)
cmd/docker/docker.go:122:18:warning: dockerCli can be github.com/docker/cli/cli/command.Cli (interfacer)
cmd/docker/docker.go:185:19:warning: dockerCli can be github.com/docker/cli/cli/command.Cli (interfacer)
```

e2e/cli-plugins/flags_test.go
```diff
@@ -17,3 +19,82 @@ func TestRunGoodArgument(t *testing.T) {
...
+		})
+	}
+}
+
+// TestUnknownGlobal checks that unknown globals report errors
+func TestUnknownGlobal(t *testing.T) {
+	run, _, cleanup := prepare(t)
+	defer cleanup()
+
+	t.Run("no-val", func(t *testing.T) {
```
thaJeztah: Perhaps;

```go
// TestUnknownGlobal checks that unknown globals report errors
func TestUnknownGlobal(t *testing.T) {
	run, _, cleanup := prepare(t)
	defer cleanup()

	for name, args := range map[string][]string{
		"no-val":       {"--unknown", "helloworld"},
		"separate-val": {"--unknown", "foo", "helloworld"},
		"joined-val":   {"--unknown=foo", "helloworld"},
	} {
		t.Run(name, func(t *testing.T) {
			res := icmd.RunCmd(run(args...))
			res.Assert(t, icmd.Expected{
				ExitCode: 125,
				Out:      "",
				Err:      "unknown flag: --unknown",
			})
		})
	}
}
```

e2e/cli-plugins/flags_test.go
```diff
@@ -17,3 +19,82 @@ func TestRunGoodArgument(t *testing.T) {
 		Out:      "Hello Cleveland!",
 	})
 }
+
+// TestClashWithGlobalArgs ensures correct behaviour when a plugin
+// has an argument with the same name as one of the globals.
+func TestClashWithGlobalArgs(t *testing.T) {
```
thaJeztah: (not a blocker for this PR); we should evaluate the naming of these tests, so make them easier to "grep" (which can be useful to run a subset of tests), i.e. give them a consistent prefix (e.g. `TestPluginGlobalArgsClash`, `TestPluginGlobalArgsUnknown`)

e2e/cli-plugins/flags_test.go
ijc: Ack, will do this. NB `Out: ""` needs to be `Out: icmd.None` to actually check it is empty, `""` means "no check". I only just discovered this, I had been doing it manually everywhere.

ijc:
> and also, as a side-effect, the first item in #1661.

I'm going to add an e2e for this case here too once I'm done with the feedback so far

ijc:
Test failure ends with
```
10:44:23 Step 2/6 : RUN apk --no-cache add shadow openssh-server &&   groupadd -f docker &&   useradd -m penguin &&   usermod -aG docker penguin &&   usermod -p $(head -c32 /dev/urandom | base64) penguin &&   chsh -s /bin/sh penguin &&   ssh-keygen -A
10:44:23  ---> Running in da24ce61c17f
10:44:23 fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/main/x86_64/APKINDEX.tar.gz
10:44:23 fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/community/x86_64/APKINDEX.tar.gz
10:44:23 (1/5) Installing openssh-keygen (7.9_p1-r4)
10:44:23 (2/5) Installing openssh-server-common (7.9_p1-r4)
10:44:23 (3/5) Installing openssh-server (7.9_p1-r4)
10:44:23 (4/5) Installing linux-pam (1.3.0-r0)
10:44:23 (5/5) Installing shadow (4.5-r0)
10:44:23 Executing busybox-1.29.3-r10.trigger
10:44:23 OK: 22 MiB in 44 packages
10:44:23 Creating mailbox file: No such file or directory
10:44:24 Password: chsh: PAM: Authentication token manipulation error
10:44:24 [0mService 'engine' failed to build: The command '/bin/sh -c apk --no-cache add shadow openssh-server &&   groupadd -f docker &&   useradd -m penguin &&   usermod -aG docker penguin &&   usermod -p $(head -c32 /dev/urandom | base64) penguin &&   chsh -s /bin/sh penguin &&   ssh-keygen -A' returned a non-zero code: 1
10:44:24 docker.Makefile:152: recipe for target 'test-e2e-connhelper-ssh' failed
10:44:24 make: *** [test-e2e-connhelper-ssh] Error 1
```
Which I don't think can be due to my changes here... I'm about to push an update so if this was a flake it'll go away then I guess.

e2e/cli-plugins/flags_test.go
thaJeztah: nice find; didn't know that; that's a bit confusing 😕

cli/cobra.go
```diff
@@ -84,6 +86,77 @@ func FlagErrorFunc(cmd *cobra.Command, err error) error {
...
+	// subcommand in order to properly deal with plugins. We rely
+	// on the root command never having any non-flag arguments.
+	flags := cmd.Flags()
+
+	// We need !interspersed to ensure we stop at the first
+	// potential command instead of accumulating it into
+	// flags.Args() and then continuing on and finding other
+	// arguments which we try and treat as globals (when they are
+	// actually arguments to the subcommand).
+	flags.SetInterspersed(false)
```
silvin-lubecki: I think this is smelly, but the only other way I see is to clone the flags, and then set the interspersed, add the flag set with persistent flags... so `HandleGlobalFlags` has no side effects anymore.
That said, the effort may not be worth it...

ijc: Ah, no, that's a good idea actually. I shall investigate.

ijc: Hrm, looks like cloning the flags is not as trivial as I would have hoped :-/ at least there is no handy looking method I can see and I'm not sure if a simple copy of the pointed to object is sufficient (but I think probably not).

silvin-lubecki: As far as I remember, there's a `VisitAll` on flags but yep it's not trivial. Maybe we can add 
 a `Clone` method upstream?

cmd/docker/docker_test.go
```diff
@@ -30,17 +32,23 @@ var discard = ioutil.NopCloser(bytes.NewBuffer(nil))
 func TestExitStatusForInvalidSubcommandWithHelpFlag(t *testing.T) {
 	cli, err := command.NewDockerCli(command.WithInputStream(discard), command.WithCombinedStreams(ioutil.Discard))
 	assert.NilError(t, err)
-	cmd := newDockerCommand(cli)
-	cmd.SetArgs([]string{"help", "invalid"})
+	tcmd := newDockerCommand(cli)
```
silvin-lubecki: Can you factorize these 5 lines in a helper test function (and [flagged](https://golang.org/pkg/testing/#T.Helper) as such)? I think they add some boiling plate code to these pretty succinct tests. (and by the way if you can

thaJeztah: Probably better to combine some of these tests and rewrite them as subtests (but I'm ok with doing that in a follow up (also see my comment around naming; https://github.com/docker/cli/pull/1722/files/9457f1f070e8fdb6028a5767c78b95d5f09d68d0#r263477744)

cli/cobra.go
ijc: @silvin-lubecki Please have a look at https://github.com/docker/cli/pull/1722/commits/73f5adb459e450475dd7b30d01e78e6a981e21c6 and see if you think it is ok.

It doesn't try to copy all of the settings from the flagset to our local one, but I think it does enough.

silvin-lubecki:
The validate task failed
```time make -B vendor
make[1]: Entering directory '/go/src/github.com/docker/cli'
rm -rf vendor
bash -c 'vndr |& grep -v -i clone'
2019/03/12 12:35:21 Collecting initial packages
Too long with no output (exceeded 10m0s)``` 😞

ijc:
> `Too long with no output (exceeded 10m0s)`

It worked for me locally in about 7m. GH was having an outage earlier, it's probably that.

cli/cobra.go
silvin-lubecki: Great, you found a nice way to construct another flag set 👍

PR #1731, opened 2019-03-12
cli: container: remove logging in wait-for-exit paths

cyphar:
The wait code can output log messages, which can end up with you seeing
"errors" due to not being able to get the exit code, when in reality the
context is timing our or being cancelled. I'm not sure how common this
is on x86, but I can see it all the time on arm64 -- and these error
messages seem quite noisy to me (they also cause the integration-cli
tests to become quite flaky in my testing).

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -166,7 +166,11 @@ func runContainer(dockerCli command.Cli, opts *runOptions, copts *containerOptio
 		defer close()
 	}
 
-	statusChan := waitExitOrRemoved(ctx, dockerCli, createResponse.ID, copts.autoRemove)
+	// We don't need to wait if we're not attaching.
+	var statusChan <-chan int
+	if attach {
+		statusChan = waitExitOrRemoved(ctx, dockerCli, createResponse.ID, copts.autoRemove)
+	}
 
 	//start the container
 	if err := client.ContainerStart(ctx, createResponse.ID, types.ContainerStartOptions{}); err != nil {
@@ -186,7 +190,7 @@ func runContainer(dockerCli command.Cli, opts *runOptions, copts *containerOptio
 		return runStartContainerErr(err)
 	}
 
-	if (config.AttachStdin || config.AttachStdout || config.AttachStderr) && config.Tty && dockerCli.Out().IsTerminal() {
+	if attach && config.Tty && dockerCli.Out().IsTerminal() {
 		if err := MonitorTtySize(ctx, dockerCli, createResponse.ID, false); err != nil {
 			fmt.Fprintln(stderr, "Error monitoring TTY size:", err)
 		}


cyphar:
This is touching some historically hairy code, sorry for awakening this beast again. :P

thaJeztah:
Oh boy, this is indeed some hairy code; wondering if `legacyWaitExitOrRemoved` is still needed in this situation?

@cpuguy83 @mlaventure ptal

cli/command/container/run.go
```diff
@@ -205,7 +209,7 @@ func runContainer(dockerCli command.Cli, opts *runOptions, copts *containerOptio
 	}
 
 	// Detached mode: wait for the id to be displayed and return.
-	if !config.AttachStdout && !config.AttachStderr {
+	if !attach {
```
cpuguy83: This is slightly different b/c `attach` is looking at AttachStdin as well. Is this intended?

cyphar: I'm not sure, my reading was that the original code (which didn't account for `AttachStdin`) was wrong -- because if you attach *just* stdin then `docker run` will exit prematurely (possibly not even accepting input) and won't wait for the container to exit.

For instance, `docker run -a stdin busybox tee /foo` will only read a single line of input before exiting. But if you want to keep the old behaviour, I can change this back.

cpuguy83: Hmm... I'm thinking if it's not material to fixing the issue you are seeing let's keep as is.

cyphar: I've removed it.

thaJeztah:
@cpuguy83 ptal

cli/command/container/run.go
```diff
@@ -166,7 +166,11 @@ func runContainer(dockerCli command.Cli, opts *runOptions, copts *containerOptio
 		defer close()
 	}
 
-	statusChan := waitExitOrRemoved(ctx, dockerCli, createResponse.ID, copts.autoRemove)
+	// We don't need to wait if we're not attaching.
+	var statusChan <-chan int
+	if attach {
```
cyphar: I'm not sure if `attach` is right here, but I really don't understand `config.Attach*` at all anyway. Right now we might get a panic if you do `-a stdin` (or in some of the error paths) because we might read from a nil chan -- but I don't see a nice way of solving that.

A simpler way of stopping the issue behind this problem is to just drop the logging code from `waitExitOrRemoved`.

cpuguy83: Reading from a nil chain would block forever, not panic.
So we either need another `if attach` before reading below (so as not to block forever) or make a channel and close it...

I agree `waitExitOrRemoved` should not be logging.
It could return a channel can handle both exit status and error messages.

I think removing logging, or some way to gate logging, might be a better approach.

PR #1740, opened 2019-03-16
build: enable platform flag for build if buildkit

tonistiigi:
Platform flag is supported if BuildKit is being used. If not (lcow) then I left it for the old behavior. I don't think we want to move this flag out of experimental for other commands like pull and run cause their implementation will change with future containerd storage integration. 

@tiborvass 

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -146,7 +146,16 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 	flags.StringVar(&options.imageIDFile, "iidfile", "", "Write the image ID to the file")
 
 	command.AddTrustVerificationFlags(flags, &options.untrusted, dockerCli.ContentTrustEnabled())
-	command.AddPlatformFlag(flags, &options.platform)
+
+	flags.StringVar(&options.platform, "platform", os.Getenv("DOCKER_DEFAULT_PLATFORM"), "Set platform if server is multi-platform capable")
+	// Platform is not experimental when BuildKit is used
+	buildkitEnabled, err := command.BuildKitEnabled(dockerCli.ServerInfo())
+	if err == nil && buildkitEnabled {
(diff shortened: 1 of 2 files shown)


cli/command/image/build.go
```diff
@@ -146,7 +146,16 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 	flags.StringVar(&options.imageIDFile, "iidfile", "", "Write the image ID to the file")
 
 	command.AddTrustVerificationFlags(flags, &options.untrusted, dockerCli.ContentTrustEnabled())
-	command.AddPlatformFlag(flags, &options.platform)
+
+	flags.StringVar(&options.platform, "platform", os.Getenv("DOCKER_DEFAULT_PLATFORM"), "Set platform if server is multi-platform capable")
```
thaJeztah: Can you document this environment-variable? https://github.com/docker/cli/blob/master/docs/reference/commandline/cli.md#environment-variables

tonistiigi: Added. Note that this PR does not add this env, but just moves around the code.

cli/command/image/build.go
```diff
@@ -146,7 +146,16 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
...
+	// Platform is not experimental when BuildKit is used
+	buildkitEnabled, err := command.BuildKitEnabled(dockerCli.ServerInfo())
```
simonferquel: I think this is buggy:
When we are creating commands and flags, we have not yet run PersistentPreRun, and so I think the CLI is not yet correctly initialized. I think `docker --context=some-context build --platform ...`  can then have a very wierd behavior (where flag support is evaluated for a daemon that is not the one used after flags are actually parsed)
WDYT @thaJeztah @vdemeester @silvin-lubecki  ?

PR #1758, opened 2019-03-19
Update some dependencies

thaJeztah:

### DIFF
--- a/vendor.conf
+++ b/vendor.conf
@@ -62,7 +62,7 @@ github.com/opencontainers/runc 12f6a991201fdb8f82579582d5e00e28fba06d0a
 github.com/opencontainers/runtime-spec 29686dbc5559d93fb1ef402eeda3e35c38d75af4 # v1.0.1-59-g29686db
 github.com/opentracing/opentracing-go 1361b9cd60be79c4c3a7fa9841b3c132e40066a7
 github.com/peterbourgon/diskv 5f041e8faa004a95c88a202771f4cc3e991971e6 # v2.0.1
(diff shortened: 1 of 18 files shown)


vendor.conf
```diff
@@ -43,10 +43,10 @@ github.com/grpc-ecosystem/grpc-gateway 1a03ca3bad1e1ebadaedd3abb76bc58d4ac8143b
 github.com/grpc-ecosystem/grpc-opentracing 8e809c8a86450a29b90dcc9efbf062d0fe6d9746
 github.com/hashicorp/golang-lru 0fb14efe8c47ae851c0034ed7a448854d3d34cf3
 github.com/hashicorp/go-version 23480c0
-github.com/imdario/mergo v0.3.6
+github.com/imdario/mergo v0.3.7
 github.com/inconshreveable/mousetrap 76626ae9c91c4f2a10f34cad8ce83ea42c93bb75 # v1.0
-github.com/json-iterator/go 1624edc4454b8682399def8740d46db5e4362ba4 # 1.1.5
-github.com/mattn/go-shellwords v1.0.3
```
silvin-lubecki: I thought the idea was to always pin the version with the sha1 and add the semver as comment ?
Shouldn't we do that each time we bump dependencies, so then all the `vendor.conf` will be homogeneous?

thaJeztah: I can update if you prefer; I kept these because they already were pinned by tag. Let me push an update

thaJeztah: updated PTAL

PR #1760, opened 2019-03-19
Search Windows CLI plugins also in ProgramFiles

StefanScherer:
**- What I did**

On Windows also search for CLI plugins in the `C:\Program Files\Docker\cli-plugins` folder.
This would simplify the installation of the docker.zip file with the PowerShell module via the DockerMsftProvider or DockerProvider. They both just unzip the file into C:\Program Files
We only have to put the CLI plugins in the zip into `docker/cli-plugins/` and they will be just right in place.

**- How I did it**

**- How to verify it**

On Windows also search for CLI plugins in the `C:\Program Files\Docker\cli-plugins` folder.

IMG_5280

### DIFF
--- a/cli-plugins/manager/manager_windows.go
+++ b/cli-plugins/manager/manager_windows.go
@@ -7,4 +7,5 @@ import (
 
 var defaultSystemPluginDirs = []string{
 	filepath.Join(os.Getenv("ProgramData"), "Docker", "cli-plugins"),
+	filepath.Join(os.Getenv("ProgramFiles"), "Docker", "cli-plugins"),
 }
--- a/docs/extend/cli_plugins.md
+++ b/docs/extend/cli_plugins.md
@@ -111,7 +111,8 @@ the `/usr/local/lib` or `/usr/local/libexec` equivalents but packages
 should not do so.
 
 Plugins distributed on Windows for system wide installation should be
-installed in `%PROGRAMDATA%\Docker\cli-plugins`.
+installed in either `%ProgramData%\Docker\cli-plugins` or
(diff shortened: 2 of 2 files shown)


StefanScherer:
@ijc A good question. I don't have a hard preference, it looked plausible for me to put ProgramFiles in front as a Windows Server installation has the plugins there.

There always is another user specific folder that is searched, so the order would be
1. $env:USERPROFILE/.docker/cli-plugins
2. $env:ProgramFiles/Docker/cli-plugins
3. $env:ProgramData/Docker/cli-plugins

Any concerns from the Desktop side about this order?

thaJeztah:
Perhaps swap 2 and 3 (as in "ProgramFiles" feels more like a default location where things are installed, and "ProgramData" more "runtime", thus "override"; this post seems to confirm that https://stackoverflow.com/a/27497662/1811501)

StefanScherer:
@thaJeztah All right, I've swapped the two pathes to have ProgramData before ProgramFiles

thaJeztah:
LGTM, thanks!

gtardif:
Yes, This will help a bit as well for Desktop Enterprise to allow deploy plugins either in `Program Files` or `ProgramData`. (no preference between the 2 or on the order)

ijc:
> All right, I've swapped the two pathes to have ProgramData before ProgramFiles

If we assume that "packaged" installs will generally statically installing things in `ProgramFiles` then that makes sense I think, it leaves `ProgramData` for "system level overrides" (be that ones done by the admin or ones done by applications dynamically somehow), the "user level overrides" in `%HOME%` still take overall precedence, which is good.

Still LGTM then!

PR #1766, opened 2019-03-21
build: allow setting buildkit outputs

tonistiigi:
Introduce `-o/--output` to `docker build` allowing configuration of how the build output should be handled on buildkit mode.

Currently only supports `local` exporter.

`-o type=local,dest=outpath`
`-o ./path/to/out` - non csv value defaults to output path

Would be good to add `tarball` exporter as well before we ship this because local exporter can not preserve file uid/gid. Other exporters that are supported by buildkit need to wait for the containerd support to land in moby first.

@tiborvass @AkihiroSuda 

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -5,6 +5,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"encoding/csv"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -73,6 +74,7 @@ type buildOptions struct {
 	untrusted      bool
 	secrets        []string
 	ssh            []string
+	outputs        []string
 }
 
 // dockerfileFromStdin returns true when the user specified that the Dockerfile
@@ -176,6 +178,11 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 	flags.StringArrayVar(&options.ssh, "ssh", []string{}, "SSH agent socket or keys to expose to the build (only if BuildKit enabled) (format: default|<id>[=<socket>|<key>[,<key>]])")
 	flags.SetAnnotation("ssh", "version", []string{"1.39"})
 	flags.SetAnnotation("ssh", "buildkit", nil)
+
+	flags.StringArrayVarP(&options.outputs, "output", "o", []string{}, "Output destination (format: type=local,dest=path)")
+	flags.SetAnnotation("output", "version", []string{"1.40"})
+	flags.SetAnnotation("output", "buildkit", nil)
+
 	return cmd
 }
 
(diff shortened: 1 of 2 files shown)


cli/command/image/build_buildkit.go
```diff
@@ -117,6 +117,22 @@ func runBuildBuildKit(dockerCli command.Cli, options buildOptions) error {
...
+		return errors.Wrapf(err, "failed to parse outputs")
+	}
+
+	for _, out := range outputs {
+		if out.Type == "local" {
+			outDir, ok := out.Attrs["dest"]
+			if !ok {
+				return errors.Errorf("dest is required for local output")
+			}
+			delete(out.Attrs, "dest")
```
tiborvass: why?

tonistiigi: cause dest is client side only and should not be sent to the daemon

cli/command/image/build.go
```diff
@@ -643,3 +650,49 @@ func imageBuildOptions(dockerCli command.Cli, options buildOptions) types.ImageB
...
+
+		out := types.ImageBuildOutput{
+			Attrs: map[string]string{},
+		}
+		for _, field := range fields {
+			parts := strings.SplitN(field, "=", 2)
+			if len(parts) != 2 {
+				return nil, errors.Errorf("invalid value %s", field)
+			}
+			key := strings.ToLower(parts[0])
```
tiborvass: @thaJeztah doesn't like ToLower. He wants it strict.

tonistiigi: too bad for @thaJeztah then https://github.com/docker/cli/blob/774d78fcb84053f7648fae2358d7657feeb6b03b/cli/command/image/build_buildkit.go#L397

cli/command/image/build.go
```diff
@@ -643,3 +650,49 @@ func imageBuildOptions(dockerCli command.Cli, options buildOptions) types.ImageB
...
+	}
+	for _, s := range inp {
+		csvReader := csv.NewReader(strings.NewReader(s))
+		fields, err := csvReader.Read()
+		if err != nil {
+			return nil, err
+		}
+		if len(fields) == 1 && fields[0] == s {
+			outs = append(outs, types.ImageBuildOutput{
+				Type: "local",
```
AkihiroSuda: off-topic: do we want to port over this default convention to buildctl?

tonistiigi: `buildctl` doesn't necessarily need to be the most user-friendly but expose all API capabilities. If this becomes common I see no reason not to do it. But I wouldn't want to go over the other flags(eg. cache import/export) of `buildctl` trying to find similar patterns atm.

cli/command/image/build.go
thaJeztah: Yes, ToNIs, noT reAlLy a FaN of toLOWER heRE

Better keep it strict, and we can loosen up if we really would see a need

cli/command/image/build.go
```diff
@@ -643,3 +650,49 @@ func imageBuildOptions(dockerCli command.Cli, options buildOptions) types.ImageB
...
+			continue
+		}
+
+		out := types.ImageBuildOutput{
+			Attrs: map[string]string{},
+		}
+		for _, field := range fields {
+			parts := strings.SplitN(field, "=", 2)
+			if len(parts) != 2 {
+				return nil, errors.Errorf("invalid value %s", field)
```
thaJeztah: We don't have (or will have) options that don't require a value? Otherwise we may defer this to the daemon, and use 
https://github.com/docker/cli/blob/9e712073274021ef1e552ecbb54595142b49edbb/opts/parse.go#L40-L53

cli/command/image/build_buildkit.go
thaJeztah: Perhaps add a comment to mention that

cli/command/image/build.go
```diff
@@ -176,6 +178,11 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 	flags.StringArrayVar(&options.ssh, "ssh", []string{}, "SSH agent socket or keys to expose to the build (only if BuildKit enabled) (format: default|<id>[=<socket>|<key>[,<key>]])")
 	flags.SetAnnotation("ssh", "version", []string{"1.39"})
 	flags.SetAnnotation("ssh", "buildkit", nil)
+
+	flags.StringArrayVarP(&options.outputs, "output", "o", []string{}, "Output destination (format: type=local,dest=path)")
```
thaJeztah: did we decide on the shorthand `-o` option? (just to be sure; I'm cautious adding shorthands if other options may be added that also start with the same letter)

thaJeztah:
@vdemeester @silvin-lubecki ptal

cli/command/image/build.go
tonistiigi: `-o` is a common convention in many other build tools for output. I think it would be weird to ever use it for anything else in `docker build`

cli/command/image/build.go
tonistiigi: What I referred above is that everything else already is case insensitive. This includes csv for build secrets, swarm secrets, swarm config, mounts, network and ports. CSV in Dockerfile flags in daemon side is also case insensitive.

Btw, I'm not in favor of documenting it.

cli/command/image/build.go
tonistiigi: The API already accepts these values in a structured form and correctly formatted. We also need to parse out the values that are handled on the client side.

PR #317, opened 2017-07-10
Cli change to pass driver specific options to docker run

abhi:
Opened this PR since docker/cli#156 was closed. 

The commit contains cli changes to support driver options for a network in docker run and docker network connect cli's.
The driver-opt, aliases is now supported in the form of csv as per network option in service commands in swarm mode since docker/cli#62 . This commit extends this support to docker run command as well. 
For docker connect command --driver-opt is added to pass driver specific options for the network the container is connecting to.

Following is supported:
```
docker run -itd --net name=docknet,alias=web1.0,driver-opt=field1=value1 alpine
docker run -itd --net name=docknet,driver-opt=field1=value1 --net-alias=web1.0 alpine
docker run -itd --net name=docknet --net-alias web1 alpine
```
```
docker connect network --driver-opt field1=value1 --alias=web2.0 docknet2 dockcontainer
```
Signed-off-by: Abhinandan Prativadi <abhi@docker.com>

PR #156, opened 2017-06-05
Cli change to pass driver specific options to docker run

abhinandanpb:
The commit contains cli changes to support driver options for a network in docker run and docker network connect cli's.
The driver-opt, aliases is now supported in the form of csv as per network option in service commands in swarm mode since docker/cli#62 . This commit extends this support to docker run command as well. 
For docker connect command --driver-opt is added to pass driver specific options for the network the container is connecting to.

Following is supported:
```
docker run -itd --net name=docknet,alias=web1.0,driver-opt=field1=value1 alpine
docker run -itd --net name=docknet,driver-opt=field1=value1 --net-alias=web1.0 alpine
docker run -itd --net name=docknet --net-alias web1 alpine
```
```
docker connect network --driver-opt field1=value1 --net-alias=web2.0 docknet2 dockcontainer
```
Signed-off-by: Abhinandan Prativadi <abhi@docker.com>

cc @mavenugo

PR #62, opened 2017-05-10
Adding network specific option in csv format for service create/update

abhinandanpb:
**- What I did**
Changed the network option to accept csv format values in service/create update

**- How I did it**
Created a new networkOpt type which accepts csv format values

**- How to verify it**
```
docker service create --name web --network name=docknet,alias=web1,driver-opt=field1=value1 nginx
docker service create --name web --network docknet nginx
docker service update web --network-add name=docknet,alias=web1,driver-opt=field1=value1
docker service update web --network-rm docknet

```
The cli change will allow users to pass driver options, alias and network name as a part of service create and service update. The driver options information will eventually trickle down as a part of endpoint create and join which will allow the drivers to use this information for any additional processing for the endpoint.
Alias information is already part of docker run as ```--net-alias```. However since service can be created and updated with multiple networks there is a need to pass network specific aliases which can also be done with this change.

Relevant PRs:
docker/swarmkit#2176 and docker/swarmkit#2183 (merged)
moby/moby#33130

PR #1767, opened 2019-03-21
[carry 317] Cli change to pass driver specific options to docker run

thaJeztah:
carries https://github.com/docker/cli/pull/317
closes https://github.com/docker/cli/pull/317
closes https://github.com/docker/cli/pull/156

- partially implements: https://github.com/moby/moby/issues/31964 (RFC: add advanced "csv" syntax for "--net" / "--network")
- relates to https://github.com/docker/libnetwork/issues/2066 feature: pass IPAM options at docker run

Adds preliminary support for multiple networks on `docker run` now that this syntax would allow this; the daemon does not yet support this, so specifying multiple networks will still produce an error

Opened this PR since docker/cli#156 was closed. 

The commit contains cli changes to support driver options for a network in docker run and docker network connect cli's.
The driver-opt, aliases is now supported in the form of csv as per network option in service commands in swarm mode since docker/cli#62 . This commit extends this support to docker run command as well. 
For docker connect command --driver-opt is added to pass driver specific options for the network the container is connecting to.

Following is supported:
```
docker run -itd --net name=docknet,alias=web1.0,driver-opt=field1=value1 alpine
docker run -itd --net name=docknet,driver-opt=field1=value1 --net-alias=web1.0 alpine
docker run -itd --net name=docknet --net-alias web1 alpine
```
```
docker connect network --driver-opt field1=value1 --alias=web2.0 docknet2 dockcontainer
```
Signed-off-by: Abhinandan Prativadi <abhi@docker.com>

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -97,7 +97,7 @@ type containerOptions struct {
 	ioMaxBandwidth     opts.MemBytes
 	ioMaxIOps          uint64
 	swappiness         int64
-	netMode            string
+	netMode            opts.NetworkOpt
 	macAddress         string
 	ipv4Address        string
 	ipv6Address        string
@@ -215,8 +215,8 @@ func addFlags(flags *pflag.FlagSet) *containerOptions {
 	flags.VarP(&copts.publish, "publish", "p", "Publish a container's port(s) to the host")
 	flags.BoolVarP(&copts.publishAll, "publish-all", "P", false, "Publish all exposed ports to random ports")
 	// We allow for both "--net" and "--network", although the latter is the recommended way.
-	flags.StringVar(&copts.netMode, "net", "default", "Connect a container to a network")
-	flags.StringVar(&copts.netMode, "network", "default", "Connect a container to a network")
+	flags.Var(&copts.netMode, "net", "Connect a container to a network")
+	flags.Var(&copts.netMode, "network", "Connect a container to a network")
 	flags.MarkHidden("net")
 	// We allow for both "--net-alias" and "--network-alias", although the latter is the recommended way.
 	flags.Var(&copts.aliases, "net-alias", "Add network-scoped alias for the container")
@@ -613,8 +613,8 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		DNSOptions:     copts.dnsOptions.GetAllOrEmpty(),
 		ExtraHosts:     copts.extraHosts.GetAll(),
 		VolumesFrom:    copts.volumesFrom.GetAll(),
-		NetworkMode:    container.NetworkMode(copts.netMode),
 		IpcMode:        container.IpcMode(copts.ipcMode),
+		NetworkMode:    container.NetworkMode(copts.netMode.NetworkMode()),
 		PidMode:        pidMode,
 		UTSMode:        utsMode,
 		UsernsMode:     usernsMode,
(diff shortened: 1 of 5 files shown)


thaJeztah:
@vdemeester @silvin-lubecki I marked this as "WIP", but perhaps it's already ready for review (possibly tests will have to be added still); I'm on PTO until Monday, but if you think this would be good to have for this release; I can add tests in a follow-up

thaJeztah:
To support multiple networks; this part will have to be removed from the engine, but otherwise (I think) that all works; https://github.com/moby/moby/blob/c7105e3c99286c88855c8d35079f957281dbf581/daemon/create.go#L308-L314

thaJeztah:
🤔 

```
WARNING: deadline exceeded by linter unparam (try increasing --deadline)
cli/command/container/hijack.go:188:1:warning: nolint directive did not match any issue (nolint)
cli/command/trust/signer_remove.go:79:1:warning: nolint directive did not match any issue (nolint)
```

thaJeztah:
ping @vdemeester @silvin-lubecki I think this should be ok now; I added tests.

Will need follow-ups for the completion-scripts (@albers) and for documentation 🤗

cli/command/container/opts.go
```diff
@@ -654,46 +655,115 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
...
+
+// parseNetworkOpts converts --network advanced options to endpoint-specs, and combines
+// them with the old --network-alias and --links. If returns an error if conflicting options
+// are found.
+//
+// this function may return _multiple_ endpoints, which is not currently supported
+// by the daemon, but may be in future; it's up to the daemon to produce an error
+// in case that is not supported.
+func parseNetworkOpts(copts *containerOptions) (map[string]*networktypes.EndpointSettings, error) {
+	var endpoints = make(map[string]*networktypes.EndpointSettings, len(copts.netMode.Value()))
```
silvin-lubecki: nit: use a var block?

thaJeztah: updated; PTAL

opts/network_test.go
```diff
@@ -69,13 +69,15 @@ func TestNetworkOptCompleteSyntax(t *testing.T) {
...
+		t.Run(tc.value, func(t *testing.T) {
+			var network NetworkOpt
+			assert.NilError(t, network.Set(tc.value))
+			assert.Check(t, is.DeepEqual(tc.expected, network.Value()))
+		})
 	}
 }
 
-func TestNetworkOptInvalidSyntax(t *testing.T) {
+func TestNetworkOptdvancedSyntaxInvalid(t *testing.T) {
```
thaJeztah: oh, typo; fixing

cli/command/container/opts.go
```diff
@@ -654,46 +655,116 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
...
+		hasUserDefined, hasNonUserDefined bool
+	)
+
+	for i, n := range copts.netMode.Value() {
+		if container.NetworkMode(n.Target).IsUserDefined() {
+			hasUserDefined = true
+		} else {
+			hasNonUserDefined = true
 		}
+		if i == 0 {
```
silvin-lubecki: nit: could you extract this special block ? Something like:
```go
if i == 0{
    checkFirstNetwork(n, copts)
}
```

thaJeztah: Was actually looking at that yes; let me update that

thaJeztah: updated; added a `applyContainerOptions` function

cli/command/container/opts.go
```diff
@@ -654,46 +655,122 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
...
+			if err := applyContainerOptions(&n, copts); err != nil {
+				return nil, err
+			}
+		}
+		ep, err := parseNetworkAttachmentOpt(n)
+		if err != nil {
+			return nil, err
+		}
+		if _, ok := endpoints[n.Target]; ok {
+			return nil, errdefs.InvalidParameter(errors.Errorf("network %q is specified multiple times", n.Target))
```
cpuguy83, line 704: Any particular reason this was added?
We have a test in integration-cli in moby/moby that tests that this exact case is successful.

cpuguy83: Opened https://github.com/moby/moby/pull/42163 to discuss removing the test case that this validation breaks.

albers:
@thaJeztah The csv syntax is a big problem in bash completion because it does not treat `,` as a word separator.
That's why I did not add this syntax for any of the commands that have it.

david-l-riley:
> To support multiple networks; this part will have to be removed from the engine, but otherwise (I think) that all works; https://github.com/moby/moby/blob/c7105e3c99286c88855c8d35079f957281dbf581/daemon/create.go#L308-L314

Is there a reason this isn't done yet? I'm happy to file a PR to make this, but given how simple it seems I just want to make sure there's not a deeper reason it hasn't been done. Seems like an arbitrary check for something the engine should be able to do just fine? Does an API spec need to be updated?

thaJeztah:
> Is there a reason this isn't done yet? I'm happy to file a PR to make this, but given how simple it seems I just want to make sure there's not a deeper reason it hasn't been done. Seems like an arbitrary check for something the engine should be able to do just fine? Does an API spec need to be updated?

No specific reason, other than that nobody has come round to it yet. When making that change, it should probably be gated by API version (i.e., keep the old behavior for existing API versions), and tests will be needed; the network code is known to be "tricky" at times (lots of corner-cases and due to complexity on setting up the network, there's various potential race-conditions).

That said; at least having a (draft) pull request that removes the check, and verifies it works never hurts.

And once it's in a release, we can possibly update docker-compose as well to consume it (connecting multiple networks instead of connecting them one at a time)

thaJeztah:
100% agreed! If it all works, it's a lot cleaner (create this container attached to these networks, with these options) all in a single API call 🎉

PR #1769, opened 2019-03-21
Fix annnotation on docker config create --template-driver

simonferquel:
**- What I did**
Fixed an annotation that was set on an incorrect flag
**- How I did it**
Spotted it while working on deferred client initialization

cc @thaJeztah


thaJeztah:
introduced in https://github.com/docker/cli/pull/896 (part of 18.03)

PR #1460, opened 2018-10-19
fixes #1459 handle archive with paxHeader or not when docker build

lifubang:
Signed-off-by: Lifubang <lifubang@acmcoder.com>

**- What I did**
If tar file with paxHeader, func IsArchive can't check it correctly.
See https://github.com/moby/moby/issues/38059 #1459
Tar file with paxHeader:
```
pax_global_header

  "presets": ["es2015"]
}

branch = True

source = measurements

[report]
exclude_lines = 
    # pragma: no cover
```

**- How I did it**
Expand archiveHeaderSize and recheck it.

**- How to verify it**
After fix:
```
root@dockerdemo:~/api# git archive --format=tar 210e8603aea57e03e3fc6066d5967c2bf53879a9 | ~/gocode/src/github.com/docker/cli/build/docker build -t test -
IsArchive Error: unexpected EOF
IsArchive Error: <nil>
IsArchive: True
 
Sending build context to Docker daemon   8.52MB
Step 1/14 : FROM python:3.5.2-slim
^C
root@dockerdemo:~/api#
```

**- Description for the changelog**
Support git archive and other tar file with paxHeader

**- A picture of a cute animal**
i67 op146 pnf9u5zhlp6zk

Issue #1459, opened 2018-10-19
Need to handle archive with paxHeader or not when docker build


lifubang:
**Description**
For https://github.com/moby/moby/issues/38059
We need to handle archive with paxHeader or not when docker build
Because git archive has paxHeader:
```
pax_global_header

  "presets": ["es2015"]
}

branch = True

source = measurements

[report]
exclude_lines = 
    # pragma: no cover
```
But regular tar file doesn't have it.

**Describe the results you received:**
See https://github.com/moby/moby/issues/38059

**Describe the results you expected:**
See https://github.com/moby/moby/issues/38059

PR #1770, opened 2019-03-21
Fix archive detection

kolyshkin:
This is a carry-over from #1460 with a simpler fix.

As pointed out in #1459, docker cli fails to detect that the input is a tarball,
in case it is generated by `git archive --format=tgz`.
    
This happens because `git archive` adds some metadata to the initial tar header,
and so it is more than 1 block (of 512 bytes) long, while we only provide 1 block
to archive/tar.Next() and it fails.
    
To fix, give it 2 blocks :)

Test case by @lifubang is added.


cli/command/image/build/context.go
```diff
@@ -89,7 +89,7 @@ func ValidateContextDirectory(srcPath string, excludes []string) error {
 func DetectArchiveReader(input io.ReadCloser) (rc io.ReadCloser, isArchive bool, err error) {
 	buf := bufio.NewReader(input)
 
-	magic, err := buf.Peek(archiveHeaderSize)
+	magic, err := buf.Peek(archiveHeaderSize * 2)
```
tiborvass: i'm surprised this works, because magic should be twice as big, which ends up going to `archive.DetectCompression` which looks for byte equality in the beginning of the magic https://github.com/docker/cli/blob/master/vendor/github.com/docker/docker/pkg/archive/archive.go#L147 Am I missing something ?

tiborvass: Or I guess it would just skip it, because of the different length https://github.com/docker/cli/blob/master/vendor/github.com/docker/docker/pkg/archive/archive.go#L143

tiborvass: Nevermind I thought this was for the compressed codepath, it's not.

PR #1773, opened 2019-03-22
add --from-current flag to context create

zappy-shu:
**- What I did**

Added `--from-current` flag to `context create` command that simplifies creation of a context from the current context, acting as an alias for `--docker from-context=true` and automatically infers whether to create the context with kubernetes as per `--kubernetes from-context=true`

**- How I did it**
Uses the context store to export the current context, then updates the details (description and orchestrator), and uses the context store to import the new context.

**- How to verify it**
Tests have been added to check the new functionality and ensure the user cannot use the `--from-current` flag with other docker/kubernetes settings


simonferquel:
ping @dhiltgen

cli/command/context/create.go
```diff
@@ -139,3 +144,41 @@ func checkContextNameForCreation(s store.Store, name string) error {
 	}
 	return nil
 }
+
+func createFromCurrent(cli command.Cli, s store.Store, stackOrchestrator command.Orchestrator, o *CreateOptions) error {
+	if len(o.Docker) != 0 || len(o.Kubernetes) != 0 {
+		return errors.New("cannot use --docker and --kubernetes flags when --from-current is set")
```
silvin-lubecki: nit: `--docker OR --kubernetes` ?

zappy-shu: makes sense. Pushed the change

cli/command/context/create.go
```diff
@@ -63,6 +64,7 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 		"Default orchestrator for stack operations to use with this context (swarm|kubernetes|all)")
 	flags.StringToStringVar(&opts.Docker, "docker", nil, "set the docker endpoint")
 	flags.StringToStringVar(&opts.Kubernetes, "kubernetes", nil, "set the kubernetes endpoint")
+	flags.BoolVar(&opts.FromCurrent, "from-current", false, "get docker endpoint from current configuration")
```
silvin-lubecki: I think `from-current` flag is misleading, we don't know from what it is: from current context, from current environment? Reading the code it's clear it's from the current context, so I think we should be explicit here and rename it to `from-current-context`.
The description should also be amended with something more explicit like: `create a new context from the current context`

ijc: In context (no pun intended) it'll be `docker context --from-current`. With the proposed change it'll be `docker context --from-current-context` which seems stuttery to me. Is the current naming not clear enough when you see it used in context?

When rewording the help we should be careful not to obscure too much the fact that if your current context is the automatic one derived from envvars then that is what you will be importing -- that's the 80% use case for this flag I think (otherwise we'd call this `context copy --from=xxx` or `context create --from=xxx` etc I think).

silvin-lubecki: We already have a "flag" (to be fair it's a CSV key) which is `from-current`, but creates a context with docker configuration from the current `environment`. Same for `kubernetes`.
That's a lot of `from-current` here, with different semantic. That's why I think we should be explicit here.

silvin-lubecki: After off-line discussion with @ijc, I agree with `docker context create --from=my-current-context` 👍 
@thaJeztah any thought about that?

simonferquel: `from-current` either as a flag or as the csv config key, always refers to the current context (being a real one, or `default`. It just applies at a different scope (e.g. either copy both, or only one of the endpoints as is).
I agree that naming might be confusing though. We should make sure it is clean before 19.03 is out, but I think we should keep both the flag and the setting names consistent.

simonferquel: I think I have an even better UI:
- `docker context create my-ctx`: with no value for --from, --docker --kubernetes -> copy the current context
- `docker context create my-ctx --from=<some context>` copy an existing context (some context can be `default`)
- `docker context create my-ctx --docker "..."` create a custom context from scratch.

Also in addition, replace from-current setting name with from=<existing context or default> in docker and kubernetes create options. wdyt ?

silvin-lubecki: I'm good with this 👍

ijc: My question would be what is the state of `--kubernetes` in the final case -- is it blank or is it inherited from the current context. IOW does passing any single option completely disable the copying behaviour of the first case for all options?

zappy-shu: maybe if `--docker` is not set and `--kubernetes` is then it acts as an alias for `--docker from-current=true --kubernetes blah=blah`

simonferquel: @ijc I guess it is a matter of opinion. My guess is if you want a custom endpoint configuration, you are in fully customized way, but that does not disallow things like:
`docker context create my-ctx --docker host=tcp://42.42.42.42:42 --kubernetes from=default ...`

ijc: I wasn't expressing (and don't really have) an opinion, but we should make a decision for one or the other.

silvin-lubecki:
@zappy-shu there is a linter error:
```
cli/command/context/create.go:72::warning: cyclomatic complexity 19 of function RunCreate() is high (> 16) (gocyclo)
```

simonferquel:
Reference documentation must be updated then :)
Can you deal with that @zappy-shu ? (I can give you pointers tomorrow on where to look at)

simonferquel:
For symmetry, we need to remove `from-current` config key in docker and kubernetes endpoints, with a `from` config key accepting a context name (which can be `default`)

thaJeztah:
> For symmetry, we need to remove `from-current` config key in docker and kubernetes endpoints, with a from config key accepting a context name (which can be `default`)

I agree, I think the functionality overlaps


Would this be a common scenario? As this is already possible (and would be symmetrical with, e.g. `docker image save hello-world | docker image load`)

```bash
docker context export default - | docker context import mynewcontext -
```

thaJeztah:
(Was also thinking of a `docker context copy` subcommand)

simonferquel:
@thaJeztah it is actually how --from is implemented. It allows to override original description / default stack orchestrator though.
Export/Import + Update would probably do the same, but the idea here is to polish the UI for common context creation tasks.
Our nominal scenario is when working with a UCP bundle:
We can easily do something like 
```
$ eval "$(<env.sh)" # or env.ps1
$ docker context create ucp
```
and be done with it.

docs/reference/commandline/context_create.md
```diff
@@ -72,4 +73,14 @@ $ export KUBECONFIG=/path/to/my/kubeconfig
 $ docker context create my-context --kubernetes "from-current=true" --docker "host=/var/run/docker.sock"
 ```
 
+By default the `context` will be created from the current context. This will include both docker and, if set, kubernetes endpoint configuration. This is equivalent to `--docker "from-current=true"` or, if the current context has a kubernetes contfiguration, `--docker "from-current=true" --kubernetes "from-current=true"`:
+```bash
+$ docker context create my-context
+```
+
+You can select a different `context` to create from by using the `--from` option:
```
thaJeztah:
```suggestion
Use the `--from=<context name>` option to create a new context from an
existing context. The example below creates a new context named
`my-context` from the existing `existing-context` context.
```

thaJeztah:
> Our nominal scenario is when working with a UCP bundle

That feels like a band-aid solution; can we include an actual _context_ in the UCP bundle, so that we can `docker context import ./ucp-context.context` ?

simonferquel:
@thaJeztah we will in the next release. But users that don't upgrade their clusters - and whose bundles don't have the dockercontext file in it -  should have something better that what we have now.

thaJeztah:
> But users that don't upgrade their clusters - and whose bundles don't have the dockercontext file in it - should have something better that what we have now.

Well, if they don't update they would still use an old client 😉, but I get the idea; as long as we make sure to document the `docker context import` as the recommended/standard approach

thaJeztah:
Also wondering if we can backport the "include context in UCP bundle" to older UCP releases (as part of a patch release)

simonferquel:
Not likely because of golang vendoring hell (vendoring a new CLI requires vendoring a new docker engine... which is not possible for patch releases in UCP - except bumping a patch release of the currently supported engine)

cli/command/context/create.go
```diff
@@ -139,3 +144,52 @@ func checkContextNameForCreation(s store.Store, name string) error {
 	}
 	return nil
 }
+
+func createFromExistingContext(s store.Store, fromContextName string, stackOrchestrator command.Orchestrator, o *CreateOptions) error {
+	if len(o.Docker) != 0 || len(o.Kubernetes) != 0 {
+		return errors.New("cannot use --docker or --kubernetes flags when --from is set")
```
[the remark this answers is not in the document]
zappy-shu: Being able to set the description makes sense to me as it is a new context. The default stack orchestrator I don't know much about. From what I understand the main purpose of `--from` is to make it easier to copy the connection details for docker/kubernetes from a context or environment variables

cli/command/context/create.go
```diff
@@ -63,6 +64,7 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 		"Default orchestrator for stack operations to use with this context (swarm|kubernetes|all)")
 	flags.StringToStringVar(&opts.Docker, "docker", nil, "set the docker endpoint")
 	flags.StringToStringVar(&opts.Kubernetes, "kubernetes", nil, "set the kubernetes endpoint")
+	flags.StringVar(&opts.From, "from", "", "create context from a named context")
```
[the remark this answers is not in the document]
zappy-shu: The default depends on whether you use `--docker`/`--kubernetes` or not. With then it will use the options given to those arguments, without it will create from the current context. Seems a bit finicky to me I'll admit.
To make it less so we could remove the defaulting entirely but then creating from the current/environment is a bit more of a pain.

docs/reference/commandline/context_create.md
```diff
@@ -23,7 +23,7 @@ Create a context
 Docker endpoint config:
 
 NAME                DESCRIPTION
-from-current        Copy current Docker endpoint configuration
+from                Copy Docker endpoint configuration from an existing context
```
thaJeztah: Do we still want the `from` for `docker` and `kubernetes`, now that we have the top-level `--from`? Basically; wondering if we want `context create` and `context update` to allow mish-mashing two unrelated contexts.

I can _slightly_ see this being something when _creating_ a new context, but not entirely sure for _updating_ an existing one.

simonferquel: I think it is not about "usefulness" but more about symetry: endpoint configuration options would be easier to learn if they are the same for both create and update.

PR #1757, opened 2019-03-19
Update compose-on-kubernetes v0.4.20

thaJeztah:
ping @simonferquel @silvin-lubecki PTAL

PR #1774, opened 2019-03-25
bump kubernetes to v1.14.0

zappy-shu:
**- What I did**
Updates kubernetes from v1.11.2 to v1.14.0

**- How I did it**
Updated vendor.conf and vendors

**- How to verify it**
Check resulting kubernetes version


vendor.conf
```diff
@@ -93,9 +92,12 @@ google.golang.org/grpc v1.12.0
...
-k8s.io/kubernetes v1.11.2
+k8s.io/api kubernetes-1.13.4
+k8s.io/apimachinery kubernetes-1.13.4
+k8s.io/client-go kubernetes-1.13.4
+k8s.io/kube-openapi c59034cc13d587f5ef4e85ca0ade0c1866ae8e1d
+k8s.io/kubernetes v1.13.4
 vbom.ml/util 256737ac55c46798123f754ab7d2c784e2c71783
+sigs.k8s.io/yaml v1.1.0
+k8s.io/klog v0.2.0
+
```
silvin-lubecki: totally nit: empty line here

silvin-lubecki:
PTAL @thaJeztah

vendor.conf
```diff
@@ -14,7 +14,7 @@ github.com/davecgh/go-spew 346938d642f2ec3594ed81d874461961cd0faa76 # v1.1.0
 github.com/dgrijalva/jwt-go a2c85815a77d0f951e33ba4db5ae93629a1530af
 github.com/docker/distribution 0d3efadf0154c2b8a4e7b6621fff9809655cc580
 github.com/docker/docker 827cb09f87964ed38b46502f22a585f2ed4a78e1
-github.com/docker/compose-on-kubernetes 356b2919c496f7e988f6e0dfe7e67d919602e14e # master w/ v1alpha3+pullsecrets+pull-policy
+github.com/docker/compose-on-kubernetes e1f7354ed0de256b6ebf5bfa5a90a76460036c52
```
thaJeztah: Should we align this with a tagged version? See the discussion on https://github.com/docker/cli/pull/1757#issuecomment-476229818

simonferquel:
I guess we should wait for https://github.com/docker/compose-on-kubernetes/pull/95 (v1.14 ga) to merge and the 0.4.21 tag and then bump everything to 1.14. @thaJeztah do you agree ?

cli/command/stack/kubernetes/watcher.go
```diff
@@ -220,13 +220,11 @@ func newPodInformer(podsClient podListWatch, stackName string, indexers cache.In
 		&cache.ListWatch{
 			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
 				options.LabelSelector = labels.SelectorForStack(stackName)
-				options.IncludeUninitialized = true
```
thaJeztah: Are these changes needed to make everything build? If so, perhaps better to include it in the same commit as the actual bump

vendor.conf
```diff
@@ -26,8 +26,8 @@ github.com/docker/go-units 47565b4f722fb6ceae66b95f853feed578a4a51c # v0.3.3
 github.com/docker/libtrust 9cbd2a1374f46905c68a4eb3694a130610adc62a
 github.com/docker/licensing 9781369abdb5281cdc07a2a446c6df01347ec793
 github.com/docker/swarmkit 415dc72789e2b733ea884f09188c286ca187d8ec
+github.com/evanphx/json-patch 5858425f75500d40c52783dce87d085a483ce135
```
thaJeztah: I see this package provides tagged releases; do we need the changes that are in master and not yet in their latest release (`v4.1.0`)? https://github.com/evanphx/json-patch/compare/v4.1.0...5858425f75500d40c52783dce87d085a483ce135 

Otherwise, I'd prefer pinning to `v4.1.0` (or ask them to tag a new release)


FWIW this is licensed [BSD3-clause](https://github.com/evanphx/json-patch/blob/master/LICENSE) so we may have to change packaging to show the license? @seemethere @justincormack

zappy-shu: @thaJeztah the revision for json-patch can from compose-on-kubernetes [gopkg.toml](https://github.com/docker/compose-on-kubernetes/blob/master/Gopkg.toml). I've checked vendoring with v4.1.0 here doesn't break any of the tests. Do you want me to make this change or stick with the revision to match compose-on-kubernetes?

thaJeztah: If it's ok with you, I'd prefer using a version that matches a tagged release for dependencies that use tags; doing so provides a bit more "context" (also preparing/taking into account users that start using `go mod` for vendoring)

zappy-shu: sure, done

cli/command/stack/kubernetes/watcher.go
zappy-shu: rebased and amended

thaJeztah:
Linking this to https://github.com/docker/cli/pull/1784 - bug (but server-side) in Kubernetes 1.14 / Golang 1.12. Probably not a blocker for this PR given that it's server-side, but thought it'd be relevant to mention

simonferquel:
@thaJeztah I confirm that once the bug is fixed in kube 1.14.1, #1784 is not needed anymore.
I think we can merge #1774, and I'll split #1784 into
- A PR that fix the selector issue (which will be mergeable as is)
- A PR with the funky client side workaround, that we will only merge if Amberjack ship with 1.14.0.

Issue #1492, opened 2018-10-31
initial windows size of docker exec -it is too small


lifubang:
**Description**
initial windows size of `docker exec -it` is too small, I need to resize the window.
This is too inconvenient for me.

**Steps to reproduce the issue:**
1. docker exec -it k8s_nginxgateway_apigateway-pxwj9_default_430f8bb0-dcb3-11e8-a770-00163e0aec56_0 /bin/bash
2. vim /usr/local/nginx/conf/nginx.conf

**Describe the results you received:**
![t0gq 6tu a s6 32wesg](https://user-images.githubusercontent.com/783424/47763588-460e1f00-dcfc-11e8-8cb4-52b11596f03d.png)

**Describe the results you expected:**
the vim window size should be equal to the window's size.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
Client:
 Version:	18.03.0-ce
 API version:	1.37
 Go version:	go1.9.4
 Git commit:	0520e24
 Built:	Wed Mar 21 23:10:01 2018
 OS/Arch:	linux/amd64
 Experimental:	false
 Orchestrator:	swarm

Server:
 Engine:
  Version:	18.03.0-ce
  API version:	1.37 (minimum version 1.12)
  Go version:	go1.9.4
  Git commit:	0520e24
  Built:	Wed Mar 21 23:08:31 2018
  OS/Arch:	linux/amd64
  Experimental:	false
```

**Output of `docker info`:**

```
Containers: 25
 Running: 24
 Paused: 0
 Stopped: 1
Images: 75
Server Version: 18.03.0-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: cfd04396dc68220d1cecbe686a6cc3aa5ce3667c
runc version: 4fc53a81fb7c994640722ac585fa9ca548971871
init version: 949e6fa
Security Options:
 apparmor
 seccomp
  Profile: default
Kernel Version: 4.4.0-105-generic
Operating System: Ubuntu 16.04.3 LTS
OSType: linux
Architecture: x86_64
[...]

PR #1529, opened 2018-11-21
fixes 1492: tty initial size error

lifubang:
Signed-off-by: Lifubang <lifubang@acmcoder.com>

fixes #1492.

For Ubuntu 16.04.2 LTS (GNU/Linux 4.4.0-62-generic x86_64), Docker version 17.12.0-ce, build c97c6d6 / 18.03.0-ce, build 0520e24
When I run: `docker exec -it ubuntu /bin/bash`
The tty's size is small than window's size.
```
root@iZ2zeesy3n96esegm7tue0Z:~# docker exec -it ubuntu /bin/bash
root@97b3e229a701:/# ls -a
.   .dockerenv  bin   dev  home  lib64  mnt  proc  run   srv  tmp  var
..  .r          boot  etc  lib   media  opt  root  sbin  sys  usr
```
After I fix the bug:
```
root@iZ2zeesy3n96esegm7tue0Z:~# ./docker-linux-amd64 exec -it ubuntu /bin/bash
root@97b3e229a701:/# ls -a
.  ..  .dockerenv  .r  bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
```
If not fix this bug, it's too inconvenient for me to use docker exec -it.

**- What I did**
1. Fix tty initial size error
I think it's a bug of dockerd, but for previous dockerd edition, I think we can fix it in cli.
2. Fix when we use --debug flag, the first shell prompt don't return to the head of line when we got an error in func resizeTtyTo!

**- How I did it**
~~func MonitorTtySize in file cli/command/container/tty.go~~
~~I move the call resizeTty() to last and run it asynchronously.~~
1. Because in sometimes when dockercli call tty resize, the exec have not finished yet. Then we got a message "Error response from daemon: no such exec", but we ignore it, so we need to throw this error and deal with it after 10 milliseconds.
2. Add `\r` when logrus.Debugf.

**- How to verify it**
It's convenient for me to use docker exec -it.
I don't need to resize the window.

**- Description for the changelog**
fixes tty initial size error

**- I have no more cute animal now**
**Before fix:**
vim /etc/mime.types
sk j _1irw za2kc78iody5

**After fix:**
vim /etc/mime.types
x z0 r5jg_i4h x6m2bny 4

PR #1377, opened 2018-09-18
add a unit test to cover container export command output file option

adshmh:
Adding a unit test to cover the output file option of the `export` command. If merged, this allows the removal of the redundant integration test `TestExportContainerWithOutputAndImportImage` from `moby`, which is the same as `TestExportContainerAndImportImage`, except for the output file option:

https://github.com/moby/moby/blob/d6a7c22f7b9aa028623ef8c7cb31d1bb04dd9d84/integration-cli/docker_cli_export_import_test.go#L12-L17

Signed-off-by: Arash Deshmeh <adeshmeh@ca.ibm.com>

**- What I did**
Added a unit test to cover the output file option of the `export` command

**- How I did it**
Added new test file `cli/command/container/export_test.go`

**- How to verify it**

**- Description for the changelog**
Added unit test for output to file option of the export command.

**- A picture of a cute animal (not mandatory but encouraged)**

PR #1775, opened 2019-03-26
[18.09 backport] fixes 1492: tty initial size error

thaJeztah:
backport of https://github.com/docker/cli/pull/1377 and https://github.com/docker/cli/pull/1529 for 18.09

fixes https://github.com/docker/cli/issues/1492

For Ubuntu 16.04.2 LTS (GNU/Linux 4.4.0-62-generic x86_64), Docker version 17.12.0-ce, build c97c6d6 / 18.03.0-ce, build 0520e24
When I run: `docker exec -it ubuntu /bin/bash`
The tty's size is small than window's size.
```
root@iZ2zeesy3n96esegm7tue0Z:~# docker exec -it ubuntu /bin/bash
root@97b3e229a701:/# ls -a
.   .dockerenv  bin   dev  home  lib64  mnt  proc  run   srv  tmp  var
..  .r          boot  etc  lib   media  opt  root  sbin  sys  usr
```
After I fix the bug:
```
root@iZ2zeesy3n96esegm7tue0Z:~# ./docker-linux-amd64 exec -it ubuntu /bin/bash
root@97b3e229a701:/# ls -a
.  ..  .dockerenv  .r  bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
```
If not fix this bug, it's too inconvenient for me to use docker exec -it.

**- What I did**
1. Fix tty initial size error
I think it's a bug of dockerd, but for previous dockerd edition, I think we can fix it in cli.
2. Fix when we use --debug flag, the first shell prompt don't return to the head of line when we got an error in func resizeTtyTo!

**- How I did it**
~~func MonitorTtySize in file cli/command/container/tty.go~~
~~I move the call resizeTty() to last and run it asynchronously.~~
1. Because in sometimes when dockercli call tty resize, the exec have not finished yet. Then we got a message "Error response from daemon: no such exec", but we ignore it, so we need to throw this error and deal with it after 10 milliseconds.
2. Add `\r` when logrus.Debugf.

**- How to verify it**
It's convenient for me to use docker exec -it.
I don't need to resize the window.

fixes tty initial size error

**- I have no more cute animal now**
**Before fix:**
vim /etc/mime.types
sk j _1irw za2kc78iody5

**After fix:**
vim /etc/mime.types
x z0 r5jg_i4h x6m2bny 4

### DIFF
--- a/cli/command/container/client_test.go
+++ b/cli/command/container/client_test.go
@@ -24,6 +24,7 @@ type fakeClient struct {
 	logFunc               func(string, types.ContainerLogsOptions) (io.ReadCloser, error)
 	waitFunc              func(string) (<-chan container.ContainerWaitOKBody, <-chan error)
 	containerListFunc     func(types.ContainerListOptions) ([]types.Container, error)
+	containerExportFunc   func(string) (io.ReadCloser, error)
 	Version               string
 }
 
@@ -123,4 +124,11 @@ func (f *fakeClient) ContainerStart(_ context.Context, container string, options
 		return f.containerStartFunc(container, options)
 	}
 	return nil
+}
+
+func (f *fakeClient) ContainerExport(_ context.Context, container string) (io.ReadCloser, error) {
+	if f.containerExportFunc != nil {
+		return f.containerExportFunc(container)
+	}
+	return nil, nil
 }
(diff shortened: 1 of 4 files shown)


thaJeztah:
ping @silvin-lubecki @kolyshkin ptal

PR #1656, opened 2019-02-02
CLI: Add support for using Configs as CredentialSpecs in services

thaJeztah:
depends on ~https://github.com/moby/moby/pull/38672~ https://github.com/moby/moby/pull/38632 (using a temporary vendoring), ~and built on top of https://github.com/docker/cli/pull/1655 (which isn't merged yet)~

- Improve validation of credential_spec in compose schema (I'll split that to a separate PR as well)
- Add compose schema 3.8
- Add `config` as option in credential_spec

PR #1781, opened 2019-03-27
Support using swarm Configs as CredentialSpecs in Services.

dperny:
closes https://github.com/docker/cli/pull/1656

**- What I did**
Builds on and obsoletes #1656. Uses the final API that went into the engine.

Additionally, in the process, discovered and fixed a bug where the `--credential-spec` flag would not have been respected on service updates.

**- How I did it**

When handling configs, add additional logic for handling runtime configs. Additionally, adds logic to replace the config name which the user passes with the config ID, which is what the engine API needs.

**- How to verify it**

Includes unit tests.

Add ability to use swarm Configs as CredentialSpecs on Services.


dperny:
Added tests, removed WIP.

dperny:
Refactored to reduce cyclomatic complexity, which should hopefully be enough to pass lint.

dperny:
Added compose support.

cli/compose/convert/service.go
```diff
@@ -357,6 +367,28 @@ func convertFileObject(
 		source = obj.Name
 	}
 
+	// if the config is a Runtime config, that means we don't mount it as a
+	// file in the container. This is used for supporting CredentialSpec
+	// configs.
+	if config.Runtime {
+		if config.Target != "" || config.UID != "" || config.GID != "" || config.Mode != nil {
```
thaJeztah: Wouldn't target and uid/gid still be configurable? (i.e., mount a config as credential-spec, but specify which uid/gid and/or path to use?

cli/compose/types/types.go
```diff
@@ -408,6 +408,11 @@ type FileReferenceConfig struct {
 	UID    string  `yaml:",omitempty" json:"uid,omitempty"`
 	GID    string  `yaml:",omitempty" json:"gid,omitempty"`
 	Mode   *uint32 `yaml:",omitempty" json:"mode,omitempty"`
+	// Runtime is set `true` if reference isn't actually used as a File in
+	// the service. Specifically, this currently only indicates that the Config
+	// is used as a CredentialSpec. If this field is set `true`, then all other
+	// fields besides `Source` must be empty.
+	Runtime bool `yaml:",omitempty" json:"runtime,omitempty"`
```
thaJeztah: Can we deduct this from the `Source` value? (IIUC, it will be prefixed with `config://` ?)

cli/compose/convert/service.go
dperny: No, because a credentialspec is never mounted into the container, or written to the filesystem. And in any case, the actual API type (which is a `ConfigReference` with a `ConfigReferenceRuntimeTarget`) does not allow setting file attributes.

cli/compose/types/types.go
dperny: No. The string `config://` should never actually be in the compose code, I think.

We could theoretically determine that a config is a `Runtime` config by looking for the absence of all other fields, but I usually think it's better to affirmatively determine something rather than make assumptions from the absence.

It would be cleaner to have a whole new type besides `FileReferenceConfig` but doing so would be more trouble than it's worth, IMHO.

cli/compose/convert/service.go
thaJeztah: 🤦‍♂️ ah, yes, makes sense

andrewhsu:
I had a quick look, but I'm not as familiar with this codebase. @cpuguy83 is this patch also something you could have a look at if you have time?

cli/command/service/update.go
```diff
@@ -754,6 +795,42 @@ func getUpdatedConfigs(apiClient client.ConfigAPIClient, flags *pflag.FlagSet, c
...
+// removeConfigs figures out which configs in the existing spec should be kept
+// after the update.
+func removeConfigs(flags *pflag.FlagSet, spec *swarm.ContainerSpec, credSpecName, credSpecID string) []*swarm.ConfigReference {
+	keepConfigs := []*swarm.ConfigReference{}
+
+	toRemove := buildToRemoveSet(flags, flagConfigRemove)
+	// all configs in spec.Configs should have both a Name and ID, because
+	// they come from an already-accepted spec.
+	for _, config := range spec.Configs {
+		// if the config is a Runtime target, make sure it's still in use right
```
cpuguy83: Is this a bug waiting to happen? Is there something we can do now to prevent someone from overlooking this in the future?

cli/compose/convert/service.go
```diff
@@ -308,11 +310,24 @@ func convertServiceConfigObjs(
 			return nil, err
 		}
 
-		file := swarm.ConfigReferenceFileTarget(obj.File)
-		refs = append(refs, &swarm.ConfigReference{
-			File:       &file,
-			ConfigName: obj.Name,
-		})
+		// if the obj.File is identical to the zero-value of
```
cpuguy83: As above, is this a bug waiting to happen?

cli/command/service/create.go
```diff
@@ -141,3 +137,42 @@ func runCreate(dockerCli command.Cli, flags *pflag.FlagSet, opts *serviceOptions
...
+		// field on the spec
+		//
+		// we check the opts instead of the service directly because there are
+		// a few layers of nullable objects in the service, which is a PITA
+		// to traverse, but the existence of the option implies that those are
+		// non-null.
+		if cs := opts.credentialSpec.Value(); cs != nil && cs.Config != "" {
+			for _, config := range configs {
+				if config.ConfigName == cs.Config {
+					service.TaskTemplate.ContainerSpec.Privileges.CredentialSpec.Config = config.ConfigID
```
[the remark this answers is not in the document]
dperny: short circuits evaluation but not required for correctness. will add one anyway though.

cli/command/service/update.go
dperny: sort of. any future use for runtime specs will probably require similar alterations to the configs logic, which means the implementer will have to look here anyway. then, they'll see this comment, and add the correct removal code.

cli/compose/convert/service.go
dperny: Possibly, but the thing is I'm not sure under what case we'll add another non-file ConfigReference type, so I can't really code defensively against it, and even if I do, it's unlikely that something else will even be implemented at any point.

cpuguy83: Off the top of my head, a value explicitly for runtime targets.

dperny: Yes, it would make sense lol

I've made this change, and added tests to boot.

cli/compose/types/types.go
cpuguy83: Would it make sense to use a string instead of a bool?
I'd like to avoid a situation where we are adding multiple bools here.

dperny: A string representing what? The target, file vs runtime?

thaJeztah: One thing to take note is that this type (`FileReferenceConfig`) is used both for configs and secrets, but we'll only use the "runtime" option for configs; so we might have to create a separate type for configs?

Thinking (UX-wise); in order to use a config as credential-spec, I would have to;

1. define the Config

      ```yaml
      version: "3.8"

      configs:
        my_credentials_spec:
          file: ./credentials-spec-file.json
      ```

2. add the Config to the Configs section of the compose file (I guess this can _only_ be with [the long syntax](https://docs.docker.com/compose/compose-file/#long-syntax), because I'd have to specify the `runtime` option);

      ```yaml
      version: "3.8"
      services:
        myservice:
          image: myimage:latest
          configs:
            - source: my_credentials_spec
              runtime: true

      configs:
        my_credentials_spec:
          file: ./credentials-spec-file.json
      ```

3. configure the service to use the config as source for the credentials-spec;

      ```yaml
      version: "3.8"
      services:
        myservice:
          image: myimage:latest
          credential_spec:
            config: my_credentials_spec
          configs:
            - source: my_credentials_spec
              runtime: true

      configs:
        my_credentials_spec:
          file: ./my-credential-spec.json
      ```

That's a lot of duplication; can we somehow skip `step 2.` ? i.e., if `credential_spec` uses `config: <some name>`, we already know that we want to use a config as source. That config must be defined in the compose-file, but we could attach it automatically to the service (if needed) (or can we do that step even "behind the scenes", daemon-side?)

dperny: Sure, why not.

dperny:
Altered the compose code to infer the need for a Config by the existence of a value in the CredentialSpec.Config field, meaning there is no need to define a `Runtime` config in the service's configs; merely specifying a CredentialSpec with a Config is sufficient.

This means a service using a CredentialSpec Config would look like this:

```yaml
version: "3.8"
services:
  myservice:
    image: myimage:latest
    credential_spec:
      config: my_credentials_spec

configs:
  my_credentials_spec:
    file: ./my-credential-spec.json|
```

whereas before this change, it would have looked like this:

```yaml
version: "3.8"
services:
  myservice:
    image: myimage:latest
    credential_spec:
      config: my_credentials_spec
    configs:
      - source: my_credentials_spec
        runtime: true

configs:
  my_credentials_spec:
    file: ./my-credential-spec.json
```

thaJeztah:
Thanks! That looks a lot more convenient (and less confusing) 🤗 
Also sorry for the long delay between review/comments. Lots of plates to keep spinning 😓 





I'll have another look at the code shortly @cpuguy83 @vdemeester if you have some time 🙏

cli/command/service/opts.go
```diff
@@ -331,12 +331,25 @@ func (c *credentialSpecOpt) Set(value string) error {
...
+		// Therefore, this isn't the definitive location for the value of
+		// Config that is passed to the API.
+		c.value.Config = strings.TrimPrefix(value, "config://")
 	case strings.HasPrefix(value, "file://"):
 		c.value.File = strings.TrimPrefix(value, "file://")
 	case strings.HasPrefix(value, "registry://"):
 		c.value.Registry = strings.TrimPrefix(value, "registry://")
+	case value == "":
+		// if the value of the flag is an empty string, that means there is no
+		// CredentialSpec needed. This is useful for removing a CredentialSpec
```
thaJeztah: Would the whole `credential_spec` not be "not present" in the update case? (wondering)

cli/compose/convert/service.go
```diff
@@ -599,13 +639,46 @@ func convertDNSConfig(DNS []string, DNSSearch []string) (*swarm.DNSConfig, error
...
+	case l == 2:
+		return nil, errors.Errorf("invalid credential spec: cannot specify both %s and %s", o[0], o[1])
+	case l > 2:
+		return nil, errors.Errorf("invalid credential spec: cannot specify both %s, and %s", strings.Join(o[:l-1], ", "), o[l-1])
 	}
 	swarmCredSpec := swarm.CredentialSpec(spec)
+	// if we're using a swarm Config for the credential spec, over-write it
+	// here with the config ID
+	if swarmCredSpec.Config != "" {
+		for _, config := range refs {
```
thaJeztah: Was wondering if we should have a `resolveConfig` / `resolveConfigID` function somewhere, but (I think) all the lookup loops are just _slightly_ different

cli/command/service/opts.go
thaJeztah: Discussing with @dperny - this is for the flag, not the compose-file, so it would be an empty string (not "not present")

cli/compose/schema/schema_test.go
```diff
@@ -104,7 +104,7 @@ func TestValidateCredentialSpecs(t *testing.T) {
 					"foo": dict{
 						"image": "busybox",
 						"credential_spec": dict{
-							tc.expectedErr: "foobar",
+							"config": "foobar",
```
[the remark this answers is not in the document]
thaJeztah: Checked this, and that would require some changes to the test, that are not worth it

dperny:
i went and googled "cat in a pope hat" to give this extra powerful blessings

image

Issue #1782, opened 2019-03-28
No "driver" field for secrets in stack deploy


sirlatrom:
**Description**

`docker stack deploy` does not support setting the driver for a secret. This has been supported in the Engine API and on the command-line since 0ee9e05d8f51169b94708711b9d2e5473a777927 (July 30 2017).

**Steps to reproduce the issue:**
1. Install a secretprovider driver plugin:
```
docker plugin install sirlatrom/docker-secretprovider-plugin-vault
```
2. Create a minimal docker-compose.yml file:
```
version: "3.7"
services:
  service:
    image: busybox
    command: tail -f /dev/null
    secrets:
      - secret
secrets:
  secret:
    driver: sirlatrom/docker-secretprovider-plugin-vault
```
3. Run `docker stack deploy --compose-file docker-compose.yml repro`

**Describe the results you received:**
```
driver Additional property driver is not allowed
```

**Describe the results you expected:**
```
Creating network repro_default                                                               
Creating service repro_service
```

**Additional information you deem important (e.g. issue happens only occasionally):**
This would be *awesome* to have in Docker Compose format 3.8 and Engine 19.03.

**Output of `docker version`:**

```
Client:
 Version:           18.09.4-rc1
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        9d6081e
 Built:             Thu Mar 21 03:16:00 2019
 OS/Arch:           linux/amd64
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          0.0.0-20190325050551-9374b8d
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.11.5
  Git commit:       9374b8d
  Built:            Mon Mar 25 06:53:44 2019
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.2.5
  GitCommit:        bb71b10fd8f58240ca47fbb579b9d1028eea7c84
 runc:
  Version:          1.0.0-rc6+dev
  GitCommit:        96ec2177ae841256168fcf76954f7177af9446eb
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Output of `docker info`:**

```
Containers: 23
 Running: 4
 Paused: 0
 Stopped: 19
Images: 261
Server Version: 0.0.0-20190325050551-9374b8d
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
[...]

PR #1746, opened 2019-03-18
[carry] Add support for `template_driver` in composefiles

thaJeztah:
Carry of https://github.com/docker/cli/pull/1100 - rebased, and moved the option to schema 3.8
closes https://github.com/docker/cli/pull/1100

This maps the `--template-driver` flag on secret and config creation.

PR #1783, opened 2019-03-28
Add driver field to top-level secret object

sirlatrom:
**- What I did**
Added support for specifying a driver for a secret in a compose file when performing a `docker stack deploy`.

**- How I did it**
Added a `driver` field to the `FileObjectConfig` in cli/compose/types/types.go.

**- How to verify it**
`make -f docker.Makefile binary` and run the repro case in #1782.

Added support for secret drivers in docker stack deploy

image


sirlatrom:
I see a similar miniscule reduction in code coverage in the analogous PR #1746, so I hope that is not going to block this PR from being accepted.

cli/command/secret/create.go
```diff
@@ -45,7 +45,7 @@ func newSecretCreateCommand(dockerCli command.Cli) *cobra.Command {
 	flags.StringVarP(&options.driver, "driver", "d", "", "Secret driver")
 	flags.SetAnnotation("driver", "version", []string{"1.31"})
 	flags.StringVar(&options.templateDriver, "template-driver", "", "Template driver")
-	flags.SetAnnotation("driver", "version", []string{"1.37"})
+	flags.SetAnnotation("template-driver", "version", []string{"1.37"})
```
thaJeztah: ahum (introduced in https://github.com/docker/cli/pull/896) 😊; looks like we fixed this for `config` but not `secrets` https://github.com/docker/cli/pull/1769

Could you open a separate PR for this change so that we can cherry-pick this

cli/compose/schema/data/config_schema_v3.8.json
```diff
@@ -538,6 +538,7 @@
           }
         },
         "labels": {"$ref": "#/definitions/list_or_dict"},
+        "driver": {"type": "string"},
```
thaJeztah: could you squash this commit together with the one that re-generates the bindata?

cli/compose/loader/loader.go
```diff
@@ -651,6 +651,10 @@ func loadFileObjectConfig(name string, objType string, obj types.FileObjectConfi
 			}
 		}
 		// if not "external: true"
+	} else if obj.Driver != "" {
```
thaJeztah: Not necessarily for this PR, but I think it may be cleaner if we start using a `switch` here instead of `if/elseif/else`;

```go
	switch {
	case obj.External.External:
		// handle deprecated external.name
		if obj.External.Name != "" {
			if obj.Name != "" {
				return obj, errors.Errorf("%[1]s %[2]s: %[1]s.external.name and %[1]s.name conflict; only use %[1]s.name", objType, name)
			}
			if versions.GreaterThanOrEqualTo(details.Version, "3.5") {
				logrus.Warnf("%[1]s %[2]s: %[1]s.external.name is deprecated in favor of %[1]s.name", objType, name)
			}
			obj.Name = obj.External.Name
			obj.External.Name = ""
		} else {
			if obj.Name == "" {
				obj.Name = name
			}
		}
	case obj.Driver != "":
		if obj.File != "" {
			return obj, errors.Errorf("%[1]s %[2]s: %[1]s.driver and %[1]s.file conflict; only use %[1]s.driver", objType, name)
		}
	default:
		obj.File = absPath(details.WorkingDir, obj.File)
	}
```

cli/compose/schema/data/config_schema_v3.8.json
```diff
@@ -539,6 +539,12 @@
         },
         "labels": {"$ref": "#/definitions/list_or_dict"},
         "driver": {"type": "string"},
+        "driver_opts": {
```
thaJeztah: can you squash this with the other schema-changes? or (perhaps cleaner) do a re-generate of the bindata in this commit as well?

cli/command/secret/create.go
sirlatrom: Sure, and sorry that slipped in, I will check my diff better next time :)

thaJeztah: no, it's good that you found it (thanks!) just if it's a separate PR, I can cherry-pick if 😄

sirlatrom: Thanks for spotting that; I opened #1785.

cli/compose/schema/data/config_schema_v3.8.json
sirlatrom: I've now split this PR in 3 commits:
1. Modifying the schema itself and updating bindata
2. Making use of the changes to the schema in the relevant Go code
3. Adding a test case

cli/compose/loader/loader.go
sirlatrom: I included this change in the new squashed/rebased commits, thanks for suggesting it; it *is* indeed easier to read.

sirlatrom:
@thaJeztah I have rebased and squashed as requested, PTAL.

thaJeztah:
ping @silvin-lubecki @vdemeester perhaps you could have a look as well? 🤗

thaJeztah:
note this will conflict with https://github.com/docker/cli/pull/1781 (which is targeted for 19.03), this will have to be rebased, and the bindata will have to be re-generated after that is merged

sirlatrom:
> note this will conflict with #1781 (which is targeted for 19.03), this will have to be rebased, and the bindata will have to be re-generated after that is merged

I'll stand by to rebase and regenerate bindata once that is merged.

sirlatrom:
@thaJeztah Since #1781 is not merged yet, I assume the regenerating of bindata will happen in that PR?

thaJeztah:
Spoke with @dperny and he's ok doing a rebase/regenerate of the spec, so let's merge 🎉

PR #1784, opened 2019-03-28
Fix hang on `docker stack deploy` with Kubernetes 1.14

simonferquel:
There were 2 issues in this logic:
- we did watch for stack with non-matching label selector: we used the
one for getting stacks child resources, instead of a field selector on
the stack name
- something in kube 1.14 has changed the behavior of watches against
aggregated API that makes context cancellation effective only when the
next event is received. I don't know exactly where the problem is, but
it is certainly on the central API Server or in Kube-Proxy, as the watch
hangs with any version of go client vendoring with a server running 1.14

This should be back-ported to older version of the CLI

**- What I did**

I fixed the way we watch for stack events on stack deploy.
- our watch query was not matching the stack we just created
- with a server running Kubernetes 1.14, the watch hanged for 7 minutes (default watch timeout)

**- How I did it**

- For the watch query, I replaced the wrong label selector with a name field selector
- For the hang, the main issue seem to be a bug in kubernetes 1.14 were watches on aggregated resources are effective only when the server sends the next event. Combined with the other bug (where we did not match the actual stack we were supposed to watch), we had a systematic hang. With the previous fix, depending on timing, the watcher could hang as well, but not all time. I fixed this by intercepting the watcher interface to bypass graceful cancellation of the server-side watcher.

**- How to verify it**

- Deploy compose-on-kubernetes on a Kubernetes v1.14 cluster
- Deploy a stack on it

Note @thaJeztah: this should be backported to older CLI versions as well
@zappy-shu https://github.com/docker/cli/pull/1774 should be rebased on this one for full 1.14 compatibility.


simonferquel:
The bug has been acknowledged upstream: https://github.com/kubernetes/kubernetes/issues/75837, investigation concluded with a bug introduced in golang 1.12 (Kube 1.14 is built with golang 1.12), reported here: https://github.com/golang/go/issues/31125, confirmed by another person hit by this issue here: https://github.com/golang/go/issues/31126, and a patch has been submitted: https://go-review.googlesource.com/c/go/+/170066/.

Thanks @liggitt for the swift handling !

In the meantime, this client-side mitigation is legit and should be applied both in master and within patch releases of currently supported versions.

simonferquel:
I spotted a new possible hang that this fix does not completely compensate, I'll update it and remove the [wip] marker.

thaJeztah:
Looks like we have to wait for Go 1.12.2 (also because it will have the fix for https://github.com/golang/go/issues/30977)

simonferquel:
Actually the bug is server side (the reverse-proxy created to relay to the stack api server is the source of the bug).
We need to discuss if:
- we want to merge a client-side workaround (and having it back-ported in relevant CLI versions)
- wait for k8s 1.14.1, which might solve the issue by either reverting to golang 1.11.x, or using a future golang 1.12 patch release

simonferquel:
https://github.com/kubernetes/kubernetes/pull/75887 <- there is a fix planned for kube 1.14.1 (working around the root issue in golang http/utils library). I guess we should drop this client-side fix.
In the mean time I'll create another PR for just fixing the selector used to watch the stacks.

simonferquel:
I stripped the weird workaround from this PR. I'll prepare a PR with it before release if we see we can't ship Desktop / UCP with 1.14.1

simonferquel:
@wsong for UCP
@gtardif for Desktop

PR #1789, opened 2019-03-29
Fix login logout when engine is down and credential helper is in use

djs55:
**- What I did**

Fixed `docker login` and `docker logout` when the engine is not running and when using a credential helper.

**- How I did it**

The logic in `ElectAuthServer` accidentally combined together checking the error from the engine with the state of the debug flag when deciding which value to return.

**- How to verify it**

I used Docker Desktop on Mac with the `docker-credential-osxkeychain` enabled (which is the default). I shutdown the app. Without this patch I ran:

```
$ docker logout
Not logged in to
```
-- note the missing server URL

```
$ docker --debug logout
Warning: failed to get default registry endpoint from daemon (Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?). Using system default: https://index.docker.io/v1/
Not logged in to https://index.docker.io/v1/
```    
-- note the server URL is present

This shows that the `--debug` flag is changing the actual behaviour of the program.

I ran `docker login` and observed an error from the credential helper
```
Error saving credentials: error storing credentials - err: no credentials server URL, out: `no credentials server URL`
```

When I rebuilt the CLI with this PR `docker login` and `docker logout` operated as expected.

I separated this PR into 2 patches:
1. the actual bugfix
2. a small refactoring which I think makes `ElectAuthServer` easier to read but which is optional

Fix `docker login` and `docker logout` when the engine is not running and a credential helper is in use.

39df0ec8dab014008eb217865bc4635b

### DIFF
--- a/cli/command/registry.go
+++ b/cli/command/registry.go
@@ -28,21 +28,22 @@ func ElectAuthServer(ctx context.Context, cli Cli) string {
 	// used. This is essential in cross-platforms environment, where for
 	// example a Linux client might be interacting with a Windows daemon, hence
 	// the default registry URL might be Windows specific.
-	serverAddress := registry.IndexServer
-	if info, err := cli.Client().Info(ctx); err != nil {
+	info, err := cli.Client().Info(ctx)
+	if err != nil {
 		// Daemon is not responding so use system default.
 		if debug.IsEnabled() {
 			// Only report the warning if we're in debug mode to prevent nagging during engine initialization workflows
-			fmt.Fprintf(cli.Err(), "Warning: failed to get default registry endpoint from daemon (%v). Using system default: %s\n", err, serverAddress)
+			fmt.Fprintf(cli.Err(), "Warning: failed to get default registry endpoint from daemon (%v). Using system default: %s\n", err, registry.IndexServer)
 		}
-	} else if info.IndexServerAddress == "" {
-		if debug.IsEnabled() {
-			fmt.Fprintf(cli.Err(), "Warning: Empty registry endpoint from daemon. Using system default: %s\n", serverAddress)
-		}
-	} else {
-		serverAddress = info.IndexServerAddress
+		return registry.IndexServer
 	}
-	return serverAddress
+	if info.IndexServerAddress == "" {
+		if debug.IsEnabled() {
+			fmt.Fprintf(cli.Err(), "Warning: Empty registry endpoint from daemon. Using system default: %s\n", registry.IndexServer)
+		}
+		return registry.IndexServer
+	}
+	return info.IndexServerAddress
 }
 
 // EncodeAuthToBase64 serializes the auth configuration as JSON base64 payload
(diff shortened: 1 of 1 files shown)


cli/command/registry.go
```diff
@@ -28,16 +28,22 @@ func ElectAuthServer(ctx context.Context, cli Cli) string {
...
 	// the default registry URL might be Windows specific.
-	serverAddress := registry.IndexServer
-	if info, err := cli.Client().Info(ctx); err != nil && debug.IsEnabled() {
-		// Only report the warning if we're in debug mode to prevent nagging during engine initialization workflows
-		fmt.Fprintf(cli.Err(), "Warning: failed to get default registry endpoint from daemon (%v). Using system default: %s\n", err, serverAddress)
-	} else if info.IndexServerAddress == "" && debug.IsEnabled() {
-		fmt.Fprintf(cli.Err(), "Warning: Empty registry endpoint from daemon. Using system default: %s\n", serverAddress)
-	} else {
-		serverAddress = info.IndexServerAddress
+	info, err := cli.Client().Info(ctx)
```
thaJeztah: Should we use `/_ping` for this? (`/info` is a bit heavy _just_ to know if the engine is up and running)

cli/command/registry.go
```diff
@@ -28,16 +28,22 @@ func ElectAuthServer(ctx context.Context, cli Cli) string {
...
+	if err != nil {
+		// Daemon is not responding so use system default.
+		if debug.IsEnabled() {
+			// Only report the warning if we're in debug mode to prevent nagging during engine initialization workflows
+			fmt.Fprintf(cli.Err(), "Warning: failed to get default registry endpoint from daemon (%v). Using system default: %s\n", err, registry.IndexServer)
+		}
+		return registry.IndexServer
+	}
+	if info.IndexServerAddress == "" {
```
thaJeztah: Oh, never mind, I see we need this information 🤔

thaJeztah: Actually looks like it's hard-coded, and never updated anywhere 🤔 https://github.com/moby/moby/blob/fc01c2b481097a6057bec3cd1ab2d7b4488c50c4/registry/config.go#L29-L48

cli/command/registry.go
```diff
@@ -28,16 +28,22 @@ func ElectAuthServer(ctx context.Context, cli Cli) string {
...
+		if debug.IsEnabled() {
```
silvin-lubecki: We definitely need a real log policy here: we should write something like `log.Warning(...)` and let the debug flag change only the output level, not spreading it everywhere. But of course that's another issue.

cli/command/registry.go
```diff
@@ -28,16 +28,22 @@ func ElectAuthServer(ctx context.Context, cli Cli) string {
 	// used. This is essential in cross-platforms environment, where for
 	// example a Linux client might be interacting with a Windows daemon, hence
```
thaJeztah: Looks like this was added in https://github.com/moby/moby/pull/19891 at a time where a temporary/special Windows registry was set up: https://github.com/moby/moby/issues/18019

Actually thinking if that handling is now just dead code and no longer needed

djs55:
Thanks for the very quick response (and merge)!

PR #1792, opened 2019-03-29
Entrypoint short example

dagelf:
Useful for people searching the help for "cmd" and who do not use the parameter format every day.

Finding this format takes some digging through the docs... this puts it on the front page.


dagelf:
It's just one line of text, and this is my first PR, I'm on a slow connection, haven't learned how to sign it yet... I did load a key onto github.

thaJeztah:
It's the DCO sign-off that's needed (`-s` lower-case), not a GPG signed commit (`-S` capital s); if you're making the changes using the GitHub UI, you need to do that manually, by adding this line as last line of your commit message:

```
Signed-off-by: Coenraad Loubser <coenraad@wish.org.za>
```


Note, however, that the `usage` output on these pages should match the output that's in `docker <command> --help`, so that would have to be modified in the code itself 😅

dagelf:
It does show that if I click the ...

And I ran that command exactly like that too, with the -s

I also added it to the first message... and it matches the format well. In fact, the format of [ARG] or string isn't defined, and it not used consistently, another option would be to clutter the top of the clean output of --help to be more like the good old man pages from the '80s.

PR #1794, opened 2019-04-01
[18.09 backport] Fix the stack informer's selector used to track deployment

thaJeztah:
Old selector was wrong (it watched for the label we applied to child
resources when reconciling the stack, instead of the stack itself)

This should be back-ported to older version of the CLI


thaJeztah:
ping @simonferquel @silvin-lubecki PTAL

PR #1800, opened 2019-04-03
vendor: update buildkit to 62e55427

tonistiigi:
Brings in latest buildkit, including new progress formatting and tar output.

`docker build -o type=tar,dest=file.tar .`
`docker build -o - . > file.tar`

@tiborvass @AkihiroSuda 


cli/command/image/build.go
```diff
@@ -662,13 +662,22 @@ func parseOutputs(inp []string) ([]types.ImageBuildOutput, error) {
...
 			return nil, err
 		}
-		if len(fields) == 1 && fields[0] == s {
-			outs = append(outs, types.ImageBuildOutput{
-				Type: "local",
-				Attrs: map[string]string{
-					"dest": s,
-				},
-			})
+		if len(fields) == 1 && fields[0] == s && !strings.HasPrefix(s, "type=") {
```
AkihiroSuda: I feel this behavior is confusing

tonistiigi: why?

AkihiroSuda: If `docker build -o -` produces tar, user would also except `docker build -o a` to produce tar

tonistiigi: csv is the main format. non csv is an output path similar to many other tools. And `-` path does not make sense as a directory name, so as many other tools it means outputting to stdout(and we can't write a directory to stdout).

> If docker build -o - produces tar, user would also except docker build -o a to produce tar

Somewhat yes, but docker build always produces a directory(list of files), not a single file. Most common (and promoted) way to use it should be `docker build -o . .` meaning input from current and output to current, while `cat in.tar | docker build -o - - > out.tar` also follows same pattern.

tonistiigi: Note that we do not support `docker build ctx.tar` either.

tiborvass:
Ping @silvin-lubecki

vendor.conf
```diff
@@ -76,7 +76,9 @@ github.com/spf13/cobra v0.0.3
 github.com/spf13/pflag 4cb166e4f25ac4e8016a3595bbf7ea2e9aa85a2c https://github.com/thaJeztah/pflag.git
 github.com/syndtr/gocapability 2c00daeb6c3b45114c80ac44119e7b8801fdd852
 github.com/theupdateframework/notary v0.6.1
-github.com/tonistiigi/fsutil 1ec1983587cde7e8ac2978e354ff5360af622464
+github.com/tonistiigi/fsutil 3bbb99cdbd76619ab717299830c60f6f2a533a6b
+github.com/jaguilar/vt100 ad4c4a5743050fb7f88ce968dca9422f72a0e3f2 git://github.com/tonistiigi/vt100.git
```
thaJeztah: This something we can contribute upstream? (I see the repo hasn't been active for 4 years, so if that doesn't work, alternatively we'd have to officially fork it?)

thaJeztah:
@tonistiigi can you do docs updates and completion-script updates in a follow up? /cc @albers

albers:
@tonistiigi, @thaJeztah Support for `-o` in bash completion is added as part of #1810.

PR #1808, opened 2019-04-04
add cli integration for unconfined systempaths

martencassel:
**- What I did**
I've implemented the proposed changes in https://github.com/docker/cli/pull/1347#issuecomment-475426162.  
Added --security-opt systempaths=unconfined support.

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -485,6 +485,8 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		return nil, err
 	}
 
+	securityOpts, maskedPaths, readonlyPaths := parseSystemPaths(securityOpts)
+
 	storageOpts, err := parseStorageOpts(copts.storageOpt.GetAll())
 	if err != nil {
 		return nil, err
@@ -635,6 +637,8 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		Sysctls:        copts.sysctls.GetAll(),
 		Runtime:        copts.runtime,
 		Mounts:         mounts,
+		MaskedPaths:    maskedPaths,
+		ReadonlyPaths:  readonlyPaths,
 	}
 
 	if copts.autoRemove && !hostConfig.RestartPolicy.IsNone() {
@@ -823,6 +827,25 @@ func parseSecurityOpts(securityOpts []string) ([]string, error) {
 	}
 
 	return securityOpts, nil
+}
+
+// parseSystemPaths checks if `systempaths=unconfined` security option is set,
+// and returns the `MaskedPaths` and `ReadonlyPaths` accordingly. An updated
+// list of security options is returned with this option removed, because the
+// `unconfined` option is handled client-side, and should not be sent to the
+// daemon.
+func parseSystemPaths(securityOpts []string) (filtered, maskedPaths, readonlyPaths []string) {
+	filtered = securityOpts[:0]
+	for _, opt := range securityOpts {
+		if opt == "systempaths=unconfined" {
+			maskedPaths = []string{}
+			readonlyPaths = []string{}
+		} else {
+			filtered = append(filtered, opt)
+		}
+	}
+
+	return filtered, maskedPaths, readonlyPaths
 }
 
 // parses storage options per container into a map
--- a/cli/command/container/opts_test.go
+++ b/cli/command/container/opts_test.go
@@ -800,3 +800,57 @@ func TestValidateDevice(t *testing.T) {
 		}
 	}
 }
+
+func TestParseSystemPaths(t *testing.T) {
+	tests := []struct {
+		doc                       string
+		in, out, masked, readonly []string
+	}{
+		{
+			doc: "not set",
+			in:  []string{},
+			out: []string{},
+		},
+		{
+			doc: "not set, preserve other options",
+			in: []string{
+				"seccomp=unconfined",
+				"apparmor=unconfined",
+				"label=user:USER",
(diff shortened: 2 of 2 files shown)


cli/command/container/opts.go
```diff
@@ -825,6 +832,23 @@ func parseSecurityOpts(securityOpts []string) ([]string, error) {
 	return securityOpts, nil
 }
 
+func parseSystemPaths(securityOpts []string) ([]string, []string, []string, error) {
+	var maskedPaths []string
+	var readonlyPaths []string
+	filtered := securityOpts[:0]
+	for _, opt := range securityOpts {
+		con := strings.SplitN(opt, "=", 2)
+		if con[0] == "systempaths" && con[1] == "unconfined" {
```
thaJeztah: Looks like we don't need the split (and don't strip whitespace etc) at this point, so this could be simplified to

```suggestion
		if opt == "systempaths=unconfined" {
```

cli/command/container/opts.go
```diff
@@ -485,6 +485,11 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		return nil, err
 	}
 
+	securityOpts, maskedPaths, readonlyPaths, err := parseSystemPaths(securityOpts)
+	if err != nil {
```
thaJeztah: I see we never return an error; although it's good practice to keep this signature, this function is not exported, so I think it's ok to drop the `error` return for now, and remove this error-handling

cli/command/container/opts.go
```diff
@@ -825,6 +832,23 @@ func parseSecurityOpts(securityOpts []string) ([]string, error) {
 	return securityOpts, nil
 }
 
+func parseSystemPaths(securityOpts []string) ([]string, []string, []string, error) {
```
thaJeztah: I realise this function is not exported, but it might be important to document what it does (especially because it filters the `securityOpts`; perhaps add a short GoDoc; something along the line of:

```suggestion
// parseSystemPaths checks if `systempaths=unconfined` security option is set,
// and returns the `MaskedPaths` and `ReadonlyPaths` accordingly. An updated
// list of security options is returned with this option removed, because the
// `unconfined` option is handled client-side, and should not be sent to the
// daemon.
func parseSystemPaths(securityOpts []string) ([]string, []string, []string, error) {
```

(suggestions for better wording welcome! 😂)


Although I'm generally not a fan of named return variables, I'm a bit on the fence if we should use them here (for readability), as three `[]string` returns might be difficult to read/grasp;

```suggestion
func parseSystemPaths(securityOpts []string) (filtered, maskedPaths, readonlyPaths []string) {
```

I'd love to get feedback from others in that as well though 😅

cli/command/container/opts_test.go
```diff
@@ -800,3 +800,37 @@ func TestValidateDevice(t *testing.T) {
 		}
 	}
 }
+
+func TestParseSystemPaths(t *testing.T) {
+	valids := []string{
```
thaJeztah: Would be good to test other combinations (i.e., make sure if doesn't touch other values if not set, and that invalid/unknown options are ignored; quick write-up for that;

```go
func TestParseSystemPaths(t *testing.T) {
	tests := []struct {
		doc                       string
		in, out, masked, readonly []string
	}{
		{
			doc: "not set",
			in:  []string{},
			out: []string{},
		},
		{
			doc: "not set, preserve other options",
			in: []string{
				"seccomp=unconfined",
				"apparmor=unconfined",
				"label=user:USER",
				"foo=bar",
			},
			out: []string{
				"seccomp=unconfined",
				"apparmor=unconfined",
				"label=user:USER",
				"foo=bar",
			},
		},
		{
			doc:      "unconfined",
			in:       []string{"systempaths=unconfined"},
			out:      []string{},
			masked:   []string{},
			readonly: []string{},
		},
		{
			doc:      "unconfined and other options",
			in:       []string{"foo=bar", "systempaths=unconfined", "bar=baz"},
			out:      []string{"foo=bar", "bar=baz"},
			masked:   []string{},
			readonly: []string{},
		},
		{
			doc: "unknown option",
			in:  []string{"foo=bar", "systempaths=unkown", "bar=baz"},
			out: []string{"foo=bar", "systempaths=unkown", "bar=baz"},
		},
	}

	for _, tc := range tests {
		securityOpts, maskedPaths, readonlyPaths, err := parseSystemPaths(tc.in)
		assert.NilError(t, err)
		assert.DeepEqual(t, securityOpts, tc.out)
		assert.DeepEqual(t, maskedPaths, tc.masked)
		assert.DeepEqual(t, readonlyPaths, tc.readonly)
	}
}
```

thaJeztah:
ping @tiborvass @jessfraz PTAL 🤗

martencassel:
I've implemented your suggested changes. I see that the linting failed due to the deliberate "unkown"  text.

thaJeztah:
Oh! The linter detected my typo 😂

thaJeztah:
Oh! when you're updating, could you fix your sign-off to use your real name (instead of your github handle)?

```
martencassel <marten.cassel@gmail.com>
```

Guessing it's `Marten Cassel`, so should then be:

```
Marten Cassel <marten.cassel@gmail.com>
```

martencassel:
Updated sign-off to Mårten Cassel <marten.cassel@gmail.com>

AkihiroSuda:
ping @tiborvass @jessfraz

PhilippWendler:
It seems the [documentation](https://docs.docker.com/engine/reference/run/#security-configuration) was forgotten to be updated and this option remains undocumented. Could this be added to the docs?

PR #1810, opened 2019-04-05
Add BuildKit specific options to bash completion

albers:
Hints for reviewers: 

### Visibility of the options is not trivial
- `--compress` and `--stream` are not available in BuildKit mode
- `--stream` is experimental

### csv format
There is no easy way to add completion to option values that expect csv format.
This is because bash completion does by default not perform word splitting on `,`.
This can only be modified globally, which will affect all installed completion scripts.

Therefore these values are treated as "any string".

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -2863,7 +2863,6 @@ _docker_image_build() {
 	"
 
 	local boolean_options="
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @tiborvass PTAL

PR #1817, opened 2019-04-09
Add warnings when DOCKER_HOST conflicts with contexts

simonferquel:
**- What I did**

For clarity, on `docker context use` or `docker context ls`, this adds a warning if the DOCKER_HOST variable is set because it overrides the active context.

**- How I did it**

Just a one liner to test if the variable is set, and print a message on stderr

**- How to verify it**

- `DOCKER_HOST=unix:///var/run/docker.sock docker context ls`

### DIFF
--- a/cli/command/context/list.go
+++ b/cli/command/context/list.go
@@ -2,6 +2,7 @@ package context
 
 import (
 	"fmt"
+	"os"
 	"sort"
 
 	"github.com/docker/cli/cli"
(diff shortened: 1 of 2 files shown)


cli/command/context/list.go
```diff
@@ -76,7 +76,14 @@ func runList(dockerCli command.Cli, opts *listOptions) error {
 	sort.Slice(contexts, func(i, j int) bool {
 		return sortorder.NaturalLess(contexts[i].Name, contexts[j].Name)
 	})
-	return format(dockerCli, opts, contexts)
+	if err := format(dockerCli, opts, contexts); err != nil {
+		return err
+	}
+	if isDockerHostSet() {
+		fmt.Fprint(dockerCli.Err(), "Warning: DOCKER_HOST environment variable overrides the active context. "+
+			"To use a context, either set the global --context flag, or unset DOCKER_HOST environment variable.\n")
```
thaJeztah: Wondering (orthogonal); could we have a `--force` flag to unset `DOCKER_HOST` ? i.e.;

```bash
export DOCKER_HOST=foo
docker context use --force my-context

echo $DOCKER_HOST
# not set
```

simonferquel: How would you unset an env variable in the parent process?

thaJeztah: LOL, good point; magic?

magic

simonferquel: cc @chris-crone !

simonferquel:
@thaJeztah, inlined the env var tests.

PR #1823, opened 2019-04-12
Regroup all kubernetes extra-fields under x-kubernetes

simonferquel:
**- What I did**

This regroup all Kubernetes extra fields for compose-on-kubernetes v1alpha3 in a single x-kubernetes object.

Also use the same naming scheme as cap_add etc. for fields inside this object.

**- How I did it**

Just a little bit of refactoring

**- How to verify it**

This is covered by unit tests


thaJeztah:
👍 definitely a lot cleaner, and likely better maintainable

@vdemeester @silvin-lubecki PTAL; I was discussing this with @simonferquel, and thought this would be a cleaner solution than adding yet-another-x-field for each thing kubernetes-specific (that isn't a "stable" API yet)

thaJeztah:
looks good to me, but waiting for others to have a look

PR #1830, opened 2019-04-13
Switch to google/shlex

thaJeztah:
The github.com/flynn-archive/go-shlex package is a fork of Google/shlex,
and the repository is now archived, so let's switch to the maintained
version.


thaJeztah:
@dnephin I couldn't find history if this package was used for a specific reason (instead of google/shlex); perhaps you know?

dnephin:
I don't remember. It's possible at the time google had not moved the code to github yet

thaJeztah:
So for those interested; it looks like they forked at this commit: https://github.com/google/shlex/commit/6d55fc2f1c3f2d7fd0f0bfadfb0f3719ded87d3d

Changes made in flynn-archive/go-shlex since that; https://github.com/flynn-archive/go-shlex/compare/6d55fc2f1c3f2d7fd0f0bfadfb0f3719ded87d3d...3f9db97f856818214da2e1057f8ad84803971cff

And changes made in google/shlex since that; https://github.com/google/shlex/compare/6d55fc2f1c3f2d7fd0f0bfadfb0f3719ded87d3d...c34317bd91bf98fab745d77b03933cf8769299fe

These two changes caught my eye;

- https://github.com/flynn-archive/go-shlex/commit/70015acf9353907700fc808ceab7f17b7b86fc7f "apparently / is not a char, i wonder what else..."
- https://github.com/flynn-archive/go-shlex/commit/3fe3f523a9b1008a219279a97e25f4b8b2bd4691  "i mean, maybe char should be the default case when parsing"

Which both seemed kinda YOLO / hacky

Google removed that functionality in https://github.com/google/shlex/commit/6f45313302b9c56850fc17f99e40caebce98c716 ("Modify shlex to remove the charRuneClass. This makes it behave more like shlex.py with whitespace_split=True")

Either way, it might be difficult to catch corner case differences between both libraries, but given that
- flynn-archive/go-shlex is no longer maintained, and google/shlex is (or at least more recently updated), 
- google/shlex is used in other parts of the code (buildkit uses it)

I think it makes sense to make the switch and use a single implementation (YOLO!) 😂

PR #1836, opened 2019-04-15
Split the context store interface

simonferquel:
**- What I did**

- I split the context store "meta"-interface into smaller ones
- I changed existing code to depend on smaller interfaces

This is to make it easier to implement support for exporting contexts in
3rd party code, or to create mocks in tests.

2 exemples where it simplify things:
- docker-app desktop-specific context decorator (which rewrites parts of
the docker context to simplify UX when using on Docker Desktop contexts)
- ucp for including a context in the connection bundle

**- How I did it**

- simple interface splitting

**- How to verify it**

- everything I changed is covered by existing unit tests

### DIFF
--- a/cli/context/store/store.go
+++ b/cli/context/store/store.go
@@ -18,15 +18,47 @@ import (
 
 // Store provides a context store for easily remembering endpoints configuration
 type Store interface {
-	ListContexts() ([]ContextMetadata, error)
-	CreateOrUpdateContext(meta ContextMetadata) error
-	RemoveContext(name string) error
+	Reader
+	Lister
+	Writer
+	StorageInfo
+}
+
+// Reader provides read-only (without list) access to context data
+type Reader interface {
 	GetContextMetadata(name string) (ContextMetadata, error)
-	ResetContextTLSMaterial(name string, data *ContextTLSData) error
-	ResetContextEndpointTLSMaterial(contextName string, endpointName string, data *EndpointTLSData) error
 	ListContextTLSFiles(name string) (map[string]EndpointFiles, error)
 	GetContextTLSData(contextName, endpointName, fileName string) ([]byte, error)
+}
+
+// Lister provides listing of contexts
+type Lister interface {
+	ListContexts() ([]ContextMetadata, error)
+}
+
+// ReaderLister combines Reader and Lister interfaces
+type ReaderLister interface {
+	Reader
+	Lister
+}
+
+// StorageInfo provides more information about storage details of contexts
+type StorageInfo interface {
 	GetContextStorageInfo(contextName string) ContextStorageInfo
+}
+
+// Writer provides write access to context data
+type Writer interface {
+	CreateOrUpdateContext(meta ContextMetadata) error
+	RemoveContext(name string) error
+	ResetContextTLSMaterial(name string, data *ContextTLSData) error
+	ResetContextEndpointTLSMaterial(contextName string, endpointName string, data *EndpointTLSData) error
+}
+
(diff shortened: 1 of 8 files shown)


simonferquel:
cc @silvin-lubecki @thaJeztah

cli/context/store/store.go
```diff
@@ -18,15 +18,41 @@ import (
...
+// Read provides read-only (without list) access to context data
+type Read interface {
+	GetContextMetadata(name string) (ContextMetadata, error)
+	ListContextTLSFiles(name string) (map[string]EndpointFiles, error)
+	GetContextTLSData(contextName, endpointName, fileName string) ([]byte, error)
+}
+
+// List provides listing of contexts on top of Read interface
+type List interface {
+	Read
```
silvin-lubecki: Is Read really useful here? I think you should let the user compose between the different interface, and maybe recreate an object which will be Reader and Lister.

cli/context/store/store.go
```diff
@@ -18,15 +18,41 @@ import (
 
 // Store provides a context store for easily remembering endpoints configuration
 type Store interface {
+	StorageInfo
+	Write
+}
+
+// Read provides read-only (without list) access to context data
+type Read interface {
```
silvin-lubecki: Nit: What about using the same terminology than io.Writer/Reader... Lister etc..?

simonferquel:
@silvin-lubecki renamed the interface and replaced interface inheritance with interface composition

simonferquel:
Being able to create a context file by just implementing the Reader interface (and calling store.Export) is very valuable. Previously, you had to guess which functions of the store.Store meta-interface was actually needed by store.Export interface.

simonferquel:
ping @thaJeztah

cli/context/store/store.go
```diff
@@ -18,15 +18,47 @@ import (
...
 	ListContexts() ([]ContextMetadata, error)
+}
+
+// ReaderLister combines Reader and Lister interfaces
+type ReaderLister interface {
+	Reader
+	Lister
+}
+
+// StorageInfo provides more invormation about storage details of contexts
```
thaJeztah:
```suggestion
// StorageInfo provides more information about storage details of contexts
```

cli/context/store/store.go
```diff
@@ -18,15 +18,47 @@ import (
...
+	Lister
+	Writer
+	StorageInfo
+}
+
+// Reader provides read-only (without list) access to context data
+type Reader interface {
+	GetContextMetadata(name string) (ContextMetadata, error)
+	ListContextTLSFiles(name string) (map[string]EndpointFiles, error)
+	GetContextTLSData(contextName, endpointName, fileName string) ([]byte, error)
```
thaJeztah: It might be too late (not sure), but I keep looking at these and think: we should remove `Context` from these names.

I did a quick find/replace to see what it would look like, and pushed it to a branch; https://github.com/thaJeztah/cli/tree/remove_context_from_names

(it's based on this PR, so look at the last commit; https://github.com/docker/cli/commit/1068242c1545a1a4ac0c7a33a7e6c36d96b6b329)

simonferquel: I aggree with this one, and if we want to do it, we have to do it before 19.03 is out I think. It will have impact on docker-app code, and on my current PR in UCP for adding a context file in the bundle though. I'll do it in a followup PR, so that we can choose to merge it or not.

simonferquel:
Fixed the typo & squashed, ping @thaJeztah (currently preparing followup PR)

thaJeztah:
ah, yes, let's merge

PR #1840, opened 2019-04-19
cli-plugins: alias an existing allowed command (only builder for now)

tiborvass:
With this patch it is possible to alias an existing allowed command.
At the moment only builder allows an alias.

This also properly puts the build command under builder, instead of image
where it was for historical reasons.

### DIFF
--- a/cli-plugins/manager/manager.go
+++ b/cli-plugins/manager/manager.go
@@ -164,6 +164,7 @@ func PluginRunCommand(dockerCli command.Cli, name string, rootcmd *cobra.Command
 			return nil, err
 		}
 		if plugin.Err != nil {
+			// TODO: why are we not returning plugin.Err?
 			return nil, errPluginNotFound(name)
 		}
 		cmd := exec.Command(plugin.Path, args...)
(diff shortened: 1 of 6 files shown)


tiborvass:
Ping @thaJeztah @vdemeester PTAL :))

thaJeztah:
> This also properly puts the build command under builder, instead of image where it was for historical reasons.

I still think `docker image build` makes sense for the current feature-set (as in; you're building an image), but I guess if the BuildKit integration progresses, we'll get other options. Is the plan to _keep_ `docker image build`? (should we in future add `docker <some-object> build`?)

One downside is that `docker builder build` is a bit double 😕

tonistiigi:
> I still think docker image build

Independently from the taste, it doesn't allow us to add new subcommands to add features to the builder. We already had this with `builder prune` where it can't be added under image. It makes sense to normalize to `builder build` as well as it follows the existing pattern and currently a lot of users do not find `builder prune` because they don't see the `builder` command. I don't see a need for removing `image build`, (not sure if anyone uses it), but we could go over the docs to make sure we prefer `builder build` over `image build` there (after release).

tonistiigi:
@thaJeztah No, but the plugin itself remains visible in the list with *

andrewhsu:
before this PR is merged, we got:
```
docker build [OPTIONS] PATH | URL | -
docker builder prune
```

after this PR is merged, we'll have:
```
docker build [OPTIONS] PATH | URL | -
docker builder prune
docker builder build [OPTIONS] PATH | URL | -
```

PR #1841, opened 2019-04-19
Do not disable sig-proxy when using a TTY

thaJeztah:
This partially reverts https://github.com/moby/moby/commit/e0b59ab52b87b8fc15dd5534c3231fdd74843f9f (https://github.com/moby/moby/pull/2426), and does not automatically disable proxying signals in TTY-mode

fixes https://github.com/moby/moby/issues/28872 (docker client doesn't pass signals when a terminal is attached)
fixes https://github.com/moby/moby/issues/3793 docker client not passing signals to dockerd

relates to:

- https://github.com/moby/moby/issues/9098 Kill `docker exec` command will not terminate the spawned process
  - This patch does not fix the `docker exec` case; it looks like there's no API to kill an exec'd process, so there's no signal-proxy for this yet

Before this change:
------------------------------------

Start a container with a TTY in one shell:

```
docker run -it --init --name repro-28872 busybox sleep 30
```

then, in another shell, kill the docker cli:

```
kill `pgrep -f repro-28872`
```

Notice that the CLI was killed, but the signal not forwarded to the container;
the container continues running

```
docker container inspect --format '{{ .State.Status }}' repro-28872
running

docker container rm -f repro-28872
```

After this change:
------------------------------------

Start a container with a TTY in one shell:

```
docker run -it --init --name repro-28872 busybox sleep 30
```

then, in another shell, kill the docker cli:

```
kill `pgrep -f repro-28872`
```

Verify that the signal was forwarded to the container, and the container exited

```
docker container inspect --format '{{ .State.Status }}' repro-28872
exited

docker container rm -f repro-28872
```

```
- Fix docker client doesn't pass signals when a terminal is attached
```

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -117,11 +117,6 @@ func runContainer(dockerCli command.Cli, opts *runOptions, copts *containerOptio
 		config.StdinOnce = false
 	}
 
-	// Disable sigProxy when in TTY mode
-	if config.Tty {
-		opts.sigProxy = false
-	}
-
 	// Telling the Windows daemon the initial size of the tty during start makes
 	// a far better user experience rather than relying on subsequent resizes
 	// to cause things to catch up.
--- a/e2e/container/proxy_signal_test.go
+++ b/e2e/container/proxy_signal_test.go
@@ -1,0 +1,56 @@
+package container
+
+import (
+	"os/exec"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/docker/cli/e2e/internal/fixtures"
+	"github.com/kr/pty"
+	"gotest.tools/assert"
+	"gotest.tools/icmd"
+	"gotest.tools/poll"
+)
+
+// TestSigProxyWithTTY tests that killing the docker CLI forwards the signal to
+// the container, and kills the container's process. Test-case for moby/moby#28872
+func TestSigProxyWithTTY(t *testing.T) {
+	_, tty, err := pty.Open()
+	assert.NilError(t, err, "could not open pty")
+	defer func() { _ = tty.Close() }()
+
+	containerName := "repro-28872"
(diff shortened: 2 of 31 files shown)


thaJeztah:
@vdemeester I tried coming up with a test for this, but I couldn't get it to work (creating a container with a TTY attached, then signal that docker cli); perhaps you have ideas?)

thaJeztah:
oooh, actually, I think I might've found it 🤗 - cleaning up a bit, and will add a test

thaJeztah:
Dang; why doesn't it work here 😞 

```
11:10:01 === FAIL: e2e/container TestSigProxyWithTTY (6.16s)
11:10:01     proxy_signal_test.go:38: terminating PID 3793
11:10:01     proxy_signal_test.go:42: timeout hit after 5s: expected status exited != running
```

thaJeztah:
Hm... it works when running `dind` (using `/var/run/docker.sock`), but fails when using `tcp://` 🤔

thaJeztah:
Ok, so my test passes with `/var/run/docker.sock`, but many other tests fail; need to look into why that's happening 🤔 

```
    revoke_test.go:59: assertion failed: 
        Command:  docker pull registry:5000/alpine:3.6
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:   Error response from daemon: Get https://registry:5000/v2/: Service Unavailable
```

thaJeztah:
If someone has any ideas on the test, help is appreciated 🤗

thaJeztah:
I'll be damned.. LOL, tests are green now 🤷‍♂ 

let me remove the testing commits from here

thaJeztah:
Well.. there you go; all green and shiny 🎉 

ping @vdemeester @tianon @silvin-lubecki PTAL

thaJeztah:
Speedy gonzales !

PR #1857, opened 2019-05-02
Support GOARM and windows .exe in binary target

cpuguy83:
This just makes it easier to build a targeted binary for the
goos/goach/goarm version.

This of course will not work for all cases but is nice to get things
going.
Specifically cross-compiling pkcs for yubikey support requires some
extra work which is not tackled here.


cpuguy83:
Ping

PR #1865, opened 2019-05-09
Not "each" instruction is run independently

browncrane:
>The ENV instruction sets the environment variable <key> to the value <value>. This value will be in the environment for all subsequent instructions in the build stage and can be replaced inline in many as well.

There may be others.

**- What I did**

**- How I did it**

**- How to verify it**


browncrane:
> Please sign your commits following these rules:
> https://github.com/moby/moby/blob/master/CONTRIBUTING.md#sign-your-work
> The easiest way to do this is to amend the last commit:
> 
> ```
> $ git clone -b "patch-1" git@github.com:browncrane/cli.git somewhere
> $ cd somewhere
> $ git commit --amend -s --no-edit
> $ git push -f
> ```
> 
> Amending updates the existing PR. You **DO NOT** need to open a new one.

I did it.
But not sure how to handle current test fail

devdreamsolution (approved):

PR #1867, opened 2019-05-10
Add logic for finding a docker context within a UCP bundle

simonferquel:
A bundle is a zip file, and the upcoming UCP version will have a
.dockercontext file in it.

This PR makes it non-mandatory to uncompress such a UCP bundle
before importing the context.

cc @dhiltgen @chris-crone @thaJeztah

### DIFF
--- a/cli/command/context/import.go
+++ b/cli/command/context/import.go
@@ -1,9 +1,13 @@
 package context
 
 import (
+	"archive/tar"
+	"archive/zip"
+	"errors"
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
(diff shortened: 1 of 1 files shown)


cli/command/context/import.go
```diff
@@ -39,11 +43,35 @@ func RunImport(dockerCli command.Cli, name string, source string) error {
 		defer f.Close()
 		reader = f
 	}
-
 	if err := store.Import(name, dockerCli.ContextStore(), reader); err != nil {
```
thaJeztah: This could fail for other reasons, in which case we shouldn't proceed; perhaps we should detect the filetype, and if it's a Zip, call a function that  returns the `reader` for the `.dockercontext` inside the zip (e.g. rename `importBundleFile` to `readContextFromBundle` (or whatever) and make it return a `reader`), and if it's a `tar` (`.dockercontext`), open a reader for the tar file

simonferquel: Hmm, yes, I suppose it could work. 
The only problem I have for now, is with stdin handling: the Zip decompression algorithm requires random read access (a io.ReaderAt interface) to work. By definition, stdin is a forward only stream. The import command allows to read the tar from stdin (allowing things like `curl <...> | docker context import new-ctx -`, which is cool). So what should we do?
- disable std-in support altogether for coherence between tar and zip handling
- report an error when we have an invalid tar header coming from stdin saying that UCP bundles cant be passed trough stdin 
- do nothing, hope for the best ?

ijc: Are bundle.zip's small enough that we could reasonably slurp in the whole thing (`ioutil.ReadAll`) and then wrap in a `bytes.Reader` to provide `io.ReaderAt`?

simonferquel: If we can trust the user to pass sensible data to us (like an actual context/ucp bundle) I guess we can... but that is putting an unusual amount of trust in the hands of an unknown user :D

ijc: The user who is running `docker context create` can already do whatever an exploit of `docker context create` would be able to do since they have the same privilege level. Being able to somehow exploit/confuse `docker context create` doesn't get them anything extra, I don't think.

Even if we assume `curl http://somethinguntrusted | docker context create -` I think Go is (supposed to be) safe enough that the process will just eventually crash/OOM, there shouldn't be any possibility of e.g. a buffer overrun exploit. If we are concerned about that we could cap the size at, say, 10M or whatever "far too large to be a legitimate bundle" we like.

simonferquel: I like the idea of capping and loading in memory. I'll do that!

ijc: I just saw @thaJeztah pointing someone to https://golang.org/pkg/io/#LimitReader elsewhere and it reminded me of this...

thaJeztah: @ijc yes, that PR is for the alternative implementation (hybrid bundles); if you have time to help review that one, that'd be appreciated 🤗

andrewhsu:
closing in favour of https://github.com/docker/cli/pull/1895

PR #1747, opened 2019-03-18
Defer API version negotiation

simonferquel:
**- What I did**

Fix #1739 by deferring the API version negotiation to the point we actually need access to either the `ServerInfo` or the API client interface.

Also introduce a `no-remote` annotation for marking commands that do not require anything on the daemon side to work. Commands annotated as such are not evaluated by the conditional hiding feature (which relies on ServerInfo to get access to the daemon OS, experimental flag, etc.).

Context commands are annotated as no-remote

**- How to verify it**

- Create a docker context to an unreachable endpoint : `docker context create faulty --docker host=tcp://42.42.42.41:4242`
- Makes it your default context: `docker context use faulty`
- Commands that require interaction with the daemon should trigger a 1m timeout (e.g `docker info`)
- Commands that do not require any interaction with the daemon should run quickly (e.g. `docker context ls`)

CC @silvin-lubecki @thaJeztah

Issue #1739, opened 2019-03-15
Docker commands take 1 minute to timeout if context endpoint is unreachable


kinghuang:
**Description**

The upcoming context feature seem to cause CLI commands to pause for a 1 minute timeout if the current context's endpoint is not reachable.

Some of my Docker endpoints are only reachable via a VPN connection. If the current context is for an endpoint that I cannot currently reach, every command takes an extra minute (60 seconds) to complete.

**Steps to reproduce the issue:**

1. Create a context for an unreachable endpoint.
2. Use that context.
3. Run CLI commands like `docker context ls` or `docker info`.

**Describe the results you received:**

When the endpoint is unreachable, commands take an extra minute to complete.

In this example, when I switch to `psc-prod-gen` without the requisite network access, Docker CLI commands go through a 1 minute timeout. Even switching back to the `default` context for my local Docker config takes a minute.

```
▸ ~ time docker context ls
NAME                DESCRIPTION                               DOCKER ENDPOINT               KUBERNETES ENDPOINT                ORCHESTRATOR
default *           Current DOCKER_HOST based configuration   unix:///var/run/docker.sock   https://localhost:6443 (default)   swarm
psc-dev-gen         PSC Development Cluster                   tcp://hidden:2376                                           
psc-onboard-gen     PSC Onboard Cluster                       tcp://hidden:2376                                          
psc-prod-gen        PSC Production Cluster                    tcp://hidden:2376                                          
psc-staging-gen     PSC Staging Cluster                       tcp://hidden:2376                                           
psc-uat-gen         PSC UAT Cluster                           tcp://hidden:2376                                          
/Users/king.huang/go/src/github.com/docker/cli/build/docker-darwin-amd64  ls  0.03s user 0.01s system 101% cpu 0.047 total

▸ ~ time docker context use psc-prod-gen
psc-prod-gen
Current context is now "psc-prod-gen"
/Users/king.huang/go/src/github.com/docker/cli/build/docker-darwin-amd64  use  0.04s user 0.02s system 94% cpu 0.060 total

▸ ~ time docker context ls
NAME                DESCRIPTION                               DOCKER ENDPOINT           KUBERNETES ENDPOINT   ORCHESTRATOR
default             Current DOCKER_HOST based configuration                                                   
psc-dev-gen         PSC Development Cluster                   tcp://hidden:2376                          
psc-onboard-gen     PSC Onboard Cluster                       tcp://hidden:2376                         
[...]

PR #1872, opened 2019-05-13
add a 10s timeout to the client object.

ijc:
This partially mitigates #1739 ("Docker commands take 1 minute to timeout if
context endpoint is unreachable") and is a simpler alternative to #1747 (which
completely defers the client connection until an actual call is attempted).

Note that the previous 60s delay was the culmination of two separate 30s
timeouts since the ping is tried twice. This with this patch the overall
timeout is 20s. https://github.com/moby/moby/pull/39206 will remove the second
ping and once that propagates to this tree the timeout will be 10s.

### DIFF
--- a/cli/context/docker/load.go
+++ b/cli/context/docker/load.go
@@ -122,6 +122,7 @@ func (c *Endpoint) ClientOpts() ([]client.Opt, error) {
 				client.WithDialContext(helper.Dialer),
 			)
 		}
+		result = append(result, client.WithTimeout(10*time.Second))
 	}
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
ok, let's get this one in; we can tweak in a follow-up

ijc:
10s is pretty conservative but really I have no data just a gut feeling. I suppose it's possible that slow DNS or perhaps an on-demand VPN link or something could take approaching that long to actually route traffic. But yes in the normal case I you'd reasonably expect something an order of magnitude faster.

PR #1874, opened 2019-05-13
context: produce consistent output on `context create`.

ijc:
Refactor `RunCreate` slightly so that all three paths always produce the same
output, namely the name of the new context of `stdout` (for scripting) and the
success log message on `stderr`.

Validate by extending the existing unit tests to always check the output is as
expected.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/context/create_test.go
+++ b/cli/command/context/create_test.go
@@ -1,6 +1,7 @@
 package context
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"testing"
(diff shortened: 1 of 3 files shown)


silvin-lubecki:
Linter is complaining
```
internal/test/cli.go:172:1:warning: exported method FakeCli.ResetOutputBuffers should have comment or be unexported (golint)
```

ijc:
Oops, I got distracted right at the point I was supposed to write that comment. Now done.

cli/command/context/create_test.go
```diff
@@ -154,6 +155,8 @@ func TestCreateOrchestratorEmpty(t *testing.T) {
 		Docker: map[string]string{},
 	})
 	assert.NilError(t, err)
+	assert.Equal(t, "test\n", cli.OutBuffer().String())
```
silvin-lubecki: Factorize those 2 duplicated lines into a test helper function ?

ijc: :+1: done

PR #1880, opened 2019-05-13
Some linting fixes and updates to yamldocs generator

thaJeztah:
some minor fixes/changes to the yamldocs generator script that were made in other repositories

### DIFF
--- a/docs/yaml/Dockerfile
+++ b/docs/yaml/Dockerfile
@@ -1,4 +1,4 @@
 FROM scratch
-COPY docs /docs
 # CMD cannot be nil so we set it to empty string
 CMD  [""]
+COPY docs /docs
--- a/docs/yaml/generate.go
+++ b/docs/yaml/generate.go
@@ -17,12 +17,12 @@ import (
 const descriptionSourcePath = "docs/reference/commandline/"
 
 func generateCliYaml(opts *options) error {
-	dockerCli, err := command.NewDockerCli()
+	dockerCLI, err := command.NewDockerCli()
 	if err != nil {
 		return err
 	}
 	cmd := &cobra.Command{Use: "docker"}
-	commands.AddCommands(cmd, dockerCli)
+	commands.AddCommands(cmd, dockerCLI)
 	disableFlagsInUseLine(cmd)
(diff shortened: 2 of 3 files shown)


thaJeztah:
ping @silvin-lubecki @tiborvass

docs/yaml/generate.go
```diff
@@ -17,12 +17,12 @@ import (
 const descriptionSourcePath = "docs/reference/commandline/"
 
 func generateCliYaml(opts *options) error {
-	dockerCli, err := command.NewDockerCli()
+	dockerCLI, err := command.NewDockerCli()
```
silvin-lubecki: Shouldn't we do this rename to the entire code then? I think I mostly used `dockerCli` 😅

thaJeztah: yes, Ideally we should; likely that would break some people though, so that's why I didn't do it as part of this PR 😊

docs/yaml/Dockerfile
```diff
@@ -1,4 +1,4 @@
 FROM scratch
-COPY docs /docs
 # CMD cannot be nil so we set it to empty string
 CMD  [""]
+COPY docs /docs
```
tiborvass: this doesn't change anything :)

docs/yaml/yaml.go
```diff
@@ -68,6 +68,9 @@ func GenYamlTreeCustom(cmd *cobra.Command, dir string, filePrepender func(string
 			return err
 		}
 	}
+	if !cmd.HasParent() {
```
tiborvass: What does this do?

docs/yaml/generate.go
tiborvass: @thaJeztah oh you mean we would have to change "NewDockerCli" to "NewDockerCLI"? I don't think it's worth changing these little things but consistency is more important. So I'd vote for keeping it as-is.

docs/yaml/yaml.go
thaJeztah: I'll have to look into this one again, I think it was needed when generating docs for plugins, but it's been a while 😓

thaJeztah:
Let me close this one for now; really have to look into it again

PR #1881, opened 2019-05-13
Fix PowerShell codehint for rouge

thaJeztah:
Rouge is case-sensitive, and only works with powershell
all lowercase.

This is making the same changes as https://github.com/docker/docker.github.io/pull/8170, but those were applied to the generated YAML files, so will be overwritten with each update of those


thaJeztah:
ping @bermudezmt PTAL

thaJeztah:
LGTM

SharpJuicyBedbug-size_restricted

PR #1887, opened 2019-05-14
Allow vendorers of docker/cli to avoid transitively pulling in a big chunk if k8s too

ijc:
The first commit is a (somewhat unrelated) minor cleanup, the second commit is the interesting one:
```
    Use underlying `NewKubernetesConfig` directly from compose-on-kubernetes.
    
    The comment on `github.com/docker/cli/kubernetes.NewKubernetesConfig` said:
    
        // Deprecated: Use github.com/docker/compose-on-kubernetes/api.NewKubernetesConfig instead
    
    By making this switch in `github.com/docker/cli/context/kubernetes/load.go` we
    break a vendoring chain:
    
    `github.com/docker/cli/cli/command`
    → `vendor/github.com/docker/cli/cli/context/kubernetes/load.go`
      → `vendor/github.com/docker/cli/kubernetes`
         → `github.com/docker/compose-on-kubernetes/api/compose/...`
    
    This means that projects which just want `github.com/docker/cli/cli/command`
    (which is itself pulled in transitively by
    `github.com/docker/cli/cli-plugins/plugin`) which do not themselves need the
    compose-on-kubernetes API avoid a huge pile of transitive dependencies.
    
    On one of my private projects the diff on the vendor dir is:
    
        280 files changed, 21 insertions(+), 211346 deletions(-)
    
    and includes dropping:
    
    * `github.com/docker/compose-on-kubernetes/api/compose/{clone,impersonation}`
    * `github.com/docker/compose-on-kubernetes/api/compose/{v1alpha3,v1beta1,v1beta2,v1beta3}`
    * `github.com/google/btree`
    * `github.com/googleapis/gnostic`
    * `github.com/gregjones/httpcache`
    * `github.com/peterbourgon/diskv`
    * `k8s.io/api/*` (_lots_ of subpackages)
    * `k8s.io/client-go/{discovery,kubernetes/scheme}`
    
    and I've gone from:
    
        $ du -sh vendor/k8s.io/
        8.1M        vendor/k8s.io/
    
    to:
    
        $ du -sh vendor/k8s.io/
        2.1M        vendor/k8s.io/
    
    (overall I went from 36M → 29M of vendor dir for this particular project)
    
    The change to `cli/command/system/version.go` is just for consistency and
    allows us to drop the now unused alias.
    
```

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -14,7 +14,7 @@ import (
 	"github.com/docker/cli/cli/config/configfile"
 	dcontext "github.com/docker/cli/cli/context"
 	"github.com/docker/cli/cli/context/docker"
-	kubcontext "github.com/docker/cli/cli/context/kubernetes"
+	kubecontext "github.com/docker/cli/cli/context/kubernetes"
 	"github.com/docker/cli/cli/context/store"
 	"github.com/docker/cli/cli/debug"
 	cliflags "github.com/docker/cli/cli/flags"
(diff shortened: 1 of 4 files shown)


ijc:
Ideally I'd also like to be able to have callers who don't need/want the kubernetes end point to avoid pulling in `./cli/context/kubernetes` which I think would mean that a vendoring project which doesn't use k8s wouldn't vendor any of it.

I don't think that can be done with the current API though. Best I can see is to allow/require callers to request the kube endpoint be included in the context explicitly (like 3rd party endpoints are). That's an API change though.

PR #1889, opened 2019-05-15
Builder docs update

tonistiigi:
@tiborvass @thaJeztah @andrewhsu @AkihiroSuda


docs/reference/commandline/build.md
```diff
@@ -489,6 +490,49 @@ FROM alpine AS production-env
 $ docker build -t mybuildimage --target build-env .
 ```
 
+### Custom build outputs
+
+By default, a local container image is created from the build result. With the `--output` flag this behavior can be overriden and a custom exporter can be used instead. For example, this is useful for generating local binaries, code generation etc.
```
thaJeztah:
```suggestion
By default, a local container image is created from the build result. With the `--output` flag this behavior can be overriden and a custom exporter can be used instead. For example, custom exporters allow you to export the build artefacts as files on the local filesystem instead of a Docker image, which can be useful for generating local binaries, code generation etc.
```

docs/reference/commandline/build.md
```diff
@@ -489,6 +490,49 @@ FROM alpine AS production-env
...
+
+The value for `--output` is a CSV-formatted string defining the exporter type and options. Currently, `local` and `tar` exporters are supported. The local exporter writes the resulting build files to a directory on the client side. The tar exporter is similar but writes the files as a single tarball.
+
+If no type is specified, the value defaults to the output directory of the local exporter. `-` can be used to write the output tarball to standard output.
+
+Examples:
+
+```
```
thaJeztah:
```suggestion
```bash
```

docs/reference/commandline/build.md
```diff
@@ -489,6 +490,49 @@ FROM alpine AS production-env
...
+docker build -o out .
```
thaJeztah: These examples show what syntax is possible, but don't explain what each form does; perhaps better to split them up into separate blocks, and a short introduction, e.g.

```suggestion

The following example builds an image using the current directory (`.`) as build context,
and exports the files to a directory named `out` in the current directory. If the directory
does not exist, Docker creates the directory automatically:

```bash
docker build -o out .
```

docs/reference/commandline/build.md
```diff
@@ -489,6 +490,49 @@ FROM alpine AS production-env
...
+docker build -o - . > out.tar
+docker build --output type=local,dest=. .
+docker build --output type=tar,dest=out.tar .
+```
+
+A common pattern for exporting only specific files is to do multi-stage builds and to copy the desired files to a new scratch stage with [`COPY --from`](../builder.md#copy).
+
+This feature requires the use of a BuildKit backend. You can either [enable BuildKit](../builder.md#buildkit) or use the [buildx](https://github.com/docker/buildx) plugin which provides more output type options.
```
thaJeztah: Perhaps we should move this to the top of this section, and make it a note;

```Markdown
> **Note**: This feature requires the use of a BuildKit backend. You can either
> [enable BuildKit](../builder.md#buildkit) or use the [buildx](https://github.com/docker/buildx)
> plugin which provides more output type options.

File: docs/reference/commandline/build.md [hunk 5]
```diff
@@ -489,6 +490,49 @@ FROM alpine AS production-env
...
+
+```
+docker build -t myname/myapp --build-arg BUILDKIT_INLINE_CACHE=1 .
+docker push myname/myapp
+
+# in another machine
+docker build --cache-from myname/myapp .
+```
+
+This feature requires the use of BuildKit backend. Previous builder has limited support for reusing cache from prepulled images.
```
thaJeztah: We might be more clear about "which feature" (I think in this case it refers to just the `BUILDKIT_INLINE_CACHE`)?

Perhaps also turn it into a note to make it stand out more

docs/reference/commandline/build.md
```diff
@@ -489,6 +490,49 @@ FROM alpine AS production-env
 $ docker build -t mybuildimage --target build-env .
 ```
 
+### Custom build outputs
```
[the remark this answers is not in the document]
tonistiigi: there is already a pr in to support multiple

docs/reference/commandline/build.md
tonistiigi: this feature = everything in this block

tiborvass:
Ping @thaJeztah @andrewhsu

docs/reference/commandline/build.md
```diff
@@ -323,7 +324,9 @@ Successfully built 99cc1ad10469
 This example shows the use of the `.dockerignore` file to exclude the `.git`
 directory from the context. Its effect can be seen in the changed size of the
 uploaded context. The builder reference contains detailed information on
-[creating a .dockerignore file](../builder.md#dockerignore-file)
+[creating a .dockerignore file](../builder.md#dockerignore-file).
+
+When using BuildKit backend, `.dockerignore` file will be first looked up relative to the Dockerfile name. For example, running `docker build -f myapp.Dockerfile .` will first look for an ignore file from `myapp.Dockerfile.dockerignore`. If such a file can't be found a global `.dockerignore` file is used. Using Dockerfile based `.dockerignore` is useful if a project contains multiple Dockerfiles that expect to ignore different sets of files.
```
adrian-plata: "When using the BuildKit backend, `docker build` searches for a `.dockerignore` file relative to the Dockerfile name. For example, running `docker build -f myapp.Dockerfile .` will first look for an ignore file from `myapp.Dockerfile.dockerignore`. If such a file can't be found, a global `.dockerignore` file is used. Using a Dockerfile based `.dockerignore` is useful if a project contains multiple Dockerfiles that expect to ignore different sets of files."

Cleaned up the text a bit. Double check the "`docker build` searches for . . . " bit for accuracy.

AkihiroSuda:
ping @tonistiigi

thaJeztah:
I had a branch I was working on to address some of my comments; pushed it here: https://github.com/docker/cli/pull/2296

PR #1890, opened 2019-05-16
Dynamically register kubernetes context store endpoint type.

ijc:
Further reduce amount of transitive vendoring required by vendorers and builds on (includes) #1887. With this my private project no longer vendors anything from k8s.io. diffstat (in addition to #1887, not cumulative) is:
```
 407 files changed, 87 insertions(+), 76812 deletions(-)
```

And it dropped transitive vendorings of
```
github.com/docker/{compose-on-kubernetes,spdystream}
github.com/elazarl/goproxy
github.com/evanphx/json-patch
github.com/gogo/protobuf
github.com/golang/groupcache
github.com/googleapis/gnostic
github.com/google/{gofuzz,uuid}
github.com/imdario/mergo
github.com/json-iterator/go
github.com/modern-go/{concurrent,reflect2}
github.com/mxk/go-flowrate
github.com/onsi/gomega
github.com/spf13/pflag
golang.org/x/net
golang.org/x/oauth2
gopkg.in/inf.v0
k8s.io/{api,apimachinery,client-go,klog,kube-openapi,utils}
sigs.k8s.io/yaml
```
for a ~2M reduction in my vendor dir (which is still 27M, but is now dominated by `golang.org/x/*` (mostly `sys`, but `text` too), `github.com/containerd/containerd`, `github.com/docker/docker`).

There's a few things I'm unsure of:
* The mechanism of `EndpointDefaultResolver interface` being implemented by the concrete endpointmeta types feels a bit iffy. Might be better to expand `NamedTypeGetter` to have an additional method alongside the `typeGetter`.
* The special casing (and necessary open coding) of kube logic in `ResolveDefaultContext` is rather gross (a layering violation of sorts). I justify it (sort of) in that while the kubernetes endpoint is notionally optional there is a coupling via the `stackOrchestrator` options which means the core needs to be somewhat aware. At first I had it where it would pass the `stackOrchestrator` as an argument to `kubernetes.ResolveDefault`, I wasn't terribly happy with that either but maybe it is better?
* I really wanted to pass a `Cli` to `ResolveDefaultContext` not a `store.Config` and use `cli.contextStoreConfig` (and `cli.configFile`) but `NewAPIClientFromFlags` doesn't have a cli to use.
* I don't much like exporting  `DefaultContextStoreConfig()` and `ResolveDefaultContext()` just for use in unit tests but I couldn't see another way. Similarly/Relatedly I'd have preferred not to need to split `TestDefaultContextInitializer` into a kube part, but couldn't avoid it while also avoiding the circular imports.


ijc:
FWIW Cumulatively #1887 and this change together result in the vendor for https://github.com/deislabs/duffle/ shrinking by:
```
529 files changed, 79 insertions(+), 253508 deletions(-)
```

Conversely it makes no difference for https://github.com/docker/app/ because that uses `github.com/docker/cli/cli/context/kubernetes` itself and so pulls in all the deps anyhow.

cli/context/kubernetes/load.go
```diff
@@ -61,6 +74,25 @@ func (c *Endpoint) KubernetesConfig() clientcmd.ClientConfig {
...
+// ResolveDefault returns endpoint metadata for the default Kubernetes
+// endpoint, which is derived from the env-based kubeconfig.
+func (c *EndpointMeta) ResolveDefault() (interface{}, *store.EndpointTLSData) {
+	kubeconfig := os.Getenv("KUBECONFIG")
+	if kubeconfig == "" {
+		kubeconfig = filepath.Join(homedir.Get(), ".kube/config")
+	}
+	kubeEP, err := FromKubeConfig(kubeconfig, "", "")
+	if err != nil {
+		return nil, nil
```
ijc: @simonferquel this is refactoring the behaviour of the original code which was to ignore failures here (commonly "file not found") was this a deliberate choice in the original?

I originally had this return the `err` but then things break if you haven't got a `$HOME/.kube` (which happens in the `test` target somewhere). I considered whitelisting some errors ("not found", maybe one or two others, like "not a director" perhaps) but for now settled on just replicating the original behaviour. WDYT?

simonferquel: Yes, we should not break the docker CLI if we have no kubeconfig file, or if it is damanged: in those cases, we should just fail running Kubernetes related commands.

ijc:
I was just experimenting with adding an `error` to `EndpointDefaultResolver`'s return so I could try and push the `wantkubernetesEP` error handling down into the kube endpoint.

With that I'll try moving the docker stuff down too. I had rationalised the current way as Docker being super-duper-core but it'd be nice to be more uniform.

simonferquel:
Just remember that we don't want the CLI error if a kubeconfig file is missing or malformed when we do things like `docker run` (that is the reason that the current default k8s endpoint resolver swallow loading errors)

ijc:
Yep, I've written that as a code comment already ;-)

ijc:
Actually, I think the previous behaviour was to error if the orchestrator was `kubernetes` or `all` even if the specific subcommand being used didn't need it, so I'll be retaining that.

simonferquel:
I also wonder if we could decouple a little bit more the packages by making the defaultendpointresolver interface unexposed and remove references to it from the implementation (replacing that with comments on the implementing function explaining which interface it implements). I haven't digged enough in the dependency graph to know if it would have any impact, but might be interesting.

ijc:
The only reason to expose it is for the:
```go
var _ command.EndpointDefaultResolver = &EndpointMeta{}
```
check. Since a package which implements an endpoint must already import `github.com/docker/cli/cli/command` to get e.g.`RegisterDefaultStoreEndpoints` it doesn't decouple anything to hide it and it removes the possibility of a useful compile time check so I think it might as well be exported.

ijc:
> The special casing (and necessary open coding) of kube logic in ResolveDefaultContext is rather gross (a layering violation of sorts)

I managed to avoid this (actually, it was easy, I'm not sure why I thought I couldn't do it before) in the commit I just pushed.

ijc:
> With that I'll try moving the docker stuff down too. I had rationalised the current way as Docker being super-duper-core but it'd be nice to be more uniform.

This caused an import loop:
```
Building statically linked build/docker-linux-amd64
import cycle not allowed
package github.com/docker/cli/cmd/docker
	imports github.com/docker/cli/cli
	imports github.com/docker/cli/cli-plugins/manager
	imports github.com/docker/cli/cli/command
	imports github.com/docker/cli/cli/context/docker
	imports github.com/docker/cli/cli/command
```

Given the amount of usage of `docker.Endpoint` in `github.com/docker/cli/cli` I think this one might be impossible to avoid.

IOW I think the special case for the docker endpoint defaulting has to stay

ijc:
THe linit failure is
```
 ---> 42ed2232ca41
Successfully built 42ed2232ca41
Successfully tagged cli-linter:25212
read tcp 172.31.0.3:39188->35.237.56.105:2376: read: connection timed out
```
but I don't seem to be allowed to hit retry.

thaJeztah:
> I don't much like exporting DefaultContextStoreConfig() and ResolveDefaultContext() just for use in unit tests but I couldn't see another way.

One possible approach for that would be to add an exported version of those functions in a `_test.go` file, that way it's only available for tests, and not outside of that.

ijc:
> One possible approach for that would be to add an exported version of those functions in a _test.go file, that way it's only available for tests, and not outside of that.

I don't think that works, IIRC, package A's `*_test.go` are not included when building the tests of package B, and that's what we would need here...

I'll rebase now #1887 is in though.

thaJeztah:
Guess the squashing is not a big deal; let's merge as-is 👍

PR #1895, opened 2019-05-20
Add .zip file support for docker context import

goksu:
**- What I did**
I've introduced a way for `docker context import` to accept `.zip` files.

**- How I did it**
I've tried to implement it in a way that it can land itself better for what future additions also bring, with addition of `Cli` enum type. Once we identify which type of import we are trying to get for context, then I go ahead and use the specific import functions, although these functions share as much functionality as possible. I've tried to also make use of an interface of `fileData` to tie these two functionalities together.

**- How to verify it**
I've updated the existing tests to define their import types as `Cli` representing the source. One can test the functionality with using the existing any Docker Context tar file and using the contents to create a `.zip` file. `docker context import` will work the same. I've also added new tests to check if the import functions properly to determine these different import types.

Introduces a `.zip` import support for `docker context`.

### DIFF
--- a/cli/command/context/import.go
+++ b/cli/command/context/import.go
@@ -14,7 +14,7 @@ import (
 func newImportCommand(dockerCli command.Cli) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "import CONTEXT FILE|-",
-		Short: "Import a context from a tar file",
+		Short: "Import a context from a tar or zip file",
 		Args:  cli.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return RunImport(dockerCli, args[0], args[1])
@@ -28,6 +28,7 @@ func RunImport(dockerCli command.Cli, name string, source string) error {
 	if err := checkContextNameForCreation(dockerCli.ContextStore(), name); err != nil {
 		return err
 	}
+
 	var reader io.Reader
 	if source == "-" {
 		reader = dockerCli.In()
@@ -43,6 +44,7 @@ func RunImport(dockerCli command.Cli, name string, source string) error {
 	if err := store.Import(name, dockerCli.ContextStore(), reader); err != nil {
 		return err
 	}
+
 	fmt.Fprintln(dockerCli.Out(), name)
 	fmt.Fprintf(dockerCli.Err(), "Successfully imported context %q\n", name)
 	return nil
--- a/cli/context/store/io_utils.go
+++ b/cli/context/store/io_utils.go
@@ -1,0 +1,29 @@
+package store
+
+import (
+	"errors"
+	"io"
+)
+
+// LimitedReader is a fork of io.LimitedReader to override Read.
+type LimitedReader struct {
+	R io.Reader
+	N int64 // max bytes remaining
+}
+
+// Read is a fork of io.LimitedReader.Read that returns an error when limit exceeded.
+func (l *LimitedReader) Read(p []byte) (n int, err error) {
+	if l.N < 0 {
+		return 0, errors.New("read exceeds the defined limit")
+	}
+	if l.N == 0 {
+		return 0, io.EOF
+	}
+	// have to cap N + 1 otherwise we won't hit limit err
+	if int64(len(p)) > l.N+1 {
+		p = p[0 : l.N+1]
+	}
+	n, err = l.R.Read(p)
+	l.N -= int64(n)
+	return n, err
+}
--- a/cli/context/store/store.go
+++ b/cli/context/store/store.go
@@ -2,12 +2,16 @@ package store
 
 import (
 	"archive/tar"
+	"archive/zip"
+	"bufio"
+	"bytes"
 	_ "crypto/sha256" // ensure ids can be computed
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"path"
 	"path/filepath"
 	"strings"
@@ -259,12 +263,44 @@ func Export(name string, s Reader) io.ReadCloser {
 	return reader
 }
 
+const (
+	maxAllowedFileSizeToImport int64  = 10 << 20
+	zipType                    string = "application/zip"
+)
+
+func getImportContentType(r *bufio.Reader) (string, error) {
+	head, err := r.Peek(512)
+	if err != nil && err != io.EOF {
+		return "", err
+	}
+
+	return http.DetectContentType(head), nil
+}
+
 // Import imports an exported context into a store
 func Import(name string, s Writer, reader io.Reader) error {
-	tr := tar.NewReader(reader)
+	// Buffered reader will not advance the buffer, needed to determine content type
+	r := bufio.NewReader(reader)
+
+	importContentType, err := getImportContentType(r)
+	if err != nil {
+		return err
+	}
+	switch importContentType {
+	case zipType:
+		return importZip(name, s, r)
+	default:
+		// Assume it's a TAR (TAR does not have a "magic number")
+		return importTar(name, s, r)
+	}
+}
+
+func importTar(name string, s Writer, reader io.Reader) error {
+	tr := tar.NewReader(&LimitedReader{R: reader, N: maxAllowedFileSizeToImport})
 	tlsData := ContextTLSData{
 		Endpoints: map[string]EndpointTLSData{},
 	}
+
 	for {
 		hdr, err := tr.Next()
 		if err == io.EOF {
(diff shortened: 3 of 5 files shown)


cli/command/context/import.go
```diff
@@ -28,10 +29,19 @@ func RunImport(dockerCli command.Cli, name string, source string) error {
...
 		return err
 	}
+
 	var reader io.Reader
+	var importType store.ImportType
 	if source == "-" {
 		reader = dockerCli.In()
+		importType = store.Cli
 	} else {
+		if strings.HasSuffix(source, ".zip") {
```
thaJeztah: I left a similar comment on https://github.com/docker/cli/pull/1867/files#r282963749, but wondering if instead we could read the first few bytes from the file to determine the compression from its header (instead of using the file-extension?)

cli/context/store/store.go
```diff
@@ -259,12 +271,111 @@ func Export(name string, s Reader) io.ReadCloser {
...
+	return data, nil
+}
+
+type tarFileData struct {
+	fileData
+	tr io.Reader
+}
+
+func (tfd *tarFileData) get() ([]byte, error) {
+	data, err := ioutil.ReadAll(tfd.tr)
```
thaJeztah: We might want to protect this against huge files being passed and being read into memory (also see https://github.com/docker/cli/pull/1867/files#r285028468)

cli/command/context/import.go
goksu: There is a way to identify the content of a file using `http.DetectContentType`. Do you think that will land itself better for this purpose? I wasn't sure brining in a dependency from `http` package in.

cli/context/store/store.go
goksu: I've added `checkFileSize` for individual files both for `.tar` & `.zip`. Let me know if it seems satisfactory for you, or if you were envisioning a different implementation.

cli/command/context/import.go
thaJeztah: Hm, looks like that is really targeted at downloading from a remote location; https://golang.org/src/net/http/sniff.go?s=646:688#L11

If we know the "magic" header for zips perhaps we can just read some bytes and check for that (but haven't looked into it deeply); @ijc @simonferquel any ideas?

cli/context/store/store.go
thaJeztah: Could use `io.LimitReader`; https://golang.org/pkg/io/#LimitReader

thaJeztah: If we want it to produce an error (instead of silently stopping at the limit), we could use this approach: https://github.com/docker/distribution/pull/2340 (but that may be a bit overcomplicating things)

goksu: Well if we in fact go with the `io.LimitReader` route, I'd have to implement the error handling as well because since reader will fail silently we might get back errors such as `invalid json object` of some sorts that will just result in red herrings.

goksu: I've added `LimtiedReadAll` as a part of `io_utils`, check it out, I think it covers what we are trying to achieve here pretty well. I am going to resolve this as it's outdated, please feel free to add additional comments on the new set of functionality.

cli/command/context/import.go
```diff
@@ -28,10 +29,19 @@ func RunImport(dockerCli command.Cli, name string, source string) error {
 	if err := checkContextNameForCreation(dockerCli.ContextStore(), name); err != nil {
 		return err
 	}
+
 	var reader io.Reader
+	var importType store.ImportType
```
silvin-lubecki: nit: use a `var(....) block`

cli/context/store/io_utils.go
```diff
@@ -0,0 +1,67 @@
...
+// LimitedReadAll will read all content of a limited reader.
+// Should be called with a Reader that's gathered from limitReaderWithErrorHandling
+// Safer than using regular ioutil.ReadAll() that may result in issues on very big files.
+func LimitedReadAll(r io.Reader, limit int64) ([]byte, error) {
+	r = limitReaderWithErrorHandling(r, limit)
+	return ioutil.ReadAll(r)
+}
+
+// ReadExceedsLimitError thrown error type when read exceeds expected limit on reader
+type ReadExceedsLimitError struct {
```
thaJeztah: Looks like this error doesn't have to be exported (there's no special handling for this error); if it's only to be used for the unit test, we could keep it simple and just return `return fmt.Errorf()` or `return error.Errorf()` and just check for the error string?

cli/command/context/import.go
goksu: Thank you, updated!

cli/context/store/io_utils.go
goksu: Changed it back to being an error string. My original thinking was around having strings and possible mistypes, but since it's only in two places and not public properties, it was an overkill.

cli/command/context/import.go
goksu: I have looked into this a bit more, and was not able to come up with a straightforward way to determine just using file headers, especially for `.zip`. Since this is a very common way of doing this I say let's go with it, but please let me know if there's issues with having a check like this in the grand scheme of things that I might be missing.

@ijc @simonferquel — Please let me know if you got something for handling this in another way, would love to change it to that.

thaJeztah:
GitHub is having issues, and looks like they have a full outage of delivering webhooks

cli/command/context/import.go
goksu: I will resolve this for now, happy to re-open if we want to pursue this further.

goksu:
@thaJeztah — All comments are resolved, and checks are passing, let me know about the next steps please, thank you! 💯

cli/context/store/io_utils.go
```diff
@@ -0,0 +1,58 @@
+package store
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+)
+
+// limitedReader a wrapper on io.Reader for error handling of ReadAll
+type limitedReader struct {
```
tiborvass: @goksu I don't understand why you're implementing your own limitReader. Isn't passing a reader to https://godoc.org/io#LimitReader sufficient?

cli/context/store/store.go
```diff
@@ -259,12 +271,113 @@ func Export(name string, s Reader) io.ReadCloser {
...
+		return importTar(name, s, reader)
+	}
+}
+
+func importZip(name string, s Writer, reader io.Reader) error {
+	body, err := ioutil.ReadAll(reader)
+	if err != nil {
+		return err
+	}
+	zr, err := zip.NewReader(bytes.NewReader(body), int64(len(body)))
```
tiborvass: just pass `reader` to zip.NewReader

cli/context/store/store.go
```diff
@@ -259,12 +271,113 @@ func Export(name string, s Reader) io.ReadCloser {
...
+		Endpoints: map[string]EndpointTLSData{},
+	}
+
+	for _, zf := range zr.File {
+		if zf.FileInfo().IsDir() {
+			// skip this entry, only taking files into account
+			continue
+		}
+
+		fd := newZipFileData(zf)
```
thaJeztah: Sorry, I've been very busy, so didn't find time to look closely and comment, but this stood out to me: here we're protecting against a big file _inside_ the ZIP (same as for `tar` below). While it's definitely that someone with ill intent crafts a Zip file that's small, but contains a file that expands to a very large file, I think the main (or _only_) thing we should be concerned about is reading the zip/tar _itself_.

So, instead we could use a limitReader for

```go
body, err := ioutil.ReadAll(reader)
```

(to protect against someone accidentally doing `docker context import ./game-of-thrones-season-1-9.zip`)

And here just read the file as usual. Doing so would probably also reduce the complexity (we wouldn't need the `newZipFileData` and `newTarFileData`

cli/context/store/store.go
```diff
@@ -259,12 +271,113 @@ func Export(name string, s Reader) io.ReadCloser {
...
+	switch importType {
+	case Zip:
+		return importZip(name, s, reader)
+	default: // Tar, Cli
+		return importTar(name, s, reader)
+	}
+}
+
+func importZip(name string, s Writer, reader io.Reader) error {
+	body, err := ioutil.ReadAll(reader)
```
thaJeztah: This would thus use a limitReader then

cli/context/store/io_utils.go
goksu: Hey Tibor, passing the reader through LimitReader is in fact sufficient. But, there was an earlier comment on the PR from @thaJeztah indicating the fact that the error handling on the reader coming back from `LimitReader` falling short on signifying to the user that the failure is in fact from file size being more than allowed amount.

So instead of failing silently with an `EOF`, I've went with this route per recommendation. If we think that is an overkill, then I can happily revert that section and just wrap the reader with a `LimitReader`.

thaJeztah: The custom version was to allow producing a more useful error message https://github.com/docker/cli/pull/1895#discussion_r286110269

thaJeztah: Ah, sorry; our comments crossed

cli/context/store/store.go
```diff
@@ -259,12 +271,113 @@ func Export(name string, s Reader) io.ReadCloser {
...
+	Zip
+)
+
 // Import imports an exported context into a store
-func Import(name string, s Writer, reader io.Reader) error {
+func Import(name string, s Writer, reader io.Reader, importType ImportType) error {
+	switch importType {
+	case Zip:
+		return importZip(name, s, reader)
+	default: // Tar, Cli
```
tiborvass: Why have both Cli and Tar if they have the same behavior?

goksu: To be honest with you, I've struggled to choose one over the other as just being the only type. Was worried that it might have cause more confusion. Would you recommend just getting rid of one of those import types, if yes, which one? Or should there be only `importType.Zip` and rest is the default value.

cli/context/store/io_utils.go
goksu: @tiborvass — With the above comments giving you more context, are you okay with keeping the implementation as is, or would you still recommend the simpler route of just using the `LimitReader`.

thaJeztah:
I took a quick stab at making some of the changes that were suggested here; pushed it as a PR, but feel free to include it here if it looks good: https://github.com/docker/cli/pull/1908

goksu:
@thaJeztah — The changes look good, I'll bring them in, just one question. I've noticed that you've used the limited reader for the file itself, but not the individual archived files. Is that on purpose, or should I also use limited reader for those files too when brining the changes in. 

Thank you for your help, let me know.

thaJeztah:
I left them out for the individual files as that would be a bit of a corner case, but I'm not against doing it for both

cli/context/store/store.go
goksu: With @thaJeztah's great direction on his PR, I've made the same updates here. We don't need the `imporType` enum anymore, it's way more cleaner.

goksu:
@thaJeztah — Following your very helpful PR, I've implemented all those changes here as well. Was not able to cherry-pick across forks, but all updates you made are here as well. I've added a test to check the content type generation & added limited reader usage for individual Zipped files. 

I will keep an eye on the CI builds, I am hoping all will pass and will land this PR for merging possibilities well. Thank you!

thaJeztah:
Thanks! Could you also squash some (or all) of the commits?

thaJeztah:
Ideally we'd have a test as well to try importing from a ZIP

goksu:
Squashed commits, added tests for specifically testing `.zip` & CI is looking good, I think we are almost there with the PR.

Let me know, thank you!

thaJeztah:
@tiborvass @silvin-lubecki PTAL

cli/context/store/io_utils.go
```diff
@@ -0,0 +1,56 @@
...
+	r     io.Reader
+	limit int64
+	n     int64
+	err   error
+}
+
+// newLimitedReader will result in a limited reader with defined byte limit.
+// This basically extends io.LimitReader with proper errors as io.LimitReader only errors with EOF.
```
tiborvass: io.LimitReader does not only error with EOF. https://golang.org/src/io/io.go#L448

I would much rather use io.LimitReader. Unless I'm misunderstanding something.

tiborvass: OK I understand now, what you want is https://golang.org/pkg/net/http/#MaxBytesReader but with `io` instead of `http`.

goksu: I think I wasn't clear in the comment there. I think the main concern here from @thaJeztah & me was the fact that, if reader hits the limit it won't tell us it's actually over the limit, but instead will just return an `EOF`.

goksu: Ah, yes, the comment in there `returns a non-EOF error for a Read beyond the limit` is basically what we wanted.

tiborvass: Ok so you can simplify this further, because all you need is just a 2 lines change around https://golang.org/src/io/io.go?s=14885:14942#L443
```Go
type LimitedReader struct {
	io.LimitedReader
}

func (l *LimitedReader) Read(p []byte) (n int, err error) {
	if l.N < 0 {
		return 0, errors.New("read limit exceeded")
	}
	if l.N == 0 {
		return 0, io.EOF
	}
	if int64(len(p)) > l.N {
		p = p[0:l.N]
	}
	n, err = l.R.Read(p)
	l.N -= int64(n)
	return
}
```

This should possibly in its own ioutils package, although I see we have on in docker/docker/pkg/ioutils.
Maybe we keep it here for now (with the risk of forgetting about it forever).

tiborvass: You don't need limitedReadAll. Just call ioutil.ReadAll inline.

goksu: I've implemented the `Read` as discussed, and called `LimitedReader` inline from all the places used for context import. I'll resolve this conversation just for the sake of the iteration causing the code being outdated. Please feel free to open another thread if need be.

thaJeztah:
Got some failure in CI;

```
=== FAIL: cli/command/context TestCreateFromContext/no-override (600.02s)
SIGQUIT: quit
PC=0x405b9a m=0 sigcode=0

goroutine 20 [running]:
runtime.chanrecv(0xc0004ca600, 0x0, 0x0, 0x0)
	/usr/local/go/src/runtime/chan.go:456 +0x6a fp=0xc0003357d0 sp=0xc000335740 pc=0x405b9a
runtime.selectnbrecv(0x0, 0xc0004ca600, 0xc00003e050)
	/usr/local/go/src/runtime/chan.go:636 +0x3a fp=0xc000335800 sp=0xc0003357d0 pc=0x40644a
io.(*pipe).Read(0xc0002a20f0, 0xc00033a000, 0x1000, 0x1000, 0x0, 0x11a8fa0, 0xc00003e050)
	/usr/local/go/src/io/pipe.go:45 +0x4d fp=0xc0003358b8 sp=0xc000335800 pc=0x468d9d
io.(*PipeReader).Read(0xc000320868, 0xc00033a000, 0x1000, 0x1000, 0x0, 0x11a8fa0, 0xc00003e050)
	/usr/local/go/src/io/pipe.go:127 +0x4c fp=0xc000335900 sp=0xc0003358b8 pc=0x4696ec
bufio.(*Reader).Read(0xc00034e2a0, 0xc000385468, 0x1f9, 0x200, 0x0, 0x11a8fa0, 0xc00003e050)
	/usr/local/go/src/bufio/bufio.go:223 +0x23e fp=0xc000335958 sp=0xc000335900 pc=0x504d9e
github.com/docker/cli/cli/context/store.(*LimitedReader).Read(0xc00047dc60, 0xc000385468, 0x1f9, 0x200, 0x0, 0x0,
```

goksu:
Pushed a fix for the build. My recent push to fix the linter on empty return statement, `err` was carried as nil, I've properly changed that statement now, tests should pass as expected.

thaJeztah:
And it's green 🎉

Issue #1897, opened 2019-05-21
docker cli experimental flag for plugins


andrewhsu:
Change request...

Have the docker cli experimental flag in `~/.docker/config.json`:
```
{ "experimental": "enabled" }
```

Control whether _experimental_ docker cli plugins are enabled.

This would imply a need for plugins to be marked as _experimental_ if desired.

Related: #1564

andrewhsu:
After going over [feedback from implementation PR](https://github.com/docker/cli/pull/1898#issuecomment-494865191), I'm putting some constraints on this change request to help with the flow of developers installing and cli plugins...

Requirements:
1. docker-ce packages install cli experimental packages by default but disabled
2. user of docker-ce must explicitly enable experimental plugins that come installed by default
3. user of docker-ce can install a developer build of plugin without having to globally enable all experimental features (this constraint is subtle, but a new one that is needed to help with developer workflow of plugins)

Implementation:
1. if a cli plugin has `-experimental` as a suffix in its binary filename, then only enable the plugin if `~/.docker/config.json` has `experimental` flag set.

In this way, a user of docker-ce can install a `foobar` cli plugin by making sure it does not have `-experimental` in its filename as a suffix and be able to use it straight away.

andrewhsu:
Talked through some of the _implementation_ steps with @tiborvass and having a scan of directories seem the way to go for keeping binaries separate.

So, if the client config `~/.docker/config.json` has `experimental` flag set then the binaries for the plugin will be searched in the following order:
1. `cliPluginsExtraDirs` directories specified in `~/.docker/config.json`
2. the `~/.docker/cli-plugins` directory
3. [defaultSystemPluginDirs](https://github.com/docker/cli/blob/master/cli-plugins/manager/manager_unix.go#L6-L7)
4. `cliPluginsExtraDirs` directories specified in `~/.docker/config.json` with `-experimental` appended
5. the `~/.docker/cli-plugins-experimental` directory
6. [defaultSystemPluginDirs](https://github.com/docker/cli/blob/master/cli-plugins/manager/manager_unix.go#L6-L7) with `-experimental` appended

If the client config `~/.docker/config.json` does not `experimental` flag set, then the search directory list for finding a plugin will just be:
1. `cliPluginsExtraDirs` directories specified in `~/.docker/config.json`
2. the `~/.docker/cli-plugins` directory
[...]

PR #1898, opened 2019-05-21
cli-plugins: add concept of experimental plugin, only enabled in experimental mode

tiborvass:
Closes: #1897

To test, add $(pwd)/build/plugins-linux-amd64 to "cliPluginsExtraDirs" config and run:
```
make plugins
make binary
HELLO_EXPERIMENTAL=1 docker helloworld
```

To show it enabled:
```
HELLO_EXPERIMENTAL=1 DOCKER_CLI_EXPERIMENTAL=enabled docker helloworld
```

### DIFF
--- a/cli-plugins/manager/candidate_test.go
+++ b/cli-plugins/manager/candidate_test.go
@@ -36,9 +36,10 @@ func TestValidateCandidate(t *testing.T) {
 		builtinName  = NamePrefix + "builtin"
 		builtinAlias = NamePrefix + "alias"
 
-		badPrefixPath  = "/usr/local/libexec/cli-plugins/wobble"
-		badNamePath    = "/usr/local/libexec/cli-plugins/docker-123456"
-		goodPluginPath = "/usr/local/libexec/cli-plugins/" + goodPluginName
+		badPrefixPath    = "/usr/local/libexec/cli-plugins/wobble"
+		badNamePath      = "/usr/local/libexec/cli-plugins/docker-123456"
+		goodPluginPath   = "/usr/local/libexec/cli-plugins/" + goodPluginName
+		metaExperimental = `{"SchemaVersion": "0.1.0", "Vendor": "e2e-testing", "Experimental": true}`
 	)
 
 	fakeroot := &cobra.Command{Use: "docker"}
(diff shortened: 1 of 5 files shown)


cpuguy83:
Not sure I understand the plugin code specifically.
I would expect a plugin would be told if experimental mode is enabled or not and let it decide how to handle it?

tiborvass:
@cpuguy83 Hmm, didn't think of that but it sounds weird to me. It'd be functionally the same though: in both cases the plugin decides whether it is experimental or not, in one case it is through a metadata field, the other would be through code.

The idea in the current PR is that if experimental is off on the cli, we do not consider experimental plugins.

cpuguy83:
What if a plugin has some experimental functionality but is not wholly experimental?

cpuguy83:
I guess it's ok to have a plugin which is entirely experimental and not list it at all if so.

cli-plugins/manager/candidate_test.go
```diff
@@ -48,7 +49,7 @@ func TestValidateCandidate(t *testing.T) {
 		},
 	})
 
-	for _, tc := range []struct {
+	for i, tc := range []struct {
```
cpuguy83: It'd be really nice to give these test cases a name and use `t.Run`.
Debugging with an index like this is painful.

tiborvass: Exactly what i was struggling with, hence the `i` that i wanted to remove.

tiborvass:
@cpuguy83 added test names in a separate commit.

cli-plugins/manager/candidate_test.go
```diff
@@ -49,39 +50,43 @@ func TestValidateCandidate(t *testing.T) {
...
+		{name: "empty vendor", c: &fakeCandidate{path: goodPluginPath, exec: true, meta: `{"SchemaVersion": "0.1.0", "Vendor": ""}`}, invalid: "plugin metadata does not define a vendor"},
+		{name: "experimental required", c: &fakeCandidate{path: goodPluginPath, exec: true, meta: `{"SchemaVersion": "0.1.0", "Vendor": "e2e-testing", "Experimental": true}`}, invalid: "requires experimental enabled"},
 		// This one should work
-		{c: &fakeCandidate{path: goodPluginPath, exec: true, meta: `{"SchemaVersion": "0.1.0", "Vendor": "e2e-testing"}`}},
+		{name: "valid", c: &fakeCandidate{path: goodPluginPath, exec: true, meta: `{"SchemaVersion": "0.1.0", "Vendor": "e2e-testing"}`}},
+		{name: "valid + allowing experimental", c: &fakeCandidate{path: goodPluginPath, exec: true, meta: `{"SchemaVersion": "0.1.0", "Vendor": "e2e-testing"}`, allowExperimental: true}},
+		{name: "experimental + allowing experimental", c: &fakeCandidate{path: goodPluginPath, exec: true, meta: `{"SchemaVersion": "0.1.0", "Vendor": "e2e-testing", "Experimental": true}`, allowExperimental: true}},
 	} {
-		p, err := newPlugin(tc.c, fakeroot)
+		p, err := newPlugin(tc.c, fakeroot, tc.c.allowExperimental)
```
cpuguy83: Can do `t.Run(tc.name, func(t *testing.T) { // the test }` and avoid annotating each error message.

tiborvass: Fixed. Thanks!

tiborvass:
@cpuguy83 it's part of the tests.

cpuguy83:
My bad! Overlooked it.

cli-plugins/manager/plugin.go
```diff
@@ -33,7 +33,9 @@ type Plugin struct {
 // is set, and is always a `pluginError`, but the `Plugin` is still
 // returned with no error. An error is only returned due to a
 // non-recoverable error.
-func newPlugin(c Candidate, rootcmd *cobra.Command) (Plugin, error) {
+//
+// nolint: gocyclo
```
silvin-lubecki: Can be done in a follow-up, but I think this function should be split into pieces as it reached the `go-cyclo` threshold.

cli-plugins/manager/metadata.go
```diff
@@ -22,4 +22,6 @@ type Metadata struct {
 	ShortDescription string `json:",omitempty"`
 	// URL is a pointer to the plugin's homepage.
 	URL string `json:",omitempty"`
+	// Experimental specifies whether the plugin is experimental
```
silvin-lubecki: nit: Add more context? Something like "and then won't be displayed unless the CLI has experimental enabled" ?

cli-plugins/manager/manager.go
```diff
@@ -27,10 +28,23 @@ func (e errPluginNotFound) Error() string {
...
 }
 
+type errPluginRequireExperimental string
+
+// Note: errPluginRequireExperimental implements notFound so that the plugin
+// is skipped when listing the plugins.
+func (e errPluginRequireExperimental) NotFound() {}
+
+func (e errPluginRequireExperimental) Error() string {
+	return fmt.Sprintf("plugin candidate %q: requires experimental enabled", string(e))
```
silvin-lubecki: `cli experimental` instead?

ijc:
Actual code changes look fine to me. I'd suggest adding some variants to the tests under `e2e/cli-plugins/` to ensure that the experimental plugins appear in the right places at the right time wrt the various variations on `docker help` `docker --help`, in particular I think we want to avoid having to have people remember to rerun the test suit with the envvar set and the `help`/`--help` stuff has been tricky to get right.

Or maybe rather than just doubling ~every test in that suite just add another stunt plugin as `e2e/cli-plugins/experimental` (using `e2e/cli-plugins/plugins/nopersistentprerun` as the ~simplest template rather than the `hello` one) and then update the various golden files to accommodate that showing up in the various places it will show up. That will mean splitting some of the `*.golden` files into two versions and conditionally adding `-experimental` to the filename in the `golden.Assert()` calls (perhaps best done with a helper/wrapper?)

Even with the stunt plugin I think you want a new case which does `docker experimental help` and `docker help experimental` (given the stunt plugin is `docker-experimental`) since that presumably needs to DTRT...

silvin-lubecki:
Linter is complaining
```
cli-plugins/manager/candidate_test.go:76:157:warning: unknown field allowExperimental in struct literal (gosimple)
cli/command/formatter/reflect_test.go:15:34:warning: nolint directive did not match any issue (nolint)
cli-plugins/manager/candidate_test.go:76:157:warning: unknown field allowExperimental in struct literal (unused)
```

silvin-lubecki:
Some tests are failing:
```
# github.com/docker/cli/cli-plugins/manager [github.com/docker/cli/cli-plugins/manager.test]
cli-plugins/manager/candidate_test.go:76:157: unknown field 'allowExperimental' in struct literal of type struct { name string; c *fakeCandidate; err string; invalid string }
cli-plugins/manager/candidate_test.go:77:127: unknown field 'allowExperimental' in struct literal of type struct { name string; c *fakeCandidate; err string; invalid string }
```

tiborvass:
@silvin-lubecki fixed. @ijc thanks for the pointers, I'll do that in a follow up.

silvin-lubecki:
ping @thaJeztah

tonistiigi:
To me, it makes much more sense if there is a separate experimental directory (or a naming scheme) for experimental plugins instead of a metadata command.

Plugins have their own versioning/revisions that clearly shows in the output. Eg. I have no desire to mark buildx experimental. If you build it or pull a binary from github there is no need to punish the user so that they need to get all the other experimental things as well and use environment the few other users are using. Adding a plugin does not make the rest of the docker experience more unstable and having buildx with its `0.x` version needing to modify the configuration while a random hello-world plugin does not, does not make sense. In practice, this means that people would just start to call `buildx` directly and not install it as a plugin at all, leading to fragmentation.

Things are different when a plugin is deployed together with Docker CE/EE suites and has a different compatibility guarantees. Then it does makes sense to not enable them automatically. And for that these suites should install the plugins in a way that only enables them when the global experimental mode is on. Eg. by putting them in a separate directory where cli looks for them or naming them `docker-buildx-experimental`.

andrewhsu:
Talked with @tonistiigi about his https://github.com/docker/cli/pull/1898#issuecomment-494865191 and i've updated the issue proposal for change request with details on the implementation https://github.com/docker/cli/issues/1897#issuecomment-494937091 to accommodate for a user installing a plugin explicitly. cc @tiborvass

tiborvass:
I opened https://github.com/docker/cli/pull/1902.

tiborvass:
I think @tonistiigi's points are valid and we should still consider the alternative I opened in #1902, but it turns out it requires more work on Desktop and it's blocking an RC release, so I'm advocating for merging this to unblock.

PR #1907, opened 2019-05-24
[19.03 backport] Dynamically register kubernetes context store endpoint type.

silvin-lubecki:
Backport of #1890 for 19.03

Further reduce amount of transitive vendoring required by vendorers and builds on (includes) #1887. With this my private project no longer vendors anything from k8s.io. diffstat (in addition to #1887, not cumulative) is:
```
 407 files changed, 87 insertions(+), 76812 deletions(-)
```

And it dropped transitive vendorings of
```
github.com/docker/{compose-on-kubernetes,spdystream}
github.com/elazarl/goproxy
github.com/evanphx/json-patch
github.com/gogo/protobuf
github.com/golang/groupcache
github.com/googleapis/gnostic
github.com/google/{gofuzz,uuid}
github.com/imdario/mergo
github.com/json-iterator/go
github.com/modern-go/{concurrent,reflect2}
github.com/mxk/go-flowrate
github.com/onsi/gomega
github.com/spf13/pflag
golang.org/x/net
golang.org/x/oauth2
gopkg.in/inf.v0
k8s.io/{api,apimachinery,client-go,klog,kube-openapi,utils}
sigs.k8s.io/yaml
```
for a ~2M reduction in my vendor dir (which is still 27M, but is now dominated by `golang.org/x/*` (mostly `sys`, but `text` too), `github.com/containerd/containerd`, `github.com/docker/docker`).

There's a few things I'm unsure of:
* The mechanism of `EndpointDefaultResolver interface` being implemented by the concrete endpointmeta types feels a bit iffy. Might be better to expand `NamedTypeGetter` to have an additional method alongside the `typeGetter`.
* The special casing (and necessary open coding) of kube logic in `ResolveDefaultContext` is rather gross (a layering violation of sorts). I justify it (sort of) in that while the kubernetes endpoint is notionally optional there is a coupling via the `stackOrchestrator` options which means the core needs to be somewhat aware. At first I had it where it would pass the `stackOrchestrator` as an argument to `kubernetes.ResolveDefault`, I wasn't terribly happy with that either but maybe it is better?
* I really wanted to pass a `Cli` to `ResolveDefaultContext` not a `store.Config` and use `cli.contextStoreConfig` (and `cli.configFile`) but `NewAPIClientFromFlags` doesn't have a cli to use.
* I don't much like exporting  `DefaultContextStoreConfig()` and `ResolveDefaultContext()` just for use in unit tests but I couldn't see another way. Similarly/Relatedly I'd have preferred not to need to split `TestDefaultContextInitializer` into a kube part, but couldn't avoid it while also avoiding the circular imports.


thaJeztah:
@silvin-lubecki can you try force-pushing again? I think the `doc-signed` check got confused because it was originally opened against the wrong branch 🤔

albers:
As the PR does not contain changes to bash completion any more, I removed myself from the reviewers.

PR #1908, opened 2019-05-25
Add .zip file support for docker context import (alternative)

thaJeztah:
Quick stab at making some changes that were suggested in the review of https://github.com/docker/cli/pull/1895

This will need squashing, and we'll likely want a test (could be in a follow-up)

**- What I did**
I've introduced a way for `docker context import` to accept `.zip` files.

**- How I did it**
I've tried to implement it in a way that it can land itself better for what future additions also bring, with addition of `Cli` enum type. Once we identify which type of import we are trying to get for context, then I go ahead and use the specific import functions, although these functions share as much functionality as possible. I've tried to also make use of an interface of `fileData` to tie these two functionalities together.

**- How to verify it**
I've updated the existing tests to define their import types as `Cli` representing the source. One can test the functionality with using the existing any Docker Context tar file and using the contents to create a `.zip` file. `docker context import` will work the same. I've also added new tests to check if the import functions properly to determine these different import types.

Introduces a `.zip` import support for `docker context`.

### DIFF
--- a/cli/context/store/store.go
+++ b/cli/context/store/store.go
@@ -3,6 +3,7 @@ package store
 import (
 	"archive/tar"
 	"archive/zip"
+	"bufio"
 	"bytes"
 	_ "crypto/sha256" // ensure ids can be computed
 	"encoding/json"
@@ -10,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"path"
 	"path/filepath"
 	"strings"
@@ -84,16 +86,6 @@ type EndpointTLSData struct {
 // ContextTLSData represents tls data for a whole context
 type ContextTLSData struct {
 	Endpoints map[string]EndpointTLSData
-}
-
-// Push will add or update endpoint files record on tls data for the context
-func (tlsData *ContextTLSData) Push(endpointName string, fileName string, data []byte) {
-	if _, ok := tlsData.Endpoints[endpointName]; !ok {
-		tlsData.Endpoints[endpointName] = EndpointTLSData{
-			Files: map[string][]byte{},
-		}
-	}
-	tlsData.Endpoints[endpointName].Files[fileName] = data
 }
 
 // New creates a store from a given directory.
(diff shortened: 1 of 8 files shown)


cli/context/store/store.go
```diff
@@ -259,9 +263,27 @@ func Export(name string, s Reader) io.ReadCloser {
 	return reader
 }
 
+const maxFileSize int64 = 1 << 20
+
 // Import imports an exported context into a store
 func Import(name string, s Writer, reader io.Reader) error {
-	tr := tar.NewReader(reader)
+	// need a buffered reader to read the header without advancing the reader
+	r := bufio.NewReader(reader)
```
thaJeztah: Decided to go with adding detection here; it adds a bit of complexity here, but by doing it here, the `Import()` signature remains unmodified. Also, by detecting, _piping_ a ZIP works (whereas previously it required a file as the extension was used).

cli/context/store/store.go
```diff
@@ -282,37 +304,108 @@ func Import(name string, s Writer, reader io.Reader) error {
...
+				return err
 			}
-			tlsData.Endpoints[endpointName].Files[fileName] = data
 		}
 	}
 	return s.ResetTLSMaterial(name, &tlsData)
 }
 
+func importZip(name string, s Writer, reader io.Reader) error {
+	body, err := limitedReadAll(reader, maxFileSize)
```
thaJeztah: Moved the limiting here; previously, this would to a `ioutils.ReadAll()`

cli/context/store/store.go
```diff
@@ -282,37 +304,108 @@ func Import(name string, s Writer, reader io.Reader) error {
...
+	tlsData := ContextTLSData{
+		Endpoints: map[string]EndpointTLSData{},
+	}
+	for _, zf := range zr.File {
+		fi := zf.FileInfo()
+		if fi.IsDir() {
+			// skip this entry, only taking files into account
+			continue
+		}
+		if zf.Name == metaFile {
```
thaJeztah: I went for just duplicating this import, instead of trying to abstract it; it felt like the abstraction was adding a bit too much complexity (e.g. the addition of the `fileData` interface, the `newZipFileData` and `newTarFileData` etc.) for a (not too) complicated function. We could probably do a bit more refactoring, but I left that for later if we want to.

cli/context/store/store.go
```diff
@@ -282,37 +304,108 @@ func Import(name string, s Writer, reader io.Reader) error {
...
+			f, err := zf.Open()
+			if err != nil {
+				return err
+			}
+			data, err := ioutil.ReadAll(f)
```
thaJeztah: This _could be_ `limitedReadAll`, but that would only be for really corner cases (specially crafted ZIP containing a sparse-file, e.g.)

cli/context/store/store.go
```diff
@@ -282,37 +304,108 @@ func Import(name string, s Writer, reader io.Reader) error {
...
+
+func importEndpointTLS(tlsData *ContextTLSData, path string, data []byte) error {
+	parts := strings.SplitN(strings.TrimPrefix(path, "tls/"), "/", 2)
+	if len(parts) != 2 {
+		// TLS endpoints require archived file directory with 2 layers
+		// i.e. tls/{endpointName}/{fileName}
+		return errors.New("archive format is invalid")
+	}
+
+	epName := parts[0]
```
thaJeztah: Removed the `tlsData.Push()` and inlined it here. `Push()` was effectively only used here (during import), so there was no need yet to add it as a new "interface"

cli/context/store/store.go
goksu: This and usage of the `DetectContentType` looks good to me.

cli/context/store/store.go
goksu: The main reason for the abstraction was in fact removing the code duplication, when I was doing it I was envisioning having multiples of different file types, but as we moved along, it's clear that it's very unlikely that we will get support in for more than tar & zip, thus I am good with removing the abstracted layer. 

As you said, if we end up needing more and more file types, then probably would be the time to re-factor and create that additional layer.

cli/context/store/store.go
goksu: I've mentioned this specific section on the other PR, I am still wondering if adding limited reader for all of these is the safest way to go, since we already have the interface implemented.

cli/context/store/store.go
goksu: The odd thing here for me was the function returning `nil`, and creating a side-affect by changing content of the arguments at the same time. I've tried to change this function to be pure, and that's why I've added the `Push`, but I am guessing it added more complexity than it should've.

cli/context/store/store.go
thaJeztah: Yes, it's a pity it wasn't easy to reuse the code, due to reading, and iterating over the files in Tar and Zip archives is quite different

cli/context/store/store.go
thaJeztah: I initially had it here, but removed it; I'm fine with having it in both places (whole file, and individual files reading from the Zip)

For the tar, we won't need it, as it's not compressed, so if the file as a whole is within limits, files inside the tar should be as well

thaJeztah:
closing, because https://github.com/docker/cli/pull/1895 was merged

PR #1914, opened 2019-05-30
container start wrong error check

rdallman:
noticed that when checking for if the user has entered the escape sequence
after starting the container, the code was checking that the error from
inspect was a term.EscapeError rather than the error from attach in order to
mask the error returned to a nil one - I think this is the error that was
intended to be checked here.

I did not add any tests, sorry, and I'm just gonna go with the social experiment to see
if I can get away with it first ;)

### DIFF
--- a/cli/command/container/start.go
+++ b/cli/command/container/start.go
@@ -154,7 +154,7 @@ func runStart(dockerCli command.Cli, opts *startOptions) error {
 			}
 		}
 		if attachErr := <-cErr; attachErr != nil {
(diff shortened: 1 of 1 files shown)


hc0503 (approved):

thaJeztah:
Looks like this was fixed in https://github.com/docker/cli/commit/9afeb6f4329b8cd6bca15fee5e5b25ade3bc89e6, and updated in https://github.com/docker/cli/commit/056e314645801598f80898029c5d7cbf9f751ec8

PR #1917, opened 2019-06-04
REMOVE MULTIPLE CREDENTIAL ERASES

### LINKED ISSUE #204: docker does not gracefully logout for non-default registry
We are developer an [ACR credential helper](https://github.com/Azure/acr-docker-credential-helper). The issue was discovered during our test on os X.

We ran
1. `az acr login -name <our registry>`
2. `az acr logout`

The log out succeeded and the credentials were removed from osxkeychain. However, We do see the following output

> WARNING: could not erase credentials: error erasing credentials - err: exit status 1, out: 'The specified item could not be found int the keychain'

After looking into the docker code. We realized this is not an ACR specific issue. This issue should repro on any non-default registry if osxkeychain was used in cred store .

Problem with the code:
https://github.com/docker/cli/blob/master/cli/command/registry/logout.go#L42-L53

In the case where `hostnameAddress` and `serverAddress` are the same, we would have duplicated entries in  `regsToTry`. Assuming you can find `serverAddress` in the config file, `Erase()` will be called twice for this `serverAddress`. Cred stores such as file store or WinCred does not complain if you try to remove an entry that does not exist, but osxkeychain would.

Rapix-x:
**- What I did**
Fixed the problem of `docker logout` trying to erase credentials multiple times, which leads to an error when credentials no longer exist. This happened with the macOS keychain as well as with pass on Ubuntu 18.04. Fixes #204 

**- How I did it**
When starting to arrange the server addresses to be checked, the program initialized the slice (regsToTry) with the original server address. If this address is already blank (without any prefix or suffix, e.g. "8.8.8.8"), it is inserted twice in the slice. I simply removed the original server address from the initialization of the slice.

**- How to verify it**
Since it is pure string manipulation, it is pretty obvious. However, all server addresses are already cleaned up and reduced to the host name, when registering the credentials on `docker login`. Therefore, uncleaned server addresses shall not occur and they don't need to be checked/erased on logout.

Remove duplicate credential erases on `docker logout`.

Rottnest-Island-Wildlife-Quokka-Australia-Animal-2676171

### DIFF
--- a/cli/command/registry/logout.go
+++ b/cli/command/registry/logout.go
@@ -31,24 +31,25 @@ func NewLogoutCommand(dockerCli command.Cli) *cobra.Command {
 
 func runLogout(dockerCli command.Cli, serverAddress string) error {
 	ctx := context.Background()
-	var isDefaultRegistry bool
-
-	if serverAddress == "" {
-		serverAddress = command.ElectAuthServer(ctx, dockerCli)
-		isDefaultRegistry = true
-	}
 
 	var (
-		loggedIn        bool
-		regsToLogout    []string
 		hostnameAddress = serverAddress
-		regsToTry       []string
+		loggedIn        bool     // is set later, when checking for credentials
+		regsToTry       []string // is set based on the type of registry
 	)
-	if !isDefaultRegistry {
+
+	// differentiate between default und private registry
+	if serverAddress == "" {
+		// if no server address given, handle case for default registry
+		serverAddress = command.ElectAuthServer(ctx, dockerCli)
+		regsToTry = []string{serverAddress}
+	} else {
+		// if server address given, handle case for private registry
 		hostnameAddress = registry.ConvertToHostname(serverAddress)
+
 		// the tries below are kept for backward compatibility where a user could have
 		// saved the registry in one of the following format.
-		regsToTry = append(regsToTry, hostnameAddress, "http://"+hostnameAddress, "https://"+hostnameAddress)
+		regsToTry = []string{hostnameAddress, "http://" + hostnameAddress, "https://" + hostnameAddress}
 	}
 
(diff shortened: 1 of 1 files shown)


cli/command/registry/logout.go
```diff
@@ -42,7 +42,7 @@ func runLogout(dockerCli command.Cli, serverAddress string) error {
 		loggedIn        bool
 		regsToLogout    []string
 		hostnameAddress = serverAddress
-		regsToTry       = []string{serverAddress}
+		regsToTry       []string
```
thaJeztah: Looks like this means that in the `isDefaultRegistry` situation we won't do any logging out, correct?

Rapix-x: You're right. Thanks for pointing this out! I handled the default registry case in my new commits.

Rapix-x:
@thaJeztah After you pointed out that the case with the default registry is not handled, I observed that the only distinction between the default and a private registry is made, when checking the server address for an empty string. I rewrote most of the function accordingly. See the commit messages for further details.
Any thoughts?

Furthermore, I do have a question. How should we handle errors in the runLogout func?
The function signature states that an error is returned. Unfortunately, at any point in the code there is only returned nil. Appearing errors are directly logged to the console instead of being returned.

cli/command/registry/logout.go
```diff
@@ -31,46 +31,44 @@ func NewLogoutCommand(dockerCli command.Cli) *cobra.Command {
 
 func runLogout(dockerCli command.Cli, serverAddress string) error {
 	ctx := context.Background()
-	var isDefaultRegistry bool
 
+	var (
+		loggedIn  bool     // is set later, when checking for credentials
```
silvin-lubecki: nit: can you move this var declaration closer to where it is actually used?

PR #1940, opened 2019-06-11
Stack: Support cap_add, cap_drop and privileged on services

olljanat:
**- What I did**
Add support for cap_add, cap_drop and privileged options in services with stack deploy

Fourth step to address moby/moby#25885

**- How I did it**
- Bump moby and swarmkit (not to latest because it looks to contain breaking changes but just enough to get this one working).
- Included logic wich converts cap_add, cap_drop and privileged to exact capabilities list

**- How to verify it**
Use stack file like this:
```yaml
version: "3.7"
services:
  test:
    image: ollijanatuinen/capsh
    networks:
      - test
    cap_add:
      - NET_ADMIN
    cap_drop:
      - CAP_MKNOD

networks:
  test:
    driver: overlay
```
or this
```yaml
version: "3.7"
services:
  test:
    image: ollijanatuinen/capsh
    networks:
      - test
    privileged: true

networks:
  test:
    driver: overlay
```

Check log with command: `docker logs <container id>` which should print capabilities from inside of container.

image

### DIFF
--- a/cli/compose/convert/service.go
+++ b/cli/compose/convert/service.go
@@ -14,6 +14,7 @@ import (
 	"github.com/docker/docker/api/types/swarm"
 	"github.com/docker/docker/api/types/versions"
 	"github.com/docker/docker/client"
+	"github.com/docker/docker/oci/caps"
 	"github.com/pkg/errors"
 )
 
@@ -117,6 +118,11 @@ func Service(
 		}
 	}
 
+	capabilities, err := convertCapabilities(apiVersion, service.CapAdd, service.CapDrop, service.Privileged)
+	if err != nil {
+		return swarm.ServiceSpec{}, err
+	}
+
 	serviceSpec := swarm.ServiceSpec{
 		Annotations: swarm.Annotations{
 			Name:   name,
@@ -147,6 +153,7 @@ func Service(
 				Isolation:       container.Isolation(service.Isolation),
 				Init:            service.Init,
 				Sysctls:         service.Sysctls,
+				Capabilities:    capabilities,
 			},
 			LogDriver:     logDriver,
 			Resources:     resources,
@@ -674,4 +681,19 @@ func convertCredentialSpec(namespace Namespace, spec composetypes.CredentialSpec
 		return nil, errors.Errorf("invalid credential spec: spec specifies config %v, but no such config can be found", swarmCredSpec.Config)
 	}
 	return &swarmCredSpec, nil
+}
+
+func convertCapabilities(apiVersion string, capAdd, capDrop []string, privileged bool) ([]string, error) {
+	capabilities := []string{}
+	if privileged || len(capAdd) > 0 || len(capDrop) > 0 {
+		if versions.LessThan(apiVersion, "1.41") {
+			return nil, errors.Errorf("Engine version does not support exact list of capabilities")
+		}
+		capabilities, err := caps.TweakCapabilities(caps.DefaultCapabilities(), capAdd, capDrop, nil, privileged)
+		if err != nil {
+			return nil, err
+		}
+		return capabilities, nil
+	}
+	return capabilities, nil
 }
--- a/cli/compose/convert/service_test.go
+++ b/cli/compose/convert/service_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/swarm"
 	"github.com/docker/docker/client"
+	"github.com/docker/docker/oci/caps"
 	"github.com/pkg/errors"
 	"gotest.tools/assert"
 	is "gotest.tools/assert/cmp"
@@ -622,4 +623,58 @@ func TestConvertUpdateConfigParallelism(t *testing.T) {
 		Parallelism: &parallel,
 	})
 	assert.Check(t, is.Equal(parallel, updateConfig.Parallelism))
+}
+
+func TestConvertServiceCapAddAndCapDrop(t *testing.T) {
+	result, err := Service("1.41", Namespace{name: "foo"}, composetypes.ServiceConfig{}, nil, nil, nil, nil)
+	assert.NilError(t, err)
+	assert.Check(t, is.DeepEqual(result.TaskTemplate.ContainerSpec.Capabilities, []string{}))
+
+	result, err = Service("1.40", Namespace{name: "foo"},
+		composetypes.ServiceConfig{CapAdd: []string{"SYS_NICE"}}, nil, nil, nil, nil)
+	assert.Error(t, err, "Engine version does not support exact list of capabilities")
+
+	service := composetypes.ServiceConfig{
+		CapAdd: []string{
+			"SYS_NICE",
+			"CAP_NET_ADMIN",
+		},
+		CapDrop: []string{
+			"CHOWN",
+			"DAC_OVERRIDE",
+			"CAP_FSETID",
+			"CAP_FOWNER",
+		},
+	}
+	expected := []string{
+		"CAP_MKNOD",
+		"CAP_NET_RAW",
+		"CAP_SETGID",
+		"CAP_SETUID",
+		"CAP_SETFCAP",
+		"CAP_SETPCAP",
+		"CAP_NET_BIND_SERVICE",
+		"CAP_SYS_CHROOT",
+		"CAP_KILL",
+		"CAP_AUDIT_WRITE",
+		"CAP_SYS_NICE",
+		"CAP_NET_ADMIN",
+	}
+	result, err = Service("1.41", Namespace{name: "foo"}, service, nil, nil, nil, nil)
+	assert.NilError(t, err)
+	assert.Check(t, is.DeepEqual(result.TaskTemplate.ContainerSpec.Capabilities, expected))
+}
+
+func TestConvertServicePrivileged(t *testing.T) {
+	service := composetypes.ServiceConfig{
+		Privileged: true,
+	}
+
+	result, err := Service("1.40", Namespace{name: "foo"}, service, nil, nil, nil, nil)
+	assert.Error(t, err, "Engine version does not support exact list of capabilities")
+
+	expected := caps.GetAllCapabilities()
+	result, err = Service("1.41", Namespace{name: "foo"}, service, nil, nil, nil, nil)
+	assert.NilError(t, err)
+	assert.Check(t, is.DeepEqual(result.TaskTemplate.ContainerSpec.Capabilities, expected))
 }
--- a/cli/compose/types/types.go
+++ b/cli/compose/types/types.go
@@ -9,8 +9,6 @@ import (
 // UnsupportedProperties not yet supported by this implementation of the compose file
 var UnsupportedProperties = []string{
 	"build",
-	"cap_add",
-	"cap_drop",
 	"cgroupns_mode",
 	"cgroup_parent",
 	"devices",
@@ -21,7 +19,6 @@ var UnsupportedProperties = []string{
 	"mac_address",
 	"network_mode",
 	"pid",
-	"privileged",
 	"restart",
 	"security_opt",
 	"shm_size",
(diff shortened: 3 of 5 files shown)


olljanat:
Looks that those moby / swarmkit bumps generates need to modify existing code so will split them to another PR later (if no one else don't do that before me).

mterron:
Why is this syntax different form the docker-compose syntax? It makes using swarm difficult. We already have a syntax for adding and removing capabilities, can't you use that?

olljanat:
@mterron very valid point. I can investigate that option too. Whole CLI solution is currently bit open that how it should be implemented so all feedback is welcome (discussion is going on https://github.com/moby/moby/issues/25885#issuecomment-501017588 )

PS. Marked this PR to WIP

olljanat:
@mterron now this works with existing cap_add/cap_drop/privileged syntax. Can you test?

@thaJeztah  PTAL

westsouthnight:
Somebody up merge

nurettin:
Is this the roadmap for this feature?
https://github.com/moby/moby/issues/25885#issuecomment-512910922

kolyshkin:
@olljanat this needs a rebase

olljanat:
@kolyshkin for vendor bump yes but can you plz review the actual implementation first?

olljanat:
Rebased. Default capabilities are duplicated here because DefaultCapabilities() -function from github.com/docker/docker/oci cannot be vendored because it would include some other references which does not pass cross compile.

Created https://github.com/moby/moby/pull/40212 which fixes that issue.

cpuguy83:
I'm conflicted on having default caps here.... or even if we reference them from moby it doesn't quite seem right because it is not necessarily in sync with the engine.

olljanat:
@cpuguy83 those default are only used during `docker stack deploy` to set capabilities list for service(s). I now also added logic that capabilities list will be only set if cap_add/cap_drop/privileged is defined on compose file. Other way it will be leave empty so engine defaults will take effect.

Only place where I can see that this would cause issue/confusion for user is if default capabilities list is changed on some future version and same stack is deployed using multiple versions of docker cli.

olljanat:
Updated to use latest version of Moby and also included tests. @thaJeztah PTAL

Zerocjx:
When can we use this feature?

olljanat:
@Zerocjx plz look https://github.com/moby/moby/issues/25885#issuecomment-557790402

cli/compose/convert/service_test.go
```diff
@@ -623,3 +624,48 @@ func TestConvertUpdateConfigParallelism(t *testing.T) {
 	})
 	assert.Check(t, is.Equal(parallel, updateConfig.Parallelism))
 }
+
+func TestConvertServiceCapAddAndCapDrop(t *testing.T) {
+	service := composetypes.ServiceConfig{
```
cpuguy83: Maybe an additional test for no cap's set?

cli/compose/convert/service_test.go
```diff
@@ -623,3 +624,48 @@ func TestConvertUpdateConfigParallelism(t *testing.T) {
...
+	assert.Check(t, is.DeepEqual(result.TaskTemplate.ContainerSpec.Capabilities, expected))
+}
+
+func TestConvertServicePrivileged(t *testing.T) {
+	service := composetypes.ServiceConfig{
+		Privileged: true,
+	}
+
+	expected := caps.GetAllCapabilities()
+	result, err := Service("1.41", Namespace{name: "foo"}, service, nil, nil, nil, nil)
```
alexellis: If you pulled 1.41 our into a const or var then it could self-describe what it is/does. I assume it means version?

cli/compose/convert/service.go
```diff
@@ -675,3 +682,15 @@ func convertCredentialSpec(namespace Namespace, spec composetypes.CredentialSpec
 	}
 	return &swarmCredSpec, nil
 }
+
+func convertCapabilities(capAdd, capDrop []string, privileged bool) ([]string, error) {
```
alexellis: Can you write a test for convertCapabilities in its current state?

Zerocjx:
I replace my `/usr/bin/docker` with https://github.com/olljanat/cli/releases/download/beta1/docker. There is no cap-add not supported message output when I use `docker stack deploy ...`. But the container log shows `nice: cannot set niceness: Permission denied`. 
(I'm using this to deploy an openvswitch container on swarm mode. Just use docker-compose works fine without swarm mode.)
My docker-compose.yaml file config is like this
```yaml
version: "3"
services:
  test:
    image: ****
    cap_add:
      - ALL
      - NET_ADMIN
```

cli/compose/convert/service_test.go
olljanat: API version 1.41 is first one which contains ability to define Capabilities list so that one is called on here: https://github.com/docker/cli/blob/968ce1ae4d45722c6ae70aa1dff6ee28d88e976a/cli/compose/convert/service.go#L58-L60

cli/compose/convert/service_test.go
olljanat: Included

cli/compose/convert/service_test.go
olljanat: Added now also another test with older API version.

cli/compose/convert/service.go
olljanat: Should be now

olljanat:
@Zerocjx sounds that you are not updated Docker engine (dockerd) to master build. If that does not help plz ping me on Docker community Slack (as that sounds more like usage issue than bug).

@cpuguy83 @alexellis Added now also check for API version because those parameters exists already on earlier engine versions but only API version >= 1.40 support them with stack which would be very confusing for users. Also added requested tests.

@cpuguy83 Starting container without no capabilities at all should works just like on `docker run` / `docker-compose` commands where user need specify all default ones to `cap_drop`list. Another options would be introduce some special capability e.g. `NONE` (like there actually is `ALL` already). However I'm not sure if there actually is any real world use case for that scenario? And even if there is probably it would be better to implement it separately?

Zerocjx:
@olljanat thx. After replace the `dockerd` with the master dockerd build, it works.

amadeuszkryze:
Hi, what is the status of cap_add and cap_drop support in docker stack?

olljanat:
@cpuguy83 @alexellis (and maybe @thaJeztah  too? ) please review again after latest changes.

olljanat:
Rebased. @thaJeztah @silvin-lubecki PTAL

olljanat:
ping @thaJeztah @silvin-lubecki

tuxity:
Oh! Exactly what I'm looking for!
I use GDB inside of my docker containers normally and I'm trying to migrate to Swarm, but GDB need STRACE cap to start

How I can test it @olljanat ? I guess I need to manually install this version of CLI and maybe dockerd ?

olljanat:
@tiborvass I can see that you have merged some PRs to here lately. Maybe you can check this one too?

contrem:
Please confirm:

A config like the following results in all capabilities instead of no capabilities:

```yaml
version: "3.8"
services:
    test:
        image: *
        cap_drop:
            - ALL
```

The capabilities array comes back as empty in [service.go line 121](https://github.com/docker/cli/pull/1940/files#diff-dc2678e1e65c05f198d9b84656ed2f00R121).

Which I think is interpreted the same as not specifying any caps (defaulting to all caps) in both the  [swarmkit pb file](https://github.com/docker/swarmkit/blob/master/api/specs.pb.go#L1023) and [docker api container.go](https://github.com/moby/moby/blob/master/api/types/swarm/container.go#L76) because it's empty. They can't distinguish between not specifying caps (no caps array) and drop all caps (empty caps array).

A special NONE value may be required.

olljanat:
@contrem a lot of time has passed after I implemented those so I cannot say sure but my supposition is that "ALL" should invalid value for cap_add/cap_drop on Moby side and give error. If it does not work like that then it is most probably bug on Moby.

If there is some real world use case for special NONE capability IMO it should be implemented as separate PR on moby side as this is part of trying get same logic to docker service which have been years available with docker-compose.

cbrherms:
I'm surprised after all this time this has still yet to be merged.

contrem:
The ALL value is handled in https://github.com/moby/moby/blob/master/oci/caps/utils.go#L120

The behavior I expected would be the same as

`docker run -it --cap-drop all ollijanatuinen/capsh`:

```pre
Current: =
Bounding set =
Securebits: 00/0x0/1'b0
 secure-noroot: no (unlocked)
 secure-no-suid-fixup: no (unlocked)
 secure-keep-caps: no (unlocked)
uid=0(root)
gid=0(root)
groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
```
Instead, the result using the previous config with image `ollijanatuinen/capsh` and `docker stack deploy` is:

```pre
Current: = cap_chown,cap_dac_override,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_net_bind_service,cap_net_raw,cap_sys_chroot,cap_mknod,cap_audit_write,cap_setfcap+eip
Bounding set =cap_chown,cap_dac_override,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_net_bind_service,cap_net_raw,cap_sys_chroot,cap_mknod,cap_audit_write,cap_setfcap
Securebits: 00/0x0/1'b0
 secure-noroot: no (unlocked)
 secure-no-suid-fixup: no (unlocked)
 secure-keep-caps: no (unlocked)
uid=0(root)
gid=0(root)
groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
```

When adding a single cap and dropping all others, it works as expected because the cap array returned is not empty:

```yaml
version: "3.8"
services:
    test:
        image: ollijanatuinen/capsh
        cap_add:
            - chown
        cap_drop:
            - all
```

```pre
Current: = cap_chown+eip
Bounding set =cap_chown
Securebits: 00/0x0/1'b0
 secure-noroot: no (unlocked)
 secure-no-suid-fixup: no (unlocked)
 secure-keep-caps: no (unlocked)
uid=0(root)
gid=0(root)
groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
```

I don't know where the empty cap array issue should be handled, but I do think the `cap_drop all` handling is unexpected behavior.

oramirite:
I am cheering for you all SO HARD on this one. This is such a crucial feature and will unlock so many use-cases for us. Thank you @olljanat for your work championing this feature and I hope we see this in the next release.

tu-nv:
While already  :+1: and ❤️ , my inner still want to post this useless comment to show that I (and many others) really hope this feature will be available soon. I will go back to cheer with many 🎉🎉🎉🎉🎉 when it is finally merged.

georgmay:
I can't believe it's still not merged...

cpuguy83:
I *think* we are going to have to revert these API changes in the engine and swarmkit since the API does not seem to hold up well. This is based on discussions in the maintainers call last Thursday.

Specifically some of the things mentioned early on in the design review here are problematic.
Namely that the client must specify the full list of caps, meanwhile the client does not know actually what caps are available on the engine host.

cpuguy83:
https://github.com/moby/moby/pull/41249 makes the neccessary changes to moby's API for correcting cap add/drop support.

akerouanton:
@olljanat Do you need some help to finish this PR? :slightly_smiling_face:

olljanat:
@akerouanton if you or someone else want take responsibility to finalize this I'm more than happy to give it forward. I did see than PR on moby side was merged but have not found time to investigate what actually is need to be changed here.

thaJeztah:
I'm working on updating the vendored version of moby in https://github.com/docker/cli/pull/2659 (opened as draft as it may need local changes). After that it should be possible to implement the changes here to wire that up

Issue #1854, opened 2019-04-30
19.03 CLI always sets `NetworkingConfig.EndpointsConfig` on any container with a network


wsong:
It appears that https://github.com/docker/cli/pull/1767 changed the behavior of the CLI to always set `NetworkingConfig.EndpointsConfig` for any container created with a `--net` flag. Previously, this was only set when you specified non-default options for the network.

We should probably match the previous behavior if possible.

PR #1963, opened 2019-06-21
Fix advanced options for backward compatibility

thaJeztah:
For backward compatibility: if no custom options are provided for the network,
and only a single network is specified, omit the endpoint-configuration
on the client (the daemon will still create it when creating the container)

This fixes an issue on older versions of legacy Swarm, which did not support
`NetworkingConfig.EndpointConfig`.

This was introduced in 5bc09639cc0aa382fdfcd9ca94d7817c8b3073c0 (#1767)

fixes https://github.com/docker/cli/issues/1854

PR #1967, opened 2019-06-24
[19.03 backport] Fix advanced options for backward compatibility

thaJeztah:
backport of https://github.com/docker/cli/pull/1963 for 19.03

For backward compatibility: if no custom options are provided for the network,
and only a single network is specified, omit the endpoint-configuration
on the client (the daemon will still create it when creating the container)

This fixes an issue on older versions of legacy Swarm, which did not support
`NetworkingConfig.EndpointConfig`.

This was introduced in 5bc09639cc0aa382fdfcd9ca94d7817c8b3073c0 (#1767)

fixes https://github.com/docker/cli/issues/1854

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path"
+	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


thaJeztah:
marking 'WIP' because the backported PR isn't merged yet

Issue #1696, opened 2019-02-27
Windows refusing to load key from delegation.key: private key file delegation.key must not be readable or writable by others


GodofGurus:
**Description**

Can not load delegation key

**Steps to reproduce the issue:**
1.openssl genrsa -out delegation.key 2048
2.openssl req -new -sha256 -key delegation.key -out delegation.csr
3.openssl x509 -req -sha256 -days 365 -in delegation.csr -signkey delegation.key -out delegation.crt
4. C:\test> docker trust key load delegation.key --name tester
Loading key from "delegation.key"...
refusing to load key from delegation.key: private key file delegation.key must not be readable or writable by others

**Describe the results you received:**
refusing to load key from delegation.key: private key file delegation.key must not be readable or writable by others

**Describe the results you expected:**
key to be added

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
docker version
Client: Docker Engine - Community
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        6247962
 Built:             Sun Feb 10 04:12:31 2019
 OS/Arch:           windows/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.2
  API version:      1.39 (minimum version 1.24)
  Go version:       go1.10.6
  Git commit:       6247962
  Built:            Sun Feb 10 04:28:48 2019
  OS/Arch:          windows/amd64
  Experimental:     true
```

**Output of `docker info`:**

```
docker info
Containers: 0
 Running: 0
 Paused: 0
 Stopped: 0
Images: 0
Server Version: 18.09.2
Storage Driver: windowsfilter
 Windows:
Logging Driver: json-file
Plugins:
 Volume: local
 Network: ics l2bridge l2tunnel nat null overlay transparent
 Log: awslogs etwlogs fluentd gelf json-file local logentries splunk syslog
Swarm: inactive
Default Isolation: process
Kernel Version: 10.0 14393 (14393.2791.amd64fre.rs1_release.190205-1511)
Operating System: Windows Server 2016 Datacenter Version 1607 (OS Build 14393.2791)
OSType: windows
Architecture: x86_64
CPUs: 8
Total Memory: 15.92GiB
Name: subzero
ID: IM7O:6MJT:GKDZ:6UHH:URZG:YE24:6VUX:4BPT:U2VM:WGWM:IS23:6WMA
[...]

PR #1968, opened 2019-06-24
Windows: skip permissions check on key

thaJeztah:
fixes https://github.com/docker/cli/issues/1696

This code was attempting to check Linux file permissions to determine if the key was accessible by other users, which doesn't work, and therefore prevented users on Windows to load keys.

Skipping this check on Windows (correspinding tests were already skipped).

```
* Fix docker refusing to load key from delegation.key on Windows
```

### DIFF
--- a/cli/command/trust/key_load.go
+++ b/cli/command/trust/key_load.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @justincormack @riyazdf ptal

thaJeztah:
ping @vdemeester ptal

PR #1973, opened 2019-06-26
Fix Rollback config type interpolation

silvin-lubecki:
**- What I did**

Rollback config type interpolation on fields "parallelism" and "max_failure_ratio" were missing, as it uses the same type as update_config.

It addresses https://github.com/docker/app/issues/559

* Fix Rollback config type interpolation

image


thaJeztah:
ping @vdemeester ptal 🤗

PR #1985, opened 2019-07-04
[19.03 backport] context: produce consistent output on `context create`.

thaJeztah:
backport of https://github.com/docker/cli/pull/1874

Refactor `RunCreate` slightly so that all three paths always produce the same
output, namely the name of the new context of `stdout` (for scripting) and the
success log message on `stderr`.

Validate by extending the existing unit tests to always check the output is as
expected.


kolyshkin:
@silvin-lubecki PTAL

thaJeztah:
Merging this one, because currently, no output is produced at all in some cases, e.g.:

```bash
docker context create --from=default mynewcontext
# no output
```

Issue #1403, opened 2018-10-01
Docker stack deploy do not undersand volumes in windows format


eiximenis:
**Description**

I am using Docker CE For Windows (Version 2.0.0.0-beta1-win75 (19925) and using **docker stack deploy** to deploy a workload against the local k8s cluster.

If I define a volume using "linux like" syntax it works OK:

```
    volumes:     
        -/c/tfs/eShopOnContainers/src/ApiGateways/Mobile.Bff.Shopping/apigw:${ESHOP_OCELOT_VOLUME_SPEC:-/app/configuration}
```

But If I use windows syntax it fails:

```
    volumes:      
        -c:\tfs\eShopOnContainers\src\ApiGateways\Mobile.Bff.Shopping\apigw:${ESHOP_OCELOT_VOLUME_SPEC:-/app/configuration}
```

Then I receive on error:

```
Stack.compose.docker.com "eshop" is invalid: eshop: Invalid value: "null": conversion to kube entities failed: c:\tfs\eShopOnContainers\ApiGateways\Mobile.Bff.Marketing\apigw: only absolute paths can be specified in mount source
```

Note that if instead of `c:\tfs\...` I use something like `\c\tfs\...` (linux-syntax with no colon but with windows backslashes) I receive another error:

```
Stack.compose.docker.com "eshop" is invalid: spec.stack.services[1].volumes[0]: Invalid value: "\\c\\tfs\\eShopOnContainers\\src\\ApiGateways\\Mobile.Bff.Shopping\\apigw": not a valid volume name in Kubernetes: a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
```

This is very unfortunate, because under windows, docker-compose changes the format of Linux paths to windows paths when using "docker-compose config". So, if I have a compose file with one volume:

```
- /c/tfs/eShopOnContainers/ApiGateways/Mobile.Bff.Marketing/apigw:${ESHOP_OCELOT_VOLUME_SPEC:-/app/configuration}
```

The output of  `docker-compose config` is (linux slashes are converted to backslashes):

```
    volumes:
    - \c\tfs\eShopOnContainers\ApiGateways\Mobile.Bff.Marketing\apigw:/app/configuration:rw
```

But this format is not supported by `docker stack deploy`. This prevents me to use `docker-compose config | docker stack deploy -c -`.

**Output of `docker version`:**

```
Client: Docker Engine - Community
[...]

PR #1990, opened 2019-07-09
Detect Windows absolute paths on non-Windows CLI

thaJeztah:
Note that there is an alternative PR in https://github.com/docker/cli/pull/1871, but that PR changes the current behaviour, and might require more discussion. This PR tries to focus on just the basic problem.

addresses https://github.com/docker/cli/issues/1403
addresses https://github.com/moby/moby/issues/33746
fixes https://github.com/moby/moby/issues/34810

When deploying a stack using a relative path as bind-mount source in the compose file, the CLI converts the relative path to an absolute path, relative to the location of the docker-compose file.

This causes a problem when deploying a stack that uses an absolute Windows path, because a non Windows client will fail to detect that the path (e.g. `C:\somedir`) is an absolute path (and not a relative directory named `C:\`).

The existing code did already take Windows clients deploying a Linux stack into account (by checking if the path had a leading slash). This patch adds the reverse, and adds detection for Windows absolute paths on non-Windows clients.

The code used to detect Windows absolute paths is copied from the Golang filepath package;
https://github.com/golang/go/blob/1d0e94b1e13d5e8a323a63cd1cc1ef95290c9c36/src/path/filepath/path_windows.go#L12-L65

### DIFF
--- a/cli/compose/loader/loader_test.go
+++ b/cli/compose/loader/loader_test.go
@@ -985,6 +985,84 @@ services:
 	assert.Error(t, err, `invalid mount config for type "bind": field Source must not be empty`)
 }
 
+func TestLoadBindMountSourceIsWindowsAbsolute(t *testing.T) {
+	tests := []struct {
+		doc      string
+		yaml     string
+		expected types.ServiceVolumeConfig
+	}{
+		{
+			doc: "Z-drive lowercase",
+			yaml: `
+version: '3.3'
+
+services:
+  windows:
+    image: mcr.microsoft.com/windows/servercore/iis:windowsservercore-ltsc2019
+    volumes:
+      - type: bind
+        source: z:\
+        target: c:\data
+`,
+			expected: types.ServiceVolumeConfig{Type: "bind", Source: `z:\`, Target: `c:\data`},
+		},
+		{
+			doc: "Z-drive uppercase",
+			yaml: `
+version: '3.3'
+
+services:
+  windows:
+    image: mcr.microsoft.com/windows/servercore/iis:windowsservercore-ltsc2019
+    volumes:
+      - type: bind
+        source: Z:\
+        target: C:\data
+`,
+			expected: types.ServiceVolumeConfig{Type: "bind", Source: `Z:\`, Target: `C:\data`},
+		},
+		{
+			doc: "Z-drive subdirectory",
+			yaml: `
+version: '3.3'
+
+services:
+  windows:
+    image: mcr.microsoft.com/windows/servercore/iis:windowsservercore-ltsc2019
+    volumes:
+      - type: bind
+        source: Z:\some-dir
+        target: C:\data
+`,
+			expected: types.ServiceVolumeConfig{Type: "bind", Source: `Z:\some-dir`, Target: `C:\data`},
+		},
+		{
+			doc: "forward-slashes",
+			yaml: `
+version: '3.3'
+
+services:
+  app:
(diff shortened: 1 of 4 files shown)


thaJeztah:
ping @silvin-lubecki @vdemeester @ddebroy PTAL

/cc @stevenfollis @david-yu @ivansharamok

thaJeztah:
Ah, booh! The linter doesn't like the Golang standard library;

```
cli/compose/loader/windows_path.go:35::warning: cyclomatic complexity 18 of function volumeNameLen() is high (> 16) (gocyclo)
```

cli/compose/loader/windows_path.go
```diff
@@ -0,0 +1,70 @@
...
+	filePath = filePath[l:]
+	if filePath == "" {
+		return false
+	}
+	return isSlash(filePath[0])
+}
+
+// volumeNameLen returns length of the leading volume name on Windows.
+// It returns 0 elsewhere.
+// nolint: gocyclo
```
thaJeztah: Added this to silence the linter

cli/compose/loader/loader_test.go
```diff
@@ -985,6 +985,84 @@ services:
...
+			doc: "C-drive lowercase",
+			yaml: `
+version: '3.3'
+
+services:
+  windows:
+    image: mcr.microsoft.com/windows/servercore/iis:windowsservercore-ltsc2019
+    volumes:
+      - type: bind
+        source: c:\
```
ddebroy: Just to be safe, can the unit tests use `E:\` or `X:\` as the source path's drive (something other than `C:\`) in some of the cases?

cli/compose/loader/windows_path.go
```diff
@@ -0,0 +1,70 @@
...
+// https://github.com/golang/go/blob/1d0e94b1e13d5e8a323a63cd1cc1ef95290c9c36/src/path/filepath/path_windows.go#L12-L65
+
+import "path"
+
+func isSlash(c uint8) bool {
+	return c == '\\' || c == '/'
+}
+
+// isAbs reports whether the path is a Windows absolute path.
+func isAbs(filePath string) (b bool) {
```
ddebroy: should we have unit tests for `isAbs`?

thaJeztah: All these were copied from the golang source code; I did have a quick peek if I could copy their tests as well, but couldn't find a test for these functions in isolation (at a glance)

cli/compose/loader/loader_test.go
thaJeztah: Yes, I can change that; let me update

cli/compose/loader/windows_path.go
thaJeztah: oh; found where they live; I'll add some tests

thaJeztah:
Per discussion with @silvin-lubecki on slack: I removed the redundant `path.IsAbs` from the `isAbs()` function, and updated the test-file to also include the copyright from upstream

thaJeztah:
Booh; this one starts failing now all of a sudden; https://github.com/docker/cli/pull/1841#issuecomment-484855473

```
10:06:29 === FAIL: e2e/container TestSigProxyWithTTY (6.52s)
10:06:29     proxy_signal_test.go:38: terminating PID 3438
10:06:29     proxy_signal_test.go:42: timeout hit after 5s: expected status exited != running
```

PR #1992, opened 2019-07-10
e2e: use stable-dind image for testing

thaJeztah:
The edge channel is deprecated and no longer updated

### DIFF
--- a/e2e/compose-env.yaml
+++ b/e2e/compose-env.yaml
@@ -5,7 +5,7 @@ services:
     image: 'registry:2'
 
   engine:
-      image: 'docker:${TEST_ENGINE_VERSION:-edge-dind}'
+      image: 'docker:${TEST_ENGINE_VERSION:-stable-dind}'
       privileged: true
       command: ['--insecure-registry=registry:5000']
 


thaJeztah:
@silvin-lubecki @vdemeester PTAL; wasn't sure what we preferred here;

- use `test-dind` (latest "RC"); gives the advantage of testing against latest/greatest, but this image may change at points so possibly cause inconsistent results
- use `stable-dind` same "problem" as above
- pin to a specific version (this PR); downside is that we don't catch regressions, but positive side is that we know what we're testing against

vdemeester:
> @silvin-lubecki @vdemeester PTAL; wasn't sure what we preferred here;
> 
>     * use `test-dind` (latest "RC"); gives the advantage of testing against latest/greatest, but this image may change at points so possibly cause inconsistent results
> 
>     * use `stable-dind` same "problem" as above
> 
>     * pin to a specific version (this PR); downside is that we don't catch regressions, but positive side is that we know what we're testing against

I think safest bet is `stable-dind`. We could have another job `test-dind` that is not a blocker (aka we would see it needs fixing but it wouldn't be required to be green to be able to merge). @thaJeztah wdyt ?

thaJeztah:
I'm good with `stable-dind` (less maintenance) let me update

thaJeztah:
updated 👍

PR #1995, opened 2019-07-10
[19.03 backport] Detect Windows absolute paths on non-Windows CLI

thaJeztah:
backport of https://github.com/docker/cli/pull/1990

Note that there is an alternative PR in https://github.com/docker/cli/pull/1871, but that PR changes the current behaviour, and might require more discussion. This PR tries to focus on just the basic problem.

addresses https://github.com/docker/cli/issues/1403
addresses https://github.com/moby/moby/issues/33746
fixes https://github.com/moby/moby/issues/34810

When deploying a stack using a relative path as bind-mount source in the compose file, the CLI converts the relative path to an absolute path, relative to the location of the docker-compose file.

This causes a problem when deploying a stack that uses an absolute Windows path, because a non Windows client will fail to detect that the path (e.g. `C:\somedir`) is an absolute path (and not a relative directory named `C:\`).

The existing code did already take Windows clients deploying a Linux stack into account (by checking if the path had a leading slash). This patch adds the reverse, and adds detection for Windows absolute paths on non-Windows clients.

The code used to detect Windows absolute paths is copied from the Golang filepath package;
https://github.com/golang/go/blob/1d0e94b1e13d5e8a323a63cd1cc1ef95290c9c36/src/path/filepath/path_windows.go#L12-L65

### DIFF
--- a/cli/compose/loader/loader.go
+++ b/cli/compose/loader/loader.go
@@ -479,12 +479,13 @@ func resolveVolumePaths(volumes []types.ServiceVolumeConfig, workingDir string, 
 		}
 
 		filePath := expandUser(volume.Source, lookupEnv)
-		// Check for a Unix absolute path first, to handle a Windows client
-		// with a Unix daemon. This handles a Windows client connecting to a
-		// Unix daemon. Note that this is not required for Docker for Windows
-		// when specifying a local Windows path, because Docker for Windows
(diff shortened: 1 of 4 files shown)


thaJeztah:
moving this to the 19.03.1 milestone (don't merge yet)

thaJeztah:
this is ok to merge now

PR #2008, opened 2019-07-18
Dockerfile: use GO_VERSION build-arg for overriding Go version

thaJeztah:
This allows overriding the version of Go without making modifications in the
source code, which can be useful to test against multiple versions.

For example:

    make GO_VERSION=1.13beta1 -f docker.Makefile binary


thaJeztah:
ping @silvin-lubecki @vdemeester ptal

/cc @seemethere

thaJeztah:
@silvin-lubecki it's good to have a default in the Dockerfile; Dockerfiles should be "buildable" without having to pass specific build-args

thaJeztah:
That default _could be_ `:latest` of course, but that would make them quite unpredictable

PR #2012, opened 2019-07-18
man/docker-build: fix/expand --build-arg

kolyshkin:
* Explain that variable value is optional and is taken from environment if not set.
* Clarify the meaning of ARG.
* Move example to the end.

### DIFF
--- a/man/docker-build.1.md
+++ b/man/docker-build.1.md
@@ -84,16 +84,19 @@ set as the **URL**, the repository is cloned locally and then sent as the contex
    Add a line to /etc/hosts. The format is hostname:ip.  The **--add-host**
 option can be set multiple times.
 
-**--build-arg**=*variable*
-   name and value of a **buildarg**.
+**--build-arg**=*name*[=*value*]
+   Set a build-time variable. The *name* is the variable name as it appears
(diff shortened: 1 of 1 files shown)


man/docker-build.1.md
```diff
@@ -84,16 +84,19 @@ set as the **URL**, the repository is cloned locally and then sent as the contex
...
-**--build-arg**=*variable*
-   name and value of a **buildarg**.
+**--build-arg**=*name*[=*value*]
+   Set a build-time variable. The *name* is the variable name as it appears
+   in an `ARG` instruction of a Dockerfile. The *value* is optional; if not
+   set, it is taken from the current environment.
+
+   These variables then are accessed like regular environment variables
+   in a `RUN` instruction of the Dockerfile. This mechanism is not meant
+   for passing secret values.
```
thaJeztah: Perhaps we should also mention that if both an `ENV`, and an `ARG` is defined, that the `ARG` takes precedence and will override the value of the `ENV`.


While working on this, could you also update the online reference docs so that the descriptions match? https://github.com/docker/cli/blob/master/docs/reference/commandline/build.md#set-build-time-variables---build-arg

devdreamsolution (approved):

PR #2028, opened 2019-08-01
Sort list of services from swarm stacks using natural sorting

pszczekutowicz:
**- What I did**
I added sorting to services list returned by command `docker stack services` from Swarm.

**- How I did it**
I did it by copying the code for `docker service ls` command.

**- How to verify it**
Execute command `docker stack services <STACK_NAME>` and list of services should be sorted.

Sort list of services from swarm stacks using natural sorting

### DIFF
--- a/cli/command/stack/swarm/services.go
+++ b/cli/command/stack/swarm/services.go
@@ -3,6 +3,7 @@ package swarm
 import (
 	"context"
 	"fmt"
+	"sort"
 
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/service"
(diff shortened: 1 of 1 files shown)


cli/command/stack/swarm/services.go
```diff
@@ -3,6 +3,9 @@ package swarm
 import (
 	"context"
 	"fmt"
+	"sort"
+
+	"vbom.ml/util/sortorder"
```
thaJeztah: nit: we don't group imports in three groups (only two), so this should be grouped (and sorted) with the ones below.

thaJeztah:
ping @silvin-lubecki PTAL

cli/command/stack/swarm/services.go
pszczekutowicz: Changed.

thaJeztah:
oh! thanks for updating; could you squash those two commits?

pszczekutowicz:
ok, all my commits are squashed.

PR #2052, opened 2019-08-16
Skip inspect of builtin networks on service create

trapier:
**- What I did**

Ported #362 to `service create`.  Note this leaves `service update` still affected.  Patching that path required investigation beyond the scope of available time.

**- How I did it**

Copied logic to skip network inspection if the network is built-in.

**- How to verify it**

Build a docker API server that returns a 404 for the "host" and "bridge" networks (such as https://github.com/docker/swarm) and run the following:

```
docker service create --network host nginx
```

Then, verify the deployed task is attached to the "host" network on the target node.

* Service create no longer inspects built-in networks

### DIFF
--- a/cli/command/service/opts.go
+++ b/cli/command/service/opts.go
@@ -613,9 +613,16 @@ func (options *serviceOptions) ToService(ctx context.Context, apiClient client.N
 
 	networks := convertNetworks(options.networks)
 	for i, net := range networks {
-		nwID, err := resolveNetworkID(ctx, apiClient, net.Target)
-		if err != nil {
-			return service, err
+		var nwID string
+		if !container.NetworkMode(net.Target).IsUserDefined() {
+			// Networks that are not user defined always exist on all nodes as
+			// local-scoped networks, so there's no need to inspect them.
+			nwID = net.Target
+		} else {
+			nwID, err = resolveNetworkID(ctx, apiClient, net.Target)
+			if err != nil {
+				return service, err
+			}
 		}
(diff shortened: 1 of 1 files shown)


cli/command/service/opts.go
```diff
@@ -613,9 +613,16 @@ func (options *serviceOptions) ToService(ctx context.Context, apiClient client.N
...
-		nwID, err := resolveNetworkID(ctx, apiClient, net.Target)
-		if err != nil {
-			return service, err
+		var nwID string
+		if !container.NetworkMode(net.Target).IsUserDefined() {
+			// Networks that are not user defined always exist on all nodes as
+			// local-scoped networks, so there's no need to inspect them.
+			nwID = net.Target
+		} else {
+			nwID, err = resolveNetworkID(ctx, apiClient, net.Target)
```
thaJeztah: Looking if we

- can roll-up this logic into `resolveNetworkID`, so that we use the same logic in all places
- should _reverse_ the logic; first try to resolve the ID and only if that fails, fall back to making an exception for a network that is (possibly) a pre-defined network. That way we have a slightly lower risk of excluding a predefined "windows" network on Linux, and vice-versa (we should fix that case though, and make both platforms aware of predefined networks of both, also to prevent (e.g.) a Linux machine creating a `"nat"` network (which is pre-defined on Windows)

thaJeztah: I'm working on a branch to explore those options

devdreamsolution (approved):

PR #2060, opened 2019-08-22
Add option to remove `env_file` entry once it's merged in the `environment` section

ulyssessouza:
This avoids having redundant output when rendering the compose file once again by adding an option to remove the `env_file` from the config when already resolved.

Causes https://github.com/docker/app/issues/495

**- What I did**
Remove `env_file` entries when the option `RemoveEnvFiles` is set to `true`

**- How I did it**
By setting the serviceConfig's `EnvFile` to `nil` when the `RemoveEnvFiles` option is set to `true`

**- How to verify it**
N/A. Since the rendered is still the same

Remove `env_file` entries when the option `RemoveEnvFiles` is set to `true`

adult-brown-mask

### DIFF
--- a/cli/compose/loader/loader.go
+++ b/cli/compose/loader/loader.go
@@ -32,6 +32,14 @@ type Options struct {
 	SkipInterpolation bool
 	// Interpolation options
 	Interpolate *interp.Options
+	// Discard 'env_file' entries after resolving to 'environment' section
+	discardEnvFiles bool
+}
+
+// WithDiscardEnvFiles sets the Options to discard the `env_file` section after resolving to
+// the `environment` section
+func WithDiscardEnvFiles(opts *Options) {
+	opts.discardEnvFiles = true
 }
 
 // ParseYAML reads the bytes from a file, parses the bytes into a mapping
@@ -105,6 +113,11 @@ func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.
 			return nil, err
 		}
 		cfg.Filename = file.Filename
+		if opts.discardEnvFiles {
+			for i := range cfg.Services {
+				cfg.Services[i].EnvFile = nil
+			}
+		}
 
 		configs = append(configs, cfg)
 	}
--- a/cli/compose/loader/loader_test.go
+++ b/cli/compose/loader/loader_test.go
@@ -759,6 +759,38 @@ services:
 	assert.Check(t, is.DeepEqual([]string{"build", "links", "pid"}, unsupported))
 }
 
+func TestDiscardEnvFileOption(t *testing.T) {
+	dict, err := ParseYAML([]byte(`version: "3"
+services:
+  web:
+    image: nginx
+    env_file:
+     - example1.env
+     - example2.env
+`))
+	expectedEnvironmentMap := types.MappingWithEquals{
+		"FOO": strPtr("foo_from_env_file"),
+		"BAZ": strPtr("baz_from_env_file"),
+		"BAR": strPtr("bar_from_env_file_2"), // Original value is overwritten by example2.env
+		"QUX": strPtr("quz_from_env_file_2"),
+	}
+	assert.NilError(t, err)
+	configDetails := buildConfigDetails(dict, nil)
+
+	// Default behavior keeps the `env_file` entries
+	configWithEnvFiles, err := Load(configDetails)
+	assert.NilError(t, err)
+	assert.DeepEqual(t, configWithEnvFiles.Services[0].EnvFile, types.StringList{"example1.env",
+		"example2.env"})
+	assert.DeepEqual(t, configWithEnvFiles.Services[0].Environment, expectedEnvironmentMap)
+
+	// Custom behavior removes the `env_file` entries
+	configWithoutEnvFiles, err := Load(configDetails, WithDiscardEnvFiles)
+	assert.NilError(t, err)
+	assert.DeepEqual(t, configWithoutEnvFiles.Services[0].EnvFile, types.StringList(nil))
+	assert.DeepEqual(t, configWithoutEnvFiles.Services[0].Environment, expectedEnvironmentMap)
+}
+
 func TestBuildProperties(t *testing.T) {
 	dict, err := ParseYAML([]byte(`
 version: "3"


thaJeztah:
Testing this; I want to verify if interpolating environment variables in the env-file is done correctly in the `docker stack` variant

thaJeztah:
Hm; at least found a bug (on master); https://github.com/docker/cli/issues/2075 LOL, haven't checked with this PR yet

cli/compose/loader/full-struct_test.go
```diff
@@ -613,9 +610,6 @@ services:
       BAZ: baz_from_service_def
       FOO: foo_from_env_file
       QUX: qux_from_environment
-    env_file:
```
thaJeztah: Still pinching my head a bit about this change, and wondering (assuming the `TestMarshallConfig` test shows that we can marshal the config to YAML) if we need a separate test to convert a Service _with_ `env_file` back to YAML.

I agree that this whole test is a bit weird (as in: it's not really symmetrical with the YAML file we _read_)

ulyssessouza: The whole `env_file` resolution is not symmetrical anyways. 
I'm even thinking about another PR to pass an option to avoid it, and by that, conserve the symmetry.

cli/compose/loader/loader.go
```diff
@@ -32,6 +32,8 @@ type Options struct {
 	SkipInterpolation bool
 	// Interpolation options
 	Interpolate *interp.Options
+	// Remove 'env_file' entries after resolving to 'environment' section
+	RemoveEnvFiles bool
```
thaJeztah: Looking at this again; wondering if `DiscardEnvFiles` or `ClearEnvFiles` would be less confusing. `Remove....` may sound as if the actual files are removed?

(open to suggestions)

cli/compose/loader/loader_test.go
```diff
@@ -759,6 +759,32 @@ services:
...
+
+	// Default behavior keeps the `env_file` entries
+	configWithEnvFiles, err := Load(configDetails)
+	assert.NilError(t, err)
+	assert.DeepEqual(t, configWithEnvFiles.Services[0].EnvFile, types.StringList{"example1.env",
+		"example2.env"})
+
+	// Custom behavior removes the `env_file` entries
+	configWithoutEnvFiles, err := Load(configDetails, func(opts *Options) {
+		opts.RemoveEnvFiles = true
```
thaJeztah: Should we create a `loader.WithDiscardEnvFiles()` for this? (perhaps create a type for `functionalists(opts @Options)`)

ulyssessouza:
@ndeloof Me too, I thought about passing a smaller structure, but then I preferred this approach because all the functions involved and the Options struct are all in the same file and intimately related.
WDYT @silvin-lubecki ?

thaJeztah:
> I just wonder this make sense to pass the whole opts through all those functions, and not just RemoveEnvFiles or maybe a subset. afaict most options don't have any meaning at the end of the function call chain.

Good point; OTOH, I really want to avoid adding a single `bool` argument (as they're generally awkward to use). However, looking at the code, I don't think we have to pass the options all the way through; we could just discard the options in `Load`.

I was reviewing, let me make that change, and post it here as a suggestion

cli/compose/loader/loader.go
```diff
@@ -100,7 +102,7 @@ func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.
 			}
 		}
 
-		cfg, err := loadSections(configDict, configDetails)
+		cfg, err := loadSections(configDict, configDetails, *opts)
```
thaJeztah: Instead of passing the option through here, we could just as well discard the env-files here

thaJeztah:
Whoops; forgot to submit the actual patch;

Patch with my suggestions; 
[suggestions.patch](https://github.com/docker/cli/files/3642336/suggestions.patch.txt)



```patch
diff --git a/cli/compose/loader/loader.go b/cli/compose/loader/loader.go
index 782a9d3a..a58b7c29 100644
--- a/cli/compose/loader/loader.go
+++ b/cli/compose/loader/loader.go
@@ -32,8 +32,8 @@ type Options struct {
 	SkipInterpolation bool
 	// Interpolation options
 	Interpolate *interp.Options
-	// Remove 'env_file' entries after resolving to 'environment' section
-	RemoveEnvFiles bool
+	// Discard 'env_file' entries after resolving to 'environment' section
+	discardEnvFiles bool
 }
 
 // ParseYAML reads the bytes from a file, parses the bytes into a mapping
@@ -102,11 +102,16 @@ func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.
 			}
 		}
 
-		cfg, err := loadSections(configDict, configDetails, *opts)
+		cfg, err := loadSections(configDict, configDetails)
 		if err != nil {
 			return nil, err
 		}
 		cfg.Filename = file.Filename
+		if opts.discardEnvFiles {
+			for i := range cfg.Services {
+				cfg.Services[i].EnvFile = nil
+			}
+		}
 
 		configs = append(configs, cfg)
 	}
@@ -126,8 +131,7 @@ func validateForbidden(configDict map[string]interface{}) error {
 	return nil
 }
 
-func loadSections(config map[string]interface{}, configDetails types.ConfigDetails,
-	opts Options) (*types.Config, error) {
+func loadSections(config map[string]interface{}, configDetails types.ConfigDetails) (*types.Config, error) {
 	var err error
 	cfg := types.Config{
 		Version: schema.Version(config),
@@ -140,8 +144,7 @@ func loadSections(config map[string]interface{}, configDetails types.ConfigDetai
 		{
 			key: "services",
 			fnc: func(config map[string]interface{}) error {
-				cfg.Services, err = LoadServices(config, configDetails.WorkingDir,
-					configDetails.LookupEnv, opts)
+				cfg.Services, err = LoadServices(config, configDetails.WorkingDir, configDetails.LookupEnv)
 				return err
 			},
 		},
@@ -381,13 +384,11 @@ func formatInvalidKeyError(keyPrefix string, key interface{}) error {
 
 // LoadServices produces a ServiceConfig map from a compose file Dict
 // the servicesDict is not validated if directly used. Use Load() to enable validation
-func LoadServices(servicesDict map[string]interface{}, workingDir string, lookupEnv template.Mapping,
-	opts Options) ([]types.ServiceConfig, error) {
+func LoadServices(servicesDict map[string]interface{}, workingDir string, lookupEnv template.Mapping) ([]types.ServiceConfig, error) {
 	var services []types.ServiceConfig
 
 	for name, serviceDef := range servicesDict {
-		serviceConfig, err := LoadService(name, serviceDef.(map[string]interface{}), workingDir,
-			lookupEnv, opts)
+		serviceConfig, err := LoadService(name, serviceDef.(map[string]interface{}), workingDir, lookupEnv)
 		if err != nil {
 			return nil, err
 		}
@@ -399,16 +400,14 @@ func LoadServices(servicesDict map[string]interface{}, workingDir string, lookup
 
 // LoadService produces a single ServiceConfig from a compose file Dict
 // the serviceDict is not validated if directly used. Use Load() to enable validation
-func LoadService(name string, serviceDict map[string]interface{}, workingDir string, lookupEnv template.Mapping,
-	opts Options) (*types.ServiceConfig, error) {
+func LoadService(name string, serviceDict map[string]interface{}, workingDir string, lookupEnv template.Mapping) (*types.ServiceConfig, error) {
 	serviceConfig := &types.ServiceConfig{}
 	if err := Transform(serviceDict, serviceConfig); err != nil {
 		return nil, err
 	}
 	serviceConfig.Name = name
 
-	if err := resolveEnvironment(serviceConfig, workingDir, lookupEnv,
-		opts); err != nil {
+	if err := resolveEnvironment(serviceConfig, workingDir, lookupEnv); err != nil {
 		return nil, err
 	}
 
@@ -453,8 +452,7 @@ func updateEnvironment(environment map[string]*string, vars map[string]*string,
 	}
 }
 
-func resolveEnvironment(serviceConfig *types.ServiceConfig, workingDir string, lookupEnv template.Mapping,
-	ops Options) error {
+func resolveEnvironment(serviceConfig *types.ServiceConfig, workingDir string, lookupEnv template.Mapping) error {
 	environment := make(map[string]*string)
 
 	if len(serviceConfig.EnvFile) > 0 {
@@ -474,9 +472,6 @@ func resolveEnvironment(serviceConfig *types.ServiceConfig, workingDir string, l
 
 	updateEnvironment(environment, serviceConfig.Environment, lookupEnv)
 	serviceConfig.Environment = environment
-	if ops.RemoveEnvFiles {
-		serviceConfig.EnvFile = nil
-	}
 	return nil
 }
 
diff --git a/cli/compose/loader/loader_test.go b/cli/compose/loader/loader_test.go
index 7e9f7c26..62be3269 100644
--- a/cli/compose/loader/loader_test.go
+++ b/cli/compose/loader/loader_test.go
@@ -759,7 +759,7 @@ services:
 	assert.Check(t, is.DeepEqual([]string{"build", "links", "pid"}, unsupported))
 }
 
-func TestRemoveEnvFileOption(t *testing.T) {
+func TestDiscardEnvFileOption(t *testing.T) {
 	dict, err := ParseYAML([]byte(`version: "3"
 services:
   web:
@@ -779,7 +779,7 @@ services:
 
 	// Custom behavior removes the `env_file` entries
 	configWithoutEnvFiles, err := Load(configDetails, func(opts *Options) {
-		opts.RemoveEnvFiles = true
+		opts.discardEnvFiles = true
 	})
 	assert.NilError(t, err)
 	assert.DeepEqual(t, configWithoutEnvFiles.Services[0].EnvFile, types.StringList(nil))
```

thaJeztah:
oh! Linting error;

```
cli/compose/loader/loader.go:39:1:warning: exported function WithDiscardEnvFiles should have comment or be unexported (golint)
```

@ulyssessouza can you add a godoc comment to that function?

PR #2066, opened 2019-08-25
add rudimentary fish completions for docker network

sanga:
Signed-off-by: Tim Sampson <tim@sampson.fi>

**- What I did**
The fish completions are rather out of date. This ever so slightly improves the situation by including some basically completions for `docker network` (which I happen to use frequently).
**- How I did it**
**- How to verify it**
Open up a fish shell and try to tab-complete various `docker network` commands
improved fish shell completions for docker network.

This is my first contribution to docker, I believe I have followed the appropriate guidelines but if I've misread or -understood something apologies beforehand,

### DIFF
--- a/contrib/completion/fish/docker.fish
+++ b/contrib/completion/fish/docker.fish
@@ -16,7 +16,7 @@
 
 function __fish_docker_no_subcommand --description 'Test if docker has yet to be given the subcommand'
     for i in (commandline -opc)
-        if contains -- $i attach build commit cp create diff events exec export history images import info inspect kill load login logout logs pause port ps pull push rename restart rm rmi run save search start stop tag top trust unpause version wait stats
+        if contains -- $i attach build commit cp create diff events exec export history images import info inspect kill load login logout logs network pause port ps pull push rename restart rm rmi run save search start stop tag top trust unpause version wait stats
             return 1
         end
     end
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thanks! I recall that the completion used to be generated by a script (see a comment in that file); https://github.com/barnybug-archive/docker-fish-completion. That script has been archived though; if updating that script would make it easier to maintain (and update these), we could consider including it in the `contrib` directory.

One problem is that that repository doesn't have a license, so we probably can't just take it, unless  @barnybug makes that contribution 🤔 (alternatively, look into contributing a Fish variant to spf13/cobra, and use that for generating; similar to https://github.com/spf13/cobra/blob/master/bash_completions.go)

sanga:
Scripting the generation of completions sounds like a much better idea than doing stuff manually. Adding fish completion generation to cobra is outside of my skillset and available freetime. However if @barnybug 's python script needs some nursing I'm happy to help there. Obviously it would need a compatible license to be committed here but as it stands I could update this commit with completions generated from it (assuming getting it working again isn't too much hassle) if that's agreeable for you @thaJeztah ? As a temp workaround, as the fish completions are currently very much out of date.

thaJeztah:
Let's get this one merged and work from there, because this changes looks good to me (and at least brings it in a slightly better state).

Fish is (unfortunately) less popular than Bash, and gets less care w.r.t. the completions scripts; and without a dedicated person helping with maintaining those scripts, I think generating them would be a good way forward (leaning towards having it integrated into Cobra here, but if we have the python script as intermediate solution works).

Our Bash completion scripts have been fully hand-written, which takes a lot of effort but definitely gives a higher quality than generated ones (and we're lucky to have people passionate about maintaining them).

With Apple's decision to switch to `zsh` as default shell in the next macOS (Catalina) release, I expect more people using that shell (we may want to consider generating those)

Let me take a few minutes to write up a tracking issue / epic for things we can work on w.r.t. improving the completion scripts; I'll post a link here once I did so.

sanga:
Btw, I had a little look at cobra and it seems like generated fish completions is _almost_ over the line https://github.com/spf13/cobra/pull/754

Issue #1018, opened 2018-04-21
docker login/pull and bearer token


vpaul:
I'm trying to use bearer token providing by dockerhub in config.json. I'm able to pull an image with some sort of scripting with curl and `docker load` but using the same Authorization header as for curl in config.json doesn't make any sense. 

Are there any guidelines how to use docker login/pull commands with tokens?

PR #2068, opened 2019-08-27
 feat(pull/push): add --registry-token option

knqyf263:
close https://github.com/docker/cli/issues/1018

**- What I did**
Add `--registry-token` option to `pull` and `push`.

**- How I did it**
Rewrite AuthConfig.

**- How to verify it**

```
$ docker push --registry-token $TOKEN [YOUR_IMAGE]
```

```
$ docker pull --registry-token $TOKEN [YOUR_IMAGE]
```

Add `--registry-token` option to `pull` and `push`.

image

### DIFF
--- a/cli/command/image/pull.go
+++ b/cli/command/image/pull.go
@@ -15,11 +15,12 @@ import (
 
 // PullOptions defines what and how to pull
 type PullOptions struct {
-	remote    string
-	all       bool
-	platform  string
-	quiet     bool
-	untrusted bool
+	remote        string
+	all           bool
+	platform      string
+	quiet         bool
+	untrusted     bool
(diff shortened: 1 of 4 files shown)


thaJeztah:
I'm unsure about this; these tokens would normally either be set through `docker login`, or set in the `~/.docker/config.json`. Adding it to only `docker push` and `docker pull` would only allow setting the token if those commands are used, but not on (e.g.) `docker run` (which can pull images), `docker stack deploy`, `docker service create` or any other command that touches the registry.

I don't think we should make this change (my 0.02c)

thaJeztah:
closing this one for now, but feel free to continue the conversation

PR #2070, opened 2019-08-27
[docs] Updated docker cp documentation

nickolai-voyage:
**- What I did**
Docs claim that ownership of files copied into a container is changed
to the default user of the container. Experimental evidence shows that
in fact the original ownership is preserved, with or without the -a
option.

**- How I did it**

**- How to verify it**

Updated docker cp documentation

baby penguin


thaJeztah:
I think this may actually be a bug; see https://github.com/moby/moby/pull/34099 (I'll need to look if we can get that one moving)

devdreamsolution (approved):

PR #2077, opened 2019-09-03
Clarifying docker config rules

adrian-plata:
https://docs.docker.com/engine/reference/commandline/cli/

Signed-off-by: Adrian Plata <adrian.plata@docker.com>

### DIFF
--- a/docs/reference/commandline/cli.md
+++ b/docs/reference/commandline/cli.md
@@ -96,28 +96,43 @@ variables.
 ### Configuration files
 
 By default, the Docker command line stores its configuration files in a
-directory called `.docker` within your `$HOME` directory. However, you can
-specify a different location via the `DOCKER_CONFIG` environment variable
-or the `--config` command line option. If both are specified, then the
-`--config` option overrides the `DOCKER_CONFIG` environment variable.
-For example:
-
-    docker --config ~/testconfigs/ ps
-
-Instructs Docker to use the configuration files in your `~/testconfigs/`
-directory when running the `ps` command.
+directory called `.docker` within your `$HOME` directory.
 
 Docker manages most of the files in the configuration directory
 and you should not modify them. However, you *can modify* the
 `config.json` file to control certain aspects of how the `docker`
 command behaves.
 
-Currently, you can modify the `docker` command behavior using environment
+You can modify the `docker` command behavior using environment
 variables or command-line options. You can also use options within
-`config.json` to modify some of the same behavior. When using these
-mechanisms, you must keep in mind the order of precedence among them. Command
-line options override environment variables and environment variables override
-properties you specify in a `config.json` file.
+`config.json` to modify some of the same behavior. If an environment variable
+and the `--config` flag are set, the flag takes precedent over the environment
+variable. Command line options override environment variables and environment
+variables override properties you specify in a `config.json` file.
+
+
+#### Change the `.docker` directory
+
+To specify a different directory, use the `DOCKER_CONFIG`
+environment variable or the `--config` command line option. If both are
+specified, then the `--config` option overrides the `DOCKER_CONFIG` environment
+variable. The example below overrides runs the `docker ps` command using a
+`config.json` file located in the `~/testconfigs/` directory.
+
+```bash
+$ docker --config ~/testconfigs/ ps
+```
+
+This flag only applies to whatever command is being ran. For persistent
+configuration, you can set the `DOCKER_CONFIG` environment variable in your
+shell (e.g. `~/.profile` or `~/.bashrc`). The example below sets the new
+directory to be `HOME/newdir/.docker`.
+
+```bash
+echo export DOCKER_CONFIG=$HOME/newdir/.docker > ~/.profile
+```
+
+#### `config.json` properties
 
 The `config.json` file stores a JSON encoding of several properties:
 
@@ -264,6 +279,23 @@ Following is a sample `config.json` file:
 }
 {% endraw %}
 ```
+
+### Experimental features
+
+To enable experimental features, edit the `config.json` file and set
+`experimental` to `enabled`. The example below enables experimental features
+in a `config.json` file that already enables a debug feature.
+
+```json
+{
+  "experimental": "enabled",
+  "debug": true
+}
+```
+
+You can also enable experimental features from the Docker Desktop menu. See the
+[Docker Desktop Getting Started page](https://docs.docker.com/docker-for-mac#experimental-features)
+for more information.
 
 ### Notary
 


adrian-plata:
@andrewhsu and @thaJeztah PTAL. This is an attempt to clarify where the `.docker/config.json` file location is.

adrian-plata:
Added information on enabling experimental features.

docs/reference/commandline/cli.md
```diff
@@ -265,6 +273,15 @@ Following is a sample `config.json` file:
 {% endraw %}
 ```
 
+### Experimental features
+
+To enable [experimental features](https://success.docker.com/article/experimental-features),
```
andrewhsu: We don't usually link to success.docker.com from the cli/docs repo because it is enterprise-related documentation. It would, however, be appropriate for success.docker.com to link to the reference docs of cli.

docs/reference/commandline/cli.md
```diff
@@ -265,6 +273,15 @@ Following is a sample `config.json` file:
...
+edit the `config.json` file and set `experimental` to `enabled`.
+
+You can also enable experimental features from the Docker Desktop menu. Click
```
andrewhsu: Since this documentation is specifically for the command-line cli, should not have desktop docs here. Experimental feature enablement is already covered in docker desktop docs https://github.com/docker/docker.github.io/blob/master/docker-for-mac/index.md#experimental-features

docs/reference/commandline/cli.md
```diff
@@ -265,6 +273,15 @@ Following is a sample `config.json` file:
 {% endraw %}
 ```
 
+### Experimental features
+
+To enable [experimental features](https://success.docker.com/article/experimental-features),
+edit the `config.json` file and set `experimental` to `enabled`.
```
andrewhsu: Should have an example of the syntax, e.g.
```
{ "experimental" : "enabled" }
```

docs/reference/commandline/cli.md
```diff
@@ -96,28 +96,36 @@ variables.
...
+
+
+#### Change the `.docker` directory
+
+To specify a different directory, use the `DOCKER_CONFIG`
+environment variable or the `--config` command line option. If both are
+specified, then the `--config` option overrides the `DOCKER_CONFIG` environment
+variable.
+
+For example:
```
thaJeztah: Might be good to shortly introduce the example, something like (just a quick draft, improve if you see fit);

```suggestion
The example below overrides runs the `docker ps` command using a `config.json` file located in the `~/testconfigs/` directory:
```

thaJeztah: Oh! I see you have something like that below; I'll comment there

docs/reference/commandline/cli.md
```diff
@@ -96,28 +96,36 @@ variables.
...
+
+    docker --config ~/testconfigs/ ps
+
+Instructs Docker to use the configuration files in your `~/testconfigs/`
```
thaJeztah: This should be `docker` (the command), not Docker (the company) 😅 

Can you also move this _above_ the example (instead of the `For example`)?

```suggestion
The following example instructs Docker to use the configuration files in your `~/testconfigs/`:
```

docs/reference/commandline/cli.md
```diff
@@ -96,28 +96,36 @@ variables.
...
+directory when running the `ps` command.
+
```
thaJeztah: Perhaps in this location, we should explain that the `--config` flag is only applying to the command that's run (i.e., it's not updating the default configuration, and only applies to the command that you ran with that flag set).

After explaining that, I think we should add a _second_ example that's using the `DOCKER_CONFIG` environment variable; we can show that example using `export DOCKER_CONFIG=........`, and below the example add a note that if they want to persist that configuration, they can set it in, e.g. their `~/.profile` (or `~/.bashrc`, or whatever flavour of shell they're using)

docs/reference/commandline/cli.md
```diff
@@ -96,28 +96,36 @@ variables.
...
+#### Change the `.docker` directory
+
+To specify a different directory, use the `DOCKER_CONFIG`
+environment variable or the `--config` command line option. If both are
+specified, then the `--config` option overrides the `DOCKER_CONFIG` environment
+variable.
+
+For example:
+
+    docker --config ~/testconfigs/ ps
```
thaJeztah: Can you add `bash` code-hints and a prompt (`$`) as we do in other shell examples?

```suggestion
```bash
$ docker --config ~/testconfigs/ ps
```

Perhaps we should use a (possibly common) location for the example (if people like to use an `XDG` directory structure (https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html);

```suggestion
```bash
$ docker --config ~/.config/docker/ ps
```

(or `docker --config ${XDG_CONFIG_HOME}/docker ps`, but that might be too specific to XDG)

docs/reference/commandline/cli.md
thaJeztah: Also make sure that we explain that the file must be valid JSON; other options may  already be present in that file, so we could show an example `config.json` with some other ("bogus") options set, e.g.;

```json
{
  "experimental": "enabled",
  "debug": true
}
```

docs/reference/commandline/cli.md
thaJeztah: Agreed; we could link to it though

docs/reference/commandline/cli.md
```diff
@@ -96,28 +96,36 @@ variables.
...
 command behaves.
 
-Currently, you can modify the `docker` command behavior using environment
+You can modify the `docker` command behavior using environment
 variables or command-line options. You can also use options within
 `config.json` to modify some of the same behavior. When using these
-mechanisms, you must keep in mind the order of precedence among them. Command
-line options override environment variables and environment variables override
-properties you specify in a `config.json` file.
+mechanisms, you **must** keep in mind the order of precedence among them.
```
thaJeztah: wondering if the **bold** `must` is a bit too much (could be italic), or rewrite to, e.g. "If both an environment variable and the `--config` flag are set, the flag takes precedence over environment variables" (or something along those lines)

docs/reference/commandline/cli.md
adrian-plata: I took a shot at the command, how does it look @thaJeztah ?

docs/reference/commandline/cli.md
adrian-plata: I think think that's too specific for this example.

docs/reference/commandline/cli.md
```diff
@@ -96,28 +96,43 @@ variables.
...
+$ docker --config ~/testconfigs/ ps
+```
+
+This flag only applies to whatever command is being ran. For persistent
+configuration, you can set the `DOCKER_CONFIG` environment variable in your
+shell (e.g. `~/.profile` or `~/.bashrc`). The example below sets the new
+directory to be `HOME/newdir/.docker`.
+
+```bash
+export DOCKER_CONFIG=$HOME/newdir/.docker > ~/.profile
```
thaJeztah: I think this is missing an `echo`

```suggestion
echo export DOCKER_CONFIG=$HOME/newdir/.docker > ~/.profile
```

docs/reference/commandline/cli.md
```diff
@@ -265,6 +280,23 @@ Following is a sample `config.json` file:
...
+
+```json
+{
+  "experimental": "enabled",
+  "debug": true
+}
+```
+
+You can also enable experimental features from the Docker Desktop menu. See the
+[Docker Desktop Getting Started page](https://github.com/docker/docker.github.io/blob/master/docker-for-mac/index.md#experimental-features)
```
thaJeztah: We should probably link to the docs.docker.com site instead (otherwise, once this is published on docs.docker.com, we direct people to GitHub instead of staying on docs.docker.com);
```suggestion
[Docker Desktop Getting Started page](https://docs.docker.com/docker-for-mac/#experimental-features)
```

PR #2078, opened 2019-09-05
Contrib: Make bash completion faster for services

artheus:
**- What I did**
In `contrib/completion/bash/docker` I changed the `__docker_services` function to use the quiet mode of `docker service ls` which responds with `ID` and `Name` of services.

Using the `-q` quiet mode, is much faster, at least when there are hundreds of services in the Swarm cluster.

I also changed from using `compgen -W` to filter all the services, to using name filtering with the `--filter` argument of `docker service ls`

**- How I did it**
I rewrote the completion code, continuously testing it by running `source ~/.bash_profile` which will reload the docker completion in my local environment. I did this until I were happy with the speed

**- How to verify it**
Source it into your local bash environment and test it by writing e.g. `docker service inspect ` and pressing tab a couple of times. *Note!* you will have to be in an environment with a docker client, connected to a swarm with existing services.

Make bash completion faster for services

image

Pokémon are animals 😄

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -561,7 +561,7 @@ __docker_services() {
 		shift
 	fi
 
-  __docker_q service ls -q --format "$format" --filter "name=$1"
+  __docker_q service ls --quiet --format "$format" "$@"
 }
 
 # __docker_complete_services applies completion of services based on the current
@@ -573,7 +573,7 @@ __docker_complete_services() {
 		current="$2"
 		shift 2
 	fi
-	COMPREPLY=( $(__docker_services "$@" "$current") )
+	COMPREPLY=( $(__docker_services "$@" --filter "name=$current") )
 }
 
 # __docker_tasks returns a list of all task IDs.
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -561,7 +561,7 @@ __docker_services() {
 		shift
 	fi
 
-  __docker_q service ls --quiet --format "$format" "$@"
+	__docker_q service ls --quiet --format "$format" "$@"
 }
 
 # __docker_complete_services applies completion of services based on the current
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -551,7 +551,7 @@ __docker_complete_nodes() {
 # precedence over the environment setting.
 __docker_services() {
 	local format='{{.Name}}'  # default: service name only
-	[ "${DOCKER_COMPLETION_SHOW_SERVICE_IDS}" = yes ] && format='{{.ID}},{{.Name}}' # ID & name
+	[ "${DOCKER_COMPLETION_SHOW_SERVICE_IDS}" = yes ] && format='{{.ID}} {{.Name}}' # ID & name
 
 	if [ "$1" = "--id" ] ; then
 		format='{{.ID}}' # IDs only
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -550,17 +550,18 @@ __docker_complete_nodes() {
 # output to the IDs or names of matching items. This setting takes
 # precedence over the environment setting.
 __docker_services() {
-	local fields='$2'  # default: service name only
-	[ "${DOCKER_COMPLETION_SHOW_SERVICE_IDS}" = yes ] && fields='$1,$2' # ID & name
+	local format='{{.Name}}'  # default: service name only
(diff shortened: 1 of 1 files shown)


contrib/completion/bash/docker
```diff
@@ -550,17 +550,18 @@ __docker_complete_nodes() {
 # output to the IDs or names of matching items. This setting takes
 # precedence over the environment setting.
 __docker_services() {
-	local fields='$2'  # default: service name only
-	[ "${DOCKER_COMPLETION_SHOW_SERVICE_IDS}" = yes ] && fields='$1,$2' # ID & name
+	local format='{{.Name}}'  # default: service name only
+	[ "${DOCKER_COMPLETION_SHOW_SERVICE_IDS}" = yes ] && format='{{.ID}},{{.Name}}' # ID & name
```
albers: This does not work as expected if you `export DOCKER_COMPLETION_SHOW_SERVICE_IDS=yes`.
You will get a list of id,name pairs. Please change like this:

```suggestion
	[ "${DOCKER_COMPLETION_SHOW_SERVICE_IDS}" = yes ] && format='{{.ID}} {{.Name}}' # ID & name
```

contrib/completion/bash/docker
```diff
@@ -550,17 +550,18 @@ __docker_complete_nodes() {
...
+		format='{{.ID}}' # IDs only
 		shift
 	elif [ "$1" = "--name" ] ; then
-		fields='$2' # names only
+		format='{{.Name}}' # names only
 		shift
 	fi
-        __docker_q service ls "$@" | awk "NR>1 {print $fields}"
+
+  __docker_q service ls -q --format "$format" --filter "name=$1"
```
albers: `--format` takes precedence over `-q`, so `-q` can be removed.

artheus: @albers Not quite. The data sent from the docker daemon is truncated when using `-q`, therefore using `-q` is much faster. You can benchmark it on swarm clusters with a lot of services running. For example at my workplace:
```
$ docker service ls -q | wc -l
194

$ time docker service ls --format "{{json .}}" 2>/dev/null

real	0m1.254s
user	0m0.225s
sys	0m0.047s

$ time docker service ls -q --format "{{json .}}" 2>/dev/null

real	0m0.378s
user	0m0.093s
sys	0m0.026s
```

In this example I take the time only once, but I have benchmarked this with 1000 requests over time (each), and it seems to be that without `-q` it takes ~1.5s and with `-q` it takes ~0.4s. So for a tab completion script, I would say that `-q` is quite crucial.

artheus: There is also one of our environments where we always get an error without `-q`:

```bash
$ docker service ls
Error response from daemon: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4835595 vs. 4194304)

$ docker service ls -q | wc -l
780
```

albers: I did some testing and can confirm that while it does not change the output (in our examples), `--quiet` in fact positively impacts the performance. I did not expect this, thanks for pointing me to this.

contrib/completion/bash/docker
```diff
@@ -550,17 +550,18 @@ __docker_complete_nodes() {
...
```
albers: Because you dropped the `"$@"` here, the behavior does not mach the documentation any more. Additional arguments are no longer considered but break the function.
I suggest you change this line to
```suggestion
  __docker_q service ls --quiet --format "$format" "$@"
```
and add `--filter "name=$current"` to the function call in `__docker_complete_services`.

This way, the responsibility for filtering services stays in `__docker_complete_services` (where it belongs) and `__docker_services` does not change its semantics.

artheus: @albers Great feedback. Will get on to doing this ASAP.

contrib/completion/bash/docker
```diff
@@ -550,17 +550,18 @@ __docker_complete_nodes() {
...
+		format='{{.ID}}' # IDs only
 		shift
 	elif [ "$1" = "--name" ] ; then
-		fields='$2' # names only
+		format='{{.Name}}' # names only
 		shift
 	fi
-        __docker_q service ls "$@" | awk "NR>1 {print $fields}"
+
+  __docker_q service ls --quiet --format "$format" "$@"
```
albers: Please fix indentation here.

albers: ping @artheus

artheus: Sorry about the delay. Will fix this right away

artheus: @albers Done

PR #2095, opened 2019-09-20
APP-214 Load Client info in getter function

aiordache:
Change required by [PR-625](https://github.com/docker/app/pull/625)

**- What I did**
Moved cli config load in ClientInfo() function such that when calling it we don't have to initialize the struct in case it isn't. 
**- How I did it**
Load cli config on the first call of ClientInfo().
**- How to verify it**
ClientInfo() should always return an object with the config data already loaded.
Load cli config on the first call to the ClientInfo() from the cli interface
8d1

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -25,6 +25,7 @@ import (
 	"github.com/docker/cli/internal/containerizedengine"
 	dopts "github.com/docker/cli/opts"
 	clitypes "github.com/docker/cli/types"
+	"github.com/docker/docker/api"
 	"github.com/docker/docker/api/types"
 	registrytypes "github.com/docker/docker/api/types/registry"
 	"github.com/docker/docker/client"
@@ -76,7 +77,7 @@ type DockerCli struct {
 	err                   io.Writer
 	client                client.APIClient
 	serverInfo            ServerInfo
-	clientInfo            ClientInfo
+	clientInfo            *ClientInfo
 	contentTrust          bool
 	newContainerizeClient func(string) (clitypes.ContainerizedClient, error)
 	contextStore          store.Store
@@ -87,7 +88,7 @@ type DockerCli struct {
 
 // DefaultVersion returns api.defaultVersion or DOCKER_API_VERSION if specified.
 func (cli *DockerCli) DefaultVersion() string {
-	return cli.clientInfo.DefaultVersion
+	return cli.ClientInfo().DefaultVersion
 }
 
 // Client returns the APIClient
@@ -126,7 +127,14 @@ func ShowHelp(err io.Writer) func(*cobra.Command, []string) error {
 
 // ConfigFile returns the ConfigFile
 func (cli *DockerCli) ConfigFile() *configfile.ConfigFile {
+	if cli.configFile == nil {
+		cli.loadConfigFile()
+	}
 	return cli.configFile
+}
+
+func (cli *DockerCli) loadConfigFile() {
+	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
 }
 
 // ServerInfo returns the server version details for the host this client is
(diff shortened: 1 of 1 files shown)


cli/command/cli.go
```diff
@@ -137,7 +142,28 @@ func (cli *DockerCli) ServerInfo() ServerInfo {
...
-	return cli.clientInfo
+	if cli.clientInfo != nil {
+		return *cli.clientInfo
+	}
+	var experimentalValue string
+	// Environment variable always overrides configuration
+	if experimentalValue = os.Getenv("DOCKER_CLI_EXPERIMENTAL"); experimentalValue == "" {
+		experimentalValue = cli.ConfigFile().Experimental
+	}
+	hasExperimental, _ := isEnabled(experimentalValue)
```
silvin-lubecki: Here we are dropping the error, so we will be silent if the `experimentalValue` is not valid. I don't want to panic here, but returning an error in `ClientInfo()` would break many people. @thaJeztah any idea?

cli/command/cli.go
```diff
@@ -126,6 +127,10 @@ func ShowHelp(err io.Writer) func(*cobra.Command, []string) error {
 
 // ConfigFile returns the ConfigFile
 func (cli *DockerCli) ConfigFile() *configfile.ConfigFile {
+	if cli.configFile != nil {
+		return cli.configFile
+	}
+	cli.configFile = cliconfig.LoadDefaultConfigFile(cli.err)
```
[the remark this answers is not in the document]
thaJeztah: Actually, looking at this; I wonder if this loads the correct config-file, because it's not taking the `--config` flag into account? 🤔 https://github.com/docker/cli/blob/0f337f1dfe574eb12eab8bb102a24f714cc79d86/cli/command/cli.go#L202-L210

cli/command/cli.go
```diff
@@ -137,7 +142,28 @@ func (cli *DockerCli) ServerInfo() ServerInfo {
...
+		return *cli.clientInfo
+	}
+	var experimentalValue string
+	// Environment variable always overrides configuration
+	if experimentalValue = os.Getenv("DOCKER_CLI_EXPERIMENTAL"); experimentalValue == "" {
+		experimentalValue = cli.ConfigFile().Experimental
+	}
+	hasExperimental, _ := isEnabled(experimentalValue)
+
+	version := ""
```
thaJeztah: this clashes with the `github.com/docker/cli/cli/version` import

```suggestion
	v := ""
```

cli/command/cli.go
```diff
@@ -137,7 +142,28 @@ func (cli *DockerCli) ServerInfo() ServerInfo {
...
+
+	if cli.client != nil {
+		version = cli.client.ClientVersion()
+	} else {
+		version = api.DefaultVersion
```
thaJeztah: Is this also a problem if this is called from the plugin? (as in; if there's no client, it will always return the default version, which may not match with  the daemon that is used

thaJeztah:
OK, pushed one possible alternative, but it's also not perfect; https://github.com/docker/cli/compare/master..thaJeztah:app-214_client_info_load_func_alt?expand=1

Description (and change) is in this commit https://github.com/docker/cli/commit/189659f90b249146ddc53bb6189dd5debaf82905;


Possible approach for client info
- split ClientInfo() into ClientInfo() and loadClientInfo()
- split ConfigFile() into ConfigFile() and loadConfigFile()
- ConfigFile() and ClientInfo() call their corresponding loadXX function
  if it has not yet been loaded; this allows them to be used before
  Initialize() was called.
- Initialize() *always* (re-)loads the configuration; this makes sure
  that the correct configuration is used when actually calling commands.

I think the problem remains that (IIUC), the client is not initialised at that point, so we're still working  with partial / incomplete data (config-dir may be incorrect, and api-client not initialised, so we don't know the API version that the daemon supports, or if the daemon has experimental features enabled).

I also explored the "functional arguments" variant for a bit, but I wasn't fully happy with that yet, because they would also have the same problem, and have to be used in the correct order (load flags, load config, load client, get clientinfo). I think it looks cleaner with those options, but that needs more work (and thinking); https://github.com/docker/cli/compare/master..thaJeztah:more_options?expand=1

cli/command/cli.go
```diff
@@ -137,7 +145,35 @@ func (cli *DockerCli) ServerInfo() ServerInfo {
 
 // ClientInfo returns the client details for the cli
 func (cli *DockerCli) ClientInfo() ClientInfo {
-	return cli.clientInfo
+
```
silvin-lubecki: nit: empty line

PR #2105, opened 2019-09-24
build: remove --stream (was experimental)

tiborvass:
--stream was always experimental and this patch removes the functionality.

Users should enable BuildKit with DOCKER_BUILDKIT=1

Related: https://github.com/moby/moby/pull/39983

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"net"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -33,7 +32,6 @@ import (
 	"github.com/docker/docker/pkg/urlutil"
 	units "github.com/docker/go-units"
 	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


cli/command/image/build.go
```diff
@@ -164,11 +161,6 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation("squash", "experimental", nil)
 	flags.SetAnnotation("squash", "version", []string{"1.25"})
 
-	flags.BoolVar(&options.stream, "stream", false, "Stream attaches to server to negotiate build context")
```
thaJeztah: Should we add a hidden, deprecated dummy flag so that we can provide a useful error message if someone still uses this?

Similar to https://github.com/moby/moby/pull/24721/files

We can remove if after 1 or 2 releases

tiborvass: This was experimental, so the best thing I can do is keep the flag to tell them it's been removed and that they should use buildkit, but I don't want to keep the feature.

AkihiroSuda: can we error out with human-friendly message when the dummy flag was specified?

tiborvass: Sure, that's what I suggested in the comment above. I'll update the PR

tiborvass:
@thaJeztah @AkihiroSuda updated, PTAL

crosbymichael:
LGTM

PR #2107, opened 2019-09-26
Fix CREATED field when listing image if date is not specified

### LINKED ISSUE #2047: Display sensible values if image creation date is unspecified
**Description**

Services that build content-addressable images (such as [Nixery](https://nixery.dev) or Bazel's [rules_docker](https://github.com/bazelbuild/rules_docker/)) do not populate the "image creation date" field in image metadata. This is an intentional choice to create reproducible images.

The Docker client displays these images as being 292 years old, which is misleading. It should instead display `<unknown>` or some other sensible value.

Example issues:

* [google/nixery#12](https://github.com/google/nixery/issues/12)
* [bazelbuild/rules_docker#52](https://github.com/bazelbuild/rules_docker/issues/52)

1. Pull an image built with a tool that does not set the image creation date. To test this you can use the public Nixery instance, e.g. `docker pull nixery.dev/hello` for an image containing GNU `hello`.
2. Run `docker images` and look at the output.

jonatasbaldin:
Signed-off-by: Jonatas Baldin <jonatas.baldin@gmail.com>

Fixes #2047.

**- What I did**
When listing images with the `docker images` command, returns `<none>` in the `CREATED` field if the date is not specified during the image build. Without this fix, a `292 years ago` string was displayed.

**- How I did it**
Added a special case in the CLI formatter. 

**- How to verify it**
- Get an image that has no created date `docker pull nixery.dev/hello`

Before the fix:
- Run `docker images`
```
REPOSITORY                           TAG                 IMAGE ID            CREATED             SIZE
nixery.dev/hello                     latest              866a9896ca80        292 years ago             28.1MB
```

After the fix:
```
REPOSITORY                           TAG                 IMAGE ID            CREATED             SIZE
nixery.dev/hello                     latest              866a9896ca80        <none>              28.1MB
```

Fix CREATED field when listing image if date is not specified.

_an alpaca an me_

### DIFF
--- a/cli/command/formatter/image.go
+++ b/cli/command/formatter/image.go
@@ -11,8 +11,8 @@ import (
 )
 
 const (
-	defaultImageTableFormat           = "table {{.Repository}}\t{{.Tag}}\t{{.ID}}\t{{.CreatedSince}}\t{{.Size}}"
-	defaultImageTableFormatWithDigest = "table {{.Repository}}\t{{.Tag}}\t{{.Digest}}\t{{.ID}}\t{{.CreatedSince}}\t{{.Size}}"
+	defaultImageTableFormat           = "table {{.Repository}}\t{{.Tag}}\t{{.ID}}\t{{if .CreatedSince }}{{.CreatedSince}}{{else}}N/A{{end}}\t{{.Size}}"
+	defaultImageTableFormatWithDigest = "table {{.Repository}}\t{{.Tag}}\t{{.Digest}}\t{{.ID}}\t{{if .CreatedSince }}{{.CreatedSince}}{{else}}N/A{{end}}\t{{.Size}}"
 
 	imageIDHeader    = "IMAGE ID"
 	repositoryHeader = "REPOSITORY"
@@ -235,6 +235,11 @@ func (c *imageContext) Digest() string {
 
 func (c *imageContext) CreatedSince() string {
 	createdAt := time.Unix(c.i.Created, 0)
+
+	if createdAt.IsZero() {
+		return ""
+	}
+
 	return units.HumanDuration(time.Now().UTC().Sub(createdAt)) + " ago"
 }
 
--- a/cli/command/formatter/image_test.go
+++ b/cli/command/formatter/image_test.go
@@ -16,6 +16,7 @@ import (
 func TestImageContext(t *testing.T) {
 	imageID := stringid.GenerateRandomID()
 	unix := time.Now().Unix()
+	zeroTime := int64(-62135596800)
 
 	var ctx imageContext
 	cases := []struct {
@@ -76,6 +77,11 @@ func TestImageContext(t *testing.T) {
 				i: types.ImageSummary{SharedSize: 5000, VirtualSize: 20000},
 			}, "15kB", ctx.UniqueSize,
 		},
+		{
+			imageContext{
+				i: types.ImageSummary{Created: zeroTime},
+			}, "", ctx.CreatedSince,
+		},
 	}
 
 	for _, c := range cases {
@@ -91,7 +97,9 @@ func TestImageContext(t *testing.T) {
 
 func TestImageContextWrite(t *testing.T) {
 	unixTime := time.Now().AddDate(0, 0, -1).Unix()
+	zeroTime := int64(-62135596800)
 	expectedTime := time.Unix(unixTime, 0).String()
+	expectedZeroTime := time.Unix(zeroTime, 0).String()
 
 	cases := []struct {
 		context  ImageContext
@@ -125,7 +133,7 @@ func TestImageContextWrite(t *testing.T) {
 			},
 			`REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
 image               tag1                imageID1            24 hours ago        0B
-image               tag2                imageID2            24 hours ago        0B
+image               tag2                imageID2            N/A                 0B
 <none>              <none>              imageID3            24 hours ago        0B
 `,
 		},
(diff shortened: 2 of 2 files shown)


jonatasbaldin:
hey there, any comments on this? 🌱

cli/command/formatter/image.go
```diff
@@ -235,6 +235,11 @@ func (c *imageContext) Digest() string {
 
 func (c *imageContext) CreatedSince() string {
 	createdAt := time.Unix(c.i.Created, 0)
+
+	if createdAt.IsZero() {
+		return "<none>"
```
thaJeztah: Thinking of a slightly different approach; I'll write down a comment / suggestion

thaJeztah:
I'm thinking if we should use this special format, or instead to return an empty string. I realise we already have some of the "special" values that we return, and have not been really consistent in that matter (unfortunately; bad choices in the past), e.g. here's an example of `docker image ls --format '{{ json . }}'` where it's clear we're not really consistent.

```json
{
  "Containers": "N/A",
  "CreatedAt": "2019-08-20 22:19:55 +0200 CEST",
  "CreatedSince": "2 months ago",
  "Digest": "<none>",
  "ID": "961769676411",
  "Repository": "alpine",
  "SharedSize": "N/A",
  "Size": "5.58MB",
  "Tag": "3.10",
  "UniqueSize": "N/A",
  "VirtualSize": "5.582MB"
}
```

I'm thinking of using an empty string instead. That way, the `json` output would be closer to what the API returns (and reduce the amount of "presentation" in that format. This would keep the flexibility for users to format it to their liking. For example;

```console
docker image ls -a --format "table {{.ID}}\t{{if .CreatedSince }}{{.CreatedSince}}{{else}}Prehistoric image, may contain dragons{{end}}"

IMAGE ID            CREATED
dc2fed1afb2c        19 hours ago
8c75c34169e5        2 days ago
48479b5deffa        2 days ago
75fed645a5a0        Prehistoric image, may contain dragons
87b2dd3fa15d        3 days ago
3a3b77f619aa        3 days ago
1544ad69a5d2        Prehistoric image, may contain dragons
9614462fec5e        3 days ago
1f61a4270ed5        4 days ago
```

If we go for this option, I'd be ok with changing the default table format to print `<none>` or `N/A`; https://github.com/docker/cli/blob/69fdd2a4ad1f8ab539d45b19e3a891951e36e157/cli/command/formatter/image.go#L14-L15


I'm _slightly_ leaning towards `N/A`, but as mentioned, we're already inconsistent, and `<none>` is also used (for images without tag or digest), so I'm open to suggestions there.

These would then look something like;

```go
	defaultImageTableFormat           = "table {{.Repository}}\t{{.Tag}}\t{{.ID}}\t{{if .CreatedSince }}{{.CreatedSince}}{{else}}N/A{{end}}\t{{.Size}}"
	defaultImageTableFormatWithDigest = "table {{.Repository}}\t{{.Tag}}\t{{.Digest}}\t{{.ID}}\t{{if .CreatedSince }}{{.CreatedSince}}{{else}}N/A{{end}}\t{{.Size}}"
```

> **Note**: Perhaps as a follow-up, we should consider taking a similar approach for existing fields, but we may need to discuss that, as there could be people relying on the format, even though the cli output format is not a _formal_ API.

thaJeztah:
Happy to hear your thoughts! If you agree with my suggestions, could you update the pull request?


As to the empty values themselves; I _think_ this was a bug in BuildKit (not 100% sure), which should be resolved, but I agree that there may still be images with that issue, so it's nice to take those into account

jonatasbaldin:
Hey there! I don't have strong opinions on this, so I'm good going with an empty string return and `N/A`.

Will make the changes!

jonatasbaldin:
Did the changes ✨  Here's the outputs:

```
$ ./build/docker-darwin-amd64  images -f reference='nixery.dev/hello'
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
nixery.dev/hello    latest              866a9896ca80        N/A                 28.1MB
```

```
$ ./build/docker-darwin-amd64 images -f reference='nixery.dev/hello' --format '{{ json .}}'      
{
  "Containers": "N/A",
  "CreatedAt": "0000-12-31 20:53:32 -0306 LMT",
  "CreatedSince": "N/A",
  "Digest": "<none>",
  "ID": "866a9896ca80",
  "Repository": "nixery.dev/hello",
  "SharedSize": "N/A",
  "Size": "28.1MB",
  "Tag": "latest",
  "UniqueSize": "N/A",
  "VirtualSize": "28.13MB"
}
```

cli/command/formatter/image.go
```diff
@@ -235,6 +235,11 @@ func (c *imageContext) Digest() string {
 
 func (c *imageContext) CreatedSince() string {
 	createdAt := time.Unix(c.i.Created, 0)
+
+	if createdAt.IsZero() {
+		return "N/A"
```
thaJeztah: Thanks! Could you change this to return an empty string, so that the `--format '{{ json .}}'` output will show the empty value as-is?

jonatasbaldin: Oh man, I'm so sorry, shouldn't do any PRs when I'm waking up hahaha Fixed.

thaJeztah: LOL no worries! 😂

jonatasbaldin:
Squashed ✨!

thaJeztah:

cli/command/formatter/image_test.go
```diff
@@ -76,6 +76,11 @@ func TestImageContext(t *testing.T) {
 				i: types.ImageSummary{SharedSize: 5000, VirtualSize: 20000},
 			}, "15kB", ctx.UniqueSize,
 		},
+		{
+			imageContext{
+				i: types.ImageSummary{Created: -62135596800},
+			}, "", ctx.CreatedSince,
```
silvin-lubecki: Can we add also a test with `"N/A"`? And maybe group all the `Created` tests together?

jonatasbaldin: Hello @silvin-lubecki!

What I did was to add `N/A` tests to the whole output tests, inside `TestImageContextWrite` function. Also, since all the other tests are mixed, I didn't see how grouping just a couple of them would make sense (I believe that grouping everything would be better).

What do u think?

jonatasbaldin:
Hey there, just a gentle nudge on this! @silvin-lubecki answered your comment in the thread ✨

silvin-lubecki:
Thank you @jonatasbaldin !!

PR #2108, opened 2019-09-26
Add notice about port binding and overriding of UFW to docker run reference

Kovah:
**- What I did**

I updated the docs for the docker run reference to include a notice about port binding, which overrides / ignores UFW firewall rules. This goes along with the pull request for the [Docker docs](https://github.com/docker/docker.github.io/pull/9519) (more details in there). I think this is important as this may lead to serious vulnerabilities if ports are open even if you closed them.

### DIFF
--- a/docs/reference/commandline/run.md
+++ b/docs/reference/commandline/run.md
@@ -352,6 +352,11 @@ machine. You can also specify `udp` and `sctp` ports.
 The [Docker User Guide](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/)
 explains in detail how to manipulate ports in Docker.
 
+Note that ports which are not bound to the host (i.e., `-p 80:80` instead of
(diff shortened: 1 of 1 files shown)


Kovah:
Any news on this?

FelixFurtmayr:
Hi guys, would be cool if you could merge, so it is in the docs. I also just faced this problem and fixed it like suggest (by adding 127.0.0.1).

I think updating the docs is a first step, but I would really change the default behaviour of docker. Like Kovah mentions it, I see the unsafe default behaviour as serious vulnerability. But I#m not an exert on docker - please give some feedback to this thoughts:

* you are an newbie and installed your machine
* you started your docker
* now you enable ufw; ufw is pretty standard and has good defaults (deny incoming)
* you think you are safe

thaJeztah:
CI failures are not related (see https://github.com/docker/cli/issues/2257), so I'll go ahead and merge this one 👍

PR #2110, opened 2019-09-26
vendor: align with engine at b6684a403c99aaf6be5b8ce0bef3c6650fcdcd12

tiborvass:
Testdata in cli/command/system/testdata needed `"OSVersion":""` to be added.

PR #2101, opened 2019-09-23
cli/config: use os.UserHomeDir instead of github.com/docker/docker/pkg/homedir

tiborvass:
Signed-off-by: Tibor Vass <tibor@docker.com>

This is to avoid too many dependencies as outlined in #2082

Note that, this PR greatly (aka over-) simplifies the homedir logic: it requires HOME envvar to be set, and no longer looks at the current user's homedir via user.Current().HomeDir, as pkg/homedir did.

PR #2111, opened 2019-09-26
config: revert to using pkg/homedir

tiborvass:
In c2626a8 (#2101) homedir logic got extremely simplified to only
checking HOME environment variable on UNIX systems.

Although this should work well enough in traditional environments,
this could break minimal containerized environments.

This patch reverts to using github.com/docker/docker/pkg/homedir
that was recently updated to have less dependencies.

~Dependent on #2110~


tiborvass:
Ping @tonistiigi @jonjohnsonjr @dmcgowan @thaJeztah @silvin-lubecki

PR #2117, opened 2019-10-01
vendor: Bump gopkg.in/yaml.v2

chris-crone:
Signed-off-by: Christopher Crone <christopher.crone@docker.com>

**- What I did**
Bumped gopkg.in/yaml.v2

Related issue: https://github.com/kubernetes/kubernetes/issues/83253

**- How I did it**
* Change `vendor.conf`
* Run `vndr`

**- How to verify it**
* Check `vendor.conf`, `vendor/`, run `make vendor`

**- Description for the changelog**

* Mitigate against YAML files that has excessive aliasing

**- A picture of a cute animal (not mandatory but encouraged)**

PR #2119, opened 2019-10-01
[19.03 backport] vendor: Bump gopkg.in/yaml.v2

thaJeztah:
backport of https://github.com/docker/cli/pull/2117

To mitigate against malicious YAML (kubernetes/kubernetes#83253), we had implemented our own patch to the yams.v2 library. Now that there's an upstream fix, this PR brings us back to using the upstream library.

### Description for the changelog
* cli: Mitigate against YAML files that has excessive aliasing

### DIFF
--- a/vendor/gopkg.in/yaml.v2/decode.go
+++ b/vendor/gopkg.in/yaml.v2/decode.go
@@ -229,6 +229,10 @@ type decoder struct {
 	mapType reflect.Type
 	terrors []string
 	strict  bool
+
+	decodeCount int
+	aliasCount  int
+	aliasDepth  int
 }
 
 var (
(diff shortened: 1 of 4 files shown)


thaJeztah:
ping @silvin-lubecki @chris-crone @vdemeester PTAL

kolyshkin:
Do we need this for 19.03.3?

vendor/gopkg.in/yaml.v2/decode.go
```diff
@@ -315,6 +319,13 @@ func (d *decoder) prepare(n *node, out reflect.Value) (newout reflect.Value, unm
 }
 
 func (d *decoder) unmarshal(n *node, out reflect.Value) (good bool) {
+	d.decodeCount++
```
thaJeztah: @kolyshkin not critical, but it probably won't hurt to have; the beef of the change is in this file (although their fix is a bit weird)

PR #2122, opened 2019-10-02
Allow username/password in config file

jonjohnsonjr:
**- What I did**

Only decode `auth` if it is present. Currently, this overwrites `username` and `password` fields, even if `auth` is empty.

Fixes https://github.com/google/go-containerregistry/issues/555

**- How I did it**

Guard `decodeAuth` with an empty string check.

**- How to verify it**

Run the unit test :)

Alternatively, modify your config file to use username/password instead of auth and try it. I have a branch [here](https://github.com/google/go-containerregistry/compare/master...jonjohnsonjr:auth-fix?expand=1) which seems to work when I try it 👍  

Fix reading config files containing "username" and "password" auth.

### DIFF
--- a/cli/config/configfile/file_test.go
+++ b/cli/config/configfile/file_test.go
@@ -389,23 +389,31 @@ func TestLoadFromReaderWithUsernamePassword(t *testing.T) {
 		Username: "user",
 		Password: "pass",
 	}
-	cf := ConfigFile{
-		AuthConfigs: map[string]types.AuthConfig{
-			"example.com/foo": want,
+
+	for _, tc := range []types.AuthConfig{
+		want,
+		types.AuthConfig{
+			Auth: encodeAuth(&want),
 		},
+	} {
+		cf := ConfigFile{
+			AuthConfigs: map[string]types.AuthConfig{
+				"example.com/foo": tc,
+			},
+		}
+
+		b, err := json.Marshal(cf)
+		assert.NilError(t, err)
+
+		err = configFile.LoadFromReader(bytes.NewReader(b))
+		assert.NilError(t, err)
+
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
ping @tiborvass 

Came across this when I switched over to using docker/cli for config file parsing. Docker seems to ignore the username/password fields from a config file if they're set directly instead of the "auth" field.

max-wittig:
@silvin-lubecki @chris-crone @vdemeester Could you please take a look at this small fix. It would really help. Thanks!

cli/config/configfile/file_test.go
```diff
@@ -380,6 +381,33 @@ func TestGetAllCredentialsCredHelperOverridesDefaultStore(t *testing.T) {
...
+	configFile := New("test-load")
+	defer os.Remove("test-load")
+
+	want := types.AuthConfig{
+		Username: "user",
+		Password: "pass",
+	}
+	cf := ConfigFile{
+		AuthConfigs: map[string]types.AuthConfig{
+			"example.com/foo": want,
```
silvin-lubecki: Can you also test when auth is empty? As this PR fixes that path, it should be tested too, not just the happy path 🤔

jonjohnsonjr: `auth` *is* empty in this case, which is why we get the Username and Password back (instead of nothing). I've added a case the encodes the `username` and `password` as `auth` to test that it still works.

silvin-lubecki: Yep sorry misread it 😅 Thanks for adding the other test case 👍

silvin-lubecki:
@jonjohnsonjr linter is complaining
> cli/config/configfile/file_test.go:1::warning: file is not gofmted with -s (gofmt)

PR #2131, opened 2019-10-10
Refactor the `stack services` command to be uniform

rumpl:
**- What I did**

Running `docker stack services <STACK> --orchestrator swarm` would yield the message "Noting found in stack: asdf" with an exit code 0. The same command with kubernetes orchestrator would yield "nothing found in stack: adsf" (note the lower-case "nothing") and a non-zero exit code. This change makes the `stack services` command uniform for both orchestrators. The logic of getting and printing services is split to reuse the same formatting code.

**- How to verify it**

Run:
```
$ docker stack services unknown --orchestrator swarm
$ docker stack services unknown --orchestrator kubernetes
```

Both commands should return `0` and show the error message `Noting found in stack: unknown`.

Stack services will return the same exit code for all orchestrators.

image

### DIFF
--- a/cli/command/stack/kubernetes/services.go
+++ b/cli/command/stack/kubernetes/services.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 
 	"github.com/docker/cli/cli/command/service"
-	"github.com/docker/cli/cli/command/stack/formatter"
 	"github.com/docker/cli/cli/command/stack/options"
 	"github.com/docker/compose-on-kubernetes/api/labels"
 	"github.com/docker/docker/api/types/filters"
@@ -79,39 +78,39 @@ func getResourcesForServiceList(dockerCli *KubeCli, filters filters.Args, labelS
 	return replicas, daemons, services, nil
 }
 
-// RunServices is the kubernetes implementation of docker stack services
-func RunServices(dockerCli *KubeCli, opts options.Services) error {
+// GetServices is the kubernetes implementation of listing stack services
+func GetServices(dockerCli *KubeCli, opts options.Services) ([]swarm.Service, map[string]service.ListInfo, error) {
 	filters := opts.Filter.Value()
 	if err := filters.Validate(supportedServicesFilters); err != nil {
-		return err
+		return nil, nil, err
 	}
 	client, err := dockerCli.composeClient()
 	if err != nil {
-		return nil
+		return nil, nil, err
 	}
 	stacks, err := client.Stacks(false)
 	if err != nil {
-		return nil
+		return nil, nil, err
 	}
 	stackName := opts.Namespace
 	_, err = stacks.Get(stackName)
 	if apierrs.IsNotFound(err) {
-		return fmt.Errorf("nothing found in stack: %s", stackName)
+		return []swarm.Service{}, nil, nil
 	}
 	if err != nil {
-		return err
+		return nil, nil, err
 	}
 
 	labelSelector := generateLabelSelector(filters, stackName)
 	replicasList, daemonsList, servicesList, err := getResourcesForServiceList(dockerCli, filters, labelSelector)
 	if err != nil {
-		return err
+		return nil, nil, err
 	}
 
 	// Convert Replicas sets and kubernetes services to swarm services and formatter information
 	services, info, err := convertToServices(replicasList, daemonsList, servicesList)
 	if err != nil {
-		return err
+		return nil, nil, err
 	}
 	services = filterServicesByName(services, filters.Get("name"), stackName)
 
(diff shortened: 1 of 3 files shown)


rumpl:
Force pushed to add comments on exported functions

cli/command/stack/kubernetes/services.go
```diff
@@ -79,60 +78,47 @@ func getResourcesForServiceList(dockerCli *KubeCli, filters filters.Args, labelS
 	return replicas, daemons, services, nil
 }
 
-// RunServices is the kubernetes implementation of docker stack services
-func RunServices(dockerCli *KubeCli, opts options.Services) error {
+// GetServices is the kubernetes implementation of listing stack services
+func GetServices(dockerCli *KubeCli, opts options.Services) ([]swarm.Service, map[string]service.ListInfo, error) {
```
thaJeztah: I'm looking at the extra return here; the `service.ListInfo` is basically generated from information that's in `[]swarm.Service`, combined with information about the number of tasks running / desired. If I see it correctly, the `service.ListInfo` here is needed, because `swarm.Service` has no field for task status ("mode" is already part of the Service spec).

I _just_ (finally) merged https://github.com/moby/moby/pull/39231, which is a PR that adds exact those fields to the `Service` struct, which means we could just return `([]swarmService, error)`, and have the presentation code either convert it to a `ListInfo` or adjust that code to use `[]swarm.Service` directly.

cli/command/stack/kubernetes/services.go
```diff
@@ -79,60 +78,47 @@ func getResourcesForServiceList(dockerCli *KubeCli, filters filters.Args, labelS
...
 	stacks, err := client.Stacks(false)
 	if err != nil {
-		return nil
+		return nil, nil, err
 	}
 	stackName := opts.Namespace
 	_, err = stacks.Get(stackName)
 	if apierrs.IsNotFound(err) {
-		return fmt.Errorf("nothing found in stack: %s", stackName)
+		return []swarm.Service{}, nil, nil
```
thaJeztah: I was looking if we should return the error here, and have it handled by the caller 🤔. That way, this function (when used in different situations) would not hide this information.

I think that's ok for a follow-up, because we should have a function to convert Kubernetes API errors to `errdef` errors. In that case we would map (if possible) / standardise these errors, so that the non-k8s specific code understands them; https://github.com/docker/cli/blob/0b6685bca8383437ebaa635437db2891fc77711d/vendor/k8s.io/apimachinery/pkg/api/errors/errors.go#L456-L459

Similar to https://github.com/moby/moby/blob/81dbed4c8b87d99824eecea68c95f10056fd3d2a/errdefs/http_helpers.go#L172-L191

```go
// statusCodeFromContainerdError returns status code for containerd errors when
// consumed directly (not through gRPC)
func statusCodeFromContainerdError(err error) int {
	switch {
	case containerderrors.IsInvalidArgument(err):
		return http.StatusBadRequest
	case containerderrors.IsNotFound(err):
		return http.StatusNotFound
	case containerderrors.IsAlreadyExists(err):
		return http.StatusConflict
	case containerderrors.IsFailedPrecondition(err):
		return http.StatusPreconditionFailed
	case containerderrors.IsUnavailable(err):
		return http.StatusServiceUnavailable
	case containerderrors.IsNotImplemented(err):
		return http.StatusNotImplemented
	default:
		return http.StatusInternalServerError
	}
}
```

cli/command/stack/services.go
```diff
@@ -1,12 +1,17 @@
...
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/command/service"
+	"github.com/docker/cli/cli/command/stack/formatter"
 	"github.com/docker/cli/cli/command/stack/kubernetes"
 	"github.com/docker/cli/cli/command/stack/options"
-	"github.com/docker/cli/cli/command/stack/swarm"
+	swarmStack "github.com/docker/cli/cli/command/stack/swarm"
 	cliopts "github.com/docker/cli/opts"
+	"github.com/docker/docker/api/types/swarm"
```
thaJeztah: (see other comment above) and change this one to

```suggestion
	swarmtypes "github.com/docker/docker/api/types/swarm"
```

cli/command/stack/services.go
```diff
@@ -1,12 +1,17 @@
...
+	"fmt"
+
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/command/service"
+	"github.com/docker/cli/cli/command/stack/formatter"
 	"github.com/docker/cli/cli/command/stack/kubernetes"
 	"github.com/docker/cli/cli/command/stack/options"
-	"github.com/docker/cli/cli/command/stack/swarm"
+	swarmStack "github.com/docker/cli/cli/command/stack/swarm"
```
thaJeztah: (see other comment below)
In other places, we used `swarmtypes` as alias for the types (instead of aliasing `stack/swarm`; would it be ok to change that? So, change this one back to:

```suggestion
	"github.com/docker/cli/cli/command/stack/swarm"
```

cli/command/stack/kubernetes/services.go
thaJeztah: Let me prepare a vendor bump to bring in the changes from  https://github.com/moby/moby/pull/39231

rumpl: Sure, I can wait for the vendor bump, thanks!

cli/command/stack/kubernetes/services.go
rumpl: The reason I didn't return an error here is because the same code for swarm returns an empty slice when no services are found. We can do it one way or the other, we should just make sure that errors are returned in the same places for kubernetes and swarm.

thaJeztah:
carrying in https://github.com/docker/cli/pull/2167 - which is now ready for review 👍

rumpl:
👍 thanks a bunch!

PR #2132, opened 2019-10-10
connhelper: use ssh multiplexing

tonistiigi:
Add multiplexing to ssh connection helper so commands that do multiple requests share the connection. Environment variables for opting out and controlling the persist time for even better connection sharing.

Before:
```
 # time docker run alpine true
docker run alpine true  0.04s user 0.03s system 1% cpu 4.373 total
 # time docker build . 2>/dev/null
docker build . 2> /dev/null  0.05s user 0.03s system 3% cpu 2.312 total
```
After
```
 # time docker run alpine true
docker run alpine true  0.04s user 0.03s system 2% cpu 2.376 total
 # time docker build . 2>/dev/null
docker build . 2> /dev/null  0.04s user 0.03s system 5% cpu 1.295 total
```

@AkihiroSuda @cpuguy83 @tiborvass 

### DIFF
--- a/cli/connhelper/connhelper.go
+++ b/cli/connhelper/connhelper.go
@@ -5,7 +5,10 @@ import (
 	"context"
 	"net"
 	"net/url"
+	"os"
+	"strconv"
 
+	"github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/connhelper/commandconn"
 	"github.com/docker/cli/cli/connhelper/ssh"
 	"github.com/pkg/errors"
@@ -34,7 +37,7 @@ func GetConnectionHelper(daemonURL string) (*ConnectionHelper, error) {
 		}
 		return &ConnectionHelper{
 			Dialer: func(ctx context.Context, network, addr string) (net.Conn, error) {
-				return commandconn.New(ctx, "ssh", append(sp.Args(), []string{"--", "docker", "system", "dial-stdio"}...)...)
+				return commandconn.New(ctx, "ssh", append(multiplexingArgs(), append(sp.Args(), []string{"--", "docker", "system", "dial-stdio"}...)...)...)
 			},
 			Host: "http://docker",
 		}, nil
@@ -52,4 +55,20 @@ func GetCommandConnectionHelper(cmd string, flags ...string) (*ConnectionHelper,
 		},
 		Host: "http://docker",
 	}, nil
+}
+
+func multiplexingArgs() []string {
+	if v := os.Getenv("DOCKER_SSH_NO_MUX"); v != "" {
+		if b, err := strconv.ParseBool(v); err == nil && b {
+			return nil
+		}
+	}
+	if err := os.MkdirAll(config.Dir(), 0700); err != nil {
+		return nil
+	}
+	args := []string{"-o", "ControlMaster=auto", "-o", "ControlPath=" + config.Dir() + "/%r@%h:%p"}
+	if v := os.Getenv("DOCKER_SSH_MUX_PERSIST"); v != "" {
+		args = append(args, "-o", "ControlPersist="+v)
+	}
+	return args
 }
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
SGTM, but IIRC we intentionally avoided having socket files, for security purpose

AkihiroSuda:
That's why we are using stdio instead of ssg -L

cli/connhelper/connhelper.go
```diff
@@ -53,3 +56,16 @@ func GetCommandConnectionHelper(cmd string, flags ...string) (*ConnectionHelper,
...
 	}, nil
 }
+
+func multiplexingArgs() []string {
+	if v := os.Getenv("DOCKER_SSH_NO_MUX"); v != "" {
+		if b, err := strconv.ParseBool(v); err == nil && b {
+			return nil
+		}
+	}
+	args := []string{"-o", "ControlMaster=auto", "-o", "ControlPath=" + config.Dir() + "/%r@%h:%p"}
```
AkihiroSuda: What happens if .ssh/config already has this?

tonistiigi: It overwrites afaik. This is the main reason I left a way to opt-out. If we want to be conservative about it, one way would be to use a different helper prefix.

cli/connhelper/connhelper.go
```diff
@@ -53,3 +56,19 @@ func GetCommandConnectionHelper(cmd string, flags ...string) (*ConnectionHelper,
...
 	}, nil
 }
+
+func multiplexingArgs() []string {
+	if v := os.Getenv("DOCKER_SSH_NO_MUX"); v != "" {
+		if b, err := strconv.ParseBool(v); err == nil && b {
+			return nil
+		}
+	}
+	if err := os.MkdirAll(config.Dir(), 0700); err != nil {
```
AkihiroSuda: Probably we should create a new directory, because when `~/.docker` already exists with other permission bits, we can't guarantee the sockets are created under a directory with 0700. 
On my Macintosh, `~/.docker` was somehow created with 0755.

But it is more likely to hit `UNIX_PATH_MAX` (108)... :cry:

AkihiroSuda: Probably we can let `commandconn` change working directory for avoiding 108 limit

AkihiroSuda:
@justincormack PTAL?

cli/connhelper/connhelper.go
tonistiigi: The socket itself is created with `0600`. This mkdir is to avoid the case where `.docker` does not exist.

cli/connhelper/connhelper.go
```diff
@@ -53,3 +56,19 @@ func GetCommandConnectionHelper(cmd string, flags ...string) (*ConnectionHelper,
...
+		return nil
+	}
+	args := []string{"-o", "ControlMaster=auto", "-o", "ControlPath=" + config.Dir() + "/%r@%h:%p"}
+	if v := os.Getenv("DOCKER_SSH_MUX_PERSIST"); v != "" {
+		args = append(args, "-o", "ControlPersist="+v)
```
justincormack: Is there any reason not to have a Docker config option for this as well?

AkihiroSuda:
ping @tonistiigi

cli/connhelper/connhelper.go
tonistiigi: If you would ever set them they would probably be more dependent on the actual `DOCKER_HOST` value, not a global client config. I'm not even sure we even want to document these, it should be very advanced usage if you want to set them.

thaJeztah: ping @justincormack pal ^

thaJeztah:
@AkihiroSuda ^ looking good for you, or still concerns / changes needed?

cli/connhelper/connhelper.go
```diff
@@ -53,3 +56,19 @@ func GetCommandConnectionHelper(cmd string, flags ...string) (*ConnectionHelper,
 		Host: "http://docker",
 	}, nil
 }
+
+func multiplexingArgs() []string {
+	if v := os.Getenv("DOCKER_SSH_NO_MUX"); v != "" {
```
thaJeztah: can you document the new environment variables in https://github.com/docker/cli/blob/master/docs/reference/commandline/cli.md#environment-variables ?

PR #2142, opened 2019-10-18
TestFormatInfo: add extra test-case

thaJeztah:
This case was in a test in the engine repository, where it is being removed, so add it to the list of existing tests here.

relates to https://github.com/moby/moby/pull/40104


thaJeztah:
ping @tiborvass @andrewhsu @kolyshkin ptal

tiborvass:
@thaJeztah test fails, probably equality vs contains

thaJeztah:
oh, think it's the escaped quotes, as I'm using a string literal (backticks)

thaJeztah:
But I was actually doubting if I should change the test to use "contains" and only look for the relevant part

thaJeztah:
@tiborvass all green now

PR #2143, opened 2019-10-18
Remove remaining uses of pkg/homedir

SamWhited:
This PR was created to address [build failures](https://travis-ci.org/moby/buildkit/jobs/599720422) when updating buildkit to vendor a more recent version of docker which is not currently usable by this package.

**- What I did**

Remove uses of [`"github.com/docker/docker/pkg/homedir".Get`](https://godoc.org/github.com/docker/docker/pkg/homedir#Get).

**- How I did it**

Replaced it with [`os.UserHomeDir`](https://golang.org/pkg/os/#UserHomeDir) which was added in Go 1.12 and is already in use in other places in this package.

**- How to verify it**

Run the tests and ensure everything builds. Comments explain why errors were unhandled to prevent confusion later.

Previous behavior can be found [here](https://github.com/moby/moby/blob/master/pkg/homedir/homedir_unix.go#L21).

n/a

### DIFF
--- a/cli/command/context/options.go
+++ b/cli/command/context/options.go
@@ -13,7 +13,6 @@ import (
 	"github.com/docker/cli/cli/context/kubernetes"
 	"github.com/docker/cli/cli/context/store"
 	"github.com/docker/docker/client"
-	"github.com/docker/docker/pkg/homedir"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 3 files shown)


thaJeztah:
This was removed in https://github.com/docker/cli/pull/2101, but later reverted in https://github.com/docker/cli/pull/2111 (see the description there for reasons why)


/cc @tiborvass @tonistiigi

SamWhited:
This is still removed from `pkg/homedir` in more recent versions of Docker, should removing that function be reverted there too? That would also solve my problem with updating docker in buildkit.

tonistiigi:
I don't understand this. The function exists in moby master https://github.com/moby/moby/blob/master/pkg/homedir/homedir_unix.go#L21 . I think you just need to vendor the correct version.

SamWhited:
That's odd, I could swear I was looking at master, and the commit appears to be right. I'll experiment and try to figure out what I did to cause the wrong version to be vendored. Thanks.

SamWhited:
Turns out that Travis was picking up the "Linux" build constraint, but not "Unix" so the file where `Get` was defined was not being included. There was a comment that said that the tag `osusergo` was required when building statically, so I've added that. Unsure why it was working before, but not now.

PR #2157, opened 2019-10-23
Services: use ServiceStatus on API v1.41 and up

thaJeztah:
depends on:

- [x] ~https://github.com/docker/cli/pull/2162 build static binaries with -tags osusergo~
- [x] moby/moby#40134 Revert "homedir: add cgo or osusergo buildtag constraints for unix"
- [x] https://github.com/docker/cli/pull/2163 Fix-up (gometalinter) linting config
- [x] https://github.com/docker/cli/pull/2158 bump docker/docker to a09e6e323e55e1a9b21df9c2c555f5668df3ac9b

#### Services: use ServiceStatus on API v1.41 and up

API v1.41 adds a new option to get the number of desired and running tasks when listing services. This patch enables this functionality, and provides a fallback mechanism when the ServiceStatus is not available, which would be when using an older API version.

Now that the swarm.Service struct captures this information, the `ListInfo` type is no longer needed, so it is removed, and the related list- and formatting functions have been modified accordingly.

To reduce repetition, sorting the services has been moved to the formatter. This is a slight change in behavior, but all calls to the formatter performed this sort first, so the change will not lead to user-facing changes.

### DIFF
--- a/cli/command/service/formatter.go
+++ b/cli/command/service/formatter.go
@@ -16,6 +16,7 @@ import (
 	"github.com/docker/docker/pkg/stringid"
 	units "github.com/docker/go-units"
 	"github.com/pkg/errors"
+	"vbom.ml/util/sortorder"
 )
 
 const serviceInspectPrettyTemplate formatter.Format = `
@@ -520,17 +521,14 @@ func NewListFormat(source string, quiet bool) formatter.Format {
 	return formatter.Format(source)
 }
 
-// ListInfo stores the information about mode and replicas to be used by template
-type ListInfo struct {
-	Mode     string
-	Replicas string
-}
-
 // ListFormatWrite writes the context
-func ListFormatWrite(ctx formatter.Context, services []swarm.Service, info map[string]ListInfo) error {
+func ListFormatWrite(ctx formatter.Context, services []swarm.Service) error {
 	render := func(format func(subContext formatter.SubContext) error) error {
+		sort.Slice(services, func(i, j int) bool {
+			return sortorder.NaturalLess(services[i].Spec.Name, services[j].Spec.Name)
+		})
 		for _, service := range services {
-			serviceCtx := &serviceContext{service: service, mode: info[service.ID].Mode, replicas: info[service.ID].Replicas}
+			serviceCtx := &serviceContext{service: service}
 			if err := format(serviceCtx); err != nil {
 				return err
 			}
@@ -551,9 +549,7 @@ func ListFormatWrite(ctx formatter.Context, services []swarm.Service, info map[s
 
 type serviceContext struct {
 	formatter.HeaderContext
-	service  swarm.Service
-	mode     string
-	replicas string
+	service swarm.Service
 }
 
 func (c *serviceContext) MarshalJSON() ([]byte, error) {
@@ -569,11 +565,35 @@ func (c *serviceContext) Name() string {
 }
 
 func (c *serviceContext) Mode() string {
-	return c.mode
+	switch {
+	case c.service.Spec.Mode.Global != nil:
+		return "global"
+	case c.service.Spec.Mode.Replicated != nil:
+		return "replicated"
+	default:
+		return ""
+	}
 }
 
 func (c *serviceContext) Replicas() string {
-	return c.replicas
+	s := &c.service
+
+	var running, desired uint64
+	if s.ServiceStatus != nil {
+		running = c.service.ServiceStatus.RunningTasks
+		desired = c.service.ServiceStatus.DesiredTasks
+	}
+	if r := c.maxReplicas(); r > 0 {
+		return fmt.Sprintf("%d/%d (max %d per node)", running, desired, r)
+	}
+	return fmt.Sprintf("%d/%d", running, desired)
+}
+
+func (c *serviceContext) maxReplicas() uint64 {
+	if c.Mode() != "replicated" || c.service.Spec.TaskTemplate.Placement == nil {
+		return 0
+	}
+	return c.service.Spec.TaskTemplate.Placement.MaxReplicas
 }
 
 func (c *serviceContext) Image() string {
--- a/cli/command/service/formatter_test.go
+++ b/cli/command/service/formatter_test.go
@@ -33,29 +33,37 @@ func TestServiceContextWrite(t *testing.T) {
 		// Table format
 		{
 			formatter.Context{Format: NewListFormat("table", false)},
-			`ID                  NAME                MODE                REPLICAS            IMAGE               PORTS
-id_baz              baz                 global              2/4                                     *:80->8080/tcp
-id_bar              bar                 replicated          2/4                                     *:80->8080/tcp
+			`ID                  NAME                MODE                REPLICAS               IMAGE               PORTS
+02_bar              bar                 replicated          2/4                                        *:80->8090/udp
+01_baz              baz                 global              1/3                                        *:80->8080/tcp
+04_qux2             qux2                replicated          3/3 (max 2 per node)                       
+03_qux10            qux10               replicated          2/3 (max 1 per node)                       
 `,
 		},
 		{
 			formatter.Context{Format: NewListFormat("table", true)},
-			`id_baz
-id_bar
+			`02_bar
+01_baz
+04_qux2
+03_qux10
 `,
 		},
 		{
-			formatter.Context{Format: NewListFormat("table {{.Name}}", false)},
-			`NAME
-baz
-bar
+			formatter.Context{Format: NewListFormat("table {{.Name}}\t{{.Mode}}", false)},
+			`NAME                MODE
+bar                 replicated
+baz                 global
+qux2                replicated
+qux10               replicated
 `,
 		},
 		{
 			formatter.Context{Format: NewListFormat("table {{.Name}}", true)},
 			`NAME
-baz
 bar
+baz
+qux2
+qux10
 `,
 		},
 		// Raw Format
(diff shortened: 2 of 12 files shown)


thaJeztah:
taking care of the breaking changes in the homedir package first, in https://github.com/docker/cli/pull/2158

cli/command/service/list.go
```diff
@@ -2,11 +2,13 @@ package service
 
 import (
 	"context"
-	"fmt"
 	"sort"
 
+	"github.com/docker/docker/client"
 	"vbom.ml/util/sortorder"
 
+	"github.com/docker/docker/api/types/versions"
```
thaJeztah: need to fix up formatting, but will do so when rebasing after the other PR's are merged

cli/command/service/formatter.go
```diff
@@ -569,11 +565,35 @@ func (c *serviceContext) Name() string {
...
+		running = c.service.ServiceStatus.RunningTasks
+		desired = c.service.ServiceStatus.DesiredTasks
+	}
+	if r := c.maxReplicas(); r > 0 {
+		return fmt.Sprintf("%d/%d (max %d per node)", running, desired, r)
+	}
+	return fmt.Sprintf("%d/%d", running, desired)
+}
+
+func (c *serviceContext) maxReplicas() uint64 {
```
thaJeztah: I want to export this one, so that it can be used in `--format`, but keeping that separate to reduce the changes a bit

cli/command/service/formatter_test.go
```diff
@@ -33,29 +33,37 @@ func TestServiceContextWrite(t *testing.T) {
...
 		{
 			formatter.Context{Format: NewListFormat("table", false)},
-			`ID                  NAME                MODE                REPLICAS            IMAGE               PORTS
-id_baz              baz                 global              2/4                                     *:80->8080/tcp
-id_bar              bar                 replicated          2/4                                     *:80->8080/tcp
+			`ID                  NAME                MODE                REPLICAS               IMAGE               PORTS
+02_bar              bar                 replicated          2/4                                        *:80->8090/udp
+01_baz              baz                 global              1/3                                        *:80->8080/tcp
+04_qux2             qux2                replicated          3/3 (max 2 per node)                       
+03_qux10            qux10               replicated          2/3 (max 1 per node)
```
thaJeztah: Added test-cases to test the `max X per node` output

cli/command/service/formatter_test.go
```diff
@@ -65,25 +73,32 @@ bar
...
+qux10
 `,
 		},
 	}
 
 	for _, testcase := range cases {
 		services := []swarm.Service{
 			{
-				ID: "id_baz",
+				ID: "01_baz",
```
thaJeztah: prefixing the ID's with numbers, so that we verify the list is sorted by the service _name_ not the service _ID_

cli/command/service/formatter_test.go
```diff
@@ -33,29 +33,37 @@ func TestServiceContextWrite(t *testing.T) {
...
-			formatter.Context{Format: NewListFormat("table {{.Name}}", false)},
-			`NAME
-baz
-bar
+			formatter.Context{Format: NewListFormat("table {{.Name}}\t{{.Mode}}", false)},
+			`NAME                MODE
+bar                 replicated
+baz                 global
+qux2                replicated
+qux10               replicated
```
thaJeztah: Sorting moved to the formatter (see the PR description), which is why these tests now have their output in a different order. I added services named `qux2` and `qux10` to verify the formatter uses naturalsort (otherwise `qux2` would come after `qux10`)

cli/command/service/formatter_test.go
```diff
@@ -95,37 +110,70 @@ bar
...
+					},
 				},
 				Endpoint: swarm.Endpoint{
 					Ports: []swarm.PortConfig{
 						{
 							PublishMode:   "ingress",
 							PublishedPort: 80,
-							TargetPort:    8080,
-							Protocol:      "tcp",
+							TargetPort:    8090,
```
thaJeztah: Not an important change, but multiple services cannot public the same port, so this is just making the test-case more realistic/correct.

cli/command/service/formatter_test.go
```diff
@@ -33,29 +33,37 @@ func TestServiceContextWrite(t *testing.T) {
 		// Table format
 		{
 			formatter.Context{Format: NewListFormat("table", false)},
-			`ID                  NAME                MODE                REPLICAS            IMAGE               PORTS
-id_baz              baz                 global              2/4                                     *:80->8080/tcp
-id_bar              bar                 replicated          2/4                                     *:80->8080/tcp
+			`ID                  NAME                MODE                REPLICAS               IMAGE               PORTS
+02_bar              bar                 replicated          2/4                                        *:80->8090/udp
+01_baz              baz                 global              1/3                                        *:80->8080/tcp
```
thaJeztah: changed the replica-count to be unique for each service, to verify we're showing the correct status

cli/command/service/list.go
```diff
@@ -89,54 +92,97 @@ func runList(dockerCli command.Cli, options listOptions) error {
...
+	if len(status) == 0 {
+		// All services have their ServiceStatus set, so we're done
+		return services, nil
+	}
+
+	tasks, err := c.TaskList(ctx, types.TaskListOptions{Filters: taskFilter})
+	if err != nil {
+		return nil, err
+	}
+	if len(tasks) == 0 {
```
thaJeztah: This optimisation _just_ tipped the function over Gocyclo 16, requiring me to add a `//nolint` 😞 not sure if it's worth factoring these lines out to a separate function just for that

cli/command/service/list.go
```diff
@@ -89,54 +92,97 @@ func runList(dockerCli command.Cli, options listOptions) error {
...
+			// No such thing as number of desired tasks for global services
+			services[i].ServiceStatus = &swarm.ServiceStatus{}
+			status[s.ID] = &swarm.ServiceStatus{}
+			taskFilter.Add("service", s.ID)
+		default:
+			// Unknown task type
 		}
 	}
+	if len(status) == 0 {
+		// All services have their ServiceStatus set, so we're done
```
thaJeztah: Theoretically, we could skip this whole function based on API version, but in situations where the API version is unknown/not set, we would not be propagating `ServiceStatus` (which happens, e.g., in our unit tests). To be on the safe side, the loop above, and this check, make sure that we always check if `ServiceStatus` is present, and otherwise fall back to the old behaviour of fetching tasks and nodes to get this information.

cli/command/service/list_test.go
```diff
@@ -22,7 +29,315 @@ func TestServiceListOrder(t *testing.T) {
 		},
 	})
 	cmd := newListCommand(cli)
+	cmd.SetArgs([]string{})
```
thaJeztah: This was to allow running the test separately in my IDE; when pre-compiling the test and running it, this test would otherwise pick-up the `-test.run` flags as arguments, and fail the  test;

```bash
go test -c -o /tmp/foo.test github.com/docker/cli/cli/command/service \
go tool test2json -t /tmp/foo.test -test.v -test.run '^TestServiceListOrder$'

...
list_test.go:34: assertion failed: error is not nil: \"ls\" accepts no arguments.\n"}
```

thaJeztah:
okay, cleaned up the commits; this should be ready for review

ping @silvin-lubecki @dperny @rumpl @vdemeester PTAL

cli/command/service/formatter_test.go
```diff
@@ -33,29 +33,37 @@ func TestServiceContextWrite(t *testing.T) {
 		// Table format
 		{
 			formatter.Context{Format: NewListFormat("table", false)},
-			`ID                  NAME                MODE                REPLICAS            IMAGE               PORTS
-id_baz              baz                 global              2/4                                     *:80->8080/tcp
-id_bar              bar                 replicated          2/4                                     *:80->8080/tcp
+			`ID                  NAME                MODE                REPLICAS               IMAGE               PORTS
+02_bar              bar                 replicated          2/4                                        *:80->8090/udp
```
silvin-lubecki: Shouldn't this be sorted by service name? 🤔

silvin-lubecki: Ok just saw your comment below 👍

dperny:
LGTM, thanks! Adding this functionality was actually on my todo list for the sprint, so I'm sorry I didn't communicate that.

thaJeztah:
@dperny no worries; it _could've_ waited, but https://github.com/docker/cli/pull/2131 came up, where I looked at the changes, and thought: better make use of this new feature, so that we don't have to change the interface / exported functions multiple times.

And with this change, that refactor became quite a bit more simple (last commit on this PR: https://github.com/docker/cli/pull/2167)

cli/command/stack/kubernetes/conversion_test.go
```diff
@@ -172,8 +165,49 @@ func makeKubeService(service, stack, uid string, serviceType apiv1.ServiceType,
 	}
 }
 
-func makeSwarmService(service, id string, ports []swarm.PortConfig) swarm.Service {
-	return swarm.Service{
+func withMode(mode string, replicas uint64) func(*swarm.Service) {
+	return func(service *swarm.Service) {
+		// service.Spec.Mode = swarm.ServiceMode{
```
silvin-lubecki: nit: commented code?

rumpl:
LGTM! Thanks @thaJeztah !!

cli/command/stack/kubernetes/conversion_test.go
thaJeztah: arf; good catch; I'll update

thaJeztah:
All nice and shiny green; I'll go ahead and merge 👍

PR #2160, opened 2019-10-24
CircleCI: add some info steps, and re-group jobs

thaJeztah:
This patch:

- Combines the shellcheck and lint stages. Free CircleCI plans allow a maximum
  of 4 concurrent jobs, and from the timing, the "lint" and "shellcheck" stages
  combined would still take less time than the other stages, so combining them
  keeps the same overall duration, but saving one machine "slot".
- Splits some steps, so that their output can be found more easily in the CI
  results. For example, separating building of Docker images from running them.
- Adds a "Docker info" step, because information about the environment can be
  useful when debugging.
- Adds the "Docker info" and "Docker version" steps to all stages, so that it's
  possible to get that information without having to find the stage in which
  it's printed.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -14,11 +14,27 @@ jobs:
           reusable: true
           exclusive: false
       - run:
+          name: "Docker version"
           command: docker version
+      - run:
+          name: "Docker info"
+          command: docker info
+      - run:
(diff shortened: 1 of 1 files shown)


thaJeztah:
@chris-crone updated, PTAL

PR #2161, opened 2019-10-24
config: don't call homedir on init()

thaJeztah:
The `homedir.Get()` function can call out to external applications to get user information.

This patch changes the package to lazily obtain the user's home-directory on first use, instead of when initializing the package.

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/cli/cli/config/credentials"
@@ -23,10 +24,15 @@ const (
 )
 
 var (
-	configDir = os.Getenv("DOCKER_CONFIG")
+	initConfigDir sync.Once
+	configDir     string
 )
 
-func init() {
+func setConfigDir() {
+	if configDir != "" {
+		return
+	}
+	configDir = os.Getenv("DOCKER_CONFIG")
 	if configDir == "" {
 		configDir = filepath.Join(homedir.Get(), configFileDir)
 	}
@@ -34,6 +40,7 @@ func init() {
 
 // Dir returns the directory the configuration file is stored in
 func Dir() string {
+	initConfigDir.Do(setConfigDir)
 	return configDir
 }
 


thaJeztah:
@silvin-lubecki @simonferquel @tiborvass PTAL

kolyshkin:
> The homedir.Get() function can call out to external applications to get user information.

Can you elaborate on this?

My $0.02: since golang 1.8 or so `user.Current()` result is cached.

thaJeztah:
> Can you elaborate on this?

arg; you're right; I was thinking the package was still using the runc / libcontainer stuff, but that was removed in https://github.com/moby/moby/commit/a8608b5b67c77169276863cf31c6bc89a9ab3d8c#diff-1b2f13bed97df1999bf7f922eb2f0909

thaJeztah:
@kolyshkin let me know what you think; I _think_ it still makes sense to get this info lazily, but happy to close otherwise

thaJeztah:
Actually; let me make one small change, and don't attempt to set the dir if it's non-empty (which could be if a user of this package calls `SetDir()` before `Dir()`

kolyshkin:
> Actually; let me make one small change, and don't attempt to set the dir if it's non-empty (which could be if a user of this package calls `SetDir()` before `Dir()`

I don't like this change -- now you have two "run once" guards, and they are scattered.

Maybe it's better to rewrite `Dir()` to encapsulate setConfigDir(), i.e. the same way that user.Current is written (except there's no need to do a copy a string being returned)? 

https://github.com/golang/go/blob/46e0d724b3f14fd0d350123bbf101e815b493791/src/os/user/lookup.go#L9-L28

thaJeztah:
@kolyshkin it's a bit problematic; tbh I think this package is way too complex 

`Dir()` (and thus `configDir`) is quite ambiguous; it's neither the "default" location, nor the location of the current config.

It's a global / package level variable that can be set to a default, but also overwritten by a "setter". 

eg, someone could set the directory; https://github.com/docker/cli/blob/7a279af43de6650d2f36c41f9d926db7be3b3567/cli/config/config.go#L46

In which case we don't want to overwrite it with the default


Someone could load the config, and specify the configDir https://github.com/docker/cli/blob/7a279af43de6650d2f36c41f9d926db7be3b3567/cli/config/config.go#L82-L85

In which case `Dir()` is not changed, and thus doesn't reflect the current config's location. But if no path is specified, it _is_ used as the default, and thus the config's location

And worst case; the specified location couldn't be found, in which case the current config location is _neither_ the one specified, not the default one, but a completely different path: https://github.com/docker/cli/blob/7a279af43de6650d2f36c41f9d926db7be3b3567/cli/config/config.go#L108-L112

jonjohnsonjr:
This tripped us up, so I had to land a workaround: https://github.com/google/go-containerregistry/pull/568

I'm in favor of this change :)

thaJeztah:
@kolyshkin PTAL

PR #2158, opened 2019-10-24
bump docker/docker to a09e6e323e55e1a9b21df9c2c555f5668df3ac9b

thaJeztah:
depends on:

- [x] ~https://github.com/docker/cli/pull/2162 build static binaries with -tags osusergo~
- [x] moby/moby#40134 Revert "homedir: add cgo or osusergo buildtag constraints for unix"
- [x] https://github.com/docker/cli/pull/2163 Fix-up (gometalinter) linting config

#### bump Microsoft/hcsshim 2226e083fc390003ae5aa8325c3c92789afa0e7a

full diff: https://github.com/Microsoft/hcsshim/compare/672e52e9209d1e53718c1b6a7d68cc9272654ab5...2226e083fc390003ae5aa8325c3c92789afa0e7a

- microsoft/hcsshim#569 Enhancement: add osversion.Build() utility
    - relates to moby/moby#39100 Use Microsoft/hcsshim constants and deprecate pkg/system.GetOsVersion()

#### bump docker/docker to a09e6e323e55e1a9b21df9c2c555f5668df3ac9b

full diff: https://github.com/docker/docker/compare/b6684a403c99aaf6be5b8ce0bef3c6650fcdcd12...a09e6e323e55e1a9b21df9c2c555f5668df3ac9b

relevant changes:

- moby/moby#39995 Update containerd binary to v1.2.10
- moby/moby#40001 Update runc to v1.0.0-rc8-92-g84373aaa (CVE-2019-16884)
- moby/moby#39999 bump golang 1.13.1 (CVE-2019-16276)
- moby/moby#40102 bump golang 1.13.3 (CVE-2019-17596)
- moby/moby#40134 Revert "homedir: add cgo or osusergo buildtag constraints for unix"
    - reverts moby/moby#39994 homedir: add cgo or osusergo buildtag constraints for unix,
      in favor of documenting when to set the `osusergo` build tag. The `osusergo`
      build-flag must be used when compiling a static binary with `cgo` enabled,
      and linking against `glibc`.
- moby/moby#39983 builder: remove legacy build's session handling
  This feature was used by docker build --stream and it was kept experimental.
  Users of this endpoint should enable BuildKit anyway by setting Version to BuilderBuildKit.
    - Related: #2105 build: remove --stream (was experimental)
- moby/moby #40045 Bump logrus 1.4.2, go-shellwords, mergo, flock, creack/pty,
  golang/gddo, gorilla/mux
[...]

PR #2162, opened 2019-10-24
build static binaries with -tags osusergo

thaJeztah:
When building a static binary, the osusergo build-tag
should be used so that a pure go implementation of the
`os/user` package is used.

relates to:

- golang/go#23265 os/user: add build tags to select cgo-vs-go
  implementation, like net package
    - https://github.com/golang/go/commit/62f0127d8104d8266d9a3fb5a87e2f09ec8b6f5b
- golang/go#24787 os/user: LookupId panics on Linux+glibc static build
- golang/go#26492 cmd/go: build: add -static flag
- golang/go#12058 cmd/go: #cgo pkg-config: add conditional --static flag to pkg-config
- moby/moby#39994 homedir: add cgo or osusergo buildtag constraints for unix

```
* Use pure Go implementation of os/user package for static binaries
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,14 +12,14 @@ clean: ## remove build artifacts
 
 .PHONY: test-unit
 test-unit: ## run unit tests, to change the output format use: GOTESTSUM_FORMAT=(dots|short|standard-quiet|short-verbose|standard-verbose) make test-unit 
-	gotestsum $(TESTFLAGS) -- $${TESTDIRS:-$(shell go list ./... | grep -vE '/vendor/|/e2e/')}
+	gotestsum $(TESTFLAGS) -- -tags osusergo $${TESTDIRS:-$(shell go list ./... | grep -vE '/vendor/|/e2e/')}
 
 .PHONY: test
 test: test-unit ## run tests
 
 .PHONY: test-coverage
 test-coverage: ## run test coverage
-	gotestsum -- -coverprofile=coverage.txt $(shell go list ./... | grep -vE '/vendor/|/e2e/')
+	gotestsum -- -tags osusergo -coverprofile=coverage.txt $(shell go list ./... | grep -vE '/vendor/|/e2e/')
 
 .PHONY: fmt
 fmt:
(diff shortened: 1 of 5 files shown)


thaJeztah:
ping @kolyshkin @silvin-lubecki @tiborvass PTAL

kolyshkin:
1. Since `CGO` is explicitly disabled here for a static build, technically this build tag is not needed. Doesn't hurt to have it, but it's essentially a no-op when `CGO=0` is set.

2. If we still want to add this, we might also add `netgo`.

kolyshkin:
So unless we want to try enabling cgo for a static build, i'd rather close this one

kolyshkin:
So, in case we would want to use CGO for static binary, we need to add `osusergo` but also `netgo` and `static_build` tags. IOW:

```diff
diff --git a/scripts/build/binary b/scripts/build/binary
index 41c4196cc..3440dfcf6 100755
--- a/scripts/build/binary
+++ b/scripts/build/binary
@@ -8,7 +8,8 @@ set -eu -o pipefail
 source ./scripts/build/.variables
 
 echo "Building statically linked $TARGET"
-export CGO_ENABLED=0
-go build -o "${TARGET}" --ldflags "${LDFLAGS}" "${SOURCE}"
+LDFLAGS="-extldflags -static $LDFLAGS"
+TAGS="-tags netgo,osusergo,static_build"
+go build -o "${TARGET}" ${TAGS} --ldflags "${LDFLAGS}" "${SOURCE}"
 
 ln -sf "$(basename "${TARGET}")" build/docker
```

but for now I don't see a reason to do that ([re-]enable CGO for static build)

thaJeztah:
@kolyshkin this is in preparation of https://github.com/docker/cli/pull/2158. Where, because of https://github.com/moby/moby/pull/39994, everything breaks if we don't set the `osusergo` tag

thaJeztah:
we should probably wait for the outcome of https://github.com/moby/moby/pull/40134

thaJeztah:
This is no longer needed, now that moby/moby#40134 Revert "homedir: add cgo or osusergo buildtag constraints for unix" is merged

kolyshkin:
OK since moby/moby#40134 is merged, I assume this one can be closed.

PR #1797, opened 2019-04-02
Replace gometalinter with Golangci lint

silvin-lubecki:
**- What I did**
gometalinter is deprecated in favour og golangci-lint which is faster and less memory hungry.
https://github.com/golangci/golangci-lint has more on the performance etc.

All the caught issues were fixed, commit by commit for better review. Happy to squash everything at the end of the review 👍 

Moved the linter Dockerfile to Multistage build.

**- A picture of a cute animal (not mandatory but encouraged)**
image

PR #2163, opened 2019-10-24
Fix-up (gometalinter) linting config

thaJeztah:
#### gometalinter: fix configuration

The configuration abused "Exclude" to exclude file-paths by filtering on the output, however, the `Skip` option was designed for that, whereas `Exclude` is for matching warnings.

An explicit "Skip" was added for "vendor", because even though the vendor directory should already be ignored by the linter, in some situations, it still seemed to warn on issues, so let's explicitly ignore it.

### DIFF
--- a/gometalinter.json
+++ b/gometalinter.json
@@ -2,12 +2,16 @@
   "Vendor": true,
   "Deadline": "2m",
   "Sort": ["linter", "severity", "path", "line"],
-  "Exclude": [
+  "Skip": [
       "cli/compose/schema/bindata.go",
       "cli/command/stack/kubernetes/api/openapi",
       "cli/command/stack/kubernetes/api/client",
       ".*generated.*",
-      "parameter .* always receives"
+      "vendor"
+  ],
+  "Exclude": [
+    "parameter .* always receives",
+    "_esc(Dir|FS|FSString|FSMustString) is unused"
   ],
   "EnableGC": true,
   "Linters": {


thaJeztah:
ping @silvin-lubecki @kolyshkin @tiborvass @vdemeester PTAL

thaJeztah:
removed one commit, as it may only be applicable with other changes

kolyshkin:
side note: why don't we switch to `golangci-lint` instead?

gometalinter.json
```diff
@@ -2,12 +2,16 @@
...
       "cli/compose/schema/bindata.go",
       "cli/command/stack/kubernetes/api/openapi",
       "cli/command/stack/kubernetes/api/client",
       ".*generated.*",
-      "parameter .* always receives"
+      "vendor"
+  ],
+  "Exclude": [
+    "parameter .* always receives",
+    "_esc(Dir|FS|FSString|FSMustString) is unused"
```
kolyshkin: the commit message does not explain where this line is coming from

thaJeztah: Ahm, yes. Those are from the `cli/compose/schema/bindata.go` file. Looks like some linters don't get passed the `Skip` configuration correctly, and still invalidate.

We really need to get rid of gometalinter in this repository; there's a WIP pull request https://github.com/docker/cli/pull/1797, but I need to find some time to carry it.

silvin-lubecki:
@kolyshkin I have a PR somewhere (#1797 ) to switch to `golangci-lint` but never had time to rebase it and fix all the linter failures 😞

PR #2165, opened 2019-10-25
fixes #1923 Update stack command docs --orchestrator

### LINKED ISSUE #1923: Stack deploy docs don't explain --orchestrator
The file [stack_deploy.md](https://github.com/docker/cli/blob/2c7822b0361dd3e78bb0745a5c03f741bc21821e/docs/reference/commandline/stack_deploy.md) mentions the `--orchestrator` flag but does not explain what it does or what the different options do. Of the three values it lists for this flag, `swarm` and `kubernetes` might be considered to be self-explanatory, but the `all` option certainly isn't. The docs either need to explain what the options mean, or link to somewhere that does.

nabeel-shakeel:
Signed-off-by: nabeel-shakeel <nabeel_shakeel@symantec.com>

**- What I did**

Since the option `all` of `--orchestrator` does not have any functionality associated with it.

https://github.com/docker/cli/blob/9e9fbf0699009ed1e52af587dbc88b19c7ba7689/cli/command/stack/common.go#L40

It return this :
https://github.com/docker/cli/blob/9e9fbf0699009ed1e52af587dbc88b19c7ba7689/cli/command/stack/cmd.go#L14
 We can either use **kubernetes** or **swarm**

**- How I did it**
I removed the `all` option to make docs consistent that only two options are supported and if someone try to input other then that, then error handling has already there.

**- How to verify it**
you can verify either looking at the docs: [Stack Deploy](https://github.com/docker/cli/blob/2c7822b0361dd3e78bb0745a5c03f741bc21821e/docs/reference/commandline/stack_deploy.md)
Or by running the command: `docker stack deploy --orchestrator swarm`

### DIFF
--- a/cli/command/stack/cmd.go
+++ b/cli/command/stack/cmd.go
@@ -69,7 +69,7 @@ func NewStackCommand(dockerCli command.Cli) *cobra.Command {
 	flags := cmd.PersistentFlags()
 	flags.String("kubeconfig", "", "Kubernetes config file")
 	flags.SetAnnotation("kubeconfig", "kubernetes", nil)
-	flags.String("orchestrator", "", "Orchestrator to use (swarm|kubernetes|all)")
+	flags.String("orchestrator", "", "Orchestrator to use (swarm|kubernetes)")
 	return cmd
 }
 
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -1136,7 +1136,7 @@ __docker_complete_stack_orchestrator_options() {
 			return 0
 			;;
 		--orchestrator)
-			COMPREPLY=( $( compgen -W "all kubernetes swarm" -- "$cur") )
+			COMPREPLY=( $( compgen -W "kubernetes swarm" -- "$cur") )
 			return 0
 			;;
 	esac
--- a/docs/reference/commandline/stack.md
+++ b/docs/reference/commandline/stack.md
@@ -23,7 +23,7 @@ Manage Docker stacks
 Options:
       --help                  Print usage
       --kubeconfig string     Kubernetes config file
-      --orchestrator string   Orchestrator to use (swarm|kubernetes|all)
+      --orchestrator string   Orchestrator to use (swarm|kubernetes)
 
 Commands:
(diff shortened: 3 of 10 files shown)


albers:
@nabeel-shakeel The build fails because you changed the expected help messages in the tests but the actual help message is still unchanged.
To make the tests pass, you also need to change the message [here](https://github.com/docker/cli/blob/69b73f7519a433691a2432a234480964b591e0d9/cli/command/stack/cmd.go#L72).

nabeel-shakeel:
@albers : Thanks a-lot in quick reviewing and point me to the location!

cli/command/stack/cmd.go
```diff
@@ -69,7 +69,7 @@ func NewStackCommand(dockerCli command.Cli) *cobra.Command {
 	flags := cmd.PersistentFlags()
 	flags.String("kubeconfig", "", "Kubernetes config file")
 	flags.SetAnnotation("kubeconfig", "kubernetes", nil)
-	flags.String("orchestrator", "", "Orchestrator to use (swarm|kubernetes|all)")
+	flags.String("orchestrator", "", "Orchestrator to use (swarm|kubernetes)")
```
thaJeztah: While it's true for commands that have to interact with a specific orchestrator (e.g. deploying, or removing a stack), I think the `--orchestrator=all` option is supported for, `docker stack ls`

@silvin-lubecki (please double check I'm right there)

nabeel-shakeel: @thaJeztah : So we have this `all` option everywhere other then `stack ls` is of no use ?

nabeel-shakeel: @thaJeztah : It would be good to know, whether this fix helps or its by design that `all` option is used only with `stack ls` and invalid option in all other stack commands

thaJeztah: It's by design that it only works for stack ls (for setups that support multiple orchestrators in parallel, which is possible with Docker Enterprise)

nabeel-shakeel:
@thaJeztah: Thanks for updating!

PR #2173, opened 2019-10-28
Replace gometalinter with Golangci lint [carry 1797]

thaJeztah:
carries https://github.com/docker/cli/pull/1797
closes https://github.com/docker/cli/pull/1797

- [x] based on top of / depends on https://github.com/docker/cli/pull/2174 compose/loader: fix TestIsAbs not testing all combinations
- [x] depends on https://github.com/docker/cli/pull/2172 "Gometalinter: raise deadline to 3 minutes"
    - in case we want to backport that change, and not the switch to gometalinter
- [x] depends on https://github.com/docker/cli/pull/2180 kubernetes/conversion_test: use test-builders package

PR #2174, opened 2019-10-29
compose/loader: fix TestIsAbs not testing all combinations

thaJeztah:
This test was intending to run all tests, but didn't, which was
caught by golangci-lint;

    cli/compose/loader/windows_path_test.go:46:17: SA4010: this result of append is never used, except maybe in other appends (staticcheck)
    	tests := append(isabstests, winisabstests...)
    	               ^

Looks like a whoopsie on my side in https://github.com/docker/cli/pull/1990 (I recall I was testing some variations, probably forgot to change back 😂). Pushing as a PR separate from https://github.com/docker/cli/pull/2173 in case we want to backport this individually.

### DIFF
--- a/cli/compose/loader/windows_path_test.go
+++ b/cli/compose/loader/windows_path_test.go
@@ -43,7 +43,8 @@ var winisabstests = []IsAbsTest{
 }
 
 func TestIsAbs(t *testing.T) {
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @vdemeester @silvin-lubecki one more small one 😅

PR #2177, opened 2019-10-29
[19.03 backport] Do not disable sig-proxy when using a TTY

thaJeztah:
backport of:

- https://github.com/docker/cli/pull/1841 Do not disable sig-proxy when using a TTY 
- https://github.com/docker/cli/pull/2016 TestSigProxyWithTTY: fix 

This partially reverts https://github.com/moby/moby/commit/e0b59ab52b87b8fc15dd5534c3231fdd74843f9f (https://github.com/moby/moby/pull/2426), and does not automatically disable proxying signals in TTY-mode

fixes https://github.com/moby/moby/issues/28872 (docker client doesn't pass signals when a terminal is attached)
fixes https://github.com/moby/moby/issues/3793 docker client not passing signals to dockerd

relates to:

- https://github.com/moby/moby/issues/9098 Kill `docker exec` command will not terminate the spawned process
  - This patch does not fix the `docker exec` case; it looks like there's no API to kill an exec'd process, so there's no signal-proxy for this yet

Before this change:
------------------------------------

Start a container with a TTY in one shell:

```
docker run -it --init --name repro-28872 busybox sleep 30
```

then, in another shell, kill the docker cli:

```
kill `pgrep -f repro-28872`
```

Notice that the CLI was killed, but the signal not forwarded to the container;
the container continues running

```
docker container inspect --format '{{ .State.Status }}' repro-28872
running

docker container rm -f repro-28872
```

After this change:
------------------------------------

Start a container with a TTY in one shell:

```
docker run -it --init --name repro-28872 busybox sleep 30
```

then, in another shell, kill the docker cli:

```
kill `pgrep -f repro-28872`
```

Verify that the signal was forwarded to the container, and the container exited

```
docker container inspect --format '{{ .State.Status }}' repro-28872
exited

docker container rm -f repro-28872
```

```
- Fix docker client doesn't pass signals when a terminal is attached
```

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -115,11 +115,6 @@ func runContainer(dockerCli command.Cli, opts *runOptions, copts *containerOptio
 		config.StdinOnce = false
 	}
 
-	// Disable sigProxy when in TTY mode
-	if config.Tty {
-		opts.sigProxy = false
-	}
-
 	// Telling the Windows daemon the initial size of the tty during start makes
 	// a far better user experience rather than relying on subsequent resizes
 	// to cause things to catch up.
--- a/e2e/container/proxy_signal_test.go
+++ b/e2e/container/proxy_signal_test.go
@@ -17,29 +17,23 @@ import (
 // TestSigProxyWithTTY tests that killing the docker CLI forwards the signal to
 // the container, and kills the container's process. Test-case for moby/moby#28872
 func TestSigProxyWithTTY(t *testing.T) {
-	_, tty, err := pty.Open()
-	assert.NilError(t, err, "could not open pty")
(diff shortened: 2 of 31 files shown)


ANogin:
Note that this change in behavior could be a very unpleasant surprise for people, particularly as it relates to SIGHUP. I am used to running very long-term jobs on a remote machine, under docker. I would ssh into the machine, start a container, suspend my laptop, and then later ssh back in and attach to the still-running container to see the job status. But suddenly all my jobs started disappearing! Turned out that this change suddenly meant that SIGHUPs from ssh connections dying were propagating into the containers and killing them! I would recommend specifically mentioning this difference in changelog.

PR #2179, opened 2019-10-30
unused-for is a deprecated synonym for until

sj26:
**- What I did**

Change `unused-for` to `until` in `docker builder prune --help` output, which should flow through to the docs site.

**- How I did it**

Modifying the command flag description.

**- How to verify it**

Run `docker builder prune --help`

Fixed deprecated docker builder prune filter example

image


thaJeztah:
Thanks! this should probably also remove the term from the completion script; https://github.com/docker/cli/blob/3fa5aef0f98d472b39bca55c1a82ca348ae0883e/contrib/completion/bash/docker#L1230

However, I'm a bit hazy on this one; I recall there were some issues with `docker builder prune` vs `docker prune` (last one could only support "generic" filters supported by all object types, so perhaps `until` was mainly added as an _alias_ so that it would work in that case). And I know this was renamed from `max-age` at some point; https://github.com/docker/cli/pull/1451 


ping @tiborvass could you help out on this one?

PR #2180, opened 2019-10-30
kubernetes/conversion_test: use test-builders package

thaJeztah:
also some small changes to the internal/test/builders

relates to https://github.com/docker/cli/pull/2173

### DIFF
--- a/cli/command/stack/kubernetes/conversion_test.go
+++ b/cli/command/stack/kubernetes/conversion_test.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"testing"
 
+	. "github.com/docker/cli/internal/test/builders" // Import builders to get the builder function as package function
 	"github.com/docker/compose-on-kubernetes/api/labels"
 	"github.com/docker/docker/api/types/swarm"
 	"gotest.tools/assert"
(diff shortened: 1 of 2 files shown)


cli/command/stack/kubernetes/conversion_test.go
```diff
@@ -165,25 +167,7 @@ func makeKubeService(service, stack, uid string, serviceType apiv1.ServiceType,
...
-			}
-			withStatus(0, replicas)
-		default:
-			service.Spec.Mode = swarm.ServiceMode{}
-			withStatus(0, 0)
-		}
-	}
-}
-
+// TODO convertToServices currently doesn't set swarm.EndpointSpec.Ports
```
thaJeztah: Added a TODO for this; looks like the conversion doesn't return the ports in the _spec_ for the service

internal/test/builders/service.go
```diff
@@ -6,20 +6,12 @@ import (
...
-		ID: "serviceID",
-		Spec: swarm.ServiceSpec{
-			Annotations: swarm.Annotations{
-				Name: "defaultServiceName",
-			},
-			EndpointSpec: &swarm.EndpointSpec{},
-		},
-	}
+	service := &swarm.Service{}
+	defaults := []func(*swarm.Service){ServiceID("serviceID"), ServiceName("defaultServiceName")}
```
thaJeztah: Actually thinking that we could even remove these defaults; looks like it's currently not used in other locations where the code depends on it being set (which means we could convert this package to something that can be used _outside_ of testing as well

thaJeztah:
@silvin-lubecki updated; PTAL

PR #1612, opened 2019-01-14
Add support for maximum replicas per node without stack

olljanat:
**- What I did**
Added support for maximum replicas per node

Third step to be able solve:
* moby/moby#26259

**- How I did it**
Added new switch --replicas-max-per-node switch to docker service

**- How to verify it**
Create two services and specify ```--replicas-max-per-node``` one of them:
```
docker service create --detach=true --name web1 --replicas 2 nginx
docker service create --detach=true --name web2 --replicas 2 --replicas-max-per-node 1 nginx
```

See difference on command outputs:
```
$ docker service ls
ID                  NAME                MODE                REPLICAS               IMAGE               PORTS
0inbv7q148nn        web1                replicated          2/2                    nginx:latest        
9kry59rk4ecr        web2                replicated          1/2 (max 1 per node)   nginx:latest
        
$ docker service ps --no-trunc web2
ID                          NAME                IMAGE                                                                                  NODE                DESIRED STATE       CURRENT STATE            ERROR                                                     PORTS
bf90bhy72o2ry2pj50xh24cfp   web2.1              nginx:latest@sha256:b543f6d0983fbc25b9874e22f4fe257a567111da96fd1d8f1b44315f1236398c   limint              Running             Running 34 seconds ago                                                             
xedop9dwtilok0r56w4g7h5jm   web2.2              nginx:latest@sha256:b543f6d0983fbc25b9874e22f4fe257a567111da96fd1d8f1b44315f1236398c                       Running             Pending 35 seconds ago   "no suitable node (max replicas per node limit exceed)"   
```

**- Description for the changelog**
- Split from #1410 to simplify review

PR #2182, opened 2019-11-04
Add interpolation type cast for max_replicas_per_node

rumpl:
Fixes https://github.com/docker/app/issues/688

**- What I did**
Added interpolation type cast mapping for missing `deploy.placement.max_replicas_per_node`.

**- How to verify it**

Changed the unit test

image

### DIFF
--- a/cli/compose/loader/loader_test.go
+++ b/cli/compose/loader/loader_test.go
@@ -582,7 +582,7 @@ volumes:
 
 func TestLoadWithInterpolationCastFull(t *testing.T) {
 	dict, err := ParseYAML([]byte(`
-version: "3.7"
+version: "3.8"
 services:
   web:
     configs:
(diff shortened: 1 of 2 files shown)


cli/compose/loader/loader_test.go
```diff
@@ -582,7 +582,7 @@ volumes:
 
 func TestLoadWithInterpolationCastFull(t *testing.T) {
 	dict, err := ParseYAML([]byte(`
-version: "3.7"
+version: "3.9"
```
thaJeztah: Could you change this to version 3.8? This feature was added in 19.03 (through https://github.com/docker/cli/pull/1612), which has compose schema 3.8 as maximum

Changing to version 3.8 would allow us to back port this fix

rumpl: Done

PR #2192, opened 2019-11-11
Update dead link and add missing info on COPY

mdkalish:
1. Fix dead URL to [Dockerfile best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache).
2. Add missing information about cache invalidation by `COPY`. It works in the same way as in the case of `ADD`. Informing only about the `ADD`s behavior is misleading as one can infer that these two directives differ in this regard.

### DIFF
--- a/docs/reference/builder.md
+++ b/docs/reference/builder.md
@@ -743,8 +743,7 @@ flag, for example `docker build --no-cache`.
 See the [`Dockerfile` Best Practices
 guide](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/) for more information.
 
-The cache for `RUN` instructions can be invalidated by `ADD` instructions. See
-[below](#add) for details.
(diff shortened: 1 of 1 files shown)


docs/reference/builder.md
```diff
@@ -691,8 +691,7 @@ flag, for example `docker build --no-cache`.
 See the [`Dockerfile` Best Practices
 guide](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/build-cache) for more information.
 
-The cache for `RUN` instructions can be invalidated by `ADD` instructions. See
-[below](#add) for details.
+The cache for `RUN` instructions can be invalidated by [`ADD`](#add) and [`COPY`](#copy) instructions. See below for details.
```
thaJeztah, line 694: I think we can remove the `See below for details.` part, as the `ADD` and `COPY` words are now links

PR #2196, opened 2019-11-12
e2e: add new test package "global" with TestTLSVerify

tiborvass:

This is to move TestRunTLSVerify from moby integration-cli tests.

### DIFF
--- a/e2e/global/cli_test.go
+++ b/e2e/global/cli_test.go
@@ -1,0 +1,24 @@
+package global
+
+import (
+	"testing"
+
+	"github.com/docker/cli/internal/test/environment"
+	"gotest.tools/icmd"
+	"gotest.tools/skip"
+)
+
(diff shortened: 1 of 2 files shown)


e2e/global/cli_test.go
```diff
@@ -0,0 +1,24 @@
...
+	"testing"
+
+	"github.com/docker/cli/internal/test/environment"
+	"gotest.tools/icmd"
+	"gotest.tools/skip"
+)
+
+func TestTLSVerify(t *testing.T) {
+	// Remote daemons use TLS and this test is not applicable when TLS is required.
+	skip.If(t, environment.RemoteDaemon())
```
thaJeztah: Perhaps put the "skip" description as third argument to `skip.If`, then it shows up in the logs as reason why it's skipped

Issue #2214, opened 2019-12-05
[proposal] change "docker push" behavior to default to ":latest" instead of "all tags"


thaJeztah:
Implementation:

- [x] API client: https://github.com/moby/moby/pull/40302 client.ImagePush(): default to ":latest" instead of "all tags"
- [ ] CLI: https://github.com/docker/cli/pull/2220 implement docker push -a/--all-tags
- [ ] docs: (no PR yet)

The `docker push` command up until [v0.9.1](https://github.com/moby/moby/blob/v0.9.1/api/client.go#L998) always pushed all tags of a given image, so `docker push foo/bar` would push (e.g.) all of  `foo/bar:latest`, `foo:/bar:v1`, `foo/bar:v1.0.0`.

Pushing all tags of an image was not desirable in many case, so docker v0.10.0 enhanced `docker push` to optionally specify a tag to push (`docker push foo/bar:v1`) (see https://github.com/moby/moby/issues/3411 and the pull request that implemented this: https://github.com/moby/moby/pull/4948).

This behavior exists up until today, and is confusing, because unlike other commands, `docker push` does not default to use the `:latest` tag when omitted, but instead makes it push "all tags of the image"

For example, in the following situation;

```
docker images

REPOSITORY          TAG                        IMAGE ID            CREATED             SIZE
thajeztah/myimage   latest                     b534869c81f0        41 hours ago        1.22MB
```

Running `docker push thajeztah/myimage` seemingly does the expected behavior (it pushes `thajeztah/myimage:latest` to Docker Hub), however, it does not so for the reason expected (`:latest` being the default tag), but because `:latest` happens to be the only tag present for the `thajeztah/myimage` image.

If another tag exists for the image:

```
docker images

REPOSITORY          TAG                        IMAGE ID            CREATED             SIZE
thajeztah/myimage   latest                     b534869c81f0        41 hours ago        1.22MB
thajeztah/myimage   v1.0.0                     b534869c81f0        41 hours ago        1.22MB
```

Running the same command (`docker push thajeztah/myimage`) will push _both_ images to Docker Hub.

> Note that the behavior described above is currently not (clearly) documented; the `docker push` reference documentation (https://docs.docker.com/engine/reference/commandline/push/) does not mention that omitting the tag will push all tags

### Proposed change

First of all, it should be noted that changing this behavior will be a breaking change, so should be clearly mentioned in the release notes.

I think changing the behavior should be reversed, and a flag should be added to opt-in to the behavior:

[...]

PR #2220, opened 2019-12-09
implement docker push -a/--all-tags

thaJeztah:
fixes https://github.com/docker/cli/issues/2214

- [x] depends on https://github.com/docker/cli/pull/2242

depends on https://github.com/moby/moby/pull/40302 client.ImagePush(): default to ":latest" instead of "all tags"

The `docker push` command up until [v0.9.1](https://github.com/moby/moby/blob/v0.9.1/api/client.go#L998)
always pushed all tags of a given image, so `docker push foo/bar` would push (e.g.)
all of  `foo/bar:latest`, `foo:/bar:v1`, `foo/bar:v1.0.0`.

Pushing all tags of an image was not desirable in many case, so docker v0.10.0
enhanced `docker push` to optionally specify a tag to push (`docker push foo/bar:v1`)
(see https://github.com/moby/moby/issues/3411 and the pull request that implemented
this: https://github.com/moby/moby/pull/4948).

This behavior exists up until today, and is confusing, because unlike other commands,
`docker push` does not default to use the `:latest` tag when omitted, but instead
makes it push "all tags of the image"

For example, in the following situation;

```
docker images

REPOSITORY          TAG                        IMAGE ID            CREATED             SIZE
thajeztah/myimage   latest                     b534869c81f0        41 hours ago        1.22MB
```

Running `docker push thajeztah/myimage` seemingly does the expected behavior (it
pushes `thajeztah/myimage:latest` to Docker Hub), however, it does not so for the
reason expected (`:latest` being the default tag), but because `:latest` happens
to be the only tag present for the `thajeztah/myimage` image.

If another tag exists for the image:

```
docker images

REPOSITORY          TAG                        IMAGE ID            CREATED             SIZE
thajeztah/myimage   latest                     b534869c81f0        41 hours ago        1.22MB
thajeztah/myimage   v1.0.0                     b534869c81f0        41 hours ago        1.22MB
```

Running the same command (`docker push thajeztah/myimage`) will push _both_ images
to Docker Hub.

> Note that the behavior described above is currently not (clearly) documented;
> the `docker push` reference documentation (https://docs.docker.com/engine/reference/commandline/push/)
does not mention that omitting the tag will push all tags

This patch changes the default behavior, and if no tag is specified, `:latest` is
assumed. To push _all_ tags, a new flag (`-a` / `--all-tags`) is added, similar
to the flag that's present on `docker pull`.

With this change:

- `docker push myname/myimage` will be the equivalent of `docker push myname/myimage:latest`
- to push all images, the user needs to set a flag (`--all-tags`), so `docker push --all-tags myname/myimage:latest`

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/image/push.go
+++ b/cli/command/image/push.go
@@ -9,12 +9,15 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/streams"
 	"github.com/docker/distribution/reference"
+	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/registry"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 type pushOptions struct {
+	all       bool
 	remote    string
 	untrusted bool
 	quiet     bool
@@ -35,6 +38,7 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
 	}
 
 	flags := cmd.Flags()
+	flags.BoolVarP(&opts.all, "all-tags", "a", false, "Push all tagged images in the repository")
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress verbose output")
 	command.AddTrustSigningFlags(flags, &opts.untrusted, dockerCli.ContentTrustEnabled())
 
@@ -44,8 +48,16 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
 // RunPush performs a push against the engine based on the specified options
 func RunPush(dockerCli command.Cli, opts pushOptions) error {
 	ref, err := reference.ParseNormalizedNamed(opts.remote)
-	if err != nil {
+	switch {
+	case err != nil:
 		return err
+	case opts.all && !reference.IsNameOnly(ref):
+		return errors.New("tag can't be used with --all-tags/-a")
+	case !opts.all && reference.IsNameOnly(ref):
+		ref = reference.TagNameOnly(ref)
+		if tagged, ok := ref.(reference.Tagged); ok && !opts.quiet {
+			_, _ = fmt.Fprintf(dockerCli.Out(), "Using default tag: %s\n", tagged.Tag())
+		}
 	}
 
 	// Resolve the Repository name from fqn to RepositoryInfo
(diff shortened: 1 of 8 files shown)


docs/reference/commandline/push.md
```diff
@@ -68,16 +69,63 @@ this, tag the image with the host name or IP address, and the port of the
...
+```
+
+The image is now tagged under multiple names:
+
+```bash
+$ docker image ls
+
+REPOSITORY                      TAG                          IMAGE ID            CREATED             SIZE
+myimage                         latest                       6d5fcfe5ff17        2 hours ago         1.22MB
+localhost:5000/myname/myimage   latest                       6d5fcfe5ff17        2 hours ago         1.22MB
```
albers: The previous step tagged the image as *registry-host*, not *localhost*.

thaJeztah: oh! copied the output from my shell, and forgot to update it with the fake "registry-host" name for the docs.

thaJeztah:
e2e looks to be flaky/broken (don't think due to this PR). I was wondering why some PR's were green, but I noticed the _Jenkins_ jobs were not triggered on all PR's. I opened https://github.com/docker/cli/issues/2256 and https://github.com/docker/cli/issues/2257 (but they probably are the same issue)

thaJeztah:
@silvin-lubecki PTAL - kicked CI, and it's green now 👍

cli/command/image/push.go
```diff
@@ -33,6 +36,7 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
 	}
 
 	flags := cmd.Flags()
+	flags.BoolVarP(&opts.all, "all-tags", "a", false, "Push all tagged images in the repository")
```
silvin-lubecki: Do we really need `-a` ? I think that `all-tags` is short enough, and is not the main use case, not even a common use case  🤔

cli/command/image/push.go
```diff
@@ -42,8 +46,16 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
...
-	if err != nil {
+	switch {
+	case err != nil:
 		return err
+	case opts.all && !reference.IsNameOnly(ref):
+		return errors.New("tag can't be used with --all-tags/-a")
+	case !opts.all && reference.IsNameOnly(ref):
+		ref = reference.TagNameOnly(ref)
+		if tagged, ok := ref.(reference.Tagged); ok && !opts.quiet {
+			_, _ = fmt.Fprintf(dockerCli.Out(), "Using default tag: %s\n", tagged.Tag())
```
silvin-lubecki: nit: Was the linter complaining about the errors not checked for `fmt.Fprintf` ? 🤔

thaJeztah: I think we disabled the linter to check that, but my IDE still complained

We should create a utility for this to use everywhere (something like `printStdOut()` `printStdErr()`)

cli/command/image/push.go
thaJeztah: Yeah, it's consistent with `docker pull` which also has both

PR #2230, opened 2019-12-20
command/container: unify list tests in a single file

thaJeztah:
just a small nit: I noticed there was still a test in `ps_test.go`, but the file for `docker ps` / `docker container ls` was renamed to `list.go`, so moving that one test together with the others.

Also renamed the test from `TestBuildContainerListOptions` to `TestContainerListBuildContainerListOptions`, so that it has the same prefix as the other tests.


thaJeztah:
@vdemeester @silvin-lubecki ptal

PR #2231, opened 2019-12-23
Change example networks

oscrx:
**- What I did**
Change the example networks to allowed private ranges (192.168.0.0/16) according to [RFC1918](https://en.wikipedia.org/wiki/Private_network).
**- How I did it**
With the editor.
**- How to verify it**
Check the diff.
Change the network example to something allowed.

Cute doggo


oscrx:
Excuse me, I must not have payed enough attention. It should be fine now.

PR #2232, opened 2019-12-24
Fix some issues with docker ps --format

thaJeztah:
#### Fix docker ps --format with templating functions

Before this patch, using a template that used templating functions (such as
`lower` or `json`) caused the command to fail in the pre-processor step (in
`buildContainerListOptions`):

    docker ps --format='{{upper .Names}}'
    template: :1:8: executing "" at <.Names>: invalid value; expected string

This problem was due to the pre-processing using a different "context" type than
was used in the actual template, and custom functions to not be defined when
instantiating the Go template.

With this patch, using functions in templates works correctly:

    docker ps --format='{{upper .Names}}'
    MUSING_NEUMANN
    ELOQUENT_MEITNER

#### Fix docker ps table headers with custom format and "split" or "join"

Update the list of overrides for table headers so that columns using split or
join will produce the correct table header.

Before this patch:

    docker ps --format='table {{split .Names "/"}}'
    [NAMES]
    [unruffled_mclean]
    [eloquent_meitner]
    [sleepy_grothendieck]

With this patch applied:

    docker ps --format='table {{split .Names "/"}}'
    NAMES
    [unruffled_mclean]
    [eloquent_meitner]
    [sleepy_grothendieck]

### DIFF
--- a/cli/command/container/list.go
+++ b/cli/command/container/list.go
@@ -10,6 +10,7 @@ import (
 	"github.com/docker/cli/opts"
 	"github.com/docker/cli/templates"
 	"github.com/docker/docker/api/types"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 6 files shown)


thaJeztah:
Ah. booh

```
cli/command/formatter/container.go:110:1: exported method `ContainerContext.MarshalJSON` should have comment or be unexported (golint)
func (c *ContainerContext) MarshalJSON() ([]byte, error) {
^
cli/command/formatter/container.go:114:1: exported method `ContainerContext.ID` should have comment or be unexported (golint)
func (c *ContainerContext) ID() string {
^
```

thaJeztah:
@chris-crone @vdemeester ptal

PR #2243, opened 2020-01-07
Detect single-value advanced syntax

### LINKED ISSUE #2058: service create/update doesn't detect single-value advanced syntax
**Description**

The `--secret` and `--config` flags accept both a "basic" and "advanced" (csv / option=value) syntax.

Docker fails to detect the "advanced" syntax if there is no comma in the options passed, and as a result, attempts to use `option=value` as name for the secret.

1. Create a config and a secret:

    ```bash
    echo hello | docker config create myconfig -
    echo hello | docker secret create mysecret -
    ```

2. Create a service that uses the config, using the advanced ("csv" / "option=value") syntax

    ```bash
    docker service create --name hello1 --config source=myconfig nginx:alpine
    
    config not found: source=myconfig
    ```

3. Create a service that uses the secret, using the advanced ("csv" / "option=value") syntax

    ```bash
    docker service create --name hello2 --secret source=mysecret nginx:alpine
    
    secret not found: source=mysecret
    ```

4. Create a service that uses the config, using the "basic" syntax (just referencing the name)

    ```bash
    docker service create --name hello3 --config myconfig nginx:alpine
    
    08oc61ewfwbkxmasstkgv5yzw
    verify: Service converged 
    ```

4. Create a service that uses the config, using the "basic" syntax (just referencing the name)

    ```bash
    docker service create --name hello4 --secret mysecret nginx:alpine
    
    k0fki16c5anjsylywuc75xxy1
    verify: Service converged 
    ```

Both `--secret mysecret` and `--secret source=mysecret` should work

Secret (and config) names are restricted to `[a-zA-Z0-9-_.]`, so `--secret source=blabla` should not be ambiguous

```bash
echo blabla | docker secret create source=blabla -
Error response from daemon: rpc error: code = InvalidArgument desc = invalid name, only 64 [a-zA-Z0-9-_.] characters allowed, and the start and end character must be [a-zA-Z0-9]
```

### Proposed fix

In the "single value" case, check if the value contains a `=`, and if that's the case, don't return early

Parsing for secrets: https://github.com/docker/cli/blob/ae1618713f83e7da07317d579d0675f578de22fa/opts/secret.go#L34-L40
Parsing for configs: https://github.com/docker/cli/blob/ae1618713f83e7da07317d579d0675f578de22fa/opts/config.go#L34-L40

We should also check other flags that accept both "basic" and "advanced" syntax, to see if they have similar issues

```
 Version:           19.03.1
 API version:       1.40
 Go version:        go1.12.5
 Git commit:        74b1e89
 Built:             Thu Jul 25 21:18:17 2019
 Experimental:      true

 Engine:
  Version:          19.03.1
  Go version:       go1.12.5
  Git commit:       74b1e89
  Built:            Thu Jul 25 21:17:52 2019
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v1.2.6
  GitCommit:        894b81a4b802e4eb2a91d1ce216b8817763c29fb
 runc:
  Version:          1.0.0-rc8
  GitCommit:        425e105d5a03fabd737a126ad93d62a9eeede87f
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

zappy-shu:
**- What I did**

Adds the ability to just specify the `source` for config and secret flags using the advanced syntax as per https://docs.docker.com/engine/reference/commandline/service_create/#create-a-service-with-secrets. When the target is not specified the target is set to the source, same as when using the simple syntax.

**- How I did it**

When checking if the user is using the simple syntax check if the field contains an `=` as well as checking if there is more than one field.

**- How to verify it**

Unit tests have been added for just setting the source for both secrets and configs (all existing test cases for secrets have been copied for configs as well).

To manually test, follow the steps in #2058 

Fixes #2058 allowing the use of the advanced syntax when setting a config or secret with only the source field.

hedgehog-2019-11-07

### DIFF
--- a/opts/config_test.go
+++ b/opts/config_test.go
@@ -33,6 +33,7 @@ func TestConfigOptionsSource(t *testing.T) {
 	assert.Assert(t, is.Len(reqs, 1))
 	req := reqs[0]
 	assert.Check(t, is.Equal("foo", req.ConfigName))
+	assert.Check(t, is.Equal("foo", req.File.Name))
 }
 
 func TestConfigOptionsSourceTarget(t *testing.T) {
--- a/opts/config_test.go
+++ b/opts/config_test.go
@@ -8,23 +8,23 @@ import (
 	is "gotest.tools/assert/cmp"
 )
 
-func TestSecretOptionsSimple(t *testing.T) {
-	var opt SecretOpt
+func TestConfigOptionsSimple(t *testing.T) {
+	var opt ConfigOpt
 
-	testCase := "app-secret"
+	testCase := "app-config"
 	assert.NilError(t, opt.Set(testCase))
 
 	reqs := opt.Value()
 	assert.Assert(t, is.Len(reqs, 1))
 	req := reqs[0]
-	assert.Check(t, is.Equal("app-secret", req.SecretName))
-	assert.Check(t, is.Equal("app-secret", req.File.Name))
+	assert.Check(t, is.Equal("app-config", req.ConfigName))
+	assert.Check(t, is.Equal("app-config", req.File.Name))
 	assert.Check(t, is.Equal("0", req.File.UID))
 	assert.Check(t, is.Equal("0", req.File.GID))
 }
 
-func TestSecretOptionsSource(t *testing.T) {
-	var opt SecretOpt
+func TestConfigOptionsSource(t *testing.T) {
+	var opt ConfigOpt
 
 	testCase := "source=foo"
 	assert.NilError(t, opt.Set(testCase))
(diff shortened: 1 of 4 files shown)


opts/config_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
+)
+
+func TestConfigOptionsSimple(t *testing.T) {
```
silvin-lubecki: I think the tests here could be refactored to use a table driven test instead 🤔

zappy-shu: Presumably the same with the Secrets tests

opts/config_test.go
```diff
@@ -8,86 +8,79 @@ import (
...
+			configName: "foo",
+			fileName:   "testing",
+			uid:        "1000",
+			gid:        "1001",
+			fileMode:   0444,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
```
thaJeztah: Can you change this to

```suggestion
		tc := tc
		t.Run(tc.name, func(t *testing.T) {
```

see https://github.com/moby/moby/pull/40115#discussion_r337301763 (and https://blog.golang.org/subtests)

opts/secret_test.go
```diff
@@ -8,86 +8,79 @@ import (
...
+			secretName: "foo",
+			fileName:   "testing",
+			uid:        "1000",
+			gid:        "1001",
+			fileMode:   0444,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
```
thaJeztah: same here

```suggestion
		tc := tc
		t.Run(tc.name, func(t *testing.T) {
```

see https://github.com/moby/moby/pull/40115#discussion_r337301763 (and https://blog.golang.org/subtests)

Issue #1954, opened 2019-06-18
Docker wait command should exit with the container exit code


liranmauda:
**Description**

`docker wait` should exit with the container exit code.

When we are using `docker wait` it's exit status is equal to 0 and not to the containers exit status.

**Steps to reproduce the issue:**
1. `$ docker run -d --name test my_image`
a92fdad61acb3e27a809231338ec13dcfbe742bf2b38935c6dc50f097685beab
2. `$ docker wait test`
1
3. `$ echo $?`
0

**Describe the results you received:**
Docker wait prints the container's exit status (1) and its own exit status is 0

**Describe the results you expected:**
`docker wait` should exit with the container exit code

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        6247962
 Built:             Sun Feb 10 04:12:39 2019
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.2
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.6
  Git commit:       6247962
  Built:            Sun Feb 10 04:13:06 2019
  OS/Arch:          linux/amd64
  Experimental:     true
```

**Output of `docker info`:**

```
Containers: 12
 Running: 0
 Paused: 0
 Stopped: 12
Images: 412
Server Version: 18.09.2
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host ipvlan macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 9754871865f7fe2f4e74d43e2fc7ccd237edcbce
runc version: 09c8266bf2fcf9519a651b04ae54c967b9ab86ec
init version: fec3683
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.9.125-linuxkit
Operating System: Docker for Mac
OSType: linux
Architecture: x86_64
CPUs: 6
Total Memory: 1.952GiB
Name: linuxkit-025000000001
[...]

PR #2244, opened 2020-01-07
Docker wait command is now exiting with the container exit code

liranmauda:
**- What I did**
Returning the result.StatusCode in the wait command 

**- How to verify it**
Run a container that will fail:
`docker run -d --name test_wait_exitstatus centos:8 testme`

See that the command failed:
/go/src/github.com/docker/cli # `docker ps -a`
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS                   PORTS               NAMES
3658455b45a9        centos:8            "testme"            29 seconds ago      Created                                      test_wait_exitstatus

Run the wait command:
/go/src/github.com/docker/cli # `docker wait test_wait_exitstatus`
127

Run `echo $?` to see the wait command exit status
/go/src/github.com/docker/cli # `echo $?`
127

Docker wait command is now exiting with the container exit code

Fixes: #1954 

Signed-off-by: Liran Mauda <liran.mauda@gmail.com>

### DIFF
--- a/cli/command/container/wait.go
+++ b/cli/command/container/wait.go
@@ -42,6 +42,9 @@ func runWait(dockerCli command.Cli, opts *waitOptions) error {
 		select {
 		case result := <-resultC:
 			fmt.Fprintf(dockerCli.Out(), "%d\n", result.StatusCode)
+			if result.StatusCode != 0 {
+				return cli.StatusError{StatusCode: int(result.StatusCode)}
+			}
 		case err := <-errC:
 			errs = append(errs, err.Error())
 		}


thaJeztah:
Thanks for contributing; unfortunately this won't work, because `docker wait` can be used with _multiple_ containers, so there is not a _single_ exit code to return.

If you need this information, using the API directly could be a possible solution for your use-case, but I'm afraid we won't be able to accept this change.

I'll close this PR because of the above, but thanks again, and feel free to continue the conversation

Issue #2250, opened 2020-01-10
docker run -p with new syntax is backwards


phxnsharp:
**Description**

If you use the new expanded syntax for publishing ports with the 'docker run' command, the "sense" is backwards from the equivalent option in 'docker service create'.

Granted, the documentation for 'docker run' does not state you can use the new syntax. However, it accepts the format and then does the wrong thing.

**Steps to reproduce the issue:**
On a one node swarm, these two commands have the same basic effect:
```
docker service create -p target=80,published=8123 nginx
docker run -p target=8123,published=80 nginx
```
Needless to say, as a new user to Docker I was a little confused for a while :)

**Describe the results you received:**
docker run publishes the target and targets the published

**Describe the results you expected:**
docker run targets the target and publishes the published

**Additional information you deem important (e.g. issue happens only occasionally):**
100% reliable

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.12.12
 Git commit:        633a0ea838
 Built:             Wed Nov 13 07:29:52 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.5
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.12
  Git commit:       633a0ea838
  Built:            Wed Nov 13 07:28:22 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Output of `docker info`:**

```
Client:
 Debug Mode: false

Server:
 Containers: 11
  Running: 0
  Paused: 0
  Stopped: 11
 Images: 9
 Server Version: 19.03.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: active
[...]

PR #2252, opened 2020-01-14
Reverse order of long-form ports

zappy-shu:
**- What I did**

Reverses the order long-form port options when converted to short-form to correctly match the documentation and `docker service create`. See https://docs.docker.com/engine/reference/commandline/service_create/#publish-service-ports-externally-to-the-swarm--p---publish

Post change `-p published=8111,target=8112` is the equivalent of `8111:8112`

**- How I did it**

Changed the order of the values when converting from long to short-form when the options are parsed

**- How to verify it**

Build and run `docker run -d --name run_test -p published=8111,target=80 nginx`. nginx should be accessible from the host with port 8111

Fixes https://github.com/docker/cli/issues/2250

cat-2019-11-19

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -808,7 +808,7 @@ func parsePortOpts(publishOpts []string) ([]string, error) {
 
 			params[opt[0]] = opt[1]
 		}
-		optsList = append(optsList, fmt.Sprintf("%s:%s/%s", params["target"], params["published"], params["protocol"]))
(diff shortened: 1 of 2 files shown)


thaJeztah:
Looks like this was added in https://github.com/docker/cli/commit/29612ccefecfe6017f32001e6289480a5d276eda, which was part of https://github.com/docker/cli/pull/1102

So this issue affects docker 18.09 and 19.03

zappy-shu:
@silvin-lubecki PTAL

PR #2277, opened 2020-01-21
Bump mergo v0.3.8

zappy-shu:
**- What I did**

Bumps mergo from v0.3.7 to v0.3.8

full diff: imdario/mergo@v0.3.7...v0.3.8

includes:

* imdario/mergo#112 Add strict override
  * fixes imdario/mergo#111 WithOverride should be able to check types
* imdario/mergo#106 Fix merging of interface types with concrete values
* imdario/mergo#120 should not overwrite pointers directly, instead check embedded values
  * fixes imdario/mergo#114 Embedded struct of pointer types will overwrite the whole destination struct
* imdario/mergo#125 added WithOverrideEmptySlice config flag

**- How I did it**

Updated the vendor and added 2 special transforms to handle a change in how pointers in maps are handled in mergo: https://github.com/imdario/mergo/compare/v0.3.7...v0.3.8#diff-a103b5fee5fb31abebf12ee067c01da8R153

These transforms ensure that:
* ulimit config overrides correctly switch between single and soft/hard instead of merging the 2 together.
* service network config aliases are replaced with the override list instead of the lists being appended.

**- How to verify it**

The existing unit tests cover the special transforms as they have been put in place to preserve the existing functionality

Bumps mergo dependency from v0.3.7 to v0.3.8

cat-2019-11-19

### DIFF
--- a/cli/compose/loader/merge_test.go
+++ b/cli/compose/loader/merge_test.go
@@ -1,9 +1,11 @@
 package loader
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/docker/cli/cli/compose/types"
+	"github.com/imdario/mergo"
 	"gotest.tools/assert"
 )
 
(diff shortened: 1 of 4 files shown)


silvin-lubecki:
The linter is complaining 🦁 
```
cli/compose/loader/merge.go:206:49: mergeUlimitsConfig - result 0 (error) is always nil (unparam)
func mergeUlimitsConfig(dst, src reflect.Value) error {
                                                ^
cli/compose/loader/merge.go:213:56: mergeServiceNetworkConfig - result 0 (error) is always nil (unparam)
func mergeServiceNetworkConfig(dst, src reflect.Value) error {
```

silvin-lubecki:
@thaJeztah PTAL

cli/compose/loader/merge_test.go
```diff
@@ -1,8 +1,11 @@
 package loader
 
 import (
+	"reflect"
 	"testing"
 
+	"github.com/imdario/mergo"
+
```
thaJeztah: nit; should be grouped with the imports below

thaJeztah:
all green; merging!

Issue #2290, opened 2020-01-27
[RFD] add configuration option to add `host.docker.internal` by default


thaJeztah:
Pull request https://github.com/moby/moby/pull/40007 implemented a "magic" `host-gateway` value to allow users to add `host.docker.internal` to `/etc/hosts` in containers, mapping it to the gateway IP address (the IP address itself can be configured in the daemon configuration).

With this feature, users running dockerd on Linux can get parity with Docker for Mac / Docker for Windows to have `host.docker.internal` resolve inside containers.

Currently, this option is "opt-in" on Linux; containers will not be able to resolve `host.docker.internal` unless they are created with `--add-host host.docker.internal:host-gateway`

### Proposal

Users may want to have the `host.docker.internal` host added _automatically_ for every container created (similar to how it's currently added automatically on Docker Desktop (Docker for Mac / Docker for Windows).

I suggest to:

- add an option to the CLI configuration file (`~/.docker/config.json`) that allows configuring a default (set of) custom hosts for new containers
- possibly (t.b.d.) an environment variable to configure this

The configuration option could potentially be used by the Docker Desktop team, as a replacement for the current (proprietary) implementation; doing so would make the implementation on Linux and Docker Desktop identical (which would make it less confusing).

rfay:
Please don't forget docker-compose as well. Thanks!

thaJeztah:
Good point, yes: I'm not sure to what extend `docker-compose` currently reads/uses the CLI configuration file, so that's something to be looked at. @ndeloof @ulyssessouza

silvin-lubecki:
I think this feature makes more sense per context basis. One global configuration won't allow fine grain control, as I guess that if you target different environments, you don't want this feature enabled everywhere (ex maybe on prod environment?).

ibaldonl:
The host is always accessible by default, this only adds a well known DNS reference to it, so it should be always there by default too I think, and all the configuration complexity is not warranted for this case.
Extreme configuration and flexibility is the enemy of simplicity so it should be a balance; in this case, I think that it shouldn't be configurable at all, it just should always exist and doing the right thing in all platforms without users worrying about it.
Just wanted to express this personal opinion for your consideration guys, thanks a lot for reading and for your time.

ibaldonl:
Sorry, let me elaborate with an example: you do a Docker Compose environment and hope it works mostly without too many tweaks and assumptions about the users environments.

Docker Compose definitions are terse and to the point, they are simple and straightforward.

[...]

PR #2293, opened 2020-01-27
Skip IPAddr validation for "host-gateway" string

arkodg:
Relates to - https://github.com/moby/moby/pull/40007
The above PR added support in moby, that detects if
a special string "host-gateway" is added to the IP
section of --add-host, and if true, replaces it with
a special IP value (value of --host-gateway-ip Daemon flag
which defaults to the IP of the default bridge).

This PR is needed to skip the validation for the above
feature

Signed-off-by: Arko Dasgupta <arko.dasgupta@docker.com>

### DIFF
--- a/opts/hosts.go
+++ b/opts/hosts.go
@@ -2,6 +2,7 @@ package opts
 
 import (
 	"fmt"
+	_ "github.com/docker/docker/daemon/network"
 	"net"
 	"net/url"
 	"strconv"
--- a/vendor.conf
+++ b/vendor.conf
@@ -12,7 +12,7 @@ github.com/creack/pty                               3a6a957789163cacdfe0e291617a
 github.com/davecgh/go-spew                          8991bc29aa16c548c550c7ff78260e27b9ab7c73 # v1.1.1
 github.com/docker/compose-on-kubernetes             78e6a00beda64ac8ccb9fec787e601fe2ce0d5bb # v0.5.0-alpha1
 github.com/docker/distribution                      0d3efadf0154c2b8a4e7b6621fff9809655cc580
-github.com/docker/docker                            a9507c6f76627fdc092edc542d5a7ef4a6df5eec
+github.com/docker/docker                            10425ed4cb855bc6d0e51b83084d2eaa1bab9ad5 
 github.com/docker/docker-credential-helpers         54f0238b6bf101fc3ad3b34114cb5520beb562f5 # v0.6.3
 github.com/docker/go                                d30aec9fd63c35133f8f79c3412ad91a3b08be06 # Contains a customized version of canonical/json and is used by Notary. The package is periodically rebased on current Go versions.
 github.com/docker/go-connections                    7395e3f8aa162843a74ed6d48e79627d9792ac55 # v0.4.0
--- a/vendor/github.com/docker/docker/api/types/client.go
+++ b/vendor/github.com/docker/docker/api/types/client.go
@@ -205,7 +205,7 @@ const (
 	// BuilderV1 is the first generation builder in docker daemon
 	BuilderV1 BuilderVersion = "1"
 	// BuilderBuildKit is builder based on moby/buildkit project
-	BuilderBuildKit = "2"
+	BuilderBuildKit BuilderVersion = "2"
 )
 
 // ImageBuildResponse holds information
--- a/vendor/github.com/docker/docker/api/types/swarm/service.go
+++ b/vendor/github.com/docker/docker/api/types/swarm/service.go
@@ -17,6 +17,10 @@ type Service struct {
 	// listing all tasks for a service, an operation that could be
 	// computation and network expensive.
 	ServiceStatus *ServiceStatus `json:",omitempty"`
+
+	// JobStatus is the status of a Service which is in one of ReplicatedJob or
+	// GlobalJob modes. It is absent on Replicated and Global services.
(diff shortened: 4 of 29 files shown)


silvin-lubecki:
Related to #2290

opts/hosts.go
```diff
@@ -24,6 +24,9 @@ var (
 	DefaultTLSHost = fmt.Sprintf("tcp://%s:%d", DefaultHTTPHost, DefaultTLSHTTPPort)
 	// DefaultNamedPipe defines the default named pipe used by docker on Windows
 	DefaultNamedPipe = `//./pipe/docker_engine`
+	// HostGatewayName defines a special string which users can append to --add-host
+	// to add an extra entry in /etc/hosts that maps host.docker.internal to the host IP
+	HostGatewayName = "host-gateway"
```
thaJeztah: This should be a `const`; also wondering if there's a const for this in moby/moby that we can use

opts/hosts.go
```diff
@@ -160,8 +163,11 @@ func ValidateExtraHost(val string) (string, error) {
 	if len(arr) != 2 || len(arr[0]) == 0 {
 		return "", fmt.Errorf("bad format for add-host: %q", val)
 	}
-	if _, err := ValidateIPAddress(arr[1]); err != nil {
-		return "", fmt.Errorf("invalid IP address in add-host: %q", arr[1])
+	// Skip IPaddr validation for "host-gateway" string
+	if arr[1] != HostGatewayName {
+		if _, err := ValidateIPAddress(arr[1]); err != nil {
```
thaJeztah: Actually slightly on the fence if we should validate _at all_ in the cli, or if we should just pass it to the daemon and have the daemon validate / produce an error (haven't checked yet what the UX looks like in that case)

opts/hosts.go
silvin-lubecki: I think this value should be used directly from [moby](https://github.com/moby/moby/pull/40007/files#diff-65c8668a717635f8e126e71c7213f537R7)

thaJeztah:
Tested what it looks like if we remove the IP address validation;

```bash
docker create --add-host foo:fooobar nginx:alpine
Error response from daemon: invalid IP address in add-host: "fooobar"

docker run --add-host foo:fooobar nginx:alpine
docker: Error response from daemon: invalid IP address in add-host: "fooobar".
```

And, I think those errors look reasonable (have not tried `docker stack deploy` though)

thaJeztah:
`docker service create` also validates on the service, but appears to be retrying creation; perhaps it's only validating the moment the container is created and not when the _service_ is created;

```bash
docker service create --host foo:foooobar nginx:alpine
opbff21havwdz5fyrjk68so11
1/1: invalid IP address in add-host: "foooobar"

...

1/1: assigned  [======================>                            ]

...

1/1: invalid IP address in add-host: "foooobar"
```

arkodg:
sure, there is already an identical validation in the daemon, and this PR is just here to unblock this feature.
we can completely remove `ValidateHost` from the `cli`, but I'm not sure what the design philosophy is for cli and daemon arg validation

thaJeztah:
Overall, I prefer delegating validation to the daemon as much as possible; at _most_ validating well-formedness of values (which in this case was possible before).

Keeping the daemon as source of truth not only keeps complexity out of the cli code, it also prevents situations where the cli talks to a different (older/newer) version of the daemon, or (eg) a daemon running on a different platform.

We should probably (as a follow up) look at the `docker service create` example though, because that UX isn't ideal (if possible, validate earlier, or perhaps give up if the config is invalid instead of keep retrying)

arkodg:
@thaJeztah IMHO for the `docker service create` case, I think keeping the logic here makes sense, updated the PR removing the `const`

silvin-lubecki:
@arkodg The vendoring looks weird:
```
These files were changed:

?? vendor/github.com/opencontainers/runc/libcontainer/cgroups/
?? vendor/github.com/opencontainers/runc/libcontainer/configs/
```

silvin-lubecki:
Please also rebase 🦁

thaJeztah:
@arkodg could you;

- move the vendor update to a separate commit (first commit)
    - only including changes needed related to that update (I see some tests failing, which could be related to changes upstream)
- move local changes to for this feature to a commit coming after that (combine the local changes in the current commits)

arkodg:
@thaJeztah the issue is with `make cross` using the latest `docker/docker` reappoint

thaJeztah:
opened https://github.com/docker/cli/pull/2333 to update vendoring

thaJeztah:
@arkodg I think this was a bad rebase, as I see no local code-changes anymore, other than adding the import 🤔 

Noticed this when I rebased the PR (also the commit message was now only about `vndr`); see https://github.com/docker/cli/compare/master...thaJeztah:carry_2293_add_host_gateway_dns_entry?expand=1

thaJeztah:
@silvin-lubecki found an old commit here; https://github.com/docker/cli/commit/709cf07122c91e111c53edd4457e48fd87777fb1

Updating my branch with that

arkodg:
@thaJeztah I removed the old commit from the PR after you suggested to rc the CI issue by separating the commits to a vendor commit and the feature commit :)

yuraku:
Hi guys. Is there any ETA on when this PR will be merged? 🤞🏻

PR #2296, opened 2020-01-28
Builder docs update [carry 1889]

thaJeztah:
closes https://github.com/docker/cli/pull/1889
carry of https://github.com/docker/cli/pull/1889 with:

- the text wrapped to 80 chars
- some "notes" update to use the "note" formatting, and to link to the BuildKit section for instructions on enabling buildkit)
- the examples expanded with some detailed description
- an example on using a separate `FROM scratch` stage to use for exporting


thaJeztah:
@tonistiigi @AkihiroSuda @tiborvass LMK if the changes I made LGTY

docs/reference/commandline/build.md
```diff
@@ -489,6 +499,137 @@ FROM alpine AS production-env
...
+for subsequent builds.
+
+Upon importing the cache, the builder will only pull the JSON metadata from the
+registry and determine possible cache hits based on that information. If there
+is a cache hit, the matched layers are pulled into the local environment.
+
+In addition to images, the cache can also be pulled from special cache manifests
+generated by [`buildx`](https://github.com/docker/buildx) or the BuildKit CLI
+(`buildctl`). These manifests allow pulling layer data for intermediate stages
+in multi-stage builds.
```
AkihiroSuda: Can we clarify that these sentences are about `mode=max` cache?

thaJeztah: (Reading up on `mode=max`);

> - `mode=min (default): only export layers for the resulting image
> - `mode=max`: export all the layers of all intermediate steps. Not supported for inline cache exporter.

So, we should mention that when using `buildctl` (and `builds`?) this only works if those images were build by those using the `mode=max` option?

thaJeztah: Something like this?

```markdown
In addition to images, the cache can also be pulled from special cache manifests
 generated by [`buildx`](https://github.com/docker/buildx) or the BuildKit CLI
 (`buildctl`) when built with the `type=inline` and `mode=max` option.
```

(my only concern there would be that people would go looking for `mode=max` on the `docker` cli

thaJeztah: oh, probably only applies to that last sentence, so;

```markdown
These manifests (when built with the `type=inline` and `mode=max` options)
allow pulling layer data for intermediate stages in multi-stage builds.
```

AkihiroSuda: max is only supported for type=registry

thaJeztah: @AkihiroSuda @tonistiigi updated; PTAL

thaJeztah:
ping @AkihiroSuda @tonistiigi PTAL 🙏

silvin-lubecki:
It's just documentation PR, so CI failure is unrelated.

PR #2307, opened 2020-02-02
Add final report to sytem prune command

### LINKED ISSUE #2306: Feature: final report with total objects deleted per resource type in system prune
Hi,

I think this will be super helpful when managing resources. The default output tends to be really huge depending on the number of resources dangling. This potential feature would add a final report at the end of the `system prune` command making it easier to read and interpret.

image

GustavoKatel:
Signed-off-by: Gustavo Sampaio <gbritosampaio@gmail.com>

This closes #2306
This is also my first PR, so let me know if I missed anything from the guidelines. Also let me know if I missed any test case. Thanks :D

**- What I did**

Added a final report into the `system prune` command

**- How I did it**

Refactored the prune function of each resource to return the number of deleted objects in the process.

**- How to verify it**

Run the following command: `./build/docker-linux-amd64 system prune -a --volumes -f` with some dangling resources.

Refactored the prune function of each resource to return the number of deleted objects in the process. Aggregated the results in a final report to be presented after the `system prune` command.

Here is my dorky cat Penumbra.
image

### DIFF
--- a/cli/command/builder/prune.go
+++ b/cli/command/builder/prune.go
@@ -29,13 +29,14 @@ func NewPruneCommand(dockerCli command.Cli) *cobra.Command {
 		Short: "Remove build cache",
 		Args:  cli.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			spaceReclaimed, output, err := runPrune(dockerCli, options)
+			objectsDeleted, spaceReclaimed, output, err := runPrune(dockerCli, options)
 			if err != nil {
 				return err
 			}
 			if output != "" {
 				fmt.Fprintln(dockerCli.Out(), output)
 			}
+			fmt.Fprintln(dockerCli.Out(), "Total caches deleted:", objectsDeleted)
 			fmt.Fprintln(dockerCli.Out(), "Total reclaimed space:", units.HumanSize(float64(spaceReclaimed)))
 			return nil
 		},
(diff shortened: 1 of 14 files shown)


devdreamsolution (approved):

thaJeztah:
Sorry for the long delay; we are evaluating the UX for these commands, and looking if we can improve the output overall.

That said, we're wondering if this additional information adds much value; do you have a specific example how you use this information? (and what you use it for?)

GustavoKatel:
> Sorry for the long delay; we are evaluating the UX for these commands, and looking if we can improve the output overall.
> 
> That said, we're wondering if this additional information adds much value; do you have a specific example how you use this information? (and what you use it for?)

to be honest, the current output shows too much info and at the same not that much. It's mostly a huge waterfall of hashes and object ids, see:

```
# docker system prune --all --volumes -f
Deleted Networks:

---- Several lines of names omitted ----

Deleted Images:

---- Several lines of hashes omitted ----

Deleted build cache objects:

---- Several lines of object ids omitted ----

Total reclaimed space: 17.38GB
```

At the end the only thing I can managed to clearly identify is the total reclaimed space. To take a sense of anything else we need to scroll back and find a way to count/visually parse then. This is not critical ofc, but I think it's a nice addition to the output and helps you to keep an eye what's been filling up. Specially for resource constrained devices.

That said and considering this was a long time ago, maybe a better way to get that info has emerged?

Anyway, happy to work on the conflicts if you all consider this has value.

PR #2308, opened 2020-02-04
Add support for Kubernetes username/password auth

simonferquel:
**- What I did**
Added support for username/password Kubernetes authentication. The main motivation being to support k3s Kubernetes distributions.

**- How I did it**
Don't discard username/password fields when loading a kubeconfig file

**- How to verify it**
The PR comes with a unit test (same style as gke and eks test cases)

- Docker stack --orchestrator=kubernetes support now support username/password Kubernetes authentication (adding support for k3s)


cli/context/kubernetes/save.go
```diff
@@ -39,6 +39,13 @@ func FromKubeConfig(kubeconfig, kubeContext, namespaceOverride string) (Endpoint
 			Key:  key,
 		}
 	}
+	var usernamePassword *UsernamePassword
+	if clientcfg.Username != "" || clientcfg.Password != "" {
```
thaJeztah: ~Do we need the check here? Of both are empty, we'll just set an empty string (which is the default for strings)~

oh, nevermind; you created a pointer so that we don't set the `UsernamePassword` field if we don't use it

cli/context/kubernetes/load.go
```diff
@@ -21,6 +21,13 @@ type EndpointMeta struct {
 	DefaultNamespace string                           `json:",omitempty"`
 	AuthProvider     *clientcmdapi.AuthProviderConfig `json:",omitempty"`
```
thaJeztah: Curious; what's `AuthProvider` used for; is username/password not some form of that? (looking at https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/api/types.go#L153-L177 it also has some sanitisation built in, so was wondering)

simonferquel: AuthProvider is for authentication plugins (such as GKE authentication plugin). They allow to put arbitrary json in the kubeconfig file and feed the plugin with that so that it can setup the rest.Client accordingly. It is not related to the basic auth scheme.

simonferquel:
@silvin-lubecki would probably want to have a look as well.

PR #2309, opened 2020-02-06
Fix a typo and enhance a script in an example

Abreto:
Hi everyone,

**- What I did**

image

I can't figure out what does `awk '{print \$2}'` mean, and actually it cannot run with `GNU Awk 4.1.4, API: 1.1 (GNU MPFR 4.0.1, GNU MP 6.1.2)`. So I think it might be a typo. It should be `awk '{print $2}'`.

What's more, if the host has several IPs on `eth0`, the script cannot get `${HOSTIP}` correctly, unless appended by `| sed -n '1p'`.

**- How I did it**

Removed the extra `\`. Appended `| sed -n 1p` at the end.

**- How to verify it**

image

### DIFF
--- a/docs/reference/commandline/run.md
+++ b/docs/reference/commandline/run.md
@@ -690,7 +690,7 @@ using IPv4 or IPv6 networking in your containers. Use the following
 flags for IPv4 address retrieval for a network device named `eth0`:
 
 ```bash
-$ HOSTIP=`ip -4 addr show scope global dev eth0 | grep inet | awk '{print \$2}' | cut -d / -f 1`
+$ HOSTIP=`ip -4 addr show scope global dev eth0 | grep inet | awk '{print $2}' | cut -d / -f 1 | sed -n 1p`
 $ docker run  --add-host=docker:${HOSTIP} --rm -it debian
 ```
 


thaJeztah:
I think the `\` was added to prevent the `$2` from being substituted (perhaps it was taken from a `docker run`), but I don't think it's needed (as it's within single quotes).

I'm not able to reproduce an issue with both cases though;

```bash
root@swarm-test-01:~# HOSTIP=`ip -4 addr show scope global dev eth0 | grep inet | awk '{print \$2}' | cut -d / -f 1`
root@swarm-test-01:~# echo $HOSTIP
10.18.0.9
root@swarm-test-01:~# HOSTIP=`ip -4 addr show scope global dev eth0 | grep inet | awk '{print $2}' | cut -d / -f 1`
root@swarm-test-01:~# echo $HOSTIP
10.18.0.9

root@swarm-test-01:~# awk --version
GNU Awk 4.1.4, API: 1.1 (GNU MPFR 4.0.1, GNU MP 6.1.2)
```

> What's more, if the host has several IPs on eth0, the script cannot get ${HOSTIP} correctly, unless appended by | sed -n '1p'.

Good point; I guess the example was added just to "illustrate" how it could be done, but it makes sense to add that; perhaps you can update this PR?

Note that a feature is worked on to make this easier (see https://github.com/moby/moby/pull/40007)

Abreto:
Well, you are right. The fault was caused by my running the command `ip -4 addr show scope global dev eth0 | grep inet | awk '{print \$2}' | cut -d / -f 1` directly in shell without ` `` `. When wrapped with ` `` `, they seem to be identical:
```
$ set -x
$ echo `echo '\$2' '$2'`
+ echo $2 $2
+ echo $2 $2
$2 $2
```

I would recover the `\` and update this PR to enhance the script in terms of handling multiple IPs.

thaJeztah:
> Should I recover the `\`?

Well, it didn't seem to add anything, so I'm ok with leaving it out 😂

Abreto:
Thank you. I might have squashed them now.

PR #855, opened 2018-01-31
Add examples for configs

mistyhacks:
Signed-off-by: Misty Stanley-Jones <misty@docker.com>

Fixes https://github.com/docker/docker.github.io/issues/5878

cc/ @mbentley 

baby starfish

PR #2312, opened 2020-02-06
Add examples for configs (carry 855)

thaJeztah:
carries https://github.com/docker/cli/pull/855
closes https://github.com/docker/cli/pull/855

fixes https://github.com/docker/docker.github.io/issues/5878

cc/ @mbentley 

baby starfish

### DIFF
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -200,11 +200,40 @@ $ docker service create --name redis \
 
 To grant a service access to multiple secrets, use multiple `--secret` flags.
 
-Secrets are located in `/run/secrets` in the container.  If no target is
-specified, the name of the secret will be used as the in memory file in the
-container.  If a target is specified, that will be the filename.  In the
-example above, two files will be created: `/run/secrets/ssh` and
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/service_create.md
```diff
@@ -200,12 +200,41 @@ $ docker service create --name redis \
...
+```bash
+$ docker service create --name redis \
+  --config source=redis-conf,target=/etc/redis/redis.conf,mode=0400 redis:3.0.6
+```
+
+To grant a service access to multiple configs, use multiple `--config` flags.
+
+Configs are located in `/` in the container if no target is specified. If no
+target is specified, the name of the config is used as the name of the file in
+the container. If a target is specified, that is the filename.
```
silvin-lubecki: `that is used as the filename` ?

thaJeztah: yeah, that was the original text, but I also struggled a bit with that sentence; let me give it a try

thaJeztah:
nit fixed! I'll merge once CI completes, and will prepare a cherry-pick

PR #2337, opened 2020-02-13
Skip IPAddr validation for "host-gateway" string

thaJeztah:
carry of https://github.com/docker/cli/pull/2293
closes https://github.com/docker/cli/pull/2293

Relates to - https://github.com/moby/moby/pull/40007
The above PR added support in moby, that detects if
a special string "host-gateway" is added to the IP
section of --add-host, and if true, replaces it with
a special IP value (value of --host-gateway-ip Daemon flag
which defaults to the IP of the default bridge).

This PR is needed to skip the validation for the above
feature

### DIFF
--- a/opts/hosts.go
+++ b/opts/hosts.go
@@ -24,6 +24,12 @@ var (
 	DefaultTLSHost = fmt.Sprintf("tcp://%s:%d", DefaultHTTPHost, DefaultTLSHTTPPort)
 	// DefaultNamedPipe defines the default named pipe used by docker on Windows
 	DefaultNamedPipe = `//./pipe/docker_engine`
+	// hostGatewayName defines a special string which users can append to --add-host
+	// to add an extra entry in /etc/hosts that maps host.docker.internal to the host IP
+	// TODO Consider moving the HostGatewayName constant defined in docker at
+	// github.com/docker/docker/daemon/network/constants.go outside of the "daemon"
+	// package, so that the CLI can consume it.
+	hostGatewayName = "host-gateway"
 )
 
 // ValidateHost validates that the specified string is a valid host and returns it.
@@ -160,8 +166,11 @@ func ValidateExtraHost(val string) (string, error) {
 	if len(arr) != 2 || len(arr[0]) == 0 {
 		return "", fmt.Errorf("bad format for add-host: %q", val)
 	}
-	if _, err := ValidateIPAddress(arr[1]); err != nil {
-		return "", fmt.Errorf("invalid IP address in add-host: %q", arr[1])
+	// Skip IPaddr validation for "host-gateway" string
+	if arr[1] != hostGatewayName {
+		if _, err := ValidateIPAddress(arr[1]); err != nil {
+			return "", fmt.Errorf("invalid IP address in add-host: %q", arr[1])
+		}
 	}
 	return val, nil
 }


opts/hosts.go
```diff
@@ -6,6 +6,8 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+
+	"github.com/docker/docker/daemon/network"
```
thaJeztah: Perhaps we should move the constant to somewhere in the client/api package; we're now importing the daemon package, which is not really ideal

thaJeztah:
Working on updated documentation for this, but splitting it in "enhancements to be backported" and "for next release"

arkodg:
@thaJeztah and @silvin-lubecki  thanks for taking care of this !

thaJeztah:
@arkodg you're welcome! And sorry for the back-and-forth on the `const`. I didn't realise it was in the "daemon" package at the time.

jcoutch:
@thaJeztah - Will support for `host-gateway` be added to [docker/compose](https://github.com/docker/compose)?  While it works using the docker cli directly, it looks like compose has it's own IP address validation.

thaJeztah:
@jcoutch it's supported by the API / daemon, so would be able to be supported by compose if the validation is removed; can you open a ticket in the docker compose repository? (or perhaps even better; a pull request if you're familiar with Python

thaJeztah:
@jcoutch I just tested docker compose against a daemon that supports this feature;

```bash
docker-compose version

docker-compose version 1.26.0-rc3, build 46118bc5
docker-py version: 4.2.0
CPython version: 3.7.6
OpenSSL version: OpenSSL 1.1.1d  10 Sep 2019
```

Using this compose file:

```yaml
version: '3.6'
services:
  nginx:
    image: nginx:alpine
    extra_hosts:
      - "foo:127.0.0.1"
      - "bar:127.0.0.2"
      - "host.docker.internal:host-gateway"
```

Trying to run it on a docker 19.03 daemon produces an error (as expected);

```
docker-compose up -d
Creating network "extra-hosts_default" with the default driver
Creating extra-hosts_nginx_1 ... error

ERROR: for extra-hosts_nginx_1  Cannot create container for service nginx: invalid IP address in add-host: "host-gateway"

ERROR: for nginx  Cannot create container for service nginx: invalid IP address in add-host: "host-gateway"
ERROR: Encountered errors while bringing up the project.
```

On a docker daemon built from master (so with support for the `host-gateway` feature);

```bash
docker-compose up -d

Creating network "extra-hosts_default" with the default driver
Creating extra-hosts_nginx_1 ... done
```

Verifying the container that's created:

```bash
docker inspect --format='{{json .HostConfig.ExtraHosts}}' extra-hosts_nginx_1
```

Which produces:

```json
[
  "foo:127.0.0.1",
  "bar:127.0.0.2",
  "host.docker.internal:host-gateway"
]
```

And verifying that the magic `host-gateway` value is replaced correctly inside the container;

```bash
docker exec extra-hosts_nginx_1 cat /etc/hosts

127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::0	ip6-localnet
ff00::0	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
127.0.0.1	foo
127.0.0.2	bar
172.18.0.1	host.docker.internal
172.20.0.2	2eb3a4602c9a
```

PR #2355, opened 2020-02-21
fix panic on single-character volumes

thaJeztah:
fixes https://github.com/moby/moby/issues/40550 single digit volume name leads to 'panic: runtime error: index out of range'

Before this change, this would cause a panic:

    docker run -it --rm -v 1:/1 alpine
    panic: runtime error: index out of range

    goroutine 1 [running]:
    github.com/docker/cli/cli/compose/loader.isFilePath(0xc42027e058, 0x1, 0x557dcb978c20)
    ...

After this change, a correct error is returned:

    docker run -it --rm -v 1:/1 alpine
    docker: Error response from daemon: create 1: volume name is too short, names should be at least two alphanumeric characters.

```
* Fix panic: runtime error: index out of range when trying to use a single-letter volume
```

### DIFF
--- a/cli/compose/loader/volume.go
+++ b/cli/compose/loader/volume.go
@@ -111,6 +111,9 @@ func isFilePath(source string) bool {
 	case '.', '/', '~':
 		return true
 	}
+	if len([]rune(source)) == 1 {
+		return false
+	}
 
 	// windows named pipes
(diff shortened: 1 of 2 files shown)


thaJeztah:
ping @ndeloof - you probably want to cherry-pick this as well

PR #2360, opened 2020-02-26
Update Golang 1.13.8

thaJeztah:
Also pinning the e2e image to the "buster" variant,
which is what's currently used, but making it explicit.


thaJeztah:
@silvin-lubecki PTAL

Issue #956, opened 2018-03-21
Review winresources autogenerated package


thaJeztah:
related issue for the daemon; https://github.com/moby/moby/issues/36652

The [docker.ico](https://github.com/docker/cli/blob/8a6ad48d2b2067f3495f973194490f052d580c1f/scripts/winresources/docker.ico) and [docker.png](https://github.com/docker/cli/blob/8a6ad48d2b2067f3495f973194490f052d580c1f/scripts/winresources/docker.png) logo's are outdated (see https://www.docker.com/brand-guidelines for the current version of the logo)

We also may want to

- verify if the [docker.exe.manifest](https://github.com/docker/cli/blob/8a6ad48d2b2067f3495f973194490f052d580c1f/scripts/winresources/docker.exe.manifest) needs updating
- add a comment to [cmd/docker/docker_windows.go](https://github.com/docker/cli/blob/b28c4822cc5ff6162d2eae9acd272af2740770ed/cmd/docker/docker_windows.go#L3) to explain the empty file only having an import

ping @FSHA who should be able to assist with the correct logos

ping @salah-khan @jhowardmsft @simonferquel, who probably are more familiar with this 😄

PR #2367, opened 2020-03-02
Update Windows ico

StefanScherer:
fixes https://github.com/docker/cli/issues/956

Signed-off-by: Stefan Scherer <stefan.scherer@docker.com>

**- What I did**

I've replaces the old Docker ico for the Windows exe files with a newer one.

I've also removed the `docker.png` file as I couldn't find any reference where it may be used. Maybe that was only an example of the ico file for people unable to view Windows ico files.

**- How I did it**

Picked it from an internal JIRA task. :-)

**- How to verify it**

Compile Windows `docker.exe` and check it in Windows Explorer and Sysinternals Process Monitor.

Updated the Windows icon.

systray-icon-win-light-frame0

See also https://github.com/moby/moby/pull/40609


thaJeztah:
The `.png` looks to be still the old logo? 😅

thaJeztah:
oh, never mind; the png is being deleted here, correct?

StefanScherer:
Yes, it's deleted in that commit. GitHub somehow doesn't show that better for such removed binary files.

StefanScherer:
The relevant change is in the `.ico` file which is used here https://github.com/docker/cli/blob/master/scripts/winresources/common.rc
The `.png` seems unused.

thaJeztah:
Ah, interesting. Actually recall looking at that at some point, but wasn't sure if the `.png` would implicitly be used somewhere

PR #2392, opened 2020-03-17
docs: update code-hints for compatibility with "rouge"

thaJeztah:
Unlike GitHub's web-UI, the "rouge" hightlighter used in our online documentation is case sensitive. As a result, code-blocks having the Dockerfile (uppercase) code-hint were not highlighted.

This changes those to use lowercase, which is supported by both.


thaJeztah:
ping @silvin-lubecki PTAL

PR #2402, opened 2020-03-20
Update deprecation policy to 1 stable release

thaJeztah:
The deprecation policy was written when we had monthly releases
(edge+stable), and stable releases were released every 3 months.

The release cadence has changed to be longer than 3 months for
stable releases, so adjusting the policy accordingly.

Note that the policy continues to be "at least x releases", so
even though we _can_ remove a feature after one release, it does
not mean we _must_ remove it: announcing deprecations early helps
users migrate away from features we do not intent to support in
future.


thaJeztah:
ping @nebuk89 @chris-crone @silvin-lubecki PTAL

PR #2405, opened 2020-03-24
Add third-party logging plugins to plugins list

ad-m:
**- What I did**
Added list of logging plugins
**- How I did it**
I added new section following of existing structure.

**- How to verify it**
Read it.

No apply.

obraz


thaJeztah:
Thanks for contributing, but this page lists the _legacy_ plugins (not plugins that can be installed through `docker plugin install`) 

Managed plugins (plugins that are installed with `docker plugin install`) can be found through Docker Hub (e.g. https://hub.docker.com/search?q=&type=plugin&category=logging), so we don't want to list those here to prevent the list from getting out of sync with what's available.

I'll close this PR because of the above, but thanks again; and feel free to continue the conversation

ad-m:
Can we point to Docker Hub then? 

BTW. Plugin ```H1 Docker Journal Plugin``` mentioned on that list is not distributed via Docker Hub and docker plugins support private registries.

PR #2411, opened 2020-03-30
Allow publishing same port as UDP and TCP using docker stack deploy

### LINKED ISSUE #2407: docker stack can't publish udp and tcp for the same port
**Description**

Using `docker stack deploy`, it is not possible to publish the same port as both UDP and TCP.  An example where this is important is DNS.

1. Deploy the following stack:

```
version: '3.7'
services:
  consul:
    image: consul
    command: agent -dev
    ports:
      # Port 8600 is the consul DNS query interface.
      # Mapping this to port 53 allows us to use this as our nameserver.
      - target: 8600
        published: 53
        protocol: tcp
      - target: 8600
        published: 53
        protocol: udp
    networks:
      - consul
networks:
  consul:
```

Only the second port (53/udp) is published.

Both ports are published.

This _may_ only occur if you specify multiple docker-compose yaml files that are merged by `docker stack deploy`.  I haven't tested with just a single compose file.

19.03.8 client and server

**Additional details:**

The problem is in this function: https://github.com/docker/cli/blob/22acbbcc4b3fb5020492b1ed9e72186191bfedc0/cli/compose/loader/merge.go#L114

When merging, ports are placed into a map keyed on the port number.  If the same port is published twice (udp and tcp), the second protocol wins because it overwrites the first.  This map should instead be keyed on `(port, protocol)`.

lneva-fastly:
fixes #2407 

**- What I did**
Simple fix for #2407, to ensure that TCP and UDP publishing for the same port doesn't go away during compose file merging.

**- How I did it**

I changed the key of the map from the port number to the full port specification (including published port number and protocol).

This may not be the ideal solution since it includes not just the protocol, but also the published port number in the map key.  I did it this way because it's all I can figure out how do do with my meager golang skills.  I also have not included a unit test for the same reason.

**- How to verify it**

Try the example in #2407.

* fix bug: now possible to publish same port as TCP and UDP using docker stack deploy

```
   |\---/|
   | ,_, |
    \_`_/-..----.
 ___/ `   ' ,""+ \  sk
(__...'   __\    |`.___.';
  (_,...'(_,.`__)/'.....+
```
(source: https://www.asciiart.eu/animals/cats)

### DIFF
--- a/cli/compose/loader/merge.go
+++ b/cli/compose/loader/merge.go
@@ -111,7 +111,7 @@ func toServicePortConfigsMap(s interface{}) (map[interface{}]interface{}, error)
 	}
 	m := map[interface{}]interface{}{}
 	for _, p := range ports {
-		m[p.Published] = p
+		m[p] = p
 	}
 	return m, nil
 }


thaJeztah:
Thanks! Looks like some tests are failing; would be good to have a new test case added for this as well

devdreamsolution (approved):

adeniyistephen:
> Thanks! Looks like some tests are failing; would be good to have a new test case added for this as well

@thaJeztah Hi, Should I work on this PR?

thaJeztah:
Yes, feel free to have a go to look if it can be fixed

adeniyistephen:
> Yes, feel free to have a go to look if it can be fixed

@thaJeztah  Yes I worked on it, but i think the testing might not be really possible check #2858
 [https://github.com/docker/cli/pull/2858](url)

toby-griffiths:
Anyone know what is happening with this?  We've got this issue with our API Platform stack, and would rather keep the default port numbers, ideally.

What's needed to get this merged?  Seems like it shouldn't be much for such a small change, but I'm unfamiliar with Go projects?

Issue #2420, opened 2020-04-06
Docker Desktop on Windows: 'docker context inspect' is very slow


karolz-ms:
**Steps to reproduce the issue:**
1. Install Docker Desktop on Windows (tested on Windows 10 Enterpirse version 1909 build 18363.720, Docker Desktop 2.2.0.4)
2. Start Docker Desktop.
3. Open command prompt and do `docker context inspect`

**Describe the results you received:**
It consistently takes more than 1 second to run this command, often more than 2 seconds.

**Describe the results you expected:**
Should take 100 ms or less. The default context, pointing to Docker Desktop engine, is the only context on the machine and it should be really straightforward to "inspect" it.

**Additional information you deem important (e.g. issue happens only occasionally):**

I took a trace (available upon request) and noticed that the majority of time is spent waiting on  results from [NetUserGetInfo Windows API](https://docs.microsoft.com/en-us/windows/win32/api/lmaccess/nf-lmaccess-netusergetinfo). On my machine this API is called for a default Docker Desktop context inspection, which has no user information. Ultimately this accounts for more than 2 seconds of latency (for that run of the command)

May be related to https://github.com/docker/cli/issues/1739

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Version:           19.03.8
 API version:       1.40
 Go version:        go1.12.17
 Git commit:        afacb8b
 Built:             Wed Mar 11 01:23:10 2020
 OS/Arch:           windows/amd64
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          19.03.8
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.17
  Git commit:       afacb8b
  Built:            Wed Mar 11 01:29:16 2020
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Output of `docker info`:**

```
Client:
 Debug Mode: false
 Plugins:
  app: Docker Application (Docker Inc., v0.8.0)
  buildx: Build with BuildKit (Docker Inc., v0.3.1-tp-docker)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
[...]

PR #2424, opened 2020-04-07
cli: perform feature detection lazily

thaJeztah:
alternative to / closes https://github.com/docker/cli/pull/1747
fixes https://github.com/docker/cli/issues/1739
fixes https://github.com/docker/cli/issues/2420

cli: perform feature detection lazily

- Docker build: check experimental --platform on pre-run instead of during intialization
- Perform feature detection when actually needed, instead of during initializing
- Version negotiation is performed either when making an API request, or when (e.g.) running `docker help` (to hide unsupported features)
- Use a 2 second timeout when 'pinging' the daemon; this should be sufficient for most cases, and when feature detection failed, the daemon will still perform validation (and produce an error if needed)
    - `context.WithTimeout()` doesn't currently work with ssh connections (connhelper), so we're only applying this timeout for tcp:// connections, otherwise keep the old behavior.

Before this change:

    time sh -c 'DOCKER_HOST=tcp://42.42.42.41:4242 docker help &> /dev/null'
    real   0m32.919s
    user   0m0.370s
    sys    0m0.227s

    time sh -c 'DOCKER_HOST=tcp://42.42.42.41:4242 docker context ls &> /dev/null'
    real   0m32.072s
    user   0m0.029s
    sys    0m0.023s

After this change:

    time sh -c 'DOCKER_HOST=tcp://42.42.42.41:4242 docker help &> /dev/null'
    real   0m 2.28s
    user   0m 0.03s
    sys    0m 0.03s

    time sh -c 'DOCKER_HOST=tcp://42.42.42.41:4242 docker context ls &> /dev/null'
    real   0m 0.13s
    user   0m 0.02s
    sys    0m 0.02s

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -113,6 +113,25 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 		},
 	}
 
+	// Wrap the global pre-run to handle non-BuildKit use of the --platform flag.
+	//
+	// We're doing it here so that we're only contacting the daemon when actually
+	// running the command, and not during initialization.
+	// TODO remove this hack once we no longer support the experimental use of --platform
+	rootFn := cmd.Root().PersistentPreRunE
+	cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		if ok, _ := command.BuildKitEnabled(dockerCli.ServerInfo()); !ok {
+			f := cmd.Flag("platform")
+			delete(f.Annotations, "buildkit")
+			f.Annotations["version"] = []string{"1.32"}
+			f.Annotations["experimental"] = nil
+		}
+		if rootFn != nil {
+			return rootFn(cmd, args)
+		}
+		return nil
+	}
+
 	flags := cmd.Flags()
 
 	flags.VarP(&options.tags, "tag", "t", "Name and optionally a tag in the 'name:tag' format")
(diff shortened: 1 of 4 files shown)


thaJeztah:
@gtardif @silvin-lubecki PTAL

cli/command/image/build.go
```diff
@@ -113,6 +113,26 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 		},
 	}
 
+	// Wrap the global pre-run to handle non-NuildKit use of the
```
silvin-lubecki, line 116: s/NuildKit/Buildkit

cmd/docker/docker.go
```diff
@@ -312,63 +312,69 @@ type versionDetails interface {
...
-	clientVersion := details.Client().ClientVersion()
-	osType := details.ServerInfo().OSType
-	hasExperimental := details.ServerInfo().HasExperimental
-	hasExperimentalCLI := details.ClientInfo().HasExperimental
-	hasBuildKit, err := command.BuildKitEnabled(details.ServerInfo())
-	if err != nil {
-		return err
-	}
+	var (
+		hasClientVersion   = func(v string) bool { return versions.LessThan(details.Client().ClientVersion(), v) }
```
silvin-lubecki, line 343: Shouldn't we use `LessThanOrEqualTo` instead?

thaJeztah: hm, yes it should be reversed here; we need to hide if lower than the version in the annotation.

Let me also rename these versions and reverse the logic because we're now doing `if !hasFeature(v) {`, which is slightly confusing

thaJeztah:
@silvin-lubecki updated; PTAL

PR #2426, opened 2020-04-09
test: make sure environment vars are reset after tests

thaJeztah:
The trust tests were not resetting the environment after they ran, which could result in tests following those tests to fail.

While at it, I also updated some other tests to use gotest.tools


thaJeztah:
@silvin-lubecki PTAL

PR #2432, opened 2020-04-10
improve "network prune" output to mention custom networks only

thaJeztah:
The `docker network prune` command removes unused custom networks,
but built-in networks won't be removed. This patch updates the
message to mention that it's only removing custom networks.

addresses https://github.com/moby/moby/issues/30364#issuecomment-611994955


thaJeztah:
@silvin-lubecki @bkline PTAL

PR #2433, opened 2020-04-10
Jenkinsfile: split e2e tests to separate stages

thaJeztah:
The non-experimental, experimental, and ssh-connhelper stages are running the same tests, which makes it difficult to find _which_ variant of the test failed if they all run in the same stage.

With this change:

Screenshot 2020-04-10 at 16 23 07

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,11 +3,23 @@ wrappedNode(label: 'linux && x86_64', cleanWorkspace: true) {
     stage "Git Checkout"
     checkout scm
 
-    stage "Run end-to-end test suite"
+    stage "Docker info"
     sh "docker version"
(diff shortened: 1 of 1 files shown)


tiborvass:
I'd rather avoid jenkinsifying too much. Can't we add it to the test name instead?

thaJeztah:
They're the same tests, only with different environment variables set; see it as similar to running tests with `CGO_ENABLED=1` and `CGO_ENABLED=0`. If we migrate to github actions, they will probably end up in a `matrix`, which is similar to this

thaJeztah:
looks like this was rolled up into https://github.com/docker/cli/pull/2538 for 19.03, so removing the "cherry-pick" label

PR #2443, opened 2020-04-15
Add "host-gateway" to tests for extra_hosts / --add-host

thaJeztah:
follow-up to https://github.com/docker/cli/pull/2337

67ebcd6dcf82c00f133b886a9a343b67124e58ec added an exception for the "host-gateway" magic value to the validation rules, but didn't add thise value to any of the tests.

This patch adds the magic value to tests, to verify the validation is skipped for this magic value.

Note that validation on the client side is "optional" and mostly done to provide a more user friendly error message for regular values (IP-addresses).


thaJeztah:
ping @silvin-lubecki @arkodg PTAL

PR #2472, opened 2020-04-28
vendor: update kubernetes and dependencies v1.16.9

thaJeztah:
just updating to the latest patch release for v1.16.x; no significant changes

### DIFF
--- a/vendor.conf
+++ b/vendor.conf
@@ -86,15 +86,15 @@ google.golang.org/grpc                              f495f5b15ae7ccda3b38c53a1bfc
 gopkg.in/inf.v0                                     d2d2541c53f18d2a059457998ce2876cc8e67cbf # v0.9.1
 gopkg.in/yaml.v2                                    53403b58ad1b561927d19068c655246f2db79d48 # v2.2.8
 gotest.tools/v3                                     bb0d8a963040ea5048dcef1a14d8f8b58a33d4b3 # v3.0.2
-k8s.io/api                                          35e20aa79eb876d1014e0383c7fcda49e52c5d76 # kubernetes-1.16.1
-k8s.io/apimachinery                                 27d36303b6556f377b4f34e64705fa9024a12b0c # kubernetes-1.16.1
-k8s.io/apiextensions-apiserver                      49e3d608220c016ce72a3bd9524eed4dcef587df # kubernetes-1.16.1
-k8s.io/apiserver                                    3c8b233e046cffa0f8b63fff9e3ba4014a1ef456 # kubernetes-1.16.1
-k8s.io/client-go                                    f68efa97b39e67ee47d4b4683d0037d721f077be # kubernetes-1.16.1
-k8s.io/component-base                               f573d376509cec44c50b8d79741bc939ec7ede56 # kubernetes-1.16.1
-k8s.io/klog                                         3ca30a56d8a775276f9cdae009ba326fdc05af7f # v0.4.0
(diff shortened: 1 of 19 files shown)


vendor.conf
```diff
@@ -86,15 +86,15 @@ google.golang.org/grpc                              f495f5b15ae7ccda3b38c53a1bfc
...
-k8s.io/component-base                               f573d376509cec44c50b8d79741bc939ec7ede56 # kubernetes-1.16.1
-k8s.io/klog                                         3ca30a56d8a775276f9cdae009ba326fdc05af7f # v0.4.0
+k8s.io/api                                          d49a3f108dab8e8d25f22c497fb48678b124efd2 # v0.16.9
+k8s.io/apimachinery                                 f336d9be0221d10a93a7d6d2ec25f6fc799f4fc6 # v0.16.9
+k8s.io/apiextensions-apiserver                      e0288c14e6fda28eefbd584da8c657447e414f96 # v0.16.9
+k8s.io/apiserver                                    3987e0d46aec47186e5f2c0d61ad84975cb79bd8 # v0.16.9
+k8s.io/client-go                                    002560d5bf54049bf5b5ae99231cb2b591f15954 # v0.16.9
+k8s.io/component-base                               dbd90d7c5d87d80fd674cc40d69ca7b14798eda5 # v0.16.9
+k8s.io/klog                                         2ca9ad30301bf30a8a6e0fa2110db6b8df699a91 # v1.0.0
 k8s.io/kube-openapi                                 0270cf2f1c1d995d34b36019a6f65d58e6e33ad4
```
silvin-lubecki, line 96: Should we bump this one too? This commit has been made on September 18 2019.

vendor.conf
```diff
@@ -86,15 +86,15 @@ google.golang.org/grpc                              f495f5b15ae7ccda3b38c53a1bfc
...
-k8s.io/kubernetes                                   d647ddbd755faf07169599a625faf302ffc34458 # v1.16.1
+k8s.io/kubernetes                                   a17149e1a189050796ced469dbd78d380f2ed5ef # v1.16.9
 k8s.io/utils                                        69764acb6e8e900b7c05296c5d3c9c19545475f9
```
silvin-lubecki, line 98: What about this one too?
https://github.com/kubernetes/utils/compare/master..69764acb6e8e900b7c05296c5d3c9c19545475f9

vendor.conf
silvin-lubecki: https://github.com/kubernetes/kube-openapi/compare/master..0270cf2f1c1d995d34b36019a6f65d58e6e33ad4

vendor.conf
thaJeztah: I looked at the `go.mod` in the kubernetes dependencies, and those use commit `581e00157fb1`, and it looked like we were already ahead of that one with 15 commits; https://github.com/kubernetes/utils/compare/581e00157fb1...69764acb6e8e900b7c05296c5d3c9c19545475f9

Updating to master would bring in 379 commits; https://github.com/kubernetes/utils/compare/69764acb6e8e900b7c05296c5d3c9c19545475f9...master

But I admit that I didn't look how many of those would affect our vendored code; WDYT?

thaJeztah:
rebased

vendor.conf
silvin-lubecki: Ok it may not be a good idea then to bump to master as we are still targeting the `1.16.X` version. Fine then 👍

PR #2485, opened 2020-05-01
zsh completion: complete capabilities after --cap-{add,drop}

pseyfert:
**- What I did**

Add a new zsh function to the completion file `contrib/completion/zsh/_docker`.

The bulk of the function is copy and paste from [the documentation of
`--cap-add`](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities) (**Subject to review**: I barely read through them, assuming the docs on the web are a good initial version. I did not investigate if the list of capabilities can be obtained in an automated manner.).

The function is called when tab completing `docker run --cap-add=` (or `drop`).

The suggested completion should look like that (depending on a user's configuration of zsh's tab completion):
```
>> docker run --cap-add=
special capability value:
ALL  -- enable/disable all capabilities
off by default capabilities:
AUDIT_CONTROL    -- Enable and disable kernel auditing; change auditing filter rules; retrieve auditing st
BLOCK_SUSPEND    -- Employ features that can block system suspend.
DAC_READ_SEARCH  -- Bypass file read permission checks and directory read and execute permission checks.
IPC_LOCK         -- Lock memory (mlock(2), mlockall(2), mmap(2), shmctl(2)).
...
```

The function checks:

 - if `--cap-add` or `--cap-drop` is completed. In the former case, off-by-default capabilities are suggested. In the latter case on-by-default capabilities are suggested (I took that idea from the bash completion).

 - if `--cap-add=ALL` or `--cap-drop=ALL` are present on the current command line. If `--cap-add=ALL` is present in the command line, then `--cap-drop=` will also suggest off-by-default capabilities (and vice versa). This should look something like that:

```
special capability value:
ALL  -- enable/disable all capabilities
on by default capabilities:
AUDIT_WRITE       -- Write records to kernel auditing log.
CHOWN             -- Make arbitrary changes to file UIDs and GIDs (see chown(2)).
...
off by default capabilities:
AUDIT_CONTROL    -- Enable and disable kernel auditing; change auditing filter rules; retrieve auditing st
BLOCK_SUSPEND    -- Employ features that can block system suspend.
DAC_READ_SEARCH  -- Bypass file read permission checks and directory read and execute permission checks.
...
```

The function does *not* check any other already-present settings of `--cap-add`. I.e. `docker run --cap-add=BLOCK_SUSPEND --cap-add=<TAB>` will still suggest `BLOCK_SUSPEND` as option.
(**Subject for discussion**. My impression is it's not worth increasing the complexity of the completion function for such a corner case.)

**- How I did it**

Sticking to zsh's builtin features. I use the `opt_args` associative array to
detect what's already specified in `--cap-add` on the command line. If the
option is given multiple times, `$opt_args[--cap-add]` will be colon separated.
I didn't add any logic to handle that case but rely on `*ALL*` matching exactly
the `ALL` value and `ALL` doesn't occur in any other capability.

**- How to verify it**

Typed interactively a few examples

```
docker run --cap-add=<TAB>

docker run --cap-drop=<TAB>

docker run --cap-add=ALL --cap-drop=<TAB>

docker run --cap-add=DAC_READ_SEARCH --cap-add=ALL --cap-add=MAC_OVERRIDE --cap-drop=<TAB>
```

Add suggestions for `docker run --cap-{add,drop}=<TAB>` in the zsh completion

DSC_7596


contrib/completion/zsh/_docker
```diff
@@ -217,6 +217,64 @@ __docker_search() {
     _wanted dockersearch expl 'available images' compadd -a result
 }
 
+__docker_run_all_capabilities() {
+    # list and description taken from https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities
+    local -a default_off_capabilities default_on_capabilities all_capabilities
+    default_off_capabilities=(
+        "SYS_MODULE:Load and unload kernel modules."
```
thaJeztah, line 224: Could you

- sort these lists alphabetically, 
- also add `ALL` to the list? 
- double check if this list matches the Bash completion; https://github.com/docker/cli/blob/master/contrib/completion/bash/docker#L834-L882
-  Perhaps change the names of `default_off_capabilities` and `default_on_capabilities` to `capabilities_droppable` / `capabilities_droppable` / `capabilities_addable` (to match the bash completion)?

pseyfert: > sort alphabetically
sure

> add `ALL` to the list
not sure i understand the suggestion. in [line 264 and 265](https://github.com/docker/cli/pull/2485/files#diff-b100a71a8a5e8e4eb980e60204bb5f6eR264) `ALL` gets added to the suggestions in any case, so I handle this special capability separate instead of adding it to both lists. Did you overlook that or do you suggest dropping the special treatment and adding `ALL` to the default_off and default_on lists?

> double check against bash
will do

> change the names
I like your suggestion

pseyfert: ✓ sorted alphabetically
✓ compared against bash: the only difference I spotted is that my alphabetic sorting put SYSLOG into a different position
✓ renamed the arrays (and rephrased the labels)
→ waiting for clarification on `ALL`. I thought back why I treated `ALL` separately in the first draft. I had decided in favor of that when I was considering to obtain the list of capabilities in an automated way rather than hardcoding them. Since I didn't go for that, the reason for three instead of two arrays is gone. Having a "suggest for drop" "suggest for add" and "suggest always" list kind of eases things up when completing `docker run --cap-add=ALL --cap-drop=` (because then `ALL` doesn't appear twice in the suggestions (in the "droppable (on by default)" and the "droppable (enabled through ALL)") but if you prefer two lists for other reasons that's fine with me

pseyfert:
ping?

devdreamsolution (approved):

pseyfert:
rebased

pseyfert:
What's your (docker's) status on this one? From my side it's good to merge. Just rebased onto current master.

pseyfert:
ping?
 - any more review comments?
 - happy with my responses to the previous round of comments?
  - happy to merge as-is or should we close this pr and start over from the current master branch?

PR #2487, opened 2020-05-03
Implement IPNetSliceValue  locally, and un-fork spf13/pflag

thaJeztah:
This flag type was not yet merged upstream (https://github.com/spf13/pflag/pull/170), so instead of using a fork of spf13/pflag, define the type locally, so that we can vendor the upstream package again.


thaJeztah:
rebased (conflict in vendor.conf because https://github.com/docker/cli/pull/2490 was merged)

@silvin-lubecki PTAL

PR #2489, opened 2020-05-03
vendor: update cpuguy83/go-md2man v2.0.0

thaJeztah:
full diff: https://github.com/cpuguy83/go-md2man/compare/v1.0.8...v2.0.0


thaJeztah:
Ah, dang; can't really do this separately, because the old Cobra expects the old go-md2man. I could vendor them both, but I'll just squash them

Issue #1631, opened 2019-01-23
Docker manifest push ignoring --insecure


stephencsnow:
**Description**

When attempting to push a manifest to an insecure registry, the command fails, indicating an issue with the registry server giving an HTTP response to a HTTPS client. The documentation states that passing the `--insecure` flag into the push command should fix this, but it does nothing.

**Steps to reproduce the issue:**
1. Create a manifest via `docker manifest create --insecure 192.168.0.20:5000/my-img:latest 192.168.0.20:5000/my-img-arm:latest 192.168.0.20:5000/my-img-x86:latest`
2. Annotate the manifest via `docker manifest annotate 192.168.0.20:5000/my-img:latest 192.168.0.20:5000/my-img-arm:latest --arch arm`
3. Attempt to push manifest via `docker -D manifest push --insecure 192.168.0.20:5000/my-img:latest`

**Describe the results you received:**

```
DEBU[0000] hostDir: /etc/docker/certs.d/192.168.0.20:5000 
DEBU[0000] hostDir: /etc/docker/certs.d/192.168.0.20:5000 
DEBU[0000] hostDir: /etc/docker/certs.d/192.168.0.20:5000 
DEBU[0000] hostDir: /etc/docker/certs.d/192.168.0.20:5000 
DEBU[0000] hostDir: /etc/docker/certs.d/192.168.0.20:5000 
DEBU[0000] hostDir: /etc/docker/certs.d/192.168.0.20:5000 
DEBU[0000] hostDir: /etc/docker/certs.d/192.168.0.20:5000 
DEBU[0000] hostDir: /etc/docker/certs.d/192.168.0.20:5000 
failed to configure transport: error pinging v2 registry: Get https://192.168.0.20:5000/v2/: http: server gave HTTP response to HTTPS client
```
This is the same exact result as when running without `--insecure`.

**Describe the results you expected:**

The manifest push succeeds.

**Additional information you deem important (e.g. issue happens only occasionally):**

I was having issues with `manifest create` as well until I updated my docker from 18.0.6 to 18.0.9. Docker pull/push with this insecure registry works fine.

**Output of `docker version`:**

```
Client:
 Version:           18.09.1
 API version:       1.39
 Go version:        go1.10.6
 Git commit:        4c52b90
 Built:             Wed Jan  9 19:35:31 2019
 OS/Arch:           linux/amd64
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          18.09.1
[...]

PR #2376, opened 2020-03-04
try http for docker manifest --insecure

davidwtf:
**- What I did**
fixes #1631

**- How I did it**
Add retry for http when https not work

**- How to verify it**
`docker manifest push  --insecure 127.0.0.1:5000/test:v1`

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #2494, opened 2020-05-05
the better way to fix: focker manifest push ignoring --insecure

earthdiaosi:
**- What I did**
the better way to fixes https://github.com/docker/cli/issues/1631

**- How I did it**

Take advantage of existing code to fix the bug with minimal changes。
- add parameter `insecure` for `func getDefaultEndpointFromRepoInfo()`.
-  assign a value for `options.InsecureRegistries` when `insecure` is `true`.

the method reference to https://github.com/docker/cli/commit/d57adbc034a69d14b1c88c63cdf07b58fb8eb253

**- How to verify it**
docker --debug manifest push --insecure 127.0.0.1:5000/test:v1

compared with https://github.com/docker/cli/pull/2376, this method has the following advantages：
1. the logic of the code is clearer;
2. more in line with the original author's thinking


earthdiaosi:
hi @thaJeztah，PTAL

thaJeztah:
I'm not near my computer right now, but should the previous changes be reverted if this replaces the approach?

earthdiaosi:
the previous changes should be reverted.

earthdiaosi:
The current method pushes the manifest successfully regardless of whether or not the previous changes are rolled back.

earthdiaosi:
@thaJeztah Is there anything I can do about the operation of rolling back the previous code？

thaJeztah:
@earthdiaosi you can do a `git revert -s 83478fe7cb4334619754236c5e35e228498e5a12` to revert the previous change, and push that as part of this PR

cli/registry/client/endpoint.go
```diff
@@ -12,6 +12,8 @@ import (
 	authtypes "github.com/docker/docker/api/types"
 	"github.com/docker/docker/registry"
 	"github.com/pkg/errors"
+
+	"github.com/sirupsen/logrus"
```
thaJeztah, line 16: Please group this import with the other ones above

cli/registry/client/endpoint.go
```diff
@@ -63,7 +69,7 @@ func getDefaultEndpointFromRepoInfo(repoInfo *registry.RepositoryInfo) (registry
 	}
 	// Default to the highest priority endpoint to return
 	endpoint := endpoints[0]
-	if !repoInfo.Index.Secure {
+	if insecure {
 		for _, ep := range endpoints {
 			if ep.URL.Scheme == "http" {
```
thaJeztah, line 74: I'm actually wondering if this original implementation is correct; looking at this code, it seems to _always_ use `http`, instead of _trying_ https and _falling back_ to `http`.

`--insecure` / `--insecure-registry` _should_ mean;

- try https
- if the certificate is self-signed, allow
- if https failed, fall back to http

thaJeztah: Oh, maybe it does, because `registryService.LookupPushEndpoints` seems to check if it can find certificates for the host, but please double check if it's doing the right thing

earthdiaosi:
ok. I will double check it tomorrow.

cli/registry/client/endpoint.go
earthdiaosi: 1. when `if !repoInfo.Index.Secure {`, it always use `https`, because `!repoInfo.Index.Secure` is always `false`;
2. when `if insecure {`, it always use `http`, 
3. the result of `registryService.LookupPushEndpoints` is `endpoints: [{false https://myregistry-other.domain.com v2 false false true 0xc0004c5800} {false http://myregistry-other.domain.com v2 false false true 0xc0004c5800}]`

earthdiaosi:
@thaJeztah the best way to fix this bug is reuse [`func (c *client) iterateEndpoints(...){...}`](https://github.com/docker/cli/blob/master/cli/registry/client/fetcher.go#L199).
 And the previous change will break the logic of [`iterateEndpoints`](https://github.com/docker/cli/blob/master/cli/registry/client/fetcher.go#L228), so the previous change should be reverted.

earthdiaosi:
@thaJeztah, please check again. I've squashed the commits. Thanks.

earthdiaosi:
hi @thaJeztah  PTAL

neverseek:
docker manifest push --insecure still doesn't work on 19.03.5.

[root@123 harbor]# docker -D manifest push --insecure 168.1.25.123:80/library/busybox
DEBU[0000] hostDir: /etc/docker/certs.d/168.1.25.123:80
DEBU[0000] hostDir: /etc/docker/certs.d/168.1.25.123:80
DEBU[0000] hostDir: /etc/docker/certs.d/168.1.25.123:80
failed to configure transport: error pinging v2 registry: Get https://168.1.25.123:80/v2/: http: server gave HTTP response to HTTPS client

docker push works so configuration in daemon.json is ok.

[root@123 harbor]# docker push 168.1.25.123:80/library/busybox-amd64
The push refers to repository [168.1.25.123:80/library/busybox-amd64]
d2421964bad1: Layer already exists
latest: digest: sha256:c9249fdf56138f0d929e2080ae98ee9cb2946f71498fc1484288e6a935b5e5bc size: 527

devdreamsolution (approved):

SJern:
Hello guys,
I see this has been approved. Any plan to merge it soon?

SJern:
> Hello guys,
> I see this has been approved. Any plan to merge it soon?
> My Docker version is `20.10.5` but #2376 doesn't seem to make any difference for me.
> `failed to configure transport: error pinging v2 registry: Get https://host.docker.internal:9000/v2/: dial tcp: lookup host.docker.internal: no such host`

never mind. My error might not have to do with the `insecure` flag. Everything works after I changed from `host.docker.internal:9000` to `0.0.0.0:9000`
For anyone who also tries to change to `0.0.0.0:9000`, just remember to add that to the `insecure-registries` list in `daemon.json`

PR #2498, opened 2020-05-06
Make healthcheck buffer size configurable

gloomybrain:
Healthcheck result is stored in a byte buffer of 4096 bytes. Making the size of this buffer configurable gives us more possibilities to rely on this feature

Please see the related [pull request to moby](https://github.com/moby/moby/pull/40911)

Signed-off-by: Denis Novozhilov <348558+gloomybrain@users.noreply.github.com>

**- What I did**
I've added a new cli option named `--health-buffer-size` intended to regulate the amount of healthcheck output stored.

**- How I did it**
Changes are pretty much trivial, please see the code changes

**- How to verify it**
This should be verified in conjunction with the corresponding moby PR (see the link above)

cute-animal

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -121,6 +121,7 @@ type containerOptions struct {
 	healthTimeout      time.Duration
 	healthStartPeriod  time.Duration
 	healthRetries      int
+	healthBufferSize   int
 	runtime            string
 	autoRemove         bool
 	init               bool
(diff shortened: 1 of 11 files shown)


cpuguy83:
I don't think we should allow this to be configurable.
First and foremost this is stored in memory in dockerd.
This also changes the image spec.
Health checks as a whole are not OCI compliant, this makes it even more so.
It also seems really sketchy to cram even 4K of output into the health check.

gloomybrain:
> I don't think we should allow this to be configurable.
> First and foremost this is stored in memory in dockerd.
> This also changes the image spec.
> Health checks as a whole are not OCI compliant, this makes it even more so.
> It also seems really sketchy to cram even 4K of output into the health check.

Not that to argue with you, more of elaborating on my POV:
I suppose there is no one buffer size that fits for everybody. Some might want to shrink this buffer, some might need it to be larger. However making this configurable won't change a thing for the widest majority of users because the default value is left intact.
Speaking of OCI compliance: I'm not aware of how much of a priority is it for the moment for docker. If health checks already are outside of the OCI scope anyway then it can't be worse than it is now ;)

cpuguy83:
What I'm saying is that even a 4KB buffer is really *too much* and a horrible design decision on our side for something that is supposed to be a signal to the health of a service, not a log of what's going on with it.

gloomybrain:
@cpuguy83 thanks for your explanation, closing this

PR #2500, opened 2020-05-07
Add "context" to "docker version" and "docker info"

thaJeztah:
This adds the currently selected "docker context" to the output of `docker version` (and `docker info`), which allows users to see which context is selected to produce the version output, and can be used (for example), to set the prompt to the currently selected context:

(in `~/.bashrc`):

```bash
function docker_context_prompt() {
        PS1="context: $(docker version --format='{{.Client.Context}}')> "
}

PROMPT_COMMAND=docker_context_prompt
```

After reloading the `~/.bashrc`, the prompt now shows the currently selected
`docker context`:

```bash
$ source ~/.bashrc
context: default> docker context create --docker host=unix:///var/run/docker.sock my-context
my-context
Successfully created context "my-context"
context: default> docker context use my-context
my-context
Current context is now "my-context"
context: my-context> docker context use default
default
context: default>
```


thaJeztah:
@silvin-lubecki @simonferquel @nebuk89 PTAL

cli/command/system/version.go
```diff
@@ -32,6 +32,7 @@ Client:{{if ne .Platform.Name ""}} {{.Platform.Name}}{{end}}
  Git commit:	{{.GitCommit}}
  Built:	{{.BuildTime}}
  OS/Arch:	{{.Os}}/{{.Arch}}
+ Context:	{{.Context}}
```
silvin-lubecki, line 35: I wonder if we really need this `context` in `version` 🤔 
I think that version should mostly print static informations, but `context` is by nature purely dynamic.
That said, it may help the user to understand which daemon she's talking to, so which daemon version are printed.
I'm a bit over the fence 🤷

thaJeztah: My main reason for putting it here, is that `docker info` requires a daemon connection (and is rather heavy-weight), so if someone wants to use it to set their prompt (as the example), they likely wouldn't want it to be taken from `docker info`.

We _could_ add, e.g., the option to call `docker context use` (without argument) to print the current context, or add a `docker context show-current` subcommand (but not sure if the last one is "great" to have a subcommand _just_ for that); WDYT?

AkihiroSuda:
Can we also print non-context DOCKER_HOST?

thaJeztah:
@AkihiroSuda Actually wanted to do that as well; didn't make the PR yet, but want to do that soon

colegatron:
Any idea when 20.10 is going to be released? I am interested in (any) solution to solve the 'docker context ls' slowness to get the current context. My prompt displays kubernetes, git, and other info blazing fast, but docker context is damn slow for remotes.

PR #2507, opened 2020-05-09
-buildmode=pie is not supported for the mips arch

XiaodongLoong:
reference: https://github.com/containerd/containerd/commit/4c99c81326f4026fb8c0b8c5e10542205d99c321

Signed-off-by: liuxiaodong <liuxiaodong@loongson.cn>

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/build/dynbinary
+++ b/scripts/build/dynbinary
@@ -9,6 +9,16 @@ source ./scripts/build/.variables
 
 echo "Building dynamically linked $TARGET"
 export CGO_ENABLED=1
-go build -o "${TARGET}" -tags pkcs11 --ldflags "${LDFLAGS}" -buildmode=pie "${SOURCE}"
+case "$(go env GOARCH)" in
+        mips*|ppc64)
+                # pie build mode is not supported on mips architectures
+                GO_BUILDMODE=""
+                ;;
+        *)
+                GO_BUILDMODE="-buildmode=pie"
+                ;;
+esac
+
+go build -o "${TARGET}" -tags pkcs11 --ldflags "${LDFLAGS}" ${GO_BUILDMODE} "${SOURCE}"
 
 ln -sf "$(basename "${TARGET}")" build/docker


XiaodongLoong:
@thaJeztah  please review, thanks.

scripts/build/dynbinary
```diff
@@ -9,6 +9,14 @@ source ./scripts/build/.variables
...
-go build -o "${TARGET}" -tags pkcs11 --ldflags "${LDFLAGS}" -buildmode=pie "${SOURCE}"
+GO_BUILDMODE="-buildmode=pie"
+#pie build mode is not supported for the mips arch 
+for arch in mips mipsle mips64 mips64le ppc64
+do
+    if [ "$(go env GOARCH)" = "${arch}" ];then
+        GO_BUILDMODE=""
+    fi
+done
+go build -o "${TARGET}" -tags pkcs11 --ldflags "${LDFLAGS}" "${GO_BUILDMODE}" "${SOURCE}"
```
thaJeztah, line 20: Perhaps use a `case`, so that we don't have to do this loop;

 ```suggestion

case "$(go env GOARCH)" in
    mips*|ppc64)
        # pie build mode is not supported on mips architectures
        GO_BUILDMODE=""
        ;;
    *)
        GO_BUILDMODE="-buildmode=pie"
        ;;
esac

go build -o "${TARGET}" -tags pkcs11 --ldflags "${LDFLAGS}" "${GO_BUILDMODE}" "${SOURCE}"
```

**Comment by thaJeztah** [comment]
I noticed this script uses spaces for indentation, which we should probably fix as well, but let's keep that for a separate PR (there might be more scripts that need to be reformatted)

File: scripts/build/dynbinary (hunk 1 above)
Comment (tianon) [reply scripts/build/dynbinary:20]: Agreed, also `${GO_BUILDMODE}` can't be quoted in this final `go build` line (or it'll pass the empty saying as an argument).

Comment (thaJeztah) [reply scripts/build/dynbinary:20]: Good catch; I was looking at it, but didn't think about the "empty" case ❤️

Comment (thaJeztah) [reply scripts/build/dynbinary:20]: (updated my example/suggestion)

Comment (XiaodongLoong) [reply scripts/build/dynbinary:20]: I  have learnt a lot，thanks！：）

**Comment by thaJeztah** [comment]
Thanks for updating! I have one final nit 😅; could you change the `signed-off-by` line to use your real name (not your GitHub handle)? So instead of

```
Signed-off-by: liuxiaodong <liuxiaodong@loongson.cn>
```

Change it to

```
Signed-off-by: Xiaodong Liu <liuxiaodong@loongson.cn>
```

(assuming that's your correct name)

silvin-lubecki:
@XiaodongLoong Do you think we also need to backport that fix on `19.03` ?

tianon:
I'm not a CLI maintainer (just an interested party here), but IMO this isn't worth backporting to the 19.03 series.  There are several other components/patches involved in Docker for MIPS to build successfully on 19.03, so I think it would be better to just wait for the imminent 20.x series where many of those patches will be already included. :innocent:

(By my own count, the only "hole" left for the things Docker needs is a change similar to https://github.com/containerd/containerd/commit/4c99c81326f4026fb8c0b8c5e10542205d99c321 in `runc`'s `Makefile` -- everything else is already a merged PR. :sweat_smile:)

XiaodongLoong:
@silvin-lubecki   I agree with tianon, because I found this issue on docker-ce version 19.03.4. There are several patches to build successfully for mips64el on version 19.03.4. I have confirmed and pushed these patches one by one, this PR is the last one.

PR #2503, opened 2020-05-07
Add support for --limit-pids on service create / update and stacks (swarm)

thaJeztah:
depends on:

- [x] https://github.com/moby/moby/pull/40937 API: split types for Resources Reservations and Limits
- [x] https://github.com/moby/moby/pull/40938 API: swarm: move PidsLimit to TaskTemplate.Resources
- [x] https://github.com/docker/cli/pull/2512 vendor: pkg/errors v0.9.1

relates to:

- https://github.com/moby/moby/pull/39882 Add API support for PidsLimit on services
- https://github.com/docker/swarmkit/pull/2415 add pid limits
- fixes https://github.com/moby/moby/issues/28618 Security Proposal: add "--pids-limit" to "docker service create/update" option list

docker-compose / stack support was not yet added; I'll try to get that done asap 

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #2206, opened 2019-11-27
Add device cgroup rules

agners:
This (re-)adds `device_cgroup_rules` to the compose schema. Currently it is not supported by the `docker stack deploy` command. This implements what has been suggested in https://github.com/docker/compose/pull/6974

PR #2517, opened 2020-05-10
compose: add build.extra_hosts to v3.9 schema

thaJeztah:
This is not currently used by the CLI, but can be used by docker compose to bring parity on this feature with the compose v2.4 schema.

relates to https://github.com/docker/compose/issues/7323


thaJeztah:
@silvin-lubecki @ndeloof PTAL

ndeloof:
Compose canonical schema should be https://github.com/compose-spec/compose-spec/blob/master/schema/config_schema_v3.9.json and such change should only take place after a public proposal to the compose-spec community

thaJeztah:
@ndeloof what I understood is that the current 3.9 spec was used as the transition to make compose-spec the canonical place because it was already worked on for the upcoming 20.03 release; if compose-spec is the canonical place, should this repository vendor the spec from that location? (also this means we have a release blocker for 20.03, pending https://github.com/docker/cli/pull/2503

agners:
@thaJeztah @ndeloof this is utterly frustrating for outside contributors. I created a merge request months ago adding `device_cgroup_rules` (see https://github.com/docker/cli/pull/2206). However it got ignored/sidelined. With the compose-spec comming along, I did all my home work and helped making it part of the upcoming spec https://github.com/compose-spec/compose-spec/pull/63. 

And now, within 4 days other properties make it in directly into CLI's compose spec? So now compose-spec essentially forked??

ndeloof:
https://github.com/compose-spec/compose-go is the exact same codebase as docker/cli for compose file loading and mdel (this is a git extract, including history). CLI _should_ be updated to adopt compose-go and throw avoid code to diverge like this. This is more a question of finding the right timing vs CLI release schedule.

agners:
@ndeloof @thaJeztah since we won't adopt compose-go in time for 20.03, could we then not just merge #2206? I would mean less divergence from compose-spec...

ndeloof:
compose-spec is going to re-introduce all v2 attributes in the schema by approval of https://github.com/compose-spec/compose-spec/issues/12
so #2206 still make sense, this won't introduce a divergence.

burka:
For anyone searching like me: It's solved in Version 3.9 https://github.com/docker/cli/pull/2517 - you can add `extra_hosts:` in `build:` now.

PR #2521, opened 2020-05-11
vendor: mitchellh/mapstructure v1.0.0

thaJeztah:
we were only one commit behind v1.0.0, so updating to that version; we can do a follow-up to update to the latest minor release (v1.3.0)

full diff: https://github.com/mitchellh/mapstructure/compare/f15292f7a699fcc1a38a80977f80a046874ba8ac...v1.0.0


thaJeztah:
@silvin-lubecki PTAL

PR #2540, opened 2020-05-21
Pass empty kubeconfig path when loading default Kubernetes configuration

### LINKED ISSUE #256: Support for in cluster kubernetes client
It seems like buildx explicitly uses the `~/.kube/config`. I'd like to run buildx in a CI pipeline that runs in kubernetes. Until #216 is implemented, users are forced to run `docker buildx create` and need the proper permissions to create deployments. To make doing this easier in a pipeline I'd like to be able to use the permissions of the service account attached to the pod I am running the above command from.

jbarrick-mesosphere:
Previously, in-cluster configuration could not be used because the
ExplicitPath was always set in the `ClientConfigLoadingRules`.

If `ExplicitPath` is not set, `ClientConfigLoadingRules` will look up
all of the standard environment variables to find the kubeconfig file,
including supporting in cluster authentication.

We leave the `kubeconfig` parameter on `FromKubeConfig` to support
explicitly setting the path and also to not change the signature.

This will fix docker/buildx#256 once the docker/cli dependency is
updated.

Signed-off-by: jbarrick@mesosphere.com <jbarrick@mesosphere.com>

Added support for in-cluster Kubernetes authentication for Kubernetes contexts.

### DIFF
--- a/cli/command/context/options.go
+++ b/cli/command/context/options.go
@@ -2,8 +2,6 @@ package context
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -13,7 +11,6 @@ import (
 	"github.com/docker/cli/cli/context/kubernetes"
 	"github.com/docker/cli/cli/context/store"
 	"github.com/docker/docker/client"
-	"github.com/docker/docker/pkg/homedir"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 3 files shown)


simonferquel (approved):
Seems legit to me

silvin-lubecki (approved):
LGTM

silvin-lubecki:
Ping @mat007 do we want this PR for the next 20.0X release? If so we need to be sure it won't break desktop.

AkihiroSuda:
PTAL @tonistiigi @thaJeztah 

Seems needed for https://github.com/docker/buildx/issues/256

hc0503 (approved):

thaJeztah:
closing because this code was removed in https://github.com/docker/cli/pull/3139

PR #2546, opened 2020-05-25
Some minor refactoring and reformatting 

thaJeztah:
### clean-up "resolve image" option logic

- change `validateResolveImageFlag()` to only perform _validation_,  and not combine it with modifying the option.
- use a `switch` instead of `if` in `validateResolveImageFlag()`
- `deployServices()`: break up some `switch` cases to make them easier to read/understand the logic.

### un-wrap some code

The wrapping made the code harder to read (and in some cases destracted from the actual code flow).

Some of these functions take too many arguments; instead of hiding that, it probably better to make it apparent that something needs to be done (and fix it :-)).

### DIFF
--- a/cli/command/stack/swarm/deploy.go
+++ b/cli/command/stack/swarm/deploy.go
@@ -25,25 +25,26 @@ const (
 func RunDeploy(dockerCli command.Cli, opts options.Deploy, cfg *composetypes.Config) error {
 	ctx := context.Background()
 
-	if err := validateResolveImageFlag(dockerCli, &opts); err != nil {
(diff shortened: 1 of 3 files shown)


thaJeztah:
boo!

```
cli/command/stack/swarm/deploy_composefile.go:178:1: cyclomatic complexity 17 of func `deployServices` is high (> 16) (gocyclo)
func deployServices(ctx context.Context, dockerCli command.Cli, services map[string]swarm.ServiceSpec, namespace convert.Namespace, sendAuth bool, resolveImage string) error {
^

Exited with code exit status 1
```

PR #2566, opened 2020-06-07
Remove duplicate container fish completion

soraxas:
**- What I did**
Fix duplicate suggestion in fish completion file.

**- How I did it**
Changed `\n` to `\t` in container list's format

**- How to verify it**
Background info: fish completion will take a text in the format of
```
foo1\tbar1\nfoo2\tbar2\n...
```
format and interpret it as having two suggestion, 1st with the name **foo1** with description **bar1**, and so on. (i.e. `\t` denotes description, and `\n` denotes new suggestion.

This is a test of having only *two* containers running.

Current behaviour:
ab

With the patch:
ac

Use container's name as description in fish completion suggestion.

53e719f9b6b9e19c5c8ab6a5f78d803f


thaJeztah:
What if I (in your example) try to use `docker stop interest<tab>`, would it complete to the _ID_ of the container, instead of the name?

soraxas:
Now that I think of it, probably using the friendly name as the completion makes more sense (and hash as description, as I was afraid of changing too much previously.

To answer your question, yes it will still complete as it considers the description as well (in fact it consider any substring afaik)

docker-show

There were 4 completions in this gif.
1. Using 2nd container's name
2. Using 2nd container's hash (beginning)
3. Using 1st container's hash (beginning)
4. Using 2nd container's name (some substring in middle)

If you agree, I will update the changes (i.e. swap `{{.ID}}` and `{{.NAME}}`

thaJeztah:
Sorry for the delay; I'm still a bit on the fence, as the completion will _replace_ something I typed, instead of _completing_ it 🤔 

/cc @silvin-lubecki any thoughts?

devdreamsolution (approved):

krobelus:
I have thought about doing the same, however, I think descriptions should be used to convey additional information about an image/container, which #1623 does quite nicely. Therefore both IDs and names should remain as completion options - they don't hurt anyone.

Currently the container IDs and names are ordered alphabetically, because that's what fish does by default.
This doesn't make a ton of sense. Docker already provides a sensible order.
My suggestion would be to use the [`--keep-order` option](https://fishshell.com/docs/current/cmds/complete.html)
to prevent fish from sorting them, and use a small snippet in `__fish_print_docker_containers` to first print all human-friendly names, and then all IDs.

```fish
set -l all_containers (docker ps -a --no-trunc --format "{{.ID}}\n{{.Names}}")
set -l ids (string match -r '^\w{64}$' $all_containers)
set -l names (string match -rv '^\w{64}$' $all_containers)
string join \n $names $ids
```

adding `--keep-order` would look like

```diff
-complete -c docker -A -f -n '__fish_seen_subcommand_from attach' -a '(__fish_print_docker_containers running)' -d 
+complete --keep-order -c docker -A -f -n '__fish_seen_subcommand_from attach' -a '(__fish_print_docker_containers running)' -d "Container"
```

PR #2569, opened 2020-06-09
Use consistent name for errors

thaJeztah:
This prevents inconsistent errors when using a symlink, or when renaming
the binary;

Before this change;

    ln -s $(which docker) toto
    ./toto rune
    docker: 'rune' is not a docker command.

    ./toto run daslkjadslkjdaslkj
    Unable to find image 'adslkjadslakdsj:latest' locally
    ./toto: Error response from daemon: pull access denied for adslkjadslakdsj, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.

After this change:

    ln -s $(which docker) toto
    ./toto rune
    docker: 'rune' is not a docker command.

    ./toto run daslkjadslkjdaslkj
    Unable to find image 'adslkjadslakdsj:latest' locally
    docker: Error response from daemon: pull access denied for adslkjadslakdsj, repository does not exist or may require 'docker login': denied: requested access to the resource is den>

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os"
 	"runtime"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @gtardif @rumpl @silvin-lubecki PTAL

Issue #1637, opened 2019-01-24
Improve / harden handling of CLI configuration files


thaJeztah:
I noticed this "WARNING" line when reading the code; https://github.com/docker/cli/blob/48bd4c6deb59784b78b860bca14b9995acef0cb3/cli/config/config.go#L117-L126

From that code, I think we're handling configuration files incorrectly, so I did some testing, and wrote down what I propose to be the correct behavior for these situations

Prepare the test;

```bash
mkdir config-example && cd config-example
mkdir ./non-existing-config

mkdir -p ./not-a-file-config/config.json

touch ./not-a-directory-config

mkdir ./invalid-config && echo "Invalid JSON" > ./invalid-config/config.json
mkdir ./empty-config && touch ./empty-config/config.json
mkdir ./whitespace-config && echo '' > ./whitespace-config/config.json

mkdir ./no-access-config && echo "{}" > ./no-access-config/config.json
chmod 600 ./no-access-config/config.json
sudo chown 0:0 ./no-access-config/config.json
```

### 1. missing configuration file

Specifying a non-existing `config.json` silently ignores the missing file;

```bash
docker --config=./non-existing-config version

Client: Docker Engine - Community
 Version:           18.09.1
 API version:       1.39
...
```

#### Proposed behavior

We should

- fail if a user explicitly specifies a configuration file to use, and the file is missing.
- if no configuration file is specified; ignore the missing file (as having a configuration file is optional)

### 2. Invalid configuration file

Using an invalid `config.json` will print a warning, but continue anyway;

```bash
docker --config=./invalid-config version

WARNING: Error loading config file: invalid-config/config.json: invalid character 'I' looking for beginning of value
Client: Docker Engine - Community
 Version:           18.09.1
 API version:       1.39
...
```

Reporting a "successful" exit-code;

```bash
echo $?
0
```

Same for non-accessible, non-file configurations;

```bash
docker --config=./not-a-file-config/ version

WARNING: Error loading config file: not-a-file-config/config.json: read not-a-file-config/config.json: is a directory
Client: Docker Engine - Community

docker --config=./no-access-config/config.json

WARNING: Error loading config file: no-access-config/config.json: open no-access-config/config.json: permission denied

docker --config=./not-a-directory-config version
WARNING: Error loading config file: not-a-directory-config/config.json: stat not-a-directory-config/config.json: not a directory
```

#### Proposed behavior

[...]

PR #2576, opened 2020-06-12
Defer error to first API call when config file references a missing context

simonferquel:
**- What I did**

~~Previously, when the "CurrentContext" field of the config file referenced a context that did not exist in the store, the CLI would fail (even for simple things like docker context ls).~~
~~This causes an issue with Docker Desktop "Contexts Syncing" feature (that uses the same store between win32 and WSL2):~~
~~If win32 cli removes the "current context" of the WSL 2 CLI (which it can do if it is not the "current context" in the win32 config file), it would make the WSL 2 CLI unusable, asking the user to modify the config file by hand.~~

When a user removes a context from the store without using the CLI (by removing the context directory directly in the store), it can trigger an issue where the CLI does not work at all (even for changing the current context) anymore without editing the config file by hand.

With Docker Desktop and WSL 2, we now link the wsl 2 distros context store to the windows one, so that a context created from Windows can be used in WSL 2 and vice-versa. However we keep the CLI config files independant (at least for now), so this issue can be triggered quite often.

This commit fixes that by changing the way we handle context missing error: if the resolved context name comes from the config file and refers a missing context, the error is reported when the API is actually called instead of at the CLI initialization. Thus, local-only commands continue to work (shuch as `context use`, `context ls`,...).

~~This commit relaxes this behavior a bit: explicitly referencing a missing context (using the --context flag or DOCKER_CONTEXT env var) will still produce an error, but implicitly referencing such a missing context (via the CurrentContext config field) will fallback to use the "default" context.~~

**- How I did it**

~~In the function resolving the current context name, instead of erroring with asking the user to modify its config file when the config file references a missing context, fallback to default~~

The function resolving the current context name does not error when the config file references a missing context. It returns a hint about the source of context name.

The check for a missing context is done in the CLI init code. If we find out that the context is missing and that the source of the selected context name is the config file, instead of failing, we initialize the api client with an always failing client (that reports the error during dialing). This allows to change the current context easily, without having to edit the config file by hand.

**- How to verify it**

~~Edit the config file, set CurrentConfig to a value that is not an existing context name, check that the CLI fallbacks to using "default".~~

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"io/ioutil"
+	"net"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -36,6 +37,16 @@ import (
 	"github.com/theupdateframework/notary"
 	notaryclient "github.com/theupdateframework/notary/client"
 	"github.com/theupdateframework/notary/passphrase"
+)
+
+type contextNameSource int
+
+const (
+	contextFallbackToDockerHost contextNameSource = iota
+	contextFromFlag
+	contextFromEnv
+	contextFromConfigFile
+	contextDefault
 )
 
 // Streams is an interface which exposes the standard input and output streams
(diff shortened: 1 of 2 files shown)


cli/command/cli.go
```diff
@@ -549,7 +549,9 @@ func resolveContextName(opts *cliflags.CommonOptions, config *configfile.ConfigF
 	if config != nil && config.CurrentContext != "" {
 		_, err := contextstore.GetMetadata(config.CurrentContext)
 		if store.IsErrContextDoesNotExist(err) {
-			return "", errors.Errorf("Current context %q is not found on the file system, please check your config file at %s", config.CurrentContext, config.Filename)
+			// If config file references a context that does not exist, fallback to default
+			// Explicitly requested contexts (flag or env-var based) will still get the error
+			return DefaultContextName, nil
```
thaJeztah, line 554: I don't think we should silently fall back, as this would lead to unexpected results (similar to outlined in https://github.com/docker/cli/issues/1637)

simonferquel: Would writing a message to stderr feel good enough ?

thaJeztah: No, I don't think a message would be enough, because that wouldn't make the command fail.

Consider a situation where I am in a specific context, and that context is (somehow) removed, and all of a sudden my command completes successfully but against the wrong server. Depending on the command I ran that could mean I just (e.g.)  _removed_ a volume or image from the wrong environment.

simonferquel: Indeed. Would deferring the error to the http call be enough?

thaJeztah: If we can defer to the moment it's actually needed, yes, that'd work (probably also when doing `docker context inspect`, but I guess that would already be the case)

hc0503 (approved):

PR #2578, opened 2020-06-12
Added support for setting OS version in docker manifest annotate.

cpuguy83:
Carry of #1461, just rebased.

### DIFF
--- a/cli/command/manifest/annotate.go
+++ b/cli/command/manifest/annotate.go
@@ -18,6 +18,7 @@ type annotateOptions struct {
 	os         string
 	arch       string
 	osFeatures []string
+	osVersion  string
 }
 
 // NewAnnotateCommand creates a new `docker manifest annotate` command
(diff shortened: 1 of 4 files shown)


tonistiigi:
@cpuguy83 fyi, this use-case is also covered with `buildx imagetools create`

sozercan:
@tonistiigi Is there an example of setting os.version using `buildx imagetools create`? I couldn't find it in https://github.com/docker/buildx#buildx-imagetools-create-options-source-source

tonistiigi:
@sozercan `-f` can point to a file containing the JSON descriptor with all the platform fields and annotations you would want that item to have in the final manifest list. But yes, I agree it could use more docs (and I'm not sure how well tested it is).

Not against this PR as well as clearly seems something that was forgotten, but generally recommend `buildx` instead of manifest command.

thaJeztah:
@tonistiigi ok to go?

sjackman:
> -f can point to a file containing the JSON descriptor with all the platform fields and annotations you would want that item to have in the final manifest list. But yes, I agree it could use more docs (and I'm not sure how well tested it is).

@tonistiigi Could you please give an example of this JSON file?

sjackman:
@tonistiigi wrote…

> fyi, this use-case is also covered with buildx imagetools create

How can `docker buildx imagetools create` set `os.version`?

sjackman:
I successfully uploaded an image index that includes `os.version` using Skopeo! https://github.com/containers/skopeo

Issue #2565, opened 2020-06-06
Docker image command Bug.


ShivamTripathi21:
I'm not sure whether it's a bug or not, but when I was going through code I found some command level discrepancy.
usually when we run:-
1. docker images
2. docker image ls 
both return the same result as expected but
3.  docker image images
is also returning the same result. is it expected? because I never heard of it.

and also when you run:- docker image -help
it doesn't show images as a subcommand but images work, so it seems like a valid bug.

ShivamTripathi21:
[Code Link](https://github.com/docker/cli/blob/82de74a5e3a745d29c96227f5c16914736411bda/cli/command/image/list.go#L53)
During calling newListCommand, you are calling NewImagesCommand first and that is why images is also working.

thaJeztah:
Hmm. yes, looks like it's "valid" from a technical perspective, but still interesting / confusing. Perhaps we're able to check if the command is called as `docker image ls` or as `docker images` (top-level), and produce an error 🤔.

Looks like a "corner case", so not top-priority, but would be "nice" to fix (contributions welcome if the fix is not adding too much complexity)

PR #2580, opened 2020-06-14
Remove 'images' alias from 'image ls' command

dominikbraun:
Signed-off-by: Dominik Braun <Dominik.Braun@nbsp.de>
Fixes: #2565 

**- What this is about**
In Issue #2565 it was found that `docker image images` is a confusing command that produces the same output as `docker image ls` or the `docker images` legacy command. @thaJeztah proposed that we [could produce an error if the command is called likes this](https://github.com/docker/cli/issues/2565#issuecomment-642616747):

> Perhaps we're able to check if the command is called as docker image ls or as docker images (top-level), and produce an error 🤔.

However, I found that the command `docker image images` is technically not the exact same command as `docker images` - it is merely an alias of `docker image ls` instead. I'm calling this 'not the exact same command' because `docker image ls` _is_ registered as an own cobra command, but nevertheless it is created using the same method as `docker images` is: `NewImagesCommand()`.

image

Since `docker image ls` and `docker images` are two distinct command objects, IMO it is ok to just throw out the `images` alias for `ls`.

**- What I did**
Removed the alias.

**- How to verify it**
Run `docker image images` and note that it isn't a valid command anymore, while `docker images` still is.

Remove 'images' alias from 'docker image ls' command

### DIFF
--- a/cli/command/image/list.go
+++ b/cli/command/image/list.go
@@ -52,7 +52,7 @@ func NewImagesCommand(dockerCli command.Cli) *cobra.Command {
 
 func newListCommand(dockerCli command.Cli) *cobra.Command {
 	cmd := *NewImagesCommand(dockerCli)
-	cmd.Aliases = []string{"images", "list"}
+	cmd.Aliases = []string{"list"}
 	cmd.Use = "ls [OPTIONS] [REPOSITORY[:TAG]]"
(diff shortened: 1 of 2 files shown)


dominikbraun:
I squashed that two commits.

PR #2583, opened 2020-06-15
Don't filter out registries to logout from with config file contents

simonferquel:
supersedes / closes

closes https://github.com/docker/cli/pull/1917
closes https://github.com/docker/cli/pull/2346
fixes https://github.com/docker/cli/issues/204

**- What I did**
Previously, if a registry AuthInfo was not present in the CLI config file, docker logout could not be used
to ask the credential helper to forget about it. It causes problem for people working with
multiple alternative config files, and it causes problems for cases like Docker Desktop w/ WSL 2, as
it uses the same win32 credential helper as the Windows CLI, but a different config file, leading to
bugs where I cannot logout from a registry from wsl2 if I logged in from Windows and vice-versa.

**- How I did it**
Change the logic to call the credential helper for each candidate entry, and comparing the number of returned errors to the number of tries.

**- How to verify it**
- `docker login` (with a credential helper)
- remove the authInfo from `~/.docker/config.json`
- ensure private pull / push operations still work
- `docker logout`
- ensure private pull / push operations fail

### DIFF
--- a/cli/command/registry/logout.go
+++ b/cli/command/registry/logout.go
@@ -39,36 +39,29 @@ func runLogout(dockerCli command.Cli, serverAddress string) error {
 	}
 
 	var (
-		loggedIn        bool
-		regsToLogout    []string
+		regsToLogout    = []string{serverAddress}
 		hostnameAddress = serverAddress
-		regsToTry       = []string{serverAddress}
 	)
 	if !isDefaultRegistry {
 		hostnameAddress = registry.ConvertToHostname(serverAddress)
 		// the tries below are kept for backward compatibility where a user could have
 		// saved the registry in one of the following format.
-		regsToTry = append(regsToTry, hostnameAddress, "http://"+hostnameAddress, "https://"+hostnameAddress)
-	}
-
-	// check if we're logged in based on the records in the config file
-	// which means it couldn't have user/pass cause they may be in the creds store
-	for _, s := range regsToTry {
-		if _, ok := dockerCli.ConfigFile().AuthConfigs[s]; ok {
-			loggedIn = true
-			regsToLogout = append(regsToLogout, s)
-		}
-	}
-
-	if !loggedIn {
-		fmt.Fprintf(dockerCli.Out(), "Not logged in to %s\n", hostnameAddress)
-		return nil
+		regsToLogout = append(regsToLogout, hostnameAddress, "http://"+hostnameAddress, "https://"+hostnameAddress)
 	}
 
 	fmt.Fprintf(dockerCli.Out(), "Removing login credentials for %s\n", hostnameAddress)
+	errs := make(map[string]error)
 	for _, r := range regsToLogout {
 		if err := dockerCli.ConfigFile().GetCredentialsStore(r).Erase(r); err != nil {
-			fmt.Fprintf(dockerCli.Err(), "WARNING: could not erase credentials: %v\n", err)
+			errs[r] = err
+		}
+	}
+
+	// if at least one removal succeeded, report success. Otherwise report errors
+	if len(errs) == len(regsToLogout) {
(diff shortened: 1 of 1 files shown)


cli/command/registry/logout.go
```diff
@@ -39,36 +39,29 @@ func runLogout(dockerCli command.Cli, serverAddress string) error {
...
+	errs := make(map[string]error)
 	for _, r := range regsToLogout {
 		if err := dockerCli.ConfigFile().GetCredentialsStore(r).Erase(r); err != nil {
-			fmt.Fprintf(dockerCli.Err(), "WARNING: could not erase credentials: %v\n", err)
+			errs[r] = err
+		}
+	}
+
+	// if at least one removal succeeded, report success. Otherwise report errors
+	if len(errs) == len(regsToLogout) {
```
thaJeztah, line 61: Should we check if `len(errs) > 0` or `len(regsToLogout) > 0` as well?

thaJeztah:
I see another PR also added some tests; https://github.com/docker/cli/pull/2346. Haven't looked closely, but could you check if those tests would be useful to have?

cli/command/registry/logout.go
```diff
@@ -39,36 +39,29 @@ func runLogout(dockerCli command.Cli, serverAddress string) error {
...
-		if _, ok := dockerCli.ConfigFile().AuthConfigs[s]; ok {
-			loggedIn = true
-			regsToLogout = append(regsToLogout, s)
-		}
-	}
-
-	if !loggedIn {
-		fmt.Fprintf(dockerCli.Out(), "Not logged in to %s\n", hostnameAddress)
-		return nil
+		regsToLogout = append(regsToLogout, hostnameAddress, "http://"+hostnameAddress, "https://"+hostnameAddress)
```
thaJeztah, line 49: Actually, given that we duplicate `hostnameAddress` and perform a logout three times for each registry, I'm wondering if we should make `regsToLogout` a `map[string][]string`, so that we (at most) generate a single error per registry

```suggestion
		regsToLogout[hostnameAddress] = []string{hostnameAddress, "http://"+hostnameAddress, "https://"+hostnameAddress}
```

wdyt? 🤔

simonferquel:
> I see another PR also added some tests; #2346. Haven't looked closely, but could you check if those tests would be useful to have?

Hmm, the thing is those tests do not really apply there I think, as it changes the errors reported here (and especially, as a side effect, logout is idempotent - at least for creds helper that don't report an error when loging out from a not logged-in registry)

cli/command/registry/logout.go
simonferquel: regsToLogout cannot be empty (it is initialized with one value, and we append other candidates in it if it is not the default registry)

cli/command/registry/logout.go
simonferquel: Actually, we logout from a single registry here. It is just that we generate a list of candidates so that `docker logout toto` tries to logout from both `toto`, `http://toto` and `https://toto`. So there will always be only one entry in this map.
I wonder though if we could only print the first error...

thaJeztah: Ah, yes, had my wires crossed, and thinking we're logging out from all registries. Perhaps it's ok for now, considering this a "these are the URLs we tried to logout from, and this is why they failed".

Only thing I'm wondering is if we produce an error if "no entry was found"; if so, and if we would be able to detect that cause, we (IMO) should ignore those, and treat it as an idempotent action.

simonferquel: The "plaintext" credstore implementation is already idempotent, and the wincred helper as well. I am not sure about the mac one, and 3rd parties. But I agree we should document the intended behavior should be idempotent (now that the CLI code is)

PR #2593, opened 2020-06-18
docs: fix example output for docker plugin ls

thaJeztah:
The output format was changed to combine tag and name in a single column, but this change was never reflected in the docs.


thaJeztah:

PR #2596, opened 2020-06-21
Add checks for dependencies that are not vendored

### LINKED ISSUE #2332: ci: add check for missing dependencies in vendor-check
relates to https://github.com/docker/cli/pull/2329

`vndr` prints missing dependencies as a `WARNING`. We currently don't appear to be catching these issues, and ignore those warnings.

We should update the vendor check to fail if a `WARNING: dependency is not vendored` is printed;

```
vndr github.com/Microsoft/hcsshim

2020/02/12 16:40:16 Collecting initial packages
2020/02/12 16:40:16 Download dependencies
2020/02/12 16:40:20 	Clone github.com/Microsoft/hcsshim, revision b3f49c06ffaeef24d09c6c08ec8ec8425a0303e2, attempt 1/20
2020/02/12 16:40:25 	Finished clone github.com/Microsoft/hcsshim
2020/02/12 16:40:25 Dependencies downloaded. Download time: 4.625264419s
2020/02/12 16:40:25 Collecting all dependencies
2020/02/12 16:40:32 WARNING: dependency is not vendored: github.com/containerd/cgroups/stats/v1
2020/02/12 16:40:32 Clean vendor dir from unused packages
2020/02/12 16:40:32 Success
2020/02/12 16:40:32 Running time: 16.278317994s
```

Madhur1997:
Signed-off-by: Madhur Batra <madhurbatra097@gmail.com>

**- What I did**
    Fix #2332: `make vendor` now regards 'WARNING: dependency is not vendored' as an
                         error and fails when some non vendored dependencies exist.

**- How I did it**
     Added a new validation script that runs against the `vndr` output to find any 
     'WARNING: dependency is not vendored' line. If found script exits with non zero status.

**- How to verify it**
     Added a new import in a go file, and ran go get to install the same in GOPATH.
     Now `vndr check` failed due to unvendored dependency.

Enhanced existing Makefile to store the `vndr` logs in a tmp file.
Ran a new script against the log to verify that there is no unvendored dependency.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,4 @@
 appveyor.yml
 build
 circle.yml
+/vndr.log
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ cli/winresources/rsrc_amd64.syso
 /docs/yaml/gen/
 coverage.txt
 profile.out
+/vndr.log
--- a/Makefile
+++ b/Makefile
@@ -64,8 +64,9 @@ dynbinary: ## build dynamically linked binary
 
 vendor: vendor.conf ## check that vendor matches vendor.conf
 	rm -rf vendor
-	bash -c 'vndr |& grep -v -i clone'
+	bash -c 'vndr |& grep -v -i clone | tee ./vndr.log'
 	scripts/validate/check-git-diff vendor
+	scripts/validate/check-all-packages-vendored
 
 .PHONY: authors
 authors: ## generate AUTHORS file from git history
--- a/scripts/validate/check-all-packages-vendored
+++ b/scripts/validate/check-all-packages-vendored
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+set -eu
+
+WARN_MISS_VENDOR='WARNING: dependency is not vendored'
+
+if grep -q "$WARN_MISS_VENDOR" "./vndr.log"; then
+	echo 
+	echo "Some dependencies are not vendored."
+	echo
+	exit 1
+fi


Makefile
```diff
@@ -64,8 +65,9 @@ dynbinary: ## build dynamically linked binary
 
 vendor: vendor.conf ## check that vendor matches vendor.conf
 	rm -rf vendor
-	bash -c 'vndr |& grep -v -i clone'
+	bash -c 'vndr |& grep -v -i clone | tee /tmp/vndr.log'
```
thaJeztah, line 68: Perhaps we should store this in `./build/` as well 🤔

scripts/validate/check-all-packages-vendored
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -eu
+
+FILE_PATH='/tmp/vndr.log'
+WARN_MISS_VENDOR='WARNING: dependency is not vendored'
+
+if grep -q $WARN_MISS_VENDOR $FILE_PATH; then
```
thaJeztah, line 8: We should quote `$FILE_PATH` here, or perhaps we should just hard-code it, because it's not something we really change

scripts/validate/check-all-packages-vendored
```diff
@@ -0,0 +1,13 @@
...
+	echo 
+	echo "Some dependencies are not vendored."
```
thaJeztah, line 10: Would it be difficult to print which dependencies? Perhaps we can print the output of those warning lines?

Makefile
Madhur1997: sure, makes sense. Will update it.

scripts/validate/check-all-packages-vendored
Madhur1997: okay

scripts/validate/check-all-packages-vendored
Madhur1997: umm, the previous run bash command already prints these lines.
bash -c 'vndr |& grep -v -i clone | tee /tmp/vndr.log'
But these lines are intertwined with lot of other logs as well. 
We can add the warning lines again or perhaps ask the user to look at the log file.
What do you think?

Makefile
Madhur1997: I see build dir doesn't exist in ci server. It fails with 'no such directory'.
I added it to the project's root dir and also added the file in .gitignore, .dockerignore

.gitignore
```diff
@@ -16,3 +16,4 @@ cli/winresources/rsrc_amd64.syso
 /docs/yaml/gen/
 coverage.txt
 profile.out
+./vndr.log
```
thaJeztah, line 19: I think either

```suggestion
/vndr.log
```

Or (to match the other ones, although they're not "strictly" correct if the intent was to only ignore at the root

```suggestion
vndr.log
```

.dockerignore
```diff
@@ -4,3 +4,4 @@
 appveyor.yml
 build
 circle.yml
+./vndr.log
```
thaJeztah, line 7: (same comment as for the other one)

```suggestion
vndr.log
```

Madhur1997:
Thanks for the review @thaJeztah. 
Updated the PR with the addressed comments.

Madhur1997:
Hey @thaJeztah 

Wanted to know if there is any other review comment or am I missing anything here?
Thanks.

PR #2600, opened 2020-06-24
docs: update for cgroup v2 and rootless

AkihiroSuda:
Updated the description toward v20.0X. (https://github.com/moby/moby/pull/40846 https://github.com/moby/moby/pull/40483)

We also need to have a description for cgroup v2 mode, but not sure where is the right place.


AkihiroSuda:
Where is the right place to add cgroup2 docs?

AkihiroSuda:
https://github.com/docker/docker.github.io/blob/master/config/containers/runmetrics.md seems the right place to add cgroup 2 docs. Will open a PR.

PR #2603, opened 2020-06-26
ReadKVEnvStrings/ReadKVStrings return nil if empty, and add tests

thaJeztah:
If no env-vars were loaded from "files", and "overrides" was nil, the code returned an empty slice instead of a `nil` value.

Also add a test for this function, as no unit test was present yet.

Noticed this when reviewing https://github.com/docker/cli/pull/2602, where adding an env-file caused the container's Env to be different in tests.

### DIFF
--- a/opts/parse.go
+++ b/opts/parse.go
@@ -23,7 +23,7 @@ func ReadKVEnvStrings(files []string, override []string) ([]string, error) {
 }
 
 func readKVStrings(files []string, override []string, emptyFn func(string) (string, bool)) ([]string, error) {
-	variables := []string{}
+	var variables []string
 	for _, ef := range files {
 		parsedVars, err := parseKeyValueFile(ef, emptyFn)
 		if err != nil {
--- a/opts/parse_test.go
+++ b/opts/parse_test.go
@@ -1,0 +1,72 @@
+package opts
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/env"
+	"gotest.tools/v3/fs"
+)
+
+func TestReadKVEnvStrings(t *testing.T) {
+	emptyEnvFile := fs.NewFile(t, t.Name())
+	defer emptyEnvFile.Remove()
+
+	// EMPTY_VAR should be set, but with an empty string as value
+	// FROM_ENV value should be substituted with FROM_ENV in the current environment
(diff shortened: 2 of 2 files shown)


opts/parse.go
```diff
@@ -23,14 +23,17 @@ func ReadKVEnvStrings(files []string, override []string) ([]string, error) {
...
-	variables := []string{}
+	var variables []string
 	for _, ef := range files {
 		parsedVars, err := parseKeyValueFile(ef, emptyFn)
 		if err != nil {
 			return nil, err
 		}
 		variables = append(variables, parsedVars...)
 	}
+	if len(variables) == 0 {
```
silvin-lubecki, line 34: Question: Why not just return the `append` which would be no-op?
```go
var variables []string
for _, ef := range files {
...
}
return append(variables, override...), nil
```

opts/parse_test.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func TestReadKVEnvStrings(t *testing.T) {
+	emptyEnvFile := fs.NewFile(t, t.Name())
+	defer emptyEnvFile.Remove()
+
+	// EMPTY_VAR should be set, but with an empty string as value
+	// FROM_ENV value should be substituted with FROM_ENV in the current environment
+	// NO_SUCH_ENV does not exist in the current environment, so should be omitted
+	envFile1 := fs.NewFile(t, t.Name(), fs.WithContent("Z1=z\nEMPTY_VAR=\nFROM_ENV\nNO_SUCH_ENV"))
```
silvin-lubecki, line 18: nit: use backquote notation and "real" indentation instead of `\n` ?

opts/parse.go
thaJeztah: Basically did that to prevent the value from changing (`[]string{}` changing to `nil` or vice-versa).

opts/parse_test.go
thaJeztah: Thought this was slightly shorter, but I can change yes (I dislike the literals a bit if they use multiple lines as it messes up indentation of the code 😂)

silvin-lubecki: Yes same here... but I think it's still more readable than `KEY=VALUE\nKEY=VALUE\n`

silvin-lubecki: and it's just nit picking BTW 😄

PR #2613, opened 2020-06-30
Fix experimental README instructions

Funcan:
The correct syntax for the "experimental" option in config.json is as a
stribe "enabled" or "disabled, not a boolean. Following the README as is
creates an error during unmarsall when running any docker cli commands.

**- What I did**
Tried to follow the instructions in the README, got an error, found the correct syntax
**- How I did it**
Read the error message
**- How to verify it**
Follow the old instructions, get an error. Follow the updated instructions and you don't

Fix syntax in experimental README

### DIFF
--- a/experimental/README.md
+++ b/experimental/README.md
@@ -19,7 +19,7 @@ To enable experimental features, start the Docker daemon with the
 
 ```json
 {
-    "experimental": true
+    "experimental": "enabled"
 }
 ```
 


experimental/README.md
```diff
@@ -19,7 +19,7 @@ To enable experimental features, start the Docker daemon with the
 
 ```json
 {
-    "experimental": true
+    "experimental": "enabled"
```
thaJeztah, line 22: Hm, this is for the _daemon_ configuration, not the CLI. The Cli uses "enabled" / "disabled", but the daemon uses a boolean

Funcan: Ah ha, got it, thanks. That's horrible that they're different, but understood. I'll close this off.

Funcan:
Closing since I misunderstood the purpose of the README. Sorry for the noise.

PR #2621, opened 2020-07-07
Optional stream suppression for "container logs" command.

federvieh:
**- What I did**
In the "container log" command it's now possible to suppress output of the stdout or stderr stream using the "--no-stdout" or "--no-stderr" option.

**- How I did it**
Exposing inversion of the already existing flags `ShowStdout` and `ShowStderr` from `types.ContainerLogsOptions` to the command line options `--no-stdout` or `--no-stderr`.

**- How to verify it**
Run `docker container logs --no-stdout` or `docker container logs --no-stderr` on a container that generated outputs to both of those streams.

Added options "--no-stdout" and "--no-stderr" to the "container log" command.

Image of Koala climbing tree

### DIFF
--- a/cli/command/container/logs.go
+++ b/cli/command/container/logs.go
@@ -18,6 +18,8 @@ type logsOptions struct {
 	timestamps bool
 	details    bool
 	tail       string
+	noStdout   bool
+	noStderr   bool
 
 	container string
 }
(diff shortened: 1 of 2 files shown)


cpuguy83:
I guess this is fine? Typically, though, you'd use shell redirects to handle this.
It doesn't hurt to pass these options to the daemon at least since it can save resources (though unfortunately as is it still has to process each log line).

federvieh:
Oh my! I'm new to docker and honestly didn't realize that "container logs" generates its output to the original streams.

thaJeztah:
Could you describe what the use-case is for this? The standard way is to redirect streams if you want to suppress them (or redirect to a different file)

federvieh:
I'm fine with this PR being closed without merging.

As I wrote yesterday, I didn't realize that the output of "container logs" goes to the original streams, so that standard shell redirection can be used. I did not find any description of this behaviour and it was not obvious to me (it could of course be obvious to others).

@cpuguy83 mentioned above that "It doesn't hurt to pass these options to the daemon at least since it can save resources (though unfortunately as is it still has to process each log line)." That's why I kept it open.

thaJeztah:
Thanks for the extra information

Discussing this with @cpuguy83 and other in our maintainers meeting, and given that there's no urgent need or use-case for this (and that it won't save resources), we'd rather not add these flags for now (in favour of the regular way to redirect streams in the shell).

I'll close this PR for that reason (but _thank you_ for contributing!). If there's a strong reason for adding this in future, we can always reconsider 👍

PR #2631, opened 2020-07-15
[19.03] handle close error on save

cpuguy83:
Backports:
- 22a291f703b2511daaa8e0d325ab48523e817558 https://github.com/docker/cli/pull/2228 config: preserve ownership and permissions on configfile
    - fixes https://github.com/moby/moby/issues/40175 Bug: sudo docker login changes .docker/config.json permissions
- d02173090ff15ffb3ec4a5427ba35dafeb7fa92b https://github.com/docker/cli/pull/2595 Handle errors on close in config file write

The first because it seemed at least relevant and the fact that it was missing caused a merge conflict (simple to fix, but still).
Can remove if needed.

### DIFF
--- a/cli/config/configfile/file_windows.go
+++ b/cli/config/configfile/file_windows.go
@@ -1,0 +1,5 @@
+package configfile
+
+func copyFilePermissions(src, dst string) {
+	// TODO implement for Windows
+}
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -196,6 +196,9 @@ func (configFile *ConfigFile) Save() error {
 		os.Remove(temp.Name())
 		return err
 	}
+	// Try copying the current config file (if any) ownership and permissions
+	copyFilePermissions(configFile.Filename, temp.Name())
+
 	return os.Rename(temp.Name(), configFile.Filename)
 }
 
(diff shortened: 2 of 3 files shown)


cli/config/configfile/file_windows.go
```diff
@@ -0,0 +1,5 @@
+package configfile
+
+func copyFilePermissions(src, dst string) {
+	// TODO implement for Windows
```
silvin-lubecki, line 4: Question: what would be the implementation for windows?

thaJeztah: Not sure how to copy permissions and/or ownership on Windows. Perhaps @simonferquel would know.

Also not sure if the problem that this was fixing happens on Windows as well

simonferquel: Hmm, I am not sure about the correct "Windows" way to do that. But anyway I don't think it applies (when you create a file on Windows, by default you inherit ACLs from the parent directory).

thaJeztah: Thx! Perhaps I should update the comment and remove "todo"

simonferquel: also renaming the function would be welcome, because on the windows implementation it does not do what it claims to do (in the general case it does not copy the permissions, which is fine in this specific case, but could be confusing in the future). Maybe renaming it to `replicateConfigFilePermissions` or something like this would be preferable.

thaJeztah: Don't think it's needed to include "configfile" in the name, as it's already in this package, and in itself, it's not specific to configfiles.

I'll have a look for changing the comment (and possibly name) on master

PR #2654, opened 2020-07-27
support emulated version detection

tonistiigi:
Arm64 on Darwin supports emulation of amd64 binaries. This makes emulation visible in `docker version` so it can be detected as a factor in bug reports.

Using `docker build --platform=local` now builds for the actual native platform, not the platform for the binary that is being emulated.

### DIFF
--- a/cli/command/image/build_buildkit.go
+++ b/cli/command/image/build_buildkit.go
@@ -34,6 +34,7 @@ import (
 	"github.com/moby/buildkit/util/progress/progressui"
 	"github.com/pkg/errors"
 	fsutiltypes "github.com/tonistiigi/fsutil/types"
+	"github.com/tonistiigi/go-rosetta"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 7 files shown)


cli/command/image/build_buildkit.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/moby/buildkit/util/progress/progressui"
 	"github.com/pkg/errors"
 	fsutiltypes "github.com/tonistiigi/fsutil/types"
+	"github.com/tonistiigi/go-rosetta"
```
thaJeztah, line 37: can you add a license to github.com/tonistiigi/go-rosetta ?

thaJeztah: Or would this be code we want to roll up into (e.g.) https://github.com/moby/moby/tree/cadd72d37d57e2dc686502d9f9ab0fc73d3b6af8/pkg/parsers

PR #2655, opened 2020-07-27
docs: remove docs for —-from=index

tonistiigi:
Naming stages is the preferred method for using
multi-stage builds.


thaJeztah:
Is the plan to deprecate the index variant?

thaJeztah:
Also should we clarify that it also accepts images?

tonistiigi:
I'm not sure if we actually need to deprecate. No need to remove and break somebody, but this is generating confusion for new users and should be avoided.

tonistiigi:
> Also should we clarify that it also accepts images?

tonistiigi:
> Also should we clarify that it also accepts images?

There's already a comment for it.

thaJeztah:
@silvin-lubecki PTAL

PR #2656, opened 2020-07-28
vendor: buildkit 4d1f260e8490ec438ab66e08bb105577aca0ce06

thaJeztah:
full diff: https://github.com/moby/buildkit/compare/df35e9818d1f9066e616e03f4b8d727c97562e5b...4d1f260e8490ec438ab66e08bb105577aca0ce06

- moby/buildkit#1551 session: track sessions with a group construct
- moby/buildkit#1534 secrets: allow providing secrets with env
- moby/buildkit#1533 git: support for token authentication
- moby/buildkit#1549 progressui: fix logs time formatting


thaJeztah:
@tonistiigi @tiborvass @AkihiroSuda PTAL if the changes for secrets and git require documentation changes and if local changes are needed to expose the functionality

thaJeztah:
Ah, guess this answers some of my questions;

```
#14 0.375 Building statically linked build/docker-linux-amd64
#14 26.34 # github.com/docker/cli/cli/command/image
#14 26.34 cli/command/image/build_buildkit.go:427:15: undefined: secretsprovider.FileSource
#14 26.34 cli/command/image/build_buildkit.go:435:16: undefined: secretsprovider.NewFileStore
#14 26.34 cli/command/image/build_buildkit.go:442:34: undefined: secretsprovider.FileSource
#14 26.34 cli/command/image/build_buildkit.go:449:8: undefined: secretsprovider.FileSource
#14 ERROR: executor failed running [/bin/sh -c ./scripts/build/binary]: runc did not terminate sucessfully
```

tonistiigi:
Spotted possible issue, fixed in https://github.com/moby/buildkit/pull/1613 . Don't need to be vendored but just same fix can be done in cli.

thaJeztah:
I'm a bit confused by the UX. Looking at it, it doesn't seem consistent.

The `--secret` flag currently supports these keys: `type`, `id`, `source`, `src`, `env`

- `id` is a required option
- `type` is full ignored, unless an invalid type is specified (anything other than "file" or "env" (the code doesn't follow different paths if `file` or `env` is specified)
- both `source`, `src` _and_ `env` can be specified at the same time.
    - this is ambiguous if no explicit "type" is set
- if `type` is set to "env", `source` / `src` acts as an alias for `env` (if unset)
- if only `id` is set (and no `source / src` or `env`), and an environment variable is found that matches that name, the default is now switched to "take value from the environment variable"
    - This is a breaking change: previously `id` would be used as filename

I _think_

- if no `type` is set, and no `source / src / env`, use the current default and use `id` as `source` (current behavior)
- if no `type` is set, but `source / src` _is_ set; implicitly assume `type=file` (could be seen as same as above, as `file` is default)
- if no `type` is set, but `env` _is_ set; implicitly assume `type=env`
- if `type=file`
    - only look at `source / src` (ignore `env`)
    - fall back to `id` if not set
    - produce error if `env` is set? (invalid option for `type=file` ?
- if `type=env`
    - look at `env` to use as "source"
    - fall back to `source / src` if not set
    - fall back to `id` if not set
    - produce error if both `env` and `source / src` are set?


However, from the above; does `env=` provide much value, given that we already have `type=env`, and that `source/src` and `id` already provide the functionality?
Effectively if would only be useful to allow switching between `file` and `env` without having to explicitly set `type=`.

One "feature" could be to explicitly allow `env` as a "fallback", so;

```
--secret id=foo-token,src=~/my-token.txt,env=FOO_TOKEN
```

Which would _either_ pick the value from `$FOO_TOKEN` or from `~/my-token.txt` (not sure in which order though)


If we would remove `env=`, the combinations would be reduced quite a bit;

- if `type=file` or if no `type` is set (default)
    - look at `source / src`
    - fall back to `id` if not set
    - produce error if the file was not found
- if `type=env`
    - look at `source / src`
    - fall back to `id` if not set
    - produce error if no environment variable with the given name was found? (TBD)

cli/command/image/build_buildkit.go
```diff
@@ -459,17 +460,24 @@ func parseSecret(value string) (*secretsprovider.FileSource, error) {
...
 			fs.ID = value
 		case "source", "src":
 			fs.FilePath = value
+		case "env":
+			fs.Env = value
 		default:
 			return nil, errors.Errorf("unexpected key '%s' in '%s'", key, field)
 		}
 	}
+	if typ == "env" {
```
thaJeztah, line 477: need to add this

```suggestion
	f typ == "env" && fs.Env == "" {
```

tonistiigi (approved):

tiborvass (approved):

PR #2526, opened 2020-05-13
document deprecation of legacy `~/.dockercfg` config-file

thaJeztah:
relates to https://github.com/docker/cli/pull/2504

The docker CLI up until v1.7.0 used the `~/.dockercfg` file to store credentials
after authenticating to a registry (`docker login`). Docker v1.7.0 replaced this
file with a new CLI configuration file, located in `~/.docker/config.json`. When
implementing the new configuration file, the old file (and file-format) was kept
as a fall-back, to assist existing users with migrating to the new file.

Given that the old file format encourages insecure storage of credentials
(credentials are stored unencrypted), and that no version of the CLI since
Docker v1.7.0 has created this file, the file is marked deprecated, and support
for this file will be removed in a future release.

Signed-off-by: Sebastiaan van Stijn <github@gone.nl>

**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #2666, opened 2020-08-03
config: print deprecation warning when falling back to ~/.dockercfg

thaJeztah:
follow-up to https://github.com/docker/cli/pull/2526

The docker CLI up until v1.7.0 used the `~/.dockercfg` file to store credentials
after authenticating to a registry (`docker login`). Docker v1.7.0 replaced this
file with a new CLI configuration file, located in `~/.docker/config.json`. When
implementing the new configuration file, the old file (and file-format) was kept
as a fall-back, to assist existing users with migrating to the new file.

Given that the old file format encourages insecure storage of credentials
(credentials are stored unencrypted), and that no version of the CLI since
Docker v1.7.0 has created this file, the file is marked deprecated, and support
for this file will be removed in a future release.

This patch adds a deprecation warning, which is printed if the CLI falls back
to using the deprecated `~/.dockercfg` file.

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -24,12 +24,12 @@ const (
 )
 
 var (
-	initConfigDir sync.Once
+	initConfigDir = new(sync.Once)
 	configDir     string
 	homeDir       string
 )
 
-// resetHomeDir is used in testing to resets the "homeDir" package variable to
+// resetHomeDir is used in testing to reset the "homeDir" package variable to
 // force re-lookup of the home directory between tests.
 func resetHomeDir() {
 	homeDir = ""
@@ -40,6 +40,13 @@ func getHomeDir() string {
 		homeDir = homedir.Get()
 	}
 	return homeDir
+}
+
+// resetConfigDir is used in testing to reset the "configDir" package variable
+// and its sync.Once to force re-lookup between tests.
+func resetConfigDir() {
+	configDir = ""
+	initConfigDir = new(sync.Once)
 }
 
 func setConfigDir() {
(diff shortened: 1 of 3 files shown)


devdreamsolution (approved):

thaJeztah:
@silvin-lubecki @tiborvass ptal

petar8903:
Tenks

cli/config/config.go
```diff
@@ -140,6 +160,11 @@ func LoadDefaultConfigFile(stderr io.Writer) *configfile.ConfigFile {
 	if err != nil {
 		fmt.Fprintf(stderr, "WARNING: Error loading config file: %v\n", err)
 	}
+	mutex.RLock()
+	defer mutex.RUnlock()
+	if printLegacyFileWarning {
+		_, _ = fmt.Fprintln(stderr, "WARNING: The legacy ~/.dockercfg configuration file and file-format are deprecated and will be removed in an upcoming release")
```
silvin-lubecki, line 166: Nit: maybe change "will be removed" by "support will be removed"? Otherwise it could be interpreted as the config file itself will be removed

silvin-lubecki (approved):
LGTM, just small nit on messaging

cli/config/config.go
thaJeztah: Something like this?

```suggestion
		_, _ = fmt.Fprintln(stderr, "WARNING: Support for the legacy ~/.dockercfg configuration file and file-format is deprecated and will be removed in an upcoming release")
```

silvin-lubecki: Yes thanks 👍

Issue #2549, opened 2020-05-25
`docker container run --g[TAB]` no `--gpus` completion for `zsh`


yantaozhao:
**Description**

**Steps to reproduce the issue:**
1. In `zsh` shell prompt, input `docker container run --g[TAB][TAB]`, there's no `--gpus` completion shows in the candidates.

**Describe the results you expected:**
In `zsh` shell prompt, when press [TAB][TAB] to trigger the completion, the `--gpus` option should shows, along with such as `--group`, in the candidates.

Moreover, followed with the value set of `all` and `device=` would be better (i.e.: --gpus [all device=] ), just like the `--network` ways.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
% docker --version 
Docker version 19.03.8, build afacb8b7f0
```

**Output of `docker info`:**

```
$ docker info     
Client:
 Debug Mode: false

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 6
 Server Version: 19.03.8
 Storage Driver: overlay2
  Backing Filesystem: <unknown>
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc version: dc9208a3303feef5b3839f4323d9beb36df0a9dd
 init version: fec3683
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.0.0-32-generic
 Operating System: Ubuntu 18.04.4LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.55GiB
 ID: WZP2:QT4J:PRAA:ZXIF:JMXR:TLOL:ZES3:MX7S:5QOQ:4C7N:7IMY:65CZ
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Registry Mirrors:
 Live Restore Enabled: false
```

[...]

PR #2672, opened 2020-08-09
Issue 2549/autocomplete group and gpus support

rchenzheng:
**- What I did**
Confirm the issues presented in https://github.com/docker/cli/issues/2549 and https://github.com/docker/cli/issues/2549#issuecomment-640055524

Fix Auto complete options available:
- Add support for --gpus to run/create container in bash and zsh
- Remove --group from run and update container as it's not a valid flag in zsh
- Add --group-add --group-rm for create and update update service in bash

**- How I did it**
Update the auto complete files

**- How to verify it**
For zsh
```
/usr/local/bin/zsh

source contrib/completion/zsh/_docker
docker container run --g[TAB][TAB]
--gpu and --group-add
docker service create --g[TAB][TAB]
--group
docker service update --g[TAB][TAB]
--group-add and --group-rm

docker container --help
docker service --help
```

For bash
```
/bin/bash
 source contrib/completion/bash/docker

docker container run --g[TAB][TAB]
--gpus and --group-add
docker service create --g[TAB][TAB]
--generic-resource and --group
docker service update --g[TAB][TAB]
--generic-resource-add  --generic-resource-rm   --group-add             --group-rm 

docker container --help
docker service --help
```

I also went through the library option and tested each command and confirmed that those options were valid for docker-cli aka no 'unknown flag:`

Fix Auto complete options available:
- Add support for --gpus to run/create container in bash and zsh
- Remove --group from run and update container as it's not a valid flag in zsh
- Add --group-add --group-rm for create and update update service in bash

https://i.pinimg.com/originals/a7/da/a1/a7daa190107fa107413036bdace84884.jpg

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -1894,6 +1894,7 @@ _docker_container_run_and_create() {
 		--env -e
 		--env-file
 		--expose
+		--gpus
 		--group-add
 		--health-cmd
 		--health-interval
--- a/contrib/completion/zsh/_docker
+++ b/contrib/completion/zsh/_docker
@@ -624,7 +624,8 @@ __docker_container_subcommand() {
         "($help)--entrypoint=[Overwrite the default entrypoint of the image]:entry point: "
         "($help)*--env-file=[Read environment variables from a file]:environment file:_files"
         "($help)*--expose=[Expose a port from the container without publishing it]: "
-        "($help)*--group=[Set one or more supplementary user groups for the container]:group:_groups"
+        "($help)*--gpus=[GPU devices to add to the container ('all' to pass all GPUs)]:device: "
+        "($help)*--group-add=[Set one or more supplementary user groups for the container]:group:_groups"
         "($help -h --hostname)"{-h=,--hostname=}"[Container host name]:hostname:_hosts"
         "($help -i --interactive)"{-i,--interactive}"[Keep stdin open even if not attached]"
         "($help)--init[Run an init inside the container that forwards signals and reaps processes]"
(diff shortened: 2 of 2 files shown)


contrib/completion/bash/docker
```diff
@@ -3842,7 +3843,7 @@ _docker_service_update_and_create() {
 			__docker_nospace
 			return
 			;;
-		--group|--group-add|--group-rm)
+		--group-add|--group-rm)
```
albers, line 3846: With this change, `docker service create --group <tab>` no longer completes groups.

rchenzheng: It still does in my case, can you confirm? 

```
source contrib/completion/bash/docker
source ~/.bash_profile

docker service create --g
bash-3.2$ docker service create --g
--generic-resource  --group             
bash-3.2$ docker service create --g
--generic-resource  --group             
bash-3.2$ docker service create --g
```

albers: @rchenzheng
> With this change, `docker service create --group <tab>` no longer completes groups.

The issue is about completing the arguments of the `--group` option.

The code above is only called when the preceding argument is either `--group`, `--group-add` or `--group-rm`. In the context of `service create`, `--group` is still valid and needs support for completing its arguments.

rchenzheng: Got it! Sorry, it's been a busy day

rchenzheng:
@albers Looks like we need to rerun the job

thaJeztah:
Could you squash your commits? Let me know if you need a hand/instructions on doing so

rchenzheng:
> Could you squash your commits? Let me know if you need a hand/instructions on doing so

I definitely can, thought it was automated at the end

thaJeztah:
We could use GitHub's "squash and merge" feature, but prefer to keep contributor's commits as they were authored by the contributor. GitHub's feature also removes GPG signing (which is used by some people), which we try to keep intact if possible.

Issue #2677, opened 2020-08-14
docker rm --force should return 0 (zero) exit-code if container doesn't exist


thaJeztah:
Relates to / noticed in https://github.com/moby/moby/pull/41228#issuecomment-673468993

When using `docker rm` / `docker container rm` with the `-f` / `--force` option, attempts to remove non-existing containers should print a warning, but should return a zero exit code ("successful").

Currently, a non-zero exit code is returned, marking the removal as "failed";

```
$ docker rm -fv 798c9471b695
Error: No such container: 798c9471b695
$ echo $?
1
```

The command should match the behavior of `rm` / `rm -f`, with the exception that a warning is printed (instead of silently ignored):

Running `rm` without `-f`, files/directories are removed, but the command returns a non-zero exit code:

```bash
touch some-file && rm no-such-file some-file; echo exit code: $?; ls -la
# rm: no-such-file: No such file or directory
# exit code: 1
# total 0
# drwxr-xr-x    2 sebastiaan  staff    64 Aug 14 12:15 .
# drwxr-xr-x  199 sebastiaan  staff  6368 Aug 14 12:13 ..

mkdir some-directory && rm -r no-such-directory some-directory; echo exit code: $?; ls -la
# rm: no-such-file: No such file or directory
# exit code: 1
# total 0
# drwxr-xr-x    2 sebastiaan  staff    64 Aug 14 12:15 .
# drwxr-xr-x  199 sebastiaan  staff  6368 Aug 14 12:13 ..
```

Running `rm` with `-f` silences output and returns a zero exit code:

```bash
touch some-file && rm -f no-such-file some-file; echo exit code: $?; ls -la
# exit code: 0
# total 0
# drwxr-xr-x    2 sebastiaan  staff    64 Aug 14 12:17 .
# drwxr-xr-x  199 sebastiaan  staff  6368 Aug 14 12:13 ..

mkdir some-directory && rm -rf no-such-directory some-directory; echo exit code: $?; ls -la
# exit code: 0
# total 0
# drwxr-xr-x    2 sebastiaan  staff    64 Aug 14 12:17 .
# drwxr-xr-x  199 sebastiaan  staff  6368 Aug 14 12:13 ..
```

Note that other reasons for a delete to fail should still result in a non-zero exit code, matching the behavior of `rm`. For instance, in the example below, the `rm` failed because directories can only be removed if the `-r` option is used;

```bash
[...]

PR #2678, opened 2020-08-14
Return zero exit-code when force-removing non-existing containers

thaJeztah:
fixes https://github.com/docker/cli/issues/2677

When using `docker rm` / `docker container rm` with the `-f` / `--force` option, attempts to remove non-existing containers should print a warning, but should return a zero exit code ("successful").

Currently, a non-zero exit code is returned, marking the removal as "failed";

	$ docker rm -fv 798c9471b695
	Error: No such container: 798c9471b695
	$ echo $?
	1

The command should match the behavior of `rm` / `rm -f`, with the exception that
a warning is printed (instead of silently ignored):

Running `rm` with `-f` silences output and returns a zero exit code:

    touch some-file && rm -f no-such-file some-file; echo exit code: $?; ls -la
    # exit code: 0
    # total 0
    # drwxr-xr-x    2 sebastiaan  staff    64 Aug 14 12:17 .
    # drwxr-xr-x  199 sebastiaan  staff  6368 Aug 14 12:13 ..

    mkdir some-directory && rm -rf no-such-directory some-directory; echo exit code: $?; ls -la
    # exit code: 0
    # total 0
    # drwxr-xr-x    2 sebastiaan  staff    64 Aug 14 12:17 .
    # drwxr-xr-x  199 sebastiaan  staff  6368 Aug 14 12:13 ..

Note that other reasons for a delete to fail should still result in a non-zero
exit code, matching the behavior of `rm`. For instance, in the example below,
the `rm` failed because directories can only be removed if the `-r` option is used;

    touch some-file && mkdir some-directory && rm -f some-directory no-such-file some-file; echo exit code: $?; ls -la
    # rm: some-directory: is a directory
    # exit code: 1
    # total 0
    # drwxr-xr-x    3 sebastiaan  staff    96 Aug 14 14:15 .
    # drwxr-xr-x  199 sebastiaan  staff  6368 Aug 14 12:13 ..
    # drwxr-xr-x    2 sebastiaan  staff    64 Aug 14 14:15 some-directory

This patch updates the `docker rm` / `docker container rm` command to not produce
an error when attempting to remove a missing containers, and instead only print
the error, but return a zero (0) exit code.

With this patch applied:

    docker create --name mycontainer busybox \
    && docker rm nosuchcontainer mycontainer; \
    echo exit code: $?; \
    docker ps -a --filter name=mycontainer
    # df23cc8573f00e97d6e948b48d9ea7d75ce3b4faaab4fe1d3458d3bfa451f39d
    # mycontainer
    # Error: No such container: nosuchcontainer
    # exit code: 0
    # CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES

**- What I did**

**- How I did it**

**- How to verify it**

```
Fix `docker rm --force` returning a non-zero exit code if one or more containers did not exist
```

### DIFF
--- a/cli/command/container/client_test.go
+++ b/cli/command/container/client_test.go
@@ -31,6 +31,7 @@ type fakeClient struct {
 	containerListFunc       func(types.ContainerListOptions) ([]types.Container, error)
 	containerExportFunc     func(string) (io.ReadCloser, error)
 	containerExecResizeFunc func(id string, options types.ResizeOptions) error
+	containerRemoveFunc     func(ctx context.Context, container string, options types.ContainerRemoveOptions) error
 	Version                 string
 }
 
@@ -78,6 +79,13 @@ func (f *fakeClient) ContainerCreate(
 		return f.createContainerFunc(config, hostConfig, networkingConfig, platform, containerName)
 	}
 	return container.ContainerCreateCreatedBody{}, nil
+}
+
+func (f *fakeClient) ContainerRemove(ctx context.Context, container string, options types.ContainerRemoveOptions) error {
+	if f.containerRemoveFunc != nil {
+		return f.containerRemoveFunc(ctx, container, options)
+	}
+	return nil
 }
 
 func (f *fakeClient) ImageCreate(ctx context.Context, parentReference string, options types.ImageCreateOptions) (io.ReadCloser, error) {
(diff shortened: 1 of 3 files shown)


thaJeztah:
@silvin-lubecki @cpuguy83 PTAL

thaJeztah:
> Probably we should do this in the API, though

hm, possibly; also feels a bit odd to have the API ignore a non-existing resource

PR #2683, opened 2020-08-19
Dockerfiles: update shellcheck v0.7.1

thaJeztah:
full diff: https://github.com/koalaman/shellcheck/compare/v0.7.0...v0.7.1

**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah:
Looks like some new rules are added. I guess we can ignore them;

```
In contrib/completion/bash/docker line 5530:
			$(__docker_to_extglob "$global_options_with_args") )
                        ^-- SC2254: Quote expansions in case patterns to match literally rather than as a glob.

For more information:
  https://www.shellcheck.net/wiki/SC2254 -- Quote expansions in case patterns...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

thaJeztah:
Actually, let me ping @albers to have a quick peek if I should update the code, or ignore the warnings

albers:
> ```
> In contrib/completion/bash/docker line 5530:
> 			$(__docker_to_extglob "$global_options_with_args") )
>                         ^-- SC2254: Quote expansions in case patterns to match literally rather than as a glob.

The purpose of `__docker_to_extglob` is to create an extglob. So quoting the result in order to prevent it from being treated as a glob is not a good idea. We need an exception here.

thaJeztah:
Thanks! I'll update the PR tomorrow and either put exceptions inline, or look if I can configure shellcheck to skip that check 👍

contrib/completion/bash/docker
```diff
@@ -688,7 +689,7 @@ __docker_pos_first_nonflag() {
 		(( counter++ ))
 	done
 
-	echo $counter
+	echo "$counter"
```
albers, line 692: Does shellcheck insist on this?

thaJeztah: Yes, it did (although a bit odd that it wasn't smart enough to detect that it's not needed 🙄;

```
In contrib/completion/bash/docker line 692:
	echo $counter
             ^------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
	echo "$counter"
```

albers: This creates the wrong impression that the output might contain spaces. But even if it did, quoting would not be necessary here.

That said, let's keep it qouted. It's not worth a shellcheck exception.

albers (approved):
LGTM

contrib/completion/bash/docker
thaJeztah: Yes, that was my train of thought as well; this looked less messy than an exception

PR #2663, opened 2020-07-30
Add capabilities support to stack/service commands

akerouanton:
closes #1940
closes #2199

**- What I did**

This PR supersedes #1940 and #2199. I removed the support of `privileged` option from the original PR as I believe privileged containers involve more than just having the complete set of capabilities (eg. no user namespace,  the unconfined AppArmor profile is used, etc...).

1. `docker stack deploy` now supports `cap_add` and `cap_drop` options ;
2. `docker service create|update` now support `--cap-add` and `--cap-drop` flags. For the `update` command and unlike most of other options, those flags don't have `-add` & `-rm` variants as this would make a poor UX (eg. `--cap-add-add` & `--cap-drop-rm`). Instead, `updateCapabilities` takes care of stripping caps provided to `--cap-add` from `CapabilityDrop` (and vice versa)  ;
3. `docker service inspect --pretty` displays the list of added caps and dropped caps ;

Closes moby/moby#25885 docker/cli#1940 docker/cli#2199 

**- How I did it**

1. The service converter transforming compose types into API types has been updated to add capabilities to `ServiceSpec`;
2. A new function with the logic described above and named `updateCapabilities()` has been introduced and is called by `updateService` ;
3. The pretty formatter has been updated to include added and dropped caps ;

**- How to verify it**

<details>
<summary><code>docker stack deploy</code></summary>

```
$ cat docker-compose.yaml
version: "3.7"
services:
  test:
    image: ollijanatuinen/capsh
    cap_add:
      - NET_ADMIN
    cap_drop:
      - CAP_MKNOD
$ build/docker stack deploy --compose-file docker-compose.yaml t1
$ docker inspect t1_test.1.m0jmdufuo93fotyyrhm00xx16
        "CapAdd": [
            "NET_ADMIN"
        ],
        "CapDrop": [
            "CAP_MKNOD"
        ],
```
</details>

<details>
<summary><code>docker service create</code></summary>

```
$ build/docker service create --name test --cap-add NET_ADMIN --cap-drop CAP_MKNOD ollijanatuinen/capsh
$ docker inspect test.1.li7llrgyp8v4mfvo5tvimhzxt
        "CapAdd": [
            "NET_ADMIN"
        ],
        "CapDrop": [
            "CAP_MKNOD"
        ],
```
</details>

<details>
<summary><code>docker service update</code></summary>

```
$ build/docker service update --cap-add CAP_MKNOD t1_test
$ docker inspect t1_test.1.92sd9omniksbenuv6wm17ft78
[...]

PR #2199, opened 2019-11-17
[WIP] Cli: Support --cap-add, --cap-drop and --privileged on services

olljanat:
**- What I did**
* Added --cap-add, --cap-drop and --privileged flags to `docker service create`
* Added --cap-add and --cap-drop to `docker service update`

Closes moby/moby#25885 moby/moby#24862 docker/swarmkit#1030

**- How I did it**
Included logic which converts -cap-add, --cap-drop and --privileged flags to exact capabilities list

**- How to verify it**
Create service based on default capabilities with minor adjustments:
```
docker service create --name test --cap-add "NET_ADMIN" --cap-drop "CAP_CHOWN" ollijanatuinen/capsh
```
Update capabilities for existing service:
```
docker service update --cap-drop "CAP_NET_RAW" --cap-add "SYS_NICE" test
```

Create service with all capabilities and drop some capabilities from it:
```
docker service create --name test2 --privileged ollijanatuinen/capsh
docker service update --cap-drop "CAP_AUDIT_READ" --cap-drop "CAP_BLOCK_SUSPEND" test2
```

**- A picture of a cute animal (not mandatory but encouraged)**
image

WIP because based on #1940 and it needs to be merged first.

PR #2687, opened 2020-08-25
[carry 2663] Add capabilities support to stack/service commands

thaJeztah:
carries https://github.com/docker/cli/pull/2663

closes #1940
closes #2199
closes #2663

Fixes moby/moby#25885

**- What I did**

This PR supersedes #1940 and #2199. I removed the support of `privileged` option from the original PR as I believe privileged containers involve more than just having the complete set of capabilities (eg. no user namespace,  the unconfined AppArmor profile is used, etc...).

1. `docker stack deploy` now supports `cap_add` and `cap_drop` options ;
2. `docker service create|update` now support `--cap-add` and `--cap-drop` flags. For the `update` command and unlike most of other options, those flags don't have `-add` & `-rm` variants as this would make a poor UX (eg. `--cap-add-add` & `--cap-drop-rm`). Instead, `updateCapabilities` takes care of stripping caps provided to `--cap-add` from `CapabilityDrop` (and vice versa)  ;
3. `docker service inspect --pretty` displays the list of added caps and dropped caps ;

**- How I did it**

1. The service converter transforming compose types into API types has been updated to add capabilities to `ServiceSpec`;
2. A new function with the logic described above and named `updateCapabilities()` has been introduced and is called by `updateService` ;
3. The pretty formatter has been updated to include added and dropped caps ;

**- How to verify it**

<details>
<summary><code>docker stack deploy</code></summary>

```
$ cat docker-compose.yaml
version: "3.7"
services:
  test:
    image: ollijanatuinen/capsh
    cap_add:
      - NET_ADMIN
    cap_drop:
      - CAP_MKNOD
$ build/docker stack deploy --compose-file docker-compose.yaml t1
$ docker inspect t1_test.1.m0jmdufuo93fotyyrhm00xx16
        "CapAdd": [
            "CAP_NET_ADMIN"
        ],
        "CapDrop": [
            "CAP_MKNOD"
        ],
```
</details>

<details>
<summary><code>docker service create</code></summary>

```
$ build/docker service create --name test --cap-add NET_ADMIN --cap-drop CAP_MKNOD ollijanatuinen/capsh
$ docker inspect test.1.li7llrgyp8v4mfvo5tvimhzxt
        "CapAdd": [
            "CAP_NET_ADMIN"
        ],
        "CapDrop": [
            "CAP_MKNOD"
        ],
```
</details>

<details>
<summary><code>docker service update</code></summary>

```
$ build/docker service update --cap-add CAP_MKNOD t1_test
$ docker inspect t1_test.1.92sd9omniksbenuv6wm17ft78
        "CapAdd": [
            "CAP_NET_ADMIN",
            "CAP_MKNOD"
        ],
        "CapDrop": null,

$ build/docker service update --cap-drop CAP_MKNOD --cap-drop NET_ADMIN t1_test
        "CapAdd": null,
        "CapDrop": [
            "CAP_MKNOD",
            "CAP_NET_ADMIN"
        ],
```
</details>

<details>
<summary><code>docker service inspect</code></summary>

```
$ build/docker service inspect t1_test
        "CapabilityAdd": [
            "CAP_NET_ADMIN"
        ],
        "CapabilityDrop": [
            "CAP_MKNOD"
        ]
$ build/docker service inspect --pretty t1_test
Capabilities:
 Add: CAP_NET_ADMIN
 Drop: CAP_MKNOD
```
</details>

Add `cap_add` and `cap_drop` support to `docker stack deploy` and `--cap-add` and `--cap-drop` flags to `docker service create|update`.

### DIFF
--- a/cli/command/service/update.go
+++ b/cli/command/service/update.go
@@ -505,7 +505,10 @@ func updateService(ctx context.Context, apiClient client.NetworkAPIClient, flags
 	}
 
 	updateString(flagStopSignal, &cspec.StopSignal)
-	updateCapabilities(flags, cspec)
+
+	if anyChanged(flags, flagCapAdd, flagCapDrop) {
+		updateCapabilities(flags, cspec)
+	}
 
 	return nil
 }
@@ -1351,40 +1354,71 @@ func updateCredSpecConfig(flags *pflag.FlagSet, containerSpec *swarm.ContainerSp
 	}
 }
 
+// updateCapabilities calculates the list of capabilities to "drop" and to "add"
+// after applying the capabilities passed through `--cap-add` and `--cap-drop`
+// to the existing list of added/dropped capabilities in the service spec.
+//
+// Adding capabilities takes precedence over "dropping" the same capability, so
+// if both `--cap-add` and `--cap-drop` are specifying the same capability, the
+// `--cap-drop` is ignored.
+//
+// Capabilities to "drop" are removed from the existing list of "added"
+// capabilities, and vice-versa (capabilities to "add" are removed from the existing
+// list of capabilities to "drop").
+//
+// Capabilities are normalized, sorted, and duplicates are removed to prevent
+// service tasks from being updated if no changes are made. If a list has the "ALL"
+// capability set, then any other capability is removed from that list.
 func updateCapabilities(flags *pflag.FlagSet, containerSpec *swarm.ContainerSpec) {
-	var addToRemove, dropToRemove map[string]struct{}
-	capAdd := containerSpec.CapabilityAdd
-	capDrop := containerSpec.CapabilityDrop
+	var (
+		toAdd, toDrop map[string]bool
 
-	// First add the capabilities passed to --cap-add to the list of requested caps
+		capDrop = opts.CapabilitiesMap(containerSpec.CapabilityDrop)
+		capAdd  = opts.CapabilitiesMap(containerSpec.CapabilityAdd)
+	)
 	if flags.Changed(flagCapAdd) {
-		caps := flags.Lookup(flagCapAdd).Value.(*opts.ListOpts).GetAll()
-		capAdd = append(capAdd, caps...)
-
-		dropToRemove = buildToRemoveSet(flags, flagCapAdd)
+		toAdd = opts.CapabilitiesMap(flags.Lookup(flagCapAdd).Value.(*opts.ListOpts).GetAll())
 	}
-
-	// And add the capabilities passed to --cap-drop to the list of dropped caps
 	if flags.Changed(flagCapDrop) {
-		caps := flags.Lookup(flagCapDrop).Value.(*opts.ListOpts).GetAll()
-		capDrop = append(capDrop, caps...)
-
-		addToRemove = buildToRemoveSet(flags, flagCapDrop)
+		toDrop = opts.CapabilitiesMap(flags.Lookup(flagCapDrop).Value.(*opts.ListOpts).GetAll())
 	}
 
-	// Then take care of removing caps passed to --cap-drop from the list of requested caps
-	containerSpec.CapabilityAdd = make([]string, 0, len(capAdd))
-	for _, cap := range capAdd {
-		if _, exists := addToRemove[cap]; !exists {
-			containerSpec.CapabilityAdd = append(containerSpec.CapabilityAdd, cap)
+	// First remove the capabilities to "drop" from the service's exiting
+	// list of capabilities to "add". If a capability is both added and dropped
+	// on update, then "adding" takes precedence.
+	for c := range toDrop {
+		if !toAdd[c] {
+			delete(capAdd, c)
+			capDrop[c] = true
 		}
 	}
 
-	// And remove the caps passed to --cap-add from the list of caps to drop
-	containerSpec.CapabilityDrop = make([]string, 0, len(capDrop))
-	for _, cap := range capDrop {
-		if _, exists := dropToRemove[cap]; !exists {
-			containerSpec.CapabilityDrop = append(containerSpec.CapabilityDrop, cap)
-		}
+	// And remove the capabilities we're "adding" from the service's existing
+	// list of capabilities to "drop".
+	//
+	// "Adding" capabilities takes precedence over "dropping" them, so if a
+	// capability is set both as "add" and "drop", remove the capability from
+	// the service's list of dropped capabilities (if present).
+	for c := range toAdd {
+		delete(capDrop, c)
+		capAdd[c] = true
 	}
+
+	// Now that the service's existing lists are updated, apply the new
+	// capabilities to add/drop to both lists. Sort the lists to prevent
+	// unneeded updates to service-tasks.
+	containerSpec.CapabilityDrop = capsList(capDrop)
+	containerSpec.CapabilityAdd = capsList(capAdd)
+}
+
+func capsList(caps map[string]bool) []string {
+	if caps[opts.AllCapabilities] {
+		return []string{opts.AllCapabilities}
+	}
+	var out []string
+	for c := range caps {
+		out = append(out, c)
+	}
+	sort.Strings(out)
+	return out
 }
--- a/cli/command/service/update.go
+++ b/cli/command/service/update.go
@@ -505,6 +505,7 @@ func updateService(ctx context.Context, apiClient client.NetworkAPIClient, flags
 	}
 
 	updateString(flagStopSignal, &cspec.StopSignal)
+	updateCapabilities(flags, cspec)
 
 	return nil
 }
@@ -1347,5 +1348,43 @@ func updateCredSpecConfig(flags *pflag.FlagSet, containerSpec *swarm.ContainerSp
 		}
 
 		containerSpec.Privileges.CredentialSpec = credSpec
+	}
+}
+
+func updateCapabilities(flags *pflag.FlagSet, containerSpec *swarm.ContainerSpec) {
+	var addToRemove, dropToRemove map[string]struct{}
+	capAdd := containerSpec.CapabilityAdd
+	capDrop := containerSpec.CapabilityDrop
+
+	// First add the capabilities passed to --cap-add to the list of requested caps
+	if flags.Changed(flagCapAdd) {
+		caps := flags.Lookup(flagCapAdd).Value.(*opts.ListOpts).GetAll()
+		capAdd = append(capAdd, caps...)
+
+		dropToRemove = buildToRemoveSet(flags, flagCapAdd)
+	}
+
+	// And add the capabilities passed to --cap-drop to the list of dropped caps
+	if flags.Changed(flagCapDrop) {
+		caps := flags.Lookup(flagCapDrop).Value.(*opts.ListOpts).GetAll()
+		capDrop = append(capDrop, caps...)
+
+		addToRemove = buildToRemoveSet(flags, flagCapDrop)
+	}
+
+	// Then take care of removing caps passed to --cap-drop from the list of requested caps
+	containerSpec.CapabilityAdd = make([]string, 0, len(capAdd))
+	for _, cap := range capAdd {
+		if _, exists := addToRemove[cap]; !exists {
+			containerSpec.CapabilityAdd = append(containerSpec.CapabilityAdd, cap)
+		}
+	}
+
+	// And remove the caps passed to --cap-add from the list of caps to drop
+	containerSpec.CapabilityDrop = make([]string, 0, len(capDrop))
+	for _, cap := range capDrop {
+		if _, exists := dropToRemove[cap]; !exists {
+			containerSpec.CapabilityDrop = append(containerSpec.CapabilityDrop, cap)
+		}
 	}
 }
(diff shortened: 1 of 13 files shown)


thaJeztah:
Ok, so I'm a bit in doubt what would be more logical; when updating a service that had previous cap-add/cap-drops, should we do:


### A. Update both "previous" and "new" state at once

Given the following service;


| CapDrop        | CapAdd        |
| -------------- | ------------- |
| CAP_SOME_CAP   |               |

When updating the service, and applying `--cap-add CAP_SOME_CAP`, the previously dropped capability is removed, and `CAP_SOME_CAP` is added:

| CapDrop        | CapAdd        |
| -------------- | ------------- |
|                | CAP_SOME_CAP  |

Downside of this is that there's no option to `reset` the cap-add/cap-drop through `docker service update` (note that this _would_ be possible through `docker stack up` as it's "declarative", and doesn't take the previous state into account).


### B. Update as a "tri-state" (for better words)

Given the following service;

| CapDrop        | CapAdd        |
| -------------- | ------------- |
| CAP_SOME_CAP   |               |

When updating the service, and applying `--cap-add CAP_SOME_CAP`, the previously dropped capability is removed:

| CapDrop        | CapAdd        |
| -------------- | ------------- |
|                |               |

When updating the service a second time, applying `--cap-add CAP_SOME_CAP`, capability is now added:

| CapDrop        | CapAdd        |
| -------------- | ------------- |
|                | CAP_SOME_CAP  |

So upside is that it's possible to "reset" previous capabilities without _adding_ new ones. Downside is that it requires multiple updates if that _is_ the desired result (given, probably a rare condition)

### C. Introduce a special "RESET" value

Given the following service;

| CapDrop        | CapAdd        |
| -------------- | ------------- |
| CAP_SOME_CAP   |               |

When updating the service, and applying `--cap-drop RESET`:

| CapDrop        | CapAdd        |
| -------------- | ------------- |
|                |               |

When updating the service, and applying `--cap-drop RESET`, combined with `--cap-add CAP_SOME_CAP` and `--cap-drop CAP_SOME_OTHER_CAP`:

| CapDrop        | CapAdd        |
| -------------- | ------------- |
| CAP_FOO_CAP    | CAP_SOME_CAP  |


### D. Combine B + C

Apply the diff to the existing list of capabilities first, but also allow using `RESET` to "start from scratch"

thaJeztah:
Need to look at some of the handling of `--cap-add=ALL` / `--cap-drop=ALL`;

- if both are specified, `--cap-add` takes precedence, so it's the equivalent of `--cap-add=ALL`
- if `--cap-drop=ALL` is set, and existing service's `CapAdd` has `ALL`, it should be removed before applying other capabilities
- if the existing service does not have  `CapAdd` `ALL`, then add it to `CapDrop`, omit other `--cap-drop`'s, and apply any `--cap-add`'s

thaJeztah:
As to my comment above, I added two commits that implement B and C

contrib/completion/bash/docker
```diff
@@ -874,6 +875,7 @@ __docker_complete_capabilities_droppable() {
 		MKNOD
 		NET_BIND_SERVICE
 		NET_RAW
+		RESET
 		SETFCAP
```
thaJeztah, line 879: I noticed that our completion scripts use the capabilities without the `CAP_` prefix; I'm on the fence what we should use as "normalised" format: with, or without the `CAP_`. From a UX perspective, possibly _without_ is more friendly (less to type, and less characters to type to use auto-completion)

@tiborvass @silvin-lubecki WDYT?

thaJeztah: Perhaps it doesn't make a big difference (as the values entered by the user will be normalised anyway, so the only difference will be in what's shown in `docker service inspect`. Looks like that's currently consistent with (e.g.) what's documented for `docker plugin` capabilities.

cli/command/service/update.go
```diff
@@ -1351,3 +1355,131 @@ func updateCredSpecConfig(flags *pflag.FlagSet, containerSpec *swarm.ContainerSp
...
+// Adding capabilities takes precedence over "dropping" the same capability, so
+// if both `--cap-add` and `--cap-drop` are specifying the same capability, the
+// `--cap-drop` is ignored.
+//
+// Capabilities to "drop" are removed from the existing list of "added"
+// capabilities, and vice-versa (capabilities to "add" are removed from the existing
+// list of capabilities to "drop").
+//
+// Capabilities are normalized, sorted, and duplicates are removed to prevent
+// service tasks from being updated if no changes are made. If the a list has
```
silvin-lubecki, line 1372: typo: the/a list

cli/command/service/update_test.go
```diff
@@ -1310,3 +1310,201 @@ func TestUpdateCredSpec(t *testing.T) {
...
+			},
+			expectedAdd: []string{"ALL"},
+		},
+		{
+			// Note that this won't be run as updateCapabilities is gated by anyChanged(flags, flagCapAdd, flagCapDrop)
+			name: "Normalize cap-drop only",
+			spec: &swarm.ContainerSpec{
+				CapabilityDrop: []string{"ALL", "CAP_MOUNT", "CAP_NET_ADMIN"},
+			},
+			expectedDrop: []string{"ALL"},
```
silvin-lubecki, line 1372: nit: should we also test `ALL` in `cap-add` AND `cap-drop`?

cli/command/service/update_test.go
```diff
@@ -1310,3 +1310,201 @@ func TestUpdateCredSpec(t *testing.T) {
...
+			spec:         &swarm.ContainerSpec{},
+			expectedDrop: []string{"CAP_NET_ADMIN"},
+		},
+		{
+			name:     "Add a previously dropped cap",
+			flagAdd:  []string{"CAP_NET_ADMIN"},
+			flagDrop: []string{},
+			spec: &swarm.ContainerSpec{
+				CapabilityDrop: []string{"CAP_NET_ADMIN"},
+			},
```
silvin-lubecki, line 1394: nit: maybe add explicit empty `expectedDrop`?

cli/command/service/update_test.go
```diff
@@ -1310,3 +1310,201 @@ func TestUpdateCredSpec(t *testing.T) {
...
+			flagAdd:  []string{"RESET"},
+			flagDrop: []string{"RESET"},
+			spec: &swarm.ContainerSpec{
+				CapabilityAdd:  []string{"CAP_AAA", "CAP_BBB", "CAP_CCC", "CAP_DDD"},
+				CapabilityDrop: []string{"CAP_WWW", "CAP_XXX", "CAP_YYY", "CAP_ZZZ"},
+			},
+		},
+		{
+			name:     "Reset capabilities, and update after",
+			flagAdd:  []string{"RESET", "CAP_ADD_ONE", "CAP_FOO"},
```
silvin-lubecki, line 1483: nit: test also with "RESET", "ALL" ?

contrib/completion/bash/docker
silvin-lubecki: I'm totally fine with values without the `CAP_` 👍

opts/capabilities_test.go
```diff
@@ -0,0 +1,119 @@
...
+		// white space handling. strictly, these could be considered "invalid",
+		// but are a likely situation, so handling these for now.
+		{in: "  ALL  ", out: "ALL"},
+		{in: "  FOO  ", out: "CAP_FOO"},
+		{in: "  CAP_FOO  ", out: "CAP_FOO"},
+		{in: " 	 ALL 	 ", out: "ALL"},
+		{in: " 	 FOO 	 ", out: "CAP_FOO"},
+		{in: " 	 CAP_FOO 	 ", out: "CAP_FOO"},
+
+		// weird values: no validation takes place currently, so these
```
silvin-lubecki, line 31: question: shouldn't the validation take place server side anyway?

opts/capabilities.go
```diff
@@ -0,0 +1,89 @@
...
+	"strings"
+)
+
+const (
+	// AllCapabilities is a special value to add or drop all capabilities
+	AllCapabilities = "ALL"
+
+	// ResetCapabilities is a special value to reset capabilities when updating.
+	// This value should only be used when updating, not used on "create".
+	ResetCapabilities = "RESET"
```
silvin-lubecki, line 14: question: Should RESET be explicitly forbidden for service create?

silvin-lubecki (approved):
LGTM, even if I'm a bit on the fence with the UX (`RESET` looks weird to me, but I can't find a better way), so let's move with that 👍

contrib/completion/bash/docker
albers: When completing signals, we offer both SIGXXX and XXX.
So for consistency, we could switch to supporting both variants here as well.
I'll create a PR for that.

albers:
I tried to test this (against current master of moby/moby), but I got

```bash
root@6c8577822fe3:/go/src/github.com/docker/docker# docker service create --name test --cap-add NET_ADMIN --cap-drop CAP_MKNOD ollijanatuinen/capsh
"--cap-add" requires API version 1.41, but the Docker daemon API version is 1.40
"--cap-drop" requires API version 1.41, but the Docker daemon API version is 1.40
```
I guess that I'm not up to date with the changes in docker development.
Can someone please give me a hint on how to spin up a current dev environment now?

contrib/completion/bash/docker
```diff
@@ -874,6 +875,7 @@ __docker_complete_capabilities_droppable() {
 		MKNOD
 		NET_BIND_SERVICE
 		NET_RAW
+		RESET
```
albers, line 878: Bash completion has two distinct sets of capabilities: one for capabilities that are enabled by default and thus can be dropped, and one for those that are not enabled by default and therefore can be added.
Any given capablility should not be contained in both lists.

contrib/completion/bash/docker
```diff
@@ -3670,6 +3672,8 @@ _docker_service_update() {
 # and `docker service update`
 _docker_service_update_and_create() {
 	local options_with_args="
+		--cap-add
+		--cap-drop
```
albers, line 3676: This enables completion of `--cap-add` and `--cap-drop`.
To add completion for the specific capabilities, you need to add [this](https://github.com/docker/cli/blob/master/contrib/completion/bash/docker#L2020-L2027) to the case statement that starts [here].(https://github.com/docker/cli/blob/10b529e2aed166263f2ae5d7d9e38524b336f6e3/contrib/completion/bash/docker#L3834)

albers (changes requested):

contrib/completion/bash/docker
albers: Created #2706

opts/capabilities_test.go
thaJeztah: Yes, validation should be done server side, so strictly speaking, I could remove these values from the test; I left them in mostly to make it explicit that we don't care about these invalid values (at least, not currently); that said, I think it'd be okay to handle _some_ early cleaning up in the CLI (trimming white-space should be mostly "ok" to do)

contrib/completion/bash/docker
thaJeztah: Oh, thanks! Yes, missed that: I'll add those lines

cli/command/service/update_test.go
thaJeztah: Yeah, guess it makes it slightly more understandable; I'll add explicit `expectedDrop` / `expectedAdd` lines

opts/capabilities.go
thaJeztah: I was thinking about it, but decided to keep it simple for now

thaJeztah:
> even if I'm a bit on the fence with the UX (RESET looks weird to me, but I can't find a better way), so let's move with that 👍

I can move the last commit to a separate PR; I don't think it's _critical_ to have it, and we can add it after that, so if it requires more discussion, we can do so separately

opts/capabilities_test.go
silvin-lubecki: Sure, makes sense 👍

opts/capabilities.go
silvin-lubecki: Fine by me 👍

contrib/completion/bash/docker
thaJeztah: Hmm.. isn't this the same as the `ALL` capability (which is also in both lists?)

thaJeztah:
@silvin-lubecki @albers updated this one, and moved the `RESET` commit to https://github.com/docker/cli/pull/2709

silvin-lubecki (approved):
LGTM

thaJeztah:
@tiborvass @cpuguy83 PTAL

contrib/completion/bash/docker
albers: Ah, yes. I didn't realize this capablity is a special one.

albers (approved):

cpuguy83 (approved):
LGTM

kaysond:
I've been waiting for this! Is it going to be in 19.03 or a newer version? Any idea when it will get released?

thaJeztah:
this will be in the upcoming 20.xx release

robertnisipeanu:
Isn't this in the master branch? I tried installing docker-ce docker-ce-cli and containerd.io from the nightly build, however I still get "Ignoring unsupported options: cap_add"

thaJeztah:
I think there's an issue with the nightly builds currently not being updated; this should be in the docker v20.10 release candidates in the testing channel (GA should be released soon)

liangyuanpeng:
Look forward it!

georgmay:
I'm getting the same Ignoring unsupported options: cap_add on 20.10

trajano:
It does not work on Docker Desktop for Windows 3.0

PR #2694, opened 2020-08-28
Replace vbom.ml/util with fvbommel/sortorder

thaJeztah:
The vanity domain is down, and the project has moved to a new location.

vendor check started failing because of this:

    Collecting initial packages
    Download dependencies
    unrecognized import path "vbom.ml/util" (https fetch: Get https://vbom.ml/util?go-get=1: dial tcp: lookup vbom.ml on 169.254.169.254:53: no such host)


thaJeztah:
Argh, looks like there's a bug in vndr; for some reason it still looks for the original URL

PR #2704, opened 2020-09-03
Reference docs: Fix CMD --ignored-param1 example

StefanScherer:
Signed-off-by: Stefan Scherer <stefan.scherer@docker.com>

addresses docker/docker.github.io#11319 

**- What I did**

I fixed the issues found in our docs in issue docker/docker.github.io#11319 by adding quotes. I also updated the sample output.

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/builder.md
+++ b/docs/reference/builder.md
@@ -1632,7 +1632,7 @@ If you forget to add `exec` to the beginning of your `ENTRYPOINT`:
 ```dockerfile
 FROM ubuntu
 ENTRYPOINT top -b
-CMD --ignored-param1
+CMD -- --ignored-param1
 ```
 
 You can then run it (giving it a name for the next step):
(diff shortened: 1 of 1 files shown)


docs/reference/builder.md
```diff
@@ -1584,20 +1584,22 @@ If you forget to add `exec` to the beginning of your `ENTRYPOINT`:
 ```dockerfile
 FROM ubuntu
 ENTRYPOINT top -b
-CMD --ignored-param1
+CMD "--ignored-param1"
```
thaJeztah, line 1587: Wondering if this is just "tricking" the parser to consider it valid (by adding the quotes), and if instead it should be

```suggestion
CMD -- --ignored-param1
```

(using `--` to terminate processing as option for `CMD`)

However, in both cases, this likely won't give the expected result, as `CMD` will still become `["/bin/sh", "-c", ...]`, so it's not really ignored

```dockerfile
FROM docker
ENTRYPOINT docker run
CMD "--help"
```

```bash
docker inspect --format='{{json .Config.Entrypoint }} {{json .Config.Cmd }}' foo
["/bin/sh","-c","docker run"] ["/bin/sh","-c","\"--help\""]
```

```dockerfile
FROM docker
ENTRYPOINT docker run
CMD -- --help
```

```bash
docker inspect --format='{{json .Config.Entrypoint }} {{json .Config.Cmd }}' foo
["/bin/sh","-c","docker run"] ["/bin/sh","-c","--help"]
```


I'm trying to understand what the purpose of the example is, because the `--ignored-param` (unless there's a binary with that name) would never give anything useful 🤔

docs/reference/builder.md
```diff
@@ -1584,20 +1584,22 @@ If you forget to add `exec` to the beginning of your `ENTRYPOINT`:
...
 
 You can then run it (giving it a name for the next step):
 
 ```bash
 $ docker run -it --name test top --ignored-param2
+top - 13:58:24 up 17 min,  0 users,  load average: 0.00, 0.00, 0.00
+Tasks:   2 total,   1 running,   1 sleeping,   0 stopped,   0 zombie
+%Cpu(s): 16.7 us, 33.3 sy,  0.0 ni, 50.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
+MiB Mem :   1990.8 total,   1354.6 free,    231.4 used,    404.7 buff/cache
+MiB Swap:   1024.0 total,   1024.0 free,      0.0 used.   1639.8 avail Mem
```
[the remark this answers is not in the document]
StefanScherer: I assume the ubuntu image has changed over the time. It's running the top command inside the container (I'm running the example on my MBP).

docs/reference/builder.md
StefanScherer: Thanks, the `CMD -- --xxx` looks much better.

Well the example shows the edge cases and it seems both ENTRYPOINT and CMD without using an array doesn't build a useful command, it always has a second `/bin/sh` 🤷

silvin-lubecki (approved):
LGTM

devdreamsolution (approved):

thaJeztah:
Did a quick rebase to kick CI again; looks like we have a flaky test (or a bug; I recall some users reporting that happened sometimes); posting here to find it back 😅 

```
--- FAIL: TestInitTtySizeErrors (0.10s)
    tty_test.go:29: assertion failed: 
        --- expectedError
        +++ →
        @@ -1,2 +1 @@
        -failed to resize tty, using default size
```

thaJeztah (approved):
LGTM

Issue #1281, opened 2018-08-09
Add visual feedback on how far copying files to and from docker container is going.


Crare:
**Description**

I'm copying files to and from docker container to edit them locally.
I would like to see how long it takes, or how many files is there to copy still.
Maybe an estimate how long it takes and how long it has been running copy command.
These could be extra command options you could choose to show the progress of copying files.
It would be useful when you're copying a lot of files.
Expecially it would seem better than what it now is. Currently it looks like the whole console is jammed up when copying because there's no visual feedback of how the copying is going.

**Steps to reproduce the issue:**
1. copy file to docker container
`docker cp ./www container_name:/var/`

2. copy files from docker container
`docker cp container_name:/var/www ./`

**Describe the results you received:**
No visual feedback how far is the copying going. It looks like the console is frozen.

**Describe the results you expected:**
Some indication of how far the progress is going on with copying files.

**Additional information you deem important (e.g. issue happens only occasionally):**
visual clue that console is not frozen, that it's still copying files.

silvin-lubecki:
Hi @Crare thank you for filling this issue! You're right it seems to be a good enhancement. PRs are welcomed 😉

Crare:
You're welcome! @silvin-lubecki I actually would find this useful elsewhere in the CLI too. Like when making a commit after a while it takes a bit longer and that looks like it's frozen too. It's a minor enhancement, but would look better and I wouldn't be so confused every-time it happens. (thinking it's frozen when it's actually not. I've sometimes restarted the whole docker because of this.)

adeniyistephen:
@Crare @silvin-lubecki Hi, I would love to work on this, Please provide pointers and input on how to work on it. Thanks

thaJeztah:
I think this is being worked on in https://github.com/docker/cli/pull/2708

PR #2708, opened 2020-09-04
Add spinner loading to docker cp command

### LINKED ISSUE #2564: Feature Proposal: Copying activity indicator for docker cp command
When running this command show an activity indicator.

`docker cp`

When running it could some sort of spinner like this.

https://www.shellscript.sh/tips/spinner/spinner.gif

If progress doesn't start or fails midstream the spinner would stop.

maximillianfx:
**- What I did**

When a user executes `docker cp` command, a `Copying` message is showed on terminal and a spinner is printed out together, with a small animation.

**- How I did it**

Add animation using a channel and an array of characters to simulate a spinner in cp.go.

**- How to verify it**

Run the environment container with `make -f docker.Makefile shell`, then run the command `./build/docker-linux-amd64 cp file container:path` to see.

Addition of animated spinner when copying files using the docker cp command

image

closes #2564 
closes https://github.com/docker/cli/issues/1281

### DIFF
--- a/cli/command/container/cp.go
+++ b/cli/command/container/cp.go
@@ -2,6 +2,7 @@ package container
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -12,6 +13,8 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/archive"
 	"github.com/docker/docker/pkg/system"
+	units "github.com/docker/go-units"
+	"github.com/morikuni/aec"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
@@ -21,6 +24,7 @@ type copyOptions struct {
 	destination string
 	followLink  bool
 	copyUIDGID  bool
+	quiet       bool
 }
 
 type copyDirection int
@@ -34,9 +38,36 @@ const (
 type cpConfig struct {
 	followLink bool
 	copyUIDGID bool
+	quiet      bool
 	sourcePath string
 	destPath   string
 	container  string
+}
+
+// copyProgressPrinter wraps io.ReadCloser to print progress information when
+// copying files to/from a container.
+type copyProgressPrinter struct {
+	io.ReadCloser
+	toContainer bool
+	total       *float64
+	writer      io.Writer
+}
+
+func (pt *copyProgressPrinter) Read(p []byte) (int, error) {
+	n, err := pt.ReadCloser.Read(p)
+	*pt.total += float64(n)
+
+	if err == nil {
+		fmt.Fprint(pt.writer, aec.Restore)
+		fmt.Fprint(pt.writer, aec.EraseLine(aec.EraseModes.All))
+		if pt.toContainer {
+			fmt.Fprintln(pt.writer, "Copying to container - "+units.HumanSize(*pt.total))
+		} else {
+			fmt.Fprintln(pt.writer, "Copying from container - "+units.HumanSize(*pt.total))
+		}
+	}
+
+	return n, err
 }
 
 // NewCopyCommand creates a new `docker cp` command
@@ -64,6 +95,10 @@ func NewCopyCommand(dockerCli command.Cli) *cobra.Command {
 			}
 			opts.source = args[0]
 			opts.destination = args[1]
+			if !cmd.Flag("quiet").Changed {
+				// User did not specify "quiet" flag; suppress output if no terminal is attached
+				opts.quiet = !dockerCli.Out().IsTerminal()
+			}
 			return runCopy(dockerCli, opts)
 		},
 	}
@@ -71,6 +106,7 @@ func NewCopyCommand(dockerCli command.Cli) *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVarP(&opts.followLink, "follow-link", "L", false, "Always follow symbol link in SRC_PATH")
 	flags.BoolVarP(&opts.copyUIDGID, "archive", "a", false, "Archive mode (copy all uid/gid information)")
+	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress progress output during copy. Progress output is automatically suppressed if no terminal is attached")
 	return cmd
 }
 
@@ -81,6 +117,7 @@ func runCopy(dockerCli command.Cli, opts copyOptions) error {
 	copyConfig := cpConfig{
 		followLink: opts.followLink,
 		copyUIDGID: opts.copyUIDGID,
+		quiet:      opts.quiet,
 		sourcePath: srcPath,
 		destPath:   destPath,
 	}
@@ -171,12 +208,34 @@ func copyFromContainer(ctx context.Context, dockerCli command.Cli, copyConfig cp
 		RebaseName: rebaseName,
 	}
 
+	var copiedSize float64
+	if !copyConfig.quiet {
+		content = &copyProgressPrinter{
+			ReadCloser:  content,
+			toContainer: false,
+			writer:      dockerCli.Err(),
+			total:       &copiedSize,
+		}
+	}
+
 	preArchive := content
 	if len(srcInfo.RebaseName) != 0 {
 		_, srcBase := archive.SplitPathDirEntry(srcInfo.Path)
 		preArchive = archive.RebaseArchiveEntries(content, srcBase, srcInfo.RebaseName)
 	}
-	return archive.CopyTo(preArchive, srcInfo, dstPath)
+
+	if copyConfig.quiet {
+		return archive.CopyTo(preArchive, srcInfo, dstPath)
+	}
+
+	fmt.Fprint(dockerCli.Err(), aec.Save)
+	fmt.Fprintln(dockerCli.Err(), "Preparing to copy...")
+	res := archive.CopyTo(preArchive, srcInfo, dstPath)
+	fmt.Fprint(dockerCli.Err(), aec.Restore)
+	fmt.Fprint(dockerCli.Err(), aec.EraseLine(aec.EraseModes.All))
+	fmt.Fprintln(dockerCli.Err(), "Successfully copied", units.HumanSize(copiedSize), "to", dstPath)
+
+	return res
 }
 
 // In order to get the copy behavior right, we need to know information
@@ -229,8 +288,9 @@ func copyToContainer(ctx context.Context, dockerCli command.Cli, copyConfig cpCo
 	}
 
 	var (
-		content         io.Reader
+		content         io.ReadCloser
 		resolvedDstPath string
+		copiedSize      float64
 	)
 
 	if srcPath == "-" {
@@ -272,13 +332,33 @@ func copyToContainer(ctx context.Context, dockerCli command.Cli, copyConfig cpCo
 
 		resolvedDstPath = dstDir
 		content = preparedArchive
+		if !copyConfig.quiet {
+			content = &copyProgressPrinter{
+				ReadCloser:  content,
+				toContainer: true,
+				writer:      dockerCli.Err(),
+				total:       &copiedSize,
+			}
+		}
 	}
 
 	options := types.CopyToContainerOptions{
 		AllowOverwriteDirWithFile: false,
 		CopyUIDGID:                copyConfig.copyUIDGID,
 	}
-	return client.CopyToContainer(ctx, copyConfig.container, resolvedDstPath, content, options)
+
+	if copyConfig.quiet {
+		return client.CopyToContainer(ctx, copyConfig.container, resolvedDstPath, content, options)
+	}
+
+	fmt.Fprint(dockerCli.Err(), aec.Save)
+	fmt.Fprintln(dockerCli.Err(), "Preparing to copy...")
+	res := client.CopyToContainer(ctx, copyConfig.container, resolvedDstPath, content, options)
+	fmt.Fprint(dockerCli.Err(), aec.Restore)
+	fmt.Fprint(dockerCli.Err(), aec.EraseLine(aec.EraseModes.All))
+	fmt.Fprintln(dockerCli.Err(), "Successfully copied", units.HumanSize(copiedSize), "to", copyConfig.container+":"+dstInfo.Path)
+
+	return res
 }
 
 // We use `:` as a delimiter between CONTAINER and PATH, but `:` could also be
--- a/cli/command/container/cp_test.go
+++ b/cli/command/container/cp_test.go
@@ -77,7 +77,7 @@ func TestRunCopyFromContainerToFilesystem(t *testing.T) {
 			return readCloser, types.ContainerPathStat{}, err
 		},
 	}
-	options := copyOptions{source: "container:/path", destination: destDir.Path()}
+	options := copyOptions{source: "container:/path", destination: destDir.Path(), quiet: true}
 	cli := test.NewFakeCli(fakeClient)
 	err := runCopy(cli, options)
 	assert.NilError(t, err)


silvin-lubecki:
Hello @maximillianfx, thank you for opening this PR, but can I ask what is the rational behind adding a spinner to this command? I guess the issue is that you don't really know what's going on and the command looks blocked (like if you cp a large file to a remote engine)?
Maybe then we shouldn't add a spinner (I don't think there's one yet in the CLI UX, so I don't think we should introduce one here), but instead have a counter of the bytes sent or received, like we do when we send the context to `docker build` or when we pull an image?

maximillianfx:
Hi @silvin-lubecki, how are you? Thanks for the quick help! Yeah, the reason for the spinner is to show to user a little animation/progression about the command, but I understand that this have some UI restrictions.

Your idea about a progress bar looks great and achieve the same (or even better) goal. Can I adjust this PR to this new approach?

silvin-lubecki:
cc @thaJeztah any thoughts?

thaJeztah:
Chatted with @silvin-lubecki and I agree that printing the amount of data that we copied would be more informative than just a spinner.

We weren't sure if it's possible to know up-front how much data _will_ be copied (in case you're copying a directory structure, the API will probably not return the total size?), so if we don't know the total, a progress-bar won't make much sense; in that case, only printing the amount to data that was copied (`copied xx kB` ... `copied xx mB`) would work

maximillianfx:
Yes, I understand, the directory scenario shows exactly the problem when displaying the amount of data that will be copied. Therefore, displaying a simple message after the copying process seems more effective. I will make these changes to this PR and we can return to the discussion soon.

maximillianfx:
Hey guys. I was working in this item yesterday, and I faced the `content` variable inside the `copyToContainer` and `copyFromContainer` functions. Getting the size from `content` from `copyToContainer` function returns more bytes than the real file size. Is this behavior right?

```golang
func getCopySize(data io.ReadCloser) (int, io.Reader) {
	content, err := ioutil.ReadAll(data.Body)
	if err != nil {
		fmt.Println(err)
	}
	
    return len(content), bytes.NewReader(content)
}
```

In my tests, the file size is 6 bytes, but `len(content)` returns 2048.

maximillianfx:
I made a mistake, the file is transformed to a Tar file. So, I'll made the size evaluation before this transformation.

maximillianfx:
Alright @silvin-lubecki @thaJeztah ! Can you run these changes and give me some feedback?

silvin-lubecki:
Hello @maximillianfx, thanks for the update! But I think there's an issue with your last rebase, the PR now includes all the last commits from master, so it's pretty hard to follow which are your commits/modifications 😅 Could you rebase properly your commits? 🙏

maximillianfx:
Yes, I'll update this.

maximillianfx:
@silvin-lubecki the current rebase is wrong? Because, I made the rebase to put my commits in front of the last commits of master, this is wrong? Can you help with the correct rebase?

silvin-lubecki:
Ok I reviewed the code using https://github.com/docker/cli/compare/master...maximillianfx:2564-add-spinner-loading
If I understood it correctly, it waits for the `cp` to be done, and then prints the size ? But it means that during the `cp` you won't see anything as it was before? I think we should have something more dynamic, updating in real time the content size transferred. WDYT?

maximillianfx:
Thanks for reviewing. I don't know, because the `cp` command generally takes some ms or seconds to complete, different from `docker pull`, where the download speed interfere in the progress (some seconds up to minutes). Could be a better experience. I'll investigate if this change will be affect more files than cp.go and the complexity to implement.

silvin-lubecki:
I guess that if you are targeting a remote daemon the `cp` will take more time 🤔

maximillianfx:
Yes, this scenario could occur. I'll investigate.

maximillianfx:
Hi @silvin-lubecki how are you? The behavior of these dynamically copying should be implemented on the engine project itself, WDYT? Because, looking into the `pull` command, the animation for the download is implemented there. In the `cli` project could be a more generic implementation, I think.

albers:
Removing myself as a reviewer because completion is not touched by this PR.

silvin-lubecki:
Hello @maximillianfx, I think that if you replace the `io.Copy` https://github.com/docker/cli/blob/master/cli/command/container/cp.go#L163 here with a "dynamic" printer, it should be enough 🤔 . WDYT?

maximillianfx:
Could be a solution, but this will affect only the copy from container and not to container, isn't it?

maximillianfx:
@silvin-lubecki did you manage to check my last question?

silvin-lubecki:
Sorry @maximillianfx I missed your question, thank you for the ping 😅 
I think we can do it the other way too https://github.com/docker/cli/blob/master/cli/command/container/cp.go#L232
`Content` is an `io.Reader`, so we could implement the interface, counting how much data was transferred here. WDYT?

maximillianfx:
Yeah, this looks great @silvin-lubecki. Can you suggest me some lecture to achieve this? I'm new into Golang  😅

maximillianfx:
I have made some changes. Added a little progress bar to indicate the copy process. Both copies, from and into container, are working. I have made this because only displaying the amount of bytes transfered could be unreadable 😕. WDYT @silvin-lubecki ?

cli-1602871556571

silvin-lubecki:
@maximillianfx Yes I like it! Can you use [units.HumanSize](https://github.com/docker/go-units/blob/master/size.go#L66) to print the size?

maximillianfx:
Hi @silvin-lubecki, thanks. Yeah, I'll do this soon 😄.

cpuguy83:
Thanks for working on this! 


Late comer here, but just some thoughts on the UI:

- It would be nice if this was more consistent with pull

Maybe something like:

```
<container>:/foo.txt: Copying [====>  ] 10KB/15KB
```

I'm not sure if the file reference there should show the remote, local... or both?

Then at the end a summary would be nice to have in the same vein as `docker pull`.

cpuguy83:
Let's also make sure to only show this stuff when stdout is a terminal.

maximillianfx:
Good points @cpuguy83 , in the next changes I'll follow your suggestions.

maximillianfx:
Hey guys @cpuguy83 @silvin-lubecki , how are you? Here are the UI updates 😄 
cli-1604098030366

Can you give me some feedback?
Thanks!

cli/command/container/cp.go
```diff
@@ -171,12 +256,46 @@ func copyFromContainer(ctx context.Context, dockerCli command.Cli, copyConfig cp
...
+	}
+
+	fmt.Println("Copying " + fileOrFolderName + " (" + units.HumanSize(float64(copySize)) + ") into host:" + copyPath + "...")
+	fmt.Print("\033[s")
+	res := archive.CopyTo(preArchive, srcInfo, dstPath)
+
+	fmt.Println()
+	fmt.Println("Status: " + fileOrFolderName + " copied to host at " + copyPath)
+
+	/*if dstPath[len(dstPath)-1] == '.' {
```
silvin-lubecki, line 283: nit: can we clean the commented code? 🙏

silvin-lubecki:
Hello @maximillianfx thanks for this hard work 🚀 👍 
I really like how it evolved.
I played with it and have few questions about the UX:
```
$ docker cp payload mycontainer:/tmp
Copying payload (79.15MB) into mycontainer:/tmp...
mycontainer:/tmp - Copying [==========] 79.15MB/79.15MB
Status: payload copied to container mycontainer at /tmp

$ docker cp mycontainer:/tmp/payload .
Copying payload (79.15MB) into host:/Users/silvin/dev/go/src/github.com/docker/cli/....
host:/Users/silvin/dev/go/src/github.com/docker/cli/. - Copying [==========] 79.15MB/79.15MB
Status: payload copied to host at /Users/silvin/dev/go/src/github.com/docker/cli/.
```
* I think we have some duplicated info here. The user already defined everything in the command, I guess we shouldn't repeat it and keep the messages simple, like `successfully copied`.
* Can we avoid `host` and use from/to container instead? It may be misleading what `host` is...

Here is an example of `docker service create`
```
$ docker service create --name foo nginx:alpine
dzbthv3lk2ndjbs718sf6omre
1/1: preparing [=================================>                 ]
...
dzbthv3lk2ndjbs718sf6omre
verify: Waiting 5 seconds to verify that tasks are stable...
...
dzbthv3lk2ndjbs718sf6omre
verify: Service converged
```
* What about updating the message too (preparing/copying/finishing), something like this:
```
$ docker cp mycontainer:/tmp/payload .
preparing copy       [                                                   ]
...
copy from container  [==============================>                    ] 59.15MB/79.15MB
...
copy from container  [==================================================>] 79.15MB/79.15MB
successfully copied 79.15MB to /Users/silvin/dev/go/src/github.com/docker/cli/.
```

Also I see that the code is not yet ready to be reviewed, let me know as soon as you want me to take a look 👍

maximillianfx:
Hi @silvin-lubecki, I have made a little mistake with the comented code and I'll remove it.
About the UX, is too much information (duplicated) really. I'll use your suggestions and see the output of other docker commands to enhance this.

Thanks for the opinions!

thaJeztah:
Didn't check if it's included in this PR yet, but we may need a `-q / --quiet` option as well, to facilitate CI setups where a terminal is present, but logging the progress-bar only would generate "noise"

maximillianfx:
This is not included yet @thaJeztah, good point, I'll cover this too.

cli/command/container/cp.go
maximillianfx: Done 😄

cpuguy83:
I like it @silvin-lubecki 
Looking at the demo with the filename in the progress, it definitely gets long and unwieldy quickly.

maximillianfx:
Hey guys @silvin-lubecki @thaJeztah.
I made the changes in the UI following the @silvin-lubecki suggestions and the `-q` option suggestion from @thaJeztah .
@silvin-lubecki can you review?

cli-1604691551859

cpuguy83:
Seems like this has some extra commits?

maximillianfx:
Yeah @cpuguy83, I have made a wrong rebase here. I'll try to fix it.

maximillianfx:
@cpuguy83 Done, fixed the commits. But now a error in the `cross` step occur. Can you help me?

silvin-lubecki:
It looks like everything is green now 👍

cli/command/container/cp.go
```diff
@@ -34,11 +39,91 @@ const (
...
 	followLink bool
 	copyUIDGID bool
+	quiet      bool
 	sourcePath string
 	destPath   string
 	container  string
 }
 
+//CopyReader is a Wrapper to content when it is a Reader
+type CopyReader struct {
```
cpuguy83, line 49: Do these reader types need to be exported?

cli/command/container/cp.go
```diff
@@ -34,11 +39,91 @@ const (
...
+	total int64
+}
+
+//CopyReadCloser is a Wrapper to content when it is a ReadCloser
+type CopyReadCloser struct {
+	io.ReadCloser
+	total int64
+}
+
+var (
```
cpuguy83, line 60: These variables should be localized to where they are used instead of singletons.

cli/command/container/cp.go
cpuguy83: Also can this just be one type?
For the case where we don't have a closer we can wrap it in `ioutil.NopCloser`

cpuguy83 (changes requested):

cli/command/container/cp.go
maximillianfx: The types don't need to be exported. We can made then private.
I have created two types because in the `copyFromContainer` method the parameter that will be copied (`content`) is a `ReadCloser` type (returned by client.CopyFromContainer), and in the method `copyToContainer` the `content` parameter is a `Reader`.
Do you have any suggestions to change this?

cli/command/container/cp.go
maximillianfx: I did not understand. Some of these variables need to be used in different scopes (functions), so I ended up making them "global" to the cp.go package.

maximillianfx:
Hi @cpuguy83, were you able to review my comments?

cli/command/container/cp.go
cpuguy83: The issue is, as I recall, that these values are being changed, which is undesirable.

cli/command/container/cp.go
cpuguy83: I would always have it take a ReadCloser and wrap the reader in a no-op closer.

maximillianfx: Ok, I'll check this.

cli/command/container/cp.go
maximillianfx: Hm, so I need move these variables to their respective functions ou could I make pointers to them?

maximillianfx: What do you think, @cpuguy83, if we move these fields to the types `CopyReader`, initialize them and in the `Read` function use them? So, with this change I make only one type to the `CopyReader` and `CopyReadCloser`.

maximillianfx: Hi @cpuguy83, how are you? I've made the changes. Now, only one type exists to perform the copy process and the variables are inside this type, because they are used during the `Read` method. Could you review?

cli/command/container/cp.go
```diff
@@ -34,11 +39,70 @@ const (
...
+	pt.total += int64(n)
+
+	percent := float64(pt.total) / float64(pt.copySize)
+
+	arrow := updateArrowLoading(percent * 100)
+
+	if err == nil && !pt.quiet {
+		fmt.Print("\033[u\033[K")
+		if pt.toContainer {
+			fmt.Printf("Copying to container %s %s/%s", arrow, units.HumanSize(float64(pt.total)), units.HumanSize(float64(pt.copySize)))
```
cpuguy83, line 97: I think the `copyReader` should take an i/o stream instead of assuming os.Stdout (As `Printf` and `Println` do).
Then we need `fmt.Fprint[ln,f]`.

cli/command/container/cp.go
```diff
@@ -34,11 +39,70 @@ const (
...
 	followLink bool
 	copyUIDGID bool
+	quiet      bool
 	sourcePath string
 	destPath   string
 	container  string
 }
 
+//copyReader is a Wrapper to content
+type copyReader struct {
```
cpuguy83, line 49: Maybe `cpWithProgress`, and a doc string that describes the usage.

cli/command/container/cp.go
```diff
@@ -171,12 +237,30 @@ func copyFromContainer(ctx context.Context, dockerCli command.Cli, copyConfig cp
...
+
+	if !copyConfig.quiet {
+		fmt.Println("Preparing to copy...")
+		fmt.Print("\033[s")
+	}
+
+	res := archive.CopyTo(preArchive, srcInfo, dstPath)
+
+	if !copyConfig.quiet {
+		fmt.Println()
```
cpuguy83, line 259: Same as the progress output, we should have an output stream that we write to rather than assuming `os.Stdout`

cpuguy83 (changes requested):
This looks much better!

The main thing now I think is it should not assume `os.Stdout` is the output stream.

cli/command/container/cp.go
maximillianfx: Right 😉

cli/command/container/cp.go
maximillianfx: Ok, I'll update this.

cli/command/container/cp.go
maximillianfx: Right.

cli/command/container/cp.go
maximillianfx: @cpuguy83 how we can sinalize to `copyReader` the right output (os.Stdout, file, etc...)?

cpuguy83: This can be passed down froM the cli object.

maximillianfx: Oh, right!

maximillianfx: Hi @cpuguy83 , I made the change in the output, can you review?

cli/command/container/cp.go
```diff
@@ -34,11 +39,74 @@ const (
...
+//showing the progress during the copy.
+type cpWithProgress struct {
+	io.ReadCloser
+	toContainer   bool
+	total         int64
+	copySize      int64
+	containerName string
+	copyPath      string
+	quiet         bool
+	dockerCli     command.Cli
```
cpuguy83, line 60: Should just be an `io.Writer`

cli/command/container/cp.go
```diff
@@ -272,13 +362,28 @@ func copyToContainer(ctx context.Context, dockerCli command.Cli, copyConfig cpCo
 
 		resolvedDstPath = dstDir
 		content = preparedArchive
+		copySize, content = getSize(content)
```
cpuguy83, line 365: Shouldn't we already have the size of this?

cpuguy83 (changes requested):
We should get the size of the content from the filesystem or API, but not read them into a buffer.

This can be racey, but reading archives into memory doesn't seem desirable.
If we don't have the total size we can omit the progress bar and just show the amount transferred.

cli/command/container/cp.go
maximillianfx: Ok. Yeah, we don't have the total size in hands, so I'll update the loading to show only the amount transferred.

cli/command/container/cp.go
maximillianfx: No, because the file props are obtained during the methods `copyFromContainer` and `copyToContainer`. But I'll remove this and only show the amount transferred.

cli/command/container/cp.go
maximillianfx: Hello @cpuguy83 , I've made the changes in the code, can you review? I've removed the progress bar because we don't have the total size of file before the copy process and changed the output from `dockerCli` to `io.Writer`.

maximillianfx:
Hey @cpuguy83, how are you? Were you able to review the latest changes?

cli/command/container/cp.go
cpuguy83: We should be able to get size, for local files as well as remote (though I'm not sure if the remote stat returns size right now).

maximillianfx: I have changed this. Now, only the transferred amount is displayed.

cli/command/container/cp.go
```diff
@@ -71,6 +105,7 @@ func NewCopyCommand(dockerCli command.Cli) *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVarP(&opts.followLink, "follow-link", "L", false, "Always follow symbol link in SRC_PATH")
 	flags.BoolVarP(&opts.copyUIDGID, "archive", "a", false, "Archive mode (copy all uid/gid information)")
+	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Does not display copying process")
```
cpuguy83, line 108: What do you think if we set the default based on if stdout is a tty?

maximillianfx: Could be a good idea.

cpuguy83: Yep.

maximillianfx: Ok!

maximillianfx: Done @cpuguy83

cli/command/container/cp.go
```diff
@@ -68,9 +102,12 @@ func NewCopyCommand(dockerCli command.Cli) *cobra.Command {
 		},
 	}
 
+	fileInfo, _ := os.Stdin.Stat()
+
 	flags := cmd.Flags()
 	flags.BoolVarP(&opts.followLink, "follow-link", "L", false, "Always follow symbol link in SRC_PATH")
 	flags.BoolVarP(&opts.copyUIDGID, "archive", "a", false, "Archive mode (copy all uid/gid information)")
+	flags.BoolVarP(&opts.quiet, "quiet", "q", (fileInfo.Mode()&os.ModeCharDevice) == 0, "Does not display copying process")
```
cpuguy83, line 110: I think `dockerCli` has a `IsTerminalOut` function (which is what we need to check here).

maximillianfx: A quick Cmd+F for `IsTerminalOut` in project results nothing 😕

cpuguy83: Found it: `dockerCli.Out().IsTerminal()`

maximillianfx: Alright!

maximillianfx: Now yes @cpuguy83

cpuguy83 (approved):
LGTM

maximillianfx:
I'll fix the test.

cli/command/container/cp.go
```diff
@@ -272,13 +327,27 @@ func copyToContainer(ctx context.Context, dockerCli command.Cli, copyConfig cpCo
...
 
 	options := types.CopyToContainerOptions{
 		AllowOverwriteDirWithFile: false,
 		CopyUIDGID:                copyConfig.copyUIDGID,
 	}
-	return client.CopyToContainer(ctx, copyConfig.container, resolvedDstPath, content, options)
+
+	if !copyConfig.quiet {
+		fmt.Fprint(dockerCli.Out(), "Preparing to copy...\n")
+		fmt.Fprint(dockerCli.Out(), "\033[s")
```
thaJeztah, line 340: Hm.. I'm bad at these; what's this sequence doing?

I have a feeling that a `cursor-up` is missing; here's the output after a copy;

```console
$ docker cp --quiet=false foo:/usr/share/nginx bla
Preparing to copy...
Copying from container - 4.608kB
Successfully copied 4.608kB to /go/src/github.com/docker/cli/bla
```

I think the `Preparing to copy...` needs to be overwritten by `Copying from container ...`

Perhaps for readability, we could use the https://github.com/morikuni/aec package, which we already use in other places, e.g.

```suggestion
		fmt.Fprint(dockerCli.Out(), "\033[s")
		fmt.Fprint(dockerCli.Out(), aec.Up(1))
```

cli/command/container/cp.go
```diff
@@ -34,11 +37,42 @@ const (
...
+	quiet         bool
+	writer        io.Writer
+}
+
+func (pt *cpWithProgress) Read(p []byte) (int, error) {
+	n, err := pt.ReadCloser.Read(p)
+	*pt.total += int64(n)
+
+	if err == nil && !pt.quiet {
+		fmt.Fprint(pt.writer, "\033[u\033[K")
```
thaJeztah, line 65: Also see my other comment/question; I _think_ this is "erase line", correct? If so, we could replace this with;

```suggestion
		fmt.Fprint(pt.writer, aec.EraseLine(aec.EraseModes.All))
```

(but double check if I'm correct 😅)

cli/command/container/cp.go
```diff
@@ -34,11 +37,42 @@ const (
...
+	total         *int64
+	containerName string
+	copyPath      string
+	quiet         bool
+	writer        io.Writer
+}
+
+func (pt *cpWithProgress) Read(p []byte) (int, error) {
+	n, err := pt.ReadCloser.Read(p)
+	*pt.total += int64(n)
```
thaJeztah, line 62: Should we change `total` to a `float64`, so that we don't have to convert for all places where we print?

```suggestion
	*pt.total += float64(n)
```

thaJeztah (changes requested):
Thanks!! Sorry for slow response/review; I had a look at your PR, and left comments inline.

I was making changes to the code while I wrote those comments, and will open a pull request against your branch if that's easier for you 😅

thaJeztah:
> I was making changes to the code while I wrote those comments, and will open a pull request against your branch if that's easier for you 😅

I opened https://github.com/maximillianfx/cli/pull/1 on your fork 👍

maximillianfx:
Thanks @thaJeztah ! Your comments are very hepful and necessary. Your PR is welcome 😄

thaJeztah:
You're welcome! Be sure to check my changes with the control-codes, because I'm horrible at those 😂

cli/command/container/cp.go
```diff
@@ -34,11 +37,42 @@ const (
...
+	quiet      bool
 	sourcePath string
 	destPath   string
 	container  string
 }
 
+//cpWithProgress is a Wrapper to the content being copied from or to container. This wrapper
+//contains information about the file or folder being copied (size and bytes - ReadCloser) as well as
+//source and destination path. When the bytes are copied, the `Read` method from io.ReadCloser is overwritten
+//showing the progress during the copy.
```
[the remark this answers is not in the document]
maximillianfx: Good!

cli/command/container/cp.go
```diff
@@ -71,6 +105,7 @@ func NewCopyCommand(dockerCli command.Cli) *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVarP(&opts.followLink, "follow-link", "L", false, "Always follow symbol link in SRC_PATH")
 	flags.BoolVarP(&opts.copyUIDGID, "archive", "a", false, "Archive mode (copy all uid/gid information)")
+	flags.BoolVarP(&opts.quiet, "quiet", "q", !dockerCli.Out().IsTerminal(), "Does not display copying process")
```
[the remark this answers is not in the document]
maximillianfx: True, good point.

cli/command/container/cp.go
```diff
@@ -34,11 +37,42 @@ const (
...
+//cpWithProgress is a Wrapper to the content being copied from or to container. This wrapper
+//contains information about the file or folder being copied (size and bytes - ReadCloser) as well as
+//source and destination path. When the bytes are copied, the `Read` method from io.ReadCloser is overwritten
+//showing the progress during the copy.
+type cpWithProgress struct {
+	io.ReadCloser
+	toContainer   bool
+	total         *int64
+	containerName string
+	copyPath      string
```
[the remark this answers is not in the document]
maximillianfx: Yeah, unused.

cli/command/container/cp.go
```diff
@@ -272,13 +327,27 @@ func copyToContainer(ctx context.Context, dockerCli command.Cli, copyConfig cpCo
...
+	}
+
+	res := client.CopyToContainer(ctx, copyConfig.container, resolvedDstPath, content, options)
+
+	if !copyConfig.quiet {
+		fmt.Fprint(dockerCli.Out(), "\n")
+		fmt.Fprint(dockerCli.Out(), "Successfully copied ", units.HumanSize(float64(copiedSize)), " to ", copyConfig.container, ":", dstInfo.Path, "\n")
+	}
+
+	return res
```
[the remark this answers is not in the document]
maximillianfx: True, good observation.

cli/command/container/cp.go
maximillianfx: Good.

cli/command/container/cp.go
```diff
@@ -171,12 +207,29 @@ func copyFromContainer(ctx context.Context, dockerCli command.Cli, copyConfig cp
...
+
 	if len(srcInfo.RebaseName) != 0 {
 		_, srcBase := archive.SplitPathDirEntry(srcInfo.Path)
 		preArchive = archive.RebaseArchiveEntries(content, srcBase, srcInfo.RebaseName)
 	}
-	return archive.CopyTo(preArchive, srcInfo, dstPath)
+
+	if !copyConfig.quiet {
+		fmt.Fprint(dockerCli.Out(), "Preparing to copy...\n")
+		fmt.Fprint(dockerCli.Out(), "\033[s")
```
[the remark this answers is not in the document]
maximillianfx: Uhm ok, I'll check this with you on your PR.

cli/command/container/cp.go
```diff
@@ -34,11 +37,42 @@ const (
...
+	copyPath      string
+	quiet         bool
+	writer        io.Writer
+}
+
+func (pt *cpWithProgress) Read(p []byte) (int, error) {
+	n, err := pt.ReadCloser.Read(p)
+	*pt.total += int64(n)
+
+	if err == nil && !pt.quiet {
```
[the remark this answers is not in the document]
maximillianfx: `copyProgressPrinter` make sense.

cli/command/container/cp.go
```diff
@@ -272,13 +327,27 @@ func copyToContainer(ctx context.Context, dockerCli command.Cli, copyConfig cpCo
 
 		resolvedDstPath = dstDir
 		content = preparedArchive
+		content = &cpWithProgress{ReadCloser: content, toContainer: true, containerName: copyConfig.container, copyPath: dstInfo.Path, quiet: copyConfig.quiet, writer: dockerCli.Out(), total: &copiedSize}
```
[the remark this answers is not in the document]
maximillianfx: Uhm, ok.

cli/command/container/cp.go
maximillianfx: Uhm, I'll check this with your on your PR.

cli/command/container/cp.go
maximillianfx: Yeah, erase line.

cli/command/container/cp.go
```diff
@@ -272,13 +333,33 @@ func copyToContainer(ctx context.Context, dockerCli command.Cli, copyConfig cpCo
...
 		CopyUIDGID:                copyConfig.copyUIDGID,
 	}
-	return client.CopyToContainer(ctx, copyConfig.container, resolvedDstPath, content, options)
+
+	if !copyConfig.quiet {
+		return client.CopyToContainer(ctx, copyConfig.container, resolvedDstPath, content, options)
+	}
+
+	fmt.Fprint(dockerCli.Err(), "Preparing to copy...\n")
+	fmt.Fprint(dockerCli.Err(), "\033[s")
```
thaJeztah, line 356: 🤦 @maximillianfx I missed updating this one

```suggestion
	fmt.Fprint(dockerCli.Err(), aec.Save)
```

maximillianfx: Okay 😄, I'll update this.

maximillianfx:
@thaJeztah Some problems arose, I'll made more tests.

Captura de Tela 2021-02-22 às 16 30 47

cli/command/container/cp.go
maximillianfx: Fixed.

maximillianfx:
I think that's ok now, PTAL @thaJeztah 😄

maximillianfx:
Hi @thaJeztah , how are you? Could you review again?
Thanks.

thaJeztah:
Ah, welp.. this one slipped off my radar again 😬. Let me have another look.

Before merging, we should squash the commits (let me have a quick look at doing a rebase and squash)

thaJeztah:
I rebased (will push in a minute); and found some last small things, which I added in an extra commit (but will squash);

1. Remove the "progress" line after copy is complete

    After copying, the "progress" line was kept:

        docker cp foo:/usr/share/nginx bla
        Preparing to copy...

        docker cp foo:/usr/share/nginx bla
        Copying from container - 512B

        docker cp foo:/usr/share/nginx bla
        Copying from container - 4.608kB
        Successfully copied 4.608kB to /go/src/github.com/docker/cli/bla

    The "Copying from container" line is a bit redundant after it's completed, so
    removing the line afterwards:

        docker cp foo:/usr/share/nginx bla
        Preparing to copy...

        docker cp foo:/usr/share/nginx bla
        Copying from container - 512B

        docker cp foo:/usr/share/nginx bla
        Successfully copied 4.608kB to /go/src/github.com/docker/cli/bla

2. Print progress lines with a trailing newline

    Without a newline, the cursor would be visible at the end of the progress
    line, which was a bit distracting, as it would be jumping if the progress
    line was not the same width for each refresh.

3. Swapped `fmt.Fprint` for `fmt.Fprintln` in some places

    Same result, just prevents having to add `\n` in the strings, and it automatically
    joins arguments with a space in between, so a slight bit cleaner.


Here's a before/after:

https://user-images.githubusercontent.com/1804568/115956932-ee815080-a4ff-11eb-810b-d63849746fe0.mov


https://user-images.githubusercontent.com/1804568/115956945-fd680300-a4ff-11eb-8683-4ee3873b19bf.mov

thaJeztah:
Pushed the squashed and rebased version; also squashed my last changes; diff below for posterity;

```patch
From c5e7af600e90b1793c789219773cdbf6ed1948a1 Mon Sep 17 00:00:00 2001
From: Sebastiaan van Stijn <github@gone.nl>
Date: Sat, 24 Apr 2021 13:14:13 +0200
Subject: [PATCH] Some small fix-ups / improvements

1. Remove the "progress" line after copy is complete

After copying, the "progress" line was kept:

    docker cp foo:/usr/share/nginx bla
    Preparing to copy...

    docker cp foo:/usr/share/nginx bla
    Copying from container - 512B

    docker cp foo:/usr/share/nginx bla
    Copying from container - 4.608kB
    Successfully copied 4.608kB to /go/src/github.com/docker/cli/bla

The "Copying from container" line is a bit redundant after it's completed, so
removing the line afterwards:

    docker cp foo:/usr/share/nginx bla
    Preparing to copy...

    docker cp foo:/usr/share/nginx bla
    Copying from container - 512B

    docker cp foo:/usr/share/nginx bla
    Successfully copied 4.608kB to /go/src/github.com/docker/cli/bla

2. Print progress lines with a trailing newline

Without a newline, the cursor would be visible at the end of the progress
line, which was a bit distracting, as it would be jumping if the progress
line was not the same width for each refresh.

3. Swapped `fmt.Fprint` for `fmt.Fprintln` in some places

Same result, just prevents having to add `\n` in the strings, and it automatically
joins arguments with a space in between, so a slight bit cleaner.

---
 cli/command/container/cp.go | 18 ++++++++++--------
 1 file changed, 10 insertions(+), 8 deletions(-)

diff --git a/cli/command/container/cp.go b/cli/command/container/cp.go
index 071becf94b..01940d5616 100644
--- a/cli/command/container/cp.go
+++ b/cli/command/container/cp.go
@@ -61,9 +61,9 @@ func (pt *copyProgressPrinter) Read(p []byte) (int, error) {
 		fmt.Fprint(pt.writer, aec.Restore)
 		fmt.Fprint(pt.writer, aec.EraseLine(aec.EraseModes.All))
 		if pt.toContainer {
-			fmt.Fprint(pt.writer, "Copying to container - "+units.HumanSize(*pt.total))
+			fmt.Fprintln(pt.writer, "Copying to container - "+units.HumanSize(*pt.total))
 		} else {
-			fmt.Fprint(pt.writer, "Copying from container - "+units.HumanSize(*pt.total))
+			fmt.Fprintln(pt.writer, "Copying from container - "+units.HumanSize(*pt.total))
 		}
 	}
 
@@ -229,10 +229,11 @@ func copyFromContainer(ctx context.Context, dockerCli command.Cli, copyConfig cp
 	}
 
 	fmt.Fprint(dockerCli.Err(), aec.Save)
-	fmt.Fprint(dockerCli.Err(), "Preparing to copy...\n")
+	fmt.Fprintln(dockerCli.Err(), "Preparing to copy...")
 	res := archive.CopyTo(preArchive, srcInfo, dstPath)
-	fmt.Fprint(dockerCli.Err(), "\n")
-	fmt.Fprint(dockerCli.Err(), "Successfully copied ", units.HumanSize(copiedSize), " to ", dstPath, "\n")
+	fmt.Fprint(dockerCli.Err(), aec.Restore)
+	fmt.Fprint(dockerCli.Err(), aec.EraseLine(aec.EraseModes.All))
+	fmt.Fprintln(dockerCli.Err(), "Successfully copied", units.HumanSize(copiedSize), "to", dstPath)
 
 	return res
 }
@@ -351,10 +352,11 @@ func copyToContainer(ctx context.Context, dockerCli command.Cli, copyConfig cpCo
 	}
 
 	fmt.Fprint(dockerCli.Err(), aec.Save)
-	fmt.Fprint(dockerCli.Err(), "Preparing to copy...\n")
+	fmt.Fprintln(dockerCli.Err(), "Preparing to copy...")
 	res := client.CopyToContainer(ctx, copyConfig.container, resolvedDstPath, content, options)
-	fmt.Fprint(dockerCli.Err(), "\n")
-	fmt.Fprint(dockerCli.Err(), "Successfully copied ", units.HumanSize(copiedSize), " to ", copyConfig.container, ":", dstInfo.Path, "\n")
+	fmt.Fprint(dockerCli.Err(), aec.Restore)
+	fmt.Fprint(dockerCli.Err(), aec.EraseLine(aec.EraseModes.All))
+	fmt.Fprintln(dockerCli.Err(), "Successfully copied", units.HumanSize(copiedSize), "to", copyConfig.container+":"+dstInfo.Path)
 
 	return res
 }
```

thaJeztah (approved):
LGTM, thanks!!!

thaJeztah:
@silvin-lubecki @cpuguy83 PTAL

silvin-lubecki (approved):
LGTM

silvin-lubecki:
Thanks a lot @maximillianfx for this awesome work 👍

maximillianxavier:
Thank you guys!

Azratosh:
May I ask what the rationale behind suddenly adding output to a previously silent command is? This didn't really break anything for me, but it's instead just *very* annoying, because `docker cp` now clogs up my terminal when it didn't before:

Screenshot_20230224_154335

What's obviously not shown in the screenshot are the thousands of other lines that the command now prints.

Therefore, why not just add this kind of behaviour through a `--progress` flag or similar? Why does the default behaviour for a command need to change? And moreover, if it now *does* have a progress indicator, fine - but why does it only show up when it's connected to a TTY? Why can I not pipe the progress output to a log file in case I explicitly want to see it, in order to, for example, reproduce an issue that happens during a large transfer?

timokz:
> May I ask what the rationale behind suddenly adding output to a previously silent command is? This didn't really break anything for me, but it's instead just _very_ annoying, because `docker cp` now clogs up my terminal when it didn't before:
> 
> Screenshot_20230224_154335
> 
> What's obviously not shown in the screenshot are the thousands of other lines that the command now prints.
> 
> Therefore, why not just add this kind of behaviour through a `--progress` flag or similar? Why does the default behaviour for a command need to change? And moreover, if it now _does_ have a progress indicator, fine - but why does it only show up when it's connected to a TTY? Why can I not pipe the progress output to a log file in case I explicitly want to see it, in order to, for example, see at which point a large transfer gets aborted?
> 
> Edit: Phrasing.

This seems very annoying and doesn't appear to be expected behavior. 
Maybe there is an error in the implementation of the new progress indication?

PR #2733, opened 2020-09-21
Add .IPAddresses as formatting option on docker ps

thaJeztah:
fixes https://github.com/moby/moby/issues/8786

This allows showing the IP address for each network that the container is attached to,
for example:

    docker network create foo
    docker run -d --name foo nginx:alpine
    docker network connect foo foo

    docker container ls --format 'table {{.ID}}\\t{{join .IPAddresses ", "}}'
    CONTAINER ID   IP ADDRESSES
    17e7d1910fc0   bridge:172.17.0.2, foo:172.19.0.2

    docker container ls --format='{{json .IPAddresses}}' | jq .
    [
      "bridge:172.17.0.2",
      "foo:172.19.0.2"
    ]

### DIFF
--- a/cli/command/formatter/container.go
+++ b/cli/command/formatter/container.go
@@ -4,8 +4,10 @@
 package formatter
 
 import (
+	"cmp"
 	"fmt"
 	"net"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -39,6 +41,16 @@ type Platform struct {
 
 func (p Platform) String() string {
 	return platforms.FormatAll(p.Platform)
+}
+
+// NetworkIP describes an IP-address and the network it's associated with.
+type NetworkIP struct {
+	Network string `json:"Network,omitempty"`
+	IP      string `json:"IP"`
+}
+
+func (p NetworkIP) String() string {
+	return p.Network + "/" + p.IP
 }
 
 // NewContainerFormat returns a Format for rendering using a Context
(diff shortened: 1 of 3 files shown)


thaJeztah:
Had this branch on my computer; not sure if we should merge (see discussion on https://github.com/moby/moby/issues/8786), but thought to at least push it to github 😂

hc0503 (approved):

codecat:
Are there any blockers (unrelated to Moby) to get this PR merged into Docker CLI? I think this would make the `--format` option for `docker ps` a lot more powerful.

ndeloof (commented):
LGTM. I don't understand the need for `ps` command uses an intermediate `ContainerContext` type and not just expose the plain `types.Container` returned by the API for formatting

rumpl (dismissed):

thaJeztah:
I want to have another look at this (also taking into account IPv6 addresses); slightly wondering if the `<network-name>:<ip address>` is the right / most usable format.

>  I don't understand the need for ps command uses an intermediate ContainerContext type and not just expose the plain types.Container returned by the API for formatting

I think this was done to decorate the struct with formatting functions (such as this one); the formatting is meant for "printing on the CLI", and is kinda separate from both "client" and "api", so may not be very suitable for inclusion in the upstream moby repository (as part of the types)

Benehiko (dismissed):

thaJeztah:
This still needs a bit of work, and I'll probably split the first commit, but was rebasing old PRs, and trying to make it work and a bit more flexible than just a  string ☺️

Issue #2680, opened 2020-08-14
Docker build output tar file not working


SpikePy:
**Description**

If you want to build a tar file of an image with `docker build --output type=tar,dest=out.tar .` like described here: https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs it just does not work. there is no file created! Also the `man docker-build` doesn't inclode information about the --output flag.

**Steps to reproduce the issue:**
1. run `docker build --output type=tar,dest=out.tar .` in a directory with a **Dockerfile**

**Describe the results you received:**
 out.tar is not created

**Describe the results you expected:**
 out.tar is created

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Version:           19.03.12
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        48a66213fe
 Built:             Mon Jun 22 15:45:50 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.12
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       48a66213fe
  Built:            Mon Jun 22 15:44:21 2020
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Output of `docker info`:**

```
Client:
 Debug Mode: false

Server:
 Containers: 2
  Running: 2
  Paused: 0
  Stopped: 0
 Images: 28
 Server Version: 19.03.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7ad184331fa3e55e52b890ea95e65ba581ae3429
[...]

PR #2736, opened 2020-09-22
build: print error if BuildKit-specific flags are used without BuildKit enabled

thaJeztah:
fixes https://github.com/docker/cli/issues/2680

With this patch, the `--progress`, `--secret`, `--ssh`, and `--output` flags
trigger an error when trying to use without BuildKit enabled;

    DOCKER_BUILDKIT=0 docker build --progress=plain .
    The --progress flag requires BuildKit to be enabled. Enable BuildKit with DOCKER_BUILDKIT=1

    DOCKER_BUILDKIT=0 docker build --output=foo .
    The --progress flag requires BuildKit to be enabled. Enable BuildKit with DOCKER_BUILDKIT=1

**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah:
@tiborvass @tonistiigi @silvin-lubecki PTAL

thaJeztah:
Pushed a second commit to try integrating it with `areFlagsSupported()` however, that function is called by `isSupported()`, which is called by `PersistentPreRunE`, but I think we override that for build; checking now

thaJeztah:
opened https://github.com/docker/cli/pull/2737 to remove the pre-run hack

silvin-lubecki (dismissed):
LGTM 👍

thaJeztah:
Squashed the commits, and rebased on top of https://github.com/docker/cli/pull/2737 (that one needs to be merged first)

silvin-lubecki:
@thaJeztah this PR needs a rebase 🦁

devdreamsolution (approved):

thaJeztah:
@tiborvass PTAL

tonistiigi:
The flags marked with `no-buildkit` annotation do not change the outcome of the build, with the exception of `security-opt` maybe. So I'm not in favor of erroring on them. Hiding should be enough to recommend not using them.

thaJeztah:
Do you think we should print a warning? e.g. if someone expects to have set memory limits, but they are not applied?

PR #2737, opened 2020-09-22
build: remove PersistentPreRunE hack for experimental --stream

thaJeztah:
This hack was added in an attempt to continue supporting the experimental
(non-buildkit) `--stream` option, by dynamically updating the API version
required if buildkit isn't enabled.

This hack didn't work, however, because at the moment the override is
added, the command is not yet attached to the "root" (`docker`) command,
and because of that, the command itself is the `root` command;
`cmd.Root()` returned the `build` command.

As a result, validation steps defined as `PersistentPreRunE` on the root
command were not executed, causing invalid flags/options to not producing
an error.

Attempts to use an alternative approach (for example, cobra supports both
a `PersistentPreRun` and `PersistentPreRunE`) did not work either, because
`PersistentPreRunE` takes precedence over `PersistentPreRun`, and only one
will be executed.

Given that the `--stream` option (without BuildKit) was experimental in
older API versions, we are "allowed" to break this behavior, so removing
this hack.

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -113,25 +113,6 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 		},
 	}
 
-	// Wrap the global pre-run to handle non-BuildKit use of the --platform flag.
-	//
-	// We're doing it here so that we're only contacting the daemon when actually
-	// running the command, and not during initialization.
-	// TODO remove this hack once we no longer support the experimental use of --platform
-	rootFn := cmd.Root().PersistentPreRunE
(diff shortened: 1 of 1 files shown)


thaJeztah:

thaJeztah:
erm.. brainfart. this was not about stream need to revise this

thaJeztah:
Need to search again which versions of the api support this with, and without buildit 😞

silvin-lubecki (approved):
I'm ok to break experimental things, so LGTM 👍

devdreamsolution (approved):

thaJeztah:
closing in favour of https://github.com/docker/cli/pull/2842

PR #2740, opened 2020-09-23
Add stack config command

### LINKED ISSUE #2365: Feature Request: Add command to print config
`docker-compose` has the `config` command [to print the full configuration](https://docs.docker.com/compose/reference/config/)

This is invaluable for many scenarios: 
 - checking passed build args
 - confirming that multiple YAML files have combined properly
 - ensuring that your CI/CD variables are making it into the YAML properly (e.g. GitLab CI/CD env variables don't properly support `$`, so variables like `$apr1$hash` get mangled before they enter the YAML file)
 - recording a record of each deployment 

Could swarm have a config command as well? 

My team has basically 100% migrated to swarm, but we keep `docker-compose` installed in our CI/CD environment specifically for this one subcommand 

**Example docker-compose config output**

```
docker-compose config
 networks:
   internal: {}
   web-ingress: {}
 services:
   dashboard:
     build:
       args:
         BUILD_FLAGS: VUE_APP_API_URL=https://api.dev.mydomain.com VUE_APP_COOKIE_DOMAIN=dev.mydomain.com
       context: /builds/mygroup/my-project
     image: 543066533799.dkr.ecr.us-east-1.amazonaws.com/mycontainer:64e3da5d
     networks:
       internal: {}
       web-ingress: {}
     ports:
     - published: 80
       target: 80
 version: '3.7'
```

SMFloris:
Add stack config command.

Fixes #2365 

Usage:

```
# output interpolated and merged config
docker stack config -c docker-compose.prod.yml -c docker-compose.dev.yml
```

```
# also possible to input from stdin, like docker stack deploy
docker stack config -c -
```

```
# output merged config without interpolation
docker stack config -c docker-compose.prod.yml -c docker-compose.dev.yml --skip-interpolation
```

```
# possible to pipe it back to docker stack deploy, like this
# interpolating twice is not a good idea, so one should skip interpolation
docker stack config -c docker-compose.prod.yml -c docker-compose.dev.yml --skip-interpolation | docker stack deploy -c -
```

**- What I did**
Added stack config command that outputs on stdout the merged and/or interpolated config files for docker stack deploy. 
Now it is no longer required to have docker-compose in order to see how configs are interpolated and merged before deploying your stack.

Currently, in our team, we currently use something like this to check our merge/compilation output:
```
docker-compose -f config1 -f config2 config | docker stack deploy -c -
```
The interpolation and merge that `docker-compose config` provides is untrue in a swarm deploy context since docker-compose and docker/cli are two separate projects and use different methods for interpolation and merge. With the latest version of docker-compose we encountered some problems with the merged/interpolated format not being accepted by docker stack deploy. I decided to bring a config command to docker/cli as well so we won't encounter this problem in the future.

Related issues with latest docker-compose version:
https://github.com/docker/compose/issues/7773
https://github.com/docker/compose/pull/7778

I also added relevant unit tests for the two new merges.

**- How I did it**
I used the existing dependencies (the yaml lib) and functionality present in docker cli (the stack loader and the compose loader) in order to load the config files, merge/interpolate them and then output the yaml to stdout.

Testing on our own docker-compose files, I also noticed that when merging with docker stack deploy some properties did not merge in an expected way and produced unexpected consequences.

* the volume property in each service, would not take into account the target of the volume when merging.
I believe that in many cases, you would have a prod compose file with named or anonymous volumes and a dev compose file with a bind mounted volume to the same target. The merge of `docker stack deploy` would just concatenate the volumes array and throw an error since you try to mount to the same target. See the example in (How to verify it section of this PR), on how this PR merges them.

* the command property and entrypoint in each service, would just be concatenated, whereas with `docker-compose config` it would've been replaced. Meaning, for example, that if in configFile1 you had a command like `cat file1.txt` and in configFile2 you have a command like `cat file2.txt` the merge of `docker stack deploy -c configFile1 -c configFile2` would result in the command: `cat file1.txt cat file2.txt` which is definitely not what most people would expect and want. Modified this as well so that the command in configFile2 if exists will override the command in configFile1. See the example in (How to verify it section of this PR), on how this PR merges the command/entrypoint.

**- How to verify it**

All you need is two very simple config files:

```
# docker-compose.test1.yml
version: '3.7'

services:
  bash:
    image: bash
    command: cat randomFile.txt
    volumes:
      - app:/app
    working_dir: /app

volumes:
  app:
```

```
# docker-compose.test2.yml
version: '3.7'

services:
  bash:
    image: bash
    command: echo $ENV_STRING
    volumes:
      - ./src/app:/app
    working_dir: /app
```

Now, if you run `$ENV_STRING=helloworld docker-local stack config -c docker-compose.test1.yml -c docker-compose.test2.yml` (this is with interpolation), you will get something like this:
```
version: "3.7"
services:
  bash:
    command:
    - echo
    - helloworld
    image: bash
    environment:
      ENV_STRING: helloworld
    volumes:
    - type: bind
      source: /Users/flow/Projects/docker-test/src/app
      target: /app
    working_dir: /app
volumes:
  app: {}
```

If you run `docker-local stack config -c docker-compose.test1.yml -c docker-compose.test2.yml --skip-interpolation` (this is without interpolation), you will get something like this:

```
version: "3.7"
services:
  bash:
    command:
    - echo
    - $ENV_STRING
    image: bash
    environment:
      ENV_STRING: helloworld
    volumes:
    - type: bind
      source: /Users/flow/Projects/docker-test/src/app
      target: /app
    working_dir: /app
volumes:
  app: {}
```

Added docker stack config command that outputs the merged and interpolated config files as utilised by stack deploy.

Here is neko ikiru san to apologize for the long read
image

P.S. I am new to Go, please let me know if I did anything in a not so Go way

### DIFF
--- a/cli/command/stack/config.go
+++ b/cli/command/stack/config.go
@@ -32,11 +32,7 @@ func newConfigCommand(dockerCli command.Cli) *cobra.Command {
 			}
 
 			_, err = fmt.Fprintf(dockerCli.Out(), "%s", cfg)
-			if err != nil {
-				return err
-			}
-
-			return nil
+			return err
 		},
 	}
 
--- a/cli/command/stack/config.go
+++ b/cli/command/stack/config.go
@@ -34,6 +34,7 @@ func newConfigCommand(dockerCli command.Cli) *cobra.Command {
 			_, err = fmt.Fprintf(dockerCli.Out(), "%s", cfg)
 			return err
 		},
+		Annotations: map[string]string{"experimental": ""},
 	}
 
 	flags := cmd.Flags()
--- a/cli/command/stack/config.go
+++ b/cli/command/stack/config.go
@@ -26,7 +26,7 @@ func newConfigCommand(dockerCli command.Cli) *cobra.Command {
 				return err
 			}
 
-			cfg, err := OutputConfig(configDetails, opts.SkipInterpolation)
+			cfg, err := outputConfig(configDetails, opts.SkipInterpolation)
 			if err != nil {
 				return err
 			}
@@ -47,8 +47,8 @@ func newConfigCommand(dockerCli command.Cli) *cobra.Command {
 	return cmd
 }
 
-// OutputConfig returns the merged and interpolated config file
-func OutputConfig(configFiles composetypes.ConfigDetails, skipInterpolation bool) (string, error) {
+// outputConfig returns the merged and interpolated config file
+func outputConfig(configFiles composetypes.ConfigDetails, skipInterpolation bool) (string, error) {
 	optsFunc := func(options *composeLoader.Options) {
 		options.SkipInterpolation = skipInterpolation
 	}
--- a/cli/command/stack/config.go
+++ b/cli/command/stack/config.go
@@ -17,10 +17,9 @@ func newConfigCommand(dockerCli command.Cli) *cobra.Command {
 	var opts options.Config
 
 	cmd := &cobra.Command{
-		Use:     "config [OPTIONS]",
-		Aliases: []string{"cfg"},
-		Short:   "Outputs the final config file, after doing merges and interpolations",
-		Args:    cli.NoArgs,
+		Use:   "config [OPTIONS]",
+		Short: "Outputs the final config file, after doing merges and interpolations",
+		Args:  cli.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			configDetails, err := loader.GetConfigDetails(opts.Composefiles, dockerCli.In())
 			if err != nil {
--- a/cli/command/stack/config.go
+++ b/cli/command/stack/config.go
@@ -31,7 +31,11 @@ func newConfigCommand(dockerCli command.Cli) *cobra.Command {
 				return err
 			}
 
-			fmt.Printf("%s", cfg)
+			_, err = fmt.Fprintf(dockerCli.Out(), "%s", cfg)
+			if err != nil {
+				return err
+			}
+
 			return nil
 		},
 	}
--- a/cli/command/stack/config.go
+++ b/cli/command/stack/config.go
@@ -1,0 +1,62 @@
+package stack
+
+import (
+	"fmt"
+
+	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/command/stack/loader"
+	"github.com/docker/cli/cli/command/stack/options"
+	composeLoader "github.com/docker/cli/cli/compose/loader"
+	composetypes "github.com/docker/cli/cli/compose/types"
+	"github.com/spf13/cobra"
+	yaml "gopkg.in/yaml.v2"
+)
+
+func newConfigCommand(dockerCli command.Cli) *cobra.Command {
+	var opts options.Config
+
+	cmd := &cobra.Command{
+		Use:     "config [OPTIONS]",
+		Aliases: []string{"cfg"},
+		Short:   "Outputs the final config file, after doing merges and interpolations",
+		Args:    cli.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			configDetails, err := loader.GetConfigDetails(opts.Composefiles, dockerCli.In())
+			if err != nil {
+				return err
+			}
+
+			cfg, err := OutputConfig(configDetails, opts.SkipInterpolation)
+			if err != nil {
+				return err
+			}
+
+			fmt.Printf("%s", cfg)
+			return nil
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringSliceVarP(&opts.Composefiles, "compose-file", "c", []string{}, `Path to a Compose file, or "-" to read from stdin`)
+	flags.SetAnnotation("compose-file", "version", []string{"1.25"})
+	flags.BoolVar(&opts.SkipInterpolation, "skip-interpolation", false, "Skip interpolation and output only merged config")
+	return cmd
+}
+
+// OutputConfig returns the merged and interpolated config file
+func OutputConfig(configFiles composetypes.ConfigDetails, skipInterpolation bool) (string, error) {
+	optsFunc := func(options *composeLoader.Options) {
+		options.SkipInterpolation = skipInterpolation
+	}
+	config, err := composeLoader.Load(configFiles, optsFunc)
+	if err != nil {
+		return "", err
+	}
+
+	d, err := yaml.Marshal(&config)
+	if err != nil {
+		return "", err
+	}
+	return string(d), nil
+}
--- a/cli/command/stack/config.go
+++ b/cli/command/stack/config.go
@@ -38,7 +38,6 @@ func newConfigCommand(dockerCli command.Cli) *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.StringSliceVarP(&opts.Composefiles, "compose-file", "c", []string{}, `Path to a Compose file, or "-" to read from stdin`)
-	flags.SetAnnotation("compose-file", "version", []string{"1.25"})
 	flags.BoolVar(&opts.SkipInterpolation, "skip-interpolation", false, "Skip interpolation and output only merged config")
 	return cmd
 }
--- a/cli/command/stack/config_test.go
+++ b/cli/command/stack/config_test.go
@@ -43,7 +43,7 @@ services:
 		"VERSION": "1.0",
 	}
 
-	cfg, err := OutputConfig(composetypes.ConfigDetails{
+	cfg, err := outputConfig(composetypes.ConfigDetails{
 		ConfigFiles: []composetypes.ConfigFile{
 			{Config: firstConfigData, Filename: "firstConfig"},
 			{Config: secondConfigData, Filename: "secondConfig"},
@@ -89,7 +89,7 @@ services:
 		"VERSION": "1.0",
 	}
 
-	cfg, err := OutputConfig(composetypes.ConfigDetails{
+	cfg, err := outputConfig(composetypes.ConfigDetails{
 		ConfigFiles: []composetypes.ConfigFile{
 			{Config: firstConfigData, Filename: "firstConfig"},
 			{Config: secondConfigData, Filename: "secondConfig"},
--- a/cli/command/stack/config_test.go
+++ b/cli/command/stack/config_test.go
@@ -1,0 +1,110 @@
+package stack
+
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/docker/cli/cli/compose/loader"
+	composetypes "github.com/docker/cli/cli/compose/types"
+	"github.com/docker/cli/internal/test"
+	"gotest.tools/v3/assert"
+)
+
+func TestConfigWithEmptyComposeFile(t *testing.T) {
+	cmd := newConfigCommand(test.NewFakeCli(&fakeClient{}))
+	cmd.SetOut(ioutil.Discard)
+
+	assert.ErrorContains(t, cmd.Execute(), `Please specify a Compose file`)
+}
+
+func TestConfigMergeUsingInterpolation(t *testing.T) {
+
+	firstConfig := []byte(`
+version: "3.7"
+services:
+  foo:
+    image: busybox:latest
+    command: cat file1.txt
+`)
+	secondConfig := []byte(`
+version: "3.7"
+services:
+  foo:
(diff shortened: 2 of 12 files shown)


thaJeztah:
Thank you! I haven't had time yet to look closely at these changes, but it came up in our slack channel.

I like the general idea; still thinking somewhat about naming of that command. It's good that it matches the `docker-compose` equivalent, but I think that name was a bit poorly chosen.

For `docker stack`, using `config` as name for the command could also easily lead to confusing it to be related to `docker config`, so need to do some thinking what would be a better alternative.

olljanat:
> For `docker stack`, using `config` as name for the command could also easily lead to confusing it to be related to `docker config`, so need to do some thinking what would be a better alternative.

Maybe `docker stack preview` or `docker stack validate` ?

SMFloris:
> > For `docker stack`, using `config` as name for the command could also easily lead to confusing it to be related to `docker config`, so need to do some thinking what would be a better alternative.
> 
> Maybe `docker stack preview` , `docker stack validate` or `docker stack print` ?

I would add, borrowing a bit from Terraform: `docker stack plan`

`docker stack config` is a term borrowed from docker-compose, but I agree it might lead to confusion with the existing `docker config` command. Still, since its already used in `docker-compose` it might be easier for people to use it and adopt it. Alternative to the alternative (ATTA): `docker stack deploy --config`

`docker stack preview` is a good candidate, since you actually preview what you would deploy in term of config files. The only downside is that when I think of the `preview` word, it also makes me think in `docker stack` terms that I would get to see what will be updated/created/left untouched. The same arguments can be said about the `docker stack plan` alternative. ATTA: `docker stack deploy --preview` or `docker stack deploy --dry-run`

`docker stack validate` is a great alternative. It's pretty clear that it is about the config files and their validation, but it also makes me think that it will either exit successfully if the config is valid, or error out when the config might not be valid. Maybe adding an `--output-config` argument or something would make it more verbose. ATTA: `docker stack deploy --validate`

`docker stack print` is for me not clear what it does from the get-go, but in the context of config files is a good alternative since it would print the merged config files.

@thaJeztah , what are your thoughts on these alternatives?

olljanat:
> I would add, borrowing a bit from Terraform: `docker stack plan`

Good point, even when Terraform architecture it selves is quite awful (IMO) that term "plan" make sense on both of thus use cases.

cli/command/stack/config.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/spf13/cobra"
+	yaml "gopkg.in/yaml.v2"
+)
+
+func newConfigCommand(dockerCli command.Cli) *cobra.Command {
+	var opts options.Config
+
+	cmd := &cobra.Command{
+		Use:     "config [OPTIONS]",
+		Aliases: []string{"cfg"},
```
silvin-lubecki, line 21: Can we remove the alias for now? I think we should start smaller and if we have some feedback about it then we will re-add that alias in a follow-up 👍

cli/command/stack/config.go
```diff
@@ -0,0 +1,62 @@
...
+			if err != nil {
+				return err
+			}
+
+			cfg, err := OutputConfig(configDetails, opts.SkipInterpolation)
+			if err != nil {
+				return err
+			}
+
+			fmt.Printf("%s", cfg)
```
silvin-lubecki, line 35: I think we should use the configured cli output instead:
```go
if err := fmt.Fprint(dockerCli.Out(), cfg); err != nil{
    return err
}
```

albers:
Removing myself as a reviewer because completion is not touched in this PR.

cli/command/stack/config.go
```diff
@@ -0,0 +1,62 @@
...
+
+	flags := cmd.Flags()
+	flags.StringSliceVarP(&opts.Composefiles, "compose-file", "c", []string{}, `Path to a Compose file, or "-" to read from stdin`)
+	flags.SetAnnotation("compose-file", "version", []string{"1.25"})
+	flags.BoolVar(&opts.SkipInterpolation, "skip-interpolation", false, "Skip interpolation and output only merged config")
+	return cmd
+}
+
+// OutputConfig returns the merged and interpolated config file
+func OutputConfig(configFiles composetypes.ConfigDetails, skipInterpolation bool) (string, error) {
```
silvin-lubecki, line 48: nit: this function shouldn't be exposed

cli/command/stack/config.go
```diff
@@ -0,0 +1,62 @@
...
+	optsFunc := func(options *composeLoader.Options) {
+		options.SkipInterpolation = skipInterpolation
```
silvin-lubecki, line 50: question: I wonder if we should also expose the `SkipValidation` option 🤔 
Any thought @thaJeztah ?

cli/command/stack/config_test.go
```diff
@@ -0,0 +1,110 @@
...
+  foo:
+    command:
+    - cat
+    - file2.txt
+    image: busybox:1.0
+`
+	assert.Equal(t, cfg, mergedConfig)
+}
+
+func TestConfigMergeSkipInterpolation(t *testing.T) {
```
silvin-lubecki, line 66: nit: I think we can merge this test with the previous one, using the [Table Driven Tests](https://github.com/golang/go/wiki/TableDrivenTests) pattern.

cli/command/stack/config.go
```diff
@@ -0,0 +1,62 @@
...
+func newConfigCommand(dockerCli command.Cli) *cobra.Command {
+	var opts options.Config
+
+	cmd := &cobra.Command{
+		Use:     "config [OPTIONS]",
+		Aliases: []string{"cfg"},
+		Short:   "Outputs the final config file, after doing merges and interpolations",
+		Args:    cli.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			configDetails, err := loader.GetConfigDetails(opts.Composefiles, dockerCli.In())
```
silvin-lubecki, line 25: question: @SMFloris why not using `LoadComposefile` function instead, which actually calls `getConfigDetails` ?

cli/command/stack/config.go
SMFloris: Removed alias.

cli/command/stack/config.go
SMFloris: Now using the configured cli output

cli/command/stack/config.go
SMFloris: Made sure to not expose this in latest push.

cli/command/stack/config_test.go
SMFloris: Merged the two tests using the table driven test pattern.

cli/command/stack/config.go
SMFloris: In my mind when wanting to see the final/merged/interpolated config you also care to see if it is valid. I don't think I've ever used the `SkipValidation` flag though, what are the use-cases for it?

cli/command/stack/config.go
SMFloris: Great question!

If I recall correctly, I didn't use it because I wanted to have control on how the compose files are actually getting loaded.
The `--skip-interpolation` flag is being sent down to the `composeLoader.load()` call, like so:

```
optsFunc := func(options *composeLoader.Options) {
	options.SkipInterpolation = skipInterpolation
}
config, err := composeLoader.Load(configFiles, optsFunc)
```
This is kinda important to the use-case of this new command, seeing the output with and without interpolation to figure out where things went wrong and if you're deploying the right thing.

Another reason is that `LoadComposefile` also does some other things besides loading compose files, it also does some validation (forbidden, deprecated and unsupported properties). I though those validations were just for deploy since the validation of the `composeLoader.load` is also used. 🤔  Also, since in my mind I'm piping the output back to `docker stack deploy -c -` I saw no reason in doing deploy validations when not doing a deploy.

The final reason, since I wasn't that familiar with the code-base and because I added a new command I didn't want to mess too much with the existing signatures of the methods 😅 

Now that I'm more familiar with the codebase. I can change the `LoadComposefile` function to pass down some options to `composeLoader.load()` call if you want me too (perhaps even adding an option to skip those other validations).

cli/command/stack/config.go
silvin-lubecki: Ok let's keep it simple then, we can add a skip validation flag in a follow-up if needed 👍

cli/command/stack/config.go
```diff
@@ -0,0 +1,65 @@
...
+			if err != nil {
+				return err
+			}
+
+			cfg, err := outputConfig(configDetails, opts.SkipInterpolation)
+			if err != nil {
+				return err
+			}
+
+			_, err = fmt.Fprintf(dockerCli.Out(), "%s", cfg)
```
silvin-lubecki, line 34: nit picking: 
```go
_, err = fmt.Fprintf(dockerCli.Out(), "%s", cfg)
return err
```

SMFloris: Solved in the last commit

cli/command/stack/config.go
silvin-lubecki: @thaJeztah any thoughts on that? I'm a bit on the fence, as if the output is pipped to the docker stack deploy command, then the config should print the warnings I guess... but it can also be pipped to the docker-compose command...
@SMFloris propose to call `LoadComposeFile` and add the `--skip-validation` option so one can disable the validation... WDYT?

SMFloris: @thaJeztah , could we get your input on the above, please? 😄 We are doing some ugly hacks with downgrading docker-compose in order to get valid docker stack deploy files

devdreamsolution (approved):

SMFloris:
@silvin-lubecki @thaJeztah , Any updates or new thoughts on this PR? I'm gonna keep updating this PR and keep it from going stale unless you guys have other thoughts 😄

SMFloris:
@silvin-lubecki @thaJeztah Any new developments on this?

thaJeztah (changes requested):
Arf, sorry for the very long delay; I had some review comment, but looks like I never posted them 😞

SMFloris:
@thaJeztah , thanks for your review. I'll take a better look over it tonight and come back 😄

cli/command/stack/config.go
```diff
@@ -0,0 +1,61 @@
...
+			return err
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringSliceVarP(&opts.Composefiles, "compose-file", "c", []string{}, `Path to a Compose file, or "-" to read from stdin`)
+	flags.SetAnnotation("compose-file", "version", []string{"1.25"})
```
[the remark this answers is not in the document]
SMFloris: Nice catch! Solved in the last commit.

docs/reference/commandline/stack_config.md
```diff
@@ -0,0 +1,68 @@
...
+```bash
+  service: webserver
+  environment:
+    REDIRECT_REGEX=http://host/redirect/$${1} 
+```
+
+With interpolation, the `stack config` command will replace the environment variable in the Compose file 
+with `REDIRECT_REGEX=http://host/redirect/${1}`, but then when piping it back to the `stack deploy` 
+command it will be interpolated again and result in undefined behavior. 
+That is why, when piping the output back to `stack deploy` one should always prefer the `--skip-interpolation` option.
```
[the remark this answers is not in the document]
SMFloris: I was just about to push this change, but actually saw that the compose loader we use also has a `--skip-interpolation` flag. It might cause confusion passing `!interpolate` to it. 

@thaJeztah , Should I keep it as it is to avoid confusion or remove `--skip-interpolation` in favor of `--interpolate`?

cli/command/stack/config.go
```diff
@@ -0,0 +1,61 @@
...
+			}
+
+			cfg, err := outputConfig(configDetails, opts.SkipInterpolation)
+			if err != nil {
+				return err
+			}
+
+			_, err = fmt.Fprintf(dockerCli.Out(), "%s", cfg)
+			return err
+		},
```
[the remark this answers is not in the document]
SMFloris: Added the experimental flag and check in e2e test for experimental env.

cli/command/stack/config.go
SMFloris: @thaJeztah , I managed to track the inclusion of these options down to https://github.com/docker/cli/pull/1128 

Maybe reading through that a bit can help jog some memories 😄

docs/reference/commandline/stack_config.md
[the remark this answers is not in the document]
SMFloris: @thaJeztah , have you thought on how to proceed?

PR #2741, opened 2020-09-23
builder: rephrase ENV section, remove examples for ENV key value without '='

thaJeztah:
relates to https://github.com/moby/buildkit/pull/1692#issuecomment-696823293

**- What I did**

The `ENV key value` form can be ambiguous, for example, the following defines
a single env-variable (`ONE`) with value `"TWO= THREE=world"`:

```dockerfile
ENV ONE TWO= THREE=world
```

While we cannot deprecate/remove that syntax (as it would break existing
Dockerfiles), we should reduce exposure of the format in our examples.

### DIFF
--- a/docs/reference/builder.md
+++ b/docs/reference/builder.md
@@ -500,10 +500,10 @@ Example (parsed representation is displayed after the `#`):
 
 ```dockerfile
 FROM busybox
-ENV foo /bar
-WORKDIR ${foo}   # WORKDIR /bar
-ADD . $foo       # ADD . /bar
-COPY \$foo /quux # COPY $foo /quux
+ENV FOO=/bar
+WORKDIR ${FOO}   # WORKDIR /bar
+ADD . $FOO       # ADD . /bar
+COPY \$FOO /quux # COPY $FOO /quux
 ```
 
 Environment variables are supported by the following list of instructions in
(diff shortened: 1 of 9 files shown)


thaJeztah:
@silvin-lubecki @tonistiigi ptal

docs/reference/builder.md
```diff
@@ -994,53 +994,74 @@ port. For detailed information, see the
...
+and may confuse users of your image.
+
+If an environment variable is only needed during build, and not in the final
+image, consider setting a value for a single command instead:
+
+```dockerfile
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y ...
+```
+ 
+Or using [`ARG`](#arg), which is not persisted in the final image:
```
silvin-lubecki, line 1039: I don't think we should enforce that, I consider it more a trick than a real feature.

silvin-lubecki (approved):
LGTM, with just one comment

docs/reference/builder.md
thaJeztah: It's a build-time argument/option, so I think it's a valid use, not different from, e.g.

```dockerfile
ARG VERSION=v1.0.0
RUN install foo-${VERSION}
```

silvin-lubecki: To me the `ARG` semantic is to expose a customization point to the user, not to set an env var only during build step. So yes it works, but should we advertise it in the doc?

docs/reference/builder.md
```diff
@@ -994,53 +994,74 @@ port. For detailed information, see the
...
+> This syntax does not allow for multiple environment-variables to be set in a
+> single `ENV` instruction, and can be confusing. For example, the following
+> sets a single environment variable (`ONE`) with value `"TWO= THREE=world"`:
+>
+> ```dockerfile
+> ENV ONE TWO= THREE=world
+> ```
+> 
+> The alternative syntax is supported for backward compatibility, but discouraged
+> for the reasons outlined above.
```
tonistiigi, line 1064: and may be removed in a future version.

thaJeztah: Updated; PTAL

thaJeztah: Opened https://github.com/docker/cli/pull/2743 if we want to officially deprecate

tonistiigi (approved):

PR #2743, opened 2020-09-23
deprecate Dockerfile legacy 'ENV name value' syntax

thaJeztah:
The Dockerfile `ENV` instruction allows values to be set using either `ENV name=value`
or `ENV name value`. The latter (`ENV name value`) form can be ambiguous, for example,
the following defines a single env-variable (`ONE`) with value `"TWO= THREE=world"`,
but may have intended to be setting three env-vars:

    ENV ONE TWO= THREE=world

This format also does not allow setting multiple environment-variables in a single
`ENV` line in the Dockerfile.

Use of the `ENV name value` syntax is discouraged, and may be removed in a future
release. Users are encouraged to update their Dockerfiles to use the `ENV name=value`
syntax, for example:

    ENV ONE="" TWO="" THREE="world"

```markdown
- deprecate Dockerfile legacy 'ENV name value' syntax
```

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -54,6 +54,7 @@ Deprecated | [Kernel memory limit](#kernel-memory-limit)                        
 Deprecated | [Classic Swarm and overlay networks using external key/value stores](#classic-swarm-and-overlay-networks-using-cluster-store)      | v20.03.0   | -
 Deprecated | [Support for the legacy `~/.dockercfg` configuration file for authentication](#support-for-legacy-dockercfg-configuration-files)   | v20.03.0   | -
 Deprecated | [CLI plugins support](#cli-plugins-support)                                                                                        | v20.03.0   | -
+Deprecated | [Dockerfile legacy `ENV name value` syntax](#dockerfile-legacy-env-name-value-syntax)                                                                                        | v20.03.0   | -
 Deprecated | [Pushing and pulling with image manifest v2 schema 1](#pushing-and-pulling-with-image-manifest-v2-schema-1)                        | v19.03.0   | v20.03.0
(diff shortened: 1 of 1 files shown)


thaJeztah:
@tonistiigi @tiborvass ptal

tonistiigi (approved):

tiborvass (approved):

tianon:
What's the end goal of deprecating this (very widely used) syntax?  Do we actually think we can remove support for it in the future without causing massive breakage?  (Again, to what end?)

Why not instead detect the ambiguous case and provide a warning?

thaJeztah:
Some more details in https://github.com/moby/buildkit/pull/1692#issuecomment-696559998

tonistiigi:
There is no timeline to remove it atm. and we'll not do it if it can cause a big breakage. This is phase 1 where we need to send a clear message that this syntax should not be used anymore.

tanzislam:
I'd recommend not removing the without-`=` syntax, as I find that much more readable when my variable values need to have `=` and space characters in them. (The cost of the extra image/builder layers due to separate `ENV` statements is minimal compared to the readability improvement.)

tianon:
Reiterating for the record that I think deprecating this syntax is a mistake. :disappointed:

Coming from the context of https://github.com/moby/buildkit/pull/4759, I would re-iterate "Why not instead detect the ambiguous case and provide a warning?"

Helping people find `Dockerfiles` that are _actually_ problematic/ambiguous/possibly not doing what they expected seems a lot more fruitful than trying to remove this syntax entirely.

tianon:
Now I'm seeing a lot of useless `Dockerfile` churn/noise for completely 100% unambiguous `ENV foo bar` lines just to silence the new warning :upside_down_face:

Can we please remove that linter warning? (or at least disable it by default somehow so users can opt into it instead?)

PR #2696, opened 2020-08-28
Refactor / cleanup Dockerfile.dev

thaJeztah:
Follow-up to https://github.com/docker/cli/pull/2693

This is just a first step (more Dockerfiles and changes to look at); see individual commits for details

PR #2744, opened 2020-09-24
Fix make -f docker.Makefile without buildkit enabled

thaJeztah:
Commit 01cd748eb6eec123d50637b7b1c835755392717e (https://github.com/docker/cli/pull/2696) started using Dockerfile features that are currently only supported with buildkit enabled.

While we enabled buildkit in our Jenkinsfile, the makefile did not do the same, causing some targets to fail if buildkit was not enabled.

This patch updates the makefile to enable buildkit.

**- What I did**

**- How I did it**

**- How to verify it**


silvin-lubecki (approved):
LGTM

PR #2750, opened 2020-09-28
[19.03 backport] builder: rephrase ENV section, remove examples for ENV key value without '='

thaJeztah:
backport of https://github.com/docker/cli/pull/2741
relates to https://github.com/moby/buildkit/pull/1692#issuecomment-696823293

**- What I did**

The `ENV key value` form can be ambiguous, for example, the following defines
a single env-variable (`ONE`) with value `"TWO= THREE=world"`:

```dockerfile
ENV ONE TWO= THREE=world
```

While we cannot deprecate/remove that syntax (as it would break existing
Dockerfiles), we should reduce exposure of the format in our examples.

### DIFF
--- a/cli/command/image/import_test.go
+++ b/cli/command/image/import_test.go
@@ -81,6 +81,14 @@ func TestNewImportCommandSuccess(t *testing.T) {
 		},
 		{
 			name: "change",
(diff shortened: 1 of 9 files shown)


silvin-lubecki (approved):
LGTM

PR #2640, opened 2020-07-20
Revert "Add CodeQL Analysis workflow"

thaJeztah:
Added in https://github.com/docker/cli/pull/2630 to try if it worked, but this feature does not appear to be generally available yet, so reverting for now.

This reverts commit 454c1927c09860179a7544b1a9ea718b108789bf.

PR #2630, opened 2020-07-15
Add CodeQL Analysis workflow

thaJeztah:
similar to https://github.com/containerd/containerd/pull/4258

PR #2768, opened 2020-10-01
Add CodeQL Analysis workflow (again)

thaJeztah:
previously tried in https://github.com/docker/cli/pull/2630, but reverted in https://github.com/docker/cli/pull/2640 because the feature was not yet generally available.

This feature now went GA, so adding this workflow again. Workflow is updated to the suggested template when enabling.


silvin-lubecki (approved):
LGTM

silvin-lubecki:
Let's see what happens enabling it 👀

PR #2770, opened 2020-10-01
GH Actions: run CodeQL only on cron-job, not on pull requests

thaJeztah:
follow-up to https://github.com/docker/cli/pull/2768

Analysis takes more than 10 minutes, which is quite long to run, so change it to only run on a cron job.

Also changing the cron to run on Thursdays instead of Fridays; no need to worry about security issues during the Weekend.


silvin-lubecki (approved):
LGTM

PR #2774, opened 2020-10-02
Always enable experimental features

thaJeztah:
replaces / closes https://github.com/docker/cli/pull/1355

The CLI disabled experimental features by default, requiring users to set a configuration option to enable them.

Disabling experimental features was a request from Enterprise users that did not want experimental features to be accessible.

We are changing this policy, and now enable experimental features by default. Experimental features may still change and/or removed, and will be highlighted in the documentation and "usage" output.

For example, the `docker manifest inspect --help` output now shows:

    EXPERIMENTAL:
      docker manifest inspect is an experimental feature.

      Experimental features provide early access to product functionality. These features
      may change between releases without warning or can be removed entirely from a future
      release. Learn more about experimental features: https://docs.docker.com/go/experimental/

### DIFF
--- a/cli-plugins/manager/metadata.go
+++ b/cli-plugins/manager/metadata.go
@@ -23,6 +23,6 @@ type Metadata struct {
 	// URL is a pointer to the plugin's homepage.
 	URL string `json:",omitempty"`
 	// Experimental specifies whether the plugin is experimental.
-	// Experimental plugins are not displayed on non-experimental CLIs.
+	// Deprecated: experimental features are now always enabled in the CLI
 	Experimental bool `json:",omitempty"`
 }
--- a/kubernetes/check.go
+++ b/kubernetes/check.go
@@ -24,18 +24,18 @@ const (
 )
 
 // GetStackAPIVersion returns the most appropriate stack API version installed.
-func GetStackAPIVersion(serverGroups discovery.ServerGroupsInterface, experimental bool) (StackVersion, error) {
+func GetStackAPIVersion(serverGroups discovery.ServerGroupsInterface) (StackVersion, error) {
 	groups, err := serverGroups.ServerGroups()
 	if err != nil {
 		return "", err
 	}
 
-	return getAPIVersion(groups, experimental)
+	return getAPIVersion(groups)
 }
 
-func getAPIVersion(groups *metav1.APIGroupList, experimental bool) (StackVersion, error) {
+func getAPIVersion(groups *metav1.APIGroupList) (StackVersion, error) {
 	switch {
-	case experimental && findVersion(apiv1alpha3.SchemeGroupVersion, groups.Groups):
+	case findVersion(apiv1alpha3.SchemeGroupVersion, groups.Groups):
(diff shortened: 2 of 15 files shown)


cli-plugins/manager/metadata.go
```diff
@@ -24,5 +24,6 @@ type Metadata struct {
 	URL string `json:",omitempty"`
 	// Experimental specifies whether the plugin is experimental.
 	// Experimental plugins are not displayed on non-experimental CLIs.
+	// TODO decide if we want to annotate plugins that are experimental
```
thaJeztah, line 27: /cc @chris-crone @justincormack

kubernetes/check.go
```diff
@@ -24,6 +24,7 @@ const (
 )
 
 // GetStackAPIVersion returns the most appropriate stack API version installed.
+// TODO remove "experimental" argument now that experimental-cli features are enabled by default
```
thaJeztah, line 27: /cc @chris-crone

kubernetes/check.go
```diff
@@ -33,6 +34,7 @@ func GetStackAPIVersion(serverGroups discovery.ServerGroupsInterface, experiment
 	return getAPIVersion(groups, experimental)
 }
 
+// TODO remove "experimental" argument now that experimental-cli features are enabled by default
```
thaJeztah, line 37: /cc @chris-crone (let me know if I should just remove it here)

cli-plugins/manager/metadata.go
justincormack: I dont think we want to annotate them.

justincormack (approved):

thaJeztah:
Looks like I missed some tests; updating 

```
You can run 'go test . -test.update-golden' to automatically update testdata/docker-help-helloworld.golden to the new expected value.'
``

This test looks flaky (saw it failing on a few runs);

```
=== Failed
=== FAIL: cli/command/container TestRemoveForce/without_force (0.00s)
    --- FAIL: TestRemoveForce/without_force (0.00s)
        rm_test.go:37: assertion failed: 
            --- removed
            +++ →
            {[]string}[?->0]:
            	-: <non-existent>
            	+: "mycontainer"
            

=== FAIL: cli/command/container TestRemoveForce (0.00s)
```

kubernetes/check.go
thaJeztah: removed

kubernetes/check.go
thaJeztah: removed

cli-plugins/manager/metadata.go
thaJeztah: Removed the TODO, and marked the field deprecated

thaJeztah:
FWIW, I preserved the `Experimental:      true` in the output of `docker version`. It's likely that there are scripts that use this output to check if the  cli is configured with experimental features, and we don't want to break those (we could mark that output as deprecated as well, and remove it at some point though)

silvin-lubecki (approved):
LGTM

PR #2775, opened 2020-10-02
Replace tab with spaces in usage output

thaJeztah:
All output of the usage / `--help` output uses spaces, and having a tab in the output can be somewhat cumbersome (e.g. our YAML docs generator doesn't like them, and copy/pasing the output in iTerm produces a warning).

This patch changes the output to use two spaces instead.

**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah:
ping @silvin-lubecki ptal

tiborvass (approved):

PR #2807, opened 2020-10-22
[Docs] Fixes wrong Docker image name references with repository

froblesmartin:
Docker images do not have a name. They have a repository name, a tag, an ID...
References to Docker image names mean actually Docker image repository names.

**- What I did**
Replace docker images `name` with docker images `repository`. 

**- How I did it**
Basic change to MardDown files.

**- How to verify it**
It is just documentation, PR review is the way to verify it.

Fixing wrong Docker image name references with repository

Cute panda bear

### DIFF
--- a/docs/reference/commandline/images.md
+++ b/docs/reference/commandline/images.md
@@ -17,8 +17,8 @@ Options:
   -f, --filter value    Filter output based on conditions provided (default [])
                         - dangling=(true|false)
                         - label=<key> or label=<key>=<value>
-                        - before=(<image-name>[:tag]|<image-id>|<image@digest>)
-                        - since=(<image-name>[:tag]|<image-id>|<image@digest>)
+                        - before=(<image-repository>[:tag]|<image-id>|<image@digest>)
+                        - since=(<image-repository>[:tag]|<image-id>|<image@digest>)
                         - reference=(pattern of an image reference)
       --format string   Pretty-print images using a Go template
       --help            Print usage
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thanks for contributing; _technically_ this is correct; I recall that we used "name" because for _most_ users, "repository" is somewhat confusing. Perhaps we should remove the `-name`, and just `<image>[:<tag>]`, that way "circumventing" both _confusion_ and "incorrect" terminology; wdyt?

froblesmartin:
> Thanks for contributing; _technically_ this is correct; I recall that we used "name" because for _most_ users, "repository" is somewhat confusing. Perhaps we should remove the `-name`, and just `<image>[:<tag>]`, that way "circumventing" both _confusion_ and "incorrect" terminology; wdyt?

Hi @thaJeztah , thanks for your fast check of the PR 😋 

I get your point, but `image-repository` is actually more accurate as an image is much more than just the repository name. In my opinion, calling things in a different way will just bring more confusion in the future. If it has been decided to call something in a specific way, everything should point in that direction. 

But anyway, I don't really have a decent view of how people use these terms, in my case I think reading `repository` and `image` or `image name` can be the same for most people. But you know better than me what would be the best for the community 😉

froblesmartin:
@thaJeztah any new on this?

devdreamsolution (approved):

froblesmartin:
@albers could you review this? :yum:

PR #2809, opened 2020-10-23
Deprecation: change --stream error to a warning, and update deprecated.md

thaJeztah:
Relates to https://github.com/docker/cli/pull/2105
Relates to https://github.com/moby/moby/pull/39983

Addresses https://github.com/docker/docker-ce/pull/660#discussion_r510536957

### builder: print deprecation warning instead of failing for `--stream`

While performance will be worse, we can safely ignore the `--stream` option when used, and print a deprecation warning instead of failing the build.

With this patch:

    echo -e "FROM scratch\nLABEL foo=bar" | docker build --stream -
    DEPRECATED: The experimental --stream flag has been removed and the build context
                will be sent non-streaming. Enable BuildKit instead with DOCKER_BUILDKIT=1
                to stream build context, see https://docs.docker.com/go/buildkit/

    Step 1/2 : FROM scratch
     --->
    Step 2/2 : LABEL foo=bar
     ---> Running in 99e4021085b6
    Removing intermediate container 99e4021085b6
     ---> 1a7a41be241f
    Successfully built 1a7a41be241f

### Deprecation: add experimental docker build --stream option

Docker v17.07 introduced an experimental `--stream` flag on `docker build` which
allowed the build-context to be incrementally sent to the daemon, instead of
unconditionally sending the whole build-context.

This functionality has been reimplemented as part of BuildKit, which uses streaming
by default and the `--stream` option will be ignored when using the classic builder,
printing a deprecation warning instead.

Users that want to use this feature are encouraged to enable BuildKit by setting
the `DOCKER_BUILDKIT=1` environment variable or through the daemon or CLI configuration
files.

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -246,7 +246,11 @@ func runBuild(dockerCli command.Cli, options buildOptions) error {
 	)
 
 	if options.stream {
-		return errors.New("Experimental flag --stream was removed, enable BuildKit instead with DOCKER_BUILDKIT=1")
+		_, _ = fmt.Fprint(dockerCli.Err(), `DEPRECATED: The experimental --stream flag has been removed and the build context
+            will be sent non-streaming. Enable BuildKit instead with DOCKER_BUILDKIT=1
+            to stream build context, see https://docs.docker.com/go/buildkit/
+
+`)
 	}
 
 	if options.dockerfileFromStdin() {
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -56,6 +56,7 @@ Deprecated | [Classic Swarm and overlay networks using external key/value stores
 Deprecated | [Support for the legacy `~/.dockercfg` configuration file for authentication](#support-for-legacy-dockercfg-configuration-files)   | v20.10     | -
 Deprecated | [CLI plugins support](#cli-plugins-support)                                                                                        | v20.10     | -
 Deprecated | [Dockerfile legacy `ENV name value` syntax](#dockerfile-legacy-env-name-value-syntax)                                              | v20.10     | -
+Removed    | [`docker build --stream` flag (experimental)](#docker-build---stream-flag-experimental)                                            | v20.10     | v20.10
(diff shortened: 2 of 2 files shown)


thaJeztah:
@tiborvass @cpuguy83 @tonistiigi PTAL

cli/command/image/build.go
```diff
@@ -246,7 +246,11 @@ func runBuild(dockerCli command.Cli, options buildOptions) error {
 	)
 
 	if options.stream {
-		return errors.New("Experimental flag --stream was removed, enable BuildKit instead with DOCKER_BUILDKIT=1")
+		_, _ = fmt.Fprint(dockerCli.Err(), `DEPRECATED: The experimental --stream flag is deprecated and the build context
```
cpuguy83, line 249: `s/deprecated/removed/`

cli/command/image/build.go
```diff
@@ -246,7 +246,11 @@ func runBuild(dockerCli command.Cli, options buildOptions) error {
...
+            will be sent non-streaming. Enable BuildKit instead with DOCKER_BUILDKIT=1
```
cpuguy83, line 250: "Consider using BuildKit instead", maybe with a doc link?

cli/command/image/build.go
thaJeztah: I was a bit in doubt, as the flag is "technically" still there, but yeah from a user's perspective, I guess we could say it's removed.

cli/command/image/build.go
thaJeztah: Good suggestion; I recently added https://github.com/docker/docker.github.io/pull/11460, and this would be a good candidate for that. Need to check if we have a good location already to point users to (on how to enable buildkit).

can then add a URL like `https://docs.docker.com/go/buildkit`

thaJeztah: Opened https://github.com/docker/docker.github.io/pull/11621 to implement `https://docs.docker.com/go/buildkit/` as a redirect

thaJeztah:
Updated; PTAL

chris-crone (approved):
LGTM

tonistiigi (approved):

tiborvass (approved):

PR #2816, opened 2020-10-26
Update kill.md

callmePicacho:
only SIGKILL，not KILL signal

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/commandline/kill.md
+++ b/docs/reference/commandline/kill.md
@@ -13,7 +13,7 @@ Kill one or more running containers
 
 Options:
       --help            Print usage
-  -s, --signal string   Signal to send to the container (default "KILL")
+  -s, --signal string   Signal to send to the container (default "SIGKILL")
 ```
 
 ## Description


silvin-lubecki (changes requested):
Hello @callmePicacho and thanks for this PR, but I think this help message is accurate.
It's the same output as `docker kill --help`. By the way both `KILL` and `SIGKILL` are accepted, it's just that the default one is `KILL` when you don't specify the flag.

devdreamsolution (approved):

thaJeztah (changes requested):
I agree. I understand it may be somewhat confusing (because `SIGKILL` (`SIGxxx`) and `KILL` are used interchangeably, but the current docs reflect the actual value that's set as default, and reflects what the cli `--help` output prints;


```console
$ docker kill --help | grep signal
  -s, --signal string   Signal to send to the container (default "KILL")
```

thaJeztah:
Let me close this PR, per the above discussion.

Thank you for contributing! Even though we didn't merge your PR, it _is_ appreciated!

PR #2819, opened 2020-10-29
registry: don't call "/info" API endpoint to get default registry

thaJeztah:
The CLI currenly calls the `/info` endpoint to get the address of the default registry to use.

This functionality was added as part of the initial Windows implementation of the engine. For legal reasons, Microsoft Windows (and thus Docker images based on Windows) were not allowed to be distributed through non-Microsoft infrastructure. As a temporary solution, a dedicated "registry-win-tp3.docker.io" registry was created to serve Windows images.

As a result, the default registry was no longer "fixed", so a helper function (`ElectAuthServer`) was added to allow the CLI to get the correct registry address from the daemon. (docker/docker PR's/issues 18019, 19891, 19973)

Using separate registries was not an ideal solution, and a more permanent solution was created by introducing "foreign image layers" in the distribution spec, after which the "registry-win-tp3.docker.io" ceased to exist, and removed from the engine through https://github.com/moby/moby/pull/21100.

However, the `ElectAuthServer` was left in place, quoting from that PR;

> make the client check which default registry the daemon uses is still more correct than leaving it up to the client, even if it won't technically matter after this PR. There may be some backward compatibility scenarios where `ElectAuthServer` [sic] is still helpful.

That comment was 5 years ago, and given that the engine and cli are released in tandem, and the default registry is not configurable, we can save the extra roundtrip to the daemon by using a fixed value.

This patch deprecates the `ElectAuthServer` function, and makes it return the default registry without calling (potentially expensive) `/info` API endpoint.


thaJeztah:
@tiborvass @tonistiigi PTAL WDYT?

silvin-lubecki (approved):
LGTM 😻

thaJeztah:
There's one thing I need to check; I'm looking at changing the default in docker/engine from the old `index.docker.io/v1/` (the `/v1` is replaced by `v2` or `v1` depending on what part of Docker Hub is used, so ignore that) to the actual `registry-1.docker.io/v2/`. I _think_ that should not cause a compatibility issue if a potential cli would still use `index.docker.io/v1/`, but I need to push a WIP branch that makes that change to double-check.

devdreamsolution (approved):

rumpl (approved):

thaJeztah:
Thx! Okay let's get this one in 👍

PR #2828, opened 2020-11-04
[19.03] 2825 Update activate help information

awmirantis:
**- What I did**
Update help information for the `activate` command to reflect the current required usage.

**- How I did it**

**- How to verify it**
run:
`docker engine activate --help`

**- Description for the changelog**
Update 'docker engine activate' help information

**- A picture of a cute animal (not mandatory but encouraged)**

cats

Issue #2213, opened 2019-12-05
[epic] Remove "engine activate/check/update" subcommands


thaJeztah:
These subcommands were created to allow upgrading a Docker Community engine to Docker Enterprise, but never really took off. With the migration of the docker "enterprise" products to Mirantis, we should remove these subcommands as releases may no longer be done in sync in future.

- [x] https://github.com/docker/cli/pull/2207 Remove "docker engine" subcommands
- [x] https://github.com/docker/docker-ce-packaging/pull/411 Remove image-based build (for docker engine activate)
- [x] https://github.com/docker/docker-ce-packaging/pull/420 Remove some remnants related to image-based builds
- [x] https://github.com/docker/docker-ce-packaging/pull/426 Remove more remnants from image-based engine
- [x] https://github.com/docker/docker.github.io/pull/9984 Remove "docker engine" commands
- [x] updated deprecated features page (https://github.com/docker/cli/pull/2319)
- [x] changes in release-repositories
- [ ] backport (packaging changes only?) to 19.03 (and 18.09?) branches

seemethere:
I believe https://github.com/docker/docker-ce-packaging/pull/411 should suffice

thaJeztah:
Thanks for looking, @seemethere 🤗

chris-crone:
@nebuk89 Do we still need to publish the EE images?

thaJeztah:
Work for this is done for "master" / 20.03, we might still consider backporting to 19.03 packaging, but no need to track that here

PR #2207, opened 2019-11-28
Remove "docker engine" subcommands

thaJeztah:
These subcommands were created to allow upgrading a Docker Community engine to Docker Enterprise, but never really took off.

This patch removes the `docker engine` subcommands, as they added quite some complexity / additional code.

Signed-off-by: Sebastiaan van Stijn <github@gone.nl>

**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #2829, opened 2020-11-05
[draft][19.03 backport] Remove "docker engine" subcommands

thaJeztah:
Backport of https://github.com/docker/cli/pull/2207
alternative to / closes https://github.com/docker/cli/pull/2828

Opening this for discussion; These subcommands were created to allow upgrading a Docker Community engine to Docker Enterprise, but with the move of the enterprise products to Mirantis, these commands were no longer working; the images used to distribute the "enterprise" engine are no longer updated, and the latest versions available on docker hub are outdated (and may have unpatched vulnerabilities, so should not be recommended to use).

Reference documentation for these subcommands were already removed from docs.docker.com, and removing these commands (even as part of a patch release) may be the best option.

addresses https://github.com/docker/cli/issues/2213


thaJeztah:
hm. CircleCI for some reason isn't triggered on this PR (Jenkins is though)

devdreamsolution (approved):

PR #2830, opened 2020-11-06
Use consistent formatting and sorted keys in all JSON examples for dockerd reference

tianon:
**- What I did**

Sorted the JSON objects and made the format (whitespace) consistent -- I went to go add some new items, and wasn't sure where to add them because the order felt so random. :sweat_smile:

**- How I did it**

I ran every JSON document through `jq -S` (which sorts the keys and consistently pretty-prints the result in a format which matches the majority of documents in this file).

**- How to verify it**

Make sure the object keys are in sorted order and the whitespace is consistent. :smile:

image

(As always, happy to adjust, amend, rebase, close, etc etc etc)

### DIFF
--- a/docs/reference/commandline/dockerd.md
+++ b/docs/reference/commandline/dockerd.md
@@ -967,18 +967,18 @@ The following is an example adding 2 runtimes via the configuration:
 
 ```json
 {
-	"default-runtime": "runc",
-	"runtimes": {
-		"runc": {
-			"path": "runc"
-		},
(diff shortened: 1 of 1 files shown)


thaJeztah:
LOL, guess I missed these when doing https://github.com/docker/cli/pull/2519 🤔  (I didn't sort them though at the time, thanks!)

thaJeztah (approved):
LGTM

PR #2831, opened 2020-11-06
Add "--containerd-namespace" daemon flags in completion and docs

tianon:
This also adds the missing `containerd` key in the example daemon configuration files.

This is a follow-up / companion to https://github.com/moby/moby/pull/41640.

image

### DIFF
--- a/contrib/completion/zsh/_docker
+++ b/contrib/completion/zsh/_docker
@@ -2646,6 +2646,8 @@ __docker_subcommand() {
                 "($help)*--cluster-store-opt=[Cluster store options]:Cluster options:->cluster-store-options" \
                 "($help)--config-file=[Path to daemon configuration file]:Config File:_files" \
                 "($help)--containerd=[Path to containerd socket]:socket:_files -g \"*.sock\"" \
+                "($help)--containerd-namespace=[Containerd namespace to use]:containerd namespace:" \
+                "($help)--containerd-plugins-namespace=[Containerd namespace to use for plugins]:containerd namespace:" \
                 "($help)--data-root=[Root directory of persisted Docker data]:path:_directories" \
                 "($help -D --debug)"{-D,--debug}"[Enable debug mode]" \
                 "($help)--default-gateway[Container default gateway IPv4 address]:IPv4 address: " \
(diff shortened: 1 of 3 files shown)


docs/reference/commandline/dockerd.md
```diff
@@ -1463,6 +1468,9 @@ This is a full example of the allowed configuration options on Windows:
   "bridge": "",
   "cluster-advertise": "",
   "cluster-store": "",
+  "containerd": "\\\\.\\pipe\\containerd-containerd",
```
tianon, line 1471: I'm not _100%_ sure this is correct, but I would say I'm about 99% sure now (with @cpuguy83's help digging :heart:).

cc @StefanScherer perhaps you can help confirm? :innocent: :pray: :heart:

tianon: (This gets passed directly to `containerd.New`, which on Linux just takes the direct socket path not a URI like `unix://...`)

contrib/completion/zsh/_docker
```diff
@@ -2646,6 +2646,8 @@ __docker_subcommand() {
                 "($help)*--cluster-store-opt=[Cluster store options]:Cluster options:->cluster-store-options" \
                 "($help)--config-file=[Path to daemon configuration file]:Config File:_files" \
                 "($help)--containerd=[Path to containerd socket]:socket:_files -g \"*.sock\"" \
+                "($help)--containerd-namespace=[Containerd namespace to use]:containerd namespace:" \
+                "($help)--containerd-plugins-namespace=[Containerd namespace to use for plugins]:containerd namespace:" \
```
tianon, line 2650: I'm less confident that this is right -- I definitely don't use `zsh` but this seems to match the examples around it.

docs/reference/commandline/dockerd.md
StefanScherer: The UNC path looks good for the named pipe, otherwise a URI would look like `npipe://...`

albers (approved):
Bash completion LGTM.

devdreamsolution (approved):

thaJeztah (approved):
LGTM, thanks!

PR #2840, opened 2020-11-13
2839-%SystemDrive% is not always C:\

Emeka-NMC:
On Windows, the OS drive is often `C:\`, but can be changed.
In other words, Windows can be installed on a different drive (e.g. `D:\`).

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/builder.md
+++ b/docs/reference/builder.md
@@ -1709,7 +1709,7 @@ Keep the following things in mind about volumes in the `Dockerfile`.
   the destination of a volume inside the container must be one of:
 
   - a non-existing or empty directory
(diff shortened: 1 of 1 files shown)


thaJeztah:
This is referring to the path inside the container; ISTR the system drive could not be changed inside the container itself (but I may be mistaken), and it's possible there's code special-casing that. 

Do you have examples of a container image using a different system drive _inside_ the container?

Emeka-NMC:
You are right. I'll abandon the PR.

thaJeztah:
Thanks! Let me close this one 👍

PR #2857, opened 2020-12-01
Help guides link

gtardif:
**- What I did**
* Add a link to docker help guides (at docs.docker.com) for general help messages 
* gate this with an env var to revert to no link if users want to avoid this additional message

link actually depends on https://github.com/docker/docker.github.io/pull/11812

**- How I did it**
Added section at the bottom of go template, and a function to check if it should display or not

**- How to verify it**
`docker help` will display it, `docker --help` also, but not `docker run --help` 

### DIFF
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -37,6 +37,7 @@ func setupCommonRootCommand(rootCmd *cobra.Command) (*cliflags.ClientOptions, *p
 	cobra.AddTemplateFunc("isPlugin", isPlugin)
 	cobra.AddTemplateFunc("isExperimental", isExperimental)
 	cobra.AddTemplateFunc("displayHelpLink", displayHelpLink)
+	cobra.AddTemplateFunc("cyan", cyan)
 	cobra.AddTemplateFunc("decoratedName", decoratedName)
 
 	rootCmd.SetUsageTemplate(usageTemplate)
@@ -210,6 +211,16 @@ func displayHelpLink(cmd *cobra.Command) bool {
 	return !cmd.HasParent() && !hideGuides
 }
 
+var cyan = color("\033[1;36m%s\033[0m")
+
+func color(colorString string) func(...interface{}) string {
+	sprint := func(args ...interface{}) string {
+		return fmt.Sprintf(colorString,
+			fmt.Sprint(args...))
+	}
+	return sprint
+}
+
 func isPlugin(cmd *cobra.Command) bool {
 	return cmd.Annotations[pluginmanager.CommandAnnotationPlugin] == "true"
 }
@@ -366,7 +377,7 @@ Invalid Plugins:
 Run '{{.CommandPath}} COMMAND --help' for more information on a command.
 {{- end}}
 {{- if displayHelpLink .}}
-To get more help with docker, check out guides at https://docs.docker.com/go/guides/
+{{ cyan "To get more help with docker, check out guides at https://docs.docker.com/go/guides/" }}
 {{- end}}
 `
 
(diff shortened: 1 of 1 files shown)


cli/cobra.go
```diff
@@ -205,6 +206,21 @@ func isExperimental(cmd *cobra.Command) bool {
 	return experimental
 }
 
+func displayHelpLink(cmd *cobra.Command) bool {
+	hideGuides := os.Getenv("DOCKER_HIDE_HELP_GUIDES") != ""
```
thaJeztah, line 210: Not sure if we need this; I think overall adding content to the `help` output should be ok

cli/cobra.go
```diff
@@ -360,6 +376,9 @@ Invalid Plugins:
 
 Run '{{.CommandPath}} COMMAND --help' for more information on a command.
 {{- end}}
+{{- if displayHelpLink .}}
+{{ cyan "To get more help with docker, check out guides at https://docs.docker.com/go/guides/" }}
```
thaJeztah, line 380: Wondering (discussed that option briefly with @chris-crone) if instead of hard-coding, we should use annotations similar to; https://github.com/docker/cli/blob/v20.10.0-rc1/cli/cobra.go#L195-L198

That way we could just print the information if its set on a specific (or "any") command

cli/cobra.go
gtardif: cc @nebuk89 can we drop this requirement ? (I agree this doesn't seem too invasive)

cli/cobra.go
chris-crone: +1 to this but we can also do it as a follow up if we add links for specific helps

gtardif: Right, added the additional help message in root annotation, so we can add specific messages to any command as we like in the future. (also renamed methods for something more homogeneous

cli/cobra.go
gtardif: removed this env var check with other changes (comment below), we can add it back later if needed

nebuk89: let's leave it out for now... I agree that it's good practice but this kind of feels like something help has just been missing rather than something very pushy :)

cli/cobra.go
```diff
@@ -47,6 +49,8 @@ func setupCommonRootCommand(rootCmd *cobra.Command) (*cliflags.ClientOptions, *p
 	rootCmd.PersistentFlags().MarkShorthandDeprecated("help", "please use --help")
 	rootCmd.PersistentFlags().Lookup("help").Hidden = true
 
+	rootCmd.Annotations = map[string]string{"additionalHelp": cyan("To get more help with docker, check out guides at https://docs.docker.com/go/guides/")}
```
chris-crone, line 52: super nit: @thaJeztah do we need the trailing slash

cli/cobra.go
```diff
@@ -205,6 +208,27 @@ func isExperimental(cmd *cobra.Command) bool {
...
+		return additionalHelp
+	}
+	return ""
+}
+
+func hasAdditionalHelp(cmd *cobra.Command) bool {
+	return additionalHelp(cmd) != ""
+}
+
+var cyan = color("\033[1;36m%s\033[0m")
```
chris-crone, line 222: @nebuk89 can you confirm what is the right color? I think the spec said "light cyan" which might be "bright cyan" [here](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors), i.e.: 96

chris-crone (approved):
Two questions but LGTM

cli/cobra.go
nebuk89: Ah yes I meant bright sorry!

gtardif: Updated

gtardif:
@thaJeztah should be OK now I guess ; still dependent on merging the redirect in docs.docker.com, though (and @chris-crone comment on the trailing slash).

cli/cobra.go
gtardif: updated

chris-crone:
@thaJeztah Got the OK from ben on this. Can we remove the trailing slash?

nebuk89:
LGTM ship it!

thaJeztah (approved):
LGTM

tiborvass (approved):

gtardif:
https://github.com/docker/docker.github.io/pull/11812 has been merged, anything blocking this one or should it be merged ?  (@chris-crone @thaJeztah)

PR #2858, opened 2020-12-02
Allow publishing same port as UDP and TCP using docker stack deploy(T…

### LINKED ISSUE #2407: docker stack can't publish udp and tcp for the same port

adeniyistephen:
fixes #2407

- What I did
Simple fix for #2407, to ensure that TCP and UDP publishing for the same port doesn't go away during compose file merging. and added some testing.

- How I did it

I changed the key of the map from the port number to the full port specification (including published port number and protocol).

This may not be the ideal solution since it includes not just the protocol, but also the published port number in the map key. I did it this way because it's all I can figure out how do do with my meager golang skills. 

- How to verify it

Try the example in #2407.

fix bug: now possible to publish same port as TCP and UDP using docker stack deploy

### DIFF
--- a/cli/compose/loader/merge.go
+++ b/cli/compose/loader/merge.go
@@ -111,7 +111,7 @@ func toServicePortConfigsMap(s interface{}) (map[interface{}]interface{}, error)
 	}
 	m := map[interface{}]interface{}{}
 	for _, p := range ports {
-		m[p.Published] = p
+		m[p] = p
 	}
 	return m, nil
 }
--- a/cli/compose/loader/merge_test.go
+++ b/cli/compose/loader/merge_test.go
@@ -237,7 +237,10 @@ func TestLoadMultipleServicePorts(t *testing.T) {
 			name: "no_override",
 			portBase: map[string]interface{}{
 				"ports": []interface{}{
-					"8080:80",
+					Mode:      "ingress",
+					Published: 8080,
+					Target:    80,
+					Protocol:  "tcp",
 				},
 			},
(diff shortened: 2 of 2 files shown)


adeniyistephen:
@thaJeztah  I think i might not be able to pull off the testing tho,you can drop inputs or pointers on how to go about it.

adeniyistephen:
Reference to #2411

hikkyXII (approved):

PR #2869, opened 2020-12-07
docs: un-wrap line to prevent YAML docs from using "compact" formatting

thaJeztah:
Looks like the YAML conversion doesn't like lines starting with `[`, and
causing it to use the "compact" formatting in the generated YAML.

This patch un-wraps these lines to prevent this.

### DIFF
--- a/docs/reference/commandline/build.md
+++ b/docs/reference/commandline/build.md
@@ -402,14 +402,12 @@ the command line.
 ### Use a custom parent cgroup (--cgroup-parent)
 
 When `docker build` is run with the `--cgroup-parent` option the containers
-used in the build will be run with the 
-[corresponding `docker run` flag](../run.md#specify-custom-cgroups).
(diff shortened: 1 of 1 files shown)


thaJeztah:
@chris-crone @silvin-lubecki ptal

chris-crone (approved):
LGTM

chris-crone:
What a weird issue!

thaJeztah:
> What a weird issue!

Yes, so the YAML is still functional, but it generates the output as `"## Description\n\nfoo\n\nbar` (etc). While that makes the file smaller, it's not really human-readable, so it's hard to review changes in the YAML files.

Other things that result in that are `<tab>` characters, and trailing whitespaces in the document, so I try to avoid those in the files as well.

silvin-lubecki (approved):
LGTM

PR #2876, opened 2020-12-09
execv into context-specific binary is context type is not "" (default)

ndeloof:
**- What I did**
Detect the active context type, and if this one is not default (docker/moby) execv into a context specific binary

**- How I did it**
bake context-specific binary by appending context type to the current executable path.

**- How to verify it**
create an `ecs` context using docker/compose-cli
run `docker --context ECS foo` 
command will fail as you obviously have no `docker-ecs` binary to handle it.

Add support for alternate context types CLI

image

### DIFF
--- a/cli/command/context/create.go
+++ b/cli/command/context/create.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/context"
 	"github.com/docker/cli/cli/context/docker"
 	"github.com/docker/cli/cli/context/kubernetes"
 	"github.com/docker/cli/cli/context/store"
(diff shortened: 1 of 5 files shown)


cli/context/store/storeconfig.go
```diff
@@ -11,15 +11,15 @@ type NamedTypeGetter struct {
...
+// EndpointTypeGetter returns a NamedTypeGetter with the specified name and getter
 func EndpointTypeGetter(name string, getter TypeGetter) NamedTypeGetter {
 	return NamedTypeGetter{
 		name:       name,
 		typeGetter: getter,
 	}
 }
 
-// Config is used to configure the metadata marshaler of the context store
+// Config is used to configure the metadata marshaller of the context store
```
thaJeztah, line 22: I think both are correct https://en.wiktionary.org/wiki/marshaler, but Golang uses single `l` in its vocabulary, e.g. https://golang.org/pkg/encoding/#TextMarshaler

ndeloof: my IDE warned me about a (supposed to be) typo and proposed to fix it.
Didn't know about those distinct-but-valid forms.

cli/command/context/create.go
```diff
@@ -47,10 +47,13 @@ func longCreateDescription() string {
 func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	opts := &CreateOptions{}
 	cmd := &cobra.Command{
-		Use:   "create [OPTIONS] CONTEXT",
+		Use:   "create [OPTIONS] [TYPE] CONTEXT",
 		Short: "Create a context",
-		Args:  cli.ExactArgs(1),
+		Args:  cli.RequiresRangeArgs(1, 2),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 2 {
```
gtardif, line 54: At this level, we'll have to find a way to accept ACI/ECS specific flags, I believe it will fail before reaching this if you run `docker context create aci my context --subscription-id xxx`. 
Maybe it will be easier to defined the create subcommands we support and simply exec from there ? (duplicating commands, but ... ). 
For now at least, we want a specified list of context types we can create, having this in `--help` will be nicer too

gtardif: (same question for login)

ndeloof: oh yes indeed, this one will require a less-naive approach

ndeloof: I've found a hack, but it's really a crappy hack :'(

gtardif:
continued in #2894 , we should close this one

PR #2877, opened 2020-12-09
generate-man.sh: Only build md2man if it's not installed

Foxboron:
Distributions uses generate-man.sh to create man pages, however it only
assumes a container environment and attempts to build binaries towards
`/go` which might be an illegal path in `fakeroot`.

This patch ensures we only build `md2man` from the vendored sources if
the command is not present. This allows distributions to supply thier
packaged `md2man`.

Signed-off-by: Morten Linderud <morten@linderud.pw>

### DIFF
--- a/scripts/docs/generate-man.sh
+++ b/scripts/docs/generate-man.sh
@@ -4,8 +4,10 @@ set -eu -o pipefail
 
 mkdir -p ./man/man1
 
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM, thanks!

thaJeztah:

silvin-lubecki (approved):
LGTM

PR #2888, opened 2020-12-16
Rename bin/md2man to bin/go-md2man

elboulangero:
In the recent PR !2877, some code was added to check if md2man is
already installed in the build environment. This is to cater to the
needs of Linux distributions.

However it turns out that Linux distributions install md2man as
bin/go-md2man instead of bin/md2man, hence the PR !2877 doesn't help
much.

This commit fixes it by settling on using the binary name go-md2man.

For reference, here the file list of the package go-md2man in several
distributions:

- Debian: <https://packages.debian.org/sid/amd64/go-md2man/filelist>
- Ubuntu: <https://packages.ubuntu.com/hirsute/amd64/go-md2man/filelist>
- Fedora: <https://fedora.pkgs.org/31/fedora-x86_64/golang-github-cpuguy83-md2man-2.0.0-0.4.20190624gitf79a8a8.fc31.x86_64.rpm.html>
- ArchLinux: <https://www.archlinux.org/packages/community/x86_64/go-md2man/>

Signed-off-by: Arnaud Rebillout <elboulangero@gmail.com>

### DIFF
--- a/man/md2man-all.sh
+++ b/man/md2man-all.sh
@@ -18,5 +18,5 @@ for FILE in *.md; do
 		continue
 	fi
 	mkdir -p "./man${num}"
-	md2man -in "$FILE" -out "./man${num}/${name}"
+	go-md2man -in "$FILE" -out "./man${num}/${name}"
 done
--- a/scripts/docs/generate-man.sh
+++ b/scripts/docs/generate-man.sh
@@ -4,9 +4,9 @@ set -eu -o pipefail
 
 mkdir -p ./man/man1
 
-if ! command -v md2man &> /dev/null; then
+if ! command -v go-md2man &> /dev/null; then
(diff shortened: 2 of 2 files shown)


elboulangero:
@Foxboron what do you think of that?

Foxboron:
I realized this after I submitted the patch, awkwardly enough. I was wondering if it was better to maybe provide `export GO_MD2MAN=${GO_MD2MAN:-go-md2man}` so there is a possibility to override it in the case where people don't use `go-md2man`.

elboulangero:
Me, I would be in favor of hard-coding `go-md2man`, and then provide a way to override as you suggest only after someone asks for it. Let's wait for the docker devs to speak :)

thaJeztah (approved):
SGTM

thaJeztah:
@cpuguy83 ptal

silvin-lubecki (approved):
LGTM

PR #2894, opened 2020-12-18
based on context type, delegate to backend binary

gtardif:
**- What I did**
* Follow up work started at https://github.com/docker/cli/pull/2876, @ndeloof is on holidays for a while
* Changed context create ACI/ECS subcommands for more explicit delegation, will be nicer for help and easier to follow.
* replaced syscall.Exec (didn't work on my mac, and eventually we will want to control process and catch exit state)
* Added `docker login azure` to delegate to backend for Azure login

Fixes https://github.com/docker/dev-tooling-team/issues/247, 
Fixes https://github.com/docker/dev-tooling-team/issues/249, 

**- How I did it**

**- How to verify it**
* place [compose-cli v1.0.6](https://github.com/docker/compose-cli/releases/tag/v1.0.6) binary in `/usr/local/lib/docker/cli-backends/`
* display help of ECS subcommand `docker context create ecs --help` 
* or run commands against ACI with `docker --context acicontext run nginx` (requires an ACI context setup)
* `docker login azure` 

### DIFF
--- a/backends/manager.go
+++ b/backends/manager.go
@@ -1,6 +1,7 @@
 package backends
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -19,9 +20,10 @@ type Backend struct {
 	Err            error    `json:"-"`
 }
 
+// BackendMetadata backend metadata
 type BackendMetadata struct {
-	Name    string
-	Version string
+	Name    string `json:",omitempty"`
+	Version string `json:",omitempty"`
 }
 
 const (
@@ -29,17 +31,28 @@ const (
 	backendSuffix = "-backend"
 )
 
+var allowedBackends = map[string][]string{"docker-compose-cli-backend": {"aci", "ecs", "local"}}
+
 // ListBackends produces a list of available backends on the system and their context types
-func ListBackends() ([]Backend, error) {
-	return listBackendsFrom(getDockerCliBackendDir(), fakeMetadata)
+func ListBackends() []Backend {
+	return listBackendsFrom(getDockerCliBackendDir(), getMetadata, allowedBackends)
+}
+
+func getMetadata(binary string) (BackendMetadata, error) {
+	output, err := shellout(binary, "backend-metadata")
+	if err != nil {
+		return BackendMetadata{}, err
+	}
+	metadata := BackendMetadata{}
+	if err := json.Unmarshal(output, &metadata); err != nil {
+		return BackendMetadata{}, err
+	}
+	return metadata, nil
 }
 
 // GetBackend get backend for a given context type
 func GetBackend(contextType string) (*Backend, error) {
-	backends, err := ListBackends()
-	if err != nil {
-		return nil, err
-	}
+	backends := ListBackends()
 	for _, backend := range backends {
 		if utils.StrSliceContains(backend.SupportedTypes, contextType) {
 			return &backend, nil
@@ -50,18 +63,13 @@ func GetBackend(contextType string) (*Backend, error) {
 
 type extractMetadataFunc func(binary string) (BackendMetadata, error)
 
-func fakeMetadata(binary string) (BackendMetadata, error) {
-	name := strings.TrimPrefix(filepath.Base(binary), backendPrefix)
-	return BackendMetadata{Name: strings.TrimSuffix(name, ".exe"), Version: "1.0"}, nil
-}
-
-func listBackendsFrom(backendDir string, extractVersion extractMetadataFunc) ([]Backend, error) {
+func listBackendsFrom(backendDir string, extractMetadata extractMetadataFunc, allowedBackendFiles map[string][]string) []Backend {
 	if fi, err := os.Stat(backendDir); err != nil || !fi.IsDir() {
-		return nil, fmt.Errorf("%q is not a directory, unable to list backends", backendDir)
+		return nil
 	}
 	dentries, err := ioutil.ReadDir(backendDir)
 	if err != nil {
-		return nil, err
+		return nil
 	}
 	result := []Backend{}
 	for _, candidate := range dentries {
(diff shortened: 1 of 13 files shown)


ndeloof (approved):

cpuguy83:
Instead of requiring backends to be in a particular dir, having it in a particular namespace with PATH seems cleaner to me.

e.g. `docker-cli-context-azure-v1` could be the name of a binary that we can lookup via `exec.LookPath`.

ndeloof:
@cpuguy83 I would tend to agree with your suggestion, but proposed approach better matches the CLI plugin delegation mechanism.

cpuguy83:
Is this using the plugin delegation? It doesn't seem like it is (and probably shouldn't since it seems like a different case).

ndeloof:
Nope, just saying CLI plugin work this way, being retrieved from CLI_PLUGINS folder and not from PATH, so make sense to me backends adopt the same approach

gtardif:
Indeed, this does not use plugin delegation, but the model for finding binaries is very similar. 
We don't want to let backends be "overriden" by other binaries in the PATH just based on the binary name, so looking up only a specific folder helps with this. 
Also, backend binaries install this way is totally independent from PATH ; with the current compose-cli shelling out to com.docker.cli, we encountered a few PATH issues with specific user setup. 
The plugin mechanism is quite proven with buildx (& more recently scan), trying to do something very similar will avoid discovering new setup issues.

cpuguy83:
@gtardif Not surprising that a user had an issue with the cute things done to make compose-cli work.

re: /usr/local/*
Packagers are not supposed to install things here.
/usr/lib/* is ok for packagers.

The binary namespacing is somewhat nondescript and will almost certainly cause problems down the line.

This whole thing is also pretty boldly mixing the concept of a context (transport) with an API implementation, whereas today context is only about transport (unfortunately there is already mixing of transport and protocol).

cpuguy83:
From a security perspective on ACI, I'd rather not giving the docker CLI full reign on all the subscriptions I have access to, but `docker login azure` seems to do this. This is actually really bad.

I see I can create a service principal out of band and set the SP details on `docker login azure`, but the default means setting up an ACI context is just flat out dangerous.

gtardif:
@cpuguy83 regarding `docker login azure`, I suppose this is actually unrelated to this PR, could you open a ticket in docker/compose-cli regarding this ? (And to be discussed with @nebuk89, @bigMorty and @macolso). 
re: concept of a context, this is also something to be discussed in compose-cli I believe, contexts have been used this way for the ACI / ECS integration (you cal also cc @nebuk89 on this subject). 

Regarding `/usr/local/*` vs `/usr/lib/*` I don't have a strong opinion on this, I wanted to align with where plugins like buildx are put ATM. 
Maybe @thaJeztah or @silvin-lubecki will have an opinion if we keep these in `/usr/local/lib` to keep things aligned with buildx and other plugins, or if we prefer to use `/usr/lib/... `.

backends/manager.go
```diff
@@ -0,0 +1,102 @@
...
+const (
+	backendPrefix = "docker-"
+	backendSuffix = "-backend"
+)
+
+var allowedBackends = map[string][]string{"docker-compose-cli-backend": {"aci", "ecs", "local"}}
+
+// ListBackends produces a list of available backends on the system and their context types
+func ListBackends() []Backend {
+	return listBackendsFrom(getDockerCliBackendDir(), getMetadata, allowedBackends)
```
silvin-lubecki, line 38: question: Shouldn't this function return an error? As it looks like there are many cases when `listBackendsFrom` returns an empty slice, for example if `os.Stat` on the backend directory returns an error. This error is ignored and maybe it could help the user to understand why we can't find any backend?

backends/manager.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+		filename := candidate.Name()
+		path := filepath.Join(backendDir, filename)
+		withoutExe := strings.TrimSuffix(filename, ".exe")
+		if !strings.HasPrefix(withoutExe, backendPrefix) || !strings.HasSuffix(withoutExe, backendSuffix) {
+			continue
+		}
+		contextTypes, allowed := allowedBackendFiles[withoutExe]
+		if !allowed {
+			fmt.Fprintf(os.Stderr, "Invalid backend : backend binary %q is not allowed", filename)
```
silvin-lubecki, line 91: nit: I wonder if we should print here on os.Stderr, or just log that in debug so it would be printed only if the user sets the `--debug` flag. 🤔

backends/manager.go
gtardif: Right, I returned an error and log the error meg to debug if there is one, that will help debuging when something goes wrong

backends/manager.go
gtardif: moved to debug, inline with previous comment

backends/shellout.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	return output, nil
+}
+
+func delegate(execBinary string) {
+	cmd := exec.Command(execBinary, os.Args[1:]...)
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Env = os.Environ()
```
silvin-lubecki, line 24: nit: I wonder if this line is useful 🤔

cli/cobra.go
```diff
@@ -161,6 +161,20 @@ func (tcmd *TopLevelCommand) Initialize(ops ...command.InitializeOpt) error {
 	return tcmd.dockerCli.Initialize(tcmd.opts, ops...)
 }
 
+// ContextType return the current context type, "" for default
```
silvin-lubecki, line 164: question: just thinking out of loud, but what about returning "engine", "docker" or "moby" by default?

cli/command/context/create.go
```diff
@@ -68,6 +72,30 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	return cmd
 }
 
+func newAciCreateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                "aci CONTEXT [flags]",
```
silvin-lubecki, line 77: nit: Should be `aci [OPTIONS] CONTEXT`

cli/command/context/create.go
```diff
@@ -68,6 +72,30 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return backends.RunBackendCLI(backends.ContextTypeACI)
+		},
+	}
+	return cmd
+}
+
+func newEcsCreateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                "ecs CONTEXT [flags]",
```
silvin-lubecki, line 89: nit: Should be `ecs [OPTIONS] CONTEXT`

cli/command/system/version.go
```diff
@@ -35,7 +36,11 @@ Client:{{if ne .Platform.Name ""}} {{.Platform.Name}}{{end}}
  OS/Arch:	{{.Os}}/{{.Arch}}
  Context:	{{.Context}}
  Experimental:	{{.Experimental}}
-{{- end}}
+ {{- range $backend := .Backends}}
+ {{$backend.Name}}:
```
silvin-lubecki, line 40: question: Should we add a sub section title `Backends` ?

backends/shellout.go
gtardif: indeed, if nil uses the parent process env :)

cli/command/context/create.go
gtardif: fixed

cli/command/context/create.go
gtardif: fixed

cli/command/system/version.go
gtardif: Good question, currently this looks like: 
```
$ ./build/docker-darwin-amd64 version
Client:
 Version:           20.10.0-dev
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        73cbd63ff
 Built:             Fri Jan  8 16:34:46 2021
 Context:           default
 Experimental:      true
 Cloud integration:
  Version:          v1.0.6-16
```
Ideally I would prefer to display a single line for each backend and version, but the backend name is unknown in the template, making it hard to align the 2nd column. Also, we could have more info later on about backends (supported context type, ...).

cli/cobra.go
gtardif: This part will have to change when rebasing this PR on top of https://github.com/docker/cli/pull/2895, adding a cleaner way to get context type.

gtardif:
Closing this, will reopen/fetch code later on if required

Issue #2718, opened 2020-09-14
Add `daemon` or `engine` subcommand


cpuguy83:
In the past, `docker` had the daemon built-in as a single binary under the `daemon` subcommand (or if you go far enough back just the `-d` flag as I recall).

I propose that we add this subcommand back... only better.
The subcommand would be responsible for setting up the daemon on the client node.

On linux this would be starting a daemon as normal.
If `docker` is run as unprivileged, it could be responsible for setting up all the rootless stuff.
Perhaps could be rootless by default or even rootless only (since it is the client after all).

On non-native platforms this could also be extended to support other cases such as spinning up a platform-native VM to run the daemon, setup socket access, etc.
But anything that we don't actually support the subcommand would just error out as not supported.

In any case this would also configure a context to connect to the daemon.

Honestly this is not a concrete proposal, just exploring some ideas.

I think having the rootless experience tied in with the CLI would be extremely interesting.

cpuguy83:
Ping @AkihiroSuda re: rootless

AkihiroSuda:
We already dockerd-rootless-setuptool.sh for setting up the systemd unit for the daemon

https://github.com/moby/moby/blob/master/contrib/dockerd-rootless-setuptool.sh

Does it cover the use case?

cpuguy83:
@AkihiroSuda Not really. As a user I want polish, I don't want to hunt down other things for making it work. I just want it to work.

I could even see having a special context for rootless (like `default` is special) and `docker run` could even handle the case of no daemon being up.

AkihiroSuda:
The "default" context SGTM, then we don't need "daemon" sub command?

I'd expect it to work like this

```
$ docker run hello-world
Docker daemon isn't running, would you want to:
1) Set up the daemon as the root
2) Set up the daemon as non-root user
3) Connect to a remote daemon

Choose 1 if you don't have an idea.

Enter the number: 2 [RET]

Running `dockerd-rootless-setuptool.sh`...done
Creating a new context "rootless" for unix://run/user/1000/docker.sock
Setting "rootless" as the new default context

Running `docker run hello-world `

Hello from Docker
```

cpuguy83:
Toying around with this a bit over the weekend.
Currently the UX would be something like:

```
docker context create --docker user my_user_context
```

Anything context created with `--docker user` will trigger, at runtime, a connection to a a local daemon managed by the CLI.
[...]

PR #2895, opened 2020-12-18
Enable type into the reading of the contexts

ulyssessouza:
**- What I did**
Added the type of the context to the listing.
Note that this only include the reading&listing of the contexts. Not it's creation or update.

**- How to verify it**
```shell
$ docker context list
```

```
Add "type" for contexts.
```

type


gtardif (approved):
LGTM

cli/command/context.go
```diff
@@ -37,8 +42,11 @@ func (dc *DockerContext) UnmarshalJSON(payload []byte) error {
 	if err := json.Unmarshal(payload, &data); err != nil {
 		return err
 	}
+	dc.Type = DefaultContextType
```
thaJeztah, line 45: Wondering if it would make sense to use `""` (empty string) as default. i.e., if no type is present, use the current default, and only require a value to be set for non-default. (We could still print "default" or "moby" when doing a `docker context ls` if we want)

gtardif: Yes, we can let "" (empty) be the default type especially for storing contexts, and only replace it when outputting context type to users, that's maybe closer to what already exists in the field for now

ulyssessouza: Cool!
PTAL

gtardif (approved):

ulyssessouza:
@thaJeztah PTAL

silvin-lubecki (approved):
LGTM

ndeloof (approved):

e2e/context/testdata/context-ls.golden
```diff
@@ -1,3 +1,3 @@
-NAME        DESCRIPTION                               DOCKER ENDPOINT               KUBERNETES ENDPOINT            ORCHESTRATOR
-default *   Current DOCKER_HOST based configuration   unix:///var/run/docker.sock   https://someserver (zoinx)     swarm
-remote      my remote cluster                         ssh://someserver              https://someserver (default)   kubernetes
+NAME        TYPE      DESCRIPTION                               DOCKER ENDPOINT               KUBERNETES ENDPOINT            ORCHESTRATOR
```
thaJeztah, line 1: Can you update the example output in https://github.com/docker/cli/blob/master/docs/reference/commandline/context_ls.md#examples to match the new format?

thaJeztah: Probably the `docker context inspect` example output should also be updated; https://github.com/docker/cli/blob/master/docs/reference/commandline/context_inspect.md#inspect-a-context-by-name

cli/command/formatter/context.go
```diff
@@ -73,6 +75,10 @@ func (c *clientContextContext) Name() string {
 	return c.c.Name
 }
 
+func (c *clientContextContext) Type() string {
+	return c.c.Type
```
thaJeztah, line 79: Thinking what would make more sense; set the default, or have this function return the default if it's empty 🤔

thaJeztah (changes requested):

cli/command/defaultcontextstore.go
```diff
@@ -14,6 +14,8 @@ import (
 const (
 	// DefaultContextName is the name reserved for the default context (config & env based)
 	DefaultContextName = "default"
+	// DefaultContextType is the default type for a context
+	DefaultContextType = "moby"
```
thaJeztah, line 18: /cc @cpuguy83 @tiborvass just to double-check; are we good with `moby` as name for the default? (or do we want it to be called different, e.g. "default" literally?)

cpuguy83: It's weird, because we don't even refer to moby as moby in moby.
It's also not very descriptive.

cpuguy83: I'd also like to have a context type for https://github.com/docker/cli/issues/2718#issuecomment-700162376

ndeloof: DefaultContextType = "default" would sound weird to me, as a non descriptive name. Should be "docker engine" but considering the moby/docker naming split I'm fine with "moby"

cpuguy83: Could be "engine".
The problem is this encodes more than just the api implementation.

cpuguy83: "com.docker.engine.v1"

thaJeztah: Alternatively, use an empty string instead? (only show a backend if it's not the default (docker engine))?

gtardif: "moby" is what is currently returned as context type, for users who have installed Docker Desktop or Linux users having installed compose-cli binary. So here we want to just prepare replacing compose-cli current installation with CLI backend mechanism, it's probably better to not change the current behaviour ?

cpuguy83: On the contrary, it seems like a great time to get it right.

e2e/context/testdata/context-ls.golden
ulyssessouza: Done

PR #2902, opened 2020-12-31
configFile: directly write changes to configFile

clwluvw:
In some situations, the config.json is volumed from somewhere else and the file is busy, move can't be performed on this file and configFile.Save() will be failed because it's trying to move another file to this file. This will write changes directly to the file.

Signed-off-by: Seena Fallah <seenafallah@gmail.com>

**- What I did**
Write config file changes directly to the file.

**- How to verify it**
Volume docker config in a `slave` mode and try to change it with `docker login` inside the container

Perform config file changes even if the file is busy

image


thaJeztah (changes requested):
Hmmm... not sure about this change. The temp-file was added for a reason, and to make sure that writes would be preformed atomically (see https://github.com/docker/cli/pull/1359). Removing this would mean we introduce that problem again.

What other process is using the file in your case that prevents it from being accessed?

clwluvw:
> What other process is using the file in your case that prevents it from being accessed?

@thaJeztah The file is volumed to a container and It can't be removed so moving another file to this file will be failed. For example, mount `~/.docker/config.json` from your host to the container and then run `docker login` inside the container and it will be failed!

clwluvw:
> Hmmm... not sure about this change. The temp-file was added for a reason, and to make sure that writes would be preformed atomically (see #1359). Removing this would mean we introduce that problem again.

@thaJeztah I used `flock` to handle concurrent writes to the config file. What's your idea?

PR #2903, opened 2021-01-01
Fix docker plugin suggestions in Zsh completion

### LINKED ISSUE #2761: zsh completion: bad math expression in __docker_plugins function
**Description**

1. Run zsh with docker installed and the docker completion working.
2. Make sure there's at least one docker plugin installed so that `docker plugin ls` shows at least one row. For example: `docker plugin install vieux/sshfs`.
3. Try to complete `docker inspect <TAB>`

The shell outputs multiple `__docker_plugins:27: bad math expression: empty string` error messages and the completion entries are all garbled.

A list of Docker objects available to be inspected.

I've tracked down the problem to the [`__docker_plugins` function](https://github.com/docker/cli/blob/88c6089300a82d3373892adf6845a4fed1a4ba8d/contrib/completion/zsh/_docker#L1585), which expects the output of `docker plugin ls` to have a `TAG` column, which it doesn't. I believe the solution is to use the `ID` column instead, but I haven't used docker plugins before so I may be wrong.

I've also seen that just modifying the function to use the `ID` column isn't enough to make the results of the function correct, since it also tries to truncate the ID hash to 7 characters, but it truncates the beginning characters of the string, not the ending ones. This is I believe what would make the function work correctly:

```diff
diff --git a/contrib/completion/zsh/_docker b/contrib/completion/zsh/_docker
index 4d28ac4..82a77de 100644
--- a/contrib/completion/zsh/_docker
+++ b/contrib/completion/zsh/_docker
@@ -1582,7 +1582,7 @@ __docker_plugins() {
     # Name
     for line in $lines; do
         s="${line[${begin[NAME]},${end[NAME]}]%% ##}"
-        s="$s:${(l:7:: :::)${${line[${begin[TAG]},${end[TAG]}]}%% ##}}"
+        s="$s:${(r:7:: :::)${${line[${begin[ID]},${end[ID]}]}%% ##}}"
         plugins=($plugins $s)
     done
```

Related issue: https://github.com/ohmyzsh/ohmyzsh/issues/9294

```
 Version:           19.03.12
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        48a66213fe
 Built:             Mon Jun 22 15:45:47 2020
 Experimental:      false

 Engine:
  Version:          19.03.12
  Go version:       go1.13.10
  Git commit:       48a66213fe
  Built:            Mon Jun 22 15:44:17 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Output of `docker info`:**

```
Client:
 Debug Mode: false

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 16
 Server Version: 19.03.12
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc version: dc9208a3303feef5b3839f4323d9beb36df0a9dd
 init version: fec3683
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 4.19.128-microsoft-standard
 Operating System: Ubuntu 19.10
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 3.803GiB
 Name: WATSON
 ID: 3F54:JRUE:ZIAJ:KJ42:DQOE:TLU6:774I:JZKT:5I3L:HAJX:US4R:NHFE
 Debug Mode: false
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

Using Ubuntu 19.10 on WSL2 but I don't think it makes a difference.

mcornella:
Fixes #2761

**- What I did**

The `__docker_plugins` function calls `docker plugin ls` to get its information, but it didn't parse it correctly before, so I fixed it. Without this change, triggering the function would produce the following errors:

```
__docker_plugins:27: bad math expression: empty string
__docker_plugins:27: bad math expression: empty string
__docker_plugins:27: bad math expression: empty string
__docker_plugins:27: bad math expression: empty string
__docker_plugins:27: bad math expression: empty string
```

**- How I did it**

The `docker plugin ls` command doesn't output a `TAG` column; instead, its name is `ID`. So one of the changes is to replace `TAG` for `ID`. Also, the relevant part of ID hash is the left one, so we need to use right-padding to trim it to 7 characters. With `(l:7:: :::)` what we get is the last 7 characters of the hash, which is not what we want.

**- How to verify it**

1. If you don't have docker plugins installed, install at least 2. For example:
   ```sh
   docker plugin install vieux/sshfs
   docker plugin install grafana/loki-docker-driver:latest --alias loki --grant-all-permissions
   ```
2. Run `docker plugin ls` to see the output:
   ```console
   $ docker plugin ls
   ID             NAME                 DESCRIPTION               ENABLED
   a237ec8092c3   loki:latest          Loki Logging Driver       true
   84dd2ad7d54e   vieux/sshfs:latest   sshFS plugin for Docker   true
   ```
3. Make sure to be using this `_docker` file for completion. In Zsh, download this `_docker` file and run this to load it:
   ```zsh
   unfunction _docker
   autoload -Uz /path/to/downloaded/_docker/file
   ```
4. Tab-complete `docker plugin inspect`:
   ```console
   $ docker plugin inspect [TAB]
   loki         -- latest:a237ec8
   vieux/sshfs  -- latest:84dd2ad
   ```

Fix `docker plugin` Zsh completion by correctly parsing the output of `docker plugin ls`.

cute sleeping baby donkeys


mcornella:
Turns out I was wrong. Took making a PR to realise that 😅

The problem is that `docker plugin ls` used to provide a TAG column with the plugin tag name, but now it's is incorporated into the NAME column. For example, if we install both the `latest` and `1.3` tags of `vieux/sshfs`, this is what `docker plugin ls` looks like:

```sh
docker plugin install vieux/sshfs:latest
docker plugin install vieux/sshfs:1.3
```

```console
$ docker plugin ls
ID             NAME                 DESCRIPTION               ENABLED
a237ec8092c3   loki:latest          Loki Logging Driver       true
23dd56eff145   vieux/sshfs:1.3      sshFS plugin for Docker   true
84dd2ad7d54e   vieux/sshfs:latest   sshFS plugin for Docker   true
```

With the current proposed change, this is what completion suggests:
```console
$ docker plugin inspect [TAB]
loki         -- latest:a237ec8
vieux/sshfs  -- 1.3:23dd56e
```

Note that the tag (latest, 1.3) appears in the right side of the completion entry, and `vieux/sshfs` is only presented once, while two entries should appear, one for each tag. This happens because `:` is a special character used by `_describe` to provide a comment for a completion entry, so instead of appearing in the suggested completion it's parsed by `_describe`. If we quote the colon, then the tag correctly appears in the suggestion:

```patch
diff --git a/contrib/completion/zsh/_docker b/contrib/completion/zsh/_docker
index a7d94a3..ee32e82 100644
--- a/contrib/completion/zsh/_docker
+++ b/contrib/completion/zsh/_docker
@@ -1581,7 +1581,7 @@ __docker_plugins() {
 
     # Name
     for line in $lines; do
-        s="${line[${begin[NAME]},${end[NAME]}]%% ##}"
+        s="${${line[${begin[NAME]},${end[NAME]}]%% ##}//:/\:}"
         s="$s:${(r:7:: :::)${${line[${begin[ID]},${end[ID]}]}%% ##}}"
         plugins=($plugins $s)
     done
```

```console
$ docker plugin inspect [TAB]
loki:latest         -- a237ec8
vieux/sshfs:1.3     -- 23dd56e
vieux/sshfs:latest  -- 84dd2ad
```

Also, the line in the `for` that tries to find the TAG column can be removed, since it doesn't provide any relevant information. This is how the completion now looks like:

```console
$ docker plugin inspect [TAB]
loki:latest         vieux/sshfs:1.3     vieux/sshfs:latest
```

I'll redo the original PR post to match this recent changes.

contrib/completion/zsh/_docker
```diff
@@ -1581,8 +1581,7 @@ __docker_plugins() {
 
     # Name
     for line in $lines; do
-        s="${line[${begin[NAME]},${end[NAME]}]%% ##}"
-        s="$s:${(l:7:: :::)${${line[${begin[TAG]},${end[TAG]}]}%% ##}}"
+        s="${${line[${begin[NAME]},${end[NAME]}]%% ##}//:/\:}"
```
thaJeztah, line 1584: Instead of using the standard output of `docker plugin ls`, would it perhaps work better to use `--format`, to output the information in a way it's most usable for the completion script?

mcornella: [Lines 1569 to 1580](https://github.com/docker/cli/pull/2903/files#diff-474a9d942d53cc2279d5641e4a9dcfb18958e412ad53940b53423a38033857d6R1569-R1580) could be removed entirely yes. Would displaying NAME be enough (1) or would you prefer having other completion information, like DESCRIPTION (2)? Also should I remove the `:latest` tag like in (2)?

1. ```console
   $ docker plugin inspect [TAB]
   loki:latest         vieux/sshfs:1.3     vieux/sshfs:latest
   ```
2. ```console
   $ docker plugin inspect [TAB]
   loki                          -- Loki Logging Driver
   vieux/sshfs:1.3  vieux/sshfs  -- sshFS plugin for Docker
   ```

mcornella: I went ahead and submitted (2) version.

mcornella:
Rebased on top of the current master. @thaJeztah this problem still exists, so this PR is still relevant.

obvionaoe:
It would help if someone could merge this! @thaJeztah

mcornella:
@thaJeztah gentle ping. This is still an issue 😊

toby-griffiths:
Is this still being considered, as I'm also suffering from this bug?

Issue #2890, opened 2020-12-17
Non-root user "docker login" results in panic SIGSEGV segmentation violation


brenc:
**Description**

Fresh install on Debian Buster. Using a non-root user who is a member of the docker group I started getting a panic when running `docker login` for the first time after install. This seems to be a new issue.

**Steps to reproduce the issue:**
1. Create a new user who is a member of the docker group
2. Run `docker login`

**Describe the results you received:**

```
$ docker login
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x55b37f2d0406]

goroutine 1 [running]:
github.com/docker/cli/cli/command.ConfigureAuth(0x55b380d12fe0, 0xc000505520, 0x0, 0x0, 0x0, 0x0, 0x0, 0x55b380c82401, 0xc0000812d0, 0xc000727968)
        /go/src/github.com/docker/cli/cli/command/registry.go:128 +0x46
github.com/docker/cli/cli/command/registry.runLogin(0x55b380d12fe0, 0xc000505520, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /go/src/github.com/docker/cli/cli/command/registry/login.go:123 +0x223
github.com/docker/cli/cli/command/registry.NewLoginCommand.func1(0xc0006a22c0, 0x55b381d1d5c8, 0x0, 0x0, 0x0, 0x0)
        /go/src/github.com/docker/cli/cli/command/registry/login.go:45 +0xcc
github.com/docker/cli/vendor/github.com/spf13/cobra.(*Command).execute(0xc0006a22c0, 0xc0006915c0, 0x0, 0x0, 0xc0006a22c0, 0xc0006915c0)
        /go/src/github.com/docker/cli/vendor/github.com/spf13/cobra/command.go:850 +0x462
github.com/docker/cli/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc00050fb80, 0xc0006915c0, 0x1, 0x1)
        /go/src/github.com/docker/cli/vendor/github.com/spf13/cobra/command.go:958 +0x34b
github.com/docker/cli/vendor/github.com/spf13/cobra.(*Command).Execute(...)
        /go/src/github.com/docker/cli/vendor/github.com/spf13/cobra/command.go:895
main.runDocker(0xc000505520, 0x55b380c85120, 0xc000010028)
[...]

PR #2818, opened 2020-10-29
GetDefaultAuthConfig: fix potential panic due to unhandled error

thaJeztah:
Saw my IDE was complaining about this, and looks like it could panic

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #2918, opened 2021-01-07
Fix panic when failing to get DefaultAuthConfig

thaJeztah:
fixes https://github.com/docker/cli/issues/2890 Non-root user "docker login" results in panic SIGSEGV segmentation violation
fixes https://github.com/moby/moby/issues/4177 Command 'docker login' does not work

Commit f32731f9020f12ae2600a3f4c90e668565220cb6 (https://github.com/docker/cli/pull/2818) fixed a potential panic when an error was returned while trying to get existing credentials.

However, other code paths currently use the result of `GetDefaultAuthConfig()` even in an error condition; this resulted in a panic, because a `nil` was returned.

We should fix the other code paths to be more resilient, and to not use the result of `GetDefaultAuthConfig()` if an error occurs, but starting with a minimal patch to fix the immediate issue.

```markdown
- Fix a panic on `docker login` if no config file is present
```

### DIFF
--- a/cli/command/registry.go
+++ b/cli/command/registry.go
@@ -63,6 +63,9 @@ func RegistryAuthenticationPrivilegedFunc(cli Cli, index *registrytypes.IndexInf
 		indexServer := registry.GetAuthConfigKey(index)
 		isDefaultRegistry := indexServer == ElectAuthServer(context.Background(), cli)
 		authConfig, err := GetDefaultAuthConfig(cli, true, indexServer, isDefaultRegistry)
+		if authConfig == nil {
+			authConfig = &types.AuthConfig{}
+		}
 		if err != nil {
 			fmt.Fprintf(cli.Err(), "Unable to retrieve stored credentials for %s, error: %s.\n", indexServer, err)
 		}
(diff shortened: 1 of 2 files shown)


thaJeztah:
For reference; the problem situation is here; https://github.com/docker/cli/blob/88c6089300a82d3373892adf6845a4fed1a4ba8d/cli/command/registry/login.go#L118-L123

If `GetDefaultAuthConfig` fails (line 118), the `authConfig` variable is still used (line 120 and/or line 123); those functions expect `authConfig` to be a `*types.AuthConfig`, and fail if it's `nil`. We should look at those code-paths, because it's incorrect to use `authConfig` if `err != nil`, but we can do so in a follow-up.

thaJeztah:

thaJeztah:
Discussing with @tiborvass @cpuguy83 - I'll update this to do a nil check outside of the function

thaJeztah:
@tiborvass @cpuguy83 updated; PTAL

cpuguy83 (approved):
LGTM

thaJeztah:
@silvin-lubecki PTAL

silvin-lubecki (approved):
LGTM

Issue #2882, opened 2020-12-14
TestRemoveForce flaky test?


elboulangero:
### Describe the bug

While running unit tests:

```
=== Failed
=== FAIL: cli/command/container TestRemoveForce/with_force (0.00s)
    rm_test.go:44: assertion failed: 
        --- removed
        +++ →
          []string{
          	"mycontainer",
        + 	"nosuchcontainer",
          }
        
    --- FAIL: TestRemoveForce/with_force (0.00s)

=== FAIL: cli/command/container TestRemoveForce (0.00s)
```

Context: building the package on the Debian infrastructure.

The issue happened on the `s390x` architecture, for the version `20.10.0`. Cf: https://buildd.debian.org/status/fetch.php?pkg=docker.io&arch=s390x&ver=20.10.0%2Bdfsg1-1&stamp=1607657904

During previous builds (version `20.10.0-rc1`), there was no issue.

I've seen this failed on the `mips` architecture as well, although I have no logs. And the last time I tried, the test passed.

All of that leads me to believe that the test might be flaky. However the code looks quite straightforward, so I'm a bit puzzled.

For background, this test was introduced in 9a071a993a7d15cc490937104b3d37fd6a339bc2

elboulangero:
I just had this test failure again, this time on `amd64` architecure: https://buildd.debian.org/status/fetch.php?pkg=docker.io&arch=amd64&ver=20.10.2%2Bdfsg1-1&stamp=1610074719

This time it's the `without_force` test which fails:

```
=== Failed
=== FAIL: cli/command/container TestRemoveForce/without_force (0.00s)
Error: Error: No such container: nosuchcontainer
    rm_test.go:37: assertion failed: 
        --- removed
        +++ →
          []string{
        + 	"mycontainer",
          	"nosuchcontainer",
          }
        
    --- FAIL: TestRemoveForce/without_force (0.00s)

=== FAIL: cli/command/container TestRemoveForce (0.00s)
```

thaJeztah:
Are tests ran with `-parallel`? The `removed` variable is global to both subtests; if they run in parallel, that would possibly explain this 🤔

elboulangero:
> Are tests ran with -parallel?

No, we run these tests with the Makefile provided, ie. `make test-unit`. In the logs, we can see that it translates to `gotestsum  -- ${TESTDIRS:-github.com/docker/cli/cli [...] github.com/docker/cli/templates}`. No `-parallel`.

[...]

PR #2929, opened 2021-01-13
Ignore SIGURG on Linux.

cpuguy83:
In go1.14+, SIGURG is used by the runtime to handle preemtable system
calls.
In practice this signal is caught *frequently*.

For reference:

https://go.googlesource.com/proposal/+/master/design/24543-non-cooperative-preemption.md
https://github.com/golang/go/issues/37942

### DIFF
--- a/cli/command/container/attach.go
+++ b/cli/command/container/attach.go
@@ -97,7 +97,8 @@ func runAttach(dockerCli command.Cli, opts *attachOptions) error {
 	}
 
 	if opts.proxy && !c.Config.Tty {
-		sigc := ForwardAllSignals(ctx, dockerCli, opts.container)
+		sigc := notfiyAllSignals()
+		go ForwardAllSignals(ctx, dockerCli, opts.container, sigc)
 		defer signal.StopCatch(sigc)
 	}
 
--- a/cli/command/container/client_test.go
+++ b/cli/command/container/client_test.go
@@ -32,6 +32,7 @@ type fakeClient struct {
 	containerExportFunc     func(string) (io.ReadCloser, error)
 	containerExecResizeFunc func(id string, options types.ResizeOptions) error
 	containerRemoveFunc     func(ctx context.Context, container string, options types.ContainerRemoveOptions) error
+	containerKillFunc       func(ctx context.Context, container, signal string) error
 	Version                 string
 }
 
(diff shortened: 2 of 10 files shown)


silvin-lubecki (commented):
Thank you for this PR @cpuguy83 👍  Code looks good, but could we add a unit test on this new behavior ? 🙏

cpuguy83:
I'm not sure we can.

1) Since it is dealing with signals it is a bit more tricky in general since it is a global resource 
2) Once we upgrade go, it is even more tricky since go used the signal internally.

silvin-lubecki:
I wonder if we can play with the channel returned by the `ForwardAllSignals`, sending the signal via this way? 
```go
ch := ForwardAllSignals(ctx, fakeCli, "id")
ch <- unix.SIGURG
// check fakeCli.ContainerKill was not called
```
WDYT?

cpuguy83:
I think unless we pass in a `<-chan os.Signal` (which we won't wire up to the real signals) to `ForwardAllSignals` we can easily end up with a blocked channel after upgrading go in the test.

thaJeztah:
Would (some of) these be useful to be included in https://github.com/moby/moby/tree/master/pkg/signal ? Or too unrelated to that package?

cpuguy83:
We could probably add an `except ...os.Signal` to the `CatchAll` function.

cpuguy83:
But I really wanted to not modify that package to fix the CLI because dependency hell.

cpuguy83:
Updated this to add a couple of tests.
I did have to refactor `ForwardAllSignals` to receive a channel instead of returning one.

silvin-lubecki (approved):
LGTM

silvin-lubecki:
@cpuguy83 the linter is complaining:
```
cli/command/container/signals_test.go:38:17: SA1025: it is not possible to use Reset's return value correctly, as there is a race condition between draining the channel and the new timer expiring (staticcheck)
	if !timer.Reset(30 * time.Second) {
```

cpuguy83:
Fixed lint issue.

Now flakey test?
```
=== Failed
=== FAIL: cli/command/container TestRemoveForce/without_force (0.00s)
Error: Error: No such container: nosuchcontainer
    --- FAIL: TestRemoveForce/without_force (0.00s)
        rm_test.go:37: assertion failed: 
            --- removed
            +++ →
            {[]string}[?->0]:
            	-: <non-existent>
            	+: "mycontainer"
```

thaJeztah:
ah, interesting; it was reported in https://github.com/docker/cli/issues/2882 to be flaky; so far don't have a clue yet why (perhaps you can spot the issue?)

thaJeztah:
kicked CI

thaJeztah (approved):
LGTM

PR #2934, opened 2021-01-18
cli/config: prevent warning if HOME is not set

thaJeztah:
fixes https://github.com/moby/moby/issues/41890 WARNING: Error loading config file: .dockercfg: $HOME is not defined

commit c2626a8270924d0cec54477fc506174cd31dc560 (https://github.com/docker/cli/pull/2101) replaced the use of github.com/docker/docker/pkg/homedir with Golang's `os.UserHomeDir()`.

This change was partially reverted in 7a279af43de6650d2f36c41f9d926db7be3b3567 (https://github.com/docker/cli/pull/2111) to account for situations where `$HOME` is not set.

In  situations where no configuration file is present in `~/.config/`, the CLI falls back to looking for the (deprecated) `~/.dockercfg` configuration file, which was still using `os.UserHomeDir()`, which produces an error/warning if `$HOME` is not set.

This patch introduces a helper function and a global variable to get the user's home-directory. The global variable is used to prevent repeatedly looking up the user's information (which, depending on the setup can be a costly operation).

```
- Fix "`WARNING: Error loading config file: .dockercfg: $HOME is not defined`"
```

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -26,7 +26,21 @@ const (
 var (
 	initConfigDir sync.Once
 	configDir     string
+	homeDir       string
 )
+
+// resetHomeDir is used in testing to resets the "homeDir" package variable to
+// force re-lookup of the home directory between tests.
+func resetHomeDir() {
+	homeDir = ""
+}
+
+func getHomeDir() string {
+	if homeDir == "" {
+		homeDir = homedir.Get()
+	}
+	return homeDir
+}
 
 func setConfigDir() {
 	if configDir != "" {
@@ -34,7 +48,7 @@ func setConfigDir() {
 	}
 	configDir = os.Getenv("DOCKER_CONFIG")
 	if configDir == "" {
-		configDir = filepath.Join(homedir.Get(), configFileDir)
+		configDir = filepath.Join(getHomeDir(), configFileDir)
 	}
 }
 
@@ -109,11 +123,7 @@ func Load(configDir string) (*configfile.ConfigFile, error) {
 	}
 
 	// Can't find latest config file so check for the old one
-	home, err := os.UserHomeDir()
-	if err != nil {
-		return configFile, errors.Wrap(err, oldConfigfile)
-	}
-	filename = filepath.Join(home, oldConfigfile)
+	filename = filepath.Join(getHomeDir(), oldConfigfile)
 	if file, err := os.Open(filename); err == nil {
 		defer file.Close()
 		if err := configFile.LegacyLoadFromReader(file); err != nil {
--- a/cli/config/config_test.go
+++ b/cli/config/config_test.go
@@ -115,6 +115,7 @@ password`: "Invalid Auth config file",
 email`: "Invalid auth configuration file",
 	}
 
+	resetHomeDir()
 	tmpHome, err := ioutil.TempDir("", "config-test")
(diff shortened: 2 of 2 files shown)


silvin-lubecki (approved):
LGTM

silvin-lubecki:
@thaJeztah some tests are failing:
```
=== Failed
=== FAIL: cli/config TestOldInvalidsAuth (0.00s)
    config_test.go:129: assertion failed: expected an error, got nil

=== FAIL: cli/config TestOldValidAuth (0.00s)
    config_test.go:150: assertion failed:  (ac.Username string) != joejoe (string)

=== FAIL: cli/config TestOldJSONInvalid (0.00s)
    config_test.go:182: Expected an error got : &{map[] map[]         map[] /tmp/config-test967963783/config.json       [] map[]   <nil>  [] map[] map[]}, <nil>

=== FAIL: cli/config TestOldJSON (0.00s)
    config_test.go:202: assertion failed:  (ac.Username string) != joejoe (string)

**Comment by thaJeztah** [comment]
> some tests are failing:

Hmmm... ah, I suspect those go wrong because it will only fetch the home-directory _once_, and the test is updating the path 🤔. Will have to have a look how to solve that test

**Comment by thaJeztah** [comment]
Fixed the tests, but it's a bit ugly; not sure if there's really a cleaner way though

File: cli/config/config.go [hunk 1]
```diff
@@ -26,15 +26,29 @@ const (
 var (
 	initConfigDir sync.Once
 	configDir     string
+	homeDir       string
 )
 
+// resetHomeDir is used in testing to resets the "homeDir" package variable to
+// force re-lookup of the home directory between tests.
+func resetHomeDir() {
```
silvin-lubecki, line 34: Maybe we can move this function in the tests files? I mean, tests are in the same package, so they have access to the `homeDir` variable.

cli/config/config_test.go
```diff
@@ -115,6 +115,7 @@ password`: "Invalid Auth config file",
 email`: "Invalid auth configuration file",
 	}
 
+	resetHomeDir()
```
silvin-lubecki, line 118: I'm a bit on the fence between calling this function for each test as first step, to make sure the path is clean for the test, OR defer it at first step, to make sure everything will be cleared for other tests. 🤔 
```go
func TestOldInvalidsAuth(t *testing.T) {
    defer resetHomeDir()
    ...
}
```
```go
func TestOldInvalidsAuth(t *testing.T) {
   resetHomeDir()
    ...
}
```

cli/config/config.go
thaJeztah: Yes, I was on the fence what would be better; I put it in the non-test file to make it more visible (I really don't like having to do this 😞)

cli/config/config_test.go
thaJeztah: It's tricky; any test that uses the `homeDir()` will initialise the package variable; my current approach was to put it in those tests that expect the variable to be re-resolved (because they patch the `HOME` variable for the test); changing the `HOME` variable is something that would only happen in tests (under normal circumstances it would not modify after the cli binary is started, so it would be safe to store it in the variable.

silvin-lubecki (approved):
Ok let's merge it this way then 👍

Issue #2946, opened 2021-01-23
Enhancement: When no context argument is specified, infer the current working directory as the context to use.


marzhall:
My apologies if this has already been opened; I did a search of "build context," and didn't see any related issues.

**Description**
This feature makes it so that if the user types `docker build`, an incantation which does not specify a context for the build, the docker client tool treats the command as if the user had typed `docker build .`.

I've a branch ready to go that includes the appropriate code changes, and just need to change the branch name to include this feature #.

** Reasoning **

This particular enhancement has been something I've meant to suggest and implement for a few years now, but because I'm a terrible person I had procrastinated. Earlier this week I waited ten minutes for a CI/CD pipeline I was working on late at night and thought I'd just finished, only to get the ever-dreaded `"docker build" requires exactly 1 argument.` and, knowing I had to make a one-character change and put off sleep for another ten minutes, after some very cathartic verbalizations, I decided this was the week I'd do the job.

As far as I've experienced, inferring the current working directory is what we're talking about matches how other regularly used cli commands behave, making it the 'unsurprising' behavior. `make` is the most similar example, I think; while I can specify a different path for `make` to use when running the command (at least in`gnu` make), by default it assumes we're talking about the current directory. `ls` is a more trivial but still notable command; while I can say `ls some/other/folder/`, if I just say `ls`, then it assumes I'd like the directory listing for the directory I'm in. Docker's client not doing the same is surprising to my 'muscle memory' when using the cli, which is likely why I frequently make the mistake. 

Requiring the user to always specify a context seems less consistent with the both the command line tools I use and also the use of docker I see than having them only specify a context in the case in which they don't mean the current working directory.

**Steps to reproduce the issue:**
1. Run `docker build` without a context specified
2. Observe the response ` "docker build" requires exactly 1 argument.`, followed by the client exiting
3. Drink.

**Describe the results you received:**

The docker client crows ` "docker build" requires exactly 1 argument.` smugly at my incompetence, reminding me that I'm so simple I fail at typing even a single character

**Describe the results you expected:**

The docker client makes the assumption - which matches every experience I at least personally have ever had - that I meant to use the current directory as my context, and instead understands implicitly that if I had wanted to specify some build context other than the current working directory, I would have done so explicitly.

**Additional information you deem important (e.g. issue happens only occasionally):**

This change does not impact the behavior of any currently-working workflows. The build context can still be passed in the way it already is, and so anyone explicitly specifying a build context will receive the results they expect.

[...]

PR #2947, opened 2021-01-23
Modify a 'docker build' command with no context specified to infer "." as the context.

marzhall:
Also update documentation to match.

Signed-off-by: Marshall Conover <marzhall.o+docker@gmail.com>

**- What I did**
Removed the requirement to specify a context when running docker build.
**- How I did it**
Modified the options parsing for `docker build` to use the current working directory as the context when no context is specified.
**- How to verify it**
In a directory with a Dockerfile, and with an updated binary, run:
```
    docker build
```
You will see a line to stdout with the message
```
    No build context was specified. Using the current working directory '<current working directory>' as our build context.
```
followed by the docker build command executing as if you had run `docker build .`.

This is in contrast to the previous behavior, in which case running `docker build` would return:
```
    "docker build" requires exactly 1 argument.
```
followed by the client exiting.

Infer the local directory as the context to use if no context is specified. `docker build` is now the same as `docker build .`

image

# Important Notes

- I was not able to figure out how to run the unit tests from within the docker container development environment; I'm cracking at it, but wanted to note that the current documentation (at least that I saw) didn't have a straightforward "just run these commands" in it. I'm happy to update the doc once I have it working.
- I'm unsure of how to test the error case in the code, as it requires os.GetCWD to fail. Pointers on how to do so would be helpful, though (like the above issue) I can continue digging.

Thanks!

(PR for #2946 )

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,0 @@
-FROM ubuntu
-
-RUN echo test
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,3 @@
+FROM ubuntu
+
+RUN echo test
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -112,13 +112,13 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 			} else {
 				currentDir, err := os.Getwd()
 				if err != nil {
-					_, _ = fmt.Fprint(dockerCli.Err(), "No build context was specified, and we were" +
-					    "unable to get the local directory's path in order to use it as a" +
-					    "build context. Consider passing the build context path as an argument.")
+					_, _ = fmt.Fprint(dockerCli.Err(), "No build context was specified, and we were"+
+						"unable to get the local directory's path in order to use it as a"+
+						"build context. Consider passing the build context path as an argument.")
 					return errors.Errorf("unable to prepare context: %s", err)
 				}
-				_, _ = fmt.Fprintln(dockerCli.Out(), "No build context was specified. Using the current " +
-				    fmt.Sprintf("working directory '%s' as our build context.", currentDir))
+				_, _ = fmt.Fprintln(dockerCli.Out(), "No build context was specified. Using the current "+
+					fmt.Sprintf("working directory '%s' as our build context.", currentDir))
 				options.context = currentDir
 			}
 			return runBuild(dockerCli, options)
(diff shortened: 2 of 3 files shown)


thaJeztah (changes requested):
Thanks for contributing! This change will need further discussion. We've had a look at similar changes in the past, and there's some reasons for not making this change;

- keeping it explicit makes sure that unintended missing paths produce an error, instead of "silently" picking the current directory. These situations happen, e.g., when using an environment variable to specify the path (`docker build $SOME_VARIABLE`)
- we want to keep the option open to use `docker build` (without specifying options or a path) to perform a more "interactive" build. No decisions were made on this yet, but changing the behavior to default to `.` would no longer allow us to move in that direction without introducing a breaking change.

For reasons above, I'm "-1" on merging this; I do understand where you're coming from though 🤗

marzhall:
Hi thaJeztah!

I don't know if that's considered the final word or not, but in case not, I'll just throw in a few thoughts:

- For the first point, that's totally reasonable; would a line to stdout or stderr saying "no context was specified, so we're defaulting to the current directory" address that? I can add that later today.
- For the second point, while the idea is interesting, I'm not sure sacrificing a clear and easy-to-implement feature that improves QoL immediately for a feature that might potentially be implemented in the future is the best priority planning-wise. In the case interactive mode was desired after this change went in, it could be implemented as a switch (e.g. `--interactive`), which I think would also better match its frequency of use by developers; intuitively, I think developers who forget to specify a context are more likely to intend to have typed `.` than to want a manual prompt. In my example of making the mistake in my builds, for example, it would've been only slightly less frustrating. Having a prompt appear when no context is specified also makes the client continue to have 'surprising' behavior vs. its other command-line cousins like `ls` who require directories be specified, and infer you meant the local directory should you not specify one.

Let me know if it would be more convenient to discuss this in another spot, and thanks for your time!

marzhall:
I've added informational output to inform the user when the context has been inferred, e.g.:

```
$ build/docker-linux-amd64 build
No build context was specified. Using the current working directory '/home/mars/fun/cli' as our build context.
Sending build context to Docker daemon  45.37MB
Step 1/2 : FROM ubuntu
 ---> f63181f19b2f
Step 2/2 : RUN echo test
 ---> Using cache
 ---> c8e2247d0739
Successfully built c8e2247d0739
```

thaJeztah:
closing for now, as this is no longer in a mergeable state

PR #2951, opened 2021-01-26
Add docs and completion for docker node ls --filter node.label

thaJeztah:
fixes https://github.com/moby/moby/issues/27231

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -4355,6 +4355,9 @@ _docker_node_ls() {
 			__docker_complete_nodes --cur "${cur##*=}" --id
 			return
 			;;
+		label|node.label)
+			return
+			;;
 		membership)
 			COMPREPLY=( $( compgen -W "accepted pending" -- "${cur##*=}" ) )
 			return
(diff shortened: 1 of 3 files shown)


thaJeztah:
@silvin-lubecki @albers ptal

contrib/completion/bash/docker
```diff
@@ -4371,7 +4371,7 @@ _docker_node_ls() {
 
 	case "$prev" in
 		--filter|-f)
-			COMPREPLY=( $( compgen -W "id label membership name role" -S = -- "$cur" ) )
+			COMPREPLY=( $( compgen -W "id label membership name node.label role" -S = -- "$cur" ) )
```
albers, line 4374: We need to add a
```bash
     node.label)
        return
        ;;
```
branch in the preceding case statement in order to silence the following completions. Otherwise `docker node ls --filter node.label=-` would suggest `--filter  --format  --help    --quiet   -f        -q `.

albers (changes requested):

contrib/completion/bash/docker
thaJeztah: Thanks for review! ❤️ I updated the PR; made it `label|node.label)` (as I expect `label` currently had the same issue)

albers: Ah,  yes, you are right!

albers (approved):
LGTM, thanks

silvin-lubecki (approved):
LGTM

PR #2958, opened 2021-02-04
[20.10 backport] cli/config: prevent warning if HOME is not set

thaJeztah:
backport of https://github.com/docker/cli/pull/2934

fixes https://github.com/moby/moby/issues/41890 WARNING: Error loading config file: .dockercfg: $HOME is not defined

commit c2626a8270924d0cec54477fc506174cd31dc560 (https://github.com/docker/cli/pull/2101) replaced the use of github.com/docker/docker/pkg/homedir with Golang's `os.UserHomeDir()`.

This change was partially reverted in 7a279af43de6650d2f36c41f9d926db7be3b3567 (https://github.com/docker/cli/pull/2111) to account for situations where `$HOME` is not set.

In  situations where no configuration file is present in `~/.config/`, the CLI falls back to looking for the (deprecated) `~/.dockercfg` configuration file, which was still using `os.UserHomeDir()`, which produces an error/warning if `$HOME` is not set.

This patch introduces a helper function and a global variable to get the user's home-directory. The global variable is used to prevent repeatedly looking up the user's information (which, depending on the setup can be a costly operation).

```
- Fix "`WARNING: Error loading config file: .dockercfg: $HOME is not defined`"
```

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -26,7 +26,21 @@ const (
 var (
 	initConfigDir sync.Once
 	configDir     string
+	homeDir       string
 )
+
+// resetHomeDir is used in testing to resets the "homeDir" package variable to
+// force re-lookup of the home directory between tests.
+func resetHomeDir() {
+	homeDir = ""
+}
+
+func getHomeDir() string {
+	if homeDir == "" {
+		homeDir = homedir.Get()
+	}
+	return homeDir
+}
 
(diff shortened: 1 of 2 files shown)


cpuguy83 (approved):
LGTM

silvin-lubecki (approved):
LGTM

PR #2959, opened 2021-02-04
[20.10 backport] Fix panic when failing to get DefaultAuthConfig

thaJeztah:
backport of https://github.com/docker/cli/pull/2918

fixes https://github.com/docker/cli/issues/2890 Non-root user "docker login" results in panic SIGSEGV segmentation violation
fixes https://github.com/moby/moby/issues/41771 Command 'docker login' does not work

Commit f32731f9020f12ae2600a3f4c90e668565220cb6 (https://github.com/docker/cli/pull/2818) fixed a potential panic when an error was returned while trying to get existing credentials.

However, other code paths currently use the result of `GetDefaultAuthConfig()` even in an error condition; this resulted in a panic, because a `nil` was returned.

We should fix the other code paths to be more resilient, and to not use the result of `GetDefaultAuthConfig()` if an error occurs, but starting with a minimal patch to fix the immediate issue.

```markdown
- Fix a panic on `docker login` if no config file is present
```


silvin-lubecki (approved):
LGTM

PR #2975, opened 2021-02-22
Slight cleanup/refactor of attachContainer

thaJeztah:
Return early in case there's an error, and declare variables closer to where they're used.


silvin-lubecki (approved):
LGTM

PR #2978, opened 2021-02-23
{WIP}Add support for * for device name in --device

aws-vrnatham:
This is still Work in Progress.
Currently for --device we support /dev/xyz1 and if there
are 0 such devices, then user has to have --device for each one.
Instead with this change in allows --device=/dev/xyz*

Signed-off-by: Venky Natham <vrnatham@amazon.com>

**- Check for * prefix for src device**
Right now the device name is taken as is. If the * prefix is there at the end then consider this as a wildcard.

**- Check for all filepaths**
In the parseLinuxDevice, get all the files that match the src device and add them to the devicemapping.

**- Add --device=/dev/neuron\* and verify it**
`ubuntu@ip-10-1-200-237:~/docker/cli$ sudo ~/docker-linux-amd64 run --name test -it --device=/dev/neuron* --cap-add IPC_LOCK neuron-rtd bash
bash-4.2# ls /dev/neuron*
                         /dev/neuron0  /dev/neuron1  /dev/neuron2  /dev/neuron3`

**- Add support for \* for device name in --device**

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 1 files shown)


thaJeztah (changes requested):
This needs to be implemented on the daemon-side, as docker uses a client-server architecture, so the CLI may not be running on the same host as the daemon is running

PR #2983, opened 2021-02-26
provide commands to remove all containers in Windows

carehart:
The command currently offered fails for those on Windows, which seemed easily solved.

### DIFF
--- a/docs/reference/commandline/rm.md
+++ b/docs/reference/commandline/rm.md
@@ -62,8 +62,12 @@ The main process inside the container referenced under the link `redis` will rec
 $ docker rm $(docker ps -a -q)
 ```
 
-This command deletes all stopped containers. The command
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thanks for contributing! I see CI is failing because the commit is missing a DCO sign-off (https://github.com/docker/cli/blob/master/CONTRIBUTING.md#sign-your-work).

Looking at this example, I think we should recommend using `docker container prune` instead; this example was added to the documentation before the `prune` commands were implemented, but for many cases they are a better alternative (at least; more portable, and easier to use). We can still show an example of "combining commands", but with some warnings/caution added.

Let me open a pull request with what I have in mind.

thaJeztah:
opened https://github.com/docker/cli/pull/3001

PR #2992, opened 2021-03-03
Implement WithDefaultContextStoreConfig() DockerCliOption

thaJeztah:
Just a minor refactor to make this slightly cleaner


ndeloof (approved):

thaJeztah:
FWIW, while rebasing this branch, I saw that `RegisterDefaultStoreEndpoints()` is now unused in our code, and `DefaultContextStoreConfig()` could potentially be un-exported.  However, both are used by Buildx, so we need to verify if that's using it for the correct reasons (or testing the now deprecated Kubernetes contexts?) 

https://github.com/docker/buildx/blob/6efcee28d59e1aa4d935e8d4055ca5c0edff0a4b/driver/kubernetes/context/load_test.go#L19

https://github.com/docker/buildx/blob/6efcee28d59e1aa4d935e8d4055ca5c0edff0a4b/driver/kubernetes/context/load.go#L41


/cc @crazy-max

Issue #1997, opened 2019-07-11
Volume or container completion is broken in zsh


ay0o:
**Description**
Completion does not work for volumes or container's id/names in ZSH

**Steps to reproduce the issue:**
`docker run -v <TAB>`
or
`docker exec -it `<TAB>`

**Describe the results you received:**
Nothing is completed

**Describe the results you expected:**
I expected to have a list of directories in the current path in the first case, or a list of containers in the second.

**Additional information you deem important:**
For example, in Bash that works like a charm. `docker run -v <TAB>` will return a list of the directories, meanwhile `docker exec -it <TAB>` will return a list of the containers.

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        6247962
 Built:             Sun Feb 10 04:12:39 2019
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.2
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.6
  Git commit:       6247962
  Built:            Sun Feb 10 04:13:06 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

**Output of `docker info`:**

```
Containers: 8
 Running: 1
 Paused: 0
 Stopped: 7
Images: 38
Server Version: 18.09.2
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
Swarm: active
 NodeID: trvdyzd4jc8jra925irru11ac
 Is Manager: true
 ClusterID: ravg59u2mm6svhkryrq61uvlv
 Managers: 1
 Nodes: 1
 Default Address Pool: 10.0.0.0/8
 SubnetSize: 24
 Orchestration:
  Task History Retention Limit: 5
 Raft:
  Snapshot Interval: 10000
  Number of Old Snapshots to Retain: 0
  Heartbeat Tick: 1
  Election Tick: 10
 Dispatcher:
  Heartbeat Period: 5 seconds
 CA Configuration:
  Expiry Duration: 3 months
  Force Rotate: 0
 Autolock Managers: false
 Root Rotation In Progress: false
[...]

PR #2998, opened 2021-03-04
[completion/zsh] add volume completion

pseyfert:
This provides completion for `docker run ... -v <TAB>`.
This closes part of #1997.

 - the completion of `docker run -it` is discussed in the issue and does not require a fix (maybe a different default, but that's out of scope here)
 - the completion of `docker run ... -v ` is addressed in this PR

As outlined [here](contrib/completion/zsh/_docker) there are multiple options passed to `_directories`. The behaviour is that:

 - `docker run -v <TAB>` inserts a `/` immediately and will suggest directories relative to `/`. On my system that means `docker run -v <TAB>` turns into `docker run -v /` and the completion list contains elements such as `usr`,`home`
 - a `:` is inserted at the end of every suggestion. depending how the user configured their completion this means they can cycle through insertions of `docker run -v /usr:` `docker run -v /home:`and so on (for me the colon is bold face).
 - the `:` does not need to be removed by the user if they want to go one level higher. hitting space or `/` will remove the `:`. I.e. hitting `/` after the completion inserted `docker run -v /usr:` turns the command line into `docker run -v /usr/`. Continued hitting of TAB continues the game and suggests `docker run -v /usr/bin:`, again with a self-removing `:`
 - the `:` persists if the user types `:`
 - after `:` in the same word there is no more completion suggestions, the user needs to type the volume mount on their own.
 - after a ` ` (such as `docker run -v /home ` or `docker run -v /home:/hosthome `) the completion continues to complete arguments to docker run just like it did before this PR.
 
Verification was trying around on the command line with `docker run -it --rm -v <stuff here> debian:testing`. I think the suggestions should be absolute paths (the original draft suggested relative paths to the cwd, but docker didn't mount them).

One can probably argue if the treatment of the trailing `:` is optimal (self removal on space and slash. i caught myself a few times that after having `docker run -it --rm -v /home:` i tried to continue with `/hosthome` and wanting to type`-v /home:/hosthome` but typed `/home/hosthome` as the `/` overwrote the visual `:`).

not an animal, but a cake:
cake

### DIFF
--- a/contrib/completion/zsh/_docker
+++ b/contrib/completion/zsh/_docker
@@ -661,7 +661,7 @@ __docker_container_subcommand() {
         "($help)*--ulimit=[ulimit options]:ulimit: "
         "($help)--userns=[Container user namespace]:user namespace:(host)"
         "($help)--tmpfs[mount tmpfs]"
-        "($help)*-v[Bind mount a volume]:volume: "
+        "($help)*-v[Bind mount a volume]:volume:_directories -W / -P '/' -S '\:' -r '/ '"
         "($help)--volume-driver=[Optional volume driver for the container]:volume driver:(local)"
         "($help)*--volumes-from=[Mount volumes from the specified container]:volume: "
         "($help -w --workdir)"{-w=,--workdir=}"[Working directory inside the container]:directory:_directories"


pseyfert:
Any maintainer out there willing to review / approve? ( @thaJeztah , tagging you as you appear in the other completion related MRs)

vvoland (approved):
Looks good to me.
However, I think that in the long run we would like to get rid of the manually maintained completions scripts and continue with the approach started by @ndeloof with https://github.com/docker/cli/pull/3429 (using Cobra's completion generation feature).
So probably it would be good to take a look how to implement this completion with the Cobra.

pseyfert:
@vvoland thanks for the feedback. I appreciate the effort going for a generated completion feature rather than the manually maintained function. That said, shall we still try to merge the MR here? It improves the completion file while it's still there. And admittedly I don't (yet?) find my way around the cobra completion, so it will take a bit until I can contribute to that.

thaJeztah:
The only thing I was wondering is if it should complete (named) volumes, and if so, how we could make it either complete those, or local paths (for bind-mounts). Not exactly sure what we currently do for the Bash completion (maybe we don't do it there either)

thaJeztah:
(close/reopen to trigger CI)

thaJeztah (approved):
LGTM

PR #3018, opened 2021-03-19
Update exec.md

KDSGYT:
I was trying to open a shell in the container but wasn't ablt to do so. 
Changed the command from bash to sh and it worked to open a shell in the container


thaJeztah (changes requested):
thanks for contributing!

I left a comment in-line, but the commit is also missing a "DCO" sign-off (see ["Sign your work"](https://github.com/moby/moby/blob/v20.10.11/CONTRIBUTING.md#sign-your-work) in the CONTRIBUTING.md in this repository), which makes CI fail.

Let me close this PR for now, but let me know if you want to contribute the suggested changes as a pull request.

PR #3025, opened 2021-03-25
config.Load() remove unneeded locks

thaJeztah:
These were added in b83bc67136e1edce2bdc225863af83d45d5472c9 (https://github.com/docker/cli/pull/2666), but I'm not sure why I added these; they're likely not needed.

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -105,18 +105,13 @@ func LoadFromReader(configData io.Reader) (*configfile.ConfigFile, error) {
 }
 
 // TODO remove this temporary hack, which is used to warn about the deprecated ~/.dockercfg file
-var (
-	mutex                  sync.RWMutex
-	printLegacyFileWarning bool
-)
+var printLegacyFileWarning bool
 
 // Load reads the configuration files in the given directory, and sets up
 // the auth config information and returns values.
 // FIXME: use the internal golang config parser
 func Load(configDir string) (*configfile.ConfigFile, error) {
-	mutex.Lock()
 	printLegacyFileWarning = false
-	mutex.Unlock()
 
 	if configDir == "" {
 		configDir = Dir()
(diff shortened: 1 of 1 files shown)


thaJeztah:
Failing because of https://github.com/docker/cli/pull/2986

```

=== Failed
=== FAIL: e2e/container TestRunAttachedFromRemoteImageAndRemove (1.91s)
    run_test.go:28: assertion failed: 
        --- expected
        +++ actual
        @@ -3,3 +3,4 @@
         Digest: sha256:641b95ddb2ea9dc2af1a0113b6b348ebc20872ba615204fbe12148e98fd6f23d
         Status: Downloaded newer image for registry:5000/alpine:test-run-pulls
        +Unsupported signal: <nil>. Discarding.
         
        
        
        You can run 'go test . -test.update-golden' to automatically update testdata/run-attached-from-remote-and-remove.golden to the new expected value.'
```

thaJeztah:

tiborvass (approved):

coryb:
@thaJeztah I think I figured out why you added this originally, I now get a race error when I run my tests with `-race`:
```
WARNING: DATA RACE
Write at 0x00000415c60d by goroutine 511:
  github.com/docker/cli/cli/config.Load()
      /go/pkg/mod/github.com/docker/cli@v20.10.7+incompatible/cli/config/config.go:114 +0x6b
  github.com/docker/cli/cli/config.LoadDefaultConfigFile()
      /go/pkg/mod/github.com/docker/cli@v20.10.7+incompatible/cli/config/config.go:152 +0xa4
```
so the mutex was preventing the mutation of global state.  Not sure if there is a better strategy to handle this or if we should add back in the mutex to prevent the race.

thaJeztah:
Ah! Nice find!

PR #3039, opened 2021-04-03
Update stop.md to mention that other stop signals can be set

cvermilion:
**- What I did**

Updates the stop.md doc to mention that the stop signal can be changed, either with the Dockerfile or via `docker run --stop-signal`. This is a real gotcha if you're not familiar with this feature and build a container that extends a container that uses `STOPSIGNAL`.

**- How I did it**

Added an explanatory sentence.

**- How to verify it**

Read it, I guess?

Updates `docker stop` documentation to mention stop signals other than SIGTERM can be set

Here is a very cute bear I spotted in my neighborhood:

Screen Shot 2016-11-01 at 9 52 03 AM

### DIFF
--- a/docs/reference/commandline/stop.md
+++ b/docs/reference/commandline/stop.md
@@ -19,7 +19,9 @@ Options:
 ## Description
 
 The main process inside the container will receive `SIGTERM`, and after a grace
-period, `SIGKILL`.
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM, thank you!

PR #2855, opened 2020-11-29
scripts/build/.variables: Support SOURCE_DATE_EPOCH

Foxboron:
Repeated builds of docker is currently unreproducible as the embedded
time does not support SOURCE_DATE_EPOCH. This patch ensures we check for
the defined variable before utilizing current date.

This has been fixed previously with manpages as well

https://github.com/docker/cli/commit/2d7091aaeb3dac6388d81cbb9b6847c75ee4b7b9

Signed-off-by: Morten Linderud <morten@linderud.pw>

**- How to verify it**
```
λ cli morten/reprobuilds» export SOURCE_DATE_EPOCH=1606660000
λ cli morten/reprobuilds» export DISABLE_WARN_OUTSIDE_CONTAINER=1
λ cli morten/reprobuilds» make
WARNING: binary creates a Linux executable. Use cross for macOS or Windows.
./scripts/build/binary
Building statically linked build/docker-linux-amd64
+ go build -o build/docker-linux-amd64 --ldflags '    -w 
  -X "github.com/docker/cli/cli/version.GitCommit=9d40c7464e" 
  -X "github.com/docker/cli/cli/version.BuildTime=2020-11-29T14:26:40Z" 
  -X "github.com/docker/cli/cli/version.Version=unknown-version"      ' github.com/docker/cli/cmd/docker
++ basename build/docker-linux-amd64
+ ln -sf docker-linux-amd64 build/docker
```

**- Description for the changelog**
Support reproducible builds by checking SOURCE_DATE_EPOCH before build.

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3041, opened 2021-04-05
Reproducibility improvement

vkleinik:
SOURCE_DATE_EPOCH is a standardised environment variable that distributions can set
centrally and have build tools consume this in order to produce reproducible output.

The change was done in the context of the [Reproducible Builds](https://reproducible-builds.org/) initiative.


cpuguy83:
Dup of #2855 
Thanks!

PR #2996, opened 2021-03-04
vendor: update x/sys to 134d130e

tonistiigi:
Makes possible to build for windows/arm64

Separated out from #2993 for easier review

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>

PR #3042, opened 2021-04-06
[20.10 backport] dockerfile based binary building

tiborvass:
Cherry-picks of #2996 and #2993 with added commit that reverses some changes in the main Makefile to limit the script changes within the same 20.10 release. The main purpose is to have darwin arm64 builds working in 20.10.


kadirselcuk:
Dep to individual organization teams

thaJeztah (approved):
LGTM

tonistiigi:
I guess we technically don't need to backport #2996 unless we really want to make a windows/arm64 release.

tonistiigi:
@cpuguy83 https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/Dockerfile#L14https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/Dockerfile#L14 does

But I guess plugins are built somewhere differently that would also need exception so no need to rush with it yet.

cpuguy83:
I really don't like the way this was done. I would also really prefer to not backport this to 20.10 because it breaks downstream builds.

tiborvass:
@tonistiigi Sorry I thought I pushed the changes to remove windows/arm64 and that sys vendor commit. Updated now.

@cpuguy83 I know it's not great to do this in 20.10 which is why I kept the Makefile as backwards compatible as possible while still building the new way. Which part specifically breaks down downstream for you?

tonistiigi (approved):

PR #3055, opened 2021-04-19
Fix typo for configuring dockerd Containerd plugins namespace.

kieran-ohara:
**- What I did**

When I ried to configure the containerd namespace for plugins via JSON, I found an error that was caused by an incorrect spelling

**- How I did it**

Updated the JSON

**- How to verify it**

I ran the daemon using the new configuration suggested in this PR and then it worked.

Fix typo for configuring dockerd Containerd plugins namespace.

### DIFF
--- a/docs/reference/commandline/dockerd.md
+++ b/docs/reference/commandline/dockerd.md
@@ -119,7 +119,7 @@ To run the daemon with debug output, use `dockerd -D` or add `"debug": true` to
 the `daemon.json` file.
 
 > **Enabling experimental features**
-> 
+>
 > Enable experimental features by starting `dockerd` with the `--experimental`
 > flag or adding `"experimental": true` to the `daemon.json` file.
 
@@ -1349,7 +1349,7 @@ This is a full example of the allowed configuration options on Linux:
   "cluster-store-opts": {},
   "containerd": "/run/containerd/containerd.sock",
   "containerd-namespace": "docker",
-  "containerd-plugin-namespace": "docker-plugins",
+  "containerd-plugins-namespace": "docker-plugins",
   "data-root": "",
   "debug": true,
   "default-address-pools": [
@@ -1472,7 +1472,7 @@ This is a full example of the allowed configuration options on Windows:
   "cluster-store": "",
   "containerd": "\\\\.\\pipe\\containerd-containerd",
   "containerd-namespace": "docker",
-  "containerd-plugin-namespace": "docker-plugins",
+  "containerd-plugins-namespace": "docker-plugins",
   "data-root": "",
   "debug": true,
   "default-ulimits": {},


docs/reference/commandline/dockerd.md
```diff
@@ -1349,7 +1349,7 @@ This is a full example of the allowed configuration options on Linux:
   "cluster-store-opts": {},
   "containerd": "/run/containerd/containerd.sock",
   "containerd-namespace": "docker",
-  "containerd-plugin-namespace": "docker-plugins",
+  "containerd-plugins-namespace": "docker-plugins",
```
thaJeztah, line 1352: Looking at the source-code, the option in JSON looks to be named `containerd-plugin-namespace` (singular "plugin"); https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/daemon/config/config.go#L266

```go
ContainerdPluginNamespace string `json:"containerd-plugin-namespace,omitempty"`
```


What error did you get when using this option in your daemon.json?

thaJeztah: (Note that the _flag_ looks to be using plural (which may be somewhat confusing); I'll have a look at adding a hidden flag to support both options; https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/cmd/dockerd/config.go#L100)

thaJeztah (changes requested):

docs/reference/commandline/dockerd.md
kieran-ohara: Hello, I am using Docker and Docker where are the daemon configuration is loaded onto the D in D image.

When I use the non plural version, I get the error that I am sharing below:

Screenshot 2021-04-22 at 15 26 49

(With configuration panel removed):

Screenshot 2021-04-22 at 15 26 55

When I do use the plural version however, the DND container stat as I would expect:

Screenshot 2021-04-22 at 15 24 19

(dont worry about those logs concerning missing containers; the point is the docker daemon is running).

kieran-ohara: For more context, this is the `docker:20.10.6-dind` image.

thaJeztah:
Whoops, sorry for the delay;

Hmmm.. that looks like a bug; the validation fails because it doesn't find a name for a flag that matches the daemon.json option.

However, when trying to use the plural (`containerd-plugins-namespace`) for the option to make validation pass, it won't set the correct value.

Doing a quick patch for debugging, where I disabled that validation, and added a debug line to return the value it'll be using as an error;

```patch
diff --git a/daemon/config/config.go b/daemon/config/config.go
index 4990727597..e6392dcf4e 100644
--- a/daemon/config/config.go
+++ b/daemon/config/config.go
@@ -445,6 +445,7 @@ func getConflictFreeConfiguration(configFile string, flags *pflag.FlagSet) (*Con
        if err := json.NewDecoder(reader).Decode(&config); err != nil {
                return nil, err
        }
+       return nil, errors.New("Plugin namespace is: " + config.ContainerdPluginNamespace)

        if config.RootDeprecated != "" {
                logrus.Warn(`The "graph" config file option is deprecated. Please use "data-root" instead.`)
@@ -479,6 +480,7 @@ func configValuesSet(config map[string]interface{}) map[string]interface{} {
 // duplicated configurations and unknown keys. It returns an error with all the conflicts if
 // it finds any.
 func findConfigurationConflicts(config map[string]interface{}, flags *pflag.FlagSet) error {
+       return nil
        // 1. Search keys from the file that we don't recognize as flags.
        unknownKeys := make(map[string]interface{})
        for key, value := range config {
```

Then, create a test `test-daemon.json` that uses `containerd-plugins-namespace`:

```bash
echo '{"containerd-plugins-namespace":"foobar"}' > /tmp/test-daemon.json
```

As can be seen; the value won't be used (Plugin namespace is empty);

```bash
dockerd --config-file=/tmp/test-daemon.json
unable to configure the Docker daemon with file /tmp/test-daemon.json: Plugin namespace is:
```

Whereas using the "correct" name (for JSON) the value (`foobar`) _is_ used;

```bash
echo '{"containerd-plugin-namespace":"foobar"}' > /tmp/test-daemon.json

dockerd --config-file=/tmp/test-daemon.json
unable to configure the Docker daemon with file /tmp/test-daemon.json: Plugin namespace is: foobar
```

PR #3070, opened 2021-04-30
Various test cleanups

thaJeztah:
### Use designated test domains (RFC2606) in tests

Some tests were using domain names that were intended to be "fake", but are
actually registered domain names (such as mycorp.com).

Even though we were not actually making connections to these domains, it's
better to use domains that are designated for testing/examples in RFC2606:
https://tools.ietf.org/html/rfc2606

### cli/config/configfile: various test cleanups

- use var/const blocks when declaring a list of variables
- use const where possible

TestCheckKubernetesConfigurationRaiseAnErrorOnInvalidValue:

- use keys when assigning values
- make sure test is dereferenced in the loop
- use subtests


thaJeztah:
@silvin-lubecki ptal

silvin-lubecki (approved):
LGTM

PR #3072, opened 2021-04-30
docs: various changes, and touch-up main CLI page

thaJeztah:
See individual commits for details

Before

After

PR #3074, opened 2021-05-01
Add support for ALL_PROXY

thaJeztah:
Depends on / relates to:

- [x] https://github.com/docker/cli/pull/3072 docs: various changes, and touch-up main CLI page
- [x] https://github.com/moby/buildkit/pull/2086 Add "ALL_PROXY" to list of default args / proxy env-vars
- [x] https://github.com/moby/moby/pull/42318 builder/dockerfile: add "ALL_PROXY" to list of default build args
- [x] revendor buildkit in moby and cli https://github.com/docker/cli/pull/3153

(changes are needed in https://github.com/docker/docker.github.io as well)

Support for ALL_PROXY as default build-arg was added recently in
buildkit and the classic builder.

This patch adds the `ALL_PROXY` environment variable to the list of
configurable proxy variables, and updates the documentation.

### DIFF
--- a/docs/reference/commandline/cli.md
+++ b/docs/reference/commandline/cli.md
@@ -214,6 +214,7 @@ be set for each environment:
 | `httpsProxy`   | Default value of `HTTPS_PROXY` and `https_proxy` for containers, and as `--build-arg` on `docker build` |
 | `ftpProxy`     | Default value of `FTP_PROXY` and `ftp_proxy` for containers, and as `--build-arg` on `docker build`     |
 | `noProxy`      | Default value of `NO_PROXY` and `no_proxy` for containers, and as `--build-arg` on `docker build`       |
+| `allProxy`     | Default value of `ALL_PROXY` and `all_proxy` for containers, and as `--build-arg` on `docker build`     |
 
 These settings are used to configure proxy settings for containers only, and not
 used as proxy settings for the `docker` CLI or the `dockerd` daemon. Refer to the
@@ -295,7 +296,8 @@ various fields:
       "httpProxy":  "http://user:pass@example.com:3128",
       "httpsProxy": "https://my-proxy.example.com:3129",
       "noProxy":    "intra.mycorp.example.com",
-      "ftpProxy":   "http://user:pass@example.com:3128"
+      "ftpProxy":   "http://user:pass@example.com:3128",
+      "allProxy":   "socks://example.com:1234"
     },
     "https://manager1.mycorp.example.com:2377": {
       "httpProxy":  "http://user:pass@example.com:3128",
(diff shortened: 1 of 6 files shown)


thaJeztah:
Looks like the buildkit changes were pulled in through https://github.com/docker/cli/pull/3153; https://github.com/docker/cli/blob/711334e9a4b70978816052564d3bfcc3b182733d/vendor/github.com/moby/buildkit/client/llb/exec.go#L236

thaJeztah:
Looks like the `gotest.tools` vanity URL went down?

```
bash -c 'vndr |& grep -v -i clone | tee ./vndr.log'
2021/07/21 07:08:10 Collecting initial packages
2021/07/21 07:08:10 Download dependencies
2021/07/21 07:08:22 unrecognized import path "gotest.tools/v3" (https fetch: Get https://gotest.tools/v3?go-get=1: read tcp 172.17.0.2:43364->216.239.32.21:443: read: connection reset by peer)
scripts/validate/check-git-diff vendor
```

thaJeztah:
@silvin-lubecki  @tonistiigi @crazy-max PTAL


/cc @StefanScherer (as a "heads up"; in case we need to make changes in Docker Desktop once this ships)

docs/reference/commandline/cli.md
```diff
@@ -214,6 +214,7 @@ be set for each environment:
 | `httpsProxy`   | Default value of `HTTPS_PROXY` and `https_proxy` for containers, and as `--build-arg` on `docker build` |
 | `ftpProxy`     | Default value of `FTP_PROXY` and `ftp_proxy` for containers, and as `--build-arg` on `docker build`     |
 | `noProxy`      | Default value of `NO_PROXY` and `no_proxy` for containers, and as `--build-arg` on `docker build`       |
+| `allProxy`     | Default value of `ALL_PROXY` and `all_proxy` for containers, and as `--build-arg` on `docker build`     |
```
silvin-lubecki, line 217: Question: should we add ALL_PROXY to the environment variables section too? https://github.com/docker/cli/pull/3074/files#diff-98c80ce7ccfb31b6e6204cc1abaa7d250c8d92ba9f882a0f2b38115081c646d1L87

silvin-lubecki (approved):

docs/reference/commandline/cli.md
thaJeztah: Hm.. not sure; that section talks about what Go itself does by default (https://pkg.go.dev/net/http#ProxyFromEnvironment), which doesn't include `ALL_PROXY` (although I _think_ we have some code in place to support it in github.com/docker/go-connections).

It also mentions "_In particular_, you may find these useful:", to highlight the most common ones.

We can have a look though if `ALL_PROXY` is indeed supported (and if so, update that section), but it's separate from this one (which is for containers, not the daemon and cli itself)

silvin-lubecki: Ok let's keep it that way then 👍

Issue #993, opened 2018-04-10
https://github.com/robbyrussell/oh-my-zsh/issues/6710


kenden:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Description**

docker zsh completion breaks when optional arguments are combined.
Example, when `-i` and `-t` are combined to `-it`

This problem was reported here: https://github.com/robbyrussell/oh-my-zsh/issues/6710

**Steps to reproduce the issue:**
1. Make sure you are using zsh and check that the docker autocomplete is enabled.  
  Type `docker` then press tab.
If it shows you:
 ```
image   -- Manage images
images  -- List images
```
then it is enabled.

2. Type the following:
`docker run -it` then press tab
--> **PROBLEM: **A list of executables on the PATH is displayed instead of possible arguments to `docker run`

**Describe the results you received:**

When typing `docker run -it` and pressing tab, a list of executables on the PATH is displayed.

**Describe the results you expected:**

When typing `docker run -it` and pressing tab, I expect to get a list of possible arguments to `docker run`

**Additional information you deem important (e.g. issue happens only occasionally):**

When not combining options, it works fine. Ex, when typing 
`docker run -i -t` and pressing tab, I get:
```
02f9d6707661  -- 18.04           in ubuntu
0458a4468cbc  -- xenial-20180123 in ubuntu
3fd9065eaf02  -- latest          in alpine
4ea294712533  -- latest          in nate/dockviz
5b117edd0b76  -- 12.04           in ubuntu
```

**Output of `docker version`:**
```
$ docker version
Client:
 Version:	18.03.0-ce
 API version:	1.37
 Go version:	go1.9.4
 Git commit:	0520e24
 Built:	Wed Mar 21 23:06:22 2018
 OS/Arch:	darwin/amd64
 Experimental:	false
 Orchestrator:	swarm

Server:
 Engine:
  Version:	18.03.0-ce
  API version:	1.37 (minimum version 1.12)
  Go version:	go1.9.4
  Git commit:	0520e24
  Built:	Wed Mar 21 23:14:32 2018
  OS/Arch:	linux/amd64
  Experimental:	true```

I tried using the current lastest version available at https://github.com/docker/cli/blob/ec412bdbb47b0cd84d85a0212c7160cec0ee9f02/contrib/completion/zsh/_docker
(put the content of that file to `~/.oh-my-zsh/plugins/docker/_docker`)
with the same result.

**Output of `docker info`:**

```
$ docker info
[...]

PR #3075, opened 2021-05-02
zsh completion: Fix completion when using `docker run/exec -it` shortcut

michelhe:
This commit adds the -it support to any command that has both --interactive and --tty.
Before, the completions plugin did not recognize the -it option and thus the subcommands completion context break whenever -it is used.

**- What I did**
Added the `-it` option in the relevant places in the zsh plugin
**- How I did it**

**- How to verify it**
on the master branch of a oh-my-zsh installation, run:
`docker run -it <TAB>` - the container list does not appear.

now replace with the plugin from this branch:
`docker run -it <TAB>` - the container name completor kicks in.

fix completion for the -it option

image

### DIFF
--- a/contrib/completion/zsh/_docker
+++ b/contrib/completion/zsh/_docker
@@ -665,6 +665,7 @@ __docker_container_subcommand() {
         "($help)--volume-driver=[Optional volume driver for the container]:volume driver:(local)"
         "($help)*--volumes-from=[Mount volumes from the specified container]:volume: "
         "($help -w --workdir)"{-w=,--workdir=}"[Working directory inside the container]:directory:_directories"
+        "($help)-it[shortcut for --interactive and --tty]"
     )
     opts_create_run_update=(
         "($help)--blkio-weight=[Block IO (relative weight), between 10 and 1000]:Block IO weight:(10 100 500 1000)"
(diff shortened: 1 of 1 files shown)


contrib/completion/zsh/_docker
```diff
@@ -665,6 +665,7 @@ __docker_container_subcommand() {
         "($help)--volume-driver=[Optional volume driver for the container]:volume driver:(local)"
         "($help)*--volumes-from=[Mount volumes from the specified container]:volume: "
         "($help -w --workdir)"{-w=,--workdir=}"[Working directory inside the container]:directory:_directories"
+        "($help)-it[shortcut for --interactive and --tty]"
```
thaJeztah, line 668: I don't think this would be the right thing to do, as any shorthand flag can be combined in any order (e.g. both `-ti` or `-it` can be used).

The completion script already has support for "short option stacking"; https://github.com/docker/cli/blob/master/contrib/completion/zsh/_docker#L42-L49

But it's disabled by default, because enabling this feature caused some regression for other completion; see the discussion on https://github.com/moby/moby/pull/17124#issuecomment-150518700

Also, related; https://github.com/ohmyzsh/ohmyzsh/issues/6710 and https://github.com/docker/cli/issues/993

thaJeztah (changes requested):

contrib/completion/zsh/_docker
michelhe: Thanks, I wasn't aware of that.

We can close this PR, in the meantime I would use it locally for my setup. :)

PR #3077, opened 2021-05-04
printServerWarningsLegacy: silence "No kernel memory limit support"; silence "No oom kill disable support" on cgroup v2

AkihiroSuda:
**- What I did**
Removed unhandleable warnings from `printServerWarningsLegacy()`.

**- How I did it**
- silence "No kernel memory limit support" (ref: https://github.com/moby/moby/pull/41254/commits/b8ca7de823acf572cad9becb74fd5cdac1748057)
- silence "No oom kill disable support" on cgroup v2 (ref:  https://github.com/moby/moby/pull/41894/commits/8086443a449af8ef5c5b0574de8957df7e52f296)

These warnings have been already silenced on the daemon side.

**- How to verify it**
Before:
```console
$ docker info
...
Cgroup Version: 2
...

WARNING: No kernel memory limit support
WARNING: No oom kill disable support
```

After:  (no warning)


thaJeztah:
I actually think it's a bug that we hit this code; instead of checking if the Warnings slice is empty, we should also skip this function based on API version used; https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/cli/command/system/info.go#L381

AkihiroSuda:
> I actually think it's a bug that we hit this code; instead of checking if the Warnings slice is empty, we should also skip this function based on API version used;

Yes, but even when `DOCKER_API_VERSION` is set to an old version, I think we should silence the unhandleable warnings.

thaJeztah (approved):
LGTM

thaJeztah:
@silvin-lubecki ptal

silvin-lubecki (approved):
LGTM

PR #3081, opened 2021-05-04
Docker build documentation: Powershell examples added for piping.

SetTrend:
Piping from STDIN is not supported on PowerShell, which is the current
REPL on Windows. I added corresponding PowerShell examples, so Windows
users can learn from and understand the documentation.

/ref: /docker/docker.github.io/issues/12794
<br/>

**A picture of a cute animal (not mandatory but encouraged)**


SetTrend:
@thaJeztah: Do you believe we'll get any progress here?

PR #3103, opened 2021-05-21
Ignore SIGURG on Unix (including Darwin)

djs55:
**- What I did**

Ignored spurious `SIGURG` signals generated by the Go 1.14+ runtime on Darwin.

Running the example in https://github.com/golang/go/issues/37942
I see lots of these signals:
```
dave@m1 sigurg % uname -ms
Darwin arm64

dave@m1 sigurg % go run main.go
received urgent I/O condition: 2021-05-21 16:03:03.482211 +0100 BST m=+0.014553751
received urgent I/O condition: 2021-05-21 16:03:03.507171 +0100 BST m=+0.039514459
```

Previously I would also see a spurious error when closing a connection:
```
dave@m1 cli % ./build/docker run -it --init ubuntu
root@7f9fc6aee300:/# exit
Unsupported signal: <nil>. Discarding.
```

**- How I did it**

#2929 already covered `SIGURG` on Linux. I extended it to cover all Unix versions via `_unix.go`.

I added an `ok` to detect the signals channel being closed and return from the forwarding function.

**- How to verify it**

The unit test for Linux should also cover Mac. On Docker Desktop if you run an interactive container:
```
dave@m1 cli % ./build/docker run -it --init ubuntu
```
and then watch the Docker API proxy log:
```
tail -f ~/Library/Containers/com.docker.docker/Data/log/host/com.docker.driver.amd64-linux.log
```
and observe no
```
2021-05-21 18:01:39.972896+0900  localhost com.docker.driver.amd64-linux[87081]: proxy >> POST /v1.41/containers/b6e4ab5bb7d4271bae2ad14dfce139f9b0e84f315ae6e9fdac8224cd93c0b780/kill?signal=URG
```
as seen on docker/for-mac#5712

- Don't inject `SIGURG` signals generated by the Go runtime into containers

darwinfrog

### DIFF
--- a/cli/command/container/signals_unix.go
+++ b/cli/command/container/signals_unix.go
@@ -1,3 +1,5 @@
+// +build !windows
+
 package container
 
 import (
--- a/cli/command/container/signals_windows.go
+++ b/cli/command/container/signals_windows.go
@@ -1,5 +1,3 @@
-// +build !linux
-
 package container
 
 import "os"
--- a/cli/command/container/signals_unix_test.go
+++ b/cli/command/container/signals_unix_test.go
@@ -1,3 +1,5 @@
+// +build !windows
+
 package container
 
 import (


cpuguy83:
Wow, that's surprising. Confirmed this is getting fired off quite rapidly on darwin.

cli/command/container/signals_sigurg.go
```diff
@@ -1,3 +1,5 @@
+// +build linux darwin
```
cpuguy83, line 1: Maybe `signals_unix.go` and make it `!windows`
I saw some issues related to bsd in the go repo about this as well.

It's a pretty good bet no one will want to forward this signal, ever.

thaJeztah: Was wondering if we should filter by name; does windows also have a "SIGURG" defined?

djs55: I've pushed an update to use `signals_unix.go` and `!windows`. Hopefully CircleCI's infra is back on line!

It looks like SIGURG isn't defined on Windows, according to https://stackoverflow.com/a/63396672/6774019 . According to the [x/sys/windows docs](https://pkg.go.dev/golang.org/x/sys/windows#Signal) there is only
```golang
const (
	// More invented values for signals
	SIGHUP  = Signal(0x1)
	SIGINT  = Signal(0x2)
	SIGQUIT = Signal(0x3)
	SIGILL  = Signal(0x4)
	SIGTRAP = Signal(0x5)
	SIGABRT = Signal(0x6)
	SIGBUS  = Signal(0x7)
	SIGFPE  = Signal(0x8)
	SIGKILL = Signal(0x9)
	SIGSEGV = Signal(0xb)
	SIGPIPE = Signal(0xd)
	SIGALRM = Signal(0xe)
	SIGTERM = Signal(0xf)
)
```
I guess this is good news 😅

cli/command/container/signals_windows.go
```diff
@@ -1,4 +1,4 @@
-// +build !linux
+// +build !linux,!darwin
```
cpuguy83, line 1: Build tags are redundant here. `_windows` implicitly makes this only build on windows.

djs55: Good point! I've repushed, removing the build tags here.

cli/command/container/signals_unix.go
```diff
@@ -1,3 +1,5 @@
+// +build linux darwin
```
cpuguy83, line 1: Sorry usually this is `!windows`. Any reason to restrict this from other unixes?

djs55: Ah, no reason! I agree having `signals_windows.go` and  `signals_unix.go` with `!windows` makes sense. For such a small patch it's taking me a while to get it right :)

cpuguy83 (approved):
LGTM

Thanks for dealing with nitpicks.

thaJeztah:
Second commit looks to be the same as https://github.com/docker/cli/pull/2986 (which removes the warning as well); let me check if that one can be merged 🤗

djs55:
I've force-pushed to remove the second commit, since it has been independently fixed in #2986

thaJeztah (approved):
LGTM, thanks!

PR #3134, opened 2021-06-11
e2e: make sure that os.environ is preserved

thaJeztah:
We updated some of these functions to make sure os.environ was preserved, but some where not.

This adds a utility to help with this, which also prevents the os.environ to be added multiple times.


thaJeztah:
@silvin-lubecki PTAL

silvin-lubecki (approved):
LGTM

PR #2544, opened 2020-05-24
Fixes #2543: bug with panic when DOCKER_CLI_EXPERIMENTAL is incorrect

qoo2p5:
**- What I did**

I've fixed the bug described in #2543.

**- How I did it**

1. I've added a check in cli initialization.
2. I've added `panic(err)` in the place which caused panic before. It will help to find similar problems easier in the future.

**- How to verify it**

The steps are described in the issue. You can see that after my commit everything works as expected.

**- Description for the changelog**

Fixed bug with panic when DOCKER_CLI_EXPERIMENTAL is set to incorrect value.

Issue #2543, opened 2020-05-24
DOCKER_CLI_EXPERIMENTAL="1" docker version fails with panic


qoo2p5:
`DOCKER_CLI_EXPERIMENTAL="1" docker version` fails with panic.

**Steps to reproduce the issue:**
1. `git clone https://github.com/docker/cli.git && cd cli`
2. `make -f docker.Makefile shell`
3. `make binary`
4. `DOCKER_CLI_EXPERIMENTAL="1" ./build/docker-linux-amd64`

**Describe the results you received:**
```panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xa275dd]

goroutine 1 [running]:
github.com/docker/cli/cli/command.(*DockerCli).ClientInfo(...)
	/go/src/github.com/docker/cli/cli/command/cli.go:149
github.com/docker/cli/cli/command.(*DockerCli).DefaultVersion(0xc0003b69c0, 0xc00014e7fc, 0x4)
	/go/src/github.com/docker/cli/cli/command/cli.go:89 +0x2d
github.com/docker/cli/cli/command/system.runVersion(0x20bff80, 0xc0003b69c0, 0xc00032b220, 0x0, 0x0)
	/go/src/github.com/docker/cli/cli/command/system/version.go:145 +0x123
github.com/docker/cli/cli/command/system.NewVersionCommand.func1(0xc0005f4dc0, 0x3075708, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/docker/cli/cli/command/system/version.go:108 +0x3c
github.com/docker/cli/vendor/github.com/spf13/cobra.(*Command).execute(0xc0005f4dc0, 0xc0005b1430, 0x0, 0x0, 0xc0005f4dc0, 0xc0005b1430)
	/go/src/github.com/docker/cli/vendor/github.com/spf13/cobra/command.go:842 +0x460
github.com/docker/cli/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc000414580, 0xc0005b1430, 0x1, 0x1)
	/go/src/github.com/docker/cli/vendor/github.com/spf13/cobra/command.go:950 +0x349
github.com/docker/cli/vendor/github.com/spf13/cobra.(*Command).Execute(...)
	/go/src/github.com/docker/cli/vendor/github.com/spf13/cobra/command.go:887
[...]

PR #3136, opened 2021-06-15
Remove ClientInfo as it is not practically used.

thaJeztah:
- relates to https://github.com/docker/for-linux/issues/1251#issuecomment-861380310

The information in this struct was basically fixed (there's
some discrepancy around the "DefaultVersion" which, probably,
should never vary, and always be set to the Default (maximum)
API version supported by the client.

Experimental is now always enabled, so this information did
not require any dynamic info as well.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -52,7 +52,6 @@ type Cli interface {
 	Apply(ops ...DockerCliOption) error
 	ConfigFile() *configfile.ConfigFile
 	ServerInfo() ServerInfo
-	ClientInfo() ClientInfo
 	NotaryClient(imgRefAndAuth trust.ImageRefAndAuth, actions []string) (notaryclient.Repository, error)
 	DefaultVersion() string
 	ManifestStore() manifeststore.Store
@@ -73,7 +72,6 @@ type DockerCli struct {
 	err                io.Writer
 	client             client.APIClient
 	serverInfo         ServerInfo
-	clientInfo         *ClientInfo
 	contentTrust       bool
 	contextStore       store.Store
 	currentContext     string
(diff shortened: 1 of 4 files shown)


cli/command/cli.go
```diff
@@ -265,11 +239,6 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...Initialize
 		}
 	}
 	cli.initializeFromClient()
-
-	if err := cli.loadClientInfo(); err != nil {
```
thaJeztah, line 269: This one was added back in https://github.com/docker/cli/pull/2544, related to https://github.com/docker/cli/issues/2543, which should also no longer be an issue.

That said; we need to verify if there's any codepath that expects the client to perform version negotiation somewhere.

cli/command/cli.go
```diff
@@ -76,17 +75,16 @@ type DockerCli struct {
...
 	currentContext     string
 	dockerEndpoint     docker.Endpoint
 	contextStoreConfig store.Config
 }
 
-// DefaultVersion returns api.defaultVersion or DOCKER_API_VERSION if specified.
+// DefaultVersion returns api.defaultVersion.
 func (cli *DockerCli) DefaultVersion() string {
-	return cli.ClientInfo().DefaultVersion
+	return api.DefaultVersion
```
thaJeztah, line 87: Relates to https://github.com/docker/for-linux/issues/1251#issuecomment-861380310. Still not sure why we changed `DefaultVersion` to be containing "negotiated version"

cli/command/cli.go
thaJeztah: Actually looks like that's already handled by `initializeFromClient()`; https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/cli/command/cli.go#L361-L370

ndeloof (approved):

PR #3128, opened 2021-06-09
e2e: update engine version

thaJeztah:
see individual commits for details

PR #3137, opened 2021-06-15
CI: update engine versions and Jenkins labels

thaJeztah:
Splitting this from https://github.com/docker/cli/pull/3128, as the "e2e" docker-in-docker version needs a bit more work.

This update makes sure that Jenkins doesn't fail when trying to run on machines with cgroups v2 enabled (which docker 19.03 doesn't work on)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,9 +21,9 @@ pipeline {
                     make -f docker.Makefile test-e2e-non-experimental"
             }
         }
-        stage("e2e (non-experimental) - 18.09 engine") {
+        stage("e2e (non-experimental) - 19.03 engine") {
             steps {
-                sh "E2E_ENGINE_VERSION=18.09-dind \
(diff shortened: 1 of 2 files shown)


thaJeztah:
@silvin-lubecki ptal 🤗

Jenkinsfile
```diff
@@ -21,9 +21,9 @@ pipeline {
                     make -f docker.Makefile test-e2e-non-experimental"
             }
         }
-        stage("e2e (non-experimental) - 18.09 engine") {
+        stage("e2e (non-experimental) - 19.03 engine") {
```
silvin-lubecki, line 24: question: can't we bump to 20.10 here too ?

thaJeztah: This one is specifically to run agains an older version of the engine (others should be running on latest)

silvin-lubecki (approved):
LGTM

PR #3140, opened 2021-06-16
e2e: update docker-compose to 1.29.2

thaJeztah:
splitting from https://github.com/docker/cli/pull/3128. After this I'll have a look at using the new compose-cli (v2) and install that as a cli plugin

Newer versions have COMPOSE_DOCKER_CLI_BUILD enabled by default,
so removing that env-var.


mat007 (approved):
Makes sense!

PR #3141, opened 2021-06-16
Makefile: allow overriding go command by environment

metux:
This is required for environments/build systems where a specific
go version / command needs to be used.

Signed-off-by: Enrico Weigelt, metux IT consult <info@metux.net>

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 #
 # github.com/docker/cli
 #
+export GO ?= go
+
 all: binary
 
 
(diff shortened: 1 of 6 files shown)


thaJeztah:
overall "SGTM", but I'm a bit scared that this would (going forward) require _any_ script or file that's used to be using the `$GO` variable. I'm curious though; wouldn't it be possible to set a custom `PATH` to use a custom / alternative version of binaries?

Go also provides some environment variables; https://golang.org/pkg/cmd/go/internal/help/, and I know those are used by (e.g.) the setup-go GitHub action to switch between go versions; https://github.com/actions/setup-go

tianon:
I'm -1 on this change -- I agree that `PATH` is the appropriate solution for pointing to a different Go version/install, and is the solution used commonly in for example Debian when a specific Go version is necessary for a package (`Build-Depends` on `golang-1.X-go` + `export PATH := /usr/lib/go-1.X/bin:$(PATH)` in `debian/rules`).  I don't think this is a common environment variable for projects to support.

thaJeztah:
Let me go ahead and close this, but feel free to comment if there is a specific reason `PATH` cannot be used for this

PR #3142, opened 2021-06-16
Add information on WORKDIR

govindrai:
**- What I did**
added information about default WORKDIR value
added information about base images passing down WORKDIR values
added best practice for WORKDIR

**- How I did it**
Updated README.md

**- How to verify it**
Use a base image that has WORKDIR set and see that WORKDIR in your image will automatically default to the base image's WORKDIR

Add information about default WORKDIR value, inherited WORKDIR value, and best practice.


thaJeztah (approved):
LGTM, thanks!

PR #3148, opened 2021-06-18
`--until` example should use negative interval

danielsparing:
**- What I did**

Followed Example featuring `--until` within logs.md and could not replicate the shown functionality, however, it does work as expected if the sign of the interval is inverted. It is also in line with the sign meaning in case of the `--since` flag.

Therefore I replaced the example to use negative interval, to match the expected behavior.

**- How I did it**

Only doc change

**- How to verify it**

Just run the example pre and post commit.

fixed logs example using --until to correctly print future log records

### DIFF
--- a/docs/reference/commandline/logs.md
+++ b/docs/reference/commandline/logs.md
@@ -65,13 +65,13 @@ fraction of a second no more than nine digits long. You can combine the
 
 ### Retrieve logs until a specific point in time
 
-In order to retrieve logs before a specific point in time, run:
+In order to retrieve logs before a specific point in time in the future, run:
 
 ```bash
 $ docker run --name test -d busybox sh -c "while true; do $(echo date); sleep 1; done"
 $ date
 Tue 14 Nov 2017 16:40:00 CET
-$ docker logs -f --until=2s test
+$ docker logs -f --until=-2s test
 Tue 14 Nov 2017 16:40:00 CET
 Tue 14 Nov 2017 16:40:01 CET
 Tue 14 Nov 2017 16:40:02 CET


docs/reference/commandline/logs.md
```diff
@@ -65,13 +65,13 @@ fraction of a second no more than nine digits long. You can combine the
...
 
-In order to retrieve logs before a specific point in time, run:
+In order to retrieve logs before a specific point in time in the future, run:
 
 ```bash
 $ docker run --name test -d busybox sh -c "while true; do $(echo date); sleep 1; done"
 $ date
 Tue 14 Nov 2017 16:40:00 CET
-$ docker logs -f --until=2s test
+$ docker logs -f --until=-2s test
```
thaJeztah, line 74: Negative should produce a date in the past, not in future.

Do you have more details about what isn't working for you? (I agree that these docs could use some further refinement in general); one reason could be that, because the time will be relative to the time _on the daemon_, possibly the `2s` is not long enough, which could be the time used to start the container, and/or in a situation where the daemon is running remotely (or in a VM) and the clock of the daemon and client are slightly out of sync. 

If that's the reason, perhaps instead, we can change the `--until` to `10s` (e.g.) to reduce that chance (assuming they're not more than 10 seconds out of sync).

danielsparing: @thaJeztah Thanks, trying out `10s` is a good idea, however when comparing, I still clearly see negative as the future (see below -- besides what you see printed out, I also see that exactly the `-10s` version is the one that prints "slowly" and only returns the prompt after ~10s).

Maybe then `docker logs` itself should be fixed to flip the sign interpretation in `--until`, and get in line with this documentation example?

```
$ docker -v
Docker version 20.10.7, build f0df350
$ docker run --name test -d busybox sh -c "while true; do $(echo date); sleep 1; done"
3b1c0a54f7ac876a25a8f41536f08058bf45e805bc8d8374f2ab15a2370e39b7
$ sleep 15
$ date ; echo '---' ; docker logs -f --until=10s test
Tue Jun 22 21:01:45 UTC 2021
---
Tue Jun 22 21:01:30 UTC 2021
Tue Jun 22 21:01:31 UTC 2021
Tue Jun 22 21:01:32 UTC 2021
Tue Jun 22 21:01:33 UTC 2021
Tue Jun 22 21:01:34 UTC 2021
$ date ; echo '---' ; docker logs -f --until=-10s test
Tue Jun 22 21:01:46 UTC 2021
---
Tue Jun 22 21:01:30 UTC 2021
Tue Jun 22 21:01:31 UTC 2021
Tue Jun 22 21:01:32 UTC 2021
Tue Jun 22 21:01:33 UTC 2021
Tue Jun 22 21:01:34 UTC 2021
Tue Jun 22 21:01:35 UTC 2021
Tue Jun 22 21:01:36 UTC 2021
Tue Jun 22 21:01:37 UTC 2021
Tue Jun 22 21:01:38 UTC 2021
Tue Jun 22 21:01:39 UTC 2021
Tue Jun 22 21:01:40 UTC 2021
Tue Jun 22 21:01:41 UTC 2021
Tue Jun 22 21:01:42 UTC 2021
Tue Jun 22 21:01:43 UTC 2021
Tue Jun 22 21:01:44 UTC 2021
Tue Jun 22 21:01:45 UTC 2021
Tue Jun 22 21:01:46 UTC 2021
Tue Jun 22 21:01:47 UTC 2021
Tue Jun 22 21:01:48 UTC 2021
Tue Jun 22 21:01:49 UTC 2021
Tue Jun 22 21:01:50 UTC 2021
Tue Jun 22 21:01:51 UTC 2021
Tue Jun 22 21:01:52 UTC 2021
Tue Jun 22 21:01:53 UTC 2021
Tue Jun 22 21:01:54 UTC 2021
Tue Jun 22 21:01:55 UTC 2021
```

StefanScherer: I think the PR is correct.
Of course it looks weird in the first place to use negative relative time to log for a specific amount of time.

But I think the intention of `--since` and `--until` was more to have a range.
Example:
```
docker logs -f --since '5m' --until '4m' containerid
```

OK, the `-f` doesn't make much sense, the CLI exits after it shows the logs from 5 minutes ago until 4 minutes ago. I looked at other commands like `docker events` which have the same.

I guess we cannot change the behaviour for `--until` to calculate in the opposite direction as `--since`.

PR #3158, opened 2021-06-23
Add completion for docker-compose plugin

ulyssessouza:
WIP: Note that this PR depends on the implementation of [completions PR](https://github.com/docker/compose-cli/pull/1773) in `docker/compose-cli` to work.

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -5486,21 +5486,21 @@ _docker_wait() {
 	_docker_container_wait
 }
 
+COMPOSE_PLUGIN_PATH=$(docker info --format '{{json .ClientInfo.Plugins}}' | sed -n 's/.*"Path":"\([^"]\+docker-compose\)".*/\1/p')
+
 _docker_compose() {
-	local composePluginPath="${HOME}/.docker/cli-plugins/docker-compose"
 	local completionCommand="__completeNoDesc"
-  local resultArray=(${composePluginPath} ${completionCommand} compose)
-	for value in "${words[@]:2}"
-	do
-		if [[ "${value}" == "" ]] ; then
+	local resultArray=($COMPOSE_PLUGIN_PATH $completionCommand compose)
+	for value in "${words[@]:2}"; do
+		if [ -z "$value" ]; then
 			resultArray+=( "''" )
 		else
-			resultArray+=( "${value}" )
+			resultArray+=( "$value" )
 		fi
 	done
-	local result=$(eval "${resultArray[*]}" 2> /dev/null)
+	local result=$(eval "${resultArray[*]}" 2> /dev/null | grep -v '^:[0-9]*$')
 
-	COMPREPLY=( $(compgen -W "${result%%:*}" -- "$current") )
+	COMPREPLY=( $(compgen -W "${result}" -- "$current") )
 }
 
 _docker() {
@@ -5572,9 +5572,11 @@ _docker() {
 		wait
 	)
 
-	local known_plugin_commands=(
-		compose
-	)
+	local known_plugin_commands=()
+
+	if [ -f "$COMPOSE_PLUGIN_PATH" ] ; then
+		known_plugin_commands+=("compose")
+	fi
 
 	local experimental_server_commands=(
 		checkpoint
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -5486,6 +5486,23 @@ _docker_wait() {
 	_docker_container_wait
 }
 
+_docker_compose() {
+	local composePluginPath="${HOME}/.docker/cli-plugins/docker-compose"
+	local completionCommand="__completeNoDesc"
+  local resultArray=(${composePluginPath} ${completionCommand} compose)
+	for value in "${words[@]:2}"
+	do
+		if [[ "${value}" == "" ]] ; then
+			resultArray+=( "''" )
+		else
+			resultArray+=( "${value}" )
+		fi
+	done
+	local result=$(eval "${resultArray[*]}" 2> /dev/null)
+
+	COMPREPLY=( $(compgen -W "${result%%:*}" -- "$current") )
+}
+
 _docker() {
 	local previous_extglob_setting=$(shopt -p extglob)
 	shopt -s extglob
@@ -5555,11 +5572,15 @@ _docker() {
 		wait
 	)
 
+	local known_plugin_commands=(
+		compose
+	)
+
 	local experimental_server_commands=(
 		checkpoint
 	)
 
-	local commands=(${management_commands[*]} ${top_level_commands[*]})
+	local commands=(${management_commands[*]} ${top_level_commands[*]} ${known_plugin_commands[*]})
 	[ -z "$DOCKER_HIDE_LEGACY_COMMANDS" ] && commands+=(${legacy_commands[*]})
 
 	# These options are valid as global options for all client commands


contrib/completion/bash/docker
```diff
@@ -5485,6 +5485,23 @@ _docker_wait() {
 	_docker_container_wait
 }
 
+_docker_compose() {
+	local composePluginPath="${HOME}/.docker/cli-plugins/docker-compose"
+	local completionCommand="__completeNoDesc"
+  local resultArray=(${composePluginPath} ${completionCommand} compose)
```
mat007, line 5491: Indent?

contrib/completion/bash/docker
```diff
@@ -5554,11 +5571,15 @@ _docker() {
 		wait
 	)
 
+	local known_plugin_commands=(
```
thaJeztah, line 5574: Should we somehow detect if it's installed (to prevent it from providing (or trying to provide) completion for the plugin when it's not installed, which I _guess_ would fail)

contrib/completion/bash/docker
```diff
@@ -5485,6 +5485,23 @@ _docker_wait() {
 	_docker_container_wait
 }
 
+_docker_compose() {
+	local composePluginPath="${HOME}/.docker/cli-plugins/docker-compose"
```
thaJeztah, line 5489: This location would only work (by default) on Docker Desktop; on Linux, plugins may be installed elsewhere.

```bash
docker info --format '{{json .ClientInfo.Plugins}}' | jq .
[
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v0.9.1-beta3",
    "ShortDescription": "Docker App",
    "Experimental": true,
    "Name": "app",
    "Path": "/usr/libexec/docker/cli-plugins/docker-app"
  },
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v0.5.1-docker",
    "ShortDescription": "Build with BuildKit",
    "Name": "buildx",
    "Path": "/usr/libexec/docker/cli-plugins/docker-buildx"
  },
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v0.7.0",
    "ShortDescription": "Docker Scan",
    "Name": "scan",
    "Path": "/usr/libexec/docker/cli-plugins/docker-scan"
  }
]
```

thaJeztah: I should not that if we _do_ want to use  the above `docker info` to get that information, and we must do that on every request, we should have a look if we can somehow get the info without the CLI also making an API call (to the `/info` endpoint) 😬

contrib/completion/bash/docker
albers: I agree that this completion should only be added if the plugin is installed.

albers (changes requested):

albers (changes requested):
I think this PR is a very good starting point to gain some experience with generated completions.
We could start with a hardcoded option to include completion for the compose plugin, as suggested in this PR.

The completion script could check the well-known install locations (`/usr/libexec/docker/cli-plugins/` and `~/.docker/cli-plugins/docker-compose` for the presence of the docker-compose plugin file. In case of a match, it should activate completion for the plugin.

In subsequent PRs, the completion script should be extended to provide a generic solution for client plugins.

The generated completion as implemented in this PR works quite well, but should also be improved in subsequent PRs. For example, I miss support for completion of local files )`docker compose -f`), and completion of service names (`docker compose logs`)

thaJeztah:
for https://github.com/docker/cli/pull/3158#discussion_r661574610

> I should not that if we _do_ want to use the above `docker info` to get that information, and we must do that on every request, we should have a look if we can somehow get the info without the CLI also making an API call (to the `/info` endpoint) 😬

I opened https://github.com/docker/cli/pull/3179, which should be able to detect if we're only using local information (in which case it will skip making an API connection to the `/info` endpoint.

contrib/completion/bash/docker
```diff
@@ -5485,10 +5485,11 @@ _docker_wait() {
 	_docker_container_wait
 }
 
+COMPOSE_PLUGIN_PATH=$(docker info --format '{{json .ClientInfo.Plugins}}' | jq '.[] | select(.Name=="compose") | .Path')
```
albers, line 5488: Sadly, we cannot rely on jq being installed.

One solution could be
```bash
docker info --format '{{json .ClientInfo.Plugins}}' | sed -n 's/.*"Path":"\([^"]\+docker-compose\)".*/\1/p'
```

If we use `docker info` we should also ensure that it is not being called multiple times during one completion.
I can take care of this based on the existing [basic result caching](https://github.com/docker/cli/blob/5f07d7d5a12423c0bc1fb507f4d006ad0cdfef42/contrib/completion/bash/docker#L603-L608).

We can avoid the whole issue by checking hard coded well known install locations, i.e. `/usr/libexec/docker/cli-plugins/` and `~/.docker/cli-plugins/`.

albers (changes requested):

contrib/completion/bash/docker
ulyssessouza: Cool! It's replaced with `sed`!

The problem with a hardcoded path solution is that new locations can be configured in `~/.docker/config.json`.

albers (changes requested):
For consistency, please avoid unneccessary braces in variables, e.g. `"${value}"` should be just `"$value"`. For the same reason please pull up the `do`.

contrib/completion/bash/docker
albers: > The problem with a hardcoded path solution is that new locations can be configured in ~/.docker/config.json.

Good point. I did not know this.

contrib/completion/bash/docker
```diff
@@ -5485,6 +5485,24 @@ _docker_wait() {
...
 }
 
+COMPOSE_PLUGIN_PATH=$(docker info --format '{{json .ClientInfo.Plugins}}' | sed -n 's/.*"Path":"\([^"]\+docker-compose\)".*/\1/p')
+
+_docker_compose() {
+	local completionCommand="__completeNoDesc"
+	local resultArray=(${COMPOSE_PLUGIN_PATH} ${completionCommand} compose)
+	for value in "${words[@]:2}"
+	do
+		if [[ "${value}" == "" ]] ; then
```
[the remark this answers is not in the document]
ulyssessouza: Done

contrib/completion/bash/docker
```diff
@@ -5554,11 +5572,17 @@ _docker() {
 		wait
 	)
 
+	local known_plugin_commands=()
+
+	if [[ "${COMPOSE_PLUGIN_PATH}" != "" && "${COMPOSE_PLUGIN_PATH}" != "null" ]]; then
```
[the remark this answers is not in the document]
ulyssessouza: Done

contrib/completion/bash/docker
```diff
@@ -5485,6 +5485,23 @@ _docker_wait() {
...
+	local completionCommand="__completeNoDesc"
+	local resultArray=($COMPOSE_PLUGIN_PATH $completionCommand compose)
+	for value in "${words[@]:2}"; do
+		if [ -z "$value" ]; then
+			resultArray+=( "''" )
+		else
+			resultArray+=( "$value" )
+		fi
+	done
+	local result=$(eval "${resultArray[*]}" 2> /dev/null)
```
albers, line 5500: If we use something like
```suggestion
	local result=$(eval "${resultArray[*]}" 2> /dev/null | grep -v '^:[0-9]*$')
```
here in order to remove the `:<number>` line, we can get rid of the `%%:*` two lines below (which I find unintuitive and a bit dangerous because it mangles every completion word which happens to include a colon (image:tag)).

Do you know what the `:4` or similar line returned by cobra means?

contrib/completion/bash/docker
```diff
@@ -5485,6 +5485,23 @@ _docker_wait() {
...
+
+	COMPREPLY=( $(compgen -W "${result%%:*}" -- "$current") )
```
albers, line 5502: see above, might mangle completions that contain a colon.

albers (changes requested):
Very well.
One last nit and a question.

albers (approved):
LGTM, thanks!
ping @thaJeztah

thaJeztah (approved):
LGTM

tried it inside a container, and looks to work ok 👍 

I also see https://github.com/spf13/cobra/pull/1146 was merged, which looks to be improving cobra's completion scripts as well (instead of generating massive scripts)

ulyssessouza:
Thanks a lot @albers & @thaJeztah !

albers:
@ulyssessouza Please keep me up to date with your further work on generated completions.

Issue #923, opened 2018-03-06
docker man pages need some love


kolyshkin:
Executive summary: there are a number of docker-* man pages available under [man/](https://github.com/docker/cli/tree/master/man), they all need some love.

### Formatting
For one thing, formatting/typesetting needs to be cleaned up. The rules (from the top of my head) are:
 - literals (like option names, literal values like true or false, or examples) should be **bold**;
 - variable parts (like option values to be substituted) should be *italic* (note this is frequently rendered as underlined when viewing a man page in a terminal);
 - optional parts should be enclosed into square brackets;
 - in general, `=` between the **--option** and the *value* should be omitted;
 - angle brackets (like `<value>`) should not be used.

For more rules and examples, check [man-pages(7)](http://man7.org/linux/man-pages/man7/man-pages.7.html#STYLE_GUIDE).

For example, here's the correct formatting:

> **--memory-limit** *value*[*SUFFIX*]
> 
> Sets the memory limit. *SUFFIX* is optional and can be one of **b** (bytes), **k** (kilobytes), **m**
> (megabytes), or **g** (gigabytes).

#### Source:
```
**--memory-limit** *value*[*SUFFIX*]

Sets the memory limit. *SUFFIX* is optional and can be one of
**b** (bytes), **k** (kilobytes), **m** (megabytes), or **g** (gigabytes).
```

As an example of such work, please see https://github.com/docker/cli/pull/922

silvin-lubecki:
Thank you for taking care of this @chriscoffee !

kolyshkin:
@chriscoffee I suggest start with a small patch, fixing just a few things first. Don't try to fix all at once.

chriscoffee:
> @chriscoffee I suggest start with a small patch, fixing just a few things first. Don't try to fix all at once.

Sure, I'll put in some smaller PRs. Thanks for the advice 👍

Joshua-Rowe:
I was thinking about working on this during the weekend. Can anyone confirm that I would not be duplicating work etc. I could not find an open PR from @chriscoffee or a branch on his fork. Not sure if there are guidelines with deciding when someone else can work on something claimed by another from a quick skim thorugh the [guidelines](https://github.com/docker/cli/blob/master/CONTRIBUTING.md#docker-community-guidelines). Thanks!

chriscoffee:
@Joshua-Rowe feel free to take this, apologies. I got some initial work done on it then got sidetracked with other things.

[...]

PR #3163, opened 2021-06-26
Cleaned up formatting/typesetting.

KGB33:
Changed backticks to bold/italics, removed angle brackets.

Updated `cli/man/src/container/attach.md` to match formatting specified in #923
By changing in-line code i.e. backticks, to bold (for literals) and italics (for variables).

### DIFF
--- a/man/src/container/attach.md
+++ b/man/src/container/attach.md
@@ -10,7 +10,7 @@ configurable key sequence. The default sequence is `CTRL-p CTRL-q`. You
 configure the key sequence using the **--detach-keys** option or a configuration
 file. See **config-json(5)** for documentation on using a configuration file.
 
-It is forbidden to redirect the standard input of a `docker attach` command while
+It is forbidden to redirect the standard input of a **docker attach** command while
 attaching to a tty-enabled container (i.e.: launched with `-t`).
 
 # Override the detach sequence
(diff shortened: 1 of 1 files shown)


KGB33:
I tested out the example and ran into an error:

```bash
ID=$(sudo docker run -d fedora /usr/bin/top -b)
docker: Error response from daemon: failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/usr/bin/top": stat /usr/bin/top: no such file or directory: unknown.
```

Changing fedora to ubuntu solved the issue without changing it too much. I also pinned the ubuntu tag to help prevent another issue from happening in the future.

StefanScherer (commented):
Thanks @KGB33 for the contribution.
Could you explain why the formatting change is needed? 
I can see that we have many similar man pages and they look good to me as they are. 🤔 

Indeed, the `fedora` image does not contain `top` command, but eg. `alpine` does or as you mentioned the `ubuntu` image as well.

KGB33:
I changed the formatting to match what was discribed in Issue #923. 

This is a very small PR, If the formatting issues that #923 discribes is still the correct I'd be happy to update more of the man pages.

StefanScherer (approved):
Thank you @KGB33 for your contribution.
Thanks for the explanation, yes making small PR's make sense.
LGTM

thaJeztah (approved):
LGTM

thanks!

PR #3177, opened 2021-07-05
[WIP] Add "platform" column to docker ps

thaJeztah:
With this change:

    docker create armhf/hello-world
    docker create hello-world

    docker ps -a
    CONTAINER ID   IMAGE               PLATFORM      COMMAND    CREATED          STATUS                      PORTS     NAMES
    a020fcae5f4d   hello-world         linux/amd64   "/hello"   5 seconds ago    Exited (0) 3 seconds ago              vigorous_yalow
    499a18874987   armhf/hello-world   linux/arm64   "/hello"   58 seconds ago   Exited (0) 57 seconds ago             hardcore_brown

    docker ps -a --format=raw
    container_id: a020fcae5f4d
    image: hello-world
    platform: linux/amd64
    command: "/hello"
    created_at: 2021-07-05 07:08:19 +0000 UTC
    state: exited
    status: Exited (0) About an hour ago
    names: vigorous_yalow
    labels:
    ports:

    container_id: 499a18874987
    image: armhf/hello-world
    platform: linux/arm64
    command: "/hello"
    created_at: 2021-07-05 07:07:26 +0000 UTC
    state: exited
    status: Exited (0) About an hour ago
   names: hardcore_brown
    labels:
    ports:

    docker ps -a --format '{{json .}}' | jq .
    {
      "Command": "\"/hello\"",
      "CreatedAt": "2021-07-05 07:08:19 +0000 UTC",
      "ID": "a020fcae5f4d",
      "Image": "hello-world",
      "Labels": "",
      "LocalVolumes": "0",
      "Mounts": "",
      "Names": "vigorous_yalow",
      "Networks": "bridge",
      "Platform": "linux/amd64",
      "Ports": "",
      "RunningFor": "52 seconds ago",
      "Size": "0B (virtual 13.3kB)",
      "State": "exited",
      "Status": "Exited (0) 51 seconds ago"
    }
    {
      "Command": "\"/hello\"",
      "CreatedAt": "2021-07-05 07:07:26 +0000 UTC",
      "ID": "499a18874987",
      "Image": "armhf/hello-world",
      "Labels": "",
      "LocalVolumes": "0",
      "Mounts": "",
      "Names": "hardcore_brown",
      "Networks": "bridge",
      "Platform": "linux/arm64",
      "Ports": "",
      "RunningFor": "About a minute ago",
      "Size": "0B (virtual 1.64kB)",
      "State": "exited",
      "Status": "Exited (0) About a minute ago"
    }

To be discussed: what should the presentation be of platform if an "os variant" is present?

For example (from `docker manifest inspect hello-world`);

```json
{
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "size": 1125,
  "digest": "sha256:90e120baffe5afa60dd5a24abcd051db49bd6aee391174da5e825ee6ee5a12a0",
  "platform": {
    "architecture": "amd64",
    "os": "windows",
    "os.version": "10.0.17763.1999"
  }
}
```


thaJeztah:
Some tests are failing because I didn't update them yet for the new output.

We need to discuss whether or not the `PLATFORM` column should be shown by default, or if it should be "optional" (as output is already quite wide).

Mixed platforms may _currently_ be something that occurs often (thinking of Docker Desktop on Apple m1), but may become less common (once arm64 containers become more "standard?"). From that perspective (architecture being useful information during the "transition"), perhaps it should remain "optional".

tianon:
IMO most users will not be running multiplatform on one system (esp. outside the M1 userbase), so we shouldn't include it by default (the default output is already v. v. wide, as you've noted).

thaJeztah:
> IMO most users will not be running multiplatform on one system (esp. outside the M1 userbase), so we shouldn't include it by default (the default output is already v. v. wide, as you've noted).

Yes, I agree. I started with this and thought "this is useful information", but when I started to update the docs, I started to doubt if that was a good idea.

Let me change this to;

- Not show by default (in the default `table` view)
- Add an _example_ that shows how to use the new information in a custom `--format` (or custom format in `~/.docker/config.json`)
- I'll _keep_ it by default in the `--format=raw` output (although I doubt many users use that formatting)


Perhaps one remaining question: what do we want `docker ps -a --format '{{json .}}' ` to show?

1. `Platform: "os/arch/variant"` (string presentation of `Platform`)
2. `Platform: {"OS": "os", "Architecture": "arch" ....}`
3. combination of the above? (e.g. `OSArch` field with combined, and `Platform` with struct?)


- `1.` makes it more easy to use (`--format '{{ . }}`), but will "hide" the nitty-gritty details
- `2.` shows all details (closer to API response), but cumbersome to use if you "just want os/arch"
- `3.` will create a lot of noise. Also wondering if it's possible (haven't tried) to make `{{ json .Platform}}` return the "struct" and `{{ .Platform}}` return the string presentation. (if anyone knows the answer to that or has an example, let me know!)

tianon:
> * `3.` will create a lot of noise. Also wondering if it's possible (haven't tried) to make `{{ json .Platform}}` return the "struct" and `{{ .Platform}}` return the string presentation. (if anyone knows the answer to that or has an example, let me know!)

Isn't that "just" a struct that also has a `String()` method?  (This would definitely be my preference.)

https://play.golang.org/p/i4FOcAfAg7q

thaJeztah:
- superseded by https://github.com/docker/cli/pull/6042

PR #3197, opened 2021-07-19
Bump go 1.16.6

thaJeztah:
Keeping the dockerfiles/Dockerfile.cross image at 1.13, as we don't
have more current versions of that image. However, I don't think it's
still used, so we should remove it.

### DIFF
--- a/dockerfiles/Dockerfile.e2e
+++ b/dockerfiles/Dockerfile.e2e
@@ -1,4 +1,4 @@
-ARG GO_VERSION=1.13.15
+ARG GO_VERSION=1.16.6
 
 # Use Debian based image as docker-compose requires glibc.
 FROM golang:${GO_VERSION}-buster
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,17 @@
 # syntax=docker/dockerfile:1.3
 
 ARG BASE_VARIANT=alpine
-ARG GO_VERSION=1.13.15
+ARG GO_VERSION=1.16.6
 
 FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-${BASE_VARIANT} AS gostable
-FROM --platform=$BUILDPLATFORM golang:1.16-${BASE_VARIANT} AS golatest	
+FROM --platform=$BUILDPLATFORM golang:1.17rc1-${BASE_VARIANT} AS golatest
 
-FROM gostable AS go-linux	
-FROM golatest AS go-darwin
-FROM golatest AS go-windows-amd64
-FROM golatest AS go-windows-386
-FROM golatest AS go-windows-arm
-FROM --platform=$BUILDPLATFORM golang:1.17rc1-${BASE_VARIANT} AS go-windows-arm64
+FROM gostable AS go-linux
+FROM gostable AS go-darwin
+FROM gostable AS go-windows-amd64
+FROM gostable AS go-windows-386
+FROM gostable AS go-windows-arm
+FROM golatest AS go-windows-arm64
 FROM go-windows-${TARGETARCH} AS go-windows
 
 FROM --platform=$BUILDPLATFORM tonistiigi/xx@sha256:620d36a9d7f1e3b102a5c7e8eff12081ac363828b3a44390f24fa8da2d49383d AS xx
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ clone_folder: c:\gopath\src\github.com\docker\cli
 
 environment:
   GOPATH: c:\gopath
-  GOVERSION: 1.13.15
+  GOVERSION: 1.16.6
   DEPVERSION: v0.4.1
 
 install:
--- a/dockerfiles/Dockerfile.binary-native
+++ b/dockerfiles/Dockerfile.binary-native
@@ -1,4 +1,4 @@
-ARG GO_VERSION=1.13.15
+ARG GO_VERSION=1.16.6
 
 FROM    golang:${GO_VERSION}-alpine
 
--- a/dockerfiles/Dockerfile.dev
+++ b/dockerfiles/Dockerfile.dev
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.3
 
-ARG GO_VERSION=1.13.15
+ARG GO_VERSION=1.16.6
 
 FROM golang:${GO_VERSION}-alpine AS golang
 ENV  CGO_ENABLED=0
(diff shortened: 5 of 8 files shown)


thaJeztah:
Interesting; looks like for some reason, the test doesn't pick up the `HTTP_PROXY` env var. In that case `transport.Proxy(request)` returns `nil`;

```
=== Failed
=== FAIL: cli/command TestNewAPIClientFromFlagsWithHttpProxyEnv (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x67a14a]

goroutine 24 [running]:
testing.tRunner.func1.2(0xaa9ca0, 0xfe73a0)
	/usr/local/go/src/testing/testing.go:1143 +0x332
testing.tRunner.func1(0xc000083380)
	/usr/local/go/src/testing/testing.go:1146 +0x4b6
panic(0xaa9ca0, 0xfe73a0)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
net/url.(*URL).String(0x0, 0xc000083380, 0x0)
	/usr/local/go/src/net/url/url.go:813 +0x4a
github.com/docker/cli/cli/command.TestNewAPIClientFromFlagsWithHttpProxyEnv(0xc000083380)
	/go/src/github.com/docker/cli/cli/command/cli_test.go:98 +0x377
testing.tRunner(0xc000083380, 0xb98b50)
	/usr/local/go/src/testing/testing.go:1193 +0xef
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1238 +0x2b3


DONE 1568 tests, 3 skipped, 1 failure in 216.761s
make: *** [Makefile:22: test-coverage] Error 1

Exited with code exit status 2
CircleCI received exit code 2
```

thaJeztah:
I suspect the problem is that stdlib uses a `sync.Once` to get the proxy;  https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/net/http/transport.go#L815-L826

So if some codepath has checked if a proxy must be used, and at that moment  no proxy env-vars were set, then it will no longer be called after that.

```
func envProxyFunc() func(*url.URL) (*url.URL, error) {
	envProxyOnce.Do(func() {
		envProxyFuncValue = httpproxy.FromEnvironment().ProxyFunc()
	})
	return envProxyFuncValue
}

// resetProxyConfig is used by tests.
func resetProxyConfig() {
	envProxyOnce = sync.Once{}
	envProxyFuncValue = nil
}
```

There is a `resetProxyConfig()` function that they used in testing, but that can't be used externally.

thaJeztah:
Hmm.. actually, this commit may be the cause; https://github.com/golang/net/commit/7b1cca2348c07eb09fef635269c8e01611260f9f

We set up a `HTTP(S)_PROXY`, but the scheme we use to connect is `tcp://`. The commit above adds an extra check and if the scheme is not `http://` it will return "no proxy" (`nil`)

thaJeztah:
This is gonna be fun, because we use `tcp://` by default. I'll be honest; I don't know why we used that in the first place (and not `http://` and `https://`, but it's there, and (IIUC) it means that with go 1.16, connections specifying one of those will ignore proxy configuration; https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/opts/hosts.go#L21-L24

thaJeztah:
> This is gonna be fun, because we use `tcp://` by default. I'll be honest; I don't know why we used that in the first place (and not `http://` and `https://`, but it's there, and (IIUC) it means that with go 1.16, connections specifying one of those will ignore proxy configuration;


Update to the above:

So, it turned out that the actual code-path that the CLI uses does not (directly)
hit this problem. The issue in `TestNewAPIClientFromFlagsWithHttpProxyEnv` was
that;

- It was a unit-test, and Golang uses a `sync.Once` to configure the proxy to
  be used. Depending on what code was executed before that test, the proxy would
  already be set (based on the first request made), and therefore potentially hit
  the wrong value (see https://github.com/docker/cli/pull/3203#issuecomment-884875230).
  Changing the test to be an integration/e2e test fixes that (https://github.com/docker/cli/pull/3221)
- The test [manually created a request using `tcp://` as scheme](https://github.com/thaJeztah/cli/blob/d9cdbb0a4c6be0e478dd850f60877816e30f3631/cli/command/cli_test.go#L94-L96).
  while that revealed that the transport itself would get the incorrect proxy
  (no proxy) when used with a non-`http(s)://` scheme, our standard code does
  not use


During initialisation;

1. `client.NewClientWithOpts()` (which is used to initialize the client sets up a default HTTP client, based on the default host (`unix:///var/run/docker.sock` on Linux, `npipe:////./pipe/docker_engine` on Windows). Users can override this with a custom option, but (see 2.), this should not matter.
https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/client/client.go#L120

2. `client.defaultClient()` parses the `host`, and gets the scheme from it. It passes that scheme to `sockets.ConfigureTransport` https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/client/client.go#L158-L169

3. `sockets.ConfigureTransport` _only_ uses the scheme to switch between setting up `unix`, `npipe`, or "other" transports https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/vendor/github.com/docker/go-connections/sockets/sockets.go#L21-L37

4. while it sets `transport.Proxy` to `http.ProxyFromEnvironment`, that function is not called until a request is made https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/vendor/github.com/docker/go-connections/sockets/sockets.go#L28

5. the _client's_ scheme is not set based on `DOCKER_HOST`. The `tcp://` scheme is simply ignored, but depending on whether or not TLS is configured, is overwritten with `http://` or `https://` https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/client/client.go#L141-L153

6. Whenever the client makes a request, it does not use the Transport's scheme, but sets the scheme for the request to what's configured on the client, which (see above) would be `unix://`, `npipe://`, or `http(s)://` (not sure what happens with `ssh://`) https://github.com/moby/moby/blob/b9ad7b96bd86e5f632e5fb4f36f98dcc145014fc/client/request.go#L99-L100

So, although there still are some bits to look into, for example, if the client is initialized with a custom scheme, using [the `WithScheme()` option](https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/client/options.go#L122-L128), the regular code-path won't be affected by the Golang change.

thaJeztah:
@silvin-lubecki @StefanScherer PTAL - this is ready for review now 🤗

dockerfiles/Dockerfile.e2e
```diff
@@ -39,6 +39,7 @@ ARG GITCOMMIT
 ENV VERSION=${VERSION}
 ENV GITCOMMIT=${GITCOMMIT}
 ENV DOCKER_BUILDKIT=1
+#ENV GO111MODULE=off
```
silvin-lubecki, line 42: question: Should we keep this commented line?

thaJeztah: Arf, nope, my mistake. Let me fix

silvin-lubecki (approved):
LGTM

PR #3203, opened 2021-07-22
cli/command: don't use client.CustomHTTPHeaders()

thaJeztah:
It's the only use of this function, and it's better to check that
the client actually sends the header.

### DIFF
--- a/cli/command/cli_test.go
+++ b/cli/command/cli_test.go
@@ -6,8 +6,11 @@ import (
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"runtime"
+	"strings"
 	"testing"
 
 	cliconfig "github.com/docker/cli/cli/config"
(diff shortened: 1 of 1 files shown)


thaJeztah:
Hmm... looks like this fails currently, because golang only resolves the proxy configuration once (using a `sync.Once`) and because this test actually makes a request, expectations for other tests fail 😞 

```
--- FAIL: TestNewAPIClientFromFlagsWithHttpProxyEnv (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x67898a]

goroutine 49 [running]:
testing.tRunner.func1(0xc000334a00)
	/usr/local/go/src/testing/testing.go:874 +0x3a3
panic(0xb708a0, 0x12844d0)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
net/url.(*URL).String(0x0, 0xc000334a00, 0x0)
	/usr/local/go/src/net/url/url.go:775 +0x4a
github.com/docker/cli/cli/command.TestNewAPIClientFromFlagsWithHttpProxyEnv(0xc000334a00)
	/go/src/github.com/docker/cli/cli/command/cli_test.go:111 +0x39c
testing.tRunner(0xc000334a00, 0xc94ba0)
	/usr/local/go/src/testing/testing.go:909 +0xc9
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:960 +0x350
FAIL	github.com/docker/cli/cli/command	0.013s
```

silvin-lubecki (approved):
LGTM 👍

PR #3213, opened 2021-07-26
cli/context: ignore linting warnings about RFC 1423 encryption

thaJeztah:
relates to https://github.com/docker/cli/issues/3212

From https://go-review.googlesource.com/c/go/+/264159

> It's unfortunate that we don't implement PKCS#8 encryption so we can't
> recommend an alternative but PEM encryption is so broken that it's worth
> deprecating outright.

When linting on Go 1.16:

    cli/context/docker/load.go:69:6: SA1019: x509.IsEncryptedPEMBlock is deprecated: Legacy PEM encryption as specified in RFC 1423 is insecure by design. Since it does not authenticate the ciphertext, it is vulnerable to padding oracle attacks that can let an attacker recover the plaintext.  (staticcheck)
            if x509.IsEncryptedPEMBlock(pemBlock) {
               ^
    cli/context/docker/load.go:70:20: SA1019: x509.DecryptPEMBlock is deprecated: Legacy PEM encryption as specified in RFC 1423 is insecure by design. Since it does not authenticate the ciphertext, it is vulnerable to padding oracle attacks that can let an attacker recover the plaintext.  (staticcheck)
                keyBytes, err = x509.DecryptPEMBlock(pemBlock, []byte(c.TLSPassword))
                                ^

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

Issue #3212, opened 2021-07-26
context: consider deprecating support for legacy PEM encryption as specified in RFC 1423


thaJeztah:
PR's https://github.com/docker/cli/pull/1501 / https://github.com/docker/cli/pull/1500 added support for TLS config using  encryption as specified in RFC 1423, however, that functionality has been deprecated in Go, and causes linting to fail on Go 1.16;

    cli/context/docker/load.go:69:6: SA1019: x509.IsEncryptedPEMBlock is deprecated: Legacy PEM encryption as specified in RFC 1423 is insecure by design. Since it does not authenticate the ciphertext, it is vulnerable to padding oracle attacks that can let an attacker recover the plaintext.  (staticcheck)
            if x509.IsEncryptedPEMBlock(pemBlock) {
               ^
    cli/context/docker/load.go:70:20: SA1019: x509.DecryptPEMBlock is deprecated: Legacy PEM encryption as specified in RFC 1423 is insecure by design. Since it does not authenticate the ciphertext, it is vulnerable to padding oracle attacks that can let an attacker recover the plaintext.  (staticcheck)
                keyBytes, err = x509.DecryptPEMBlock(pemBlock, []byte(c.TLSPassword))
                                ^

From https://go-review.googlesource.com/c/go/+/264159

> It's unfortunate that we don't implement PKCS#8 encryption so we can't
> recommend an alternative but PEM encryption is so broken that it's worth
> deprecating outright.

We should consider deprecating this functionality (log a warning and/or produce an error)

thaJeztah:
Relates to https://github.com/docker/cli/pull/3213

justincormack:
This is being used to decrypt an encrypted private key with a supplied password. This is pretty pointless as the key is sitting next to the password in the filesystem, so encrypting the key (with broken encryption) serves no purpose. I would remove the support for decrypting keys and error requiring the key is in plaintext.

thaJeztah:
closing; this was done in https://github.com/docker/cli/pull/3220

PR #3218, opened 2021-07-28
context: deprecate support for encrypted TLS private keys

thaJeztah:
relates to https://github.com/docker/cli/issues/3212
follow-up to https://github.com/docker/cli/pull/3213

> Legacy PEM encryption as specified in RFC 1423 is insecure by design. Since
> it does not authenticate the ciphertext, it is vulnerable to padding oracle
> attacks that can let an attacker recover the plaintext

From https://go-review.googlesource.com/c/go/+/264159

> It's unfortunate that we don't implement PKCS#8 encryption so we can't
> recommend an alternative but PEM encryption is so broken that it's worth
> deprecating outright.

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -255,7 +255,7 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...Initialize
 		if tlsconfig.IsErrEncryptedKey(err) {
 			passRetriever := passphrase.PromptRetrieverWithInOut(cli.In(), cli.Out(), nil)
 			newClient := func(password string) (client.APIClient, error) {
-				cli.dockerEndpoint.TLSPassword = password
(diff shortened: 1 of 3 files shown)


silvin-lubecki (approved):
LGTM

silvin-lubecki:
@thaJeztah this PR needs to be rebased on top of `main` to remove the lint errors 🙏

thaJeztah:
oh! 🤦 no it's because I deprecated the `TLSPassword` field, and we still use it, so need to add another `//nolint`

```
cli/command/cli.go:258:5: SA1019: cli.dockerEndpoint.TLSPassword is deprecated: Use of encrypted TLS private keys has been deprecated, and will be removed in a future release. Golang has deprecated support for legacy PEM encryption (as specified in RFC 1423), as it is insecure by design (see https://go-review.googlesource.com/c/go/+/264159).  (staticcheck)
				cli.dockerEndpoint.TLSPassword = password
				^
```

thaJeztah:
Whoop; it's green now @silvin-lubecki ptal

silvin-lubecki (approved):

PR #3219, opened 2021-07-28
[20.10 backport] context: deprecate support for encrypted TLS private keys

thaJeztah:
backport of https://github.com/docker/cli/pull/3218 and https://github.com/docker/cli/pull/3213

relates to https://github.com/docker/cli/issues/3212

> Legacy PEM encryption as specified in RFC 1423 is insecure by design. Since
> it does not authenticate the ciphertext, it is vulnerable to padding oracle
> attacks that can let an attacker recover the plaintext

From https://go-review.googlesource.com/c/go/+/264159

> It's unfortunate that we don't implement PKCS#8 encryption so we can't
> recommend an alternative but PEM encryption is so broken that it's worth
> deprecating outright.


thaJeztah:
@silvin-lubecki @StefanScherer ptal

silvin-lubecki (approved):
LGTM

PR #3230, opened 2021-08-03
info: skip client-side warning about seccomp profile on API >= 1.42

thaJeztah:
This warning will be moved to the daemon-side, similar to how it returns
other warnings. There's work in progress to change the name of the default
profile, so we may need to backport this change to prevent existing clients
from printing an incorrect warning if they're connecting to a newer daemon.

relates to https://github.com/moby/moby/pull/42481


thaJeztah:
@justincormack @silvin-lubecki PTAL

silvin-lubecki (approved):

cpuguy83 (changes requested):
I think we should still show a warning for older API versions.
Also lets make sure to hold off on this until the change is in the engine.

cli/command/system/info.go
```diff
@@ -254,9 +255,6 @@ func prettyPrintServerInfo(dockerCli command.Cli, info types.Info) []error {
 					for _, o := range so.Options {
 						switch o.Key {
 						case "profile":
-							if o.Value != "default" {
```
thaJeztah, line 257: @cpuguy83 updated; PTAL

Alternatively, I could do:

```suggestion
							if versions.LessThan(dockerCli.Client().ClientVersion(), "1.42") && o.Value != "default" && o.Value != "builtin" {
```

Which would bring a smaller patch, but I tried to also group the warning with the other warnings (printed at the end) instead of having it "half-way" the `docker info` output here. Happy to change though if you think that's an issue.

thaJeztah:
ping @cpuguy83 PTAL 🤗

cpuguy83 (approved):
LGTM

PR #3240, opened 2021-08-10
vendor: update docker to current master (API v1.42)

thaJeztah:
full diff: https://github.com/docker/docker/compare/25917217cab38eab40c3db0010b915258f4a8491...343665850e3a7ee160989b33b39477b6dd3f15cb

also updates swarmkit (no local changes) https://github.com/docker/swarmkit/compare/2dcf70aafdc9ea55af3aaaeca440638cde0ecda6...3629f50980f6c0dd5ccd7dbfa0956b57ea0cd78d

**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3243, opened 2021-08-10
build: fix AddDockerfileToBuildContext not de-referencing tar header template

thaJeztah:
relates to https://github.com/docker/cli/pull/3240#issuecomment-896303278

Commit https://github.com/docker/docker/commit/73aef6edfe5ae533f9cf547f924c98a9b0f7be59 (https://github.com/moby/moby/pull/42515) modified archive.ReplaceFileTarWrapper to set the Name field in the tar header, if the field was not set.

That change exposed an issue in how a Dockerfile from stdin was sent to the daemon. When attempting to build using a build-context, and a Dockerfile from stdin, the following happened:

```bash
mkdir build-stdin && cd build-stdin && echo hello > hello.txt

DOCKER_BUILDKIT=0 docker build --no-cache -t foo -f- . <<'EOF'
FROM alpine
COPY . .
EOF

Sending build context to Docker daemon  2.607kB
Error response from daemon: dockerfile parse error line 1: unknown instruction: .DOCKERIGNORE
```

Removing the `-t foo`, oddly lead to a different failure:

```bash
DOCKER_BUILDKIT=0 docker build --no-cache -f- . <<'EOF'
FROM alpine
COPY . .
EOF

Sending build context to Docker daemon  2.581kB
Error response from daemon: Cannot locate specified Dockerfile: .dockerfile.701d0d71fb1497d6a7ce
```

From the above, it looks like the tar headers got mangled, causing (in the first case) the daemon to use the build-context tar as a plain-text file, and therefore parsing it as Dockerfile, and in the second case, causing it to not being able to find the Dockerfile in the context.

I noticed that both TarModifierFuncs were using the same `hdrTmpl` struct, which looks to caused them to step on each other's toes. Changing them to each initialize their own struct made the issue go away.

After this change:

```bash
DOCKER_BUILDKIT=0 docker build --no-cache -t foo -f- . <<'EOF'
FROM alpine
COPY . .
EOF
Sending build context to Docker daemon  2.607kB
Step 1/2 : FROM alpine
 ---> d4ff818577bc
Step 2/2 : COPY . .
 ---> 556f745e6938
Successfully built 556f745e6938
Successfully tagged foo:latest

DOCKER_BUILDKIT=0 docker build --no-cache -f- . <<'EOF'
FROM alpine
COPY . .
EOF

Sending build context to Docker daemon  2.607kB
Step 1/2 : FROM alpine
 ---> d4ff818577bc
Step 2/2 : COPY . .
 ---> aaaee43bec5e
Successfully built aaaee43bec5e
```

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/image/build/context.go
+++ b/cli/command/image/build/context.go
@@ -378,26 +378,32 @@ func AddDockerfileToBuildContext(dockerfileCtx io.ReadCloser, buildCtx io.ReadCl
 		return nil, "", err
 	}
 	now := time.Now()
-	hdrTmpl := &tar.Header{
-		Mode:       0600,
-		Uid:        0,
-		Gid:        0,
-		ModTime:    now,
-		Typeflag:   tar.TypeReg,
-		AccessTime: now,
-		ChangeTime: now,
-	}
 	randomName := ".dockerfile." + stringid.GenerateRandomID()[:20]
 
 	buildCtx = archive.ReplaceFileTarWrapper(buildCtx, map[string]archive.TarModifierFunc{
 		// Add the dockerfile with a random filename
-		randomName: func(_ string, h *tar.Header, content io.Reader) (*tar.Header, []byte, error) {
-			return hdrTmpl, file, nil
+		randomName: func(_ string, _ *tar.Header, _ io.Reader) (*tar.Header, []byte, error) {
+			header := &tar.Header{
+				Name:       randomName,
+				Mode:       0600,
+				ModTime:    now,
+				Typeflag:   tar.TypeReg,
+				AccessTime: now,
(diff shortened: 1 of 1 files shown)


thaJeztah:

tonistiigi (approved):
Moby code should make a copy if it modifies the properties in certain conditions.

thaJeztah:
> Moby code should make a copy if it modifies the properties in certain conditions.

Thanks for reviewing! Yes, I wanted to look more into the moby code as well. This change made sense on its own though (and looks to fix the issue)

crazy-max (approved):

PR #3245, opened 2021-08-11
create/run: remove default --stop-signal

thaJeztah:
relates to https://github.com/moby/moby/pull/42717

The DefaultStopSignal const has been deprecated, because the daemon already
handles a default value. The current code did not actually send the default
value unless the flag was set, which also made the flag description incorrect,
because in that case, the _daemon's_ default would be used, which could
potentially be different as was specified here.

This patch removes the default value from the flag, leaving it to the daemon
to set a default.

```
Remove default value from `--stop-signal` flag, as it may not reflect the actual default used by the daemon.
```

### DIFF
--- a/man/docker-run.1.md
+++ b/man/docker-run.1.md
@@ -621,13 +621,16 @@ incompatible with any restart policy other than `none`.
    For the `overlay2` storage driver, the size option is only available if the backing fs is `xfs` and mounted with the `pquota` mount option.
    Under these conditions, user can pass any size less than the backing fs size.
 
-**--stop-signal**=*SIGTERM*
-   Signal to stop the container. Default is SIGTERM.
+**--stop-signal**=""
+   Signal to stop the container.
 
    The `--stop-signal` flag sets the system call signal that will be sent to the
    container to exit. This signal can be a signal name in the format `SIG<NAME>`,
    for instance `SIGKILL`, or an unsigned number that matches a position in the
    kernel's syscall table, for instance `9`.
+
+   The default is defined by `STOPSIGNAL` in the image, or `SIGTERM` if the image
+   has no `STOPSIGNAL` defined.
 
 **--stop-timeout**
   Timeout (in seconds) to stop a container, or **-1** to disable timeout.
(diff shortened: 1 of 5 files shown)


man/docker-run.1.md
```diff
@@ -621,7 +621,7 @@ incompatible with any restart policy other than `none`.
    For the `overlay2` storage driver, the size option is only available if the backing fs is `xfs` and mounted with the `pquota` mount option.
    Under these conditions, user can pass any size less than the backing fs size.
 
-**--stop-signal**=*SIGTERM*
+**--stop-signal**=""
   Signal to stop a container. Default is SIGTERM.
```
tianon, line 625: Should this maybe note the fact that the default can be defined by the image or something? (`STOPSIGNAL` in `Dockerfile`)

```suggestion
  Signal to stop a container. Default is defined by `STOPSIGNAL` in the image or `SIGTERM` if not defined.
```

thaJeztah: Ah, yes, makes sense to add, let me add some of that; thanks!

thaJeztah:
I did some further rewriting of the docs, and moved a large part to https://github.com/docker/cli/pull/3261 (included in this PR; first two commits are from there)

cli/command/container/kill.go
```diff
@@ -32,7 +32,7 @@ func NewKillCommand(dockerCli command.Cli) *cobra.Command {
 	}
 
 	flags := cmd.Flags()
-	flags.StringVarP(&opts.signal, "signal", "s", "KILL", "Signal to send to the container")
+	flags.StringVarP(&opts.signal, "signal", "s", "SIGKILL", "Signal to send to the container")
```
thaJeztah, line 35: I'm considering to remove this default from the client as well, but keeping that for a follow-up

thaJeztah:
Rebased, to removed the changes from https://github.com/docker/cli/pull/3261 from this PR, and moved the `KILL -> SIGKILL` changes separate as well, as it's not directly related https://github.com/docker/cli/pull/3271

thaJeztah:
@tianon @silvin-lubecki PTAL

tianon (approved):
:eyes:

thaJeztah:
@silvin-lubecki PTAL 🤗

thaJeztah:
@silvin-lubecki @tonistiigi PTAL 🤗

silvin-lubecki (approved):
LGTM

PR #3254, opened 2021-08-15
Add top-level --completion flag, and integrate completion for plugins

### LINKED ISSUE #1257: Docker completion scripts should be easier to install
There is no simple and built-in way to install the completion scripts (see [the bash script installation](https://github.com/docker/cli/blob/master/contrib/completion/bash/docker#L20)).

A better way would be to output the scripts with a docker command, like [kubectl does](https://kubernetes.io/docs/tasks/tools/install-kubectl/#enabling-shell-autocompletion): 
```sh
$ docker completion bash
...
# install the zsh completion
$ echo "source <(docker completion zsh)" >> ~/.zshrc
```

thaJeztah:
fixes https://github.com/docker/cli/issues/1257

This patch adds a new top-level `--completion` flag, which allows the user to
print the completion script for their shell of choice.

We currently maintain hand-written completion scripts for bash, fish, and zsh,
of which the bash script is in a good state, but the fish and zsh scripts are
less well maintained.

We should review the quality of the fish and zsh scripts, and decide wether or
not we want to use the hand-written versions for those, or to replace them with
the auto-generated completion provided by Cobra.

This patch embeds the hand-written completion scripts using Go 1.16's "embed"
functionality, but also adds options to try/test/compare the Cobra completion
scripts. Currently, this patch allows for the following:

    docker --completion=bash        outputs the hand-written bash completion script
    docker --completion=bash-cobra  outputs cobra's generated completion script
    docker --completion=fish        outputs the hand-written fish completion script
    docker --completion=fish-cobra  outputs cobra's generated completion script
    docker --completion=powershell  outputs cobra's generated completion script
    docker --completion=zsh         outputs the hand-written bash completion script
    docker --completion=zsh-cobra   outputs cobra's generated completion script

While the hand-written bash completion script has definitions for commands and
options provided by the main CLI, it does not provide completions for commands
and options provided by plug-ins. Commits 1148163c3ec90fc78b8f385e065984989b018e8b
and 5a8d7d506cbbe7cd0df3ed3903cb27a90a51e142 (https://github.com/docker/cli/pull/3158) added support for completions for
the compose cli plug-in, but do so by directly calling the plug-in binary.

Cobra's completion scripts use two hidden subcommands that provide dynamic
generating of completion scripts: `__complete` and `__completeNoDesc`. This
patch adds support for both these subcommands, and implements code to extend
these commands to be plug-in aware: if either of those commands request completion
for an unknown command, the cli will look if a plug-in is install that provides
the command, and will call the `__complete` or `__completeNoDesc` command on the
plug-in.

With that, we are able to automatically provide completion scripts for all plug-
ins. For example:

    docker __complete compose r
    restart	Restart containers
    rm	Removes stopped service containers
    run	Run a one-off command on a service.
    :4
    Completion ended with directive: ShellCompDirectiveNoFileComp

    docker __complete scan --
    --accept-license	Accept using a third party scanning provider
    --dependency-tree	Show dependency tree with scan results
    --exclude-base	Exclude base image from vulnerability scanning (requires --file)
    --file	Dockerfile associated with image, provides more detailed results
    --file=	Dockerfile associated with image, provides more detailed results
    --group-issues	Aggregate duplicated vulnerabilities and group them to a single one (requires --json)
    --json	Output results in JSON format
    --login	Authenticate to the scan provider using an optional token (with --token), or web base token if empty
    --reject-license	Reject using a third party scanning provider
    --severity	Only report vulnerabilities of provided level or higher (low|medium|high)
    --severity=	Only report vulnerabilities of provided level or higher (low|medium|high)
    --token	Authentication token to login to the third party scanning provider
    --token=	Authentication token to login to the third party scanning provider
    --version	Display version of the scan plugin
    :0
    Completion ended with directive: ShellCompDirectiveDefault

    docker __completeNoDesc scan --
    --accept-license
    --dependency-tree
    --exclude-base
    --file
    --file=
    --group-issues
    --json
    --login
    --reject-license
    --severity
    --severity=
    --token
    --token=
    --version
    :0
    Completion ended with directive: ShellCompDirectiveDefault

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -11,7 +11,6 @@ import (
 	pluginmanager "github.com/docker/cli/cli-plugins/manager"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/commands"
-	cliflags "github.com/docker/cli/cli/flags"
 	"github.com/docker/cli/cli/version"
 	"github.com/docker/docker/api/types/versions"
 	"github.com/docker/docker/client"
@@ -27,12 +26,6 @@ var allowedAliases = map[string]struct{}{
 }
 
 func newDockerCommand(dockerCli *command.DockerCli) *cli.TopLevelCommand {
-	var (
-		opts    *cliflags.ClientOptions
-		flags   *pflag.FlagSet
-		helpCmd *cobra.Command
-	)
-
 	cmd := &cobra.Command{
 		Use:              "docker [OPTIONS] COMMAND [ARG...]",
 		Short:            "A self-sufficient runtime for containers",
@@ -44,15 +37,32 @@ func newDockerCommand(dockerCli *command.DockerCli) *cli.TopLevelCommand {
 				return command.ShowHelp(dockerCli.Err())(cmd, args)
 			}
 			return fmt.Errorf("docker: '%s' is not a docker command.\nSee 'docker --help'", args[0])
-
 		},
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			// Handle shell completion for plugin commands by forwarding the
+			// arguments to the plugin. Supporting completion is currently
+			// optional for plugins, so we ignore errors (if any).
+			if len(args) > 0 && cmd.Name() == cobra.ShellCompRequestCmd || cmd.Name() == cobra.ShellCompNoDescRequestCmd {
+				if _, _, err := cmd.Root().Find(args); err != nil {
+					// No such command. Try if we have a plugin that has this command
+					if err := tryPluginRun(dockerCli, cmd, args[0]); err == nil {
+						// We're done. Exit here to prevent the regular command
+						// from being executed.
+						os.Exit(0)
+					}
+					// Didn't find a plugin for this command either. Continue as
+					// usual, and have the default __complete / __completeNoDesc
+					// command handle the completion.
+					return nil
+				}
+			}
 			return isSupported(cmd, dockerCli)
 		},
 		Version:               fmt.Sprintf("%s, build %s", version.Version, version.GitCommit),
 		DisableFlagsInUseLine: true,
 	}
-	opts, flags, helpCmd = cli.SetupRootCommand(cmd)
+	opts, flags, helpCmd := cli.SetupRootCommand(cmd)
+	flags.String("completion", "", "Print the shell completion script for the specified shell (bash, fish, powershell, or zsh) and quit")
 	flags.BoolP("version", "v", false, "Print version information and quit")
 
 	setFlagErrorFunc(dockerCli, cmd)
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -5486,11 +5486,8 @@ _docker_wait() {
 	_docker_container_wait
 }
 
-COMPOSE_PLUGIN_PATH=$(docker info --format '{{json .ClientInfo.Plugins}}' | sed -n 's/.*"Path":"\([^"]\+docker-compose\)".*/\1/p')
-
-_docker_compose() {
-	local completionCommand="__completeNoDesc"
-	local resultArray=($COMPOSE_PLUGIN_PATH $completionCommand compose)
+__completeNoDesc() {
+	local resultArray=(docker __completeNoDesc $1)
 	for value in "${words[@]:2}"; do
 		if [ -z "$value" ]; then
 			resultArray+=( "''" )
@@ -5501,6 +5498,18 @@ _docker_compose() {
 	local result=$(eval "${resultArray[*]}" 2> /dev/null | grep -v '^:[0-9]*$')
 
 	COMPREPLY=( $(compgen -W "${result}" -- "$current") )
+}
+
+_docker_buildx() {
+	__completeNoDesc buildx
+}
+
+_docker_compose() {
+	__completeNoDesc compose
+}
+
+_docker_scan() {
+	__completeNoDesc scan
 }
 
 _docker() {
@@ -5572,11 +5581,8 @@ _docker() {
 		wait
 	)
 
-	local known_plugin_commands=()
-
-	if [ -f "$COMPOSE_PLUGIN_PATH" ] ; then
-		known_plugin_commands+=("compose")
-	fi
+	PLUGINS_INSTALLED=$(docker info --format '{{range .ClientInfo.Plugins}}{{ .Name }} {{end}}')
+	local known_plugin_commands=($PLUGINS_INSTALLED)
 
 	local experimental_server_commands=(
 		checkpoint
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -9,6 +9,7 @@ import (
 	"github.com/docker/cli/cli/command"
 	cliconfig "github.com/docker/cli/cli/config"
 	cliflags "github.com/docker/cli/cli/flags"
+	"github.com/docker/cli/contrib/completion"
 	"github.com/moby/term"
 	"github.com/morikuni/aec"
 	"github.com/pkg/errors"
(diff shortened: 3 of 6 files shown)


contrib/completion/bash/docker
```diff
@@ -5486,11 +5486,8 @@ _docker_wait() {
 	_docker_container_wait
 }
 
-COMPOSE_PLUGIN_PATH=$(docker info --format '{{json .ClientInfo.Plugins}}' | sed -n 's/.*"Path":"\([^"]\+docker-compose\)".*/\1/p')
-
-_docker_compose() {
-	local completionCommand="__completeNoDesc"
-	local resultArray=($COMPOSE_PLUGIN_PATH $completionCommand compose)
+__completeNoDesc() {
```
thaJeztah, line 5489: I need to check if I updated this correctly. The intent is to:

- change `/usr/local/lib/docker/cli-plugins/docker-compose __completeNoDesc <command>` to `docker __completeNoDesc <command>`
- the same should now work for all plugins, so, we should be able to just use all plugins:

        docker info --format '{{range .ClientInfo.Plugins}}{{ .Name }} {{end}}'
        buildx compose scan

thaJeztah:
@silvin-lubecki @albers @ulyssessouza ptal

contrib/completion/bash/docker
```diff
@@ -5503,6 +5500,18 @@ _docker_compose() {
 	COMPREPLY=( $(compgen -W "${result}" -- "$current") )
 }
 
+_docker_compose() {
```
silvin-lubecki, line 5503: typo: ~~_docker_compose~~ _docker_buildx

contrib/completion/bash/docker
```diff
@@ -5572,11 +5581,8 @@ _docker() {
 		wait
 	)
 
-	local known_plugin_commands=()
-
-	if [ -f "$COMPOSE_PLUGIN_PATH" ] ; then
-		known_plugin_commands+=("compose")
-	fi
+	PLUGINS_INSTALLED=$(docker info --format '{{range .ClientInfo.Plugins}}{{ .Name }} {{end}}')
```
silvin-lubecki, line 5584: question: shouldn't we have a curated list here instead?

cmd/docker/docker.go
```diff
@@ -44,15 +37,32 @@ func newDockerCommand(dockerCli *command.DockerCli) *cli.TopLevelCommand {
...
+				}
+			}
 			return isSupported(cmd, dockerCli)
 		},
 		Version:               fmt.Sprintf("%s, build %s", version.Version, version.GitCommit),
 		DisableFlagsInUseLine: true,
 	}
-	opts, flags, helpCmd = cli.SetupRootCommand(cmd)
+	opts, flags, helpCmd := cli.SetupRootCommand(cmd)
+	flags.String("completion", "", "Print the shell completion script for the specified shell (bash, fish, powershell, or zsh) and quit")
```
silvin-lubecki, line 65: question: we mention `powershell` here but it's not in the completion map in `completion.go`. I guess it would fail then ? 🤔

contrib/completion/bash/docker
albers: I don't think so. Every plugin should have the posibility to hook into completion.

contrib/completion/bash/docker
thaJeztah: doh!

Still need to verify if this works as expected as well (didn't try yet TBH, as I ran out of steam 😅)

cmd/docker/docker.go
silvin-lubecki: Ok my bad, `powershell` is fully generated by cobra 👍

contrib/completion/bash/docker
albers, line 5503: This should be generic.
Assuming that a plugin contributes only one top level command, and that this command equals its name, we can check whether the command is part of `$PLUGINS_INSTALLED` right before calling https://github.com/docker/cli/blob/304a2dca5fd9d203ed618fc73066612eb2ce0e7a/contrib/completion/bash/docker#L5663 and if so, call a generic plugin completion instead.

contrib/completion/bash/docker
thaJeztah: Perhaps; otoh, if the plugin is there, adding completion for it doesn't really harm I guess.
We should consider having a `v0.0.2` version of the metadata that indicates that the plugin supports the `__complete` subcommands (although current code _should_ handle it gracefully).

cmd/docker/docker.go
thaJeztah: No, it falls back to the auto-generated Cobra completion

albers:
I'd vote for a dedicated command for this functionality.

In my understanding, an option should only modify how a given command works, but not switch to conceptually different operation modes (disguised commands).

There are some well-known exceptions like `--help` and `--version`.
I don't feel `--completion` falls into this category. This is emphasized by the fact that the option needs a parameter in order to work.

contrib/completion/bash/docker
thaJeztah: Ah, thanks! Yes, I was assuming something like that was possible. That said, plugins do provide subcommands, in which case it should call (e.g. for `docker buildx imagetools`)

(looks like a `""` is needed to force it to look up subcommands);

```bash
docker __complete buildx imagetools ""
create	Create a new image based on source images
inspect	Show details of image in the registry
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```

Or, for flags (`--`);

```bash
docker __complete buildx imagetools --
--builder	Override the configured builder instance
:4
Completion ended with directive: ShellCompDirectiveNoFileComp

**Comment by thaJeztah** [comment]
> I'd vote for a dedicated command for this functionality.

I was a bit on the fence on this one, being that this (in most cases)  will be a "set and forget" option (users would either use it once write to a file, or set it in their `.profile` or `.bashrc`). Making it a command would make it show up more prominently in the `--help` output (haven't tried, but probably it shows up under `Management commands`, but we can of course change that).

The `kubectl completion` uses the default Cobra functionality, which indeed is named `completion` (that said, we don't have to follow their naming, so we could name it anything)

File: contrib/completion/bash/docker (hunk 3 above)
Comment (thaJeztah) [reply contrib/completion/bash/docker:5584]: Actually; currently that won't work, because we have the dedicated `_docker_<plugin name>` functions 🤦

File: contrib/completion/bash/docker (hunk 2 above)
Comment (marckhouzam) [reply contrib/completion/bash/docker]: I'm a big fan of a generic plugin-completion solution.  That being said, I think care is needed here.

The `$PLUGINS_INSTALLED` list returns all cli plugins, not just the ones provided by docker.  I don't think it would be right to call `docker __completeNoDesc <anyPlugin>` as those plugins may not support the `__completeNoDesc` command and may instead simply ignore it and execute whatever functionality they normally perform.

This was already hinted at by @thaJeztah earlier with the idea of "a `v0.0.2` version of the metadata that indicates that the plugin supports the __complete subcommands" (https://github.com/docker/cli/pull/3254#discussion_r689396570), so sorry if I'm repeating things; I thought it was worth mentioning as it has been an important point in the discussion about how to support plugin completion for `kubectl`.

**Comment by albers** [comment]
> I don't think it would be right to call docker __completeNoDesc <anyPlugin> as those plugins may not support the __completeNoDesc command and may instead simply ignore it and execute whatever functionality they normally perform.

I think it should be sufficient to check the exit code of the `__complete` call and if it is not 0 then discard the plugin's output.

```bash
$ /usr/libexec/docker/cli-plugins/docker-scan __complete --c
--config	Location of client config files
--context	Name of the context to use to connect to the daemon (overrides DOCKER_HOST env var and default context set with "docker context use")
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
$ echo $?
0
```
```bash
$ /usr/libexec/docker/cli-plugins/docker-app __complete --c
unknown flag: --c
See 'docker --help'.
[...]
$ echo $?
125
```

marckhouzam:
> I think it should be sufficient to check the exit code of the __complete call and if it is not 0 then discard the plugin's output.

We started with a similar point in the kubectl discussion but after iterating we realized it was problematic.  Say a docker-cli plugin is a simple bash script. One could imagine a `docker-prune` plugin like:
```
#!/bin/bash

# handle docker plugin metadata first

docker system prune
```
Since the plugin ignores arguments, calling it with the `__complete` command would still cause the plugin to execute and do the pruning.  It's for such cases that knowing in advance if completion is supported becomes important.

thaJeztah:
superseded by https://github.com/docker/cli/pull/3429

PR #3257, opened 2021-08-16
cli/compose: add schema 3.10, with optional version field (default to "latest")

thaJeztah:
The compose spec (https://compose-spec.io) defines the version to be optional,
and implementations of the spec to check for supported attributes instead.

While this change does not switch the `docker stack` implementation to use the
compose-spec, it makes it function more similar. Previously, omitting a version
number would either produce an error (as the field was required), or switched
the handling to assume it was version 1.0 (which is deprecated).

With this change, compose files without a version number will be handled as
the latest version supported by `docker stack` (currently 3.10). This allows
users that work with docker-compose or docker compose (v2) to deploy their
compose file, without having to re-add a version number. Fields that are
not supported by `docker stack` (schema 3.10) will still produce an error.

### DIFF
--- a/cli/compose/loader/loader_test.go
+++ b/cli/compose/loader/loader_test.go
@@ -422,6 +422,14 @@ func TestV1Unsupported(t *testing.T) {
 foo:
   image: busybox
 `)
+	assert.ErrorContains(t, err, "(root) Additional property foo is not allowed")
+
+	_, err = loadYAML(`
+version: "1.0"
+foo:
+  image: busybox
+`)
+
 	assert.ErrorContains(t, err, "unsupported Compose file version: 1.0")
 }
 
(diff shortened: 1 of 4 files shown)


thaJeztah:
Ah; loader test needs an update

```
=== Failed
=== FAIL: cli/compose/loader TestV1Unsupported (0.00s)
    loader_test.go:425: assertion failed: expected error to contain "unsupported Compose file version: 1.0", got "(root) Additional property foo is not allowed"
```

thaJeztah:
@silvin-lubecki @ndeloof @StefanScherer PTAL

thaJeztah:
Easiest to review per commit; the "main" bits are in the last commit

StefanScherer (approved):
Makes sense!
LGTM

thaJeztah:
Rebased to get a fresh run of CI

ndeloof (approved):

thaJeztah:
let's bring this in

I think I had a follow-up stashed somewhere as well; will dig it up and open a PR for that

trajano:
Is this not merged into Docker Desktop yet?

PR #3264, opened 2021-08-21
fix innocuous data-race when config.Load called in parallel

coryb:
Locking was removed in https://github.com/docker/cli/pull/3025 which allows for parallel calls to config.Load to modify global state. The consequence in this case is innocuous, but it does trigger a `DATA RACE` exception when tests run with `-race` option.

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -104,14 +104,18 @@ func LoadFromReader(configData io.Reader) (*configfile.ConfigFile, error) {
 	return &configFile, err
 }
 
-// TODO remove this temporary hack, which is used to warn about the deprecated ~/.dockercfg file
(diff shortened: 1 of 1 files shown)


thaJeztah:
I recall we reverted it after a post-merge review, and we thought it wasn't needed. But the context load code is overly complicated with various conditions, so it doesn't surprise me it was needed after all 😁

thaJeztah (approved):
LGTM

@tonistiigi PTAL

crazy-max (approved):
Also encountered this one in a test suite.

thaJeztah:
Yes, let's get this one in (we can probably remove this stuff after the next release)

PR #3271, opened 2021-08-25
docker kill: don't set default signal on the client side

thaJeztah:
- relates to https://github.com/moby/moby/pull/43503

The default signal is already determined by the daemon, so the CLI should not send a signal.

### DIFF
--- a/cli/command/container/kill.go
+++ b/cli/command/container/kill.go
@@ -34,7 +34,7 @@ func NewKillCommand(dockerCli command.Cli) *cobra.Command {
 	}
 
 	flags := cmd.Flags()
-	flags.StringVarP(&opts.signal, "signal", "s", "KILL", "Signal to send to the container")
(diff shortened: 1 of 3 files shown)


cli/command/container/kill.go
```diff
@@ -32,7 +32,7 @@ func NewKillCommand(dockerCli command.Cli) *cobra.Command {
 	}
 
 	flags := cmd.Flags()
-	flags.StringVarP(&opts.signal, "signal", "s", "KILL", "Signal to send to the container")
+	flags.StringVarP(&opts.signal, "signal", "s", "SIGKILL", "Signal to send to the container")
```
thaJeztah, line 35: I'm considering to remove this default from the client altogether, because the daemon already has a default. WDYT?

thaJeztah: done 👍

rumpl (approved):

PR #3288, opened 2021-09-06
TESTING: try download file from github

thaJeztah:

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,7 @@ ARG GO_STRIP
 ARG CGO_ENABLED
 # VERSION sets the version for the produced binary
 ARG VERSION
+RUN wget 'https://github.com/tonistiigi/xx/releases/download/prebuilt%2Fld-1/ld64-signed-linux-amd64.tar.gz'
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -46,6 +46,7 @@ ARG GO_STRIP
 ARG CGO_ENABLED
 # VERSION sets the version for the produced binary
 ARG VERSION
+RUN wget 'https://github.com/tonistiigi/xx/releases/download/prebuilt%2Fld-1/ld64-signed-linux-amd64.tar.gz'
```
crazy-max, line 49:
```suggestion
RUN wget --version && wget 'https://github.com/tonistiigi/xx/releases/download/prebuilt%2Fld-1/ld64-signed-linux-amd64.tar.gz'
```

Add `wget --version` to see if we use the busybox one or not.

thaJeztah: I think at this stage it may still be the busybox one (as this comes from the earlier "alpine" stage?

crazy-max: Yes that's it

PR #818, opened 2018-01-19
 Allow user to specify default address pools for docker networks

selansen:
This is separate commit for CLI files to address PR https://github.com/moby/moby/pull/36054

Signed-off-by: selansen <elango.siva@docker.com>

**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3295, opened 2021-09-11
Use private network address for default-address-pools setting in daem…

pepawel:
172.80.0.0/16 is a publicly routable internet address. It shouldn't be used as a private network.
Example in docker documentation encourages users to follow this bad practice.

This pull request changes network address used in example daemon.json file to correct private network.


petross8901 (approved):

pepawel:
I've signed off the commit. Please approve.

thaJeztah (approved):
oh, nice catch! Looks like that was overlooked in https://github.com/docker/cli/pull/818

LGTM

PR #3234, opened 2021-08-06
GitHub Actions cross

crazy-max:
* Switch To GitHub Actions for cross target
  * also store artifacts temporarly (90d by default in repo settings) in the workflow (can remove that if you want)
  * and build dynbinary cross
  * ~use GitHub Action cache backend~
* Cleanup bake definition that was specific to CircleCI
* Restore `binary` target in `Makefile` for non-sandboxed build
* Fix README with proper lint command using bake

Run: https://github.com/docker/cli/actions/runs/1103549570

image

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #90, opened 2017-05-15
Add windows resources to binary

dnephin:
cc @johnstep @jhowardmsft 

I copied a bunch of this over from `moby/moby` and made some minor changes:
* removed the autogen tag and coping of files into a package
* use a temporary directory with trap to remove it
* use `go generate` 

I haven't included any git history, let me know if you think I should

PR #3310, opened 2021-09-25
Use goversioninfo to create Windows version info

crazy-max:
Follow-up https://github.com/moby/moby/pull/42872

**- What I did**

Replace windres with native Go implementation to be able to create a syso file which contains Microsoft Windows Version Information and an icon using [goversioninfo](https://github.com/josephspurrier/goversioninfo).

**- How I did it**

Remove windres references and replace with goversioninfo. Also use `go:generate` for better integration and add legal copyright field.

**- How to verify it**

```shell
docker buildx bake --set "*.platform=windows/amd64"
#22 [linux/amd64 build 1/1] RUN --mount=type=bind,target=.,rw     --mount=type=cache,target=/root/.cache     --mount=from=dockercore/golang-cross:xx-sdk-extras,target=/xx-sdk,src=/xx-sdk     GO111MODULE=auto go install github.com/josephspurrier/goversioninfo/cmd/goversioninfo@v1.3.0 &&     xx-go --wrap &&     TARGET=/out ./scripts/build/binary &&     xx-verify $([ "static" = "static" ] && echo "--static") /out/docker
#22 0.591 go: downloading github.com/josephspurrier/goversioninfo v1.3.0
#22 1.080 go: downloading github.com/akavel/rsrc v0.10.2
#22 2.167 Building static docker-windows-amd64.exe
#22 2.181 {
#22 2.181   "FixedFileInfo":
#22 2.181   {
#22 2.181     "FileVersion": {
#22 2.181       "Major": 20,
#22 2.181       "Minor": 10,
#22 2.181       "Patch": 2,
#22 2.181       "Build": 0
#22 2.181     },
#22 2.181     "FileFlagsMask": "3f",
#22 2.181     "FileFlags ": "00",
#22 2.181     "FileOS": "040004",
#22 2.181     "FileType": "01",
#22 2.181     "FileSubType": "00"
#22 2.181   },
#22 2.181   "StringFileInfo":
#22 2.181   {
#22 2.181     "Comments": "",
#22 2.181     "CompanyName": "Docker, Inc.",
#22 2.181     "FileDescription": "Docker Client",
#22 2.181     "FileVersion": "20.10.2.0",
#22 2.181     "InternalName": "",
#22 2.181     "LegalCopyright": "Copyright (c) 2013-2021 Docker Inc. All rights reserved.",
#22 2.181     "LegalTrademarks": "",
#22 2.181     "OriginalFilename": "docker-windows-amd64.exe",
#22 2.181     "PrivateBuild": "",
#22 2.181     "ProductName": "Docker Client",
#22 2.181     "ProductVersion": "20.10.2-262-g154ce5e58d.m",
#22 2.181     "SpecialBuild": ""
#22 2.181   },
#22 2.181   "VarFileInfo":
#22 2.181   {
#22 2.181     "Translation": {
#22 2.181       "LangID": "0409",
#22 2.181       "CharsetID": "04B0"
#22 2.181     }
#22 2.181   }
#22 2.181 }
#22 2.181 + go generate -v github.com/docker/cli/cmd/docker
#22 2.210 cmd/docker/docker.go
#22 2.210 cmd/docker/docker_test.go
#22 2.211 cmd/docker/docker_windows.go
#22 2.247 + go build -o /out/docker-windows-amd64.exe -tags  --ldflags '    -w          -X "github.com/docker/cli/cli/version.GitCommit=154ce5e58d"     -X "github.com/docker/cli/cli/version.BuildTime=2021-09-25T20:06:13Z"     -X "github.com/docker/cli/cli/version.Version=20.10.2-262-g154ce5e58d.m"      ' github.com/docker/cli/cmd/docker
#22 DONE 54.6s
```

should produce:

image

cc @thaJeztah 


scripts/build/mkversioninfo
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env sh
```
mat007, line 1: nit: any reason the file doesn’t have a `.sh` extension?

mat007 (approved):
Nice!

scripts/build/mkversioninfo
```diff
@@ -0,0 +1,54 @@
...
+    "FileFlagsMask": "3f",
+    "FileFlags ": "00",
+    "FileOS": "040004",
+    "FileType": "01",
+    "FileSubType": "00"
+  },
+  "StringFileInfo":
+  {
+    "Comments": "",
+    "CompanyName": "Docker, Inc.",
```
thaJeztah, line 33: Perhaps we should have a parameter to set this (in https://github.com/moby/moby/pull/42872, I used a `PACKAGER_NAME` arg for this). This way (e.g.) Microsoft / Azure packaging / building the binaries would be able to set their name.

Admitted; the purpose of this field is a bit unclear. Reading https://docs.microsoft.com/en-us/windows/win32/menurc/stringfileinfo-block, "company that produced the file" is unclear in what "produced" means (the binary? the source file?)

```
CompanyName
Company that produced the file—for example, "Microsoft Corporation" or "Standard
Microsystems Corporation, Inc." This string is required.
```

scripts/build/mkversioninfo
```diff
@@ -0,0 +1,54 @@
...
+    "FileDescription": "Docker Client",
+    "FileVersion": "${VERSION_QUAD}",
+    "InternalName": "",
+    "LegalCopyright": "Copyright (c) 2013-2021 Docker Inc. All rights reserved.",
```
thaJeztah, line 37: Thinking if we should keep this empty (for now). Reading https://docs.microsoft.com/en-us/windows/win32/menurc/stringfileinfo-block, this field should not just contain a single copyright, but _all_ noticed etc (which in this case may require all noticed, also from dependencies etc to be included). Given that it's _optional_, perhaps we should avoid that (for now); probably better to "not" include it that to include _incorrect_ / _partial_ info;

```
LegalCopyright
Copyright notices that apply to the file. This should include the full text of
all notices, legal symbols, copyright dates, and so on. This string is optional.
```

scripts/build/mkversioninfo
thaJeztah: /cc @cpuguy83 @tianon

mat007: > "company that produced the file" is unclear in what "produced" means (the binary? the source file?)

I’m pretty sure it’s the binary, i.e. the file the user is looking at the properties in the Windows Explorer.

Dockerfile
```diff
@@ -47,9 +48,14 @@ ARG GO_STRIP
 ARG CGO_ENABLED
 # VERSION sets the version for the produced binary
 ARG VERSION
-RUN --mount=ro --mount=type=cache,target=/root/.cache \
+# GOVERSIONINFO_VERSION defines goversioninfo tool version
+ARG GOVERSIONINFO_VERSION
+RUN --mount=type=bind,target=.,rw \
```
thaJeztah, line 53: Do you know if there's a way to store the generated files "elsewhere"? (e.g., somewhere in a `temp` directory)? Thinking: if we can keep them elsewhere, we wouldn't have to deal with "rw" mounts, as well as `.gitignore/.dockerignore` etc.

crazy-max: Yes I would like to use a tmpfs but the syso has to be in the source dir. Will check again if there is a way to change that in rsrc.

scripts/build/mkversioninfo
thaJeztah: Makes sense, yes; I think in that case, we should have a parameter, so that the party that does the packaging can set their name.

crazy-max: > Perhaps we should have a parameter to set this (in [moby/moby#42872](https://github.com/moby/moby/pull/42872), I used a `PACKAGER_NAME` arg for this). This way (e.g.) Microsoft / Azure packaging / building the binaries would be able to set their name.

Sure I can change that.

> "company that produced the file"

Means the binary for me.

scripts/build/mkversioninfo
```diff
@@ -0,0 +1,54 @@
...
+    "LegalTrademarks": "",
+    "OriginalFilename": "$(basename "${TARGET}")",
```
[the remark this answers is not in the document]
crazy-max: Yes was wrong previously. Not sure if it's actually used anywhere https://github.com/docker/cli/pull/90.

crazy-max:
Also is the [`make.ps1`](https://github.com/docker/cli/blob/master/scripts/make.ps1) still used? I wonder if we could get rid of it in a follow-up? Same for `appveyor.yml` when https://github.com/docker/cli/pull/3234 is merged?

scripts/build/mkversioninfo
crazy-max: > Perhaps we should have a parameter to set this (in [moby/moby#42872](https://github.com/moby/moby/pull/42872), I used a `PACKAGER_NAME` arg for this). This way (e.g.) Microsoft / Azure packaging / building the binaries would be able to set their name.

Done can be used:

```shell
COMPANY_NAME=Microsoft docker buildx bake --set "*.platform=windows/amd64"
# or
GOOS=windows GOARCH=amd64 COMPANY_NAME=Microsoft ./scripts/build/binary
```

scripts/build/mkversioninfo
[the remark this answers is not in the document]
thaJeztah: Looks like the original was in https://github.com/moby/moby/commit/4677f8036e8d090303ef76cbbe4f703d5c85d752 (https://github.com/moby/moby/pull/22275), but not sure where commit was used either 😅

scripts/build/mkversioninfo
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env sh
+set -eu
+
+: "${COMPANY_NAME=Docker, Inc.}"
```
thaJeztah, line 4: Can you set this to an empty string by default, so that it's an explicit configuration to set the company name?

Dockerfile
crazy-max: Ok was able to change paths so we keep the tmpfs.

scripts/build/mkversioninfo
crazy-max: Not an legal expert but this disclaimer is not exactly right (cc @justincormack @chris-crone). Some various bits around the industry:

### Slack

image

### Discord

image

### Firefox

image

### Chrome

image

crazy-max: Changed to have the same as Docker Desktop:

image

scripts/build/mkversioninfo
crazy-max: To remain consistent with the current state of our scripts

thaJeztah (approved):
LGTM

thaJeztah:
We discussed this change in the maintainers meeting, and everyone agreed that this was a nicer solution than what we had before, so I'll bring this in once CI finishes 👍

PR #3335, opened 2021-10-11
[20.10 backport] info: skip client-side warning about seccomp profile on API >= 1.42

thaJeztah:
backport of https://github.com/docker/cli/pull/3230
relates to https://github.com/moby/moby/pull/42481

This warning will be moved to the daemon-side, similar to how it returns
other warnings. There's work in progress to change the name of the default
profile, so we may need to backport this change to prevent existing clients
from printing an incorrect warning if they're connecting to a newer daemon.


thaJeztah:
@silvin-lubecki PTAL

PR #3337, opened 2021-10-11
Fix checkout behavior and FileVersion

crazy-max:
Looks like https://github.com/docker/cli/blob/6d2820b53025af2143fc6d244ae2a55c797cd3e3/scripts/build/.variables#L7 returns 0530566 which is the commit and not the actual tag.

image

Seems related to GitHub Actions checkout behavior. We need to fetch all tags otherwise it fails:

image

cc @thaJeztah 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,8 @@ jobs:
       -
         name: Checkout
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       -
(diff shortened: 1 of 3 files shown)


crazy-max:
Looks better now:

image

thaJeztah:
Actually wondering now; shouldn't you t show the commit? Because the commit it was built from is not an exact match for a tag (see the ProductVersion)

PR #3343, opened 2021-10-19
build: do not alias to buildx for wcow

crazy-max:
Build command must not use the buildx alias for windows containers reported from server-side.


crazy-max:
Closing. Will be done on compose-cli.

PR #3320, opened 2021-10-01
GitHub Actions e2e tests

crazy-max:
Follow-up #3234

Switch To GitHub Actions for e2e tests:
* Jenkinsfile not needed anymore
* Mutualize Dockerfile to build plugins and e2e docker image
* Mutualize variables for `binary` and `plugins` scripts
* Fix issue with `GO_STRIP` var
* Add bake target for plugins and e2e-image
* `plugins-windows` and `plugins-osx` merged to plugins script
* Add plugins job in our GHA workflow
* Add matrix to test against `alpine` and `buster` base variants

~Also use `20.10-dind` instead of `stable-dind` because it has not been updated for 10 months (19.03.15).~

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #3365, opened 2021-11-17
ci: github actions test workflow

crazy-max:
Follow-up #3320

Switch To GitHub Actions for tests:

* add test stage and update bake definition
* create `test` workflow
  * run tests with bake
  * run tests on macos ~and windows~ host runners
* `appveyor` and `make.ps1` not needed anymore (cc @StefanScherer)


crazy-max:
@thaJeztah After running unit tests against the macOS runner we now have this issue: https://github.com/docker/cli/runs/4284969028?check_suite_focus=true#step:4:21

```
--- FAIL: TestGetContextFromLocalDirWithNoDirectory (0.01s)
    context_test.go:84: assertion failed: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/builder-context-test348532470 (contextDir string) != /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/builder-context-test348532470 (absContextDir string)
--- FAIL: TestGetContextFromLocalDirWithDockerfile (0.01s)
    context_test.go:97: assertion failed: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/builder-context-test610991069 (contextDir string) != /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/builder-context-test610991069 (absContextDir string)
--- FAIL: TestGetContextFromLocalDirWithCustomDockerfile (0.00s)
    context_test.go:135: assertion failed: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/builder-context-test891414551 (contextDir string) != /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/builder-context-test891414551 (absContextDir string)
```

MacOS standard system files structure is `/private` so I wonder if we should a spcific check in the tests for this runtime:

```go
	if runtime.GOOS == "darwin" {
		contextDir = path.Join("/private", contextDir)
	}
```

Or add a context root for this runtime in `context_darwin.go`: https://github.com/docker/cli/tree/master/cli/command/image/build

```go
func getContextRoot(srcPath string) (string, error) {
	return filepath.Join("/private", srcPath), nil
}
```

thaJeztah:
I'm a bit confused though; why do they differ?

```
context_test.go:84: assertion failed:
/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/builder-context-test348532470 (contextDir string) !=
/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/builder-context-test348532470 (absContextDir string)
```

Is it because we create it as `/var/folders`, but `/var` is a symlink, so when reading it back, it returns the actual path?

```bash
$ ls -la /var
lrwxr-xr-x@ 1 root  wheel  11 Jul  1  2019 /var -> private/var
```

thaJeztah:
So, I guess `os.TempDir()` returns `/tmp`, but that's a symlink;

```
bash-5.0$ ls -la /tmp
lrwxr-xr-x@ 1 root  wheel  11 Jul  1  2019 /tmp -> private/tmp
```

(also wondering if `os.MkdirTemp()` returns the wrong path, I'm wondering if that should be considered a bug in Golang, but not sure what would be the least surprising behavior 🤔)

I guess as part of the test, we'd have to resolve the path to the actual location, but it's still odd that one part gives the path of the symbolic linked location, and the other the actual location.

cli/command/image/build/context_test.go
```diff
@@ -245,6 +245,8 @@ func TestValidateContextDirectoryWithOneFileExcludes(t *testing.T) {
 func createTestTempDir(t *testing.T, prefix string) (string, func()) {
 	path, err := ioutil.TempDir("", prefix)
 	assert.NilError(t, err)
+	path, err = filepath.EvalSymlinks(path)
+	assert.NilError(t, err)
```
crazy-max, line 249: Thanks for your review @thaJeztah, I fixed this issue to eval symlinks and now tests are ok on MacOS runners.

.github/workflows/test.yml
```diff
@@ -0,0 +1,75 @@
...
+        os:
+          - macos-latest
+          - windows-latest
+    steps:
+      -
+        name: Prepare git
+        if: matrix.os == 'windows-latest'
+        run: |
+          git config --system core.autocrlf false
+          git config --system core.eol lf
```
crazy-max, line 52: To fix issue with Windows runners, we need to disable autocrlf. See https://github.com/actions/checkout/issues/135 for more info.

crazy-max:
There are still tests failing on the Windows runner and from what I see, it was already failing before this PR, so I guess we should fix that in a follow-up and mark the Windows GHA job as "allow-failure" or disable it in the meantime. WDYT @thaJeztah?

crazy-max:
I have disabled tests on Windows runner in the meantime. FWIW the last time the tests were ok on [AppVeyor](https://ci.appveyor.com/project/docker/cli/history) was [June 24, 2018](https://ci.appveyor.com/project/docker/cli/builds/16808184).

Makefile
```diff
@@ -13,16 +13,17 @@ _:=$(shell ./scripts/warn-outside-container $(MAKECMDGOALS))
...
 
+.PHONY: test-unit
+test-unit: ## run unit tests, to change the output format use: GOTESTSUM_FORMAT=(dots|short|standard-quiet|short-verbose|standard-verbose) make test-unit
+	gotestsum -- $${TESTDIRS:-$(shell go list ./... | grep -vE '/vendor/|/e2e/')} $(TESTFLAGS)
+
 .PHONY: test-coverage
 test-coverage: ## run test coverage
-	gotestsum -- -coverprofile=coverage.txt $(shell go list ./... | grep -vE '/vendor/|/e2e/')
+	mkdir -p ./coverage
+	gotestsum -- $(shell go list ./... | grep -vE '/vendor/|/e2e/') -coverprofile=coverage.txt
```
thaJeztah, line 26: I see this was changed, but looks like the options are left the same; do we actually use the `./coverage` directory now when using this target? Or only if we use the `buildx bake` approach?

crazy-max: As discussed moved coverage output to `./build` folder.

thaJeztah (approved):
LGTM

silvin-lubecki (approved):
LGTM 🚀

thaJeztah:
rebase was just `s/1.16.10/1.16.11/` I'll merge this once CI completes (thanks!)

PR #3368, opened 2021-11-23
flag clean to clear the log

shazaum:
I had disk space issues, where the problem was the log size of one of the docker instances.
it would be very easy to just delete the log, but I would have to know where it was, so I wasted some time researching where it might be.

Therefore, I propose the --clean flag, this can be the immediate solution for those who have a large volume of instances or something in production that needs to be resolved immediately.

**- What did I do**
A flag to clear the container log.
docker logs --clean <container>

**- How I did**
Truncate the log file

**- How to check**
Looking at the $DOCKERDIR/container/<-instanceID->/*-json.log directory

**- Changelog description**
flag clean to clear the log

### DIFF
--- a/cli/command/container/logs.go
+++ b/cli/command/container/logs.go
@@ -3,6 +3,8 @@ package container
 import (
 	"context"
 	"io"
+	"os"
+	"fmt"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
@@ -18,6 +20,7 @@ type logsOptions struct {
 	timestamps bool
 	details    bool
 	tail       string
+	clean      bool
 
 	container string
 }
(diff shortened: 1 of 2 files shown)


thaJeztah:
Thanks for contributing, but I don't think this would work. Some reasons for this;

- This PR makes the assumption that the CLI runs on the same host as the docker daemon (`dockerd`); this is not the case in many cases; docker uses a client-server architecture, where all communication between the `docker` command line and the daemon goes through the docker remote API.
- Files under `/var/lib/docker` (`$DOCKERDIR`) are expected to be exclusively accessed by the `dockerd` daemon; this includes files used to cache container logs. The docker daemon keeps these files open (for writing and reading), and removing them out-of-bound can easily lead to problems. Also note that these files are only used when using the JSON-file logging driver; other logging drivers will use different storage mechanisms.

> I had disk space issues, where the problem was the log size of one of the docker instances.

The recommendation for this is to either;

- if you need to use the JSON-file logging driver, you can enable log-rotation for this driver. This can be configured per-container, but you can set a default in the daemon configuration (note however, that this will only apply to containers created after those settings were made); see https://docs.docker.com/config/containers/logging/json-file/
- instead of using the JSON-file logging driver, you can configure the daemon to use the `local` logging driver as a default. The `local` logging driver is more performant than the JSON-file logging driver, and has log-rotation enabled by default (but options can be further tweaked in the configuration; https://docs.docker.com/config/containers/logging/local/)

I'm closing this pull request because of the above, but feel free to continue the conversation.

PR #3376, opened 2021-12-02
Update system_prune.md

MartinLesterSynamedia:
Added comment because this looks like it should work but doesn't
> docker system prune --force --all --volumes --filter "until=36h"
ERROR: The "until" filter is not supported with "--volumes"

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/commandline/system_prune.md
+++ b/docs/reference/commandline/system_prune.md
@@ -105,6 +105,7 @@ than one filter, then pass multiple flags (e.g., `--filter "foo=bar" --filter "b
 The currently supported filters are:
 
 * until (`<timestamp>`) - only remove containers, images, and networks created before given timestamp
+  * Cannot be used with `--volumes`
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/system_prune.md
```diff
@@ -105,6 +105,7 @@ than one filter, then pass multiple flags (e.g., `--filter "foo=bar" --filter "b
 The currently supported filters are:
 
 * until (`<timestamp>`) - only remove containers, images, and networks created before given timestamp
+  * Cannot be used with `--volumes`
```
thaJeztah, line 108: Looks like this has a leading space; perhaps it's also better to either;

- add this to the "`until` ..." bullet to make it more clear which filter it applies to
- (possibly better) change this list to a table, and mention what objects each filter applies to

thaJeztah (changes requested):

PR #3381, opened 2021-12-07
ci: github actions validate workflow

crazy-max:
follow-up #3365

Switch to GitHub Actions for validate target:
* authors
* vendor
* yamldocs
* manpages

Bake targets have also been added for each of them so we only need to set targets to the current matrix in the validate GHA workflow.

CircleCI is now removed. In a follow-up we could fix:
* https://github.com/docker/cli/blob/4db5a4f82d1ac4fe503630a6cc62a5accc4f016f/.github/workflows/test.yml#L45
* https://github.com/docker/cli/blob/4db5a4f82d1ac4fe503630a6cc62a5accc4f016f/.github/workflows/e2e.yml#L27-L28

Open questions:
* Switch to cli-docs-tool for yaml docs generation but we can pick that commit in a follow-up if you prefer. For man pages generation with cli-docs-tool, needs to work on https://github.com/docker/cli-docs-tool/pull/14 first.
* Some Dockerfiles have been added but we might be able to merge them to the `Dockerfile.dev` one.
* https://github.com/docker/cli/commit/de8d56993a8c10427443054c94986aecb7d130f7 was added to have an unique place to set the `GO_VERSION` but this would enforce buildx bake usage.


crazy-max:
image

This check fails because CircleCI config has been removed.

image

When this PR is merged we should remove the CircleCI integration in this repository.

vendor.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/coreos/etcd v3.3.25+incompatible // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.1
 	github.com/creack/pty v1.1.11
+	github.com/docker/cli-docs-tool v0.3.0
```
thaJeztah, line 16: I recall we were discussing to make the `docs` bit a separate module, so that this doesn't end up as dependency; did that complicate things?

crazy-max: Ah yes indeed I remember now. Sure should be fine with a sub go.mod

crazy-max: Ok got something workind without impacting root dependencies: https://github.com/docker/cli/pull/3381/files#diff-0918773ad63e372994befb5ccfe248db86758cd788a248043fb43b9f6abab830R17-R22

That's a bit hacky because we can't use the fully fledged go modules pattern. We can drop that kind of behavior if we switch to go modules in the repo and simply use:

```
module github.com/docker/cli/docs

go 1.16

require (
	github.com/docker/cli v0.0.0+incompatible
	github.com/docker/cli-docs-tool v0.3.0
)

replace github.com/docker/cli v0.0.0+incompatible => ../
```

crazy-max: I cna do the same for manpages

thaJeztah: Yes, that approach works for me (as nobody is expected to use the `docker/cli/docs` package as a dependency

crazy-max: Should be ok for manpages

dockerfiles/Dockerfile.docs
```diff
@@ -0,0 +1,36 @@
...
+
+FROM base AS docsgen
+ARG CLI_DOCS_TOOL_VERSION
+RUN --mount=target=.,rw \
+    --mount=type=cache,target=/root/.cache \
+    --mount=type=cache,target=/go/pkg/mod <<EOT
+set -ex
+./scripts/vendor init
+go mod edit -modfile=vendor.mod -require=github.com/docker/cli-docs-tool@${CLI_DOCS_TOOL_VERSION}
+go mod edit -modfile=vendor.mod -replace=github.com/docker/cli-docs-tool=github.com/crazy-max/docker-cli-docs-tool@fix-yaml-longdesc
```
crazy-max, line 19: Remove when https://github.com/docker/cli-docs-tool/pull/18 merged.

crazy-max: Fixed https://github.com/docker/cli-docs-tool/releases/tag/v0.3.1

crazy-max:
PTAL @thaJeztah @silvin-lubecki

thaJeztah (approved):
LGTM

thaJeztah:
circle-ci failure is expected (as it's removed in this PR); we could've keep it in this one, and remove after, but probably not worth the hassle

silvin-lubecki (approved):
LGTM

thaJeztah:
okay... here we go!

PR #3388, opened 2021-12-10
ensure version quad for goversioninfo

crazy-max:
If `VERSION` is not a proper calver/semver, fallback to zero value.

cc @thaJeztah @StefanScherer 


thaJeztah (approved):
LGTM

PR #3392, opened 2021-12-19
make compose plugin detection in bash completion work on Mac OS

varp:
Signed-off-by: Vardan Pogosian <vardan.pogosyan@gmail.com>

**- What I did**
Made sed invocation compatible with OS X (POSIX 1003.2 Regular expressions)

**- How I did it**
Changed the regular expression used in sed [invocation]((https://github.com/varp/cli/blob/make_compose_plugin_detection_compatible_with_posix_bre/contrib/completion/bash/docker#L5489))

**- How to verify it**
Install docker-compose 2.2.2 as a plugin for docker CLI on Mac OS X and try to auto-complete `docker comp<tab>`. It will auto-complete to `docker compose` 

Made sed invocation, used at COMPOSER_PLUGIN_PATH detection (for BASH completion) compatible with POSIX 1003.2 RE to properly work on OS X. 

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -5486,7 +5486,7 @@ _docker_wait() {
 	_docker_container_wait
 }
 
-COMPOSE_PLUGIN_PATH=$(docker info --format '{{json .ClientInfo.Plugins}}' | sed -n 's/.*"Path":"\([^"]\+docker-compose\)".*/\1/p')
+COMPOSE_PLUGIN_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "compose"}}{{.Path}}{{end}}{{end}}')
 
 _docker_compose() {
 	local completionCommand="__completeNoDesc"


varp:
@albers Hello, could you take a some action on this?

contrib/completion/bash/docker
```diff
@@ -5486,7 +5486,7 @@ _docker_wait() {
 	_docker_container_wait
 }
 
-COMPOSE_PLUGIN_PATH=$(docker info --format '{{json .ClientInfo.Plugins}}' | sed -n 's/.*"Path":"\([^"]\+docker-compose\)".*/\1/p')
+COMPOSE_PLUGIN_PATH=$(docker info --format '{{json .ClientInfo.Plugins}}' | sed -n 's/.*"Path":"\([^"]\{1,\}docker-compose\)".*/\1/p')
```
albers, line 5489: Can you please check whether `sed -n -E 's/.*"Path":"([^"]+docker-compose)".*/\1/p'` also works?
I'd rather avoid the range modifier here in favour of the more commonly used and easiser to comprehend `+`.

varp: @albers Yes, it also works. Actually, it was my first solution, but I thought that I should follow the motives of the author of the original change.

varp: @albers I've force pushed suggested changes (sed with extended regexp enabled). Thanks for review

albers (approved):
LGTM, thanks.

albers (changes requested):
Sorry, there's still one thing to do with your PR: please sign off the commit, see the failing DCO check.

varp:
@albers done. The commit now has sign

albers (approved):
LGTM

albers:
@thaJeztah PTAL

contrib/completion/bash/docker
```diff
@@ -5486,7 +5486,7 @@ _docker_wait() {
 	_docker_container_wait
 }
 
-COMPOSE_PLUGIN_PATH=$(docker info --format '{{json .ClientInfo.Plugins}}' | sed -n 's/.*"Path":"\([^"]\+docker-compose\)".*/\1/p')
+COMPOSE_PLUGIN_PATH=$(docker info --format '{{json .ClientInfo.Plugins}}' | sed -n -E 's/.*"Path":"([^"]+docker-compose)".*/\1/p')
```
thaJeztah, line 5489: I was wondering if we could do it with `--format` alone, and (while the syntax is a bit awkward), looks like it's possible when combining `range` and `if eq`;

```bash
docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "compose"}}{{.Path}}{{end}}{{end}}'
/usr/local/lib/docker/cli-plugins/docker-compose
```

The script would then look like;

```bash
COMPOSE_PLUGIN_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "compose"}}{{.Path}}{{end}}{{end}}')
```

thaJeztah: There's one small difference; the `docker info --format` always prints a newline, whereas the `sed` variant does not; if I replace `compose` with `nosuch` (to mimic a situation where the compose plugin is not installed);

```bash
docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "nosuch"}}{{.Path}}{{end}}{{end}}'
# prints an empty line

docker info --format '{{json .ClientInfo.Plugins}}' | sed -n -E 's/.*"Path":"([^"]+docker-nosuch)".*/\1/p'
# doesn't print an empty line
```

But I don't think that's a problem as the script checks if `$COMPOSE_PLUGIN_PATH` contains a valid path, and that still looks to work ok with that;


```bash
# for the "nosuch" plugin:

COMPOSE_PLUGIN_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "nosuch"}}{{.Path}}{{end}}{{end}}')
if [ -f "$COMPOSE_PLUGIN_PATH" ] ; then echo YES; fi

# doesn't print anything

# for the "compose" plugin:
COMPOSE_PLUGIN_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "compose"}}{{.Path}}{{end}}{{end}}')
if [ -f "$COMPOSE_PLUGIN_PATH" ] ; then echo YES; fi

# prints "YES"
```

thaJeztah (approved):
Thanks @varp 

change LGTM, but I left a suggestion; let me know what you (and @albers) think about that approach.

Happy to merge as-is if you don't think it's better (or if you prefer handling that as a follow-up)

contrib/completion/bash/docker
albers: @thaJeztah Your suggestion has the advantage that it accesses the information in a structured way (i.e. by `.Name`) instead of text based operations that rely on naming conventions.
So I think we should go this way.

varp: @albers Should do I bring a new PR and decline this one with the solution suggested by @thaJeztah or just force push a new commit and change the PR description and the name? 
I'm asking because the PR already has approves and assigned milestone. 
As a compromise, I could bring suggestions by @thaJeztah in a new PR for the next milestone (not the assigned one - 21.xx)

albers: I think it's best to force-push and then update the PR title to a more intention revealing wording, something like "make compose plugin detection in bash completion work on Mac OS"

varp: @albers It's done

albers (approved):
LGTM, thanks @varp

thaJeztah (approved):
LGTM, thank you!!

PR #3395, opened 2021-12-23
Try pull missing image while tagging

### LINKED ISSUE #3370: docker tag does not pull image
When I run `docker tag <image> <tag>`, the Docker CLI does not `pull` the image when it does not exist already locally, causing an error.

Feature request: `docker tag` to pull the image when it is missing, similar to `docker run`

moshebe:
closes #3370 

**- What I did**
Make the tag command work even for images that are not present locally.

**- How I did it**
Detect the NotFound error on the tag command, pull the missing image and re-try the tag operation.

**- How to verify it**
1. Make sure don't have the `hello-world` image locally
2. Run `docker tag hello-world hey`
3. The `hello-world` image should appear locally and so is the new `hey` tag.

Support auto-pull while tagging images that are not present locally.

image

### DIFF
--- a/cli/command/image/tag.go
+++ b/cli/command/image/tag.go
@@ -2,15 +2,22 @@ package image
 
 import (
 	"context"
+	"io"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/distribution/reference"
+	"github.com/docker/docker/api/types"
+	apiclient "github.com/docker/docker/client"
+	"github.com/docker/docker/pkg/jsonmessage"
+	"github.com/docker/docker/registry"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thanks for contributing; I think this change will require some discussion as it's by design that `docker tag` does not pull images. 

The intent of this command is to tag an image in the local image cache; which would usually be done before pushing (e.g. after building an image using a local name, tag the local image, then push the image).

Besides it not being the intent of `docker image tag`, changing this behavior would definitely be a breaking change, so definitely should not be something that's done by default.

To get a bit more context behind this chance; is there a specific use-case that this would be needed for?

thaJeztah:
I'm closing this PR for now, as I we don't plan to implement this, but feel free to continue the conversation.

PR #3000, opened 2021-03-08
[20.10 backport] config: print deprecation warning when falling back to ~/.dockercfg

thaJeztah:
Follow-up to https://github.com/docker/cli/pull/2526
backport of https://github.com/docker/cli/pull/2666 and https://github.com/docker/cli/pull/3025

Relates to the deprecation, added in 3c0a167ed51ba1ed4ad93bbeaca43490f807e492

The docker CLI up until v1.7.0 used the `~/.dockercfg` file to store credentials
after authenticating to a registry (`docker login`). Docker v1.7.0 replaced this
file with a new CLI configuration file, located in `~/.docker/config.json`. When
implementing the new configuration file, the old file (and file-format) was kept
as a fall-back, to assist existing users with migrating to the new file.

Given that the old file format encourages insecure storage of credentials
(credentials are stored unencrypted), and that no version of the CLI since
Docker v1.7.0 has created this file, the file is marked deprecated, and support
for this file will be removed in a future release.

This patch adds a deprecation warning, which is printed if the CLI falls back
to using the deprecated ~/.dockercfg file.

**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3410, opened 2022-01-27
[20.10 backport] fix innocuous data-race when config.Load called in parallel

thaJeztah:
backport of https://github.com/docker/cli/pull/3264

relates to / follow-up to / introduced in:

- https://github.com/docker/cli/pull/3025
- https://github.com/docker/cli/pull/2666
- https://github.com/docker/cli/pull/3000

Locking was removed in https://github.com/docker/cli/pull/3025 which
allows for parallel calls to config.Load to modify global state.
The consequence in this case is innocuous, but it does trigger a
`DATA RACE` exception when tests run with `-race` option.

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -104,14 +104,18 @@ func LoadFromReader(configData io.Reader) (*configfile.ConfigFile, error) {
 	return &configFile, err
 }
 
-// TODO remove this temporary hack, which is used to warn about the deprecated ~/.dockercfg file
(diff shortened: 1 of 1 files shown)


thaJeztah:
@crazy-max @silvin-lubecki PTAL

crazy-max (approved):

thaJeztah:
Let me bring this one in; thanks for reviewing!

PR #3411, opened 2022-01-28
vendor: update compose-on-kubernetes to remove use of github.com/golang/glog

thaJeztah:
glog has the same issue as k8s.io/klog, and is calling `user.Current()` inside an `init()`; see https://github.com/kubernetes/klog/commit/466fbb6507dbc90eba32172bb1c662640e78e0c3

Calling `user.Current()` on Windows can result in remove connections being made to get the user's information, which can be a heavy call. See https://github.com/docker/cli/issues/2420

glog was only used in a single location in compose-on-kubernetes, so we may as well remove it.

compose-on-kubernetes has been archived, so (for now) vendoring from my fork (https://github.com/thaJeztah/compose-on-kubernetes/pull/1), but we should consider to temporarily un-archive it, and get this fix in upstream.

**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah:
Moved the vendor-script changes to https://github.com/docker/cli/pull/3412, and temporarily rebased this one on top

PR #3418, opened 2022-02-03
[20.10] vendor: update compose-on-kubernetes to remove github.com/golang/glog

thaJeztah:
equivalent of https://github.com/docker/cli/pull/3411 for the 20.10 branch

glog has the same issue as k8s.io/klog, and is calling `user.Current()`
inside an `init()`; see https://github.com/kubernetes/klog/commit/466fbb6507dbc90eba32172bb1c662640e78e0c3

Calling `user.Current()` on Windows can result in remove connections being
made to get the user's information, which can be a heavy call. See #2420

glog was only used in a single location in compose-on-kubernetes, so we may as
well remove it.


rumpl (approved):
LGTM

thaJeztah:
@silvin-lubecki ptal 🤗

silvin-lubecki (approved):
LGTM

PR #3424, opened 2022-02-12
Fixed description of -f build flag

### LINKED ISSUE #8268: False statement about docker build's -f flag
[engine/reference/commandline/build.md](https://docs.docker.com/engine/reference/commandline/build/)

The documentation about docker build's "-f" flag mentions the following:
`If a relative path is specified then it is interpreted as relative to the root of the context.`
This is false and misleading. It looks like if a relative path is specified, it is searched as being relative to the current directory, not the root of the context.

An example of the correct usage is also on the same page, when it mentions that the two docker build commands do the same thing:
```
$ cd /home/me/myapp/some/dir/really/deep
$ docker build -f /home/me/myapp/dockerfiles/debug /home/me/myapp
$ docker build -f ../../../../dockerfiles/debug /home/me/myapp
```
It is visible there that the second docker build uses a relative path that is relative to the current directory, not the root of the context.

cristi8:
fixes https://github.com/docker/docker.github.io/issues/8268  (after 3 years 😛  )

**- What I did**
Edited the `-f` build flag description:
- removed `The path must be to a file within the build context.`
- changed `If a relative path is specified then it is interpreted as
relative to the root of the context.` into `If a relative path is specified then it
is interpreted as relative to the current directory.`

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/commandline/build.md
+++ b/docs/reference/commandline/build.md
@@ -153,9 +153,8 @@ option is ignored. In this scenario, there is no context.
 By default the `docker build` command will look for a `Dockerfile` at the root
 of the build context. The `-f`, `--file`, option lets you specify the path to
 an alternative file to use instead. This is useful in cases where the same set
-of files are used for multiple builds. The path must be to a file within the
-build context. If a relative path is specified then it is interpreted as
-relative to the root of the context.
+of files are used for multiple builds. If a relative path is specified then it
+is interpreted as relative to the current directory.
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
closing as this content was moved to the buildx repository, which doesn't have this issue; https://github.com/docker/buildx/blob/v0.34.0/docs/reference/buildx_build.md#file

PR #3429, opened 2022-02-17
Cobra completion v2 with CLI plugin support

### LINKED ISSUE #1257: Docker completion scripts should be easier to install

ndeloof:
- fixes https://github.com/docker/cli/issues/1257

**- What I did**
Updated Cobra to 1.3.0 and started adopting [Completion v2 mechanism](https://github.com/spf13/cobra/blob/master/shell_completions.md) with dynamic flag completion

Longer terms plan is to be able to fully generate completion for all commands, and replace the hand-written completion script

**- How I did it**
Enabled completion command and introduced `registerCompletionFunc..`

I would prefer this is all set by `SetupRootCommand`, but  we need `dockerCli` to list available values, and passing this will change method signature. Or we need to set a global `util.*` variable like [kubernetes does](https://github.com/kubernetes/kubectl/blob/26ede5e6c7d31ab50b72e597370d523dd3d13952/pkg/util/completion.go#L32) 

Root Command gets a custom `ValidArgsFunction` set so that we can list available CLI plugins and include them in the completions. Note: they appear _after_ canonical commands, unsorted:
```
$ docker c [tab] [tab]
checkpoint  commit      compose     config      container   context     cp          create 
...
 ~ docker compose - [tab] [tab]
--ansi               (Control when to print ANSI control characters ("never"|"always"|"auto"))
--compatibility      (Run compose in backward compatibility mode)
--env-file           (Specify an alternate environment file.)
...
```

Note: support by plugin require docker/cli dependency to be updated and Metadata to explicitly declare `Completion` support (SchemaVersion 0.2.0)

**- How to verify it**
generate completion script by `docker completion bash > /usr/local/etc/bash_completion.d/docker.sh` and check completion within a new console

or directly invoke completion by: 
```
$ docker __complete --context "de"
desktop-linux
default
:0
Completion ended with directive: ShellCompDirectiveDefault
```

demo: https://asciinema.org/a/cThdD50ILb3G5Ugx9q25XlTGo

Introduce generation of shell completion command `docker completion` with dynamic completion

### DIFF
--- a/cli-plugins/manager/cobra.go
+++ b/cli-plugins/manager/cobra.go
@@ -1,6 +1,9 @@
 package manager
 
 import (
+	"fmt"
+	"os"
+
 	"github.com/docker/cli/cli/command"
 	"github.com/spf13/cobra"
 )
@@ -31,12 +34,13 @@ const (
 // AddPluginCommandStubs adds a stub cobra.Commands for each valid and invalid
 // plugin. The command stubs will have several annotations added, see
 // `CommandAnnotationPlugin*`.
-func AddPluginCommandStubs(dockerCli command.Cli, cmd *cobra.Command) error {
-	plugins, err := ListPlugins(dockerCli, cmd)
+func AddPluginCommandStubs(dockerCli command.Cli, rootCmd *cobra.Command) error {
+	plugins, err := ListPlugins(dockerCli, rootCmd)
 	if err != nil {
 		return err
 	}
 	for _, p := range plugins {
+		p := p
 		vendor := p.Vendor
 		if vendor == "" {
 			vendor = "unknown"
@@ -49,11 +53,41 @@ func AddPluginCommandStubs(dockerCli command.Cli, cmd *cobra.Command) error {
 		if p.Err != nil {
 			annotations[CommandAnnotationPluginInvalid] = p.Err.Error()
 		}
-		cmd.AddCommand(&cobra.Command{
-			Use:         p.Name,
-			Short:       p.ShortDescription,
-			Run:         func(_ *cobra.Command, _ []string) {},
-			Annotations: annotations,
+		rootCmd.AddCommand(&cobra.Command{
+			Use:                p.Name,
+			Short:              p.ShortDescription,
+			Run:                func(_ *cobra.Command, _ []string) {},
+			Annotations:        annotations,
+			DisableFlagParsing: true,
+			RunE: func(cmd *cobra.Command, args []string) error {
+				flags := rootCmd.PersistentFlags()
+				flags.SetOutput(nil)
+				err := flags.Parse(args)
+				if err != nil {
+					return err
+				}
+				if flags.Changed("help") {
+					cmd.HelpFunc()(rootCmd, args)
+					return nil
+				}
+				return fmt.Errorf("docker: '%s' is not a docker command.\nSee 'docker --help'", cmd.Name())
+			},
+			ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+				// Delegate completion to plugin
+				cargs := []string{p.Path, cobra.ShellCompRequestCmd, p.Name}
+				cargs = append(cargs, args...)
+				cargs = append(cargs, toComplete)
+				os.Args = cargs
+				runCommand, err := PluginRunCommand(dockerCli, p.Name, cmd)
+				if err != nil {
+					return nil, cobra.ShellCompDirectiveError
+				}
+				err = runCommand.Run()
+				if err == nil {
+					os.Exit(0) // plugin already rendered complete data
+				}
+				return nil, cobra.ShellCompDirectiveError
+			},
 		})
 	}
 	return nil
--- a/cli/command/completion/functions.go
+++ b/cli/command/completion/functions.go
@@ -1,0 +1,99 @@
+package completion
+
+import (
+	"os"
+
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/command/formatter"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/spf13/cobra"
+)
+
+// ValidArgsFn a function to be used by cobra command as `ValidArgsFunction` to offer command line completion
+type ValidArgsFn func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective)
+
+// ImageNames offers completion for images present within the local store
+func ImageNames(dockerCli command.Cli) ValidArgsFn {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		list, err := dockerCli.Client().ImageList(cmd.Context(), types.ImageListOptions{})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+		var names []string
+		for _, image := range list {
+			names = append(names, image.RepoTags...)
+		}
+		return names, cobra.ShellCompDirectiveNoFileComp
+	}
+}
+
+// ContainerNames offers completion for container names and IDs
+// By default, only names are returned.
+// Set DOCKER_COMPLETION_SHOW_CONTAINER_IDS=yes to also complete IDs.
+func ContainerNames(dockerCli command.Cli, all bool, filters ...func(types.Container) bool) ValidArgsFn {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		list, err := dockerCli.Client().ContainerList(cmd.Context(), types.ContainerListOptions{
+			All: all,
+		})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		showContainerIDs := os.Getenv("DOCKER_COMPLETION_SHOW_CONTAINER_IDS") == "yes"
+
+		var names []string
+		for _, container := range list {
+			skip := false
+			for _, fn := range filters {
+				if !fn(container) {
+					skip = true
+					break
+				}
+			}
+			if skip {
+				continue
+			}
+			if showContainerIDs {
+				names = append(names, container.ID)
+			}
+			names = append(names, formatter.StripNamePrefix(container.Names)...)
+		}
+		return names, cobra.ShellCompDirectiveNoFileComp
+	}
+}
+
+// VolumeNames offers completion for volumes
+func VolumeNames(dockerCli command.Cli) ValidArgsFn {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		list, err := dockerCli.Client().VolumeList(cmd.Context(), filters.Args{})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+		var names []string
+		for _, volume := range list.Volumes {
+			names = append(names, volume.Name)
+		}
+		return names, cobra.ShellCompDirectiveNoFileComp
+	}
+}
+
+// NetworkNames offers completion for networks
+func NetworkNames(dockerCli command.Cli) ValidArgsFn {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		list, err := dockerCli.Client().NetworkList(cmd.Context(), types.NetworkListOptions{})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+		var names []string
+		for _, network := range list {
+			names = append(names, network.Name)
+		}
+		return names, cobra.ShellCompDirectiveNoFileComp
+	}
+}
+
+// NoComplete is used for commands where there's no relevant completion
+func NoComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return nil, cobra.ShellCompDirectiveNoFileComp
+}
--- a/cli/command/config/cmd.go
+++ b/cli/command/config/cmd.go
@@ -1,10 +1,11 @@
 package config
 
 import (
-	"github.com/spf13/cobra"
-
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/command/completion"
+	"github.com/docker/docker/api/types"
+	"github.com/spf13/cobra"
 )
 
 // NewConfigCommand returns a cobra command for `config` subcommands
@@ -26,4 +27,19 @@ func NewConfigCommand(dockerCli command.Cli) *cobra.Command {
 		newConfigRemoveCommand(dockerCli),
 	)
 	return cmd
+}
+
+// completeNames offers completion for swarm configs
+func completeNames(dockerCli command.Cli) completion.ValidArgsFn {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		list, err := dockerCli.Client().ConfigList(cmd.Context(), types.ConfigListOptions{})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+		var names []string
+		for _, config := range list {
+			names = append(names, config.ID)
+		}
+		return names, cobra.ShellCompDirectiveNoFileComp
+	}
 }
--- a/cli/command/config/ls.go
+++ b/cli/command/config/ls.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/command/completion"
 	"github.com/docker/cli/cli/command/formatter"
 	flagsHelper "github.com/docker/cli/cli/flags"
 	"github.com/docker/cli/opts"
@@ -32,6 +33,7 @@ func newConfigListCommand(dockerCli command.Cli) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return RunConfigList(dockerCli, listOpts)
 		},
+		ValidArgsFunction: completion.NoComplete,
 	}
 
 	flags := cmd.Flags()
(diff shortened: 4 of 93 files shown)


cli-plugins/manager/metadata.go
```diff
@@ -25,4 +25,6 @@ type Metadata struct {
 	// Experimental specifies whether the plugin is experimental.
 	// Deprecated: experimental features are now always enabled in the CLI
 	Experimental bool `json:",omitempty"`
+	// Completions specifies whether the plugin supports command completion
+	Completions bool `json:",omitempty"`
```
thaJeztah, line 29: Perhaps we should consider updating the SchemaVersion as well

```
// SchemaVersion describes the version of this struct. Mandatory, must be "0.1.0"
```

Do you think we can combine this PR with https://github.com/docker/cli/pull/3254 ? (feel free to take that branch of course, if there's useful bits in it)

cmd/docker/completions.go
```diff
@@ -0,0 +1,40 @@
...
+func registerCompletionFuncForGlobalFlags(dockerCli *command.DockerCli, cmd *cobra.Command) {
+	cmd.RegisterFlagCompletionFunc(
+		"context",
+		func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			names, err := store.Names(dockerCli.ContextStore())
+			if err != nil {
+				return nil, cobra.ShellCompDirectiveError
+			}
+
+			return filterForCompletion(names, toComplete), cobra.ShellCompDirectiveDefault
```
marckhouzam, line 20: If there are no context that match what the user has typed, do you want the shell to suggest file name?
If not, you should return `cobra.ShellCompDirectiveNoFileComp`.  For example:
```
$ docker --context D[tab][tab]  # There is a context starting with 'D'
Development
$ docker --context Do[tab][tab] # There is NO context starting with 'Do' so we would get file completion
Documents/   Dockerfile    Dockerfile.new   Dockerfile.ori
```

cli-plugins/manager/metadata.go
ndeloof: I'm not sure about #3254. I understand we want to keep backward compatibility for users, but on the other hand using cobra completion v2 bring more flexibility and CLI plugins support, so should be preferred. If we support both, we will have to offer double maintenance...

ndeloof:
current limitation is that we can't offer completion for plugin flags, 
i.e `docker compose up -- [tab]` will actually list the root `docker` command flags, has flag completion is implemented without ability to plugin some custom hook

marckhouzam:
> current limitation is that we can't offer completion for plugin flags, i.e `docker compose up -- [tab]` will actually list the root `docker` command flags, has flag completion is implemented without ability to plugin some custom hook

This should work with Cobra 1.3.

Cobra will suggestion the root `docker` persistent flags, but will also call `ValidArgsFunction` for the `up` command, where you can provide flag names.  For this to work though, the `up` command must set the field `DisableFlagParsing` to `true`.

ndeloof:
@marckhouzam the challenge here, due to CLI plugin architecture, is that the `__complete` command is ran on `docker` CLI, and we need to redirect this call to `docker-compose` plugin. But as this command is dynamically created by cobra's `initCompleteCmd` and doesn't expose any hook in its execution, the plugin is not even executed when invoked with  `compose - [tab]`

ndeloof:
Seems I've found a solution by registering a plugin command stub, with `DisableFlagParsing: true`
 and a custom `ValidArgsFunction` to redirect __complete command to plugin

marckhouzam:
> Seems I've found a solution by registering a plugin command stub, with `DisableFlagParsing: true` and a custom `ValidArgsFunction` to redirect __complete command to plugin

That is excellent.  That is what `helm` does and what `kubectl` will be doing for plugins.

thaJeztah:
carrying the first commit in https://github.com/docker/cli/pull/3492, as I saw there was a minor merge conflict in `vendor.sum`, and so that this PR can focus on the local changes 👍

cli/command/completions.go
```diff
@@ -0,0 +1,92 @@
...
+var ShowContainerIDs = os.Getenv("DOCKER_COMPLETION_SHOW_CONTAINER_IDS") == "yes"
+
+// CompletionFn a function to be used by cobra command to offer command line completion
+type CompletionFn func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective)
+
+// ContainerNamesCompletionFn offers completion for container names and IDs
+// By default, only names are returned.
+// Set DOCKER_COMPLETION_SHOW_CONTAINER_IDS=yes to also complete IDs.
+func ContainerNamesCompletionFn(dockerCli Cli, all bool, filters ...func(types.Container) bool) CompletionFn {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
marckhouzam, line 21: I haven't tested so apologies if I'm not reading the code right.
I believe this function will complete container names, no matter what the `args` field is.
This would mean that doing
```
docker exec myContainer <TAB>
```
would continue suggesting other container names.
You may want to have a `repeat bool` argument for this function which would indicate if container names should continue to be suggested, or if they should only be suggested for the first argument.  And then base your logic on `len(args)`.

This may apply to other completion function as well.

cmd/docker/docker.go
```diff
@@ -47,10 +47,15 @@ func newDockerCommand(dockerCli *command.DockerCli) *cli.TopLevelCommand {
 		},
 		Version:               fmt.Sprintf("%s, build %s", version.Version, version.GitCommit),
 		DisableFlagsInUseLine: true,
+		CompletionOptions: cobra.CompletionOptions{
+			DisableDefaultCmd:   false,
+			HiddenDefaultCmd:    true,
+			DisableDescriptions: true,
```
marckhouzam, line 53: Out of curiosity, you don't want descriptions?
I notice that the current docker completion system does provide descriptions.

ndeloof: Some description are too long and break, haven't check the actual constraint to propose a fix

thaJeztah (changes requested):
Thanks! So overall, this looks like a good start. While testing it, I found some issues (also commented inline), and (perhaps I'm doing something wrong) `docker buildx <tab>` didn't want to complete for me (I was testing it inside the dev container, and installed bash-completion; it worked for the other commands).

Looking at what's implemented, I think there's still some things to be done to make it fully feature-compared with the existing completion script, for example, things like this are not (yet) implemented; https://github.com/docker/cli/blob/8d91c1442c8bef621ad4e7c2ccf129b9a8a11b1d/contrib/completion/bash/docker#L154-L156

Overall (besides code/review), I think it'd be ok to ship this, but considering it a "first implementation" (not yet complete), and at least in the meantime continuing shipping the existing completion scripts as well.

thaJeztah:
I noticed GItHub is hiding some of my review comments, so make sure to expand them / load more comments 😞

cli/command/completions.go
```diff
@@ -0,0 +1,92 @@
...
+		list, err := dockerCli.Client().ContainerList(cmd.Context(), types.ContainerListOptions{
+			All: all,
+		})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+		var names []string
```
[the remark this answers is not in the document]
ndeloof: IMHO pre-allocating slice with slice only make sense when 1. size is actually known (and here, we don't know as filter might result in a significant subset) 2. code is critical regarding performances, as this creates some "noise" in the codebase

cli-plugins/manager/cobra.go
```diff
@@ -49,11 +53,42 @@ func AddPluginCommandStubs(dockerCli command.Cli, cmd *cobra.Command) error {
...
+				cargs = append(cargs, toComplete)
+				os.Args = cargs
+				runCommand, err := PluginRunCommand(dockerCli, p.Name, cmd)
+				if err != nil {
+					return nil, cobra.ShellCompDirectiveError
+				}
+				err = runCommand.Run()
+				if err == nil {
+					os.Exit(0) // plugin already rendered complete data
+				}
```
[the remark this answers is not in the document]
ndeloof: child process indeed already returned as we run this.
The main reason doing so is to enforce no extra output is sent to stdout, which would break bash-completion protocol

e2e/cli-plugins/run_test.go
```diff
@@ -83,7 +83,7 @@ func TestHelpBad(t *testing.T) {
 
 	res := icmd.RunCmd(run("help", "badmeta"))
 	res.Assert(t, icmd.Expected{
-		ExitCode: 1,
+		ExitCode: 0,
```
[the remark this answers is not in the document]
ndeloof: We didn't had an homogeneous UX regarding `docker help doesnotexists` and `docker help invalidplugin`. The former one exit with status 0, while the latter return 1.

cli-plugins/manager/cobra.go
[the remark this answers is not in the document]
ndeloof: the reason for doing `if err == nil` is that go force us to have some `return` line whenever we exit. So would have to include a useless `return nil, cobra.ShellCompDirectiveError` that is just the same as the error case

cli/command/container/exec.go
```diff
@@ -73,6 +75,19 @@ func NewExecCommand(dockerCli command.Cli) *cobra.Command {
...
+	cmd.RegisterFlagCompletionFunc(
+		"env",
+		func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return os.Environ(), cobra.ShellCompDirectiveNoFileComp
+		},
+	)
+	cmd.RegisterFlagCompletionFunc(
+		"env-file",
+		func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return nil, cobra.ShellCompDirectiveDefault // _filedir
```
rumpl, line 87: What does `_filedir` mean here?

rumpl (approved):

cli/command/config/cmd.go
```diff
@@ -27,3 +27,18 @@ func NewConfigCommand(dockerCli command.Cli) *cobra.Command {
 	)
 	return cmd
 }
+
+// CompletionFn offers completion for swarm secrets
+func CompletionFn(dockerCli command.Cli) command.CompletionFn {
```
thaJeztah, line 32: One thing; so it looks like all completion functions (with the exception of `image.CompletionFn()` and `container.CompletionFn()`) are only used within the package itself;

Screenshot 2022-05-02 at 16 29 41

I generally like to keep things un-exported, unless needed; perhaps we should 

- un-export all of them (except for the `image` and `container` ones)
- should we have more descriptive names (for the exported ones at least)? (e.g., `CompleteNames()` ?); mostly wondering if there will be more "specialized" completion functions (per my earlier comments w.r.t. "in specific states" etc), in which case a more descriptive name would help.

e2e/cli-plugins/run_test.go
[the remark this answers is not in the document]
thaJeztah: Still a bit on the fence on this one; it feels like a "non-success" exist code is correct for non-existing commands (not a show-stopper, but perhaps something we should look into if it's fixable)

cli/command/container/exec.go
ndeloof: Default directive will list files in current dir as completion option, which is  relevant for this flag

ndeloof:
@thaJeztah I've addressed some of the reported completion issue, there's many places we can improve.
Typically `docker run --network=<TAB>` is easy to fix with 
```go
cmd.RegisterFlagCompletionFunc(
		"network",
		network.CompleteNames(dockerCli),
	)
```
but there's many other flags to address :)

thaJeztah:
I was about to give this one another look, but I see there's an issue in the latest iteration, @ndeloof 

```
#0 3.704 + go build -o /out/docker-windows-386.exe -tags ' osusergo' -ldflags ' -w -X "github.com/docker/cli/cli/version.GitCommit=5ea118aab0" -X "github.com/docker/cli/cli/version.BuildTime=2022-05-06T09:40:21Z" -X "github.com/docker/cli/cli/version.Version=20.10.2-683-g5ea118aab0.m"' github.com/docker/cli/cmd/docker
#80 6.592 package github.com/docker/cli/cmd/docker
#80 6.592 	imports github.com/docker/cli/cli/command/commands
#80 6.592 	imports github.com/docker/cli/cli/command/checkpoint
#80 6.592 	imports github.com/docker/cli/cli/command/container
#80 6.592 	imports github.com/docker/cli/cli/command/network
#80 6.592 	imports github.com/docker/cli/cli/command/container: import cycle not allowed
```

cli/command/completion/functions.go
```diff
@@ -0,0 +1,94 @@
...
+			ok := true
+			for _, fn := range filters {
+				if !fn(container) {
+					ok = false
+					break
+				}
+			}
+			if !ok {
+				continue
+			}
```
thaJeztah, line 56: Perhaps more readable (the double negative is a bit hard to read); reverse this and to rename the variable to `skip` (e.g.);

```suggestion
			skip := false
			for _, fn := range filters {
				if !fn(container) {
					skip = true
					break
				}
			}
			if skip {
				continue
			}
```

cli/command/completion/functions.go
```diff
@@ -0,0 +1,94 @@
...
+			names = append(names, image.RepoTags...)
+		}
+		return names, cobra.ShellCompDirectiveNoFileComp
+	}
+}
+
+// CompleteContainerNames offers completion for container names and IDs
+// By default, only names are returned.
+// Set DOCKER_COMPLETION_SHOW_CONTAINER_IDS=yes to also complete IDs.
+func CompleteContainerNames(dockerCli command.Cli, all bool, filters ...func(types.Container) bool) ValidArgsFn {
```
thaJeztah, line 34: I'm ok with keeping this for now, and iterate on this, but somewhat on the fence on the `bool` argument (which could also be a "filter" / "option") and  doing the filtering client-side

I _think_ most (if not all) of the filtering we need can be done with a `--filter` (although in some cases, we'd have to pass multiple filters). For example;

```bash
docker ps --filter status=exited --filter status=created
CONTAINER ID   IMAGE     COMMAND   CREATED          STATUS                      PORTS     NAMES
7b1647808d7b   busybox   "sh"      56 minutes ago   Exited (0) 56 minutes ago             stopped-container
2ee1ae95ecf4   busybox   "sh"      56 minutes ago   Created                               created-container
```

We could still look at implementing `!=` as filter (`--filter status!=running`), but in most cases, we'd still have to be more specific (e.g. a `paused` container cannot be started, so only excluding `running` likely won't work)

cli/command/config/ls.go
```diff
@@ -32,6 +32,9 @@ func newConfigListCommand(dockerCli command.Cli) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return RunConfigList(dockerCli, listOpts)
 		},
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return nil, cobra.ShellCompDirectiveError
+		},
```
thaJeztah, line 37: nit: perhaps we should consider having a `completion.NoComplete()` function, so that we can just reuse that for all of these;

```suggestion
		ValidArgsFunction: completion.NoComplete,
```

cli/command/container/list.go
```diff
@@ -40,6 +40,9 @@ func NewPsCommand(dockerCli command.Cli) *cobra.Command {
 		Annotations: map[string]string{
 			"category-top": "3",
 		},
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return nil, cobra.ShellCompDirectiveError
```
thaJeztah, line 44: Should this use `cobra.ShellCompDirectiveNoFileComp` ? Looks like currently it falls back to default (complete filenames); changing it to `ShellCompDirectiveNoFileComp` disables completion (probably same for some others; also see my comment on considering to have a reusable function for this perhaps 🤔 .

```suggestion
			return nil, cobra.ShellCompDirectiveNoFileComp
```

thaJeztah (changes requested):

cli/command/container/unpause.go
```diff
@@ -27,6 +29,9 @@ func NewUnpauseCommand(dockerCli command.Cli) *cobra.Command {
 			opts.containers = args
 			return runUnpause(dockerCli, &opts)
 		},
+		ValidArgsFunction: completion.CompleteContainerNames(dockerCli, false, func(container types.Container) bool {
+			return container.Status != "pause"
```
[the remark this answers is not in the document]
thaJeztah: > So we should move the human-readable translation in the CLI and deprecate this ASAP

Yeah, it's slightly complicated; the field can include (as shown above) `Up XX minutes`. For the CLI to show the correct duration (there's a ticket for that somewhere; I can try find it), we need it to know about the server time. So we either have to include that in some endpoints, or (alternatively) could return a Header in all responses that can be used by the CLI to do these calculations.

Another example is the `Created` (I think?) field, which is currently changed into a "human friendly" relative time, but this may show the wrong "XX minutes ago" if the daemon and cli's clocks are out of sync.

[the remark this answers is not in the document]
rumpl: Maybe the server should return a duration instead of a date/time.

rumpl:
Looks like all the comments were fixed, maybe we should merge this one?

thaJeztah:
LOL completion is funny; wondering if there's an option to prevent this (each `<TAB>` adds a new `paused-container` to the list; it probably completes automatically because there's only one match;


```bash
docker unpause paused-container paused-container paused-container paused-container paused-container paused-container paused-container paused-container
```

thaJeztah:
For the list of "improvements"; we should exclude default (built-in) networks;

```bash
docker network rm <TAB>
bridge                  docker_gwbridge         docker_mynetwork  host                    none
```

thaJeztah (approved):
LGTM, thanks!

PR #3434, opened 2022-02-22
Fix incorrect pointer inputs to `json.Unmarshal`

howardjohn:
See https://github.com/howardjohn/go-unmarshal-double-pointer for more
info on why this is not safe and how this is detected.

### DIFF
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -120,7 +120,7 @@ func (configFile *ConfigFile) LegacyLoadFromReader(configData io.Reader) error {
 // LoadFromReader reads the configuration data given and sets up the auth config
 // information with given directory and populates the receiver object
 func (configFile *ConfigFile) LoadFromReader(configData io.Reader) error {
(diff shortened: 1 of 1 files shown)


thaJeztah:
Oh, nice find!!

I see you have that checker; do you know if there's any (existing) linter that would detect these? We currently use `golangci-lint` (although not with all linters enabled); would be great if these could be caught in CI.

Are you planning to contribute your checker as a linter and/or as addition to an existing linter?

howardjohn:
@thaJeztah I have been in touch with some more knowledgeable Golang folks at my company and they have a plan to implement a linter but I don't think its a top priority. I tried myself but ran into trouble since it can be tricky when a method gets an `interface{}` type, for example - we lose a lot of the cases where it can happen which is why I did the runtime checker.

thaJeztah:
oh! I see I never replied to you ☺️ 😓. Thanks for the extra info, and no worries; if it would ever see the light as a linter; let me know!

PR #3435, opened 2022-02-22
bring back and expose BuildKitEnabled func

crazy-max:
Bring back `BuildKitEnabled` that is used by compose: https://github.com/docker/compose/blob/598b59f8bf0a86f3ba2a135c0489b63bb4285973/pkg/compose/build.go#L216-L219. @ndeloof Signature has changed, let me know if it sounds good to you.

Second commit fixes an issue where legacy builder was not used if wcow was reported server-side (see #3314).

### DIFF
--- a/cmd/docker/builder.go
+++ b/cmd/docker/builder.go
@@ -41,8 +41,7 @@ func newBuilderError(warn bool, err error) error {
 }
 
 func processBuilder(dockerCli command.Cli, cmd *cobra.Command, args, osargs []string) ([]string, []string, error) {
-	var useLegacy bool
-	var useBuilder bool
+	var useLegacy, useBuilder bool
 
 	// check DOCKER_BUILDKIT env var is present and
 	// if not assume we want to use the builder component
@@ -70,6 +69,12 @@ func processBuilder(dockerCli command.Cli, cmd *cobra.Command, args, osargs []st
 	// is this a build that should be forwarded to the builder?
 	fwargs, fwosargs, forwarded := forwardBuilder(builderAlias, args, osargs)
 	if !forwarded {
+		return args, osargs, nil
+	}
+
+	// wcow build command must use the legacy builder
+	// if not opt-in through a builder component
+	if !useBuilder && dockerCli.ServerInfo().OSType == "windows" {
 		return args, osargs, nil
 	}
 
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -58,6 +59,7 @@ type Cli interface {
 	ManifestStore() manifeststore.Store
 	RegistryClient(bool) registryclient.RegistryClient
 	ContentTrustEnabled() bool
+	BuildKitEnabled() (bool, error)
 	ContextStore() store.Store
 	CurrentContext() string
 	DockerEndpoint() docker.Endpoint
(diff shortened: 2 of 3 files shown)


cmd/docker/builder.go
```diff
@@ -73,6 +73,12 @@ func processBuilder(dockerCli command.Cli, cmd *cobra.Command, args, osargs []st
 		return args, osargs, nil
 	}
 
+	// wcow build command must use the legacy builder
+	// if not opt-in through a builder component
+	if !useBuilder && dockerCli.ServerInfo().OSType == "windows" {
+		return args, osargs, nil
+	}
```
crazy-max, line 80: Repro

```console
# on WSL2
docker buildx bake binary --set *.platform=windows/amd64
```

_Switch to Windows containers_ on Docker Desktop

```powershell
# on powerhell or cmd
> .\build\docker-windows-amd64.exe info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.7.1)
  compose: Docker Compose (Docker Inc., v2.2.3)
  extension: Manages Docker extensions (Docker Inc., v0.0.1-alpha.4)
  hub: Docker Hub (Docker Inc., 346ece73cb-dirty)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 4
 Server Version: 20.10.12
 Storage Driver: windowsfilter
  Windows:
 Logging Driver: json-file
 Plugins:
  Volume: local
  Network: ics internal l2bridge l2tunnel nat null overlay private transparent
  Log: awslogs etwlogs fluentd gcplogs gelf json-file local logentries splunk syslog
 Swarm: inactive
 Default Isolation: hyperv
 Kernel Version: 10.0 22000 (22000.1.amd64fre.co_release.210604-1628)
 Operating System: Windows 10 Pro Version 2009 (OS Build 22000.527)
 OSType: windows
 Architecture: x86_64
 CPUs: 32
 Total Memory: 63.93GiB
 Name: pc-sff
 ID: BJ23:IQXR:IXL3:FAEE:TGKX:5XWD:7ZKB:6GXF:P5BL:LU3E:T45O:MULU
 Docker Root Dir: C:\ProgramData\Docker
 Debug Mode: false
 Username: ...
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
 Product License: Community Engine
```

```dockerfile
FROM mcr.microsoft.com/windows/nanoserver:ltsc2022
WORKDIR /app
RUN ver
```

```powershell
> docker build .
Step 1/3 : FROM mcr.microsoft.com/windows/nanoserver:ltsc2022
ltsc2022: Pulling from windows/nanoserver
3ab33c1d9cc1: Pull complete
Digest: sha256:8aa115e244add588e9a71a38d302e8fc801f2e986be83a58d5e015e4cba8f613
Status: Downloaded newer image for mcr.microsoft.com/windows/nanoserver:ltsc2022
 ---> 6dde50c651d0
Step 2/3 : WORKDIR /app
 ---> Running in ed30b4ce8844
Removing intermediate container ed30b4ce8844
 ---> b261db46d93b
Step 3/3 : RUN ver
 ---> Running in ff8390d7dd39

Microsoft Windows [Version 10.0.20348.524]
Removing intermediate container ff8390d7dd39
 ---> f3e0e001378a
Successfully built f3e0e001378a
```

@thaJeztah Side note about what is reported by `docker info` in _Server_ with WCOW:

```
 Operating System: Windows 10 Pro Version 2009 (OS Build 22000.527)
```

This one is wrong. Should be smth like:

```
 Operating System: Windows 11 Pro Version 21H2 (OS Build 22000.527)
```

ndeloof (approved):

ndeloof:
we can manage the signature change on docker/compose, thanks for this

cmd/docker/builder.go
thaJeztah: @crazy-max could you open a ticket for that in https://github.com/moby/moby ?

crazy-max: yes sure

cmd/docker/builder.go
```diff
@@ -73,6 +73,12 @@ func processBuilder(dockerCli command.Cli, cmd *cobra.Command, args, osargs []st
 		return args, osargs, nil
 	}
 
+	// wcow build command must use the legacy builder
+	// if not opt-in through a builder component
+	if !useBuilder && dockerCli.ServerInfo().OSType == "windows" {
```
thaJeztah, line 78: I'm slightly concerned that there's now 2 implementations to do the BuildKit check (one in `BuildKitEnabled()`); I don't want to necessarily block this PR on that, but wondering if we can use the same code for both (so that we're sure the logic won't diverge)

crazy-max: Yes, I was thinking of moving some logic to `BuildKitEnabled` but in `processBuilder` we need to know if `DOCKER_BUILDKIT` is being used/enforced.

thaJeztah: could we extract the common bits to a non-exported function perhaps? (haven't checked), so something like;

```go
func (cli *DockerCli) BuildKitEnabled() (bool, error) {
    foo, _, err := doTheBuildKitCheck(cli)
    return foo, err
}
```

crazy-max: hum yeah we could that, let me see.

thaJeztah (approved):
LGTM

for those following along; @crazy-max did an implementation with the "common bits" extracted to a function, but because things are in separate packages, that meant the function had to be exported (so we couldn't put it in a non-exported / internal function).

Because of that, we decided that duplicating the code/logic for handling the `DOCKER_BUILDKIT` env-var (and other options) would be the better alternative for now.

PR #3437, opened 2022-02-23
introduce plugin.RunningStandalone utility func

ndeloof:
**- What I did**
introduced plugin.RunningStandalone

**- How I did it**
code from https://github.com/docker/buildx/blob/10debb577e0f2981ff5b4d09772be73a6267cebb/cmd/buildx/main.go#L46-L47

**- How to verify it**
This code is already in used, just made it reusable by other CLI plugins (typically, https://github.com/docker/compose/blob/740276f550ecb7da444b20dc21860ce2f2c547d6/cmd/main.go#L71)

image

### DIFF
--- a/cli-plugins/plugin/plugin.go
+++ b/cli-plugins/plugin/plugin.go
@@ -160,3 +160,11 @@ func newMetadataSubcommand(plugin *cobra.Command, meta manager.Metadata) *cobra.
 	}
 	return cmd
 }
+
+// RunningStandalone tells a CLI plugin it is run standalone by direct execution
+func RunningStandalone() bool {
+	if os.Getenv(manager.ReexecEnvvar) != "" {
+		return false
+	}
+	return len(os.Args) < 2 || os.Args[1] != manager.MetadataSubcommandName
+}


thaJeztah:
> code from https://github.com/docker/buildx/blob/10debb577e0f2981ff5b4d09772be73a6267cebb/cmd/buildx/main.go#L46-L47

I got a bit confused by that, as that code doesn't look like the code in this PR; did you link to the wrong code?

Here's the code you linked to (https://github.com/docker/buildx/blob/10debb577e0f2981ff5b4d09772be73a6267cebb/cmd/buildx/main.go#L46-L61)

```go
if os.Getenv("DOCKER_CLI_PLUGIN_ORIGINAL_CLI_COMMAND") == "" {
    if len(os.Args) < 2 || os.Args[1] != manager.MetadataSubcommandName {
        dockerCli, err := command.NewDockerCli()
        if err != nil {
            fmt.Fprintln(os.Stderr, err)
            os.Exit(1)
        }
        opts := cliflags.NewClientOptions()
        dockerCli.Initialize(opts)
        rootCmd := commands.NewRootCmd(os.Args[0], false, dockerCli)
        if err := rootCmd.Execute(); err != nil {
            os.Exit(1)
        }
        os.Exit(0)
    }
}
```

That code seems to be using the `` env-var (tbh, I forgot we had that one, and I'd have to look again how exactly it's used); https://github.com/docker/cli/blob/c780f7c4abaf67034ecfaa0611e03695cf9e4a3e/cli-plugins/manager/manager.go#L17-L21

thaJeztah:
Oh! Nevermind, I didn't look closely; so the code is different because it does basically the reverse, and (from looking up that `DOCKER_CLI_PLUGIN_ORIGINAL_CLI_COMMAND `. I just see that's the cons you're using (instead of using the literal string)

I need more coffee ☕

thaJeztah (approved):
LGTM

thaJeztah:
okay, let's get this one in

PR #3443, opened 2022-02-25
Remove uses of deprecated io/ioutil, and use new t.TempDir() and t.Cleanup() in tests

thaJeztah:
I had a "WIP" branch for this locally, and decided to finish the work; there's quite some uses of this package, so I split this up into commits per-package for easier reviewing.

Current versions of Go also added `t.TempDIr()` and `t.Cleanup()` for tests, so while at it, I updated tests to use those (which simplified the tests).

Finally, I updated the `golangci-lint` configuration to prevent the package from accidentally being added again.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,6 +2,7 @@ linters:
   enable:
     - bodyclose
     - deadcode
+    - depguard
     - dogsled
     - gocyclo
     - goimports
@@ -33,6 +34,13 @@ run:
     - .*generated.*
 
 linters-settings:
+  depguard:
+    list-type: blacklist
+    include-go-root: true
+    packages:
+      # The io/ioutil package has been deprecated.
+      # https://go.dev/doc/go1.16#ioutil
+      - io/ioutil
   gocyclo:
     min-complexity: 16
   govet:
--- a/cli-plugins/manager/manager.go
+++ b/cli-plugins/manager/manager.go
@@ -1,7 +1,6 @@
 package manager
 
 import (
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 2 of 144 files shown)


thaJeztah:
When reviewing, I recommend doing so per commit (as github doesn't like showing the whole diff 😅) most of them should be easy to review, but perhaps some may need closer looking (where I refactored or removed some test-utils for `t.TempDir()` etc.

thaJeztah:
@silvin-lubecki @ndeloof @crazy-max ptal 🤗

.golangci.yml
```diff
@@ -33,6 +34,13 @@ run:
     - .*generated.*
 
 linters-settings:
+  depguard:
+    list-type: blacklist
+    include-go-root: true
+    packages:
+      # The io/ioutil package has been deprecated.
+      # https://go.dev/doc/go1.16#ioutil
+      - io/ioutil
```
crazy-max, line 43: we should have that in buildkit too (cc @tonistiigi)

crazy-max (approved):
LGTM

ndeloof (approved):
LGTM
I like use of `t.TempDir()` that don't force us to manage a `cleanup` function 👍

thaJeztah:
Yes! `t.TempDir()` is really nice!!

I also want to have a look at uses of `defer` altogether in tests (not really a priority though), and perhaps change them to use `t.Cleanup()` instead. Probably doesn't make a lot of difference, but `t.Cleanup()` was designed for that, so _perhaps_ there's advantages (e.g. perhaps it takes time taken by those into account when reporting test-durations etc)

thaJeztah:
ok, let me get this one in; thanks for reviewing!

PR #3444, opened 2022-02-25
dummy circleci config

crazy-max:
follow-up #3381 

adds a dummy CircleCI config file to avoid GitHub status failures reported on branches that don't use CircleCI. This file should be deleted when all branches are no longer dependent on CircleCI.

cc @thaJeztah 


thaJeztah (approved):
LGTM, thanks!

PR #3445, opened 2022-02-26
cli/compose/schema: make version: "3" equivalent to "3.x" (latest)

thaJeztah:
follow-up to https://github.com/docker/cli/pull/3257

Previously, `version: "3"` was equivalent to `version: "3.0"`, which
caused confusion for many users, as they expected it to be "3.x".

docker-compose and docker compose (v2) have adopted the compose-spec
(https://compose-spec.io), which no longer has a version field in
the compose file, and always picks the "latest" supported version.

This changes how `docker stack` interprets "major" version numbers
specified in compose-files:

When only the major version ("3") is specified, it is now equivalent
to "3.x" (latest supported v3 schema).

Compose-files that specify both major and minor version (e.g. "3.0"
or "3.1") continue to use the existing behavior; validation is down-
graded to the specified version and will produce an error if options
are used that are not supported in that schema version. This allows
users to locally verify that a composse-file does not use options
that are not supported in the intended deployment environment (for
example if the deploy environment only supports older versions of
the schema).

### DIFF
--- a/cli/compose/loader/loader_test.go
+++ b/cli/compose/loader/loader_test.go
@@ -180,7 +180,7 @@ func strPtr(val string) *string {
 }
 
 var sampleConfig = types.Config{
-	Version: "3.0",
+	Version: "3.10",
 	Services: []types.ServiceConfig{
 		{
 			Name:        "foo",
--- a/cli/compose/schema/schema.go
+++ b/cli/compose/schema/schema.go
@@ -40,7 +40,8 @@ func init() {
 }
 
 // Version returns the version of the config, defaulting to the latest "3.x"
-// version (3.10).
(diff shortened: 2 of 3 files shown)


ndeloof (approved):
+10000000

cli/compose/loader/loader_test.go
```diff
@@ -180,7 +180,7 @@ func strPtr(val string) *string {
 }
 
 var sampleConfig = types.Config{
-	Version: "3.0",
+	Version: "3.10",
```
thaJeztah, line 183: This was the config I forgot to update that made the test fail; all green now

thaJeztah:
Let's get this one in 👍

PR #3377, opened 2021-12-03
cli/command: Add quiet option for create and run

ch33hau:
**- What I did**
fixes #3181 
The newly added `quiet` option for `docker run` and `docker create` will suppress `pull`'s output, other outputs remain.

**- How I did it**
The main changes are on `create.go`, `run.go` will make use of it when calling `createContainer`.

**- How to verify it**
```
# see the usage of --quiet
docker run --help
docker create --help

# run
docker rmi hello-world
docker run -ti --quiet hello-world

# create 
docker rmi hello-world
docker create --quiet hello-world
```

Below output shouldn't show unless without the `--quiet` option:
```
Unable to find image 'hello-world:latest' locally
latest: Pulling from library/hello-world
2db29710123e: Already exists
Digest: sha256:cc15c5b292d8525effc0f89cb299f1804f3a725c8d05e158653a563f15e4f685
Status: Downloaded newer image for hello-world:latest
```

**- Description for the changelog**

Add `quiet` option for `create` and `run` commands to suppress `pull`'s output.

PR #3449, opened 2022-03-01
docs: add missing documentation for --pull flag

thaJeztah:
addresses https://github.com/docker/docker.github.io/issues/14135

These options were added in 22cd418967f09179520dc61f49fa3697bea0daf8 (https://github.com/docker/cli/pull/1498), but did not update the documentation. That PR was already included in docker 20.10, so adding the `cherry-pick` label as well.

I noticed this when looking at https://github.com/docker/cli/pull/3377 (I started this PR with that commit)

### DIFF
--- a/docs/reference/commandline/create.md
+++ b/docs/reference/commandline/create.md
@@ -98,6 +98,7 @@ Options:
       --privileged                    Give extended privileges to this container
   -p, --publish value                 Publish a container's port(s) to the host (default [])
   -P, --publish-all                   Publish all exposed ports to random ports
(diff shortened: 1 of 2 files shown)


ndeloof (approved):
doc is also aligned withhttps://github.com/compose-spec/compose-spec/blob/master/spec.md#pull_policy

thaJeztah:
Ah, nice, yes. Forgot for a bit that it also defines it (but I recall we had a discussion on making them match as much as possible)

Perhaps https://github.com/docker/cli/pull/3377 is also a candidate to add to `docker compose run` btw

PR #3450, opened 2022-03-01
docs: unify "docker create" and "docker run" reference

thaJeztah:
The `docker create` command shares most (all) of its options with `docker run`,
which uses `docker create` under the hood. The `docker create` reference docs
already referred users to the `docker run` sections for details, but some
flags were only documented on the `docker create` page.

This patch:

- moves those flags from the `docker create` to the `docker run` page
- does some minor rephrasing and touch-ups.

### DIFF
--- a/docs/reference/commandline/create.md
+++ b/docs/reference/commandline/create.md
@@ -134,35 +134,51 @@ Options:
 
 ## Description
 
-The `docker create` command creates a writeable container layer over the
-specified image and prepares it for running the specified command.  The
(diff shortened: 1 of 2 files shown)


thaJeztah:
@ndeloof @rumpl ptal 🤗

ndeloof (approved):

thaJeztah:
Thanks for review; I'll bring this in, and do some rebasing (and might prepare cherry-picks of some docs changes)

PR #3452, opened 2022-03-01
vendor: reduce replace rules (upgrades some packages)

thaJeztah:
### vendor: remove replace rules for k8s.io packages

They're no longer used, so we can remove the replace rules.

### vendor: golang.org/x/sys v0.0.0-20211025201205-69cdffdb9359

remove the replace rule to update it to the actual version specified:

full diff: https://github.com/golang/sys/compare/63515b42dcdf...69cdffdb9359

### vendor: golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1

remove the replace rule to use the actual version (no changes in vendored code)

### vendor: golang.org/x/text v0.3.4

remove the replace rule to update it to the actual version specified:

full diff: https://github.com/golang/text/compare/v0.3.3...v0.3.4

### vendor: gopkg.in/yaml.v2 v2.4.0

full diff: https://github.com/go-yaml/yaml/compare/v2.2.8...v2.4.0

### vendor: gotest.tools/v3 v3.0.3

https://github.com/gotestyourself/gotest.tools/compare/v3.0.2...v3.0.3

### vendor: move github.com/moby/sys packages to "requires" ("indirect")

These were an "upgrade" not a "downgrade" (It was put in vendor.conf
to push it ahead of the version resolved by go mod). Let's move it
to the "requires" section as an indirect, that way it will dissolve
once other modules require this version or up.

### vendor: remove redundant replace rules

These dependencies were either unused, or go modules already resolved them
to the same version as specified in the replace rule, so those could be
removed.

### vendor: github.com/google/go-cmp v0.5.5

it was downgraded to v0.2.0, but should be safe to upgrade

full diff: https://github.com/google/go-cmp/compare/v0.2.0...v0.5.5


thaJeztah:
This removes most of the replace rules; after this we can do some updates of dependencies.

thaJeztah:
@crazy-max @tonistiigi ptal

vendor.mod
```diff
@@ -47,37 +49,10 @@ require (
 )
 
 replace (
-	cloud.google.com/go => cloud.google.com/go v0.44.3
 	github.com/docker/distribution => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
```
crazy-max, line 52: maybe this one should be removed from `replace` too? and use latest v2 stable?

crazy-max (approved):

vendor.mod
thaJeztah: Yes, I wanted to wait for v2.8.1, then update to that version

thaJeztah:
let me bring this one in 👍

PR #3467, opened 2022-03-13
docs: dockerd: fix an unclickable and non-direct link in Markdown's blockquote area

pzhlkj6612:
**- What I did**

Hello! In the docs:

1. I put the link behind a piece of text. The current webpage shows the URL in plaintext ([Wayback Machine archive](http://web.archive.org/web/20220313051005/https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file)). I think this may fix it.
2. I replaced the old link with non-redirect one.

**- How I did it**

Just editing Markdown.

**- How to verify it**

I have no environment to build the docs, but I guess this fix will work correctly.

No changelog entity required.

none.


thaJeztah (approved):
LGTM, thanks!

Issue #3421, opened 2022-02-08
Column size calculation seems off when there are non-English characters in a description


masterl:
**Description**

Looks like when using docker search if a description has non-English characters, the column sizes are calculated wrong.

**Steps to reproduce the issue:**
1. `docker search gitlab`

**Describe the results you received:**

```
NAME                                         DESCRIPTION                                     STARS     OFFICIAL   AUTOMATED
gitlab/gitlab-ce                             GitLab Community Edition docker image based …   3518                 [OK]
sameersbn/gitlab                             Dockerfile to build a GitLab image for the D…   1183                 [OK]
gitlab/gitlab-runner                         GitLab CI Multi Runner used to fetch and run…   761                  [OK]
gitlab/gitlab-ee                             GitLab Enterprise Edition docker image based…   304                  
twang2218/gitlab-ce-zh                       汉化的 GitLab 社区版 Docker Image                     254                  [OK]
jangrewe/gitlab-ci-android                   GitLab CI image for building Android apps       63                   [OK]
```

**Describe the results you expected:**

```
NAME                                         DESCRIPTION                                     STARS     OFFICIAL   AUTOMATED
gitlab/gitlab-ce                             GitLab Community Edition docker image based …   3518                 [OK]
sameersbn/gitlab                             Dockerfile to build a GitLab image for the D…   1183                 [OK]
gitlab/gitlab-runner                         GitLab CI Multi Runner used to fetch and run…   761                  [OK]
gitlab/gitlab-ee                             GitLab Enterprise Edition docker image based…   304                  
twang2218/gitlab-ce-zh                       汉化的 GitLab 社区版 Docker Image                 254                  [OK]
jangrewe/gitlab-ci-android                   GitLab CI image for building Android apps       63                   [OK]
```

**Additional information you deem important (e.g. issue happens only occasionally):**

-

**Output of `docker version`:**

```
Client:
 Version:           20.10.12
 API version:       1.41
 Go version:        go1.17.5
 Git commit:        e91ed5707e
 Built:             Mon Dec 13 22:31:40 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.12
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.17.5
  Git commit:       459d0dfbbb
  Built:            Mon Dec 13 22:30:43 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.5.9
[...]

PR #3468, opened 2022-03-14
adopt github.com/WeiZhang555/tabwriter to support non-Latin rune size

ndeloof:
closes https://github.com/docker/cli/issues/3421
also reported https://github.com/golang/go/issues/51650, maybe we will get a fix in standard library at some point?

**- What I did**
replaced golang's tabwrittter with github.com/WeiZhang555/tabwriter to support non-latin character width

**- How I did it**
replace imports (github.com/WeiZhang555/tabwriter is a fork from golang's tabwriter)

**- How to verify it**
`docker search twang2218`

docker table output supports non-latin characters to compute column sizes


thaJeztah:
Couple of things (just from a very cursory glance);

- the https://github.com/WeiZhang555/tabwriter package looks to be a fork of the stdlib tabwriter, but it's based on a version from 6 years ago; we should check if there's no fixes we're missing that were made since then; https://github.com/golang/go/commits/master/src/text/tabwriter/tabwriter.go)
- I see it originally used `golang.org/x/text/width`, but there was an issue with that, which made it switch to use `github.com/mattn/go-runewidth` - slightly wondering if that was a _bug_ in `golang.org/x/text` (and if so, if that either was fixed upstream, or if there's something to fix?)
- Overall, it would be nice if golang's own `text/tabwriter` would do this correct "out of the box' (or if out of scope for that, if this would be something that could be part of `golang.org/x/text`)

ndeloof:
I've reported https://github.com/golang/go/issues/51650 to (maybe?) get some feedback on this regarding the standard library and see what would be the best option for this.

rumpl:
`text/tabwriter` is frozen, no new features and bugfixes will get into it. It also doesn't support non-latin things and never will, I think it's a great candidate to remove and use something else that fixes our issues

ndeloof:
working on https://github.com/docker/cli/pull/3568 as an alternative

PR #3469, opened 2022-03-14
Handle relative source mounts

### LINKED ISSUE #1203: Support relative paths with bind mounts (volumes)
As previously discussed in https://github.com/moby/moby/issues/4830, it would be useful to be able to bind-mount relative paths like this:

```
docker run --mount type=bind,source=host-dir,target=/app ...
```

> Note: The original issue suggests `docker run -v host-dir:/app ...` which might (probably?) be confused with named volumes so I think it's more likely to expect it from the newer `--mount` option.

The two main motivations for this are:

- Consistency with docker-compose.
- Ability to write cross-platform `docker run` commands, e.g., we have a npm script that currently does something like `"start": "docker run -v $PWD:/app ..."`. It's not easy at all to write a command that would work across Linux, macOS and Windows today.

rumpl:
**- What I did**

With this change it is now possible to give a relative path to the --volume and
--mount flags.

```console
$ docker run --mount type=bind,source=./,target=/test ...
```

```console
$ docker run -v .:/test ...
```

Fixes #1203

**- How I did it**

Updated opts parsing for volumes and mounts

**- How to verify it**

Run `docker run --rm -v .:/test busybox ls -l /test` and you should see the listing of your current working dir

The run command accepts relative source paths in `-v/--volume` and `-m/--mount` flags

image

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"strconv"
@@ -15,6 +16,7 @@ import (
 	"github.com/docker/cli/cli/compose/loader"
 	"github.com/docker/cli/opts"
 	"github.com/docker/docker/api/types/container"
+	mounttypes "github.com/docker/docker/api/types/mount"
 	networktypes "github.com/docker/docker/api/types/network"
 	"github.com/docker/docker/api/types/strslice"
 	"github.com/docker/docker/api/types/versions"
@@ -348,10 +350,25 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 	// add any bind targets to the list of container volumes
 	for bind := range copts.volumes.GetMap() {
 		parsed, _ := loader.ParseVolume(bind)
+
 		if parsed.Source != "" {
+			toBind := bind
+
+			if parsed.Type == string(mounttypes.TypeBind) {
+				if arr := strings.SplitN(bind, ":", 2); len(arr) == 2 {
+					hostPart := arr[0]
+					if strings.HasPrefix(hostPart, "."+string(filepath.Separator)) || hostPart == "." {
+						if absHostPart, err := filepath.Abs(hostPart); err == nil {
+							hostPart = absHostPart
+						}
+					}
+					toBind = hostPart + ":" + arr[1]
+				}
+			}
+
 			// after creating the bind mount we want to delete it from the copts.volumes values because
 			// we do not want bind mounts being committed to image configs
-			binds = append(binds, bind)
+			binds = append(binds, toBind)
 			// We should delete from the map (`volumes`) here, as deleting from copts.volumes will not work if
 			// there are duplicates entries.
 			delete(volumes, bind)
--- a/opts/mount.go
+++ b/opts/mount.go
@@ -4,6 +4,7 @@ import (
 	"encoding/csv"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -92,6 +93,11 @@ func (m *MountOpt) Set(value string) error {
 			mount.Type = mounttypes.Type(strings.ToLower(value))
 		case "source", "src":
 			mount.Source = value
+			if strings.HasPrefix(value, "."+string(filepath.Separator)) || value == "." {
+				if abs, err := filepath.Abs(value); err == nil {
+					mount.Source = abs
+				}
+			}
 		case "target", "dst", "destination":
 			mount.Target = value
 		case "readonly", "ro":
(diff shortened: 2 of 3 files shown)


opts/mount.go
```diff
@@ -92,6 +93,11 @@ func (m *MountOpt) Set(value string) error {
 			mount.Type = mounttypes.Type(strings.ToLower(value))
 		case "source", "src":
 			mount.Source = value
+			if strings.HasPrefix(value, "."+string(filepath.Separator)) || value == "." {
+				if abs, err := filepath.Abs(value); err == nil {
```
thaJeztah, line 97: Do we need to return the error if this fails? (Not entirely sure if needed; it seems like the only case this would fail is if it's failing to resolve `os.GetCWD()` on Linux (although the Windows code seems to have more possible error conditions)

opts/mount_test.go
```diff
@@ -28,6 +29,22 @@ func TestMountOptString(t *testing.T) {
 	assert.Check(t, is.Equal(expected, mount.String()))
 }
 
+func TestMountRelative(t *testing.T) {
+	var mount MountOpt
+
+	assert.NilError(t, mount.Set("type=bind,source=.,target=/target"))
```
thaJeztah, line 35: Can you add test-cases for `./` as well? Might be worth creating an array/table with test-cases (and using sub-tests) if that helps making it less DRY.

ndeloof (approved):

opts/mount.go
```diff
@@ -92,6 +93,11 @@ func (m *MountOpt) Set(value string) error {
 			mount.Type = mounttypes.Type(strings.ToLower(value))
 		case "source", "src":
 			mount.Source = value
+			if strings.HasPrefix(value, "."+string(filepath.Separator)) || value == "." {
```
thaJeztah, line 96: I guess as a follow-up, we should consider;

- also allowing `../` (relative parent directories)
- wondering if we should always do a `filepath.Clean()`

cli/command/container/opts.go
```diff
@@ -348,10 +349,23 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 	// add any bind targets to the list of container volumes
 	for bind := range copts.volumes.GetMap() {
 		parsed, _ := loader.ParseVolume(bind)
+
 		if parsed.Source != "" {
+			toBind := bind
```
thaJeztah, line 354: This code is a bit hairy (not your fault), and could use some cleaning up at least (we're parsing the `bind` into a structured type, but effectively only to do some mild "validation" (however, we're throwing away any error caused by `loader.ParseVolume()`.

One thing we _can_ do already (if I see it correctly), is to use the `parsed.Type` field to only run this code if it's a bind-mount.

Ideally (but that bit might be ok to be done separate), we should

- possibly make have `loader.ParseVolume()` already do the resolving (I guess that won't work for the compose case, as it would have to resolve _relative to_ the compose-file)
- if the above is not an option because of compose-file, we could update the `parsed.Source`, and;
- change `types.ServiceVolumeConfig` (which is returned by `loader.ParseVolume()`) to have a `.String()` (or a separate utility for that), which returns `parsed` in its canonical `string` representation.

opts/mount.go
rumpl: For the second point, `filepath.Abs` already calls `Clean`

cli/command/container/opts.go
rumpl: I added the check for the `parsed.Type`, I didn't really want to touch the loader code since it touches the compose/swarm things and figured it would surely break something else. I do agree we could have something somewhere that can return a canonical string representation, we should also try and think about how we can make volume and mounts options parsing be in the same place.

opts/mount_test.go
rumpl: Done

thaJeztah (approved):
LGTM

jdruizip:
Although the fix seems to work with paths such as `.` or `./test`, a longer path such as `./test/case` seems to always trigger an error on Windows.

When I try
`docker run --rm -v ./foo/bar:/test busybox ls -l /test`
I always encounter the error:
`docker: Error response from daemon: create foo/bar: "foo/bar" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.`

I am on Windows, and a path such as:

`docker run --rm -v .\foo\bar:/test busybox ls -l /test` (using `\` instead of `/`)

Will run successfully, but then it is no longer valid on unix.  It was my understanding that this change was implemented to help cross-platform compatibility, and to sidestep the need for shell-specific `pwd` calls.  docker-compose is able to handle this well, using `/` on Windows without issue.  Even `-v %cd%/foo/bar:/foo/bar` works well on Windows, so its really not obvious why only backslashes would work in this context.

In any case, even this isn't fixed to be fully cross-platform, the errors should be updated to reflect that not only absolute paths are usable.

thaJeztah:
@jdruizip thanks for reporting; could you open a ticket for that? Things can easily get lost when commented on merged pull-requests 🙈

PR #3383, opened 2021-12-08
update go to 1.17

thaJeztah:
carries / closes https://github.com/docker/cli/pull/3258

Removes the platform based switch between different versions.

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3473, opened 2022-03-16
Upgrade to Go v1.18

ndeloof:
**- What I did**
Bump go to v1.18 in project

**- How I did it**
Just changed go version and updated vendor, using CI to check anything breaks

**- How to verify it**
isn't this what CI is for :P 

docker CLI is now built with Go v1.18


DasSkelett:
I suspect the work in https://github.com/docker/cli/pull/3383 has to be completed before an update to 1.18 can be considered.

I also don't think blindly bumping the go directive in go.mod makes sense, until features from Go 1.18 are used.

ndeloof:
as noted, the goal of this draft PR was just to check the potential impact, so we have a better idea about the missing bits before we can benefit from go 1.18 features

PR #3486, opened 2022-03-25
vendor: github.com/docker/docker 4a26fdda76d996708aa8100dd23ea90215984451

thaJeztah:
some follow-ups coming; keeping this to just updating the vendor code

full diff: https://github.com/docker/docker/compare/83b51522df43866e39f7c757a6ab84ef4050eeb3...4a26fdda76d996708aa8100dd23ea90215984451


vendor.mod
```diff
@@ -28,6 +28,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.0.2
 	github.com/pkg/errors v0.9.1
+	github.com/rootless-containers/rootlesskit v0.14.6 // indirect
```
thaJeztah, line 31: This is something we need to fix in moby, as we shouldn't get this additional dependency.

I went looking "why" we get this dependency, and it looks like [`GetRootlessKitClient()`](https://github.com/moby/moby/blob/6b9b445af6c7908992632cff6c30cbf6a4c617ac/rootless/rootless.go#L31-L39) uses this dependency. While that function is only used on the daemon side, it's in the same package as [`RunningWithRootlessKit()`](https://github.com/moby/moby/blob/6b9b445af6c7908992632cff6c30cbf6a4c617ac/rootless/rootless.go#L22-L29), which is used in the CLI through [`registry.CertsDir()`](https://github.com/moby/moby/blob/2b70006e3bfa492b8641ff443493983d832955f4/registry/config_unix.go#L13-L24), which is used by [`registry.HostCertsDir()`](https://github.com/moby/moby/blob/7cba4ffa30466156b0915e9effe0a5d8c085e75e/registry/registry.go#L18-L21) which is called in various locations.

ndeloof: do you want to postpone this PR waiting for this to be fixed?

thaJeztah:
@ndeloof @crazy-max this should be ready for review; I updated to latest upstream, so that the new dependency is no longer there


not sure what's going on with the `vendor` validate though; it appears to be hanging on all pull requests 😞

crazy-max:
> not sure what's going on with the `vendor` validate though; it appears to be hanging on all pull requests 😞

Yeah stuck with this git repo has some issues:

```
#11 81.23 get "contrib.go.opencensus.io/exporter/ocagent": found meta tag vcs.metaImport{Prefix:"contrib.go.opencensus.io/exporter/ocagent", VCS:"git", RepoRoot:"https://github.com/census-ecosystem/opencensus-go-exporter-ocagent"} at //contrib.go.opencensus.io/exporter/ocagent?go-get=1
#11 81.76 get "contrib.go.opencensus.io/resource": found meta tag vcs.metaImport{Prefix:"contrib.go.opencensus.io/resource", VCS:"git", RepoRoot:"https://github.com/census-ecosystem/opencensus-go-resource"} at //contrib.go.opencensus.io/resource?go-get=1
```

We should change `GOXPROXY` to `GOXPROXY=https://goproxy.io,direct`.

thaJeztah:
Ah, cool; looks like we got it un-stuck

@ndeloof @rumpl ptal

ndeloof (approved):

PR #3493, opened 2022-03-26
Update to debian bullseye

thaJeztah:
follow-up to https://github.com/docker/cli/pull/3383

this currently fails in cross-compile, so we should find out what's causing the failures

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -22,7 +22,7 @@ jobs:
           - connhelper-ssh
         base:
           - alpine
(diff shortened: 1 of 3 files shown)


crazy-max (approved):

thaJeztah:
I swapped the order of commits to not break `git bisect` if we'd ever need it (checked in https://github.com/docker/cli/pull/3503)

Thanks Tõnis for the fix! Let's get this one in

PR #3495, opened 2022-03-26
gofmt with go1.17

thaJeztah:
**- How to verify it**


opts/hosts_windows.go
```diff
@@ -1,3 +1,4 @@
+//go:build windows
 // +build windows
```
thaJeztah, line 2: same here

crazy-max (approved):

PR #3492, opened 2022-03-26
Bump cobra to v1.4.0

thaJeztah:
taking the first commit from https://github.com/docker/cli/pull/3429, which had a merge conflict in `vendor.sum`, so I did a quick rebase; moving it out of that PR to get this in already, and so that that PR can focus on the local changes 👍 

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3497, opened 2022-03-26
e2e: cleanup TestGlobalHelp() to be less brittle

thaJeztah:
- relates to https://github.com/docker/cli/pull/3492
- relates to https://github.com/docker/cli/pull/3429

- don't make the checks positional (just match that we find them, and  that we don't find them multiple times)
- account for leading whitespace to change instead of hard-coding the   number of spaces before output.
- change the badopt check; I think it should be sufficient to check   that the bad option was printed and that "run --help" output is  printed.

**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah:
@ndeloof ptal (this one may conflict slightly with https://github.com/docker/cli/pull/3429, but helps with updating cobra to 1.4 before that)

ndeloof (approved):

PR #3499, opened 2022-03-27
CLI flag for running container as the current user

### LINKED ISSUE #3498: Feature request: CLI flag for running container as the current user
There are many use cases where Docker containers are used as development environments to bundle dependencies and provide consistent development experience. Docker daemon [uses such an approach as well](https://github.com/moby/moby/blob/master/docs/contributing/set-up-dev-env.md)). In order to allow developers to modify the files in the IDE outside the container, the project’s working directory is often mounted inside the container like this:

```bash
$ docker run -v $(pwd):/mount my_container /bin/bash
```

However, this creates a problem where all the files, created by the process inside the container, are owned by this process’ user, which is often root. Removing root-owned files from the project directory requires juggling `sudo` commands around and is tedious in general. 

Often it’s not even necessary to run the containerized process as root at all, and the following command can be executed to run it as the current user:

```bash
$ docker run --user $(id -u):$(id -g) -v $(pwd):/mount my_container /bin/bash
```

However, this is cumbersome to type in the terminal and bloats the scripts in which it’s used. It would be neat to have a dedicated boolean flag, for instance `--current-user`, which would fetch the current user’s UID and GID (using [Go’s standard library](https://pkg.go.dev/os/user#User)) and use them for running the container’s process.

This would shorten the above example to:

```bash
$ docker run --current-user -v $(pwd):/mount my_container /bin/bash
```

kponichtera:
Signed-off-by: Konrad Ponichtera <konpon96@gmail.com>

Closes #3498

**- What I did**

I’ve added the `--current-user` boolean flag to the `docker run` command to avoid `docker run -u $(id -u):$(id -g)` construct which often appears in the scripts that use Docker containers for development.

**- How I did it**

The user parameter of the Docker API call is now built by the dedicated `parseUser` function in `opts.go` file. It checks if `--current-user` flag was set. In such case, it fetches current user UID and GID using [Go’s standard library](https://pkg.go.dev/os/user#Current), builds the string `UID:GID`, and passes it to the Docker API. The `--user` flag has precedence over `--current-user` - if both are specified, the `--user` one will be taken. 

There is also a unit test, which verifies the behavior of the function that builds the API call user field. I’ve also updated bash and zsh autocompletion files, as well as the Markdown reference document for the `run` command.

If there are some things that should be done and which I missed, please let me know and I will address them.

**- How to verify it**

1. Build repo with `docker buildx bake`
2. Run `./build/docker run -it --current-user ubuntu /bin/bash` - the Ubuntu container will run the process inside with the same UID and GID as the CLI caller’s process.
3. Run `./build/docker run -it --current-user -u 1234:1234 ubuntu /bin/bash` - the `-u` flag will take precedence

Added `--current-user` flag to the `docker run` command.

cute_animal

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/user"
 	"path"
 	"path/filepath"
 	"reflect"
@@ -86,6 +87,7 @@ type containerOptions struct {
 	memorySwap         opts.MemSwapBytes
 	kernelMemory       opts.MemBytes
 	user               string
+	currentUser        bool
 	workingDir         string
 	cpuCount           int64
 	cpuShares          int64
@@ -191,6 +193,7 @@ func addFlags(flags *pflag.FlagSet) *containerOptions {
 	flags.BoolVarP(&copts.tty, "tty", "t", false, "Allocate a pseudo-TTY")
 	flags.Var(copts.ulimits, "ulimit", "Ulimit options")
 	flags.StringVarP(&copts.user, "user", "u", "", "Username or UID (format: <name|uid>[:<group|gid>])")
+	flags.BoolVarP(&copts.currentUser, "current-user", "", false, "Run as the user that called the CLI")
 	flags.StringVarP(&copts.workingDir, "workdir", "w", "", "Working directory inside the container")
 	flags.BoolVar(&copts.autoRemove, "rm", false, "Automatically remove the container when it exits")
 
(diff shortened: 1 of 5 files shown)


kponichtera:
I have modified the code and implemented the suggestion, mentioned in https://github.com/docker/cli/issues/3498. Instead of new `--current-user` flag, running the container as current user is now done though setting already existing `--user` flag to value `current`.

petross8901 (approved):

Cornul11 (approved):

albers:
Please see the recent discussion in https://github.com/docker/cli/issues/3498#issuecomment-1080344498

kponichtera:
@albers Looking at the discussion in https://github.com/docker/cli/issues/3498 it seems like a dedicated `--current-user` flag would be a better solution than trying to find a safe value for the `--user` one. Should I revert this PR to state with the `--current-user` flag?

albers:
> Should I revert this PR to state with the --current-user flag?

I think, yes.
@thaJeztah @ndeloof WDYT?

PR #3500, opened 2022-03-27
Change "COMPANY_NAME" to "PACKAGER_NAME"

thaJeztah:
follow-up to https://github.com/docker/cli/pull/3310

The COMPANY_NAME currently sets the "CompanyName" field in the metadata
of Windows binaries. Our intent of this field is this field to contain
information about the company/party that produced the binary.

Also from [FileVersionInfo.CompanyName][FileVersionInfo.CompanyName]:

> Gets the name of the company that produced the file

Based on the above, "PACKAGER_NAME" is a bit more generic, and clearer
on intent, and we may at some point re-use this same information to
propagate equivalent fields on other platforms (rpms, debs)

[FileVersionInfo.CompanyName]: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.fileversioninfo.companyname

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,8 +52,8 @@ ARG GO_STRIP
 ARG CGO_ENABLED
 # VERSION sets the version for the produced binary
 ARG VERSION
-# COMPANY_NAME sets the company that produced the windows binary
-ARG COMPANY_NAME
+# PACKAGER_NAME sets the company that produced the windows binary
+ARG PACKAGER_NAME
 COPY --from=goversioninfo /out/goversioninfo /usr/bin/goversioninfo
 RUN --mount=type=bind,target=.,ro \
     --mount=type=cache,target=/root/.cache \
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 #
 
 # Sets the name of the company that produced the windows binary.
-COMPANY_NAME ?=
+PACKAGER_NAME ?=
 
 all: binary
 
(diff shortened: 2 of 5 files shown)


thaJeztah:
We should also have a look at the `LegalTrademarks` field; looks like we currently leave it empty, but perhaps would be good to (at some point) also set it.

It's a bit ambiguous though what should be in there (initially I thought that should contain something like  "docker, the docker logo, and... are trademarks of docker. inc.", but the description from https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.fileversioninfo.legaltrademarks?view=net-6.0 describes it as:

> The legal trademarks include the full text of all notices, legal symbols, and trademark numbers.

Which would potentially mean _all_ `NOTICE` files of source we use, so we need to check what should be in that field.

thaJeztah:
Alpine having an outage?

```
#26 [linux/amd64 build-base-alpine 3/4] RUN apk add --no-cache bash clang lld llvm file git
#26 130.1 WARNING: Ignoring https://dl-cdn.alpinelinux.org/alpine/v3.15/main: network error (check Internet connection and firewall)
#26 130.1 fetch https://dl-cdn.alpinelinux.org/alpine/v3.15/community/x86_64/APKINDEX.tar.gz
#26 261.2 WARNING: Ignoring https://dl-cdn.alpinelinux.org/alpine/v3.15/community: network error (check Internet connection and firewall)
#26 261.2 ERROR: unable to select packages:
#26 261.2   bash (no such package):
#26 261.2     required by: world[bash]
```

crazy-max (approved):

PR #3501, opened 2022-03-27
update go to 1.18.0

thaJeztah:
- [x] depends on https://github.com/docker/cli/pull/3502

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3502, opened 2022-03-27
lint: update golangci-lint to v1.45.2

thaJeztah:
- relates to https://github.com/docker/cli/pull/3501

### DIFF
--- a/cli/command/container/formatter_stats.go
+++ b/cli/command/container/formatter_stats.go
@@ -181,14 +181,14 @@ func (c *statsContext) ID() string {
 
 func (c *statsContext) CPUPerc() string {
 	if c.s.IsInvalid {
-		return fmt.Sprintf("--")
+		return "--"
 	}
 	return fmt.Sprintf("%.2f%%", c.s.CPUPercentage)
 }
 
 func (c *statsContext) MemUsage() string {
 	if c.s.IsInvalid {
-		return fmt.Sprintf("-- / --")
+		return "-- / --"
 	}
 	if c.os == winOSType {
 		return units.BytesSize(c.s.Memory)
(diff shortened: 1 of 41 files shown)


cli/command/formatter/formatter.go
```diff
@@ -73,7 +73,7 @@ func (c *Context) preFormat() {
 func (c *Context) parseFormat() (*template.Template, error) {
 	tmpl, err := templates.Parse(c.finalFormat)
 	if err != nil {
-		return tmpl, errors.Errorf("Template parsing error: %v\n", err)
+		return tmpl, errors.Wrap(err, "template parsing error")
```
thaJeztah, line 76: FWIW; although the error changed (removing the newline at the end); when printed on the command-line, there's still a newline printed, so not sure why this was added in the first place 🤷‍♂️

thaJeztah:
whoop, green now; @silvin-lubecki @crazy-max PTAL

cli/command/container/formatter_stats.go
```diff
@@ -181,45 +181,45 @@ func (c *statsContext) ID() string {
...
 
 func (c *statsContext) MemUsage() string {
 	if c.s.IsInvalid {
-		return fmt.Sprintf("-- / --")
+		return "-- / --"
 	}
 	if c.os == winOSType {
 		return units.BytesSize(c.s.Memory)
 	}
-	return fmt.Sprintf("%s / %s", units.BytesSize(c.s.Memory), units.BytesSize(c.s.MemoryLimit))
```
silvin-lubecki, line 196: Nit: I'm not really enthousiast with this modification, even if I understand why. I think the sprintf version os easier to read and review. Anyway it's just nitpicking here.

silvin-lubecki (approved):

cli/command/container/formatter_stats.go
thaJeztah: Let me do those separate, as they're not strictly needed for the linter, but given that this stats output is called in a loop (to refresh stats), I thought it would be worth to change them. Looking at the file, I now see some other ones that could be adjusted in that respect, so let me change these back and do it separately.

thaJeztah: oh! of course it helps if I ping you after that; updated 👍 PTAL

silvin-lubecki: Still LGTM 🤗

PR #3471, opened 2022-03-15
search: remove client-side default for --limit

thaJeztah:
- [x] depends on https://github.com/moby/moby/pull/43380
- [x] depends on https://github.com/docker/cli/pull/3486

The daemon (and registry) already have a default limit. This patch
removes the default from the client side, to not duplicate setting
these defaults.

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3504, opened 2022-03-28
Fix flaky TestContainerList tests

thaJeztah:
relates to https://github.com/docker/cli/pull/3471#issuecomment-1080989454

These tests were creating a stub container, using the current timestamp as
created date. However, if CI was slow to run the test, `Less than a second ago`
would change into `1 second ago`, causing the test to fail:

    --- FAIL: TestContainerListNoTrunc (0.00s)
        list_test.go:198: assertion failed:
            --- expected
            +++ actual
            @@ -1,4 +1,4 @@
            -CONTAINER ID   IMAGE            COMMAND   CREATED                  STATUS        PORTS     NAMES
            -container_id   busybox:latest   "top"     Less than a second ago   Up 1 second             c1
            -container_id   busybox:latest   "top"     Less than a second ago   Up 1 second             c2,foo/bar
            +CONTAINER ID   IMAGE            COMMAND   CREATED        STATUS        PORTS     NAMES
            +container_id   busybox:latest   "top"     1 second ago   Up 1 second             c1
            +container_id   busybox:latest   "top"     1 second ago   Up 1 second             c2,foo/bar

This patch changes the "created" time of the container to be a minute ago. This
will result in `About a minute ago`, with a margin of 1 minute.

### DIFF
--- a/cli/command/container/testdata/container-list-without-format-no-trunc.golden
+++ b/cli/command/container/testdata/container-list-without-format-no-trunc.golden
@@ -1,3 +1,3 @@
-CONTAINER ID   IMAGE            COMMAND   CREATED                  STATUS        PORTS     NAMES
-container_id   busybox:latest   "top"     Less than a second ago   Up 1 second             c1
-container_id   busybox:latest   "top"     Less than a second ago   Up 1 second             c2,foo/bar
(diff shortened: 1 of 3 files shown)


thaJeztah:
@ndeloof @rumpl PTAL

PR #3507, opened 2022-03-29
Dockerfile: update buildx to v0.8.1

thaJeztah:
This only affects the version used in the dev container, but I noticed it was an older version of buildx.


thaJeztah:
@crazy-max @ndeloof PTAL

PR #3508, opened 2022-03-30
Tweak description of some commands

thaJeztah:
Changed the description of some commands to (hopefully) be a bit clearer / more descriptive


cli/command/image/pull.go
```diff
@@ -28,7 +28,7 @@ func NewPullCommand(dockerCli command.Cli) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "pull [OPTIONS] NAME[:TAG|@DIGEST]",
-		Short: "Pull an image or a repository from a registry",
+		Short: "Download an image from an image registry into the daemon's image store",
```
thaJeztah, line 31: These ones (pull/push) were a bit tricky;

- trying to keep the description brief, so that it fits the terminal width and doesn't wrap in the `--help` output
- initially I had "intro the local image store", but that may be confusing, as it's "local to the daemon", which may not be "your local machine"
- at the same time, for most users, `daemon` should be mostly an implementation detail (albeit _important_ for some situations)
- "image registry" could be just "registry" (happy to hear if we think "registry" is sufficient) (we can't use "Docker Hub" here, because that just _one_ possible registry)

usha-mandya: How about "Download an image from a registry"? Do we need to say where it's stored? For a user, the pull command just downloads the image. I'm not sure if they need to know the details on where it's stored. I think it's more of an implementation detail?

cli/command/image/push.go
```diff
@@ -29,7 +29,7 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "push [OPTIONS] NAME[:TAG]",
-		Short: "Push an image or a repository to a registry",
+		Short: "Upload an image from the daemon's image store to an image registry",
```
usha-mandya, line 32: We could also simplify this and say "Upload an image to a registry"

cli/command/image/pull.go
thaJeztah: Yes, perhaps we should keep it simple; let me update 👍

usha-mandya (approved):
LGTM. Thank you @thaJeztah

PR #3513, opened 2022-03-31
use GO_LDFLAGS instead of LDFLAGS to prevent inheriting unrelated options

thaJeztah:
relates to https://github.com/docker/docker-ce-packaging/pull/666

When building on Fedora 36, the build failed. I suspect this is because the
rpm tools also set LDFLAGS, but with options that cannot be used;

    GO_LINKMODE=dynamic
    + ./scripts/build/binary
      /go/src/github.com/docker/cli ~/rpmbuild/BUILD/src
      Building dynamic docker-linux-arm64
      + go build -o build/docker-linux-arm64 -tags ' pkcs11' -ldflags '-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -Wl,--build-id=sha1 -Wl,-dT,/root/rpmbuild/BUILD/src/.package_note-docker-ce-cli-0.0.0.20220330082637.68cad50-0.fc36.aarch64.ld -w -X "github.com/docker/cli/cli/version.GitCommit=68cad50" -X "github.com/docker/cli/cli/version.BuildTime=2022-03-30T20:05:36Z" -X "github.com/docker/cli/cli/version.Version=0.0.0-20220330082637-68cad50" -X "github.com/docker/cli/cli/version.PlatformName=Docker Engine - Community"' -buildmode=pie github.com/docker/cli/cmd/docker
    # github.com/docker/cli/cmd/docker
    flag provided but not defined: -Wl,-z,relro
    usage: link [options] main.o

This patch changes the variable we use to `GO_LDFLAGS`, taking a similar approach
as containerd, and various other projects using this name: https://grep.app/search?q=GO_LDFLAGS

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/build/.variables
+++ b/scripts/build/.variables
@@ -63,23 +63,23 @@ if [ "$CGO_ENABLED" = "1" ] && [ "$(go env GOOS)" != "windows" ]; then
 fi
 export GO_BUILDMODE
 
-LDFLAGS="${LDFLAGS:-} -w"
-LDFLAGS="$LDFLAGS -X \"github.com/docker/cli/cli/version.GitCommit=${GITCOMMIT}\""
-LDFLAGS="$LDFLAGS -X \"github.com/docker/cli/cli/version.BuildTime=${BUILDTIME}\""
-LDFLAGS="$LDFLAGS -X \"github.com/docker/cli/cli/version.Version=${VERSION}\""
+GO_LDFLAGS="${GO_LDFLAGS:-} -w"
+GO_LDFLAGS="$GO_LDFLAGS -X \"github.com/docker/cli/cli/version.GitCommit=${GITCOMMIT}\""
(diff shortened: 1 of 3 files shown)


crazy-max (approved):
Indeed we call [`./scripts/build/binary` directly on docker-ce-packaging](https://github.com/docker/docker-ce-packaging/blob/b742b3a0810e6862193bd7967a850cf14e34118a/rpm/SPECS/docker-ce-cli.spec#L74) so we are not sandboxed unfortunately.

thaJeztah:
Right, so I _think_ moby/moby _resets_ `LDFLAGS` at the start (would have to double check), so that would avoid the problem as well, but I was thinking that "have a `GO_LDFLAGS` allows people to (if needed for anything) set an initial set of flags (before our scripts add the options we set), which could be useful.

PR #3531, opened 2022-04-04
tweak description of login/logout

thaJeztah:
follow-up to https://github.com/docker/cli/pull/3508

Remove "Docker" from registry, as the registry specification is no
longer docker-specific, but part of the OCI distribution spec.

Also removed "Register" from one of the docs pages, as the login
command hasn't supported creating a new acccount on Docker Hub for
a long time.

I'm wondering if we should be more explicit about what log in / out
does (effectively; authenticate, and on success store the credentials
or token, and on log out; remove credentials/token).

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -80,6 +80,6 @@ End-to-end test should run the `docker` binary using
 and make assertions about the exit code, stdout, stderr, and local file system.
 
 Any Docker image or registry operations should use `registry:5000/<image name>`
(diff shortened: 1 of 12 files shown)


thaJeztah:
@usha-mandya PTAL

usha-mandya (approved):
LGTM. Thank you @thaJeztah

thaJeztah:
Thanks! let me get this one in

Issue #3441, opened 2022-02-24
Flaky test: TestInitTtySizeErrors() (on macOS "latest")


thaJeztah:
Looks like this test is flaky on macOS ("latest"); could be due to macOS updates perhaps (but the "resize" code has been known to be somewhat racy/flaky, so perhaps some improvements are needed)

```
--- FAIL: TestInitTtySizeErrors (0.10s)
    tty_test.go:29: assertion failed: 
        --- expectedError
        +++ →
        @@ -1,2 +1 @@
        -failed to resize tty, using default size
```

thaJeztah:
Test flakiness was fixed in https://github.com/docker/cli/pull/3573 and https://github.com/docker/cli/pull/3584

There's some remaining work to be done, which is tracked through https://github.com/docker/cli/issues/3554

PR #3536, opened 2022-04-06
Clear go.mod after vendor scripts ran

benja-M-1:
**- What I did**

This PR adds an option to remove the go.mod stub when executing a `./script/vendor` command. It avoids `go build` to fail because of unconsistencies between go.mod and vendor/modules.txt files.

This PR also
- fixes the `./scripts/vendor` to display the usage command when no command is passed
- adds the `clean` command called after yaml and man docs have been generated

**- How I did it**

When the following commands are ran
```bash
make vendor
make binary
```

`go build` fails with the message
```bash
go build -o build/docker-linux-arm64 -tags  -ldflags ' -w -X "github.com/docker/cli/cli/version.GitCommit=b82d9a7742" -X "github.com/docker/cli/cli/version.BuildTime=2022-04-06T11:45:13Z" -X "github.com/docker/cli/cli/version.Version=20.10.0-dev"' github.com/docker/cli/cmd/docker
go: inconsistent vendoring in /go/src/github.com/docker/cli:
	github.com/Azure/go-ansiterm@v0.0.0-20210617225240-d185dfc1b5a1: is marked as explicit in vendor/modules.txt, but not explicitly required in go.mod
...
```

**- How to verify it**

Clean up `go.mod` file after vendors have been updated, validated or list outdated.

### DIFF
--- a/scripts/vendor
+++ b/scripts/vendor
@@ -13,6 +13,10 @@ if [ -z "$TYP" ]; then
   usage
 fi
 
+clean() {
+  rm go.mod
+}
+
 init() {
   # create dummy go.mod, see comment in vendor.mod
   cat > go.mod <<EOL
@@ -43,6 +47,7 @@ outdated() {
   (set -x ; go list -mod=vendor -mod=readonly -modfile=vendor.mod -u -m -json all | go-mod-outdated -update -direct)
 }
 
+
 case $TYP in
   "init")
     init
@@ -50,15 +55,18 @@ case $TYP in
   "update")
     init
     update
+    clean
     ;;
   "validate")
     init
     update
+    clean
     validate
     ;;
   "outdated")
     init
     outdated
+    clean
     ;;
   *)
     echo >&2 "Unknown type $TYP"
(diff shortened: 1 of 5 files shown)


benja-M-1:
@thaJeztah I finally followed up on this WIP PR/ Somehow the `test/host (macos-latest)` check fails but I fail to understand why. If ever you can help me on that I could mark this PR as reviewable! :)

thaJeztah:
> Somehow the test/host (macos-latest) check fails but I fail to understand why. 

Ah, you can ignore that one; we have a flaky test; see https://github.com/docker/cli/issues/3441 and https://github.com/docker/cli/issues/3554 for details.

scripts/vendor
```diff
@@ -60,8 +68,21 @@ case $TYP in
...
+
+while getopts ":c" opt; do
+  case $opt in
+    c)
+      clean
+      ;;
+    *)
+      ;;
+  esac
+done
```
thaJeztah, line 88: Perhaps we should reverse the option; by default, remove the `go.mod`, and only if it's needed after it ran, pass an option to `--keep` (or whatever) the file.

Looking at the changes, I think all (but one) calls of this script expect the `go.mod` to be deleted afterwards; WDYT?

PR #3540, opened 2022-04-06
fix: remove asterisk from docker command suggestions

mcornella:
**- What I did**

Some commands in the output of `docker` show up with an asterisk, like `app`, `build`, `buildx` or `scan`.

image

This tweak removes them so that the asterisk is not filled in when choosing those commands.

**- How I did it**

In `__docker_commands`, the output of `docker` is parsed to get the list of valid commands. The function already trims whitespace from left and right. I just added a zsh-regex pattern for an asterisk on the right-hand side to remove it as well.

The zsh regex `\*#` is equivalent to POSIX regex `\**` (0 or more occurrences of `*`) (note that this requires `setopt extendedglob`).

**- How to verify it**

1. Clone this PR and `cd` into the project root.
2. Run `unfunction _docker` if docker completion has been attempted before.
3. Run `autoload -Uz $PWD/contrib/completion/zsh/_docker`.
4. Attempt to complete `docker `.

Suggestion entries should appear without a trailing asterisk:

image

Remove trailing asterisk from zsh completion for docker commands.

kitten at computer

### DIFF
--- a/contrib/completion/zsh/_docker
+++ b/contrib/completion/zsh/_docker
@@ -2653,7 +2653,7 @@ __docker_commands() {
     then
         local -a lines
         lines=(${(f)"$(_call_program commands docker 2>&1)"})
-        _docker_subcommands=(${${${(M)${lines[$((${lines[(i)*Commands:]} + 1)),-1]}:# *}## #}/ ##/:})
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM, thanks!

mcornella:
Thanks!

PR #3542, opened 2022-04-07
update status of deprecated features for 22.04 (docs, and some code)

thaJeztah:
Mostly docs, but some code-changes

### docs: reformat table for compatibility

My IDE's linter kept complaining:

> For compatibility reasons all table rows should have borders (pipe
> symbols) at the start and at the end.

### update deprecation status of --kernel-memory(-tcp)

Update the flag annotations and description

### docs: update deprecation status of "experimental" cli

Change future tense to current tense.

### cli/command/system: remove printStorageDriverWarnings()

These warnings were for features that are no longer supported (overlay
on a backingFS without `d_type` support), or related to the deprecated
devicemapper storage driver.

Removing this function for that reason.

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3543, opened 2022-04-07
remove "Experimental" client field from "docker version"

thaJeztah:
- [x] follow-up to https://github.com/docker/cli/pull/3542

Configuration (enabling/disabling) of Experimental client features was deprecated in Docker 19.03, and removed in 20.10. Experimental CLI features are now always enabled. In Docker 20.10, the Experimental field in `docker version` was kept (but always true).

This patch removes the field from the output (both "pretty" output and the JSON struct).


rumpl (approved):
👍

PR #3555, opened 2022-04-12
cli/command/image: remove use of docker/docker/pkg/urlutil

thaJeztah:
- relates to https://github.com/moby/moby/pull/43477

pkg/urlutil (despite its poorly chosen name) is not really intended as a
generic utility to handle URLs, and should only be used by the builder to
handle (remote) build contexts.

The `IsURL()` function only does a very rudimentary check for `http(s)://`
prefixes, without any other validation, but due to its name may give
incorrect expectations.

As we're deprecating this package for uses other than for build-contexts,
this patch replaces this instance of the utility for a local function.

### DIFF
--- a/cli/command/image/import.go
+++ b/cli/command/image/import.go
@@ -2,15 +2,14 @@ package image
 
 import (
 	"context"
-	"io"
 	"os"
+	"strings"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	dockeropts "github.com/docker/cli/opts"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/jsonmessage"
-	"github.com/docker/docker/pkg/urlutil"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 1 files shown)


cli/command/image/import.go
```diff
@@ -88,3 +88,7 @@ func runImport(dockerCli command.Cli, options importOptions) error {
 
 	return jsonmessage.DisplayJSONMessagesToStream(responseBody, dockerCli.Out(), nil)
 }
+
+func isURL(str string) bool {
+	return strings.HasPrefix(str, "https://") || strings.HasPrefix(str, "http://")
```
ndeloof, line 93: why not use https://github.com/moby/moby/blob/master/builder/remotecontext/urlutil/urlutil.go#L20 (as recommended by deprecation notice)? This introduces code duplication
Also method name sounds weird to me, as there's more than just http to be actual URLs

thaJeztah: I moved it to `builder/remotecontext` as that utility should really only be used for that purpose, and not as a generic utility to check "if it's a URL", as `docker build` may have different "rules" than other users.

thaJeztah: I removed the local `isURL()` function (to not give the illusion we're validating it's an actual URL), and changed the logic to use a switch, to make it slightly more clearer what the criteria are for each.

ndeloof (approved):

thaJeztah:
Thx! Let me get this one in

Issue #3425, opened 2022-02-13
Command: docker context show does not exist for Linux client


mattKaczorowski:
**Description**

When typing `docker context`, the help menu shows the `context show` command is missing.

```
$ docker context

Usage:  docker context COMMAND

Manage contexts

Commands:
  create      Create a context
  export      Export a context to a tar or kubeconfig file
  import      Import a context from a tar or zip file
  inspect     Display detailed information on one or more contexts
  ls          List contexts
  rm          Remove one or more contexts
  update      Update a context
  use         Set the current docker context

Run 'docker context COMMAND --help' for more information on a command.

```

**Steps to reproduce the issue:**
1. Download docker for a linux machine (in my case Ubuntu)
2. Run: docker context show

**Describe the results you received:**
Docker client shows the help screen for `docker context`

**Describe the results you expected:**
I expect to see the output describing the name of my current context: `default`
I also expect the command to exist.

**Additional information you deem important (e.g. issue happens only occasionally):**
I have the same client version installed on an M1 mac and the expected command is there.

**Output of `docker version`:**

Client Version Info:
```
$ docker version
Client: Docker Engine - Community
 Version:           20.10.12
 API version:       1.41
 Go version:        go1.16.12
 Git commit:        e91ed57
 Built:             Mon Dec 13 11:45:33 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.12
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.12
  Git commit:       459d0df
  Built:            Mon Dec 13 11:43:42 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.12
  GitCommit:        7b11cfaabd73bb80907dd23182b9347b4245eb5d
 runc:
  Version:          1.0.2
  GitCommit:        v1.0.2-0-g52b36a2
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

```

**Output of `docker info`:**

```
$ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.7.1-docker)
  scan: Docker Scan (Docker Inc., v0.12.0)

Server:
 Containers: 7
  Running: 0
  Paused: 0
  Stopped: 7
 Images: 430
[...]

PR #3567, opened 2022-04-25
introduce `context show` command

ndeloof:
closes [#xxxx"](https://github.com/docker/cli/issues/3425)

**- What I did**
introduce `context show` command, aligned with same command introduce in compose-cli

**- How I did it**
new cobra command, dump current context name after checking it exists

**- How to verify it**
`docker context show`

introduce `docker context show` command

image


rumpl (approved):

cli/command/context/show_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+func TestShow(t *testing.T) {
+	cli := makeFakeCli(t)
+	createTestContext(t, cli, "current")
+	cli.SetCurrentContext("current")
+
+	cli.OutBuffer().Reset()
+	assert.NilError(t, runShow(cli))
+	expected := string(golden.Get(t, "show.golden"))
+	assert.Equal(t, cli.OutBuffer().String(), expected)
```
thaJeztah, line 18: In other places we use this (I _think_ using that approach also allows re-generating the golden files);

```suggestion
	golden.Assert(t, cli.OutBuffer().String(), "show.golden")
```

That said, I wonder if using a `golden` file for this is a bit too much, we can probably just check for a string;

```suggestion
	assert.Equal(t, cli.OutBuffer().String(), "current")
```

thaJeztah: I tried some variations; do you think we should also have some tests for when an override is used? Or are such cases already covered enough?)


```bash
$ docker context show
default

$ docker context create current
current
Successfully created context "current"

$ docker context use current
current
Current context is now "current"

$ docker context show
current

$ DOCKER_CONTEXT=default docker context show
default

$ docker --context=default context show
default
```

thaJeztah (changes requested):
Left some comments.

Now that we have the `docker context show` command, perhaps we should

- add a `docs/reference/context_show.md` file
- move the examples from https://github.com/docker/cli/pull/2500/files#diff-3589954942dcb3aa5aa11ca30bee4e904ebd9c35d1d9dfa73192f1655cdc5250R81 
- (or keep the heading that's there, and link to the `context show` reference)

cli/command/context/show_test.go
ndeloof: testing DOCKER_CONTEXT / --context would actually test `dockerCli.CurrentContext()`, not the `show` command, and would require an actual e2e test, not a unit test relying on `fakecli`

ndeloof: use of a golden file here is bit over-engineering, but I prefer this is aligned with other tests checking console output + it makes it easier to manage empty line at console end (otherwise we need `assert.Equal(t, out, "current\n")` which is not nice to read imho)

thaJeztah (approved):
LGTM, thanks!

PR #3470, opened 2022-03-15
search: do not sort results by stars on client side

thaJeztah:
Once upon a time, there was a website named ["The Docker index"][2]; a complimentary
service for users of Docker, provided by dotCloud. The Docker Index was the place
to find and explore pre-made container images, and allowed you to [share your
images and download them][1]. The Docker Index evolved rapidly, and gained new
features, such as [Trusted Images][3], and "stars" to rank your favorite images.

The website also provided an API, which allowed you to search images, even from
the comfort of your `docker` CLI. Things moved fast in container-land, and while
there was an API to use, it was still a work in progress. While the Docker Index
allowed users to "star" images, the search API did not rank results accordingly.

As any engineer knows, there's no problem that can't be solved with some elbow-
grease and a piece of Duct tape, so while the Docker Index team worked on making
the search API better, the `docker` engine [fixed the problem on the client side][4]

Years went by, and the Docker Index API became the "registry V1" specification,
including search. The registry got a major "v2" rewrite and became the [OCI Distribution
Spec][5], and Docker Index became Docker Hub, which included V2 and V3 search APIs.
The V1 search API continued to be supported, as it was the only documented API
for registries, but improvements were made, including ranking of search results.

Duct tape is durable, and even though improvements were made, the Docker client
continued to sort the results as well. Unfortunately, this meant that search
results on the command-line were ranked different from their equivalent on the
registry (such as Docker Hub).

This patch removes the client-side sorting of results, using the order in which
the search API returned them to (finally) celebrate the work of the engineers
working on the search API, also when used from the command-line.

[1]: https://web.archive.org/web/20130708004229/http://docker.io/
[2]: https://web.archive.org/web/20130623223614/https://index.docker.io/
[3]: https://web.archive.org/web/20140208001647/https://index.docker.io/
[4]: https://github.com/docker/docker/commit/1669b802cc3e7ce30f05e61630542c88696bbca1
[5]: https://github.com/opencontainers/distribution-spec

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

image

PR #3568, opened 2022-04-25
use our own version of text/tabwriter

ndeloof:
closes #https://github.com/docker/cli/issues/3421
supersede https://github.com/docker/cli/pull/3468

**- What I did**
add support for tab output with non-Latin characters

**- How I did it**
use a tabwriter for and relie on mattn/go-runewidth to compute columns width

**- How to verify it**
```
$ build/docker search twang2218
NAME                       DESCRIPTION                                     STARS     OFFICIAL   AUTOMATED
twang2218/gitlab-ce-zh     汉化的 GitLab 社区版 Docker Image               255                  [OK]
twang2218/mysql            MySQL Replication Master/Slave cluster examp…   1                    
twang2218/spark            Apache Spark Docker Image                       0                    [OK]
twang2218/lnmp-nginx       这是 LNMP 示例中的 nginx 镜像                   23                   [OK]

```
docker tab-aligned output supports non-Latin characters

image


rumpl (approved):

thaJeztah:
LOL. wanted to post before/after, but looks like GitHub doesn't render it correctly. So with screenshots;

Before:

Screenshot 2022-04-25 at 14 49 14

After:

Screenshot 2022-04-25 at 14 49 22


Looks like I should check with the Docker Hub team on the sorting BTW; I removed client-side sorting in https://github.com/docker/cli/pull/3470 (as the result from hub should be in the correct order), but looking at that output, I'm not sure what sorting is used there (at least it's not sorted by `stars` ⭐ ⭐ ⭐ )

thaJeztah:
> Looks like I should check with the Docker Hub team on the sorting BTW; I removed client-side sorting in https://github.com/docker/cli/pull/3470 (as the result from hub should be in the correct order), but looking at that output, I'm not sure what sorting is used there (at least it's not sorted by stars ⭐ ⭐ ⭐ )

So apparently, it's sorted by `pull_count`, which is indeed returned by the search response, but not shown. Will need to check if that was part of the original v1 search API, or a later addition. If it was already present, we can consider to present pull count instead of stars (e.g.), or show both

Issue #3554, opened 2022-04-12
investigate terminal resize issues, and (if possible) set size on creation


thaJeztah:
relates to:

- https://github.com/docker/cli/issues/3231
- https://github.com/docker/cli/issues/3441
- https://github.com/moby/moby/issues/42707

The `TestInitTtySizeErrors` test is failing frequently (see https://github.com/docker/cli/issues/3441), and it looks like it's not just a flaky test, but a bug / regression in the CLI:

```
--- FAIL: TestInitTtySizeErrors (0.16s)
    tty_test.go:29: assertion failed: 
        --- expectedError
        +++ →
        @@ -1,2 +1 @@
        -failed to resize tty, using default size
```

We had issues in this area before;

- https://github.com/moby/moby/issues/35407
- https://github.com/docker/cli/issues/1492
- https://github.com/docker/cli/pull/1529
- https://github.com/moby/moby/issues/33794
- various other issues

Previous "fixes" implemented a retry loop to try resizing the terminal if it didn't work the first time (there's some race between the container being created and it being receptacle for resizing), so the existing code continues to be a "best effort"; https://github.com/docker/cli/blob/b68db383d389ae7c0302aaa3dc33f3155b6fe9a3/cli/command/container/run.go#L189-L193

### Reproduction steps

The issue can be reproduced with this script (which just starts a lot of containers in a loop); I haven't counted the success/failure ratio, but it's failing _very_ frequently:

```bash
#!/bin/bash

A=0
while true; do
    for a in $(seq 1 15); do
        com.docker.cli run --rm -t alpine echo $A/$a &
    done

    wait
    date
    A=$((A+1))
    echo $A
done
```

Coincidentally, when using a Windows machine, the CLI sets the initial size when _creating_ the container. 

https://github.com/docker/cli/blob/b68db383d389ae7c0302aaa3dc33f3155b6fe9a3/cli/command/container/run.go#L122-L127

The code above looks wrong though; 
 
- it checks for the operating system on the CLI side, but does not take into account that the _daemon_ could be either a Linux or a Windows daemon
- which could be either an "omission" (initial size supported on both Linux and Windows daemons)
- or (if Linux daemons do not support this) result in Linux container shells never getting the right size when started from a Windows CLI.
- or "redundant" (initial size already connect, but after this, doing a resize to the same size again)

### What's needed?

[...]

PR #3572, opened 2022-04-28
Always set the console size

rumpl:
**- What I did**

Removed the check for the OS when setting the console size on container creation. This check doesn't really make sense because the client doesn't know on what OS the daemon is really running. The daemon uses the [console size on creation when available](https://github.com/moby/moby/blob/master/daemon/oci_windows.go#L127-L132) (on windows).

**- How I did it**
Checked the daemon code that it uses the console size only when available.

**- How to verify it**
`docker run ...` should still work

image


ndeloof (approved):
nice catch.

thaJeztah:
relates to https://github.com/docker/cli/issues/3554 (see the top description)

This change looks correct indeed (at least from the client side); what I noticed though is that the code you linked to is "on windows only"; https://github.com/moby/moby/blob/master/daemon/oci_windows.go#L127-L132


I see the corresponding implementation for linux does not take these values into account; https://github.com/moby/moby/blob/cb07afa3cc1cbc7adfab3ac65d92c497a3271b50/daemon/oci_linux.go#L1001

A quick look at the runtime-spec shows that the `ConsoleSize` field is not marked as platform-specific (i.e., it's available on both Windows and Linux; https://github.com/opencontainers/runtime-spec/blob/2cf6663ca299f8454318ea7c797a8039d013fe8c/specs-go/config.go#L35-L39

And the description of it also doesn't mention it to be "Windows only"; https://github.com/opencontainers/runtime-spec/blob/a3c33d663ebc56c4d35dbceaa447c7bf37f6fab3/config.md#process


So perhaps we should (as a follow-up);

1. check if the option is "punched through" all the way in containerd and runc (verify if it takes effect if set)
2. add a `WithConsoleSize()` option (to be used in the Linux implementation at least) in moby/moby
3. after the above, we could make the "retry resize" loop "optional", based on API version (if the CLI connects with a daemon that has  `2.` above, then skip the resize retry code

It may be worth adding a comment to the lines changed here, that they currently are ignored on Linux daemons

cli/command/container/run.go
```diff
@@ -119,12 +118,7 @@ func runContainer(dockerCli command.Cli, opts *runOptions, copts *containerOptio
 		config.StdinOnce = false
 	}
 
-	// Telling the Windows daemon the initial size of the tty during start makes
-	// a far better user experience rather than relying on subsequent resizes
-	// to cause things to catch up.
-	if runtime.GOOS == "windows" {
-		hostConfig.ConsoleSize[0], hostConfig.ConsoleSize[1] = dockerCli.Out().GetTtySize()
-	}
+	hostConfig.ConsoleSize[0], hostConfig.ConsoleSize[1] = dockerCli.Out().GetTtySize()
```
thaJeztah, line 121: See my comment above; perhaps we should (until fixed) add a comment here that this option is currently ignored on Linux daemons

rumpl: Added a comment

thaJeztah (approved):
LGTM

PR #3573, opened 2022-04-28
Retry initializing TTY size a bit more

rumpl:
**- What I did**

I some cases, for example if there is a heavy load, the initialization of the TTY size would fail. This change makes the cli retry more times, 10 instead of 5 and we wait incrementally from 10ms to 100ms between two calls to resize the TTY.

Running 150 containers (see script below) takes ~2 minutes to complete without this change, with this change the time to complete was ~1m55s and 2m12s so I don't think this change will impact users.

Running one container `docker run --rm -t hello-world` takes the same amount of time (1.5s on my machine) with or without this change.

Relates to #3554

**- How I did it**
Changed the retry logic to retry 10 times instead of 5 and increase the wait between retries to 100ms

**- How to verify it**

Running this script shouldn't show any `failed to resize tty, using default size` errors.

```bash
#!/bin/bash

for A in $(seq 1 10); do
    echo $A

    for a in $(seq 1 15); do
        com.docker.cli run --rm -t alpine echo $A/$a &
    done

    wait
    date
done
```

### DIFF
--- a/cli/command/container/tty.go
+++ b/cli/command/container/tty.go
@@ -45,7 +45,7 @@ func resizeTty(ctx context.Context, cli command.Cli, id string, isExec bool) err
 	return resizeTtyTo(ctx, cli.Client(), id, height, width, isExec)
 }
 
-// initTtySize is to init the tty's size to the same as the window, if there is an error, it will retry 5 times.
+// initTtySize is to init the tty's size to the same as the window, if there is an error, it will retry 10 times.
 func initTtySize(ctx context.Context, cli command.Cli, id string, isExec bool, resizeTtyFunc func(ctx context.Context, cli command.Cli, id string, isExec bool) error) {
 	rttyFunc := resizeTtyFunc
 	if rttyFunc == nil {
@@ -54,8 +54,8 @@ func initTtySize(ctx context.Context, cli command.Cli, id string, isExec bool, r
 	if err := rttyFunc(ctx, cli, id, isExec); err != nil {
 		go func() {
 			var err error
-			for retry := 0; retry < 5; retry++ {
-				time.Sleep(10 * time.Millisecond)
+			for retry := 0; retry < 10; retry++ {
+				time.Sleep(time.Duration(retry+1) * 10 * time.Millisecond)
 				if err = rttyFunc(ctx, cli, id, isExec); err == nil {
 					break
 				}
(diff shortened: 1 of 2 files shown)


thaJeztah:
Let me comment here, as we were discussing this;

- First of all, this loop is for the _initial_ resize (docker run -> resize to desired dimensions), and not used for the "monitor for terminal to be resized" case
- The retry loop is only hit after we tried a resize immediately (line 54 does a resize _immediately_ after attaching, and we only hit the _retry_ loop if that fails)
- so: incrementing this sleep does not delay the initial resize (which is good)

Thinking about this a bit more;

1. for a "daemon under load", it may be expected that we need slightly longer (or more retries)
2. if the resize is taking longer for _other_ (daemon not under load) situations, it may be that there's a regression in performance
3. so even if we fix the initial size (to be defined when creating the container, and without a "resize" after), the "monitor for resize" would still need to trigger resizes
4. so: if there's a regression in performance, that "monitor" case would still be affected by the regression (if any), so should still be looked into. (at least, we can keep a tracking ticket to do so)

cli/command/container/tty.go
```diff
@@ -54,8 +54,8 @@ func initTtySize(ctx context.Context, cli command.Cli, id string, isExec bool, r
 	if err := rttyFunc(ctx, cli, id, isExec); err != nil {
 		go func() {
 			var err error
-			for retry := 0; retry < 5; retry++ {
-				time.Sleep(10 * time.Millisecond)
+			for retry := 0; retry < 10; retry++ {
+				time.Sleep(100 * time.Millisecond)
```
thaJeztah, line 58: Wondering if we should consider making the delay incremental;

- retry after 10 milliseconds ("happy path" for when the daemon wasn't able to handle it immediately)
- if failed; add 10 milliseconds (so, retry after 20 milliseconds)
- if failed; add 10 milliseconds (so, retry after 30)
- and so on, until we reach 10 attempts (which would be using 100 milliseconds) 

If incrementing by 10 is not "aggressive" enough, perhaps we could make it exponential (but with a "cap", which could be 100ms);

10 -> 20 -> 40 -> 80 -> 100 -> 100 -> 100


Doing so would allow the resize to be handled "faster" (in the happy case), while still _allowing_ it to take longer (if the daemon happens to be under load)

WDYT?

rumpl: Done!

thaJeztah (approved):
LGTM, thanks!

thaJeztah:
Hmm.... CI fails on the related test, so wondering if incrementing with `10` is not aggressive enough after all 😞 

```
--- FAIL: TestInitTtySizeErrors (0.65s)
    tty_test.go:29: assertion failed: 
        --- expectedError
        +++ →
        @@ -1,2 +1 @@
        -failed to resize tty, using default size
```

rumpl:
> 

It's the other way around, the test is about whether we return that error if we can't resize, to the trick is to sleep enough time for the loop to finish and the function to return and error. It should be good now

thaJeztah:
Ah, whoop!

Let's get this one merged 👍

Issue #3565, opened 2022-04-21
Feature: `docker kill --all` terminate all containers at once


bchadwic:
**Description**

As a user when I'm building new projects I tend to spin up and kill a lot of containers for troubleshooting purposes.

Typically I do something like this `alias dka='docker kill $(docker ps -q)'`, but I wonder why there isn't a flag for either `docker stop` or `docker kill` that does this.

If `docker ps -q` returns nothing, then naturally `docker kill` will display the help. Thus extra steps are needed in scripting to avoid this output.

**Proposal**

Could `docker kill` and/or `docker stop` have a `-a/--all` flag that terminates any available running containers? I imagine the `-s` flag could be used in combination like `docker kill -a -s SIGQUIT`

Another option would be `docker killall` which would allow for more in depth options like `killall`. This isn't necessary for my use case though.

If this is a valid feature, I'd like to give it a shot at implementing it 👍 

Thank you

ndeloof:
I'm also running `docker rm -f $(docker ps -aq)` many time a day on my workstation.
Such a command with a single `-a` could be dangerous for production servers, so maybe short `-a` should be avoided to prevent misuse, but I'm fine we introduce `kill --all` and `rm --all`.

bchadwic:
I agree with your reasoning to avoid `-a`

I'll open up a PR later this week for both `kill --all` and `rm --all` 👍

PR #3580, opened 2022-05-01
Add functionality for `docker kill --all`

bchadwic:
**- What I did**
Added `--all` to `docker kill`. This allows a user to easily send a signal to all running containers. 

**- How I did it**
I had made a new func `RequiresArgOrAllFlag` which makes sure a cmd has either args passed to it or has the `--all` flag set. This will be especially useful should `stop` and `rm` get an `--all` flag. Other than that I used the functionality of `ps` to get the containers using the `status=running` filter.

**- How to verify it**
I added new unit test file to cover the `kill --all` functionality as well as the already existing `kill` functionality.

functionality looks like this:
```bash
$ for i in $(seq 3); do docker run -itd alpine sh; done;
b776410ee32895527d804ec7ee08ca7d3ef20c9a4fefc1ebe1384cbfdf8bd051
56245ce81708d8d3eae8c7bc920f353deb2bb6b62d2fa8668b3ebc8ba2c6581e
bd8a998578173134eed2266d51572f861bea3f33aa79b2292a967052329c8b7e
$ docker kill --all
b776410ee32895527d804ec7ee08ca7d3ef20c9a4fefc1ebe1384cbfdf8bd051
56245ce81708d8d3eae8c7bc920f353deb2bb6b62d2fa8668b3ebc8ba2c6581e
bd8a998578173134eed2266d51572f861bea3f33aa79b2292a967052329c8b7e
$ docker kill --all -s SIGINT
no containers running to send SIGINT signal to
$ docker kill --all some-container
"docker kill" requires at least 1 argument or --all flag.
See 'docker kill --help'.

Usage:  docker kill [OPTIONS] CONTAINER [CONTAINER...]

Kill one or more running containers
```

Introduce  `kill --all` to send a signal to all running containers

Fixes https://github.com/docker/cli/issues/3565

***Additional Information***
I decided not to make a `rm --all` like the issue conversation suggests because I think there might be a functionality conflict. `docker container prune` requires either a confirmation or a `--force` flag. I think that making `docker rm --all` remove all containers without confirmation could lead to an unpredictable experience for the end user. I'd be open to suggestions, but I'm not sure if it is necessary given `docker prune` exists. Maybe I'm missing something; I would be happy to hear your thoughts about this and about `docker stop --all` :smiley: 

image


olivierdalang:
Thanks for the effort on this !

> I would be happy to hear your thoughts about this and about docker stop --all

IMO you should at least also include `docker stop --all` for consistency since so close to `docker kill --all` in terms of scope.

About `docker rm --all`, I would also include it albeit indeed more debatable.

I think the rationale for not including it by comparing with `prune` is not so strong: `prune` is about removing unused/dangling stuff, so there indeed the warning makes sense. When doing `docker rm --all`, it's very clear that it's about removing containers. Of course it's possible to make mistakes, but in the end we just do `docker rm $(docker ps -qa)` which is annoying to type but that's not much of a security against mistakes. And whoever looses data due to `docker rm --all` has some reading to do about how to work with Docker and will loose data sooner or later anyway.

PR #3581, opened 2022-05-01
docs: reference: version: improve description

thaJeztah:
See individual commits for details

### DIFF
--- a/docs/reference/commandline/version.md
+++ b/docs/reference/commandline/version.md
@@ -18,30 +18,84 @@ Options:
 
 ## Description
 
-By default, this will render all version information in an easy to read
-layout. If a format is specified, the given template will be executed instead.
+The version command prints the current version number for all independently
+versioned Docker components. Use the [`--format`](#format) option to customize
+the output.
 
-Go's [text/template](https://golang.org/pkg/text/template/) package
-describes all the details of the format.
-
-## Examples
+The version command (`docker version`) outputs the version numbers of Docker
+components, while the `--version` flag (`docker --version`) outputs the version
+number of the Docker CLI you are using.
 
 ### Default output
+
+The default output renders all version information divided into two sections;
+the "Client" section contains information about the Docker CLI and client
+components, and the "Server" section contains information about the Docker
+Engine and components used by the Engine, such as the "Containerd" and "Runc"
+OCI Runtimes.
+
+The information shown may differ depending on how you installed Docker and
+what components are in use. The following example shows the output on a macOS
+machine running Docker Desktop:
 
 ```console
 $ docker version
 
 Client:
- Version:           19.03.8
- API version:       1.40
- Go version:        go1.12.17
- Git commit:        afacb8b
- Built:             Wed Mar 11 01:21:11 2020
+ Version:           20.10.16
+ API version:       1.41
+ Go version:        go1.17.10
+ Git commit:        aa7e414
+ Built:             Thu May 12 09:17:28 2022
  OS/Arch:           darwin/amd64
  Context:           default
- Experimental:      true
 
-Server:
+Server: Docker Desktop 4.8.2 (77141)
+ Engine:
+  Version:          20.10.16
+  API version:      1.41 (minimum version 1.12)
+  Go version:       go1.17.10
+  Git commit:       f756502
+  Built:            Thu May 12 09:15:33 2022
+  OS/Arch:          linux/amd64
+  Experimental:     false
+ containerd:
+  Version:          1.6.4
+  GitCommit:        212e8b6fa2f44b9c21b2798135fc6fb7c53efc16
+ runc:
+  Version:          1.1.1
+  GitCommit:        v1.1.1-0-g52de29d
+ docker-init:
+  Version:          0.19.0
+  GitCommit:        de40ad0
+```
+
+### Client and server versions
+
+Docker uses a client/server architecture, which allows you to use the Docker CLI
+on your local machine to control a Docker Engine running on a remote machine,
+which can be (for example) a machine running in the Cloud or inside a Virtual Machine.
+
+The following example switches the Docker CLI to use a [context](context.md)
+named "remote-test-server", which runs an older version of the Docker Engine
+on a Linux server:
+
+```console
+$ docker context use remote-test-server
+remote-test-server
+
+$ docker version
+
+Client:
+ Version:           20.10.16
+ API version:       1.40 (downgraded from 1.41)
+ Go version:        go1.17.10
+ Git commit:        aa7e414
+ Built:             Thu May 12 09:17:28 2022
+ OS/Arch:           darwin/amd64
+ Context:           remote-test-server
+
+Server: Docker Engine - Community
  Engine:
   Version:          19.03.8
   API version:      1.40 (minimum version 1.12)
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/version.md
```diff
@@ -18,38 +18,97 @@ Options:
...
+
+### Client and server versions
+
+In most cases, the `docker` CLI is used to control Docker Engine running on your
+local machine (for example, the Docker Engine running in Docker Desktop).
+Docker uses a client/server architecture, which allows you to use the `docker`
+CLI on your local machine to control a Docker Engine running on a remote machine,
+which can be (for example) a machine running in the Cloud or inside a Virtual Machine.
+
+The following example switches the `docker` CLI to use a [context](context.md)
```
thaJeztah, line 85: Bit on the fence on how to refer to the CLI; `docker` (the command) CLI, or if we should use "Docker CLI" (guess at some point we should "pick one".

docs/reference/commandline/version.md
```diff
@@ -61,22 +120,74 @@ Server:
...
 
-19.03.8
+20.10.14
 ```
 
 ### Get the client API version
 
+The following example prints the API version that is used by the client:
+
 ```console
```
thaJeztah, line 181: This is something that probably should be done for various other examples; make sure that examples have a short introduction / description that outlines it, instead of only having a "heading" immediately followed by a command-line example.

(admitted; sometimes the introductions are a bit "boilerplate", so we should find a good balance)

docs/reference/commandline/version.md
dockertopia: IMH I think the CLI tool should be referred in full for disambiguation ("Docker CLI" also instead of "docker CLI") and the command whenever possible formatted as code like you show above and sometimes as "docker command" for example, when explaining samples.

dockertopia: I've seen tutorials referring it to as the Docker Engine CLI, it's not shorter but I do like how the term provides clarification and room to refer to subset of commands as extensions to the Docker Engine CLI. Something like "with version 2 of Compose, Docker extends the Engine CLI by adding the Compose CLI, a family of commands with all the functionality provided by the docker-compose binary.)".
If we consider each of these extensions each as a CLI in its own terms, then in a sense they are all Docker CLIs (Compose, Build, etc.). 
So in listing the commands in a Ref guide (thinking mostly of navigation) I'd maybe not advocate for any distinctions but explain in the chapter's intro all of these nuances of what is available in what context and why.

docs/reference/commandline/version.md
```diff
@@ -18,38 +18,97 @@ Options:
 
 ## Description
 
-By default, this will render all version information in an easy to read
-layout. If a format is specified, the given template will be executed instead.
+The `version` command prints detailed information about the version of the
+`docker` CLI, the Docker Engine (`dockerd`), as well as other components.
 
-Go's [text/template](https://golang.org/pkg/text/template/) package
```
dockertopia, line 24: I think this might be a good xref to use here https://docs.docker.com/config/formatting/

docs/reference/commandline/version.md
```diff
@@ -18,38 +18,97 @@ Options:
...
-describes all the details of the format.
+By default, this renders all version information in an easy to read
```
dockertopia, line 24: "easy to read layout" is not neutral in the sense of some users might disagree, I'd steer away from using such expressions. I'd just keep the second sentence like "Use the [`--format`](#format) option to customize the default output layout with a Go template".

docs/reference/commandline/version.md
```diff
@@ -18,38 +18,97 @@ Options:
...
+layout. Use the [`--format`](#format) option to customize the output with a Go
+template.
 
-## Examples
+> **Note**
```
dockertopia, line 28: Note that the version command [`Docker version`] outputs version numbers for Docker components, while the --version flag [`Docker --version`] output the version number for Docker Engine CLI you are using. Can we say Docker Engine CLI instead of Docker CLI or is it innacurate?

docs/reference/commandline/version.md
```diff
@@ -61,12 +120,24 @@ Server:
   GitCommit:        fec3683
 ```
 
+## Examples
+
+### <a name=format></a> Format the output (--format)
+
+The formatting option (`--format`) pretty-prints container output using a Go
```
dockertopia, line 161: I'd maybe just link https://docs.docker.com/config/formatting/ instead of repeating the explanation. But I also would not expect flag description to be again in the examples section but maybe just provide two very different examples.

dockertopia (commented):
Here are some comments which will mostly likely provide the opening for @thaJeztah to teach me stuff : ).

docs/reference/commandline/version.md
thaJeztah: > I've seen tutorials referring it to as the Docker Engine CLI

Yes, so with (e.g.) "Compose" and "Buildx" all being part of the same CLI (but in  some cases through a plugin, but this should mostly be an implementation detail), I think we should avoid _Engine_ CLI, and (unless ambiguous inn a specific context), use "Docker CLI".

dockertopia: Sure, agreed. 
Thinking of communicating with users, do you think there is there the need to clarify what we call the whole vs. some parts (from previous comments I understood there was some level of uncertainty)? Or is it really just an implementation detail?
I wondering if we need that  i.e. for a set of designations to refer clearly to the:
1) whole (all docker commands);
2) (engine CLI?) set of base commands before any integrations;
3) "families" of docker commands or CLI plugins (compose, build, or others) whenever they actually can be obtained and used stand-alone in contexts other than a) Docker CLI b) Docker Desktop.

docs/reference/commandline/version.md
dockertopia: We can work on one command that is frequently used and has more information, optimize that, define based on that one what we'd like to improve on there CLI reference and then revisit each page with this in mind.

docs/reference/commandline/version.md
```diff
@@ -61,22 +116,71 @@ Server:
...
 ```
 
+### API version and version negotiation
+
+The API version used by the client depends on the Docker Engine that the Docker
+CLI is connecting with. When connecting with the Docker Engine, the Docker CLI
+and Docker Engine perform API version negotiation, and select the highest API
+version that is supported by both the Docker CLI and the Docker Engine.
+
+For example, if the CLI is connecting with a Docker 19.03 engine, it downgrades
```
dockertopia, line 126: This would be interesting to show a table for a few versions ( docker engine v (API v) | docker CLI v (API v) | = API v their will be talking with each other.

docs/reference/commandline/version.md
```diff
@@ -18,38 +18,93 @@ Options:
...
+  Version:          1.1.1
+  GitCommit:        v1.1.1-0-g52de29d
+ docker-init:
+  Version:          0.19.0
+  GitCommit:        de40ad0
+```
+
+### Client and server versions
+
+In most cases, the Docker CLI is used to control Docker Engine running on your
```
dockertopia, line 75: As a user in general I have a problem with vague statements. Expressions such as "in most cases", "probably", etc does that to some statements. Makes them vague. I can imagine the case sometime being not wanting to list all possible scenarios. But maybe it's fair for everyone to describe with certain the "mainstream" case or cases (which by definition excludes edge cases).

dockertopia (approved):
This version works better. As a very general feedback I think a lot of our refs are either bare or very narrative and I think I would make more list and tables to just describe possibilities, rather than describing them in prose.

docs/reference/commandline/version.md
dockertopia: Oh it exists! Yes, the link would be helpful.

PR #3139, opened 2021-06-16
Drop support for (archived) Compose-on-Kubernetes

ndeloof:
closes https://github.com/docker/cli/issues/2967
closes https://github.com/docker/cli/pull/3389
(https://docker.atlassian.net/browse/IL-257 for internal tracker)

**- What I did**
Removed support for Kubernetes orchestrator, as required [compose-on-kubernetes](https://github.com/docker/compose-on-kubernetes) backend is archived

**- How I did it**
Code removal

**- How to verify it**
isn't this what CI is for ? :P

**- Description for the changelog**
Removed support for deprecated Compose-on-Kubernetes

**- A picture of a cute animal (not mandatory but encouraged)**
image

PR #3582, opened 2022-05-01
remove some swarm annotations from flags

thaJeztah:
- follow-up to https://github.com/docker/cli/pull/3139

These annotations were added because these options were not supported
when using kubernetes as an orchestrator. Now that this feature was
removed, we can remove these annotations.


thaJeztah:
@ndeloof @rumpl PTAL

ndeloof (approved):

PR #3588, opened 2022-05-05
Loads .env file if existing at the very beginning

ulyssessouza:
**- What I did**
Load .env files when existing

**- How I did it**
Load it as the very first thing we do in `main`

**- How to verify it**
Setting `DOCKER_BUILDKIT=X` or `DOCKER_HOST=XXXXX` in a `.env` file in `$PWD`

Load .env files when existing

### DIFF
--- a/cli/command/stack/loader/loader.go
+++ b/cli/command/stack/loader/loader.go
@@ -13,6 +13,7 @@ import (
 	"github.com/docker/cli/cli/compose/loader"
 	"github.com/docker/cli/cli/compose/schema"
 	composetypes "github.com/docker/cli/cli/compose/types"
+	"github.com/docker/cli/opts"
 	"github.com/pkg/errors"
 )
 
@@ -97,21 +98,8 @@ func GetConfigDetails(composefiles []string, stdin io.Reader) (composetypes.Conf
 	}
 	// Take the first file version (2 files can't have different version)
 	details.Version = schema.Version(details.ConfigFiles[0].Config)
-	details.Environment, err = buildEnvironment(os.Environ())
+	details.Environment, err = opts.BuildEnvironment(os.Environ())
 	return details, err
-}
-
-func buildEnvironment(env []string) (map[string]string, error) {
-	result := make(map[string]string, len(env))
-	for _, s := range env {
-		// if value is empty, s is like "K=", not "K".
-		if !strings.Contains(s, "=") {
-			return result, errors.Errorf("unexpected environment %q", s)
-		}
-		kv := strings.SplitN(s, "=", 2)
-		result[kv[0]] = kv[1]
-	}
-	return result, nil
 }
 
 func loadConfigFiles(filenames []string, stdin io.Reader) ([]composetypes.ConfigFile, error) {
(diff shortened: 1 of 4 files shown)


thaJeztah:
Curious; why is this needed? Wouldn't compose handle the environment, and set the environment for things it calls out to? (`os.Exec("docker cli")` with the appropriate env-vars?)

I know there's a lot of issues with `.env` that need to be looked into; IIUC, this makes `docker` use `.env`  for all commands (which would be quite a breaking change)

StefanScherer:
It came out of a discussion that it would feel natural to have support for more `docker` commands as compose is now a cli plugins with `docker compose ...` commands supporting the `.env` file.

I understand your concerns. Maybe we can try it later once the cli release pipeline is more decoupled from engine.
For now I suggest to close the PR.

Or what do you think @ndeloof I think you were involved in that discussion as well?

thaJeztah:
> It came out of a discussion that it would feel natural to have support for more docker ... commands as compose is now a cli plugins with docker compose ... commands supporting the .env file.
 
Honestly, I think we should go back to the drawing board for `.env` files. I can see the convenience for them, but they're a cause for many, many issues, and I'd _really_ prefer to at least acknowledge those, before we increase the scope of that feature; just some of those from the top of my head

- `.env` vs `--env-file` (compose) vs `--env-file` (`docker run`); it's already confusing users within compose itself, and even more when taking `docker run --env-file` into account, and the behavior (order of processing, scope) is under-defined.
- `.env` files are not a "docker" concept; other tooling uses `.env` files as well, which is a "feature" to some (I can re-use my `.env` file for compose and share the same options) and a problem for others (my project's `.env` file is also picked up by `docker`, and I must not pass a flag to make `docker` ignore it).
- location of `.env` files; relative to working-directory, project directory (`compose`), etc. Some of those depend on a compose-file being present (`docker` will traverse the filesystem, which can even mean a potential security issue), or depend on `--project-dir` (which is a compose-specific option).
- "conceptually"; `.env` files were introduced when compose was still in early stages (one compose file, one env-file), which somewhat made sense at the time, but features evolved, and we now have projects containing multiple compose files, multiple `--profile`'s for compose files, override files (`docker-compose.override.yaml`), etc. Having a "single" `.env` file "to rule them all" doesn't scale well.
- Adding to that (within scope of "compose") it's slightly odd to have a config file (`.env`) to configure a config-file (`docker-compose.yaml`), besides just the overlap with override-files (which have a similar purpose).

I know the feature has been around in compose for a long time, so we can't _remove_ it (for compose), but we definitely must look at it before increasing the scope. I also see benefits from the feature for various use-cases, and those must be looked at if we would be going to the drawing board. In that light, perhaps we should also have a look at

- [parameter files](https://github.com/docker/app#writing-an-app-definition) as were used for `docker app` (although that wouldn't directly benefit non-compose things)
- enhancing `docker context` (which is in need of a bit of an overhaul) to make it more useful beyond _just_ connection settings
- a better concept of "projects" (e.g. borrow some of the ideas from the [hackday demo I once did](https://github.com/thajeztah/cli/tree/init_easy_hackday_demo))

StefanScherer:
Thanks for the great writeup @thaJeztah ! With compose it's fine to support .env, and there's no actual pressure to support it for other docker commands.  
Let's close this draft.

Issue #3593, opened 2022-05-06
`docker info --format` panic


rumpl:
**Description**

The cli panics when I run `docker info --format '{{.LoggingDriver}}'` while the daemon is not running.

**Steps to reproduce the issue:**
1. stop the daemon
2. run `docker info --format '{{.LoggingDriver}}'`

**Describe the results you received:**

```
panic: reflect: indirection through nil pointer to embedded struct [recovered]
	panic: reflect: indirection through nil pointer to embedded struct

goroutine 1 [running]:
text/template.errRecover(0xc0007972e8)
	/usr/local/go/src/text/template/exec.go:169 +0x1c5
panic(0x10ec3d780, 0x10ef37608)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
reflect.Value.FieldByIndex(0x10edd5720, 0xc000a10100, 0x99, 0xc0007b8230, 0x2, 0x2, 0x0, 0x10efc66e8, 0x10ec3d780)
	/usr/local/go/src/reflect/value.go:889 +0x325
text/template.(*state).evalField(0xc000797258, 0x10edd5720, 0xc000a10100, 0x99, 0x7ff7b2c56537, 0xd, 0x10efb0300, 0xc000a143c0, 0xc000230120, 0x1, ...)
	/usr/local/go/src/text/template/exec.go:616 +0x4ed
text/template.(*state).evalFieldChain(0xc000797258, 0x10edd5720, 0xc000a10100, 0x99, 0x10edd5720, 0xc000a10100, 0x99, 0x10efb0300, 0xc000a143c0, 0xc000230110, ...)
	/usr/local/go/src/text/template/exec.go:569 +0x225
text/template.(*state).evalFieldNode(0xc000797258, 0x10edd5720, 0xc000a10100, 0x99, 0xc000a143c0, 0xc000230120, 0x1, 0x1, 0x10eccbe40, 0x10fd457b0, ...)
	/usr/local/go/src/text/template/exec.go:533 +0x113
text/template.(*state).evalCommand(0xc000797258, 0x10edd5720, 0xc000a10100, 0x99, 0xc000a14390, 0x10eccbe40, 0x10fd457b0, 0x99, 0xd0, 0xc0007b2820, ...)
	/usr/local/go/src/text/template/exec.go:461 +0x8f0
[...]

PR #3179, opened 2021-07-06
docker info: skip API connection if possible

thaJeztah:
The docker info output contains both "local" and "remote" (daemon-side) information.
The API endpoint to collect daemon information (`/info`) is known to be "heavy",
and (depending on what information is needed) not needed.

This patch checks if the template (`--format`) used requires information from the
daemon, and if not, omits making an API request.

This will improve performance if (for example), the current "context" is requested
from `docker info` or if only plugin information is requested.

Before:

    time docker info --format '{{range  .ClientInfo.Plugins}}Plugin: {{.Name}}, {{end}}'
    Plugin: buildx, Plugin: compose, Plugin: scan,

    ________________________________________________________
    Executed in  301.91 millis    fish           external
       usr time  168.64 millis   82.00 micros  168.56 millis
       sys time  113.72 millis  811.00 micros  112.91 millis

    time docker info --format '{{json .ClientInfo.Plugins}}'

    time docker info --format '{{.ClientInfo.Context}}'
    default

    ________________________________________________________
    Executed in  334.38 millis    fish           external
       usr time  177.23 millis   93.00 micros  177.13 millis
       sys time  124.90 millis  927.00 micros  123.97 millis

    docker context use remote-ssh-daemon
    time docker info --format '{{.ClientInfo.Context}}'
    remote-ssh-daemon

    ________________________________________________________
    Executed in    1.22 secs   fish           external
       usr time  116.93 millis  110.00 micros  116.82 millis
       sys time  144.36 millis  887.00 micros  143.47 millis

And daemon logs:

    Jul 06 12:42:12 remote-ssh-daemon dockerd[14377]: time="2021-07-06T12:42:12.139529947Z" level=debug msg="Calling HEAD /_ping"
    Jul 06 12:42:12 remote-ssh-daemon dockerd[14377]: time="2021-07-06T12:42:12.140772052Z" level=debug msg="Calling HEAD /_ping"
    Jul 06 12:42:12 remote-ssh-daemon dockerd[14377]: time="2021-07-06T12:42:12.163832016Z" level=debug msg="Calling GET /v1.41/info"

After:

    time ./build/docker info --format '{{range  .ClientInfo.Plugins}}Plugin: {{.Name}}, {{end}}'
    Plugin: buildx, Plugin: compose, Plugin: scan,

    ________________________________________________________
    Executed in  139.84 millis    fish           external
       usr time   76.53 millis   62.00 micros   76.46 millis
[...]

PR #3594, opened 2022-05-06
info: improve handling of empty Info

thaJeztah:
- relates to https://github.com/docker/cli/issues/3593
- relates to https://github.com/docker/cli/pull/3179

Before this change, the function could print an error in some cases, for example;

```bash
docker -H tcp://127.0.0.1:2375 info --format '{{.LoggingDriver}}'

template: :1:2: executing "" at <.LoggingDriver>: reflect: indirection through nil pointer to embedded struct field Info
```

With this patch applied, the error is handled gracefully, and when failing to
connect with the daemon, the error is logged;

```bash
docker -H tcp://127.0.0.1:2375 info --format '{{.LoggingDriver}}'
Cannot connect to the Docker daemon at tcp://127.0.0.1:2375. Is the docker daemon running?

docker -H tcp://127.0.0.1:2375 info --format '{{json .}}'
Cannot connect to the Docker daemon at tcp://127.0.0.1:2375. Is the docker daemon running?
{"ID":"","Containers":0,"..."}}
```

Note that the connection error is also included in the JSON `ServerErrors` field,
so that the information does not get lost, even if STDERR would be redirected;

```bash
docker -H tcp://127.0.0.1:2375 info --format '{{json .ServerErrors}}' 2> /dev/null
["Cannot connect to the Docker daemon at tcp://127.0.0.1:2375. Is the docker daemon running?"]
```

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/system/info.go
+++ b/cli/command/system/info.go
@@ -85,8 +85,17 @@ func runInfo(cmd *cobra.Command, dockerCli command.Cli, opts *infoOptions) error
 		if dinfo, err := dockerCli.Client().Info(ctx); err == nil {
 			info.Info = &dinfo
 		} else {
-			fmt.Fprintln(dockerCli.Err(), err)
 			info.ServerErrors = append(info.ServerErrors, err.Error())
+			if opts.format == "" {
+				// reset the server info to prevent printing "empty" Server info
+				// and warnings, but don't reset it if a custom format was specified
+				// to prevent errors from Go's template parsing during format.
+				info.Info = nil
+			} else {
+				// if a format is provided, print the error, as it may be hidden
+				// otherwise if the template doesn't include the ServerErrors field.
+				fmt.Fprintln(dockerCli.Err(), err)
+			}
 		}
 	}
 
--- a/cli/command/system/info.go
+++ b/cli/command/system/info.go
@@ -67,11 +67,12 @@ func NewInfoCommand(dockerCli command.Cli) *cobra.Command {
 }
 
 func runInfo(cmd *cobra.Command, dockerCli command.Cli, opts *infoOptions) error {
-	var info info
-
-	info.ClientInfo = &clientInfo{
-		Context: dockerCli.CurrentContext(),
-		Debug:   debug.IsEnabled(),
+	info := info{
+		ClientInfo: &clientInfo{
+			Context: dockerCli.CurrentContext(),
+			Debug:   debug.IsEnabled(),
+		},
+		Info: &types.Info{},
 	}
 	if plugins, err := pluginmanager.ListPlugins(dockerCli, cmd.Root()); err == nil {
 		info.ClientInfo.Plugins = plugins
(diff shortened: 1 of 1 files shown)


rumpl (approved):

cli/command/system/info.go
```diff
@@ -84,6 +85,7 @@ func runInfo(cmd *cobra.Command, dockerCli command.Cli, opts *infoOptions) error
 		if dinfo, err := dockerCli.Client().Info(ctx); err == nil {
 			info.Info = &dinfo
 		} else {
+			fmt.Fprintln(dockerCli.Err(), err)
```
rumpl, line 88: Shouldn't we return on error?

thaJeztah: I think this case was to allow at least the Client info to be printed (Server info would not be printed, but "failed to connect") 🤔

rumpl: Ah right!

thaJeztah: Gave it a quick try; here's if we don't return an error

```bash
docker -H tcp://127.0.0.1:2375 info
Cannot connect to the Docker daemon at tcp://127.0.0.1:2375. Is the docker daemon running?
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.2)
  compose: Docker Compose (Docker Inc., v2.4.1)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Plugins:
  Volume:
  Network:
  Log:
 Swarm:
  NodeID:
  Is Manager: false
  Node Address:
 CPUs: 0
 Total Memory: 0B
 Docker Root Dir:
 Debug Mode: false
 Experimental: false
 Live Restore Enabled: false

WARNING: No memory limit support
WARNING: No swap limit support
WARNING: No oom kill disable support
WARNING: No cpu cfs quota support
WARNING: No cpu cfs period support
WARNING: No cpu shares support
WARNING: No cpuset support
WARNING: IPv4 forwarding is disabled
WARNING: bridge-nf-call-iptables is disabled
WARNING: bridge-nf-call-ip6tables is disabled
ERROR: Cannot connect to the Docker daemon at tcp://127.0.0.1:2375. Is the docker daemon running?
errors pretty printing info
```

And if we return the error;

```bash
docker -H tcp://127.0.0.1:2375 info
Cannot connect to the Docker daemon at tcp://127.0.0.1:2375. Is the docker daemon running?
```

thaJeztah: From some perspective "either" can make sense, although I think it's useful to still have the Client information (even if we can't connect to the daemon), which is the same behavior as we have for `docker version`.

That said; to further improve;

- We should not print the `WARNING` output if we can't connect to the daemon (as they're inaccurate in that case). Those warnings are printed by [`printServerWarningsLegacy()`](https://github.com/docker/cli/blob/b68db383d389ae7c0302aaa3dc33f3155b6fe9a3/cli/command/system/info.go#L458-L463), which is only to be used for older daemons (and should not be called if we have no daemon to connect with).
- Similar to the `docker version` output, we should consider _not_ printing the `Server:` output if we can't connect, and instead only print the error for that part.
- ^^ perhaps an exception to the above would be if a `--format` flag is specified (?) (e.g. `--format '{{.Containers}}'`) (not sure)

thaJeztah:
@rumpl I pushed a commit that improves the "failed to connect to server" situation; it now no longer prints the Server information (and instead only shows the error if no `--format` flag is printed)

ndeloof (approved):

PR #3599, opened 2022-05-10
use local ConvertToHostname() implementation

thaJeztah:
Commit 27b2797f7deb3ca5b7f80371d825113deb1faca1 (https://github.com/docker/cli/pull/1642) added a local implementation of this function, so let's use the local variant to (slightly) reduce the dependency on moby's registry package.

Also made some minor cleanups.

### DIFF
--- a/cli/command/registry.go
+++ b/cli/command/registry.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/distribution/reference"
 	"github.com/docker/cli/cli/config/configfile"
+	"github.com/docker/cli/cli/config/credentials"
 	configtypes "github.com/docker/cli/cli/config/types"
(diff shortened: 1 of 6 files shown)


thaJeztah:
@rumpl @ndeloof ptal 🤗

ndeloof:
I don't think "Copied from github.com/docker/docker/registry.ConvertToHostname to reduce dependencies" makes our codebase better.
1. we still (obviously) depend on docker/docker as the API is not an isolated project vs engine backend
2. we now have two implementations of the same `ConvertToHostname` function to maintain (for sure, low maintenance expected here)

_imho_ it would make more sense for such utility func to be shared between cli and engine within one of the `api` package. I'm not sure about the initial intent in https://github.com/docker/cli/pull/1642

PR #3129, opened 2021-06-09
Proposed Cluster (CSI) Volume Command

dperny:
This PR is a **not fully functional draft proposal** of the CLI for Swarm CSI Cluster Volumes.

This proposal involves merging the cluster volume specific options and commands with the existing volume subcommand. There are a handful of new create flags which are only applicable to cluster volumes. There is also a new flag for `list`, which lists only cluster volumes and formats the table differently.

Examples of the output can be found in the new `.golden` files.

This PR does not build as-is, but instead is to show what a combined volumes command look like, in order to assess its ergonomics and UI.

PR #3606, opened 2022-05-13
Proposed Cluster (CSI) Volume Command (rebase)

thaJeztah:
- rebase of https://github.com/docker/cli/pull/3129 
- cli-side changes for https://github.com/moby/moby/pull/41982
- relates to https://github.com/moby/moby/issues/31923
- relates to https://github.com/moby/moby/issues/39624
---

This proposal involves merging the cluster volume specific options and commands with the existing volume subcommand. There are a handful of new create flags which are only applicable to cluster volumes. There is also a new flag for `list`, which lists only cluster volumes and formats the table differently.

Examples of the output can be found in the new `.golden` files.


thaJeztah:
looks like most of the intermediate commits no longer built (due to the vendoring changes), so I squashed them

thaJeztah:
Made some changes;

Annotated `docker volume update` and the cluster-volume flags on `docker volume create` both for documentation, and to hide them if swarm is not enabled (or the daemon is too old); for example;

```
docker volume update
docker volume update requires API version 1.42, but the Docker daemon API version is 1.41
```

And I removed the `-g` shorthand for `--group` (for now); we're trying to be conservative on shorthand flags, as they easily "conflict" with other options, and to reserve them for options that are _very_ frequently used. It's easier to add later, than to remove/deprecated.

Worth noting that `docker stack deploy` does not yet support this feature, and may need changes in the compose-spec (?); we probably need a follow-up after this to add support for it.

And of course, we need to do a full review of the UX before GA (are we ok with all the flags? should some be "combined"?)

cli/command/volume/list.go
```diff
@@ -38,6 +43,7 @@ func newListCommand(dockerCli command.Cli) *cobra.Command {
 	flags.BoolVarP(&options.quiet, "quiet", "q", false, "Only display volume names")
 	flags.StringVar(&options.format, "format", "", flagsHelper.FormatHelp)
 	flags.VarP(&options.filter, "filter", "f", "Provide filter values (e.g. 'dangling=true')")
+	flags.BoolVar(&options.cluster, "cluster", false, "Display only cluster volumes, and use cluster volume list formatting")
```
thaJeztah, line 47: Let me annotate this one as well

Perhaps we should have a `--filter` for this, but we can discuss that further

thaJeztah:
@dperny @corhere @tianon @cpuguy83 PTAL

cli/command/volume/create.go
```diff
@@ -37,6 +55,7 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 				}
 				options.name = args[0]
 			}
+			options.cluster = hasClusterVolumeOptionSet(cmd.Flags())
```
corhere, line 58: I think creating a cluster volume should be explicitly set by the user, rather than implied by setting any of the cluster-volume options so that users don't unintentionally create cluster volumes by setting a generically-named flag. Strawman: the `--group` flag pulls double-duty as the "create a cluster volume" flag by renaming it to <tt>&#8209;&#8209;cluster&#8209;volume&nbsp;_GROUPNAME_</tt> and making it an error to set any of the other cluster-only options unless that flag is also set. If a user wanted for whatever reason to create a cluster volume in the empty-string group, they could do so explicitly: <tt>docker volume create &#8209;&#8209;cluster&#8209;volume&nbsp;""</tt>

cli/command/volume/create.go
```diff
@@ -48,16 +67,130 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
...
 	return cmd
 }
 
+// hasClusterVolumeOptionSet returns true if any of the cluster-specific
+// options are set.
+func hasClusterVolumeOptionSet(flags *pflag.FlagSet) bool {
+	return flags.Changed("group") || flags.Changed("scope") ||
+		flags.Changed("sharing") || flags.Changed("availability") ||
+		flags.Changed("type") || flags.Changed("secrets") ||
+		flags.Changed("limit-bytes") || flags.Changed("required-bytes")
```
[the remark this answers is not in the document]
thaJeztah: Yes, there's some things to clean up in this area. I had a variant where I defined a slice of flag-names that are related, and used the slice for both setting the annotations and for this code, but in the end decided to keep my changes on top of @dperny's PR minimal

My goal with this rebase was to have  the PR in a "mergeable" state, so that I could do a test-build/test-release of 22.06 with at least _something_ that exposes the new feature. So I only fixed the type renames from upstream, and added the second commit to have the correct annotations on the flags.

dperny:
This looks fine to me for now. I'm more than happy to make any UI changes desired. UI is the easy/fun part.

olljanat:
FYI. I have build some example scripts and guidance to https://github.com/olljanat/csi-plugins-for-docker-swarm about how to make existing CSI plugins working with Swarm which those who following this discussion might find useful.

PR #3614, opened 2022-05-17
add --signal option to stop and restart

thaJeztah:
- relates to https://github.com/moby/moby/pull/43206

Wording and documentation still need to be updated, but will do
so in a follow-up.

Also removing the default "10 seconds" from the timeout flags, as
this default is not actually used, and may not match the actual
default (which is defined on the daemon side).

### DIFF
--- a/cli/command/container/restart.go
+++ b/cli/command/container/restart.go
@@ -14,8 +14,9 @@ import (
 )
 
 type restartOptions struct {
-	nSeconds        int
-	nSecondsChanged bool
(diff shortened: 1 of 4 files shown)


thaJeztah:
I have a branch with "WIP" documentation changes, and changes to the completion scripts, but will push that as a follow-up pull request (some more work needed).

ndeloof (approved):

rumpl (approved):

PR #3616, opened 2022-05-17
export ParseSecurityOpts so this can be used by docker/compose

ndeloof:
**- What I did**
Made `ParseSecurityOpts` a public function and introduced relative path solver function as parameter, so it can be used by docker/compose.
This will avoid code duplication [here](https://github.com/docker/compose/tree/master/pkg/compose/create.go#L393)

**- How I did it**
Use `filepath.Abs` by default within docker/cli

**- How to verify it**
This is what CI is for, isn't it :P 

image


thaJeztah:
Generally not against exporting some of these; some "quick" thoughts;

- should this be in https://github.com/docker/cli/tree/master/opts ? (Not 100% sure, as those are more for options that can be used directly for `flag.Var()`)
- I'm wondering if this function is doing _too much_ (as in; should it really read the value as part of "parsing"?); wondering if that functionality should (ideally) be split; so 1. Parse Options (only check if the format is correct) (which could return a more _structured__ output, and 2. handle the options from the struct (which can involve reading the file)

cli/command/container/opts.go
```diff
@@ -850,8 +850,8 @@ func parseLoggingOpts(loggingDriver string, loggingOpts []string) (map[string]st
 	return loggingOptsMap, nil
 }
 
-// takes a local seccomp daemon, reads the file contents for sending to the daemon
-func parseSecurityOpts(securityOpts []string) ([]string, error) {
+// ParseSecurityOpts takes a local seccomp daemon, reads the file contents for sending to the daemon
+func ParseSecurityOpts(securityOpts []string, rel func(path string) (string, error)) ([]string, error) {
```
milas, line 854: The motivation for exporting this came from a Compose PR that wanted to make changes to this method because it mutates the input.

Can't comment directly on the line (873 in the before diff):
```go
securityOpts[key] = fmt.Sprintf("seccomp=%s", b.Bytes())
```

I think it'd be good to create a new slice to return from this method rather than mutating the input, particularly now that it's exported.

PR #3619, opened 2022-05-18
cli/container: Fill ConsoleSize in create

vvoland:
Related to https://github.com/moby/moby/pull/43593

**- What I did**
Fill HostConfig.ConsoleSize for both `docker run` and `docker create`

**- How I did it**
Move the code that sets the ConsoleSize from runContainer to createContainer.
This does not impact the runContainer, as it calls createContainer anyway.

**- How to verify it**
```
# Change terminal size from the default 24x80
docker create -t ubuntu sh -c 'tput lines; tput cols'
docker start <id>
docker logs <id>
# Check if the printed size is correct
```

### DIFF
--- a/cli/command/container/create.go
+++ b/cli/command/container/create.go
@@ -261,6 +261,8 @@ func createContainer(ctx context.Context, dockerCli command.Cli, containerConfig
 		}
 	}
 
+	hostConfig.ConsoleSize[0], hostConfig.ConsoleSize[1] = dockerCli.Out().GetTtySize()
+
 	response, err := dockerCli.Client().ContainerCreate(ctx, config, hostConfig, networkingConfig, platform, opts.name)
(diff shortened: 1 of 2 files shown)


cli/command/container/create.go
```diff
@@ -261,6 +261,8 @@ func createContainer(ctx context.Context, dockerCli command.Cli, containerConfig
 		}
 	}
 
+	hostConfig.ConsoleSize[0], hostConfig.ConsoleSize[1] = dockerCli.Out().GetTtySize()
```
vvoland, line 264: Shouldn't this be filled only if config.Tty?

thaJeztah (approved):
LGTM

vvoland:
Just wondering if this makes sense if config.Tty is false?

PR #3627, opened 2022-05-20
command/exec: Fill ConsoleSize

vvoland:
Depends on:
- https://github.com/moby/moby/pull/43622 (adds execConfig.ConsoleSize in vendor/github.com/docker/docker)
- https://github.com/docker/cli/pull/3688

Relates to:
- https://github.com/docker/cli/issues/3554

This makes the exec'd process started with console dimensions matching the users terminal.

**- How I did it**
Set the exec config ConsoleSize to cli stdout tty size

**- How to verify it**
```
# resize terminal from the default size

cid=$(docker run -d ubuntu sleep infinity)

docker exec -it $cid stty size
# check if the printed size is as expected

docker exec -it $cid top -n 1
# check if top UI is rendered at the full terminal size
```

`docker exec -t` now sets the console size for the exec'd process immediately when it's created

### DIFF
--- a/cli/command/container/exec.go
+++ b/cli/command/container/exec.go
@@ -117,6 +117,8 @@ func RunExec(dockerCli command.Cli, options ExecOptions) error {
 		}
 	}
 
+	fillConsoleSize(execConfig, dockerCli)
+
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM

PR #3638, opened 2022-05-30
feat(cli): add --http-header option

mapleeit:
Signed-off-by: Zhaofeng Miao <522856232@qq.com>

**- What I did**

Add `--http-header` option to CLI to allow adding custom http headers when doing `docker command` 

**- How I did it**

See the diff

**- How to verify it**

I tested it locally, if you think it's good to continue, I'll add test for it and polish the code.

Screen Shot 2022-05-30 at 4 19 23 PM

Add `--http-header` option to CLI to allow specifying http headers. `docker --http-header "key: value" version`

: )

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -229,7 +229,7 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...Initialize
 	}
 
 	if cli.client == nil {
-		cli.client, err = newAPIClientFromEndpoint(cli.dockerEndpoint, cli.configFile)
+		cli.client, err = newAPIClientFromEndpoint(opts.Common, cli.dockerEndpoint, cli.configFile)
 		if err != nil {
 			return err
 		}
@@ -255,18 +255,26 @@ func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to resolve docker endpoint")
 	}
-	return newAPIClientFromEndpoint(endpoint, configFile)
+	return newAPIClientFromEndpoint(opts, endpoint, configFile)
 }
 
-func newAPIClientFromEndpoint(ep docker.Endpoint, configFile *configfile.ConfigFile) (client.APIClient, error) {
+func newAPIClientFromEndpoint(opts *cliflags.CommonOptions, ep docker.Endpoint, configFile *configfile.ConfigFile) (client.APIClient, error) {
 	clientOpts, err := ep.ClientOpts()
 	if err != nil {
 		return nil, err
 	}
-	customHeaders := make(map[string]string, len(configFile.HTTPHeaders))
+	customHeaders := make(map[string]string, len(configFile.HTTPHeaders)+len(opts.HttpHeaders))
 	for k, v := range configFile.HTTPHeaders {
 		customHeaders[k] = v
 	}
+
+	for i := 0; i < len(opts.HttpHeaders); i++ {
+		split := strings.Split(opts.HttpHeaders[i], ":")
+		k := strings.TrimSpace(split[0])
+		v := strings.TrimSpace(split[1])
+		customHeaders[k] = v
+	}
+
 	customHeaders["User-Agent"] = UserAgent()
 	clientOpts = append(clientOpts, client.WithHTTPHeaders(customHeaders))
 	return client.NewClientWithOpts(clientOpts...)
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -269,7 +269,7 @@ func newAPIClientFromEndpoint(opts *cliflags.CommonOptions, ep docker.Endpoint, 
 	}
 
 	for i := 0; i < len(opts.HttpHeaders); i++ {
-		split := strings.Split(opts.HttpHeaders[i], ":")
+		split := strings.SplitN(opts.HttpHeaders[i], ":", 2)
 		k := strings.TrimSpace(split[0])
 		v := strings.TrimSpace(split[1])
 		customHeaders[k] = v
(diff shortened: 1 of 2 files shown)


mapleeit:
@thaJeztah It's good to have your opinion here. Thanks in advance.

thaJeztah:
Could you describe the use-case for this feature? The CLI already supports a `HttpHeaders` option in the cli configuration file (https://docs.docker.com/engine/reference/commandline/cli/#custom-http-headers), so I'm wondering if that would be an alternative to use.

mapleeit:
Thanks for the quick response.

We are treating docker daemon as a microservice in our system. So we need to append `x-request-id` http header to the each request, so that we can trace and route them as we need. And it will change every time apparently. Writing it to the `config.json` doesn't work in concurrent cases.

thaJeztah:
> And it will change every time apparently. Writing it to the config.json doesn't work in concurrent cases.

Ah, right, gotcha, yes, understand the use-case.

cli/command/cli.go
```diff
@@ -255,18 +255,26 @@ func NewAPIClientFromFlags(opts *cliflags.CommonOptions, configFile *configfile.
...
 		return nil, err
 	}
-	customHeaders := make(map[string]string, len(configFile.HTTPHeaders))
+	customHeaders := make(map[string]string, len(configFile.HTTPHeaders)+len(opts.HttpHeaders))
 	for k, v := range configFile.HTTPHeaders {
 		customHeaders[k] = v
 	}
+
+	for i := 0; i < len(opts.HttpHeaders); i++ {
+		split := strings.Split(opts.HttpHeaders[i], ":")
```
thaJeztah, line 272: I have yet to have a closer look, but some quick thoughts;

- we should probably use `=` as separator (`--http-header My-Header=myvalue`)
- use `strings.SplitN(.., ..., 2)` (as we expect only a single separator); this prevents issues if the value would contain `:`, which could be if (just thinking out lout) for example, an IPv6 address is passed (`Foo-Header=[::]:123`)
- We probably need to do this in some other places as well (), but perhaps we should also use [`http. CanonicalHeaderKey()`](https://pkg.go.dev/net/http#CanonicalHeaderKey) (after stripping whitespace), to prevent duplicate headers, and to prevent `User-Agent` from being overridden (which we don't want); it looks like the [client already does normalising](https://github.com/moby/moby/blob/f1dd6bf84e28930e1ccd903361f9284fb22d3b8a/client/request.go#L245) (I just notice it doesn't trim whitespace, and probably should though), but it doesn't hurt to prevent possible duplicates here

Overall, I still want to look if this function (`newAPIClientFromEndpoint()`) is the best place to set this; I'm considering if perhaps we should merge the config-file and flag options earlier, perhaps as part of [`DockerCLI.Initialize()`](https://github.com/docker/cli/blob/d0df532a25ad18e8b4325edcf46f3d0f05b47708/cli/command/cli.go#L195) ? (I'll have to have a look if that would work)

mapleeit:
Do you think I should continue to add unit test and polish the code now?

cli/flags/common.go
```diff
@@ -45,13 +45,14 @@ var (
...
-	TLSOptions *tlsconfig.Options
-	Context    string
+	Debug       bool
+	Hosts       []string
+	LogLevel    string
+	TLS         bool
+	TLSVerify   bool
+	TLSOptions  *tlsconfig.Options
+	Context     string
+	HttpHeaders []string
```
thaJeztah, line 55: w.r.t. the above (parsing) in mind I'm thinking if perhaps the value parsing should be part of a custom type (so that this would be an (e.g.) `opts.HttpHeadersOpt` with its own validation function, similar fo [`opts.ListOpts`](https://github.com/docker/cli/blob/c780f7c4abaf67034ecfaa0611e03695cf9e4a3e/opts/opts.go#L22) and [`opts. NamedListOpts`](https://github.com/docker/cli/blob/c780f7c4abaf67034ecfaa0611e03695cf9e4a3e/opts/opts.go#L132)

Thinking of which (sorry, bit hazy on that one), but I _think_ the `NamedListOpts` and `NamedOpts` types were used for command-line flags that also have an equivalent in `~/.docker/config.json`. Which (in this case) could be considered the case (the `--http-headers` being an equivalent for the `HttpHeaders` option in the config file)

(Hope this helps, but happy to do some further digging in this)

thaJeztah:
> Do you think I should continue to add unit test and polish the code now? or still need consideration or discussion.

I was writing down some first comments (let me post them below); overall, I'm not against adding this as a top-level flag (but of course things can still be discussed with other maintainers/reviewers as well).

cli/command/cli.go
mapleeit: > we should probably use = as separator (--http-header My-Header=myvalue)

I'm trying to keep consistent with [`curl`](https://stackoverflow.com/questions/356705/how-to-send-a-header-using-a-http-request-through-a-curl-call) since it has been widely accepted. And I'm not sure if there are traps if we use `=`

mapleeit: > I'm considering if perhaps we should merge the config-file and flag options earlier, perhaps as part of [DockerCLI.Initialize()](https://github.com/docker/cli/blob/d0df532a25ad18e8b4325edcf46f3d0f05b47708/cli/command/cli.go#L195)

Agree. This will make the whole configing procedure more clear

cli/flags/common.go
mapleeit: Basically, you want to use `opts.NamedMapOpts` and according validator, right?

cli/command/cli.go
mapleeit: > use strings.SplitN(.., ..., 2)

Addressed

mapleeit:
What should I do to push this PR forward? guys

PR #3641, opened 2022-05-31
Deflake TestVolumeCreateClusterOpts

### LINKED ISSUE #3636: Flaky test: TestVolumeCreateClusterOpts
Looks like we have a new flaky test; I'm guessing something is using a map, and therefore randomized by Go. Introduced in https://github.com/docker/cli/pull/3606

```
--- FAIL: TestVolumeCreateClusterOpts (0.00s)
    create_test.go:195: assertion failed: 
        --- body
        +++ expectedBody
          volume.CreateOptions{
          	ClusterVolumeSpec: &volume.ClusterVolumeSpec{
          		... // 2 identical fields
          		AccessibilityRequirements: &{Requisite: {{Segments: {"region": "R1", "zone": "Z1"}}, {Segments: {"region": "R1", "zone": "Z2"}}, {Segments: {"region": "R1", "zone": "Z3"}}}, Preferred: {{Segments: {"region": "R1", "zone": "Z2"}}, {Segments: {"region": "R1", "zone": "Z3"}}}},
          		CapacityRange:             &{RequiredBytes: 1234, LimitBytes: 567890},
          		Secrets: []volume.Secret{
        - 			{Key: "key2", Secret: "secret2"},
          			{Key: "key1", Secret: "secret1"},
        + 			{Key: "key2", Secret: "secret2"},
          		},
          		Availability: "active",
          	},
          	Driver:     "csi",
          	DriverOpts: {},
          	... // 2 identical fields
          }
```

/cc @dperny

photra:
**- What I did**
Fixes #3636

**- How I did it**
Added sorting logic to test to ensure consistent ordering of `Secrets`

**- How to verify it**
Run the full test suite multiple times to validate this fix

Deflaked TestVolumeCreateClusterOpts

image

### DIFF
--- a/cli/command/volume/create_test.go
+++ b/cli/command/volume/create_test.go
@@ -3,6 +3,7 @@ package volume
 import (
 	"io"
 	"reflect"
+	"sort"
 	"strings"
 	"testing"
 
@@ -192,6 +193,9 @@ func TestVolumeCreateClusterOpts(t *testing.T) {
 
 	cli := test.NewFakeCli(&fakeClient{
 		volumeCreateFunc: func(body volume.CreateOptions) (volume.Volume, error) {
+			sort.SliceStable(body.ClusterVolumeSpec.Secrets, func(i, j int) bool {
+				return body.ClusterVolumeSpec.Secrets[i].Key < body.ClusterVolumeSpec.Secrets[j].Key
+			})
 			assert.DeepEqual(t, body, expectedBody)
 			return volume.Volume{}, nil
 		},
--- a/cli/command/volume/create.go
+++ b/cli/command/volume/create.go
@@ -3,6 +3,7 @@ package volume
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/docker/cli/cli"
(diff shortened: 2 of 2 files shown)


cli/command/volume/create_test.go
```diff
@@ -192,6 +193,11 @@ func TestVolumeCreateClusterOpts(t *testing.T) {
 
 	cli := test.NewFakeCli(&fakeClient{
 		volumeCreateFunc: func(body volume.CreateOptions) (volume.Volume, error) {
+			// sort Secrets to avoid flaky test
+			secrets := body.ClusterVolumeSpec.Secrets
+			sort.SliceStable(secrets, func(i, j int) bool {
```
thaJeztah, line 198: Perhaps remove the intermediate variable to prevent confusing the `secrets` argument from the other `secrets` variable;

```suggestion
  			sort.SliceStable(body.ClusterVolumeSpec.Secrets, func(i, j int) bool {
```

thaJeztah: We should probably also update the `runCreate()` function to also sort these before making the request; https://github.com/docker/cli/blob/7f073ab823ed25ab2dfd3f800212cdab50501313/cli/command/volume/create.go#L147-L155


Wondering if we should add a `GetSlice()` (or something like that) method on `MapOpt`, which would return a sorted slice of the map; https://github.com/docker/cli/blob/c780f7c4abaf67034ecfaa0611e03695cf9e4a3e/opts/opts.go#L177-L180

thaJeztah (commented):
Thanks!

Left two suggestions / thoughts; it looks like the PR has some unrelated commits (probably from the change you were working on); could you remove those from this PR?

cli/command/volume/create_test.go
photra: I will update the `runCreate()` function as well!

As for `getSlice()`, I think it's a good idea. It will be useful in similar testing situations and also prevent flakes in the future. Should we add it to this PR?

thaJeztah (approved):
LGTM, thanks!

I did a quick squash of the commits; will merge once CI completes 👍

thaJeztah:
done! thank you!

PR #3660, opened 2022-06-06
Add env file quotes handling

### LINKED ISSUE #3630: Handle quotes in --env-file values consistently with Linux/WSL2 "source"ing
Hi there,

I use an `.env` file to supply env vars for my application both when running locally and in a Docker container.

When running my application in Linux/WSL2 (Ubuntu), I `source` the `.env` file before running.

Example:
```
set -o allexport
source ./myVars.env
set +o allexport
```

And when running my application in Docker Desktop on Windows I use the --env-file option to specify the `.env` file:
```
docker run --env-file  .\myVars.env <etc>
```

Unfortunately, some of the env vars in the `.env` file need to contain spaces. This means they must be double-quoted to work correctly when `source`ing in Linux. However when quoted, the parsing in `docker run` does not remove the quotes making the values invalid as they contain literal quote marks at beginning and end.

The consequence of this is I need to maintain a second (mostly duplicate) `.env` file for Docker use.

I would much prefer if `docker run` could parse `.env` files in the same way as my Linux `source`ing does and in line with the general shell concept that quotes are a guide to how a string should be interpreted - not part of the string itself.

From searching, it seems this has been fixed in `docker compose` (https://github.com/docker/compose/issues/8388). Please can you also fix it in `docker run`?

photra:
**- What I did**
Closes #3630 

**- How I did it**
Added quotes trimming for env files

**- How to verify it**
Run the updated unit test TestParseEnvFileGoodFile

Added --env-file quotes handling

### DIFF
--- a/opts/file.go
+++ b/opts/file.go
@@ -10,7 +10,10 @@ import (
 	"unicode/utf8"
 )
 
-const whiteSpaces = " \t"
+const (
+	whiteSpaces = " \t"
+	quotes      = "'\""
+)
 
 // ErrBadKey typed error for bad environment variable
 type ErrBadKey struct {
@@ -58,8 +61,12 @@ func parseKeyValueFile(filename string, emptyFn func(string) (string, bool)) ([]
 			}
 
 			if len(data) > 1 {
-				// pass the value through, no trimming
-				lines = append(lines, fmt.Sprintf("%s=%s", variable, data[1]))
+				// pass the value through, trimming leading and trailing quotes
+				value := data[1]
+				value = strings.TrimLeft(value, quotes)
+				value = strings.TrimRight(value, quotes)
+
+				lines = append(lines, fmt.Sprintf("%s=%s", variable, value))
 			} else {
 				var value string
 				var present bool
--- a/opts/file.go
+++ b/opts/file.go
@@ -10,10 +10,7 @@ import (
 	"unicode/utf8"
 )
 
-const (
-	whiteSpaces = " \t"
-	quotes      = "'\""
-)
+const whiteSpaces = " \t"
 
 // ErrBadKey typed error for bad environment variable
 type ErrBadKey struct {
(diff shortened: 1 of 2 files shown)


opts/file.go
```diff
@@ -58,8 +61,12 @@ func parseKeyValueFile(filename string, emptyFn func(string) (string, bool)) ([]
 			}
 
 			if len(data) > 1 {
-				// pass the value through, no trimming
-				lines = append(lines, fmt.Sprintf("%s=%s", variable, data[1]))
+				// pass the value through, trimming leading and trailing quotes
+				value := data[1]
+				value = strings.TrimLeft(value, quotes)
+				value = strings.TrimRight(value, quotes)
```
thaJeztah, line 67: We should probably make sure that we only trim the "outer" quotes, and only if "left" and "right" have the same quote; I was writing along some ideas for that, and push it as a PR against your branch https://github.com/photra/cli/pull/2

Overall, we should look if there's potential side-effects / changes in behavior because of this though 😅 - perhaps there's other expectations (such as, when using single quotes, don't expand other env-vars inside the value).

I know the original design for the env-file was explicitly to only split on `=`, and nothing else.

thaJeztah (changes requested):

opts/file.go
thaJeztah: Given that `parseKeyValueFile()` is also used in other parts of the code, we may need to check if we want the behavior to change for those areas as well

photra: That's a good point! I checked `parseKeyValueFile()` and after evaluating `readKVStrings()` and `ParseEnvFile()`, syntactically there shouldn't be side effects given their signatures.

Additionally, I read our [documentation](https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file) on `--env-file`, and it looks like we do not perform any operations on the variable or its value. Where we do this is within the [Dockerfile](https://docs.docker.com/engine/reference/builder/#run) when using shell form: 

`"When using the exec form and executing a shell directly, as in the case for the shell form, it is the shell that is doing the environment variable expansion, not docker."`

I think we should be OK on `env-file` behavior, but we should look closer for `Dockerfile` image building. I'll update here if I find anything. Let me know if you found something!

phrfpeixoto:
This is sitting here for 2 years now?

ysmood:
Any progress?

caelchristian:
> This is sitting here for 2 years now?

4 years, now. Absurd.

PR #3661, opened 2022-06-06
docs: various (minor) changes and rewording

thaJeztah:
Had a couple of these in some branches I was working on, so let's move them separately in a PR 😅 

See individual commits for details

### DIFF
--- a/docs/reference/commandline/attach.md
+++ b/docs/reference/commandline/attach.md
@@ -48,12 +48,12 @@ a container and leave it running using the `CTRL-p CTRL-q` key sequence.
 > so.
 
 It is forbidden to redirect the standard input of a `docker attach` command
-while attaching to a tty-enabled container (i.e.: launched with `-t`).
+while attaching to a TTY-enabled container (using the `-i` and `-t` options).
 
-While a client is connected to container's stdio using `docker attach`, Docker
-uses a ~1MB memory buffer to maximize the throughput of the application. If
-this buffer is filled, the speed of the API connection will start to have an
-effect on the process output writing speed. This is similar to other
+While a client is connected to container's `stdio` using `docker attach`, Docker
+uses a ~1MB memory buffer to maximize the throughput of the application.
+Once this buffer is full, the speed of the API connection is affected, and so
+this impacts the output process' writing speed. This is similar to other
 applications like SSH. Because of this, it is not recommended to run
 performance critical applications that generate a lot of output in the
 foreground over a slow client connection. Instead, users should use the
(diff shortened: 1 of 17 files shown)


thaJeztah:
@dockertopia PTAL (might be easiest to review per-commit, as each of them describes what changed 😅)

docs/reference/commandline/info.md
```diff
@@ -40,64 +40,76 @@ available on the volume where `/var/lib/docker` is mounted.
 
 ### Show output
 
-The example below shows the output for a daemon running on Red Hat Enterprise Linux,
-using the `devicemapper` storage driver. As can be seen in the output, additional
-information about the `devicemapper` storage driver is shown:
+The example below shows the output for a daemon running on ubuntu Linux,
```
vvoland, line 43: Nit: ubuntu => Ubuntu

```suggestion
The example below shows the output for a daemon running on Ubuntu Linux,
```

thaJeztah: Ah, nice catch!

Actually, let me clean up this section; it's about `--debug`, but then continues to show "a machine with swarm enabled", which is quite irrelevant for this; we could just describe; additional information, such a the `Debug` section.

dockertopia:
@thaJeztah I still have this under my radar, but have not been able to get to it yet. Sorry.

docs/reference/commandline/attach.md
```diff
@@ -48,11 +48,11 @@ a container and leave it running using the `CTRL-p CTRL-q` key sequence.
 > so.
 
 It is forbidden to redirect the standard input of a `docker attach` command
-while attaching to a tty-enabled container (i.e.: launched with `-t`).
+while attaching to a TTY-enabled container (i.e., launched with `-i` and `-t`).
 
 While a client is connected to container's stdio using `docker attach`, Docker
```
dockertopia, line 53:
```suggestion
While a client is connected to container's `stdio` using `docker attach`, Docker
```
Considering it's a file.

docs/reference/commandline/attach.md
```diff
@@ -48,11 +48,11 @@ a container and leave it running using the `CTRL-p CTRL-q` key sequence.
 > so.
 
 It is forbidden to redirect the standard input of a `docker attach` command
-while attaching to a tty-enabled container (i.e.: launched with `-t`).
+while attaching to a TTY-enabled container (i.e., launched with `-i` and `-t`).
```
dockertopia, line 51: WDYT about losing the i.e. for readability, I think it is implied in a aside like: "(using the `-i` or `-t` options)" or even "(`-i`, `-t` options).

docs/reference/commandline/attach.md
```diff
@@ -48,11 +48,11 @@ a container and leave it running using the `CTRL-p CTRL-q` key sequence.
...
 
 While a client is connected to container's stdio using `docker attach`, Docker
 uses a ~1MB memory buffer to maximize the throughput of the application. If
-this buffer is filled, the speed of the API connection will start to have an
+this buffer is filled, the speed of the API connection will have an
 effect on the process output writing speed. This is similar to other
 applications like SSH. Because of this, it is not recommended to run
 performance critical applications that generate a lot of output in the
```
dockertopia, line 58: (After removing the "If" in line 54,)
```suggestion
Once this buffer is full, the speed of the API connection is affected, and so this impacts the output process' writing speed. 
```

docs/reference/commandline/attach.md
```diff
@@ -87,45 +87,68 @@ containers, see [**Configuration file** section](cli.md#configuration-files).
...
+MiB Mem :   3934.3 total,    770.1 free,    674.2 used,   2490.1 buff/cache
+MiB Swap:   1024.0 total,    839.3 free,    184.7 used.   2814.0 avail Mem
+
+  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
+    1 root      20   0    7180   2896   2568 R   0.0   0.1   0:00.02 top
+```
 
-PID USER      PR  NI  VIRT  RES  SHR S %CPU %MEM    TIME+  COMMAND
- 1 root      20   0 17200 1116  912 R    0  0.3   0:00.03 top
+As the container was started without the `-i`, and `-t` options, signals are
```
dockertopia, line 108: Do we need to use -i and -t in conjunction? Or is the case of using one or the other?

docs/reference/commandline/attach.md
```diff
@@ -87,45 +87,68 @@ containers, see [**Configuration file** section](cli.md#configuration-files).
...
+forwarded to the attached process, which means that the default `CTRL-p CTRL-q`
+detach key sequence takes no effect, and pressing `CTRL-c` will terminate the
+container:
```
dockertopia, line 111:
```suggestion
forwarded to the attached process, which means that the default `CTRL+P`,`CTRL+Q`
detach key sequence produces no effect, but pressing `CTRL+C` terminates the
container:
```

docs/reference/commandline/attach.md
```diff
@@ -87,45 +87,68 @@ containers, see [**Configuration file** section](cli.md#configuration-files).
...
+$ docker run -dit --name topdemo2 ubuntu:22.04 /usr/bin/top -b
+```
 
-$ echo $?
-0
-$ docker ps -a | grep topdemo
+Now, when attaching to the container, and pressing the `CTRL-p CTRL-q` sequence,
+the Docker CLI is handling the detach sequence ("read escape sequence"), and the
+`attach` command is detached from the container. Checking the container's status
+with `docker ps` shows that the container is still running in the background:
```
dockertopia, line 134:
```suggestion
Now, when attaching to the container, and pressing the `CTRL+P CTRL+Q` ("read escape sequence"),
the Docker CLI is handling the detach sequence, and the
`attach` command is detached from the container. Checking the container's status
with `docker ps` shows that the container is still running in the background:
```

docs/reference/commandline/events.md
```diff
@@ -343,8 +343,8 @@ $ docker events --filter 'type=network'
 
 $ docker events --filter 'container=container_1' --filter 'container=container_2'
 
-2014-09-03T15:49:29.999999999Z07:00 container die 4386fb97867d (image=ubuntu-1:14.04)
-2014-05-10T17:42:14.999999999Z07:00 container stop 4386fb97867d (image=ubuntu-1:14.04)
+2014-09-03T15:49:29.999999999Z07:00 container die 4386fb97867d (image=ubuntu:22.04  )
+2014-05-10T17:42:14.999999999Z07:00 container stop 4386fb97867d (image=ubuntu:22.04  )
```
dockertopia, line 347:
```suggestion
2014-09-03T15:49:29.999999999Z07:00 container die 4386fb97867d (image=ubuntu:22.04)
2014-05-10T17:42:14.999999999Z07:00 container stop 4386fb97867d (image=ubuntu:22.04)
```

docs/reference/commandline/exec.md
```diff
@@ -35,12 +35,12 @@ The command started using `docker exec` only runs while the container's primary
...
-in its Dockerfile, this will be used instead.
+COMMAND runs in the default directory of the container. If the underlying image
+has a custom directory specified with the WORKDIR directive in its Dockerfile,
+this directory is used instead.
 
-COMMAND should be an executable, a chained or a quoted command
-will not work. Example: `docker exec -ti my_container "echo a && echo b"` will
+COMMAND should be an executable, a chained or a quoted command does not work.
+Example: `docker exec -ti my_container "echo a && echo b"` will
 not work, but `docker exec -ti my_container sh -c "echo a && echo b"` will.
```
dockertopia, line 44:
```suggestion
COMMAND must be an executable. A chained or a quoted command does not work.
Example: `docker exec -ti my_container sh -c "echo a && echo b"` works, but `docker exec -ti my_container "echo a && echo b"` does not.
```

docs/reference/commandline/exec.md
```diff
@@ -50,70 +50,90 @@ not work, but `docker exec -ti my_container sh -c "echo a && echo b"` will.
...
-$ docker run --name ubuntu_bash --rm -i -t ubuntu bash
+$ docker run --name mycontainer -d -i -t alpine /bin/sh
 ```
 
-This will create a container named `ubuntu_bash` and start a Bash session.
+This creates and starts a container named `mycontainer` from an `alpine` image
+with an `sh` shell as its main process. The `-d` option is set to run the container
+in the background ("detached" mode) with a pseudo-TTY attached (`-t`). The `-i`
+option is set to keep STDIN attached (`-i`), which prevents the `sh` process from
+exiting immediately.
```
dockertopia, line 61:
```suggestion
This creates and starts a container named `mycontainer` from an `alpine` image
with an `sh` shell as its main process. The `-d` (shorthand for the  `--detach`) option sets the container
to run in the background, in detached mode, with a pseudo-TTY attached (`-t`). The `-i`
option is set to keep `STDIN` attached (`-i`), which prevents the `sh` process from
exiting immediately.

```

docs/reference/commandline/exec.md
```diff
@@ -50,70 +50,90 @@ not work, but `docker exec -ti my_container sh -c "echo a && echo b"` will.
...
-Next, execute an interactive `bash` shell on the container.
+Next, execute an interactive `sh` shell on the container.
 
 ```console
-$ docker exec -it ubuntu_bash bash
+$ docker exec -it mycontainer sh
 ```
 
-This will create a new Bash session in the container `ubuntu_bash`.
+This starts a new `sh` shell session in the container `mycontainer`.
```
dockertopia, line 77:
```suggestion
This starts a new shell session in the container `mycontainer`.
```

docs/reference/commandline/exec.md
```diff
@@ -50,70 +50,90 @@ not work, but `docker exec -ti my_container sh -c "echo a && echo b"` will.
...
+
+### <a name=env></a> Set environment variables for the exec process (--env, -e)
 
 Next, set environment variables in the current bash session.
+By default `docker exec` command inherits the environment variables that are set
+when the container was created. Use the `--env` (or `-e` shorthand) to override
+environment variables, or to set additional environment variables for the process
+started by `docker exec`.
```
dockertopia, line 85:
```suggestion
By default, the `docker exec` command, inherits the environment variables that are set
at the time the container is created. Use the `--env` (or the `-e` shorthand) to override
global environment variables, or to set additional environment variables for the process
started by `docker exec`.
```

docs/reference/commandline/exec.md
```diff
@@ -50,70 +50,90 @@ not work, but `docker exec -ti my_container sh -c "echo a && echo b"` will.
...
+
+The example below creates a new shell session in the container `mycontainer` with
+environment variables `$VAR_A` and `$VAR_B` set to "1" and "2" respectively.
+These environment variables are only valid for the `sh` process started by the
+`docker exec` command, and not available to other processes running inside the
+container.
```
dockertopia, line 91:
```suggestion
These environment variables are only valid for the `sh` process started by that
`docker exec` command, and are not available to other processes running inside the
container.
```

docs/reference/commandline/exec.md
```diff
@@ -50,70 +50,90 @@ not work, but `docker exec -ti my_container sh -c "echo a && echo b"` will.
...
 
 ```console
-$ docker exec -it ubuntu_bash pwd
+$ docker exec -it mycontainer pwd
 /
 ```
 
-You can select working directory for the command to execute into
+You can specify an alternative working directory for the command to execute 
+using the `--workdir` (or `-w` shorthand) option:
```
dockertopia, line 113:
```suggestion
You can specify an alternative working directory for the command to execute 
using the `--workdir` option (or the `-w` shorthand):
```

docs/reference/commandline/pull.md
```diff
@@ -198,33 +197,35 @@ By default, `docker pull` pulls a *single* image from the registry. A repository
...
-Status: Downloaded newer image for fedora
+Status: Downloaded newer image for ubuntu
 ```
 
-After the pull has completed use the `docker images` command to see the
-images that were pulled. The example below shows all the `fedora` images
-that are present locally:
+After the pull has completed use the `docker image ls` (or `docker images` shorthand)
+command to see the images that were pulled. The example below shows all the `ubuntu`
+images that are present locally:
```
dockertopia, line 217:
```suggestion
After the pull has completed, use the `docker image ls` command (or the `docker images` shorthand) to see the images that were pulled. The example below shows all the `ubuntu`
images that are present locally:
```

docs/reference/commandline/pull.md
```diff
@@ -233,18 +234,16 @@ Killing the `docker pull` process, for example by pressing `CTRL-c` while it is
...
-> **Note**
->
-> The Engine terminates a pull operation when the connection between the Docker
-> Engine daemon and the Docker Engine client initiating the pull is lost. If the
-> connection with the Engine daemon is lost for other reasons than a manual
-> interaction, the pull is also aborted.
+The Engine terminates a pull operation when the connection between the Docker
+Engine daemon and the Docker Engine client initiating the pull is lost. If the
+connection with the Engine daemon is lost for other reasons than a manual
+interaction, the pull is also aborted.
```
dockertopia, line 249: The formulation is bit unclear. I would formulate it like the pull is aborted when: 1) connection between daemon and client is cut or lost or the operation is manually terminated.

```suggestion
The Engine terminates a pull operation when the connection between the daemon and the client (initiating the pull) is cut or lost for any reason or the command is manually terminated. 
```

man/src/container/attach.md
```diff
@@ -41,12 +41,12 @@ containers, see **docker(1)**.
 
 ## Attaching to a container
 
-In this example the top command is run inside a container, from an image called
-fedora, in detached mode. The ID from the container is passed into the **docker
-attach** command:
+In this example the top command is run inside a container, from an ubuntu image,
+in detached mode, then attaches to it, and finally terminating the container
+with `CTRL-c`:
```
dockertopia, line 46:
```suggestion
In this example the top command is run inside a container, from an ubuntu image,
in detached mode, then attaches to it, and then terminates the container
with `CTRL+C`:
```

dockertopia (changes requested):
:) This PR refreshes the versions, examples and console outputs, and improves some explanations or provides them in examples where there weren't any previously. Text is also more clear when we refrain from using the future tense.

Some comments/suggestions:

1) metadata extends/config
In the description field of the frontmatter, where is reads "How develop and read(...)" should read " How to develop and read(...)".

2) Line 35 in attach.md - couldn't make suggestion to untouched lines. 
"You can attach to the same contained process multiple times simultaneously, from different sessions on the Docker host."
I'd clarify, I'm not sure if we mean a container as a process or a process in a container, if the object of the multiple (and simultaneous) attachments is the same or not.

3) In general replaced notation key combinations capitalizing Keys and following MS style guide. 
So CTRL+P instead of CTRL-p, for example. 

4) And made inline suggestion for improving readability or clarifying text.

docs/reference/commandline/info.md
dockertopia: @thaJeztah could you please let me know if you're fine with the changes. I can push them to branch myself if that works better. Thanks.

PR #3640, opened 2022-05-31
Fix psFormat's Size handling in config file

photra:
**- What I did**
Fixed #3632 

**- How I did it**
1. Added check for `Size` usage in config file's `PsFormat`, and moved this logic before calling `containerList`
2. Added `WithSize` function in test builder to allow specifying container size in unit tests
3. Amended `TestContainerListWithConfigFormat` to test for changes above

**- How to verify it**
Run the full test suite via `docker buildx bake test`, then attempt to reproduce issue above

**- Description for the changelog**
Fixed incorrect size handling in config file when using docker ps

**- A picture of a cute animal (not mandatory but encouraged)**
image

PR #3666, opened 2022-06-07
[rfc] docker ps: always use --quiet, also combined with --format

thaJeztah:
- follow-up to https://github.com/docker/cli/pull/3640
- see the discussion on https://github.com/photra/cli/pull/1#issuecomment-1148364089

Previously, the formatter would ignore the quiet option if a custom format
was passed; this situation was handled in runPs(), where custom formats
would only be applied if the quiet option was not set, but only if the
format was set in the CLI's config.

This patch updates NewContainerFormat() to do the same, even if a `--format`
was passed on the command-line.

This is a change in behavior, so may need some discussion; possible alternatives;

- produce an error if both `--format` and `--quiet` are passed
- print a warning if both are passed (but use the logic from this patch)

Before this patch:

```console
docker ps --format '{{.Image}}'
ubuntu:22.04
alpine

docker ps --format '{{.Image}}' --quiet
ubuntu:22.04
alpine
```

With this patch applied:

```console
docker ps --format '{{.Image}}'
ubuntu:22.04
alpine

docker ps --format '{{.Image}}' --quiet
40111f61d5c5
482efdf39fac
```


vdemeester (approved):

thaJeztah:
@vvoland @rumpl @laurazard PTAL (I added the warning we discussed during one of our triage sessions)

vvoland (approved):

thaJeztah:
Thx! Let me bring this one in 👍

PR #3669, opened 2022-06-08
cli/command/info: Print OSVersion

vvoland:
Add the Info.OSVersion to the `docker info` output

Signed-off-by: Paweł Gronowski <pawel.gronowski@docker.com>

**- What I did**
Add operating system version to the `docker info` output
**- How I did it**
Print info.OSVersion
**- How to verify it**
`docker info`
Added operating system version to docker info output

### DIFF
--- a/cli/command/system/info.go
+++ b/cli/command/system/info.go
@@ -283,6 +283,7 @@ func prettyPrintServerInfo(dockerCli command.Cli, info types.Info) []error {
 
 	fprintlnNonEmpty(dockerCli.Out(), " Kernel Version:", info.KernelVersion)
 	fprintlnNonEmpty(dockerCli.Out(), " Operating System:", info.OperatingSystem)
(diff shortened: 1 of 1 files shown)


thaJeztah:
Overall, this seems fine; ISTR there was some discussion at the time about `OSVersion` and "variant" (which is what's used as part of the image-spec to resolve the right variant of an image) 🤔 Only concern I'd have is if we would ever add variant, if it would be confusing to have both.

vvoland:
OSVersion is just some numeric'ish version of the OS we are running on, like `36` for Fedora 36 or `10.0.17763` for some version of Windows Server 2019. I don't see how this would be confusing with `variant`. Or maybe you mean something else than https://github.com/opencontainers/image-spec/blob/bac4452888988085f263a61379c521abfd3472cf/specs-go/v1/descriptor.go#L68 ?

thaJeztah:
Closing for now, until we figure out what to do with this field (and what information to include in it).

PR #3685, opened 2022-06-23
Removed feminine reference

rouralberto:
This is a minor change, but the documentation was assuming a developer woman instead of just a developer using Docker.

**- What I did**
Update a minor detail in the documentation

**- How I did it**
Changing the wording

**- How to verify it**
Reading the change

The `docker run` documentation was assuming a developer woman instead of just a developer using Docker.


petariordanoo72 (commented):
Start

Issue #3681, opened 2022-06-21
Validate `docker run`'s `--pull` values


tbroyer:
**Description**

The value of the `--pull` flag on the `run` command is not validated against expected values (`missing`, `always`, `never`), this is error prone and can cause unexpected behavior *and* ignore other flags. This is particularly true given `docker build` has a `--pull` flag too with a different syntax.

We've been hit by this using a command like `docker run --pull --rm …`, which is actually being parsed as `--pull=--rm` and therefore a) not behaving as expected (we'd expected it to behave like `--pull=always`, just like `docker build --pull`) and b) ignoring/*eating* the `--rm` flag

**Steps to reproduce the issue:**
1. `docker run --pull --rm hello-world`

This step alone should be enough to reproduce (but behavior depends on context), but you can add:

2. `docker pull hello-world`
3. `docker run --pull --rm hello-world`
4. `docker rmi hello-world`
5. `docker rmi -f hello-workd`
6. `docker run --pull --rm hello-world`

**Describe the results you received:**

```
docker: Error response from daemon: No such image: hello-world:latest.
See 'docker run --help'.
```

Had I have the image locally, it would have used it without pulling it, and would have **not** removed the container.
With the above additional steps:

2. pulls the image
3. prints the hello world (but does not remove the container afterwards)
4. fails:
    ```
    $ docker rmi hello-world
    Error response from daemon: conflict: unable to remove repository reference "hello-world" (must force) - container a2824695f764 is 
    using its referenced image feb5d9fea6a5
    ```
5. removes the image
6. fails with the error above (`No such image: hello-world:latest`)

**Describe the results you expected:**

I expected this to pull the `hello-world` image (actually behaving like `--pull=always`, as we later discovered through `docker run --help`), or to tell me `--rm` is not a valid value for `--pull`.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
Client:
 Version:           20.10.17
 API version:       1.41
 Go version:        go1.18.3
 Git commit:        100c70180f
 Built:             Sat Jun 11 23:27:28 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.17
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.3
  Git commit:       a89b84221c
  Built:            Sat Jun 11 23:27:14 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
[...]

PR #3692, opened 2022-06-27
container: validate --pull option on create and run

thaJeztah:
- fixes https://github.com/docker/cli/issues/3681

Before this change, specifying the `--pull` flag without a value, could
result in the flag after it, or the positional argument to be used as
value.

This patch makes sure that the value is an expected value;

    docker create --pull --rm hello-world
    docker: invalid pull option: '--rm': must be one of "always", "missing" or "never".

    docker run --pull --rm hello-world
    docker: invalid pull option: '--rm': must be one of "always", "missing" or "never".

    docker run --pull hello-world
    docker: invalid pull option: 'hello-world': must be one of "always", "missing" or "never".

### DIFF
--- a/cli/command/container/create.go
+++ b/cli/command/container/create.go
@@ -79,6 +79,10 @@ func NewCreateCommand(dockerCli command.Cli) *cobra.Command {
 }
 
 func runCreate(dockerCli command.Cli, flags *pflag.FlagSet, options *createOptions, copts *containerOptions) error {
+	if err := validatePullOpt(options.pull); err != nil {
+		reportError(dockerCli.Err(), "create", err.Error(), true)
+		return cli.StatusError{StatusCode: 125}
+	}
 	proxyConfig := dockerCli.ConfigFile().ParseProxyConfig(dockerCli.Client().DaemonHost(), opts.ConvertKVStringsToMapWithNil(copts.env.GetAll()))
 	newEnv := []string{}
 	for k, v := range proxyConfig {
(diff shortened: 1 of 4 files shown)


thaJeztah:
@ndeloof @vvoland PTAL

cli/command/container/create.go
```diff
@@ -100,7 +100,8 @@ func runCreate(dockerCli command.Cli, flags *pflag.FlagSet, options *createOptio
 	}
 	response, err := createContainer(context.Background(), dockerCli, containerConfig, options)
 	if err != nil {
-		return err
+		reportError(dockerCli.Err(), "create", err.Error(), true)
+		return cli.StatusError{StatusCode: 125}
```
vvoland, line 104: Some tests (TestTrustedCreateFromUnreachableTrustServer, TestTrustedCreateFromBadTrustServer, TestNewCreateCommandWithContentTrustErrors) expect exit code 1 here, so they need to be adjusted to account for this change.

vvoland (changes requested):

cli/command/container/create.go
thaJeztah: Hm.. good one; thinking of that, I may have been a bit too eager to make this a `125`, as `createContainer` may produce an error for other reasons. I had this part initially elsewhere, but decided to move the validation to `createContainer`. Let me have a look

vvoland (approved):
LGTM

thaJeztah:
thx! bringing this in

PR #3693, opened 2022-06-28
cli: print full command as aliases in usage output

thaJeztah:
The default output for Cobra aliases only shows the subcommand as alias, which
is not very intuitive. This patch changes the output to print the full command
as it would be called by the user.

Note that there's still some improvements to be made; due to how aliases must be
set-up in Cobra, aliases at different "levels" are still not shown. So for example,
`docker ps --help` will not show `docker container ps` as alias, and vice-versa.
This will require additional changes, and can possibly be resolved using custom
metadata/annotations.

Before this patch:

    docker container ls --help

    Usage:  docker container ls [OPTIONS]

    List containers

    Aliases:
      ls, ps, list

After this patch:

    docker container ls --help

    Usage:  docker container ls [OPTIONS]

    List containers

    Aliases:
      docker container ls, docker container ps, docker container list

```
Usage output for command aliases now prints the full name for each alias.
```

### DIFF
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -37,6 +37,7 @@ func setupCommonRootCommand(rootCmd *cobra.Command) (*cliflags.ClientOptions, *p
 	cobra.AddTemplateFunc("hasSwarmSubCommands", hasSwarmSubCommands)
 	cobra.AddTemplateFunc("hasInvalidPlugins", hasInvalidPlugins)
 	cobra.AddTemplateFunc("topCommands", topCommands)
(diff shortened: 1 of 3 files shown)


thaJeztah:
@crazy-max @ndeloof @rumpl PTAL

(follow-up is in https://github.com/docker/cli/pull/3694, but also want to make some changes to our YAML generator)

crazy-max (approved):
SGTM

rumpl (approved):

PR #3694, opened 2022-06-28
cli: use custom annotation for aliases

thaJeztah:
- follow-up to https://github.com/docker/cli/pull/3693
- relates to https://github.com/docker/cli/pull/2389
- relates to https://github.com/docker/docker.github.io/pull/10471

Cobra allows for aliases to be defined for a command, but only allows these
to be defined at the same level (for example, `docker image ls` as alias for
`docker image list`). Our CLI has some commands that are available both as a
top-level shorthand as well as `docker <object> <verb>` subcommands. For example,
`docker ps` is a shorthand for `docker container ps` / `docker container ls`.

This patch introduces a custom "aliases" annotation that can be used to print
all available aliases for a command. While this requires these aliases to be
defined manually, in practice the list of aliases rarely changes, so maintenance
should be minimal.

As a convention, we could consider the first command in this list to be the
canonical command, so that we can use this information to add redirects in
our documentation in future.

Before this patch:

    docker images --help

    Usage:  docker images [OPTIONS] [REPOSITORY[:TAG]]

    List images

    Options:
      -a, --all             Show all images (default hides intermediate images)
      ...

With this patch:

    docker images --help

    Usage:  docker images [OPTIONS] [REPOSITORY[:TAG]]

    List images

    Aliases:
      docker image ls, docker image list, docker images

    Options:
      -a, --all             Show all images (default hides intermediate images)
      ...

### DIFF
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -31,6 +31,7 @@ func setupCommonRootCommand(rootCmd *cobra.Command) (*cliflags.ClientOptions, *p
 	opts.Common.InstallFlags(flags)
 
 	cobra.AddTemplateFunc("add", func(a, b int) int { return a + b })
+	cobra.AddTemplateFunc("hasAliases", hasAliases)
 	cobra.AddTemplateFunc("hasSubCommands", hasSubCommands)
 	cobra.AddTemplateFunc("hasTopCommands", hasTopCommands)
 	cobra.AddTemplateFunc("hasManagementSubCommands", hasManagementSubCommands)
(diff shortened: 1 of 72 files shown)


cli/cobra.go
```diff
@@ -258,6 +264,24 @@ func hasTopCommands(cmd *cobra.Command) bool {
 	return len(topCommands(cmd)) > 0
 }
 
+// commandAliases is a templating function to return aliases for the command,
+// formatted as the full command as they're called (contrary to the default
+// Aliases function, which only returns the subcommand).
+func commandAliases(cmd *cobra.Command) string {
+	if cmd.Annotations["aliases"] != "" {
```
crazy-max, line 271: Should we expose the "aliases" annotation in cobra pkg to be used in cli plugins projects? Pretty much like https://github.com/docker/cli-docs-tool/blob/main/annotation/annotation.go?

thaJeztah: Hm.. yes, so was a bit on the fence where to define it, as it's not "just" docs of course (we're also using it for the cli usage output).

Perhaps we should define a list of consts for annotations though. Was also thinking "should it have a prefix", but we already have used (e.g.) `version` for a long time, so I think that ship has sailed.

crazy-max: Ah yes indeed there is the cli usage output

tianon (approved):
:eyes:

thaJeztah:
thx! Let me bring this one in

Issue #3621, opened 2022-05-18
Performance regression on every command invocation


tonistiigi:
```
cbec75e2f3f5afd3334db693829a6b150612076c is the first bad commit
commit cbec75e2f3f5afd3334db693829a6b150612076c
Author: Nicolas De Loof <nicolas.deloof@gmail.com>
Date:   Wed Mar 30 15:27:25 2022 +0200

    Adopt Cobra completion v2 to support completion by CLI plugins

    Signed-off-by: Nicolas De Loof <nicolas.deloof@gmail.com>

 cli-plugins/manager/cobra.go        | 48 +++++++++++++++---
 cli-plugins/plugin/plugin.go        |  5 ++
...
```

After https://github.com/docker/cli/pull/3429 every command invocation takes ~200ms longer. Even when command just redirects to another process. 

Before:

```
 # time /tmp/docker buildx version
github.com/docker/buildx v0.8.0-92-g908ce2d2.m 908ce2d20618a069ddd06fc4f044505899932fe2.m
/tmp/docker buildx version  0.06s user 0.04s system 98% cpu 0.098 total
 # time /tmp/docker buildx version
github.com/docker/buildx v0.8.0-92-g908ce2d2.m 908ce2d20618a069ddd06fc4f044505899932fe2.m
/tmp/docker buildx version  0.06s user 0.03s system 99% cpu 0.095 total
 # time /tmp/docker buildx version
github.com/docker/buildx v0.8.0-92-g908ce2d2.m 908ce2d20618a069ddd06fc4f044505899932fe2.m
/tmp/docker buildx version  0.06s user 0.03s system 94% cpu 0.096 total
 # time /tmp/docker --version
Docker version unknown-version, build unknown-commit
/tmp/docker --version  0.02s user 0.01s system 71% cpu 0.038 total
 # time /tmp/docker --version
Docker version unknown-version, build unknown-commit
/tmp/docker --version  0.02s user 0.01s system 70% cpu 0.036 total
```

After:
```
github.com/docker/buildx v0.8.0-92-g908ce2d2.m 908ce2d20618a069ddd06fc4f044505899932fe2.m
/tmp/docker buildx version  0.19s user 0.10s system 89% cpu 0.322 total
 # time /tmp/docker buildx version
github.com/docker/buildx v0.8.0-92-g908ce2d2.m 908ce2d20618a069ddd06fc4f044505899932fe2.m
[...]

PR #78, opened 2017-05-12
Fix cross compile build

dnephin:
Fixes #26
Fixes #75

`-tags pkcs11` is working for the dynbinary on linux and osx. It's not working on windows. I'm not sure if we need it on windows.

The OSX cross has some warnings "was built for newer OSX version (10.11) than being linked (10.6)". I'm using the same OSX SDK as docker/docker, so I'm guessing these have existed for a while?

`linux/ppc64le` is broken because of `pkg/term`, needs a fix in `moby/moby`

`linux/s390x` is blocked on https://github.com/mitchellh/gox/pull/85. We should just run a fork of gox to support it. However we might hit the same  problem as `ppc64le` with `pkg/term`

PR #3717, opened 2022-07-23
missing DWARF symbol table

crazy-max:
Regression from https://github.com/docker/cli/pull/3320 which always strips DWARF symbol table.

I encounter this issue while trying to debug https://github.com/docker/cli/issues/3621 but also to improve our shell out operations with `debug/buildinfo` pkg (available since Go 1.18) for plugins discovery.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,12 +21,19 @@ RUN xx-apk add --no-cache musl-dev gcc
 
 FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-bullseye AS build-base-bullseye
 COPY --from=xx / /
-RUN apt-get update && apt-get install --no-install-recommends -y bash clang lld file
+RUN apt-get update && apt-get install --no-install-recommends -y bash clang lld llvm file
 WORKDIR /go/src/github.com/docker/cli
 
 FROM build-base-bullseye AS build-bullseye
 ARG TARGETPLATFORM
 RUN xx-apt-get install --no-install-recommends -y libc6-dev libgcc-10-dev
+# workaround for issue with llvm 11 for darwin/amd64 platform:
+#  # github.com/docker/cli/cmd/docker
+#  /usr/local/go/pkg/tool/linux_amd64/link: /usr/local/go/pkg/tool/linux_amd64/link: running strip failed: exit status 1
+#  llvm-strip: error: unsupported load command (cmd=0x5)
+# more info: https://github.com/docker/cli/pull/3717
+# FIXME: remove once llvm 12 available on debian
+RUN [ "$TARGETPLATFORM" != "darwin/amd64" ] || ln -sfnT /bin/true /usr/bin/llvm-strip
 
 FROM build-base-${BASE_VARIANT} AS goversioninfo
 ARG GOVERSIONINFO_VERSION
--- a/scripts/build/.variables
+++ b/scripts/build/.variables
@@ -86,6 +86,12 @@ if [ "$CGO_ENABLED" = "1" ] && [ "$GO_LINKMODE" = "static" ]; then
     GO_BUILDTAGS="$GO_BUILDTAGS osusergo"
 fi
 if [ -n "$GO_STRIP" ]; then
+    # if stripping enabled and building with llvm < 12 against darwin/amd64
+    # platform, it will fail with:
+    #  # github.com/docker/cli/cmd/docker
+    #  /usr/local/go/pkg/tool/linux_amd64/link: /usr/local/go/pkg/tool/linux_amd64/link: running strip failed: exit status 1
+    #  llvm-strip: error: unsupported load command (cmd=0x5)
+    # more info: https://github.com/docker/cli/pull/3717
     GO_LDFLAGS="$GO_LDFLAGS -s -w"
 fi
 export GO_LDFLAGS="$GO_LDFLAGS" # https://github.com/koalaman/shellcheck/issues/2064
--- a/scripts/build/.variables
+++ b/scripts/build/.variables
@@ -71,7 +71,7 @@ if [ "$CGO_ENABLED" = "1" ] && [ "$(go env GOOS)" != "windows" ]; then
 fi
 export GO_BUILDMODE
 
-GO_LDFLAGS="${GO_LDFLAGS:-} -w"
+GO_LDFLAGS="${GO_LDFLAGS:-}"
 GO_LDFLAGS="$GO_LDFLAGS -X \"github.com/docker/cli/cli/version.GitCommit=${GITCOMMIT}\""
 GO_LDFLAGS="$GO_LDFLAGS -X \"github.com/docker/cli/cli/version.BuildTime=${BUILDTIME}\""
 GO_LDFLAGS="$GO_LDFLAGS -X \"github.com/docker/cli/cli/version.Version=${VERSION}\""
@@ -86,7 +86,7 @@ if [ "$CGO_ENABLED" = "1" ] && [ "$GO_LINKMODE" = "static" ]; then
     GO_BUILDTAGS="$GO_BUILDTAGS osusergo"
 fi
 if [ -n "$GO_STRIP" ]; then
-    GO_LDFLAGS="$GO_LDFLAGS -s"
+    GO_LDFLAGS="$GO_LDFLAGS -s -w"
 fi
 export GO_LDFLAGS="$GO_LDFLAGS" # https://github.com/koalaman/shellcheck/issues/2064
 


crazy-max:
Ah ok now I remember why we removed DWARF symbol table: https://github.com/docker/cli/runs/7479620604?check_suite_focus=true#step:4:2067

```
 > [linux/amd64->darwin/arm64 build 3/3] RUN --mount=type=bind,target=.,ro     --mount=type=cache,target=/root/.cache     --mount=from=dockercore/golang-cross:xx-sdk-extras,target=/xx-sdk,src=/xx-sdk     --mount=type=tmpfs,target=cli/winresources     xx-go --wrap &&     TARGET=/out ./scripts/build/binary &&     xx-verify $([ "static" = "static" ] && echo "--static") /out/docker:
#0 1.584 Building static docker-darwin-arm64
#0 1.603 + go build -o /out/docker-darwin-arm64 -tags  osusergo pkcs11 -ldflags  -X "github.com/docker/cli/cli/version.GitCommit=f1d40cae1b" -X "github.com/docker/cli/cli/version.BuildTime=2022-07-23T07:51:49Z" -X "github.com/docker/cli/cli/version.Version=22.06.0-beta.0-56-gf1d40cae1b.m" -buildmode=pie github.com/docker/cli/cmd/docker
#67 517.7 # github.com/docker/cli/cmd/docker
#67 517.7 /usr/local/go/pkg/tool/linux_amd64/link: /usr/local/go/pkg/tool/linux_amd64/link: running dsymutil failed: exec: "dsymutil": executable file not found in $PATH
#67 517.7 
```

By default on Darwin the linker will invoke `dsymutil` to put the debug info directly into the executable. This invocation of `dsymutil` is disabled by the linker's `-w` or `-s` option :persevere:

I think our osxsdk used for cross comp darwin binaries might be too old: https://github.com/docker/cli/blob/f1615facb1ca44e4336ab20e621315fc2cfb845a/Dockerfile#L62

I will take a look

crazy-max:
Ok so we have to install the `llvm` debian package when using glibc but minimal version required with osxsdk is llvm 12 and `bullseye` is late: `llvm (1:11.0-51+nmu5)`.

I took a look at https://apt.llvm.org/ but does not seem to include `dsymutil` :disappointed::

```dockerfile
FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-bullseye AS build-base-bullseye
COPY --from=xx / /
RUN apt-get update && apt-get install --no-install-recommends -y bash clang lld file lsb-release software-properties-common
RUN wget https://apt.llvm.org/llvm.sh && chmod +x llvm.sh && ./llvm.sh 12 all && rm -f llvm.sh
WORKDIR /go/src/github.com/docker/cli
```

```
 > [linux/amd64->darwin/arm64 build 3/3] RUN --mount=type=bind,target=.,ro     --mount=type=cache,target=/root/.cache     --mount=from=dockercore/golang-cross:xx-sdk-extras,target=/xx-sdk,src=/xx-sdk     --mount=type=tmpfs,target=cli/winresources     xx-go --wrap &&     TARGET=/out ./scripts/build/binary &&     xx-verify $([ "static" = "static" ] && echo "--static") /out/docker:
#0 0.369 Building static docker-darwin-arm64
#0 0.371 + go build -o /out/docker-darwin-arm64 -tags  osusergo pkcs11 -ldflags  -X "github.com/docker/cli/cli/version.GitCommit=d02816e4fa" -X "github.com/docker/cli/cli/version.BuildTime=2022-07-23T11:44:35Z" -X "github.com/docker/cli/cli/version.Version=22.06.0-beta.0-55-gd02816e4fa.m" -buildmode=pie github.com/docker/cli/cmd/docker
#0 2.803 # github.com/docker/cli/cmd/docker
#0 2.803 /usr/local/go/pkg/tool/linux_amd64/link: /usr/local/go/pkg/tool/linux_amd64/link: running dsymutil failed: exec: "dsymutil": executable file not found in $PATH
#0 2.803
```

So I think we have to use the [`osxcross` toolchain](https://github.com/tpoechtrager/osxcross#how-does-it-work) that includes all the tooling we need. Unfortunately our `dockercore/golang-cross:xx-sdk-extras` image only includes the osxsdk but I have created an image that includes the osxcross toolchain and the osxsdk: https://github.com/crazy-max/docker-osxcross#usage and everything looks good now using both with alpine and debian.

@tonistiigi Maybe instead of the debian variant we could switch to `ubuntu:22.04` that supports llvm 12 and includes `dsymutil`: https://packages.ubuntu.com/jammy/amd64/llvm-12/filelist but `armel` cross pkgs are dropped in this release :persevere:: https://github.com/moby/moby/pull/43529#discussion_r925519608

Dockerfile
```diff
@@ -59,13 +66,25 @@ COPY --from=goversioninfo /out/goversioninfo /usr/bin/goversioninfo
...
+  set -e
+  if [ "$BASE_VARIANT" != "alpine" ] && [ "$(xx-info os)" = "darwin" ]; then
+    # on debian, use osxcross toolchain that includes the llvm-dsymutil tool
+    # and right minimal llvm version required
+    # more info: https://github.com/docker/cli/pull/3717
+    export PATH="/osxcross/bin:$PATH"
+    export LD_LIBRARY_PATH="/osxcross/lib:$LD_LIBRARY_PATH"
+    export CC="o64-clang"
+    export CXX="o64-clang++"
+  fi
```
crazy-max, line 81: @tonistiigi If you're ok with `osxcross` we could add a detection folder like `/xx-sdk` on xx so it could set itself the required env vars to use `osxcross` and we could remove this part in the Dockerfile? Maybe also the ability to support `macports` like https://github.com/crazy-max/goxx/blob/main/rootfs/usr/local/bin/goxx-macports. WDYT?

Dockerfile
```diff
@@ -6,9 +6,14 @@ ARG XX_VERSION=1.1.1
 ARG GOVERSIONINFO_VERSION=v1.3.0
 ARG GOTESTSUM_VERSION=v1.7.0
 ARG BUILDX_VERSION=0.8.2
+ARG OSXCROSS_VERSION=11.3-r7
 
 FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
 
+FROM crazymax/osxcross:${OSXCROSS_VERSION}-alpine AS osxcross-alpine
```
thaJeztah, line 13: So, is this similar to https://github.com/docker/golang-cross (from your other comments?)

Dockerfile
```diff
@@ -57,15 +63,27 @@ ARG PACKAGER_NAME
 COPY --from=goversioninfo /out/goversioninfo /usr/bin/goversioninfo
 # in bullseye arm64 target does not link with lld so configure it to use ld instead
 RUN [ ! -f /etc/alpine-release ] && xx-info is-cross && [ "$(xx-info arch)" = "arm64" ] && XX_CC_PREFER_LINKER=ld xx-clang --setup-target-triple || true
-RUN --mount=type=bind,target=.,ro \
+RUN --mount=type=bind,target=. \
```
thaJeztah, line 66: What parts need to be writable here? (could that be solved with a separate `tmpfs` mount?

thaJeztah (commented):
> By default on Darwin the linker will invoke dsymutil to put the debug info directly into the executable. This invocation of dsymutil is disabled by the linker's -w or -s option 

Is this something that could be fixed in Go?

Dockerfile
crazy-max: Oh it's just `ro` is the default behavior so don't think we need to keep it

crazy-max:
@tonistiigi Can't repro everytime but we currently have this issue with BuildKit: https://github.com/docker/cli/runs/7496406536?check_suite_focus=true#step:4:389

```
#59 [linux/amd64->windows/amd64 build 3/3] RUN --mount=type=bind,target=.     --mount=type=cache,target=/root/.cache     --mount=from=osxcross,src=/osxcross,target=/osxcross     --mount=from=osxcross,src=/osxsdk,target=/xx-sdk     --mount=type=tmpfs,target=cli/winresources <<EOT (set -e...)
#59 ERROR: failed to calculate checksum of ref m922mgc9ac45dw652zvglr6tl::w8dtv6wnagnwlcaa7arrzhowq: "/osxsdk": not found
```

But `/osxsdk` folder exists for `windows/amd64` platform:

```
$ undock --rm-dist --platform=windows/amd64 crazymax/osxcross:11.3-r7-alpine ./extract
$ tree -nh ./extract
[ 512]  ./extract/
├── [ 512]  osxcross
└── [ 512]  osxsdk

2 directories, 0 files
```

I wonder is this is not an linked to the infamous empty layer issue we had in the past or https://github.com/moby/buildkit/issues/2973.

crazy-max:
> I wonder is this is not linked to the infamous empty layer issue we had in the past or https://github.com/moby/buildkit/issues/2973.

Looking at https://github.com/docker/cli/runs/7497061435?check_suite_focus=true#step:4:330:

```
#36 [linux/amd64 goversioninfo 1/1] RUN --mount=type=cache,target=/root/.cache/go-build     --mount=type=cache,target=/go/pkg/mod     GOBIN=/out GO111MODULE=on go install "github.com/josephspurrier/goversioninfo/cmd/goversioninfo@v1.3.0"
#36 CACHED
```

It seems some instructions are cached even if no cache exporters are in place and the builder is a new one with a fresh BuildKit state in the GHA workflow.

With cache disabled everything looks good: https://github.com/docker/cli/actions/runs/2731390831

crazy-max:
Ok so after bisect we actually [strip DWARF symbol table in 20.10](https://github.com/docker/cli/blob/9d28e08a2dc7e4decf0eea363f6910fd8bde58eb/scripts/build/.variables#L17). So not a regression but this is needed for https://github.com/docker/compose-cli/pull/2151.

Needs to noop `llvm-strip` for `darwin/amd64` platform with glibc as discussed with @tonistiigi. Can't repro with alpine and LLVM 12.

scripts/build/.variables
```diff
@@ -71,7 +71,7 @@ if [ "$CGO_ENABLED" = "1" ] && [ "$(go env GOOS)" != "windows" ]; then
 fi
 export GO_BUILDMODE
 
-GO_LDFLAGS="${GO_LDFLAGS:-} -w"
+GO_LDFLAGS="${GO_LDFLAGS:-}"
```
thaJeztah, line 74: https://github.com/docker/cli/pull/3717#issuecomment-1194551026

> Ok so after bisect we actually [strip DWARF symbol table in 20.10](https://github.com/docker/cli/blob/9d28e08a2dc7e4decf0eea363f6910fd8bde58eb/scripts/build/.variables#L17). So not a regression but this is needed for [docker/compose-cli#2151](https://github.com/docker/compose-cli/pull/2151). Needs to noop `llvm-strip` for `darwin/amd64` platform with glibc as discussed with @tonistiigi. Can't repro with alpine and LLVM 12.


Did a quick search on "why" the `-w` was added by default; I see the `-w` was added in https://github.com/docker/cli/commit/a787cbc93bf6571ea7186e372bfffddf682f6f61. Before that commit, a `scripts/build/ldflags` was sourced, which didn't have the option set. The commit is part of https://github.com/docker/cli/pull/78, but that specific commit is described as only "clean up" (but actually changed options).

There's no mention _at all_ about the change, and I can only assume reviewers missed it as it was part of a "clean up" that removed `scripts/build/ldflags` in favor of inlining it.

scripts/build/.variables
```diff
@@ -86,7 +86,7 @@ if [ "$CGO_ENABLED" = "1" ] && [ "$GO_LINKMODE" = "static" ]; then
     GO_BUILDTAGS="$GO_BUILDTAGS osusergo"
 fi
 if [ -n "$GO_STRIP" ]; then
-    GO_LDFLAGS="$GO_LDFLAGS -s"
+    GO_LDFLAGS="$GO_LDFLAGS -s -w"
```
thaJeztah, line 89: As we now only set `-w` when `GO_STRIP` (which looks to be set through `STRIP_TARGET` in the bakefile) is enabled, do we still need the `llvm-strip` hack?

If "not stripping" is the fix for Darwin, could we instead 

- document this portion (and the corresponding `STRIP_TARGET`) to describe that it's not supported for Darwin binaries? (when using llvm < 12)
- or unset `-w` if the target is darwin (or make sure `STRIP_TARGET` won't be set

Dockerfile
```diff
@@ -21,12 +21,23 @@ RUN xx-apk add --no-cache musl-dev gcc
...
 FROM build-base-bullseye AS build-bullseye
 ARG TARGETPLATFORM
 RUN xx-apt-get install --no-install-recommends -y libc6-dev libgcc-10-dev
+# workaround for issue with llvm 11 for darwin/amd64 platform
+# #90 5.435 # github.com/docker/cli/cmd/docker
+# #90 5.435 /usr/local/go/pkg/tool/linux_amd64/link: /usr/local/go/pkg/tool/linux_amd64/link: running strip failed: exit status 1
+# #90 5.435 llvm-strip: error: unsupported load command (cmd=0x5)
+# remove once llvm 12 available on debian
+RUN <<EOT
+  if [ "$TARGETPLATFORM" = "darwin/amd64" ]; then
```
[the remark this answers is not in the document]
crazy-max: Yes only attended for `darwin/amd64` explicitly.

scripts/build/.variables
crazy-max: > do we still need the `llvm-strip` hack?

yes this is a bug with llvm that has been fixed in llvm 12 and only occurs with `darwin/amd64`. not sure why `llvm-strip` is always called even if `-s` and `-w` flags are not specified at all. I also can't repro that behavior with alpine base image because it uses llvm 12.

> If "not stripping" is the fix for Darwin, could we instead
> 
>     * document this portion (and the corresponding `STRIP_TARGET`) to describe that it's not supported for Darwin binaries? (when using llvm < 12)
> 
>     * or unset `-w` if the target is darwin (or make sure `STRIP_TARGET` won't be set

yes a comment lgtm, will do that.

crazy-max:
@thaJeztah Comment added. We also discussed with @tonistiigi about this and we could fallback to [strip from the cctools](https://github.com/tpoechtrager/cctools-port/blob/master/cctools/misc/strip.c) for dawrin if llvm upstream doesn't work. this could be bundled in xx with ld64 so we could remove this hack.

scripts/build/.variables
thaJeztah: > not sure why llvm-strip is always called even if -s and -w flags are not specified at all.

Ah, right; so removing `-s` and/or `-w` for `darwin/amd64` won't work around the issue; gotcha.

thaJeztah (approved):
LGTM

thaJeztah:
@tonistiigi this LGTY?

tonistiigi (approved):
We should look more deeply into what that darwin issue is. Is it something broken in old LLVM etc. If needed we can get the Apple `strip` from https://github.com/tpoechtrager/cctools-port/blob/master/cctools/misc/strip.c with `ld64` in `xx`.

PR #3721, opened 2022-07-29
context use: don't create/update config file and directories if not needed

thaJeztah:
- relates to https://github.com/docker/cli/pull/3668

### context use: skip validation for "default" context

This code was handling validation and parsing, only to discard the results if it was the default context.

### context use: don't create/update config file and directories if not needed

Avoid updating the config-file if nothing changed. This also prevents creating
the file and config-directory if the default is used and no config-file existed
yet.

`config.Save()` performs various steps (creating the directory, updating
or copying permissions, etc etc), which are not needed if the defaults are
used; https://github.com/docker/cli/blob/a445d97c2536f0de37469d0ea9881288d6c49cbf/cli/config/configfile/file.go#L135-L176

```
The client no longer creates or updates the CLI config file when running `docker context use` and the selected context is the current context.
```

### DIFF
--- a/cli/command/context/use_test.go
+++ b/cli/command/context/use_test.go
@@ -2,6 +2,9 @@ package context
 
 import (
 	"bytes"
+	"errors"
+	"io"
+	"os"
 	"path/filepath"
 	"testing"
 
@@ -10,6 +13,7 @@ import (
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/cli/cli/context/store"
 	"github.com/docker/cli/cli/flags"
+	"github.com/docker/docker/pkg/homedir"
 	"gotest.tools/v3/assert"
 	is "gotest.tools/v3/assert/cmp"
 )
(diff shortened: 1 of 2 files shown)


thaJeztah:
@nicks @rumpl @vvoland PTAL

cli/command/context/use_test.go
```diff
@@ -46,6 +50,35 @@ func TestUseNoExist(t *testing.T) {
...
+	// Verify config-dir and -file don't exist before
+	_, err := os.Stat(configDir)
+	assert.Check(t, errors.Is(err, os.ErrNotExist))
+	_, err = os.Stat(configFilePath)
+	assert.Check(t, errors.Is(err, os.ErrNotExist))
+
+	cli, err := command.NewDockerCli(command.WithCombinedStreams(io.Discard))
+	assert.NilError(t, err)
+	assert.NilError(t, newUseCommand(cli).RunE(nil, []string{"default"}))
+	assert.NilError(t, err)
```
vvoland, line 73: Redundant last check

vvoland (approved):

cli/command/context/use_test.go
thaJeztah: nice catch; sloppy copy/paste 😅 

fixed!

thaJeztah:
All green; let's get this one in

Issue #3652, opened 2022-06-03
cli hangs on dead connection


nicks:
**Description**

**Steps to reproduce the issue:**
1. Create a socket file with a dead listener by running:

```
nc -lkU /home/nick/test.sock
```

2. Run any docker command against the socket

```
DOCKER_HOST="unix:///home/nick/test.sock" docker ps
```

**Describe the results you received:**

The CLI should eventually time out and/or error

**Describe the results you expected:**

The CLI hangs indefinitely.

**Additional information you deem important (e.g. issue happens only occasionally):**

This is distilled from a longer user report and stack trace upstream:
https://github.com/tilt-dev/tilt/issues/5841

**Output of `docker version`:**

```
docker version
Client: Docker Engine - Community
 Cloud integration: v1.0.24
 Version:           20.10.16
 API version:       1.41
 Go version:        go1.17.10
 Git commit:        aa7e414
 Built:             Thu May 12 09:17:23 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**
Linux

nicks:
hmmmm....this is still reproducible for me, but i thought we fixed it

i see a stack trace like:

```
	/usr/local/go/src/net/http/client.go:590
github.com/docker/cli/vendor/github.com/docker/docker/client.(*Client).doRequest(0xc000258b40, 0xc000258b40?)
	/go/src/github.com/docker/cli/vendor/github.com/docker/docker/client/request.go:140 +0x7c fp=0xc00063f0b8 sp=0xc00063efd0 pc=0x5609e8a2a55c
github.com/docker/cli/vendor/github.com/docker/docker/client.(*Client).Ping(0xc000258b40, {0x5609e9506940, 0x5609e9d51840})
	/go/src/github.com/docker/cli/vendor/github.com/docker/docker/client/ping.go:28 +0x145 fp=0xc00063f390 sp=0xc00063f0b8 pc=0x5609e8a253a5
github.com/docker/cli/vendor/github.com/docker/docker/client.(*Client).NegotiateAPIVersion(0xc000258b40, {0x5609e9506940?, 0x5609e9d51840?})
	/go/src/github.com/docker/cli/vendor/github.com/docker/docker/client/client.go:310 +0x45 fp=0xc00063f468 sp=0xc00063f390 pc=0x5609e8a0a505
[...]

PR #2303, opened 2020-01-30
revert "connhelper: use ssh multiplexing"

thaJeztah:
this reverts https://github.com/docker/cli/pull/2132, which looks to be responsible for flakiness in CI; 

closes https://github.com/docker/cli/issues/2256
closes https://github.com/docker/cli/issues/2257 

https://ci.docker.com/public/blue/organizations/jenkins/cli/detail/PR-2132/3/pipeline

```
=== Failed
=== FAIL: e2e/image TestBuildFromContextDirectoryWithTag (0.32s)
    build_test.go:37: assertion failed: 
        Command:  docker build -t myimage .
        ExitCode: 1
        Error:    exit status 1
        Stdout:   Sending build context to Docker daemon  4.608kB

        
        Stderr:   error during connect: Post http://docker/v1.40/build?buildargs=%7B%7D&cachefrom=%5B%5D&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&labels=%7B%7D&memory=0&memswap=0&networkmode=default&rm=1&shmsize=0&t=myimage&target=&ulimits=null&version=1: command [ssh -o ControlMaster=auto -o ControlPath=/root/.docker/%r@%h:%p -l penguin 172.20.0.2 -- docker system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=
        
        
        Failures:
        ExitCode was 1 expected 0
        Expected stderr to contain "[NOTHING]"
        Expected no error

=== FAIL: e2e/image TestPullWithContentTrustUsesCacheWhenNotaryUnavailable (3.39s)
    pull_test.go:51: assertion failed: 
        Command:  docker image rm registry:5000/trust-pull-unreachable:latest
        ExitCode: 1
        Error:    exit status 1
        Stdout:   
        Stderr:   error during connect: Delete http://docker/v1.40/images/registry:5000/trust-pull-unreachable:latest: command [ssh -o ControlMaster=auto -o ControlPath=/root/.docker/%r@%h:%p -l penguin 172.20.0.2 -- docker system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=
        
        
        Failures:
        ExitCode was 1 expected 0
        Expected no error
```

PR #3722, opened 2022-07-29
cli: set timeout connection ping on sockets as well

thaJeztah:
- carries https://github.com/docker/cli/pull/3663 (rebased and squashed)

Note that this does not fully fix the referenced issue, but
at least makes sure that API clients don't hang forever on
the initialization step.

See: https://github.com/docker/cli/issues/3652

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -35,6 +35,8 @@ import (
 	notaryclient "github.com/theupdateframework/notary/client"
 )
 
+const defaultInitTimeout = 2 * time.Second
+
 // Streams is an interface which exposes the standard input and output streams
 type Streams interface {
 	In() *streams.In
@@ -77,6 +79,7 @@ type DockerCli struct {
 	currentContext     string
 	dockerEndpoint     docker.Endpoint
 	contextStoreConfig store.Config
+	initTimeout        time.Duration
 }
 
 // DefaultVersion returns api.defaultVersion.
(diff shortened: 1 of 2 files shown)


cli/command/cli.go
```diff
@@ -313,13 +316,20 @@ func resolveDefaultDockerEndpoint(opts *cliflags.CommonOptions) (docker.Endpoint
...
+	}
+	return defaultInitTimeout
+}
+
 func (cli *DockerCli) initializeFromClient() {
 	ctx := context.Background()
-	if strings.HasPrefix(cli.DockerEndpoint().Host, "tcp://") {
+	if !strings.HasPrefix(cli.DockerEndpoint().Host, "ssh://") {
 		// @FIXME context.WithTimeout doesn't work with connhelper / ssh connections
 		// time="2020-04-10T10:16:26Z" level=warning msg="commandConn.CloseWrite: commandconn: failed to wait: signal: killed"
```
vvoland, line 330: I know it isn't a part of this PR, but maybe you remember how does context.WithTimeout has anything to do with ssh being used? The PR that introduced this doesn't mention any details here.

vvoland: Just curious - so if you remember then just some quick tldr will be enough 😄

vvoland (approved):

cli/command/cli.go
thaJeztah: I'm not sure how much time I spent at the time digging into the exact cause; the timeout was helping with an issue (see https://github.com/docker/cli/pull/2424), and I'm not sure I was able to consistently reproduce the ssh problem locally (or only in CI?). Ultimately decided that the SSH connections were less common, so to keep diving into the issue for later (there's some other issues remaining with the ssh connection helper as well).

In case useful;

- ssh implementation; https://github.com/docker/cli/pull/1014
- multiplexing change that we reverted due to CI issues (but should look into to have it merged again) https://github.com/docker/cli/pull/2303

thaJeztah: It may be just missing functionality in the ssh connecting helper (it not handling context cancelation), but yeah, never found the time really to get myself familiar with that code or to dig into the issue.

vvoland: Thanks!

thaJeztah:
Let me bring this one in

WillisZhang:
I don't know why docker login alway fail! 
since last year , I had to build / pull docker image in a Amazon vhost , and docker save to a tar-ball , and download the tarball, and docker load , into the local env.
I am really stumbled by this mechanism of docker login,. why just pull a image , need to login ?!
and over 1 year, it has nothing changed.

WillisZhang:
is this a way to deloy a server side component ?

WillisZhang:
docker company's business model is so stupid.

thaJeztah:
@WillisZhang you're commenting on a 3 Year old pull request; if you suspect there's a bug at hand that's causing login to fail, please open a ticket instead, and provide details that are requested in the ticket template, and steps to reproduce.

PR #3724, opened 2022-08-01
display port name field in 'service ls'

aep:
**- What I did**

in `docker service ls` we durrently format the port as "*:8080->9000" with the asterisk being hardcoded,
it would be useful if the api server could return a dns name on the port, for api servers that can expose ports directly as ingress. 

**- How I did it**

the least invasive way to implement that appears to be to use the existing Name field.
We could add an additional field, if that's easy to upstream. That would have the benefit of not changing anything unless the new field is set.

**- How to verify it**

`docker service ls`

display port name field in 'service ls'

stickers


vvoland (commented):
Thanks, looks good to me!
The linter check is failing though, please give `gofmt` a run :)

Looks like a good improvement over a previous behaviour. Not sure if this is something that shouldn't change due to backwards compatibility though. But since this is CLI and not Engine API - I think it's fine.
 @ndeloof  @thaJeztah  WDYT?

thaJeztah:
I'm not sure if the information shown is accurate; the `*:` is printed to indicate for services that they're published on the ingress network, and because of that listening on all IP addresses of the swarm cluster. IIUC, in your case, there's additional code or an alternative API that uses this information to register services in a DNS, and those services (and ports) to be accessible using that, which would not be the default behavior, and exposing the `Name` here would not work for regular uses.


The `Name` field here originates from SwarmKit (added in https://github.com/moby/swarmkit/pull/353, but corresponding changes appear to have been in a separate PR), and is described as;

```
// Name for the port. If provided the port information can
// be queried using the name as in a DNS SRV query.
```

aep:
>  and because of that listening on all IP addresses of the swarm cluster. 

yes. this change would let the cluster specifically indicate that it's not all IP addresses, but a specific one

> exposing the Name here would not work for regular uses.

not sure i understand what you mean by that. The existing behavior will not change. At least i haven't seen Name actually be used.

Of course if you're suggesting you would prefer the alternative option of adding a new field, i could do that.

Issue #3727, opened 2022-08-03
Docs example error: `Template parsing error`


maxmorozoff:
**Description**

I need to [List all port bindings](https://docs.docker.com/engine/reference/commandline/inspect/#list-all-port-bindings) in Docker so I found an example in documentation but in my case it throws an error.

**Steps to reproduce the issue:**

```console
$ docker run --rm -itd --name documentation-issue mysql:5.7 bash
d24af5958f714cede788ae7db04abae2fc4e515d8fa32ce038583eebe8a9d540
$ docker inspect format='{{range $p, $conf := .NetworkSettings.Ports}}{{$p}} -> {(index $conf 0).HostPort}}{{end}}' documentation-issue
``` 
To stop container, execute:

```console
$ docker stop documentation-issue 
documentation-issue
```

**Describe the results you received:**

```
Template parsing error: template: :1:75: executing "" at <index $conf 0>: error calling index: index of untyped nil
```

**Describe the results you expected:**

In go templates there is [`with` action](https://pkg.go.dev/text/template#hdr-Actions), using that everything works as expected:

```bash
$ docker inspect --format='{{range $p, $conf := .NetworkSettings.Ports}} {{$p}} -> {{with $conf}}{{(index . 0).HostPort}}{{else}}none{{end}} {{end}}' documentation-issue
mysql:5.7 3306/tcp -> none  33060/tcp -> none 
```

```bash
$ docker inspect --format='{{range $p, $conf := .NetworkSettings.Ports}} {{$p}} -> {{with $conf}}{{(index . 0).HostPort}}{{end}} {{end}}' documentation-issue
mysql:5.7 3306/tcp ->   33060/tcp ->  
```

**Output of `docker version`:**

```
Client:
 Version:           20.10.17
 API version:       1.41
 Go version:        go1.17.11
 Git commit:        100c701
 Built:             Mon Jun  6 22:56:42 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.17
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.17.11
  Git commit:       a89b842
  Built:            Mon Jun  6 23:01:23 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.6
  GitCommit:        10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc:
  Version:          1.1.2
[...]

PR #3728, opened 2022-08-03
docs: Fix template error in cli example (#3727)

maxmorozoff:
Fixes issue #3727 

```
Template parsing error: template: :1:75: executing "" at <index $conf 0>: error calling index: index of untyped nil
```

### DIFF
--- a/docs/reference/commandline/inspect.md
+++ b/docs/reference/commandline/inspect.md
@@ -107,7 +107,7 @@ $ docker inspect --format='{{.Config.Image}}' $INSTANCE_ID
 You can loop over arrays and maps in the results to produce simple text output:
 
 ```console
-$ docker inspect --format='{{range $p, $conf := .NetworkSettings.Ports}} {{$p}} -> {{(index $conf 0).HostPort}} {{end}}' $INSTANCE_ID
+$ docker inspect --format='{{range $p, $conf := .NetworkSettings.Ports}} {{$p}} -> {{with $conf}}{{(index . 0).HostPort}}{{else}}none{{end}} {{end}}' $INSTANCE_ID
 ```
 
 ### Find a specific port mapping


thaJeztah:
Hmm... right, so I think the problem in your case was that you were using the example on a container that did not map any ports (i.e., no `-p` or `-P` options were set). In that case the template fails as it tries to iterate over an empty (or missing) slice.

For example;

```bash
docker run -d --name foo nginx:alpine
6757f08870ace82b937eecadda2e8d689cf1ea4eb263af65afa8add4576ad953

docker inspect --format='{{range $p, $conf := .NetworkSettings.Ports}} {{$p}} -> {{(index $conf 0).HostPort}} {{end}}' foo

Template parsing error: template: :1:59: executing "" at <index $conf 0>: error calling index: index of untyped nil
```

Doing the same but with a container that _does_ publish ports will show the expected output;


```bash
docker run -d --name foo2 -p 8080:80 -p 8081:443 nginx:alpine
5788ced50b6fb9ea726d6ffd36403fcb84a981e4ebc7950f9288bd3562fd7200

docker inspect --format='{{range $p, $conf := .NetworkSettings.Ports}} {{$p}} -> {{(index $conf 0).HostPort}} {{end}}' foo2
 443/tcp -> 8081  80/tcp -> 8080
```

I'm a bit confused about the addition of `{{.Config.Image}}` in the template though 🤔

maxmorozoff:
> I'm a bit confused about the addition of `{{.Config.Image}}` in the template though

You are right, it's not working with `{{.Config.Image}}`, I'm sorry, my fault.
Copy-pasted this from the wrong part of my bash script.

It should be like this:

```console
$ docker inspect --format='{{range $p, $conf := .NetworkSettings.Ports}} {{$p}} -> {{with $conf}}{{(index . 0).HostPort}}{{else}}none{{end}} {{end}}' foo
3306/tcp -> none  33060/tcp -> none
```

In this case it will not throw an error in your first example and will give the same output as in your second example.

I'm going to fix my copy-pasting error in pr soon. Sorry about that. And thanks for your work!

thaJeztah:
Sorry for the delay; thank you!  The change itself looks good, but some issues with the commit(s).

- Could you squash the two commits? (let me know if you need detailed instructions)
- It appears your commit message is missing a DCO sign-off,

```
```

commit.

    ```bash
    ```

2. Clone your fork locally

    ```bash
    ```

maxmorozoff:
I'm sorry for the delay..

I squashed and signed off my commit. 
Hope that it's ok now.

Thanks!

thaJeztah (approved):
LGTM

vvoland (approved):

PR #3744, opened 2022-08-18
ci: update github actions

crazy-max:
update github actions to latest major releases and also pin runner os version to `ubuntu-20.04` (current latest: https://github.com/actions/runner-images#available-images)

also:
* adds dependabot config to just keep gha actions up to date
* adds concurrency check

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "area/testing"
+      - "status/2-code-review"
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,16 +1,6 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
-name: "CodeQL"
+name: codeql
 
 on:
-#  push:
-#    branches: [master]
-#  pull_request:
-#    # The branches below must be a subset of the branches above
-#    branches: [master]
   schedule:
     #        ┌───────────── minute (0 - 59)
     #        │ ┌───────────── hour (0 - 23)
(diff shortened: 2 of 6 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "dependencies"
+      - "bot"
```
crazy-max, line 9: @thaJeztah I use arbitrary labels. Let me know what we should use.

.github/workflows/codeql-analysis.yml
```diff
@@ -1,80 +1,31 @@
...
   schedule:
-    #        ┌───────────── minute (0 - 59)
-    #        │ ┌───────────── hour (0 - 23)
-    #        │ │ ┌───────────── day of the month (1 - 31)
-    #        │ │ │ ┌───────────── month (1 - 12)
-    #        │ │ │ │ ┌───────────── day of the week (0 - 6) (Sunday to Saturday)
-    #        │ │ │ │ │
-    #        │ │ │ │ │
-    #        │ │ │ │ │
-    #        * * * * *
```
thaJeztah, line 23: Is it ok to keep this one?

.github/workflows/test.yml
```diff
@@ -41,8 +45,8 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-latest
-#          - windows-latest # FIXME: some tests are failing on the Windows runner, as well as on Appveyor since June 24, 2018: https://ci.appveyor.com/project/docker/cli/history
+          - macos-11
+#          - windows-2022 # FIXME: some tests are failing on the Windows runner, as well as on Appveyor since June 24, 2018: https://ci.appveyor.com/project/docker/cli/history
```
thaJeztah, line 49: Perhaps we should have a draft PR with this enabled (to see what needs fixing)

.github/dependabot.yml
thaJeztah: Perhaps the `bot` label is not needed.

If this is only updating the CI actions, perhaps;

- `area/testing` would be ok (I tend to use that for fixes in tests and CI)
- `status/2-code-review`


Do you know if there's now a way to make dependant only run on this repository, and not on forks? (e.g. can we restrict this on repository name?)

thaJeztah (changes requested):

.github/dependabot.yml
crazy-max: > Do you know if there's now a way to make dependant only run on this repository, and not on forks? (e.g. can we restrict this on repository name?)

Doesn't look like it :disappointed:
* https://github.com/dependabot/dependabot-core/issues/2198
* https://github.com/dependabot/dependabot-core/issues/2804

thaJeztah: I was hoping because this one is a schedule; perhaps we could disable the schedule if repo != docker/cli (which would be slightly different than a default dependabot check?). But yeah, if it's not possible, then, well, perhaps it is what it is (and it'll just open 1500 useless pull requests in all forks 😂)

thaJeztah (approved):
LGTM

PR #3755, opened 2022-08-27
replace pkg/system Sequential funcs with moby/sys/sequential

thaJeztah:
- depends on https://github.com/moby/sys/pull/121
- relates to https://github.com/moby/moby/pull/44051
- relates to https://github.com/containerd/containerd/pull/7334

Migrating these functions to allow them being shared between moby, docker/cli,
and containerd, and to allow using them without importing all of sys / system,
which (in containerd) also depends on hcsshim and more.

### DIFF
--- a/cli/command/config/create.go
+++ b/cli/command/config/create.go
@@ -10,7 +10,7 @@ import (
 	"github.com/docker/cli/cli/command/completion"
 	"github.com/docker/cli/opts"
 	"github.com/docker/docker/api/types/swarm"
-	"github.com/docker/docker/pkg/system"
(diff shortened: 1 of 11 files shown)


thaJeztah:
For reviewers looking at the "+ LOC";

 Screenshot 2022-08-27 at 15 42 40


- most of the added lines from the new module is because it'll vendor another copy of the Apache LICENSE
- this should all be temporary; once other dependencies have migrated to this new package, the extra copy of these functions should go away

sam-thibault (approved):
LGTM

thaJeztah:
Thx! Bringing this one in 👍

PR #3757, opened 2022-08-29
cli: use DisplayJSONMessagesToStream

thaJeztah:
This utility function is a wrapper for DisplayJSONMessagesStream, and
already used in various other places, so updating this for consistency.

FWIW; we should look if this function is actually better located inside the CLI code, as only the CLI uses this

### DIFF
--- a/cli/command/container/create.go
+++ b/cli/command/container/create.go
@@ -199,10 +199,9 @@ func createContainer(ctx context.Context, dockerCli command.Cli, containerConfig
 	config := containerConfig.Config
 	hostConfig := containerConfig.HostConfig
 	networkingConfig := containerConfig.NetworkingConfig
-	stderr := dockerCli.Err()
 
-	warnOnOomKillDisable(*hostConfig, stderr)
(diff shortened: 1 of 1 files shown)


cli/command/container/create.go
```diff
@@ -141,12 +142,7 @@ func pullImage(ctx context.Context, dockerCli command.Cli, image string, platfor
 	}
 	defer responseBody.Close()
 
-	return jsonmessage.DisplayJSONMessagesStream(
-		responseBody,
-		out,
-		dockerCli.Out().FD(),
-		dockerCli.Out().IsTerminal(),
-		nil)
+	return jsonmessage.DisplayJSONMessagesToStream(responseBody, streams.NewOut(out), nil)
```
thaJeztah, line 145: Hmm... actually this may not work, because `out` can be overridden to be `io.discard` if `--quiet` is set, and terminal detection would likely not be correct in that case, so detection must be done through the CLI's `Out()`, not `out` passed here.

PR #3758, opened 2022-08-29
feat: add 'docker containers' alias to match the 'docker images' logic

piotrostr:
Signed-off-by: Piotr Ostrowski <piotrostr@google.com>

Added another alias for `docker container ls` command. 

With an already existing `docker images`, `docker containers` would be a nice alias to match the logic and make a consistent command line interface.


thaJeztah:
closing, because the plural form of `docker images` is a legacy command; the newer `docker <object> <verb>` syntax is the canonical form, and we don't want to introduce new commands / aliases using the legacy form.

PR #3761, opened 2022-08-31
cli/command: NewDockerCli(): use WithStandardStreams()

thaJeztah:
`NewDockerCli` was configuring the standard streams using local code; this patch
instead uses the available `WithStandardStreams()` option to do the same.

There is slight difference in the order of events;

Previously, user-provided options would be applied first, after which NewDockerCli
would check if any of "in", "out", or "err" were nil, and if so set them to the
default stream (or writer) for that output.

The new code unconditionally sets the defaults _before_ applying user-provided
options. In practive, howver, this makes no difference; the fields set are not
exported, and the only functions updating them are `WithStandardStreams`,
`WithInputStream`, and `WithCombinedStream`, neither of which checks the old
value (so always overrides).

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -29,7 +29,6 @@ import (
 	"github.com/docker/docker/api/types/swarm"
 	"github.com/docker/docker/client"
 	"github.com/docker/go-connections/tlsconfig"
-	"github.com/moby/term"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


rumpl (approved):
Nice!

thaJeztah:
Yes, I seem to recall I had another branch somewhere doing more things like this (may have been for "contexts") but forgot which one it was 😂 . It'll probably pop-up some time, and I'll open a PR

Issue #2995, opened 2021-03-03
Improve UX for images created "40 years ago"


imjasonh:
Tools like [`ko`](https://github.com/google/ko), [Bazel's rules_docker](https://github.com/bazelbuild/rules_docker) and [CNCF Buildpacks](https://buildpacks.io) (and many others) build container images reproducibly by default, and as such set the creation date of the image as `1970-01-01T00:00:00Z`, according to advice from https://reproducible-builds.org.

This can lead to confusion among users consuming images built by these tools when they use `docker inspect` (etc) and see that it happily prints that the images were created "`40 years ago`". Each project above has some documentation describing why users shouldn't be worried/confused by this ([example](https://buildpacks.io/docs/reference/reproducibility/), [example](https://github.com/google/ko#why-are-my-images-all-created-in-1970)), but ultimately it's a band-aid.

Since, short of a time machine, it's incredibly unlikely that a container image was _actually_ built in 1970, I think it could make sense for `docker` to interpret `1970-01-01T00:00:00Z` as a special value and print some sentinel string in the output of `docker history` (etc), instead of `40 years ago`. That string could be something like `unset`, `zero`, `omitted`, `reproducible`, or honestly basically anything that when googled led to a page on docker.com describing why that value might be unset. (`-H=false` should still show the full timestamp)

The code change seems relatively simple, here:
https://github.com/docker/cli/blob/70a00157f161b109be77cd4f30ce0662bfe8cc32/cli/command/image/formatter_history.go#L89

Has this been discussed and rejected before? I couldn't find any issues related to this topic.

jchester:
A quick nitpick: buildpacks use 1980 as their source epoch date.

[This is because zipfiles, notably including JARs, have 1st January 1980 as their epoch date](https://github.com/buildpacks/pack/issues/130).

ImJasonH:
> A quick nitpick: buildpacks use 1980 as their source epoch date.

It's true. Buildpacks produces images built "`41 years ago`", while `ko` etc produce images built "`51 years ago`", for roughly the same reasons. Thanks for pointing that out @jchester 

I'd like to treat both of these dates (`1980-01-01T00:00:01Z` and `1970-01-01T00:00:00Z`) as special, and show some other string instead of `XX years ago`.

[...]

PR #3778, opened 2022-09-16
cli: hide old CREATED times, instead of "52 years ago"

imjasonh:
Fixes https://github.com/docker/cli/issues/2995

**- What I did**

Changed formatting code for the `CREATED` column to display no text if the date being formatted is older than the year 2000. This value was chosen because it is (currently) unlikely that an image was actually created before that date.

**- How I did it**

I updated `formatter_history.go` to handle this new case when displaying dates for human consumption (the default).

**- How to verify it**

I added unit tests, which pass in my environment. I'm happy to add any other tests that reviewers may find useful.

Hide `CREATED` dates that are older than the year 2000, where normal relative dates (e.g., "52 years ago") are likely misleading.

tinyoctopus

### DIFF
--- a/cli/command/image/formatter_history.go
+++ b/cli/command/image/formatter_history.go
@@ -82,9 +82,15 @@ func (c *historyContext) CreatedAt() string {
 	return time.Unix(c.h.Created, 0).Format(time.RFC3339)
 }
 
+// epoch is the time before which created-at dates are not displayed with human units.
+var epoch = time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC).Unix()
+
 func (c *historyContext) CreatedSince() string {
 	if !c.human {
 		return c.CreatedAt()
+	}
+	if c.h.Created <= epoch {
+		return ""
 	}
 	created := units.HumanDuration(time.Now().UTC().Sub(time.Unix(c.h.Created, 0)))
 	return created + " ago"
--- a/cli/command/image/formatter_history_test.go
+++ b/cli/command/image/formatter_history_test.go
@@ -70,6 +70,30 @@ func TestHistoryContext_CreatedSince(t *testing.T) {
 				human: false,
 			}, dateStr, ctx.CreatedSince,
 		},
+		{
+			// The zero time is not displayed.
+			historyContext{
+				h:     image.HistoryResponseItem{Created: 0},
+				trunc: false,
+				human: true,
+			}, "", ctx.CreatedSince,
+		},
+		{
+			// A time before the year 2000 is not displayed.
+			historyContext{
+				h:     image.HistoryResponseItem{Created: time.Date(1980, time.November, 10, 10, 23, 0, 0, time.UTC).Unix()},
+				trunc: false,
+				human: true,
+			}, "", ctx.CreatedSince,
+		},
+		{
+			// A time after 2000 is displayed.
+			historyContext{
+				h:     image.HistoryResponseItem{Created: time.Date(2005, time.September, 27, 4, 37, 0, 0, time.UTC).Unix()},
+				trunc: false,
+				human: true,
+			}, "", ctx.CreatedSince,
+		},
 	}
 
 	for _, c := range cases {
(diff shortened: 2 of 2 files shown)


cli/command/image/formatter_history.go
```diff
@@ -82,10 +82,16 @@ func (c *historyContext) CreatedAt() string {
 	return time.Unix(c.h.Created, 0).Format(time.RFC3339)
 }
 
+// epoch is the time before which created-at dates are not displayed with human units.
+var epoch = time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC).Unix()
```
thaJeztah, line 86: Anything before [March 15, 2013](https://pyvideo.org/pycon-us-2013/the-future-of-linux-containers.html) would be unlikely, so I guess we're safe with 2000 yes 😄

thaJeztah (approved):
LGTM

@justincormack @rumpl PTAL

rumpl (approved):
🎉

cli/command/image/formatter_history.go
tianon: I've got several container images where I intentionally set the date to something meaningful but still before 2013 -- how many of these "really old" dates are there?  Would it be tenable to have a small list of exceptions instead of hard-coding a specific "cut-off" date?

(Also for consistency, shouldn't it be replaced with `N/A` instead of the empty string?)

tonistiigi: Why do we need to hide anything but zero value at all? What is the case where non-zero value is set but should be not shown?

imjasonh: > Why do we need to hide anything but zero value at all? What is the case where non-zero value is set but should be not shown?

Some builders use 1980 as the magical reproducibility date (see https://github.com/docker/cli/issues/2995#issuecomment-790055255). Other than that I don't know any other magic values that denote special dates, but there may be more.

Rather than interpreting this change as "hide special dates chosen for reproducibility" I think this change is more about hiding values of "created" that clearly don't represent the actual date the image was created. If you actually created a Docker image on July 3, 1985, it's only because you had access to a time machine, which you're using for very mundane purposes. 😆

That being said, I'm happy to make whatever change to this will upset the fewest folks. The only date I really care about is 1/1/1970 (and 1980 for the buildpacks folks' sake). If the consensus is that we should special-case those dates, that's okay too.

(I can also change this to `N/A` if that's the consensus)

Thanks for merging this @thaJeztah ! 🙏

tonistiigi: > it's only because you had access to a time machine

You don't need a time machine to set that date. Any image author can set this date to any value they like. My assumption is that if they set it to some weird value they had some reason for it, and that reason can't be that they didn't want this date to ever show up because then they would have used zero value instead. And if they accidentally set it to a weird value it is also beneficial to show it so that they can notice the bug and fix it.

imjasonh: Fair enough.

However, if they set a specific meaningful date, they presumably want to see that the date is _exactly_ that date and not somewhere near that date. In that case, they probably want to check the full timestamp string, instead of `37 years ago`, which will become `38 years ago` eventually. Displaying the full timestamp string is unchanged by this PR.

dims:
release note should say "don't use tools like awk to parse 'docker ps'" :)

imjasonh:
> release note should say "don't use tools like awk to parse 'docker ps'" :)

I think @tianon 's suggestion to use `N/A` instead is the solution to this. I'll send a followup PR for that.

cli/command/image/formatter_history.go
kaniini: FWIW, 2000-01-01 was just a date I picked out of whimsy for the epoch, and has no specific importance other than that it looked like a good date, based on the assumption that any legacy epoch dates of interest would likely fall before then.  We could easily just exempt 1970-01-01, 1980-01-01 and other epoch dates if people would rather do that :)

PR #3781, opened 2022-09-19
un-skip history test and fix golden mismatches

imjasonh:
Signed-off-by: Jason Hall <jason@chainguard.dev>

As discovered in https://github.com/docker/cli/pull/3779#discussion_r974281026, these tests were skipped because CI doesn't run in UTC. The failures don't seem to require UTC, only updating the golden file's column spacing to match current reality.

**- What I did**

Un-skipped tests in history_test.go

**- How I did it**

Removed `skip` from `history_test.go`, added `t.Run` wrappers so that failed assertions didn't stop other tests from running. Updated golden files to match test expectations.

**- How to verify it**

`make dev && make test-unit`

Image history tests are no longer skipped on CI.

octoeyes

cc @thaJeztah

### DIFF
--- a/cli/command/image/history_test.go
+++ b/cli/command/image/history_test.go
@@ -44,12 +44,15 @@ func TestNewHistoryCommandErrors(t *testing.T) {
 }
 
 func notUTCTimezone() bool {
-	now := time.Now()
-	return now != now.UTC()
+	if _, offset := time.Now().Zone(); offset != 0 {
+		return true
+	}
+	return false
 }
 
 func TestNewHistoryCommandSuccess(t *testing.T) {
 	skip.If(t, notUTCTimezone, "expected output requires UTC timezone")
+
 	testCases := []struct {
 		name             string
 		args             []string
@@ -93,13 +96,16 @@ func TestNewHistoryCommandSuccess(t *testing.T) {
 		},
 	}
 	for _, tc := range testCases {
-		cli := test.NewFakeCli(&fakeClient{imageHistoryFunc: tc.imageHistoryFunc})
-		cmd := NewHistoryCommand(cli)
-		cmd.SetOut(io.Discard)
-		cmd.SetArgs(tc.args)
-		err := cmd.Execute()
-		assert.NilError(t, err)
-		actual := cli.OutBuffer().String()
-		golden.Assert(t, actual, fmt.Sprintf("history-command-success.%s.golden", tc.name))
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			cli := test.NewFakeCli(&fakeClient{imageHistoryFunc: tc.imageHistoryFunc})
+			cmd := NewHistoryCommand(cli)
+			cmd.SetOut(io.Discard)
+			cmd.SetArgs(tc.args)
+			err := cmd.Execute()
+			assert.NilError(t, err)
+			actual := cli.OutBuffer().String()
+			golden.Assert(t, actual, fmt.Sprintf("history-command-success.%s.golden", tc.name))
+		})
 	}
 }
--- a/cli/command/image/testdata/history-command-success.non-human.golden
+++ b/cli/command/image/testdata/history-command-success.non-human.golden
@@ -1,2 +1,2 @@
-IMAGE               CREATED AT             CREATED BY          SIZE                COMMENT
-abcdef              2017-01-01T12:00:03Z   rose                0                   new history item!
+IMAGE     CREATED AT             CREATED BY   SIZE      COMMENT
+abcdef    2017-01-01T12:00:03Z   rose         0         new history item!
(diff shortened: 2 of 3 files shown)


cli/command/image/history_test.go
```diff
@@ -93,13 +91,15 @@ func TestNewHistoryCommandSuccess(t *testing.T) {
...
 	for _, tc := range testCases {
-		cli := test.NewFakeCli(&fakeClient{imageHistoryFunc: tc.imageHistoryFunc})
-		cmd := NewHistoryCommand(cli)
-		cmd.SetOut(io.Discard)
-		cmd.SetArgs(tc.args)
-		err := cmd.Execute()
-		assert.NilError(t, err)
-		actual := cli.OutBuffer().String()
-		golden.Assert(t, actual, fmt.Sprintf("history-command-success.%s.golden", tc.name))
+		t.Run(tc.name, func(t *testing.T) {
```
thaJeztah, line 94: <3 subtests

I think it'd be good to capture the `tc` variable in the loop here, in case we'd ever decide to use `t.Parallel()`;

```suggestion
		tc := tc
		t.Run(tc.name, func(t *testing.T) {
```

cli/command/image/history_test.go
```diff
@@ -49,7 +48,6 @@ func notUTCTimezone() bool {
 }
 
 func TestNewHistoryCommandSuccess(t *testing.T) {
-	skip.If(t, notUTCTimezone, "expected output requires UTC timezone")
```
thaJeztah, line 52: Tried this locally, and looks like we may have the same issue as in the other test; the output is formatted for the local timezone, which may not be UTC, in which case the `.golden` files won't match;

```bash
=== RUN   TestNewHistoryCommandSuccess/non-human
    history_test.go:102: assertion failed:
        --- expected
        +++ actual
        @@ -1,3 +1,3 @@
        -IMAGE     CREATED AT             CREATED BY   SIZE      COMMENT
        -abcdef    2017-01-01T12:00:03Z   rose         0         new history item!
        +IMAGE     CREATED AT                  CREATED BY   SIZE      COMMENT
        +abcdef    2017-01-01T13:00:03+01:00   rose         0         new history item!

```


Not sure what the best solution is for this one though (don't think the `.golden` files allow for placeholders and/or fuzzy matching 🤔

cli/command/image/history_test.go
imjasonh: I'm not sure this is strictly needed with `t.Parallel`, but I also don't think it hurts either.

cli/command/image/history_test.go
imjasonh: I'm running these tests in the dev env on macOS (`make dev`), which seems to set the timezone to UTC. Can we have tests run in UTC so they pass consistently?

(I'm also fine with just continuing to skip this test under non-UTC, but we'd still need to fix it so it _passes_ under UTC)

thaJeztah: Ah, yes, correct; running these tests inside the dev environment will run them inside a container (which default to use UTC). Running them directly on the host may not be the "common" way to run them, but (e.g.) to run a "one-off" test from within an IDE may be run that way.

I would suggest to either add back the skip for the whole test, or (now that we have subtests) to add a skip for that specific test (although I'd be fine with just adding back the `t.Skip` if we don't want to complicate things too much).

thaJeztah:
Looks like the second commit is missing a DCO sign-off; when you're updating, could you squash the commits, and fix the sign-off to make the DCO check happy?

Thanks again! Sorry for the delay (had a lot of catching up to do on notifications 😅)

cli/command/image/history_test.go
imjasonh: So it turned out that even inside `make dev` these were being skipped, even though the dev env's timezone was UTC. 🤔 

I changed the semantics of `notUTCTimezone` to check for an offset != 0, which means these run now in the dev env.

https://go.dev/play/p/9o5SSnHUONR

Regardless, whether we skip the test case or not, I think the bigger issue is that it gets skipped in CI, so there's no safeguard against bugs creeping in for non-UTC users (statistically, most of them 🌎). So I'd like to see if we can make this unskipped in CI somehow at least.

imjasonh:
Friendly ping. Anything else you'd like to see?

imjasonh:
I'm not sure why DCO isn't reporting. I'll try closing/reopening and/or re-pushing the change to see if that wakes it up.

imjasonh:
Another friendly ping. I'm not in any rush to get this done, I just keep noticing it's still open. Anything else blocking?

vvoland (approved):

imjasonh:
> Another friendly ping. I'm not in any rush to get this done, I just keep noticing it's still open. Anything else blocking?

Another one. I'm also fine just dropping this if there isn't interest in testing this. If that's the case though, I can also just delete the test.

thaJeztah (approved):
LGTM

thaJeztah:
let me close/reopen to have  a fresh run of CI 🙈

thaJeztah:
Whoops; looks like I never pressed the "merge" button on this one.

I was just working on a similar test, and it looks like there's an workaround that just never occurred to me; adding a `t.SetEnv("TZ", "UTC")` to the test will make the test use UTC. I'll do a quick follow-up after this one's merged 👍

PR #3787, opened 2022-09-24
Add bindMaps feature (#3786)

### LINKED ISSUE #3786: Auto-mappings for host paths in bind mount option
### Description

It would be nice if `.docker/config.json` has a configuration to convert

```
docker run -v /path/to/somewhere/foo:/workspace
```

to 

```
docker run -v /anotherpath/foo:/workspace
```

automatically.

**Background**

My machine is Windows and I'm running a docker daemon on WSL2 distribution (Ubuntu-20.04). That's not Docker Desktop but a standard docker daemon for Ubuntu, installed following the instructions in https://docs.docker.com/engine/install/ubuntu/ .
WSL2 distribution can access Windows filesystems on `/mnt/c` (via SMB). That means,

```
docker run -v /mnt/c/Users/ikedam/Desktop/work:/workspace ...
```

allows launch a container accessing files on Windows. I want to launch the container from Windows with:

```
docker run -v "C:\\Users\\ikedam\\Desktop\\work:/workspace" ...
```

or:

```
docker run -v ".\\work:/workspace" ...
```

This feature allows that.

**Compared to Docker Desktop**

Docker Desktop (for Windows) also provides a feature to bind-mount windows filesystems.

Pros in Docker Desktop:

* Rich UI.
* Provides everything for docker suite. It contains not only docker daemon, but also `docker` cli and `docker-compose` cli.
* Easy to setup. Users don't have to know about details of docker backends.
* Docker feature gets available not only to Windows, but also to all WSL2 distributions.
* Auto port mapping for published port to the public IP address.
    * Note: `127.0.0.1` on WSL2 is accessible also from Windows via `127.0.0.1`. So this pros in Docker Desktop is only when the user wants to publish ports to a public network interface.
* Transparent conversion between filesystems. E.g. running `docker inspect` on Windows outputs `C:\\...` for bind mounts. This feature doesn't support that.

Cons in Docker Desktop

* Slow to start. (Maybe for GUI features?)
* Requires updating all components to update a specific component. For example, when I want to introduce the latest `docker-compose` for a bug fix, I have to wait for the next release of Docker Desktop and have to update not only `docker-compose` but whole Docker Desktop.
* No way to downgrade.
    * Unfortunately, new versions of Docker Desktop so often introduce new bugs, and they are often critical ones. You can't downgrade to the previous version once you upgrade to a new version. And the only way to recover your Docker Desktop is to uninstall it once and reinstall the previous one. Of course, all data in docker (images, containers, volumes) are removed.
    * Updating Docker Desktop is really risky for this reason.
    * Docker Desktop is not a open-source and developers can't fix the problem nor find a workaround by themselves.

I believe Docker Desktop is useful, but often costs too much.

**Configuration example**

Just like the `proxies` member in `config.json` (https://docs.docker.com/network/proxy/#configure-the-docker-client):

```
{
  "bindMaps": {
    "default": {
        "C:\\:/mnt/c/"
    }
  }
}
```

ikedam:
- implements https://github.com/docker/cli/issues/3786

**- What I did**

This change allows host paths in bind mounts specified in `docker` command line automatically mapped to the paths suitable in hosts where dockerd runs.

**- How I did it**

I introduced a new `"bindMaps`" member in `.docker/config.json`.
It defines paths to map.

For example, if your local path `/home/ikedam` is actually located on `/exports/ikedam` on the host where dockerd run, your `.docker/config.json` would be like this:

```
{
  "bindMaps": {
    "/home/ikedam": "/exports/ikedam"
  }
}
```

That configuration converts `docker run -v "/home/ikedam/work:/worksapce" ...` to `docker run -v "/exports/ikedam/work:/workspace", and that would work correct on the remote machine.
It would be especially useful when used with `docker run -v "./work:/workspace" ...` or `docker run -v "$(pwd)/work:/workspace" ...`.

**- How to verify it**

Instructions to test on Windows and WSL2:

* Build `docker` command:

    * checkout this branch.
    * Build a command for Windows (This must be performed on an environment with docker):

        ```
        docker buildx bake --set binary.platform=windows/amd64
        ```

    * `build\docker-windows-amd64.exe` will be available and rename that to `docker.exe`.

* Set up WSL2 distribution (You can skip here if you have already an appropriate WSL2 distribution available):
    * Set the version of WSL to 2.

        ```
        wsl --set-default-version 2
        ```

    * Install Ubuntu-20.04.
        ```
         wsl --install -d Ubuntu-20.04
        ```
    * Set up your unix user.

* Install dockerd on WSL2:

    * Install Docker Engine following the instructions in https://docs.docker.com/engine/install/ubuntu/
        * Note: dockerd doesn't launch automatically.
    * Configure dockerd to accept 127.0.0.1:2375 (for access from Windows)

        ```
        $ cat /etc/default/docker
        ...
        DOCKER_OPTS="-H unix:///var/run/docker.sock -H tcp://127.0.0.1:2375"
        ...
        ```

    * Start dockerd. NOTE: dockerd doesn't start automatically, and you must start it every time you log onto Windows.

        ```
        $ sudo service docker start
        ```

* Test bindMaps feature from Windows:

    * Create %USERPROFILE%\.docker\config.json like this:

        ```
        {
          "bindMaps": {
            "default": {
              "C:\\": "/mnt/c"
            }
          }
        }
        ```

    * Configure DOCKER_HOST to connect to dockerd on WSL2:

        ```
        > ${ENV:DOCKER_HOST}="tcp://127.0.0.1:2375"
        ```

        * expected to be run on Powershell.

    * Run `docker` command. Consider built `docker` command is available in the current directory:

        ```
        > echo "test!" >index.html
        > .\docker run --rm -p 8080:80 -v ".:/usr/local/apache2/htdocs" --name httpd -d httpd:2.4-alpine
        > Invoke-WebRequest http://127.0.0.1:8080
        > .\docker stop httpd
        ```

        * The output from `Invoke-WebResuest` would contain "test!". That means bindMaps feature allows you to specify bind mount with Windows paths, even though the docker daemon run on a WSL2 distribution.

Add bindMaps feature: host paths in bind mounts can be mapped to another paths, like actual paths on the remote machine

closes #3786

### DIFF
--- a/cli/command/container/create.go
+++ b/cli/command/container/create.go
@@ -100,6 +100,8 @@ func runCreate(dockerCli command.Cli, flags *pflag.FlagSet, options *createOptio
 		reportError(dockerCli.Err(), "create", err.Error(), true)
 		return cli.StatusError{StatusCode: 125}
 	}
+	dockerCli.ConfigFile().ApplyBindMap(dockerCli.Client().DaemonHost(), containerConfig.HostConfig.Binds)
+	dockerCli.ConfigFile().ApplyBindMapToMounts(dockerCli.Client().DaemonHost(), containerConfig.HostConfig.Mounts)
 	if err = validateAPIVersion(containerConfig, dockerCli.Client().ClientVersion()); err != nil {
 		reportError(dockerCli.Err(), "create", err.Error(), true)
 		return cli.StatusError{StatusCode: 125}
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -111,6 +111,8 @@ func runRun(dockerCli command.Cli, flags *pflag.FlagSet, ropts *runOptions, copt
 		reportError(dockerCli.Err(), "run", err.Error(), true)
 		return cli.StatusError{StatusCode: 125}
 	}
+	dockerCli.ConfigFile().ApplyBindMap(dockerCli.Client().DaemonHost(), containerConfig.HostConfig.Binds)
+	dockerCli.ConfigFile().ApplyBindMapToMounts(dockerCli.Client().DaemonHost(), containerConfig.HostConfig.Mounts)
 	if err = validateAPIVersion(containerConfig, dockerCli.CurrentVersion()); err != nil {
 		reportError(dockerCli.Err(), "run", err.Error(), true)
 		return cli.StatusError{StatusCode: 125}
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/docker/cli/cli/config/credentials"
 	"github.com/docker/cli/cli/config/types"
+	"github.com/docker/docker/api/types/mount"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -42,6 +43,7 @@ type ConfigFile struct {
 	CLIPluginsExtraDirs  []string                     `json:"cliPluginsExtraDirs,omitempty"`
 	Plugins              map[string]map[string]string `json:"plugins,omitempty"`
 	Aliases              map[string]string            `json:"aliases,omitempty"`
+	BindMaps             map[string]map[string]string `json:"bindMaps,omitempty"`
 }
 
 // ProxyConfig contains proxy configuration settings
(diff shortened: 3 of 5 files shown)


ikedam:
I found this didn't work with `--mount` option like:
```
docker run --rm -p 8080:80 --mount "type=bind,source=$(pwd),target=/usr/local/apache2/htdocs,consistency=cached" --name httpd -d httpd:2.4-alpine
```
 (VSCode Dev Containers extension uses this option)

I update the implementation and now this supports `--mount` options.
And also resolved the conflict by rebasing to master.
The old implementation is available in: https://github.com/ikedam/cli/commits/3786-bindMaps.old

PR #3790, opened 2022-09-29
context: various cleanups and improvements

thaJeztah:
### cli/context/store: metadataStore.get(), .remove(): accept name instead of ID

This allows callers to just pass the name, and handle the conversion to ID and
path internally.

### cli/context/store: listRecursivelyMetadataDirs(): use filepath.Join()

Looks like the intent here is to get the path of a subdirectory.

### cli/context/store: removeAllContextData(): accept name instead of ID

This allows callers to just pass the name, and handle the conversion to ID and
path internally. This also fixes a test which incorrectly used "names" as
pseudo-IDs.

### cli/context/store: TestTlsCreateUpdateGetRemove(): use correct ID

This test was depending on the fact that contextDir's are a string,
and for the test is was using the context _name_ as a pseudo-ID.

This patch updates the test to be more explicit where ID's and where
names are used.

### cli/context/store: removeAllEndpointData(): accept name instead of ID

This allows callers to just pass the name, and handle the conversion to ID and
path internally.

### cli/context/store: createOrUpdate(): accept name instead of ID

This allows callers to just pass the name, and handle the conversion to ID and
path internally.

### cli/context/store: getData(): accept name instead of ID

This allows callers to just pass the name, and handle the conversion to ID and
path internally.

### cli/context/store: remove(): accept name instead of ID

This allows callers to just pass the name, and handle the conversion to ID and
path internally.

### cli/context/store: listContextData(): accept name instead of ID

This allows callers to just pass the name, and handle the conversion to ID and
path internally.

### cli/context/store: remove endpointDir(), filePath()

These were just to produce the full path of subdirectories within the context
directory; removing the extra abstraction.

### cli/context/store: simplify error handling, and make it more idiomatic

The package defined various special errors; these errors existed for two reasons;

- being able to distinguish "not found" errors from other errors (as "not found"
  errors can be ignored in various cases).
- to be able to update the context _name_ in the error message after the error
  was created. This was needed in cases where the name was not available at the
  location where the error was produced (e.g. only the "id" was present), and
  the helpers to detect "not found" errors did not support wrapped errors (so
  wrapping the error with a "name" could break logic); a `setContextName` interface
  and corresponding `patchErrContextName()` utility was created for this (which
  was a "creative", but not very standard approach).

This patch:

- Removes the special error-types, replacing them with errdefs definitions (which
  is a more common approach in our code-base to detect error types / classes).
- Removes the internal utilities for error-handling, and deprecates the exported
  utilities (to allow external consumers to adjust their code).
- Some errors have been enriched with detailed information (which may be useful
  for debugging / problem solving).
- Note that in some cases, `patchErrContextName()` was called, but the code
  producing the error would never return a `setContextName` error, so would
  never update the error message.

### cli/context/store: New(): return concrete type

Go conventions are for interfaces to be defined on the receiver side,
and for producers to return concrete types. This patch changes the
constructor to return a concrete type.

### cli/context/store: List(): don't interrupt listing for not-found errors

There's no reason to stop listing contexts if a context does not exist
while iterating over the directories,

### DIFF
--- a/cli/context/store/tlsstore.go
+++ b/cli/context/store/tlsstore.go
@@ -36,8 +36,8 @@ func (s *tlsStore) createOrUpdate(name, endpointName, filename string, data []by
 	return os.WriteFile(s.filePath(contextID, endpointName, filename), data, 0600)
 }
 
-func (s *tlsStore) getData(contextID contextdir, endpointName, filename string) ([]byte, error) {
-	data, err := os.ReadFile(s.filePath(contextID, endpointName, filename))
+func (s *tlsStore) getData(name, endpointName, filename string) ([]byte, error) {
+	data, err := os.ReadFile(s.filePath(contextdirOf(name), endpointName, filename))
 	if err != nil {
 		return nil, convertTLSDataDoesNotExist(endpointName, filename, err)
 	}
--- a/cli/context/store/tlsstore.go
+++ b/cli/context/store/tlsstore.go
@@ -23,7 +23,8 @@ func (s *tlsStore) filePath(contextID contextdir, endpointName, filename string)
 	return filepath.Join(s.root, string(contextID), endpointName, filename)
 }
 
-func (s *tlsStore) createOrUpdate(contextID contextdir, endpointName, filename string, data []byte) error {
+func (s *tlsStore) createOrUpdate(name, endpointName, filename string, data []byte) error {
+	contextID := contextdirOf(name)
 	epdir := s.endpointDir(contextID, endpointName)
 	parentOfRoot := filepath.Dir(s.root)
 	if err := os.MkdirAll(parentOfRoot, 0755); err != nil {
--- a/cli/context/store/tlsstore.go
+++ b/cli/context/store/tlsstore.go
@@ -62,7 +62,8 @@ func (s *tlsStore) removeAllContextData(name string) error {
 	return os.RemoveAll(s.contextDir(contextdirOf(name)))
 }
 
-func (s *tlsStore) listContextData(contextID contextdir) (map[string]EndpointFiles, error) {
+func (s *tlsStore) listContextData(name string) (map[string]EndpointFiles, error) {
+	contextID := contextdirOf(name)
 	epFSs, err := os.ReadDir(s.contextDir(contextID))
 	if err != nil {
 		if os.IsNotExist(err) {
--- a/cli/context/store/tlsstore.go
+++ b/cli/context/store/tlsstore.go
@@ -44,8 +44,10 @@ func (s *tlsStore) getData(name, endpointName, filename string) ([]byte, error) 
 	return data, nil
 }
 
-func (s *tlsStore) remove(contextID contextdir, endpointName, filename string) error {
-	err := os.Remove(s.filePath(contextID, endpointName, filename))
+// remove removes a TLS data from an endpoint
+// TODO(thaJeztah) tlsStore.remove() is not used anywhere outside of tests; should we use removeAllEndpointData() only?
+func (s *tlsStore) remove(name, endpointName, filename string) error {
+	err := os.Remove(s.filePath(contextdirOf(name), endpointName, filename))
 	if os.IsNotExist(err) {
 		return nil
 	}
--- a/cli/context/store/tlsstore.go
+++ b/cli/context/store/tlsstore.go
@@ -3,6 +3,9 @@ package store
 import (
 	"os"
 	"path/filepath"
+
+	"github.com/docker/docker/errdefs"
+	"github.com/pkg/errors"
 )
 
 const tlsDir = "tls"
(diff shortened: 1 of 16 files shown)


thaJeztah:
Ah, dang; linter now starts complaining about the missing GoDoc

```
#17 87.64 cli/context/store/store.go:112:6: exported: exported type ContextStore should have comment or be unexported (revive)
275
#17 87.64 type ContextStore struct {
#17 87.64      ^
277
#17 87.64 cli/context/store/store.go:117:1: exported: exported method ContextStore.List should have comment or be unexported (revive)
278
#17 87.64 func (s *ContextStore) List() ([]Metadata, error) {
#17 87.64 ^
280
#17 87.64 cli/context/store/store.go:134:1: exported: exported method ContextStore.CreateOrUpdate should have comment or be unexported (revive)
281
#17 87.64 func (s *ContextStore) CreateOrUpdate(meta Metadata) error {
#17 87.64 ^
283
#17 87.64 cli/context/store/store.go:138:1: exported: exported method ContextStore.Remove should have comment or be unexported (revive)
284
#17 87.64 func (s *ContextStore) Remove(name string) error {
#17 87.64 ^
286
#17 87.64 cli/context/store/store.go:145:1: exported: exported method ContextStore.GetMetadata should have comment or be unexported (revive)
287
#17 87.64 func (s *ContextStore) GetMetadata(name string) (Metadata, error) {
#17 87.64 ^
289
#17 87.64 cli/context/store/store.go:149:1: exported: exported method ContextStore.ResetTLSMaterial should have comment or be unexported (revive)
290
#17 87.64 func (s *ContextStore) ResetTLSMaterial(name string, data *ContextTLSData) error {
#17 87.64 ^
292
#17 87.64 cli/context/store/store.go:166:1: exported: exported method ContextStore.ResetEndpointTLSMaterial should have comment or be unexported (revive)
293
#17 87.64 func (s *ContextStore) ResetEndpointTLSMaterial(contextName string, endpointName string, data *EndpointTLSData) error {
#17 87.64 ^
295
#17 87.64 cli/context/store/store.go:181:1: exported: exported method ContextStore.ListTLSFiles should have comment or be unexported (revive)
296
#17 87.64 func (s *ContextStore) ListTLSFiles(name string) (map[string]EndpointFiles, error) {
#17 87.64 ^
298
#17 87.64 cli/context/store/store.go:189:1: exported: exported method ContextStore.GetTLSData should have comment or be unexported (revive)
299
#17 87.64 func (s *ContextStore) GetTLSData(contextName, endpointName, fileName string) ([]byte, error) {
#17 87.64 ^
301
#17 87.64 cli/context/store/store.go:197:1: exported: exported method ContextStore.GetStorageInfo should have comment or be unexported (revive)
302
#17 87.64 func (s *ContextStore) GetStorageInfo(contextName string) StorageInfo {
#17 87.64 ^
304
#17 ERROR: executor failed running [/bin/sh -c golangci-lint run]: exit code: 1
```

cli/context/store/tlsstore.go
```diff
@@ -15,29 +15,20 @@ func (s *tlsStore) contextDir(id contextdir) string {
 	return filepath.Join(s.root, string(id))
 }
 
-func (s *tlsStore) endpointDir(contextID contextdir, name string) string {
-	return filepath.Join(s.root, string(contextID), name)
-}
-
-func (s *tlsStore) filePath(contextID contextdir, endpointName, filename string) string {
```
vvoland, line 22: I'm slightly hesitant on removing these. IMO there's some value in having these helper functions due to the signature prompting you on what you should supply to obtain the final directory.
If I would be writing a new code and need to obtain the file path, then I would have to search for other code that does this to know how to construct the whole path with `filepath.Join`.
With those conversion functions I just need to know that there is this function and then the arguments to the functions will guide me on what I need to provide. For example if I miss the `endpointName` then the compiler will yell at me, whereas if I forget it in `filepath.Join` then I won't know that at the compile time.
WDYT?

thaJeztah: Let me have a look. My main reasons were that;

- Each of them would be reconstructing the parent directory, so I initially wanted to re-use `contextDir()` and `endpointDir()`
- The `fileDir()` one felt really redundant
- The `endpointDir()` as used in `listContextData()` was really silly, as we already were traversing the directory, and then were re-constructing the name … to get the name of the directory we were in.

Perhaps I can bring back `endpointDir()`, and only remove `filePath()`; let me see how that looks.

There were some other changes I was considering;

- As the `contextDir` depends on the _context_ store, I want to look at passing that directory in (effectively, calculating that path should be something done by the _context_ store)
- Given that for the TLS-store we _only_ want `ca.pem`, `cert.pem` and `key.pem`, perhaps add accessors for those instead of trying to keep it “too generic”

A lot of the code was written to be really “generic”, but that’s not how it’s used, and perhaps we should rewrite more code to be specific to how it’s _actually_ used.

vvoland: Makes sense to me, thanks for the explanation! 👍

vvoland (approved):

thaJeztah:
@vvoland I updated, and restored the `getEndpointDir()`, so only removing the `filePath()`

vvoland (approved):

Issue #3789, opened 2022-09-28
[master / 22.06] docker buildx install breaks alias check


thaJeztah:
### Description

Not sure when this broke (did not yet do a bisect), but I noticed that if the `~/.docker/config.json` contains an alias for `builder` (which happens after running `docker buildx install`), the CLI is broken;

```bash
docker version
not allowed to alias with builtin "buildx" as target
```

### Reproduce

```bash
make -f docker.Makefile binary
```

Without alias installed:

```bash
rm ~/.docker/config.json
./build/docker version
Client:
 Version:           22.06.0-beta.0-140-g3dad26ca2.m
 API version:       1.42
 Go version:        go1.19.1
 Git commit:        3dad26ca2
 Built:             Wed Sep 28 22:36:09 2022
 OS/Arch:           darwin/arm64
 Context:           default
...
```

After running `docker buildx install`;

```bash
./build/docker buildx install

cat ~/.docker/config.json
{
	"auths": {},
	"aliases": {
		"builder": "buildx"
	}
}

./build/docker version
not allowed to alias with builtin "buildx" as target
```

The error occurs in this part of the code inside `processAliases()`; https://github.com/docker/cli/blob/c780f7c4abaf67034ecfaa0611e03695cf9e4a3e/cmd/docker/aliases.go#L29-L31

Because the `cmd.Find("buildx"))` returns a `cmd`. From a quick test the `cmd` it finds may be the "stub" command we create for plugins; when disabling the `buildx` plugin (renaming the file), the problem goes away;

```bash
mv /usr/local/lib/docker/cli-plugins/docker-buildx /usr/local/lib/docker/cli-plugins/docker-buildx-disabled
 
./build/docker version
Client:
 Version:           22.06.0-beta.0-140-g3dad26ca2.m
 API version:       1.42
 Go version:        go1.19.1
 Git commit:        3dad26ca2
 Built:             Wed Sep 28 22:36:09 2022
 OS/Arch:           darwin/arm64
 Context:           default
...
```

### docker version

```bash
current master (built from https://github.com/docker/cli/commit/3dad26ca2d418092b8c4e01b03d0455d583bec86)

Client:
 Version:           22.06.0-beta.0-140-g3dad26ca2.m
 API version:       1.42
 Go version:        go1.19.1
 Git commit:        3dad26ca2
 Built:             Wed Sep 28 22:36:09 2022
 OS/Arch:           darwin/arm64
 Context:           default
```

### docker info

```bash
not relevant
```

[...]

PR #3795, opened 2022-09-29
fix broken alias check is buildx is installed as alias for builder

thaJeztah:
- fixes https://github.com/docker/cli/issues/3789
- relates to https://github.com/docker/cli/pull/3429
- relates to https://github.com/docker/cli/issues/3621

Commit cbec75e2f3f5afd3334db693829a6b150612076c (https://github.com/docker/cli/pull/3429) updated `runDocker()` to load plugin-stubs before `processAliases()` was executed. As a result, plugin stubs were considered as "builtin commands", causing the alias verification to fail;

Without alias installed:

```bash
docker version
Client:
 Version:           22.06.0-beta.0-140-g3dad26ca2.m
 API version:       1.42
 Go version:        go1.19.1
 Git commit:        3dad26ca2
 Built:             Wed Sep 28 22:36:09 2022
 OS/Arch:           darwin/arm64
 Context:           default
...
```

After running `docker buildx install`;

```bash
./build/docker buildx install

cat ~/.docker/config.json
{
    "aliases": {
        "builder": "buildx"
    }
}

./build/docker version
not allowed to alias with builtin "buildx" as target
```

This patch moves loading the stubs _after_ the call to `processAliases()`, so that verification passes. As an extra precaution, the `processAliases()` function is also updated to exclude plugin-stub commands.

Note that cbec75e2f3f5afd3334db693829a6b150612076c (https://github.com/docker/cli/pull/3429) also introduced a performance regression (https://github.com/docker/cli/issues/3621), which may be related to the early loading of plugins (and creating stubs); it looks like various other code locations may also be loading plugins, for example `tryPluginRun()` calls `pluginmanager.PluginRunCommand()`, which also traverses plugin directories.

We should look under what circumstances the plugin stub-commands are actually needed, and make sure that they're only created in those situations.

### DIFF
--- a/cmd/docker/aliases.go
+++ b/cmd/docker/aliases.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"strings"
 
+	pluginmanager "github.com/docker/cli/cli-plugins/manager"
 	"github.com/docker/cli/cli/command"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


crazy-max (approved):
LGTM

thaJeztah:
thx; let me bring this one in 👍

PR #3770, opened 2022-09-03
golangci-lint: enable more linters

thaJeztah:
- depends on https://github.com/docker/cli/pull/3769
- depends on https://github.com/docker/cli/pull/3768

PR #3798, opened 2022-09-30
Use gofumpt if available, and enable gofumpt linter

thaJeztah:
- follow-up to https://github.com/docker/cli/pull/3794
- relates to https://github.com/docker/cli/pull/3770

gofumpt (https://github.com/mvdan/gofumpt) provides a supserset of gofmt / go fmt, but not every developer may have
it installed, so for situations where it's not available, fall back to gofmt.

As our code has been formatted with gofumpt already, in most cases contributions
will follow those formatting rules, but in some cases there may be a difference,
which would already be flagged by manual code review, but let's also enable the
gofumpt linter.

With this change, `make fmt` will use gofumpt is available; gofumpt has been
added to the dev-container, so `make -f docker.Makefile fmt` will always use it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,11 @@
 # Sets the name of the company that produced the windows binary.
 PACKAGER_NAME ?=
 
+# The repository doesn't have a go.mod, but "go list", and "gotestsum"
+# expect to be run from a module.
+GO111MODULE=auto
+export GO111MODULE
+
 all: binary
 
 _:=$(shell ./scripts/warn-outside-container $(MAKECMDGOALS))
@@ -45,8 +50,12 @@ shellcheck: ## run shellcheck validation
 	find scripts/ contrib/completion/bash -type f | grep -v scripts/winresources | grep -v '.*.ps1' | xargs shellcheck
 
 .PHONY: fmt
-fmt: ## run gofmt
-	go list -f {{.Dir}} ./... | xargs gofmt -w -s -d
+fmt: ## run gofumpt (if present) or gofmt
+	@if command -v gofumpt > /dev/null; then \
+		gofumpt -w -d -lang=1.19 . ; \
+	else \
+		go list -f {{.Dir}} ./... | xargs gofmt -w -s -d ; \
+	fi
 
 .PHONY: binary
 binary: ## build executable for Linux
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,7 @@ linters:
     - depguard
     - dogsled
     - gocyclo
+    - gofumpt
     - goimports
     - gosec
     - gosimple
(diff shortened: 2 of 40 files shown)


thaJeztah:
@crazy-max @vvoland ptal 😄

Makefile
```diff
@@ -45,8 +50,12 @@ shellcheck: ## run shellcheck validation
 	find scripts/ contrib/completion/bash -type f | grep -v scripts/winresources | grep -v '.*.ps1' | xargs shellcheck
 
 .PHONY: fmt
-fmt: ## run gofmt
-	go list -f {{.Dir}} ./... | xargs gofmt -w -s -d
+fmt: ## run gofumpt (if present) or gofmt
+	@if command -v gofumpt > /dev/null; then \
+		gofumpt -w -d . ; \
```
thaJeztah, line 55: `gofumpt` automatically excludes `vendor/` so no need to use `go list`

thaJeztah:
Hm... interesting; somehow my `gofumpt` didn't format all files?

```
#17 87.07 e2e/plugin/basic/basic.go:17: File is not `gofumpt`-ed (gofumpt)
#17 87.07 	if err := os.MkdirAll(p, 0755); err != nil {
#17 87.07 cli/context/store/metadatastore.go:31: File is not `gofumpt`-ed (gofumpt)
#17 87.07 	if err := os.MkdirAll(contextDir, 0755); err != nil {
#17 87.07 cli/context/store/metadatastore.go:38: File is not `gofumpt`-ed (gofumpt)
#17 87.07 	return os.WriteFile(filepath.Join(contextDir, metaFile), bytes, 0644)
#17 87.07 cli/context/store/store.go:249: File is not `gofumpt`-ed (gofumpt)
```

Makefile
```diff
@@ -45,8 +50,12 @@ shellcheck: ## run shellcheck validation
 	find scripts/ contrib/completion/bash -type f | grep -v scripts/winresources | grep -v '.*.ps1' | xargs shellcheck
 
 .PHONY: fmt
-fmt: ## run gofmt
-	go list -f {{.Dir}} ./... | xargs gofmt -w -s -d
+fmt: ## run gofumpt (if present) or gofmt
+	@if command -v gofumpt > /dev/null; then \
+		gofumpt -w -d -lang=1.19 . ; \
```
thaJeztah, line 55: Looks like I need to add an explicit `-lang=1.19`, otherwise it doesn't pick up the go version for which to format. Probably because we don't have a go.mod

thaJeztah:
@crazy-max ptal 🤗

crazy-max (approved):
LGTM if opt-in.

Do we want a scheduled worklow to fix-up and open PR as semi-automation process?

thaJeztah:
Let me bring this one in 👍

PR #3805, opened 2022-10-09
docs: use correct separator in --security-opt

debfx:
> Security options with `:` as a separator are deprecated and will be completely unsupported in 17.04, use `=` instead.

Signed-off-by: Felix Geyer <debfx@fobos.de>


thaJeztah (approved):
nice catch, thanks!

LGTM

PR #3812, opened 2022-10-12
add ability to override ints in compose files with 0

### LINKED ISSUE #3293: Docker stack not scaling to 0 on overridden yml
ln order to deploy our services over docker swarm we use 2 different yml files : the base one and the customized one
Here it is an example of our base.yml file:
```
version: '3.6'
services:
    hello-server-api:
      image: hello-server
      environment:
        - ID=1
      deploy:
        mode: replicated
        replicas: 1
        update_config:
          delay: 50s
        restart_policy:
          condition: on-failure
          max_attempts: 3
```
Here it is an example of our customized.yml file:
```
version: '3.6'
services:
    hello-server-api:
      deploy:
        mode: replicated
        replicas: 0
        update_config:
          delay: 50s
        restart_policy:
          condition: on-failure
          max_attempts: 3
```
we use the customized yml file to manage variables from production and the number of replicas of each container.

docker command to deploy:
```
docker stack deploy -c base.yml -c customized.yml server
```
This approach on docker-ce 19.03.11 was working correctly, in fact, the docker service is always deployed 0/0.

Upgrading the docker-ce version to 20.10.0 (we also tried 20.10.8) the behavior is changed and the replicas are always 1/1, it seems like it is not possible to set 0 in the customized.yml file.

However, if we put replicas: 2 in the base.yml and replicas: 1 in the customized.yml it is working correctly (it is working properly in every increasing and decreasing scenario, except when 0 value is used in the customized one ).

mspiess:
fixes #3293

**- What I did**
Allow the number for `replicas` to be overridden with 0 when using multiple compose files.

**- How I did it**
I added a custom merge function for the type `*uint64`. Maybe there is a more elegant way, but I was struggling to find documentation or examples for the used library "mergo". [This issue](https://github.com/imdario/mergo/issues/131) of the library might be related.

**- How to verify it**
See the reproduction steps in the linked issue.

Fix overriding ints with 0 when using multiple compose files

:whale: 
Signed-off-by: Marco Spiess <marco.spiess@hotmail.de>

### DIFF
--- a/cli/compose/loader/merge.go
+++ b/cli/compose/loader/merge.go
@@ -61,6 +61,7 @@ func mergeServices(base, override []types.ServiceConfig) ([]types.ServiceConfig,
 			reflect.TypeOf([]types.ServiceVolumeConfig{}):    mergeSlice(toServiceVolumeConfigsMap, toServiceVolumeConfigsSlice),
 			reflect.TypeOf(types.ShellCommand{}):             mergeShellCommand,
 			reflect.TypeOf(&types.ServiceNetworkConfig{}):    mergeServiceNetworkConfig,
+			reflect.PointerTo(reflect.TypeOf(uint64(1))):     mergeUint64,
 		},
 	}
 	for name, overrideService := range overrideServices {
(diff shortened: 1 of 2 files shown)


thaJeztah (approved):
LGTM, thanks!

PR #3821, opened 2022-10-19
Remove "experimental" gates around "--platform" in bash completion

tianon:
The `--platform` flag has been out of experimental for a while now. 🎉

(I'd also like to upgrade this to add reasonable completion for `--platform` values, but that's a little more involved :sweat_smile:)


albers (approved):
LGTM, thanks!

thaJeztah (approved):
LGTM, thanks!

PR #3845, opened 2022-11-04
Fix bug where incorrect response is returned

### LINKED ISSUE #3844: Incorrect error message when server is unreachable
### Description

When the docker daemon is unreachable, if we run a command that needs to check the server version, we receive an error saying the command is not supported on this server version/type instead of the cannot connect to server error. 

See https://github.com/checkpoint-restore/criu/issues/1977

### Reproduce

Run without sudo as a user not part of the docker group.
`docker checkpoint create test test`

### Expected behavior

The above command should error out with `Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock` instead of `docker checkpoint create is only supported on a Docker daemon running on linux, but the Docker daemon is running on`

### docker version

```bash
 Version:           20.10.20
 API version:       1.41
 Go version:        go1.18.7
 Git commit:        9fdeb9c
 Built:             Tue Oct 18 18:20:44 2022
 Context:           default
 Experimental:      true

 Engine:
  Version:          20.10.20
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.7
  Git commit:       03df974
  Built:            Tue Oct 18 18:18:35 2022
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          1.6.8
  GitCommit:        9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.9.1-docker)
  compose: Docker Compose (Docker Inc., v2.12.0)

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 8
 Server Version: 20.10.20
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 io.containerd.runtime.v1.linux runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 runc version: v1.1.4-0-g5fd4c4d
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.10.0-19-cloud-amd64
 Operating System: Debian GNU/Linux 11 (bullseye)
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 3.839GiB
 Name: ip-172-31-11-198
 ID: QJKF:IFSV:OKVX:PXZR:23KB:32YD:N6WU:AO24:OMMH:XVT6:GX4Z:JP2J
 Debug Mode: false
 Labels:
 Experimental: true
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

### Additional Info

_No response_

adyanth:
**- What I did**
Check for server reachability when `areSubcommandsSupported` fails.

**- How I did it**
By calling the `/_ping` API using `APIClient.Ping()` and reporting the error.

**- How to verify it**
Run `docker checkpoint create test test` as a user without privilege to access the docker socket.

Fixes #3844 incorrect error message when server unreachable.

### DIFF
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
@@ -374,8 +375,12 @@ func findCommand(cmd *cobra.Command, commands []string) bool {
 	return findCommand(cmd.Parent(), commands)
 }
 
-func isSupported(cmd *cobra.Command, details versionDetails) error {
+func isSupported(cmd *cobra.Command, details command.Cli) error {
 	if err := areSubcommandsSupported(cmd, details); err != nil {
+		// Check if the command not being supported is due to Ping failing
+		if _, connectError := details.Client().Ping(context.Background()); connectError != nil {
+			return connectError
+		}
 		return err
 	}
 	return areFlagsSupported(cmd, details)


thaJeztah (commented):
Thanks for contributing!

I saw this PR, but wanted to check if we could possibly avoid the additional "ping". I _think_ this code is also used for outputting `docker help` etc, and I know we have some code that falls back to "use defaults if we weren't able to ping" as for those it's not _critical_ if the connection failed.

That said, when _executing_ the command (as the issue this one is fixing), we should indeed show the correct error.

With the above in mind; perhaps we need to look at better disambiguating "we successfully obtained Information from "ping"" from "we couldn't ping, so we're assuming it's available".

I'll try to make some time to have a look at the code flow, but feel free to give me a ping if I don't get back to you ❤️

adyanth:
Thanks for taking the time to review!

Yes, this code is also used for help, which is why the "ping" only happens if we are about to return an error anyway, which would be an incorrect one if it was due to the ping failing.

Do let me know if you think of something else!

adyanth:
Hey @thaJeztah, were you able to take a look?

adyanth:
Hey @thaJeztah just checking back on this one :)

adyanth:
Yeah, I saw that default to experimental check, which felt out of place! 

I see that you have switched from checking if the Ping returned an error to checking if the OSType is empty or not. I did not want to do that because I am unaware of what would happen when the daemon actually returned an empty string due to an error/bug.

Thank you! Do you think we should close this PR in favor of #3901?

thaJeztah:
> Do you think we should close this PR in favor of https://github.com/docker/cli/pull/3901?

I added a "closes" in the other PR, so this one will automatically be closed if the other one gets merged.

Debugging a test right now, which fails in CI, but didn't fail locally; looks like that test may be incorrect, but trying to find out what is happening exactly 😅

thaJeztah:
>  added a "closes" in the other PR, so this one will automatically be closed if the other one gets merged.

Looks like GitHub didn't trigger that part; https://github.com/docker/cli/pull/3901 was merged, so let me close this one; thanks for contributing!!

Issue #3096, opened 2021-05-13
[BUG] Docker using context is unusable if it cannot connect to daemon  - can't change context, print version, build image


ieugen:
**Description**

I have setup a docker context that connects via ssh to a server that is available behind OpenVPN. 
I `docker context use` that context and do some work while connected to VPN.
The connection dies off or I stop the VPN then docker is unusable. 

I can't even change context back to default - local docker install. 
I can't build docker images either. 

All of these things should not depend on a docker daemon being present. 

I believe this is a bug.
Every docker command hangs if the context is not accessible. 

I believe being so reliant on the docker connection is a bug. 

**Steps to reproduce the issue:**
1. Create a context to a remote server 
2. Use that context
3. Stop the connection to that context
4. Weep :( . 

**Describe the results you received:**

Docker commands hang.

**Describe the results you expected:**

I would like to have the ability to change to default context.
I would like to have the ability to build images.  

**Additional information you deem important (e.g. issue happens only occasionally):**

Happens every time.

**Output of `docker version`:**

```
 docker version 
Client: Docker Engine - Community
 Version:           20.10.6
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        370c289
 Built:             Fri Apr  9 22:46:45 2021
 OS/Arch:           linux/amd64
 Context:           dre-main
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.5
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       363e9a8
  Built:            Tue Mar  2 20:16:15 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

```

**Output of `docker info`:**

```
docker info 
Client:
 Context:    dre-main
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  scan: Docker Scan (Docker Inc., v0.7.0)

Server:
 Containers: 17
  Running: 6
  Paused: 0
  Stopped: 11
 Images: 350
[...]

Issue #1955, opened 2019-06-18
Difficult to execute commands when currentContext is invalid


kinghuang:
**Description**

If the current context is not valid, it is not easy to run any CLI commands, including switching to a valid context.

**Steps to reproduce the issue:**

Assume there are contexts named `default` and `context-two`.

1. Make `context-two` the current context.
2. Set `DOCKER_CONTEXT=default` to override it in the current shell.
3. Remove `context-two`.
4. Attempt to do something with the Docker CLI, like `docker version`.

```sh
$ docker context ls --format 'table {{ .Name }}\t{{ .Current }}'
NAME                <no value>
context-two         true
default             false
$ export DOCKER_CONTEXT=default
$ docker context rm context-two
context-two
```

**Describe the results you received:**

Any CLI operation results in an error. It is not possible to switch to a valid context using `docker context use …`.

```sh
$ docker version
Current context "context-two" is not found on the file system, please check your config file at /Users/king.huang/.docker/config.json
$ docker context use default
Current context "context-two" is not found on the file system, please check your config file at /Users/king.huang/.docker/config.json
```

**Describe the results you expected:**

It should be possible to at least list and switch to a valid context.

**Additional information you deem important (e.g. issue happens only occasionally):**

When removing the current context specified in `config.json`, perhaps the CLI should change the value of `currentContext` to something like `default`.

The current context can be fixed by editing `config.json` manually, or using the `DOCKER_CONTEXT` environment variable to switch to a valid context, and then using a valid context.

```sh
$ export DOCKER_CONTEXT=default
$ docker context use default
default
Current context is now "default"
```

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Version:           19.03.0-rc2
 API version:       1.40
 Go version:        go1.12.5
 Git commit:        f97efcc
 Built:             Wed Jun  5 01:37:53 2019
 OS/Arch:           darwin/amd64
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          19.03.0-rc2
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.5
  Git commit:       f97efcc
  Built:            Wed Jun  5 01:42:10 2019
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v1.2.6
  GitCommit:        894b81a4b802e4eb2a91d1ce216b8817763c29fb
 runc:
  Version:          1.0.0-rc8
[...]

Issue #2584, opened 2020-06-16
'docker context ls' or 'docker context inspect' slow responses with remote servers


aduzsardi:
* [x] This is a bug report
* [x] This is a feature request
* [x] I searched existing issues before opening this one

### Expected behavior
Return current used docker context faster

### Actual behavior
while using a remote docker host context through a VPN or over the internet, the listing and inspecting of the contexts is very slow
```
$ time docker context ls
NAME                DESCRIPTION                               DOCKER ENDPOINT                 KUBERNETES ENDPOINT   ORCHESTRATOR
bb-dev *            BB Dev Environment                        ssh://deploy@192.168.213.214                         swarm
default             Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                           swarm

real	0m2.547s
user	0m0.020s
sys	0m0.015s
``` 
If using the default (local) docker context , everything seems to be ok
```
$ docker context use default
Current context is now "default"
$ time docker context ls
NAME                DESCRIPTION                               DOCKER ENDPOINT                 KUBERNETES ENDPOINT   ORCHESTRATOR
bb-dev              BB Dev Environment                        ssh://deploy@192.168.213.214                         swarm
default *           Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                           swarm

real	0m0.026s
user	0m0.024s
sys	0m0.010s
```

### Steps to reproduce the behavior

**Output of `docker version`:**
```
lient: Docker Engine - Community
 Version:           19.03.11
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        42e35e61f3
 Built:             Mon Jun  1 09:12:34 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.11
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       42e35e61f3
  Built:            Mon Jun  1 09:11:07 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Output of `docker info`:**

```
Client:
 Debug Mode: false

Server:
 Containers: 1
[...]

PR #3847, opened 2022-11-04
context: implement lazy loading, and other improvements

thaJeztah:
- fixes https://github.com/docker/cli/issues/1955
- fixes https://github.com/docker/cli/issues/3096
- relates to https://github.com/docker/cli/issues/2584

### cli/command: load context, and initialize client lazily

This allows commands that don't require a client connection (such as `context use`)
to be functional, but still produces an error when trying to run a command that
needs to connect with the API;

    mkdir -p ~/.docker/ && echo '{"currentContext":"nosuchcontext"}' >  ~/.docker/config.json
    docker version
    Failed to initialize: unable to resolve docker endpoint: load context "nosuchcontext": context does not exist: open /root/.docker/contexts/meta/8bfef2a74c7d06add4bf4c73b0af97d9f79c76fe151ae0e18b9d7e57104c149b/meta.json: no such file or directory

    docker context use default
    default

    docker version
    Client:
     Version:           22.06.0-dev
     API version:       1.42
     ...

### cli/command/context: context ls: add ERROR column, and don't fail early

This updates `docker context ls` to:

- not abort listing contexts when failing one (or more) contexts
- instead, adding an ERROR column to inform the user there was
  an issue loading the context.

### cli/command/context: context ls: always show current context

if a context is set (e.g. through DOCKER_CONTEXT or the CLI config file), but
wasn't found, then a "stub" context is added, including an error message that
the context doesn't exist.

    DOCKER_CONTEXT=nosuchcontext docker context ls
    NAME              DESCRIPTION                               DOCKER ENDPOINT               ERROR
    default           Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
    nosuchcontext *                                                                           load context "nosuchcontext": context does n…

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -152,7 +152,7 @@ func (cli *DockerCli) ConfigFile() *configfile.ConfigFile {
 // ServerInfo returns the server version details for the host this client is
 // connected to
 func (cli *DockerCli) ServerInfo() ServerInfo {
-	// TODO(thaJeztah) make ServerInfo() lazily load the info (ping only when needed)
+	_ = cli.initialize()
 	return cli.serverInfo
 }
 
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -2,12 +2,14 @@ package command
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/docker/cli/cli/config"
@@ -78,6 +80,8 @@ type DockerCli struct {
 	contentTrust       bool
 	contextStore       store.Store
 	currentContext     string
+	init               sync.Once
+	initErr            error
 	dockerEndpoint     docker.Endpoint
 	contextStoreConfig store.Config
 	initTimeout        time.Duration
@@ -91,6 +95,7 @@ func (cli *DockerCli) DefaultVersion() string {
 // CurrentVersion returns the API version currently negotiated, or the default
 // version otherwise.
 func (cli *DockerCli) CurrentVersion() string {
+	_ = cli.initialize()
 	if cli.client == nil {
 		return api.DefaultVersion
 	}
@@ -99,6 +104,10 @@ func (cli *DockerCli) CurrentVersion() string {
 
 // Client returns the APIClient
 func (cli *DockerCli) Client() client.APIClient {
+	if err := cli.initialize(); err != nil {
+		_, _ = fmt.Fprintf(cli.Err(), "Failed to initialize: %s\n", err)
+		os.Exit(1)
+	}
 	return cli.client
 }
 
@@ -203,8 +212,6 @@ func WithInitializeClient(makeClient func(dockerCli *DockerCli) (client.APIClien
 // Initialize the dockerCli runs initialization that must happen after command
 // line flags are parsed.
 func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...InitializeOpt) error {
-	var err error
-
 	for _, o := range ops {
 		if err := o(cli); err != nil {
 			return err
@@ -232,18 +239,6 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...Initialize
 			return ResolveDefaultContext(cli.options, cli.contextStoreConfig)
 		},
 	}
-	cli.dockerEndpoint, err = resolveDockerEndpoint(cli.contextStore, resolveContextName(opts, cli.configFile))
-	if err != nil {
-		return errors.Wrap(err, "unable to resolve docker endpoint")
-	}
-
-	if cli.client == nil {
-		cli.client, err = newAPIClientFromEndpoint(cli.dockerEndpoint, cli.configFile)
-		if err != nil {
-			return err
-		}
-	}
-	cli.initializeFromClient()
 	return nil
 }
 
@@ -282,6 +277,9 @@ func newAPIClientFromEndpoint(ep docker.Endpoint, configFile *configfile.ConfigF
 }
 
 func resolveDockerEndpoint(s store.Reader, contextName string) (docker.Endpoint, error) {
+	if s == nil {
+		return docker.Endpoint{}, fmt.Errorf("no context store initialized")
+	}
 	ctxMeta, err := s.GetMetadata(contextName)
 	if err != nil {
 		return docker.Endpoint{}, err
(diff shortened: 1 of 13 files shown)


cli/command/context/show.go
```diff
@@ -25,10 +25,10 @@ func newShowCommand(dockerCli command.Cli) *cobra.Command {
...
 	context := dockerCli.CurrentContext()
-	metadata, err := dockerCli.ContextStore().GetMetadata(context)
+	fmt.Fprintln(dockerCli.Out(), context)
+	_, err := dockerCli.ContextStore().GetMetadata(context)
 	if err != nil {
 		return err
 	}
-	fmt.Fprintln(dockerCli.Out(), metadata.Name)
 	return nil
 }
```
thaJeztah, line 34: For this one, I'm actually wondering if it would make sense to "just show" what's currently configured as context, and skip the validation altogether. As `docker show` could be commonly used to set the prompt, there were some concerns about performance here; https://github.com/docker/cli/issues/2584#issuecomment-1222720408

cli/command/cli.go
```diff
@@ -324,7 +324,7 @@ func (cli *DockerCli) getInitTimeout() time.Duration {
 
 func (cli *DockerCli) initializeFromClient() {
 	ctx := context.Background()
-	if !strings.HasPrefix(cli.DockerEndpoint().Host, "ssh://") {
+	if !strings.HasPrefix(cli.client.DaemonHost(), "ssh://") {
```
thaJeztah, line 334: Hm... looks like something to fix; perhaps `DaemonHost()` does not contain the `ssh://` protocol, in which case it ignores this block 🤔 

```
#5 [1/4] FROM docker.io/library/docker:test-dind@sha256:af3033aaad3b67cfd2fcefed035ea60b70627d3e7636987ba03f7b3f09f3fc3b
Pulling registry (registry:2)...
2: Pulling from library/registry
Digest: sha256:2e830e8b682d73a1b70cac4343a6a541a87d5271617841d87eeb67a824a5b3f2
Status: Downloaded newer image for registry:2
Creating cliendtoendsuite_evil-notary-server_1 ... 
Creating cliendtoendsuite_notary-server_1      ... 
Creating cliendtoendsuite_engine_1             ... 
Creating cliendtoendsuite_registry_1           ... 
Creating cliendtoendsuite_engine_1             ... done
Creating cliendtoendsuite_evil-notary-server_1 ... done
Creating cliendtoendsuite_registry_1           ... done
Creating cliendtoendsuite_notary-server_1      ... done
Waiting for docker daemon to become available at ssh://penguin@172.18.0.3
time="2022-11-04T15:46:30Z" level=warning msg="commandConn.CloseWrite: commandconn: failed to wait: signal: killed"
time="2022-11-04T15:46:30Z" level=warning msg="commandConn.CloseWrite: commandconn: failed to wait: signal: killed"
Client:
 Version:           22.06.0-dev
 API version:       1.40 (downgraded from 1.42)
 Go version:        go1.19.2
 Git commit:        d9474bc
 Built:             Fri Nov  4 15:43:50 2022
 Context:           default

 Engine:
  Version:          19.03.13
  Go version:       go1.13.15
  Git commit:       4484c46
  Built:            Wed Sep 16 17:04:43 2020
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v1.3.7
  GitCommit:        8fba4e9a7d01810a393d5d25a3621dc101981175
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
time="2022-11-04T15:46:30Z" level=warning msg="commandConn.CloseWrite: commandconn: failed to wait: signal: killed"
error during connect: Post "http://docker.example.com/v1.40/images/create?fromImage=alpine&tag=sha256%3Af006ecbb824d87947d0b51ab8488634bf69fe4094959d935c0c103f4820a417d": command [ssh -l penguin -- 172.18.0.3 docker system dial-stdio] has exited with signal: killed, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=
make: *** [docker.Makefile:145: test-e2e-connhelper-ssh] Error 1
Error: Process completed with exit code 2.
```

thaJeztah: Hm... yes, looks like it. We set up a dialer, but configure the client itself to use "a host" (`http://docker.example.com`);
https://github.com/docker/cli/blob/759007c8ca0cbd6c38e74ec32d75a9720eedd9b2/cli/connhelper/connhelper.go#L43-L54

thaJeztah:
Okay; moved this one out of draft; @vvoland @silvin-lubecki PTAL

cli/command/context/testdata/list-with-error.golden
```diff
@@ -0,0 +1,3 @@
+NAME              DESCRIPTION                               DOCKER ENDPOINT               ERROR
+default           Current DOCKER_HOST based configuration   unix:///var/run/docker.sock   
+nosuchcontext *                                                                           load context "nosuchcontext": context does n…
```
vvoland, line 3: My first reaction was that it feels weird to have `list` output a non-existing context. On the second thought it  presents the real state more explicitly, so I think it's okay.

One thing that could be better IMO is the visibility of the "context does not exist error", which seems to be hidden because of the ellipsis.
Could we shorten the error to "context does not exist" in this case? The name is already in the `NAME` column, so I think that `load context "nosuchcontext"` doesn't add much value here.
This could still make the "context does not exist" invisible, if the name is long though.

We could also the default order of the columns to: NAME, ENDPOINT, ERROR, DESCRIPTON.
Name and endpoint seem to be the most important bits, error in most cases should be empty so shouldn't take up much space, and truncating description is probably the least harmful.

WDYT?

vvoland (approved):

cli/command/context/testdata/list-with-error.golden
thaJeztah: > Could we shorten the error to "context does not exist" in this case? The name is already in the NAME column, so I think that load context "nosuchcontext" doesn't add much value here.
> This could still make the "context does not exist" invisible, if the name is long though.

Yes, I was looking at that originally; the tricky bit here was that the error is produced in a part of the code that's called in quite some places, and I couldn't immediately judge if removing the name from the error would hide relevant information (which context failed to load?) in all of those cases. We can definitely look at this again, and perhaps improve. I also want to add the `--no-trunc` option (have some of that stashed locally), but will keep that for a follow-up.

I tweaked things a bit; changedc the `load context "name of context"` to `context "name of context"`, and changed `does not exist` to `not found`;

```
NAME              DESCRIPTION                               DOCKER ENDPOINT               ERROR
default           Current DOCKER_HOST based configuration   unix:///var/run/docker.sock   
nosuchcontext *                                                                           context "nosuchcontext": context not found: …
```

> We could also the default order of the columns to: NAME, ENDPOINT, ERROR, DESCRIPTION.
> Name and endpoint seem to be the most important bits, error in most cases should be empty so shouldn't take up much space, and truncating description is probably the least harmful.

I stuck with current conventions of putting the ERROR as last column, but also agree that I always was a bit on the fence for `DESCRIPTION` altogether; perhaps `NAME, ENDPOINT, DESCRIPTION, ERROR`? Let's fine-tune this in a follow up.

thaJeztah: > the tricky bit here was that the error is produced in a part of the code that's called in quite some places

Oh, actually; we can change it; I can check if it's an `errdefs.IsErrNotFound()`, and in that case replace it with `context not found`.

thaJeztah: Oh, but if we do, then we're hiding the underlying error as well.

Alright; let's look at improving in a follow-up.

PR #3851, opened 2022-11-05
[20.10] update to Go 1.18.8 to address CVE-2022-41716

thaJeztah:
On Windows, syscall.StartProcess and os/exec.Cmd did not properly
    check for invalid environment variable values. A malicious
    environment variable value could exploit this behavior to set a
    value for a different environment variable. For example, the
    environment variable string "A=B\x00C=D" set the variables "A=B" and
    "C=D".

    Thanks to RyotaK (https://twitter.com/ryotkak) for reporting this
    issue.

    This is CVE-2022-41716 and Go issue https://go.dev/issue/56284.

This Go release also fixes https://github.com/golang/go/issues/56309, a runtime bug which can cause random memory corruption when a goroutine exits with runtime.LockOSThread() set. This fix is necessary to unblock work to replace certain uses of pkg/reexec with unshared OS threads.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1
 
 ARG BASE_VARIANT=alpine
-ARG GO_VERSION=1.18.7
+ARG GO_VERSION=1.18.8
 ARG XX_VERSION=1.1.0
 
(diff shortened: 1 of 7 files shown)


thaJeztah:
Let me bring this one in; as we already updated the other repositories

PR #3867, opened 2022-11-17
Some fixes in handling YAML files, and updates to tests

thaJeztah:
Some changes to assist in updating dependencies; see individual commits for details.

### cli/compose/types: UlimitsConfig.MarshalYAML() fix recursion

When marshaling the type with `gopkg.in/yaml.v3`, unmarshaling would
recursively call the type's `MarshalYAML()` function, which ultimately
resulted in a crash:

    runtime: goroutine stack exceeds 1000000000-byte limit
    runtime: sp=0x140202e0430 stack=[0x140202e0000, 0x140402e0000]
    fatal error: stack overflow

This applies a similar fix as was implemented in e7788d6f9a61d67566518efe11e394c414a21173
for the `MarshalJSON()` implementation. An alternative would be to use
a type alias (to remove the `MarshalYAML()`), but keeping it simple.

### cli-plugins/manager: TestPluginError: don't use yaml.Marshal

The test used  `gopkg.in/yaml.v2` to verify the TextMarshaller implementation,
which was implemented to allow printing the errors in JSON formatted output;

> This exists primarily to implement encoding.TextMarshaller such that
> rendering a plugin as JSON (e.g. for `docker info -f '{{json .CLIPlugins}}'`)
> renders the Err field as a useful string and not just `{}`.

Given that both yaml.Marshal and json.Marshal use this, we may as well use
Go's stdlib.

While updating, also changed some of the assertions to checks, so that we don't
fail the test early.

### cli/compose/loader: fix error message, and various tests

Various fixes:

- Don't capitalize error message
- Rename variables that collided with imports or types
- Prefer assert.Check over assert.Assert to prevent tests covering multiple
  cases from failing early
- Fix inconsistent order of expected <--> actual, which made it difficult to
  check which output was the expected output.

### cli/command/stack: TestConfigMergeInterpolation: various fixes
- 
- Make the package-level configMergeTests local to the test itself.
- Rename fields to better describe intent
- Remove some redundant variables
- Reverse "expected" and "actual" fields for consistency
- Use assert.Check() to not fail early

### DIFF
--- a/cli-plugins/manager/error_test.go
+++ b/cli-plugins/manager/error_test.go
@@ -1,24 +1,24 @@
 package manager
 
 import (
+	"encoding/json"
 	"fmt"
 	"testing"
 
-	"github.com/pkg/errors"
-	"gopkg.in/yaml.v2"
 	"gotest.tools/v3/assert"
+	is "gotest.tools/v3/assert/cmp"
 )
 
 func TestPluginError(t *testing.T) {
 	err := NewPluginError("new error")
-	assert.Error(t, err, "new error")
+	assert.Check(t, is.Error(err, "new error"))
 
 	inner := fmt.Errorf("testing")
 	err = wrapAsPluginError(inner, "wrapping")
-	assert.Error(t, err, "wrapping: testing")
-	assert.Assert(t, errors.Is(err, inner))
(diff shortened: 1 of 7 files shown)


thaJeztah:
Thx! Bringing this one in

PR #3868, opened 2022-11-17
context: improvements for handling non-existing contexts and errors

thaJeztah:
- This is a subset of https://github.com/docker/cli/pull/3847, without the lazy loading part


cli/command/context/show.go
```diff
@@ -25,10 +25,10 @@ func newShowCommand(dockerCli command.Cli) *cobra.Command {
...
 	context := dockerCli.CurrentContext()
-	metadata, err := dockerCli.ContextStore().GetMetadata(context)
+	fmt.Fprintln(dockerCli.Out(), context)
+	_, err := dockerCli.ContextStore().GetMetadata(context)
 	if err != nil {
 		return err
 	}
-	fmt.Fprintln(dockerCli.Out(), metadata.Name)
 	return nil
 }
```
thaJeztah, line 34: As this is a new comment, we can still change; do we think we should produce an error _at all_ if the context wasn't found, or just "show the name of the context" (basically: this is what's _configured_ to use, but you'll learn if it works once you use it)?

We could still opt for printing the error (on STDERR) as a warning, but make the command always return a zero exit code ("success").

sam-thibault: It seems like it should mimic existing behavior for other commands.  e.g. for `inspect`:
```
if dockerCli.CurrentContext() == "" {
	return errors.New("no context specified")
}
```

cli/command/context/testdata/list.golden
```diff
@@ -1,5 +1,5 @@
-NAME        DESCRIPTION                               DOCKER ENDPOINT
-current *   description of current                    https://someswarmserver.example.com
-default     Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
-other       description of other                      https://someswarmserver.example.com
-unset       description of unset                      https://someswarmserver.example.com
+NAME        DESCRIPTION                               DOCKER ENDPOINT                       ERROR
```
sam-thibault, line 1: `ERROR` looks like a copy/paste mistake here

cli/command/context/show.go
thaJeztah: Hm... right; so `docker context inspect` expects a positional argument. It looks like there's a bug in the `compose-cli` wrapper, as it hijacks the `docker context` subcommands, and it looks like the usage information is incorrect;

This is the output for compose-cli;

```bash
docker context inspect --help
Display detailed information on one or more contexts

Usage:
  docker context inspect [flags]

Flags:
  -f, --format string   Format the output using the given Go template
  -h, --help            Help for inspect
```

And the output for the actual docker cli;

```bash
com.docker.cli context inspect --help

Usage:  docker context inspect [OPTIONS] [CONTEXT] [CONTEXT...]

Display detailed information on one or more contexts

Options:
  -f, --format string   Format the output using the given Go template
```

thaJeztah: As to the `docker context show`; that command was added in https://github.com/docker/cli/pull/3567. Basically its sole purpose is to show what context is _configured_, so that it can be used in prompts. An earlier PR (https://github.com/docker/cli/pull/2500) added it to `docker info` for that purpose, but `docker context show` is more convenient.

As it (likely) will be used for prompts, the question is, should it always validate? Without validation, the _value_ is still valid (it shows what's configured), but even if we check if the context is valid (can be found, and decoded), it may still be pointing to a host that's not reachable.

cli/command/context/testdata/list.golden
thaJeztah: It's the title for the `ERROR` column that was added, but looking at this, it looks like I have more work to do. The context store itself already fails "full stop" if it encounters an error, so without some of the other changes from https://github.com/docker/cli/pull/3847, it looks like this doesn't work 😞

thaJeztah:
Hm... more bugs (or at least differences in behaviour) in compose cli; looks like it's handling contexts differently, and happily goes ahead with empty contexts :thinking:

```bash
docker context ls
NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT                                   KUBERNETES ENDPOINT   ORCHESTRATOR
                    moby
default             moby                Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                                             swarm
desktop-linux *     moby                                                          unix:///Users/thajeztah/.docker/run/docker.sock
```

```bash
com.docker.cli context ls
cannot find docker endpoint in context
```

thaJeztah:
And ... buildx does it different as well;

```bash
make shell
# build dockerfile from stdin so that we don't send the build-context; source is bind-mounted in the development environment
cat ./dockerfiles/Dockerfile.dev | docker build  --build-arg=GO_VERSION -t docker-cli-dev -
ERROR: use `docker --context= buildx` to switch to context
make[1]: *** [build_docker_image] Error 1
make: *** [dev] Error 2
```

Looks to be related to context store handling / ignoring name, which then causes it to use "" as context name, instead of the default; https://github.com/docker/buildx/blob/6e9b743296d8ba029871516370909c6f809dd935/commands/util.go#L201-L225

thaJeztah:
Closing as all of these were merged

PR #3870, opened 2022-11-17
Update version to v23.0.0-dev

thaJeztah:
As a stepping-stone towards switching to go modules, we're switching the
versioning scheme to be SemVer(ish) instead of CalVer. Note that this
currently still is `+incompatible` version, as changing to a version > 1
requires the module to be renamed, but changes in that area will arrive
in a future release.


vvoland (approved):

thaJeztah:
thx! Let me get this one in at least

PR #3884, opened 2022-11-28
cli/command: DockerCli: CurrentContext(): resolve dynamically

thaJeztah:
- subset of https://github.com/docker/cli/pull/3847

Store a reference to the options in the client, so that they can be used for later use. Now that resolveContextName() is a very lightweight function we can also dynamically get its value and use it for cli.CurrentContext().

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -68,6 +68,7 @@ type Cli interface {
 // Instances of the client can be returned from NewDockerCli.
 type DockerCli struct {
 	configFile         *configfile.ConfigFile
+	options            *cliflags.ClientOptions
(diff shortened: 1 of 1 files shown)


thaJeztah:
FWIW; moved this back to draft, as I'm still on the fence if we should resolve the context once at startup, or dynamically. Reason is that other properties (e.g. "DockerEndpoint" may _not_ be resolved dynamically, so it would not make much sense to resolve this one dynamically.

So I may reduce this change back to only storing the options.

thaJeztah:
closing in favour of https://github.com/docker/cli/pull/3847

PR #3886, opened 2022-11-28
cli/command/context: "docker context show": don't validate context

thaJeztah:
- subset of https://github.com/docker/cli/pull/3847

The "docker context show" command is intended to show the currently configured
context. While the context that's configured may not be valid (e.g., in case
an environment variable was set to configure the context, or if the context
was removed from the filesystem), we should still be able to _show_ the
context.

This patch removes the context validation, and instead only shows the context.
This can help in cases where the context is used to (e.g.) set the command-
prompt, but the user removed the context. With this change, the context name
can still be shown, but commands that _require_ the context will still fail.

### DIFF
--- a/cli/command/context/show.go
+++ b/cli/command/context/show.go
@@ -16,19 +16,14 @@ func newShowCommand(dockerCli command.Cli) *cobra.Command {
 		Short: "Print the name of the current context",
 		Args:  cli.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runShow(dockerCli)
+			runShow(dockerCli)
+			return nil
 		},
 		ValidArgsFunction: completion.NoComplete,
 	}
 	return cmd
 }
 
-func runShow(dockerCli command.Cli) error {
-	context := dockerCli.CurrentContext()
-	metadata, err := dockerCli.ContextStore().GetMetadata(context)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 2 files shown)


cli/command/context/show.go
```diff
@@ -25,10 +25,10 @@ func newShowCommand(dockerCli command.Cli) *cobra.Command {
 
 func runShow(dockerCli command.Cli) error {
 	context := dockerCli.CurrentContext()
-	metadata, err := dockerCli.ContextStore().GetMetadata(context)
+	fmt.Fprintln(dockerCli.Out(), context)
+	_, err := dockerCli.ContextStore().GetMetadata(context)
 	if err != nil {
 		return err
 	}
```
thaJeztah, line 32: From the discussion on https://github.com/docker/cli/pull/3868#discussion_r1025219461

> As this is a new command, we can still change; do we think we should produce an error _at all_ if the context wasn't found, or just "show the name of the context" (basically: this is what's _configured_ to use, but you'll learn if it works once you use it)?
> 
> We could still opt for printing the error (on STDERR) as a warning, but make the command always return a zero exit code ("success").

silvin-lubecki (approved):

thaJeztah:
Thx! Bringing this one in. We'll probably need to do follow-ups to align compose-cli with the changes we made in the cli itself.

PR #3901, opened 2022-12-05
Fix bug where incorrect response is returned [carry 3845]

thaJeztah:
- depends on https://github.com/docker/cli/pull/3903
- replaces https://github.com/docker/cli/issues/3845
- fixes https://github.com/docker/cli/issues/3844

When server is unreachable and docker checkpoint (or any command that needs to check the server type) is run, incorrect error was returned.

When checking if the daemon had the right OS, we compared the OSType from the clients ServerInfo(). In situations where the client cannot connect to the daemon, a "stub" Info is used for this, in which we assume the daemon has experimental enabled, and is running the latest API version.

However, we cannot fill in the correct OSType, so this field is empty in this situation.

This patch only compares the OSType if the field is non-empty, otherwise assumes the platform matches.

before this:

    docker -H unix:///no/such/socket.sock checkpoint create test test
    docker checkpoint create is only supported on a Docker daemon running on linux, but the Docker daemon is running on

with this patch:

    docker -H unix:///no/such/socket.sock checkpoint create test test
    Cannot connect to the Docker daemon at unix:///no/such/socket.sock. Is the docker daemon running?

### DIFF
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -306,7 +306,7 @@ func hideSubcommandIf(subcmd *cobra.Command, condition func(string) bool, annota
 func hideUnsupportedFeatures(cmd *cobra.Command, details versionDetails) error {
 	var (
 		notExperimental = func(_ string) bool { return !details.ServerInfo().HasExperimental }
-		notOSType       = func(v string) bool { return v != details.ServerInfo().OSType }
+		notOSType       = func(v string) bool { return details.ServerInfo().OSType != "" && v != details.ServerInfo().OSType }
 		notSwarmStatus  = func(v string) bool {
 			s := details.ServerInfo().SwarmStatus
 			if s == nil {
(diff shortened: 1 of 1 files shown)


cmd/docker/docker.go
```diff
@@ -419,10 +419,11 @@ func areSubcommandsSupported(cmd *cobra.Command, details versionDetails) error {
 		if cmdVersion, ok := curr.Annotations["version"]; ok && versions.LessThan(details.CurrentVersion(), cmdVersion) {
 			return fmt.Errorf("%s requires API version %s, but the Docker daemon API version is %s", cmd.CommandPath(), cmdVersion, details.CurrentVersion())
 		}
-		if ost, ok := curr.Annotations["ostype"]; ok && ost != details.ServerInfo().OSType {
-			return fmt.Errorf("%s is only supported on a Docker daemon running on %s, but the Docker daemon is running on %s", cmd.CommandPath(), ost, details.ServerInfo().OSType)
+		si := details.ServerInfo()
+		if ost, ok := curr.Annotations["ostype"]; ok && si.OSType != "" && ost != si.OSType {
```
thaJeztah, line 423: In cases where we were not able to connect with the daemon, we set `Experimental` to "true", but as we don't know the daemon's `OSType`, we can't set that value, so in that case we should ignore these; https://github.com/docker/cli/blob/7240f708087dbea6120477ceb2dcf766d4407804/cli/command/cli.go#L342-L343


We can probably optimise and detect that we didn't have a valid ping, and skip all of these checks, but leaving that for a follow-up.

thaJeztah:
Looks like I have one test to fix / verify;

```
#19 64.01 === FAIL: cmd/docker TestBuildkitDisabled (0.00s)
#19 64.01     builder_test.go:137: line 0: expected "" to end with "DEPRECATED: The legacy builder is deprecated and will be removed in a future release."
#19 64.01     builder_test.go:137:
```

thaJeztah:
Hm.. right, so the "unit" test is actually using a docker client, which tries to connect to a daemon (which isn't there);

```
=== FAIL: cmd/docker TestBuildkitDisabled (0.00s)
#19 76.40 ping err: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

Then the check if buildkit must be used is based on the (non-existing) ping result; if that's "windows", it's okay, but it it's non-windows, then it should print an error. As (at this stage) we don't know if the daemon is windows or linux, we probably shouldn't produce an error (which is what this PR does), but the test expects that error to happen.

vvoland (approved):

thaJeztah:
Thx!

cmd/docker/docker.go
```diff
@@ -419,10 +419,11 @@ func areSubcommandsSupported(cmd *cobra.Command, details versionDetails) error {
 		if cmdVersion, ok := curr.Annotations["version"]; ok && versions.LessThan(details.CurrentVersion(), cmdVersion) {
 			return fmt.Errorf("%s requires API version %s, but the Docker daemon API version is %s", cmd.CommandPath(), cmdVersion, details.CurrentVersion())
 		}
-		if ost, ok := curr.Annotations["ostype"]; ok && ost != details.ServerInfo().OSType {
-			return fmt.Errorf("%s is only supported on a Docker daemon running on %s, but the Docker daemon is running on %s", cmd.CommandPath(), ost, details.ServerInfo().OSType)
+		si := details.ServerInfo()
```
thaJeztah, line 422: Whoops, and I just realised I introduced a regression here; this was deliberately put inline to allow it to be executed conditionally (see https://github.com/docker/cli/commit/b39739123b845f872549e91be184cc583f5b387c / https://github.com/docker/cli/pull/2424)

Opening a PR to fix that 😅

PR #3885, opened 2022-11-28
cli/command: add Cli.CurrentVersion() function

thaJeztah:
- subset of https://github.com/docker/cli/pull/3847

This internalizes constructing the Client(), which allows us to provide fallbacks when trying to determin the current API version.

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3902, opened 2022-12-05
cli/command: use client.CurrentVersion()

thaJeztah:
Make sure the new CurrentVersion() function, which was added in a7e2c3ea1edd195df98cb62bc353fcbf034764cd (https://github.com/docker/cli/pull/3885), but it looks like I overlooked these uses of what it was replacing 😅 


silvin-lubecki (approved):
LGTM

thaJeztah:
Thanks! I want to have another look at this; there's many places left where we call this (through `Client().ClientVersion()`), but in most places the `Client()` is also needed for other purposes (so it wouldn't make much sense to replace it).

That said, I want to look what makes most sense for various locations;

- Pass the whole `DockerCLI` around (which includes the `Client()` API Client
- Or if we should exactly do the reverse; only pass the APIClient

The reason the whole DockerCLI is passed around (in _MOST_ cases), is only to get the CLI's `.Out()` or `.Err()` for printing messages. Perhaps we should pass something through the context or use another approach.

PR #3903, opened 2022-12-05
cil/command: use dummy client for build-tests

thaJeztah:
- relates to https://github.com/docker/cli/pull/3901

These tests were using the default client, which would try to make a connection
with the daemon (which isn't running). Some of these test subsequently had
tests that depended on the result of that connection (i.e., "ping" result).

This patch updates the test to use a dummy client, so that the ping result is
predictable.

### DIFF
--- a/cli/command/cli_options.go
+++ b/cli/command/cli_options.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 
 	"github.com/docker/cli/cli/streams"
+	"github.com/docker/docker/client"
 	"github.com/moby/term"
 )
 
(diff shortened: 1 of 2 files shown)


thaJeztah:
@crazy-max PTAL 🤗 (relates to https://github.com/docker/cli/pull/3901#issuecomment-1338140950)

cli/command/cli_options.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strconv"
 
 	"github.com/docker/cli/cli/streams"
+	"github.com/docker/docker/client"
```
crazy-max, line 9: Not sure if this is correct but plugins are going to import the whole API client pkg now right?

thaJeztah: hm... let me check if it makes a difference (I somewhat expect it to already be in here, but maybe not).

crazy-max (approved):

thaJeztah:
Thx!

PR #3905, opened 2022-12-06
cmd/docker: improve error message if BUILDKIT_ENABLED=0

thaJeztah:
Before this change, the error would suggest installing buildx:

    echo "FROM scratch" | DOCKER_BUILDKIT=0  docker build -
    DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
                Install the buildx component to build images with BuildKit:
                https://docs.docker.com/go/buildx/

    ...

However, this error would also be shown if buildx is actually installed, but disabled through "DOCKER_BUILDKIT=0";

    docker buildx version
    github.com/docker/buildx v0.9.1 ed00243

With this patch, it reports that it's disabled, and how to fix:

    echo "FROM scratch" | DOCKER_BUILDKIT=0  docker build -
    DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
                BuildKit is currently disabled; enabled it by removing the DOCKER_BUILDKIT=0
                environment-variable.

### DIFF
--- a/cmd/docker/builder.go
+++ b/cmd/docker/builder.go
@@ -18,34 +18,30 @@ const (
 	builderDefaultPlugin = "buildx"
 	buildxMissingWarning = `DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
             Install the buildx component to build images with BuildKit:
-            https://docs.docker.com/go/buildx/
-`
+            https://docs.docker.com/go/buildx/`
+
+	buildkitDisabledWarning = `DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
+            BuildKit is currently disabled; enabled it by removing the DOCKER_BUILDKIT=0
+            environment-variable.`
 
 	buildxMissingError = `ERROR: BuildKit is enabled but the buildx component is missing or broken.
        Install the buildx component to build images with BuildKit:
-       https://docs.docker.com/go/buildx/
-`
+       https://docs.docker.com/go/buildx/`
 )
 
-func newBuilderError(warn bool, err error) error {
-	var errorMsg string
-	if warn {
-		errorMsg = buildxMissingWarning
-	} else {
-		errorMsg = buildxMissingError
-	}
-	if pluginmanager.IsNotFound(err) {
+func newBuilderError(errorMsg string, pluginLoadErr error) error {
+	if pluginmanager.IsNotFound(pluginLoadErr) {
 		return errors.New(errorMsg)
 	}
-	if err != nil {
-		return fmt.Errorf("%w\n\n%s", err, errorMsg)
+	if pluginLoadErr != nil {
+		return fmt.Errorf("%w\n\n%s", pluginLoadErr, errorMsg)
 	}
-	return fmt.Errorf("%s", errorMsg)
+	return errors.New(errorMsg)
 }
 
 //nolint:gocyclo
 func processBuilder(dockerCli command.Cli, cmd *cobra.Command, args, osargs []string) ([]string, []string, []string, error) {
-	var useLegacy, useBuilder, useAlias bool
+	var buildKitDisabled, useBuilder, useAlias bool
 	var envs []string
 
 	// check DOCKER_BUILDKIT env var is present and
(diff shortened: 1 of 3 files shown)


thaJeztah:
Sigh

```
#14 488.7 	gopkg.in/yaml.v2@v2.4.0: reading gopkg.in/yaml.v2/go.mod at revision v2.4.0: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /go/pkg/mod/cache/vcs/748bced43cf7672b862fbc52430e98581510f4f2c34fb30c0064b7102a68ae2c: exit status 128:
#14 488.7 	remote: Cannot obtain refs from GitHub: cannot talk to GitHub: Get https://github.com/go-yaml/yaml.git/info/refs?service=git-upload-pack: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
#14 488.7 	fatal: unable to access 'https://gopkg.in/yaml.v2/': The requested URL returned error: 502
```

thaJeztah:
@crazy-max @vvoland ptal 🤗

vvoland (approved):
LGTM

The calls with boolean arguments
```newBuilderError(false, false, perr)```
are a bit hard to understand though.

The inlined messages seem more readable to me. The current version might be version for writing new error cases though.

Anyway, this could be mostly subjective, so feel free to ignore this 😄 .
(I made the suggestions just for my understanding of which message will be used)

**Comment by thaJeztah** [comment]
> The call with boolean arguments `newBuilderError(false, false, perr)` are a bit hard to understand though.
> 
> The inlined messages seem more readable to me. The current version might be version for writing new error cases though.
> 
> Anyway, this could be mostly subjective, so feel free to ignore this 😄 . (I made the suggestions just for my understanding of which message will be used)

yes, this whole code is kinda ugly; I also looked at whether I should change it or not; let me have a quick look at your suggestions locally to see what it looks like.

File: cmd/docker/builder.go [hunk 1]
```diff
@@ -100,10 +108,10 @@ func processBuilder(dockerCli command.Cli, cmd *cobra.Command, args, osargs []st
 	if perr != nil {
 		// if builder enforced with DOCKER_BUILDKIT=1, cmd must fail if plugin missing or broken
 		if useBuilder {
-			return args, osargs, nil, newBuilderError(false, perr)
+			return args, osargs, nil, newBuilderError(false, false, perr)
 		}
 		// otherwise, display warning and continue
-		_, _ = fmt.Fprintln(dockerCli.Err(), newBuilderError(true, perr))
+		_, _ = fmt.Fprintln(dockerCli.Err(), newBuilderError(true, buildKitDisabled, perr))
```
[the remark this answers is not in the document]
thaJeztah: Actually, looking at this part again, I don't think we should print the "buildit is disabled through DOCKER_BUILDKIT=0" warning here; that case is already handled above (where we print the warning for non-windows daemons and return).

If we reach this point it means buildkit was _not_ disabled, but we failed to find the plugin, so we should print the "buildx is not is not installed message".

I applied your suggestions locally, and agree that it looks better; let me have a look at this specific case, and push a commit

cmd/docker/builder.go
```diff
@@ -18,34 +18,30 @@ const (
 	builderDefaultPlugin = "buildx"
 	buildxMissingWarning = `DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
             Install the buildx component to build images with BuildKit:
-            https://docs.docker.com/go/buildx/
-`
```
thaJeztah, line 22: Had to remove the extra newline here, otherwise it was complaining if we used `Println` with a message that contains a newline at the end.

cmd/docker/builder.go
```diff
@@ -84,10 +80,10 @@ func processBuilder(dockerCli command.Cli, cmd *cobra.Command, args, osargs []st
 		return args, osargs, nil, nil
 	}
 
-	if useLegacy {
+	if buildKitDisabled {
 		// display warning if not wcow and continue
 		if dockerCli.ServerInfo().OSType != "windows" {
-			_, _ = fmt.Fprintln(dockerCli.Err(), newBuilderError(true, nil))
+			_, _ = fmt.Fprintf(dockerCli.Err(), "%s\n\n", buildkitDisabledWarning)
```
thaJeztah, line 86: ... so I added the extra newlines back here when _printing_ (if returned as an error, we don't have the trailing newline.

vvoland (approved):
LGTM

PR #3910, opened 2022-12-07
docs: document what AuthConfig.Auth is

davidhsingyuchen:
PR enables folks to understand what `Auth` is without the need to trace code to: 

https://github.com/docker/cli/blob/65d3f7830d8f73aeabe649ca17386f5ae5655210/cli/config/configfile/file.go#L228-L229

Signed-off-by: Hsing-Yu (David) Chen <davidhsingyuchen@gmail.com>

### DIFF
--- a/cli/config/types/authconfig.go
+++ b/cli/config/types/authconfig.go
@@ -4,7 +4,8 @@ package types
 type AuthConfig struct {
 	Username string `json:"username,omitempty"`
 	Password string `json:"password,omitempty"`
-	Auth     string `json:"auth,omitempty"`
+	// Auth is the base64 encoding of the concatenation of username and password.
+	Auth string `json:"auth,omitempty"`
 
 	// Email is an optional value associated with the username.
 	// This field is deprecated and will be removed in a later


thaJeztah:
Thank you!

I want to double-check how this field is used, to be sure it's documented correctly as there's been some changes over time, and I'm not sure currently if this would always be `username:password`. These fields would be used to authenticate with different registries, and Authentication is not the same for all of them (which probably is one of the the reasons the field hasn't been documented so far).

I did a first search in history, but didn't completely finish that yet, but let me copy my notes here (in case I want to continue searching or someone else wants to);

- This type (in docker/cli) was taken from the engine's API types (done in https://github.com/docker/cli/pull/1642 to simplify some dependency issues). 
- At some point the `IdentityToken` was introduced (https://github.com/docker/engine-api/pull/108), but I need to verify if that implementation actually was fully implemented all the way through. ISTR there were situations where the _password_ field is used to store a token, and https://github.com/docker/engine-api/pull/108#discussion_r55111412 also mentions some things about this. And, there's the `RegistryToken` field (which is confusing to have _both_ here); https://github.com/moby/moby/commit/8dce8e9901501863a80caf47d71713d4d36925a3
- The type moved around various times (e.g., https://github.com/moby/moby/pull/18637, https://github.com/moby/moby/commit/5b321e328769cc93c3454e82ec3fe07672156f2e, https://github.com/moby/moby/commit/bb9da6ba9294a8eab8f4dfaf7cf07c57959fe608), which makes it a bit more difficult to get the full picture.

PR #3923, opened 2022-12-18
Added missing backslash to documentation cli snippet

dozjul:
I think the cli code block misses a backslash to brevent line break when copy/pasting it to a terminal. I doubt that this is intentional, if it is, feel free to reject the pr.

Signed-off-by: Julian <gitea+julian@ic.thejulian.uk>

**- What I did**

I added a (likely missing) backslash to a console code block, I notice while reading through the page on docs.docker.com
I think the code block should be copy-pastable, without the backslash it interprets the line break as end of command, which likely isn't intended (*if it is, feel free to reject the pull request*).

**- How I did it**

added a backslash (nothing other then that changed)

**- How to verify it**

don't know look at it 🤷 

added missing backslash `\` to code block in documentation for `docker service create`

### DIFF
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -291,7 +291,7 @@ service with two labels:
 ```console
 $ docker service create \
   --name redis_2 \
-  --label com.example.foo="bar"
+  --label com.example.foo="bar" \
   --label bar=baz \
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM, nice catch, thanks!

PR #3943, opened 2023-01-04
cli/rm_test: Fix TestRemoveForce race condition

vvoland:
Fixes:
- https://github.com/docker/cli/issues/3916

There was already an attempt to fix this test by https://github.com/docker/cli/pull/3520 but it seems that it only accounted for the races between the test cases, and not the `container rm` itself calling `containerRemoveFunc` concurrently.

Synchronize append on the `removed` slice with mutex because `containerRemoveFunc` is called in parallel for each removed container by `container rm` cli command.
Also reduced the shared access area by separating the scopes of test cases.

Signed-off-by: Paweł Gronowski <pawel.gronowski@docker.com>

### DIFF
--- a/cli/command/container/rm_test.go
+++ b/cli/command/container/rm_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"sort"
+	"sync"
 	"testing"
 
 	"github.com/docker/cli/internal/test"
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM, thanks!

PR #3952, opened 2023-01-07
docs: inspect: remove trailing whitespace from example

thaJeztah:
Current versions of the docs generator take this into account, but on the 20.10 branch, the trailing whitespace can make the YAML generator switch to use "compact" formatting, which is hard to read, and hard to review diffs when updating.


crazy-max (approved):

thaJeztah:
Thx! Bringing this one in 👍

PR #3965, opened 2023-01-12
contrib/completion: bash `nounset` mode fixes

scop:
**- What I did**

Fixed all instances I've come across that produce errors with the bash completion and the shell in `nounset` mode.

**- How I did it**

Guarded variable accesses like `${something-}` instead of directly accessing `$something` or `${something}` where the variable can be unset.

**- How to verify it**

Set bash in `nounset` mode (`set -u` or `set -o nounset`), use the bash completion. Observe errors before this PR, no errors after.

A number of problems using the bash completion with the shell in `nounset` mode have been fixed.

:=}

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -84,13 +84,13 @@ __docker_q() {
 # precedence over the environment setting.
 __docker_configs() {
 	local format
-	if [ "$1" = "--id" ] ; then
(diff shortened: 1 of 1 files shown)


vvoland (approved):
LGTM, thanks!

albers (approved):
LGTM, thanks for your contribution.

PR #3968, opened 2023-01-13
README: fix badges

crazy-max:
GHA badges are sad in our README.


thaJeztah (approved):
LGTM


Curious; did we always use the wrong URL, or did GitHub change URLs?

PR #3971, opened 2023-01-15
cli: pass dockerCLI's in/out/err to cobra cmds

thaJeztah:
Both the DockerCLI and Cobra Commands provide accessors for Input, Output, and Error streams (usually STDIN, STDOUT, STDERR). While we were already passing DockerCLI's Output to Cobra, we were not doing so for the other streams (and were passing none for plugin commands), potentially resulting in DockerCLI output/input to mean something else than a Cobra Command's intput/output/error.

This patch sets them to the same streams when constructing the Cobra command.


thaJeztah:
Thx! Bringing this one in 👍

PR #3980, opened 2023-01-19
run plugin with context so they can be canceled by SIGTERM

ndeloof:
closes https://github.com/docker/compose/issues/10179

**- What I did**
Create plugin command with context, so that child process get stopped on SIGTERM

**- How to verify it**

create compose.yaml file
```yaml
services:
  test:
    image: alpine
    command: ping localhost
```
run `docker compose logs -f`
on a second terminal, send TERM signal to `docker` CLI process `kill -s SIGTERM <ID>`

image

### DIFF
--- a/cli-plugins/manager/manager.go
+++ b/cli-plugins/manager/manager.go
@@ -9,6 +9,7 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/config"
 	"github.com/fvbommel/sortorder"
+	"github.com/moby/buildkit/util/appcontext"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


glours (approved):
Works for me and fix https://github.com/docker/compose/issues/10179 👍

ndeloof:
doc says "The provided context is used to kill the process (by calling os.Process.Kill)" which is not what we expect either, as SIGINT/SIGTERM signal should be propagated to child process, not replaced by SIGKILL

`tryPluginRun` explicitly override the signal handler "so we let the plugin shut down first" but it seems as signal is sent by a `kill` command it is not automatically propagated to child process, while a signal sent by shell is.

PR #3989, opened 2023-01-25
feat(zsh): support for gain-privileges (automatic sudo)

vincentbernat:
Currently, this is not possible to make completion work when using sudo. Zsh supports the "gain-priveleges" feature for this exact use case. By using the `-p` argument of `_call_program`, if correctly set with zstyles, Zsh automatically tries to invoke sudo or doas if the command itself is invoked with sudo or doas. This makes the completion works in this context:

    sudo docker rm a<tab>

It seems to be a good practice to distinguish between the various cases where sudo can be invoked by customizing the tag provided to `_call_program`. Instead of `commands`, we therefore use `docker-X` where X is the subcommand invoked.

To optin this feature, a user needs something like this in its configuration file:

    zstyle ':completion:*:docker/*' gain-privileges yes
    zstyle ':completion:*:docker-*/*' gain-privileges yes

Signed-off-by: Vincent Bernat <vincent@bernat.ch>

Support for gain-privileges style for Zsh completion

### DIFF
--- a/contrib/completion/zsh/_docker
+++ b/contrib/completion/zsh/_docker
@@ -42,6 +42,10 @@
 # Short-option stacking can be enabled with:
 #  zstyle ':completion:*:*:docker:*' option-stacking yes
 #  zstyle ':completion:*:*:docker-*:*' option-stacking yes
+# Automatic sudo for completion can be enabled with:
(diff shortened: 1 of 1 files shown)


vvoland (approved):
LGTM

PR #3992, opened 2023-01-27
docs: add note about MKE CA rotation, which is potentially dangerous

neersighted:
- Follow-up to https://github.com/docker/docs/pull/16596

Add a call-out for Mirantis Kubernetes Engine/Docker UCP users re: Swarm CA rotation, which is not supported in the general case and usually dangerous/indicative of other cluster health problems.


neersighted:
cc @dvdksn

thaJeztah (approved):
LGTM, thanks!

PR #3994, opened 2023-01-30
Add manifest merging support

jedevc:
Requires https://github.com/docker/cli/pull/3990. 

**- What I did**

Introduced support for to merge manifest lists and OCI indexes to the `docker manifest` command.

**- How I did it**

We refactor the manifest store to save and load groups of manifests, instead of only one. On disk, these manifests take the form of suffixed indexes, e.g. "_2", "_3", etc (note that we ignore "_1", which is permitted to allow a backwards-compatible change with the previous disk format).
    
Then, only the commands need small updates - "create" now looks up multiple manifests and saves multiple manifests, while "annotate" now annotates all the matching manifests.

**- How to verify it**

Use docker manifest with multiple docker manifests - e.g. `docker manifest create library/ubuntu library/alpine`

Add support for merging manifest lists to docker manifest

image


thaJeztah:
This one probably can be rebased now (as the other PR was merged)

cli/command/manifest/annotate.go
```diff
@@ -58,38 +58,40 @@ func runManifestAnnotate(dockerCli command.Cli, opts annotateOptions) error {
...
-	if opts.variant != "" {
-		imageManifest.Descriptor.Platform.Variant = opts.variant
-	}
-	if opts.osVersion != "" {
-		imageManifest.Descriptor.Platform.OSVersion = opts.osVersion
-	}
-
-	if !isValidOSArch(imageManifest.Descriptor.Platform.OS, imageManifest.Descriptor.Platform.Architecture) {
-		return errors.Errorf("manifest entry for image has unsupported os/arch combination: %s/%s", opts.os, opts.arch)
+	for i, imageManifest := range imageManifests {
```
jedevc, line 69: Hm, in hindsight, I'm not convinced about annotate looping through every manifest.

e.g. assuming an image with BuildKit attestations, the platform would be `unknown/unknown`, so updating the platform is probably not the right idea?

I guess annotation only makes sense over a single manifest, so when multiple are found, we should error out?

cli/manifest/store/store.go
```diff
@@ -152,28 +151,3 @@ func makeFilesafeName(ref string) string {
...
-func newNotFoundError(ref string) *notFoundError {
-	return &notFoundError{object: ref}
-}
-
-func (n *notFoundError) Error() string {
-	return fmt.Sprintf("No such manifest: %s", n.object)
-}
-
-// NotFound interface
-func (n *notFoundError) NotFound() {}
```
thaJeztah, line 169: Wondering; instead of a new error-type, could we change the errors to use these, and use `errdefs.IsNotFound()` (etc)? https://github.com/docker/cli/blob/c1733165159c08101adb0e1f120c7181533550ef/vendor/github.com/docker/docker/errdefs/is.go#L31-L35

Issue #3716, opened 2022-07-22
`docker build` handles credentials differently than `docker pull`


pbecotte:
If a credHelper is defined, 'docker build' with a private FROM image fails, while `docker pull` of the same image works correctly.

**Steps to reproduce the issue:**
1.  'docker login' to your private repository (we are using artifactory)
2. Add a credHelper to config.json - for this I went with '{"credHelpers": {"blah.com": "false"}}'
3. Make a dockerfile with a FROM for a non-existant tag in a private image of your private repo
4. `docker build` - you will get the error "the client does not have permission for the manifest"
5. `docker pull` - you will get the error "The named manifest is not known to the registry"
6. Remove the credHelper - now `docker build` will get "The named manifest is not known to the registry"

**Describe the results you expected:**
'docker build' to be able to pull images the same as 'docker pull'

**Additional information you deem important (e.g. issue happens only occasionally):**
From our nginx logs, we can see that when the daemon requests a token it makes a request like '/v2/token?account=username' when running 'docker pull', but '/v2/token?scope=repository...' when running 'docker build'. I do not have access to the daemon logs, but the token request on 'docker build' appears to be completely unauthenticated. Removing the credHelper (whether it works or doesn't seems to be immaterial) changes the behavior- in that case, the daemon submits a correctly authenticated request for a token.

**Output of `docker version`:**

Client-
Version: 20.10.2
API: 1.39
OS: linux/amd64

Server:
Version: 20.10.2
API: 1.41

thaJeztah:
I see you're running an older version of docker 20.10;

- are you still able to reproduce this issue on the latest patch release of 20.10?
- could you update the information provided to include the _full_ output (it's ok to redact sensitive data) of `docker version` and `docker info`?
- are you building with buildkit enabled (`DOCKER_BUILDKIT=1`), or using the classic builder?

pbecotte:
Our environment is quite locked down. The base OS is a frozen rhel8 image with an immutable root filesystem- can't upgrade the daemon. I also cannot copy and paste more than a couple lines of text, which is why I tried to include the important parts, but will take the time to get the rest later.

Using the classic builder.

pbecotte:
Actually tracked it down- is still in the source code.

https://github.com/docker/docker-ce/blob/master/components/cli/cli/config/configfile/file.go#L357

[...]

PR #3973, opened 2023-01-15
cli: additionalHelp() don't decorate output if it's piped, and add extra newline

thaJeztah:
Relates to:

- https://github.com/docker/cli/pull/2857
- https://github.com/docker/cli/pull/2884
- https://github.com/docker/docs/issues/12488
- https://github.com/docker/docs/issues/13146
- https://github.com/docker/docs/issues/13838
- https://github.com/docker/docs/issues/14008
- https://github.com/docker/docs/issues/15583
- https://github.com/docker/docs/issues/15584
- https://github.com/docker/docs/issues/15599
- https://github.com/docker/docs/issues/16464

### cli: additionalHelp() don't decorate output if it's piped

This prevents the escape-characters being included when piping the
output, e.g. `docker --help > output.txt`, or `docker --help | something`.
These control-characters could cause issues if users copy/pasted the URL
from the output, resulting in them becoming part of the URL they tried
to visit, which would fail, e.g. when copying the output from:

    To get more help with docker, check out our guides at https://docs.docker.com/go/guides/

Users ended up on URLs like;

    https://docs.docker.com/go/guides/ESC
    https://docs.docker.com/go/guides/%1B[0m

Before this patch, control characters ("bold") would be printed, even if
no TTY was attached;

    docker --help > output.txt
    cat output.txt | grep 'For more help' | od -c
    0000000 033   [   1   m   F   o   r       m   o   r   e       h   e   l
    0000020   p       o   n       h   o   w       t   o       u   s   e
    0000040   D   o   c   k   e   r   ,       h   e   a   d       t   o
    0000060   h   t   t   p   s   :   /   /   d   o   c   s   .   d   o   c
    0000100   k   e   r   .   c   o   m   /   g   o   /   g   u   i   d   e
    0000120   s   / 033   [   0   m  \n
    0000127

    docker --help | grep 'For more help' | od -c
    0000000 033   [   1   m   F   o   r       m   o   r   e       h   e   l
    0000020   p       o   n       h   o   w       t   o       u   s   e
    0000040   D   o   c   k   e   r   ,       h   e   a   d       t   o
    0000060   h   t   t   p   s   :   /   /   d   o   c   s   .   d   o   c
    0000100   k   e   r   .   c   o   m   /   g   o   /   g   u   i   d   e
    0000120   s   / 033   [   0   m  \n
    0000127

With this patch, no control characters are included:

    docker --help > output.txt
    cat output.txt | grep 'For more help' | od -c
[...]

PR #3996, opened 2023-01-31
Fix issue where one bad credential helper causes no credentials to be returned

laurazard:

fixes https://github.com/docker/cli/issues/3716

**- What I did**

Instead of returning no credentials when a configured credential helper throws an error, skip the bad credential helper, return the remaining credentials, and warn the user about the error.

**- How to verify it**

Refer to the reproduction steps in https://github.com/docker/cli/issues/3716, or run the added unit tests 😄 

image

### DIFF
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -300,7 +300,8 @@ func (configFile *ConfigFile) GetAllCredentials() (map[string]types.AuthConfig, 
 	for registryHostname := range configFile.CredentialHelpers {
 		newAuth, err := configFile.GetAuthConfig(registryHostname)
 		if err != nil {
-			return nil, err
+			logrus.WithError(err).Warnf("Failed to get credentials for registry: %s", registryHostname)
+			continue
 		}
 		auths[registryHostname] = newAuth
 	}
--- a/cli/config/configfile/file_test.go
+++ b/cli/config/configfile/file_test.go
@@ -3,6 +3,7 @@ package configfile
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"os"
 	"testing"
 
@@ -166,8 +167,9 @@ func TestConfigFile(t *testing.T) {
 }
 
 type mockNativeStore struct {
-	GetAllCallCount int
-	authConfigs     map[string]types.AuthConfig
+	GetAllCallCount  int
+	authConfigs      map[string]types.AuthConfig
+	authConfigErrors map[string]error
 }
 
 func (c *mockNativeStore) Erase(registryHostname string) error {
@@ -176,7 +178,7 @@ func (c *mockNativeStore) Erase(registryHostname string) error {
 }
 
 func (c *mockNativeStore) Get(registryHostname string) (types.AuthConfig, error) {
-	return c.authConfigs[registryHostname], nil
+	return c.authConfigs[registryHostname], c.authConfigErrors[registryHostname]
 }
 
 func (c *mockNativeStore) GetAll() (map[string]types.AuthConfig, error) {
@@ -191,8 +193,8 @@ func (c *mockNativeStore) Store(authConfig types.AuthConfig) error {
 // make sure it satisfies the interface
 var _ credentials.Store = (*mockNativeStore)(nil)
 
-func NewMockNativeStore(authConfigs map[string]types.AuthConfig) credentials.Store {
-	return &mockNativeStore{authConfigs: authConfigs}
+func NewMockNativeStore(authConfigs map[string]types.AuthConfig, authConfigErrors map[string]error) credentials.Store {
+	return &mockNativeStore{authConfigs: authConfigs, authConfigErrors: authConfigErrors}
 }
 
 func TestGetAllCredentialsFileStoreOnly(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cli/config/configfile/file_test.go
```diff
@@ -176,7 +178,8 @@ func (c *mockNativeStore) Erase(registryHostname string) error {
 }
 
 func (c *mockNativeStore) Get(registryHostname string) (types.AuthConfig, error) {
-	return c.authConfigs[registryHostname], nil
+	println("hello it is me")
```
thaJeztah: hello

cli/config/configfile/file.go
```diff
@@ -300,7 +300,8 @@ func (configFile *ConfigFile) GetAllCredentials() (map[string]types.AuthConfig,
 	for registryHostname := range configFile.CredentialHelpers {
 		newAuth, err := configFile.GetAuthConfig(registryHostname)
 		if err != nil {
```
thaJeztah, line 302: Do we know what errors this can return? Can we exclude "not found" errors here? (which may be "expected")?

cli/config/configfile/file_test.go
laurazard: hahahahaha. I noticed it and already removed it 😅

cli/config/configfile/file.go
laurazard: We can definitely treat `CredentialsNotFound` error specially, but other errors can occur, and I wonder if we shouldn't skip them regardless.

thaJeztah: Yes, was thinking about the logs mostly (as it would produce some logs that are mostly expected)

laurazard: Sounds good, I'll change that.

cli/config/configfile/file.go
```diff
@@ -300,7 +300,8 @@ func (configFile *ConfigFile) GetAllCredentials() (map[string]types.AuthConfig,
...
-			return nil, err
+			logrus.WithError(err).Warnf("Failed to get credentials for registry: %s", registryHostname)
```
thaJeztah, line 303: Was looking at this again, and while we should look at using the logger more, we should probably look at the output for it. Currently, the output looks like; 

```
WARN[0000] Failed to get credentials for registry: foo   error="something went wrong"
```

Still readable, but perhaps not as "friendly". I see elsewhere we just print our own, e.g.; https://github.com/docker/cli/blob/3ae101f41ef4b1ce7ed774bd0848780dffd07472/cli/config/config.go#L144

Which may look like;

```
WARNING: Error loading config file: /root/.docker/config.json: json: cannot unmarshal string into Go struct field ConfigFile.auths of type types.AuthConfig
```

Only issue is that this location of the code currently doesn't have a reference to the CLI's stdout or stderr, so if we change it, we probably need to hardcode it to `os.Stderr` (not ideal either).

thaJeztah: Ideas / suggestions (perhaps someone else has thoughts) welcome.

Don't think it's necessarily a blocker, if there's no better solution (for now)

laurazard: That's a good point! In general I think having the more standardized logging is good (vs just calling `fmt.Fprintf`), but the friendliness is a good point. Maybe we can look into seeing if we can configuring the output for it. I don't have better ideas right now though 😅

thaJeztah: Yes, there's a lot to look into in this area;

- There's `fmt.PrintXX` in many places
- While it's "generally" ok to use the defaults (I suspect `logrus` is not configured anywhere to use something other than `os.StdOut` / `os.StdErr`), ideally we'd have better controls for that
- ^^ sometimes we need to use a different approach depending on (e.g.) if an actual TTY is attached (see https://github.com/docker/cli/pull/3973)
- ^^ or (e.g.) if we want the output to be "quiet" (could be in the CLI itself, or other projects using parts of the CLI as a "library")
- I also went looking for _some_ of that, e.g. https://github.com/docker/cli/pull/3797
- (as mentioned) logrus (or other loggers) are great for logs, but the default output format often isn't great for printing "one-off", user-facing messages

So yes, this needs work (perhaps I should create a tracking ticket somewhere); e.g.

- Consider using a context-logger to pass a logger around
- That logger could (default) be a logger with a format that's "friendly enough"
- Look into best approach to pass around the streams (STDIN/STDOUT/STDERR), but when doing so, try to prevent coupling things too tightly (i.e., we may not want to pass cobra.Command _everywhere_).

cli/config/configfile/file_test.go
```diff
@@ -237,6 +239,48 @@ func TestGetAllCredentialsCredsStore(t *testing.T) {
 	assert.Check(t, is.Equal(1, testCredsStore.(*mockNativeStore).GetAllCallCount))
 }
 
+func TestGetAllCredentialsCredStoreErrorHandling(t *testing.T) {
+	const (
+		workingHelperRegistryHostname = "working-helper.com"
+		brokenHelperRegistryHostname  = "broken-helper.com"
```
thaJeztah, line 245: Sorry 🤦 jumping between etoomanythings; while these domains probably would never be actually hit, we should avoid using "fake" domains that (potentially) are not "fake". It's better to use domains that are designated for this purpose (see [RFC 2606 (section 2 and 3)](https://www.rfc-editor.org/rfc/rfc2606.html#section-2)). I personally tend to use `example.com` / `example.org` for these (although the `.test` TLD _SHOULD_ be reserved as well, but I've encountered setups where people used it as internal domains).

These could be (e.g.) `working-helper.example.com`,  `broken-helper.example.com`

laurazard: This is a good point! I tend to play it a bit loose with unit tests, but better safe than sorry.

thaJeztah (approved):
LGTM, thanks!

Let's get this one in; we can bike-shed on the output format and refactoring later (most users (hopefully) wouldn't see this message).

PR #3999, opened 2023-02-03
Fix bad ThrottleDevice path

akerouanton:
Fixes moby/moby#44904.

This error was introduced in 6c39bc1 (#3936) and affects only v23.0.


thaJeztah (approved):
LGTM, thanks!

PR #4017, opened 2023-02-09
ci: fix branch filter pattern

crazy-max:
**- What I did**

Similar to https://github.com/moby/moby/pull/44158, found out workflows are not triggered on `23.0` branch. Looking at the docs https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet, the branch filter pattern does not support matching a defined number of consecutive characters.

**- How I did it**

Update branch filter pattern in workflows

**- How to verify it**

CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ on:
   push:
     branches:
       - 'master'
-      - '[0-9]+.[0-9]{2}'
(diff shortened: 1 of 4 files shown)


thaJeztah (approved):
LGTM, thanks!

PR #4022, opened 2023-02-11
Fix styling of arguments

ashlymat:
Signed-off-by: Ashly Mathew <ashlymathew93@gmail.com>

**- What I did**
Small contribution to #923 - changing arguments of options to italics

**- How I did it**

**- How to verify it**

### DIFF
--- a/man/dockerd.8.md
+++ b/man/dockerd.8.md
@@ -151,7 +151,7 @@ $ sudo dockerd --add-runtime runc=runc --add-runtime custom=/usr/local/bin/my-ru
   Set parent cgroup for all containers. Default is "/docker" for fs cgroup
   driver and "system.slice" for systemd cgroup driver.
 
-**--config-file**="/etc/docker/daemon.json"
+**--config-file**=*"/etc/docker/daemon.json"*
   Specifies the JSON file path to load the configuration from.
 
 **--containerd**=""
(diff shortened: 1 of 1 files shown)


ashlymat:
@kolyshkin Could you do a review?

thaJeztah:
Looks like CI got stuck (probably because some changes were made); let me try a close/reopen to trigger it

thaJeztah (approved):
changes LGTM, thanks!

man/dockerd.8.md
```diff
@@ -151,7 +151,7 @@ $ sudo dockerd --add-runtime runc=runc --add-runtime custom=/usr/local/bin/my-ru
   Set parent cgroup for all containers. Default is "/docker" for fs cgroup
   driver and "system.slice" for systemd cgroup driver.
 
-**--config-file**="/etc/docker/daemon.json"
+**--config-file**=*"/etc/docker/daemon.json"*
```
kolyshkin, line 154: 1. This is wrong to put a default value (like done here) or an example value in synopsis. The proper way would be to say something like

```markdown
**--config-file**=*"path"*
```

and add
```markdown
Default is  */etc/docker/daemon.json*.
```
to the option description.

kolyshkin: To explain the last part -- while a file name (such as `/etc/docker/daemon.json`) is a literal, the man page convention is to typeset it as italic.

man/dockerd.8.md
```diff
@@ -165,7 +165,7 @@ $ sudo dockerd --add-runtime runc=runc --add-runtime custom=/usr/local/bin/my-ru
 **-D**, **--debug**=*true*|*false*
   Enable debug mode. Default is false.
 
-**--default-cgroupns-mode**="**host**|**private**"
+**--default-cgroupns-mode**="*host*|*private*"
```
kolyshkin, line 168: Here, both `host` and `private` are literals, i.e. they should be used as-is. Therefore, those are typeset in bold and should not be changed.

OTOH, if that would be something like `default-cgroup-mode=MODE`, the MODE should be in italic, as it's a value to be substituted.

kolyshkin:
Opened https://github.com/docker/cli/pull/4215 to fix the issues from this PR (described above).

PR #4046, opened 2023-02-23
formatter: Consider empty RepoTags and RepoDigests as dangling

vvoland:
- Related to: https://github.com/moby/moby/pull/45068

From API 1.43 version `RepoTags` and `RepoDigests` will be empty for dangling images, instead of being an one-item array with magic constant string (`<none>:<none>`/`<none>@<none>`).

CLI output is not impacted and will still display <none> strings.

**- What I did**
Consider images with empty `RepoTags` and `RepoDigests` as dangling.

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/formatter/image.go
+++ b/cli/command/formatter/image.go
@@ -27,6 +27,9 @@ type ImageContext struct {
 }
 
 func isDangling(image types.ImageSummary) bool {
+	if len(image.RepoTags) == 0 && len(image.RepoDigests) == 0 {
+		return true
+	}
 	return len(image.RepoTags) == 1 && image.RepoTags[0] == "<none>:<none>" && len(image.RepoDigests) == 1 && image.RepoDigests[0] == "<none>@<none>"
 }
 


cli/command/image/client_test.go
```diff
@@ -91,7 +91,7 @@ func (cli *fakeClient) ImageList(ctx context.Context, options types.ImageListOpt
 	if cli.imageListFunc != nil {
 		return cli.imageListFunc(options)
 	}
-	return []types.ImageSummary{{}}, nil
```
vvoland, line 94: So yeah, this fixes the tests, but I'm not sure what was the reason behind this.
With the original changes, a completely empty `ImageSummary` is also considered a dangling image, which causes it to appear in the `docker images` output - but obviously it's not a valid image.

Was there a reason this is was not a plain empty slice from the beginning?
Wondering if the fake `ImageList` call not creating a completely empty slice was a deliberate choice?
Can the server really output empty `ImageSummary` and the listing code should ignore it?

\cc @thaJeztah

thaJeztah: Arf.. was writing up a comment, but didn't put it here. Hold on

thaJeztah: Yes, I agree, this seems like it was not intentional (or at least I could not find a direct reason).

Looks like all of these were added as part of the same commit (https://github.com/docker/cli/commit/e7793092a24a38df7a6f76381eab6ac1c2f5c0b0), that commit was migrated from the moby/moby repository; original one was https://github.com/moby/moby/commit/b2551c619d30f163c3ab8c9ee53cdb09bfbeaa55 (https://github.com/moby/moby/pull/32248)

The only _possible_ reasoning I could think of was if it was some attempt to match that the API returns always returns a slice (even if there's no images); the response of the API (tested on docker 17.06 to be sure) is;

```
curl --unix-socket /var/run/docker.sock 'http://foo/images/json'
[]
```

But that also doesn't match, because it doesn't return (e.g.) `[{}]` (single entry).

Other explanation could be to work with the assertion library used at the time; there's a mention that we switched libraries for that at the time; https://github.com/moby/moby/pull/32248#issuecomment-297159567 so perhaps the old one had some issue?

thaJeztah:
I always get confused on the “ordering” of commits on GitHub;

- so the “test fix” commit was needed to make CI pass _after_ the other commit?
- if that's the case, then the commits should probably be in the reverse order
- but maybe "cleaner" to open a separate PR with _only_ the "test fix" commit to show that it (in itself) does not make CI fail?

GitHub shows the first (?) commit as failing currently;


Screenshot 2023-02-24 at 14 21 53

vvoland:
Yeah I always get confused about the order too 😄 The order is just a consequence of the fact that I first pushed the empty RepoTags/RepoDigests change and then fixed the failing tests as another commit.

I extracted the test change: https://github.com/docker/cli/pull/4050 and will rebase this one.

thaJeztah (approved):
LGTM, thanks!

could you prepare cherry-pick(s) for this one as well?

PR #4050, opened 2023-02-24
test/cli: Use empty array as empty output of images/json

vvoland:
Tests mocking the output of GET images/json with fakeClient used an array with one empty element as an empty response. Change it to just an empty array.

See discussion: https://github.com/docker/cli/pull/4046#discussion_r1116709729


thaJeztah (approved):
LGTM, thanks!

PR #4032, opened 2023-02-14
[20.10 backport] update to go1.19.7

thaJeztah:
Backports of:

- https://github.com/docker/cli/pull/3767
- https://github.com/docker/cli/pull/3799
- https://github.com/docker/cli/pull/3850
- https://github.com/docker/cli/pull/3907
- https://github.com/docker/cli/pull/3958
- https://github.com/docker/cli/pull/4038
- https://github.com/docker/cli/pull/4085

Note that most cherry-picks needed some adjusting to account for the 20.10 branch diverging from master; I used cherry-picks to keep the original commit messages.

PR #3768, opened 2022-09-02
update to golangci-lint v1.49.0 for compatibilty with go1.19

thaJeztah:
Remove the "deadcode", "structcheck", and "varcheck" linters, as they are
deprecated:

    WARN [runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused.
    WARN [runner] The linter 'structcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused.
    WARN [runner] The linter 'varcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused.
    WARN [linters context] structcheck is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649.

And ignore gosec G113, which only affects gp < 1.16.14. and go < 1.17.7

    opts/opts.go:398:13: G113: Potential uncontrolled memory consumption in Rat.SetString (https://github.com/advisories/GHSA-q99m-p7hq-5v4f) (gosec)
        cpu, ok := new(big.Rat).SetString(value)
                   ^

PR #4055, opened 2023-02-27
[20.10 backport] Prepare for Go 1.19 upgrade

corhere:
**- What I did**

Prerequisite work to unblock upgrading to Go 1.19

**- How I did it**

Backported:
- https://github.com/docker/cli/pull/3443
  - skipping https://github.com/docker/cli/pull/3443/commits/8c3ae386cd263c5f4fb4c8f7519febedaf1e3281 as golangci-lint v1.49.0 has a linter which flags uses of any deprecated code without needing any configuration
- https://github.com/docker/cli/pull/3708
- https://github.com/docker/cli/pull/3768

**- How to verify it**

- Rebase https://github.com/docker/cli/pull/4032 on top of this branch
- `make -f docker.Makefile lint test-unit` passes


cli/command/config/inspect.go
```diff
@@ -52,7 +52,7 @@ func RunConfigInspect(dockerCli command.Cli, opts InspectOptions) error {
 	// check if the user is trying to apply a template to the pretty format, which
 	// is not supported
 	if strings.HasPrefix(f, "pretty") && f != "pretty" {
-		return fmt.Errorf("Cannot supply extra formatting options to the pretty template")
+		return errors.New("cannot supply extra formatting options to the pretty template")
```
thaJeztah, line 55: Slightly concerned about this commit, as it changes the output of errors. For sure, people "shouldn't" string match these, but it's _possible_, and may be unexpected in a patch release.

thaJeztah (commented):
overall looks good; some considerations;

- perhaps we should consider skipping the commit that changes the error messages 
- for the io/ioutil changes; I'd also be fine to add a exclusion rule (I think GolangCi-lint has one for this); while it's marked deprecated, (AFAIK), it won't be removed until Go 2, so in reality, it's only to satisfy the linter, so if we prefer to keep a smaller diff, it's worth considering.

cli/command/config/inspect.go
corhere: That's fair; I'll replace that commit with a golangci-lint exclude rule

corhere:
> * for the io/ioutil changes; I'd also be fine to add a exclusion rule (I think GolangCi-lint has one for this); while it's marked deprecated, (AFAIK), it won't be removed until Go 2, so in reality, it's only to satisfy the linter, so if we prefer to keep a smaller diff, it's worth considering.

I figure it'd be nice to have to reduce the delta between branches for backports, hopefully making it slightly easier on @neersighted and myself to maintain this branch over the coming months. But that's a trivial matter and the work to backport those changes is a sunk cost so I have no strong opinion one way or the other. I'll leave it up to you and the other reviewers to decide.

neersighted (approved):

thaJeztah:
Thanks! Yes, I'm fine reducing the diff; the `ioutil` changes are well-understood, and should have no risk in itself. Let me get this one in 👍

PR #4057, opened 2023-02-28
Don't calculate container size unless explicitly asked for

laurazard:
**- What I did**

Previously, we were doing special logic to figure out if the template passed in to `docker ps --format=...` requires the size field, and if so automatically applying the `--size` option to the request.
This causes an issue where, if a user wants a full json representation of the output of `docker ps`, they are unable to opt-out of getting the size. This has big performance implication, and can make `docker ps --format=...` unusable to the point of taking tens of seconds to complete.

This commit removes the special logic for this parameter, to enable users to get a json representation of the container withous `Size`, which can still be returned by setting `--size` alongside with `--format=...`.

I also noticed an inconsistency where previously we would validate the template passed in `--format=[template]` but only inside this edge case logic, which made it so we only validated the template if `--quiet` and `--size` were not passed in. This PR makes it so that if `--format` is used, we always validate the template for consistency.

**- How I did it**

With Neovim 👀 

**- How to verify it**

Run a container, then run:
- `docker ps --format={{json .}}`
- `docker ps --size --format={{json .}}`
- compare the output

```
Container size is no longer calculated by default when using a template which requires it.
```

image

### DIFF
--- a/cli/command/container/list.go
+++ b/cli/command/container/list.go
@@ -2,6 +2,7 @@ package container
 
 import (
 	"context"
+	"io"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
@@ -9,7 +10,9 @@ import (
 	"github.com/docker/cli/cli/command/formatter"
 	flagsHelper "github.com/docker/cli/cli/flags"
 	"github.com/docker/cli/opts"
+	"github.com/docker/cli/templates"
 	"github.com/docker/docker/api/types"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
@@ -73,6 +76,19 @@ func buildContainerListOptions(opts *psOptions) (*types.ContainerListOptions, er
 
 	if opts.nLatest && opts.last == -1 {
 		options.Limit = 1
+	}
+
+	// if `--format` is used, check if template is valid
+	if len(opts.format) > 0 {
+		tmpl, err := templates.NewParse("", opts.format)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to parse template")
+		}
+
+		optionsProcessor := formatter.NewContainerContext()
+		if err := tmpl.Execute(io.Discard, optionsProcessor); err != nil {
+			return nil, errors.Wrap(err, "failed to execute template")
+		}
 	}
 
 	return options, nil
--- a/cli/command/container/list.go
+++ b/cli/command/container/list.go
@@ -2,7 +2,6 @@ package container
 
 import (
 	"context"
-	"io"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
@@ -10,9 +9,7 @@ import (
 	"github.com/docker/cli/cli/command/formatter"
 	flagsHelper "github.com/docker/cli/cli/flags"
 	"github.com/docker/cli/opts"
-	"github.com/docker/cli/templates"
 	"github.com/docker/docker/api/types"
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


thaJeztah:
I realise we may already be doing "too much magic", but wondering if we should special case the "JSON format without explicit fields" (`--format={{json .}}` and `--format=json`) for this

Before this patch, specifying `.Size` as part of a template would make sure that the size is calculated;

```bash
docker ps --format='table {{.Names}}\t{{.Size}}'
NAMES     SIZE
bar       5.33MB (virtual 50.9MB)
foo       25.3MB (virtual 70.8MB)
```

With this patch, the size is omitted, which might be "unexpected";

```bash
./build/docker ps --format='table {{.Names}}\t{{.Size}}'
NAMES     SIZE
bar       0B
foo       0B
```


Hmf.. but looks like we're not doing the same for custom templates defined in the config-file :disappointed:

```bash
 cat ~/.docker/config.json
{
  "psFormat": "table {{.Names}}\t{{.Size}}"
}
```

Which gets silently ignored both before and after this patch :dissapointed:

```bash
docker ps
NAMES     SIZE
bar       0B
foo       0B

./build/docker ps
NAMES     SIZE
bar       0B
foo       0B
```

I guess some of that also relates to the (rather "silly") way we made the `SIZE` _column_ as a whole "conditional"; if the default "table" template is used, we conditionally update the template to also add the size column;

```bash
docker ps
CONTAINER ID   IMAGE          COMMAND                  CREATED          STATUS          PORTS                  NAMES
418827a442b9   nginx:alpine   "/docker-entrypoint.…"   23 minutes ago   Up 23 minutes   80/tcp                 bar
a4f37de806ac   nginx:alpine   "/docker-entrypoint.…"   47 hours ago     Up 31 minutes   0.0.0.0:1234->80/tcp   foo

docker ps --size
CONTAINER ID   IMAGE          COMMAND                  CREATED          STATUS          PORTS                  NAMES     SIZE
418827a442b9   nginx:alpine   "/docker-entrypoint.…"   23 minutes ago   Up 23 minutes   80/tcp                 bar       5.33MB (virtual 50.9MB)
a4f37de806ac   nginx:alpine   "/docker-entrypoint.…"   47 hours ago     Up 31 minutes   0.0.0.0:1234->80/tcp   foo       25.3MB (virtual 70.8MB)
```

But that only happens if it's the default table; specifying a custom table format doesn't get that treatment;


```bash
cat ~/.docker/config.json
{
  "psFormat": "table {{.Names}}"
}
```

```bash
docker ps
NAMES
bar
foo

docker ps --size
NAMES
bar
foo
```

thaJeztah:
Not sure what's best (ideally the above would at least be fixed; that definitely feels like a bug(ish)).


A non-breaking change to consider (and which would allow for a workaround for the linked ticket);

- to check if the `--size` flag is passed
- if it's _explicitly_ passed, and set to `false` (`docker ps --format='{{json .}} --size=false`), then disable the "automatic detection"

That change would not break existing scenarios.


The actual reason we do this is effectively to work around performance issues. Not sure if there's good ways to resolve those (cache results for XX seconds? but "how long" can we cache? do we now need a `--recalculate` option? Can we use something like `fsnotify` on the daemon side to invalidate cache?)

Some options to consider;

- Treat the "blanket" `--format=json` and `--format='{{json .}}'` the same as the default (and `--format=table`, `--format=raw`) templates; for those, the `size` is only included when specified explicitly
- For the `json` and `{{json .}}` option (for which the above will be a "breaking" change); perhaps print a warning (do we need a warning in all scenarios?)
- If the template contains an _explicit_ `.Size`, then also collect the size
- Consider using a more explicit indicator that the size _was not calculated_ (e.g., use `-`, or `-1` to indicate that: currently `0` can either mean "zero" or "not calculated"))
- We could (should?) have separate template functions to capture the _intent_ in the template;
   - `SizeAlways` ?
   - make `Size` accept an argument (`Size true`)?
   - have a `SizeIfPresent` (only show size if `--size` was passed, and omit otherwise)
   - ^^ horrible names (naming is HARD!)

laurazard:
> I realise we may already be doing "too much magic", but wondering if we should special case the "JSON format without explicit fields" (--format={{json .}} and --format=json) for this
> Before this patch, specifying .Size as part of a template would make sure that the size is calculated;

Yeah, maybe I misunderstood but I thought that's what we discussed 😅. Regardless, another way to work around this as you mentioned is seeing if `--size=false` was explicitly set, and otherwise keep the special logic. I'll work that into this PR, at least that allows for a workaround. I still feel like it's bad that `--format` is a specific case where the user needs to go out of their way to exclude size, when the logic works the other way around in the other cases (user needs to explicitly ask for size). This will make things a bit more confusing.

> If the template contains an explicit .Size, then also collect the size

I'm not sure if there's any good way to do this other than checking if the template string contains `.Size` which we can do but feels a bit weird and adding even more special logic. I was hoping we could do away with some of that instead of increasing it.

PR #4067, opened 2023-03-02
Don't automatically request size if `--size` was explicitly set to `false`

laurazard:
**- What I did**

Check if the `--size` flag was explicitly set to false (with `--size=false`) and don't automatically request the size when that's the case.

**- How I did it**

With Neovim 👀 

**- How to verify it**

Run `docker ps --format={{json .}} --size=false` and check that size isn't calculated

image


thaJeztah:
Oh! Was looking if we could update one of the unit tests to test this, but forgot to post; looks like we have some test for testing the auto-detect; https://github.com/docker/cli/blob/cb5463a72869c72349622a04b49ec58ccf8bdf1f/cli/command/container/list_test.go#L233-L243


I guess we could make that a test-table to test the various options; I gave that a quick go;

```go
func TestContainerListFormatSizeSetsOption(t *testing.T) {
	tests := []struct {
		doc, format, sizeFlag string
		sizeExpected          bool
	}{
		{
			doc:          "detect with all fields",
			format:       `{{json .}}`,
			sizeExpected: true,
		},
		{
			doc:          "detect with explicit field",
			format:       `{{.Size}}`,
			sizeExpected: true,
		},
		{
			doc:          "detect no size",
			format:       `{{.Names}}`,
			sizeExpected: false,
		},
		{
			doc:          "override enable",
			format:       `{{.Names}}`,
			sizeFlag:     "true",
			sizeExpected: true,
		},
		{
			doc:          "override disable",
			format:       `{{.Size}}`,
			sizeFlag:     "false",
			sizeExpected: false,
		},
	}

	for _, tc := range tests {
		tc := tc
		t.Run(tc.doc, func(t *testing.T) {
				containerListFunc: func(options types.ContainerListOptions) ([]types.Container, error) {
					assert.Check(t, is.Equal(options.Size, tc.sizeExpected))
					return []types.Container{}, nil
				},
			})
			cmd := newListCommand(cli)
			cmd.Flags().Set("format", tc.format)
			if tc.sizeFlag != "" {
				cmd.Flags().Set("size", tc.sizeFlag)
			}
			assert.NilError(t, cmd.Execute())
		})
	}
}
```

laurazard:
That's great! I thought about adding a little test case and ended up forgetting about it, but this is more comprehensive. I'll add it in.

thaJeztah:
Thanks! Probably fine to squash those commits to keep the changes together

Issue #4078, opened 2023-03-08
docker stack deploy Fails When Run From cmd.exe on Windows with CLI v23+


twinter-amosfivesix:
### Description

Trying to deploy a stack (`docker stack deploy`) when using version 23+ of the Docker CLI under a process started by Windows `cmd.exe` causes an error message similar to:
```
unexpected environment "=::=::\\"
```

This also affects `docker stack config`.

### My System
The failure happens in Docker CLI 23.0.0 and 23.0.1. I'm using Windows 10 Enterprise 21H2 19044.2604.

### The Key Part
`cmd.exe` has to be involved in starting the `docker` process in some manner. So you could run `cmd.exe` and then `docker`. Or `cmd.exe` and then `pwsh.exe` and then `docker`.

### The Relevant Code

[This appears to be the relevant code](https://github.com/docker/cli/blob/master/cli/command/stack/loader/loader.go#L100-L115). Environment variables are read and iterated through. If one is found with a blank name/key, the error message is printed and docker aborts.

### Why Does This Happen
[The Old New Thing](https://devblogs.microsoft.com/oldnewthing/20100506-00/?p=14133) has the best explanation of this. `cmd.exe` basically keeps special environment variables with blank names to support old MS-DOS functionality. These are the environment variables that Docker stumbles over.

[Stack Overflow question about this](https://stackoverflow.com/questions/10431689/what-are-these-strange-environment-variables). Here you'll see talk of the `=::=::\\` value that I commonly see.

### Reproduce

1. Be using the Docker CLI version 23.0.0 or 23.0.1 on Windows.
2. Have a simple stack file named `Stack.yml`. See example below.
3. Start a Windows command prompt (`cmd.exe`) - This is key!
4. Run this: `docker stack deploy -c Stack.yml Stack`
5. The output is:
```
unexpected environment "=::=::\\"
```

#### Example Stack File:
```
services:
  hello:
    image: hello-world:latest
```

### Expected behavior

The stack should deploy without error.

For comparison purposes, you can follow these steps to see the stack deploy correctly:
1. Be using the Docker CLI version 23.0.0 or 23.0.1 on Windows.
2. Have a simple stack file named `Stack.yml`. See example in the **Reproduce** section.
3. Start a PowerShell 7 session **without** going through `cmd.exe`. The reliable way I've found to do this is search for `pwsh.exe` from the Start menu. Using icons in the Start menu doesn't seem to work all the time.
[...]

PR #4081, opened 2023-03-09
stack/loader: Ignore cmd.exe special env variables

vvoland:
- Fixes: https://github.com/docker/cli/issues/4078

On Windows, ignore all variables that start with "=" when building an environment variables map for stack.

For MS-DOS compatibility cmd.exe can set some special environment variables that start with a "=" characters, which breaks the general assumption that the first encountered "=" separates a variable name from variable value and causes trouble when parsing.

These variables don't seem to be documented anywhere, but they are described by some third-party sources and confirmed empirically on my Windows installation.

Useful sources:
https://devblogs.microsoft.com/oldnewthing/20100506-00/?p=14133
https://ss64.com/nt/syntax-variables.html

Known variables:

- `=ExitCode` stores the exit code returned by external command (in hex format)
- `=ExitCodeAscii` - same as above, except the value is the ASCII representation of the code (so exit code 65 (0x41) becomes 'A').
- `=::=::\` and friends - store drive specific working directory. There is one env variable for each separate drive letter that was accessed in the shell session and stores the working directory for that specific drive. The general format for these is: `=<DRIVE_LETTER>:=<CWD>`  (key=`=<DRIVE_LETTER>:`, value=`<CWD>`) where <CWD> is a working directory for the drive that is assigned to the letter <DRIVE_LETTER>

  A couple of examples: `=C:=C:\some\dir`  (key: `=C:`, value: `C:\some\dir`) `=D:=D:\some\other\dir`  (key: `=C:`, value: `C:\some\dir`) `=Z:=Z:\`  (key: `=Z:`, value: `Z:\`)

  `=::=::\` is the one that seems to be always set and I'm not exactly sure what this one is for (what's drive `::`?). Others are set as soon as you CD to a path on some drive. Considering that you start a cmd.exe also has some working directory, there are 2 of these on start.

All these variables can be safely ignored because they can't be deliberately set by the user, their meaning is only relevant to the cmd.exe session and they're all are related to the MS-DOS/Batch feature that are irrelevant for us.

**- What I did**

**- How I did it**

**- How to verify it**
TestBuildEnvironment test

Fix `docker stack deploy` failing when running from `cmd.exe` on Windows

### DIFF
--- a/cli/command/stack/loader/loader.go
+++ b/cli/command/stack/loader/loader.go
@@ -120,7 +120,7 @@ func buildEnvironment(env []string) (map[string]string, error) {
 
 		k, v, ok := strings.Cut(s, "=")
 		if !ok || k == "" {
-			return result, errors.Errorf("unexpected environment %q", s)
+			return result, errors.Errorf("unexpected environment variable '%s'", s)
 		}
 		// value may be set, but empty if "s" is like "K=", not "K".
 		result[k] = v
--- a/cli/command/stack/loader/loader.go
+++ b/cli/command/stack/loader/loader.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 
@@ -104,6 +105,19 @@ func GetConfigDetails(composefiles []string, stdin io.Reader) (composetypes.Conf
 func buildEnvironment(env []string) (map[string]string, error) {
 	result := make(map[string]string, len(env))
 	for _, s := range env {
+		if runtime.GOOS == "windows" && len(s) > 0 {
+			// cmd.exe can have special environment variables which names start with "=".
+			// They are only there for MS-DOS compatibility and we should ignore them.
+			// See TestBuildEnvironment for examples.
+			//
+			// https://ss64.com/nt/syntax-variables.html
+			// https://devblogs.microsoft.com/oldnewthing/20100506-00/?p=14133
+			// https://github.com/docker/cli/issues/4078
+			if s[0] == '=' {
+				continue
+			}
+		}
+
 		k, v, ok := strings.Cut(s, "=")
 		if !ok || k == "" {
 			return result, errors.Errorf("unexpected environment %q", s)
--- a/cli/command/stack/loader/loader_test.go
+++ b/cli/command/stack/loader/loader_test.go
@@ -3,6 +3,7 @@ package loader
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
@@ -44,4 +45,35 @@ services:
 	assert.Assert(t, is.Len(details.ConfigFiles, 1))
 	assert.Check(t, is.Equal("3.0", details.ConfigFiles[0].Config["version"]))
 	assert.Check(t, is.Len(details.Environment, len(os.Environ())))
+}
+
+func TestBuildEnvironment(t *testing.T) {
+	inputEnv := []string{
+		"LEGIT_VAR=LEGIT_VALUE",
+		"EMPTY_VARIABLE=",
+	}
+
+	if runtime.GOOS == "windows" {
+		inputEnv = []string{
+			"LEGIT_VAR=LEGIT_VALUE",
+
+			// cmd.exe has some special environment variables which start with "=".
+			// These should be ignored as they're only there for MS-DOS compatibility.
+			"=ExitCode=00000041",
+			"=ExitCodeAscii=A",
+			`=C:=C:\some\dir`,
+			`=D:=D:\some\different\dir`,
+			`=X:=X:\`,
+			`=::=::\`,
+
+			"EMPTY_VARIABLE=",
+		}
+	}
+
+	env, err := buildEnvironment(inputEnv)
+	assert.NilError(t, err)
+
+	assert.Check(t, is.Len(env, 2))
+	assert.Check(t, is.Equal("LEGIT_VALUE", env["LEGIT_VAR"]))
+	assert.Check(t, is.Equal("", env["EMPTY_VARIABLE"]))
 }


cli/command/stack/loader/loader.go
```diff
@@ -104,6 +105,19 @@ func GetConfigDetails(composefiles []string, stdin io.Reader) (composetypes.Conf
...
 	for _, s := range env {
+		if runtime.GOOS == "windows" && len(s) > 0 {
+			// cmd.exe can have special environment variables which names start with "=".
+			// They are only there for MS-DOS compatibility and we should ignore them.
+			// See TestBuildEnvironment for examples.
+			//
+			// https://ss64.com/nt/syntax-variables.html
+			// https://devblogs.microsoft.com/oldnewthing/20100506-00/?p=14133
+			// https://github.com/docker/cli/issues/4078
+			if s[0] == '=' {
```
vvoland, line 116: I didn't make this one an explicit blacklist of variables, because:
1. MS documentation explicitly states that:
> The name of an environment variable cannot include an equal sign (=).
https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables

so user can't deliberately set such variable nor anyone other than `cmd.exe` should actually set such variables.

2. I don't think we want to implement the parser for the drive-specific CWD format (or hardcode every possible drive letter).
3. We don't want to play whack-a-mole if there are some other undocumented variables that can start with =.

vvoland:
```
#17 [lint 4/4] RUN --mount=type=bind,target=.     --mount=type=cache,target=/root/.cache         golangci-lint run
#17 75.52 cli/command/stack/loader/loader_test.go:65:20: `ther` is a misspelling of `there` (misspell)
#17 75.52 			"=D:=D:\\some\\other\\dir",
#17 75.52 			                ^
#17 ERROR: executor failed running [/bin/sh -c golangci-lint run]: exit code: 1

#14 [internal] load build context
------
#17 75.52 cli/command/stack/loader/loader_test.go:65:20: `ther` is a misspelling of `there` (misspell)
#17 75.52 			"=D:=D:\\some\\other\\dir",
#17 75.52 			                ^
------
```

Hmm, even the linter barks at this blasphemous environment variable 😄 .

Not sure why, though:
```go
fmt.Println("=D:=D:\\some\\other\\dir")
```
is: ```=D:=D:\some\other\dir```

Will try with raw literals.

cli/command/stack/loader/loader.go
```diff
@@ -104,6 +105,19 @@ func GetConfigDetails(composefiles []string, stdin io.Reader) (composetypes.Conf
...
+				continue
+			}
+		}
+
 		k, v, ok := strings.Cut(s, "=")
 		if !ok || k == "" {
 			return result, errors.Errorf("unexpected environment %q", s)
```
twinter-amosfivesix, line 123: I would love to see this message changed to "unexpected environment **variable** %q". See [my comment on the bug report](https://github.com/docker/cli/issues/4078#issuecomment-1462258810).

vvoland:
Changing `=D:=D:\some\other\dir` to `=D:=D:\some\different\dir` does the trick! 😄

cli/command/stack/loader/loader.go
thaJeztah: Possibly we could use `'%s'` as well; the `%q` is nice as it escapes output, but in some cases that makes the output harder to read (`error something "\"\\\\foo\\\\bar\"` etc), so having it print "as-is" could make that more readable.

cli/command/stack/loader/loader.go
thaJeztah: > 1. MS documentation explicitly states that:
> 
> > The name of an environment variable cannot include an equal sign (=).
> > https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables

Interesting, thanks!

So;

- Perhaps we should include that link (and description that a user cannot set it) in the comment
- From the above, I wonder indeed if this is something that `os.Environ()` should take into account (if that would make sense) and possibly filter out those "env-vars" (might be worth opening a ticket for in Golang, and see if there's opinions on that)

cli/command/stack/loader/loader.go
```diff
@@ -104,6 +105,19 @@ func GetConfigDetails(composefiles []string, stdin io.Reader) (composetypes.Conf
...
+			// https://ss64.com/nt/syntax-variables.html
+			// https://devblogs.microsoft.com/oldnewthing/20100506-00/?p=14133
+			// https://github.com/docker/cli/issues/4078
+			if s[0] == '=' {
+				continue
+			}
+		}
+
 		k, v, ok := strings.Cut(s, "=")
 		if !ok || k == "" {
```
thaJeztah, line 122: I was wondering if the case above could also be checked as;

```go
if ok && k == "" && runtime.GOOS == "windows" {
    // windows handling here
```

But not sure if that would make it actually more readable than your implementation, and won't give us a performance benefice, so .. bad suggestion I guess 😂

cli/command/stack/loader/loader.go
vvoland: Done. Not entirely sure about using %s instead of %q though (because if the var starts with `=` which isn't explicitly not allowed, can we really assume it's safe to print without escaping?)

thaJeztah: I think we should be fine? Or were you thinking of these containing control characters (`\033[1;42m`) that would mess up the terminal?

thaJeztah (approved):
code LGTM

left one comment earlier about possibly referring to the MS Docs (not a show-stopper, but let me know if you think it makes sense to include)  https://github.com/docker/cli/pull/4081#discussion_r1131251913

cli/command/stack/loader/loader.go
twinter-amosfivesix: There look to be existing Golang issues related to this:
* https://github.com/golang/go/issues/49886
* https://github.com/golang/go/issues/49882

neersighted (approved):

Issue #3864, opened 2022-11-16
Add CDI support to --devices 


klueska:
### Description

**Feature Request:**
Add support to the `--device` flag to process fully-qualified CDI device names in addition to standard device node paths.

Support for this was added in `podman` [`v4.1.0`](https://github.com/containers/podman/blob/main/RELEASE_NOTES.md#410) in May 2022 and it would be good to have feature parity with this in `docker`.

**Background:**
The [Container Device Interface (CDI)](https://github.com/container-orchestrated-devices/container-device-interface) is a CNCF sponsored initiative to standardize the way in which complex devices are exposed to containers. It is based on the Container Networking Interface (CNI) model and specification.

With CDI, a "device" is more than just a single device node under `/dev`.

Instead, CDI defines a _**device**_ as a high-level concept mapped to specific set of OCI runtime spec modifications. These modifications include not just the inclusion of device nodes, but also filesystem mounts, environment variables, and container lifecycle hooks.

Vendors are responsible for writing CDI specs that define their _**devices**_  in terms of these modifications, and CDI-enabled container runtimes are responsible for reading these specs and making the proper modifications when requested to do to.

The list of container runtimes that **already** include support for CDI are:
* [podman](https://github.com/containers/podman/releases/tag/v3.2.0)
* [cri-o](https://github.com/cri-o/cri-o/releases/tag/v1.24.0)
* [containerd](https://github.com/containerd/containerd/releases/tag/v1.7.0-beta.0)

Moreover, a new feature in Kubernetes called [Dynamic Resource Allocation (DRA)](https://github.com/kubernetes/kubernetes/pull/111023) uses CDI under the hood to do its device injection (with the assumption that a CDI-enabled runtime such as `cri-o` or `containerd` is there to support it).

As a concrete example, consider the following abbreviated CDI spec for injecting an NVIDIA GPU into a container:
```
---
cdiVersion: 0.4.0
kind: nvidia.com/gpu
devices:
- name: gpu0
  containerEdits:
   deviceNodes:
   - path: /dev/nvidia0
   deviceNodes:
   - path: /dev/nvidiactl
   mounts:
   - containerPath: /usr/lib/x86_64-linux-gnu/libnvidia-ml.so.460.91.03
     hostPath: /usr/lib/x86_64-linux-gnu/libnvidia-ml.so.460.91.03
     options: [ro, nosuid, nodev, bind]
   - containerPath: /usr/lib/x86_64-linux-gnu/libcuda.so.460.91.03
[...]

PR #4084, opened 2023-03-10
Support CDI devices in --device flag

elezar:
**- What I did**

Added support for specifying CDI devices in the `--device` flag. See https://github.com/docker/cli/issues/3864

Note that support for specifying CDI devices in the `--device` flag has been merged into Podman, Singularity (OCI mode), and Containerd's `ctr` command.

All of these rely on the fact that fully-qualified CDI devices names provide a unique mechanism to identify these devices names which does not conflict with devices nodes on Linux systems or [devices identifiers](https://github.com/opencontainers/runtime-spec/blob/main/config-windows.md#devices) on Windows systems.

**- How I did it**

Before performing the platform-specific validation and parsing of the `--device` arguments, the specified device is checked to see whether it is a fully-qualified device name of the form: `vendor.com/class=name`. If this is the case, it is considered a CDI device.

A device request (similar to the `--gpus` flag) is constructed from the specified CDI device names and added to the `HostConfig` to be passed to the API. This requires https://github.com/moby/moby/pull/45134 to be regognized and processed properly by the daemon.

**- How to verify it**

Prerequisites:
1. The Daemon includes the changes from https://github.com/moby/moby/pull/45134
2. The Daemon is configured in experimental mode
3. The CLI includes the changes from this PR

Create a CDI specification for testing:
```
cat > vendor.yaml << EOF
cdiVersion: 0.3.0
kind: vendor1.com/device
devices:
- name: test
  containerEdits:
    env:
    - FOO="injected by cdi"
EOF
```

Copy the generated spec to `/var/run/cdi`:
```
sudo mkdir -p /var/run/cdi
sudo cp vendor.yaml /var/run/cdi
```

Run the docker CLI requesting the device:
```
docker run --rm -ti --device vendor1.com/device=test  busybox  env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=e9c8e63f6ce2
TERM=xterm
FOO="injected by cdi"
HOME=/root
```

Running with an undefined cdi device name shows:
```
docker run --rm -ti --device vendor1.com/device=invalid  busybox  env
docker: Error response from daemon: CDI device injection failed: unresolvable CDI devices vendor1.com/device=invalid.
```

If the daemon does not support CDI (e.g. is in experimental mode), the following will be shown:
```
docker run --rm -ti --device vendor1.com/device=test  busybox  env
docker: Error response from daemon: could not select device driver "cdi" with capabilities: [[cdi]].
```

Add support for requesting CDI devices using the `--device` flag will show an error indicating a missing `cdi` driver.

### DIFF
--- a/cli/command/container/opts_test.go
+++ b/cli/command/container/opts_test.go
@@ -417,39 +417,91 @@ func TestParseWithExpose(t *testing.T) {
 
 func TestParseDevice(t *testing.T) {
 	skip.If(t, runtime.GOOS != "linux") // Windows and macOS validate server-side
-	valids := map[string]container.DeviceMapping{
-		"/dev/snd": {
-			PathOnHost:        "/dev/snd",
-			PathInContainer:   "/dev/snd",
-			CgroupPermissions: "rwm",
+	testCases := []struct {
+		devices        []string
+		deviceMapping  *container.DeviceMapping
+		deviceRequests []container.DeviceRequest
+	}{
+		{
+			devices: []string{"/dev/snd"},
+			deviceMapping: &container.DeviceMapping{
+				PathOnHost:        "/dev/snd",
+				PathInContainer:   "/dev/snd",
+				CgroupPermissions: "rwm",
+			},
 		},
-		"/dev/snd:rw": {
-			PathOnHost:        "/dev/snd",
-			PathInContainer:   "/dev/snd",
-			CgroupPermissions: "rw",
+		{
+			devices: []string{"/dev/snd:rw"},
+			deviceMapping: &container.DeviceMapping{
+				PathOnHost:        "/dev/snd",
+				PathInContainer:   "/dev/snd",
+				CgroupPermissions: "rw",
+			},
 		},
-		"/dev/snd:/something": {
-			PathOnHost:        "/dev/snd",
-			PathInContainer:   "/something",
-			CgroupPermissions: "rwm",
+		{
+			devices: []string{"/dev/snd:/something"},
+			deviceMapping: &container.DeviceMapping{
+				PathOnHost:        "/dev/snd",
+				PathInContainer:   "/something",
+				CgroupPermissions: "rwm",
+			},
 		},
-		"/dev/snd:/something:rw": {
-			PathOnHost:        "/dev/snd",
-			PathInContainer:   "/something",
-			CgroupPermissions: "rw",
+		{
+			devices: []string{"/dev/snd:/something:rw"},
+			deviceMapping: &container.DeviceMapping{
+				PathOnHost:        "/dev/snd",
+				PathInContainer:   "/something",
+				CgroupPermissions: "rw",
+			},
+		},
+		{
+			devices:       []string{"vendor.com/class=name"},
+			deviceMapping: nil,
+			deviceRequests: []container.DeviceRequest{
+				{
+					Driver:    "cdi",
+					DeviceIDs: []string{"vendor.com/class=name"},
+				},
+			},
+		},
+		{
+			devices: []string{"vendor.com/class=name", "/dev/snd:/something:rw"},
+			deviceMapping: &container.DeviceMapping{
+				PathOnHost:        "/dev/snd",
+				PathInContainer:   "/something",
+				CgroupPermissions: "rw",
+			},
+			deviceRequests: []container.DeviceRequest{
+				{
+					Driver:    "cdi",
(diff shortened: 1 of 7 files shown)


corhere (commented):
Pulling in nearly 40,000 lines of vendored code just to use `cdi.IsQualifiedName` is unlikely to be acceptable. Go is fairly good at only vendoring the parts of the package graph which are reachable from the main module, so splitting the name parsing logic from `github.com/container-orchestrated-devices/container-device-interface/pkg/cdi` into a separate package (within the same module) might be sufficient to cut out the unnecessary vendoring.

elezar:
> Pulling in nearly 40,000 lines of vendored code just to use `cdi.IsQualifiedName` is unlikely to be acceptable. Go is fairly good at only vendoring the parts of the package graph which are reachable from the main module, so splitting the name parsing logic from `github.com/container-orchestrated-devices/container-device-interface/pkg/cdi` into a separate package (within the same module) might be sufficient to cut out the unnecessary vendoring.

Yes, there have been discussions in the COD working group regarding better handling CDI dependencies. I will chat to @klihub as well to see whether we can get something done for this.

elezar:
@corhere thanks for the suggestion with regards to reworking the package upstream. I have reorganised it and the dependencies that are now pulled in are more limited.

cli/command/container/opts_test.go
```diff
@@ -417,39 +417,90 @@ func TestParseWithExpose(t *testing.T) {
...
+			args = append(args, "img", "cmd")
+
+			_, hostconfig, _, err := parseRun(args)
+
+			assert.NilError(t, err)
+
+			if tc.deviceMapping != nil {
+				assert.Equal(t, len(hostconfig.Devices), 1, "Expected 1 devices, got %v", hostconfig.Devices)
+				assert.Check(t, is.DeepEqual(*tc.deviceMapping, hostconfig.Devices[0]))
+			}
```
corhere, line 500:
```suggestion
			if tc.deviceMapping != nil {
				if assert.Check(t, is.Len(hostconfig.Devices, 1)) {
					assert.Check(t, is.DeepEqual(*tc.deviceMapping, hostconfig.Devices[0]))
				}
			} else {
				assert.Check(t, is.Len(hostconfig.Devices, 0))
			}
```

corhere (approved):

cli/command/container/opts_test.go
elezar: Thanks. Updated. I'm used to using `requires` in other projects.

klueska:
The PR to add CDI support to `moby` has been merged: https://github.com/moby/moby/pull/45134
@corhere is there something still needed to move this PR along to ensure that it makes the train for Docker 25?

cpuguy83 (approved):
I'm a little apprehensive to shove this into the existing `--device` flag due to past issues with the`-v` flag....
But if it works it works and I don't want to hold it up since even technically `-v` works with all the things we've overloaded it with... so LGTM.

thaJeztah:
We'll have a look at this; want to have a quick check with some other people on the UX 👍

elezar:
Thanks @thaJeztah and others for the review. If there are UX changes required, please let us know. Note that we currently have the same support in Podman, Singularity, and CTR and it would be good to have the same UX across all CLI clients so that we don't end up in the situation we have now with the `--gpu` flag. If we do choose some other flag or representation, these other clients would have to be updated to also support the new flag. 

With these changes the `--device` flag handles the following cases:
* Linux device nodes following the pattern: `/abs/path/on/host[:/abs/path/in/container][:permissions]`
* Windows devices following the pattern: `class/{{UUID}}`
* CDI devices following the pattern: `{{VENDOR_ID}}/{{DEVICE_CLASS}}={{DEVICE_ID}}` (where none of the elements can be blank or contain slashes).

From my perspective, the risk for ambiguity across these three cases is low and is outweighed by the simplicity of the UX in reasoning about devices. That is to say, users are used to using the `--device` flag to request access to devices.

elezar:
> We'll have a look at this; want to have a quick check with some other people on the UX 👍

@thaJeztah has there been any feedback on possible UX issues?

elezar:
@corhere @cpuguy83 @neersighted @thaJeztah 

I've updated the vendoring again. PTAL.

thaJeztah (approved):
LGTM, sorry for the delay!

PR #4089, opened 2023-03-12
Add bash completion for docker-buildx plugin

crazy-max:
**- What I did**

I was working on targets completion for the `buildx bake` command in https://github.com/docker/buildx/pull/1674 and found out we don't have any bash completion for docker-buildx plugin.

**- How I did it**

**- How to verify it**

```
$ mkdir -p ~/.local/share/bash-completion/completions
$ cp ./contrib/completion/bash/docker ~/.local/share/bash-completion/completions/docker
$ _completion_loader docker

$ docker build [tab][tab]
build    builder  buildx

$ docker buildx [tab][tab]
bake        create      help        inspect     prune       stop        version     
build       du          imagetools  ls          rm          use

$ docker buildx b [tab][tab]
bake   build
```

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -5395,6 +5395,23 @@ _docker_wait() {
 	_docker_container_wait
 }
 
+BUILDX_PLUGIN_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "buildx"}}{{.Path}}{{end}}{{end}}')
+
+_docker_buildx() {
+	local completionCommand="__completeNoDesc"
+	local resultArray=($BUILDX_PLUGIN_PATH $completionCommand)
+	for value in "${words[@]:2}"; do
+		if [ -z "$value" ]; then
+			resultArray+=( "''" )
+		else
+			resultArray+=( "$value" )
+		fi
+	done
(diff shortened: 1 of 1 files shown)


thaJeztah:
Guess it won't hurt to have it in this script, although we now also have the Cobra-generated completion in the CLI on master (and 23.0+) since https://github.com/docker/cli/pull/3429

crazy-max:
> Guess it won't hurt to have it in this script, although we now also have the Cobra-generated completion in the CLI on master (and 23.0+) since #3429

Yeah but it doesn't work for plugins afaik.

contrib/completion/bash/docker
```diff
@@ -5395,6 +5395,23 @@ _docker_wait() {
 	_docker_container_wait
 }
 
+BUILDX_PLUGIN_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "buildx"}}{{.Path}}{{end}}{{end}}')
```
silvin-lubecki, line 5398: Maybe for a follow-up, but I think we should try to make it generic for plugins. We should call `docker info` only once and process the result, depending the plugin name.
It might be tricky though, so maybe for a follow-up 🤔

crazy-max: Thanks for your feedback. I will make it more generic to support all plugins. Keep you posted.

PR #4094, opened 2023-03-13
Add bash completion for available plugins

crazy-max:
closes https://github.com/docker/cli/pull/4089

**- What I did**

This is a follow-up of https://github.com/docker/cli/pull/4089#discussion_r1133670411 to enable bash completion for available plugins.

**- How I did it**

**- How to verify it**

```
$ mkdir -p ~/.local/share/bash-completion/completions
$ cp ./contrib/completion/bash/docker ~/.local/share/bash-completion/completions/docker
$ _completion_loader docker

$ docker info --format '{{json .ClientInfo.Plugins}}' | jq
[
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v0.10.0-rc2-175-g00bb9c5c.m",
    "ShortDescription": "Docker Buildx",
    "Name": "buildx",
    "Path": "/home/crazy/.docker/cli-plugins/docker-buildx",
    "ShadowedPaths": [
      "/usr/libexec/docker/cli-plugins/docker-buildx"
    ]
  },
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v2.16.0",
    "ShortDescription": "Docker Compose",
    "Name": "compose",
    "Path": "/usr/libexec/docker/cli-plugins/docker-compose"
  },
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v0.23.0",
    "ShortDescription": "Docker Scan",
    "Name": "scan",
    "Path": "/usr/libexec/docker/cli-plugins/docker-scan"
  }
]

$ docker build [tab][tab]
build    builder  buildx

$ docker buildx [tab][tab]
bake        create      help        inspect     prune       stop        version     
build       du          imagetools  ls          rm          use

$ docker buildx b [tab][tab]
bake   build

$ docker co [tab][tab]
commit     compose    config     container  context

$ docker compose [tab][tab]
build    cp       down     exec     kill     ls       port     pull     restart  run      stop     unpause  version  
config   create   events   images   logs     pause    ps       push     rm       start    top      up
```

To make it works with buildx plugin you need https://github.com/docker/buildx/pull/1674

cc @silvin-lubecki @ulyssessouza

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -1142,6 +1142,29 @@ __docker_complete_user_group() {
 	fi
 }
 
+DOCKER_PLUGINS_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{.Path}}:{{end}}')
+
+__docker_complete_plugin() {
+	local path=$1
+	local completionCommand="__completeNoDesc"
+	local resultArray=($path $completionCommand)
+	for value in "${words[@]:2}"; do
+		if [ -z "$value" ]; then
+			resultArray+=( "''" )
+		else
+			resultArray+=( "$value" )
+		fi
+	done
+	local result=$(eval "${resultArray[*]}" 2> /dev/null | grep -v '^:[0-9]*$')
+
+	# if result empty, just use filename completion as fallback
+	if [ -z "$result" ]; then
+		_filedir
+	else
+		COMPREPLY=( $(compgen -W "${result}" -- "${current-}") )
+	fi
+}
+
 _docker_docker() {
 	# global options that may appear after the docker command
 	local boolean_options="
@@ -5395,23 +5418,6 @@ _docker_wait() {
 	_docker_container_wait
 }
 
-COMPOSE_PLUGIN_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "compose"}}{{.Path}}{{end}}{{end}}')
-
-_docker_compose() {
-	local completionCommand="__completeNoDesc"
-	local resultArray=($COMPOSE_PLUGIN_PATH $completionCommand compose)
-	for value in "${words[@]:2}"; do
-		if [ -z "$value" ]; then
-			resultArray+=( "''" )
-		else
-			resultArray+=( "$value" )
-		fi
-	done
-	local result=$(eval "${resultArray[*]}" 2> /dev/null | grep -v '^:[0-9]*$')
-
-	COMPREPLY=( $(compgen -W "${result}" -- "${current-}") )
-}
-
 _docker() {
 	local previous_extglob_setting=$(shopt -p extglob)
 	shopt -s extglob
@@ -5481,11 +5487,16 @@ _docker() {
 		wait
 	)
 
+	# Create completion functions for all registered plugins
 	local known_plugin_commands=()
-
-	if [ -f "$COMPOSE_PLUGIN_PATH" ] ; then
-		known_plugin_commands+=("compose")
-	fi
+	local plugin_name=""
+	for plugin_path in ${DOCKER_PLUGINS_PATH//:/ }; do
+		plugin_name=$(basename "$plugin_path" | sed 's/ *$//')
+		plugin_name=${plugin_name#docker-}
+		plugin_name=${plugin_name%%.*}
+		eval "_docker_${plugin_name}() { __docker_complete_plugin \"${plugin_path}\"; }"
+		known_plugin_commands+=(${plugin_name})
+	done
 
 	local experimental_server_commands=(
 		checkpoint


silvin-lubecki (approved):
Not a shell script expert here, but the logic looks good 👍

contrib/completion/bash/docker
```diff
@@ -1142,6 +1142,29 @@ __docker_complete_user_group() {
 	fi
 }
 
+PLUGINS_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{.Path}},{{end}}')
```
albers, line 1145: Two things here:

- this defines a global variable (accessable in the current shell). Maybe its name should be more specific.
- `,` as a path list delimiter looks a bit unexpected to me, would you mind using a `:` for that purpose?
```suggestion
DOCKER_PLUGINS_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{.Path}}:{{end}}')
```

(Corresponding change in https://github.com/docker/cli/blob/5c3a594bcbc414c86e47d8fc32d2729d6ee564d9/contrib/completion/bash/docker#L5492 required)

contrib/completion/bash/docker
```diff
@@ -5482,10 +5488,14 @@ _docker() {
 	)
 
 	local known_plugin_commands=()
-
-	if [ -f "$COMPOSE_PLUGIN_PATH" ] ; then
-		known_plugin_commands+=("compose")
-	fi
+	local plugin_name=""
+	for plugin_path in ${PLUGINS_PATH//,/ }; do
```
albers, line 5492: see above.
```suggestion
	for plugin_path in ${DOCKER_PLUGINS_PATH//:/ }; do
```

contrib/completion/bash/docker
```diff
@@ -5482,10 +5488,14 @@ _docker() {
...
+		plugin_name=$(basename "$plugin_path" | sed 's/ *$//g')
```
albers, line 5493: The global flag in the sed expression does not make sense for a pattern that is nailed to end of line.
```suggestion
		plugin_name=$(basename "$plugin_path" | sed 's/ *$//')
```

contrib/completion/bash/docker
```diff
@@ -5482,10 +5488,14 @@ _docker() {
 	)
 
 	local known_plugin_commands=()
-
-	if [ -f "$COMPOSE_PLUGIN_PATH" ] ; then
-		known_plugin_commands+=("compose")
-	fi
+	local plugin_name=""
```
albers, line 5491: IMHO this code deserves a descriptive comment, e.g. "Create completion functions for all registered plugins"

albers (changes requested):
This is really, really cool.

I have to admit that I do not fully understand `__docker_complete_plugin()` but I assume this is basically an idiom for cobra generated completions and should be correct.

I did not find any issues with the completions.

Thanks @crazy-max.

crazy-max:
@albers Thanks for the review. I made the suggested changed.

I think in follow-up we could use the same logic but for all docker commands so we would need a single function to handle completion like the `__docker_complete_plugin()` one does.

I also have a branch to handle this for `fish` and `zsh`.

e.g., for fish:

```
# plugins
set PLUGINS_PATH (docker info --format '{{range .ClientInfo.Plugins}}{{.Path}},{{end}}')

function __fish_docker_plugin_complete --description 'Print plugins commands and subcomands' -a plugin_path
    set -l resultArray $plugin_path "__completeNoDesc"
    for i in (commandline -opc)
        if set -q i
            set -a resultArray "''"
        else
            set -a resultArray $i
        end
    end
    set -l result (eval printf '%s ' $resultArray 2> /dev/null | grep -v '^:[0-9]*$')
    if set -q i
        return 1
    end
    echo $result
    return 0
end

for plugin_path in (string split ',' $PLUGINS_PATH)
    set plugin_name (string trim $plugin_path)
    set plugin_name (string trim -l 'docker-' $plugin_name)
    set plugin_name (echo $plugin_name | sed 's/\.[^.]*$//')
    complete -c docker -f -n '__fish_docker_no_subcommand' -a $plugin_name
    complete -c docker -A -f -n "__fish_seen_subcommand_from $plugin_name" -a "(__fish_docker_plugin_complete $plugin_path)"
end
```

Not tested and needs more work I think. I'm not an expert of this shell :sweat_smile:.

albers (approved):
Thanks @crazy-max. Very impressive.

crazy-max:
PTAL @thaJeztah @vvoland :pray:

vvoland (approved):
Just tested and it seems to work perfectly, thanks!

ulyssessouza:
LGTM!

ulyssessouza (approved):
LGTM! 🥳

PR #4109, opened 2023-03-22
Config ls examples corrections in reference documentation

vice:
I just edited directly in the GitHub interface hence I do not have signed off commits. I think those are minor corrections but I can redo them with the Signed off added.


thaJeztah:
included in https://github.com/docker/cli/pull/6980

PR #4010, opened 2023-02-07
vendor: update docker/docker to master (v24.0.0-dev)

thaJeztah:
see individual commits for details

PR #4111, opened 2023-03-22
Remove registry Service use

thaJeztah:
- relates to https://github.com/docker/cli/pull/4010
- relates to https://github.com/moby/moby/pull/45086 (https://github.com/moby/moby/commit/7b3acdff5d01c6bbac5ddd38d3bc01277f06ee64 removed the interface, so we can no longer use that).

Looks like the service is not needed, as we always pass an empty config, so we can use ParseRepositoryInfo instead.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/plugin/install.go
+++ b/cli/command/plugin/install.go
@@ -54,24 +54,12 @@ func newInstallCommand(dockerCli command.Cli) *cobra.Command {
 	return cmd
 }
 
-type pluginRegistryService struct {
-	registry.Service
-}
-
-func (s pluginRegistryService) ResolveRepository(name reference.Named) (*registry.RepositoryInfo, error) {
-	repoInfo, err := s.Service.ResolveRepository(name)
+func resolvePlugnRepository(name reference.Named) (*registry.RepositoryInfo, error) {
+	repoInfo, err := registry.ParseRepositoryInfo(name)
 	if repoInfo != nil {
 		repoInfo.Class = "plugin"
 	}
 	return repoInfo, err
-}
-
-func newRegistryService() (registry.Service, error) {
-	svc, err := registry.NewService(registry.ServiceOptions{})
-	if err != nil {
-		return nil, err
-	}
-	return pluginRegistryService{Service: svc}, nil
 }
 
 func buildPullConfig(ctx context.Context, dockerCli command.Cli, opts pluginOptions, cmdName string) (types.PluginInstallOptions, error) {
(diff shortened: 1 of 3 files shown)


cli/command/plugin/install.go
```diff
@@ -54,26 +54,14 @@ func newInstallCommand(dockerCli command.Cli) *cobra.Command {
...
-	registry.Service
-}
-
-func (s pluginRegistryService) ResolveRepository(name reference.Named) (*registry.RepositoryInfo, error) {
-	repoInfo, err := s.Service.ResolveRepository(name)
+func resolvePlugnRepository(name reference.Named) (*registry.RepositoryInfo, error) {
+	repoInfo, err := registry.ParseRepositoryInfo(name)
 	if repoInfo != nil {
 		repoInfo.Class = "plugin"
 	}
```
thaJeztah, line 61: We need to check if this is still needed; ISTR this was some snafu on Docker Hub, and may not even be needed anymore. /cc @tianon (ISTR you were on that discussion, but I don't recall the status 🤔)

cli/command/plugin/install.go
```diff
@@ -54,26 +54,14 @@ func newInstallCommand(dockerCli command.Cli) *cobra.Command {
...
 }
 
-type pluginRegistryService struct {
-	registry.Service
-}
-
-func (s pluginRegistryService) ResolveRepository(name reference.Named) (*registry.RepositoryInfo, error) {
-	repoInfo, err := s.Service.ResolveRepository(name)
+func resolvePlugnRepository(name reference.Named) (*registry.RepositoryInfo, error) {
+	repoInfo, err := registry.ParseRepositoryInfo(name)
```
thaJeztah, line 58: For reference; this is what `ParseRepositoryInfo` does; https://github.com/docker/cli/blob/14482589df194a86b2ee07df643ba3277b40df7d/vendor/github.com/docker/docker/registry/config.go#L413-L432

It uses an `emptyServiceConfig`, which is needed to mark `localhost` as "insecure", and `docker.io` as "official"
https://github.com/docker/cli/blob/14482589df194a86b2ee07df643ba3277b40df7d/vendor/github.com/docker/docker/registry/config.go#L57

cli/trust/trust.go
```diff
@@ -299,9 +299,12 @@ type ImageRefAndAuth struct {
 	digest     digest.Digest
 }
 
+// RepositoryInfoResolver returns repository info for the given reference.
+type RepositoryInfoResolver func(name reference.Named) (*registry.RepositoryInfo, error)
```
thaJeztah, line 303: I was a bit on the fence wether to define this as a type, or to just inline the signature in the function below (like we do for `authResolver`).

cli/command/plugin/install.go
tianon: That's correct!  Docker Hub still allows the auth class, but it works fine either way (ie, it is no longer _required_ like it used to be).

thaJeztah: If that's the case, perhaps we can remove all of this altogether 🤔 🎉 

I tried to be a bit more careful as I'm not even sure this code-path is tested 😬 (plugins are a bit of a corner-case, so it wouldn't surprise me if there isn't a e2e test for this).

thaJeztah: Gave that a quick go in https://github.com/docker/cli/pull/4114 (still in draft; if that variant would work, I can update this PR 😅)

thaJeztah:
Closing this in favour of https://github.com/docker/cli/pull/4114

PR #4113, opened 2023-03-22
Allow extensions in `build` field in `docker-compose.yaml` schema

zkulbeda:
Enable support of [x-bake extension](https://docs.docker.com/build/bake/compose-file/#extension-field-with-x-bake) in json schema.

**- What I did**
Add `patternProperties` in latest json schema for `build` field.

Add support for `x-` fields in `build` field in docker-compose.yaml

```
               ((`\
            ___ \\ '--._
         .'`   `'    o  )
        /    \   '. __.'
       _|    /_  \ \_\_
jgs   {_\______\-'\__\_\
```

### DIFF
--- a/cli/compose/schema/data/config_schema_v3.9.json
+++ b/cli/compose/schema/data/config_schema_v3.9.json
@@ -91,7 +91,8 @@
                 "shm_size": {"type": ["integer", "string"]},
                 "extra_hosts": {"$ref": "#/definitions/list_or_dict"}
               },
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thanks! We discussed this change, and decided that instead, we'll allow all additional properties, because stack does not support build, and already ignores it, so we might as well ignore all properties underneath it.

See https://github.com/docker/cli/pull/4214

PR #4114, opened 2023-03-22
cli/trust: remove special handling for "plugin" Class

thaJeztah:
- relates to https://github.com/docker/cli/pull/4010
- relates to https://github.com/moby/moby/pull/45086 (https://github.com/moby/moby/commit/7b3acdff5d01c6bbac5ddd38d3bc01277f06ee64 removed the interface, so we can no longer use that).
- relates to https://github.com/docker/hub-feedback/issues/2107
- replaces / closes https://github.com/docker/cli/pull/4111

This code depended on the registry Service interface, which has been removed,
so needed to be refactored. Digging further into the reason this code existed,
it looked like the Class=plugin was previously required on Docker Hub to handle
plugins, but this requirement is no longer there, so we can remove this special
handling.

This patch removes the special handling to both remove the use of the registry.Service
interface, as well as removing complexity that is no longer needed.

### DIFF
--- a/cli/trust/trust.go
+++ b/cli/trust/trust.go
@@ -160,7 +160,7 @@ func GetNotaryRepository(in io.Reader, out io.Writer, userAgent string, repoInfo
 	scope := auth.RepositoryScope{
 		Repository: repoInfo.Name.Name(),
 		Actions:    actions,
-		Class:      repoInfo.Class,
+		Class:      repoInfo.Class, // TODO(thaJeztah): Class is no longer needed for plugins and can likely be removed; see https://github.com/docker/cli/pull/4114#discussion_r1145430825
 	}
 	creds := simpleCredentialStore{auth: *authConfig}
 	tokenHandlerOptions := auth.TokenHandlerOptions{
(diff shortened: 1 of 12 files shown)


cli/trust/trust.go
```diff
@@ -160,7 +160,7 @@ func GetNotaryRepository(in io.Reader, out io.Writer, userAgent string, repoInfo
 	scope := auth.RepositoryScope{
 		Repository: repoInfo.Name.Name(),
 		Actions:    actions,
-		Class:      repoInfo.Class,
+		Class:      repoInfo.Class, // TODO(thaJeztah): if Class is no longer needed for plugins, does that mean we can remove it as a whole?
```
thaJeztah, line 163: ... and now I'm wondering if this `Class` is needed at all anymore (of there's only "image" or "").

I do see some handling in "distribution"; https://github.com/distribution/distribution/blob/e5d5810851d1f17a5070e9b6f940d8af98ea3c29/registry/client/auth/session.go#L154-L164

```go

// String returns the string representation of the repository
// using the scope grammar
func (rs RepositoryScope) String() string {
	repoType := "repository"
	// Keep existing format for image class to maintain backwards compatibility
	// with authorization servers which do not support the expanded grammar.
	if rs.Class != "" && rs.Class != "image" {
		repoType = fmt.Sprintf("%s(%s)", repoType, rs.Class)
	}
	return fmt.Sprintf("%s:%s:%s", repoType, rs.Repository, strings.Join(rs.Actions, ","))
}
```

And that looks to be "only if it's _NOT_ "" or "image"

cpuguy83 (approved):

PR #4117, opened 2023-03-24
Dockerfile: align go version

crazy-max:
**- What I did**

Looks like we forgot to update the Go version in `Dockerfile.vendor`.

Also use [null values](https://docs.docker.com/build/bake/file-definition/#null-values) for variables in the bake definition.

**- How I did it**

**- How to verify it**

cc @thaJeztah


thaJeztah (approved):
LGTM, thanks!

PR #4118, opened 2023-03-24
Dockerfile: build binary if no target specified

crazy-max:
**- What I did**

Binary should be built if no target is specified when building the Dockerfile.

**- How I did it**

**- How to verify it**


vvoland (approved):

laurazard (approved):

thaJeztah (approved):
LGTM

PR #4122, opened 2023-03-27
don't use null values in the bake definition

crazy-max:
- follow-up https://github.com/docker/docker-ce-packaging/pull/866#issuecomment-1484745221
- updates https://github.com/docker/cli/pull/4117

**- What I did**

Remove null values in the bake definition to avoid issues with environments with old Buildx version (< 0.10).

**- How I did it**

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #4123, opened 2023-03-27
e2e: update notary certificates

crazy-max:
follow-up https://github.com/docker/cli/pull/4122#issuecomment-1484908622

**- What I did**

Update notary certificates for e2e tests and increase tests certificates duration to 10 years (3 years previously).

Also moves script to generate certificates to a dedicated Dockerfile to be sandboxed:

```
$ docker buildx bake e2e-gencerts
```

**- How I did it**

**- How to verify it**


thaJeztah (approved):
oh! fancy 😄 

LGTM (if CI is happy)

thaJeztah:
If we need this for 20.10, we probably need to modify it a bit when backporting, but that's for a later concern (worst case, we just copy over the certs to that branch)

crazy-max:
I can just have an extra commit with just the certs to ease backport if yo uwant?

thaJeztah:
Whoop! All green; thank you!

PR #4130, opened 2023-03-28
ci: enhanced build workflow

crazy-max:
**- What I did**

Split binaries job so now each platform is built on its own runner.

Before ~15m

image

Now ~3m

image

It spawns a significant amount of jobs but at least reduces build time.

@thaJeztah We also need to update the required GitHub Checks in the repo settings. I think we should just remove required for build imo.

**- How I did it**

**- How to verify it**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,14 +15,35 @@ on:
   pull_request:
 
 jobs:
+  prepare:
+    runs-on: ubuntu-20.04
+    outputs:
+      matrix: ${{ steps.platforms.outputs.matrix }}
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Create matrix
+        id: platforms
+        run: |
+          echo "matrix=$(docker buildx bake cross --print | jq -cr '.target."cross".platforms')" >>${GITHUB_OUTPUT}
+      -
+        name: Show matrix
+        run: |
+          echo ${{ steps.platforms.outputs.matrix }}
+
   build:
     runs-on: ubuntu-20.04
+    needs:
+      - prepare
     strategy:
       fail-fast: false
       matrix:
         target:
-          - cross
-          - dynbinary-cross
+          - binary
+          - dynbinary
+        platform: ${{ fromJson(needs.prepare.outputs.matrix) }}
         use_glibc:
           - ""
           - glibc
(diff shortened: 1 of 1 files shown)


crazy-max:
> * does this spin up a new agent for every `platform` + `arch` combination?

Yes and also glibc comibnation, 3 dim matrix :sweat_smile:

> * if so; would that mean we _could_ (potentially) run into "queues"?

I think it's fine as it takes max 3m to build one target.

> * ^^ was wondering what it would look like (how bad / long the time would be if we split "per arch" (i.e., combine `linux/arm/v7` + `linux/arm/v7 + glibc` + `binary` + `dynbinary`), assuming those can share (some of) the build-cache

Maybe we could combine glibc yeah, let me know what you prefer.

thaJeztah:
(FWIW, happy to take this as-is, just thought, let's do a quick check ❤️)

laurazard:
No strong opinion one way or the other on the splits – I guess there's the speed vs. Github UI balance there, but if that's the only difference I wouldn't go out of my way to adapt to the actions UI (who knows when that'll change) 😅

laurazard (approved):

vvoland (approved):
I'm fine either way. However if it would take just a simple 1-2 lines change to make them combined "per-arch" AND it wouldn't be much slower then I think I would have a slight preference towards having less checks 😅

vvoland:
BTW, are these expected to be stuck?

thaJeztah:
> However if it would take just a simple 1-2 lines change to make them combined "per-arch" AND it wouldn't be much slower then I think I would have a slight preference towards having less checks 😅

@crazy-max how much work is that? (if it's more than just 1-2 lines changes, then we can probably merge as-is)

crazy-max:
> > However if it would take just a simple 1-2 lines change to make them combined "per-arch" AND it wouldn't be much slower then I think I would have a slight preference towards having less checks 😅
> 
> @crazy-max how much work is that? (if it's more than just 1-2 lines changes, then we can probably merge as-is)

I'm afraid it's more than that, I can do that in follow-up if that's fine with you

thaJeztah:
No worries; looks like it's "merge as-is" then 👍

thaJeztah:
Removed the (now obsolete) tasks from branch protection; I'll add the new ones after this is merged

PR #4136, opened 2023-03-29
Fix plugin completion results parsing for `ShellCompDirectiveFilterFileExt`

laurazard:
Compose V2 (and other plugins eventually) provide the extensions they want the shell to filter the autocompletion of files for:
```console
$ /usr/libexec/docker/cli-plugins/docker-compose __completeNoDesc -f ""
yaml
yml
:8
Completion ended with directive: ShellCompDirectiveFilterFileExt
```

The current script does not process these results correctly, and instead we just provide a `y` (common beginning between `yaml` and `yml`)

```console
$ docker compose -f <TAB><TAB>
yaml
yml
```

which is not the desired outcome.

**- What I did**

Check if the completion flag is `:8` (`ShellCompDirectiveFilterFileExt`) and format a valid glob pattern to feed into `_filedir` when this is the case

**- How to verify it**

Load the new completion script, and type
```
docker compose -f <TAB><TAB>
```
verify that completions are provided for files with the extension `.yaml` or `.yml`,

### DIFF
--- a/contrib/completion/bash/docker
+++ b/contrib/completion/bash/docker
@@ -1155,7 +1155,20 @@ __docker_complete_plugin() {
 			resultArray+=( "$value" )
 		fi
 	done
-	local result=$(eval "${resultArray[*]}" 2> /dev/null | grep -v '^:[0-9]*$')
+	local rawResult=$(eval "${resultArray[*]}" 2> /dev/null)
+	local result=$(grep -v '^:[0-9]*$' <<< "$rawResult")
+
+	# Compose V2 completions sometimes returns returns `:8` (ShellCompDirectiveFilterFileExt)
+	# with the expected file extensions (such as `yml`, `yaml`) to indicate that the shell should
+	# provide autocompletions for files with matching extensions
+	local completionFlag=$(tail -1 <<< "$rawResult")
+	if [ "$completionFlag" == ":8" ]; then
+		# format a valid glob pattern for the provided file extensions
+		local filePattern=$(tr '\n' '|' <<< "$result")
+
+		_filedir "$filePattern"
+		return
+	fi
 
 	# if result empty, just use filename completion as fallback
 	if [ -z "$result" ]; then


thaJeztah:
@tianon @albers ptal 🤗

contrib/completion/bash/docker
```diff
@@ -1155,7 +1155,20 @@ __docker_complete_plugin() {
...
 		fi
 	done
-	local result=$(eval "${resultArray[*]}" 2> /dev/null | grep -v '^:[0-9]*$')
+	local rawResult=$(eval "${resultArray[*]}" 2> /dev/null)
+	local result=$(grep -v '^:[0-9]*$' <<< "$rawResult")
+
+	# Compose V2 completions sometimes returns returns `:8` (ShellCompDirectiveFilterFileExt)
+	# with the expected file extensions (such as `yml`, `yaml`) to indicate that the shell should
+	# provide autocompletions for files with matching extensions
+	local completionFlag=$(echo "${rawResult[@]}" | tail -1)
```
tianon, line 1164: Since `rawResult` is just a string, this can be a little simpler (and avoid some of `echo`'s potentially pathologic behavior):
```suggestion
	local completionFlag=$(tail -1 <<< "$rawResult")
```

(Is it _always_ the last entry?  I guess we'd never actually have a file extension like `:0-9`)

contrib/completion/bash/docker
```diff
@@ -1155,7 +1155,20 @@ __docker_complete_plugin() {
 			resultArray+=( "$value" )
 		fi
 	done
-	local result=$(eval "${resultArray[*]}" 2> /dev/null | grep -v '^:[0-9]*$')
+	local rawResult=$(eval "${resultArray[*]}" 2> /dev/null)
+	local result=$(grep -v '^:[0-9]*$' <<< "$rawResult")
+
+	# Compose V2 completions sometimes returns returns `:8` (ShellCompDirectiveFilterFileExt)
```
tianon, line 1161: If anyone else reviewing this was unfamiliar with what this means, it's a Cobra thing, and https://github.com/spf13/cobra/blob/4dd4b25de38418174a6e859e8a32eaccca32dccc/completions_test.go#L1046-L1047 is the best reference I could find for it being `:8` (there's probably a better one, but that's what my search brought me to :sweat_smile:).

tianon (approved):

thaJeztah:
I'll skip merging this one for now, to have a look if we want to add  Tianon's suggestions (as it's marked for cherry-picking, it makes it a bit cleaner to do the backport "in one go" instead of a PR-and-follow-up).

contrib/completion/bash/docker
laurazard: Thanks! I'd thought about adding some link explaining it better and then totally forgot :|

albers (approved):
After finally realizing that the `Completion ended with directive: ShellCompDirectiveFilterFileExt` line actually goes to stderr: LGTM, thanks @laurazard

thaJeztah:
Just checking up on this; Is the suggestion from @tianon (echo -> tail) something to consider, or "too minor to bother")? https://github.com/docker/cli/pull/4136#discussion_r1154685069

contrib/completion/bash/docker
laurazard: @thaJeztah I was just going to take 5 mins to try the changes @tianon suggested before amending, but since I'm out I'm okay if you want to do that/commit the changes and merge yourself

thaJeztah: No worries, I'll have a look! Let's wait for that before merging (assuming @tianon's changes look good to everyone) /cc @albers

vvoland (approved):

thaJeztah (approved):
LGTM, thank you!

PR #4138, opened 2023-03-30
gha: align stray go 1.19.4 version

thaJeztah:
looks like this one was forgotten to be updated :)

**- What I did**

**- How I did it**

**- How to verify it**


vvoland (approved):

rumpl (approved):
So many places this is defined... :)

thaJeztah:
> So many places this is defined... :)

Yeah, we need to look at this; it was easier in the past where we had 1 or two `GO_VERSION` build-args (which is easier to grep for)

PR #4156, opened 2023-04-03
cli/command/container: implement `docker run --annotation`

AkihiroSuda:
**- What I did**
Implemented `docker run --annotation`.

For moby/moby#45025 (Docker v24, API v1.43).

`docker run --annotation foo=bar` is similar to `podman run --annotation foo=bar`, however, unlike Podman, Docker implementation also accepts an annotation with an empty value. (`docker run --annotation foo`)

**- How I did it**

Added the `--annotation` flag as an `*opts.MapOpts`.

**- How to verify it**

Run `docker run --annotation foo=bar`, and run `docker inspect` to confirm that the annotation is set.

cli/command/container: implement `docker run --annotation`

🐧


neersighted (approved):

AkihiroSuda:
> Thanks! This should _probably_ be fine, but before we merge, I want to do a quick check look; ISTR we discussed some of this, and whether or not to add this "as-is", have a "porcelain" option (for runtime annotations), etc. (But it's been some time, so I need to read up / read back on some of that).

This feature has been already implemented in `podman run --annotation`, so my opinion is to just follow their CLI design so that their users can easily adopt `docker run --annotation` too

albers (approved):
Completions LGTM, thanks.

AkihiroSuda:
@thaJeztah Can we merge this?

thaJeztah:
We discussed this in the maintainers meeting, and generally are ok with the current implementation. There were some concerns;

- Is "annotation" confusing (vs "label")? We could probably borrow some of the wording from https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/ to explain the differences
- Is it clear enough this is passed through to the runtime? We discussed describing it as "stored on the container, and passed through to the OCI runtime" 
- We discussed the `=` being optional; we don't expect these to (like `--env`) ever take values from the environment, and that to be ambiguous, so we're good with that (easier to use)
- We _do_ want @Oyabun to have a quick look before merging.

Thanks!

cli/command/container/opts.go
```diff
@@ -297,6 +299,10 @@ func addFlags(flags *pflag.FlagSet) *containerOptions {
 
 	flags.BoolVar(&copts.init, "init", false, "Run an init inside the container that forwards signals and reaps processes")
 	flags.SetAnnotation("init", "version", []string{"1.25"})
+
+	flags.Var(copts.annotations, "annotation", "Add annotations to container")
```
tianon, line 303: Something like this, maybe?

```suggestion
	flags.Var(copts.annotations, "annotation", "Add annotations to the container (passed through to the OCI runtime)")
```

(for the relevant item in https://github.com/docker/cli/pull/4156#issuecomment-1507468378)

thaJeztah: ^^ LGTM (forgot to write that down, but we discussed that as description for the flag) (thanks!)

contrib/completion/fish/docker.fish
```diff
@@ -178,6 +178,7 @@ complete -c docker -A -f -n '__fish_seen_subcommand_from cp' -l help -d 'Print u
 # create
 complete -c docker -f -n '__fish_docker_no_subcommand' -a create -d 'Create a new container'
 complete -c docker -A -f -n '__fish_seen_subcommand_from create' -l add-host -d 'Add a custom host-to-IP mapping (host:ip)'
+complete -c docker -A -f -n '__fish_seen_subcommand_from create' -l annotation -d 'Add annotations to container'
```
thaJeztah, line 181: (don't forget to update these as well when updating the flag description) ❤️

cli/command/container/opts.go
AkihiroSuda: done

contrib/completion/fish/docker.fish
AkihiroSuda: done

tianon (approved):

thaJeztah (approved):
LGTM

I haven't seen strong voices about alternative naming for the flag, so I'll bring this one in. Thanks!

PR #4165, opened 2023-04-06
Add unit information about `docker builder prune --filter until=XXX`

lyrixx:
ref: 
* https://github.com/moby/moby/blob/master/builder/builder-next/builder.go#L611-L636
* https://pkg.go.dev/time#ParseDuration


lyrixx:
What should I do to finish this PR?

PR #4166, opened 2023-04-06
Add networking reference link

ChrisChinchilla:
- addresses https://github.com/docker/docs/issues/16890

**- What I did**

Added a link to `--networking` flag as requested in this issue https://github.com/docker/docs/issues/16890

Made it an "external" link for better usage in multiple locations.

### DIFF
--- a/docs/reference/commandline/build.md
+++ b/docs/reference/commandline/build.md
@@ -29,7 +29,7 @@ Build an image from a Dockerfile
 | `--label`                           | `list`        |           | Set metadata for an image                                         |
 | `-m`, `--memory`                    | `bytes`       | `0`       | Memory limit                                                      |
 | `--memory-swap`                     | `bytes`       | `0`       | Swap limit equal to memory plus swap: -1 to enable unlimited swap |
-| `--network`                         | `string`      | `default` | Set the networking mode for the RUN instructions during build     |
+| [`--network`](#network)             | `string`      | `default` | Set the networking mode for the RUN instructions during build     |
 | `--no-cache`                        |               |           | Do not use cache when building the image                          |
 | `--platform`                        | `string`      |           | Set platform if server is multi-platform capable                  |
 | `--pull`                            |               |           | Always attempt to pull a newer version of the image               |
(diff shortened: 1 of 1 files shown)


akerouanton (approved):
LGTM

docs/reference/commandline/build.md
```diff
@@ -29,7 +29,7 @@ Build an image from a Dockerfile
 | `--label`                           | `list`        |           | Set metadata for an image                                         |
 | `-m`, `--memory`                    | `bytes`       | `0`       | Memory limit                                                      |
 | `--memory-swap`                     | `bytes`       | `0`       | Swap limit equal to memory plus swap: -1 to enable unlimited swap |
-| `--network`                         | `string`      | `default` | Set the networking mode for the RUN instructions during build     |
+| `--network`                         | `string`      | `default` | Set [the networking mode](https://docs.docker.com/network/) for the RUN instructions during build     |
```
thaJeztah, line 32: This table is actually generated from the source code (I thought we had a verify step for this one to check it, but looks like it doesn't flag this 🤔). The table in _markdown_ is only used on GitHub (and generated from the YAML properties on docs.docker.com).

What I think we need here instead is to either add a section for this  option further down (having the right anchor tag to match the flag name), and add the link for details in there, or perhaps we need to add a link to the `buildx build` documentation as alternative link (if that page _does_ have a description).

thaJeztah (changes requested):
left a comment (I can catch up with you if needed)

thaJeztah:
Looks like `buildx build` also doesn't have a flag description for this, so we probably need to add one on this page (or both, if there's differences between BuildKit's support for `--network` and non-buildkit (see https://github.com/moby/moby/issues/40379))


@crazy-max do you know from the top of your head if BuildKit now has full feature-parity on `--network` ?

docs/reference/commandline/build.md
ChrisChinchilla: Hmm, there don't actually appear to be many sections for more explanation in this doc. Is there any reason we can't add links to code comments? Or we'd just rather not?

ChrisChinchilla:
@thaJeztah and @crazy-max Turns out my last change was linking people to the wrong place anyway and digging more into this it seems that `--network` is basically the only non-experimental flag with the build/buildx disparity. So I think this latest change is correct now

thaJeztah:
Rebased, and squashed the commits

thaJeztah (approved):
LGTM if CI is happy

PR #2936, opened 2021-01-18
Add --format=json to inspect and list commands

silvin-lubecki:
**- What I did**
* Added a specific format key "json" to the inspector with the same behavior as empty string (outputing json by default). I also set "json" as a default value to all the format flags in inspect commands.
* Added a specific format key "json" to the formatter, that is just an alias to the Golang template `{{json .}}` which outputs the inspected object as pure json.

⚠️ TODO: missing updating the format flag inspect commands help message to add the `"json"` case. 

**- How to verify it**

- For ls commands:
```sh
$ docker context ls --format=json
{"Current":true,"Description":"Current DOCKER_HOST based configuration","DockerEndpoint":"unix:///var/run/docker.sock","KubernetesEndpoint":"https://kubernetes.docker.internal:6443 (default)","Name":"default","StackOrchestrator":"swarm"}
```
- For inspect commands:
```sh
$  docker inspect 14c505d03da8 --format=json
[
    {
        "Id": "14c505d03da83e4d0c33b4cdbd2068429ebac253d2de05530d48918a6bed9782",
        "Created": "2020-11-06T09:44:52.387624Z",
        "Path": "/entrypoint.sh",
        "Args": [
            "/etc/docker/registry/config.yml"
        ],
...
```

**- Description for the changelog**
* Add --format=json to inspect and list commands

**- A picture of a cute animal (not mandatory but encouraged)**
image

PR #4168, opened 2023-04-09
fix docker info, docker version --format=json not outputting json format

thaJeztah:
- relates to https://github.com/docker/cli/pull/2936

### fix docker info --format=json not outputting json format

The --format=json option was added for all inspect commands, but was not implemented for "docker info". This patch implements the missing option.

Before this patch:

    docker info --format=json
    json

With this patch applied:

    docker info --format=json
    {"ID":"80c2f18a-2c88-4e4a-ba69-dca0eea59835","Containers":7,"ContainersRunning":"..."}

### fix docker version --format=json not outputting json format

The `--format=json` option was added for all inspect commands, but was not
implemented for "docker version". This patch implements the missing option.

Before this patch:

    docker version --format=json
    json

With this patch:

    docker version --format=json
    {"Client":{"Platform":{"Name":""},"Version":"24.0.0-dev","ApiVersion":"..."}}


thaJeztah:
@silvin-lubecki @vvoland @laurazard ptal

cli/command/system/version.go
rumpl, line 1: Sneaky little commit that's unrelated :)

cli/command/system/info.go
```diff
@@ -507,6 +506,10 @@ func printServerWarningsLegacy(dockerCli command.Cli, info types.Info) {
 }
 
 func formatInfo(dockerCli command.Cli, info info, format string) error {
+	if format == formatter.JSONFormatKey {
```
rumpl, line 509: This is unfortunate, changing the parameter on fly is not great

cli/command/system/info.go
rumpl, line 1: Don't we want the system info formatting to go to the formatting package?

cli/command/system/info.go
thaJeztah: Yeah, the `system info` printing is quite involved currently. I can use a new variable for it if you prefer but thought to keep it simple (it's not passed by reference in either case, so only used here).

cli/command/system/info.go
thaJeztah: Not sure, we'd have to look. I think there's been some back-and-forth in the past; a single "formatting" package would also have to import all the types used by all objects, so I think it originally was a single package, then got split to each object type. to allow for specific logic for each where needed.

Definitely something to look at what the current state  is, and if things are all in the right place.

cli/command/system/version.go
thaJeztah: 😂 so I thought you meant that JSON file, but it was referring to the "make it a const" commit. I had that in another PR, but doing it later would mean we had more code-churn, as the lines using it are touched by this PR, so I moved it first to prevent that.

rumpl (approved):

PR #4187, opened 2023-04-11
ci: add validation for generated docs

### LINKED ISSUE #4171: CI: validate generated Markdown as part of CI
### Description

Some of the markdown is generated from source code, and while we have make targets for this, we currently do not validate if the Markdown pages match the generated content.

Relates to

- https://github.com/docker/cli/pull/3924#issue-1502002409
- https://github.com/docker/cli/pull/4166#discussion_r1159660161
- https://github.com/docker/cli/pull/4169#issuecomment-1501159399

dvdksn:
Signed-off-by: David Karlsson <david.karlsson@docker.com>

**- What I did**

Added a step for validating the generated markdown/yaml files, based on the generation output vs what's checked in to the repo

closes #4171

**- How I did it**

Added a job to the `validate` GHA workflow

**- How to verify it**

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -29,8 +29,6 @@ jobs:
       -
         name: Checkout
         uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
       -
         name: Run
         uses: docker/bake-action@v2
@@ -70,8 +68,6 @@ jobs:
       -
         name: Checkout
         uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
       -
         name: Run
         shell: 'script --return --quiet --command "bash {0}"'


dvdksn:
Well that failed spectacularly, love it

.github/workflows/validate.yml
```diff
@@ -56,3 +56,11 @@ jobs:
         shell: 'script --return --quiet --command "bash {0}"'
         run: |
           make -f docker.Makefile ${{ matrix.target }}
+      -
+        name: Validate generated # check that checked-in and generated files match
```
thaJeztah, line 60: I guess we need to skip the yaml docs, because we don't commit those into source control currently (we do commit the markdown files though)

thaJeztah:
Awesome! Let me also call in @crazy-max to check 🤗

.github/workflows/validate.yml
```diff
@@ -37,6 +37,30 @@ jobs:
...
 
+  # check that the generated Markdown and the checked-in files match
+  validate-md:
+    runs-on: ubuntu-20.04
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
crazy-max, line 48: nit: don't think this is needed (will speed up fetch time)

thaJeztah: LOL, bit confused by your comment; Will _removing_ it improve fetch time or _keeping_ it will? (I assumed "0" normally would safe time, but perhaps with caching / different checks doing without this, things may be different?)

crazy-max: Default is `1`: https://github.com/actions/checkout/blob/8f4b7f84864484a7bf31766abe9204da3cbe65b3/action.yml#L56-L58

> 0 indicates all history for all branches and tags.

thaJeztah: Oh! Good catch. I also would've assumed `0` for "shallow", not the reverse 😬 😂

dvdksn: fixed, PTAL

crazy-max (approved):

thaJeztah (commented):
Do we need the `fetch-depth: 0` at line 33, or can that one also be removed?

If we need to keep it for that one, we should probably add a comment to explain "why" it's needed for that one (which may help prevent others copying that when creating new checks).

./cc @crazy-max

crazy-max:
> Do we need the `fetch-depth: 0` at line 33, or can that one also be removed?
> 
> If we need to keep it for that one, we should probably add a comment to explain "why" it's needed for that one (which may help prevent others copying that when creating new checks).
> 
> ./cc @crazy-max

It looks safe to remove 👍

.github/workflows/validate.yml
```diff
@@ -44,8 +44,6 @@ jobs:
       -
         name: Checkout
         uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
```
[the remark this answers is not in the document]
dvdksn: sorted it out 😆

thaJeztah (approved):
LGTM, thanks!

PR #4188, opened 2023-04-11
Explain the auto-removal of anonymous volumes at the description of 'create --rm' and 'run --rm'

finalchild:
Original:
Automatically remove the container when it exits
Change to:
Automatically remove the container and its associated anonymous volumes when it exits

The behavior is currently not properly documented and can be a source of confusion. Only docs/reference/run.md has a note about it.  `podman run --rm` has similar behavior and it [is documented](https://docs.podman.io/en/latest/markdown/podman-run.1.html#rm).

It is the same as calling `docker rm -v` on container exit. [code](https://github.com/docker/engine/blob/8955d8da8951695a98eb7e15bead19d402c6eb27/daemon/start.go#L137)


thaJeztah:
rebased; looks like most of this was done in https://github.com/docker/cli/pull/5084, but this PR had a couple of remaining spots.

thaJeztah (approved):
LGTM

vvoland (approved):

PR #4190, opened 2023-04-11
cli/command: some cleanups / refactoring, and fixes related to auth

thaJeztah:
### cli/command: ConfigureAuth: fix terminal state not being restored on error

ConfigureAuth used the readInput() utility to read the username and password.
However, this utility did not return errors it encountered, but instead did
an os.Exit(1). A result of this was that the terminal was not restored if
an error happened. When reading the password, the terminal is configured to
disable echo (i.e. characters are not printed), and failing to restore
the previous state means that the terminal is now "non-functional".

This patch:

- changes readInput() to return errors it encounters
- uses a defer() to restore terminal state

### cli/command: ConfigureAuth: trim whitespace both for username and password

changes readInput() to trim whitespace. The existing code tried to be
conservative and only trimmed whitespace for username (not for password).
Passwords with leading/trailing whitespace would be _very_ unlikely, and
trimming whitespace is generally accepted.

### cli/command/registry: remove intermediate var that collided

This also simplifies the code a bit.

### cli/command: replace EncodeAuthToBase64 for registry.EncodeAuthConfig

Replace uses of this function in favor of the implementation in the
API types, so that we have a single, canonical implementation.

### cli/command/container/create: pullImage(): use RetrieveAuthTokenFromImage

replace the local code with RetrieveAuthTokenFromImage, which does exactly the same;
https://github.com/docker/cli/blob/623356001f6310fef7a8cca0b9ba0ac1b735bb38/cli/command/registry.go#L163-L188

### cli/command/container: pullImage: use DisplayJSONMessagesToStream utility

This utility provides the same logic as was implemented here (and using it
aligns with the "docker pull" equivalent).

Also added a TODO to replace this function with the regular "docker pull"
code.

### DIFF
--- a/cli/command/container/create.go
+++ b/cli/command/container/create.go
@@ -12,6 +12,7 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/completion"
 	"github.com/docker/cli/cli/command/image"
+	"github.com/docker/cli/cli/streams"
 	"github.com/docker/cli/opts"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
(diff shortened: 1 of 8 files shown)


thaJeztah:
@rumpl @laurazard ptal 🤗

thaJeztah:
oh! I think I forgot to push some of the changes. let me dig them up

thaJeztah:
all green now 👍

vvoland (approved):

PR #4199, opened 2023-04-12
docs/reference: update dockerd docs for changes in 23.0

thaJeztah:
### docs/reference: remove mention of DOCKER_NOWARN_KERNEL_VERSION

Support for this environment variable was removed in docker 23.0 in
- https://github.com/moby/moby/commit/1240f8b41d34abe9252b882bddd42e140b62b83b (https://github.com/moby/moby/pull/43253)

From that patch:

> All regular, non-EOL Linux distros now come with more recent kernels
> out of the box. There may still be users trying to run on kernel 3.10
> or older (some embedded systems, e.g.), but those should be a rare
> exception, which we don't have to take into account.
>
> This patch removes the kernel version check on Linux, and the corresponding
> DOCKER_NOWARN_KERNEL_VERSION environment that was there to skip this
> check.

### docs/reference: reformat env-vars table, and simplify

No need to mention that the env-var may be removed at that point to keep
the description more to-the-point.

### docs/reference: use table for proxy env-vars

Adding a description based on the Go documentation.

### docs/reference: dockerd: update seccomp flag description

Update the description for the changes made in;
- https://github.com/moby/moby/commit/68e96f88ee1598563a66a1f53b8844291423fc88 (https://github.com/moby/moby/pull/42481)

### docs/reference: dockerd: remove default oom-score-adjust

The daemon no longer adjusts its oom-score by default; see

- https://github.com/moby/moby/commit/2b8e68ef06ca3daf2b0e17e51ba72999773b6270 (https://github.com/moby/moby/pull/41528)
- https://github.com/moby/moby/commit/cf7a5be0f2ee8bb4a3f7961682e13b89bc625299 (https://github.com/moby/moby/pull/41527)

### docs: update dockerd usage output for new proxy-options

Adds documentation for the options that were added in
- https://github.com/moby/moby/commit/427c7cc5f86364466c7173e8ca59b97c3876471d (https://github.com/moby/moby/pull/42835)

### DIFF
--- a/docs/reference/commandline/cli.md
+++ b/docs/reference/commandline/cli.md
@@ -123,19 +123,19 @@ the [installation](https://docs.docker.com/install/) instructions for your opera
 The following list of environment variables are supported by the `docker` command
 line:
 
-| Variable                      | Description                                                                                                                             |
-|:------------------------------|:----------------------------------------------------------------------------------------------------------------------------------------|
-| `DOCKER_API_VERSION`          | Override the negotiated API version to use for debugging (e.g. `1.19`)                                                                  |
-| `DOCKER_CERT_PATH`            | Location of your authentication keys. This variable is used both by the `docker` CLI and the [`dockerd` daemon](dockerd.md)             |
-| `DOCKER_CONFIG`               | The location of your client configuration files.                                                                                        |
-| `DOCKER_CONTENT_TRUST_SERVER` | The URL of the Notary server to use. Defaults to the same URL as the registry.                                                          |
-| `DOCKER_CONTENT_TRUST`        | When set Docker uses notary to sign and verify images. Equates to `--disable-content-trust=false` for build, create, pull, push, run.   |
-| `DOCKER_CONTEXT`              | Name of the `docker context` to use (overrides `DOCKER_HOST` env var and default context set with `docker context use`)                 |
-| `DOCKER_DEFAULT_PLATFORM`     | Default platform for commands that take the `--platform` flag.                                                                          |
-| `DOCKER_HIDE_LEGACY_COMMANDS` | When set, Docker hides "legacy" top-level commands (such as `docker rm`, and `docker pull`) in `docker help` output, and only `Management commands` per object-type (e.g., `docker container`) are printed. This may become the default in a future release, at which point this environment-variable is removed. |
-| `DOCKER_HOST`                 | Daemon socket to connect to.                                                                                                            |
-| `DOCKER_TLS_VERIFY`           | When set Docker uses TLS and verifies the remote. This variable is used both by the `docker` CLI and the [`dockerd` daemon](dockerd.md) |
-| `BUILDKIT_PROGRESS`           | Set type of progress output (`auto`, `plain`, `tty`) when [building](build.md) with [BuildKit backend](https://docs.docker.com/build/buildkit/). Use plain to show container output (default `auto`). |
+| Variable                      | Description                                                                                                                                                                                                                                                  |
+|:------------------------------|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `DOCKER_API_VERSION`          | Override the negotiated API version to use for debugging (e.g. `1.19`)                                                                                                                                                                                       |
+| `DOCKER_CERT_PATH`            | Location of your authentication keys. This variable is used both by the `docker` CLI and the [`dockerd` daemon](dockerd.md)                                                                                                                                  |
+| `DOCKER_CONFIG`               | The location of your client configuration files.                                                                                                                                                                                                             |
+| `DOCKER_CONTENT_TRUST_SERVER` | The URL of the Notary server to use. Defaults to the same URL as the registry.                                                                                                                                                                               |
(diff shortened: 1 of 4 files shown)


docs/reference/commandline/dockerd.md
```diff
@@ -35,14 +35,14 @@ Options:
...
+                                              parent cgroup for all containers (not supported with cgroups v2)
       --cpu-rt-runtime int                    Limit the CPU real-time runtime in microseconds for the
-                                              parent cgroup for all containers
+                                              parent cgroup for all containers (not supported with cgroups v2)
       --cri-containerd                        start containerd with cri
       --data-root string                      Root directory of persistent Docker state (default "/var/lib/docker")
   -D, --debug                                 Enable debug mode
       --default-address-pool pool-options     Default address pools for node specific local networks
-      --default-cgroupns-mode string          Default mode for containers cgroup namespace ("host" | "private") (default "host")
+      --default-cgroupns-mode string          Default mode for containers cgroup namespace ("host" | "private") (default "private")
```
thaJeztah, line 45: We still need to look at some of these, because the default depends on the host running either cgroups v2 or v1. Cgroups v2 should be the most common now, so updated the output for that.

thaJeztah:
@dvdksn ptal

docs/reference/commandline/dockerd.md
```diff
@@ -81,16 +83,17 @@ Options:
...
+      --no-proxy string                       Comma-separated list of hosts or IP addresses for which the proxy is skipped
       --node-generic-resource list            Advertise user-defined resource
-      --oom-score-adjust int                  Set the oom_score_adj for the daemon (default -500)
+      --oom-score-adjust int                  Set the oom_score_adj for the daemon
   -p, --pidfile string                        Path to use for daemon PID file (default "/var/run/docker.pid")
       --raw-logs                              Full timestamps without ANSI coloring
       --registry-mirror list                  Preferred registry mirror
       --rootless                              Enable rootless mode; typically used with RootlessKit
-      --seccomp-profile string                Path to seccomp profile
+      --seccomp-profile string                Path to seccomp profile. Use "unconfined" to disable the default seccomp profile (default "builtin")
```
dvdksn, line 96:
```suggestion
      --seccomp-profile string                Path to seccomp profile. Set to "unconfined" to disable the default seccomp profile (default "builtin")
```

docs/reference/commandline/dockerd.md
```diff
@@ -1222,6 +1251,9 @@ This is a full example of the allowed configuration options on Linux:
   "fixed-cidr-v6": "",
   "group": "",
   "hosts": [],
+  "http-proxy": "http://proxy.example.com:80",
+  "https-proxy": "https://proxy.example.com:443",
```
dvdksn, line 1255: this seems to always be a contentious point: the type of HTTP request to use for the proxy, and whether to use use different protocols/ports for HTTP_PROXY and HTTPS_PROXY. We should probably update the [proxy configuration page](https://docs.docker.com/network/proxy/) in a follow-up, to use the same example values, so we're consistent across the board.

docs/reference/commandline/dockerd.md
thaJeztah: Oh, yes, we can change; this was the actual output of the daemon, so I would need to make code-changes for that in moby/moby 😅

dvdksn: Yeah I figured it might've come from the CLI, let's update in moby then :)

docs/reference/commandline/cli.md
```diff
@@ -123,30 +123,31 @@ the [installation](https://docs.docker.com/install/) instructions for your opera
...
+| `DOCKER_HOST`                 | Daemon socket to connect to.                                                                                                                                                                                                                                 |
+| `DOCKER_TLS_VERIFY`           | When set Docker uses TLS and verifies the remote. This variable is used both by the `docker` CLI and the [`dockerd` daemon](dockerd.md)                                                                                                                      |
+| `BUILDKIT_PROGRESS`           | Set type of progress output (`auto`, `plain`, `tty`) when [building](build.md) with [BuildKit backend](https://docs.docker.com/build/buildkit/). Use plain to show container output (default `auto`).                                                        |
 
 Because Docker is developed using Go, you can also use any environment
 variables used by the Go runtime. In particular, you may find these useful:
```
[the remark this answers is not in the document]
thaJeztah: Yeah, I was wondering about that a bit. I think the main reason it's still mentioned is that _there is no standard_ for these variables; they're a common convention, but implementations differ (with sometimes the devil being in the details).

Some tools use `http_socket` (lowercase), some `HTTP_SOCKET` (uppercase), some _both_ (but order of preference may differ), and formats accepted for `NO_PROXY` ... differ very much.

See https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/
and https://github.com/docker/docs/pull/13786 (I still need to finish up that one)

thaJeztah:
Note to self; after this we also need to document the `--validate` option;

- https://github.com/moby/moby/pull/42393

docs/reference/commandline/dockerd.md
thaJeztah: I opened a PR;

- https://github.com/moby/moby/pull/45321

cpuguy83 (approved):

neersighted (approved):

PR #4200, opened 2023-04-12
cli/compose/types: ServiceVolumeConfig: use strong type for Type and Consistency fields

thaJeztah:
### cli/compose/types: ServiceVolumeConfig: use strong type for Type field

Using a strong type for this field, to make it clearer what's supported,
and to allow assigning values without having to convert them to a string.

Also renamed a variable that collided with an import.

### cli/compose/types: ServiceVolumeConfig: use strong type for Consistency field

Using a strong type for this field, to make it clearer what's supported,
and to allow assigning values without having to cast them.

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -380,7 +380,7 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		if parsed.Source != "" {
 			toBind := bind
 
-			if parsed.Type == string(mount.TypeBind) {
+			if parsed.Type == mount.TypeBind {
 				if hostPart, targetPath, ok := strings.Cut(bind, ":"); ok {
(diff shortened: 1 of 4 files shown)


cli/compose/types/types.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"time"
+
+	"github.com/docker/docker/api/types/mount"
```
thaJeztah, line 8: I think the reason these types were not imported was to make the compose-types not directly depend on docker API types, and to keep it "agnostic".

Now that the compose-spec exists, I don't think that should still be a concern, as the compose implementation in this repository is only for `docker stack`; users that want a more agnostic implementation are better off using the compose-spec as a library than the types from this repo.

Issue #4093, opened 2023-03-13
`docker ps -q` does not work if the config file contains a `psFormat` 


exhuma:
### Description

If the config file contains a customised "ps format", the `-q` option is ignored on the CLI, breaking tools relying on it.

### Reproduce

1. Run a container (so we have something for `ps` to show)
2. Add the following to `~/.docker/config.json`:
   ```
   {
        "psFormat": "table {{.Image}}\t{{.Status}}\t{{.Names}}\t{{.Ports}}"
}
   ```
3. Run `docker ps -q -a`

### Expected behavior

The CLI output should only show the contianer IDs.

Instead it shows the output from the `psFormat`.

### docker version

```bash
Client:
 Version:           23.0.1
 API version:       1.41 (downgraded from 1.42)
 Go version:        go1.20
 Git commit:        a5ee5b1dfc
 Built:             Sat Feb 11 13:58:04 2023
 OS/Arch:           linux/amd64
 Context:           default

Server:
 Engine:
  Version:          20.10.17
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.5
  Git commit:       a89b842
  Built:            Mon Oct 17 06:20:42 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.6.6
  GitCommit:        10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc:
  Version:          1.1.2
  GitCommit:        
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc.)
    Version:  2.16.0
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 5
 Server Version: 20.10.17
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2 io.containerd.runtime.v1.linux
[...]

PR #4040, opened 2023-02-17
cli: do not apply configfile's psFormat when --quiet is passed

lionelnicolas:
**- What I did**

My vscode was failing to start/create devcontainers after I upgrade docker-ce from 20.10 to 23.0. I realized that table headers were shown in the logs when vscode was running `docker ps -q -a --filter label=vsch.local.folder=.......`. Manually running `docker ps -aq` confirmed the issue (the full formatted container list was shown).

**- How I did it**

As I have a custom `psFormat` in my `~/.docker/config.json`, I removed it and after that `docker ps -aq` behaved correctly.

**- How to verify it**

For example, add `"psFormat": "table {{.ID}}\t{{.Image}}\t{{.Status}}\t{{.Ports}}\t{{.Names}}"` to your `~/.docker/config.json` and try to run `docker ps -aq`.

**- Description for the changelog**

Fix `--quiet` flag being ignored when a custom `psFormat` is set in the config file

**- A picture of a cute animal (not mandatory but encouraged)**

image

PR #4201, opened 2023-04-13
[23.0 backport] docker ps: always use --quiet, also combined with --format

thaJeztah:
- backport of https://github.com/docker/cli/pull/3666

----

- follow-up to https://github.com/docker/cli/pull/3640
- see the discussion on https://github.com/photra/cli/pull/1#issuecomment-1148364089
- fixes https://github.com/docker/cli/issues/4093
- supersedes / closes https://github.com/docker/cli/pull/4040
 

Previously, the formatter would ignore the quiet option if a custom format
was passed; this situation was handled in runPs(), where custom formats
would only be applied if the quiet option was not set, but only if the
format was set in the CLI's config.

This patch updates NewContainerFormat() to do the same, even if a `--format`
was passed on the command-line.

This is a change in behavior, so may need some discussion; possible alternatives;

- produce an error if both `--format` and `--quiet` are passed
- print a warning if both are passed (but use the logic from this patch)

Before this patch:

```console
docker ps --format '{{.Image}}'
ubuntu:22.04
alpine

docker ps --format '{{.Image}}' --quiet
ubuntu:22.04
alpine

mkdir -p ~/.docker/
echo '{"psFormat": "{{.Image}}" > ~/.docker/config.json

docker ps
ubuntu:22.04
alpine

docker ps --quiet
ubuntu:22.04
alpine
```

With this patch applied:

```console
docker ps --format '{{.Image}}'
ubuntu:22.04
alpine

docker ps --format '{{.Image}}' --quiet
40111f61d5c5
482efdf39fac

mkdir -p ~/.docker/
echo '{"psFormat": "{{.Image}}" > ~/.docker/config.json

docker ps
ubuntu:22.04
alpine

docker ps --quiet
40111f61d5c5
482efdf39fac
```

### docker ps: print warning if both --format and --quiet are set

Of both "--quiet" and "--format" are set, --quiet takes precedence. This
patch adds a warning to inform the user that their custom format is not
used:

    docker ps --format='{{.Image}}'
    ubuntu:22.04
    alpine
    
    docker ps --format='{{.Image}}' --quiet
    WARNING: Ignoring custom format, because both --format and --quiet are set.
    40111f61d5c5
    482efdf39fac
    
The warning is printed on STDERR, so can be redirected:

    docker ps --format='{{.Image}}' --quiet 2> /dev/null
    40111f61d5c5
    482efdf39fac

The warning is only shown if the format is set using the "--format" option.
No warning is shown if a custom format is set through the CLI configuration
file:

    mkdir -p ~/.docker/
    echo '{"psFormat": "{{.Image}}"}' > ~/.docker/config.json

    docker ps
    ubuntu:22.04
    alpine
    
    docker ps --quiet
    40111f61d5c5
    482efdf39fac

### DIFF
--- a/cli/command/container/list.go
+++ b/cli/command/container/list.go
@@ -120,6 +120,8 @@ func runPs(dockerCli command.Cli, options *psOptions) error {
 	if len(options.format) == 0 {
 		// load custom psFormat from CLI config (if any)
 		options.format = dockerCli.ConfigFile().PsFormat
+	} else if options.quiet {
+		_, _ = dockerCli.Err().Write([]byte("WARNING: Ignoring custom format, because both --format and --quiet are set.\n"))
 	}
 
 	listOptions, err := buildContainerListOptions(options)
--- a/cli/command/container/list_test.go
+++ b/cli/command/container/list_test.go
@@ -324,6 +324,7 @@ func TestContainerListWithFormat(t *testing.T) {
 		assert.Check(t, cmd.Flags().Set("format", "{{ .Names }} {{ .Image }} {{ .Labels }}"))
 		assert.Check(t, cmd.Flags().Set("quiet", "true"))
 		assert.NilError(t, cmd.Execute())
+		assert.Equal(t, cli.ErrBuffer().String(), "WARNING: Ignoring custom format, because both --format and --quiet are set.\n")
 		golden.Assert(t, cli.OutBuffer().String(), "container-list-quiet.golden")
 	})
 }
--- a/cli/command/container/list_test.go
+++ b/cli/command/container/list_test.go
@@ -309,8 +309,21 @@ func TestContainerListWithFormat(t *testing.T) {
 			}, nil
 		},
 	})
-	cmd := newListCommand(cli)
-	cmd.Flags().Set("format", "{{ .Names }} {{ .Image }} {{ .Labels }}")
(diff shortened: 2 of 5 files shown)


thaJeztah:
Discussing with @neersighted whether or not to backport this to v23.0;

👍 fixes `docker ps --quiet` with a custom format in `~/.docker/config.json`
👎 it's a change in behaviour (notably for `docker ps --format {''some format"}} --quiet`)


/cc @tianon @vvoland @cpuguy83 @AkihiroSuda WDYT?

tianon:
No strong feelings - I don't think it _needs_ a backport, but I think the new behavior is much more correct (and matches the user expectation better), so it's totally reasonable to backport if 23.x is going to continue to live for a while.

thaJeztah:
FWIW; I discussed with @neersighted and he was okay with this change, but wanted to have a quick peek from others

AkihiroSuda:
For v23, it might be safer to just retain the old behavior but print a warning message

thaJeztah:
Yeah, I guess the "complicated" one is the case where both as passed through `--flags`. The case where a default is configured in the `~/.docker/config.json` (and `--quiet` not working) is easier.

I could consider adding a commit that only changes the behaviour for `~/.docker/config.json` for 23.0, and we can decide on the other ones later (or keep it)

Issue #3793, opened 2022-09-29
docker info reports an index server url as a registry url


x-yuri:
### Description

```
$ docker info | grep Registry:
 Registry: https://index.docker.io/v1/

$ curl -sS https://index.docker.io/v1/ -o /dev/null -w '%{http_code}'
404
```

> IndexServerAddress
> string
> Default: "https://index.docker.io/v1/"
>
> Address / URL of the index server that is used for image search, and as a default for user authentication for Docker Hub and Docker Cloud.

https://docs.docker.com/engine/api/v1.41/#tag/System/operation/SystemInfo

I don't see how `index.docker.io` is used for authentication. Can you possibly provide steps to reproduce and some links to the source code?

From what I can [see][e] `docker` uses [`registry-1.docker.io`][a] to contact the [registry][b]. And what's used for [authentication][f] is `auth.docker.io`:

```
$ curl -sS https://registry-1.docker.io/v2/ -o /dev/null -w '%header{www-authenticate}' |& less
Bearer realm="https://auth.docker.io/token",service="registry.docker.io"
```

Not to mention that to get the `registry-1.docker.io` value I had to consult the source code.

It's not clear if there's a [difference][c] between a `docker` registry and an index server.

[Here][d] the domain is called legacy.

This whole thing is pretty damn confusing. Particularly the fact that `docker info` calls an index server a registry and provides a broken URL.

[a]: https://github.com/moby/moby/blob/v20.10.18/registry/service_v2.go#L33-L39
[b]: https://github.com/moby/moby/blob/v22.06.0-beta.0/registry/config.go#L25-L55
[c]: https://stackoverflow.com/questions/28394930/what-is-the-differenece-between-docker-registry-and-docker-index
[d]: https://github.com/moby/moby/blob/v20.10.18/vendor/github.com/docker/distribution/reference/normalize.go#L13
[e]: https://stackoverflow.com/a/73893304/52499
[f]: https://docs.docker.com/registry/spec/auth/token/

### Reproduce

```
$ docker info | grep Registry:
```

### Expected behavior

`docker info` shouldn't confuse users. It should be documented what's the difference between an index server and a docker registry. And if there's a difference, `docker info` shouldn't call an index server a registry. Especially provide a broken URL. Is it even used these days?

### docker version

```bash
Client:
 Version:           20.10.18
 API version:       1.41
 Go version:        go1.19.1
 Git commit:        b40c2f6b5d
 Built:             Thu Sep 29 08:07:17 2022
[...]

PR #4110, opened 2023-03-22
cli/command: remove deprecated ElectAuthServer()

thaJeztah:
- relates to https://github.com/docker/cli/pull/2819

This function was deprecated in b4ca1c7368daeead400fcc1b8f2d61951a0d9d1e (https://github.com/docker/cli/pull/2819), which is part of the v23.0 release, and is no longer used, so we can remove it.

**- Description for the changelog**

```markdown
- cli/command: remove deprecated ElectAuthServer()
```

**- A picture of a cute animal (not mandatory but encouraged)**

PR #4204, opened 2023-04-13
docker info: don't print "registry" (IndexServerAddress)

thaJeztah:
- fixes https://github.com/docker/cli/issues/3793
- relates to https://github.com/docker/cli/pull/2819
- relates to https://github.com/docker/cli/pull/4110

The IndexServerAddress field was  as part of the initial Windows implementation of the engine. For legal reasons, Microsoft Windows (and thus Docker images based on Windows) were not allowed to be distributed through non-Microsoft infrastructure. As a temporary solution, a dedicated "registry-win-tp3.docker.io" registry was created to serve Windows images.

Currently, this field always shows "https://index.docker.io/v1/", which is confusing, because that address is not used for the registry (only for authentication and "v1" search).

    docker info
    ...

Starting with b4ca1c7368daeead400fcc1b8f2d61951a0d9d1e, this field is also no longer used during authentication, and a3d56e7d067b38b98548a864436880eef526f971 removed the (deprecated) ElectAuthServer() which was previously used to query it.

Given that there's currently no practical use for this information, and it only adds "noise" (and confusion), this patch removes it from the default output.

For now, the field is (still) available for those that want to use it;

    docker info --format '{{.IndexServerAddress}}'
    https://index.docker.io/v1/

But it won't be printed by default.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/system/info.go
+++ b/cli/command/system/info.go
@@ -324,7 +324,6 @@ func prettyPrintServerInfo(dockerCli command.Cli, info types.Info) []error {
 		if len(u) > 0 {
 			fmt.Fprintln(dockerCli.Out(), " Username:", u)
 		}
-		fmt.Fprintln(dockerCli.Out(), " Registry:", info.IndexServerAddress)
 	}
 
 	if len(info.Labels) > 0 {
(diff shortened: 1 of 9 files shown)


vvoland (approved):

dvdksn:
We can remove the `docs/revisit` label as docker/docs#17106 takes care of this.

thaJeztah:
Thanks!

PR #4205, opened 2023-04-13
cli/command/system: prettyPrintServerInfo: don't depend on IndexServerAddress and credential-store

thaJeztah:
- relates to https://github.com/docker/cli/pull/2819

### cli/command/system: prettyPrintServerInfo: simplify username

Starting with b4ca1c7368daeead400fcc1b8f2d61951a0d9d1e (https://github.com/docker/cli/pull/2819), docker login  no longer depends on info.IndexServerAddress to determine the default registry.

The prettyPrintServerInfo() still depended on this information, which could potentially show the wrong information.

This patch changes it to also depend on the same information as docker login now does.

### cli/command/system: prettyPrintServerInfo: move out collecting username

Make this function only _print_ the info we have, and not read the username
from the credential-store.

This patch adds a Username field to the (local) `info` type, and sets it
when needed, so that prettyPrintServerInfo only has to format and print
the information, instead of calling out to the credential-store.

### cli/command/system: printServerWarnings: use client API version from info

Set the client's API version that's used in the info, instead of requesting
it as part of printing.

### DIFF
--- a/cli/command/system/info.go
+++ b/cli/command/system/info.go
@@ -19,6 +19,7 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/swarm"
 	"github.com/docker/docker/api/types/versions"
+	"github.com/docker/docker/registry"
 	"github.com/docker/go-units"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


cli/command/system/info.go
```diff
@@ -318,14 +322,7 @@ func prettyPrintServerInfo(dockerCli command.Cli, info types.Info) []error {
...
 	fprintlnNonEmpty(dockerCli.Out(), " No Proxy:", info.NoProxy)
-
-	if info.IndexServerAddress != "" {
-		u := dockerCli.ConfigFile().AuthConfigs[info.IndexServerAddress].Username
-		if len(u) > 0 {
-			fmt.Fprintln(dockerCli.Out(), " Username:", u)
-		}
-	}
-
+	fprintlnNonEmpty(dockerCli.Out(), " Username:", info.UserName)
```
thaJeztah, line 325: FWIW, I'm also very much leaning towards moving this information to the `Client:` section, because nothing about it is part of the "Server:".

I _think_ this was historically because it _depended_ on the server (?) but no real idea.

Also;

- because it's not part of the actual `info.Info` struct (i.e., the `Server` part), it's currently not even possible to use `docker info --format <template>` to show the name
- ^^ so `docker info --format=json` won't show it
- ^^ _AND_ if a credentials-helper is used, it's not even filled in _at all_ (it only looks in the config file
- 🙃 🤷

thaJeztah: So ... perhaps we should either remove it altogether, or make it use the actual credential-store in use, BUT reading info from that is not always .. fast, and I don't want `docker info` to become slow 😓

PR #4209, opened 2023-04-15
docs: deprecation: OOM-score adjust for the daemon

thaJeztah:
- relates to https://github.com/moby/moby/pull/45315

The `oom-score-adjust` option was added to prevent the daemon from being OOM-killed before other processes. This option was mostly added as a convenience, as running the daemon as a systemd unit was not yet common.

Having the daemon set its own limits is not best-practice, and something better handled by the process-manager starting the daemon.

Docker 20.10 and newer no longer adjust the daemon's OOM score by default, instead setting the OOM-score to the systemd unit (OOMScoreAdjust) that's shipped with the packages.

Users currently depending on this feature are recommended to adjust the daemon's OOM score using systemd or through other means, when starting the daemon.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -50,6 +50,7 @@ The table below provides an overview of the current status of deprecated feature
 
 | Status     | Feature                                                                                                                            | Deprecated | Remove  |
 |------------|------------------------------------------------------------------------------------------------------------------------------------|------------|---------|
+| Deprecated | [OOM-score adjust for the daemon](#oom-score-adjust-for-the-daemon)                                                                | v24.0.0    | v25.0.0 |
 | Removed    | [Buildkit build information](#buildkit-build-information)                                                                          | v23.0.0    | v24.0.0 |
 | Deprecated | [Legacy builder for Linux images](#legacy-builder-for-linux-images)                                                                | v23.0.0    | -       |
 | Deprecated | [Legacy builder fallback](#legacy-builder-fallback)                                                                                | v23.0.0    | -       |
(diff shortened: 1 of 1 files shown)


docs/deprecated.md
```diff
@@ -105,6 +106,26 @@ The table below provides an overview of the current status of deprecated feature
 | Removed    | [`--run` flag on `docker commit`](#--run-flag-on-docker-commit)                                                                    | v0.10      | v1.13   |
 | Removed    | [Three arguments form in `docker import`](#three-arguments-form-in-docker-import)                                                  | v0.6.7     | v1.12   |
 
+### OOM-score adjust for the daemon
+
+**Deprecated in Release: v24.0.0**
+**Target For Removal In Release: v25.0.0**
```
dvdksn, line 112:
```suggestion
**Subject for Removal in Release: v25.0.0**
```

dvdksn:
I just noticed that the formatting of these "Deprecated in Release" and "Removed in Release" phrases are a bit off. They use title case and bold, which isn't according to our style guide. And there's no blank line between these two lines, which renders them as one paragraph.

I will raise a follow-up to this to improve the formatting of these.

thaJeztah:
Yeah, it would use some tweaking; I used the existing approach, but definitely could use some touching up

docs/deprecated.md
thaJeztah: I'll keep this one to be the same, but we should indeed touch up this page a bit

dvdksn (approved):

PR #4211, opened 2023-04-17
scripts/warn-outside-container: fix escapes

kolyshkin:
By default, echo does not interpret escape sequences, resulting in the following output (literal):

	\033[1mWARNING\033[0m: you are not in a container.

I don't think this was intended.

Add -e option to echo so it can work as intended.

Also, use bash instead of sh because in POSIX shell echo does not take any options.

Fixes: 94e08f2e2db2eb9

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/scripts/warn-outside-container
+++ b/scripts/warn-outside-container
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 set -eu
 
 target="${1:-}"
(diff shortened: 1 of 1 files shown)


scripts/warn-outside-container
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
thaJeztah, line 1: Was there anything in the script requiring bash now, or would sh still work?

kolyshkin: Quoting myself

> Also, use bash instead of sh because in POSIX shell echo does not take any options.

kolyshkin: The alternative is to change all `echo` calls to `printf` (and add `\n` everywhere).

Issue #4183, opened 2023-04-10
[BUG] Additional property additional_contexts is not allowed when using docker stack deploy


aaguirre:
### Description

When using `additional_contexts` option in a docker compose file, it is not possible to use the same compose file to deploy a stack in swarm using the 'docker stack deploy' command. 

The following error message is generated:

`services.userregistry.build Additional property additional_contexts is not allowed
`

### Reproduce

1. Create a compose file with a service definition that includes an additional_contexts section inside the build block.
2. Deploy a stack using the command: docker stack deploy --compose-file ./compose-files/userregistry.yml --with-registry-auth <stack_name>

### Expected behavior

docker stack deploy should be able to support `additional_contexts` docker compose syntax 

### docker version

```bash
Client:
 Cloud integration: v1.0.31
 Version:           20.10.24
 API version:       1.41
 Go version:        go1.19.7
 Git commit:        297e128
 Built:             Tue Apr  4 18:21:21 2023
 OS/Arch:           darwin/amd64
 Context:           desktop-linux
 Experimental:      true
```

### docker info

```bash
Client:
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.10.4)
  compose: Docker Compose (Docker Inc., v2.17.2)
  dev: Docker Dev Environments (Docker Inc., v0.1.0)
  extension: Manages Docker extensions (Docker Inc., v0.2.19)
  init: Creates Docker-related starter files for your project (Docker Inc., v0.1.0-beta.2)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.25.0)
  scout: Command line tool for Docker Scout (Docker Inc., v0.9.0)

Server:
 Containers: 99
  Running: 48
  Paused: 0
  Stopped: 51
 Images: 38
 Server Version: 22.06.0-beta.0-926-g914b02ebaf.m
 Storage Driver: stargz
  driver-type: io.containerd.snapshotter.v1
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: active
  NodeID: 7zby7rersxcwl9fwtu2gapxn5
  Is Manager: true
[...]

PR #4214, opened 2023-04-18
Swarm: Add schema 3.11, and Allow additional properties in `build`

laurazard:
**- What I did**

Since Swarm does not use the `build` section, there's no reason to validate properties here.
This makes it so we don't have to keep updating the schema in the CLI to support properties added in the Compose Spec for build, and does not imply any new feature support since Swarm does not consider this section.

**- How I did it**

Update the schema in the CLI to allow `additionalProperties` in `build`.

**- How to verify it**

Write a Compose file such as
```yaml
services:
  a:
    image: alpine
    build:
      whatever: "meow"
```

and deploy it with `docker stack deploy -c <file> [name]`

**related**: 
- closes https://github.com/docker/cli/issues/4183
- supersedes / closes https://github.com/docker/cli/pull/4113

image

### DIFF
--- a/cli/compose/loader/loader_test.go
+++ b/cli/compose/loader/loader_test.go
@@ -179,7 +179,7 @@ func strPtr(val string) *string {
 }
 
 var sampleConfig = types.Config{
-	Version: "3.10",
+	Version: "3.11",
 	Services: []types.ServiceConfig{
 		{
 			Name:        "foo",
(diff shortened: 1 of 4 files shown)


laurazard:
Yup! Sorry, should have left this as a draft, I wanted to look at the tests but started looking at something else in the meantime and got distracted 😅

thaJeztah (approved):
LGTM

vvoland (approved):

PR #4079, opened 2023-03-09
Update docs/command output for volume prune to specify when named volumes will be deleted

dwminer:
In previous versions of the Docker API, `system prune --volumes` and `volume prune` would remove all dangling volumes. With API v1.42, this was changed so that only anonymous volumes would be removed unless the `all` filter was specified. Update the docs and command output to reflect this.

Fixes #4028 

**- What I did**
Updated documentation/warning output for the `system_prune` and `volume_prune` commands to reflect that only anonymous volumes are removed by default.
**- How I did it**
* Updated doc files
* In `volume/prune.go`, specify that *anonymous* volumes will be removed. If `all=1` is set, indicate that named volumes will be removed too.

**- How to verify it**
`docker volume prune`
`docker volume prune --filter all=1`

**- Description for the changelog**
Update `volume prune` output to be more specific about whether named volumes will be removed.

**- A picture of a cute animal (not mandatory but encouraged)**

image

Issue #4015, opened 2023-02-09
docker volume prune keeps dangling volumes


CTaeumel:
### Description

System: Ubuntu Server 20.04/22.04 - tested on fresh install as well as old systems with current package updates

When running docker volume prune, it deletes no volumes.

### Reproduce

1. docker volume create test
2. docker volume prune
--> test-volume is not deleted

### Expected behavior

test-volume is deleted

### docker version

```bash
Client: Docker Engine - Community
 Version:           23.0.0
 API version:       1.42
 Go version:        go1.19.5
 Git commit:        e92dd87
 Built:             Wed Feb  1 17:49:08 2023
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          23.0.0
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.19.5
  Git commit:       d7573ab
  Built:            Wed Feb  1 17:49:08 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.16
  GitCommit:        31aa4358a36870b21a992d3ad2bef29e1d693bec
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.2
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.15.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose
  scan: Docker Scan (Docker Inc.)
    Version:  v0.23.0
    Path:     /usr/libexec/docker/cli-plugins/docker-scan

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 0
 Server Version: 23.0.0
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

PR #4218, opened 2023-04-20
volumes: prune: add --all / -a option

thaJeztah:
- fixes / addresses https://github.com/docker/cli/issues/4028#issuecomment-1429538149
- fixes / addresses https://github.com/docker/cli/issues/4015
- relates to https://github.com/docker/cli/pull/4079
- relates to https://github.com/moby/moby/pull/44259
- relates to https://github.com/moby/moby/pull/44216

**- What I did**

**- How I did it**

**- How to verify it**


contrib/completion/zsh/_docker
```diff
@@ -2527,6 +2527,8 @@ __docker_volume_subcommand() {
         (prune)
             _arguments $(__docker_arguments) \
                 $opts_help \
+                "($help -a --all)"{-a,--all}"[Remove all unused volumes, not just anonymous ones]" \
```
felipecruz91, line 2530: nit: perhaps it's a bit redundant, but just to be consistent with the warning messages which indicates the term `local`:

```suggestion
                "($help -a --all)"{-a,--all}"[Remove all unused local volumes, not just anonymous ones]" \
```

felipecruz91 (approved):
LGTM

cli/command/volume/prune.go
```diff
@@ -41,18 +45,37 @@ func NewPruneCommand(dockerCli command.Cli) *cobra.Command {
...
+	flags.SetAnnotation("all", "version", []string{"1.42"})
 	flags.BoolVarP(&options.force, "force", "f", false, "Do not prompt for confirmation")
 	flags.Var(&options.filter, "filter", `Provide filter values (e.g. "label=<label>")`)
 
 	return cmd
 }
 
-const warning = `WARNING! This will remove all local volumes not used by at least one container.
+const (
+	unusedVolumesWarning = `WARNING! This will remove all local anonymous volumes not used by at least one container.
```
tianon, line 57: Can we change this message around a little so that the all vs non-all difference is more obvious?

```suggestion
	unusedVolumesWarning = `WARNING! This will remove anonymous local volumes not used by at least one container.
```

For comparison, the "all" message:

```go
	allVolumesWarning = `WARNING! This will remove all local volumes not used by at least one container.
```

thaJeztah: Ah, yes, that's better.

I was even considering if we need the  warning at all for the anonymous volumes (and only a warning for the `--all` case), but left it for now

contrib/completion/zsh/_docker
thaJeztah: done 👍

thaJeztah:
updated; PTAL

thaJeztah:
I'm wondering if we should backport this one for 23.0. It's a new flag, but I think we really just forgot to add this flag to the CLI (and it was the  original intent to add it, we just overlooked it).

Basically, I want to avoid users having to get used to `--filter all=1` (only for it to become mostly "obsolete" after 23.0)

tianon (approved):

tianon:
A backport to 23.x sounds sane to me :sweat_smile:

cpuguy83 (approved):

cpuguy83:
Backport also SGTM.

albers (commented):
Completion LGTM

Issue #4221, opened 2023-04-20
docker.example.com is down, so docker compose doesn't work anymore


crawlchange:
### Description

Doing anything using docker compose on a clean server breaks because "docker.example.com" is down, and, unfortunately, it seems that docker compose does not work without that website (which is already bad).

error during connect: Get "http://docker.example.com/v1.24/version": command [ssh -l root -- **ip** docker system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=ssh: connect to host **ip** port 22: Connection refused

### Steps To Reproduce

-Use the appropriate context to point to the machine

-Run "docker compose build"

### Compose Version

```Text
Docker Compose version v2.17.2
```

### Docker Environment

```Text
Docker version 23.0.4, build f480fb1
```

thaJeztah:
I'd have to look up history, but I guess we used this one because "a" domain is needed for this (but "anything goes"). The `example.com` was probably used to prevent any possibility of hitting an _actual_ domain (and maybe `localhost` was confusing for this feature, as it would be a remote server).

Looking at https://www.rfc-editor.org/rfc/rfc2606.html#section-2

Perhaps something like `docker.localhost` or `api.localhost` could still work (suggestions welcome!)

crawlchange:
In my particular case, the actual issue generating that error message was a digital ocean droplet without enough resources. Presumably, the server shut down connections.

Docker's response should probably be something like "the remote connection was shut down", instead of that message. That would have saved many hours, since that message led me to completely unrelated issues.

laurazard:
Right, thanks for the context!
The last part of that error is actually very relevant since it includes the ssh command that the CLI used to connect to the remote host and what the error returned by that command was (`ssh: connect to host ip port 22: Connection refused`), which seems like it would be the most important for a user trying to debug an issue with their setup.
The issue seems to be that the beginning is misleading, as we're not actually making an HTTP Get call to some random host, but I'm working on some of the remote connection/ssh stuff, I'll try to make the error handling a bit better :)

PR #4226, opened 2023-04-24
ssh: fix error on commandconn close, add ping and default timeout

laurazard:
**- Related**
- https://github.com/docker/cli/pull/3722
- https://github.com/docker/cli/issues/3652
- https://github.com/docker/cli/pull/3900
- https://github.com/docker/cli/issues/4221

**- TLDR**

During CLI initialization we ping the configured host. Initially, we set no context timeout for these connections, and after https://github.com/docker/cli/commit/b39739123b845f872549e91be184cc583f5b387c we set timeouts but only for TCP connections (which then got changed in https://github.com/docker/cli/pull/3722 to only specifically exclude SSH connections and not sockets).

This looks to be due to _errory_ looking logs from `cli/connhelper/commandconn/commandconn.go` on `Close()` – which gets called by `http.Transport` when the request context times out.
However, these logs are being thrown erroneously – `commandconn.kill()` states that it will return nil if the command terminates regardless of it's exit status:
 https://github.com/docker/cli/blob/26a73575f1e3a0d1dc6ee6592db5926dcb907feb/cli/connhelper/commandconn/commandconn.go#L113
But the implementation does not hold up to this contract:
https://github.com/docker/cli/blob/26a73575f1e3a0d1dc6ee6592db5926dcb907feb/cli/connhelper/commandconn/commandconn.go#L130
since `ProcessState.Exited()` returns true only if the program exited due to calling `exit` and not if the program was terminated by a signal (this probably does works on Windows since here `ProcessState.Exited()` returns true if the program has exited either way, but not on POSIXy OSs).

If `commandconn.kill()` is fixed to always return `nil` when the command has terminated, setting timeouts on contexts passed to SSH connections works fine and we no longer need special handling.

Other than that, I also added `-o ConnectTimeout=30` to the created SSH command, which serves to align with the timeout from the client returned for non-SSH connections in https://github.com/docker/cli/blob/26a73575f1e3a0d1dc6ee6592db5926dcb907feb/cli/context/docker/load.go#L134 since it will be used only for "dialing" i.e., ssh command is unable to establish a connection within 30s it will error out, but if it can establish a connection it will keep running indefinitely, so commands like `docker stats` or an `attach` won't time out after 30s.

During this I also found a racy call to `cmd.Wait()` between `Close()` and `onEOF()`, since calling `Close` will close the command's stdin/out pipes, which will throw an EOF to pending reads, triggering a call to `onEOF()`, but afterwards `Close` will also call `kill()` which will also call `cmd.Wait()`

**- What I did**
- *`cli/connhelper/commandconn/commandconn.go`*:
    - `kill()`: Fix implementation so we don't return an error when process is successfully closed
    - `Write()`/`Read()`: Prevent race by checking if connection is being closed
- *`cli/command/cli.go`*:
    - `initializeFromClient()`: remove SSH special handling so we also add timeout for SSH connections
- *`cli/context/docker/load.go`*:
    - `ClientOpts()`: add 30s timeout to align with non-SSH client

**- How to verify it**
- `DOCKER_HOST=ssh://example.com docker`: no longer hangs forever, now it returns after the default 2s CLI init timeout
- `DOCKER_HOST=ssh://example.com docker version`: no longer hangs forever, now it will timeout after 30s (can be more if the host resolves to more than 1 IP address)

image

### DIFF
--- a/cli/connhelper/commandconn/commandconn.go
+++ b/cli/connhelper/commandconn/commandconn.go
@@ -24,6 +24,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -64,81 +65,68 @@ func New(_ context.Context, cmd string, args ...string) (net.Conn, error) {
 
 // commandConn implements net.Conn
 type commandConn struct {
-	cmd           *exec.Cmd
-	cmdExited     bool
-	cmdWaitErr    error
-	cmdMutex      sync.Mutex
-	stdin         io.WriteCloser
-	stdout        io.ReadCloser
-	stderrMu      sync.Mutex
-	stderr        bytes.Buffer
-	stdioClosedMu sync.Mutex // for stdinClosed and stdoutClosed
-	stdinClosed   bool
-	stdoutClosed  bool
-	localAddr     net.Addr
-	remoteAddr    net.Addr
+	cmdMutex     sync.Mutex // for cmd, cmdWaitErr
+	cmd          *exec.Cmd
+	cmdWaitErr   error
+	cmdExited    atomic.Bool
+	stdin        io.WriteCloser
+	stdout       io.ReadCloser
+	stderrMu     sync.Mutex // for stderr
+	stderr       bytes.Buffer
+	stdinClosed  atomic.Bool
+	stdoutClosed atomic.Bool
+	closing      atomic.Bool
+	localAddr    net.Addr
+	remoteAddr   net.Addr
 }
 
-// killIfStdioClosed kills the cmd if both stdin and stdout are closed.
-func (c *commandConn) killIfStdioClosed() error {
-	c.stdioClosedMu.Lock()
-	stdioClosed := c.stdoutClosed && c.stdinClosed
-	c.stdioClosedMu.Unlock()
-	if !stdioClosed {
-		return nil
+// kill terminates the process. On Windows it kills the process directly,
+// whereas on other platforms, a SIGTERM is sent, before forcefully terminating
+// the process after 3 seconds.
+func (c *commandConn) kill() {
+	if c.cmdExited.Load() {
+		return
 	}
-	return c.kill()
-}
-
-// killAndWait tries sending SIGTERM to the process before sending SIGKILL.
-func killAndWait(cmd *exec.Cmd) error {
+	c.cmdMutex.Lock()
 	var werr error
 	if runtime.GOOS != "windows" {
 		werrCh := make(chan error)
-		go func() { werrCh <- cmd.Wait() }()
-		cmd.Process.Signal(syscall.SIGTERM)
+		go func() { werrCh <- c.cmd.Wait() }()
+		_ = c.cmd.Process.Signal(syscall.SIGTERM)
 		select {
 		case werr = <-werrCh:
 		case <-time.After(3 * time.Second):
-			cmd.Process.Kill()
+			_ = c.cmd.Process.Kill()
 			werr = <-werrCh
 		}
 	} else {
-		cmd.Process.Kill()
-		werr = cmd.Wait()
+		_ = c.cmd.Process.Kill()
+		werr = c.cmd.Wait()
 	}
-	return werr
-}
-
-// kill returns nil if the command terminated, regardless to the exit status.
-func (c *commandConn) kill() error {
-	var werr error
-	c.cmdMutex.Lock()
-	if c.cmdExited {
-		werr = c.cmdWaitErr
-	} else {
-		werr = killAndWait(c.cmd)
-		c.cmdWaitErr = werr
-		c.cmdExited = true
-	}
+	c.cmdWaitErr = werr
 	c.cmdMutex.Unlock()
-	if werr == nil {
-		return nil
-	}
-	wExitErr, ok := werr.(*exec.ExitError)
-	if ok {
-		if wExitErr.ProcessState.Exited() {
-			return nil
-		}
-	}
-	return errors.Wrapf(werr, "commandconn: failed to wait")
+	c.cmdExited.Store(true)
 }
 
-func (c *commandConn) onEOF(eof error) error {
-	// when we got EOF, the command is going to be terminated
-	var werr error
+// handleEOF handles io.EOF errors while reading or writing from the underlying
+// command pipes.
+//
+// When we've received an EOF we expect that the command will
+// be terminated soon. As such, we call Wait() on the command
+// and return EOF or the error depending on whether the command
+// exited with an error.
+//
+// If Wait() does not return within 10s, an error is returned
+func (c *commandConn) handleEOF(err error) error {
+	if err != io.EOF {
+		return err
+	}
+
 	c.cmdMutex.Lock()
-	if c.cmdExited {
+	defer c.cmdMutex.Unlock()
+
+	var werr error
+	if c.cmdExited.Load() {
 		werr = c.cmdWaitErr
 	} else {
 		werrCh := make(chan error)
@@ -146,18 +134,17 @@ func (c *commandConn) onEOF(eof error) error {
 		select {
 		case werr = <-werrCh:
 			c.cmdWaitErr = werr
-			c.cmdExited = true
+			c.cmdExited.Store(true)
 		case <-time.After(10 * time.Second):
-			c.cmdMutex.Unlock()
 			c.stderrMu.Lock()
 			stderr := c.stderr.String()
 			c.stderrMu.Unlock()
-			return errors.Errorf("command %v did not exit after %v: stderr=%q", c.cmd.Args, eof, stderr)
+			return errors.Errorf("command %v did not exit after %v: stderr=%q", c.cmd.Args, err, stderr)
 		}
 	}
-	c.cmdMutex.Unlock()
+
 	if werr == nil {
-		return eof
+		return err
 	}
 	c.stderrMu.Lock()
 	stderr := c.stderr.String()
@@ -166,71 +153,88 @@ func (c *commandConn) onEOF(eof error) error {
 }
 
 func ignorableCloseError(err error) bool {
-	errS := err.Error()
-	ss := []string{
-		os.ErrClosed.Error(),
-	}
-	for _, s := range ss {
-		if strings.Contains(errS, s) {
-			return true
-		}
-	}
-	return false
-}
-
-func (c *commandConn) CloseRead() error {
-	// NOTE: maybe already closed here
-	if err := c.stdout.Close(); err != nil && !ignorableCloseError(err) {
-		logrus.Warnf("commandConn.CloseRead: %v", err)
-	}
-	c.stdioClosedMu.Lock()
-	c.stdoutClosed = true
-	c.stdioClosedMu.Unlock()
-	if err := c.killIfStdioClosed(); err != nil {
-		logrus.Warnf("commandConn.CloseRead: %v", err)
-	}
-	return nil
+	return strings.Contains(err.Error(), os.ErrClosed.Error())
 }
 
 func (c *commandConn) Read(p []byte) (int, error) {
 	n, err := c.stdout.Read(p)
-	if err == io.EOF {
-		err = c.onEOF(err)
+	// check after the call to Read, since
+	// it is blocking, and while waiting on it
+	// Close might get called
+	if c.closing.Load() {
+		// If we're currently closing the connection
+		// we don't want to call onEOF, but we do want
+		// to return an io.EOF
+		return 0, io.EOF
 	}
-	return n, err
-}
 
-func (c *commandConn) CloseWrite() error {
-	// NOTE: maybe already closed here
-	if err := c.stdin.Close(); err != nil && !ignorableCloseError(err) {
-		logrus.Warnf("commandConn.CloseWrite: %v", err)
-	}
-	c.stdioClosedMu.Lock()
-	c.stdinClosed = true
-	c.stdioClosedMu.Unlock()
-	if err := c.killIfStdioClosed(); err != nil {
-		logrus.Warnf("commandConn.CloseWrite: %v", err)
-	}
-	return nil
+	return n, c.handleEOF(err)
 }
 
 func (c *commandConn) Write(p []byte) (int, error) {
 	n, err := c.stdin.Write(p)
-	if err == io.EOF {
-		err = c.onEOF(err)
+	// check after the call to Write, since
+	// it is blocking, and while waiting on it
+	// Close might get called
+	if c.closing.Load() {
+		// If we're currently closing the connection
+		// we don't want to call onEOF, but we do want
+		// to return an io.EOF
+		return 0, io.EOF
 	}
-	return n, err
+
+	return n, c.handleEOF(err)
 }
 
+// CloseRead allows commandConn to implement halfCloser
+func (c *commandConn) CloseRead() error {
+	// NOTE: maybe already closed here
+	if err := c.stdout.Close(); err != nil && !ignorableCloseError(err) {
+		return err
+	}
+	c.stdoutClosed.Store(true)
+
+	if c.stdinClosed.Load() {
+		c.kill()
+	}
+
+	return nil
+}
+
+// CloseWrite allows commandConn to implement halfCloser
+func (c *commandConn) CloseWrite() error {
+	// NOTE: maybe already closed here
+	if err := c.stdin.Close(); err != nil && !ignorableCloseError(err) {
+		return err
+	}
+	c.stdinClosed.Store(true)
+
+	if c.stdoutClosed.Load() {
+		c.kill()
+	}
+	return nil
+}
+
+// Close is the net.Conn func that gets called
+// by the transport when a dial is cancelled
+// due to it's context timing out. Any blocked
+// Read or Write calls will be unblocked and
+// return errors. It will block until the underlying
+// command has terminated.
 func (c *commandConn) Close() error {
-	var err error
-	if err = c.CloseRead(); err != nil {
+	c.closing.Store(true)
+	defer c.closing.Store(false)
+
+	if err := c.CloseRead(); err != nil {
 		logrus.Warnf("commandConn.Close: CloseRead: %v", err)
+		return err
 	}
-	if err = c.CloseWrite(); err != nil {
+	if err := c.CloseWrite(); err != nil {
 		logrus.Warnf("commandConn.Close: CloseWrite: %v", err)
+		return err
 	}
-	return err
+
+	return nil
 }
 
 func (c *commandConn) LocalAddr() net.Addr {
--- a/cli/connhelper/commandconn/commandconn_unix_test.go
+++ b/cli/connhelper/commandconn/commandconn_unix_test.go
@@ -6,7 +6,9 @@ import (
 	"context"
 	"io"
 	"testing"
+	"time"
 
+	"github.com/docker/docker/pkg/process"
 	"gotest.tools/v3/assert"
 	is "gotest.tools/v3/assert/cmp"
 )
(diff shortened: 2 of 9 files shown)


cli/connhelper/commandconn/commandconn.go
```diff
@@ -90,48 +90,31 @@ func (c *commandConn) killIfStdioClosed() error {
...
 }
 
-// killAndWait tries sending SIGTERM to the process before sending SIGKILL.
-func killAndWait(cmd *exec.Cmd) error {
+// kill returns nil if the command terminated, regardless to the exit status.
+func (c *commandConn) kill() error {
+	c.cmdMutex.Lock()
+	if c.cmdExited {
+		return nil
+	}
```
thaJeztah, line 98: It looks like we're not releasing the `.Lock()`; I think that lock is only for the `cmdExited`? If so we can release it in both cases;

```suggestion
	c.cmdMutex.Lock()
	if c.cmdExited {
		c.cmdMutex.Unlock()
		return nil
	}
	c.cmdMutex.Unlock()
```

laurazard: Yup! I meant to clean that part up a little more, thanks for pointing me in the right direction!

I need to write some more comprehensive tests for this part too, there's also https://github.com/laurazard/cli/blob/fix-connhelper-docker-example/cli/connhelper/commandconn/commandconn.go#L108 where we can hang forever if we never get another message in the channel since we try to read again and don't have any escape condition 😅

cli/context/docker/load.go
```diff
@@ -103,6 +103,7 @@ func (c *Endpoint) ClientOpts() ([]client.Opt, error) {
 			httpClient := &http.Client{
 				// No tls
 				// No proxy
+				Timeout: 30 * time.Second,
```
thaJeztah, line 106: 🤔 Looks like this always cancels the connection, also for commands that may have to be running longer, for example;


```bash
time ./build/docker -H ssh://<remote machine ip> stats

...

CONTAINER ID   NAME                  CPU %     MEM USAGE / LIMIT     MEM %     NET I/O         BLOCK I/O         PIDS
4e67d3197705   nifty_hopper          0.00%     1.961MiB / 969.4MiB   0.20%     796B / 0B       8.19kB / 12.3kB   2
4aa507efd17b   relaxed_agnesi        0.00%     1.957MiB / 969.4MiB   0.20%     796B / 0B       0B / 12.3kB       2
5abb0f50e460   sweet_jang            0.00%     1.965MiB / 969.4MiB   0.20%     796B / 0B       8.19kB / 12.3kB   2
0f55fc6abd68   serene_elion          0.00%     1.953MiB / 969.4MiB   0.20%     796B / 0B       0B / 12.3kB       2
52e68d5c6261   stoic_wu              0.00%     1.957MiB / 969.4MiB   0.20%     796B / 0B       0B / 12.3kB       2
d268bc02bf6a   strange_ellis         0.00%     2.082MiB / 969.4MiB   0.21%     796B / 0B       164kB / 12.3kB    2
c7fd2fcd83eb   funny_albattani       0.00%     2.023MiB / 969.4MiB   0.21%     796B / 0B       73.7kB / 12.3kB   2
6521caf5828c   frosty_sammet         0.00%     2.148MiB / 969.4MiB   0.22%     796B / 0B       315kB / 12.3kB    2
ee9a3f87920c   sweet_feynman         0.00%     2.023MiB / 969.4MiB   0.21%     796B / 0B       77.8kB / 12.3kB   2
b1ecdb1b5eea   vigorous_meitner      0.00%     1.957MiB / 969.4MiB   0.20%     796B / 0B       24.6kB / 12.3kB   2
fd0f075961a3   stupefied_antonelli   0.00%     5.25MiB / 969.4MiB    0.54%     796B / 0B       5.2MB / 12.3kB    2
270a62d2d482   foo                   0.00%     3.16MiB / 969.4MiB    0.33%     170kB / 222kB   5.78MB / 12.3kB   2
context deadline exceeded (Client.Timeout or context cancellation while reading body)

________________________________________________________
Executed in   30.65 secs      fish           external
   usr time  204.56 millis    0.20 millis  204.36 millis
   sys time  192.99 millis    2.09 millis  190.90 millis
```

(same probably applies to pulling images which take longer than 30 seconds?)

Trying after the image was pulled initially works, but keeping the command running it produces an error after 30 seconds as well. While that error is printed, it's only printed once, but after that the shell still is functional 🤔

```bash
time ./build/docker -H ssh://<remote machine ip> run -it --rm alpine
/ # date
Fri Apr 28 14:30:33 UTC 2023
/ # ERRO[0031] error waiting for container:

/ # date
Fri Apr 28 14:31:05 UTC 2023
/ #
```

I also ran into this error, but only once, so not sure what caused it; in this case the image didn't exist locally, so wondering if something is because multiple API calls are made? (create  -> pull -> create -> attach?). I tried to reproduce it again, but wasn't able to, so perhaps it was just a networking glitch or something.


```bash
time ./build/docker -H ssh://<remote machine ip> run -it --rm alpine
Unable to find image 'alpine:latest' locally
f56be85fc22e: Already exists
Digest: sha256:124c7d2707904eea7431fffe91522a01e5a861a624ee31d03372cc1d138a3126
Status: Downloaded newer image for alpine:latest
command [ssh -- <remote machine ip> docker system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=kex_exchange_identification: Connection closed by remote host


________________________________________________________
Executed in    3.91 secs      fish           external
   usr time   51.93 millis    0.21 millis   51.72 millis
   sys time  104.69 millis    2.39 millis  102.30 millis
```

thaJeztah: ^^^ just double-checked with the current CLI before this change, and that doesn't expose this issue

cli/connhelper/commandconn/commandconn.go
```diff
@@ -200,16 +212,32 @@ func (c *commandConn) CloseWrite() error {
...
+// due to it's context timing out. Any blocked
+// Read or Write calls will be unblocked and
+// return errors. It will block until the underlying
+// command has terminated.
 func (c *commandConn) Close() error {
+	c.closing.Store(true)
+	defer c.closing.Store(false)
+
 	var err error
 	if err = c.CloseRead(); err != nil {
```
thaJeztah, line 242: Not new, but this code looks a bit iffy;

- `c.CloseRead()` is assigned to `err`, and logs it if it went wrong
- `c.CloseWrite()` is then unconditionally overwriting `err` (and logs if it went wrong)
- the function as a whole returns the last result (which could be `nil` if `c.CloseWrite()` succeeded, but `c.CloseRead()` failed

Wondering what the expected result is here;

- should we ignore the first error if the second was successful? 
- do we need to return an error if _either_ failed?
- do we need a `multiErr`, an errgroup (or variant thereof?)

cli/connhelper/commandconn/commandconn.go
```diff
@@ -113,24 +111,32 @@ func (c *commandConn) kill() error {
...
+// When we've received an EOF we expect that the command will
+// be terminated soon. As such, we call Wait() on the command
+// and return EOF or the error depending on whether the command
+// exited with an error.
+//
+// If Wait() does not return within 10s, an error is returned
 func (c *commandConn) onEOF(eof error) error {
-	// when we got EOF, the command is going to be terminated
 	var werr error
 	c.cmdMutex.Lock()
```
thaJeztah, line 130: Curious now; should we keep the lock for the whole function, or do we only need it to protect `c.cmdWaitErr` ?

In the latter case, we could change these to

```go
c.cmdMutex.Lock()
werr = c.cmdWaitErr
c.cmdMutex.Unlock()
```

and (second "if branch")

```go
c.cmdMutex.Lock()
c.cmdWaitErr = werr
c.cmdMutex.Unlock()
```

cli/connhelper/commandconn/commandconn.go
```diff
@@ -178,7 +182,17 @@ func (c *commandConn) CloseRead() error {
 
 func (c *commandConn) Read(p []byte) (int, error) {
 	n, err := c.stdout.Read(p)
-	if err == io.EOF || errors.Is(err, os.ErrClosed) {
+	// check after the call to Read, since
+	// it is blocking, and while waiting on it
+	// Close might get called
+	if c.closing.Load() {
```
thaJeztah: Some observations (but may be intentional)

- `os.ErrClosed` is no longer handled in the new code
-  is it safe to discard `err` if `c.closing.Load()` is true? (not sure if there's other errors that may return)

cli/connhelper/commandconn/commandconn.go
```diff
@@ -200,16 +212,32 @@ func (c *commandConn) CloseWrite() error {
 
 func (c *commandConn) Write(p []byte) (int, error) {
 	n, err := c.stdin.Write(p)
-	if err == io.EOF || errors.Is(err, os.ErrClosed) {
```
thaJeztah: (same here)

cli/connhelper/connhelper.go
```diff
@@ -47,7 +47,7 @@ func getConnectionHelper(daemonURL string, sshFlags []string) (*ConnectionHelper
 		}
 		return &ConnectionHelper{
 			Dialer: func(ctx context.Context, network, addr string) (net.Conn, error) {
-				return commandconn.New(ctx, "ssh", append(sshFlags, sp.Args("docker", "system", "dial-stdio")...)...)
+				return commandconn.New(ctx, "ssh", append([]string{"-o ConnectTimeout=30"}, append(sshFlags, sp.Args("docker", "system", "dial-stdio")...)...)...)
```
thaJeztah, line 50: This reads rather awkward; also looking;

- `getConnectionHelper()` is called both by `GetConnectionHelper()` and `GetConnectionHelperWithSSHOpts()`
- `-o ConnectTimeout=30` is an `sshFlag` ?

Wondering; should this option be set _both_ for `GetConnectionHelper` and for `GetConnectionHelperWithSSHOpts` (which we currently don't use ourselves, but others may be using it)

Some options;

If only for `GetConnectionHelper`, then we could pass it as option when calling `getConnectionHelper()`

```go
func GetConnectionHelper(daemonURL string) (*ConnectionHelper, error) {
	return getConnectionHelper(daemonURL, []string{"-o ConnectTimeout=30"})
}
```

If it _must_ be set for both, then we could add it here to `sshFlags`;

```go
		sshFlags = append([]string{"-o ConnectTimeout=30"}, sshFlags...)
		return &ConnectionHelper{
				return commandconn.New(ctx, "ssh", append(sshFlags, sp.Args("docker", "system", "dial-stdio")...)...)
			},
			Host: "http://docker.example.com",
		}, nil
```


⚠️ ☝️ what happens if someone uses `GetConnectionHelperWithSSHOpts()`, and sets `-o ConnectTimeout=30` as custom option?

```go
GetConnectionHelperWithSSHOpts("ssh://foo", []string{"-o ConnectTimeout=60"})
```

In that case (with this PR), `ssh` will be called with _both_ `-o ConnectTimeout=30` and `-o ConnectTimeout=60`; does that produce an error?

thaJeztah: > In that case (with this PR), `ssh` will be called with both `-o ConnectTimeout=30` and `-o ConnectTimeout=60`; does that produce an error?


Looks like it doesn't complain (not sure which timeout is used though? first? last?)

```bash
ssh -o ConnectTimeout=30 -o ConnectTimeout=60 swarm-test-01
...
```

💁 If we decide to apply it to both, it may be worth mentioning the default timeout in the function's descriptions.

cli/connhelper/commandconn/commandconn.go
```diff
@@ -64,89 +65,78 @@ func New(_ context.Context, cmd string, args ...string) (net.Conn, error) {
...
-	stderrMu      sync.Mutex
-	stderr        bytes.Buffer
-	stdioClosedMu sync.Mutex // for stdinClosed and stdoutClosed
-	stdinClosed   bool
-	stdoutClosed  bool
-	localAddr     net.Addr
-	remoteAddr    net.Addr
+	cmd          *exec.Cmd
+	cmdWaitErr   error
+	cmdMutex     sync.Mutex // for cmd, cmdWaitErr
```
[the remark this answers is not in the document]
laurazard: ah, thanks!

cli/connhelper/commandconn/commandconn.go
```diff
@@ -64,89 +65,78 @@ func New(_ context.Context, cmd string, args ...string) (net.Conn, error) {
...
-		return nil
-	}
-	wExitErr, ok := werr.(*exec.ExitError)
-	if ok {
-		if wExitErr.ProcessState.Exited() {
-			return nil
-		}
-	}
-	return errors.Wrapf(werr, "commandconn: failed to wait")
+	c.cmdExited.Store(true)
```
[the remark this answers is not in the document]
laurazard: yup! I thought about it during that change, and ended up leaving it as it was, but definitely happy to make that change.

cli/connhelper/commandconn/commandconn.go
laurazard: That code isn't from the last commit – actually comparing to the pre-this-PR code, we were never handling `os.ErrClosed()` (although we should have at the time, since this is mostly cause by us closing the connection ourselves).
It should be safe to discard errors coming from the pipe when the connection is closing, since we've already called `Close()` if there is any issue terminating the command they will be picked up there.

cli/connhelper/commandconn/commandconn.go
laurazard: I answered above (https://github.com/docker/cli/pull/4226#discussion_r1182794147)

cli/connhelper/commandconn/commandconn.go
laurazard: As an aside, this discussion really makes me think this whole code could use a big refactor. I started one while I was digging around and got a fair bit into it, but then realized I had already wasted too much time on it and went back to my previous branch and just applied these few changes. Maybe it's worth it though, I'm definitely keeping this in the back pocket for when I'm a bit bored (hah)

cli/connhelper/commandconn/commandconn.go
laurazard: Ahhhh yeah, there's a flow here where we don't actually manage to terminate the command and yet return nil – I'll write a test for that, and change up the code.
If the 1st call fails and doesn't close `stdOut`, the 2nd call can work fine, close `stdIn`, and not call `kill()`, and then it doesn't return an error but we effectively haven't killed the connection.
This is probably fine 99% of the time (or more), but we can make it better for sure.

cli/connhelper/connhelper.go
laurazard:
```go
sshFlags = append([]string{"-o ConnectTimeout=30"}, sshFlags...)
		return &ConnectionHelper{
				return commandconn.New(ctx, "ssh", append(sshFlags, sp.Args("docker", "system", "dial-stdio")...)...)
			},
			Host: "http://docker.example.com",
		}, nil
```
Fun fact, if you do it this way and then a connection gets dialed 3 times you end with
```console
$ ssh -o ConnectTimeout=30 -o ConnectTimeout=30 -o ConnectTimeout=30 ...
```
(ask me how I know)

In reality though, I iterated through a few different things while making this change (a `getSSHTimeoutArgs` func that also adds an env var for users to be able to disable/adjust it, checking if the command passed in already includes a `ConnectTimeout` option and only adding it if it wasn't there already, etc), but ended up just using this since it didn't seem like this was getting called by anyone else.

I should've considered that other people may be calling it though! I'll take a whack at it again, maybe go with the simple implementation and only add it if it wasn't already specified.

cli/connhelper/commandconn/commandconn_unix_test.go
```diff
@@ -43,3 +46,134 @@ func TestEOFWithoutError(t *testing.T) {
...
+	err = cmdConn.Close()
+	assert.NilError(t, err)
+	assert.Check(t, !checkProcess(cmdConn.cmd.Process))
+}
+
+// check whether process is running by sending
+// signal 0 and checking if an error is returned
+func checkProcess(process *os.Process) bool {
+	return process.Signal(syscall.Signal(0)) == nil
+}
```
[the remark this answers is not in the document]
thaJeztah: Yeah, definitely not a blocker, but I saw the code, and thought; hey, I worked on some code that handled exactly this. And yes, we were already importing it so 🤷‍♂️

cli/connhelper/commandconn/commandconn.go
thaJeztah: Yes; this was all just from looking at it; I haven't considered what the consequences are for the incorrect handling (and perhaps it's okay for a follow up). 😅

cli/connhelper/commandconn/commandconn.go
```diff
@@ -64,98 +65,79 @@ func New(_ context.Context, cmd string, args ...string) (net.Conn, error) {
...
+// onEOF gets called if we receive an io.EOF while reading
+// or writing from the undelying command pipes.
+//
+// When we've received an EOF we expect that the command will
+// be terminated soon. As such, we call Wait() on the command
+// and return EOF or the error depending on whether the command
+// exited with an error.
+//
+// If Wait() does not return within 10s, an error is returned
 func (c *commandConn) onEOF(eof error) error {
```
thaJeztah, line 119: Considering if it would make for a more easy-to-consume interface to have this function check the error that's passed, and handle it if it was an EOF, and otherwise return as-is.

```suggestion
func (c *commandConn) handleEOF(err error) error {
	if err != io.EOF {
		return err
	}
```

Then callers can use it as;

```go
return n, c.handleEOF(err)
```

thaJeztah (commented):
Looking good (thanks!); could use some additional set of eyes on this in case I missed things.

I left some comments/suggestions inline; let me open a PR with those suggestions (feel free to ignore/update/modify them if you want)

thaJeztah:
opened https://github.com/laurazard/cli/pull/1 (please squash with your commits, and (as mentioned) feel free to update/ignore/whatever).

cli/connhelper/commandconn/commandconn.go
```diff
@@ -64,98 +65,79 @@ func New(_ context.Context, cmd string, args ...string) (net.Conn, error) {
...
 }
 
-// killIfStdioClosed kills the cmd if both stdin and stdout are closed.
-func (c *commandConn) killIfStdioClosed() error {
-	c.stdioClosedMu.Lock()
-	stdioClosed := c.stdoutClosed && c.stdinClosed
-	c.stdioClosedMu.Unlock()
-	if !stdioClosed {
+// kill returns nil if the command terminated, regardless to the exit status.
+func (c *commandConn) kill() error {
```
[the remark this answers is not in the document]
laurazard: On UNIXy systems, we can signal the process (with a `SIGTERM` in this case) and wait for a few seconds to see if it gracefully exits, which we can't do on Windows (since, signals on windows aren't quite a thing as such). From [`os/exec_windows.go`](https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/os/exec_windows.go) it looks like the only "signal" that's handled for processes on Windows is `SIGKILL`:
```go
func (p *Process) signal(sig Signal) error {
	handle := atomic.LoadUintptr(&p.handle)
	if handle == uintptr(syscall.InvalidHandle) {
		return syscall.EINVAL
	}
	if p.done() {
		return ErrProcessDone
	}
	if sig == Kill {
		var terminationHandle syscall.Handle
		e := syscall.DuplicateHandle(^syscall.Handle(0), syscall.Handle(handle), ^syscall.Handle(0), &terminationHandle, syscall.PROCESS_TERMINATE, false, 0)
		if e != nil {
			return NewSyscallError("DuplicateHandle", e)
		}
		runtime.KeepAlive(p)
		defer syscall.CloseHandle(terminationHandle)
		e = syscall.TerminateProcess(syscall.Handle(terminationHandle), 1)
		return NewSyscallError("TerminateProcess", e)
	}
	// TODO(rsc): Handle Interrupt too?
	return syscall.Errno(syscall.EWINDOWS)
}
```
which is doing the same thing we do in the function when we call `c.cmd.Process.Kill()`

laurazard:
@thaJeztah I merged your changes (thank you!)

thaJeztah (approved):
LGTM

thaJeztah:
@rumpl @AkihiroSuda ptal 🤗

vvoland (approved):

cli/connhelper/commandconn/commandconn_unix_test.go
```diff
@@ -42,3 +44,169 @@ func TestEOFWithoutError(t *testing.T) {
...
+			assert.Check(t, is.Equal(0, n))
+		}
+	}()
+
+	err = cmdConn.Close()
+	assert.NilError(t, err)
+	assert.Check(t, !process.Alive(cmdConn.cmd.Process.Pid))
+
+	readErr := <-readErrC
+	assert.ErrorContains(t, readErr, "EOF")
```
AkihiroSuda, line 211: Can't we use `errors.Is(readErr, io.EOF)`?

cli/connhelper/commandconn/commandconn.go
```diff
@@ -64,100 +65,86 @@ func New(_ context.Context, cmd string, args ...string) (net.Conn, error) {
...
-	stderr        bytes.Buffer
-	stdioClosedMu sync.Mutex // for stdinClosed and stdoutClosed
-	stdinClosed   bool
-	stdoutClosed  bool
-	localAddr     net.Addr
-	remoteAddr    net.Addr
+	cmdMutex     sync.Mutex // for cmd, cmdWaitErr
+	cmd          *exec.Cmd
+	cmdWaitErr   error
+	cmdExited    atomic.Bool
```
tianon, line 71: Similar to https://github.com/moby/moby/pull/45966#discussion_r1334862901, this means we should update our `go 1.18` line in `vendor.mod` to at least 1.19 :sweat_smile: 

```console
# github.com/docker/cli/cli/connhelper/commandconn
cli/connhelper/commandconn/commandconn.go:71:22: undefined: atomic.Bool
cli/connhelper/commandconn/commandconn.go:76:22: undefined: atomic.Bool
cli/connhelper/commandconn/commandconn.go:77:22: undefined: atomic.Bool
cli/connhelper/commandconn/commandconn.go:78:22: undefined: atomic.Bool
```

thaJeztah: Hello @tianon, are you interested in contributing to open source and opening a PR? 😉

tianon: Naw, open source isn't really my thing :joy:

(totally fair callout, I'll see what I can do)

PR #4228, opened 2023-04-26
docs: include required tools in source tree

### LINKED ISSUE #4212: Documentation generation without internet connectivity
### Description

When trying to build the `docker` cli man pages, the scripts for building the man pages (since v23.0.0) require an internet connection because the handful of packages needed to generate the man pages are not vendored and thus need to be pulled at build time. The same applies for the yaml and markdown docs.

This is not possible in the build environment of some distributions (at the very least, openSUSE doesn't permit internet access to build machines but I suspect this is true for a fair number of distributions). The net result is that we cannot release any version of Docker >=23.0.0 at the moment because we cannot package the man pages for Docker.

I've spent a while trying to figure out a way to work around this in various ways, but it seems the simplest solution would be to back to vendoring the packages. This change was added in #3381, but it seems like the purpose of that PR was to simply add a way of building the documentation during CI to validate the build works. The switch to requiring internet access for the build doesn't seem necessary to me and there was no discussion in the PR about this change (indicating that the downstream impact of this change wasn't noticed at the time).

Would you accept a PR that switches back to vendoring the packages needed for documentation generation, removing the need for internet access?

cyphar:
In order to be able to build the documentation without internet access
(as is required by some distribution build systems), all of the source
code needed for the build needs to be available in the source tarball.

This used to be possible with the docker-cli sources but was
accidentally broken with some CI changes that switched to downloading
the tools (by modifying go.mod as part of the docs build script).

This pattern also maked documentation builds less reproducible since the
tool version used was not based on the source code version.

Fixes: commit 7dc35c0 ("validate manpages target")
Fixes: commit a650f4d ("switch to cli-docs-tool for yaml docs generation")
Fixes #4212
See #3381


cyphar:
I'm confused why the doc building validation is failing with Go compilation errors:

```
go: cannot find main module, but -modfile was set.
	-modfile cannot be used to set the module root directory.
```

On my machine, with this patch both `make manpages` and `make -f docker.Makefile manpages` work without issues. Same goes for `yamldocs` and `mddocs`. I'm also using go1.20.3.

thaJeztah:
It needs a temporary `go.mod`, otherwise `-modfile=vendor.mod` won't work.


But we need to look at this change; I know we split these to separate modules to not add the tool's dependencies as dependencies for the main module  / vendor, and because we needed these separate modules for our migration to Hugo for the docs website. ./cc @crazy-max @dvdksn

dvdksn:
I don't think [my change](https://github.com/docker/cli/commit/1e3622c50cc80bb578c59dc242630ee815e678fd) had any impact in this respect - it was only moving the doc generation to a subdirectory so it wouldn't confuse Hugo when fetching content files from this repo.

The `cli-docs-tool` module needs to be an external dep since it's used by other projects as well. I don't suppose there's much harm in vendoring them here -- other than 23k lines of bloat.

cyphar:
> It needs a temporary go.mod, otherwise -modfile=vendor.mod won't work.

Ah. I can add the temporary copying logic for these builds if it's necessary.

> I know we split these to separate modules to not add the tool's dependencies as dependencies for the main module / vendor

Unfortunately that's exactly what I'm trying to revert -- removing this from vendoring means that we cannot build Docker's man pages in openSUSE's OBS (there is no internet access in OBS when building packages).

The other alternative would be to include the generated of the man pages in the source tree (verifying they are kept up-to-date in the CI), so distributions can just copy them without needing to use a tool to build them during packaging.

cyphar:
@thaJeztah I've fixed the `go.mod` issue, it all works now. :smile_cat:

cyphar:
Rebased.

@thaJeztah Did you want me to go the route of making the documentation be pre-built and then verified in CI? I think packaging things this way is better for distributions (it Just Works :tm: with the distributed tarballs) and the bloat here is only to the tarballs not the final binary, but if it is an issue I can go the other route. Rebasing this patchset for every Docker patch release is getting a little tiring. :sweat_smile:

tianon:
Gentle ping @thaJeztah @crazy-max :bow:

(having the script that generates the manpages _also_ be responsible for downloading and building the tools necessary is definitely problematic :sob:)

tianon:
If we want to avoid having the tools in the "main" package dependencies/vendoring, perhaps we could have a separate `tools` directory with a `vendor.mod` that includes these as a separate vendoring tree?  Or within that "manpage generation" code itself (further isolating it in the directory tree if necessary)?

thaJeztah:
Ah, yes, I tried doing a rebase, and some updates in https://github.com/docker/cli/pull/4903 some time ago, but now that the cli-docs-tool also should be able to do the man-page generating (without building an `go-md2man` binary), we may be able to get rid of that part; but I didn't get round to that yet 😓 

I think @crazy-max is on PTO this week, but maybe @krissetto or @Benehiko (also on PTO this week 🙈) are interested to have a peek.

cyphar:
I've been rebasing this for the SUSE packages, I can push a new version here if you like.

cyphar:
I've rebased it again (this bump was a bit more complicated than previous ones, fwiw). I didn't remove the build of the `go-md2man` binary, but I can take a look at that if that's what's needed to get this merged so I can stop rebasing it :crying_cat_face:

Benehiko (approved):
To me this looks fine, I tested it locally on my side and seems all good. I think someone else should also just verify if the changes won't break anything.

scripts/docs/generate-man.sh
```diff
@@ -4,32 +4,18 @@ set -eu
 
 : "${MD2MAN_VERSION=v2.0.5}"
 
-export GO111MODULE=auto
-
-function clean {
-  rm -rf "$buildir"
+function clean() {
+	rm -f go.mod
```
tianon, line 8: Shouldn't these scripts only delete `go.mod` if it didn't exist when this script started?
(ie, only if this script running `./scripts/vendor init` is what created it)

thaJeztah: Good point; technically, there "shouldn't" be one, but I know I tend to have one present (as a symlink); let me try how it handles that.

thaJeztah:
Hm.. so @tianon's comment made me try if this works well with an existing `go.mod`, and it looks like it has some issues there.

:warning: this is not the "standard" situation, but while we're working on (finally!) getting  our repos to  have an actual `go.mod`, a trick I (and others use) is to create symlinks; 

```bash
ls -la go.*
lrwxr-xr-x 1 root root 10 Jan 10 21:44 go.mod -> vendor.mod
lrwxr-xr-x 1 root root 10 Jan 10 21:44 go.sum -> vendor.sum
```

With the above symlinks present, I ran `make mdddocs` (tried both inside the `make dev` dev-container and on my host), and it looks like it doesn't handle that gracefully;

```bash
make mddocs
scripts/docs/generate-md.sh
go: inconsistent vendoring in /go/src/github.com/docker/cli:
	dario.cat/mergo@v1.0.1: is marked as explicit in vendor/modules.txt, but not explicitly required in go.mod
	github.com/Azure/go-ansiterm@v0.0.0-20230124172434-306776ec8161: is marked as explicit in vendor/modules.txt, but not explicitly required in go.mod
...
```

After the above, I had modifications in my git, and it looks like it overwrote `vendor.mod`, emptying it (which would explain the error above);

```bash
git status
On branch docs-no-internet
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   vendor.mod
```

```patch
diff --git a/vendor.mod b/vendor.mod
index 3badee9b8..a8ea3c017 100644
--- a/vendor.mod
+++ b/vendor.mod
@@ -1,109 +1,3 @@
 module github.com/docker/cli

-// 'vendor.mod' enables use of 'go mod vendor' to managed 'vendor/' directory.
-// There is no 'go.mod' file, as that would imply opting in for all the rules
-// around SemVer, which this repo cannot abide by as it uses CalVer.
-
 go 1.22.0
-
-require (
-	dario.cat/mergo v1.0.1
-	github.com/containerd/platforms v0.2.1
-	github.com/creack/pty v1.1.21
-	github.com/distribution/reference v0.6.0
-	github.com/docker/cli-docs-tool v0.8.0
```

It also removed the `go.mod` symlink, but kept the `go.sum` one;

```bash
ls -la go.*
lrwxr-xr-x 1 root root 10 Jan 10 21:44 go.sum -> vendor.sum
```

Curious if we could use something like the script that's in moby/moby, which seemed to detect the "already have a `go.mod`" situation; it will produce a warning ("you shouldn't do this!" - I know), but otherwise handle it gracefully;
https://github.com/moby/moby/blob/d34a5f5d72ace96c2fbd2a7484ee4a9ddfa36e42/hack/with-go-mod.sh

thaJeztah:
Thanks for updating! ❤️ 

Overall, this looks good; I tried it in a "standard" situation (no `go.mod` / `go.mod` symlinks), and it all works well. It would be good if it could handle the "existing go.mod" either gracefully (like moby; warn, but continue), or hard failure ("can't have a `go.mod` present).

PR #4258, opened 2023-05-05
Add support for --detach flag in stack deploy

gmargaritis:
Part of #373 along with https://github.com/docker/cli/pull/4259

Added `--detach` and `--quiet` / `-q` flags to `stack deploy`. Setting `--detach=false` waits until all of the services have converged. Shows progress bars for each individual task, unless  `--quiet` / `-q` is specified.

I have used the `WaitOnService` under the hood, so that the behavior is in line with `service update` or `service create` while using the `--detach` flag.

PR #4259, opened 2023-05-05
Add support for --detach flag in stack rm

gmargaritis:
Part of #373 along with https://github.com/docker/cli/pull/4258

Added --detach to stack rm. Setting --detach=false waits until all of the stack tasks have reached a terminal state.

### DIFF
--- a/cli/command/stack/remove.go
+++ b/cli/command/stack/remove.go
@@ -27,5 +27,8 @@ func newRemoveCommand(dockerCli command.Cli) *cobra.Command {
 			return completeNames(dockerCli)(cmd, args, toComplete)
 		},
 	}
+
+	flags := cmd.Flags()
+	flags.BoolVarP(&opts.Detach, "detach", "d", true, "Do not wait for stack removal")
 	return cmd
 }
--- a/cli/command/stack/swarm/remove.go
+++ b/cli/command/stack/swarm/remove.go
@@ -11,6 +11,7 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/swarm"
 	"github.com/docker/docker/api/types/versions"
+	apiclient "github.com/docker/docker/client"
 	"github.com/pkg/errors"
 )
 
@@ -58,6 +59,14 @@ func RunRemove(ctx context.Context, dockerCli command.Cli, opts options.Remove) 
 
 		if hasError {
 			errs = append(errs, fmt.Sprintf("Failed to remove some resources from stack: %s", namespace))
+			continue
+		}
+
+		if !opts.Detach {
+			err = waitOnTasks(ctx, client, namespace)
+			if err != nil {
+				errs = append(errs, fmt.Sprintf("Failed to wait on tasks of stack: %s: %s", namespace, err))
+			}
 		}
 	}
 
@@ -136,4 +145,46 @@ func removeConfigs(
 		}
 	}
 	return hasError
+}
+
+var numberedStates = map[swarm.TaskState]int64{
+	swarm.TaskStateNew:       1,
+	swarm.TaskStateAllocated: 2,
+	swarm.TaskStatePending:   3,
+	swarm.TaskStateAssigned:  4,
+	swarm.TaskStateAccepted:  5,
+	swarm.TaskStatePreparing: 6,
+	swarm.TaskStateReady:     7,
+	swarm.TaskStateStarting:  8,
+	swarm.TaskStateRunning:   9,
+	swarm.TaskStateComplete:  10,
+	swarm.TaskStateShutdown:  11,
+	swarm.TaskStateFailed:    12,
+	swarm.TaskStateRejected:  13,
+}
+
+func terminalState(state swarm.TaskState) bool {
+	return numberedStates[state] > numberedStates[swarm.TaskStateRunning]
+}
+
+func waitOnTasks(ctx context.Context, client apiclient.APIClient, namespace string) error {
+	terminalStatesReached := 0
+	for {
+		tasks, err := getStackTasks(ctx, client, namespace)
+		if err != nil {
+			return fmt.Errorf("failed to get tasks: %w", err)
+		}
+
+		for _, task := range tasks {
+			if terminalState(task.Status.State) {
+				terminalStatesReached++
+				break
+			}
+		}
+
+		if terminalStatesReached == len(tasks) {
+			break
+		}
+	}
+	return nil
 }
--- a/docs/reference/commandline/stack_rm.md
+++ b/docs/reference/commandline/stack_rm.md
@@ -7,6 +7,12 @@ Remove one or more stacks
 
 `docker stack rm`, `docker stack remove`, `docker stack down`
 
+### Options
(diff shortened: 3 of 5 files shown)


neersighted:
PTAL @corhere @dperny

cli/command/stack/swarm/remove.go
```diff
@@ -138,3 +146,45 @@ func removeConfigs(
...
+	swarm.TaskStateAccepted:  5,
+	swarm.TaskStatePreparing: 6,
+	swarm.TaskStateReady:     7,
+	swarm.TaskStateStarting:  8,
+	swarm.TaskStateRunning:   9,
+	swarm.TaskStateComplete:  10,
+	swarm.TaskStateShutdown:  11,
+	swarm.TaskStateFailed:    12,
+	swarm.TaskStateRejected:  13,
+}
```
neersighted, line 164: We should rely on importing SwarmKit here instead of incorporating magic enums.

Also it's unclear if this logic even belongs on the client; this looks dangerously close to leaking concerns across layers.

dperny: The reason Swarm states are numbered like they are is so that intermediate states can be inserted at any time without disrupting the ordering of states committed to the object store already. If the code handling states isn't persistent, then there's no such concern. If we ever added an intermediate state, we can just renumber all of these harmlessly.

dperny:
Looks fine to me.

gmargaritis:
Hey @thaJeztah and @neersighted

Is there anything else I can do in order to move this forward?

cli/command/stack/remove.go
```diff
@@ -27,5 +27,8 @@ func newRemoveCommand(dockerCli command.Cli) *cobra.Command {
 			return completeNames(dockerCli)(cmd, args, toComplete)
 		},
 	}
+
+	flags := cmd.Flags()
+	flags.BoolVarP(&opts.Detach, "detach", "d", true, "Exit immediately instead of waiting for the stack tasks to converge")
```
cpuguy83, line 32:
```suggestion
	flags.BoolVarP(&opts.Detach, "detach", "d", true, "Do not wait for stack removal")
```

cpuguy83 (changes requested):

cpuguy83:
It looks like this is changing the default behavior of `stack rm` to wait for removal?

cpuguy83:
Oh nevermind I see the default for the flag is `true`.
This makes usage a little weird because you have to `--detach=false`.

🤔

gmargaritis:
@cpuguy83 Thank you for your feedback! 

I see where you're coming from regarding the `--detach` flag behavior. We set `--detach` to `true` by default, in order to keep the user experience of the command the same. This approach was actually based on the implementation of the introduction of the `--detach` flag in `service create` and `service update`. In this case, the default was `--detach=true` and in a later release it was decided to change the default behavior of the commands to `--detach=false`.

If the plan is to switch the default behavior of `stack rm` to non-detachable in a future release, we can add a message indicating so, just like we did in https://github.com/moby/moby/pull/31144#issuecomment-289600844 and in https://github.com/docker/cli/pull/4258#discussion_r1410353240.

I hope this clears things up and takes us closer to also merging https://github.com/docker/cli/pull/4258 and closing https://github.com/docker/cli/issues/373 😄

/cc @thaJeztah

docs/reference/commandline/stack_rm.md
```diff
@@ -7,6 +7,12 @@ Remove one or more stacks
 
 `docker stack rm`, `docker stack remove`, `docker stack down`
 
+### Options
+
+| Name             | Type | Default | Description                   |
+|:-----------------|:-----|:--------|:------------------------------|
+| `-d`, `--detach` |      |         | Do not wait for stack removal |
```
thaJeztah, line 14: Interesting; for some reason it doesn't show the default here 🤔 


`--help` shows the correct output, but the markdown generator doesn't 🤔 Any ideas @crazy-max ?

```bash
./build/docker stack rm --help

Usage:  docker stack rm [OPTIONS] STACK [STACK...]

Remove one or more stacks

Aliases:
  docker stack rm, docker stack remove, docker stack down

Options:
  -d, --detach   Do not wait for stack removal (default true)
```

crazy-max: I think it makes sense to not show default as this is a bool flag anyway? (like `--pull` or `--push`)

crazy-max: Oh my bad this is true by default so opt-out flag right?

gmargaritis: @crazy-max Yes, it's true by default and the user can use `--detach=false` if they want to wait for stack removal to complete

gmargaritis: It's the same for `service create/update` where we already have implemented this flag in https://github.com/moby/moby/pull/31144

https://github.com/docker/cli/blob/79fa65e7b58b081dfc2316a1a3c0f602f4d4c01a/docs/reference/commandline/service_create.md?plain=1#L6-L16

cpuguy83: Crazy thought, how about "--wait"?

thaJeztah: So the intent here is to (like `docker service create`) flip the default at some point, so make `stack deploy` act the same as `docker compose up`.

I don't think the markdown here is a real blocker for the PR (we can fix in "post"), but I noticed it, and was curious what went wrong here (looks like a bug in the markdown generator potentially?)

gmargaritis: I agree, we can investigate further in a separate issue, as it seems like a generalized bug in the markdown generator.

Is there anything else left to be done in this PR?

gmargaritis:
@thaJeztah @crazy-max 👋

Is there anything left to be done in this one?

neersighted (approved):
Overall LGTM/agreed that we can worry about the markdown later, given this is no more broken then other commands.

gmargaritis:
@neersighted We have fixed the markdown issue in https://github.com/docker/cli-docs-tool/pull/48 ✌️

thaJeztah:
Did a quick rebase and re-generated the markdown as we updated the `cli-docs-tool` with a fix for the boolean flags;


```patch
diff --git a/docs/reference/commandline/stack_rm.md b/docs/reference/commandline/stack_rm.md
index 8dd872b7c..5cfbfaab1 100644
--- a/docs/reference/commandline/stack_rm.md
+++ b/docs/reference/commandline/stack_rm.md
@@ -9,9 +9,9 @@ Remove one or more stacks

 ### Options

-| Name             | Type | Default | Description                   |
-|:-----------------|:-----|:--------|:------------------------------|
-| `-d`, `--detach` |      |         | Do not wait for stack removal |
+| Name             | Type   | Default | Description                   |
+|:-----------------|:-------|:--------|:------------------------------|
+| `-d`, `--detach` | `bool` | `true`  | Do not wait for stack removal |
```

I also opened a tracking ticket to look at follow-up improvements discussed above; https://github.com/docker/cli/issues/4907

PR #4260, opened 2023-05-05
docs: add description and examples for docker -H

dvdksn:
Signed-off-by: David Karlsson <david.karlsson@docker.com>

**- What I did**

Added a description for the `docker -H` flag. Since we don't really have a better place to put this atm, I used the `cli.md` file.

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/flags/options.go
+++ b/cli/flags/options.go
@@ -83,7 +83,7 @@ func (o *ClientOptions) InstallFlags(flags *pflag.FlagSet) {
 
 	// opts.ValidateHost is not used here, so as to allow connection helpers
 	hostOpt := opts.NewNamedListOptsRef("hosts", &o.Hosts, nil)
-	flags.VarP(hostOpt, "host", "H", "Daemon socket(s) to connect to")
+	flags.VarP(hostOpt, "host", "H", "Daemon socket to connect to")
 	flags.StringVarP(&o.Context, "context", "c", "",
 		`Name of the context to use to connect to the daemon (overrides `+client.EnvOverrideHost+` env var and default context set with "docker context use")`)
 }
--- a/docs/reference/commandline/cli.md
+++ b/docs/reference/commandline/cli.md
@@ -92,18 +92,18 @@ The base command for the Docker CLI.
 
 ### Options
 
-| Name                | Type     | Default                  | Description                                                                                                                           |
-|:--------------------|:---------|:-------------------------|:--------------------------------------------------------------------------------------------------------------------------------------|
-| `--config`          | `string` | `/root/.docker`          | Location of client config files                                                                                                       |
-| `-c`, `--context`   | `string` |                          | Name of the context to use to connect to the daemon (overrides DOCKER_HOST env var and default context set with `docker context use`) |
-| `-D`, `--debug`     |          |                          | Enable debug mode                                                                                                                     |
-| `-H`, `--host`      | `list`   |                          | Daemon socket(s) to connect to                                                                                                        |
-| `-l`, `--log-level` | `string` | `info`                   | Set the logging level (`debug`, `info`, `warn`, `error`, `fatal`)                                                                     |
-| `--tls`             |          |                          | Use TLS; implied by --tlsverify                                                                                                       |
-| `--tlscacert`       | `string` | `/root/.docker/ca.pem`   | Trust certs signed only by this CA                                                                                                    |
-| `--tlscert`         | `string` | `/root/.docker/cert.pem` | Path to TLS certificate file                                                                                                          |
-| `--tlskey`          | `string` | `/root/.docker/key.pem`  | Path to TLS key file                                                                                                                  |
-| `--tlsverify`       |          |                          | Use TLS and verify the remote                                                                                                         |
+| Name                             | Type     | Default                  | Description                                                                                                                           |
+|:---------------------------------|:---------|:-------------------------|:--------------------------------------------------------------------------------------------------------------------------------------|
+| `--config`                       | `string` | `/root/.docker`          | Location of client config files                                                                                                       |
+| `-c`, `--context`                | `string` |                          | Name of the context to use to connect to the daemon (overrides DOCKER_HOST env var and default context set with `docker context use`) |
+| `-D`, `--debug`                  |          |                          | Enable debug mode                                                                                                                     |
+| [`-H`](#host), [`--host`](#host) | `list`   |                          | Daemon socket to connect to                                                                                                           |
+| `-l`, `--log-level`              | `string` | `info`                   | Set the logging level (`debug`, `info`, `warn`, `error`, `fatal`)                                                                     |
+| `--tls`                          |          |                          | Use TLS; implied by --tlsverify                                                                                                       |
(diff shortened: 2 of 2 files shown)


docs/reference/commandline/cli.md
```diff
@@ -378,6 +378,37 @@ list of root Certificate Authorities.
...
+
+- `unix://`
+- `tcp://`
+- `ssh://`
+
+The following example shows how to invoke `docker ps` over TCP, to a remote
+daemon with IP address `174.17.0.1`, listening on port `3131`:
+
+```console
+$ docker -H tcp://174.17.0.1:3131 ps
```
thaJeztah, line 394: perhaps use `:2376` here to encourage using the secure connection.

We have 3 ports registered; https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=docker

- `2375`: tcp, non-TLS (insecure)
- `2376`: tcp (TLS)
- `2377`: used for swarm (secure)

So only the first 2 are relevant for this one, but perhaps we need to mention them somewhere

docs/reference/commandline/cli.md
```diff
@@ -378,6 +378,37 @@ list of root Certificate Authorities.
...
 ## Examples
 
+### <a name="host"></a> Specify daemon host (-H, --host)
+
+You can use the `-H`, `--host` flag to specify one or more sockets to use when
+you invoke a `docker` command. You can use the following protocols:
+
+- `unix://`
+- `tcp://`
+- `ssh://`
```
thaJeztah, line 388: There's also `npipe://` on Windows. Perhaps we should use a table for this, something like;


```markdown
| Scheme                                 | Description               | Example                          |
|----------------------------------------|---------------------------|----------------------------------|
| `unix://[<path>]`                      | Unix socket (Linux only)  | `unix:///var/run/docker.sock`    |
| `tcp://[<IP or host>[:port]]`          | TCP connection            | `tcp://174.17.0.1:2376`          |
| `ssh://[username@]<IP or host>[:port]` | SSH connection            | `ssh://user@192.168.64.5`        |
| `npipe://[<name>]`                     | Named pipe (Windows only) | `npipe:////./pipe/docker_engine` |
```

| Scheme                                 | Description               | Example                          |
|----------------------------------------|---------------------------|----------------------------------|
| `unix://[<path>]`                      | Unix socket (Linux only)  | `unix:///var/run/docker.sock`    |
| `tcp://[<IP or host>[:port]]`          | TCP connection            | `tcp://174.17.0.1:2376`          |
| `ssh://[username@]<IP or host>[:port]` | SSH connection            | `ssh://user@192.168.64.5`        |
| `npipe://[<name>]`                     | Named pipe (Windows only) | `npipe:////./pipe/docker_engine` |


Perhaps we also need to mention what the defaults are, and that only the scheme is required when setting it as option;

- if nothing is specified, the default is `unix:///var/run/docker.sock` on Linux and macOS, and `npipe:////./pipe/docker_engine` on Windows
- `-H tcp://` uses default `tcp://localhost:2375` (or `tcp://localhost:2376` if TLS options are set)
- `-H unix://`  uses default `unix:///var/run/docker.sock`
- `-H npipe://` uses default  `npipe:////./pipe/docker_engine`

docs/reference/commandline/cli.md
```diff
@@ -378,6 +378,37 @@ list of root Certificate Authorities.
...
+daemon with IP address `174.17.0.1`, listening on port `3131`:
+
+```console
+$ docker -H tcp://174.17.0.1:3131 ps
+```
+
+When you use SSH invoke a command on a remote daemon, the request gets forwarded
+to the `/var/run/docker.sock` Unix socket on the SSH host.
+
+```
```
thaJeztah, line 400: missing `console` code-hint here 😅

docs/reference/commandline/cli.md
```diff
@@ -378,6 +378,37 @@ list of root Certificate Authorities.
...
+$ docker -H ssh://user@192.168.64.5 ps
+```
+
+To achieve a similar effect without having to specify the `-H` flag for every
+command, could also [create a context](context_create.md),
+or alternatively, use the
+[`DOCKER_HOST` environment variable](#environment-variables).
```
thaJeztah, line 407: missing "you" before `could also`.

Perhaps aa more active voice would work;

```suggestion
You can [create a context](context_create.md), or alternatively, use the
[`DOCKER_HOST` environment variable](#environment-variables) to achieve
a similar effect without having to specify the `-H` flag for every command.
```

docs/reference/commandline/cli.md
```diff
@@ -378,6 +378,37 @@ list of root Certificate Authorities.
...
+
+For more information about the `-H` flag, see
+[Daemon socket option](dockerd.md#daemon-socket-option)
```
thaJeztah, line 410:
```suggestion
[Daemon socket option](dockerd.md#daemon-socket-option).
```

docs/reference/commandline/cli.md
```diff
@@ -378,6 +378,37 @@ list of root Certificate Authorities.
 
 ## Examples
 
+### <a name="host"></a> Specify daemon host (-H, --host)
+
+You can use the `-H`, `--host` flag to specify one or more sockets to use when
```
thaJeztah, line 383: I don't think "one or more" works here (for the CLI); only works on the daemon (to make it _listen_ on multiple 😅)

So.. hm.. good find actually; looks like we never updated the flag description (I think it was that way because previously that flag was shared between the daemon and the cli);

```
```

thaJeztah (commented):
Thanks! Left some suggestions / comments

thaJeztah:
- Also a heads-up that https://github.com/docker/cli/pull/4073 was merged (which will be in v24.0.0, so probably best to handle in a separate PR after this one)

docs/reference/commandline/cli.md
dvdksn: I added a small note about it here, but we might want to add a similar description for this somewhere else too. Let's take it in a followup. LMK if the note I added seems OK to you.

docs/reference/commandline/cli.md
dvdksn: Good idea, I added this

docs/reference/commandline/cli.md
dvdksn: I updated the CLI help as well, removing the plural

cli/flags/options.go
```diff
@@ -83,7 +83,7 @@ func (o *ClientOptions) InstallFlags(flags *pflag.FlagSet) {
 
 	// opts.ValidateHost is not used here, so as to allow connection helpers
 	hostOpt := opts.NewNamedListOptsRef("hosts", &o.Hosts, nil)
-	flags.VarP(hostOpt, "host", "H", "Daemon socket(s) to connect to")
+	flags.VarP(hostOpt, "host", "H", "Daemon socket to connect to")
```
thaJeztah, line 86: For a follow up we need to look at better wording for this, as this is not only for "sockets".

Okay for now, because we use this terminology in other places, so something to look at in a wider scope

thaJeztah (approved):
LGTM, thanks!

Issue #4265, opened 2023-05-06
Nested compose variable interpolation not working


gilbsgilbs:
### Description

Interpolation of nested variables in compose files is not supported/bugged.

### Reproduce

1. take the following compose file:
```yaml
services:
  my-service:
    image: "alpine"
    command: ["echo", "${HELLO:-${UNUSED}}"]
```
2. Init swarm `docker swarm init`
3. Deploy a stack with this compose file: `HELLO=hello docker stack deploy --compose-file docker-compose.yml foo`
4. Show the logs of the service: `docker service logs foo_my-service`

The logs print `hello}`, meaning that the variable wasn't properly interpolated.

### Expected behavior

Either interpolate to `hello`, just like Compose does, or at least output an error such as `nested variables interpolation is not supported`. Currently, the interpolation is incorrect which is confusing and surprising to the user.

### docker version

```bash
Client:
 Version:           23.0.5
 API version:       1.42
 Go version:        go1.20.4
 Git commit:        bc4487a59e
 Built:             Tue May  2 21:53:09 2023
 OS/Arch:           linux/amd64
 Context:           default

Server:
 Engine:
  Version:          23.0.4
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.20.3
  Git commit:       cbce331930
  Built:            Fri Apr 21 22:05:37 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.7.0
  GitCommit:        1fbd70374134b891f97ce19c70b6e50c7b9f4e0d.m
 runc:
  Version:          1.1.7
  GitCommit:
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.10.4
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.17.3
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 33
  Running: 3
  Paused: 0
  Stopped: 30
 Images: 276
 Server Version: 23.0.4
 Storage Driver: btrfs
  Btrfs:
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
[...]

Issue #2527, opened 2020-05-14
Adopt compose-spec/compose-go for compose file support


ndeloof:
https://compose-spec.io/ is now the place to define Compose file format and evolution. Changes like https://github.com/docker/cli/pull/2517 should not happen within this repo anymore without a sibling issue/PR on compose-spec.

http://github.com/compose-spec/compose-go was contributed as a git extraction for the `types` and `loader` packages from docker/cli. To avoid code to diverge, docker/cli should adopt compose-go ASAP

smidge84:
Is there a way we can bump this issue?

It's been open for over a year and a half with no comments, but there's a steadily increasing list of related issues which I think shows desire for this issue to be resolved.

Is there a view of the Docker Cli roadmap and what work is in-scope?

ndeloof:
compose-go is not just a drop-in replacement for legacy compose support in docker/cli, so this change will require some extra testing for backward compatibility.
https://github.com/docker/cli/pull/3257 will provide a temporary workaround

Radiergummi:
So... different Docker components aren't compatible anymore, nobody thought of marking this as a breaking change, lots of users report this, and... nothing happens in more than two years?

I don't even know how it'd be possible to use Swarm secrets without versioning them, which in turn isn't possible without variables in the compose file, which in turn isn't possible with `stack deploy`, which in turn requires the `config` command to return a preprocessed version of the compose file, which in turn requires it to include a `version` or `stack deploy` to follow the spec.

As someone actually paying for Docker and uses Swarm in production, I cannot wrap my head around why this happens, or why I actually continue to bother with this.

ndeloof:
> I don't even know how it'd be possible to use Swarm secrets without versioning them, which in turn isn't possible without variables in the compose file ...

A potential fix is for Swarm maintainers to introduce support for variables in the legacy compose file parser

Radiergummi:
> A potential fix is for Swarm maintainers to introduce support for variables in the legacy compose file parser

Please no. I don't trust them to not introduce a heap of new bugs and inconsistencies while doing so, making stuff even worse.  

Would that include interpolation in curly braces? Trimming rules inherited from bash? Default values? What if the variable is defined, but an empty string? What if it is used in an illegal location? What if it breaks Yaml parsing?

I'd rather they just finally implement the compose spec proper and get this over with. But I guess that won't happen.

Leopere:
So is this going to be permanent?

dazinator:
[...]

PR #4268, opened 2023-05-08
Nested variables interpolation

gilbsgilbs:
**- What I did**

Fix nested variables interpolation which [should be allowed](https://docs.docker.com/compose/compose-file/12-interpolation/) as per compose specification.

**- How I did it**

Use [compose-go](https://github.com/compose-spec/compose-go/) for variables interpolation in compose files instead of the current implementation.

I've also considered copying the code from compose-go as it is very similar to the existing implementation (and was probably initially copied from this project), but I assumed it would be better to use the reference implementation. Feel free to tell me if you don't want to add this new dependency.

**- How to verify it**

Run the test suite or use the repro described in #4265

Fix nested variables interpolation in compose files.

image


gilbsgilbs:
I just rebased this PR as apparently this is still an issue: https://github.com/docker/cli/issues/4265#issuecomment-2579177595 .

cli/compose/interpolation/interpolation.go
```diff
@@ -7,7 +7,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/docker/cli/cli/compose/template"
+	"github.com/compose-spec/compose-go/v2/template"
```
ndeloof, line 10: Using compose-go interpolation while still not using compose-go to parse compose.yaml seems to me a terrible hybrid solution. Why not eventually adopt then compose-spec?
(https://github.com/docker/cli/issues/2527)

PR #4293, opened 2023-05-15
docs/refactor: Update the documentation for --rm flag in run.md 

tumunshaily:
The current description for the --rm flag doesn't talk about removing the anonyms volume deletion. As add the flag is also responsible for deleting the anonyms volume created with the container.

Signed-off-by: tumunshaily <tumunshaily@gmail.com>


thaJeztah:
Thanks for contributing! I'm slightly on the fence whether `Auto-remove` is the right term to start the flag description with. Perhaps "Automatically remove the container and its anonymous volumes upon exit."


This section of the documentation is automatically generated from the source-code, so if we want to update the description, the code should be updated.
https://github.com/docker/cli/blob/adc9d1f9212973c9be96912ecd70d56fe320cb10/cli/command/container/opts.go#L195

We should probably (also) update the section that describes the `--rm` flag;
https://github.com/docker/cli/blob/adc9d1f9212973c9be96912ecd70d56fe320cb10/docs/reference/run.md?plain=1#L653

There's also some other parts that mention the flag description, e.g.;

- https://github.com/docker/cli/blob/adc9d1f9212973c9be96912ecd70d56fe320cb10/contrib/completion/fish/docker.fish#L257
- https://github.com/docker/cli/blob/adc9d1f9212973c9be96912ecd70d56fe320cb10/man/docker-run.1.md?plain=1#L599
 

@dvdksn ptal

dvdksn:
The revision suggested by @thaJeztah SGTM: "Automatically remove [...]"

thaJeztah:
superseded by https://github.com/docker/cli/pull/5084

PR #4316, opened 2023-05-29
mount: add `bind-recursive=<bool|string>` and deprecate `bind-nonrecursive=<bool>`

AkihiroSuda:
For:
- moby/moby#45278

See `opts/mount_test.go:TestMountOptSetBindRecursive()` for the behavior:

- `bind-recursive=true`, `bind-recursive=enabled`: read-only mounts are recursively read-only if Engine >= v25 && kernel >= v5.12
- `bind-recursive=false`, `bind-recursive=disabled`: alias of `bind-nonrecursive=true`. Now this form is preferred over `bind-nonrecursive=true`.
- `bind-recursive=writable`: conforms to Docker v24; read-only mounts are recursively mounted but not recursively read-only
- `bind-recursive=readonly`: force recursively read-only, or raise an error

    
Documentation will be added separately after reaching consensus on the design.

<details><summary>Backup (2023-05-29)</summary>
<p>

https://github.com/AkihiroSuda/cli/commit/6ebf6ba27841a15021f8fc4e1da36eda108b5f80

**- What I did**

Added the CLI and the docs for:
- moby/moby#45278

**- How I did it**

The `-v` options are implemented on the daemon side.
- `ro-non-recursive`
- `ro-force-recursive` (alias: `rro`)

The following "CSV" options for `--mount` are newly added in this CLI-side PR:
- `bind-readonly-nonrecursive` (alias: `bind-ro-nonrecursive`) (Do not confuse this with the exiswting `bind-nonrecursive`)
- `bind-readonly-forcerecursive` (alias: `bind-ro-forcerecursive`)

**- How to verify it**
Run `docker run --mount type=bind,source=/mnt,target=/mnt,bind-ro-forcerecursive,bind-propagation=rprivate`, and make sure the mount is made recursively read-only.

Add CLI and docs for recursively read-only mounts

🐧 

</p>

</details>

### DIFF
--- a/opts/mount.go
+++ b/opts/mount.go
@@ -2,6 +2,7 @@ package opts
 
 import (
 	"encoding/csv"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -10,6 +11,7 @@ import (
 
 	mounttypes "github.com/docker/docker/api/types/mount"
 	"github.com/docker/go-units"
+	"github.com/sirupsen/logrus"
 )
 
 // MountOpt is a Value type for parsing mounts
@@ -112,6 +114,32 @@ func (m *MountOpt) Set(value string) error {
 			if err != nil {
 				return fmt.Errorf("invalid value for %s: %s", key, val)
 			}
+			logrus.Warn("bind-nonrecursive is deprecated, use bind-recursive=disabled instead")
+		case "bind-recursive":
+			valS := val
+			// Allow boolean as an alias to "enabled" or "disabled"
+			if b, err := strconv.ParseBool(valS); err == nil {
+				if b {
+					valS = "enabled"
+				} else {
+					valS = "disabled"
+				}
+			}
+			switch valS {
+			case "enabled": // read-only mounts are recursively read-only if Engine >= v25 && kernel >= v5.12, otherwise writable
+				// NOP
+			case "disabled": // alias of bind-nonrecursive=true
+				bindOptions().NonRecursive = true
+			case "writable": // conforms to the default read-only bind-mount of Docker v24; read-only mounts are recursively mounted but not recursively read-only
+				bindOptions().ReadOnlyNonRecursive = true
+			case "readonly": // force recursively read-only, or raise an error
+				bindOptions().ReadOnlyForceRecursive = true
+				// TODO: implicitly set propagation and error if the user specifies a propagation in a future refactor/UX polish pass
+				// https://github.com/docker/cli/pull/4316#discussion_r1341974730
+			default:
+				return fmt.Errorf("invalid value for %s: %s (must be \"enabled\", \"disabled\", \"writable\", or \"readonly\")",
+					key, val)
+			}
 		case "volume-nocopy":
 			volumeOptions().NoCopy, err = strconv.ParseBool(val)
 			if err != nil {
@@ -159,6 +187,22 @@ func (m *MountOpt) Set(value string) error {
 	}
 	if mount.TmpfsOptions != nil && mount.Type != mounttypes.TypeTmpfs {
 		return fmt.Errorf("cannot mix 'tmpfs-*' options with mount type '%s'", mount.Type)
+	}
+
+	if mount.BindOptions != nil {
+		if mount.BindOptions.ReadOnlyNonRecursive {
+			if !mount.ReadOnly {
+				return errors.New("option 'bind-recursive=writable' requires 'readonly' to be specified in conjunction")
+			}
+		}
+		if mount.BindOptions.ReadOnlyForceRecursive {
+			if !mount.ReadOnly {
+				return errors.New("option 'bind-recursive=readonly' requires 'readonly' to be specified in conjunction")
+			}
+			if mount.BindOptions.Propagation != mounttypes.PropagationRPrivate {
+				return errors.New("option 'bind-recursive=readonly' requires 'bind-propagation=rprivate' to be specified in conjunction")
+			}
+		}
 	}
 
 	m.values = append(m.values, mount)
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -14,13 +14,13 @@ import (
 	"time"
 
 	cdi "github.com/container-orchestrated-devices/container-device-interface/pkg/parser"
+	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/compose/loader"
 	"github.com/docker/cli/opts"
 	"github.com/docker/docker/api/types/container"
 	mounttypes "github.com/docker/docker/api/types/mount"
 	networktypes "github.com/docker/docker/api/types/network"
 	"github.com/docker/docker/api/types/strslice"
-	"github.com/docker/docker/api/types/versions"
 	"github.com/docker/docker/errdefs"
 	"github.com/docker/go-connections/nat"
 	"github.com/pkg/errors"
@@ -1119,8 +1119,8 @@ func validateAttach(val string) (string, error) {
 
 func validateAPIVersion(c *containerConfig, serverAPIVersion string) error {
 	for _, m := range c.HostConfig.Mounts {
-		if m.BindOptions != nil && m.BindOptions.NonRecursive && versions.LessThan(serverAPIVersion, "1.40") {
-			return errors.Errorf("bind-nonrecursive requires API v1.40 or later")
+		if err := command.ValidateMountWithAPIVersion(m, serverAPIVersion); err != nil {
+			return err
 		}
 	}
 	return nil
--- a/cli/command/service/opts.go
+++ b/cli/command/service/opts.go
@@ -8,11 +8,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/opts"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/swarm"
-	"github.com/docker/docker/api/types/versions"
 	"github.com/docker/docker/client"
 	gogotypes "github.com/gogo/protobuf/types"
 	"github.com/google/shlex"
@@ -1043,8 +1043,8 @@ const (
 
 func validateAPIVersion(c swarm.ServiceSpec, serverAPIVersion string) error {
 	for _, m := range c.TaskTemplate.ContainerSpec.Mounts {
-		if m.BindOptions != nil && m.BindOptions.NonRecursive && versions.LessThan(serverAPIVersion, "1.40") {
-			return errors.Errorf("bind-nonrecursive requires API v1.40 or later")
+		if err := command.ValidateMountWithAPIVersion(m, serverAPIVersion); err != nil {
+			return err
 		}
 	}
 	return nil
--- a/cli/command/utils.go
+++ b/cli/command/utils.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/docker/cli/cli/streams"
 	"github.com/docker/docker/api/types/filters"
+	mounttypes "github.com/docker/docker/api/types/mount"
+	"github.com/docker/docker/api/types/versions"
 	"github.com/moby/sys/sequential"
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
@@ -194,4 +196,18 @@ func StringSliceReplaceAt(s, old, new []string, requireIndex int) ([]string, boo
 	out = append(out, new...)
 	out = append(out, s[idx+len(old):]...)
 	return out, true
+}
+
+// ValidateMountWithAPIVersion validates a mount with the server API version.
+func ValidateMountWithAPIVersion(m mounttypes.Mount, serverAPIVersion string) error {
+	if m.BindOptions != nil {
+		if m.BindOptions.NonRecursive && versions.LessThan(serverAPIVersion, "1.40") {
+			return errors.Errorf("bind-recursive=disabled requires API v1.40 or later")
+		}
+		// ReadOnlyNonRecursive can be safely ignored when API < 1.44
+		if m.BindOptions.ReadOnlyForceRecursive && versions.LessThan(serverAPIVersion, "1.44") {
+			return errors.Errorf("bind-recursive=readonly requires API v1.44 or later")
+		}
+	}
+	return nil
 }
--- a/opts/mount_test.go
+++ b/opts/mount_test.go
@@ -217,3 +217,86 @@ func TestMountOptSetTmpfsError(t *testing.T) {
 	assert.ErrorContains(t, m.Set("type=tmpfs,target=/foo,tmpfs-mode=foo"), "invalid value for tmpfs-mode")
 	assert.ErrorContains(t, m.Set("type=tmpfs"), "target is required")
 }
+
+func TestMountOptSetBindNonRecursive(t *testing.T) {
+	var mount MountOpt
+	assert.NilError(t, mount.Set("type=bind,source=/foo,target=/bar,bind-nonrecursive"))
+	assert.Check(t, is.DeepEqual([]mounttypes.Mount{
+		{
+			Type:   mounttypes.TypeBind,
+			Source: "/foo",
+			Target: "/bar",
+			BindOptions: &mounttypes.BindOptions{
+				NonRecursive: true,
+			},
+		},
+	}, mount.Value()))
+}
+
+func TestMountOptSetBindRecursive(t *testing.T) {
+	t.Run("enabled", func(t *testing.T) {
+		var mount MountOpt
+		assert.NilError(t, mount.Set("type=bind,source=/foo,target=/bar,bind-recursive=enabled"))
+		assert.Check(t, is.DeepEqual([]mounttypes.Mount{
+			{
+				Type:   mounttypes.TypeBind,
+				Source: "/foo",
+				Target: "/bar",
+			},
+		}, mount.Value()))
+	})
+
+	t.Run("disabled", func(t *testing.T) {
+		var mount MountOpt
+		assert.NilError(t, mount.Set("type=bind,source=/foo,target=/bar,bind-recursive=disabled"))
+		assert.Check(t, is.DeepEqual([]mounttypes.Mount{
+			{
+				Type:   mounttypes.TypeBind,
+				Source: "/foo",
+				Target: "/bar",
+				BindOptions: &mounttypes.BindOptions{
+					NonRecursive: true,
+				},
+			},
+		}, mount.Value()))
+	})
+
+	t.Run("writable", func(t *testing.T) {
+		var mount MountOpt
+		assert.Error(t, mount.Set("type=bind,source=/foo,target=/bar,bind-recursive=writable"),
+			"option 'bind-recursive=writable' requires 'readonly' to be specified in conjunction")
+		assert.NilError(t, mount.Set("type=bind,source=/foo,target=/bar,bind-recursive=writable,readonly"))
+		assert.Check(t, is.DeepEqual([]mounttypes.Mount{
+			{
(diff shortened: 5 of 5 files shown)


AkihiroSuda:
@thaJeztah PTAL 🙏

thaJeztah:
Oh! Thanks for the ping, had this one on my list, started looking, and then forgot 😅

docs/reference/commandline/service_create.md
```diff
@@ -434,7 +435,8 @@ The following options can only be used for bind mounts (`type=bind`):
     <td><b>bind-nonrecursive</b></td>
     <td>
       By default, submounts are recursively bind-mounted as well. However, this behavior can be confusing when a
-      bind mount is configured with <tt>readonly</tt> option, because submounts are not mounted as read-only.
+      bind mount is configured with <tt>readonly</tt> option, because submounts are not mounted as read-only
+      if Docker Engine is older than v25, or Linux kernel is older than v5.12.
```
thaJeztah, line 439: We should probably use a full version for these `v25.0.0`, but probably we should consider rewriting this section to assume "latest" (it now works).

docs/reference/run.md
```diff
@@ -1714,13 +1714,21 @@ $ docker run -d --tmpfs /run:rw,noexec,nosuid,size=65536k my_image
 ### VOLUME (shared filesystems)
 
     -v, --volume=[host-src:]container-dest[:<options>]: Bind mount a volume.
-    The comma-delimited `options` are [rw|ro], [z|Z],
+    The comma-delimited `options` are [rw|ro|ro-non-recursive|ro-force-recursive|rro], [z|Z],
```
thaJeztah, line 1717: I'm on the fence if we want the advanced options on this (we currently also don't have `bind-nonrecursive`).


IIUC, currently we consider `ro` to be "recursive read-only" (best effort) on v25.0.0, and "read-only" (non-recursive) on v24.0.0 and lower.

- How would `rro` and `ro-force-recursive` differ?
- `ro-non-recursive` would be for keeping the "pre-v25.0.0" behavior (read-only, but not for submounts?)

For the last one, I think that would _only_ apply to implicit submounts (i.e., I bind-mount `/foo` and there happens to be a `/foo/some-mount` present, which is a mount), but it would _not_ apply to an explicit mount inside such a path, so wouldn't apply to;

```bash
docker run \ 
  - v /foo:/foo:ro
  -v /bla:/foo/some-mount:rw \
 busybox
```

docs/reference/run.md
```diff
@@ -1714,13 +1714,21 @@ $ docker run -d --tmpfs /run:rw,noexec,nosuid,size=65536k my_image
...
 
     If neither 'rw' or 'ro' is specified then the volume is mounted in
     read-write mode.
 
+    Starting with Docker Engine v25, the `ro` mode makes its submounts read-only when running on
+    Linux kernel v5.12 or newer.
+    To fall back to the behavior of Docker v24, specify `ro-non-recursive`.
+    To explicitly make the mount recursively read-only, specify `ro-force-recursive`
+    or `rro`.
+    The `ro-force-recursive` (`rro`) mode should be used in conjunction with `bind-propagation=rprivate`.
```
thaJeztah, line 1729: Isn't `rprivate` the default?

> The `bind-propagation` option defaults to `rprivate` for both bind mounts and
> volume mounts, and is only configurable for bind mounts. In other words, named
> volumes do not support bind propagation.

If so, we should probably describe that this cannot be used when specifying a custom bind-propagation (other than `rprivate`.

Do we produce an error on the API side if such a combination is set?

docs/reference/commandline/service_create.md
```diff
@@ -445,6 +447,36 @@ The following options can only be used for bind mounts (`type=bind`):
       </ul>
     </td>
   </tr>
+  <tr>
+    <td><b>bind-readonly-nonrecursive</b> or <b>bind-ro-nonrecursive</b></td>
```
thaJeztah, line 451: I'm considering if we should (soft) deprecated `bind-nonrecursive`, and instead have a "recursive mounts" option that takes a (non-boolean) value, e.g. something like;

- `recursive-mounts=enabled` (default)
- `recursive-mounts=disabled`
- `recursive-mounts=readonly` (should error if `bind-propagation` is set by user (to anything other than `rprivate`), but perhaps handled by API)
- `recursive-mounts=force-readonly` (should error if `bind-propagation` is set by user (to anything other than `rprivate`), but perhaps handled by API)

docs/reference/run.md
AkihiroSuda: These are defined on the daemon side:
https://github.com/moby/moby/blob/78cf11575dcf83092cbad255c49ab1e8b7f00d4d/volume/mounts/parser.go#L14-L21

> How would rro and ro-force-recursive differ?

Same.

> ro-non-recursive would be for keeping the "pre-v25.0.0" behavior (read-only, but not for submounts?)

Yes.

AkihiroSuda: (But maybe want rename `ro-non-recursive` to `ro-nonrecursive`, for consistency with `bind-nonrecursive` in the CSV form)

thaJeztah: > These are defined on the daemon side:
> https://github.com/moby/moby/blob/78cf11575dcf83092cbad255c49ab1e8b7f00d4d/volume/mounts/parser.go#L14-L21

Arf.. that legacy API keeps haunting me. I'd like to move away from moving that API altogether (and do conversion on the client-side to use the more structured new API.

I think these options are "advanced" options, and for those, I'd rather use the `--mount` CSV flags, and keep the short-hand flags for the common cases. The shorthand micro-format is already overloaded.

vendor.mod
```diff
@@ -62,6 +62,7 @@ require (
 	github.com/miekg/pkcs11 v1.1.1 // indirect
 	github.com/moby/sys/symlink v0.2.0 // indirect
 	github.com/opencontainers/runc v1.1.7 // indirect
+	github.com/opencontainers/runtime-spec v1.1.0-rc.2 // indirect
```
thaJeztah, line 65: FWIW I was also updating the docker/docker dependency (https://github.com/docker/cli/pull/4341), and didn't get this indirect. I was curious why, but looks like https://github.com/moby/moby/commit/d222bf097cd4c8121900c292f0f6ff243773c3c5 (https://github.com/moby/moby/pull/43980) removes this dependency from `api/types`

AkihiroSuda: This PR no longer touches `vendor.mod`

docs/reference/commandline/service_create.md
AkihiroSuda: What about this?
```go
                case "bind-recursive":
                        valS := val
                        // Allow boolean as an alias to "enabled" or "disabled"
                        if b, err := strconv.ParseBool(valS); err == nil {
                                if b {
                                        valS = "enabled"
                                } else {
                                        valS = "disabled"
                                }
                        }
                        switch valS {
                        case "enabled": // read-only mounts are recursively read-only if Engine >= v25 && kernel >= v5.
12
                                // NOP
                        case "disabled": // alias of bind-nonrecursive=true
                                bindOptions().NonRecursive = true
                        case "readonly-classic": // conforms to Docker v24; read-only mounts are recursively mounted bu
t not recursively read-only
                                bindOptions().ReadOnlyNonRecursive = true
                                mount.ReadOnly = true
                        case "readonly-force": // force recursively read-only, or raise an error
                                bindOptions().ReadOnlyForceRecursive = true
                                bindOptions().Propagation = mounttypes.PropagationRPrivate
                                mount.ReadOnly = true
                        default:
                                return fmt.Errorf("invalid value for %s: %s (must be \"enabled\", \"disabled\", \"reado
nly-classic\", or \"readonly-force\")",
                                        key, val)
                        }
```

docs/reference/run.md
AkihiroSuda: Removed the short form option from the PR. Can be discussed again in a separate PR if somebody needs it.

AkihiroSuda:
@thaJeztah Could you take a look?

thaJeztah:
Sorry, should've posted some comments here. I've been starting to review this one multiple times, and trying to construct a table with possible combinations (and what their result would be), and my head exploded each time :joy:

However, the other day it occurred to me what I _think_ was the part that I had trouble with;

- the advanced syntax provides _two_ separate options to control the "amount of read-only" _and_ whether recursive mounts are included in the mount (`readonly` and `bind-recursive`)
- :point_up: the confusing bit there was "how do these relate"?

So the goal we're aiming for here is;

1. provide the "easy" option (`-v host-path:container-path:ro`) that covers the "90% of use-cases"
2. provide the same through `--mount`, but allow more advanced options for the 10% of use-cases (giving them more control)

_However_ I think (and that's the part I need to look at again, as I was considering this away from my computer), to achieve the same result as `1.`, but using the advanced syntax (`2.`), I now MUST provide all the options (i.e. when using `--mount` ).

While it's good to _allow_ the user to have more fine-grained control, my ideal would be to make that "optional"; even with the advanced `--mount` syntax.

So my thinking is; this:

```bash
--mount type=bind,source=host-path,target=container-path,readonly
```

To be the equivalent of:

```bash
-v host-path:container-path:ro
```

(i.e.; try recursive-readonly, and fallback to non-recursive)

But if the user needs to have more fine-grained control, then allow them to set the `bind-recursive` options;

e.g.: force recursive read-only:

```bash
--mount type=bind,source=host-path,target=container-path,readonly,bind-recursive=readonly-force
```

Maybe we don't even need the `readonly-force`, and it could just be `readonly` (as it's now limited to only the recursive parts)


e.g.: read-only, but no recursive mounts:

```bash
--mount type=bind,source=host-path,target=container-path,readonly,bind-recursive=disabled
```

AkihiroSuda:
Thanks, renamed `readonly-force` to `readonly`.

opts/mount.go
```diff
@@ -112,6 +113,33 @@ func (m *MountOpt) Set(value string) error {
...
+				// NOP
+			case "disabled": // alias of bind-nonrecursive=true
+				bindOptions().NonRecursive = true
+			case "readonly-classic": // conforms to Docker v24; read-only mounts are recursively mounted but not recursively read-only
+				bindOptions().ReadOnlyNonRecursive = true
+				mount.ReadOnly = true
+			case "readonly": // force recursively read-only, or raise an error
+				bindOptions().ReadOnlyForceRecursive = true
+				bindOptions().Propagation = mounttypes.PropagationRPrivate
+				mount.ReadOnly = true
```
thaJeztah, line 138: I think we're getting close (sorry had to jump out earlier as I had some calls); if we consider the `bind-recursive` option to be controlling options for _recursive mounts_ (and not for the "main" mount), should we update `mount.Readonly` at all here?

My thinking (please reply if I'm completely off the mark ❤️);

- `readonly` is the "overall" option; when enabled, by default it makes both the "main" mount, and any nested mounts read-only (i.e. "best-effort recursive-readonly"
- `bind-recursive=disabled` disables nested mounts
- `bind-recursive=readonly` forces nested mounts to be read-only
    - 👆 this can also be done separate from `readonly`, so `--mount type=bind,source=host-path,target=container-path,bind-recursive=readonly` would then mean: the main mount is read-write, but nested mounts (if any) would be mounted read-only
    - 👆 ❓ is that combination technically possible?
- which leaves `readonly-classic` if we don't set `mount.ReadOnly` there, would that make it the same as `enabled` ? (i.e.; nested mounts are mounted, but "read-write") ? If that's the case, then _maybe_ we don't need that option 🤔

AkihiroSuda: `mount.ReadOnly = true` is set here, as `mount.ReadOnly == false && bindOptions().ReadOnlyXXX = true` doesn't seem a valid configuration.

thaJeztah: That's the part I was wondering; is it (technically) possible to make the "main" mount (so `target`) read-write, but making _nested_ mounts read-only?

thaJeztah: Basically; my thinking here is to make nested mounts "inherit" the `readonly` option as a default, with the option to differentiate (through the `bind-recursive` options). In that case it's (I _think_) easier to comprehend (by default, they'll all be attempted the same, but if I need them to be different, I can control that through this option)

AkihiroSuda: > That's the part I was wondering; is it (technically) possible to make the "main" mount (so `target`) read-write, but making _nested_ mounts read-only?

Theoretically possible by scanning procfs, but I don't think we want to implement it

opts/mount.go
```diff
@@ -112,6 +113,33 @@ func (m *MountOpt) Set(value string) error {
...
+				} else {
+					valS = "disabled"
+				}
+			}
+			switch valS {
+			case "enabled": // read-only mounts are recursively read-only if Engine >= v25 && kernel >= v5.12
+				// NOP
+			case "disabled": // alias of bind-nonrecursive=true
+				bindOptions().NonRecursive = true
+			case "readonly-classic": // conforms to Docker v24; read-only mounts are recursively mounted but not recursively read-only
```
neersighted, line 132: `readonly-classic` doesn't feel right to me, as we're trying to split the "recursion" from the "readonlyness" with this.

How about `writable`, which makes it clear that recursive mounts will be writable/that was the legacy behavior?

---

I also think setting `ReadOnly` here is a smell -- we should instead have some check that errors when `ReadOnly` has not been set by the user (e.g. `readonly=false,bind-recursive=readonly` should be an error in validation).

opts/mount.go
neersighted, line 138: Similar to the above, I don't want to set `ReadOnly` for the user here -- a lack of `ReadOnly` should instead be a validation issue.

opts/mount.go
```diff
@@ -112,6 +113,33 @@ func (m *MountOpt) Set(value string) error {
...
+				bindOptions().ReadOnlyNonRecursive = true
+				mount.ReadOnly = true
+			case "readonly": // force recursively read-only, or raise an error
+				bindOptions().ReadOnlyForceRecursive = true
+				bindOptions().Propagation = mounttypes.PropagationRPrivate
```
neersighted, line 137: I don't think we should overwrite the propagation here unconditionally. I think it's fine to set it implicitly from a UX perspective, but if the user specifies any different propagation it should be a validation error.

neersighted (changes requested):
I think this is a lot better UX-wise, with the recursion divorced from the readonly option. Nonetheless, I think it's not quite there, mostly because confusing combinations are possible and should be caught in validation instead of silently overriding each other.

That's hard, since we don't have validation yet/the way this code is structured makes it rather obtuse. Still, we could e.g. use a fixed source order with a slice that defines a strong ordering, and check that way, or we could think of another method.

Let me know if my comments make sense and I can help with figuring out how to implement them.

opts/mount.go
AkihiroSuda: What about "classic" or "legacy"?

opts/mount.go
AkihiroSuda: Updated

opts/mount.go
AkihiroSuda: Updated

opts/mount.go
neersighted: I still think `classic` or `legacy` don't really describe what it **does**, and when documenting this option and the default for old engine versions, it really hammers home the old behavior. I don't object to an alias under those names, but I would prefer avoiding it as I think they're meaningless unless you have an excellent understanding of the kernel/engine history around this.

opts/mount.go
neersighted: Hmm, I do like the trivial validation overall, but I'm not sure about this case. I don't think that requiring the user to set the propagation explicitly is good UX; I'd prefer this to be set implicitly, but to error if the user explicitly sets a different propagation. Unfortunately that doesn't work with the simple validation you've added, but maybe you have some ideas to that end?

opts/mount.go
```diff
@@ -161,6 +187,15 @@ func (m *MountOpt) Set(value string) error {
 		return fmt.Errorf("cannot mix 'tmpfs-*' options with mount type '%s'", mount.Type)
 	}
 
+	if mount.BindOptions != nil && mount.BindOptions.ReadOnlyForceRecursive {
```
neersighted, line 190: This probably needs a check for `ReadOnlyNonRecursive` as well, as `readonly=false,bind-recursive=readonly` makes no sense.

thaJeztah: > as `readonly=false,bind-recursive=readonly`

Ah, yes, that was the case I mentioned above (and was hoping it would work, as I could see use-cases for a "writable" bind-mount with _nested mounts_ read-only), but currently not a combination that was possible; https://github.com/docker/cli/pull/4316#discussion_r1340348214

thaJeztah: Just to be sure; in the current implementation, does `--mount type=bind,readonly,.....` do the same as `-v source:target:ro` (so best-effort recursive), or do we need to make changes to the `readonly` option to make that happen (I lost track again on which combination did that, and what parts we did on the daemon side)

thaJeztah:
FWIW; we discussed this PR Yesterday, and w.r.t. validation it's possible that we need to make some follow-up changes, because of the way these options are parsed (i.e., order can matter if multiple options interact with the same "API" options behind the scenes).

We are looking at doing a first beta for v25.0 soon (to get the release train warmed up), and I think for that the _main_ goal is to have the UX "sorted" (as good as possible), but after that we can still tweak / fix validation corner cases etc.

opts/mount.go
AkihiroSuda: The default implicit "rprivate" propagation sometimes flips to "rslave":
- https://github.com/moby/moby/pull/36055

So I guess we should demand explicit "rprivate" here

opts/mount.go
AkihiroSuda: > This probably needs a check for `ReadOnlyNonRecursive` as well, as `readonly=false,bind-recursive=readonly` makes no sense.

Done (I assume you meant `readonly=false,bind-recursive=writable`)

AkihiroSuda: > Just to be sure; in the current implementation, does `--mount type=bind,readonly,.....` do the same as `-v source:target:ro` (so best-effort recursive), or do we need to make changes to the `readonly` option to make that happen (I lost track again on which combination did that, and what parts we did on the daemon side)

The former one.

opts/mount.go
neersighted: Yeah, I had that in mind. What I meant to say is that we want to be explicit at the API level, but I'm not sure that we want to require the user to explicitly specify a propagation mode to use `bind-recursive=readonly`. However with the current validation, this is messy.

I am okay with what you have; it makes the feature usable, and `bind-recursive=readonly` is definitely a power-user feature/we're "upgrading" users by default to the new `enabled` behavior anyway.

How about a `// TODO:` to implicitly set propagation and error if the user specifies a propagation in a future refactor/UX polish pass?

neersighted (approved):
Overall LGTM with reservations (about the need for the user to explicitly set the propagation mode); as per my review comment, I think a `TODO` would be fine here, and we should get this in.

opts/mount.go
AkihiroSuda: Added

thaJeztah (approved):
LGTM

added a docs/revisit label, as we will have to do some docs follow-ups, as well as looking at completion scripts (not sure if we provide completion for these currently)

AkihiroSuda:
docs:
- https://github.com/docker/cli/pull/4606

opts/mount.go
```diff
@@ -112,6 +114,32 @@ func (m *MountOpt) Set(value string) error {
...
+			valS := val
+			// Allow boolean as an alias to "enabled" or "disabled"
+			if b, err := strconv.ParseBool(valS); err == nil {
+				if b {
+					valS = "enabled"
+				} else {
+					valS = "disabled"
+				}
+			}
+			switch valS {
```
thaJeztah, line 128: Opened a PR for discussion to remove the boolean variants (motivation outlined in that PR);

- https://github.com/docker/cli/pull/4671

PR #4337, opened 2023-06-08
cli/command/container: don't mutate ConfigFile.DetachKeys

thaJeztah:
### cli/command/container: createContainer(): return container-ID

This function returned the whole response, but we already handled the
warnings included in the response as part of the function. All consumers
of this function only used the container-ID, so let's simplify and return
just that (it's a non-exported func, so we can change the signature again
if we really need it).

### cli/command/container: don't mutate ConfigFile.DetachKeys

This code was introduced in https://github.com/moby/moby/commit/15aa2a663b47b6126a66efefcadb64edfbffb9f5 (https://github.com/moby/moby/pull/15666),
but from those changes, it appears that overwriting the config value was
merely out of convenience, and that struct being used as an intermediate.

While changing the config here should be mostly ephemeral, and not written
back to the config-file, let's be clear on intent, and not mutatte the config
as part of this code.

### DIFF
--- a/cli/command/container/attach.go
+++ b/cli/command/container/attach.go
@@ -86,8 +86,9 @@ func runAttach(dockerCli command.Cli, opts *attachOptions) error {
 		return err
 	}
 
+	detachKeys := dockerCli.ConfigFile().DetachKeys
 	if opts.detachKeys != "" {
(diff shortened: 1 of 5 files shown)


laurazard (approved):
LGTM

PR #4344, opened 2023-06-12
docs: update the runtime configuration section

dvdksn:
relates to: docker/docs#17517

**- What I did**

Updated the section about runtime options, it now includes how to configure
containerd shims using the `runtimeType` and `options` fields.

Because there are now two ways to configure runtimes, it also demanded that I
refactor the entire section.

I've tried to add justification for when to use `runtimeType`, and when to use
`path`. It's quite difficult to accurately capture the distinction, while also
keeping the section somewhat intelligible. I tried.

**- How I did it**

**- How to verify it**

🌭

### DIFF
--- a/docs/reference/commandline/dockerd.md
+++ b/docs/reference/commandline/dockerd.md
@@ -396,36 +396,197 @@ Defaults to 20G.
 C:\> dockerd --storage-opt size=40G
 ```
 
-### Docker runtime execution options
+### Runtime options
 
 The Docker daemon relies on a
 [OCI](https://github.com/opencontainers/runtime-spec) compliant runtime
 (invoked via the `containerd` daemon) as its interface to the Linux
 kernel `namespaces`, `cgroups`, and `SELinux`.
 
-By default, the Docker daemon automatically starts `containerd`. If you want to
-control `containerd` startup, manually start `containerd` and pass the path to
-the `containerd` socket using the `--containerd` flag. For example:
+#### Configure container runtimes
+
+By default, the Docker daemon uses runc as a container runtime.
+You can configure the daemon to add additional runtimes.
+
+containerd shims installed on `PATH` can be used directly, without the need
+to edit the daemon's configuration. For example, if you install the Kata
+Containers shim (`containerd-shim-kata-v2`) on `PATH`, then you can select that
+runtime with `docker run` without having to edit the daemon's configuration:
 
 ```console
-$ sudo dockerd --containerd /var/run/dev/docker-containerd.sock
+$ docker run --runtime io.containerd.kata.v2
 ```
 
-Runtimes can be registered with the daemon either via the
-configuration file or using the `--add-runtime` command line argument.
+Container runtimes that don't implement containerd shims, or containerd shims
+installed outside of `PATH`, must be registered with the daemon, either via the
+configuration file or using the `--add-runtime` command line flag.
 
-The following is an example adding 2 runtimes via the configuration:
+For examples on how to use other container runtimes, see
+[Alternative container runtimes](https://docs.docker.com/engine/alternative-container-runtimes)
+
+##### Configure runtimes using `daemon.json`
+
+To register and configure container runtimes using the daemon's configuration
+file, add the runtimes as entries under `runtimes`:
 
 ```json
 {
-  "default-runtime": "runc",
   "runtimes": {
-    "custom": {
-      "path": "/usr/local/bin/my-runc-replacement",
-      "runtimeArgs": [
-        "--debug"
-      ]
+    "<runtime>": {}
+  }
+}
+```
+
+The key of the entry (`<runtime>` in the previous example) represents the name
+of the runtime. This is the name that you reference when you run a container,
+using `docker run --runtime <runtime>`.
+
+The runtime entry contains an object specifying the configuration for your
+runtime. The properties of the object depends on what kind of runtime you're
+looking to register:
+
+- If the runtime implements its own containerd shim, the object shall contain
+  a `runtimeType` field and an optional `options` field.
+
+  ```json
+  {
+    "runtimes": {
+      "<runtime>": {
+        "runtimeType": "<name-or-path>",
+        "options": {}
+      }
+    }
+  }
+  ```
+
+  See [Configure shims](#configure-containerd-shims).
+
+- If the runtime is designed to be a drop-in replacement for runc,
+  the object contains a `path` field, and an optional `runtimeArgs` field.
+
+  ```json
+  {
+    "runtimes": {
+      "<runtime>": {
+        "path": "/path/to/bin",
+        "runtimeArgs": ["...args"]
+      }
+    }
+  }
+  ```
+
+  See [Configure runc drop-in replacements](#configure-runc-drop-in-replacements).
+
+After changing the runtimes configuration in the configuration file,
+you must reload or restart the daemon for changes to take effect:
+
+```console
+$ sudo systemctl reload dockerd
+```
+
+##### Configure containerd shims
+
+If the runtime that you want to register implements a containerd shim,
+or if you want to register a runtime which uses the runc shim,
+use the following format for the runtime entry:
+
+```json
+{
+  "runtimes": {
+    "<runtime>": {
+      "runtimeType": "<name-or-path>",
+      "options": {}
+    }
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/dockerd.md
```diff
@@ -396,61 +396,240 @@ Defaults to 20G.
...
+
+- If the runtime implements its own containerd shim, or if you want to
+  configure the default runc shim, the object shall contain a `runtimeType`
+  field, and an optional `options` field.
+
+  ```json
+  {
+    "runtimes": {
+      "<name>": {
+        "runtimeType": "<containerd-shim>",
```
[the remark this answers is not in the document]
dvdksn: I converted this to a gvisor example instead, like the one below it. I think it should be clearer now.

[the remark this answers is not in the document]
dvdksn: oh, grr, scratch that. I changed in the wrong place...

docs/reference/commandline/dockerd.md
```diff
@@ -396,61 +396,241 @@ Defaults to 20G.
...
+Entries with this format use the default containerd runc shim to invoke a
+custom runtime binary, with optional CLI args. For an example, see
+[Alternative container runtimes > youki](https://docs.docker.com/engine/alternative-runtimes/#youki)
+
+##### Register runtimes using the CLI
+
+You can use the `--add-runtime` CLI flag to register runtimes on daemon startup.
+
+You can specify either the name of a fully qualified containerd runtime shim,
+or the name of a registered runtime. The executable must be installed on `PATH`.
```
corhere, line 585: This whole paragraph is incorrect. The `--add-runtime` flag can only be used to register runC drop-in replacements. The flag `--add-runtime myruntime=/path/to/binary` is exactly equivalent to the JSON config
```json
{
  "runtimes": {
    "myruntime": {
      "path": "/path/to/runtime"
    }
  }
}
```

Aside from being unable to define runtime arguments, the `--add-runtime` flag supports the exact same values as the JSON runtime path config. Notably, absolute paths are accepted, and paths which do not contain slashes are looked for in the `PATH` directories, so e.g. `--add-runtime crun=crun` or
```json
{
  "runtimes": {
    "crun": {
      "path": "crun"
    }
  }
}
```
is legal. In fact, the daemon's built-in default runtime configuration is equivalent to `--add-runtime runc=runc`.

docs/reference/commandline/dockerd.md
```diff
@@ -396,61 +396,241 @@ Defaults to 20G.
...
 
 > **Note**
 >
 > Defining runtime arguments via the command line is not supported.
 
-#### Options for the runtime
+##### Configure the default container runtime
+
+You can specify either the name of a fully qualified containerd runtime shim,
+or the name of a registered runtime. The executable must be installed on `PATH`.
```
corhere, line 600: The default runtime option sets the default value for the `docker run --runtime` flag, and therefore has exactly the same acceptable range of values. That is, the default runtime can be set to either
1. the name of a registered runtime, or
2. the fully-qualified name of a shim whose executable is installed on `PATH`.

As currently phrased, it could be read that _a registered runtime name_ can only be set as the default runtime when the runtime executable is installed on `PATH`, which is not true. If it's registered, it can be set as the default, full stop.

docs/reference/commandline/dockerd.md
```diff
@@ -396,61 +396,241 @@ Defaults to 20G.
...
-By default, the Docker daemon automatically starts `containerd`. If you want to
-control `containerd` startup, manually start `containerd` and pass the path to
-the `containerd` socket using the `--containerd` flag. For example:
+#### Configure container runtimes
+
+By default, the Docker daemon uses runc as a container runtime.
+You can configure the daemon to add additional runtimes.
+
+containerd shims installed on `PATH` can be used directly, without the need
+to edit the daemon's configuration.
```
corhere, line 412: How does one use them directly? The only mention of `docker run --runtime` in this page is in the context of using a _registered_ runtime so users may not make the connection that `docker run --runtime io.containerd.kata.v2` can be used to run containers with Kata once `containerd-shim-kata-v2` is installed on `PATH`, without any further configuration.

corhere (changes requested):
Looking good! Just needs a couple of tweaks first

dvdksn:
Thanks @corhere and @thaJeztah, I've addressed the comments, hopefully we can merge now

docs/reference/commandline/dockerd.md
```diff
@@ -396,61 +396,258 @@ Defaults to 20G.
...
+- Using the daemon configuration file
+
+  ```json
+  {
+    "default-runtime": "io.containerd.runsc.v1"
+  }
+  ```
 
-#### Options for the runtime
+- Using the `dockerd` CLI
```
thaJeztah, line 629:
```suggestion
When you use the `--default-runtime` CLI flag, use the following format:
```

docs/reference/commandline/dockerd.md
```diff
@@ -396,61 +396,258 @@ Defaults to 20G.
...
-You can configure the runtime using options specified
-with the `--exec-opt` flag. All the flag's options have the `native` prefix. A
-single `native.cgroupdriver` option is available.
+  ```console
+  $ dockerd --default-runtime io.containerd.runsc.v1
+  ```
 
-The `native.cgroupdriver` option specifies the management of the container's
-cgroups. You can only specify `cgroupfs` or `systemd`. If you specify
+#### Use a manually controlled containerd daemon
```
thaJeztah, line 635:
```suggestion
#### <a name="containerd"></a> Use a manually controlled containerd daemon (--containerd)
```

(see my other comment) not sure if "manually controlled" is the right wording here; this option would usually be if containerd is running as its own service (which is the default nowadays).

docs/reference/commandline/dockerd.md
```diff
@@ -396,61 +396,258 @@ Defaults to 20G.
...
 
 The Docker daemon relies on a
 [OCI](https://github.com/opencontainers/runtime-spec) compliant runtime
 (invoked via the `containerd` daemon) as its interface to the Linux
 kernel `namespaces`, `cgroups`, and `SELinux`.
 
-By default, the Docker daemon automatically starts `containerd`. If you want to
-control `containerd` startup, manually start `containerd` and pass the path to
-the `containerd` socket using the `--containerd` flag. For example:
+#### Configure container runtimes
```
thaJeztah, line 406:
```suggestion
#### <a name="runtime"></a> Configure container runtimes (--runtime)
```

neersighted (commented):
I don't have time for a complete review yet, but I'd like to register my intention to dig in to this in the next day or so and provide feedback.

docs/reference/commandline/dockerd.md
```diff
@@ -396,61 +396,258 @@ Defaults to 20G.
...
+[Alternative container runtimes > youki](https://docs.docker.com/engine/alternative-runtimes/#youki)
+
+##### Configure the default container runtime
+
+You can specify either the name of a fully qualified containerd runtime shim,
+or the name of a registered runtime.
+
+The following examples set the gVisor shim as the default runtime:
+
+- Using the daemon configuration file
```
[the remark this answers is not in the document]
dvdksn: good idea, thanks for spotting

docs/reference/commandline/dockerd.md
```diff
@@ -396,61 +396,258 @@ Defaults to 20G.
...
+  $ dockerd --default-runtime io.containerd.runsc.v1
+  ```
 
-The `native.cgroupdriver` option specifies the management of the container's
-cgroups. You can only specify `cgroupfs` or `systemd`. If you specify
+#### Use a manually controlled containerd daemon
+
+By default, the Docker daemon automatically starts `containerd`. If you want to
+control `containerd` startup, manually start `containerd` and pass the path to
+the `containerd` socket using the `--containerd` flag. For example:
```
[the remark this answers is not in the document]
dvdksn: OK, this is quite good information. I didn't actually change this section when updating the doc; I only moved it around. So for that reason (and because this is still under discussion) I will leave it as is for now.

docs/reference/commandline/dockerd.md
dvdksn: "Run containerd standalone"?

thaJeztah (approved):
LGTM, thanks!

PR #4374, opened 2023-06-26
docs: fix static ip example, network needs a subnet

dvdksn:
- relates to https://github.com/moby/moby/issues/45806

**- What I did**

Clarify that static container IPs require networks with an explicit subnet block

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/commandline/run.md
+++ b/docs/reference/commandline/run.md
@@ -485,10 +485,12 @@ $ docker run -itd --network=my-net busybox
 ```
 
 You can also choose the IP addresses for the container with `--ip` and `--ip6`
-flags when you start the container on a user-defined network.
+flags when you start the container on a user-defined network. To assign a
+static IP to containers, you must specify subnet block for the network.
 
 ```console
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM

thaJeztah:
@dvdksn can you do a cherry-pick as well for 24.0?

wushu037:
@thaJeztah , thank you for your handling.

In addition to this, there is a bug: networks created with `-subnet` do not automatically set the gateway, you need to specify it with `-gateway`. 
However, [this document](https://docs.docker.com/engine/reference/commandline/network_create/#specify-advanced-options) mentions that if `-gateway` is omitted, the engine will automatically set the gateway.
> If you omit the --gateway flag the Engine selects one for you from inside a preferred pool.

The reason for saying "not automatically set" is that you can't see the `gateway` when you look at the network information with the `docker inspect` command.
However, if you add the container to the network, you can see the gateway address again by running `route -n` inside the container, which is `.1` of `subnet` by default, e.g. `172.10.0.1`.

PR #4376, opened 2023-06-26
Introduce support for CLI plugin hooks

laurazard:
**- What I did**

Introduce a "hooks" mechanism for CLI plugins, which enables CLI plugins adding useful, contextual messages/actions after a command has been executed.

This works by defining a special plugin command (such as the metadata command) which the CLI calls with some information about the just-executed command (currently, command and flag names). The plugin then prints a json-encoded to it's stdout with a template for the message it would like the CLI to print.

Using such a system, a plugin to help new docker users could, for example, declare intent to hook into `docker build`, and return a templated message instructing the user how to run their newly-built image.

A templating approach was chosen to allow plugin messages to include information from the original command execution (such as the tag used for the image in `build`), so that messages can provide more useful information (such as printing "Run your image with `docker run my-image`" after a `docker build -t my-image .`
Helpers are provided in the form of `manager.TemplateReplaceFlagValue("tag")` for this case, and use Go Templates under the hood.

The templating approach chosen also keeps sensitive data from leaving the CLI process.

This PR also introduces a "features" map into the CLI `config.json`, to allow for global enabling/disabling of this (and future) features.

<details>
  <summary><b>Example Plugin</b></summary>

```go
func main() {
	plugin.Run(func(dockerCli command.Cli) *cobra.Command {
		cmd := RootCommand(dockerCli)
		originalPreRun := cmd.PersistentPreRunE
		cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
			if err := plugin.PersistentPreRunE(cmd, args); err != nil {
				//nolint: wrapcheck
				return err
			}
			if originalPreRun != nil {
				return originalPreRun(cmd, args)
			}
			return nil
		}
		return cmd
	},
		manager.Metadata{
			SchemaVersion: "0.1.0",
			Vendor:        "Docker Inc.",
			Version:       "0.1",
			HookCommands:  []string{"build"},  // declare intent to hook into `build` subcommand invocations
		})
}

func RootCommand(dockerCli command.Cli) *cobra.Command {
	rootCmd := &cobra.Command{
		Short:            "Docker Hints",
		Use:              "hints",
		TraverseChildren: true,
	}

	rootCmd.AddCommand(hookCommand(dockerCli))
	return rootCmd
}

func hookCommand(dockerCli command.Cli) *cobra.Command {
	hookCmd := &cobra.Command{
		Use:   manager.HookSubcommandName,  // register hook command that the CLI will invoke
		Short: "runs the plugins hooks",
		RunE: utils.Adapt(func(ctx context.Context, args []string) error {
			runHooks(dockerCli.Out(), []byte(args[0]))
			return nil
		}),
		Args: cobra.ExactArgs(1),
	}

	return hookCmd
}

func runHooks(out io.Writer, input []byte) {
	var c manager.HookPluginData
	_ = json.Unmarshal(input, &c)

	hint := "Run this image with `docker run " + manager.TemplateReplaceFlagValue("tag") + "`" // template helper
	returnType := manager.HookMessage{
		Template: hint,
	}
	enc := json.NewEncoder(out)
	enc.SetEscapeHTML(false)
	enc.SetIndent("", "     ")
	_ = enc.Encode(returnType)
}
```

Which results in:
image

</details>

-------------------------------

<details>
<summary><b>Performance Callouts</b></summary>

Edited: for current approach + benchmarks see https://github.com/docker/cli/pull/4376#issuecomment-2008580277

</details>
<details>
  <summary><b>Security Callouts</b></summary>

The current approach works by letting plugins declare a list of strings which are the CLI commands (such as `build`, `run`, etc.) whose execution the plugin wants to be hooked into. The CLI then only invokes each plugin for the commands the plugin has declared support for.

Since a plugin might want to be invoked for more than one command, some information must be passed into the plugin when it is invoked so that it can decide what information it wants to add – which for right now, is codified into the `HookPluginData` type and contains only a string representing the command being executed. Richer output is made possible by employing a _templating_ approach, wherein the plugin can return a template string with special values which the CLI then processes – e.g.
```
$ docker build -t my-image .

# a plugin returning
You can run the image you just built with `docker run {docker-cli-plugin-hook-flag-tag}`

# gets parsed into
You can run the image you just built with `docker run my-image`
```

This allows for plugins to provide useful/contextual messages, without passing any potentially sensitive information to any other binary.

Possible future developments to make this more secure are things such as always being explicit when a plugin is invoked, so that the user is aware, or prompting the user whether they are okay with a plugin hooking into a command execution on first execution.
</details>

----------------------

image

### DIFF
--- a/cli-plugins/hooks/printer.go
+++ b/cli-plugins/hooks/printer.go
@@ -1,0 +1,18 @@
+package hooks
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/morikuni/aec"
+)
+
+func PrintNextSteps(out io.Writer, messages []string) {
+	if len(messages) == 0 {
+		return
+	}
+	fmt.Fprintln(out, aec.Bold.Apply("\nWhat's next:"))
+	for _, n := range messages {
+		_, _ = fmt.Fprintf(out, "    %s\n", n)
+	}
+}
--- a/cli-plugins/hooks/template.go
+++ b/cli-plugins/hooks/template.go
@@ -1,0 +1,115 @@
+package hooks
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"strconv"
+	"text/template"
+
+	"github.com/spf13/cobra"
+)
+
+type HookType int
+
+const (
+	NextSteps = iota
+)
+
+// HookMessage represents a plugin hook response. Plugins
+// declaring support for CLI hooks need to print a json
+// representation of this type when their hook subcommand
+// is invoked.
+type HookMessage struct {
+	Type     HookType
+	Template string
+}
+
+// TemplateReplaceSubcommandName returns a hook template string
+// that will be replaced by the CLI subcommand being executed
+//
+// Example:
+//
+// "you ran the subcommand: " + TemplateReplaceSubcommandName()
+//
+// when being executed after the command:
+// `docker run --name "my-container" alpine`
+// will result in the message:
+// `you ran the subcommand: run`
+func TemplateReplaceSubcommandName() string {
+	return hookTemplateCommandName
+}
+
+// TemplateReplaceFlagValue returns a hook template string
+// that will be replaced by the flags value.
+//
+// Example:
+//
+// "you ran a container named: " + TemplateReplaceFlagValue("name")
+//
+// when being executed after the command:
+// `docker run --name "my-container" alpine`
+// will result in the message:
+// `you ran a container named: my-container`
+func TemplateReplaceFlagValue(flag string) string {
+	return fmt.Sprintf(hookTemplateFlagValue, flag)
+}
+
+// TemplateReplaceArg takes an index i and returns a hook
+// template string that the CLI will replace the template with
+// the ith argument, after processing the passed flags.
+//
+// Example:
+//
+// "run this image with `docker run " + TemplateReplaceArg(0) + "`"
+//
+// when being executed after the command:
+// `docker pull alpine`
+// will result in the message:
+// "Run this image with `docker run alpine`"
+func TemplateReplaceArg(i int) string {
+	return fmt.Sprintf(hookTemplateArg, strconv.Itoa(i))
+}
+
+func ParseTemplate(hookTemplate string, cmd *cobra.Command) (string, error) {
+	tmpl := template.New("").Funcs(commandFunctions)
+	tmpl, err := tmpl.Parse(hookTemplate)
+	if err != nil {
+		return "", err
+	}
+	b := bytes.Buffer{}
+	err = tmpl.Execute(&b, cmd)
+	if err != nil {
+		return "", err
+	}
+	return b.String(), nil
+}
+
+var ErrHookTemplateParse = errors.New("failed to parse hook template")
+
+const (
+	hookTemplateCommandName = "{{.Name}}"
+	hookTemplateFlagValue   = `{{flag . "%s"}}`
+	hookTemplateArg         = "{{arg . %s}}"
+)
+
+var commandFunctions = template.FuncMap{
+	"flag": getFlagValue,
+	"arg":  getArgValue,
+}
+
+func getFlagValue(cmd *cobra.Command, flag string) (string, error) {
+	cmdFlag := cmd.Flag(flag)
+	if cmdFlag == nil {
+		return "", ErrHookTemplateParse
+	}
+	return cmdFlag.Value.String(), nil
+}
+
+func getArgValue(cmd *cobra.Command, i int) (string, error) {
+	flags := cmd.Flags()
+	if flags == nil {
+		return "", ErrHookTemplateParse
+	}
+	return flags.Arg(i), nil
+}
--- a/cli-plugins/manager/hooks.go
+++ b/cli-plugins/manager/hooks.go
@@ -1,0 +1,127 @@
+package manager
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/docker/cli/cli-plugins/hooks"
+	"github.com/docker/cli/cli/command"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// HookPluginData is the type representing the information
+// that plugins declaring support for hooks get passed when
+// being invoked following a CLI command execution.
+type HookPluginData struct {
+	RootCmd string
+	Flags   map[string]string
+}
+
+// RunPluginHooks calls the hook subcommand for all present
+// CLI plugins that declare support for hooks in their metadata
+// and parses/prints their responses.
+func RunPluginHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command, plugin string, args []string) error {
+	subCmdName := subCommand.Name()
+	if plugin != "" {
+		subCmdName = plugin
+	}
+	var flags map[string]string
+	if plugin == "" {
+		flags = getCommandFlags(subCommand)
+	} else {
+		flags = getNaiveFlags(args)
+	}
+	nextSteps := invokeAndCollectHooks(dockerCli, rootCmd, subCommand, subCmdName, flags)
+
+	hooks.PrintNextSteps(dockerCli.Err(), nextSteps)
+	return nil
+}
+
+func invokeAndCollectHooks(dockerCli command.Cli, rootCmd, subCmd *cobra.Command, hookCmdName string, flags map[string]string) []string {
+	pluginsCfg := dockerCli.ConfigFile().Plugins
+	if pluginsCfg == nil {
+		return nil
+	}
+
+	nextSteps := make([]string, 0, len(pluginsCfg))
+	for pluginName, cfg := range pluginsCfg {
+		if !registersHook(cfg, hookCmdName) {
+			continue
+		}
+
+		p, err := GetPlugin(pluginName, dockerCli, rootCmd)
+		if err != nil {
+			continue
+		}
+
+		hookReturn, err := p.RunHook(hookCmdName, flags)
+		if err != nil {
+			// skip misbehaving plugins, but don't halt execution
+			continue
+		}
+
+		var hookMessageData hooks.HookMessage
+		err = json.Unmarshal(hookReturn, &hookMessageData)
+		if err != nil {
+			continue
+		}
+
+		// currently the only hook type
+		if hookMessageData.Type != hooks.NextSteps {
+			continue
+		}
+
+		processedHook, err := hooks.ParseTemplate(hookMessageData.Template, subCmd)
+		if err != nil {
+			continue
+		}
+		nextSteps = append(nextSteps, processedHook)
+	}
+	return nextSteps
+}
+
+func registersHook(pluginCfg map[string]string, subCmdName string) bool {
+	hookCmdStr, ok := pluginCfg["hooks"]
+	if !ok {
+		return false
+	}
+	commands := strings.Split(hookCmdStr, ",")
+	for _, hookCmd := range commands {
+		if hookCmd == subCmdName {
+			return true
+		}
+	}
+	return false
+}
+
+func getCommandFlags(cmd *cobra.Command) map[string]string {
+	flags := make(map[string]string)
+	cmd.Flags().Visit(func(f *pflag.Flag) {
+		var fValue string
+		if f.Value.Type() == "bool" {
+			fValue = f.Value.String()
+		}
+		flags[f.Name] = fValue
+	})
+	return flags
+}
+
+// getNaiveFlags string-matches argv and parses them into a map.
+// This is used when calling hooks after a plugin command, since
+// in this case we can't rely on the cobra command tree to parse
+// flags in this case. In this case, no values are ever passed,
+// since we don't have enough information to process them.
+func getNaiveFlags(args []string) map[string]string {
+	flags := make(map[string]string)
+	for _, arg := range args {
+		if strings.HasPrefix(arg, "--") {
+			flags[arg[2:]] = ""
+			continue
+		}
+		if strings.HasPrefix(arg, "-") {
+			flags[arg[1:]] = ""
+		}
+	}
+	return flags
+}
--- a/cli-plugins/manager/metadata.go
+++ b/cli-plugins/manager/metadata.go
@@ -8,6 +8,11 @@ const (
 	// which must be supported by every plugin and returns the
 	// plugin metadata.
 	MetadataSubcommandName = "docker-cli-plugin-metadata"
+
+	// HookSubcommandName is the name of the plugin subcommand
+	// which must be implemented by plugins declaring support
+	// for hooks in their metadata.
+	HookSubcommandName = "docker-cli-plugin-hooks"
 )
 
 // Metadata provided by the plugin.
--- a/cli-plugins/manager/plugin.go
+++ b/cli-plugins/manager/plugin.go
@@ -2,6 +2,7 @@ package manager
 
 import (
 	"encoding/json"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -99,4 +100,23 @@ func newPlugin(c Candidate, cmds []*cobra.Command) (Plugin, error) {
 		return p, nil
 	}
 	return p, nil
+}
+
+// RunHook executes the plugin's hooks command
+// and returns its unprocessed output.
+func (p *Plugin) RunHook(cmdName string, flags map[string]string) ([]byte, error) {
+	hDataBytes, err := json.Marshal(HookPluginData{
+		RootCmd: cmdName,
+		Flags:   flags,
+	})
+	if err != nil {
+		return nil, wrapAsPluginError(err, "failed to marshall hook data")
+	}
+
+	hookCmdOutput, err := exec.Command(p.Path, p.Name, HookSubcommandName, string(hDataBytes)).Output()
+	if err != nil {
+		return nil, wrapAsPluginError(err, "failed to execute plugin hook subcommand")
+	}
+
+	return hookCmdOutput, nil
 }
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -187,6 +187,36 @@ func (cli *DockerCli) BuildKitEnabled() (bool, error) {
 	return cli.ServerInfo().OSType != "windows", nil
 }
 
+// HooksEnabled returns whether plugin hooks are enabled.
+func (cli *DockerCli) HooksEnabled() bool {
+	// legacy support DOCKER_CLI_HINTS env var
+	if v := os.Getenv("DOCKER_CLI_HINTS"); v != "" {
+		enabled, err := strconv.ParseBool(v)
+		if err != nil {
+			return false
+		}
+		return enabled
+	}
+	// use DOCKER_CLI_HOOKS env var value if set and not empty
+	if v := os.Getenv("DOCKER_CLI_HOOKS"); v != "" {
+		enabled, err := strconv.ParseBool(v)
+		if err != nil {
+			return false
+		}
+		return enabled
+	}
+	featuresMap := cli.ConfigFile().Features
+	if v, ok := featuresMap["hooks"]; ok {
+		enabled, err := strconv.ParseBool(v)
+		if err != nil {
+			return false
+		}
+		return enabled
+	}
+	// default to false
+	return false
+}
+
 // ManifestStore returns a store for local manifests
 func (cli *DockerCli) ManifestStore() manifeststore.Store {
 	// TODO: support override default location from config file
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -278,6 +278,7 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string, 
 	return nil
 }
 
+//nolint:gocyclo
 func runDocker(dockerCli *command.DockerCli) error {
 	tcmd := newDockerCommand(dockerCli)
 
@@ -306,23 +307,42 @@ func runDocker(dockerCli *command.DockerCli) error {
 		}
 	}
 
+	var subCommand *cobra.Command
 	if len(args) > 0 {
 		ccmd, _, err := cmd.Find(args)
+		subCommand = ccmd
 		if err != nil || pluginmanager.IsPluginCommand(ccmd) {
 			err := tryPluginRun(dockerCli, cmd, args[0], envs)
+			if err == nil {
+				if dockerCli.HooksEnabled() && dockerCli.Out().IsTerminal() && ccmd != nil {
+					_ = pluginmanager.RunPluginHooks(dockerCli, cmd, ccmd, args[0], args)
+				}
+				return nil
+			}
 			if !pluginmanager.IsNotFound(err) {
+				// For plugin not found we fall through to
+				// cmd.Execute() which deals with reporting
+				// "command not found" in a consistent way.
 				return err
 			}
-			// For plugin not found we fall through to
-			// cmd.Execute() which deals with reporting
-			// "command not found" in a consistent way.
 		}
 	}
 
 	// We've parsed global args already, so reset args to those
 	// which remain.
 	cmd.SetArgs(args)
-	return cmd.Execute()
+	err = cmd.Execute()
+	if err != nil {
+		return err
+	}
+
+	// If the command is being executed in an interactive terminal,
+	// run the plugin hooks (but don't throw an error if something misbehaves)
+	if dockerCli.HooksEnabled() && dockerCli.Out().IsTerminal() && subCommand != nil {
+		_ = pluginmanager.RunPluginHooks(dockerCli, cmd, subCommand, "", args)
+	}
+
+	return nil
 }
 
 type versionDetails interface {
--- a/cli-plugins/hooks/printer_test.go
+++ b/cli-plugins/hooks/printer_test.go
@@ -1,0 +1,38 @@
+package hooks
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/morikuni/aec"
+	"gotest.tools/v3/assert"
+)
+
+func TestPrintHookMessages(t *testing.T) {
+	testCases := []struct {
+		messages       []string
+		expectedOutput string
+	}{
+		{
+			messages:       []string{},
+			expectedOutput: "",
+		},
+		{
+			messages: []string{"Bork!"},
+			expectedOutput: aec.Bold.Apply("\nWhat's next:") + "\n" +
+				"    Bork!\n",
+		},
+		{
+			messages: []string{"Foo", "bar"},
+			expectedOutput: aec.Bold.Apply("\nWhat's next:") + "\n" +
+				"    Foo\n" +
+				"    bar\n",
+		},
+	}
+
+	for _, tc := range testCases {
+		w := bytes.Buffer{}
+		PrintNextSteps(&w, tc.messages)
+		assert.Equal(t, w.String(), tc.expectedOutput)
+	}
+}
--- a/cli-plugins/hooks/template_test.go
+++ b/cli-plugins/hooks/template_test.go
@@ -1,0 +1,82 @@
+package hooks
+
+import (
+	"testing"
+
+	"github.com/spf13/cobra"
+	"gotest.tools/v3/assert"
+)
+
+func TestParseTemplate(t *testing.T) {
+	type testFlag struct {
+		name  string
+		value string
+	}
+	testCases := []struct {
+		template       string
+		flags          []testFlag
+		args           []string
+		expectedOutput string
+	}{
+		{
+			template:       "",
+			expectedOutput: "",
+		},
+		{
+			template:       "a plain template message",
+			expectedOutput: "a plain template message",
+		},
+		{
+			template: TemplateReplaceFlagValue("tag"),
+			flags: []testFlag{
+				{
+					name:  "tag",
+					value: "my-tag",
+				},
+			},
+			expectedOutput: "my-tag",
+		},
+		{
+			template: TemplateReplaceFlagValue("test-one") + " " + TemplateReplaceFlagValue("test2"),
+			flags: []testFlag{
+				{
+					name:  "test-one",
+					value: "value",
+				},
+				{
+					name:  "test2",
+					value: "value2",
+				},
+			},
+			expectedOutput: "value value2",
+		},
+		{
+			template:       TemplateReplaceArg(0) + " " + TemplateReplaceArg(1),
+			args:           []string{"zero", "one"},
+			expectedOutput: "zero one",
+		},
+		{
+			template:       "You just pulled " + TemplateReplaceArg(0),
+			args:           []string{"alpine"},
+			expectedOutput: "You just pulled alpine",
+		},
+	}
+
+	for _, tc := range testCases {
+		testCmd := &cobra.Command{
+			Use:  "pull",
+			Args: cobra.ExactArgs(len(tc.args)),
+		}
+		for _, f := range tc.flags {
+			_ = testCmd.Flags().String(f.name, "", "")
+			err := testCmd.Flag(f.name).Value.Set(f.value)
+			assert.NilError(t, err)
+		}
+		err := testCmd.Flags().Parse(tc.args)
+		assert.NilError(t, err)
+
+		out, err := ParseTemplate(tc.template, testCmd)
+		assert.NilError(t, err)
+		assert.Equal(t, out, tc.expectedOutput)
+	}
+}
--- a/cli-plugins/manager/error.go
+++ b/cli-plugins/manager/error.go
@@ -41,6 +41,9 @@ func (e *pluginError) MarshalText() (text []byte, err error) {
 // wrapAsPluginError wraps an error in a pluginError with an
 // additional message, analogous to errors.Wrapf.
 func wrapAsPluginError(err error, msg string) error {
+	if err == nil {
+		return nil
+	}
 	return &pluginError{cause: errors.Wrap(err, msg)}
 }
 
--- a/cli-plugins/manager/hooks_test.go
+++ b/cli-plugins/manager/hooks_test.go
@@ -1,0 +1,38 @@
+package manager
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestGetNaiveFlags(t *testing.T) {
+	testCases := []struct {
+		args          []string
+		expectedFlags map[string]string
+	}{
+		{
+			args:          []string{"docker"},
+			expectedFlags: map[string]string{},
+		},
+		{
+			args: []string{"docker", "build", "-q", "--file", "test.Dockerfile", "."},
+			expectedFlags: map[string]string{
+				"q":    "",
+				"file": "",
+			},
+		},
+		{
+			args: []string{"docker", "--context", "a-context", "pull", "-q", "--progress", "auto", "alpine"},
+			expectedFlags: map[string]string{
+				"context":  "",
+				"q":        "",
+				"progress": "",
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		assert.DeepEqual(t, getNaiveFlags(tc.args), tc.expectedFlags)
+	}
+}
--- a/cli/command/cli_test.go
+++ b/cli/command/cli_test.go
@@ -307,3 +307,56 @@ func TestInitializeShouldAlwaysCreateTheContextStore(t *testing.T) {
 	})))
 	assert.Check(t, cli.ContextStore() != nil)
 }
+
+func TestHooksEnabled(t *testing.T) {
+	t.Run("disabled by default", func(t *testing.T) {
+		cli, err := NewDockerCli()
+		assert.NilError(t, err)
+
+		assert.Check(t, !cli.HooksEnabled())
+	})
+
+	t.Run("enabled in configFile", func(t *testing.T) {
+		configFile := `{
+    "features": {
+      "hooks": "true"
+    }}`
+		dir := fs.NewDir(t, "", fs.WithFile("config.json", configFile))
+		defer dir.Remove()
+		cli, err := NewDockerCli()
+		assert.NilError(t, err)
+		config.SetDir(dir.Path())
+
+		assert.Check(t, cli.HooksEnabled())
+	})
+
+	t.Run("env var overrides configFile", func(t *testing.T) {
+		configFile := `{
+    "features": {
+      "hooks": "true"
+    }}`
+		t.Setenv("DOCKER_CLI_HOOKS", "false")
+		dir := fs.NewDir(t, "", fs.WithFile("config.json", configFile))
+		defer dir.Remove()
+		cli, err := NewDockerCli()
+		assert.NilError(t, err)
+		config.SetDir(dir.Path())
+
+		assert.Check(t, !cli.HooksEnabled())
+	})
+
+	t.Run("legacy env var overrides configFile", func(t *testing.T) {
+		configFile := `{
+    "features": {
+      "hooks": "true"
+    }}`
+		t.Setenv("DOCKER_CLI_HINTS", "false")
+		dir := fs.NewDir(t, "", fs.WithFile("config.json", configFile))
+		defer dir.Remove()
+		cli, err := NewDockerCli()
+		assert.NilError(t, err)
+		config.SetDir(dir.Path())
+
+		assert.Check(t, !cli.HooksEnabled())
+	})
+}
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -41,6 +41,7 @@ type ConfigFile struct {
 	CLIPluginsExtraDirs  []string                     `json:"cliPluginsExtraDirs,omitempty"`
 	Plugins              map[string]map[string]string `json:"plugins,omitempty"`
 	Aliases              map[string]string            `json:"aliases,omitempty"`
+	Features             map[string]string            `json:"features,omitempty"`
 }
 
 // ProxyConfig contains proxy configuration settings


cmd/docker/docker.go
```diff
@@ -259,7 +260,18 @@ func runDocker(dockerCli *command.DockerCli) error {
...
+		return err
+	}
+
+	// If the command is being executed in an interactive terminal,
+	// run the plugin hooks (but don't throw an error if something misbehaves)
+	if term.IsTerminal(dockerCli.Out().FD()) {
+		_ = pluginmanager.RunPluginHooks(dockerCli, cmd, args)
+	}
+
+	return nil
```
vvoland, line 274: Could we also pass an error to the hooks? This could allow hooks to react to known/common user errors and suggest fixes or correct usage.

laurazard: Oh, I didn't think about that but it's a good idea! Will have to investigate a little bit to see how much information we pass in the errors though – passing a lot of information about the command being executed into the plugin is part of what I was trying to avoid with the "templating" approach.

laurazard: thoughts @thaJeztah @eunomie ?

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,140 @@
...
+// `you ran a container named: my-container`
+func TemplateReplaceFlagValue(flag string) string {
+	return fmt.Sprintf(hookTemplateFlagValue, flag)
+}
+
+// RunPluginHooks calls the hook subcommand for all present
+// CLI plugins that declare support for hooks in their metadata
+// and parses/prints their responses
+func RunPluginHooks(dockerCli command.Cli, cmd *cobra.Command, args []string) error {
+	plugins, err := ListPlugins(dockerCli, cmd)
```
thaJeztah, line 68: Wondering if this is something that's already done before we reach this (i.e., don't we scan for plugins on startup somewhere?).

Considering if we already have a list of plugins we know about, if we can re-use that list (assuming here that listing / scanning for plugins is potentially "heavy".

thaJeztah: Ah, hm. Had quick look, and it appears that we don't call `ListPlugins` in all cases, but only when (e.g.) doing `docker --help` (to create "stubs"), and `docker info` (to list plugins).

However, in those cases, we _would_ be running twice, but in other cases, we'd run `PluginRunCommand` (and/or `GetPlugin`); both of which also run `getPluginDirs` and `addPluginCandidatesFromDir`, and now I'm wondering if we need to memoize some of that (at least, we should look at benchmarking some of it).

cmd/docker/docker.go
```diff
@@ -259,7 +260,18 @@ func runDocker(dockerCli *command.DockerCli) error {
...
 	cmd.SetArgs(args)
-	return cmd.Execute()
+	err = cmd.Execute()
+	if err != nil {
+		return err
+	}
+
+	// If the command is being executed in an interactive terminal,
+	// run the plugin hooks (but don't throw an error if something misbehaves)
+	if term.IsTerminal(dockerCli.Out().FD()) {
```
thaJeztah, line 270: can use `Out.IsTerminal()` for this one;

```suggestion
	if dockerCli.Out().IsTerminal() {
```

cmd/docker/docker.go
thaJeztah: Oh! Thought I replied to this one;

Yes, I think it could make sense to provide the hook access to that information. I can indeed see cases where (potentially) a hook could enrich errors as well, or it would allow experimenting with an out-of-band binary to improve error messages (and provide hints).

One thing I'm wondering is if we should change the signature of this function to accept some struct as last argument; that struct could contain "optional" information, which would allow adding more "context" in future, without having to change the signature.

Also considering if we need to pass a `context.Context` in the new functions; wdyt?

thaJeztah: I just noticed (some debugging) that we don't call hooks after plugin commands (?) was that intentional (i.e., is it expected for plugins to call hooks?) or something we overlooked?

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,140 @@
...
+	if len(matches) == 0 {
+		return template, false
+	}
+	for _, m := range matches[1:] {
+		flag := cmd.Flag(m)
+		if flag == nil {
+			continue
+		}
+		value := flag.Value.String()
+		template = strings.ReplaceAll(template, TemplateReplaceFlagValue(m), value)
```
thaJeztah, line 126: Should we use an actual Go template for this? (it just occurred to me that we are doing a DYI variant of that here?)

laurazard: When implementing this my thought was just to provide some simple, very easy to use API that plugins could call into to build their templates. We can definitely replace that with actual Go templates, though I do worry it might result in more complex implementation and, in that case, we should try to figure out how to best expose/make clear to Hooks API users what information they can/cannot access with Go templates.

Another option is to keep the current API and just use Go templates under the hood for implementation, which might make sense (but I'd need to dig into it a bit more).

cli-plugins/manager/hooks.go
laurazard: Yeah! Re: benchmarking, I worked up some general ones into the `#Performance Callouts` part of the PR description. For discussion/validation purposes, I think the implementation as-is works well, also given the rest of the comments in that section of the description – i.e., the performance penalty here only applies in interactive environments, etc.

However, I do think we should consider ways to improve performance here – and there, memoization might come in handy (or, as pointed out in that section, caching plugin metadata on disk.

cmd/docker/docker.go
laurazard: > One thing I'm wondering is if we should change the signature of this function to accept some struct as last argument; that struct could contain "optional" information, which would allow adding more "context" in future, without having to change the signature.

Could you provide an example? Re: `pluginmanager.RunPluginHooks`, I'm not sure what we would pass in for "optional" information right now, and if we need it in the future we can just change the signature, right? This is a function that we're calling from inside the CLI, not from plugins, so changing it's signature shouldn't be an ordeal.

> Also considering if we need to pass a context.Context in the new functions; wdyt?

I'm not sure. If we're talking about `pluginmanager.RunPluginHooks`, for now, the benefits it brings are limited – we don't need it inside, and we don't need it to be able to cancel execution, since it's getting called synchronously. However, if we have a misbehaving plugin, there might be some benefit to passing the context in so that we can pass it to the plugin command being executed to make it easier to kill it? Merits some looking into I guess.

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,140 @@
...
+
+// RunPluginHooks calls the hook subcommand for all present
+// CLI plugins that declare support for hooks in their metadata
+// and parses/prints their responses
+func RunPluginHooks(dockerCli command.Cli, cmd *cobra.Command, args []string) error {
+	plugins, err := ListPlugins(dockerCli, cmd)
+	if err != nil {
+		return err
+	}
+	subCommand, _, err := cmd.Find(args)
```
thaJeztah, line 72: I just noticed that we use `args` to resolve the subcommand (`ccmd`);

```go
	if len(args) > 0 {
		ccmd, _, err := cmd.Find(args)
```

But in `RunPluginHooks` we do that again;

```go
subCommand, _, err := cmd.Find(args)
```

Would it work if we'd pass `subCommand` as argument? _Could_ be low-hanging fruit to optimize (I have not benchmarked that though, but know that `.Find()` may be traversing all of Cobra).

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return err
+	}
+
+	hookMessages := make([]string, 0, len(plugins))
+	for _, p := range plugins {
+		p := p
+		if !p.supportsHookCommand(subCommand.Name()) {
```
thaJeztah, line 80: Perhaps we should store `Name()` in a variable; we're calling it in a loop here, and I noticed that it's not a plain "getter", but doing `strings.Index()` and generating the name on-the-fly.

Might be "micro-optimisation", but could be low-hanging fruit as well 😅

cmd/docker/docker.go
thaJeztah: > I'm not sure what we would pass in for "optional" information right now,

Yeah, perhaps not needed yet. I noticed the `args` and wondered what other information we may want to pass (and could already have available).

> for now, the benefits it brings are limited 
> if we have a misbehaving plugin

Was having both OTEL and misbehaving plugins in mind yes. Mostly OTEL (which could potentially allow us to measure performance as well?), but also things like passing a logger around, etc.

cli-plugins/manager/hooks.go
laurazard: "resolving" this one since we've since had other conversations re: plugin metadata caching and socket communication.

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	return template, true
+}
+
+func printHookMessages(out io.Writer, messages []string) {
+	if len(messages) == 0 {
+		return
+	}
+
+	fmt.Fprintln(out, aec.LightBlueF.Apply("\nNext steps:"))
```
vvoland, line 176: I like the idea of having different message types:
https://github.com/docker/cli/blob/10e34442b14df1fe57f689673f4f98901e49ae24/cli-plugins/manager/hooks.go#L27-L28

I think it makes sense to implement this now, with only one kind `NextSteps`, so that it's clear the the CLI is "the host" here and specifies the layout. If some plugin needs a different kind of layout, the CLI needs to be aware of that.

WDYT?

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,180 @@
...
+	for _, m := range messages {
+		fmt.Fprintf(out, "    - %s\n", m)
```
vvoland, line 178: Looks like "list" layout is enforced here. Should we also make sure that each message is 1-line only?

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,180 @@
...
+
+const (
+	hookTemplateCommandName = "{docker-cli-plugin-hook-cmd-name}"
+	hookTemplateFlagValue   = "{docker-cli-plugin-hook-flag-%s}"
+	hookTemplateArg         = "{docker-cli-plugin-hook-arg-%s}"
+)
+
+func parseHooks(template string, cmd *cobra.Command) string {
+	ret := template
+	for {
```
krissetto, line 115: nit: maybe we could leave a short comment here about why we run `parseFlagValue` and `parseArg` in a loop that we exit when we don't find any more matches, since the reason seems to be how the regex lib itself works inside of those funcs. WDYT?

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,180 @@
...
+	return nil
+}
+
+const (
+	hookTemplateCommandName = "{docker-cli-plugin-hook-cmd-name}"
+	hookTemplateFlagValue   = "{docker-cli-plugin-hook-flag-%s}"
+	hookTemplateArg         = "{docker-cli-plugin-hook-arg-%s}"
+)
+
+func parseHooks(template string, cmd *cobra.Command) string {
```
krissetto, line 113: nit: i think  `parseHookTemplate` is a better name here

cli-plugins/manager/hooks.go
krissetto: This could be nice to have, especially once we start working more on the cli ux itself. Not sure if we should consider the enum to represent the "type" of hook, but surely it'd be good to have some differentiation for ux purposes

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,180 @@
...
+// and parses/prints their responses
+func RunPluginHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command) error {
+	plugins, err := ListPlugins(dockerCli, rootCmd)
+	if err != nil {
+		return err
+	}
+	subCmdName := subCommand.Name()
+
+	hookMessages := make([]string, 0, len(plugins))
+	for _, p := range plugins {
```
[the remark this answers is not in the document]
laurazard: Good point! I do think that, at least for now, ordered hints should not be a concern – none of the current plugin authors expect their hints to happen in any specific order, and I'm not even sure it'll ever be a requirement. Your comment does make me think that we should call this out explicitly in the API and define the contract as (ordering is not guaranteed).
In the future, if necessary, we can introduce some sort of priority system, where plugins can request a certain priority. I think that can be bucketed in a larger body of work re: smarter hints.

cli-plugins/manager/hooks.go
laurazard: Agree! Will change.

cli-plugins/manager/hooks.go
laurazard: Yeah, I'm of that mind as well. I'll update the PR to remove the TODO and codify this in.

cli-plugins/manager/hooks.go
laurazard: Hmmmm, I'm generally of the opinion that comments explaining what code is doing (not why it is doing something) are a yellow flag: if we feel the need to do it, we should definitely consider whether the code couldn't be refactored to be clearer. I just added some hooks tests, so I'll make a run at a refactor here and if it's still unclear I'll add comments/change some stuff around.

cli-plugins/manager/hooks.go
krissetto: Maybe we could format each line of a message (after the first) so they at least start with the same indentaton? (a small terminal would wrap the lines, so in that case it'd look ugly anyway)

laurazard: > Maybe we could format each line of a message (after the first) so they at least start with the same indentaton?
```
Next steps:
    - Keep in mind that lorem ipsum dolor sit amet.
      (also, pellentesque ac eleifend tellus, ac malesuada dolor)
    - Nullam luctus rutrum fringilla
```
Like this? Not sure if this is clearer or not for me.

krissetto: Exactly! I think it looks much better (but that might just be a strong bias on my side 😅)

cli-plugins/manager/hooks.go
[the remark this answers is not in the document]
krissetto: Sounds good to me! Let's not mix up hooks with hints though, as I think hooks could go much further in terms of potential functionality. For hints, and for the current pr, i agree with calling out "execution order not guaranteed" in the api

cli-plugins/manager/hooks.go
eunomie: Today in `docker scout` we don't display them as list items, even if there's multiple items:
image

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,180 @@
...
+		i, err := strconv.Atoi(m)
+		if err != nil {
+			continue
+		}
+		if i >= len(cmd.Flags().Args()) {
+			return template, false
+		}
+		template = strings.ReplaceAll(template, TemplateReplaceArg(i), cmd.Flags().Arg(i))
+	}
+	return template, true
```
Benehiko, line 168: I'm not so sure about this, returning true here wouldn't necessarily mean that the matches we found were valid and actually had an impact on the template string.  

`matches` could have only 1 match, leading to the for being ignored, or `strconv` could've failed exiting the for loop (if no more matches) ending up with us returning the template string and `true`.

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,180 @@
...
+// Example:
+//
+// "run this image with `docker run " + TemplateReplaceArg(0) + "`"
+//
+// when being executed after the command:
+// `docker pull alpine`
+// will result in the message:
+// "Run this image with `docker run alpine`"
+func TemplateReplaceArg(i int) string {
+	return fmt.Sprintf(hookTemplateArg, strconv.Itoa(i))
```
Benehiko, line 75: Maybe we should accept a string here instead? Doesn't seem like using an int serves much purpose other than being converted to a string.

Benehiko (changes requested):

cli-plugins/manager/hooks.go
laurazard: Receiving an int serves to indicate to the caller that an int is expected to be used here. If an int were passed in, we'd ultimately need to attempt to convert to an int to ensure that the string represents an int, and return an error if it didn't to signal that the function is being used incorrectly. This negates any benefit associated with receiving a string.

cli-plugins/manager/plugin.go
```diff
@@ -100,3 +101,38 @@ func newPlugin(c Candidate, cmds []*cobra.Command) (Plugin, error) {
...
+
+	var hookMsg HookMessage
+	err = json.Unmarshal(hookCmdOutput, &hookMsg)
+	return hookMsg, wrapAsPluginError(err, "error parsing plugin hook message")
+}
+
+func (p *Plugin) supportsHookCommand(command string) bool {
+	if len(p.HookCommands) == 0 {
+		return false
+	}
```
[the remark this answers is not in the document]
laurazard: Oh! Good catch – this is a leftover of a different previous implementation, will remove.

cli-plugins/manager/hooks.go
laurazard: Did a large refactor here, PTAL and let me know if you like the new implementation better :)

cli-plugins/manager/hooks.go
laurazard: Same as below, did a larger refactor here (now under `cli-plugins/hooks/template.go`, PTAL and let me know if the current implementation is clearer.

cli-plugins/hooks/template.go
```diff
@@ -0,0 +1,127 @@
...
+	ret = parseArg(ret, cmd)
+	ret = strings.ReplaceAll(ret, hookTemplateCommandName, cmd.Name())
+	return ret
+}
+
+func parseFlagValue(hookTemplate string, cmd *cobra.Command) string {
+	matches := flagRegex.FindAllStringSubmatch(hookTemplate, -1)
+	for _, m := range matches {
+		flagTemplate := m[0]
+		flag := cmd.Flag(m[1])
```
Benehiko, line 105: maybe we should just make sure `m` is truly a length of 2 just in case.

cli-plugins/hooks/template.go
```diff
@@ -0,0 +1,127 @@
...
+		hookTemplate = strings.ReplaceAll(hookTemplate, flagTemplate, value)
+	}
+	return hookTemplate
+}
+
+func parseArg(hookTemplate string, cmd *cobra.Command) string {
+	matches := argRegex.FindAllStringSubmatch(hookTemplate, -1)
+	for _, m := range matches {
+		argTemplate := m[0]
+		argIndex, err := strconv.Atoi(m[1])
```
Benehiko, line 119: same here

Benehiko (changes requested):
The code is a lot clearer! thank you :)

cli-plugins/hooks/template.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+// HookMessage represents a plugin hook response. Plugins
+// declaring support for CLI hooks need to print a json
+// representation of this type when their hook subcommand
+// is invoked.
+type HookMessage struct {
+	Type     HookType
+	Template string
+}
```
thaJeztah, line 25: Do we expect a single invocation of the hook to return a single message, or could there be situations where we want multiple to be returned? (And in that case, should we have some struct with a slice of HookMessages to be returned?

cli-plugins/hooks/printer.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/morikuni/aec"
+)
+
+func PrintNextSteps(out io.Writer, messages []string) {
+	if len(messages) == 0 {
+		return
+	}
+	fmt.Fprintln(out, aec.LightBlueF.Apply("\nNext steps:"))
+	for _, n := range messages {
+		fmt.Fprintf(out, "    - %s\n", n)
```
thaJeztah, line 16: nit: as we're creating a wrapper here, should we appease the linters, and `_, _ = ` for these?

cli-plugins/hooks/template.go
laurazard: Not sure if that makes sense: the way the regex is constructed, we shouldn't get a match unless we also get a subexpression match. I'll confirm this is the case, and can add a comment in that case, but if that's the case I'd rather not add a check for something that is guaranteed not to happen.

cli-plugins/hooks/template.go
laurazard: same as above.

cli-plugins/manager/hooks.go
laurazard: I'll take a jab at it – I like the idea, but slightly concerned about totally breaking formatting with the indentation in smaller terminals.

cli-plugins/manager/hooks.go
laurazard: @thaJeztah switched to using Go templates! :)

tonistiigi:
```
~ $ time /tmp/docker-pr4376 version

real    0m0.162s
user    0m0.183s
sys     0m0.102s
~ $ time /tmp/docker-pr4376 version

real    0m0.209s
user    0m0.164s
sys     0m0.086s
~ $ time /tmp/docker-pr4376 version

real    0m0.145s
user    0m0.188s
sys     0m0.060s

~ $ time /tmp/docker-pr4376 version
real    0m0.153s
user    0m0.187s
sys     0m0.066s


~ $ time /tmp/docker-pr4376 version

real    0m0.019s
user    0m0.008s
sys     0m0.006s
~ $ time /tmp/docker-26.0.0-rc2 version

real    0m0.021s
user    0m0.010s
sys     0m0.005s

~ $ time /tmp/docker-26.0.0-rc2 version

real    0m0.019s
user    0m0.002s
sys     0m0.013s

~ $ time /tmp/docker-26.0.0-rc2 version

real    0m0.020s
user    0m0.006s
sys     0m0.009s

```

```

 $ time /tmp/docker-26.0.0-rc2 version

real    0m0.284s
user    0m0.014s
sys     0m0.010s
 $ sudo bash -c 'echo 3 > /proc/sys/vm/drop_caches'
 $ time /tmp/docker-26.0.0-rc2 version

real    0m0.318s
user    0m0.009s
sys     0m0.013s


 $ sudo bash -c 'echo 3 > /proc/sys/vm/drop_caches'
 $ time /tmp/docker-pr4376 version

real    0m8.053s
user    0m0.315s
sys     0m0.282s
 $ sudo bash -c 'echo 3 > /proc/sys/vm/drop_caches'
 $ time /tmp/docker-pr4376 version

real    0m8.151s
user    0m0.237s
sys     0m0.320s
```

Note that `hyperfine` does not show correct numbers as this feature is disabled in non-terminal in the current PR.

laurazard:
I called that out in the performance section of the PR description, and took steps to avoid that (running with a fake terminal), no?

Anyway, I'm looking from my phone but significant deviation from the numbers in the PR description, it's probably due to me pushing a version using Go Templates vs the previous regex approach. I meant to post a comment showing the comparison after I got back to my keyboard.

tonistiigi:
>  it's probably due to me pushing a version using Go Templates vs the previous regex approach.

I haven't debugged this but that seems unlikely. I do not have any actual hooks configured for anything so I guess there is no parsing. I'm just running empty defaults (where I think we should not have performance regressions). Some CPU overhead should be negligible compared to overhead running each plugin process for every command. The plugin itself should not matter more than that bigger binaries take more time to load to memory, before they can respond that they don't implement hooks.

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,61 @@
...
+
+// RunPluginHooks calls the hook subcommand for all present
+// CLI plugins that declare support for hooks in their metadata
+// and parses/prints their responses.
+func RunPluginHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command) error {
+	plugins, err := ListPlugins(dockerCli, rootCmd)
+	if err != nil {
+		return err
+	}
+	subCmdName := subCommand.Name()
```
tonistiigi, line 26: This line seems to cause a panic for command like `docker -v`

```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x104da7788]

goroutine 1 [running]:
github.com/docker/cli/vendor/github.com/spf13/cobra.(*Command).Name(0x105649148?)
	/Users/tonistiigi/gocode/src/github.com/docker/cli/vendor/github.com/spf13/cobra/command.go:1497 +0x18
github.com/docker/cli/cli-plugins/manager.RunPluginHooks({0x105649148, 0x140001b1400}, 0x0?, 0x14000463310?)
	/Users/tonistiigi/gocode/src/github.com/docker/cli/cli-plugins/manager/hooks.go:26 +0x44
main.runDocker(0x140001b1400)
	/Users/tonistiigi/gocode/src/github.com/docker/cli/cmd/docker/docker.go:336 +0x44c
main.main()
	/Users/tonistiigi/gocode/src/github.com/docker/cli/cmd/docker/docker.go:35 +0x98
```

laurazard:
Alright, I'm back at a keyboard now:
> Note that hyperfine does not show correct numbers as this feature is disabled in non-terminal in the current PR.

Right, that's why in the PR description I called used `script` to fake a terminal. I specifically called this out:
> The logic introduced in this PR to load/loop over plugins to invoke and collect their hook output only kicks in when the CLI binary is being invoked in an interactive terminal. As such, when invoking the CLI from a script, no performance impact is noticed:
> ```
> ⭑ hyperfine --prepare 'sudo sh -c "sync; echo 1 > /proc/sys/vm/drop_caches"' 'docker -v' './docker -v'
> Benchmark 1: docker -v
>   Time (mean ± σ):      19.3 ms ±   0.5 ms    [User: 4.9 ms, System: 13.2 ms]
>   Range (min … max):    18.6 ms …  20.3 ms    12 runs
>
> Benchmark 2: ./docker -v
>   Time (mean ± σ):      18.9 ms ±   0.6 ms    [User: 4.7 ms, System: 11.5 ms]
>   Range (min … max):    17.6 ms …  20.2 ms    40 runs
> 
> Summary
>   ./docker -v ran
>     1.02 ± 0.04 times faster than docker -v
> ```
> However, this doesn't apply when running normal commands from a terminal:
> ```
> ⭑ hyperfine --prepare 'sudo sh -c "sync; echo 1 > /proc/sys/vm/drop_caches"' 'script --return --quiet -c "docker -v" /dev/null
> ' 'script --return --quiet -c "./docker -v" /dev/null'
> Benchmark 1: script --return --quiet -c "docker -v" /dev/null
>   Time (mean ± σ):      34.2 ms ±   1.1 ms    [User: 4.6 ms, System: 15.3 ms]
>   Range (min … max):    31.9 ms …  36.4 ms    29 runs
> 
> Benchmark 2: script --return --quiet -c "./docker -v" /dev/null
>   Time (mean ± σ):      78.2 ms ±   9.3 ms    [User: 40.2 ms, System: 70.6 ms]
>   Range (min … max):    72.2 ms … 118.3 ms    22 runs
> 
> Summary
>   script --return --quiet -c "docker -v" /dev/null ran
>     2.29 ± 0.28 times faster than script --return --quiet -c "./docker -v" /dev/null
> ```
> In this scenario, on my machine I see an impact of, on average, ~44ms. The culprit here, is, mainly, the time it takes for the CLI to load/retrieve the metadata of all present CLI plugins, or more specifically, this will always be the same as the slowest CLI plugin. If I remove all other CLI plugins and re-run the benchmarks we get:
> 
> ```
> ⭑ hyperfine --prepare 'sudo sh -c "sync; echo 1 > /proc/sys/vm/drop_caches"' 'script --return --quiet -c "docker -v" /dev/null' 'script --return --quiet -c "./docker -v" /dev/null'
> Benchmark 1: script --return --quiet -c "docker -v" /dev/null
>   Time (mean ± σ):      32.5 ms ±   1.2 ms    [User: 5.2 ms, System: 13.6 ms]
>   Range (min … max):    30.1 ms …  35.1 ms    29 runs
> 
> Benchmark 2: script --return --quiet -c "./docker -v" /dev/null
>   Time (mean ± σ):      42.9 ms ±   2.0 ms    [User: 7.0 ms, System: 18.9 ms]
>   Range (min … max):    39.6 ms …  51.6 ms    31 runs
> 
> Summary
>   script --return --quiet -c "docker -v" /dev/null ran
>     1.32 ± 0.08 times faster than script --return --quiet -c "./docker -v" /dev/null
> which is a more reasonable average of ~10ms.
> ```
> 
> In my case, the worst offenders were Compose and Buildx, which take around 35ms to load.

Still, I can't replicate the 8 seconds that you ran into. In my environment:
```
 ⭑ time ./build/docker version
./build/docker version  0.06s user 0.12s system 126% cpu 0.141 total

⭑ time ./build/docker version
./build/docker version  0.07s user 0.06s system 188% cpu 0.067 total

 ⭑ sudo bash -c 'echo 3 > /proc/sys/vm/drop_caches'
⭑ time ./build/docker version

./build/docker version  0.06s user 0.17s system 119% cpu 0.194 total
```

It's a bit slower to be sure, but nowhere near your case.

cli-plugins/manager/hooks.go
laurazard: Thanks! I'll TAL

tonistiigi:
> script --return --quiet -c "docker -v" /dev/null ran

`docker -v` does not call into this code-path. Your latest comment uses `docker version` that does.

> Still, I can't replicate the 8 seconds that you ran into. In my environment:

This is from codespaces and can be replicated there. Just make sure to install desktop to get the plugins bundle as only compose&buildx are installed by default.

8 seconds is indeed unexpected and it doesn't fit linear scaling from plugins count(considering rc2 baseline takes 300ms on same mode), but I think any double-digit percentage (=measurable) performance regression should be avoided.

I added some debug on that drop_cache run. Clearly still coming from that function. I think previously I had one extra plugin leftover and that causes the 8s -> 6s difference.

```
03:41:47.268036 > RunPluginHooks
03:41:47.284216 < listPluginCandidates done 
03:41:47.284630 > newPlugin /usr/lib/docker/cli-plugins/docker-scout
03:41:47.284628 > newPlugin /usr/lib/docker/cli-plugins/docker-extension
03:41:47.284761 > newPlugin /usr/lib/docker/cli-plugins/docker-feedback
03:41:47.284873 > newPlugin /usr/lib/docker/cli-plugins/docker-init
03:41:47.284909 > newPlugin /usr/lib/docker/cli-plugins/docker-compose
03:41:47.284935 > newPlugin /usr/lib/docker/cli-plugins/docker-dev
03:41:47.285002 > newPlugin /usr/lib/docker/cli-plugins/docker-sbom
03:41:47.285086 > newPlugin /usr/lib/docker/cli-plugins/docker-buildx
03:41:47.285099 > newPlugin /usr/lib/docker/cli-plugins/docker-debug
03:41:51.179446 < newPlugin /usr/lib/docker/cli-plugins/docker-feedback done
03:41:51.300104 < newPlugin /usr/lib/docker/cli-plugins/docker-dev done
03:41:51.357775 < newPlugin /usr/lib/docker/cli-plugins/docker-init done
03:41:51.401657 < newPlugin /usr/lib/docker/cli-plugins/docker-extension done
03:41:51.820787 < newPlugin /usr/lib/docker/cli-plugins/docker-debug done
03:41:52.002857 < newPlugin /usr/lib/docker/cli-plugins/docker-sbom done
03:41:52.955011 < newPlugin /usr/lib/docker/cli-plugins/docker-buildx done
03:41:52.959200 < newPlugin /usr/lib/docker/cli-plugins/docker-compose done
03:41:52.963600 < newPlugin /usr/lib/docker/cli-plugins/docker-scout done
03:41:52.963630 > ListPlugins done
03:41:52.963644 < RunPluginHooks 5.695607869s

real    0m6.049s
user    0m0.214s
sys     0m0.222s
```

Benehiko (approved):
Looks a lot better! LGTM

cli-plugins/hooks/template.go
eunomie: Just to say it's quite common inside docker scout cli to display multiple hooks from a single command. It's different as it's not in the main CLI, but just to share the idea that multiple hooks can be interesting at some point.

cli-plugins/hooks/printer.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func PrintNextSteps(out io.Writer, messages []string) {
+	if len(messages) == 0 {
+		return
+	}
+	fmt.Fprintln(out, aec.LightBlueF.Apply("\nNext steps:"))
+	for _, n := range messages {
+		_, _ = fmt.Fprintf(out, "    - %s\n", n)
+	}
```
eunomie, line 17: Current implementation of hooks looks like this:
image

And with multiple messages:
image

Basically two main differences:
- `Next steps:` vs `What's Next?` and colors
- `-` in front of each lines

I can change the scout CLI to add `-` on each line, so we ensure it's done the same way everywhere.
About the title I wonder if `What's next?` isn't a bit better. But I'll defer to @jalonsogo

laurazard:
Replying here to provide a coherent response to the concerns brought up by @tonistiigi:

To properly test performance, I made the following changes to `cmd/docker/docker.go`:

```diff
diff --git a/cmd/docker/docker.go b/cmd/docker/docker.go
index 431b1e027..ca9142151 100644
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -306,7 +306,7 @@ func runDocker(dockerCli *command.DockerCli) error {
                }
        }

-       var subCommand *cobra.Command
+       // var subCommand *cobra.Command
        if len(args) > 0 {
                ccmd, _, err := cmd.Find(args)
                if err != nil || pluginmanager.IsPluginCommand(ccmd) {
@@ -318,7 +318,7 @@ func runDocker(dockerCli *command.DockerCli) error {
                        // cmd.Execute() which deals with reporting
                        // "command not found" in a consistent way.
                } else {
-                       subCommand = ccmd
+                       // subCommand = ccmd
                }
        }

@@ -332,9 +332,9 @@ func runDocker(dockerCli *command.DockerCli) error {

        // If the command is being executed in an interactive terminal,
        // run the plugin hooks (but don't throw an error if something misbehaves)
-       if dockerCli.Out().IsTerminal() && subCommand != nil {
-               _ = pluginmanager.RunPluginHooks(dockerCli, cmd, subCommand)
-       }
+       // if dockerCli.Out().IsTerminal() && subCommand != nil {
         _ = pluginmanager.RunPluginHooks(dockerCli, cmd, subCommand)
+       // }

        return nil
 }
```

And built a version, with all the same code, with and without hooks. After that, I benchmarked with:
```console
hyperfine --prepare 'sudo sh -c "sync; echo 1 > /proc/sys/vm/drop_caches"' './build/docker version' './build/docker-with-hooks version'
```

-----------------------

So, lets test this:
> This was with the plugins bundle that ships with Docker Desktop (9) as it clearly depends on the number of plugins in the discovery directories.

We'll test with a number of different plugins:
```go
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
  hints: Docker Hints (Docker Inc.)
  long: Docker Long Running Example Plugin (Docker Inc.)

Benchmark 1: ./build/docker version
  Time (mean ± σ):      28.9 ms ±   7.2 ms    [User: 3.5 ms, System: 11.9 ms]
  Range (min … max):    25.5 ms …  61.1 ms    23 runs

Benchmark 2: ./build/docker-with-hooks version
  Time (mean ± σ):      66.8 ms ±   3.6 ms    [User: 32.8 ms, System: 65.8 ms]
  Range (min … max):    64.1 ms …  78.1 ms    28 runs

Summary
  ./build/docker version ran
    2.31 ± 0.59 times faster than ./build/docker-with-hooks version
```
So that's <40ms difference. With more plugins:
```go
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
  hints: Docker Hints (Docker Inc.)
  long: Docker Long Running Example Plugin (Docker Inc.)


Benchmark 1: ./build/docker version
  Time (mean ± σ):      30.0 ms ±   5.8 ms    [User: 4.4 ms, System: 11.7 ms]
  Range (min … max):    24.8 ms …  52.3 ms    21 runs

Benchmark 2: ./build/docker-with-hooks version
  Time (mean ± σ):      71.9 ms ±   8.1 ms    [User: 38.3 ms, System: 83.8 ms]
  Range (min … max):    66.6 ms … 103.4 ms    22 runs

  Warning: The first benchmarking run for this command was significantly slower than the rest (103.4 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You are already using the '--prepare' option which can be used to clear caches. If you did not use a cache-clearing command with '--prepare', you can either try that or consider using the '--warmup' option to fill those caches before the actual benchmark.

Summary
  ./build/docker version ran
    2.40 ± 0.54 times faster than ./build/docker-with-hooks version
```
Huh that's about 40ms too. Lets try more:
```go
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
  hints: Docker Hints (Docker Inc.)
  long: Docker Long Running Example Plugin (Docker Inc.)


Benchmark 1: ./build/docker version
  Time (mean ± σ):      27.1 ms ±   1.8 ms    [User: 4.1 ms, System: 10.9 ms]
  Range (min … max):    25.2 ms …  35.7 ms    30 runs

Benchmark 2: ./build/docker-with-hooks version
  Time (mean ± σ):      71.9 ms ±   4.2 ms    [User: 45.8 ms, System: 90.9 ms]
  Range (min … max):    66.4 ms …  85.0 ms    31 runs

Summary
  ./build/docker version ran
    2.66 ± 0.24 times faster than ./build/docker-with-hooks version
```
It's still about the same. Almost seems like the performance penalty does not depend on the number of plugins, as we thought (from the original PR text):
> or more specifically, this will always be the same as the slowest CLI plugin.

Adding scout:
```
  debug: Get a shell into any image or container. (Docker Inc.)
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
  hints: Docker Hints (Docker Inc.)
  long: Docker Long Running Example Plugin (Docker Inc.)
  scout: Docker Scout (Docker Inc.)


Benchmark 1: ./build/docker version
  Time (mean ± σ):      27.3 ms ±   1.9 ms    [User: 3.8 ms, System: 11.0 ms]
  Range (min … max):    25.5 ms …  36.0 ms    26 runs

Benchmark 2: ./build/docker-with-hooks version
  Time (mean ± σ):      88.6 ms ±  10.6 ms    [User: 72.4 ms, System: 137.5 ms]
  Range (min … max):    81.0 ms … 134.6 ms    23 runs

Summary
  ./build/docker version ran
    3.25 ± 0.45 times faster than ./build/docker-with-hooks version
```
That's 60ms, 20ms slower than without. I guess that's because
```
~/.docker/cli-plugins/docker-scout docker-cli-plugin-metadata  0.05s user 0.01s system 116% cpu 0.058 total
```
whereas other plugins are a good bit faster
```
~/.docker/cli-plugins/docker-buildx  0.03s user 0.01s system 104% cpu 0.039 total
~/.docker/cli-plugins/docker-compose  0.02s user 0.02s system 110% cpu 0.037 total
~/.docker/cli-plugins/docker-sbom  0.01s user 0.02s system 121% cpu 0.022 total
```

This checks out with what we expect.

Currently, the performance penalty is **as we expect**, barring any strange edge-cases. As called out in the PR description, a mitigation (that we're planning on working on after merging this) for this issue is to cache plugin metadata so that we don't always have to execute every plugin to see if they support hooks.

I'll also be adding a very naive timeout to the plugin scanning logic, to make sure we cap that out and never let it go longer than 80 or 100ms

tonistiigi:
> It's still about the same. It seems like the performance penalty does not depend on the number of plugins, as we thought (from the original PR text):

You can see in your output how user/system gradually grows with every added plugin, spinning the CPU more each time. In the last example, the resource usage is 14x of baseline, 9x before scout is added.

As these plugins all run in their own goroutine, the real time depends on cpu count, OS's exec implementation(including malware detection mechanisms), prewarmed cache, system load, and other factors.

Docker is run by millions of users, with various os/arch/perf configurations. Afaik Desktop does not come with a free 10+ core machine. If you compare absolute numbers then obviously users with low perf and low cpu count would see the impact more (and CLI is released to some really low-powered machines) than machines with a lot of resources to spare. To compare the absolute numbers, you would need a big test matrix of all the OS(maybe even kernels & infra providers), different ranges of cpu counts and perf generations, memory pressure, invocation frequency etc. But we don't need to compare the absolute numbers, just that previously we needed resources to run 1 binary and now we need resources to run 10 (and more with any future update or extra plugin the user has added). Unless you can defy some laws of physics, the latter will be slower for the users.

> Adding scout:

Scout binary is more than 4x bigger than dev/debug/init/feedback etc small plugins. Minimally, the extra resources need to be spent to load these binaries into memory, but it is not impossible that there is some `init()` rule that can be optimized as well(and other bottlenecks for spawning binaries in parallel are likely OS implementation-specific). Compared to buildx, what is only slightly smaller, scout's metadata cmd takes 25ms vs 15ms. How much impact a bigger binary or `init()` function has when loading a big bundle of binaries may not be universal in different configurations. This does show another weakness of this approach though - should any other binary make a mistake or just be unnecessarily bloated, it slows down the whole docker experience for any command.

 
> Currently, the performance penalty is as we expect,

That's another position where we differ. To me, any measurable performance regression from spinning the CPU needlessly should be avoided. Comparing this to other historic regressions I've either fixed or reported in the past, I think this is the biggest one yet (at least if the high multi-core machines are excluded).

laurazard:
This (current) performance discussion is moot as we'll be making some implementation changes that allow us to not have to scan/execute all the plugins, but for discussion's sake:

> Docker is run by millions of users, with various os/arch/perf configurations. Afaik Desktop does not come with a free 10+ core machine. If you compare absolute numbers then obviously users with low perf and low cpu count would see the impact more (and CLI is released to some really low-powered machines) than machines with a lot of resources to spare.

Yes, it is. That still does not mean that we cannot incur a performance hit ever, especially when this is done as a step in the direction of overall improving performance.

> This does show another weakness of this approach though - should any other binary make a mistake or just be unnecessarily bloated, it slows down the whole docker experience for any command.

Maybe I wasn't clear in my previous reply, but this sentence – `I'll also be adding a very naive timeout to the plugin scanning logic, to make sure we cap that out and never let it go longer than 80 or 100ms` – refers to the plugin execution time as well.

> To me, any measurable performance regression from spinning the CPU needlessly should be avoided

I agree. Which is why we're implementing this – so we can remove the wrapper and the "needless cpu spinning" that it does.

laurazard:
Pushed a change reworking the plugin hook listing+scanning logic – instead of scanning the plugin folders and executing every candidate to check their metadata for hook support, we now expect plugins with hooks to be configured in the CLI's `config.json`. For example:

```json
{
  "auths": {},
  "plugins": {
    "hints": {
      "hooks": "pull,build"
    }
  }
}
```

The new logic looks only for plugins configured this way, which significantly reduces execution time (and makes it so that performance does not get worse with the amount of plugins bundled).

Applying the above patch (executes plugins even if not attached to a tty):

```go
Benchmark 1: ./build/docker version
  Time (mean ± σ):       9.3 ms ±   1.0 ms    [User: 3.7 ms, System: 3.1 ms]
  Range (min … max):     6.8 ms …  13.7 ms    211 runs

Benchmark 2: ./build/docker-without-hooks version
  Time (mean ± σ):       9.4 ms ±   2.0 ms    [User: 3.6 ms, System: 3.2 ms]
  Range (min … max):     6.8 ms …  40.6 ms    288 runs

Summary
  ./build/docker version ran
    1.01 ± 0.24 times faster than ./build/docker-without-hooks version
```

That is for `docker version`, which no plugins register hook from. Measuring w/ a plugin enabled – with the example plugin, and instrumenting `RunPluginHooks` to time it's execution:

```go
 ./build/docker pull alpine
Using default tag: latest
Digest: sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b
Status: Image is up to date for alpine:latest
docker.io/library/alpine:latest

Next steps:
    - Run this image with `docker run {docker-cli-plugin-hook-arg-0}`
RunPluginHooks took 12.409638ms
...
RunPluginHooks took 8.642533ms
...
RunPluginHooks took 14.853688ms
```

from end to end, including calling the hook plugin, which is a lot more reasonable.

Still want to do a quick run through of some codebases to check no plugin is going to break if we do this.

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/spf13/cobra"
+)
+
+// HookPluginData is the type representing the information
+// that plugins declaring support for hooks get passed when
+// being invoked following a CLI command execution.
+type HookPluginData struct {
+	RootCmd string
+	Flags   []string
+}
```
joe0BAB, line 18: Sry if I'm missing something, but couldn't spot any clarification about this. How can hooks access arguments that are not passed as flag? Eg `docker run -it my-app sh` -> how can the hook access `my-app` and `sh` or `docker pull nginx` how can the hook access `nginx` value? Or is this passed as `RootCmd` string?

joe0BAB: Ok, `TemplateReplaceArg` does the trick I guess, marking as resolved..

neersighted (approved):

cli/command/cli.go
```diff
@@ -187,6 +187,36 @@ func (cli *DockerCli) BuildKitEnabled() (bool, error) {
...
+	// legacy support DOCKER_CLI_HINTS env var
+	if v := os.Getenv("DOCKER_CLI_HINTS"); v != "" {
+		enabled, err := strconv.ParseBool(v)
+		if err != nil {
+			return false
+		}
+		return enabled
+	}
+	// use DOCKER_CLI_HOOKS env var value if set and not empty
+	if v := os.Getenv("DOCKER_CLI_HOOKS"); v != "" {
```
krissetto, line 201: @thaJeztah same argument for the env var name as you mentioned [here](https://github.com/docker/cli/pull/4960)?

krissetto (approved):

cli/command/cli.go
laurazard: no, different one – this is user facing, used to override/enable/disable hooks if the user wants.

cli-plugins/manager/hooks.go
```diff
@@ -0,0 +1,128 @@
...
+	Flags   map[string]string
+}
+
+// RunPluginHooks calls the hook subcommand for all present
+// CLI plugins that declare support for hooks in their metadata
+// and parses/prints their responses.
+func RunPluginHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command, plugin string, args []string) error {
+	subCmdName := subCommand.Name()
+	if plugin != "" {
+		// TODO: handle plugin alias (build = buildx)
```
Benehiko, line 27: is this `TODO` for a follow up PR? Maybe link an issue if the intent is to do it in the _future_?

Benehiko (approved):
LGTM

cli/command/cli.go
krissetto: Are we sure? Hints are user facing, not hooks (even though they are the same thing now, i still think the naming could be problematic)

laurazard: > Are we sure?

I am, but maybe I'm missing something. Can you tell me why you believe this env var to enable/disable hooks shouldn't be user facing?

krissetto: Because plugins themselves aren't __really__ user facing, that's my worry here. Also, I can imagine that a user might want to disable hints without disabling other "hooks" functionality (admittedly, we don't have other functionality right now, but even just the name of the settings I need to change being HOOKS over HINTS would be confusing to me as a user)

krissetto: So for the user to be aware of hooks, they should also be aware of the cli plugins, or else we're just adding to the confusion IMHO

laurazard: Users who would be aware of this env var would also be aware of hooks. Also, clearly we want to advertise a way to turn this off, to offer an easy escape hatch if we accidentally bork things with hooks. This is very different from an env var that a user would "never" touch, such as the one we were discussing with the sockets.

laurazard: > So for the user to be aware of hooks, they should also be aware of the cli plugins, or else we're just adding to the confusion IMHO

Also, some users are. We sometimes tell users to download x binary of Compose and place it in the `~/.docker/cli-plugins` folder if they quickly need a new version. Most users aren't but also this is a public fact about how the docker CLI works. There's separate packages in debian repos, etc.

laurazard: > Also, I can imagine that a user might want to disable hints without disabling other "hooks" functionality

Also, this is [supported](https://github.com/docker/cli/pull/4376#issuecomment-2008580277). Would be a matter of removing the `"hooks"` line for a particular plugin.

krissetto: > Users who would be aware of this env var would also be aware of hooks. Also, clearly we want to advertise a way to turn this off, to offer an easy escape hatch if we accidentally bork things with hooks. This is very different from an env var that a user would "never" touch, such as the one we were discussing with the sockets.

Fair, so maybe in the future we could add an env var specific to turning off HINTS if we end up adding more functionality to hooks in general (the annoyance to the user was what i was imagining the env var was for, not the scenario of hooks borking stuff)

krissetto: > Also, this is https://github.com/docker/cli/pull/4376#issuecomment-2008580277. Would be a matter of removing the "hooks" line for a particular plugin.

Argh true..I still have to wrap my head around all these configs, as i'm not yet totally convinced of using env vars for some things, the config.json for others, etc especially when the features are so closely related. I'll resolve this thread

laurazard: > (the annoyance to the user was what i was imagining the env var was for, not the scenario of hooks borking stuff)

This isn't even a huge point because it'll ship off by default, and the general "off/on" switch will be the `features` map in `config.json`
```json
{
  "auths": {},
  "features": {
    "hooks": "true",
  },
  "plugins": {
    "hints": {
      "hooks": "pull,build"
    }
  }
}
```

the env var just provides the same ability, but it can be more useful to set in some environments/it's also just good practice to provide env vars to override config for these things.

laurazard: But yeah, I think the rest of that we can consider in the future when we have more substantial things to figure out and hooks are doing more things.

cli-plugins/manager/hooks.go
laurazard: Good callout! I put that there to remember to take care of it and then forgot 😅 glad you caught it. I'll create a ticket and clean it up.

cli/command/cli.go
laurazard: > as i'm not yet totally convinced of using env vars for some things, the config.json for others, etc especially when the features are so closely related.

I'm happy to continue this conversation elsewhere though, but there shouldn't be confusion here – there is only one env var being introduced `DOCKER_CLI_HOOKS`, which is a general override regardless of what is configured.

`DOCKER_CLI_HINTS` was carried over from the previous implementation, and does the same thing ("big override button"). Other than that (those are big overrides, not configuration), things are configured in the `config.json`.

cli-plugins/manager/metadata.go
```diff
@@ -22,4 +27,6 @@ type Metadata struct {
 	ShortDescription string `json:",omitempty"`
 	// URL is a pointer to the plugin's homepage.
 	URL string `json:",omitempty"`
+	// Hooks specifies whether the plugin supports CLI hooks
+	HookCommands []string `json:",omitempty"`
```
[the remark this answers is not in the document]
laurazard: Resolving this since the implementation has drastically changed since, and also since I think is is good conversation to have re: a v2 of hooks.

PR #4378, opened 2023-06-26
rename and update mergo module

### LINKED ISSUE #4377: github.com/imdario/mergo module renamed to dario.cat/mergo
### Description

Please rename `github.com/imdario/mergo` to their new vanity URL `dario.cat/mergo`, v1.0.0. Since you are using old module name, this prevents "go get -u all" to work correctly in projects that depend on `github.com/docker/cli`. This is not a bug so I reported it here in features request, but it would be nice if you can fix it.

acim:
**- What I did**

I renamed module `github.com/imdario/mergo` to `dario.cat/mergo` and updated to the latest version. This fixes #4377.

**- How I did it**

I renamed the import in `merge.go` and `merge_test.go` and ran `make vendor.`

**- How to verify it**

Run tests.


thaJeztah:
Thanks! I'm wondering what motivated the move to a vanity URL; I hope there's good hosting behind it, as I've had bad experience with vanity URLs not always being very reliable

acim:
Well, I don't know, I am not the author of that module.

neersighted (approved):
It's not quite inherently safe since we do (among other things) check that vendoring is correct in CI, which requires network access. That being said, we should allow `GOPROXY` for that -- in the past we had it disabled for `$REASONS`, but it is designed to solve this problem (validation/fetching, assuming you have an up-to-date `go.sum`).

Basically, any issues we have are largely self-inflicted and we can work them out. LGTM!

thaJeztah:
Yeah, it's definitely not a blocker, just bad experience with vanity URLs going AWOL. While `GOPROXY` is useful for temporary downtime of those, it's not a permanent fix; I've had situations where vanity URLs actually went down (permanently), in which case it's a ticking time bomb until `GOPROXY`'s cache expires. Of course using the upstream code repository (github.com) does not help for situations where project source is actually removed, but for other cases I have more trust in github.com's availability / SLA, than domains hosted by a single individual.

thaJeztah (approved):
LGTM

darccio:
@thaJeztah Allow me to indicate that the hosting is CloudFlare, and I plan to keep the domain. It shouldn't be an issue.

PR #4379, opened 2023-06-27
fixes #2413 - cli/command: fix docstring for ContainerFormat.CreatedAt and add CreatedAtTimestamp

### LINKED ISSUE #2413: ContainerFormat's CreatedAt has wrong documentation
https://github.com/docker/cli/blob/389fa742ff0ebfeda5d94735ac0ca09eac8ecc4f/cli/command/formatter/container.go#L179-L182

time.Unix().String() will not create a unix timestamp (number), but instead will create an string as explained in the documentation: https://golang.org/pkg/time/#Time.String .

Maybe add another variant, eg CreatedAtTimestamp to support this option nonetheless.

JonathanJamesRasmussen:
**- What I did**
Fixed docstring error and added function as described in https://github.com/docker/cli/issues/2413

**- How I did it**
Same as above

**- How to verify it**
See code diff

Fixed ContainerFormat.CreatedAt() docstring and added ContainerFormat.CreatedAtTimestamp()

baby turtle

### DIFF
--- a/cli/command/formatter/container.go
+++ b/cli/command/formatter/container.go
@@ -189,9 +189,14 @@ func (c *ContainerContext) Command() string {
 	return strconv.Quote(command)
 }
 
-// CreatedAt returns the "Created" date/time of the container as a unix timestamp.
+// CreatedAt returns the "Created" date/time of the container as a stringified unix timestamp.
 func (c *ContainerContext) CreatedAt() string {
 	return time.Unix(c.c.Created, 0).String()
+}
+
+// CreatedAtTimestamp returns the "Created" date/time of the container as a unix timestamp.
+func (c *ContainerContext) CreatedAtTimestamp() time.Time {
+	return time.Unix(c.c.Created, 0)
 }
 
 // RunningFor returns a human-readable representation of the duration for which


cli/command/formatter/container.go
```diff
@@ -189,11 +189,16 @@ func (c *ContainerContext) Command() string {
...
 
-// CreatedAt returns the "Created" date/time of the container as a unix timestamp.
+// CreatedAt returns the "Created" date/time of the container as a stringified unix timestamp.
 func (c *ContainerContext) CreatedAt() string {
 	return time.Unix(c.c.Created, 0).String()
 }
 
+// CreatedAtTimestamp returns the "Created" date/time of the container as a unix timestamp.
+func (c *ContainerContext) CreatedAtTimestamp() time.Time {
+	return time.Unix(c.c.Created, 0)
```
thaJeztah, line 199: I wonder if this should just return `c.Created` as-is (and perhaps the function be named `Created()`, and probably needs an entry in the "headers" (when used for table-view). Do we need to look at other formatters?

I'm slightly considering if we should have an option to expose the Container itself for formatting (have ` .Container()` function that returns `c.c`)

JonathanJamesRasmussen: Good point. What benefit would there be to simply returning the `Container` directly as opposed to presenting these APIs? That feels like a potential leaky abstraction to me, but that's just a gut feeling.

I could definitely see these as redundant since the `Container` does have `Created` accessible, but having other modules do the formatting seems like we could end up with a lot of duplication unless we had an interface to standardize that. Thoughts?

PR #4281, opened 2023-05-10
cli/config: remove warning for deprecated ~/.dockercfg file

thaJeztah:
### cli/config: Load: remove outdated FIXME

This FIXME was added in 2013 in https://github.com/moby/moby/commit/c72ff318d3fb6ab887af21cf4410e1bc8104663a
and it's both unclear which "internal golang config parser" is referred to here. Given that 10 Years have passed, this will unlikely happen, and doesn't warrant a FIXME here.

### cli/config: remove warning for deprecated ~/.dockercfg file

The `~/.dockercfg` file was replaced by `~/.docker/config.json` in 2015
(github.com/docker/docker/commit/18c9b6c6455f116ae59cde8544413b3d7d294a5e).

Commit b83bc67136e1edce2bdc225863af83d45d5472c9 (https://github.com/docker/cli/pull/2666) (v23.0.0, but backported to v20.10) added a warning if no "current" config file was found but a legacy file was, and if the CLI would fall back to using the deprecated file.

Commit ee218fa89e58c7296805fe7396cee959d74ce64b (https://github.com/docker/cli/pull/2504) removed support for the legacy file, but kept a warning in place if a legacy file was in place, and now ignored.

This patch removes the warning as well, fully deprecating the legacy
`~/.dockercfg` file.

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #4384, opened 2023-06-28
cli/config: add synchronisation for configDir (Dir, SetDir)

thaJeztah:
commit 8a30653ed5276179cc4dc0978af13ea2068c7d4e (https://github.com/docker/cli/pull/2161) introduced a sync.Once to allow for the config-directory (and home-dir) to be looked up lazily instead of in an `init()`.

However, the package-level `configDir` variable can be set through two separate paths; implicitly (through `config.Dir()`), and explicitly, through `config.SetDir()`. The existing code had no synchronisation for this, which could lead to a potential race-condition (code requesting `config.Dir()` and code setting a custom path through `config.SetDir()`).

This patch adds synchronisation by triggering the `sync.Once` as part of `config.SetDir()` to prevent it being triggered later (overwriting the value that was set). It also restores the `resetConfigDir()` utility that was removed in 379122b033774d00b3d223fee8d62b0ed4b917e3 (https://github.com/docker/cli/pull/4281), to allow resetting the `sync.Once` for this test.

In general, we should get rid of this package-level variable, and store it as a config on the client (passing the option to locations where its used instead).

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -27,6 +27,13 @@ var (
 	configDir     string
 )
 
+// resetConfigDir is used in testing to reset the "configDir" package variable
+// and its sync.Once to force re-lookup between tests.
+func resetConfigDir() {
+	configDir = ""
+	initConfigDir = new(sync.Once)
+}
+
 // Dir returns the directory the configuration file is stored in
 func Dir() string {
 	initConfigDir.Do(func() {
@@ -45,6 +52,8 @@ func ContextStoreDir() string {
 
 // SetDir sets the directory the configuration file is stored in
 func SetDir(dir string) {
+	// trigger the sync.Once to synchronise with Dir()
+	initConfigDir.Do(func() {})
 	configDir = filepath.Clean(dir)
 }
 
--- a/cli/config/config_test.go
+++ b/cli/config/config_test.go
@@ -382,3 +382,12 @@ func TestConfigPath(t *testing.T) {
 
 	SetDir(oldDir)
 }
+
+// TestSetDir verifies that Dir() does not overwrite the value set through
+// SetDir() if it has not been run before.
+func TestSetDir(t *testing.T) {
+	const expected = "my_config_dir"
+	resetConfigDir()
+	SetDir(expected)
+	assert.Check(t, is.Equal(Dir(), expected))
+}


cli/config/config.go
```diff
@@ -45,6 +52,8 @@ func ContextStoreDir() string {
 
 // SetDir sets the directory the configuration file is stored in
 func SetDir(dir string) {
+	// trigger the sync.Once to synchronise with Dir()
+	initConfigDir.Do(func() {})
```
vvoland, line 56: Didn't look too closely how these are called, but if both `Dir` and `SetDir` are called at the same time from different goroutines, we still might have a race condition:


1. Thread X enters `SetDir`
```go
56: initConfigDir.Do(func() {})
```

2. Thread Y enters `Dir` 
```go
39: initConfigDir.Do(func() {
	configDir = os.Getenv("DOCKER_CONFIG")
		if configDir == "" {
			configDir = filepath.Join(homedir.Get(), configFileDir)
		}
	})
	return configDir
}
```

3. Thread X enters the `Do` and "locks" `initConfigDir` first.
4. Thread Y is blocked and waits for X to finish the `Do` block.
5. Thread X leaves the `initConfigDir` and unblocks Y, but doesn't enter the `57: configDir = filepath.Clean(dir)` yet
6. Thread Y executes `45: return configDir` and returns empty directory.

vvoland: Can be ignored if we definitely don't (and won't) call these concurrently though.

thaJeztah: I don't _think_ we would hit that case in our code, so I'm "fine" with it not 100% being race-free (per our discussion on slack; there's no deadlock, so not "worse" than before).

FWIW; I went through a couple of options;

- a mutex for the value
- using `sync/atomic.Value`

Both of those performed a lot less, and both are still somewhat tapering over the "big elephant in the room"; we shouldn't use a package-level variable for this.

So I'm also looking at removing this variable altogether in favour of this being a config for the client that's configured, as (looking at the code) that's where it's used, and I think the expectation is to set the location _for the client_ that's constructed; https://github.com/docker/cli/blob/a26e6011f076481679ebd003eebfc988e55f0295/cli/command/cli.go#L209-L221

vvoland (approved):

PR #4388, opened 2023-06-28
cli: move "config" flag to cli/flags/ClientOptions.InstallFlags()

thaJeztah:
This flag was kept separate from the other flags, because at the time, the CLI code and Daemon code still used the same codebase, and shared some parts. This option only applied to the `docker` CLI, and thus was kept separate when migrating to Cobra in https://github.com/moby/moby/commit/0452ff5a4dd1b8fba707235d6f12a4038208f34b (https://github.com/moby/moby/pull/25354)

Now that this code is only used for the CLI (and plugins), we can move this flag together with the other flags.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -9,7 +9,6 @@ import (
 
 	pluginmanager "github.com/docker/cli/cli-plugins/manager"
 	"github.com/docker/cli/cli/command"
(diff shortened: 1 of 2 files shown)


neersighted (approved):
LGTM

PR #4389, opened 2023-06-28
cli: SetupRootCommand: remove redundant flags return

thaJeztah:
- [x] follow-up to https://github.com/docker/cli/pull/4388

The flag-set that was returned is a pointer to the command's Flags(), which
is in itself passed by reference (as it is modified / set up).

This patch removes the flags return, to prevent assuming it's different than
the command's flags.

While SetupRootCommand is exported, a search showed that it's only used internally,
so changing the signature should not be a problem.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli-plugins/plugin/plugin.go
+++ b/cli-plugins/plugin/plugin.go
@@ -131,7 +131,7 @@ func newPluginCommand(dockerCli *command.DockerCli, plugin *cobra.Command, meta 
 			DisableDescriptions: true,
 		},
 	}
-	opts, flags := cli.SetupPluginRootCommand(cmd)
+	opts, _ := cli.SetupPluginRootCommand(cmd)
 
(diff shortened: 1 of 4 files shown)


cli-plugins/plugin/plugin.go
```diff
@@ -131,7 +131,7 @@ func newPluginCommand(dockerCli *command.DockerCli, plugin *cobra.Command, meta
 			DisableDescriptions: true,
 		},
 	}
-	opts, flags := cli.SetupPluginRootCommand(cmd)
+	opts, _ := cli.SetupPluginRootCommand(cmd)
```
thaJeztah, line 134: FWIW; it could also be removed from `SetupPluginRootCommand()` but that one has external consumers, so I didn't want to break them in this PR 😅

Issue #4356, opened 2023-06-14
Flaky test: TestCloseWhileWriting, TestCloseWhileReading


thaJeztah:
### Description

- tests were added in https://github.com/docker/cli/pull/4226

https://github.com/docker/cli/actions/runs/5271207127/jobs/9531750668?pr=4355

```
--- FAIL: TestCloseWhileWriting (0.00s)
    commandconn_unix_test.go:181: assertion failed: expected error to contain "EOF", got "write |1: file already closed"
--- FAIL: TestCloseWhileReading (0.00s)
    commandconn_unix_test.go:211: assertion failed: expected error to contain "EOF", got "read |0: file already closed"
```

thaJeztah:
Also seen failing on Linux https://github.com/docker/cli/pull/4366#issuecomment-1600529828

PR #4394, opened 2023-06-30
commandconn: return original error while closing

laurazard:
- fixes https://github.com/docker/cli/issues/4356

**- What I did**

Fixes the flaky `TestCloseWhileWriting` and `TestCloseWhileReading` tests.

**- How I did it**

Changes the `Read` and `Write` error handling logic to return the original error while closing the connection. We still skip calling `handleEOF` if already closing the connection.

**- How to verify it**

`go test -v ./...`, and make sure everything still works.

### DIFF
--- a/cli/connhelper/commandconn/commandconn.go
+++ b/cli/connhelper/commandconn/commandconn.go
@@ -163,9 +163,8 @@ func (c *commandConn) Read(p []byte) (int, error) {
 	// Close might get called
 	if c.closing.Load() {
 		// If we're currently closing the connection
-		// we don't want to call onEOF, but we do want
-		// to return an io.EOF
-		return 0, io.EOF
+		// we don't want to call onEOF
+		return n, err
 	}
 
 	return n, c.handleEOF(err)
@@ -178,9 +177,8 @@ func (c *commandConn) Write(p []byte) (int, error) {
 	// Close might get called
 	if c.closing.Load() {
 		// If we're currently closing the connection
-		// we don't want to call onEOF, but we do want
-		// to return an io.EOF
-		return 0, io.EOF
+		// we don't want to call onEOF
+		return n, err
 	}
 
 	return n, c.handleEOF(err)
--- a/cli/connhelper/commandconn/commandconn_unix_test.go
+++ b/cli/connhelper/commandconn/commandconn_unix_test.go
@@ -5,6 +5,7 @@ package commandconn
 import (
 	"context"
 	"io"
+	"io/fs"
 	"testing"
 	"time"
 
@@ -178,7 +179,8 @@ func TestCloseWhileWriting(t *testing.T) {
 	assert.Check(t, !process.Alive(cmdConn.cmd.Process.Pid))
 
 	writeErr := <-writeErrC
-	assert.ErrorContains(t, writeErr, "EOF")
+	assert.ErrorContains(t, writeErr, "file already closed")
+	assert.Check(t, is.ErrorIs(writeErr, fs.ErrClosed))
 }
 
 func TestCloseWhileReading(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cli/connhelper/commandconn/commandconn.go
```diff
@@ -178,9 +177,8 @@ func (c *commandConn) Write(p []byte) (int, error) {
 	// Close might get called
 	if c.closing.Load() {
 		// If we're currently closing the connection
-		// we don't want to call onEOF, but we do want
-		// to return an io.EOF
-		return 0, io.EOF
+		// we don't want to call onEOF
+		return n, err
```
thaJeztah, line 181: Err may be `nil` here; do we need to return a `io.EOF` in that case? (Thinking out loud if we need an `io.EOF` for calling code)

```go
if err == nil {
    return io.EOF
}
```

laurazard: No, if the read didn't return an error then it's fine to not return one here. This was the original behavior, and subsequent reads will fail regardless with a file closed/pipe closed error.

laurazard: In normal operation, it will only be closing if an EOF was already returned triggering the transport layer to call Close regardless. This special handling is only meant to keep us from calling handleEOF twice.

thaJeztah: Thanks for looking! The original comment triggered my curiosity, so wanted to check 😅  👍

cli/connhelper/commandconn/commandconn_unix_test.go
```diff
@@ -178,7 +178,7 @@ func TestCloseWhileWriting(t *testing.T) {
 	assert.Check(t, !process.Alive(cmdConn.cmd.Process.Pid))
 
 	writeErr := <-writeErrC
-	assert.ErrorContains(t, writeErr, "EOF")
+	assert.ErrorContains(t, writeErr, "file already closed")
```
thaJeztah, line 181: Should we be using one of these for this?

https://github.com/golang/go/blob/18e17e2cb12837ea2c8582ecdb0cc780f49a1aac/src/io/fs/fs.go#L141
https://github.com/golang/go/blob/18e17e2cb12837ea2c8582ecdb0cc780f49a1aac/src/os/error.go#L24


```suggestion
	assert.Check(t, is.ErrorIs(writeErr, fs.ErrClosed))
```

cli/connhelper/commandconn/commandconn_unix_test.go
```diff
@@ -208,5 +208,5 @@ func TestCloseWhileReading(t *testing.T) {
 	assert.Check(t, !process.Alive(cmdConn.cmd.Process.Pid))
 
 	readErr := <-readErrC
-	assert.ErrorContains(t, readErr, "EOF")
+	assert.ErrorContains(t, readErr, "file already closed")
```
thaJeztah, line 211:
```suggestion
	assert.Check(t, is.ErrorIs(readErr, fs.ErrClosed))
```

thaJeztah (commented):
minor comment on the test, but otherwise LGTM

thaJeztah (approved):
I applied my suggestions in the test

LGTM

laurazard:
Thanks! I was away from a keyboard, appreciate it ❤️

thaJeztah:
Preparing a cherry-pick; will be up in a minute 👍

thaJeztah:
- Backport is up; https://github.com/docker/cli/pull/4395

vvoland:
Looks like it's still flaky - I just got a test failure in a freshly rebased PR: https://github.com/docker/cli/actions/runs/5454066267/jobs/9923693780?pr=4331

```
 85.59 === Failed
85.59 === FAIL: cli/connhelper/commandconn TestCloseWhileWriting (0.00s)
85.59     commandconn_unix_test.go:182: assertion failed: expected error to contain "file already closed", got "write |1: broken pipe"
```

\cc @laurazard

PR #4397, opened 2023-07-03
Add hyperlink to Formatting

dapeleg-dn:
**- What I did**
Change the text of `https://docs.docker.com/go/formatting/` to a hyperlink.

**- How I did it**
Change the text to Markdown hyperlink format.

**- How to verify it**
Make sure you see the text as a link and that it's clickable and takes you to the desired page.

Change the text of `https://docs.docker.com/go/formatting/` to a hyperlink.


docs/reference/commandline/config_inspect.md
```diff
@@ -7,7 +7,7 @@ Display detailed information on one or more configs
...
 |:---------------------------------------|:---------|:--------|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| [`-f`](#format), [`--format`](#format) | `string` |         | Format output using a custom template:<br>'json':             Print in JSON format<br>'TEMPLATE':         Print output using the given Go template.<br>Refer to https://docs.docker.com/go/formatting/ for more information about formatting output with templates |
+| [`-f`](#format), [`--format`](#format) | `string` |         | Format output using a custom template:<br>'json':             Print in JSON format<br>'TEMPLATE':         Print output using the given Go template.<br>Refer to [Formatting](https://docs.docker.com/go/formatting/) for more information about formatting output with templates |
```
thaJeztah, line 10: This part of the markdown is generated from source-code, and in this case the URL is part of the flag description.

We'll probably need to think how to fix this specific case; on the command-line (assuming we want to include the URL), we need to print the URL "as-is", but in the generated documentation, we want to make the link clickable.

The alternative could be to 

- _not_ include a link as part of the flag description
- make sure all commands have at least a basic section in the docs that describes the flag (like on https://docs.docker.com/engine/reference/commandline/info/)
- and then have a link inside that section that links to the formatting section
- and the `--help` output for the command to print a link to the reference docs

dapeleg-dn: Thanks, @thaJeztah, for your comment. I am unfamiliar with this project, but if these files of the documentation section are generated from the source code, why are they committed to Git instead of just being generated on-the-fly as part of the CI?

dapeleg-dn: And, by the way, I came to this fix from the `edit this page` link on `https://docs.docker.com/engine/reference/commandline/ps/` page. 

However, I see in `docs/README.md` in this repo that:
```
The documentation for Docker Engine has been merged into
[the general documentation repo](https://github.com/docker/docker.github.io). 
```

So maybe the change needs to be done in `https://github.com/docker/docs` repo. Therefore, the `edit this page` link needs to be updated accordingly on the public website.

PR #4399, opened 2023-07-03
docs: document special host-gateway value for add-host

dvdksn:

**- What I did**

Add some docs on `--add-host host.docker.internal:host-gateway` and `dockerd --host-gateway-ip`

**- How I did it**

**- How to verify it**

PXL_20230617_072054134

### DIFF
--- a/docs/reference/commandline/run.md
+++ b/docs/reference/commandline/run.md
@@ -759,23 +759,27 @@ PING docker (93.184.216.34): 56 data bytes
 round-trip min/avg/max = 92.209/92.495/93.052 ms
 ```
 
-Sometimes you need to connect to the Docker host from within your
-container. To enable this, pass the Docker host's IP address to
-the container using the `--add-host` flag. To find the host's address,
-use the `ip addr show` command.
+The `--add-host` flag supports a special `host-gateway` value that resolves to
+the internal IP address of the host. This is useful when you want containers to
+connect to services running on the host machine.
 
-The flags you pass to `ip addr show` depend on whether you are
-using IPv4 or IPv6 networking in your containers. Use the following
-flags for IPv4 address retrieval for a network device named `eth0`:
+It's conventional to use `host.docker.internal` as the hostname referring to
+`host-gateway`. Docker Desktop automatically resolves this hostname, see
+[Explore networking features](https://docs.docker.com/desktop/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host).
+
+The following example shows how the special `host-gateway` value works. The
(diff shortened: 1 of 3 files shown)


dvdksn:
@thaJeztah does this LGTY?

docs/reference/commandline/run.md
```diff
@@ -759,24 +759,27 @@ PING docker (93.184.216.34): 56 data bytes
...
+connect to services running on the host machine.
 
-The flags you pass to `ip addr show` depend on whether you are
-using IPv4 or IPv6 networking in your containers. Use the following
-flags for IPv4 address retrieval for a network device named `eth0`:
+It's conventional to use `host.docker.internal` as the hostname referring to
+`host-gateway`. Docker Desktop automatically resolves this hostname, see
+[Explore networking features](https://docs.docker.com/desktop/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host).
+
+The following example creates a simple HTTP server that serves a file from the
```
thaJeztah, line 770: perhaps avoid `simple` :)

docs/reference/commandline/run.md
```diff
@@ -759,24 +759,27 @@ PING docker (93.184.216.34): 56 data bytes
...
+host to a container.
 
 ```console
-$ HOSTIP=`ip -4 addr show scope global dev eth0 | grep inet | awk '{print $2}' | cut -d / -f 1 | sed -n 1p`
-$ docker run  --add-host=docker:${HOSTIP} --rm -it debian
+$ echo "hello from host!" > ./hello
+$ python3 -m http.server 8000
```
thaJeztah, line 775: Was looking if we could avoid `python`, and avoid having to create a file, but looks like `nc` is _just_ different enough between mac and linux for it to not work great as example (Linux requires the `-N` option, and macOS doesn't, and doesn't support it).

```bash
echo -ne "HTTP/1.0 200 OK\r\n\r\nhello from host\r\n" | nc -N -l 8080
```


Do you know if python allows for the content to be read from stdin?

dvdksn: I don't think this python module can read from stdin - only a directory. Would also prefer a better option... But I couldn't come up with one :D Happy to take suggestions..

thaJeztah: Yeah, don't think there's a better one. At most we could just `grep` the `/etc/hosts` file inside the container to "show" that it's there, but showing what it's used for is nicer.

We could add some wording "to illustrate" or something (as in; don't need to start installing `python3` if you don't have it installed... just to see "it works").

thaJeztah (approved):
LGTM, thanks!

PR #4403, opened 2023-07-06
docker swarm init: fix help printing

### LINKED ISSUE #4400: Help message for "docker swarm init" shows text of "docker init"
### Description

When attempting to read the help text for swarm init using `docker swarm init --help`, the CLI will show the following output:
```shell
docker swarm init --help

docker-init is a Docker CLI plugin

Commands:
  init                       Docker Init

Global Options:
      --config string      Location of client config files (default "/Users/moritz/.docker")
  -D, --debug              Enable debug mode
  -l, --log-level string   Set the logging level ("debug", "info", "warn", "error", "fatal") (default "info")
      --tls                Use TLS; implied by --tlsverify
      --tlscacert string   Trust certs signed only by this CA (default "/Users/moritz/.docker/ca.pem")
      --tlscert string     Path to TLS certificate file (default "/Users/moritz/.docker/cert.pem")
      --tlskey string      Path to TLS key file (default "/Users/moritz/.docker/key.pem")
      --tlsverify          Use TLS and verify the remote

Run 'docker COMMAND --help' for more information on a command.

For more help on how to use Docker, head to https://docs.docker.com/go/guides/
```

### Reproduce

1. Call `docker swarm init --help`

### Expected behavior

It should show the swarm init help text; I needed the reference for address advertisement.

### docker version

```bash
Client:
 Cloud integration: v1.0.33
 Version:           24.0.2
 API version:       1.43
 Go version:        go1.20.4
 Git commit:        cb74dfc
 Built:             Thu May 25 21:51:16 2023
 Context:           desktop-linux

Server: Docker Desktop 4.20.1 (110738)
 Engine:
  Version:          24.0.2
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.4
  Git commit:       659604f
  Built:            Thu May 25 21:52:17 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.21
  GitCommit:        3dce8eb055cbb6872793272b4f20ed16117344f8
 runc:
  Version:          1.1.7
  GitCommit:        v1.1.7-0-g860f061
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client:
 Version:    24.0.2
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
    Version:  v0.10.5
    Path:     /Users/moritz/.docker/cli-plugins/docker-buildx
    Version:  v2.18.1
    Path:     /Users/moritz/.docker/cli-plugins/docker-compose
    Version:  v0.1.0
    Path:     /Users/moritz/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.19
    Path:     /Users/moritz/.docker/cli-plugins/docker-extension
    Version:  v0.1.0-beta.4
    Path:     /Users/moritz/.docker/cli-plugins/docker-init
    Version:  0.6.0
    Path:     /Users/moritz/.docker/cli-plugins/docker-sbom
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
    Path:     /Users/moritz/.docker/cli-plugins/docker-scan
  scout: Command line tool for Docker Scout (Docker Inc.)
    Version:  v0.12.0
    Path:     /Users/moritz/.docker/cli-plugins/docker-scout

Server:
 Containers: 8
  Running: 8
  Paused: 0
  Stopped: 0
 Images: 49
 Server Version: 24.0.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: active
  NodeID: swb1uxtvrc222v3pnatozs5vy
  Is Manager: true
  ClusterID: dkf4zgwrhtk7399yisy2xtzk1
  Managers: 1
  Nodes: 1
  Default Address Pool: 10.0.0.0/8  
  SubnetSize: 24
  Data Path Port: 4789
  Orchestration:
   Task History Retention Limit: 5
  Raft:
   Snapshot Interval: 10000
   Number of Old Snapshots to Retain: 0
   Heartbeat Tick: 1
   Election Tick: 10
  Dispatcher:
   Heartbeat Period: 5 seconds
  CA Configuration:
   Expiry Duration: 3 months
   Force Rotate: 0
  Autolock Managers: false
  Root Rotation In Progress: false
  Node Address: 192.168.65.4
  Manager Addresses:
   192.168.65.4:2377
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3dce8eb055cbb6872793272b4f20ed16117344f8
 runc version: v1.1.7-0-g860f061
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 5.15.49-linuxkit-pr
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 6
 Total Memory: 7.772GiB
 Name: docker-desktop
 ID: 7ec1dcd4-5229-4cf8-9f79-8596deb0c066
 Debug Mode: false
 HTTP Proxy: http.docker.internal:3128
 HTTPS Proxy: http.docker.internal:3128
 No Proxy: hubproxy.docker.internal
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5555
  127.0.0.0/8
 Live Restore Enabled: false
```

### Additional Info

_No response_

hernandanielg:
fixes #4400

On the `setHelpFunc` function it tries to execute plugins's helps at first based in the ammount of arguments, the help for `docker swarm init --help` was getting matched with `docker-init` plugin, thus printing pluging help to output

To fix this we make sure that even thought we have 1 or more arguments it will check if the command it's first level of command by checking that the root command is the same as the parent command 

**- What I did**

**- How I did it**

**- How to verify it**

* compile with `docker buildx bake`
* execute `swarm init` command with `--help` flag `./build/docker swarm init --help`

expected output
```
➜  cli git:(fix-swarm-init-help) ./build/docker swarm init --help

Usage:  docker swarm init [OPTIONS]

Initialize a swarm

Options:
      --advertise-addr string                  Advertised address (format: "<ip|interface>[:port]")
      --autolock                               Enable manager autolocking (requiring an unlock key to start a stopped manager)
      --availability string                    Availability of the node ("active", "pause", "drain") (default "active")
      --cert-expiry duration                   Validity period for node certificates (ns|us|ms|s|m|h) (default 2160h0m0s)
      --data-path-addr string                  Address or interface to use for data path traffic (format: "<ip|interface>")
      --data-path-port uint32                  Port number to use for data path traffic (1024 - 49151). If no value is set or is set to 0, the default port
                                               (4789) is used.
      --default-addr-pool ipNetSlice           default address pool in CIDR format (default [])
      --default-addr-pool-mask-length uint32   default address pool subnet mask length (default 24)
      --dispatcher-heartbeat duration          Dispatcher heartbeat period (ns|us|ms|s|m|h) (default 5s)
      --external-ca external-ca                Specifications of one or more certificate signing endpoints
      --force-new-cluster                      Force create a new cluster from current state
      --listen-addr node-addr                  Listen address (format: "<ip|interface>[:port]") (default 0.0.0.0:2377)
      --max-snapshots uint                     Number of additional Raft snapshots to retain
      --snapshot-interval uint                 Number of log entries between Raft snapshots (default 10000)
      --task-history-limit int                 Task history retention limit (default 5)
```

fix to prevent wrong help message for commands with same name than docker plugins

ebdde703e6208406edeeabccf29ec2d6

### DIFF
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -229,14 +229,11 @@ func setupHelpCommand(dockerCli command.Cli, rootCmd, helpCmd *cobra.Command) {
 	}
 }
 
-func tryRunPluginHelp(dockerCli command.Cli, ccmd *cobra.Command, cargs []string) error {
-	root := ccmd.Root()
-
-	cmd, _, err := root.Traverse(cargs)
-	if err != nil {
-		return err
+func tryRunPluginHelp(dockerCli command.Cli, ccmd *cobra.Command) error {
+	if !pluginmanager.IsPluginCommand(ccmd) {
+		return errdefs.ErrNotFound
 	}
-	helpcmd, err := pluginmanager.PluginRunCommand(dockerCli, cmd.Name(), root)
+	helpcmd, err := pluginmanager.PluginRunCommand(dockerCli, ccmd.Name(), ccmd.Root())
 	if err != nil {
 		return err
 	}
@@ -251,8 +248,8 @@ func setHelpFunc(dockerCli command.Cli, cmd *cobra.Command) {
 			return
 		}
 
-		if len(args) >= 1 {
-			err := tryRunPluginHelp(dockerCli, ccmd, args)
+		if pluginmanager.IsPluginCommand(ccmd) {
+			err := tryRunPluginHelp(dockerCli, ccmd)
 			if err == nil {
 				return
 			}
--- a/cmd/docker/docker_test.go
+++ b/cmd/docker/docker_test.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
+	"strings"
 	"syscall"
 	"testing"
 	"time"
@@ -204,4 +206,40 @@ func TestVisitAll(t *testing.T) {
 	})
 	expected := []string{"sub1sub1", "sub1sub2", "sub1", "sub2", "root"}
 	assert.DeepEqual(t, expected, visited)
+}
+
+func TestSwarmInitHelpNotDelegatedToInitPlugin(t *testing.T) {
+	tmpDir := t.TempDir()
+	pluginScript := `#!/bin/sh
+if [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
+	echo "PLUGIN INIT HELP"
+	exit 0
+fi
+printf '%s' '{"SchemaVersion":"0.1.0","ShortDescription":"Docker Init"}'
+`
+	pluginPath := filepath.Join(tmpDir, "docker-init")
+	err := os.WriteFile(pluginPath, []byte(pluginScript), 0o755)
+	assert.NilError(t, err)
+
+	var b bytes.Buffer
+	ctx := context.Background()
+	cli, err := command.NewDockerCli(
+		command.WithBaseContext(ctx),
+		command.WithCombinedStreams(&b),
+	)
+	assert.NilError(t, err)
+	cli.ConfigFile().CLIPluginsExtraDirs = []string{tmpDir}
+
+	tcmd := newDockerCommand(cli)
+	tcmd.SetArgs([]string{"swarm", "init", "--help"})
+	cmd, args, err := tcmd.HandleGlobalFlags()
+	assert.NilError(t, err)
+	assert.NilError(t, tcmd.Initialize())
+	cmd.SetArgs(args)
+	err = cmd.Execute()
+	assert.NilError(t, err)
+
+	out := b.String()
+	assert.Assert(t, is.Contains(out, "Initialize a swarm"))
+	assert.Assert(t, !strings.Contains(out, "PLUGIN INIT HELP"), "output: %s", out)
 }


cmd/docker/docker.go
```diff
@@ -137,7 +137,7 @@ func setHelpFunc(dockerCli command.Cli, cmd *cobra.Command) {
 			return
 		}
 
-		if len(args) >= 1 {
+		if len(args) >= 1 && ccmd.Root() == ccmd.Parent() {
```
thaJeztah, line 140: Thank you for working on this!!

I'm slightly trying to get my head around the fix; and probably it could need a comment in the code, describing what we're doing here, and a similar comment in the commit message.

Also (not yet fully grasping the fix 😓 , so perhaps silly question); this only takes `.Parent()` into account, does that affect cases where "nested" commands could run into this?

hernandanielg: Hey! 👋🏻  I enjoyed working on this one, even tho it's an easy fix I struggled to make `dlv` debugger to work in vscode and then to find the issue, but it was fun 

So yeah I know, I need to add comments and maybe do it in a better way like create a function that does the check and returns a boolean

The idea is to identify if the command is first level command, in the case of `docker swarm init --help`
```
docker: is the root command
swarm : 1st level command
init  : 2nd level command
--help: is a flag
```

Why to idenfity 1st only? because docker plugins are in the form `docker-<plugin name>`, for this reason it was taking 2nd command `init` and making it `docker-init`

So how do we identify if it's the 1st level? checking parent command and root command, if both are the same then my command it's 1st level command - in the `init` case, root cmd is `docker` and parent is `swarm` so it will fail, no need to check great parent or nested

Yeah definitely will improve the commit, this was my first shot - thanks for the review!

thaJeztah: (quick comment, in a call right now)

> and maybe do it in a better way like create a function that does the check and returns a boolean

I'm fine with keeping the code in-line. Abstracting things can also hide logic, and sometimes it's fine to just have the logic in plain sight (not "shoved under the rug" to hide it away)

hernandanielg: added comments and rewrote commit msg, any help with wording will be highly appreciated 🙏🏻

hernandanielg: Hello @thaJeztah changes has been performed, can this be reviewed again? 🙏🏻

hernandanielg:
Hey @vvoland @thaJeztah !!

Long time no see :) - this long standing oldie PR seems to be forgotten, I've used Cursor to provide a solution would you mind to check if the change is accepted? 

Thanks in advance!

hernandanielg:
@vvoland @thaJeztah bump 🙃

docker-agent (commented):
### Assessment: 🟢 APPROVE

The fix correctly addresses the reported issue. The old `setHelpFunc` guard (`len(args) >= 1`) caused built-in subcommands like `swarm init` to have their help incorrectly delegated to a same-named plugin (e.g., `docker-init`). The new `pluginmanager.IsPluginCommand(ccmd)` gate is the right predicate — it directly checks whether the current cobra command is a plugin stub, making the delegation decision based on the command's nature rather than the presence of residual arguments.

The refactored `tryRunPluginHelp` signature is cleaner and the test `TestSwarmInitHelpNotDelegatedToInitPlugin` provides solid regression coverage for the bug scenario.

vvoland:
Looks like the test passes even before the fix. The fix looks fine though

hernandanielg:
and then? 🙃 
giphy

cmd/docker/docker_test.go
```diff
@@ -205,3 +207,39 @@ func TestVisitAll(t *testing.T) {
...
+	assert.NilError(t, err)
+
+	var b bytes.Buffer
+	ctx := context.Background()
+	cli, err := command.NewDockerCli(
+		command.WithBaseContext(ctx),
+		command.WithCombinedStreams(&b),
+	)
+	assert.NilError(t, err)
+	cli.ConfigFile().CLIPluginsExtraDirs = []string{tmpDir}
```
vvoland, line 231: `CLIPluginsExtraDirs` is assigned before `tcmd.Initialize()` which reloads the CLI configuration and discards it.

cmd/docker/docker_test.go
```diff
@@ -205,3 +207,39 @@ func TestVisitAll(t *testing.T) {
...
 }
+
+func TestSwarmInitHelpNotDelegatedToInitPlugin(t *testing.T) {
+	tmpDir := t.TempDir()
+	pluginScript := `#!/bin/sh
+if [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
+	echo "PLUGIN INIT HELP"
+	exit 0
+fi
+printf '%s' '{"SchemaVersion":"0.1.0","ShortDescription":"Docker Init"}'
```
vvoland, line 218: `Vendor` is required field but is missing so the plugin is considered invalid and not loaded.

vvoland (changes requested):
The unit test needs fixing so it actually reproduces the initial issue

PR #4410, opened 2023-07-10
Dockerfile: update to debian bookworm

crazy-max:
**- What I did**

Update to latest stable Debian distro (bookworm).

Also adds extra commit to use `debian` in stages instead of distro name. We need to update GitHub checks required (or just remove them imo):

image

**- How I did it**

**- How to verify it**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,10 @@
 # syntax=docker/dockerfile:1
 
 ARG BASE_VARIANT=alpine
-ARG GO_VERSION=1.21.3
 ARG ALPINE_VERSION=3.17
+ARG BASE_DEBIAN_DISTRO=bookworm
+
+ARG GO_VERSION=1.21.3
 ARG XX_VERSION=1.2.1
 ARG GOVERSIONINFO_VERSION=v1.3.0
 ARG GOTESTSUM_VERSION=v1.10.0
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -20,21 +22,14 @@ ARG TARGETPLATFORM
...
+FROM build-base-debian AS build-debian
 ARG TARGETPLATFORM
-RUN xx-apt-get install --no-install-recommends -y libc6-dev libgcc-10-dev
-# workaround for issue with llvm 11 for darwin/amd64 platform:
-#  # github.com/docker/cli/cmd/docker
-#  /usr/local/go/pkg/tool/linux_amd64/link: /usr/local/go/pkg/tool/linux_amd64/link: running strip failed: exit status 1
-#  llvm-strip: error: unsupported load command (cmd=0x5)
-# more info: https://github.com/docker/cli/pull/3717
-# FIXME: remove once llvm 12 available on debian
-RUN [ "$TARGETPLATFORM" != "darwin/amd64" ] || ln -sfnT /bin/true /usr/bin/llvm-strip
```
crazy-max, line 37: Not required anymore (llvm 14 since bookworm)

Dockerfile
```diff
@@ -63,8 +58,6 @@ ARG VERSION
 # PACKAGER_NAME sets the company that produced the windows binary
 ARG PACKAGER_NAME
 COPY --from=goversioninfo /out/goversioninfo /usr/bin/goversioninfo
-# in bullseye arm64 target does not link with lld so configure it to use ld instead
-RUN [ ! -f /etc/alpine-release ] && xx-info is-cross && [ "$(xx-info arch)" = "arm64" ] && XX_CC_PREFER_LINKER=ld xx-clang --setup-target-triple || true
```
crazy-max, line 67: Not required anymore

Dockerfile
```diff
@@ -1,16 +1,18 @@
 # syntax=docker/dockerfile:1
 
 ARG BASE_VARIANT=alpine
+ARG BASE_DEBIAN_DISTRO=bookworm
+ARG BASE_ALPINE_VERSION=3.17
```
thaJeztah, line 5: Curious; was there a reason for renaming the arg? If there's no specific reason, I'd like to keep it the same (also consistent with the moby repo where we use `ALPINE_VERSION`

Also considering if we should use `_VERSION` for debian as well, but I see we use `BASE_DEBIAN_DISTRO` in moby. 

Also can you put the alpine version above the debian version? I think we may update the alpine version more often, and having it "separate" from the GO_VERSION can potentially reduce conflicts when cherry-picking go version updates to release branches.

neersighted (approved):

thaJeztah:
I did a quick rebase to fix merge conflicts because of the `--link` we added in some places

thaJeztah (approved):
LGTM

thaJeztah:
> We need to update GitHub checks required (or just remove them imo):

removed them for now, but I'd like to add them back, so that (e.g.) "auto-merge" works as expected

PR #4419, opened 2023-07-13
Add missing opts to --network advanced syntax

akerouanton:
**- What I did**

Related to:

- https://github.com/moby/moby/pull/45905

The new advanced `--network` syntax introduced in docker/cli#1767 is lacking support for `link-local-ip` and `mac-address` fields. This PR adds both.

**- How to verify it**

With a daemon built out of moby/moby#45905:

```
$ docker run --rm --network=name=testnet1,mac-address=02:32:1c:23:00:04 busybox ip -o link show
```

The advanced syntax for `--network` now supports `link-local-ip` and `mac-address` fields.

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -712,6 +712,12 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		return nil, err
 	}
 
+	// Put the endpoint-specific MacAddress of the "main" network attachment into the container Config for backward
+	// compatibility with older daemons.
+	if nw, ok := networkingConfig.EndpointsConfig[hostConfig.NetworkMode.NetworkName()]; ok {
+		config.MacAddress = nw.MacAddress
+	}
+
 	return &containerConfig{
 		Config:           config,
 		HostConfig:       hostConfig,
@@ -773,8 +779,7 @@ func parseNetworkOpts(copts *containerOptions) (map[string]*networktypes.Endpoin
 	return endpoints, nil
 }
 
-func applyContainerOptions(n *opts.NetworkAttachmentOpts, copts *containerOptions) error {
-	// TODO should copts.MacAddress actually be set on the first network? (currently it's not)
+func applyContainerOptions(n *opts.NetworkAttachmentOpts, copts *containerOptions) error { //nolint:gocyclo
 	// TODO should we error if _any_ advanced option is used? (i.e. forbid to combine advanced notation with the "old" flags (`--network-alias`, `--link`, `--ip`, `--ip6`)?
 	if len(n.Aliases) > 0 && copts.aliases.Len() > 0 {
 		return errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --network-alias and per-network alias"))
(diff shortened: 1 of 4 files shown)


cli/command/container/opts.go
```diff
@@ -679,6 +679,14 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		return nil, err
 	}
 
+	if hostConfig.NetworkMode.IsBridge() || hostConfig.NetworkMode.IsUserDefined() {
```
thaJeztah, line 682: Before I forget to comment; we need to check if this does what we expect it to.


I recall I wanted to use some of these functions on the client-side, but the code was written "platform specific", so `IsUserDefined()` on a Linux client potentially wouldn't detect a "user-defined" network if the daemon is Windows (or vice-versa).

thaJeztah (changes requested):
OHMAN this one is complicated I tried to make changes what I (think) is needed while reviewing, and opened a PR;

- https://github.com/akerouanton/docker/pull/3

However, it may need checking, and there's some other comments I left (including "migrating containers during restore"), which I didn't look into yet.

cli/command/container/opts.go
```diff
@@ -788,6 +793,12 @@ func applyContainerOptions(n *opts.NetworkAttachmentOpts, copts *containerOption
 	if n.IPv6Address != "" && copts.ipv6Address != "" {
 		return errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --ip6 and per-network IPv6 address"))
 	}
+	if n.MacAddress != "" && copts.macAddress != "" {
+		return errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --mac-address and per-network MAC address"))
+	}
```
thaJeztah, line 798: Haven't checked out the code locally yet, but I wonder now if these errors are always correct; this error is _legit_ if I would do the following;


This is a conflict, because I'm setting the `mac-address` _TWICE_ for the default (`bridge`) network;

```bash
docker run \
    --mac-address=foo \
    --network name=bridge,mac-address=bar \
   myimage
```


This is _NOT_ a conflict, because I'm setting the `mac-address` twice, but for different networks; for `bridge`, I set it to `foo` and for `mynetwork`, I set it to `bar`;

```bash
docker run \
    --mac-address=foo \
    --network bridge \
    --network name=mynetwork,mac-address=bar \
   myimage
```


This one is more tricky though, as this one could be seen in 2 different ways;

- an implicit `--network=bridge` with a mac-address specified for the default (`bridge`) network
- a conflict (because I specified the network to be `mynetwork`, NOT `bridge`, and specified the mac-address _twice_ for that network);

```bash
docker run \
    --mac-address=foo \
    --network name=mynetwork,mac-address=bar \
   myimage
```

(Taking `mac-address` as example, but the same applies to other options that also have a separate `--flag`)

akerouanton: I think the original idea was to make sure no potentially ambiguous notations could be used. I think it will avoid some mistakes, so it's probably better to stick with that UX.

As it's mostly unrelated to my PR, could you open an issue if you think it's worth revisiting?

thaJeztah (approved):
LGTM

PR #4427, opened 2023-07-15
cli/context/docker: Endpoint.ClientOpts, withHTTPClient: inline variables

thaJeztah:
Remove some intermediate variables, and inline them.


laurazard (approved):
LGTM

PR #4453, opened 2023-07-19
cli/command/container: waitExitOrRemoved: take APIClient as argument

thaJeztah:
### cli/command/container: waitExitOrRemoved: take APIClient as argument

It only needs the API client, not the whole DockerCLI.

### DIFF
--- a/cli/command/container/utils_test.go
+++ b/cli/command/container/utils_test.go
@@ -2,13 +2,12 @@ package container
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"testing"
 
-	"github.com/docker/cli/internal/test"
 	"github.com/docker/docker/api"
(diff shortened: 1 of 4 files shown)


vvoland (approved):

cli/command/container/utils_test.go
```diff
@@ -63,7 +63,7 @@ func TestWaitExitOrRemoved(t *testing.T) {
 
 	client := test.NewFakeCli(&fakeClient{waitFunc: waitFn, Version: api.DefaultVersion})
```
thaJeztah, line 64: Oh, I guess I can potentially remove the `NewFakeCli` wrapper here; let me check.

thaJeztah: yup! removed.

vvoland (approved):

PR #4454, opened 2023-07-20
[do not merge] Set endpoint Priority

akerouanton:
Related to:

* https://github.com/moby/moby/pull/46036
* https://github.com/moby/moby/pull/45906

**- What I did**

The new `Priority` field added to `EndpointSettings` by moby/moby#46036 is used when multiple `--network` options are passed to `docker create` / `docker run`.

This first commit manually edit `github.com/docker/docker/api/types/network`. A proper revendoring will be needed once moby/moby#46036 is merged.

**- How to verify it**

CI is green


akerouanton:
Moby PR https://github.com/moby/moby/pull/46036 has been closed in favor of https://github.com/moby/moby/pull/48936. Moreover, after discussing this internally, we realize that setting the new `GwPriority` based on the order of `--network` flags would be a breaking change, so better not implement that. We might reconsider this decision in the future based on user feedback. For now, let me close this PR.

PR #4455, opened 2023-07-20
vendor: update buildkit to v0.12.0, containerd v1.7.2

thaJeztah:
### vendor: github.com/containerd/containerd v1.7.2

full diff: https://github.com/containerd/containerd/compare/v1.6.21..v1.7.2

### vendor: github.com/moby/buildkit v0.12.0

full diff: https://github.com/moby/buildkit/compare/v0.11.6..v0.12.0

**- A picture of a cute animal (not mandatory but encouraged)**

Issue #4402, opened 2023-07-05
sigterm handling behavior is odd


nicks:
### Description

Docker plugins have some really odd behavior when they get a SIGTERM signal:
- they always ignore the first sigterm
- after 3 sigterms, they force a shutdown

### Reproduce

Example code:
https://github.com/nicks/docker-sleep/blob/31f7ba2ffd0d48f7c1ba99ef4def3bd1c2429bcb/main.go#L1

1) Install my sample plugin, `docker sleep`
```
git clone git@github.com:nicks/docker-sleep
cd docker-sleep
go build -o ~/.docker/cli-plugins/docker-sleep .
docker sleep
```

2) Run `docker sleep`

3) In a separate terminal, run `kill` 3 times on the main `docker` process

### Expected behavior

The ideal behavior would be for the main `docker` cli process to forward the SIGTERM to the plugin, then exit when the plugin exits. But I think other behaviors are arguable?

It seems bizarre that it ignores the first signal, swallows it without telling the plugin, then dies on the third signal

### docker version

```bash
Client: Docker Engine - Community
 Cloud integration: v1.0.35
 Version:           24.0.2
 API version:       1.43
 Go version:        go1.20.4
 Git commit:        cb74dfc
 Built:             Thu May 25 21:51:00 2023
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          24.0.2
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.4
  Git commit:       659604f
  Built:            Thu May 25 21:51:00 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.21
  GitCommit:        3dce8eb055cbb6872793272b4f20ed16117344f8
 runc:
  Version:          1.1.7
  GitCommit:        v1.1.7-0-g860f061
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client: Docker Engine - Community
 Version:    24.0.2
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.0-cloud-driver+002
    Path:     /home/nick/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.19.0-cloud
    Path:     /home/nick/.docker/cli-plugins/docker-compose
[...]

PR #4457, opened 2023-07-20
remove buildkit as dependency from the CLI (integrate github.com/moby/buildkit/util/appcontext)

thaJeztah:
- relates to https://github.com/docker/cli/pull/4455#issuecomment-1644032209
- relates to https://github.com/moby/patternmatcher/pull/1

This copies the github.com/moby/buildkit/util/appcontext
package as an internal package. The appcontext package from
BuildKit was the only remaining dependency on BuildKit, and
while we may need some of its functionality, the implementation
is not correct for how it's used in docker/cli (so would need
a rewrite).

Moving a copy of the code into the docker/cli (but as internal
package to prevent others from depending on it) is a first step
in that process, and removes the circular dependency between
BuildKit and the CLi.

We are only using these:

    tree vendor/github.com/moby/buildkit
    vendor/github.com/moby/buildkit
    ├── AUTHORS
    ├── LICENSE
    └── util
        └── appcontext
            ├── appcontext.go
            ├── appcontext_unix.go
            ├── appcontext_windows.go
            └── register.go

    3 directories, 6 files

Before this:

    go mod graph | grep ' github.com/docker/cli'
    github.com/moby/buildkit@v0.11.6 github.com/docker/cli@v23.0.0-rc.1+incompatible

After this:

    go mod graph | grep ' github.com/docker/cli'
    # (nothing)

But indirect dependencies and circular dependency on moby/moby -> buildkit -> moby/moby and docker/cli can cause "hell".


cmd/docker/appcontext/appcontext.go
```diff
@@ -9,8 +9,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var appContextCache context.Context
-var appContextOnce sync.Once
+var (
+	appContextCache context.Context
+	appContextOnce  sync.Once
+)
```
thaJeztah, line 15: oh! that was `gofumpt` (which I have enabled in my IDE), well 🤷

thaJeztah:
Quick blurb, because we discussed this in one of the maintainers calls with @tonistiigi @neersighted (and others, but they may recall the exact details);

- the current use of `appcontext` in the CLI is not correct (also see https://github.com/docker/cli/issues/4402)
- ... but we may want to keep _parts_ of the behavior
- and.. this is where my memory fails me 🙈 😊 because I don't recall the exact parts we should _keep_ and which parts to _remove_
- I _think_ the "CTRL-C x3 to _force_ stop" was something that was desirable
- But ... I forgot the other details 🙈

thaJeztah:
@tonistiigi @neersighted PTAL; this keeps the status-quo on the functionality and does not fix the issues that were discussed about appcontext, but does remove the BuildKit dependency, which I think may be worth considering.

I updated the PR to move the appcontext package to `internal` (at least until we fixed the issues that were discussed).

neersighted (approved):

thaJeztah:
Let me bring this one in; we can use https://github.com/docker/cli/issues/4402 as tracking issue for the remaining work

PR #4483, opened 2023-08-06
fix: make options case-sensitive

beyazit:
**- What I did**
This pull request makes the options case-sensitive.

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/swarm/opts.go
+++ b/cli/command/swarm/opts.go
@@ -180,8 +180,7 @@ func parseExternalCA(caSpec string) (*swarm.ExternalCA, error) {
 			return nil, errors.Errorf("invalid field '%s' must be a key=value pair", field)
 		}
 
-		// TODO(thaJeztah): these options should not be case-insensitive.
-		switch strings.ToLower(key) {
+		switch key {
 		case "protocol":
 			hasProtocol = true
 			if strings.ToLower(value) == string(swarm.ExternalCAProtocolCFSSL) {
--- a/opts/config.go
+++ b/opts/config.go
@@ -45,8 +45,7 @@ func (o *ConfigOpt) Set(value string) error {
 			return fmt.Errorf("invalid field '%s' must be a key=value pair", field)
 		}
 
-		// TODO(thaJeztah): these options should not be case-insensitive.
-		switch strings.ToLower(key) {
+		switch key {
 		case "source", "src":
 			options.ConfigName = val
 		case "target":
--- a/opts/mount.go
+++ b/opts/mount.go
@@ -67,9 +67,6 @@ func (m *MountOpt) Set(value string) error {
 	for _, field := range fields {
 		key, val, ok := strings.Cut(field, "=")
 
-		// TODO(thaJeztah): these options should not be case-insensitive.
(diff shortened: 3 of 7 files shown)


thaJeztah:
Thanks for working on this! I recall I left those TODO's as "TODO" because, while I think we _should_ make these case sensitive, changing this will be a breaking change for users that use the ("supported", but undocumented) options.

So before we change the behavior we may need to

- first mark these as "deprecated" for at least 1 release (adding a deprecation notice in [docs/deprecated.md](https://github.com/docker/cli/blob/v24.0.5/docs/deprecated.md))
- _possibly_ print a deprecation _warning_ if a user uses one of these options with a "non-standard" case (but still make them work)

After those changes have shipped in a release, we can proceed with the changes in this PR to change the warning into an error.

thaJeztah:
Oh! It appears your commit message is missing a DCO sign-off,

```
```

commit.

    ```bash
    ```

2. Clone your fork locally

    ```bash
    ```

0bl:
Is there a version to be removed in the future?

thaJeztah:
> Is there a version to be removed in the future? Or will it just be deprecated and a warning printed?

I don't expect many users to depend on this behavior, because it's not documented (with millions of users, there always will be _some_ 😂), so I think "one release"  would be enough of a windows for users to be notified, so I would suggest;

- deprecate, and print a warning in the v25.0 release (which is the first upcoming release based on the master/main branch in this repository)
- "remove" (make it an error) in the v26.0 release.

As to the implementation; I had a quick peek at the code, and unfortunately it looks like we don't have direct access to `DockerCli.Err()` in these locations to know where to print to (which would be "better" as it's technically configurable).

I think it would be fine to just hard-code `os.Stderr` for the warning for this one (again, I'm somewhat anticipating that only very few users would depend on this (famous last words 😂)).


As most of these changes are in the same package, perhaps add a small utility would work for that. Something like;


```go
// printCaseInsensitiveWarning prints a warning when an option is used in the
// wrong case.
//
// TODO: remove case-insensitive handling for v26.0.0, and remove this utility.
func printCaseInsensitiveWarning(orig string) {
	if lower := strings.ToLower(orig); lower != orig {
		_, _ = fmt.Fprintf(os.Stderr, "DEPRECATED: the '%s' option is deprecated and will be removed in the next release; use '%s' instead\n", orig, lower)
	}
}
```


Which could be used something like this;

```go
// TODO(thaJeztah): these options should not be case-insensitive.
switch strings.ToLower(key) {
case "valid":
	// do valid stuff
case "also valid":
	// do valid stuff
default:
	return fmt.Errorf("invalid foobar option: %s", key)
}
printCaseInsensitiveWarning(key)
```

thaJeztah:
> I think it would be fine to just hard-code `os.Stderr` for the warning for this one (again, I'm somewhat anticipating that only very few users would depend on this (famous last words 😂)).


I should add that that's my $0.02c - it's other maintainers disagree (and maybe they consider this too much of an edge-case that just announcing it deprecated in the docs is enough, and _printing_ a warning is not needed), but we can discuss that during code review 👍

PR #4490, opened 2023-08-09
added headings on some functions

ibilalkayy:
**- What I did**

I saw that many functions are missing headings. That's why I added in some files in order to check whether you accept them or not.

If you accept them and guide me more, I will add more headings.

**- How I did it**

**- How to verify it**


vvoland (commented):
Thanks for the contribution!

IMO Godocs are definitely nice to have, but only if they add something that can't be easily deduced from the name/signature of the documented element.
I don't think it's worth the effort to add comments which simply repeat the function signature using a full sentence.

PR #4493, opened 2023-08-10
Create default EndpointSettings if no --network provided

akerouanton:
**- What I did**

Following flags are silently ignored when they're passed with no `--network` specified (ie. when the default network is used):

- `--network-alias`
- `--ip`
- `--ip6`
- `--link-local-ip`

This is not really an issue right now since the first 3 parameters are not allowed on the default bridge network. However, with [moby/moby#45905][1], the container-wide MacAddress parameter will be deprecated and dismissed. Because of that, with [docker/cli#4419][2], it's currently not possible to use the `--mac-address` flag with no default network specified.

Morever, `docker network connect --link-local-ip ...` works properly, so it should also work on `docker container create`. This also lay the ground for making the default bridge network just a "normal" network.

Since the 3 parameters in the list above aren't ignored anymore, if users provide them, moby's ContainerStart endpoint will complain about those. To provide better UX, [moby/moby#46183][3] make sure these invalid parameters lead to a proper error message on `docker container create` / `docker run`.

[1]: https://github.com/moby/moby/pull/45905
[2]: https://github.com/docker/cli/pull/4419
[3]: https://github.com/moby/moby/pull/46183

**- How to verify it**

```
# Before:
$ docker container create --link-local-ip 169.254.169.254 --name test busybox
$ docker inspect -f "{{ json .NetworkSettings.Networks.bridge.IPAMConfig }}" test
null

# After:
$ docker container create --link-local-ip 169.254.169.254 --name test busybox
$ docker inspect -f "{{ json .NetworkSettings.Networks.bridge.IPAMConfig }}" test
{"LinkLocalIPs":["169.254.169.254"]}
```

* `--link-local-ip` is now supported by `docker container create` and `docker run` when no `--network` parameter is specified


cli/command/container/opts.go
```diff
@@ -816,9 +816,6 @@ func parseNetworkAttachmentOpt(ep opts.NetworkAttachmentOpts) (*networktypes.End
 		if len(ep.Aliases) > 0 {
 			return nil, errors.New("network-scoped aliases are only supported for user-defined networks")
 		}
-		if len(ep.Links) > 0 {
-			return nil, errors.New("links are only supported for user-defined networks")
-		}
```
thaJeztah, line 821: I dug up my original PR / commit that changed this code; https://github.com/docker/cli/commit/5bc09639cc0aa382fdfcd9ca94d7817c8b3073c0

A couple of things there that I was curious about (but it's a long time); on this line, there's an error that treats `container.links` and `ep.links` as separate things; https://github.com/docker/cli/commit/5bc09639cc0aa382fdfcd9ca94d7817c8b3073c0#diff-e546b85d27f9dbea4fd41e83a58e36bb540bcda3e304355d56c92a119cd5aa2aR721

```go
	if len(n.Links) > 0 && copts.links.Len() > 0 {
		return errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --link and per-network links"))
	}
```



It's valid to use `--link` on the default ("bridge") network, but for that network (currently) only with the legacy link implementation (hard-wired links). 

Do you know if the daemon-side code makes this distinction as well when using the network-config make / slice? (i.e. does it look "if its the `default` (bridge) network, then use legacy links, and otherwise the DNS-based links?) or would the daemon only look at `container.links` to create legacy links?

thaJeztah: note that the above validation is only performed on the first network in the list (which holds data of the "default" network)

cli/command/container/opts.go
```diff
@@ -729,6 +729,20 @@ func parseNetworkOpts(copts *containerOptions) (map[string]*networktypes.Endpoin
 		hasUserDefined, hasNonUserDefined bool
 	)
 
+	if len(copts.netMode.Value()) == 0 {
+		n := opts.NetworkAttachmentOpts{
+			Target: "default",
```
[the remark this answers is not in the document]
akerouanton: See:

- https://github.com/moby/moby/pull/46447

cli/command/container/opts.go
akerouanton: You're right:

* `copts.links` will be put into `HostConfig.Links`, which is the way to use legacy links;
* `n.Links` is used for DNS-based links.

I removed the first commit (ie. the one removing this check) and instead I changed the condition under which `copts.links` is copied into `n.Links` (ie. to not do that on "default" network).

thaJeztah (approved):
LGTM

PR #4497, opened 2023-08-12
Update docs/command output for volume pruning

edmorley:
In previous versions of the Docker API, `system prune --volumes` and `volume prune` would remove all dangling volumes. With API v1.42, this was changed so that only anonymous volumes would be removed unless the all filter was specified.

Some of the docs were updated in #4218, however, there were a couple of places left that didn't make the anonymous vs named volumes distinction clear.

This replaces #4079, which was bitrotted by #4218. See also #4028.

Closes #4079.

**- What I did**
* Updated the `docker system prune --help` usage text's description of the `--volumes` option
* Updated the command log output for `docker system prune --volumes` to be clearer that only anonymous volumes will be removed
* Updated the `docker volume prune --help` usage text, so the overall command description no longer confusingly uses the word "all", given that by default it doesn't remove all volumes

**- How I did it**
See above

**- How to verify it**
Run `docker system prune --help`, `docker system prune --volumes` and `docker volume prune --help`.

N/A (since these are only single word edits to usage text/misc command output, I don't think it needs a changelog entry - happy to add one if preferred though).

### DIFF
--- a/cli/command/system/prune.go
+++ b/cli/command/system/prune.go
@@ -48,7 +48,7 @@ func newPruneCommand(dockerCli command.Cli) *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVarP(&options.force, "force", "f", false, "Do not prompt for confirmation")
 	flags.BoolVarP(&options.all, "all", "a", false, "Remove all unused images not just dangling ones")
-	flags.BoolVar(&options.pruneVolumes, "volumes", false, "Prune volumes")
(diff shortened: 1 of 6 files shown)


neersighted (approved):

cpuguy83 (approved):

edmorley:
The `ctn` CI job failed, but the failure seems unrelated to the changes in this PR? Is it a flaky test?
https://github.com/docker/cli/actions/runs/5841658109/job/15953254996#step:4:386

vvoland (approved):

Issue #3889, opened 2022-11-29
Bash completion slows down bash login sessions


hoxu:
- [x] I have tried with the latest version of Docker Desktop
  - [x] I have tried disabling enabled experimental features
  - [ ] I have uploaded Diagnostics
  - Diagnostics ID:

### Expected behavior

`bash_completion.d/docker` should be fast to execute. It probably should not make any `docker` calls until actually needed.

### Actual behavior

Executing the bash completion script causes a noticeable pause when opening new bash sessions.

The slowest part is a call to `docker`:
```
+++ [[ -r /usr/local/etc/bash_completion.d/docker ]]
+++ . /usr/local/etc/bash_completion.d/docker
+++++ shopt -p extglob
++++ __docker_previous_extglob_setting='shopt -s extglob'
++++ shopt -s extglob
+++++ docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "compose"}}{{.Path}}{{end}}{{end}}'
```

This seems to be:

```
COMPOSE_PLUGIN_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "compose"}}{{.Path}}{{end}}{{end}}')
```

```
$ time docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "compose"}}{{.Path}}{{end}}{{end}}'
/usr/local/lib/docker/cli-plugins/docker-compose

real    0m2.538s
user    0m0.142s
sys     0m0.085s
```

### Information
  - macOS Version:
  - Intel chip or Apple chip:
  - Docker Desktop Version: 4.14.1 (91661)

### Output of `/Applications/Docker.app/Contents/MacOS/com.docker.diagnose check`

N/A

### Steps to reproduce the behavior

Open a new bash login session.

To visually see what is taking up time:

```
bash -lx
```

thaJeztah:
> I know some improvements were made after that, but not sure if they can be included in the 20.10 version

Did a quick comparison;

```bash
# on docker 20.10;
time $(docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "compose"}}{{.Path}}{{end}}{{end}}')
real	0m1.034s
user	0m0.130s
sys	0m0.094s

# on docker 23.0.0-dev
time $(docker info --format '{{range .ClientInfo.Plugins}}{{if eq .Name "compose"}}{{.Path}}{{end}}{{end}}')
real	0m0.196s
user	0m0.136s
[...]

PR #11, opened 2017-04-27
Update cli folder with newer changes from moby/moby

tiborvass:
For the record, here is what I did from the moby/moby repo:
- `git filter-branch -f --prune-empty --subdirectory-filter cli`
- `git rebase -s recursive -Xsubtree=cli  dockercli/master` where `dockercli/master` is the master branch from the docker/cli repo (a git remote that I added)
- fix the rebases, and `git rebase --continue`
- `git merge -s ours --no-commit HEAD`
- `git read-tree --reset -u HEAD`
- `git commit`

PR #4505, opened 2023-08-21
Docker cli slows bash init

guss77:
**- What I did**

Instead of caching the docker plugins paths during bash init, put it in a function to be called as needed.

Fixes issue #3889 

**- How I did it**

This is a very simply patch that should be self evident.

**- How to verify it**

Check that it doesn't affect completion results with plugins installed

- Lookup plugins path for completion only on demand instead of caching it during bash initialization (which may be slow).

```
  ʌ___ʌ
  (O.O)
  /)__)
--''--''--
```


cpuguy83:
This needs a DCO sign-off.

`git commit --amend -s --no-edit && git push -f origin docker-cli-slows-bash-init`

vvoland (approved):

thaJeztah:
> I think the CLI could still improve here because it looks like it is making an API call to the _ping endpoint

Oh! I see where that one comes from; needs a similar fix as https://github.com/docker/cli/commit/006c9463894581a69845d3f25131acf15b73727f (https://github.com/docker/cli/pull/3904)

Opening a PR

vvoland:
One complain from the shellcheck:
```
#11 5.639 In contrib/completion/bash/docker line 1147:
#11 5.639 	echo ${DOCKER_PLUGINS_PATH//:/ }
#11 5.639              ^-------------------------^ SC2086: Double quote to prevent globbing and word splitting.
#11 5.639 
#11 5.639 Did you mean: 
#11 5.639 	echo "${DOCKER_PLUGINS_PATH//:/ }"
#11 5.639 
```

thaJeztah:
- Opened https://github.com/docker/cli/pull/4506 to fix the redundant `/_ping`

guss77:
> One complain from the shellcheck:
> #11 5.639              ^-------------------------^ SC2086: Double quote to prevent globbing and word splitting.

I understand the general sentiment, but it is not correct generally and specifically in this case it is wrong:

1. It doesn't actually matter: I'm echoing the result and whoever receives it is going to see multiple words regardless.
2. The entire reason for converting `:` to spaces is to cause word spitting, which is then used in `for plugin_path in $(__docker_plugins_path); do` to iterate over the words.

I can change it to have the double quotes just to silence shellcheck, but that is just pasting over the real problem, which is shellcheck being over enthusiastic about recommendations that are not universally correct.

contrib/completion/bash/docker
```diff
@@ -1142,7 +1142,10 @@ __docker_complete_user_group() {
 	fi
 }
 
-DOCKER_PLUGINS_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{.Path}}:{{end}}')
+__docker_plugins_path() {
+	DOCKER_PLUGINS_PATH=$(docker info --format '{{range .ClientInfo.Plugins}}{{.Path}}:{{end}}')
```
albers, line 1146: `DOCKER_PLUGINS_PATH` very much looks like an environment variable, which is a bit confusing.
In the context of this refactoring, I'd prefer a lowercased name with `local` scope, e.g.

`local docker_plugins_path`

albers (changes requested):
Thanks very much for the contribution.
Functionally LGTM, but I'd like a naming change, see comment.

thaJeztah:
For the linting error (assuming the current code is technically correct (I'm _always_ confused by quotes in bash)); I think it'd be ok to add a `# shellcheck disable=SC2086` in the code-block.

(unless there's another option that's both "good' and keeps the linter at bay of course)

contrib/completion/bash/docker
guss77: OK

guss77:
> I don't think shellcheck is being completely incorrect here - the `${DOCKER_PLUGINS_PATH//:/ }` is intended to be passed as a `STRING` argument of echo (message to be printed) and not as a whole command line to the `echo`.

OK

vvoland (approved):
LGTM. Could you also squash your commits into one please?

albers (approved):
LGTM, thanks.
I'm looking forward to this change becoming GA.

thaJeztah (approved):
LGTM, thank you!

thaJeztah:
Opened backports:

- https://github.com/docker/cli/pull/4517
- https://github.com/docker/cli/pull/4518
- https://github.com/docker/cli/pull/4519

PR #4506, opened 2023-08-22
cmd/docker: areFlagsSupported: don't Ping if not needed

thaJeztah:
- relates to https://github.com/docker/cli/pull/3904
- addresses https://github.com/docker/cli/pull/4505#issuecomment-1686863361

### cmd/docker: areFlagsSupported: don't Ping if not needed

This is a similar fix as 006c9463894581a69845d3f25131acf15b73727f (https://github.com/docker/cli/pull/3904), which
fixed this for detection of commands that were executed. Make sure we don't
call the "/_ping" endpoint if we don't need to.


thaJeztah:
@vvoland ptal 🤗  (also opened backports for this one)

vvoland (approved):

PR #4516, opened 2023-08-23
update flag-description for --cgroup-parent

thaJeztah:
relates to:

- https://github.com/docker/buildx/pull/1905
- https://github.com/moby/moby/issues/34469#issuecomment-362042644

This attempts to make it clearer that the --cgroup-parent option is only used for the containers used during build. Instead of mentioning "build container", I opted for using "RUN instructions" (to match the --network description), although this may not be ideal (as it assumes the "Dockerfile" front-end, which of course may not be the case).

**- What I did**

**- How I did it**

**- How to verify it**


dvdksn (approved):

rumpl (approved):

thaJeztah:
Perhaps we need a backport for this one as well, because (I _think_) the `buildx` docs are taken from `master`, so because that one's merged, the flag descriptions would currently not match.

PR #4534, opened 2023-08-28
vendor: github.com/docker/docker 5b53ddfcdd1c (v25.0.0-dev)

thaJeztah:
- [x] depends on https://github.com/docker/cli/pull/4533
- [x] depends on https://github.com/moby/moby/pull/46352
- includes https://github.com/moby/moby/pull/44704

full diff: https://github.com/docker/docker/compare/4b19b2f4babd...5b53ddfcdd1c0721f875d3e237078ab7de891a57

**- A picture of a cute animal (not mandatory but encouraged)**

PR #4535, opened 2023-08-28
refactor parsing restart-policies

thaJeztah:
- relates to https://github.com/moby/moby/pull/44704
- [x] depends on https://github.com/docker/cli/pull/4534

----

- opts.ParseRestartPolicy: use a struct-literal and clarify that this
  function only parses, but does not validate invalid combinations.
- cli/compose/convert: convertRestartPolicy: update doc to be more
  clear on the order of preference and intent.
- cli/compose/convert: convertRestartPolicy: use a switch based on
  known values to allow the exhaustive linter to catch missing options.
- cli/compose/convert: convertRestartPolicy: add validation for negative
  values when converting the legacy service.restart policy.
- cli/compose/convert: convertRestartPolicy: always set MaxAttempts
  and leave validation to the daemon.

### DIFF
--- a/cli/compose/convert/service_test.go
+++ b/cli/compose/convert/service_test.go
@@ -31,7 +31,7 @@ func TestConvertRestartPolicy(t *testing.T) {
 		},
 		{
 			input:    "unknown",
-			expError: "unknown restart policy: unknown",
+			expError: "invalid restart policy: unknown policy 'unknown'",
 		},
 		{
 			input: "always",
--- a/cli/compose/convert/service_test.go
+++ b/cli/compose/convert/service_test.go
@@ -18,35 +18,52 @@ import (
 	is "gotest.tools/v3/assert/cmp"
 )
 
-func TestConvertRestartPolicyFromNone(t *testing.T) {
-	policy, err := convertRestartPolicy("no", nil)
-	assert.NilError(t, err)
-	assert.Check(t, is.DeepEqual((*swarm.RestartPolicy)(nil), policy))
-}
-
-func TestConvertRestartPolicyFromUnknown(t *testing.T) {
-	_, err := convertRestartPolicy("unknown", nil)
-	assert.Error(t, err, "unknown restart policy: unknown")
-}
-
-func TestConvertRestartPolicyFromAlways(t *testing.T) {
-	policy, err := convertRestartPolicy("always", nil)
-	expected := &swarm.RestartPolicy{
-		Condition: swarm.RestartPolicyConditionAny,
-	}
-	assert.NilError(t, err)
-	assert.Check(t, is.DeepEqual(expected, policy))
-}
-
-func TestConvertRestartPolicyFromFailure(t *testing.T) {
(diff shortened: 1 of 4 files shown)


thaJeztah:
⚠️ For reviewers; I'm leaning two ways on this PR;

- 👍 validate early
- 👎 there IS  a potential (although _unlikely_ that these combinations could be supported by a future API version
- 👍 then again, the CLI does not support "future" versions of the API
- 👍 and .. there's already code invalidating "unknown" restart policies
- 🤷‍♂️ so I guess this doesn't make things worse

vvoland (dismissed):

vvoland:
I'm kinda neutral here - even if we'd change anything, that would still be gated by the API version which would require a newer CLI anyway.

But on the other hand... We will have the error message even without it, so there's no real gain for the UX.

vvoland:
@thaJeztah  are we still considering this?

thaJeztah:
OK; I went the reverse: make the parsing more agnostic, and leave validation to the daemon side, except for the "stack" (compose file) conversion, because for those we still need to map "container" restart policies to "swarm" restart policies.

cli/compose/convert/service_test.go
```diff
@@ -18,35 +19,52 @@ import (
...
+		}
+		t.Run(name, func(t *testing.T) {
+			policy, err := convertRestartPolicy(tc.input, nil)
+			if tc.expError != "" {
+				assert.Check(t, is.ErrorContains(err, tc.expError))
+				assert.Check(t, is.Nil(policy))
+				return
+			}
+			assert.NilError(t, err)
+			assert.Check(t, cmp.Equal(policy, tc.expected))
```
vvoland: Missing `is` alias

thaJeztah: whoops; fixed 👍

vvoland:
```

#18 69.72 
#18 69.72 === Failed
#18 69.72 === FAIL: cli/compose/convert TestConvertRestartPolicy/always (0.00s)
#18 69.72     service_test.go:64: assertion failed: &{any <nil> <nil> <nil>} (policy *swarm.RestartPolicy) != &{any <nil> <nil> <nil>} (tc.expected *swarm.RestartPolicy)
#18 69.72 
#18 69.72 === FAIL: cli/compose/convert TestConvertRestartPolicy/on-failure:4 (0.00s)
#18 69.72     service_test.go:64: assertion failed: &{on-failure <nil> 0x371e56d2c610 <nil>} (policy *swarm.RestartPolicy) != &{on-failure <nil> 0x371e56c16f60 <nil>} (tc.expected *swarm.RestartPolicy)
#18 69.72 
```

thaJeztah:
That's weird; I think it was green before and I only updated the import? 🤔


Will have a look when I'm at my computer

thaJeztah:
Fixed; looks like I had changed a  `is.DeepEqual` to an `is.Equal`; I probably was trying to simplify the check; should be good now 👍

vvoland (approved):

PR #4555, opened 2023-09-08
Fix extremely minor typo in docker-run man page

briantracy:
The latin phrase "vice versa" can be pronounced "vice-a-versa", but should not be rendered as such.

**- What I did**

Fixed an extremely minor typo in the docker-run man page.

**- How I did it**

Made a 6 character edit to the docker-run markdown documentation page.

**- How to verify it**

First, check out the definition of [vice versa](https://dictionary.cambridge.org/dictionary/english/vice-versa). Then note the definition of [vice-a-versa](https://www.wordnik.com/words/vice-a-versa) indicates that it is a "phonetic alternate spelling". This irks the extremely pedantic as the original latin phrase can be kept intact.

Now, while my one semester of college linguistics taught me that there is no "right" way to pronounce something, and that the fluidity of language is not only to be expected, but should not have a moral value ascribed to it, I think that if the original phrase is both remembered and widely used, we should defer to that.

Fixed an extremely minor typo in the docker-run man page.

```
  .
 ..^____/
`-. ___ )
  ||  ||
```

### DIFF
--- a/man/docker-run.1.md
+++ b/man/docker-run.1.md
@@ -761,10 +761,10 @@ The `Z` option tells Docker to label the content with a private unshared label.
 Only the current container can use a private volume.
 
 By default bind mounted volumes are `private`. That means any mounts done
-inside container will not be visible on host and vice-a-versa. One can change
+inside container will not be visible on host and vice versa. One can change
 this behavior by specifying a volume mount propagation property. Making a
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
Looks-a-good-to-me 😄, thanks!

PR #4556, opened 2023-09-09
Add otel support

cpuguy83:
This is broken into 3 commits:

- Plumb contexts through commands
- Revendor moby/moby to get client updates for otel support
- Add otel tracing as a persistent pre-run to all CLI commands.


vendor.mod
```diff
@@ -47,10 +47,16 @@ require (
 	gotest.tools/v3 v3.5.0
 )
 
+require (
+	go.opentelemetry.io/otel v1.4.1
+	go.opentelemetry.io/otel/trace v1.4.1
+)
```
thaJeztah, line 53: These ended up in the wrong block

cpuguy83:
Just a heads up, this is causing `docker build` to slow down quite a bit which I need to look into.

vendor.mod
cpuguy83: 😱 ☁️

PR #4566, opened 2023-09-15
docs: tidy markup

dvdksn:
**- What I did**

- Replace `<a name="">` with {#} for document that's imported without
  preprocessing. This markup is also used for the CLI reference docs,
  but the YAML unmarshalling handles the conversion from `<a name>` to {#}

- Remove raw/endraw liquid tags

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/commandline/cli.md
+++ b/docs/reference/commandline/cli.md
@@ -313,7 +313,6 @@ Following is a sample `config.json` file to illustrate the format used for
 various fields:
 
 ```json
-{% raw %}
 {
   "HttpHeaders": {
     "MyHeader": "MyValue"
@@ -356,7 +355,6 @@ various fields:
     }
   }
 }
-{% endraw %}
 ```
 
 ### Experimental features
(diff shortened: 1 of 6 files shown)


docs/reference/commandline/cli.md
```diff
@@ -379,7 +377,7 @@ list of root Certificate Authorities.
 
 ## Examples
 
-### <a name="host"></a> Specify daemon host (-H, --host)
+### Specify daemon host (-H, --host) {#host}
```
crazy-max, line 380: Should we do the same for other commands like https://github.com/docker/cli/blob/master/docs/reference/commandline/build.md?plain=1#L318?

crazy-max: > Not sure if cli-docs-tool would need any updating?

Hum I think so otherwise it would not generate link in the markdown table: https://github.com/docker/cli/actions/runs/6197667729/job/16826660657?pr=4566#step:4:35

crazy-max: But I think it would break linking in GitHub Flavored Markdown

dvdksn: Looks like GFM doesn't support it; https://gist.github.com/asabaylus/3071099?permalink_comment_id=4585378#gistcomment-4585378

PR #4571, opened 2023-09-18
docs: updates documents to describe default target behaviour

chrsalx:
**- What I did**
* I updated the references to the `--target` flag so it now explains what happens when the flag is not passed.

**- How I did it**
I searched all mentions of the `--target` flag and replaced the docs where it suited.

**- How to verify it**
You can test this by running `docker build -t demo --progress plain .` on any directory with the following image.

```Dockerfile
FROM ubuntu:latest as stagea
RUN echo "stagea"

FROM ubuntu:latest as stageb
RUN echo "stageb"
```

You should see something like this in the logs

```
#6 [stageb 2/2] RUN echo "stageb"
#6 0.097 imageb
#6 DONE 0.2s
```

Here's a pic of my dad's rottie. She's the loveliest being on earth. ♥️🐶
image


chrsalx:
Bumping this.

PR #4574, opened 2023-09-20
cli-plugins: include plugin metadata in User-Agent

milas:
**- What I did**
Automatically add plugin metadata to the CLI/API client `User-Agent`.

**- How I did it**
Add support to the `cli/command` package to accept a custom User Agent to pass to the underlying client.

Use said support to automatically append CLI plugin-specific info (vendor, name, version) to the `User-Agent` value.

For example, for a hypothetical CLI plugin from `vendor` named `plugin` with version `1.0.0` would have a user agent of something like:
```
Docker-Client/25.0.0 (darwin) vendor-plugin/1.0.0
```

Plugins must provide their name via the metadata by populating the field and raising their schema version from 0.1.0 to 0.2.0.

**- How to verify it**
Build a CLI plugin against latest `HEAD`, update its `SchemaVersion` to `0.2.0`, and inspect network traffic.

```
- CLI plugin name and version included in User-Agent for HTTP requests.
```

several baby skunks with bushy tails


cli-plugins/manager/metadata.go
```diff
@@ -23,3 +29,29 @@ type Metadata struct {
...
+	if meta.SchemaVersion == "" {
+		return fmt.Errorf("%q field is required", "SchemaVersion")
+	}
+
+	switch meta.SchemaVersion {
+	case "0.1.0":
+		// clear fields not supported by version
+		meta.Name = ""
+	case "0.2.0":
+		if meta.Name == "" {
```
thaJeztah, line 45: We should be careful with this one (yes, I've been looking at that gnarly "must be 0.1.0" as well), but if we'd ship a plugin with 0.2.0, it would mean it cannot be used with any CLI other than docker 25.0.0

Which also means that anyone installing an older version of the deb/rpm packages will now be broken.

cli/command/cli.go
```diff
@@ -82,6 +82,7 @@ type DockerCli struct {
 	dockerEndpoint     docker.Endpoint
 	contextStoreConfig store.Config
 	initTimeout        time.Duration
+	userAgent          string
```
thaJeztah, line 85: - I recall I opened https://github.com/moby/moby/pull/45677 to add an option to the APIClient; wondering if we can punch through to that one, without _also_ having to store it in the DockerCLI

milas: `DockerCli` uses that helper for API client construction! But `DockerCli` also lazily creates `RegistryClient` and `NotaryClient` instances and passes along a user agent to them, so it does need it.

I agree that we might want to add a `WithAPIClientOption` passthrough for future tweaks to the API client without needing 2x PRs each time

cli-plugins/manager/metadata.go
milas: Gah! You're absolutely right.

milas: I guess the `0.1.0` threw me for a loop and made me think "oh, semver, I should update this!"

I suppose from the perspective of a JSON-ish API, field additions are typically acceptable from a compatibility standpoint. How do you feel if I rollback the `SchemaVersion` part? Keep it as `0.1.0` but start sending `Name`.

cli/command/cli.go
thaJeztah: Ah! Right, yeah, I hate it that we _also_ re-invent our own "registry" client here (and I _think_ all that 💩  is for, erm, mostly legacy stuff).

I'll have a closer look if I can come up with some solutions (otherwise this looks fine to me). Things I still have my eyes on (not related to your changes per-se, but a bit of a sore point already);


- not a fan of requiring a fixed argument to pass around "user-agent" everywhere
- ☝️ even more if that is "hard-coded" (hard-coded, as in: a function-call that is hard-coded)
- ☝️ at least your PR improves _that_ part

But indeed persisting in 2 separate locations isn't ideal; also considering; is this something we should pass over the context (at least that would allow passing this around without requiring it to be passed as a dedicated argument that "somehow" needs to find its way somewhere)

cli-plugins/manager/metadata.go
milas: A possible alternative: change the signature of `plugin.Run()` to be `(name, meta, factoryFn)`. It avoids changing `Metadata` at all at the expense changing the API. (That will guarantee that all plugins start reporting UA once they upgrade at least...)

cli/command/cli.go
milas: Clarification: when you say context, do you mean Docker context or Go `context.Context`?

It could canonically live inside the API client, but then _that_ interface would need to gain a getter so that `DockerCli` could read it back out to pass to registry & notary client.

cli-plugins/manager/metadata.go
thaJeztah: So, I think we can change the code to remove the fixed requirement (we can set a minimum version of v0.1.0)

That won't fix existing cli's, but at least removes the weird bit for future.

Other than that, I think it's fine to add new fields, which means that we can detect if new fields are set (older CLI's would ignore them), and we could use that for feature detection.

(possibly a field that indicates what features are supported, if that helps avoid ambiguity between "empty field" and "feature supported but some field is empty")

cli/command/cli.go
thaJeztah: Doh! Sorry, somehow missed your last comment (also still need to look at this PR again 🙈)

I meant `context.Context`  and to (ab)use the context to pass this information, which would allow functions that need a user-agent to get access to it (which could potentially avoid having to punch-through a `UserAgent` argument to many places.

I haven't given that idea much thought yet though; I was considering that approach for API version (but for API version it may be more "clear-cut" that the API version to use would usually be "per request", and as such "per context"; for User-Agent that's more of a "grey area" I guess 😂

cli-plugins/plugin/plugin.go
```diff
@@ -53,7 +53,12 @@ func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager
 
 // Run is the top-level entry point to the CLI plugin framework. It should be called from your plugin's `main()` function.
 func Run(makeCmd func(command.Cli) *cobra.Command, meta manager.Metadata) {
-	dockerCli, err := command.NewDockerCli()
+	var dockerCliOpts []command.DockerCliOption
+	if ua, ok := userAgent(meta); ok {
+		dockerCliOpts = append(dockerCliOpts, command.WithUserAgent(ua))
+	}
+
+	dockerCli, err := command.NewDockerCli(dockerCliOpts...)
```
thaJeztah, line 61: Sorry for the delay (again), this one dropped off my list.

I was just wondering; do we actually need the Metadata to have the new field? Would it also work if the plugin would use `dockerCli.Apply()` and set the user agent? 

Looking at compose; https://github.com/docker/compose/blob/e5cd265abbdd9d67d76558d3b8d2e7fec68b7573/cmd/main.go#L36-L39

It gets the dockerCLI passed; which (with the option you added in this PR) could use that function to set a custom user-agent;

```go
func pluginMain() {
	plugin.Run(func(dockerCli command.Cli) *cobra.Command {
		_ = dockerCli.Apply(command.WithUserAgent("whatever"))
```

The net-result would be the same, but without the `Metadata` as intermediate (because that one also gets used for the JSON response, where it's not used?)

milas:
@thaJeztah PTAL - I refactored and there's no changes in the plugin metadata needed any longer.

I'm using the `plugin.Name()` now, which is the Cobra subcommand (e.g. `compose`, `buildx`). The UA will have `docker-cli-plugin-` prepended.

[See updated PR description for more details]

thaJeztah (approved):
LGTM

I rewrote this to take advantage of new options we have on plugin.Run

PR #4593, opened 2023-10-11
stacks: Add support for start interval

cpuguy83:


neersighted:
cc @dperny

neersighted (approved):

thaJeztah (approved):
LGTM

neersighted (changes requested):
Oops, looks like we need to create a new schema version in cli/compose/schema.

ndeloof:
docker/cli should **not** define some `3.10`+ schema  for compose file format, as this schema is maintained by https://github.com/compose-spec/compose-spec. This looks like some unofficial, unmaintained compose format fork, which deserve adoption of recent compose file features. Changes introduced by this schema could easily be supported by just adopting compose-go for compose file parsing, in replacement for the legacy parser.

thaJeztah:
The official v3 compose specification  has always been maintained in this repository; compose happened to have a copy in its repository, and the compose-spec was forked from that.

ndeloof:
compose file v3 indeed _used to be maintained_ here, but this was before we (Docker) [announced](https://www.docker.com/blog/announcing-the-compose-specification/) compose specification 4 years ago

PR #4599, opened 2023-10-12
cli-plugins: terminate plugin when CLI exits

laurazard:
**- What I did**

Previously, long lived CLI plugin processes weren't properly handled (see: https://github.com/docker/cli/issues/4402) resulting in plugin processes being left behind running, after the CLI process exits.

**- How I did it**

Change the plugin handling code to open an abstract unix socket before running the plugin and passing it to the plugin process, and changes the signal handling on the CLI side to close this socket which tells the plugin that it should exit.

The 3 SIGINTs/SIGTERMs logic is copied over/kept from `cmd/docker/internal/appcontext` (previously `github.com/moby/buildkit/util/appcontext`) (see https://github.com/docker/cli/pull/4457).

This implementation makes use of sockets instead of simply setting PDEATHSIG on the plugin process so that it will work on most platforms (including Windows :'))

**- How to verify it**

Before:

https://github.com/docker/cli/assets/70572044/46dc21f9-8bb7-4af5-83ff-087eb321ba3e

After:

https://github.com/docker/cli/assets/70572044/497b2704-1f4f-4781-984d-b15baec214ed

Or, build your plugin with the CLI code from my branch :)

```
Fix issue with long lived CLI plugin processes not being terminated after the CLI exits in some cases.
```


cli-plugins/plugin/plugin.go
```diff
@@ -24,12 +32,44 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		for {
+			b := make([]byte, 1)
+			_, err := udsSock.Read(b)
+			if errors.Is(err, io.EOF) {
+				// TODO(laurazard): remove this, debugging purposes only
```
laurazard, line 54: of course I forgot to remove this before committing 😅

cmd/docker/docker.go
```diff
@@ -194,9 +197,46 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
 	}
 	plugincmd.Env = append(envs, plugincmd.Env...)
 
+	// TODO(laurazard): this will break if we run this plugin more than once,
+	// as we will collide with the already-in-use socket name. We should retry
+	// a couple of times and add a counter or some other unique element to them
+	// socket name
+	pluginCloseSocketAddr := net.UnixAddr{Name: "@docker_cli_" + subcommand, Net: "unix"}
```
laurazard, line 204: @thaJeztah @vvoland @neersighted thoughts?

laurazard:
Have to check some CI failures, a lot of tests calling Compose breaking (not sure if due to parallel tests running into the issue described above or due to other issues with listening on an abstract namespace sockets in CI).

cmd/docker/docker.go
thaJeztah: Are there any constraints on these? i.e., could we just a GUID (`docker_cli_<GUID>`)?

neersighted: It's an arbitrary string that has to be globally unique; indeed a UUID or similar (e.g. a mktemp template) should be just fine.

cmd/docker/docker.go
```diff
@@ -194,9 +197,48 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
...
+	listener, err := net.ListenUnix("unix", &pluginCloseSocketAddr)
+	if err != nil {
+		return errors.New("bork")
+	}
+	defer listener.Close()
+	plugincmd.Env = append(envs, plugin.CliPluginCloseSocketEnvKey+"="+pluginCloseSocketAddr.Name)
```
thaJeztah, line 210: Could the CI failures be related to setting the `.Env` field? ISTR "empty" or "nil" means; keep the environment variables, but setting the field to any other value will discard the environment

cmd/docker/docker.go
```diff
@@ -194,9 +197,48 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
...
 	plugincmd.Env = append(envs, plugincmd.Env...)
 
+	// TODO(laurazard): this will break if we run this plugin more than once,
+	// as we will collide with the already-in-use socket name. We should retry
+	// a couple of times and add a counter or some other unique element to them
+	// socket name
+	pluginCloseSocketAddr := net.UnixAddr{Name: "@docker_cli_" + subcommand, Net: "unix"}
+	listener, err := net.ListenUnix("unix", &pluginCloseSocketAddr)
+	if err != nil {
+		return errors.New("bork")
```
neersighted, line 207: 🐩

cmd/docker/docker.go
neersighted: We already modify the environment image (which indeed, needs to be constructed from whole cloth if you modify it) above; this code just sets it to a new slice with a new `append()`.

thaJeztah: https://pkg.go.dev/os/exec#Cmd.Env

```go
// Env specifies the environment of the process.
	// Each entry is of the form "key=value".
	// If Env is nil, the new process uses the current process's
	// environment.
	// If Env contains duplicate environment keys, only the last
	// value in the slice for each duplicate key is used.
	// As a special case on Windows, SYSTEMROOT is always added if
	// missing and not explicitly set to the empty string.
	Env [][string](https://pkg.go.dev/builtin#string)
```


So possibly need to append to `os.Environ()` or `plugincmd.Environ()` https://pkg.go.dev/os/exec#example-Cmd.Environ

thaJeztah: > We already modify the environment image (which indeed, needs to be constructed from whole cloth if you modify it) above

Yes, was looking at that, but wondering if it _actually_ set anything (if `env` is nil, it would potentially still be nil(ish?)

cmd/docker/docker.go
thaJeztah: We really need to order a bork® T-shirt

thaJeztah: (And a børk® 🇸🇪 for David)

cmd/docker/docker.go
laurazard: It does (and appending to a nil slice works too!), but maybe that's not enough everywhere? It works on my ~~machine~~ Ubuntu VM

cmd/docker/internal/signals/signals_unix.go
```diff
@@ -0,0 +1,9 @@
+package signals
```
neersighted, line 1: `_unix` is not recognized in filenames; this should need a manual buildtag: https://github.com/golang/go/issues/51572

cmd/docker/docker.go
```diff
@@ -194,9 +197,48 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
...
+				// previous behaviour if we couldn't set up socket connection
+				println("got a weird error!", err.Error())
+			}
+		}()
+
+		go func() {
+			for {
+				<-signals
+				if err := conn.Close(); err != nil {
+					logrus.Error("failed to signal plugin to close: " + err.Error())
```
neersighted, line 233:
```suggestion
					logrus.Errorf("failed to signal plugin to close: %v", err)
```

cmd/docker/docker.go
```diff
@@ -194,9 +197,48 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
...
+		return errors.New("bork")
+	}
+	defer listener.Close()
+	plugincmd.Env = append(plugincmd.Env, plugin.CliPluginCloseSocketEnvKey+"="+pluginCloseSocketAddr.Name)
+
 	go func() {
-		// override SIGTERM handler so we let the plugin shut down first
-		<-appcontext.Context().Done()
+		signals := make(chan os.Signal, 2048)
+		signal.Notify(signals, platformsignals.TerminationSignals...)
```
neersighted, line 214: Probably should be a sprintf.

neersighted (commented):
Overall looks really promising, I'm glad that my ramble on solving this using a socket was helpful.

For others, Win32 job control can technically be used to solve this on Windows, but macOS still exists.

We could poll `getppid(2)`, but we'd need a busy loop, spam system calls (causing context switching), and it's full of edge-cases like `PR_SET_CHILD_SUBREAPER` or the fact that Solaris and Android have different init PIDs in some cases.

Letting the kernel notify us when the parent has either died, or has asked us to exit (solving the non-interactive SIGINT/SIGTERM case) seems best to me.

I'm still not sure if the 'three signal then exit' logic should apply in the non-interactive case, but we do just fine with it in dockerd as @laurazard pointed to me. Similarly, I'm not sure if we should handle SIGINT ourselves when there is a TTY in the loop, but I think this should be good enough as-is, and we can figure out further edge-cases once we get this in and have others kick the tires.

cmd/docker/docker.go
laurazard: 🐶

cmd/docker/docker.go
```diff
@@ -187,17 +191,58 @@ func setValidateArgs(dockerCli command.Cli, cmd *cobra.Command) {
...
-	go func() {
-		// override SIGTERM handler so we let the plugin shut down first
-		<-appcontext.Context().Done()
-	}()
+	listener, err := setupPluginSocket()
+	// if we somehow failed to set up the socket to alert the plugin
+	// when the CLI process wants to exit, fallback to old behaviour
+	if err == nil {
+		defer listener.Close()
+		plugincmd.Env = append(plugincmd.Env, plugin.CliPluginCloseSocketEnvKey+"="+listener.Addr().String())
```
neersighted, line 212: Probably should be a sprintf.

cmd/docker/docker.go
laurazard: I found https://github.com/distribution/distribution/blob/main/uuid/uuid.go (which we're already importing) which results in sockets that look like
```console
☆｡･:*:･ﾟ⭑ laurabrehm ✿ cli [plugin-signal-handling] ⭑ ss -ax | grep docker_cli
u_str ESTAB  0      0                                        @docker_cli_c264cbff-49e2-4cab-912f-175481090216 276544            * 274216
u_str LISTEN 0      4096                                     @docker_cli_c264cbff-49e2-4cab-912f-175481090216 273236            * 0
```

which looks fine for me, but I do wonder if that goes over some arbitrary character limit in any platform.

neersighted: Per `unix(7)`, the `sun_path` field is 108 bytes long. Post https://github.com/golang/go/issues/21965, we should have all 108 bytes to work with.

Other platforms are likely compatible with this length limit, as it's been in place since the early versions of Berkley sockets as best I can tell. We could add some tests for max length to make sure this works across all platforms, if desired.

cmd/docker/docker.go
thaJeztah: Not sure we should use a sprintf for a basic concatenation; sprintf definitely comes with an overhead

neersighted: For something that happens once during the lifetime of a process, I think the readability win is big, but I concede that it is strictly less efficient.

Still, premature optimization is often the greatest enemy of maintainability, even if this example is trivial.

thaJeztah:
Failures in e2e;

```
got a weird error! accept unix @docker_cli_9bd0ef8a-2f24-441f-85d9-fcab0b79d535: use of closed network connection
```

erm... @neersighted  would this be because of ... 

- https://github.com/moby/moby/pull/44562

neersighted:
No, this is `AF_UNIX` in the abstract socket domain (read: not a path, NUL bytes don't delimit the socket name, instead the `addrlen` defines when the socket name ends), not `AF_VSOCK`.

I'm not sure why we're getting a `net.ErrClosed` here; we should get `EINVAL` if the socket isn't listening.

laurazard:
I was going to say, it seems to be failing only when running inside of a container. I still want to change that code to fallback to the previous behaviour when we can't establish a connection, so that should be okay after those changes, but I'd like to figure out why it's not working.

thaJeztah:
Yeah, had my wires crossed on the AF_VSOCK; ended up on that as I was wondering if there's something different when running in a container 🥹☺️

neersighted:
Not sure what's going on there either, it will take some digging into. containerd's v1 shim API uses abstract sockets, so it should work in the general case DinD...

laurazard:
Adding this here as it seems like the most relevant place:

At the last maintainers call, it was suggested that a more elegant solution to this could be built, relying on another OS primitive – unnamed pipes – for communication between the CLI and the plugin process. In this implementation, instead of generating a random abstract socket name/passing it as an environment variable, we'd create an unnamed pipe on the side of the CLI process, and set the FD 3 (or another non-0/1/2 FD) of the plugin process as that pipe, which could then use it to be alerted to exit.

-------------------------

> [!NOTE]  
> This is my current understanding of the issue, which is definitely incomplete as my knowledge of Windows internals is extremely limited 😅 

This would be a great solution, but in practice it falls short as Go does not have built-in support for inheriting file descriptors on Windows. In truth, "file descriptors" aren't a concept on Windows, and although this _should_ be possible to do using file handles, there's a lot of complexity regarding how to pass the desired file handle to the child process.

While, on Unix*y* platforms, we can do:
```go
reader, writer, err :=  os.Pipe()
if err == nil {
    defer writer.Close()
    plugincmd.ExtraFiles = append(plugincmd.ExtraFiles, reader)
}
```

and then simply access FD 3 on the child process, a Windows implementation would have to look like:
```go
reader, writer, err :=  os.Pipe()
if err == nil {
    defer writer.Close()
    plugincmd.SysProcAttr = &syscall.SysProcAttr{AdditionalInheritedHandles: []syscall.Handle{syscall.Handle(reader.Fd())}}
}
```

**and** we would still have to pass the file handle as an environment variable/some other way so that the child process knows where to access it, as we can't just "access FD 3".

This is because `cmd.ExtraFiles` is not supported on Windows – more on this issue in https://github.com/golang/go/issues/21085#issuecomment-320853922.

(For the more curious, check out https://go-review.googlesource.com/c/go/+/288299, the Go CL where `os.Pipe`s were made inheritable on Windows, and how the feature is tested – *it involves passing the pipe handle as an argument when running the child command*)

We *could* implement this feature that way (more likely, passing the handle as an environment variable), but then we end up on the same different implementation for each platform place that we're trying to avoid.

I think, taking this into account, the abstract socket/`AF_UNIX` implementation is preferable.

(cc @corhere @cpuguy83 @thaJeztah)

----------------------------

Other alternatives, for future reference:
- Named pipes suffer from much the same issue, as we'd have to write our own implementation/rely on `go-winio` for them on Windows
- `PDEATHSIG` works on Unix, but not Windows, and the same/similar behaviour can be obtained on Windows with [Job Objects](https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects), but then we still need a solution for macOS.

corhere:
> **and** we would still have to pass the file handle as an environment variable/some other way so that the child process knows where to access it, as we can't just "access FD 3".

Yeah, it is annoying that [we can't predict the Windows handle value at compile-time](https://learn.microsoft.com/en-us/windows/win32/procthread/inheritance?redirectedfrom=MSDN#:~:text=To%20use%20a%20handle%2C%20the%20child%20process%20must%20retrieve%20the%20handle%20value%20and%20%22know%22%20the%20object%20to%20which%20it%20refers.%20Usually%2C%20the%20parent%20process%20communicates%20this%20information%20to%20the%20child%20process%20through%20its%20command%20line%2C%20environment%20block%2C%20or%20some%20form%20of%20interprocess%20communication.), unlike the tricks which can be played with POSIX file descriptors. We would have to have separate implementations of anonymous pipe IPC for Windows and Unix, but at least Linux and macOS could share one implementation.

One distinct advantage of anonymous pipes over abstract UNIX sockets is their anonymity: there is no name by which unrelated processes could reference it, intentionally or accidentally. However this is not a dealbreaker for abstract sockets at this time as the IPC channel is only being used for lifecycle management. The worst a malicious process could reasonably do is DoS a plugin. I'd suggest keeping the anonymous-pipe design in your back pocket just in case abstract sockets don't work out for some reason. Until then, I completely agree with Keeping It Simple using AF_UNIX.

cli-plugins/plugin/plugin.go
```diff
@@ -14,6 +17,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// CliPluginCloseSocketEnvKey is used to pass the plugin being
+// executed the abstract socket name it should listen on to know
+// when the CLI has exited.
+const CliPluginCloseSocketEnvKey = "CLI_PLUGIN_EXIT_SOCKET"
```
corhere, line 23:
```suggestion
const CliPluginCloseSocketEnvKey = "DOCKER_CLI_PLUGIN_EXIT_SOCKET"
```

cmd/docker/docker.go
```diff
@@ -187,16 +191,53 @@ func setValidateArgs(dockerCli command.Cli, cmd *cobra.Command) {
...
+	listener, err := setupPluginSocket()
+	if err == nil {
+		defer listener.Close()
+		plugincmd.Env = append(plugincmd.Env, plugin.CliPluginCloseSocketEnvKey+"="+listener.Addr().String())
+
+		go func() {
+			// ignore error here, if we failed to accept a connection,
+			// conn is nil and we fallback to previous behavior
+			conn, _ = listener.AcceptUnix()
+		}()
```
corhere, line 217: Either accept in a loop, or close the listener after the first connection is accepted. Make it well-defined whether or not the plugin socket can be dialed more than once.

cmd/docker/docker.go
```diff
@@ -187,16 +191,53 @@ func setValidateArgs(dockerCli command.Cli, cmd *cobra.Command) {
...
+	}
+
+	signals := make(chan os.Signal, 2048)
```
corhere, line 220:
```suggestion
	signals := make(chan os.Signal, exitLimit)
```
There's no need to buffer any more signals as the process will have exited after processing at most three of them.

cli-plugins/plugin/plugin.go
```diff
@@ -24,12 +32,45 @@ import (
...
+		return
+	}
+	cliCloseConn, err := net.DialUnix("unix", nil, addr)
+	if err != nil {
+		return
+	}
+
+	go func() {
+		for {
+			b := make([]byte, 1)
```
corhere, line 58: Be nice to the garbage collector; reuse your scratch buffer.
```suggestion
		b := make([]byte, 1)
		for {
```

cli-plugins/plugin/plugin.go
```diff
@@ -24,12 +32,45 @@ import (
...
+			_, err := cliCloseConn.Read(b)
+			if errors.Is(err, io.EOF) {
+				os.Exit(0)
```
corhere, line 61: `os.Exit` is very abrupt; no goroutine stacks are unwound and no deferred calls are invoked. Should the plugin [be given a chance to clean up and exit gracefully?](https://pkg.go.dev/github.com/spf13/cobra#Command.ExecuteContext) Any plugin that is fine with ungraceful termination [could easily opt in](https://pkg.go.dev/context#AfterFunc):
```go
context.AfterFunc(cmd.Context(), func() { os.Exit(0) })
```

thaJeztah:
> We could implement this feature that way (more likely, passing the handle as an environment variable), but then we end up on the same different implementation for each platform place that we're trying to avoid.

~Let me interject for a moment. What you guys are referring to as Linux ..~ (edit: sorry, damn you, autocomplete)

While I think an identical implementation would be "great", I would not consider it to be a _strict requirement_.

I know there's been many cases we (and the Go team itself for stdlib funcionality) have tried to provide a platform-agnostic approach, this also has resulted in many situations where trying to shoehorn Windows into Unix semantics resulted in poor choices.

While I think it's still _preferable_ if we implement library functions (it's great if we and can provide a platform-agnostic API / signature), we should not limit / force ourselves to do so if that means a more complicated and/or less idiomatic and/or less optimized/stable implementation.

In some cases there _are_ significant differences between Windows and Unix ("non-Windows"), and for some of those, it may be better to admit those differences, and choose the _best_ / _most stable_ solution for each.

Given that this particular code is an _implementation_ and not a "library package" (or at least: it's an internal package, and does not require a public Go API), I'm _not_ against having platform-specific implementations if that means we can optimize each with platform-specific semantics (potentially making it more stable). 


I think the goal here is to provide the same / similar experience from a _user's perspective_; if that requires a different implementation under the hood for each, that's an implementation detail.

(All my $0.02c :sweat_smile:)

cli-plugins/plugin/plugin.go
```diff
@@ -24,12 +32,45 @@ import (
 // called.
 var PersistentPreRunE func(*cobra.Command, []string) error
 
+// closeOnParentSocketClose connects to the socket specified
+// by the CLI_PLUGIN_EXIT_SOCKET env var, if present, and attempts
```
thaJeztah, line 36: Don't forget to update this comment if we decide to change the name (see other comment)

cli-plugins/plugin/plugin.go
```diff
@@ -24,12 +32,45 @@ import (
...
+// to read from it until it receives an EOF, which signals that
+// the CLI is going to exit and the plugin should also exit.
+func closeOnParentSocketClose() {
```
thaJeztah, line 39: Would it be difficult to (could be in a follow-up) write an integration test for this? I recall we already have some tests that build a test-plugin; perhaps we should have a test with a plugin that doesn't terminate immediately (e.g.)

laurazard: I think we should create a follow up to this to investigate how to best do that. I'd like to do it, but I don't think it should be a blocker since this change ultimately falls back to the previous behaviour if it fails.

cmd/docker/internal/signals/signals_windows.go
```diff
@@ -0,0 +1,7 @@
+package signals
+
+import "os"
+
+// TerminationSignals represents the list of signals we
+// want to special-case handle, on this platform.
+var TerminationSignals = []os.Signal{os.Interrupt}
```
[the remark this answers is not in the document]
thaJeztah: Yup, perfectly fine for a follow-up. I started looking at the code, and got curious. It's ok to keep the current signals, and keep looking into them after that.

cli-plugins/plugin/plugin.go
thaJeztah: Yup, also fine for after this PR. I was curious if you already had tried (and if you already had ideas for testing)

cli-plugins/plugin/plugin.go
laurazard: That's a fair point – however, making the new behavior opt-in sounds like a straightforward way to have all plugins ignore it and keep misbehaving. How does opt-out sound? We could change the API for `plugin.Run` and force plugins to pass a "shutdown" function, that will be called when the CLI is going to exit, and have that be `os.Exit(0)` if the plugin passes in `nil`. Or we can have `plugin.Run` take a channel, or some other notification mechanism, as long as that involves an API change that forces plugins to explicitly opt-out/overwrite shutdown behavior, and not just ignore it.

laurazard: @thaJeztah thoughts? (re: opt-in vs opt-out, and how the API could look)

corhere: I don't think it would be worth changing the plugin API just to guard against plugins that do not respect the graceful termination signal. A plugin could opt-in to graceful shutdown and hang anyway, e.g. because of a deadlock. If plugins misbehave then the plugins should be fixed. Aren't they all first-party plugins anyway?

Consider Option C: give all plugins a chance to exit upon context cancelation, with a deadline. That would cover both the "plugin ignores cancelation" and "plugin cancelation handling is bugged" cases, without any API changes.
```go
go func() {
    time.Sleep(gracefulExitGracePeriod) // e.g. 15s
    logrus.Fatal("plugin did not gracefully terminate in time")
}()
```

laurazard: Why is changing the API a negative, though? Compared to "Option C", changing the API will necessarily make plugin authors aware of the change in behavior and that they should now handle shutdown behavior (by being forced to pass in a shutdown function, for example), whereas with "Option C" all plugins will happily still compile, but since they won't necessarily be handling the case where their `CommandContext` gets cancelled, they won't shutdown and will cause the
```go
go func() {
    time.Sleep(gracefulExitGracePeriod) // e.g. 15s
    logrus.Fatal("plugin did not gracefully terminate in time")
}()
```
block to execute and print the message, at which point they'll have to dig in to the changes, find this PR and that they now have to handle the case where their `CommandContext` gets cancelled, and then make the appropriate changes.

I think communicating contract changes via API changes is a lot clearer, but there are probably some downsides to that change that I'm missing (please let me know 😅).

cli-plugins/plugin/plugin.go
laurazard: Changed name! Now:

```go
const CLIPluginSocketEnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
```

cmd/docker/docker.go
laurazard: Although I can't currently imagine a scenario where it would be different from the `Name` we set up, I feel in general (and for future-proofing's sake, too) it's safer to use `listener.Addr().String()` than to assume it will always be that in all platforms we build for.

cli-plugins/plugin/plugin.go
thaJeztah: Works for me! Keeping it more generic, so that we could still decide to use the socket for actual communication in future 👍

cli-plugins/plugin/plugin.go
thaJeztah: opened https://github.com/docker/cli/issues/4692 as placeholder for that

cli-plugins/plugin/plugin.go
vvoland: > We could change the API for plugin.Run and force plugins to pass a "shutdown" function, that will be called when the CLI is going to exit, and have that be os.Exit(0) if the plugin passes in nil. Or we can have plugin.Run take a channel, or some other notification mechanism, as long as that involves an API change that forces plugins to explicitly opt-out/overwrite shutdown behavior, and not just ignore it.

Not sure if a "shutdown" function at the top level would be convenient to use. Looking at what [buildx](https://github.com/docker/buildx/blob/5b5c4c8c9df55e133c39cce8153e0fa9fc6f60c4/commands/create.go#L50) and [compose](https://github.com/docker/compose/blob/30e80d244003761d5f1563f7f6dcd8aab515de0a/cmd/compose/compose.go#L74) - they create their own context that gets cancelled when the plugins receives the termination signal. So to handle the graceful exit they would need to introduce additional glue between the that shutdown function and their custom context.

Having the `CommandContext` being cancelled would be much more convenient way for the consumers of this API to handle the graceful exit correctly - assuming they pass the `CommandContext` and use the `context` idiomatically, they probably wouldn't even need to modify their code at all in most cases.
It will stop every cli plugin to invent its own shutdown mechanism and would provide a more unified behavior.

I wouldn't drop the `os.Exit` completely though - I think we still want to do it in case the plugin ignores the context cancellation and doesn't stop in a specified time (10s?).

I think that's what Cory described as "Option C" 😄. IMO that's the best solution, not because it doesn't break the API (I think it's fine to have a breaking changes in that API, specifically because they're mostly first party), but because it's simply easier and more idiomatic for the consumers.

corhere: I couldn't have said it better myself @vvoland. Also, #4700 is very timely and relevant. Contexts need to be plumbed through everywhere for tracing, regardless. If a plugin has been missed, a failure to gracefully exit is a much more visible sign than otel traces being less detailed than expected that contexts have not been plumbed through properly.

laurazard: > I think that's what Cory described as "Option C" 😄. IMO that's the best solution, not because it doesn't break the API (I think it's fine to have a breaking changes in that API, specifically because they're mostly first party), but because it's simply easier and more idiomatic for the consumers.

Gotcha, this sounds good to me. Thanks all for the discussion

cli-plugins/plugin/plugin.go
```diff
@@ -24,14 +34,57 @@ import (
...
 func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager.Metadata) error {
 	tcmd := newPluginCommand(dockerCli, plugin, meta)
 
 	var persistentPreRunOnce sync.Once
-	PersistentPreRunE = func(_ *cobra.Command, _ []string) error {
+	PersistentPreRunE = func(cmd *cobra.Command, _ []string) error {
 		var err error
 		persistentPreRunOnce.Do(func() {
+			ctx, cancel := context.WithCancel(cmd.Context())
+			cmd.SetContext(ctx)
```
thaJeztah, line 85: I was curious if `cmd.Context()` would always be set, but it looksthat cobra itself already triggers something to set it, so we should be good (afaics).

thaJeztah: Hm... although; I see a mention of "can be nil" in https://github.com/docker/cli/pull/4698 (https://github.com/docker/cli/pull/4698/commits/77f2f0fd67ce5fb2789748bcc991aa7496db593e#diff-3dcd6325b70a5f5db393a3ecc08060b69007fcef371fb4dabf3164da95646fb6R80-R82)

So now starting to question myself if there's possible code-paths where it could be nil 🤔

laurazard: I don't think it should happen unless someone calls `cmd.SetContext(nil)`, but we can add a check for that to rule out any possible issues I guess. I'm not sure how we would handle it though – we could either create a context and force it in, but then I imagine if a `cmd.Context` is nil, it might be that way on purpose, or we could skip this whole logic if it is nil.

laurazard: OTOH, we don't really want it to be nil (see #4698), so 🤷‍♀️

cmd/docker/internal/signals/signals_unix.go
```diff
@@ -0,0 +1,14 @@
+//go:build !windows
+// +build !windows
```
neersighted, line 2: `!windows` is technically not the same as `unix`... I'd honestly rather buildtag this Unix and provide a compilation error on unknown platforms:
```suggestion
//go:build unix
// +build unix
```

laurazard:
Made some changes after some discussions w/ @neersighted and @thaJeztah:
- removed "graceful shutdown period", as there are scenarios (such as with Compose) where a user would explicitly want/expect the plugin to get an arbitrary amount of time to gracefully shutdown/tear down things, and setting any timeout here on the CLI side would be counterproductive
- instead, changed up the "3 SIGINTS and we hard exit the CLI process" behaviour to, before exiting, send a SIGKILL to the plugin process. With this change, we allow wanted behaviour such as Compose taking longer to gracefully shutdown, but preserve the ability for a user to kill a misbehaving plugin by sending 3 SIGINTs to the CLI process, preserving user expectations regarding what happens when 3 SIGINTs are sent (as opposed to leaving the plugin process behind running).

neersighted:
As @laurazard mentioned, I want to address two cases of "stuck plugin" behavior:
* Parent CLI binary is killed, often non-gracefully, and the plugin keeps going/is "orphaned."
* Parent CLI binary is signaled to exit, but even with a prior version of this change, the plugin binary keeps going.

I think the current version addresses both, and notably helps address the issue with the "context cancellation must be respected" contract change by adding a fallback behavior that does what the user expects ("just kill everything now").

cli-plugins/plugin/plugin.go
laurazard: @neersighted wdyt about this?

thaJeztah: > context.WithCancel will already panic() if called with nil :) 

Ha! Fair point. Yes my consideration was that if one created a plugin, and no context was set, we should fail hard and loud. But of course that would only make sense if we can confidently check that it's not set.

From a quick check, it looks like Cobra is automatically doing that somewhere in the call tree, which is "good", but of course assuming it _always_ handles this (the comment on the other PR made me a bit concerned if there's code-paths where it would _not_ do so, potentially resulting in a random `panic` 😞.

I don't have a good answer to this one (honestly would've preferred Cobra to fail hard, and require a context to be set at all times if you want to use it, but I guess that complicated things on their side).

neersighted (approved):

cli-plugins/plugin/plugin.go
neersighted: Looks like I forgot to hit "Comment" yesterday:

I'm personally okay with this as-is; it seems like nullifying the context should be a fairly niche/obviously dubious thing to do, and will result in an appropriate (if potentially opaque) failure.

laurazard: I just missed your comment before pushing 😅. I added some very, very edge case_y_ code to cover this, that shouldn't change anything regardless because if somehow the context was nil before `cmd.Execute()`, Cobra will create a new context for it anyway.

vvoland (approved):

thaJeztah (approved):
LGTM! ❤️

cli-plugins/plugin/plugin.go
```diff
@@ -24,14 +33,56 @@ import (
...
+		return
+	}
+
+	go func() {
+		b := make([]byte, 1)
+		for {
+			_, err := cliCloseConn.Read(b)
+			if errors.Is(err, io.EOF) {
+				cancel()
+			}
```
thaJeztah, line 63: Looks like we missed this during review ☺️ 🙈 ;

- https://github.com/docker/cli/pull/4805

Issue #4591, opened 2023-10-09
Docker CLI images vs image


sozercan:
### Description

`docker image ls` (newer) and `docker images ls` (older) are different commands. This is very confusing for the user as it is easy to mistake `images` with `image`. Docker CLI does not provide any warnings.

```
$ docker images ls
REPOSITORY   TAG       IMAGE ID   CREATED   SIZE

$ docker image ls
REPOSITORY   TAG                                IMAGE ID       CREATED          SIZE
alpine       latest                             8ca4688f4f35   10 days ago      7.33MB
ubuntu       latest                             3565a89d9e81   2 weeks ago      77.8MB
```

```
$ docker version
Client:
 Version:           24.0.6-2

Server:
 Engine:
  Version:          24.0.6-4
```

rafaelucena:
That happens because `docker images` is actually an alias of `docker image ls`, so, essentially, when you're running `docker images ls` what you're really doing is running `docker image ls ls`.

The effect is that it filters the images that have ls as part of it's names.

Example:
```bash
$ docker images node
REPOSITORY   TAG         IMAGE ID       CREATED        SIZE
node         18          cf58080ad70c   22 hours ago   1.09GB
node         18-alpine   59b532238836   2 weeks ago    177MB
node         14-alpine   0dac3dc27b1a   6 months ago   119MB

$ docker image ls node
REPOSITORY   TAG         IMAGE ID       CREATED        SIZE
node         18          cf58080ad70c   22 hours ago   1.09GB
node         18-alpine   59b532238836   2 weeks ago    177MB
node         14-alpine   0dac3dc27b1a   6 months ago   119MB

```

thaJeztah:
Yes, this is a bit unfortunate, and definitely has confused others.

The `docker images` command was introduced when there were only "two" things that were managed by the CLI: containers and images; 

- containers could be listed through `docker ps` and removed with `docker rm`
- images could be listed through `docker images` and removed with `docker rmi`, and `docker images` had a positional argument to filter.

This UX didn't scale well when other kind of objects (volumes, networks (and nodes, services, secrets, configs for swarm)) were introduced, so we switched to using `docker <object> <command>` syntax going forward (`docker container ls` (or `docker container list`), `docker container rm`, `docker image ls`, `docker volume ls` etc), but we could not _remove_ the old commands because they were already widely used, so kept them as aliases.

[...]

PR #4601, opened 2023-10-12
4600-cli/command: adding `docker volumes` alias

rafaelucena:
**- What I did**
I've added a new command / alias for the `docker volume ls`, being `docker volumes`

**- How I did it**
By altering the existing files and doing a very similar job to what the `docker images` does, just with the volume instead

**- How to verify it**
Running `docker volumes` on the command line

Added a new alias for the `docker volume ls` command: `docker volumes`

image

**- Side notes**
This is my first PR here, please be kind, and sorry in advance :)

### DIFF
--- a/cli/command/commands/commands.go
+++ b/cli/command/commands/commands.go
@@ -41,6 +41,7 @@ func AddCommands(cmd *cobra.Command, dockerCli command.Cli) {
 		registry.NewSearchCommand(dockerCli),
 		system.NewVersionCommand(dockerCli),
 		system.NewInfoCommand(dockerCli),
+		volume.NewVolumesCommand(dockerCli),
 
 		// management commands
 		builder.NewBuilderCommand(dockerCli),
--- a/cli/command/volume/list.go
+++ b/cli/command/volume/list.go
@@ -26,16 +26,20 @@ type listOptions struct {
 	filter  opts.FilterOpt
 }
 
-func newListCommand(dockerCli command.Cli) *cobra.Command {
+// NewVolumesCommand creates a new `docker volumes` command
+func NewVolumesCommand(dockerCli command.Cli) *cobra.Command {
 	options := listOptions{filter: opts.NewFilterOpt()}
 
 	cmd := &cobra.Command{
-		Use:     "ls [OPTIONS]",
-		Aliases: []string{"list"},
+		Use:     "volumes [OPTIONS]",
 		Short:   "List volumes",
 		Args:    cli.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runList(dockerCli, options)
+		},
+		Annotations: map[string]string{
+			"category-top": "13",
+			"aliases":      "docker volume ls, docker volume list, docker volumes",
 		},
 		ValidArgsFunction: completion.NoComplete,
 	}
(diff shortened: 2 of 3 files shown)


rafaelucena:
I have not updated the md files for this change. I wasn't sure anybody would like the idea so I didn't want to go through the trouble until this got any progress stamp, but it can be done :)

thaJeztah:
Thanks for contributing! I don't think we should add this alias though; the `images` (plural) alias is mostly a left-over from the old UX (before `docker image ls`) existed, and was kept for backward compatibility, but has caused confusion for many users (I coincidentally just left a comment on one of those cases, and see you were on that ticket as well: https://github.com/docker/cli/issues/4591#issuecomment-1761160327)

rafaelucena:
Hi @thaJeztah! Yeah indeed, and I realize the mistake it can cause, but at the same time I feel like this is a good and a bad move.

I feel like having `docker images` and not having `docker volumes` means it lacks consistency. But I also feel like if `docker images` is causing a confusion then we probably could consider including a deprecation mark on it for future usage, assuming it is meant to be removed.

Somehow I feel like the confusion users feel while mistaking `docker images ls` for `docker image ls` could be comparable to the confusion me and others feel when mistaking `docker volumes` wanting to use `docker volume ls` while it doesn't exist.

It's possible it causes a similar annoyance :)

thaJeztah:
Yeah, the original plan was to hide (soft-deprecate) these, and there's an option to hide them through an environment variable (`DOCKER_HIDE_LEGACY_COMMANDS=1`), but it was never turned on by default.


```bash
DOCKER_HIDE_LEGACY_COMMANDS=1 docker --help

A self-sufficient runtime for containers

Common Commands:
  run         Create and run a new container from an image
  exec        Execute a command in a running container
  ps          List containers
  build       Build an image from a Dockerfile
  pull        Download an image from a registry
  push        Upload an image to a registry
  images      List images
  login       Log in to a registry
  logout      Log out from a registry
  search      Search Docker Hub for images
  version     Show the Docker version information
  info        Display system-wide information

Management Commands:
  builder     Manage builds
  buildx*     Docker Buildx (Docker Inc., v0.11.2-desktop.5)
  compose*    Docker Compose (Docker Inc., v2.22.0-desktop.2)
  container   Manage containers
  context     Manage contexts
  desktop*    Commands to interact with Docker Desktop (Docker Inc.)
  dev*        Docker Dev Environments (Docker Inc., v0.1.0)
  extension*  Manages Docker extensions (Docker Inc., v0.2.20)
  image       Manage images
  init*       Creates Docker-related starter files for your project (Docker Inc., v0.1.0-beta.8)
  manifest    Manage Docker image manifests and manifest lists
  network     Manage networks
  plugin      Manage plugins
  sbom*       View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan*       Docker Scan (Docker Inc., v0.26.0)
  scout*      Docker Scout (Docker Inc., v1.0.7)
  shell*      Open a browser shell on the Docker Host. (thaJeztah, v0.0.1)
  system      Manage Docker
  trust       Manage trust on Docker images
  volume      Manage volumes

Swarm Commands:
  swarm       Manage Swarm
```

rafaelucena:
I can understand why it's going away then. Long live the `docker images`. :(

Would preventing arguments on the `docker images` one not "resolve" the issues currently with people mistaking one by the other, though?

thaJeztah:
closing this PR for reasons outlined above and in https://github.com/docker/cli/pull/3758#issuecomment-4446764123

PR #4615, opened 2023-10-20
docs: refactor cli run reference

dvdksn:
- docs: improve introduction to docker run
- docs: move info about fg/bg flags to run reference

**- What I did**

**- How I did it**

**- How to verify it**


docs/reference/run.md
```diff
@@ -59,97 +51,59 @@ To learn how to interpret the types of `[OPTIONS]`, see
...
+{"level":"info","ts":1697798309.0285645,"msg":"shutdown complete","signal":"SIGINT","exit_code":0}
+```
+
+For more information about `docker run` flags related to foreground and
+background modes, see:
+
+- [`docker run --detach`](commandline/run.md#detach): run container in background
+- [`docker run --attach`](commandline/run.md#attach): attach to `stdin`, `stdout`, and `stderr`
+- [`docker run --tty`](commandline/run.md#attach): allocate a pseudo-tty
+- [`docker run --interactive`](commandline/run.md#attach): keep `stdin` open even if not attached
```
dvdksn, line 103: these links need fixing

thaJeztah (commented):
just some quick comments

docs/reference/run.md
```diff
@@ -51,78 +51,59 @@ $ docker run -it IMAGE sh
...
-order to allocate a tty for the container process. `-i -t` is often written `-it`
-as you'll see in later examples.  Specifying `-t` is forbidden when the client
-is receiving its standard input from a pipe, as in:
-
-```console
-$ echo test | docker run -i busybox cat
-```
+When you start a container, the container runs in the foreground by default.
+Your terminal attaches to the standard input, output, and standard error of the
+container.
```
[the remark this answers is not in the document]
dvdksn: oh yeah good catch... Let me try to rephrase

docs/reference/run.md
```diff
@@ -1,54 +1,46 @@
...
 
-    $ docker run [OPTIONS] IMAGE[:TAG|@DIGEST] [COMMAND] [ARG...]
+```console
+$ docker run [OPTIONS] IMAGE[:TAG|@DIGEST] [COMMAND] [ARG...]
+```
 
-The `docker run` command must specify an [*IMAGE*](https://docs.docker.com/glossary/#image)
-to derive the container from. An image developer can define image
-defaults related to:
+The `docker run` command must specify an [`IMAGE`](https://docs.docker.com/glossary/#image)
```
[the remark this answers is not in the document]
dvdksn: yeah probably shouldn't be backticks here...... I think it was a knee-jerk reaction on my part to please the linter which complains about the use of italics. I think just sentence casing this and omitting font-style changes is better here

docs/reference/commandline/run.md
```diff
@@ -186,6 +186,64 @@ flag makes Docker attempt to create a new file and write the container ID to it.
...
+
+In certain cases you want your container to share the host's process namespace,
+basically allowing processes within the container to see all of the processes
+on the system. For example, you could build a container with debugging tools
+like `strace` or `gdb`, but want to use these tools when debugging processes
+within the container.
+
+#### Example: run htop inside a container
+
+Create this Dockerfile:
```
thaJeztah, line 211: For a follow-up; this example could use an introduction to outline what we're about to do (run `htop` inside a container, but the host's PID namespace, which shows processes on the host (not just in the container itself).

And perhaps we should show _both_ (once with `--pid=host`, once without, to show the difference)

docs/reference/commandline/run.md
```diff
@@ -186,6 +186,64 @@ flag makes Docker attempt to create a new file and write the container ID to it.
...
+
+```dockerfile
+FROM alpine:latest
+RUN apk add --update htop && rm -rf /var/cache/apk/*
```
thaJeztah, line 215: We should use `apk add --no-cache` here, so that we don't have to deal with the `rm -rf`.


Honestly, we could even omit the `docker build` at all (alpine is really fast with these steps), if we want to keep it simple.

docs/reference/commandline/run.md
```diff
@@ -186,6 +186,64 @@ flag makes Docker attempt to create a new file and write the container ID to it.
...
+$ docker run -it --rm --pid=host myhtop
+```
+
+#### Example, join another container's PID namespace
+
+Joining another container's PID namespace can be useful for debugging that
+container. Start a container running a Redis server:
+
+```console
+$ docker run --name my-redis -d redis
```
thaJeztah, line 237: "note to self"; see how many `redis` examples we have, and whether we should consider `nginx` / `nginx:alpine` for these if we don't really need `redis`.

docs/reference/commandline/run.md
```diff
@@ -244,6 +244,23 @@ $ docker run -it --pid=container:my-redis my_strace_docker_image bash
...
+
+```text
+--uts=""  : Set the UTS namespace mode for the container
+            'host': use the host's UTS namespace inside the container
+```
+
+The UTS namespace is for setting the hostname and the domain that's visible to
+running processes in that namespace. By default, all containers, including
+those with `--network=host`, have their own UTS namespace. The `host` setting
+will result in the container using the same UTS namespace as the host. Note
```
thaJeztah, line 257: We should avoid future tense.

docs/reference/commandline/run.md
```diff
@@ -244,6 +244,23 @@ $ docker run -it --pid=container:my-redis my_strace_docker_image bash
...
+that `--hostname` and `--domainname` are invalid in `host` UTS mode.
```
thaJeztah, line 258: Perhaps also consider rephrasing "Note that" (which is a "Note"). Perhaps it needs context (why!? because `--network=host` runs in the host's network namespace, and the container would otherwise attempt to change the hosts' configuration. I guess that's what the paragraph below was intended for, but they're now two parts of the same thing 🤔 

Maybe we need a separate heading/example example combining both (with some warnings added, because it's changing the host)

docs/reference/commandline/run.md
```diff
@@ -261,6 +261,38 @@ You may wish to share the UTS namespace with the host if you would like the
...
+| "none"                     | Own private IPC namespace, with /dev/shm not mounted.                             |
+| "private"                  | Own private IPC namespace.                                                        |
+| "shareable"                | Own private IPC namespace, with a possibility to share it with other containers.  |
+| "container:<_name-or-ID_>" | Join another ("shareable") container's IPC namespace.                             |
+| "host"                     | Use the host system's IPC namespace.                                              |
+
+If not specified, daemon default is used, which can either be `"private"`
+or `"shareable"`, depending on the daemon version and configuration.
+
+IPC (POSIX/SysV IPC) namespace provides separation of named shared memory
```
thaJeztah, line 284: For a follow-up, we should mention `Inter-process communication` on first use of `IPC` to explain the acronym (perhaps we should provide a link to the `ipc(5)` man-page.

docs/reference/run.md
```diff
@@ -161,196 +161,33 @@ $ docker ps -q --filter ancestor=nginx:alpine
...
-provided in the format of `--network container:<name|id>`. Note that `--add-host`
-`--hostname` `--dns` `--dns-search` `--dns-option` and `--mac-address` are
-invalid in `container` netmode, and `--publish` `--publish-all` `--expose` are
-also invalid in `container` netmode.
-
-Example running a Redis container with Redis binding to `localhost` then
-running the `redis-cli` command and connecting to the Redis server over the
-`localhost` interface.
+When multiple containers are attached to the same custom network, they can
+communicate with each other using the container names as a DNS hostname:
```
thaJeztah, line 172: Perhaps outline the example (creates a custom network named `my-net`,  .. etc).

Mostly because there's 3 steps in the example, and each of them may be "new" to the user.

docs/reference/run.md
```diff
@@ -161,196 +161,33 @@ $ docker ps -q --filter ancestor=nginx:alpine
...
-#### User-defined network
-
-You can create a network using a Docker network driver or an external network
-driver plugin. You can connect multiple containers to the same network. Once
-connected to a user-defined network, the containers can communicate easily using
-only another container's IP address or name.
-
-For `overlay` networks or custom plugins that support multi-host connectivity,
-containers connected to the same multi-host network but launched from different
-Engines can also communicate in this way.
```
thaJeztah, line 458: All of this is covered in the networking section now? (I know it has in-depth information, mostly curious if we have a short "overview" of the general concepts there)

docs/reference/commandline/run.md
```diff
@@ -981,9 +981,57 @@ $ docker run --restart=always redis
 This will run the `redis` container with a restart policy of **always**
 so that if the container exits, Docker restarts it.
 
-You can find more detailed information on restart policies in the
-[Restart Policies (--restart)](../run.md#restart-policies---restart)
-section of the Docker run reference page.
+When a restart policy is active on a container, it will be shown as either `Up`
```
thaJeztah, line 1004: We should avoid future tense here.


Also; this commit mentions "docs: move --restart to docker run reference", but I don't see the content being removed from the other document in this commit 🤔

docs/reference/commandline/run.md
```diff
@@ -981,9 +981,57 @@ $ docker run --restart=always redis
...
+
+Or, to get the last time the container was (re)started;
+
+```console
+$ docker inspect -f "{{ .State.StartedAt }}" my-container
+2015-03-04T23:47:07.691840179Z
+```
+
+Combining `--restart` (restart policy) with the `--rm` (clean up) flag results
+in an error. On container restart, attached clients are disconnected.
```
thaJeztah, line 1054: note to self: Need to verify if this is still current (ISTR we had some handling for these, but may be mistaken)

docs/reference/run.md
```diff
@@ -298,13 +298,15 @@ non-zero exit status more than 10 times in a row Docker will abort trying to
...
 **on-failure** policy.
 
-## Exit Status
+## Exit status
 
 The exit code from `docker run` gives information about why the container
-failed to run or why it exited.  When `docker run` exits with a non-zero code,
-the exit codes follow the `chroot` standard, see below:
+failed to run or why it exited. When `docker run` exits with a non-zero code,
+the exit codes follow the `chroot` standard.
```
thaJeztah, line 381: I think the "see below" added _some_ value (it links the standard with the description below).

Perhaps, ", which is described below" or "as follows", or .. LOL something


(either that, or we should look if there's a link with relevant documentation about this "chroot standard")

docs/reference/commandline/run.md
```diff
@@ -1167,6 +1167,63 @@ in the image, or `SIGTERM` if the image has no `STOPSIGNAL` defined.
...
+| `--security-opt="label=type:TYPE"`        | Set the label type for the container                                      |
+| `--security-opt="label=level:LEVEL"`      | Set the label level for the container                                     |
+| `--security-opt="label=disable"`          | Turn off label confinement for the container                              |
+| `--security-opt="apparmor=PROFILE"`       | Set the apparmor profile to be applied to the container                   |
+| `--security-opt="no-new-privileges=true"` | Disable container processes from gaining new privileges                   |
+| `--security-opt="seccomp=unconfined"`     | Turn off seccomp confinement for the container                            |
+| `--security-opt="seccomp=profile.json"`   | White-listed syscalls seccomp Json file to be used as a seccomp filter    |
+
+
+You can override the default labeling scheme for each container by specifying
```
thaJeztah, line 1228: For a follow-up; "You can .. " always feels a bit awkward. Perhaps something like "Use the `--security-opt` option to override ...", or "The `--security-opt` option (allows you to) .."

docs/reference/commandline/run.md
```diff
@@ -1167,6 +1167,63 @@ in the image, or `SIGTERM` if the image has no `STOPSIGNAL` defined.
...
+```console
+$ docker run --security-opt label=level:s0:c100,c200 -it fedora bash
+```
+
+> **Note**
+>
+> Automatic translation of MLS labels is not currently supported.
+
+To disable the security labeling for this container versus running with the
+`--privileged` flag, use the following command:
```
thaJeztah, line 1241: This sentence is also a bit awkward

LOL; this whole section actually ("if you .. you can" etc; could use more active voice).

docs/reference/commandline/run.md
```diff
@@ -874,6 +874,17 @@ $ docker run --rm -i busybox echo "foo bar baz" \
...
+
+You can use the `--init` flag to indicate that an init process should be used as
+the PID 1 in the container. Specifying an init process ensures the usual
+responsibilities of an init system, such as reaping zombie processes, are
+performed inside the created container.
+
+The default init process used is the first `docker-init` executable found in the
+system path of the Docker daemon process. This `docker-init` binary, included in
+the default installation, is backed by [tini](https://github.com/krallin/tini).
+
```
thaJeztah, line 889: For a follow-up; I _think_ the init-binary to use is configurable on the daemon (if so, we may need a link to that part of the daemon config)

docs/reference/commandline/run.md
```diff
@@ -921,6 +921,13 @@ You can use the `-t` flag without `-i` flag. This still allocates a pseudo-TTY
 to the container, but with no way of writing to `STDIN`. The only time this
 might be useful is if the output of the container requires a TTY environment.
 
+## <a name="cgroup-parent"></a> Specify custom cgroups
+
+Using the `--cgroup-parent` flag, you can pass a specific cgroup to run a
+container in. This allows you to create and manage cgroups on their own. You can
+define custom resources for those cgroups and put containers under a common
+parent group.
```
thaJeztah, line 931: For a follow-up; this section may need a rewrite (but it's pretty-much "power-user" area). There's some (not so) subtle issues with group-parent when using cgroups v2, besides currently this section doesn't describe at all how to use it 😞 (We may have a tracking issue for that somewhere, but not sure if it's still open)

docs/reference/commandline/run.md
```diff
@@ -1123,6 +1123,31 @@ $ docker run \
...
 
+## <a name="log-driver"></a> Logging drivers (--log-driver)
+
+The container can have a different logging driver than the Docker daemon. Use
+the `--log-driver=VALUE` with the `docker run` command to configure the
+container's logging driver. The following options are supported:
+
+| Driver       | Description                                                                                                                    |
+|:-------------|:-------------------------------------------------------------------------------------------------------------------------------|
+| `none`       | Disables any logging for the container. `docker logs` won't be available with this driver.                                     |
```
thaJeztah, line 1136: Perhaps we should remove the overview here (we could limit to describe `none` "to disable logging"). That way we can keep the logging section as the canonical overview.

docs/reference/commandline/run.md
```diff
@@ -1123,6 +1123,31 @@ $ docker run \
...
+| `journald`   | Journald logging driver for Docker. Writes log messages to `journald`.                                                         |
+| `gelf`       | Graylog Extended Log Format (GELF) logging driver for Docker. Writes log messages to a GELF endpoint likeGraylog or Logstash.  |
+| `fluentd`    | Fluentd logging driver for Docker. Writes log messages to `fluentd` (forward input).                                           |
+| `awslogs`    | Amazon CloudWatch Logs logging driver for Docker. Writes log messages to Amazon CloudWatch Logs.                               |
+| `splunk`     | Splunk logging driver for Docker. Writes log messages to `splunk` using Event Http Collector.                                  |
+| `etwlogs`    | Event Tracing for Windows (ETW) events. Writes log messages as Event Tracing for Windows (ETW) events. Only Windows platforms. |
+| `gcplogs`    | Google Cloud Platform (GCP) Logging. Writes log messages to Google Cloud Platform (GCP) Logging.                               |
+| `logentries` | Rapid7 Logentries. Writes log messages to Rapid7 Logentries.                                                                   |
```
thaJeztah, line 1147: Related to the above; the `logentries` driver is deprecated (well; the service no longer exists, so it's no longer functional, but we didn't remove the driver yet).

docs/reference/commandline/run.md
```diff
@@ -1123,6 +1123,31 @@ $ docker run \
...
+
+The `docker logs` command is available only for the `json-file` and `journald`
+logging drivers. For detailed information on working with logging drivers, see
+[Configure logging drivers](https://docs.docker.com/config/containers/logging/configure/).
```
thaJeztah, line 1151: This is no longer correct. We now enable "dual logging" by default, which allows using `docker logs` with any driver;

- https://github.com/docker/docs/pull/12139
- https://github.com/moby/moby/pull/40543

thaJeztah (commented):
I'm gonna take a break for a bit; I looked at commits until "docs: rewrite section on overriding image defaults"

Most comments are probably for follow-ups, but some could be part of this PR (for some follow-ups we may want to create tracking tickets)

docs/reference/run.md
```diff
@@ -1,54 +1,46 @@
...
 
- * detached or foreground running
- * container identification
- * network settings
- * runtime constraints on CPU and memory
+`[OPTIONS]` let you configure runtime options for the container. For example,
+you can give the container a name (`--name`), or run it as a background process
+(`-d`). You can also set options to control things like resource constraints
+and networking. Learn more about the available `[OPTIONS]` in [*Option
+types*](commandline/cli.md#option-types).
```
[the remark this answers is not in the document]
dvdksn: I will remove this link I think. It's not really helpful in this context.

docs/reference/commandline/run.md
```diff
@@ -541,38 +541,34 @@ content label. Shared volume labels allow all containers to read/write content.
...
-still store what's written to `STDERR` and `STDOUT`.
+This succeeds in starting the `nginx` service inside the container. However, it
+fails the detached container paradigm in that, the root process (`service nginx
+start`) returns and the detached container stops as designed. As a result, the
+`nginx` service starts but can't be used. Instead, to start a process such as
+the `nginx` web server do the following:
 
 ```console
-$ cat somefile | docker run -i -a stdin mybuilder dobuild
+$ docker run -d -p 80:80 my_image nginx -g 'daemon off;'
```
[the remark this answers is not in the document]
dvdksn: even more confusing here is that we're using a `my_image` custom image for all this without any information about what that image contains.

docs/reference/commandline/run.md
dvdksn: Good idea. One thing that came to my mind, I noticed that this option shares the VM process namespace when using DD for instance. Might be worth calling out?

docs/reference/commandline/run.md
dvdksn: Agree, we need to look into this at a broader level, but I'm changing this one to nginx for now

docs/reference/run.md
dvdksn: Yes; I've got a 2nd PR in docker/docs to add the missing bits from here: https://github.com/docker/docs/pull/18753

docs/reference/commandline/run.md
dvdksn: oops

docs/reference/run.md
dvdksn: Not sure even calling out a `chroot` standard is very useful. I will simplify.

docs/reference/commandline/run.md
```diff
@@ -128,27 +128,55 @@ Use `docker ps -a` to view a list of all containers, including those that are st
...
 ```
 
-This example runs a container named `test` using the `debian:latest`
-image. The `-it` instructs Docker to allocate a pseudo-TTY connected to
-the container's stdin; creating an interactive `bash` shell in the container.
-The example quits the `bash` shell by entering
-`exit 13`, passing the exit code on to the caller of
-`docker run`, and recording it in the `test` container's metadata.
+If you don't specify a custom name using the `--name` flag, the daemon assigns
+a randomly generated name, such as `vibrant_cannon`, to the container. Using a
```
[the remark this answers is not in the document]
thaJeztah: > I think we're also not adding auto-generated names to DNS, so if a container is not given a name, it cannot be resolved

- relates to https://github.com/moby/moby/issues/45183

docs/reference/run.md
```diff
@@ -970,59 +970,31 @@ $ docker run -it --entrypoint="" mysql bash
...
-
-                   (use 'docker port' to see the actual mapping)
-
-    --link=""  : Add link to another container (<name or id>:alias or <name or id>)
-
-With the exception of the `EXPOSE` directive, an image developer hasn't
-got much control over networking. The `EXPOSE` instruction defines the
-initial incoming ports that provide services. These ports are available
-to processes inside the container. An operator can use the `--expose`
-option to add to the exposed ports.
```
thaJeztah, line 1533: I notice we remove `--expose` altogether here. Although admitted, `--expose` is of  little value for most use-cases (i.e., don't need `--expose` to `--publish`)

docs/reference/run.md
```diff
@@ -970,59 +970,31 @@ $ docker run -it --entrypoint="" mysql bash
...
+You can start the container with the `-P` or `-p` flags to expose its ports:
+
+- The `-P` (or `--publish-all`) flag publishes all the exposed ports to the
+  host. Docker binds each exposed port to a random port on the host.
+
+  > **Note**
+  >
+  > The `-P` flag only publishes port numbers that are explicitly flagged as
+  > exposed, either using the Dockerfile `EXPOSE` instruction or the `--expose`
+  > flag for the `docker run` command.
```
thaJeztah, line 1062: Bit on the fence if we want this as a note, or just part of the running text (it maps each port that is "exposed")

docs/reference/run.md
```diff
@@ -970,59 +970,31 @@ $ docker run -it --entrypoint="" mysql bash
...
-To expose a container's internal port, an operator can start the
-container with the `-P` or `-p` flag. The exposed port is accessible on
-the host and the ports are available to any client that can reach the
-host.
-
-The `-P` option publishes all the ports to the host interfaces. Docker
-binds each exposed port to a random port on the host. The range of
-ports are within an *ephemeral port range* defined by
-`/proc/sys/net/ipv4/ip_local_port_range`. Use the `-p` flag to
-explicitly map a single port or range of ports.
```
thaJeztah, line 1544: Do we still document this information elsewhere? the `/proc/sys/net/ipv4/ip_local_port_range` information may be relevant information.

docs/reference/run.md
```diff
@@ -970,59 +970,31 @@ $ docker run -it --entrypoint="" mysql bash
...
-and the exposed ports, use `docker port`.
-
-If the operator uses `--link` when starting a new client container in the
-default bridge network, then the client container can access the exposed
-port via a private networking interface.
-If `--link` is used when starting a container in a user-defined network as
-described in [*Networking overview*](https://docs.docker.com/network/),
-it will provide a named alias for the container being linked to.
+By default, when you run a container, none of the container's ports are exposed
+to the host. This means you won't be able to access any ports that the
```
thaJeztah, line 1049: "exposed" or "published" (mapped)?

docs/reference/run.md
```diff
@@ -1232,7 +1232,13 @@ The followings examples are all valid:
...
-root directory (`/`). It is possible to set a different working directory with the
-Dockerfile `WORKDIR` command. The operator can override this with:
+root directory (`/`). The default working directory of an image is set using
+the Dockerfile `WORKDIR` command. You can override the default working
+directory for an image using the `-w` (or `--workdir`) flag for the `docker
+run` command:
 
-    -w="", --workdir="": Working directory inside the container
+```text
+$ docker run -w /home/$USER -it alpine
```
thaJeztah, line 1241: We should probably avoid `$USER` here, as that depends on the variable being set on the host. Perhaps some other example that doesn't depend on an env-var?

Perhaps also make the example illustrate the result;


```bash
$ docker run --rm -w /my/workdir alpine pwd
/my/workdir
```

docs/reference/run.md
dvdksn: yeah, it didn't feel necessary including that here

docs/reference/run.md
dvdksn: agree

docs/reference/run.md
dvdksn: IMO, covering this is going into too much detail. If you really want to know about how the high ports are assigned, you can see the source code for the port allocator.

thaJeztah: Do we have this documented elsewhere? I know it's been a frequent topic for various setups with customers, and we should document how to prevent docker from conflicting with other tools (which may mean re-configuring `/proc/sys/net/ipv4/ip_local_port_rang`

dvdksn: OK; I added a `--publish-all` section in the CLI reference page, which included the same bits as highlighted here, plus this addition about `ip_local_port_rang`

docs/reference/commandline/run.md
```diff
@@ -372,16 +372,18 @@ Under these conditions, you can pass any size less than the backing filesystem s
...
 
-
 ### <a name="tmpfs"></a> Mount tmpfs (--tmpfs)
 
+The `--tmpfs` flag lets you create a `tmpfs` mount.
+
+The options that you can pass to `--tmpfs` are identical to the Linux `mount -t
+tmpfs -o` command. The following example mounts an empty `tmpfs` into the
+container with the `rw`, `noexec`, `nosuid`, `size=65536k` options.
+
```
thaJeztah, line 382: For a follow-up, we can consider cross-linking this to `--mount type=tmpfs` https://docs.docker.com/storage/tmpfs/#tmpfs-mounts

docs/reference/run.md
```diff
@@ -189,6 +189,82 @@ round-trip min/avg/max = 0.257/0.288/0.326 ms
...
+By default, the data in a container is stored in an ephemeral, writable
+container layer. Removing the container also removes its data. If you want to
+use persistent data with containers, you can use filesystem mounts to store the
+data persistently on the host system. Filesystem mounts can also let you share
+data between containers and the host.
+
+Docker supports two main categories of mounts:
+
+- Volume mounts
+- Bind mounts
```
thaJeztah, line 203: Related to my previous comment. Looks like this section is not as complete as https://docs.docker.com/storage/, and misses at least the `tmpfs` mounts.

docs/reference/run.md
```diff
@@ -1207,21 +1207,25 @@ The health status is also displayed in the `docker ps` output.
 
 ### User
 
-`root` (id = 0) is the default user within a container. The image developer can
-create additional users. Those users are accessible by name.  When passing a numeric
-ID, the user does not have to exist in the container.
+The default user within a container is `root` (id = 0). You can set a default
```
thaJeztah, line 1210: nit: `id` should probably be `uid` (and maybe mention `gid`


No need to change in this PR; looks like we should ideally also have some mention of "windows containers", where all of this will be quite different (not sure now from the top of my head if we even support `--user` for native Windows containers).

docs/reference/run.md
```diff
@@ -1230,7 +1230,14 @@ The followings examples are all valid:
...
+directory for an image using the `-w` (or `--workdir`) flag for the `docker
+run` command:
 
-    -w="", --workdir="": Working directory inside the container
+```text
+$ docker run --rm -w /my/workdir alpine pwd
+/my/workdir
+```
+
+If the directory doesn't already exist in the container, it will be created.
```
thaJeztah, line 1243: `s/wil be/is/`

No need to change; we should make a pass at removing future tense globally.

docs/reference/commandline/run.md
```diff
@@ -483,26 +483,47 @@ $ docker run -t -i --mount type=bind,src=/data,dst=/data busybox sh
...
+The `-P`, or `--publish-all`, flag publishes all the exposed ports to the host.
+Docker binds each exposed port to a random port on the host.
+
+The `-P` flag only publishes port numbers that are explicitly flagged as
+exposed, either using the Dockerfile `EXPOSE` instruction or the `--expose`
+flag for the `docker run` command.
+
+The range of ports are within an *ephemeral port range* defined by
+`/proc/sys/net/ipv4/ip_local_port_range`. Use the `-p` flag to explicitly map a
+single port or range of ports.
```
thaJeztah, line 525: Thanks! Sorry for being picky on that one, but I know it's been troublesome in the past, so trying to preserve the information.


Related to that, I need to check with @akerouanton and @robmry what options we currently have to influence these; I know there's some ports used by Windows itself that are within the ephemeral port range, and I know there's been issues with Swarm services picking their own range(s).

Perhaps we need configuration options for this on the daemon (not just a "range", but also options to _exclude_ range(s) or individual ports from being used).

docs/reference/commandline/run.md
```diff
@@ -483,26 +483,47 @@ $ docker run -t -i --mount type=bind,src=/data,dst=/data busybox sh
...
+with Docker.
 
-Note that ports which are not bound to the host (i.e., `-p 80:80` instead of
-`-p 127.0.0.1:80:80`) are externally accessible. This also applies if
-you configured UFW to block this specific port, as Docker manages its
-own iptables rules. [Read more](https://docs.docker.com/network/iptables/)
+> **Note**
+>
+> If you don't specify an IP address (i.e., `-p 80:80` instead of `-p
+> 127.0.0.1:80:80`) when publishing a container's ports, Docker publishes the
```
thaJeztah, line 497: Also for a follow-up; we need to make a pass at reviewing examples like this in light of IPv6 (i.e., both `127.0.0.x` and `::1` becoming more relevant). /cc @akerouanton @robmry

thaJeztah (approved):
LGTM!!

docs/reference/commandline/run.md
dvdksn: carried in https://github.com/docker/cli/pull/4712

docs/reference/commandline/run.md
dvdksn: carried in #4712

docs/reference/commandline/run.md
dvdksn: carried in #4712

docs/reference/run.md
dvdksn: yeah I didn't intend to make everything on this page exhaustive; it could put us back where this page started. My thinking is:

- General information here
- All the details in the reference docs

docs/reference/run.md
dvdksn: carried in https://github.com/docker/cli/pull/4712

docs/reference/commandline/run.md
dvdksn: carried in https://github.com/docker/cli/pull/4712

PR #4622, opened 2023-10-24
cli/command/network: runCreate: inline types.NetworkCreate

thaJeztah:
Some left-over commit from other changes; looks like I hadn't pushed this one


laurazard (approved):
LGTM

PR #4635, opened 2023-11-03
docs: add default-network-opt daemon option

dvdksn:

**- What I did**

Added a description for the default-network-opt flag and daemon config option

- Follow-up to moby/moby#43197

**- How I did it**

**- How to verify it**

moose

### DIFF
--- a/docs/reference/commandline/dockerd.md
+++ b/docs/reference/commandline/dockerd.md
@@ -46,6 +46,7 @@ Options:
       --default-gateway ip                    Container default gateway IPv4 address
       --default-gateway-v6 ip                 Container default gateway IPv6 address
       --default-ipc-mode string               Default mode for containers ipc ("shareable" | "private") (default "private")
+      --default-network-opt mapmap            Default network options (default map[])
       --default-runtime string                Default OCI runtime for containers (default "runc")
       --default-shm-size bytes                Default shm size for containers (default 64MiB)
       --default-ulimit ulimit                 Default ulimits for containers (default [])
@@ -1017,6 +1018,7 @@ This is a full example of the allowed configuration options on Linux:
   "default-cgroupns-mode": "private",
   "default-gateway": "",
   "default-gateway-v6": "",
+  "default-network-opts": {},
   "default-runtime": "runc",
   "default-shm-size": "64M",
   "default-ulimits": {
@@ -1134,6 +1136,7 @@ This is a full example of the allowed configuration options on Windows:
   "containerd-plugin-namespace": "docker-plugins",
   "data-root": "",
   "debug": true,
+  "default-network-opts": {},
   "default-runtime": "",
   "default-ulimits": {},
   "dns": [],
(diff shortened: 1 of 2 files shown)


dvdksn:
@thaJeztah I noticed this flag is not included in the man page either. Where do I update that?

thaJeztah:
@dvdksn dockerd man page is here; https://github.com/docker/cli/blob/5d6612798a195b30edf4244ff06f18e80b1167d5/man/dockerd.8.md#synopsis

You can add the option to the "synopsis" section and the list of options under it

docs/reference/commandline/dockerd.md
```diff
@@ -1279,3 +1282,25 @@ $ sudo dockerd \
         --data-root=/var/lib/docker-bootstrap \
         --exec-root=/var/run/docker-bootstrap
 ```
+
+### Default network options
+
+The `--default-network-opt` flag, and the equivalent `default-network-opts` key
+in the configuratio file, lets you specify custom default options for custom
```
thaJeztah, line 1289: "custom" twice in this sentence reads a bit awkward. Perhaps drop the first one;

```suggestion
in the configuratio file, lets you specify default options for custom
```

docs/reference/commandline/dockerd.md
```diff
@@ -1279,3 +1282,25 @@ $ sudo dockerd \
...
+networks.
+
+This option accepts a value with the format `driver=opt=value`, for example:
```
thaJeztah, line 1292: It may be good to mention that these options are driver-specific, and perhaps we should cross-reference where to find those options.

For the bridge driver, there's a table here; https://docs.docker.com/network/drivers/bridge/#options, but I haven't checked if there's a similar list for all of the other drivers, and what a good page would be to link to.

We could link to the overview page of networking;

```markdown
Available network options are specific to the network-driver. Refer to the 
[network drivers section](https://docs.docker.com/network/drivers/) to learn
about available options.
```

But that would still require users to try to find options for each driver (perhaps that's fine?), or we could of add a link to this specific example;

```markdown
This example uses the "bridge" network driver. Refer to the 
[bridge network driver page](https://docs.docker.com/network/drivers/bridge/#options)
for an overview of available driver options.
```

docs/reference/commandline/dockerd.md
```diff
@@ -1279,3 +1282,25 @@ $ sudo dockerd \
...
+
+```console
+$ sudo dockerd \
+  --default-network-opt bridge=com.docker.network.bridge.host_binding_ipv4=127.0.0.1 \
+  --default-network-opt bridge=com.docker.network.bridge.mtu=1234
+```
```
thaJeztah, line 1298: No need to change, because this matches current examples, but we should consider re-writing the examples to use `daemon.json`.

manually starting the daemon with command-line options works for debugging, but these examples can also be somewhat problematic if they're tried on a machine that already has `dockerd` running (most likely as a systemd service). In that scenario, running the example would try to start a _second_ instance of `dockerd`, which can conflict with the already running one; "best case scenario" that would produce an error, but "worst case", it may run, and potentially interfere / modify state of the running daemon.

I'm not sure what the best approach is for this though, because we also don't want to repeating "this is how you create or edit a `daemon.json` (for each platform, including Docker Desktop (?)), and this is how you start/restart/reload the daemon for testing" (in addition to some warnings "always try configuration changes in a test-environment first, etc).
 
Perhaps we need a tracking ticket for that, wdyt, @dvdksn ?

docs/reference/commandline/dockerd.md
```diff
@@ -1279,3 +1282,25 @@ $ sudo dockerd \
...
+
+New networks that you create inherit the specified options by default.
```
thaJeztah, line 1300: "by default" is a bit confusing here (when are they _not_ inherited?); perhaps "as default" would work? 

I was slightly wondering as well if "inherit" could be interpreted incorrectly, as it feels like "dynamically inherit options from the default"; which would be incorrect, because the defaults are "baked" into the newly created network. We have the same "issue" for logging-driver and logging-driver options; I see for those we have a "note" https://docs.docker.com/config/containers/logging/configure/#configure-the-default-logging-driver

> Changing the default logging driver or logging driver options in the daemon configuration only affects containers that are created after the configuration is changed. Existing containers retain the logging driver options that were used when they were created. 


But honestly that's also a bit awkward; perhaps we can find a better way to make it clear from the start that "these are used as defaults when creating new networks" (and make clear that they are "baked in").

thaJeztah:
Perhaps we should also have the networking section updated to (similar to the logging-driver options that can be set in `daemon.json`) mention that these defaults can be configured on the daemon.

docs/reference/commandline/dockerd.md
dvdksn: also a typo :D

docs/reference/commandline/dockerd.md
dvdksn: afaict, only the bridge and macvlan driver actually seem to have options? At least those are the only docs pages that list any options. Is that right @akerouanton ?

docs/reference/commandline/dockerd.md
dvdksn: yeah we can use a daemon.json example in this case instead

I don't think we need to go into the details of how to use this file here and everywhere else we mention it. But we should have better "entrypoint" documentation for how to work with this file. I would connect this to the grander plan of "reworking" of daemon configuration. Do we have a tracking for that?

dvdksn:
revised—PTAL @thaJeztah :)

docs/reference/commandline/dockerd.md
```diff
@@ -1017,6 +1018,7 @@ This is a full example of the allowed configuration options on Linux:
   "default-cgroupns-mode": "private",
   "default-gateway": "",
   "default-gateway-v6": "",
+  "default-network-opts": "",
```
akerouanton, line 1021:
```suggestion
  "default-network-opts": {},
```

docs/reference/commandline/dockerd.md
```diff
@@ -1134,6 +1136,7 @@ This is a full example of the allowed configuration options on Windows:
   "containerd-plugin-namespace": "docker-plugins",
   "data-root": "",
   "debug": true,
+  "default-network-opts": "",
```
akerouanton, line 1139:
```suggestion
  "default-network-opts": {},
```

akerouanton (changes requested):

akerouanton (approved):
LGTM! Thanks

thaJeztah (approved):
LGTM

thaJeztah:
@dvdksn if you have time to do a cherry-pick for the 24.0 branch, then this will make its way to the docs 😅

PR #4643, opened 2023-11-07
WIP: rewrite runAttach to be exported

thaJeztah:
- exploring alternatives to https://github.com/docker/cli/pull/4637


cli/command/container/attach.go
```diff
@@ -65,52 +77,53 @@ func NewAttachCommand(dockerCli command.Cli) *cobra.Command {
...
 	}
 
-	options := container.AttachOptions{
-		Stream:     true,
-		Stdin:      !opts.noStdin && c.Config.OpenStdin,
-		Stdout:     true,
-		Stderr:     true,
-		DetachKeys: detachKeys,
+	if opts.DetachKeys == "" {
+		opts.DetachKeys = dockerCli.ConfigFile().DetachKeys
```
thaJeztah, line 111: Ideally, we'd move this elsewhere so that this function can potentially only depend on the API-client, not the CLI 🤔

PR #4646, opened 2023-11-08
TestParseWithMacAddress: check endpoint-scoped MacAddress field

akerouanton:
- Supersedes / Closes https://github.com/docker/cli/pull/4645
- relates to https://github.com/moby/moby/pull/45905

This is a follow-up of https://github.com/docker/cli/pull/4419. That PR leveraged the fact that EndpointSettings.MacAddress is already available, although not used by the CreateNetwork endpoint.

TestParseWithMacAddress was testing whether the container-wide MacAddress field is set, and we still need to test that to ensure backward compatibility. But we now also need to test whether the endpoint-specific MacAddress is set.


thaJeztah (approved):
LGTM

PR #4655, opened 2023-11-10
Fix setting ServerAddress property in NativeStore

StealthyCoder:
This will return the ServerAddress property when using the NativeStore. This happens when you use docker credential helpers, not the credential store.

The reason this fix is needed is because it needs to be propagated properly down towards `moby/moby` project in the following logic:

```golang
func authorizationCredsFromAuthConfig(authConfig registrytypes.AuthConfig) docker.AuthorizerOpt {
	cfgHost := registry.ConvertToHostname(authConfig.ServerAddress)
	if cfgHost == "" || cfgHost == registry.IndexHostname {
		cfgHost = registry.DefaultRegistryHost
	}

	return docker.WithAuthCreds(func(host string) (string, string, error) {
		if cfgHost != host {
			logrus.WithFields(logrus.Fields{
				"host":    host,
				"cfgHost": cfgHost,
			}).Warn("Host doesn't match")
			return "", "", nil
		}
		if authConfig.IdentityToken != "" {
			return "", authConfig.IdentityToken, nil
		}
		return authConfig.Username, authConfig.Password, nil
	})
}
```
This logic resides in the following file :
`daemon/containerd/resolver.go` .

In the case when using the containerd storage feature when setting the `cfgHost` variable from the `authConfig.ServerAddress` it will always be empty. Since it will never be returned from the NativeStore currently. Therefore Docker Hub images will work fine, but anything else will fail since the `cfgHost` will always be the `registry.DefaultRegistryHost`.

**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah:
Thanks! We also merged a fix on the daemon-side in https://github.com/moby/moby/pull/46779

Perhaps _both_ won't do harm though


/cc @dmcgowan

cli/config/credentials/native_store.go
```diff
@@ -76,6 +77,7 @@ func (c *nativeStore) GetAll() (map[string]types.AuthConfig, error) {
 		ac.Username = creds.Username
 		ac.Password = creds.Password
 		ac.IdentityToken = creds.IdentityToken
+		ac.ServerAddress = creds.ServerAddress
```
dmcgowan, line 80: is `ac.ServerAddress` ever a non-empty value?

dmcgowan:
@thaJeztah agreed, both changes make sense

cli/config/credentials/native_store.go
StealthyCoder: I do not believe so no. Since in both stores the code always sets it to what the Registry Hostname is. Is your idea to just have creds be returned there directly?

thaJeztah: @dmcgowan do you mean you prefer to make this conditional to keep that option open?

Something like this?;

```go
if ac.ServerAddress  == "" {
    ac.ServerAddress = creds.ServerAddress
}
```

StealthyCoder: I had not read it that way yet. That makes sense. So we only set it when it is empty and not accidentally overwrite a potential value. I can add that real quick.

dmcgowan (approved):

thaJeztah (approved):
LGTM, thanks!

And thanks for the extra eyes, @dmcgowan

PR #4661, opened 2023-11-13
update authors and mailmap

thaJeztah:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,6 +2,7 @@
 # This file lists all contributors to the repository.
 # See scripts/docs/generate-authors.sh to make modifications.
 
+A. Lester Buck III <github-reg@nbolt.com>
 Aanand Prasad <aanand.prasad@gmail.com>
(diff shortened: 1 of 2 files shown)


AUTHORS
```diff
@@ -210,6 +223,7 @@ Denis Defreyne <denis@soundcloud.com>
 Denis Gladkikh <denis@gladkikh.email>
 Denis Ollier <larchunix@users.noreply.github.com>
 Dennis Docter <dennis@d23.nl>
+dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
```
sam-thibault, line 226: Hmm.. should we exclude bots?

thaJeztah: Yeah, was looking if there's an option for that, but don't think there is; https://git-scm.com/docs/gitmailmap

laurazard (approved):
LGTM

PR #4606, opened 2023-10-16
docs: add `bind-recursive` mount option

AkihiroSuda:
Follow-up to 
- #4316

PR #4662, opened 2023-11-16
[24.0 Backport] docs: add `bind-recursive` mount option

dvdksn:
- backport of #4606


sam-thibault (approved):

neersighted (changes requested):
https://github.com/moby/moby/pull/45278 was merged for v25; this should not be backported as the option does not exist.

PR #4685, opened 2023-11-30
export RunStat, StatsOptions, and add Filters option

ndeloof:
The filter option is not currently exposed on the command-line,
but can be added as a flag in future. It will be used by compose
to filter the list of containers to include based on compose
labels.

**- What I did**
~introduce `--filter` flag to select containers to be observed~
export `RunStat` and `StatOptions` so this command can be shared with Docker Compose to implement new `stats` command (https://github.com/docker/compose/issues/11234)

**- How I did it**

**- How to verify it**

// only for internal purpose, not expected to be "public" as a public API

image

### DIFF
--- a/cli/command/container/stats.go
+++ b/cli/command/container/stats.go
@@ -21,40 +21,58 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// statsOptions defines options for runStats.
-type statsOptions struct {
-	// all allows including both running and stopped containers. The default
+// StatsOptions defines options for [RunStats].
+type StatsOptions struct {
+	// All allows including both running and stopped containers. The default
 	// is to only include running containers.
-	all bool
+	All bool
 
-	// noStream disables streaming stats. If enabled, stats are collected once,
+	// NoStream disables streaming stats. If enabled, stats are collected once,
 	// and the result is printed.
-	noStream bool
+	NoStream bool
 
-	// noTrunc disables truncating the output. The default is to truncate
+	// NoTrunc disables truncating the output. The default is to truncate
 	// output such as container-IDs.
-	noTrunc bool
+	NoTrunc bool
 
-	// format is a custom template to use for presenting the stats.
+	// Format is a custom template to use for presenting the stats.
 	// Refer to [flagsHelper.FormatHelp] for accepted formats.
-	format string
+	Format string
 
-	// containers is the list of container names or IDs to include in the stats.
-	// If empty, all containers are included.
-	containers []string
+	// Containers is the list of container names or IDs to include in the stats.
+	// If empty, all containers are included. It is mutually exclusive with the
+	// Filters option, and an error is produced if both are set.
+	Containers []string
+
+	// Filters provides optional filters to filter the list of containers and their
+	// associated container-events to include in the stats if no list of containers
+	// is set. If no filter is provided, all containers are included. Filters and
+	// Containers are currently mutually exclusive, and setting both options
+	// produces an error.
+	//
+	// These filters are used both to collect the initial list of containers and
+	// to refresh the list of containers based on container-events, accepted
+	// filters are limited to the intersection of filters accepted by "events"
+	// and "container list".
+	//
+	// Currently only "label" / "label=value" filters are accepted. Additional
+	// filter options may be added in future (within the constraints described
+	// above), but may require daemon-side validation as the list of accepted
+	// filters can differ between daemon- and API versions.
+	Filters *filters.Args
 }
 
 // NewStatsCommand creates a new [cobra.Command] for "docker stats".
 func NewStatsCommand(dockerCLI command.Cli) *cobra.Command {
-	var options statsOptions
+	options := StatsOptions{}
 
 	cmd := &cobra.Command{
 		Use:   "stats [OPTIONS] [CONTAINER...]",
 		Short: "Display a live stream of container(s) resource usage statistics",
 		Args:  cli.RequiresMinArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			options.containers = args
-			return runStats(cmd.Context(), dockerCLI, &options)
+			options.Containers = args
+			return RunStats(cmd.Context(), dockerCLI, &options)
 		},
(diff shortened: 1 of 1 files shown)


cli/command/container/stats.go
```diff
@@ -48,27 +56,28 @@ func NewStatsCommand(dockerCli command.Cli) *cobra.Command {
...
+	showAll := len(options.Containers) == 0
 	closeChan := make(chan error)
 
 	ctx := context.Background()
 
 	// monitorContainerEvents watches for container creation and removal (only
 	// used when calling `docker stats` without arguments).
 	monitorContainerEvents := func(started chan<- struct{}, c chan events.Message, stopped <-chan struct{}) {
-		f := filters.NewArgs()
+		f := options.Filter.Value()
```
thaJeztah, line 80: I'm slightly confused here; should the `--filter` option apply to the _events_, or should the filtering be used for _listing_ the containers? (`getContainerList` below)?


For compose, I could imaging we want to get a list of containers that are part of the compose file, or does compose pass a list of container ID's to look at?

thaJeztah: (I guess I may have to look at all logic more closely)

ndeloof: you're right, filter must also apply to `getContainerList`
passing  `--filter label=compose.docker.compose.project=xx` allows to only select containers and events related to a compose project (also could reduce scope to only focus on a service)

thaJeztah: Was considering if we should make the "getContainerList" function something that can be passed in. Currently it's a bit of an odd combination of either passing in a hard-coded list of containers, or "all" containers, which uses that function.

If we would pass that function in as argument, perhaps that would allow compose (or other consumers) to do whatever logic they want to get the list of containers.

But honestly, haven't given that much thought yet (or looked what that would look like)

ndeloof:
My intent was for the filter to primarily be used "internally" by compose to filter containers by project label, but I had the feeling it could be useful for this command to select a subset of resources, same as `docker ps` does

thaJeztah:
> My intent was for the filter to primarily be used "internally" by compose to filter containers by project label, but I had the feeling it could be useful for this command to select a subset of resources, same as docker ps does

Perhaps, based on the above, it'd be best to move that change separate; would you be able to move that change to a separate PR, so that we could move this one forward for v25?

(also if you could check my examples, as I _think_ some parts didn't work as expected, so there could be a bug at hand)

ndeloof:
sure, updated to drop public `--filter` flag until we get some consensus on such a feature

thaJeztah:
Did a quick rebase and amended the commit message to remove the `--filter` mention, but it looks like there's still something wrong; with this build, `docker stats` (without arguments) hangs, and doesn't show stats;

```bash
./build/docker stats
```

With the v24.0 CLI (same daemon, same state);

```bash
CONTAINER ID   NAME                              CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O         PIDS
af4c1c63924e   zen_payne                         0.00%     8.098MiB / 7.663GiB   0.10%     2.54MB / 30.4kB   0B / 1.68MB       11
3e6d3e37862a   foo.1.wm6sej6mfaldkgb51me88rrww   0.00%     8.066MiB / 7.663GiB   0.10%     6.7kB / 0B        0B / 12.3kB       11
1f2ed82385d3   foobar                            0.00%     9.148MiB / 7.663GiB   0.12%     8.87kB / 0B       4.56MB / 442kB    11
c0d124d2465d   mypostgres                        0.01%     21.36MiB / 7.663GiB   0.27%     23.4kB / 0B       1.46MB / 57.6MB   6
477f387c7224   nomatch                           0.00%     6.906MiB / 7.663GiB   0.09%     23.5kB / 0B       0B / 1.24MB       11
a01e96c62d11   three                             0.00%     6.926MiB / 7.663GiB   0.09%     23.6kB / 0B       0B / 1.25MB       11
c5969117edfe   two                               0.00%     7.066MiB / 7.663GiB   0.09%     23.6kB / 0B       0B / 1.23MB       11
49db05bcc543   one                               0.00%     9.129MiB / 7.663GiB   0.12%     23.7kB / 0B       7.77MB / 1.23MB   11
```

thaJeztah:
Okay, I dug into this a bit, and I found the issue, but not sure yet how to resolve the problem and make filtering work as expected.

While looking at the code, I also did some refactoring / cleanup; I'll open a separate PR for that (currently included in this branch).

thaJeztah:
Hm... perhaps I was looking in the wrong direction; events allow filtering by label (so we have some mapping behind the scenes), and so does `container list` (which we call for the initial list of containers).

Perhaps it's due to both using the same filter, and these lines causing the filter to be updated, causing the `docker container list` to try to filter on "type";

```go
f := options.Filter.Value()
f.Add("type", string(events.ContainerEventType))
```

thaJeztah:
Yup, that's it; the container-list API call fails, which gets sent to the closeChannel, but doesn't surface anywhere; if I print the error, it shows;

```
./build/docker stats
Error response from daemon: invalid filter 'type'
```

thaJeztah:
I fixed this one, and it's now working, but I made some changes for the initial implementation;

- the `Filters` option is now a pointer (to make it more explicitly "optional"; leave it nil to not use a filter)
- I changed the `Filters` type back to a `filters.Args`; that is the filter type "only", without the added features to make it _directly_ usable for a cobra Flag. I did so trying to avoid it being too tightly coupled to "command-line", and it's easy to use an intermediate option for that purpose if needed later.
- currently, passing _BOTH_ a list of containers (IDs/Names) _AND_ a filter produces an error (we currently don't support that combination)
- due to how the filters are used, we can only use the subset of filters supported by_both_ `container list` AND `docker events`. We can somehow find out what is supported by making additional API calls, but I wanted to limit complexity for now, so I hard-coded the list of supported filters to `label` only (for now). I _THINK_ Compose is currently interested in `label` only for filtering, but we _could_ add additional ones if we really need it immediately.

thaJeztah (approved):
'LGTM"

last touch-ups were done by me, but I got an "LGTM" on slack

PR #4691, opened 2023-12-04
cleanup: don't use pkg/errors

imjasonh:
**- What I did**

Replaced uses of `github.com/pkg/errors` in this repo.

That package has been archived for 2 years, after `fmt.Errorf` added support for wrapping errors with `%w` in Go 1.13, more than 3 years ago.

There are still indirect dependencies on `pkg/errors` via github.com/docker/docker -- I can try to clean those up as well, if you're interested.

**- How I did it**

Mostly I ran this: https://github.com/zchee/go-analyzer/tree/main/pkgerrors

```
pkgerrors -fix ./...
```

In order to get that to work though, I first needed to find cases where `errors.Wrap` was called like `errors.Wrap(err, msg)` (where `msg` is an identifier and not a const string expression) or `errors.Wrap(err, foo())` (where `foo` is a method call and not a const string expression).

After that I replaced any remaining `errors.Is` calls to use Go's stdlib `errors.Is`, which now works with the `%w` added by the analyzer

**- How to verify it**

Thorough code review and reliance on tests

Remove use of deprecated/archived module github.com/pkg/errors

**TODO** (if you're interested in proceeding)

- [ ] fix lint findings now that `fmt.Errorf` with capitalized message trips the linter
- [ ] do github.com/docker/docker too

### DIFF
--- a/cli-plugins/manager/error.go
+++ b/cli-plugins/manager/error.go
@@ -1,7 +1,7 @@
 package manager
 
 import (
-	"github.com/pkg/errors"
+	"fmt"
 )
 
 // pluginError is set as Plugin.Err by NewPlugin if the plugin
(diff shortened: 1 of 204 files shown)


thaJeztah:
Thanks for contributing; sorry this took a while, but there were some discussions before this; this work was done in https://github.com/docker/cli/pull/6441 (and some related PRs), so I'm closing this PR.

PR #4698, opened 2023-12-06
Add otel support (carry 4556)

thaJeztah:
- carry / rebase of https://github.com/docker/cli/pull/4556

This is broken into 2 commits:

- Plumb contexts through commands
- Add otel tracing as a persistent pre-run to all CLI commands.

**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah:
I'll move the first commit (punching-through contexts) to a separate PR

cli/command/container/stats.go
```diff
@@ -125,13 +127,16 @@ func runStats(ctx context.Context, dockerCli command.Cli, opts *statsOptions) er
...
+		span := trace.SpanFromContext(ctx)
+
 		// If no names were specified, start a long running goroutine which
 		// monitors container events. We make sure we're subscribed before
 		// retrieving the list of running containers to avoid a race where we
 		// would "miss" a creation.
 		started := make(chan struct{})
 		eh := command.InitEventHandler()
 		eh.Handle(events.ActionCreate, func(e events.Message) {
+			span.AddEvent(string(events.ActionCreate), trace.WithAttributes(attribute.String("ID", e.ID)))
```
thaJeztah, line 139: Note to self; forgot to update this one to use `e.Actor.ID` (same as `e.ID`, but that's officially deprecated)

cli/command/container/stats.go
```diff
@@ -150,6 +156,7 @@ func runStats(ctx context.Context, dockerCli command.Cli, opts *statsOptions) er
 		})
 
 		eh.Handle(events.ActionDie, func(e events.Message) {
+			span.AddEvent(string(events.ActionDie), trace.WithAttributes(attribute.String("ID", e.ID)))
```
thaJeztah, line 159: Same here

cli/command/cli.go
```diff
@@ -92,6 +92,12 @@ type DockerCli struct {
 	baseCtx context.Context
 }
 
+// WithContext sets the base context for the cli
+// This is used internally for operations may require a context to propagate tracing.
+func (cli *DockerCli) WithContext(ctx context.Context) {
+	cli.baseCtx = ctx
+}
+
```
thaJeztah: Pushed a commit that re-implements this as a CLIOpt

cli/command/container/stats.go
thaJeztah: done 👍

cli/command/container/stats.go
thaJeztah: also done 👍

thaJeztah:
Thanks to @milas ❤️  I pushed a commit that swaps BuildKit detect with the OTEL SDK 🎉

thaJeztah:
I'll cleanup / squash commits a bit after review; just pushing the changes here first

cmd/docker/docker.go
```diff
@@ -44,7 +52,23 @@ func newDockerCommand(dockerCli *command.DockerCli) *cli.TopLevelCommand {
...
+			if err := isSupported(cmd, dockerCli); err != nil {
+				return err
+			}
+
+			name := cmd.Name()
+			for p := cmd.Parent(); p != nil && p != cmd.Root(); p = p.Parent() {
+				name = p.Name() + " " + name
+			}
+
+			ctx, _ := otel.Tracer("").Start(cmd.Context(), name)
```
thaJeztah, line 64: Just discussing with @milas that we don't have a good convention yet for passing the command name; compose currently uses this;

```
	ctx, cmdSpan := tracing.Tracer.Start(
		ctx,
		"cli/"+strings.Join(commandName(cmd), "-"),
	)
	cmdSpan.SetAttributes(attribute.StringSlice("cli.args", args))
	cmdSpan.SetAttributes(attribute.StringSlice("cli.flags", getFlags(cmd.Flags())))
```


that's what Compose is doing so it's slightly different, e.g. `docker compose alpha publish` -> `cli/alpha-publish`

thaJeztah: We can keep this as-is for now, and look at improving later; I could add the flags and args if we think it's useful to add.

cmd/docker/docker.go
```diff
@@ -309,6 +353,39 @@ func runDocker(dockerCli *command.DockerCli) error {
...
+	// `service.name` to avoid https://github.com/open-telemetry/opentelemetry-go/issues/2341
+	res, err := resource.Merge(
+		resource.Default(),
+		resource.NewSchemaless(
+			attribute.String("service.name", "docker-cli"),
+		),
+	)
+	if err != nil {
+		return fmt.Errorf("merging resource: %w", err)
+	}
```
milas, line 378: I realized this is actually redundant because we do `os.SetEnv("OTEL_SERVICE_NAME", cmd.Name())` if it's not set

You can just `res := resource.Default()`

milas (commented):
LGTM but you will need to add a call to `Shutdown()` to ensure the spans get flushed out when the process exits

milas:
Screenshot 2024-01-04 at 5 22 20 PM

Using my branch off your PR 🎉 

(also notice I snuck in some new attributes for the CLI version metadata 👀)

Start a Jaeger instance:
```
docker run --rm --name jaeger \
  -p 16686:16686 \
  -p 4317:4317 \
  -p 4318:4318 \
  jaegertracing/all-in-one:1.52
```
(adapted from https://www.jaegertracing.io/docs/latest/getting-started/#all-in-one)

```shell
make -f docker.Makefile build
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 ./build/docker --debug version
```

cmd/docker/docker.go
```diff
@@ -58,6 +82,10 @@ func newDockerCommand(dockerCli *command.DockerCli) *cli.TopLevelCommand {
 	cmd.SetOut(dockerCli.Out())
 	cmd.SetErr(dockerCli.Err())
 
+	// Cobra's context may be nil in some cases, so initialize it here.
```
cpuguy83, line 84: Cobra's context should _never_ be nil.\
A context should never be nil, ever.
Did something change upstream in cobra?
I remember specifically discussing this when contexts were added.

cmd/docker/docker.go
thaJeztah: Ah, nice! Just pushed a commit (to be squash) with what I think you meant 👍

cmd/docker/docker.go
```diff
@@ -274,6 +302,22 @@ func runDocker(dockerCli *command.DockerCli) error {
 		return err
 	}
 
+	// Buildkit's detect package currently follows the old otel spec which defaulted to gRPC.
+	// Since the spec changed to default to http/protobuf.
+	// If these env vars are not set then we set them to the new default so detect will give us the expected protocol.
+	// This is the same as on the dockerd side.
+	// This can be removed after buildkit's detect package is updated.
+	if os.Getenv("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL") == "" && os.Getenv("OTEL_EXPORTER_OTLP_PROTOCOL") == "" {
+		os.Setenv("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "http/protobuf")
```
cpuguy83, line 310: This probably is not needed since we aren't using detect anymore?

cli/command/container/stats.go
```diff
@@ -123,6 +125,8 @@ func RunStats(ctx context.Context, dockerCLI command.Cli, options *StatsOptions)
 
 	showAll := len(options.Containers) == 0
 	if showAll {
+		span := trace.SpanFromContext(ctx)
```
corhere, line 128: When in doubt, [create a nested span](https://opentelemetry.io/docs/instrumentation/go/manual/#create-nested-spans). It provides more detailed tracing insights than reusing a single top-level span for the entire lifetime of the process.

cmd/docker/docker.go
```diff
@@ -357,6 +353,39 @@ func runDocker(dockerCli *command.DockerCli) error {
 	return cmd.Execute()
 }
 
+func initializeTracing() error {
```
thaJeztah, line 355: Should we pass the context here that was set on the command?

cmd/docker/docker.go
```diff
@@ -44,7 +51,23 @@ func newDockerCommand(dockerCli *command.DockerCli) *cli.TopLevelCommand {
...
+			}
+
+			name := cmd.Name()
+			for p := cmd.Parent(); p != nil && p != cmd.Root(); p = p.Parent() {
+				name = p.Name() + " " + name
+			}
+
+			ctx, _ := otel.Tracer("").Start(cmd.Context(), name)
+			cmd.SetContext(ctx)
+			_ = dockerCli.Apply(command.WithContext(ctx))
```
thaJeztah, line 65: I.e. this one

corhere (changes requested):
Needs more spans. And I suggest splitting up the instrumentation and propagation work (with the OTEL API `go.opentelemetry.io/otel/...`) into separate commits from the exporter work (with the OTEL SDK `go.opentelemetry.io/otel/sdk/...`).

cmd/docker/docker.go
thaJeztah: Ha! This was from your original PR, but we were discussing this as well in https://github.com/docker/cli/pull/4599#discussion_r1422487271

I can remove this part 👍

thaJeztah:
Thanks all! I pushed a couple of commits (to be squashed if ok)

I'll be calling it a day for now, but if someone wants to push to this branch; feel free to.


Pushing additional commits should be fine; we can cleanup / squash later (I didn't bother cleaning up commits yet while we're moving this still)

cmd/docker/docker.go
milas: Yeah, it should go away - I verified things are still good without it

(@thaJeztah I deleted it over in https://github.com/thaJeztah/cli/pull/4/files#diff-3dcd6325b70a5f5db393a3ecc08060b69007fcef371fb4dabf3164da95646fb6L305-L312 )

cmd/docker/docker.go
```diff
@@ -309,6 +348,22 @@ func runDocker(dockerCli *command.DockerCli) error {
 	return cmd.Execute()
 }
 
+func initializeTracing(ctx context.Context) error {
```
thaJeztah, line 351: I added a context here

cmd/docker/docker.go
cpuguy83: Ugh... maybe I hit a case where it was nil.

2RXV3M
2nselq

cmd/docker/docker.go
```diff
@@ -274,6 +297,22 @@ func runDocker(dockerCli *command.DockerCli) error {
...
+	// This is the same as on the dockerd side.
+	// This can be removed after buildkit's detect package is updated.
+	if os.Getenv("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL") == "" && os.Getenv("OTEL_EXPORTER_OTLP_PROTOCOL") == "" {
+		os.Setenv("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "http/protobuf")
+	}
+	if v := os.Getenv("OTEL_SERVICE_NAME"); v == "" {
+		os.Setenv("OTEL_SERVICE_NAME", cmd.Root().Name())
+	}
+
+	if err := initializeTracing(cmd.Context()); err != nil {
```
thaJeztah, line 312: And passed `cmd.Context()` here, but double check if that's indeed the correct thing to do, or if it was deliberately a separate context 🙈

thaJeztah:
HA!

```
#19 53.89 === FAIL: cmd/docker TestClientDebugEnabled (0.00s)
#19 53.89 panic: cannot create context from nil parent [recovered]
#19 53.89 	panic: cannot create context from nil parent
```

cli/command/cli_options.go
```diff
@@ -115,3 +115,12 @@ func WithAPIClient(c client.APIClient) CLIOption {
...
 }
+
+// WithContext sets the base context for the cli
+// This is used internally for operations may require a context to propagate tracing.
+func WithContext(ctx context.Context) CLIOption {
+	return func(cli *DockerCli) error {
+		cli.baseCtx = ctx
+		return nil
+	}
+}
```
corhere, line 126: This is just a clone of `WithBaseContext` by another name

cmd/docker/docker.go
milas: If we have a `context.Context` available, it's probably worth passing through.

That said...I stepped through the source - neither `otlphttp` or `otlpgrpc` actually use it 🙃

cmd/docker/docker.go
milas: Docs on it are minimal - it appears to be intended for exporter setup, e.g. if it needs to dial a connection up-front or something I guess? [see my other comment though, it's actually unused currently by both the HTTP/gRPC exporters]

cli/command/cli_options.go
thaJeztah: LOLZ You're right! I forgot that was added already; I created this one because I didn't want to extend the DockerCli signature further.

Yup; this can be removed and we can use that one instead 👍

cmd/docker/docker.go
corhere: https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/

cmd/docker/docker.go
```diff
@@ -44,7 +51,23 @@ func newDockerCommand(dockerCli *command.DockerCli) *cli.TopLevelCommand {
...
+			}
+
+			ctx, _ := otel.Tracer("").Start(cmd.Context(), name)
+			cmd.SetContext(ctx)
+			_ = dockerCli.Apply(command.WithContext(ctx))
+			return nil
+		},
+		PersistentPostRun: func(cmd *cobra.Command, args []string) {
+			// TODO: There doesn't seem to be a way to determine if the command returned an error, so we can set the span status here.
+			trace.SpanFromContext(cmd.Context()).End()
```
corhere, line 70: You're fighting an uphill trying to use cobra lifecycle hooks for this. See below.

cmd/docker/docker.go
corhere, line 312: Initialize tracing in `main()`. Here's too late to be able to export traces from early in the process lifecycle.

cmd/docker/docker.go
```diff
@@ -309,6 +348,22 @@ func runDocker(dockerCli *command.DockerCli) error {
...
+
+	exporter, err := autoexport.NewSpanExporter(ctx)
+	if err != nil {
+		return fmt.Errorf("creating span exporter: %w", err)
+	}
+
+	otel.SetTracerProvider(sdktrace.NewTracerProvider(
+		sdktrace.WithSpanProcessor(sdktrace.NewBatchSpanProcessor(exporter)),
+		sdktrace.WithResource(resource.Default()),
+	))
```
corhere, line 362: The tracer provider needs to be shut down before the process exits to ensure the recorded spans are flushed to the exporter.

cmd/docker/docker.go
```diff
@@ -309,6 +348,22 @@ func runDocker(dockerCli *command.DockerCli) error {
 	return cmd.Execute()
```
corhere, line 348:
```suggestion
	ctx, span := otel.Tracer("github.com/docker/docker/cli/cmd/docker").Start(
		cmd.Context(),
		"runDocker",
	)
	defer span.End()
	err = cmd.ExecuteContext(ctx)
	if err != nil {
		span.SetStatus(codes.Error, "command errored")
		span.RecordError(err)
	}
	return err
```

It'd be even better to start the span at the top of the function, so the latency before the command is executed gets captured in the traces.

corhere (changes requested):

cmd/docker/docker.go
thaJeztah: > https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/

OH! That looks great, and from the looks of it that pretty much describes the things we'd want to include (and we wouldn't have to come up with our own conventions, which is a big 💯)

@milas @jsternberg You'll probably be interested in this (if you didn't know about these).

From the looks of it; most of these we could set using a generic helper (if none exists for it). We _MAY_ have to consider if we should sanitise arguments though (i.e. do we want `--env MYSQL_ROOT_PASSWORD=<pass>` (or `--build-arg`) to end up in traces? 😬)

If we decide to sanitise, perhaps we need an option to opt-out of sanitising (for those that _DO_ want to include all of that in their traces).

Let me add a screenshot to save others from having to follow the link, and to add context to my comment;

Screenshot 2024-01-05 at 12 28 37

PR #4705, opened 2023-12-09
Output valid json for docker ps --format='{{json .}}'

bdpiprava:
Attempt to fix https://github.com/moby/moby/issues/46906

**- What I did**
Extended [formatter.go](https://github.com/docker/cli/blob/master/cli/command/formatter/formatter.go) to output valid json array

**- How I did it**
Modified built-in functions in the [formatter.go](https://github.com/docker/cli/blob/master/cli/command/formatter/formatter.go)  to add required prefix and suffix to generate valid json array.

**- How to verify it**
1. Must have multiple docker containers in any state
2. Run `docker ps -a --format='{{json .}}'`

- Output valid json for `docker ps --format={{json .}}`

### DIFF
--- a/cli/command/formatter/formatter.go
+++ b/cli/command/formatter/formatter.go
@@ -63,6 +63,7 @@ func (c *Context) preFormat() {
 		c.finalFormat = c.finalFormat[len(TableFormatKey):]
 	case c.Format.IsJSON():
 		c.finalFormat = JSONFormat
+		c.buffer.WriteString("[")
 	}
 
 	c.finalFormat = strings.Trim(c.finalFormat, " ")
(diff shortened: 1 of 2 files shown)


bdpiprava:
@thaJeztah  - May I request some help on the PR?

AkihiroSuda (changes requested):
For compatibility reason, I suggest closing this.

The docs might need an update to explain the current behavior.

bdpiprava:
@AkihiroSuda - I understand that this change may cause the compatibility to break but from the use-case point of view when we say `docker ps --format json` the expected output should be a valid json array as `docker ps` is to list the containers. WDYT?

P.S. May I ask if you can elaborate more on `compatibility reason` ?

AkihiroSuda:
> @AkihiroSuda - I understand that this change may cause the compatibility to break but from the use-case point of view when we say `docker ps --format json` the expected output should be a valid json array as `docker ps` is to list the containers. WDYT?

Users have been depending on the current behavior, so it can't be fixed, regardless to its validity

> P.S. May I ask if you can elaborate more on `compatibility reason` ?

Changing the behavior will break existing users' scripts, so it is unlikely acceptable

bdpiprava:
As explained by @AkihiroSuda, I am closing the MR.

PR #4709, opened 2023-12-13
docs/deprecated: mark logentries log-driver for removal

thaJeztah:
relates to:

- https://github.com/moby/moby/pull/46925
- https://github.com/moby/moby/pull/46926

The service has been discontinued on November 15, 2022:

> Dear Logentries user,
>
> We have identified you as the owner of, or collaborator of, a Logentries
> account.
>
> The Logentries service will be discontinued on November 15th, 2022. This
> means that your Logentries account access will be removed and all your
> log data will be permanently deleted on this date.
>
> Next Steps
> If you are interested in an alternative Rapid7 log management solution,
> InsightOps will be available for purchase through December 16th, 2022.
> Please note, there is no support to migrate your existing Logentries
> account to InsightOps.
>
> Thank you for being a valued user of Logentries.
>
> Thank you,
> Rapid7 Customer Success

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -51,6 +51,7 @@ The table below provides an overview of the current status of deprecated feature
 | Status     | Feature                                                                                                                            | Deprecated | Remove |
 |------------|------------------------------------------------------------------------------------------------------------------------------------|------------|--------|
 | Deprecated | [IsAutomated field, and "is-automated" filter on docker search](#isautomated-field--and-is--automated-filter-on-docker-search)     | v25.0      | -      |
+| Deprecated | [logentries logging driver](#logentries-logging-driver)                                                                            | v24.0      | v25.0  |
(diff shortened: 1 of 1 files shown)


docs/deprecated.md
```diff
@@ -122,6 +123,14 @@ The `AUTOMATED` column has been removed from the default `docker search`
 and `docker image search` output in v25.0, and the corresponding `IsAutomated`
 templating option will be removed in v26.0.
 
+### Logentries logging driver
+
+**Target For Removal In Release: v25.0**
```
dvdksn, line 128: Add deprecated in

dvdksn (approved):
Nit/ maybe you left it out intentionally

docs/deprecated.md
thaJeztah: Ah, yes, I initially left it out, because I wasn't sure what version to pick; the service was discontinued and we never called it out explicitly to be deprecated because of that, so I picked "v24.0" for the table, but the most truthful would be "November 15, 2022" when the service was discontinued.

Let me add v24.0

Issue #4746, opened 2024-01-01
`docker manifest ls` command to list manifest list draft from local storage


wayne-cheng:
### Description

Firstly, happy new year to everyone on this first day of 2024!

Getting to the point, whenever I attempt to push or delete a manifest list, I find it challenging to identify which manifest lists have already been created in local storage. While I know they are stored in the directory `~/.docker/manifests`, the filenames in the directory are not human-readable. These names cannot guarantee a straightforward conversion to the correct manifest list name because the char `/` is replaced with underline `_`. When you attempt to reverse this conversion to obtain the manifest list name, you cannot be certain whether the underline `_` should be replaced with `/` or left unchanged.

To implement the command `manifest list`, it is essential for the directory names stored in ~/.docker/manifests to be uniquely identifiable. So I decide to replace the char '/' with double underlines `__` and the colon `:` with double dashs (`--`). Since image names  are not allowed to  contain double underlines or double dashs, this change does not impact other manifest subcommands.

In the end, I have successfully implemented the desired command locally, like this:

```bash
# docker manifest ls
REPOSITORY          TAG
example.com/first    1
example.com/second   2

# docker manifest ls -q
example.com/first:1
example.com/second:2

```

After using the `docker manifest ls` command, I can easily find the target manifest list to push or delete. It's very convenient！

I am willing to submit a PR, and it could be my first contribution to the Docker Cli project.

wayne-cheng:
@thaJeztah Hello, I created the PR #4751, please take a look.

PR #4751, opened 2024-01-04
Add `docker manifest ls` command

wayne-cheng:
**- What I did**
When you want to push or delete a manifest list, you need to know the name of the manifest list you have created. So I implemented a new sub-command `docker manifest ls`：

```bash
$ docker manifest ls --help
Usage:  docker manifest ls [OPTIONS]

List local manifest lists

EXPERIMENTAL:
  docker manifest ls is an experimental feature.
  Experimental features provide early access to product functionality. These
  features may change between releases without warning, or can be removed from a
  future release. Learn more about experimental features in our documentation:
  https://docs.docker.com/go/experimental/

Aliases:
  docker manifest ls, docker manifest list

Options:
      --format string   Format output using a custom template:
                        'table':            Print output in table format with column headers (default)
                        'table TEMPLATE':   Print output in table format using the given Go template
                        'json':             Print in JSON format
                        'TEMPLATE':         Print output using the given Go template.
                        Refer to https://docs.docker.com/go/formatting/ for more information about formatting output with templates
  -q, --quiet           Only show manifest list NAMEs
```

**- How I did it**

More detail, see the issue #4746 that I opened.

**- How to verify it**
The command output like this:

```bash
REPOSITORY                                   TAG
example.com/library/test-busybox             v1
example.com/library/test-busybox             v2
```

Add sub-command docker manifest ls


contrib/completion/bash/docker
```diff
@@ -4140,6 +4140,11 @@ _docker_manifest() {
 		inspect
 		push
 		rm
+		ls
+	"
+
+	local aliases="
+		list
 	"
 	__docker_subcommands "$subcommands" && return
```
albers, line 4149:
```suggestion
	__docker_subcommands "$subcommands $aliases" && return
```

contrib/completion/bash/docker
```diff
@@ -4140,6 +4140,11 @@ _docker_manifest() {
 		inspect
 		push
 		rm
+		ls
```
albers, line 4143: please insert in alphabetic order.

contrib/completion/bash/docker
```diff
@@ -4249,6 +4254,14 @@ _docker_manifest_rm() {
 	esac
 }
```
albers, line 4256: Please insert
```shell script
_docker_manifest_list() {
  _docker_manifest_ls
}
```
here so that the options for `list` are also completed.

contrib/completion/bash/docker
```diff
@@ -4249,6 +4254,14 @@ _docker_manifest_rm() {
...
 
+_docker_manifest_ls() {
```
albers, line 4257: Please insert
```bash
	case "$prev" in
		--format)
			return
			;;
	esac
```
because `--format` has an argument and we do not want completion to trigger here.

albers (changes requested):
Looks sane, but I'm not a go expert.

Bash completion needs some additions, see comments.

contrib/completion/bash/docker
wayne-cheng: ok

contrib/completion/bash/docker
wayne-cheng: done

contrib/completion/bash/docker
wayne-cheng: done

contrib/completion/bash/docker
wayne-cheng: done

albers (approved):
Perfect, thanks!

wayne-cheng:
About the replacement of special characters, I make a mistake. On Docker Hub or other registry like Harbor, the image repo name does not allow repeat special characters (`-` and `_`),  but naming images locally allows the use of repeat special characters, possibly due to historical reasons. So the method I used for special character replacement was not good enough. I think it is safer to replace `/` with `#` and `:` with `%`, and I have fixed this now. 

BTW,  platforms information is useful, and I print it now, the output:

```bash
REPOSITORY           TAG       PLATFORMS
example.com/first    1         linux/amd64, linux/arm64
example.com/second   2         linux/amd64
```

neersighted (changes requested):
Adding a quick blocking review here; we in general want to rip this out in favor of something more like the `docker buildx imagetools` implementation (but first there are UX/compatibility questions to answer, as well as looking at adding support for local manifests when the containerd storage backend is in use).

wayne-cheng:
@neersighted I agree, but the command  `docker buildx imagetools` works on images in remote **registry**, and currently only the `docker manifest` could manage **local** manifest lists. Additionally, if using containerd as the storage backend, does it require using the OCI image index type instead of docker manifest list? This could require some adaptation work.

neersighted:
The 'local manifests' are stored in `~/.docker` and isn't really a well-supported/stable mechanism; indeed, when I talk about `imagetools` I do expect that we would extend it to manage local (stored in the containerd backend) content.

Regarding the mediaType, the only (effective) difference is mediaType and we support both transparently. There is no major implementation challenge with manifest lists vs OCI index.

PR #4752, opened 2024-01-04
ci: Add bin-image workflow 

vvoland:
**- What I did**
Build and push an image containing a static CLI binary for master branch and every release branch and tag.

This is a slightly adjusted copy of the bin-image workflow from moby/moby (by @crazy-max).

**- How I did it**
See individual commits.

**- How to verify it**
Tested against my fork:
## `v29.0.0` tag: https://github.com/vvoland/cli/actions/runs/7410112486/job/20161818586
```bash
$ printf 'FROM alpine
   COPY --from=pawelgronowski465/cli-bin:29.0.0 /docker /docker
   RUN /docker version' | docker build -
...
0.119 Client:
0.119  Version:           29.0.0
0.119  API version:       1.44
0.119  Go version:        go1.21.5
0.119  Git commit:        d1b8d3ce50
0.119  Built:             Thu Jan  4 13:26:07 2024
0.119  OS/Arch:           linux/arm64
0.119  Context:           default
```
## `29.0` branch: https://github.com/vvoland/cli/actions/runs/7409929402/job/20161236975
```bash
$ printf 'FROM alpine
   COPY --from=pawelgronowski465/cli-bin:29.0 /docker /docker
   RUN /docker version' | docker build -
...
0.162 Client:
0.162  Version:           29.0
0.162  API version:       1.44
0.162  Go version:        go1.21.5
0.162  Git commit:        a1f9f0d681
0.162  Built:             Thu Jan  4 13:10:11 2024
0.162  OS/Arch:           linux/arm64
0.162  Context:           default
...
```

## `ci-bin-image-test` branch: https://github.com/vvoland/cli/actions/runs/7410075016/job/20161631291
```bash
$ rintf 'FROM alpine
   COPY --from=pawelgronowski465/cli-bin:ci-bin-image-test /docker /docker
   RUN /docker version' | docker build -
...
0.153 Client:
0.153  Version:           ci-bin-image-test
0.153  API version:       1.44
0.153  Go version:        go1.21.5
0.153  Git commit:        d1b8d3ce50
0.153  Built:             Thu Jan  4 13:22:07 2024
0.153  OS/Arch:           linux/arm64
0.153  Context:           default
...
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  VERSION: ${{ github.ref }}
+
 on:
   workflow_dispatch:
   push:
@@ -85,6 +88,50 @@ jobs:
           name: ${{ env.ARTIFACT_NAME }}
           path: /tmp/out/*
           if-no-files-found: error
+
+  bin-image:
+    runs-on: ubuntu-20.04
+    if: ${{ github.event_name != 'pull_request' && github.repository == 'docker/cli' }}
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v4
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: dockereng/cli-bin
+          tags: |
+            type=semver,pattern={{version}}
+            type=ref,event=branch
+            type=ref,event=pr
+            type=sha
+      -
+        name: Login to DockerHub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_CLIBIN_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_CLIBIN_TOKEN }}
+      -
+        name: Build and push image
+        uses: docker/bake-action@v4
+        with:
+          files: |
+            ./docker-bake.hcl
+            ${{ steps.meta.outputs.bake-file }}
+          targets: bin-image-cross
+          push: ${{ github.event_name != 'pull_request' }}
+          set: |
+            *.cache-from=type=gha,scope=bin-image
+            *.cache-to=type=gha,scope=bin-image,mode=max
 
   prepare-plugins:
     runs-on: ubuntu-20.04
--- a/Dockerfile
+++ b/Dockerfile
@@ -124,4 +124,4 @@ FROM scratch AS plugins
 COPY --from=build-plugins /out .
 
 FROM scratch AS binary
-COPY --from=build /out .
+COPY --from=build /out/docker /docker
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -165,3 +165,26 @@ target "e2e-gencerts" {
     dockerfile = "./e2e/testdata/Dockerfile.gencerts"
     output = ["./e2e/testdata"]
 }
+
+target "docker-metadata-action" {
+  tags = ["cli-bin:local"]
+}
+
+target "bin-image" {
+  inherits = ["binary", "docker-metadata-action"]
+  output = ["type=docker"]
+}
+
+target "bin-image-cross" {
+  inherits = ["bin-image"]
+  output = ["type=image"]
+  platforms = [
+    "linux/amd64",
+    "linux/arm/v6",
+    "linux/arm/v7",
+    "linux/arm64",
+    "linux/ppc64le",
+    "linux/s390x",
+    "windows/amd64"
+  ]
+}
--- a/scripts/build/.variables
+++ b/scripts/build/.variables
@@ -22,6 +22,13 @@ else
     BUILDTIME=${BUILDTIME:-$(TZ=UTC date -u --date="@${SOURCE_DATE_EPOCH:-$(date +%s)}" +"%Y-%m-%dT%H:%M:%SZ")}
 fi
 
+case "$VERSION" in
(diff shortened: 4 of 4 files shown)


.github/workflows/bin-image.yml
```diff
@@ -0,0 +1,181 @@
...
+        name: Upload digest
+        if: github.event_name != 'pull_request' && github.repository == ${{ env.GITHUB_REPO }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: digests
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  merge:
```
crazy-max, line 142: Don't think we need to distribute the build, it's fast enough when running on the smae runner imo and would simplify the workflow.

See buildx one https://github.com/docker/buildx/blob/671347dc35039f44e2eeb4578995e7f48e7fd4d0/.github/workflows/build.yml#L192-L242

crazy-max: So can just be part of build workflow https://github.com/docker/cli/blob/master/.github/workflows/build.yml

vvoland: Right, changed. Thanks!

.github/workflows/build.yml
```diff
@@ -86,6 +90,49 @@ jobs:
           path: /tmp/out/*
           if-no-files-found: error
 
+  bin-image:
+    runs-on: ubuntu-20.04
+    if: ${{ github.event_name != 'pull_request' && github.repository == 'docker/cli' }}
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v4
```
crazy-max, line 99: You might want the same fetch-depth input as build so correct ref is returned in https://github.com/docker/cli/blob/8812e0ad3ad786d6aa9f6289df4656f096d493a0/scripts/build/.variables#L13

```suggestion
        uses: actions/checkout@v4
        with:
          fetch-depth: 0
```

crazy-max: Oh my bad didn't catch you were setting `VERSION` env. Looks good then!

vvoland: Yeah, I did set the `VERSION` explicitly, so we shouldn't need the whole history.

.github/workflows/build.yml
```diff
@@ -86,6 +90,49 @@ jobs:
...
+        uses: docker/metadata-action@v5
+        with:
+          images: dockereng/cli-bin
+          tags: |
+            type=semver,pattern={{version}}
+            type=ref,event=branch
+            type=ref,event=pr
+            type=sha
+      -
+        name: Login to DockerHub
```
crazy-max, line 118: Needs to skip login on PR:

```suggestion
        name: Login to DockerHub
        if: github.event_name != 'pull_request'
```

.github/workflows/build.yml
```diff
@@ -86,6 +90,49 @@ jobs:
...
+          password: ${{ secrets.DOCKERHUB_CLIBIN_TOKEN }}
+      -
+        name: Build and push image
+        uses: docker/bake-action@v4
+        with:
+          files: |
+            ./docker-bake.hcl
+            ${{ steps.meta.outputs.bake-file }}
+          targets: bin-image-cross
+          push: true
```
crazy-max, line 131: Same here for push

```suggestion
          push: ${{ github.event_name != 'pull_request' }}
```

.github/workflows/build.yml
vvoland: Yes it will be removed once the PR is undrafted, this is all a part of a separate debug commit that just enables the worflow to run for this PR: https://github.com/docker/cli/pull/4752/commits/23c5809c47244cf363630d7361b8bb5fc58fb370

Dockerfile
```diff
@@ -124,4 +124,4 @@ FROM scratch AS plugins
 COPY --from=build-plugins /out .
 
 FROM scratch AS binary
-COPY --from=build /out .
+COPY --from=build /out/docker /docker
```
thaJeztah, line 127: Curious why this `COPY` had to be changed; did we add more things in `/out` ? Where possible,  I like “copy all the things” so that the part copying things doesn’t need to know what to look for

(basically "`/out`" being the "contract" / API (for better words)

vvoland: `scripts/build/binary`  produces two files, first is `docker-<platform>-<arch>` which is an actual binary, and the second is a `docker` which is a symlink to the former.

So without this change the image content would be:
```
/docker-<platform>-<arch>        # the actual binary
/docker                          # symlink to the above
```


The `COPY` copies the /docker which dereferences into the actual binary and puts that directly as `/docker` in the final image. So the image only contains the `docker` binary directly.

It's not strictly needed, because it should work with `COPY --from=dockereng/cli /docker /docker` anyway, but I'm not sure if we really need that symlink in the final image.


Alternatively I could look into changing the `scripts/build/binary` so that it doesn't output the symlink, but that would probably require changing other things.

thaJeztah: Ah, yes, forgot about the symlink; yeah. not a blocker, just something "nice if we can improve"

vvoland:
FYI, the secrets are filled and the PR is ready for review.

crazy-max (approved):
LGTM

thaJeztah (approved):
LGTM; let's get this one in and check if it all works 👍

Dockerfile
thaJeztah: LOL, and of course I didn't think of docker-ce-packaging using this output, and now failng;

```
dest=$PWD/build/mac; cd /home/ubuntu/workspace/release-packaging_ce-nightly/packaging/src/github.com/docker/cli/build && for platform in *; do \
	arch=$(echo $platform | cut -d_ -f2); \
	mkdir -p $dest/$arch/docker; \
	cp $platform/docker-darwin-* $dest/$arch/docker/docker && \
	tar -C $dest/$arch -c -z -f $dest/$arch/docker-25.0.0-rc.2.tgz docker; \
done
cp: cannot stat 'darwin_amd64/docker-darwin-*': No such file or directory
cp: cannot stat 'darwin_arm64/docker-darwin-*': No such file or directory
```

PR #4754, opened 2024-01-05
cli/command/container: parseSecurityOpts: fix --security-opt seccomp=builtin

thaJeztah:
- relates to https://github.com/moby/moby/pull/42481

Docker v23.0 and up allow the daemon to be configured to have seccomp disabled by default (using the "unconfined" profile as default), and introduced a new "builtin" profile-name for the default (see [moby@f8795ed364586acd][1] and [mnoby@ac449d6b5ad29a50][2]).

However, the CLI had no special handling for the "builtin" profile, which resulted in it trying to load it as a file, which would fail;

    docker run -it --rm --security-opt seccomp=builtin busybox
    docker: opening seccomp profile (builtin) failed: open builtin: no such file or directory.
    See 'docker run --help'.

This patch adds a special case for the "builtin" profile, to allow using the default profile on daemons with seccomp disabled (unconfined) by default.

[1]: https://github.com/moby/moby/commit/f8795ed364586acd93f72e206a409e7e0e27edcc
[2]: https://github.com/moby/moby/commit/ac449d6b5ad29a5086824729ce54eec6b0cc8545

**- How to verify it**

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -28,6 +28,20 @@ import (
 	cdi "tags.cncf.io/container-device-interface/pkg/parser"
 )
 
+const (
+	// TODO(thaJeztah): define these in the API-types, or query available defaults
+	//  from the daemon, or require "local" profiles to be an absolute path or
+	//  relative paths starting with "./". The daemon-config has consts for this
+	//  but we don't want to import that package:
+	//  https://github.com/moby/moby/blob/v23.0.0/daemon/config/config.go#L63-L67
+
+	// seccompProfileDefault is the built-in default seccomp profile.
+	seccompProfileDefault = "builtin"
+	// seccompProfileUnconfined is a special profile name for seccomp to use an
+	// "unconfined" seccomp profile.
+	seccompProfileUnconfined = "unconfined"
+)
+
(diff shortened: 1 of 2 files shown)


cli/command/container/opts.go
```diff
@@ -914,16 +928,24 @@ func parseSecurityOpts(securityOpts []string) ([]string, error) {
...
-				return securityOpts, errors.Errorf("opening seccomp profile (%s) failed: %v", v, err)
-			}
-			b := bytes.NewBuffer(nil)
-			if err := json.Compact(b, f); err != nil {
-				return securityOpts, errors.Errorf("compacting json for seccomp profile (%s) failed: %v", v, err)
+		if k == "seccomp" {
+			switch v {
+			case seccompProfileDefault, seccompProfileUnconfined:
+				// known special names for built-in profiles
+				securityOpts[key] = "seccomp=" + v
```
vvoland, line 935: This will be a no-op right?  If so, maybe we can just leave this branch empty?

thaJeztah: Doh! Yes, you're right. Honestly I got confused, and later on found that `securityOpts[key]` is _NOT_ a map, but an array, and it's directly manipulating the input variable.

I really want to change that, as that feels ... icky (in either case, change `key` to `i` or something!)

I'll remove this line 👍

thaJeztah: Fixed 👍

PR #4760, opened 2024-01-08
cli/compose/loader: use golden.Assert() for readability

thaJeztah:
golden.AssertBytes prints the failure as a bytes-array, which makes it not human-readable; let's compare strings instead.

**- What I did**

**- How I did it**

**- How to verify it**


laurazard (approved):
LGTM

PR #4764, opened 2024-01-09
cli/config/configfile(autoRemove): New config.json option.

### LINKED ISSUE #4763: Allow setting AutoRemove (`--rm`) by default in `config.json`
### Description

As a user, i'd like to be able to set something like `"autoRemove": true` in my `~/.docker/config.json` file, so that i can run a command as follows:

```
docker run -it debian
```

...and have it be equivalent to:

```
docker run --rm -it debian
```

Because i always forget the `--rm` flag, and end up with hundreds of dangling containers in my Docker dashboard.

I'm not proposing changing the default, just allowing me a persistent way to set the flag.  I've had a look at the source and i don't believe there's any environment variable that i could set that would have this effect.

I'll open a small PR that enables the feature i suggest, but i'm sure it won't be ready as-is, because i'll have forgotten cases or whatever, but i'd like to have a starting point for the discussion.

toothbrush:
Allow user to set `"autoRemove": true` in their `~/.docker/config.json` to have `--rm=true` be set by default for `docker run -it <containter> -- sh` commands.

**- What I did**

I surfaced an option in the `~/.docker/config.json` to always have `--rm=true` when running containers, so that my container list doesn't get cluttered.  It's called `"autoRemove": true` to match the code, but that's perhaps not the best name.

Fixes #4763.

**- How I did it**

I added a field in the ConfigFile struct, and used it in the `container run` command.

**- How to verify it**

I ran the binary with and without `"autoRemove": true` in the `config.json` file, and observed that the flag was set correctly, both with "printf debugging" and observing my stopped containers list.

* Added `autoRemove` boolean option to `config.json`, so that `--rm` is implicit with running `docker run ...`.

image
(Canada Goose from https://commons.wikimedia.org/wiki/File:Canada_goose_gosling_-_natures_pics.jpg)

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -104,7 +104,9 @@ func runRun(ctx context.Context, dockerCli command.Cli, flags *pflag.FlagSet, ro
 		}
 	}
 	copts.env = *opts.NewListOptsRef(&newEnv, nil)
-	copts.autoRemove = dockerCli.ConfigFile().AutoRemove
+    if !flags.Changed("rm") {
+   	copts.autoRemove = dockerCli.ConfigFile().AutoRemove
+   }
 	containerCfg, err := parse(flags, copts, dockerCli.ServerInfo().OSType)
 	// just in case the parse does not exit
 	if err != nil {
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -104,6 +104,7 @@ func runRun(ctx context.Context, dockerCli command.Cli, flags *pflag.FlagSet, ro
 		}
 	}
 	copts.env = *opts.NewListOptsRef(&newEnv, nil)
+	copts.autoRemove = dockerCli.ConfigFile().AutoRemove
 	containerCfg, err := parse(flags, copts, dockerCli.ServerInfo().OSType)
 	// just in case the parse does not exit
 	if err != nil {
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -41,6 +41,7 @@ type ConfigFile struct {
 	CLIPluginsExtraDirs  []string                     `json:"cliPluginsExtraDirs,omitempty"`
 	Plugins              map[string]map[string]string `json:"plugins,omitempty"`
 	Aliases              map[string]string            `json:"aliases,omitempty"`
+	AutoRemove           bool                         `json:"autoRemove,omitempty"`
 }
 
 // ProxyConfig contains proxy configuration settings


cli/command/container/run.go
```diff
@@ -104,6 +104,7 @@ func runRun(ctx context.Context, dockerCli command.Cli, flags *pflag.FlagSet, ro
 		}
 	}
 	copts.env = *opts.NewListOptsRef(&newEnv, nil)
+	copts.autoRemove = dockerCli.ConfigFile().AutoRemove
```
vvoland, line 107: This will always overwrite the `autoRemove` with the value from `config.json` (so `false` if it's not available). regardless of the passed CLI flags.

We should still allow the user to pass `--rm=false` to override the default `autoRemove: true`.

```suggestion
    if !flags.Changed("rm") {
   	copts.autoRemove = dockerCli.ConfigFile().AutoRemove
   }
```

cli/config/configfile/file.go
```diff
@@ -41,6 +41,7 @@ type ConfigFile struct {
 	CLIPluginsExtraDirs  []string                     `json:"cliPluginsExtraDirs,omitempty"`
 	Plugins              map[string]map[string]string `json:"plugins,omitempty"`
 	Aliases              map[string]string            `json:"aliases,omitempty"`
+	AutoRemove           bool                         `json:"autoRemove,omitempty"`
```
vvoland, line 44: Considering the above, would `DefaultAutoRemove` be a better name for this?

```suggestion
	DefaultAutoRemove           bool                         `json:"defaultAutoRemove,omitempty"`
```

cli/command/container/run.go
toothbrush: Ah! Good point. I always get a bit confused with Cobra options 🐍

cli/config/configfile/file.go
toothbrush: Having said that i think it'd be worth thinking about how to model other CLI flags in config.json – i'm not wild about the idea of a special "defaultXYZ" struct because of the idiom i refer to above, but on the other hand i would also understand not wanting to clutter the `ConfigFile` struct 🤔 Up to you!

thaJeztah (changes requested):
I think this needs a wider discussion, and I'd rather not add an "ad-hoc" field to this config. More so because this would currently work for `docker run`, but won't set it for `docker create`, nor (e.g.) `docker compose run`.

I know that "defaults" have been discussed in the past, but we need to think over the implications, which options we want to allow having a default, but also make sure that these defaults are discoverable, otherwise behavior will be confusing (different behavior depending on this config, which can be a huge head-ache trying to find "what happened" when debugging).

toothbrush:
Thanks for the thoughtful responses so far!

> I think this needs a wider discussion, and I'd rather not add an "ad-hoc" field to this config. More so because this would currently work for `docker run`, but won't set it for `docker create`, nor (e.g.) `docker compose run`.
> 
> I know that "defaults" have been discussed in the past, but we need to think over the implications, which options we want to allow having a default, but also make sure that these defaults are discoverable, otherwise behavior will be confusing (different behavior depending on this config, which can be a huge head-ache trying to find "what happened" when debugging).

Yes, completely understand and agree.  I am keen for this option though – what would be (my?) next steps to help drive something like this along?  I'm happy to provide strawman PRs and so on.

PR #4767, opened 2024-01-10
docs(plugins): fix typos and formatting

dvdksn:
**- What I did**

Fix typos and formatting cruft for the plugin docs

Diff is proper messy but it's all indentation and double spaces and typos, things like that

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/extend/index.md
+++ b/docs/extend/index.md
@@ -1,24 +1,14 @@
 ---
+title: Docker Engine managed plugin system
 description: Develop and use a plugin with the managed plugin system
 keywords: "API, Usage, plugins, documentation, developer"
 ---
-
-<!-- This file is maintained within the docker/cli GitHub
-     repository at https://github.com/docker/cli/. Make all
-     pull requests against that repo. If you see this file in
-     another repository, consider it read-only there, as it will
-     periodically be overwritten by the definitive file. Pull
-     requests which include edits to this file in other repositories
-     will be rejected.
--->
-
-# Docker Engine managed plugin system
 
 - [Installing and using a plugin](index.md#installing-and-using-a-plugin)
 - [Developing a plugin](index.md#developing-a-plugin)
 - [Debugging plugins](index.md#debugging-plugins)
 
-Docker Engine's plugin system allows you to install, start, stop, and remove
+Docker Engine's plugin system lets you install, start, stop, and remove
 plugins using Docker Engine.
 
 For information about legacy (non-managed) plugins, refer to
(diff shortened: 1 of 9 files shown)


docs/extend/index.md
```diff
@@ -1,24 +1,14 @@
...
 ---
 
-<!-- This file is maintained within the docker/cli GitHub
-     repository at https://github.com/docker/cli/. Make all
-     pull requests against that repo. If you see this file in
-     another repository, consider it read-only there, as it will
-     periodically be overwritten by the definitive file. Pull
-     requests which include edits to this file in other repositories
-     will be rejected.
--->
```
thaJeztah, line 13: If we don't need these anymore; perhaps you can open a PR to remove them in all files.

(wondering though if this file vendored as-is, so this would still be visible?)

dvdksn: yeah it's vendored as markdown, but they're in the _vendor directory and we have a contribution guideline outlining this. So I think it's fine to remove.

docs/extend/plugin_api.md
```diff
@@ -43,26 +33,26 @@ user or container tries to use one by name.
 
 There are three types of files which can be put in the plugin directory.
 
-* `.sock` files are UNIX domain sockets.
+* `.sock` files are Unix domain sockets.
```
thaJeztah, line 36: Ah! TIL `UNIX` is the trademarked name, and `Unix` is the "common" name ("Unix-like"); https://unix.stackexchange.com/a/532341/78656

docs/extend/plugin_api.md
```diff
@@ -43,26 +33,26 @@ user or container tries to use one by name.
...
 * `.spec` files are text files containing a URL, such as `unix:///other.sock` or `tcp://localhost:8080`.
 * `.json` files are text files containing a full json specification for the plugin.
 
-Plugins with UNIX domain socket files must run on the same docker host, whereas
-plugins with spec or json files can run on a different host if a remote URL is specified.
+Plugins with Unix domain socket files must run on the same Docker host.
```
thaJeztah, line 40: Perhaps use something similar here as well ("same host as the Docker daemon"); we should probably avoid "Docker host", as it's not very clear what it is.

docs/extend/plugins_logging.md
```diff
@@ -46,76 +36,77 @@ receiving logs for.
...
-```
-[size][message]
-```
-
-Where `size` is a 4-byte big endian binary encoded uint32. `size` in this case
-defines the size of the next message. `message` is the actual log entry.
-
-A reference golang implementation of a stream encoder/decoder can be found
-[here](https://github.com/docker/docker/blob/master/api/types/plugins/logdriver/io.go)
+- Request
```
thaJeztah, line 39: Slight bit on the fence on using bullet-lists for this; do you have an example what it looks like when rendered in the docs? Does it significantly improve readability?

dvdksn: not super happy with the bullet list either but we shouldn't use a bolded paragraph as a heading. Here's what it currently looks like: 
image

dvdksn: I don't want to make them actual headings either... so without actually restructuring this stuff more this was the best I could come up with. I think it's fine, especially for these pages.

thaJeztah: Would "plain" (non-bold) work? e.g.


```
Request example:

```json
{
  "File": "/path/to/file/stream"
}
```

On a successful request, the `Err` field is left empty:

```json
{
  "Err": ""
}
```

If an error occurred .......

```json
{
  "Err": "an error occurred"
}
```

dvdksn: yeah that would work. I can change it

dvdksn:
changed the bullet lists to just `Request:` / `Response:`, think it's good now @thaJeztah

thaJeztah (approved):
LGTM

PR #4769, opened 2024-01-10
plugins: run plugin with new process group ID

laurazard:
Changes were made in https://github.com/docker/cli/pull/4599 to provide a mechanism for the CLI to notify running plugin processes that they should exit, in order to improve the general CLI/plugin UX. The current implementation boils down to:
1. The CLI creates a socket
2. The CLI executes the plugin
3. The plugin connects to the socket
4. (When) the CLI receives a termination signal, it uses the socket to notify the plugin that it should exit
5. The plugin's gets notified via the socket, and cancels it's `cmd.Context`, which then gets handled appropriately

This change works in most cases and fixes the issue it sets out to solve (see: https://github.com/docker/compose/pull/11292) however, in the case where the user has a TTY attached and the plugin is not already handling received signals, steps 4+ changes:
4. (When) the CLI receives a termination signal, before it can use the socket to notify the plugin that it should exit, the plugin process also receives a signal due to sharing the pgid with the CLI

Since we now have a proper "job control" mechanism, we can simplify the scenarios by executing the plugins with their own process group id, thereby removing the "double notification" issue and making it so that plugins can handle the same whether attached to a TTY or not.

In order to make this change "plugin-binary" backwards-compatible, in the case that a plugin does not connect to the socket, the CLI passes the signal to the plugin process.

-------------------

**- What I did**

Execute plugins with their own process group id. If a(n older) plugin does not connect to the CLI socket, pass received signals onto the plugin process, so as to preserve compatibility with older plugin binaries.

**- How to verify it**

IMG_2069

### DIFF
--- a/cli-plugins/manager/manager_unix.go
+++ b/cli-plugins/manager/manager_unix.go
@@ -2,7 +2,21 @@
 
 package manager
 
+import (
+	"os/exec"
+	"syscall"
+)
+
 var defaultSystemPluginDirs = []string{
 	"/usr/local/lib/docker/cli-plugins", "/usr/local/libexec/docker/cli-plugins",
 	"/usr/lib/docker/cli-plugins", "/usr/libexec/docker/cli-plugins",
+}
+
+func configureOSSpecificCommand(cmd *exec.Cmd) {
+	// Spawn the plugin process in a new process group, so that signals are not forwarded by the OS.
+	// The foreground process group is e.g. sent a SIGINT when Ctrl-C is input to the TTY, but we
+	// implement our own job control for the plugin.
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
+	}
 }
--- a/cli-plugins/manager/manager.go
+++ b/cli-plugins/manager/manager.go
@@ -232,6 +232,7 @@ func PluginRunCommand(dockerCli command.Cli, name string, rootcmd *cobra.Command
 		cmd.Stdin = os.Stdin
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
+		configureOSSpecificCommand(cmd)
 
 		cmd.Env = os.Environ()
 		cmd.Env = append(cmd.Env, ReexecEnvvar+"="+os.Args[0])
--- a/cli-plugins/manager/manager_windows.go
+++ b/cli-plugins/manager/manager_windows.go
@@ -2,10 +2,15 @@ package manager
 
 import (
 	"os"
+	"os/exec"
 	"path/filepath"
 )
 
 var defaultSystemPluginDirs = []string{
 	filepath.Join(os.Getenv("ProgramData"), "Docker", "cli-plugins"),
 	filepath.Join(os.Getenv("ProgramFiles"), "Docker", "cli-plugins"),
+}
+
+func configureOSSpecificCommand(cmd *exec.Cmd) {
+	// no-op
 }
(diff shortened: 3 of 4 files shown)


cli-plugins/manager/manager_unix.go
```diff
@@ -2,7 +2,18 @@
...
+)
+
 var defaultSystemPluginDirs = []string{
 	"/usr/local/lib/docker/cli-plugins", "/usr/local/libexec/docker/cli-plugins",
 	"/usr/lib/docker/cli-plugins", "/usr/libexec/docker/cli-plugins",
 }
+
+func configureOSSpecificCommand(cmd *exec.Cmd) {
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
```
thaJeztah, line 17: I was hoping this was abstracted in the `syscall` package (and that to be the reason to use `syscall` instead of `golang.org/x/sys`), but looks like it's not 😢 (otherwise we could've dropped the abstraction)

Well.. it is what it is 🤷‍♂️

thaJeztah:
(GitHub _STILL_ doesn't allow commenting on commit messages, so putting it here);


Screenshot 2024-01-11 at 14 07 55


It's worth considering to refer to the _commit_ (`1554ac3b5f38147301c518c60da16f3f80c1717b`) in the commit message, instead of the PR. Doing so prevents some noise if commits are cherry-picked or merged into other branches (which can be in any fork), plus it allows finding back the related changes without requiring access to GitHub.

(I usually refer to the commit in the commit message, and add a link to the PR in the GitHub description of the PR).

cli-plugins/manager/manager_unix.go
thaJeztah: Would it be worth adding a comment why we're doing this (writing down the intent; it looks like you already have a good description in the commit message)? Or do you think it's clear from the code alone?

(A good way to judge: would _you_ understand (without looking in history) 1 Year from now when looking at the code?)

thaJeztah (commented):
code changes look good afaics; left some comments  / suggestions 😅

laurazard:
> It's worth considering to refer to the commit (1554ac3b5f38147301c518c60da16f3f80c1717b) in the commit message, instead of the PR. Doing so prevents some noise if commits are cherry-picked or merged into other branches (which can be in any fork), plus it allows finding back the related changes without requiring access to GitHub.

That's funny, I usually do this because almost every time I'm doing Github archaeology, if I have to find the commit I want to go to the PR to get additional context. However, the noise and accessibility outside of Github are good arguments, and Github makes it pretty easy now to find the PR a commit came from, so I'll start doing what you're suggesting!

cli-plugins/manager/manager_unix.go
laurazard: I'm too used to having to do dig through many years old commits to understand code that I don't even notice anymore 😞 I'll add some more comments!

thaJeztah: Yup, I'm pretty sure we'd be able to trace it back if needed. I started to pay more attention to these things (sorry for being the victim of that), for all those times I had to spend "too much time" finding back the reason.

Having a comment also can help deciding making changes ("oh! thanks for the description; this is soooooo last year and really no longer needed. I can confidently remove this part").

neersighted (approved):
This looks excellent to me, especially with the addition of explicit signal-forwarding. Like @thaJeztah, my only suggestion is that we add comments explaining why we set the pgid/why we forward the signal in some cases, but not all.

thaJeztah:
Yup; I'm good with merging after the commit message is updated and a comment added 👍 (thought I'd wait for you to look at that)

neersighted:
I've carried this with the requested comments and commit message change.

thaJeztah (approved):
LGTM, thanks!

thaJeztah:
Looks like this introduced a regression with buildx running with `BUILDX_EXPERIMENTAL` enabled;

- https://github.com/moby/moby/issues/47073

PR #4774, opened 2024-01-11
improve passing context

thaJeztah:
Relates to OTEL changes and graceful exit changes

See below PRs
- https://github.com/docker/cli/pull/4698
- https://github.com/docker/cli/pull/4993

Create a base context in the main function and use it as the base context in cobra commands. This allows context cancellation throughout the CLI. 

### DIFF
--- a/cli-plugins/plugin/plugin.go
+++ b/cli-plugins/plugin/plugin.go
@@ -36,13 +36,7 @@ func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager
 	PersistentPreRunE = func(cmd *cobra.Command, _ []string) error {
 		var err error
 		persistentPreRunOnce.Do(func() {
-			cmdContext := cmd.Context()
-			// TODO: revisit and make sure this check makes sense
-			// see: https://github.com/docker/cli/pull/4599#discussion_r1422487271
-			if cmdContext == nil {
-				cmdContext = context.TODO()
-			}
-			ctx, cancel := context.WithCancel(cmdContext)
+			ctx, cancel := context.WithCancel(cmd.Context())
 			cmd.SetContext(ctx)
 			// Set up the context to cancel based on signalling via CLI socket.
 			socket.ConnectAndWait(cancel)
--- a/cmd/docker/builder_test.go
+++ b/cmd/docker/builder_test.go
@@ -23,6 +23,9 @@ import (
 var pluginFilename = "docker-buildx"
 
 func TestBuildWithBuilder(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
 	testcases := []struct {
 		name         string
 		context      string
@@ -64,12 +67,16 @@ echo '{"SchemaVersion":"0.1.0","Vendor":"Docker Inc.","Version":"v0.6.3","ShortD
 	for _, tt := range testcases {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
+			ctx2, cancel2 := context.WithCancel(ctx)
+			defer cancel2()
+
 			if tt.builder != "" {
 				t.Setenv("BUILDX_BUILDER", tt.builder)
 			}
 
 			var b bytes.Buffer
 			dockerCli, err := command.NewDockerCli(
+				command.WithBaseContext(ctx2),
 				command.WithAPIClient(&fakeClient{}),
 				command.WithInputStream(discard),
 				command.WithCombinedStreams(&b),
@@ -126,6 +133,9 @@ func (c *fakeClient) Ping(_ context.Context) (types.Ping, error) {
 }
 
 func TestBuildkitDisabled(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
 	t.Setenv("DOCKER_BUILDKIT", "0")
 
 	dir := fs.NewDir(t, t.Name(),
@@ -136,6 +146,7 @@ func TestBuildkitDisabled(t *testing.T) {
 	b := bytes.NewBuffer(nil)
 
 	dockerCli, err := command.NewDockerCli(
+		command.WithBaseContext(ctx),
 		command.WithAPIClient(&fakeClient{}),
 		command.WithInputStream(discard),
 		command.WithCombinedStreams(b),
@@ -163,6 +174,9 @@ func TestBuildkitDisabled(t *testing.T) {
 }
 
 func TestBuilderBroken(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
 	dir := fs.NewDir(t, t.Name(),
 		fs.WithFile(pluginFilename, `#!/bin/sh exit 1`, fs.WithMode(0o777)),
 	)
@@ -171,6 +185,7 @@ func TestBuilderBroken(t *testing.T) {
 	b := bytes.NewBuffer(nil)
 
 	dockerCli, err := command.NewDockerCli(
+		command.WithBaseContext(ctx),
 		command.WithAPIClient(&fakeClient{}),
 		command.WithInputStream(discard),
 		command.WithCombinedStreams(b),
@@ -199,6 +214,8 @@ func TestBuilderBroken(t *testing.T) {
 
 func TestBuilderBrokenEnforced(t *testing.T) {
 	t.Setenv("DOCKER_BUILDKIT", "1")
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
 
 	dir := fs.NewDir(t, t.Name(),
 		fs.WithFile(pluginFilename, `#!/bin/sh exit 1`, fs.WithMode(0o777)),
@@ -208,6 +225,7 @@ func TestBuilderBrokenEnforced(t *testing.T) {
 	b := bytes.NewBuffer(nil)
 
 	dockerCli, err := command.NewDockerCli(
+		command.WithBaseContext(ctx),
 		command.WithAPIClient(&fakeClient{}),
 		command.WithInputStream(discard),
 		command.WithCombinedStreams(b),
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -29,6 +29,7 @@ import (
 
 func main() {
 	ctx := context.Background()
+
 	dockerCli, err := command.NewDockerCli(command.WithBaseContext(ctx))
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
@@ -352,7 +353,7 @@ func runDocker(ctx context.Context, dockerCli *command.DockerCli) error {
 	// We've parsed global args already, so reset args to those
 	// which remain.
 	cmd.SetArgs(args)
-	err = cmd.Execute()
+	err = cmd.ExecuteContext(ctx)
 
 	// If the command is being executed in an interactive terminal
 	// and hook are enabled, run the plugin hooks.
--- a/cmd/docker/docker_test.go
+++ b/cmd/docker/docker_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"io"
 	"os"
 	"testing"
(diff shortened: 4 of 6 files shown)


cmd/docker/builder_test.go
```diff
@@ -144,7 +145,8 @@ func TestBuildkitDisabled(t *testing.T) {
 	assert.NilError(t, dockerCli.Initialize(flags.NewClientOptions()))
 	dockerCli.ConfigFile().CLIPluginsExtraDirs = []string{dir.Path()}
 
-	tcmd := newDockerCommand(dockerCli)
+	ctx := context.TODO()
```
Benehiko, line 148: I think we need to use the `fakeClient.Ping()` context here instead of a new context. If we want the `newDockerCommand` to have it's own context with a cancel, we could wrap it with `context.WithCancel(parentCtx)`.

cmd/docker/builder_test.go
```diff
@@ -98,7 +98,8 @@ echo '{"SchemaVersion":"0.1.0","Vendor":"Docker Inc.","Version":"v0.6.3","ShortD
 				dockerCli.ConfigFile().Aliases = map[string]string{"builder": "buildx"}
 			}
 
-			tcmd := newDockerCommand(dockerCli)
+			ctx := context.TODO()
```
Benehiko, line 101: I think we should create a `context.Background()` instance at the top of the test function and have it consumed by each function that requires a context.  

> Background returns a non-nil, empty [Context](https://pkg.go.dev/context#Context). It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.

cmd/docker/docker_test.go
```diff
@@ -18,7 +19,8 @@ func TestClientDebugEnabled(t *testing.T) {
 
 	cli, err := command.NewDockerCli()
 	assert.NilError(t, err)
-	tcmd := newDockerCommand(cli)
+	ctx := context.TODO()
```
Benehiko, line 22: same here, I think using `context.Background()` is a valid approach for tests.

cmd/docker/builder_test.go
thaJeztah: I think (but it's been a while since I wrote this) the idea here was that we may be wiring up OTEL contexts in tests at some point, so the `TODO` context is a reminder for those to be wired up (I recall we used those for find (/replace) in moby/moby).

From a "functional" perspective, they should be identical (empty context) I think?


I can have a look if we're creating this one in the right spot; I may have kept them in this spot as a reminder to set a span for each of the sub-tests, e.g. see https://github.com/moby/moby/blob/cba87125b25a637f5ab02ae28c87c71e83913c6b/integration/build/build_test.go#L24

cli-plugins/plugin/plugin.go
```diff
@@ -27,21 +27,15 @@ import (
...
 		persistentPreRunOnce.Do(func() {
-			cmdContext := cmd.Context()
-			// TODO: revisit and make sure this check makes sense
-			// see: https://github.com/docker/cli/pull/4599#discussion_r1422487271
-			if cmdContext == nil {
-				cmdContext = context.TODO()
-			}
-			ctx, cancel := context.WithCancel(cmdContext)
-			cmd.SetContext(ctx)
+			ctx2, cancel := context.WithCancel(cmd.Context())
```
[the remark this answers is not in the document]
Benehiko: no i think we should use the cobra context here, since cobra context will be set to the context passed in to the `RunPlugin` function here https://github.com/docker/cli/pull/4774/files/7f65f3a4ffc644ac23f4c519ac3d98971ca61ccb#diff-5525052fe59ca68d6a9837832220a8c362270aa47a8520c1e2e4e43798c41dc8R27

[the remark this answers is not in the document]
laurazard: Maybe another way to think about this is "if not for setting the context that is going to be used for my commands later on, _**what is**_ happening with this `ctx`? What happens if I cancel it?
cc @thaJeztah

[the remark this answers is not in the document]
laurazard: Oh I misread the change (that's what I get for reviewing from my phone) 😓 

Sounds good, ty!

laurazard (approved):
LGTM

Benehiko (approved):

laurazard:
@thaJeztah can you TA(nother)L?

krissetto (approved):

cli-plugins/plugin/plugin.go
```diff
@@ -27,21 +27,15 @@ import (
 var PersistentPreRunE func(*cobra.Command, []string) error
 
 // RunPlugin executes the specified plugin command
-func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager.Metadata) error {
-	tcmd := newPluginCommand(dockerCli, plugin, meta)
+func RunPlugin(ctx context.Context, dockerCli *command.DockerCli, plugin *cobra.Command, meta manager.Metadata) error {
```
vvoland, line 30: Are we fine with a breaking change in the public function signature that's used by every CLI plugin?

vvoland (dismissed):
Leaving request changes for now; I don't think we should merge this for v26 since it's a breaking change to the `plugin.Run` function.

cli-plugins/plugin/plugin.go
Benehiko: Yeah this will break plugins on the next version. This is a tricky one since **not** passing a context could cause problems further down the line (like graceful cleanup). And I'd like to move away from using `context.TODO()` or `context.Background()` instead in these kind of places. 

The next best thing would be to use the `dockerCli.baseCtx` but this is an anti-pattern in terms of context and we should actually look to remove this from the `command.DockerCli` struct.

cli-plugins/plugin/plugin.go
```diff
@@ -67,18 +61,18 @@ func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager
 }
 
 // Run is the top-level entry point to the CLI plugin framework. It should be called from your plugin's `main()` function.
-func Run(makeCmd func(command.Cli) *cobra.Command, meta manager.Metadata) {
+func Run(ctx context.Context, makeCmd func(command.Cli) *cobra.Command, meta manager.Metadata) {
 	otel.SetErrorHandler(debug.OTELErrorHandler)
 
-	dockerCli, err := command.NewDockerCli()
+	dockerCli, err := command.NewDockerCli(command.WithBaseContext(ctx))
```
Benehiko, line 67: Do plugins add their own ctx to the dockerCli with the `makeCmd`?

krissetto: compose seems to use whatever cobra's context is (https://github.com/docker/compose/blob/85567ae09246ea79191e35f060d230aa5b7c0e5b/cmd/compose/compose.go#L384)

buildx uses this func defined in buildkit though, but watch out since it already handles signals (https://github.com/moby/buildkit/blob/master/util/appcontext/appcontext.go)

Benehiko: I see, yeah this will be breaking for plugins. I'll see what I can do to work around that.  

>but watch out since it already handles signals

that's fine, this PR doesn't do any signal handling here, but this is a good point and will work around this

cmd/docker/docker.go
```diff
@@ -352,7 +353,7 @@ func runDocker(ctx context.Context, dockerCli *command.DockerCli) error {
 	// We've parsed global args already, so reset args to those
 	// which remain.
 	cmd.SetArgs(args)
-	err = cmd.Execute()
+	err = cmd.ExecuteContext(ctx)
```
Benehiko, line 356: take note, instead of passing `ctx` to the root command through convoluted function calls I've cleaned it up a bit by instead passing context to cobra when we call `ExecuteContext`. This _should_ only affect the CLI and not any plugins that are run through the CLI.

cli/cobra.go
```diff
@@ -23,8 +22,7 @@ import (
 
 // setupCommonRootCommand contains the setup common to
 // SetupRootCommand and SetupPluginRootCommand.
-func setupCommonRootCommand(ctx context.Context, rootCmd *cobra.Command) (*cliflags.ClientOptions, *cobra.Command) {
-	rootCmd.SetContext(ctx)
```
Benehiko: removed this here since we can use the cobra `ExecuteContext` instead. This prevents us from passing in `ctx` to the plugins through `RunPlugin`.

cli-plugins/plugin/plugin.go
Benehiko: removed this - see new comments on the PR for more details.

vvoland (approved):

cli-plugins/plugin/plugin.go
```diff
@@ -36,14 +36,8 @@ func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager
...
 		persistentPreRunOnce.Do(func() {
-			cmdContext := cmd.Context()
-			// TODO: revisit and make sure this check makes sense
-			// see: https://github.com/docker/cli/pull/4599#discussion_r1422487271
-			if cmdContext == nil {
-				cmdContext = context.TODO()
-			}
-			ctx, cancel := context.WithCancel(cmdContext)
-			cmd.SetContext(ctx)
+			ctx2, cancel := context.WithCancel(cmd.Context())
```
krissetto, line 39: nit: i guess this can be called `ctx` again

krissetto (approved):
LGTM

cli-plugins/plugin/plugin.go
Benehiko: oh true

cmd/docker/docker.go
```diff
@@ -352,7 +353,10 @@ func runDocker(ctx context.Context, dockerCli *command.DockerCli) error {
...
+	if err != nil {
+		return err
```
laurazard, line 358: Did this get borked during a rebase? It's returning early, should be returning after hooks get called I think

cmd/docker/docker.go
thaJeztah: Ah, yes, this was one of the things to look into!

(Still catching up on latest changes in this PR, but yes, this was one where my approach was probably not right)

cmd/docker/docker.go
thaJeztah: Oh! Should it still handle context errors early though? (cancelled context? or timeout?) I guess in that case the hooks should _not_ be called?

laurazard: I think it's fine to do it regardless in any case here? The plugin can decide if it wants to print something or not. Even if it's cancelled, a hook might want to add some hint/explain something.

laurazard (approved):
LGTM

cmd/docker/docker.go
thaJeztah: Yeah, not sure if a cancelled context should still be used; isn't context cancelation basically; "cancel my request" (on the first possible occasion) ? In some cases that may mean some work is needed to complete the transaction (flush to disk, e.g.) but calling hooks (and execute other binaries) feels like that should not be part of that.

laurazard: that's valid, sure. @Benehiko wdyt

krissetto: mmh i'm not sure if hooks should be called after a cancellation, it'd feel wrong to me

Benehiko: I don't have too much opinion on this either way. We currently execute the hooks even when the cobra `Execute` function returns an error (which is a similar thing to context cancellation).

Cancelling the context early will still execute the cobra commands, but the commands checking for context cancellation will just return (with maybe an error?). 

See below an example of a cobra command not acting on the context cancellation. Even if it was acting and returning an error it will still execute.


```go
package main

import (
	"context"
	"fmt"

	"github.com/spf13/cobra"
)

func main() {
	ctx, cancel := context.WithCancel(context.Background())

	rootCmd := &cobra.Command{
		Use: "",
			cmd.Help()
			fmt.Printf("Cobra command ctx: %v\n", cmd.Context().Err())
			return nil
		},
	}

	cancel()

	fmt.Printf("Before Cobra ctx: %v\n", ctx.Err())

	err := rootCmd.ExecuteContext(ctx)
	fmt.Printf("%v\n", err)
}
```

```sh
⋊> ~/G/test-cobra go run .                        09:14:06
Before Cobra ctx: context canceled
Usage:
   [flags]

Flags:
  -h, --help   help for this command
Cobra command ctx: context canceled
<nil>
```

Benehiko: See here what the code currently looks like on the master branch
https://github.com/docker/cli/blob/master/cmd/docker/docker.go#L355-L367

vvoland:
Ah, one more thing, can you please add the changelog description with highlighting the API change?

Benehiko:
> Ah, one more thing, can you please add the changelog description with highlighting the API change?
> 
> EDIT: Sorry, I confused PRs 🙈 There was no API change in the end here (we didn't alter the `RunPlugin` signature), so no need for changelog!

@vvoland  You had the PR right, I just reverted those changes in the PR and squashed the commits

Issue #4780, opened 2024-01-12
[25.0.0-rc.1] regression: CLI leaves behind  plugin socket mount-points


thaJeztah:
### Description

The plugin-sockets added in https://github.com/docker/cli/pull/4599 appear to be leaving behind mount-points

### Reproduce

Create an empty directory, and run a command that executes a plugin;

```bash
mkdir -p myproject && cd myproject

docker buildx version
github.com/docker/buildx v0.12.0-desktop.2 c5a13b51c1ae9358eb691e9a21c955590e26d0a0
```

Check the content of the directory;

```bash
ls -l
srwxr-xr-x  1 thajeztah  staff   0 Jan 12 10:51 @docker_cli_43243a8d-599e-4844-a193-b617dc5cd222
```

When building without BuildKit enabled, these also cause warnings to be printed as they end up in the local directory (which usually is the build-context);

```bash
DOCKER_BUILDKIT=0 docker build --no-cache -f- . <<'EOF'
FROm alpine
EOF

DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            BuildKit is currently disabled; enable it by removing the DOCKER_BUILDKIT=0
            environment-variable.

ERRO[0000] Can't add file /Users/thajeztah/Projects/myproject/@docker_cli_43243a8d-599e-4844-a193-b617dc5cd222 to tar: archive/tar: sockets not supported
Sending build context to Docker daemon  1.583kB
Step 1/1 : FROm alpine
 ---> 34871e729050
Successfully built 34871e729050
```

### docker version

```bash
docker version
Client:
 Cloud integration: v1.0.35+desktop.7
 Version:           25.0.0-rc.1
 API version:       1.44
 Go version:        go1.21.5
 Git commit:        8812e0a
 Built:             Thu Jan  4 16:27:00 2024
 OS/Arch:           darwin/arm64
 Context:           desktop-linux
```

### docker info

```bash
not relevant
```

### Additional Info

We should also check what happens if a plugin would run a container that bind-mounts the current directory. For example, if Docker Compose runs a container, and bind-mounts the directory, it would likely bind-mount the socket into the container (with all possible fun consequences of that?)

PR #4783, opened 2024-01-12
cli-plugins: don't use abstract sockets on macOS

laurazard:
- fixes https://github.com/docker/cli/issues/4780

macOS does not support abstract sockets, and as such the current implementation is problematic since the sockets are created in the current working directory, which alters build contexts and can break builds if bind-mounted.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli-plugins/plugin/plugin.go
+++ b/cli-plugins/plugin/plugin.go
@@ -3,25 +3,18 @@ package plugin
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
-	"io"
-	"net"
 	"os"
 	"sync"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli-plugins/manager"
+	"github.com/docker/cli/cli-plugins/socket"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/connhelper"
 	"github.com/docker/docker/client"
 	"github.com/spf13/cobra"
 )
-
-// CLIPluginSocketEnvKey is used to pass the plugin being
-// executed the abstract socket name it should listen on to know
-// when the CLI has exited.
-const CLIPluginSocketEnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
 
 // PersistentPreRunE must be called by any plugin command (or
 // subcommand) which uses the cobra `PersistentPreRun*` hook. Plugins
@@ -32,38 +25,6 @@ const CLIPluginSocketEnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
 // PersistentPreRunE hook and must not be run unless Run has been
 // called.
 var PersistentPreRunE func(*cobra.Command, []string) error
-
-// closeOnCLISocketClose connects to the socket specified
-// by the DOCKER_CLI_PLUGIN_SOCKET env var, if present, and attempts
-// to read from it until it receives an EOF, which signals that
-// the CLI is going to exit and the plugin should also exit.
-func closeOnCLISocketClose(cancel func()) {
-	socketAddr, ok := os.LookupEnv(CLIPluginSocketEnvKey)
-	if !ok {
-		// if a plugin compiled against a more recent version of docker/cli
-		// is executed by an older CLI binary, ignore missing environment
-		// variable and behave as usual
-		return
-	}
-	addr, err := net.ResolveUnixAddr("unix", socketAddr)
-	if err != nil {
-		return
-	}
-	cliCloseConn, err := net.DialUnix("unix", nil, addr)
-	if err != nil {
-		return
-	}
-
-	go func() {
-		b := make([]byte, 1)
-		for {
-			_, err := cliCloseConn.Read(b)
-			if errors.Is(err, io.EOF) {
-				cancel()
-			}
-		}
-	}()
-}
 
 // RunPlugin executes the specified plugin command
 func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager.Metadata) error {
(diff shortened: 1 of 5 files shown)


laurazard:
I didn't tag you yet/add tags because I still wanted to do a little refactor, but wanted to get WIP out in the meantime.

laurazard:
@neersighted

neersighted:
I've pushed up a proposed refactor/restructure to go with the non-abstract socket, PTAL @thaJeztah @laurazard

cli-plugins/socket/socket_nodarwin.go
```diff
@@ -0,0 +1,19 @@
...
+
+func listen(socketname string) (*net.UnixListener, error) {
+	return net.ListenUnix("unix", &net.UnixAddr{
+		Name: "@" + socketname,
+		Net:  "unix",
+	})
+}
+
+func onAccept(conn *net.UnixConn, listener *net.UnixListener) {
+	syscall.Unlink(listener.Addr().String())
```
whalelines, line 18: Wasn't the socket getting automatically cleaned up on Linux but not on macOS? Since you are still using the namespaces socket on Linux, i.e., prefixing it with "@", won't it still get cleaned up? Shouldn't this unlink be in the socket_darwin onAccept?

cli-plugins/plugin/plugin.go
```diff
@@ -16,9 +16,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// Deprecated: CLIPluginSocketEnvKey is an alias for socket.EnvKey.
-const CLIPluginSocketEnvKey = socket.EnvKey
-
```
thaJeztah: LOL, just wrote a comment for the previous commits;

> This one has not been in a stable release (https://github.com/docker/cli/commit/1554ac3b5f38147301c518c60da16f3f80c1717b is only in v25.0.0-beta and up)
>
> So I would just skip the alias and deprecation cycle

cli-plugins/socket/socket_nodarwin.go
neersighted: Shoot, good catch! Looks like I forgot that correction with the `git add -p`.

To test the behavior of this, it's sufficient to build a copy of the CLI and run a plugin command, and check that the sockets are no longer leaked.

To test the core functionality of the socket, something like `docker compose up` works (checking that after Ctrl-C, there is no detached docker-compose process still running).

neersighted:
Pushed a squashed version of the refactor per @thaJeztah's request.

cli-plugins/socket/socket_nodarwin.go
thaJeztah: I saw @neersighted mention some reasoning behind this for Linux, but I agree that it's not obvious, so I would suggest adding a comment outlining this. (ISTR he had some references with details)

thaJeztah: LOL.. browsers and race-conditions... ignore my comment

neersighted: I'm not quite sure what you mean; the logic is Darwin-specific and not Linux/Windows-specific. Do you mean linking to the manpage for unix(7) where it is suggested that unlink(2) can be used to forward-remove a socket present in the filesystem?

whalelines: According to this https://docker.slack.com/archives/C0W4XJVFW/p1705067942173219?thread_ts=1705055721.030329&cid=C0W4XJVFW , Go will try to remove sockets not starting with "@" upon close. So perhaps the delete is not necessary in either case since no "@" is being prepended on darwin?

I was asking about automated tests…

neersighted: We're doing a forward removal so that even if the CLI process crashes/is killed, we don't leave a socket on "disk" (read: in $TMPDIR). You're right that it's not strictly needed if we exit gracefully, but we also don't need the socket to stick around longer than the first client connects as this is more-or-less a "handshake" or "rendezvous."

Regarding tests, we don't have any right now; I don't think it would be too terrible to integration test this, but we should leave that for a follow-up as this is a pretty painful regression and a blocker for 25.0.

neersighted: Also with regards to that manpage/the reason for doing the forward removal, the relevant text is here: https://man7.org/linux/man-pages/man7/unix.7.html#:~:text=Binding%20to%20a,it%20is%20closed.

cli-plugins/socket/socket_darwin.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"net"
+	"os"
+	"path/filepath"
+	"syscall"
+)
+
+func listen(socketname string) (*net.UnixListener, error) {
+	return net.ListenUnix("unix", &net.UnixAddr{
+		Name: filepath.Join(os.TempDir(), socketname),
```
thaJeztah, line 12: This was changed from `config.Path("run")` correct?


LOL, also don't look at the windows implementation of `os.TempDir()` (which is a lot more complex than expected; https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/os/file_windows.go;l=310-333)

cli-plugins/socket/socket_nodarwin.go
```diff
@@ -0,0 +1,18 @@
...
+
+func listen(socketname string) (*net.UnixListener, error) {
+	return net.ListenUnix("unix", &net.UnixAddr{
+		Name: "@" + socketname,
+		Net:  "unix",
+	})
+}
+
+func onAccept(conn *net.UnixConn, listener *net.UnixListener) {
+	// do nothing
```
whalelines, line 17:
```suggestion
	// no need to unlink because namespaced sockets do not appear in the file system
```

cli-plugins/socket/socket_darwin.go
neersighted: Yeah, that is one difference from @laurazard's original version; I didn't see any reason to not just use the TMPDIR after verifying length should not be an issue.

cli-plugins/socket/socket.go
```diff
@@ -0,0 +1,72 @@
...
+	"io"
+	"net"
+	"os"
+
+	"github.com/docker/distribution/uuid"
+)
+
+// EnvKey represents the well-known environment variable used to pass the plugin being
+// executed the socket name it should listen on to coordinate with the host CLI.
+const EnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
```
thaJeztah, line 14: cc @milas @glours FYI this const was moved; in case code changes are needed when updating the module in compose

thaJeztah (approved):
LGTM

PR #4791, opened 2024-01-15
ci: Add bin-image workflow (2nd approach)

vvoland:
- follow up to: https://github.com/docker/cli/pull/4752
- follow up to: https://github.com/docker/cli/pull/4785

Same as https://github.com/docker/cli/pull/4752 except it has a separate build target instead of altering the original `binary` target.


thaJeztah (approved):
LGTM, thanks!


sorry for the revert; it was the quickest way to  unblock building 😅

vvoland:
No worries, I would have done the same in that case 😄

PR #4792, opened 2024-01-15
cli/plugins: use same pgid + skip signal forwarding when attached to a TTY

laurazard:
- Fixes (or attempts to): https://github.com/moby/moby/issues/47073

**- What I did**

Reverted https://github.com/docker/cli/commit/ef5e5fa03f0603f48678bfd7039c5b8121d98df1 and solved the "double notification" issue by only using the socket to notify the plugin process when not attached to a tty/not already going to receive a signal.

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli-plugins/socket/socket.go
+++ b/cli-plugins/socket/socket.go
@@ -14,21 +14,20 @@ import (
 const EnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
 
 // SetupConn sets up a Unix socket listener, establishes a goroutine to handle connections
-// and update the conn pointer, and returns the environment variable to pass to the plugin.
-func SetupConn(conn **net.UnixConn) (string, error) {
+// and update the conn pointer, and returns the listener for the socket (which the caller
+// is responsible for closing when it's no longer needed).
+func SetupConn(conn **net.UnixConn) (*net.UnixListener, error) {
 	listener, err := listen("docker_cli_" + uuid.Generate().String())
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
 	accept(listener, conn)
 
-	return EnvKey + "=" + listener.Addr().String(), nil
+	return listener, nil
 }
 
 func accept(listener *net.UnixListener, conn **net.UnixConn) {
-	defer listener.Close()
-
 	go func() {
 		for {
 			// ignore error here, if we failed to accept a connection,
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -241,6 +241,11 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string, 
 	go func() {
 		retries := 0
 		for range signals {
+			if dockerCli.Out().IsTerminal() {
+				// running attached to a terminal, so the plugin will already
+				// receive signals due to sharing a pgid with the parent CLI
+				continue
+			}
 			if conn != nil {
 				if err := conn.Close(); err != nil {
 					_, _ = fmt.Fprintf(dockerCli.Err(), "failed to signal plugin to close: %v\n", err)
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -240,16 +240,12 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string, 
 	// we send a SIGKILL to the plugin process and exit
 	go func() {
 		retries := 0
-		for s := range signals {
+		for range signals {
 			if conn != nil {
 				if err := conn.Close(); err != nil {
 					_, _ = fmt.Fprintf(dockerCli.Err(), "failed to signal plugin to close: %v\n", err)
 				}
 				conn = nil
-			} else {
-				// When the plugin is communicating via socket with the host CLI, we perform job control via the socket.
-				// However, if the plugin is an old version that is not socket-aware, we need to explicitly forward termination signals.
-				plugincmd.Process.Signal(s)
 			}
 			retries++
 			if retries >= exitLimit {
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -223,9 +223,10 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string, 
 
 	// Establish the plugin socket, adding it to the environment under a well-known key if successful.
 	var conn *net.UnixConn
-	socketenv, err := socket.SetupConn(&conn)
+	listener, err := socket.SetupConn(&conn)
 	if err == nil {
-		envs = append(envs, socketenv)
+		envs = append(envs, socket.EnvKey+"="+listener.Addr().String())
+		defer listener.Close()
 	}
 
 	plugincmd.Env = append(envs, plugincmd.Env...)
--- a/cli-plugins/manager/manager.go
+++ b/cli-plugins/manager/manager.go
@@ -233,7 +233,6 @@ func PluginRunCommand(dockerCli command.Cli, name string, rootcmd *cobra.Command
 		cmd.Stdin = os.Stdin
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
-		configureOSSpecificCommand(cmd)
 
 		cmd.Env = os.Environ()
 		cmd.Env = append(cmd.Env, ReexecEnvvar+"="+os.Args[0])
--- a/cli-plugins/manager/manager_unix.go
+++ b/cli-plugins/manager/manager_unix.go
@@ -2,21 +2,7 @@
 
 package manager
 
-import (
-	"os/exec"
-	"syscall"
-)
-
 var defaultSystemPluginDirs = []string{
 	"/usr/local/lib/docker/cli-plugins", "/usr/local/libexec/docker/cli-plugins",
 	"/usr/lib/docker/cli-plugins", "/usr/libexec/docker/cli-plugins",
-}
-
-func configureOSSpecificCommand(cmd *exec.Cmd) {
-	// Spawn the plugin process in a new process group, so that signals are not forwarded by the OS.
-	// The foreground process group is e.g. sent a SIGINT when Ctrl-C is input to the TTY, but we
-	// implement our own job control for the plugin.
-	cmd.SysProcAttr = &syscall.SysProcAttr{
-		Setpgid: true,
-	}
 }
--- a/cli-plugins/manager/manager_windows.go
+++ b/cli-plugins/manager/manager_windows.go
@@ -2,15 +2,10 @@ package manager
 
 import (
 	"os"
-	"os/exec"
 	"path/filepath"
 )
 
 var defaultSystemPluginDirs = []string{
 	filepath.Join(os.Getenv("ProgramData"), "Docker", "cli-plugins"),
 	filepath.Join(os.Getenv("ProgramFiles"), "Docker", "cli-plugins"),
-}
-
-func configureOSSpecificCommand(cmd *exec.Cmd) {
-	// no-op
 }


laurazard:
@thaJeztah @crazy-max

laurazard:
Ah I missed DCO on the revert commit 🤦‍♀️

crazy-max (approved):
LGTM

Tested with:

```
$ ./build/docker-linux-amd64 version
Client:
 Version:           24.0.0-rc.2-628-g5f6c55a724.m
 API version:       1.44
 Go version:        go1.21.6
 Git commit:        5f6c55a724
 Built:             Mon Jan 15 13:41:28 2024
 Context:           default

Server: Docker Desktop
 Engine:
  Version:          25.0.0-rc.1
  Go version:       go1.21.5
  Git commit:       9cebefa
  Built:            Thu Jan  4 16:30:11 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.26
  GitCommit:        3dd1e886e55dd695541fdcd67420c2888645a495
 runc:
  Version:          1.1.10
  GitCommit:        v1.1.10-0-g18a0cb0
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

```
BUILDX_EXPERIMENTAL=1 ./build/docker-linux-amd64 build .
#0 building with "default" instance using docker driver

#1 [internal] connecting to local controller
#1 DONE 0.0s

#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 4.90kB 0.0s done
#2 DONE 0.0s

#3 resolve image config for docker.io/docker/dockerfile:1
#3 DONE 0.5s

#4 docker-image://docker.io/docker/dockerfile:1@sha256:ac85f380a63b13dfcefa89046420e1781752bab202122f8f50032edf31be0021
#4 CACHED

...
```

laurazard:
I was testing with Compose and noticed that we'd lost the desired "notify plugin to exit via socket if detached" behavior and mostly sure I nailed it down to https://github.com/docker/cli/blob/dbf992f91f605b5fffa6b08172e6ed1cc7e84d80/cli-plugins/socket/socket.go#L36-L46

specifically, the `defer listener.Close()` line since it gets called before all those gouroutine executions.

With the `defer listener.Close()` line and logging the error:
```go
func accept(listener *net.UnixListener, conn **net.UnixConn) {
	defer listener.Close()

	go func() {
		for {
			// ignore error here, if we failed to accept a connection,
			// conn is nil and we fallback to previous behavior
			newConn, err := listener.AcceptUnix()
			*conn = newConn
			if err != nil {
				println(err.Error())
			}
			// perform any platform-specific actions on accept (e.g. unlink non-abstract sockets)
			onAccept(*conn, listener)
		}
	}()
}
```
```
accept unix @docker_cli_1c8d52df-4d23-4c5a-9a1a-6a91b798da88: use of closed network connection
accept unix @docker_cli_1c8d52df-4d23-4c5a-9a1a-6a91b798da88: use of closed network connection
accept unix @docker_cli_1c8d52df-4d23-4c5a-9a1a-6a91b798da88: use of closed network connection
accept unix @docker_cli_1c8d52df-4d23-4c5a-9a1a-6a91b798da88: use of closed network connection
accept unix @docker_cli_1c8d52df-4d23-4c5a-9a1a-6a91b798da88: use of closed network connection
accept unix @docker_cli_1c8d52df-4d23-4c5a-9a1a-6a91b798da88: use of closed network connection
accept unix @docker_cli_1c8d52df-4d23-4c5a-9a1a-6a91b798da88: use of closed network connection
accept unix @docker_cli_1c8d52df-4d23-4c5a-9a1a-6a91b798da88: use of closed network connection
accept unix @docker_cli_1c8d52df-4d23-4c5a-9a1a-6a91b798da88: use of closed network connection
...
```

Looks like during the refactor we moved that `defer` and now it gets called too early. I'll move it back.

cli-plugins/socket/socket.go
```diff
@@ -14,21 +14,23 @@ import (
...
 	if err != nil {
-		return "", err
+		return "", nil, err
 	}
 
 	accept(listener, conn)
 
-	return EnvKey + "=" + listener.Addr().String(), nil
+	closeListener := func() {
+		listener.Close()
```
thaJeztah, line 28: Perhaps take the opportunity to make linters and IDEs "happy";

```suggestion
		_ = listener.Close()
```

cli-plugins/socket/socket.go
```diff
@@ -14,21 +14,23 @@ import (
...
+	}
+	return EnvKey + "=" + listener.Addr().String(), closeListener, nil
```
thaJeztah, line 30: Starting to wonder now if it was a good choice to change the signature to return a string (`env=address`) or if we should revert to making it return the actual listener (and have the caller decide what to do with it, and to handle the `Close()`.


No strong opinion either way; mostly thinking out loud here; what do you think feels more "natural" as a signature?

laurazard: Yeah, I debated with the same question while fixing the early defer – since we don't have to abstract the "env value determination process" away in platform-specific implementations, it might be a lot cleaner to just return the listener.

laurazard:
Also tested with Compose v2.24.0 – both "attached" to a TTY and not-attached seem to be working correctly ✅

laurazard:
amended with comments, can you TAL @thaJeztah?

cli-plugins/socket/socket.go
laurazard: Oh! My linters don't complain about this, I wonder if it's Goland (maybe we can add something to `.golangci.yml` so that CI/people not using Goland would pick it up too). Btw, does it also complain about this in a `defer`?

cli-plugins/socket/socket.go
```diff
@@ -14,21 +14,20 @@ import (
 const EnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
 
 // SetupConn sets up a Unix socket listener, establishes a goroutine to handle connections
-// and update the conn pointer, and returns the environment variable to pass to the plugin.
-func SetupConn(conn **net.UnixConn) (string, error) {
+// and update the conn pointer, and returns the environment variable to pass to the plugin
```
neersighted, line 17: This line is no longer correct.

neersighted (changes requested):
Good catch on the `defer` 😅 

I'm not entirely convinced this is the right approach: now, each plugin will have to implement the 'graceful termination' behavior; which is a thing we can do/build into the `cli-plugins` package, but still, I'm not sure having two copies of that logic that run conditionally really makes sense.

It feels doubly dubious when the logic built into the parent CLI will almost never trigger in a way that is useful (given that it won't trigger when a TTY is present).

I do feel like the pgid change is a beneficial one, unless there is a definite reason that plugins must be able to read from the controlling TTY (they will still be able to read from a non-TTY stdin like a file).

laurazard:
That's a valid thought (and one I had while discussing this). The impetus to make this change, as is, now, was:
1. The current state is not great (`BUILDX_EXPERIMENTAL` is completely broken, and users will/have noticed it)
2. Simply reverting the `pgid` commit and shipping is a worse solution as it will cause `docker compose up` to break, as it will receive double `SIGINT`s on every CTRL+C so when the user tries to teardown we will always force-kill and not let the containers exit, which imo is even worse than having Buildx broken when using the experimental env var
3. We want to get a change out quickly so we can make another RC

 The other way we could explore this would be to build functionality into the CLI plugins API so that plugins can use that instead of reading from TTY – however, that's not trivial, should be well-explored, and not something we rush out to get a fix in. Since that's not an option for _right now_, and reverting the pgid changes without another patch (such as this one) leads to an even worse situation for right now, I think the current approach (or something similar) is the only solution for right now – even if we then take some time and figure out that functionality later so we can ultimatively move over for that.

Moreover, re: implementing an API for this in the CLI, we do need to think that all of these new features will be gated behind an "if running through the CLI" (for Buildx at least) and the plugin will still have to read directly from the TTY when running standalone, so those new features will have to be thoroughly thought out/not something we can do out of the gate.

laurazard:
@neersighted we should decide on this to unblock v25 – we either leave it as-is and own that we'll break buildx, or merge this change now and work on getting the pgid changes in later (together with an API for plugins to access the API).

neersighted:
To me I think it depends on how critical reading from the TTY is to the buildx folks; I'm okay with a fast revert if we think it's necessary to buildx (for now). However, if this is not a critical behavior and buildx could be changed, I'd suggest that it's better to adopt the pgid change now to prevent future issues of this nature (and I'd suggest that the regression with older buildx is fine given it's under `_EXPERIMENTAL`).

thaJeztah:
I think we all agree that this is not _ideal_, but the options we currently have limit our choices;

1. keep the pgid patch; `BUILDX_EXPERIMENTAL=1` is completely broken. While it's "experimental" for sure, it's also something that looks to have been around since buildx v0.11 (added in https://github.com/docker/buildx/pull/1296), and breaking that feature means that it'll be broken in `docker` (but still works with other tools, such as `nerdctl`). Of course ideally, we'd fix this in `buildx`, but I don't know what the timeline will be to have such a fix in a released version of buildx.
2. revert the pgid patch; docker compose will be very much broken (terminating immediately on `CTRL-C`), which is pretty bad as well. We _could_ possibly apply some fixes to compose, but  that would have to be looked into.
3. use the conditional code from this PR; it's not _ideal_, and we should continue looking at fixing buildx so that we can re-apply the pgid improvement, but AFIU, at least it would make _both_ buildx and compose "work" (and signal handling improved compared to "pre v25").


Our of options above, I'm leaning towards 3


(LOL, and after writing... my list of options is ... well.. basically fully the same as the list @laurazard posted in https://github.com/docker/cli/pull/4792#issuecomment-1892411388 above)


@docker/runtime-team any thoughts?

krissetto:
LGTM. Considering the options available, I'd also lean towards n.3, using the conditional code from this PR.

Shall we also open an issue to track this "suboptimal" solution so we don't forget about looking into it for too long? Or do we track these things somewhere else?

neersighted:
I'm really not sure how nerdctl factors in, as it has a completely separate implementation that wraps `buildctl`: https://github.com/containerd/nerdctl/blob/main/pkg/buildkitutil/buildkitutil.go

In any case, it seems like #3 is the consensus; while with the compromise here my first thought is we can remove the 3-signal-to-exit feature from the host CLI (as it's no longer doing much good), if we plan to resolve the signaling/IO issues in the near future such that the functionality can primarily live in the host CLI, it likely makes sense to take this as is.

crazy-max:
@neersighted `BUILDX_EXPERIMENTAL` runs controller build instead of the basic one: https://github.com/docker/buildx/blob/78adfc80a90ede124ae30457dc0f3c2f688f5121/commands/build.go#L293

There is some pipe handling when detach mode is enabled to handle debugging but in the case of this issue it will use the local one: "connecting to local controller".

This might be related to https://github.com/docker/buildx/blob/78adfc80a90ede124ae30457dc0f3c2f688f5121/controller/processes/processes.go#L101

I will take a look. @ktock @jedevc if you have an idea in any case :pray:

neersighted (approved):
Looks like an okay band-aid to me

thaJeztah:
Let me post a quick update on discussions we had;

- Yes, this PR isn't ideal
- It looks like work is done in buildx that may help with this issue (https://github.com/docker/buildx/pull/2196)
- But that work has not been reviewed / tested / shipped yet, and even when released, users may not be using "latest" buildx immediately
- We don't consider the current patch to be the "final" solution, but it works "for now", and we can ship an improved version in a (patch) release, as it's not immediately user-facing (just improved implementation)

I'll merge this PR to unblock the v25.0.0 release, but we'll look at follow-ups :+1:

thaJeztah (approved):
LGTM

tonistiigi:
> revert the pgid patch; docker compose will be very much broken (terminating immediately on CTRL-C), which is pretty bad as well. We could possibly apply some fixes to compose, but that would have to be looked into.

@thaJeztah What is this issue and how is that a regression? I assume if compose "terminates immediately", then it can only be if it is lacking support for handling signals.

cmd/docker/docker.go
```diff
@@ -240,16 +241,17 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
 	// we send a SIGKILL to the plugin process and exit
 	go func() {
 		retries := 0
-		for s := range signals {
+		for range signals {
+			if dockerCli.Out().IsTerminal() {
```
tonistiigi, line 245: I'm not sure if `dockerCli.Out()` is ideal here, eg. for a command like `build -o type=tar . > t.tar` . Although afaics, the actual behavior change from this line is minimal.

laurazard:
> What is this issue and how is that a regression? I assume if compose "terminates immediately", then it can only be if it is lacking support for handling signals.

No, that wasn't the issue. The issue is well-described in the linked commit in the PR description – https://github.com/docker/cli/commit/ef5e5fa03f0603f48678bfd7039c5b8121d98df1

For convenience:
> Changes were made in https://github.com/docker/cli/commit/1554ac3b5f38147301c518c60da16f3f80c1717b to provide
a mechanism for the CLI to notify running plugin processes that they
should exit, in order to improve the general CLI/plugin UX. The current
implementation boils down to:
> 1. The CLI creates a socket
> 2. The CLI executes the plugin
> 3. The plugin connects to the socket
> 4. (When) the CLI receives a termination signal, it uses the socket to
   notify the plugin that it should exit
> 5. The plugin's gets notified via the socket, and cancels it's `cmd.Context`,
   which then gets handled appropriately
> 
> This change works in most cases and fixes the issue it sets out to solve
(see: https://github.com/docker/compose/pull/11292) however, in the case
where the user has a TTY attached and the plugin is not already handling
received signals, steps 4+ changes:
> 4. (When) the CLI receives a termination signal, before it can use the
   socket to notify the plugin that it should exit, the plugin process
   also receives a signal due to sharing the pgid with the CLI

Since Compose up works like: 1st CTRL-C - start "graceful" teardown / 2nd CTRL-C - forceful teardown/exit, the double notification (via the plugin socket, and the process itself getting signalled) resulted in every time the user hits CTRL-C, Compose receives two "cancel signals" and skips the graceful teardown directly to the forceful exit.

That same "double notification" issue was what the process group id changes tried to address (fix the double notification by running the plugin process under a new process group id, so it doesn't also get signalled) and this PR solves it with a different approach (by not notifying via the socket when the plugin is attached to a TTY and hence already going to receive a signal).

tonistiigi:
By "double notification" issue, do you mean this line https://github.com/docker/cli/pull/4792/files#diff-3dcd6325b70a5f5db393a3ecc08060b69007fcef371fb4dabf3164da95646fb6L252 that has now been reverted by this PR and does not exist in v24. There is no "double notification" issue in v24.

I think what makes sense is to leave everything as was in v24 (plugins are in same processgroup, cli does not duplicate its signals). Additionally, this unix socket can be used to send custom termination request. Handling that request is optional and it doesn't cause any conflict for the plugin as when a plugin decides to support the socket, it can make sure that if it does its own signal counting logic, the new socket mechanism does not cause it to go out of sync. I believe this is mostly the current state after this PR, ignoring some `isTerminal` checks. From the previous comment, it seemed that there was some issue between this PR and compose, but if not then all good from my side.

laurazard:
Ah, no. The previous comments were discussing whether to "simply revert the pgid patch" (which would have reintroduced that issue with Compose) or to go with this change, which reverts that but also adds the `isTerminal` logic.

PR #4793, opened 2024-01-15
gha: update runners to Ubuntu 22.04

thaJeztah:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   prepare:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     outputs:
       matrix: ${{ steps.platforms.outputs.matrix }}
     steps:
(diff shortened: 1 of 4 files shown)


thaJeztah:
Ah, fun; I guess we need to update the engine that's used, which is still on some very old version;

```
Engine:
  Version:          19.03.14
```

That version doesn't like cgroups v2;

```
=== FAIL: e2e/container TestAttachExitCode (0.87s)
    attach_test.go:15: assertion failed: 
        Command:  docker run -d -i --rm registry:5000/alpine:frozen sh -c read; exit 21
        ExitCode: 125
        Error:    exit status 125
        Stdout:   aa2bb27139717fd522a124f12c0b56f9d865ed659692b70bc776ab24ff3d3b5c
        
        Stderr:   docker: Error response from daemon: cgroups: cgroup mountpoint does not exist: unknown.
        
        
        Failures:
        ExitCode was 125 expected 0
        Expected no error
```

But updating the engine was causing some other tests to fail :sweat_smile: 

- https://github.com/docker/cli/pull/3128

PR #4796, opened 2024-01-16
deprecate old API versions (1.23 and lower)

thaJeztah:
The Docker Daemon provides a versioned API for backward compatibility with old clients. When a client connects, it can perform API-version negotiation to select the most recent API version supported by the daemon (downgrading to and older version of the API when necessary). API version negotiation was introduced in Docker v1.12.0 (API 1.24), and clients before that used a fixed API version.

Docker Engine up until v25.0 provide support for all [API versions](https://docs.docker.com/engine/api/#api-version-matrix) included in stable releases for a given platform. For Linux daemons, the oldest supported API version is 1.12 (corresponding with Docker Engine v1.0.0), whereas for Windows Daemons, the oldest supported API version is 1.24 (corresponding with Docker Engine v1.12.0).

Support for legacy API versions (providing old API versions on current versions of the Docker Engine) is primarily intended to provide compatibility with recent, but still supported versions of the client, which is a common scenario (the Docker Daemon may be updated to the latest release, but not all clients may be up-to-date or vice versa). Support for API versions before that (API versions provided by EOL versions of the Docker Daemon) is provided on a "best effort" base.

Use of old API versions is very rare, and support for legacy API versions involves significant complexity (Docker 1.0.0 was released 10 Years ago). Because of this, we'll start deprecating support for legacy API versions.

Docker Engine v25.0 by default disables API version older than 1.24 (aligning the minimum supported API version between Linux and Windows daemons). When connecting with a client that uses an API version version older than 1.24, the daemon returns an error. The following example configures the docker CLI to use API version 1.23, which produces an error:

    DOCKER_API_VERSION=1.23 docker version
    Error response from daemon: client version 1.23 is too old. Minimum supported API version is 1.24, please upgrade your client to a newer version

An environment variable (`DOCKER_MIN_API_VERSION`) is introduced that allows re-enabling older API versions in the daemon. This environment-variable must be set in the daemon's environment (for example, through a [systemd override file](https://docs.docker.com/config/daemon/systemd/)), and the specified API version must be supported by the daemon (`1.12` or higher on Linux, or `1.24` or higher on Windows).

Support for API versions lower than `1.24` will be permanently removed in Docker Engine v26, and the minimum supported API version will be incrementally raised in releases following that.

We do not recommend depending on the `DOCKER_MIN_API_VERSION` environment variable other than for exceptional cases where it's not possible to update old clients, and those clients must be supported.

**- What I did**

**- How I did it**

**- How to verify it**


docs/deprecated.md
```diff
@@ -109,6 +110,60 @@ The table below provides an overview of the current status of deprecated feature
...
+connecting with a client that uses an API version version older than 1.24,
+the daemon returns an error. The following example configures the docker
+CLI to use API version 1.23, which produces an error:
+
+```console
+DOCKER_API_VERSION=1.23 docker version
+Error response from daemon: client version 1.23 is too old. Minimum supported API version is 1.24, please upgrade your client to a newer version
+```
+
+An environment variable (`DOCKER_MIN_API_VERSION`) is introduced that allows
```
thaJeztah, line 153: FWIW, I decided to not (yet?) document this env-var in the dockerd reference; https://docs.docker.com/engine/reference/commandline/dockerd/#environment-variables

My motivation;

- This is not a recommended env-var
- We _could_ add a separate table "for debugging purposes" (or similar) to better separate "ok to use" and "not recommended, other than exceptional cases" env-vars
- ☝️ while looking at that, I also discovered various other undocumented or (almost not documented) env-vars, such as `DEBUG`, `DOCKER_WINDOWSFILTER_NOREEXEC`, `DOCKER_CLI_PLUGIN_USE_DIAL_STDIO`, `ROOTLESSKIT_STATE_DIR`), some of which are for the "daeemon", some for the "cli", and some are _intentionally_ un-documented (intended for use in CI, e.g.), so we need to collect all of those, and decide whether to document them, and "how".

dvdksn: ugh. We really need to do an inventory of all env vars and ensure they are properly and consistently namespaced.

thaJeztah: I think the `ROOTLESSKIT` one may be because the library we're using defined it, but yes; leaning toward documenting _all of them_ even if they should NOT be used (but doing so gives us a clear way to tell users "don't use this!")

docs/deprecated.md
```diff
@@ -109,6 +110,60 @@ The table below provides an overview of the current status of deprecated feature
...
+**Target For Removal In Release: v26.0**
+
+The Docker Daemon provides a versioned API for backward compatibility with old
+clients. When a client connects, it can perform API-version negotiation to
+select the most recent API version supported by the daemon (downgrading to
+and older version of the API when necessary). API version negotiation was
+introduced in Docker v1.12.0 (API 1.24), and clients before that used a fixed
+API version.
+
+Docker Engine up until v25.0 provide support for all [API versions](https://docs.docker.com/engine/api/#api-version-matrix)
```
neersighted, line 125:
```suggestion
Docker Engine versions through v25.0 provide support for all [API versions](https://docs.docker.com/engine/api/#api-version-matrix)
```

docs/deprecated.md
```diff
@@ -109,6 +110,60 @@ The table below provides an overview of the current status of deprecated feature
...
+supported API version is 1.12 (corresponding with Docker Engine v1.0.0), whereas
+for Windows Daemons, the oldest supported API version is 1.24 (corresponding
+with Docker Engine v1.12.0).
+
+Support for legacy API versions (providing old API versions on current versions
+of the Docker Engine) is primarily intended to provide compatibility with recent,
+but still supported versions of the client, which is a common scenario (the Docker
+Daemon may be updated to the latest release, but not all clients may be up-to-date
+or vice versa). Support for API versions before that (API versions provided by EOL
+versions of the Docker Daemon) is provided on a "best effort" base.
```
neersighted, line 136:
```suggestion
versions of the Docker Daemon) is provided on a "best effort" basis.
```

docs/deprecated.md
```diff
@@ -109,6 +110,60 @@ The table below provides an overview of the current status of deprecated feature
...
+
+Use of old API versions is very rare, and support for legacy API versions
+involves significant complexity (Docker 1.0.0 was released 10 Years ago).
```
neersighted, line 139:
```suggestion
involves significant complexity (Docker 1.0.0 heaving been released 10 years ago).
```

docs/deprecated.md
```diff
@@ -109,6 +110,60 @@ The table below provides an overview of the current status of deprecated feature
...
+the daemon returns an error. The following example configures the docker
+CLI to use API version 1.23, which produces an error:
+
+```console
+DOCKER_API_VERSION=1.23 docker version
+Error response from daemon: client version 1.23 is too old. Minimum supported API version is 1.24, please upgrade your client to a newer version
+```
+
+An environment variable (`DOCKER_MIN_API_VERSION`) is introduced that allows
+re-enabling older API versions in the daemon. This environment-variable must
```
neersighted, line 154:
```suggestion
re-enabling older API versions in the daemon. This environment variable must
```

docs/deprecated.md
neersighted: Technically `ROOTLESSKIT` is because it's a "third party" (read: re-used in multiple places) codebase being configured, and the uniformity makes sense (especially when you consider it is an external binary that multiple packages may re-use).

thaJeztah:
@dvdksn @neersighted I _THINK_ I addressed your comments, but take a good look to see if I did 🤗

thaJeztah:
Thanks! I'll bring this one in, but happy to touch-up things in a follow up where we see a need 👍

PR #4805, opened 2024-01-20
socket: return from loop after EOF

tonistiigi:
After a plugin receives the termination request via socket and closes the context, it atm goes into a tight loop that never returns.


thaJeztah (approved):
Oof! Nice find, thanks!

LGTM

(cc @laurazard @neersighted)

PR #4808, opened 2024-01-20
info: remove printServerWarningsLegacy

thaJeztah:
Relates to:

- https://github.com/moby/moby/pull/37502
- https://github.com/docker/cli/pull/1225

Docker Engine 18.09 (API v1.39) introduced a Warnings field in the into response. This enhancement was not gated by API version (see [moby/moby@a3d4238]), and will be returned by Docker Engine 18.09 and up, regardless of the API version chosen.

Likewise, the client-side code was written to prefer warnings returned by the daemon, but to fall back on client-side detection of missing features based on information in the Info response (see [docker/cli@3c27ce2]).

Thse warnings are purely informational, and given that Docker Engine versions before 18.09 have reached EOL 6 Years ago, and any current version of the Engine now returns the Warnings, it should be safe to remove the client-side fall back logic.

This patch removes the client-side fall back code for warnings that was added in 3c27ce21c905f55c4fa06d28ca42caf7e960db18.

[moby/moby@a3d4238]: https://github.com/moby/moby/commit/a3d4238b9ce653d2863fbc93057ed4162a83221e
[docker/cli@3c27ce2]: https://github.com/docker/cli/commit/3c27ce21c905f55c4fa06d28ca42caf7e960db18


krissetto (approved):
LGTM.  

How do we handle comms when removing cli features necessary for EOL versions of the engine? Anything special that needs to be done?

thaJeztah:
> How do we handle comms when removing cli features necessary for EOL versions of the engine? Anything special that needs to be done?

So far, the CLI "supported" for all versions of docker since 1.0.0; most of that is based on API version; the client can downgrade to older versions of the API, where needed changing API calls, or disabling features not yet supported. And the CLI itself has some detection as well; there's some flags and commands that are hidden when the CLI is connected to an older engine.

Most of that was introduced to help transitioning to newer daemons, so for situations where a user updated the CLI to the latest version, but the server they connected to has not yet been updated (a more likely scenario in production setups, where you may not update your daemon to "new version" on _day one_). Before API version negotiation, users often ran into "this daemon is too old" errors.

API versions before "previous" should work (as mentioned, all the way back to v1.0.0), but that should really be considered "best effort" (it should mostly work , but it's definitely not _tested_).

If we actively _deprecate_ features, we add those to the `deprecated.md` page in this repository; on that page, we announce a feature to be deprecated, and an (estimated) version in which the feature will be removed. It depends a bit on the feature how long there's between deprecating and removing; some options are "ok" to remove fast, others may need a proper deprecation cycle, which may (in the extreme case) be multiple releases;

1. announce deprecation
2. print warnings (with link to relevant information in the docs)
3. disable by default (but an escape hatch)
4. remove but produce custom, informative error
5. full removal

In this case, the removed functionality was not essential, and would only be produced for (_very_) old, and long EOL versions of the daemon. Anyone still running those has a MUCH bigger problem.

krissetto:
Thanks for the thorough explanation @thaJeztah!

thaJeztah:
You're welcome! We should consider having some of this actually written up (I don't think expectations have ever been properly documented) /cc @dvdksn (low priority, just one to put in the back of your mind)

PR #4810, opened 2024-01-20
cli/command: remove deprecated DockerCliOption, InitializeOpt

thaJeztah:
- relates to https://github.com/docker/cli/pull/4387

These types were deprecated in 7af509c7f12e2a541aedb9f26ec3e498c64eee8f (v25.0), in favor of CLIOption, and are no longer used.

This patch removes the deprecated type-aliases, and while updating, also improves the documentation for the CLIOption type.

With this patch:

Screenshot 2024-01-20 at 21 37 36

**- What I did**

**- How I did it**

**- How to verify it**

```
cli/command: remove deprecated 1DockerCliOption`, `InitializeOpt`
```

### DIFF
--- a/cli/command/cli_options.go
+++ b/cli/command/cli_options.go
@@ -11,18 +11,10 @@ import (
 	"github.com/moby/term"
 )
 
-// CLIOption applies a modification on a DockerCli.
+// CLIOption is a functional argument to apply options to a [DockerCli]. These
+// options can be passed to [NewDockerCli] to initialize a new CLI, or
+// applied with [DockerCli.Initialize] or [DockerCli.Apply].
 type CLIOption func(cli *DockerCli) error
-
-// DockerCliOption applies a modification on a DockerCli.
-//
-// Deprecated: use [CLIOption] instead.
-type DockerCliOption = CLIOption
-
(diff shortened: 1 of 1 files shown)


krissetto (approved):
LGTM.

cli/command/cli_options.go
```diff
@@ -11,19 +11,11 @@ import (
 	"github.com/moby/term"
 )
 
-// CLIOption applies a modification on a DockerCli.
+// CLIOption is a functional argument to apply options to a [DockerCli]. These
+// options can be passed to [NewDockerCli] to initialize a new CLI, or
+// applied with [DockerCli.Initialize] or [DockerCli.Apply].
```
krissetto, line 16: Just for clarity, could It make sense to mention here that it replaces the deprecated DockerCliOption/InitializeOpt types?

thaJeztah: I don't think we've done that before. We usually;

- Deprecate in 1 release, and add aliases to the new location (to give users a pointer where to look for the new thing)
- Remove in the release after.

Admitted; if we're gonna ramp up release cadence, then 1 release "deprecated' may potentially be too short. OTOH; if we move to Go modules, every major release is ... a major release, so technically "anything goes!"

PR #4828, opened 2024-01-25
Add tests for CLI/plugin communication (i.e. let's not break everyone's plugins 🥲)

laurazard:
**- What I did**

Added unit tests for the CLI-plugin socket code, and added e2e tests covering the new functionality as well as backwards compatibility with older plugin binaries, as well as regressions for https://github.com/moby/moby/issues/47073.

OH! There's also a commit making it so that we don't hardcode `amd64` in the plugin e2e scripts, so that this works on arm too :)

**- How I did it**

With a small dog on my lap! And by adding a new e2e test plugin.

**- How to verify it**

(not the aforementioned dog)

Screenshot 2024-01-29 at 02 11 51

### DIFF
--- a/cli-plugins/socket/socket_test.go
+++ b/cli-plugins/socket/socket_test.go
@@ -1,0 +1,129 @@
+package socket
+
+import (
+	"io/fs"
+	"net"
+	"os"
+	"runtime"
+	"testing"
+	"time"
+
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/poll"
+)
+
+func TestSetupConn(t *testing.T) {
+	t.Run("updates conn when connected", func(t *testing.T) {
+		var conn *net.UnixConn
+		listener, err := SetupConn(&conn)
+		assert.NilError(t, err)
+		assert.Check(t, listener != nil, "returned nil listener but no error")
+		addr, err := net.ResolveUnixAddr("unix", listener.Addr().String())
+		assert.NilError(t, err, "failed to resolve listener address")
+
+		_, err = net.DialUnix("unix", nil, addr)
+		assert.NilError(t, err, "failed to dial returned listener")
+
+		pollConnNotNil(t, &conn)
+	})
+
+	t.Run("allows reconnects", func(t *testing.T) {
+		var conn *net.UnixConn
+		listener, err := SetupConn(&conn)
+		assert.NilError(t, err)
+		assert.Check(t, listener != nil, "returned nil listener but no error")
+		addr, err := net.ResolveUnixAddr("unix", listener.Addr().String())
+		assert.NilError(t, err, "failed to resolve listener address")
+
+		otherConn, err := net.DialUnix("unix", nil, addr)
+		assert.NilError(t, err, "failed to dial returned listener")
+
+		otherConn.Close()
+
+		_, err = net.DialUnix("unix", nil, addr)
+		assert.NilError(t, err, "failed to redial listener")
+	})
+
+	t.Run("does not leak sockets to local directory", func(t *testing.T) {
+		var conn *net.UnixConn
+		listener, err := SetupConn(&conn)
+		assert.NilError(t, err)
+		assert.Check(t, listener != nil, "returned nil listener but no error")
+		checkDirClean(t)
+
+		addr, err := net.ResolveUnixAddr("unix", listener.Addr().String())
+		assert.NilError(t, err, "failed to resolve listener address")
+		_, err = net.DialUnix("unix", nil, addr)
+		assert.NilError(t, err, "failed to dial returned listener")
+		checkDirClean(t)
+	})
+}
+
+func checkDirClean(t *testing.T) {
+	t.Helper()
+
+	files, err := os.ReadDir(".")
+	assert.NilError(t, err, "failed to list files in dir to check for leaked sockets")
+
+	for _, f := range files {
+		info, err := f.Info()
+		assert.NilError(t, err, "failed to check file info")
+		if info.Mode().Type() == fs.ModeSocket {
+			t.Fatal("found socket in a local directory")
+		}
+	}
+}
+
+func TestConnectAndWait(t *testing.T) {
+	t.Run("calls cancel func on EOF", func(t *testing.T) {
+		var conn *net.UnixConn
+		listener, err := SetupConn(&conn)
+		assert.NilError(t, err, "failed to setup listener")
+
+		done := make(chan struct{})
+		t.Setenv(EnvKey, listener.Addr().String())
+		cancelFunc := func() {
+			done <- struct{}{}
+		}
+		ConnectAndWait(cancelFunc)
+		pollConnNotNil(t, &conn)
+		conn.Close()
+
+		select {
+		case <-done:
+		case <-time.After(10 * time.Millisecond):
+			t.Fatal("cancel function not closed after 10ms")
+		}
+	})
+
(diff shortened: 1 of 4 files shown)


thaJeztah:
Ah, nice; it's green now! 🎉

laurazard:
Yes! I have more tests incoming (and need to fix the one that's commented out), but we're getting there! (And it'll be so nice to have all these things tested)

cli-plugins/socket/socket_test.go
```diff
@@ -0,0 +1,129 @@
...
+	})
+}
+
+func checkDirClean(t *testing.T) {
+	t.Helper()
+
+	files, err := os.ReadDir(".")
+	assert.NilError(t, err, "failed to list files in dir to check for leaked sockets")
+
+	for _, f := range files {
```
thaJeztah, line 68: Could this limit our options to run this with `t.Parallel()`? (i.e. other tests in parallel creating sockets)?

Curious if we need to check for a specific socket (created for the test calling this helper)

cli-plugins/socket/socket_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	t.Run("connect goroutine exits after EOF", func(t *testing.T) {
+		var conn *net.UnixConn
+		listener, err := SetupConn(&conn)
+		assert.NilError(t, err, "failed to setup listener")
+		t.Setenv(EnvKey, listener.Addr().String())
+		numGoroutines := runtime.NumGoroutine()
+
+		ConnectAndWait(func() {})
+		assert.Equal(t, runtime.NumGoroutine(), numGoroutines+1)
```
thaJeztah, line 107: Looks like this function can never be run with `t.Parallel()`. I wish Go's test package had a utility for this, but we could consider implementing something similar to what @vvoland did in https://github.com/moby/moby/pull/47088 on the CLI as well (and error when trying to run with `t.Parallel()` enabled).

scripts/test/e2e/run
```diff
@@ -69,7 +71,7 @@ runtests() {
         GOPATH="$GOPATH" \
         PATH="$PWD/build/:/usr/bin:/usr/local/bin:/usr/local/go/bin" \
         HOME="$HOME" \
-        DOCKER_CLI_E2E_PLUGINS_EXTRA_DIRS="$PWD/build/plugins-linux-amd64" \
+        DOCKER_CLI_E2E_PLUGINS_EXTRA_DIRS="$PWD/build/plugins-linux-${GOARCH}" \
```
thaJeztah, line 74: Had to double-check but looks like we run this script with `set -eu -o pipefail`, so it should fail if `GOARCH` is not set for whatever reason 👍

e2e/cli-plugins/socket_test.go
```diff
@@ -0,0 +1,232 @@
...
+// that does not connect to the CLI plugin socket, simulating
+// a plugin compiled against an older version of the CLI, and
+// ensures that backwards compatibility is maintained.
+func TestPluginSocketBackwardsCompatible(t *testing.T) {
+	run, _, cleanup := prepare(t)
+	defer cleanup()
+
+	t.Run("attached", func(t *testing.T) {
+		t.Run("the plugin gets signalled if attached to a TTY", func(t *testing.T) {
+			t.SkipNow()
```
thaJeztah, line 26: Should this skip be removed?

e2e/cli-plugins/socket_test.go
```diff
@@ -0,0 +1,232 @@
...
+					t.Fatal("failed to get command output")
+				}
+			case <-time.After(5 * time.Second):
+				t.Fatal("timed out! plugin process probably stuck")
+			}
+		})
+	})
+
+	t.Run("detached", func(t *testing.T) {
+		t.SkipNow()
```
thaJeztah, line 82: Same here?

thaJeztah (changes requested):
overall looks great! I found 2 tests that are currently skipped though, and I _think_ that was unintentional?

I also left 2 comments w.r.t. tests not being suitable for `t.Parallel()`; those are not a blocker, but we could look at that in a follow-up. In the meantime, we could (for those we don't address) consider adding a `TODO` to make it visible that we _know_ these cannot be run with `t.Parallel()`.

e2e/cli-plugins/socket_test.go
laurazard: Oops! Was testing a specific test and left it there.

scripts/test/e2e/run
laurazard: Yup! This was a pain to deal with, and that seemed the easiest way to get `GOARCH` in there so I could run it locally as well.

cli-plugins/socket/socket_test.go
laurazard: Ah, that's a good point. I wonder if this is simple enough that we should just add a comment calling it out and to not run it with `t.Parallel()`, but I'm okay with adding something like that too to give us a little bit more security.

OTOH, if we have a commend and the test starts failing it should be pretty easy to understand why?

cli-plugins/socket/socket_test.go
laurazard: Ah, that's a good point – better regardless to check for the specific socket and make sure we don't bump into any other tests running.

cli-plugins/socket/socket_test.go
thaJeztah: Yes, I don't think there's a real need to widen the scope of this PR, and I'm fine with a TODO comment for now. Having a comment at least helps discovery if we run into the test being flaky, we'd have a "I told you so!" from our "past selves" 😂

laurazard:
Ty @thaJeztah! Fixed the missed skipped tests, changed the one socket test to look for the plugin socket specifically, and added a TODO to the other test. Could you PTA(nother)L?

cli-plugins/socket/socket_test.go
laurazard: Definitely 😅

e2e/cli-plugins/socket_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+			go func() {
+				<-time.After(time.Second)
+				// we're signalling the parent process directly and not
+				// the process group, since we're testing the case where
+				// the process is detached and not simulating a CTRL-C
+				// from a TTY
+				err := syscall.Kill(command.Process.Pid, syscall.SIGINT)
+				assert.NilError(t, err, "failed to signal process group")
+				time.Sleep(50 * time.Millisecond)
```
laurazard, line 123: Oh, just remembered this @thaJeztah. I noticed if we signal the CLI multiple times without a little delay in between signals, the CLI doesn't consistently pick up all of them. I had no time at all to dig in to whether this is expected, or a symptom that we should look into the signal handling on the CLI (and in practice this is not a concern because users aren't really programatically signalling the CLI with `SIGINT`s 3 times in very very quick succession),  but I thought it was interesting

thaJeztah: Oh! Good question (and looks like I missed the "sleep" 😅).

That sounds like a TODO worthy as well, to figure out why the delay is needed. Agreed that it's not a blocker, but may be good to learn "why" so that we can make a more conscious decision (or a better comment that it's _expected_ if it is)

laurazard: Added TODO.

thaJeztah (approved):
LGTM, thanks!

PR #4834, opened 2024-01-29
cli/command: don't use pflags.XXXVarP for flags without shorthand

thaJeztah:
Use the `XXXVar` equivalent for flags that don't have a shorthand flag instead of passing an empty string for the shorthand flag.


thaJeztah:
Was working on some things in this area, and noticed we were using these, so let's get this out of the way first 😅

krissetto (approved):
LGTM

Issue #887, opened 2018-02-19
docker images list does not list any images


itilitymaarten:
**Description**
`docker image list` provides me with a list of images, e.g.:
```
$ docker image list
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
simple-graphql      latest              86b0df65e859        16 seconds ago      231MB
<none>              <none>              689940be3277        45 minutes ago      231MB
<none>              <none>              99be4d0ea751        About an hour ago   231MB
node                slim                04ab59975cba        43 hours ago        231MB
hello-world         latest              f2a91732366c        3 months ago        1.85kB
training/webapp     latest              6fae60ef3446        2 years ago         349MB
```

However, when running `docker images list` (which, as far as I know, should be an alias for `docker image list`, I get only the headers:
```
$ docker images list
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
```
I would expect these two to be the same; or is there an actual difference between `docker image` and `docker images`?

**Output of `docker version`:**

```
Docker version 18.02.0-ce, build fc4de447b5
```

**Output of `docker info`:**

```
Containers: 0
 Running: 0
 Paused: 0
 Stopped: 0
Images: 10
Server Version: 18.02.0-ce
Storage Driver: aufs
 Root Dir: /mnt/sda1/var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 23
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 9b55aab90508bd389d7654c4baf173a981477d55
runc version: 9f9c96235cc97674e935002fc3d78361b696a69e
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.4.115-boot2docker
Operating System: Boot2Docker 18.02.0-ce (TCL 8.2.1); HEAD : 99245f4 - Thu Feb  8 17:43:39 UTC 2018
OSType: linux
Architecture: x86_64
CPUs: 1
Total Memory: 995.9MiB
Name: default
ID: DWHU:76V2:7I3N:6DLZ:EK7E:XRYJ:BASW:IHDV:EXA4:7RLJ:TU4P:GCKV
[...]

Issue #4600, opened 2023-10-12
docker images is a great alias for docker image ls, while docker volumes ls does not exist


rafaelucena:
### Description

Using docker for a few years now, every few months or years I like to do a cleanup on the volumes and images and start over, and it keeps bugging me that for showing the images we can run `docker images` but for volumes we can't use `docker volumes`. I'm aiming to fix that.

When having at least one volume:

```shell
docker volumes
```

### Expected behavior

The output should be the same as:

```shell
docker volume ls
```

### docker version

(not strictly necessary?)

### docker info

(not strictly necessary?)

### Additional Info

This is my first issue, please be kind (sorry)

nicks:
i think `docker images` is considered a legacy command. it's there for backwards compatibility and we generally discourage it. see this great comment! https://github.com/docker/cli/issues/887#issuecomment-366654909

rafaelucena:
Yeah, @thaJeztah mentioned it on the PR and he explained it in another issue from another user also. It's a shame really, my favourite command is expect to depart :(

Well, it is what it is. Oh well, I'll probably just handle some bash aliases locally for that :)

RIP.

thaJeztah:
Yes, closing this one, because this was by design.

PR #4849, opened 2024-02-03
images: print hint when invoking "docker images" with ambiguous argument

thaJeztah:
- addresses / closes https://github.com/docker/cli/issues/887
- addresses / closes https://github.com/docker/cli/issues/4591
- related https://github.com/docker/cli/pull/4601
- related https://github.com/docker/cli/issues/4600
- related https://github.com/docker/cli/pull/3758

### images: print hint when invoking "docker images" with ambiguous argument

The `docker images` top-level subcommand predates the `docker <object> <verb>`
convention (e.g. `docker image ls`), but accepts a positional argument to
search/filter images by name (globbing). It's common for users to accidentally
mistake these commands, and to use (e.g.) `docker images ls`, expecting
to see all images, but ending up with an empty list because no image named
"ls" was found.

Disallowing these search-terms would be a breaking change, but we can print
and informational message to help the users correct their mistake.

Before this patch:

    docker images ls
    REPOSITORY   TAG       IMAGE ID   CREATED   SIZE

With this patch applied:

    docker images ls
    REPOSITORY   TAG       IMAGE ID   CREATED   SIZE
    
    No images found matching "ls": did you mean "docker image ls"?

```
print hint when invoking "docker images" with ambiguous argument
```

### DIFF
--- a/cli/command/image/list.go
+++ b/cli/command/image/list.go
@@ -2,6 +2,8 @@ package image
 
 import (
 	"context"
+	"fmt"
+	"io"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
@@ -21,6 +23,7 @@ type imagesOptions struct {
 	showDigests bool
 	format      string
 	filter      opts.FilterOpt
+	calledAs    string
 }
 
 // NewImagesCommand creates a new `docker images` command
@@ -35,6 +38,10 @@ func NewImagesCommand(dockerCLI command.Cli) *cobra.Command {
 			if len(args) > 0 {
 				options.matchName = args[0]
 			}
+			// Pass through how the command was invoked. We use this to print
+			// warnings when an ambiguous argument was passed when using the
+			// legacy (top-level) "docker images" subcommand.
+			options.calledAs = cmd.CalledAs()
 			return runImages(cmd.Context(), dockerCLI, options)
 		},
 		Annotations: map[string]string{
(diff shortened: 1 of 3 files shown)


cli/command/image/list.go
```diff
@@ -55,45 +62,82 @@ func NewImagesCommand(dockerCli command.Cli) *cobra.Command {
...
 		Context: formatter.Context{
-			Output: dockerCli.Out(),
+			Output: dockerCLI.Out(),
 			Format: formatter.NewImageFormat(format, options.quiet, options.showDigests),
 			Trunc:  !options.noTrunc,
 		},
 		Digest: options.showDigests,
 	}
-	return formatter.ImageWrite(imageCtx, images)
+	if err := formatter.ImageWrite(imageCtx, images); err != nil {
```
cpuguy83, line 103: Why print the empty image list before checking if its ambiguous?
I guess it can make it a little more obvious what could be happening.

cpuguy83 (approved):
Overall LGTM.
Glad to have something in place for this.

cli/command/image/list.go
thaJeztah: I continued treating as a regular search that didn't return a result. eg, a search for an image named `build` could still be legit, so keeping the behavior as usual, just the extra message on stderr

thaJeztah:
@laurazard @jalonsogo ptal 🤗

laurazard (approved):
LGTM!

vvoland (approved):

thaJeztah:
Thanks all! Let me bring this in; we can tweak the hint if we want to in follow ups.

PR #4854, opened 2024-02-06
Uniform output of volume prune cmd with other prune cmds

### LINKED ISSUE #4851: Canceling "volume prune" prints "Total reclaimed space: 0B" but likely shouldn't
### Description

I noticed that canceling `docker volume prune` (using `N`, or pressing `ENTER`) prints a summary of the amount of data reclainmed. Given that the user asked to _cancel_ the command (don't perform `prune`), it likely should not print that message. If it _does_ print a message, that message should reflect the course of action taken (i.e.; the command was canceled, not executed).

```bash
$ docker volume prune
Total reclaimed space: 0B
```

Suggested changes:

- Consider omitting the message, OR printing an alternative message (canceled the command).
- And/or consider printing `N` if the user presses `ENTER` (which is equivalent to "use the default (`N`))

Perhaps just printing the taken option would be enough (and minimal amount of noise)? i.e. after pressing `ENTER`, it would show;

```bash
$ docker volume prune
Are you sure you want to continue? [y/N] N
$
```

I also (briefly) went looking for "prior art". It looks like Bash (and other shells), simply print a prompt, but without printing default, nor printing the action taken (looks to be default is `no`);

```bash
$ touch some-file.txt
$ rm -i some-file.txt
rm: remove regular empty file 'some-file.txt'?
$ ls -a
.  .. some-file.txt
$ rm -i some-file.txt
rm: remove regular empty file 'some-file.txt'? y
$ ls -a
.  ..
```

```zsh
thajeztah@m1book cli % rm -i some-file.txt
remove some-file.txt?

thajeztah@m1book cli % ls
.  .. some-file.txt
```

I'm personally not a fan of that approach, as

- it's unclear "what happened" (what was the default? what happened when I pressed `ENTER`?)
- what are valid options? (I assumed `y`, `n`, but it doesn't show)

(But open to better suggestions!)

krissetto:
**- What I did**  

Make the `docker volume prune` cmd out uniform with the other `prune` commands in case of user cancellation at the confirmation prompt.   

Basically, don't show the `Total reclaimed space: 0 ` text when the user says "n" or simply presses ENTER at the confirmation prompt.

Closes #4851 

**- How I did it**
- Return error when user refuses at confirmation prompt
- Avoid sending space freed msg if user cancelled
- Fixed unit tests assertions and logic, some "success" tests were actually going through the failure path, and we were none the wiser :)

**- How to verify it**

photo_5913688840516780494_y

### DIFF
--- a/cli/command/volume/prune_test.go
+++ b/cli/command/volume/prune_test.go
@@ -73,13 +73,15 @@ func TestVolumePruneSuccess(t *testing.T) {
 	testCases := []struct {
 		name            string
 		args            []string
+		input           string
 		volumePruneFunc func(args filters.Args) (types.VolumesPruneReport, error)
 	}{
 		{
-			name: "all",
-			args: []string{"--all"},
+			name:  "all",
+			args:  []string{"--all"},
+			input: "y",
 			volumePruneFunc: func(pruneFilter filters.Args) (types.VolumesPruneReport, error) {
-				assert.Check(t, is.Equal([]string{"true"}, pruneFilter.Get("all")))
+				assert.Check(t, is.DeepEqual([]string{"true"}, pruneFilter.Get("all")))
 				return types.VolumesPruneReport{}, nil
 			},
 		},
@@ -91,10 +93,11 @@ func TestVolumePruneSuccess(t *testing.T) {
 			},
 		},
 		{
-			name: "label-filter",
-			args: []string{"--filter", "label=foobar"},
+			name:  "label-filter",
+			args:  []string{"--filter", "label=foobar"},
+			input: "y",
 			volumePruneFunc: func(pruneFilter filters.Args) (types.VolumesPruneReport, error) {
-				assert.Check(t, is.Equal([]string{"foobar"}, pruneFilter.Get("label")))
+				assert.Check(t, is.DeepEqual([]string{"foobar"}, pruneFilter.Get("label")))
 				return types.VolumesPruneReport{}, nil
 			},
 		},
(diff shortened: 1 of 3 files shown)


cli/command/volume/prune_test.go
```diff
@@ -73,13 +73,15 @@ func TestVolumePruneSuccess(t *testing.T) {
...
 		volumePruneFunc func(args filters.Args) (types.VolumesPruneReport, error)
 	}{
 		{
-			name: "all",
-			args: []string{"--all"},
+			name:  "all",
+			args:  []string{"--all"},
+			input: "y",
 			volumePruneFunc: func(pruneFilter filters.Args) (types.VolumesPruneReport, error) {
-				assert.Check(t, is.Equal([]string{"true"}, pruneFilter.Get("all")))
```
krissetto, line 82: I noticed we weren't checking these properly previously, so I added an input param for stdin in the test cases and made this quick fix for the comparison since we can't directly compare `[]string`. We can clean it up better if you think it could be worth it

cli/command/volume/prune_test.go
```diff
@@ -73,13 +73,15 @@ func TestVolumePruneSuccess(t *testing.T) {
...
+				assert.Check(t, is.Equal("true", pruneFilter.Get("all")[0]))
```
thaJeztah, line 84: Perhaps `is.DeepEqual` would work for this one, it we also want to make sure we only have a single value 🤔

I'm a bit surprised that `is.Equal` didn't complain about this (wondering if it should do so (non-comparable types))? Perhaps worth looking if we can contribute a fix for that in `gotest.tools`

```suggestion
				assert.Check(t, is.DeepEqual([]string{"true"}, pruneFilter.Get("all")))
```

krissetto: `is.Equal` does seem to complain, but only at runtime, and previously we were actually not even running that code (without running `cli.SetIn()` with an input, we were actually entering the code path where the user presses `ENTER` or `'n'/'N'` during the confirmation dialog. And yes, the output messages were the same since the prune affected 0B in either case 😅

thaJeztah: 🤦 😂

thaJeztah (approved):
LGTM, thanks!

PR #4860, opened 2024-02-07
docker stack: allow '=' separator in extra_hosts

### LINKED ISSUE #4859: 'extra_hosts' not working anymore in 'docker stack deploy'
### Description

So far, it was possible to pass docker compose config output to docker stack deploy without problems.

Today I upgraded to docker 25.0.2, and discovered loss of the extra_hosts mapping. The tag is completely ignored with the command, and nothing is written to /etc/hosts anymore!

This might be related to:
https://github.com/compose-spec/compose-go/pull/499
https://github.com/compose-spec/compose-spec/issues/156
https://github.com/docker/buildx/issues/2201

### Reproduce

cat docker-compose.yml
```
version: '3.8'

services:
  app:
    image: httpd:2.4
    extra_hosts:
      - "host.docker.internal:host-gateway"
```

Run:
```
docker compose config | (sed '/^name:/d') | docker stack deploy -c - httpd
```

### Expected behavior

`/etc/hosts` should show the entry from `extra_hosts`.

### docker version

```bash
 Version:           25.0.2
 API version:       1.44
 Go version:        go1.21.6
 Git commit:        29cf629
 Built:             Thu Feb  1 00:23:19 2024
 Context:           default

 Engine:
  Version:          25.0.2
  Go version:       go1.21.6
  Git commit:       fce6e0c
  Built:            Thu Feb  1 00:23:19 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.28
  GitCommit:        ae07eda36dd25f8a1b98dfbf587313b99c0190bb
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
 Version:    25.0.2
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.12.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
    Version:  v2.24.5
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 10
 Server Version: 25.0.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: active
  NodeID: j9wtq70drkn7uz09ve5cx7iru
  Is Manager: true
  ClusterID: 3kmynxgzfqluz4h5wv46zpu4j
  Managers: 1
  Nodes: 1
  Default Address Pool: 10.0.0.0/8  
  SubnetSize: 24
  Data Path Port: 4789
  Orchestration:
   Task History Retention Limit: 5
  Raft:
   Snapshot Interval: 10000
   Number of Old Snapshots to Retain: 0
   Heartbeat Tick: 1
   Election Tick: 10
  Dispatcher:
   Heartbeat Period: 5 seconds
  CA Configuration:
   Expiry Duration: 3 months
   Force Rotate: 0
  Autolock Managers: false
  Root Rotation In Progress: false
  Node Address: 127.0.0.1
  Manager Addresses:
   127.0.0.1:2377
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: ae07eda36dd25f8a1b98dfbf587313b99c0190bb
 runc version: v1.1.12-0-g51d5e94
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.2.0-39-generic
 Operating System: Ubuntu 23.04
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 62GiB
 Name:
 ID: 
 Debug Mode: false
 Username: 
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

### Additional Info

_No response_

robmry:
**- What I did**

The compose file format allows `=` as a separator in extra hosts, and `docker compose config` uses it instead of `:`.

Accept either separator in `docker stack deploy`, converting to `:` for the engine's API.

Also, remove brackets from IP addresses, as they're allowed by the compose file (although not generated by `docker compose config`).

Fixes #4859

**- How I did it**

Similar to changes in the CLI's `--add-hosts`, buildx and compose - check for an `=` separator first, then try `:`, and strip the brackets.

**- How to verify it**

As described in #4859 - plus updated unit tests.

Accept '=' separators and '[ipv6]' in compose files for 'docker stack deploy'.

### DIFF
--- a/cli/compose/loader/loader.go
+++ b/cli/compose/loader/loader.go
@@ -328,7 +328,7 @@ func createTransformHook(additionalTransformers ...Transformer) mapstructure.Dec
 		reflect.TypeOf(types.MappingWithEquals{}):                transformMappingOrListFunc("=", true),
 		reflect.TypeOf(types.Labels{}):                           transformMappingOrListFunc("=", false),
 		reflect.TypeOf(types.MappingWithColon{}):                 transformMappingOrListFunc(":", false),
-		reflect.TypeOf(types.HostsList{}):                        transformListOrMappingFunc(":", false),
+		reflect.TypeOf(types.HostsList{}):                        transformHostsList,
 		reflect.TypeOf(types.ServiceVolumeConfig{}):              transformServiceVolumeConfig,
 		reflect.TypeOf(types.BuildConfig{}):                      transformBuildConfig,
 		reflect.TypeOf(types.Duration(0)):                        transformStringToDuration,
@@ -808,26 +808,56 @@ var transformStringList TransformerFunc = func(data any) (any, error) {
 	}
 }
 
-func transformMappingOrListFunc(sep string, allowNil bool) TransformerFunc {
-	return func(data any) (any, error) {
-		return transformMappingOrList(data, sep, allowNil), nil
+var transformHostsList TransformerFunc = func(data any) (any, error) {
+	hl := transformListOrMapping(data, ":", false, []string{"=", ":"})
+
+	// Remove brackets from IP addresses if present (for example "[::1]" -> "::1").
+	result := make([]string, 0, len(hl))
+	for _, hip := range hl {
+		host, ip, _ := strings.Cut(hip, ":")
+		if len(ip) > 2 && ip[0] == '[' && ip[len(ip)-1] == ']' {
+			ip = ip[1 : len(ip)-1]
+		}
+		result = append(result, fmt.Sprintf("%s:%s", host, ip))
 	}
+	return result, nil
 }
 
-func transformListOrMappingFunc(sep string, allowNil bool) TransformerFunc {
-	return func(data any) (any, error) {
-		return transformListOrMapping(data, sep, allowNil), nil
-	}
-}
-
-func transformListOrMapping(listOrMapping any, sep string, allowNil bool) any {
+// transformListOrMapping transforms pairs of strings that may be represented as
+// a map, or a list of '=' or ':' separated strings, into a list of ':' separated
+// strings.
+func transformListOrMapping(listOrMapping any, sep string, allowNil bool, allowSeps []string) []string {
 	switch value := listOrMapping.(type) {
 	case map[string]any:
 		return toStringList(value, sep, allowNil)
 	case []any:
-		return listOrMapping
+		result := make([]string, 0, len(value))
+		for _, entry := range value {
+			for i, allowSep := range allowSeps {
+				entry := fmt.Sprint(entry)
+				k, v, ok := strings.Cut(entry, allowSep)
+				if ok {
+					// Entry uses this allowed separator. Add it to the result, using
+					// sep as a separator.
+					result = append(result, fmt.Sprintf("%s%s%s", k, sep, v))
+					break
+				} else if i == len(allowSeps)-1 {
+					// No more separators to try, keep the entry if allowNil.
+					if allowNil {
+						result = append(result, k)
+					}
+				}
+			}
+		}
+		return result
 	}
 	panic(errors.Errorf("expected a map or a list, got %T: %#v", listOrMapping, listOrMapping))
+}
+
+func transformMappingOrListFunc(sep string, allowNil bool) TransformerFunc {
+	return func(data any) (any, error) {
+		return transformMappingOrList(data, sep, allowNil), nil
+	}
 }
 
 func transformMappingOrList(mappingOrList any, sep string, allowNil bool) any {
--- a/cli/compose/loader/loader_test.go
+++ b/cli/compose/loader/loader_test.go
@@ -1302,12 +1302,14 @@ services:
     extra_hosts:
       "zulu": "162.242.195.82"
       "alpha": "50.31.209.229"
+      "beta": "[fd20:f8a7:6e5b::2]"
       "host.docker.internal": "host-gateway"
 `)
 	assert.NilError(t, err)
 
 	expected := types.HostsList{
 		"alpha:50.31.209.229",
+		"beta:fd20:f8a7:6e5b::2",
 		"host.docker.internal:host-gateway",
 		"zulu:162.242.195.82",
 	}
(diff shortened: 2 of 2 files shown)


cli/compose/loader/loader.go
```diff
@@ -808,28 +808,58 @@ var transformStringList TransformerFunc = func(data any) (any, error) {
...
+		if len(ip) > 2 && ip[0] == '[' && ip[len(ip)-1] == ']' {
+			ip = ip[1 : len(ip)-1]
+		}
+		result = append(result, fmt.Sprintf("%s:%s", host, ip))
 	}
+	return result, nil
 }
 
-func transformListOrMapping(listOrMapping any, sep string, allowNil bool) any {
+func transformListOrMapping(listOrMapping any, sep string, allowNil bool, allowSeps []string) any {
```
krissetto, line 829: Could it make sense to change the return type here to `[]string` instead of `any` now that these are the only options?

cli/compose/loader/loader.go
```diff
@@ -808,28 +808,58 @@ var transformStringList TransformerFunc = func(data any) (any, error) {
 	}
 }
 
-func transformMappingOrListFunc(sep string, allowNil bool) TransformerFunc {
-	return func(data any) (any, error) {
-		return transformMappingOrList(data, sep, allowNil), nil
-	}
-}
+var transformHostsList TransformerFunc = func(data any) (any, error) {
+	// Transform pairs of host and IP addr that may be represented as a map, or a list
```
krissetto, line 812: This description could be moved to be a proper docstring for the `transformListOrMapping` func below for better IDE integration

cli/compose/loader/loader_test.go
```diff
@@ -1324,16 +1326,23 @@ services:
     image: busybox
     extra_hosts:
       - "zulu:162.242.195.82"
+      - "whiskey=162.242.195.83"
       - "alpha:50.31.209.229"
       - "zulu:ff02::1"
-      - "host.docker.internal:host-gateway"
+      - "whiskey=ff02::2"
+      - "foxtrot=[ff02::3]"
```
krissetto, line 1333: Do we support ipv6 addresses with  `[]` brackets and the `:` separator? If so we're missing a test for that

cli/compose/loader/loader_test.go
```diff
@@ -1324,16 +1326,23 @@ services:
     image: busybox
     extra_hosts:
```
krissetto, line 1327: Generally, maybe it could make sense to think about writing a small func that takes in the valid separators etc and generates the list of all possible (acceptable) permutations, just to avoid maybe forgetting some possibilities behind and eventually hitting unexpected bugs. What do you think?

cli/compose/loader/loader_test.go
```diff
@@ -1324,16 +1326,23 @@ services:
...
       - "zulu:162.242.195.82"
+      - "whiskey=162.242.195.83"
       - "alpha:50.31.209.229"
       - "zulu:ff02::1"
-      - "host.docker.internal:host-gateway"
+      - "whiskey=ff02::2"
```
sam-thibault, line 1332: is having two hosts named "whiskey" (one ipv4 and one ipv6) intentional for this test?

robmry: Yes, it's the same as the two existing "zulu" entries. In an '/etc/hosts' file, hosts get separate lines for IPv4/IPv6 addresses. (And, more whiskey is always good.)

cli/compose/loader/loader.go
robmry: Yes, might as well - done.

cli/compose/loader/loader.go
robmry: Done.

cli/compose/loader/loader_test.go
robmry: I think maybe that can wait until there are more permutations to deal with? (None of this code actually cares that the addresses are addresses, so the IPv4/IPv6 distinction isn't really material ... it's just sorting out the separator and removing brackets.)

cli/compose/loader/loader.go
```diff
@@ -808,28 +808,58 @@ var transformStringList TransformerFunc = func(data any) (any, error) {
...
+func transformListOrMapping(listOrMapping any, sep string, allowNil bool, allowSeps []string) []string {
 	switch value := listOrMapping.(type) {
 	case map[string]any:
 		return toStringList(value, sep, allowNil)
 	case []any:
-		return listOrMapping
+		result := make([]string, 0, len(value))
+		for _, entry := range value {
+			for i, allowSep := range allowSeps {
+				entry := fmt.Sprint(entry)
```
krissetto, line 837: I'm guessing this is for stringify-ing `entry` since it's of type `any` in this switch case. What are the differences between casting it to string rather than using this method?   

I'd have personally used a cast something like  `entry, ok = entry.(string)` so we could handle eventual errors, but my go experience isn't all that profound and i might be missing details :)

robmry: Yes, it's probably a string, so casting it should work ... but, the old code didn't depend on it and I didn't want to change that. The map-handling half of this function calls `toStringList()`, which `Sprintf`'s the `any` value to turn it into a string - so I went with that.

sam-thibault (approved):

cli/compose/loader/loader.go
krissetto: Yeah I agree It's clearly not a big issue since the other parts have been using the strategy to begin with, but maybe we could adjust those as well down the line.. I'm not sure if there are any performance implications either. Just food for thought

krissetto (approved):

cli/compose/loader/loader.go
```diff
@@ -808,28 +808,58 @@ var transformStringList TransformerFunc = func(data any) (any, error) {
 	}
 }
 
-func transformMappingOrListFunc(sep string, allowNil bool) TransformerFunc {
-	return func(data any) (any, error) {
-		return transformMappingOrList(data, sep, allowNil), nil
-	}
-}
+var transformHostsList TransformerFunc = func(data any) (any, error) {
```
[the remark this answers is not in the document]
robmry: It's the pattern used for all of the functions in this table ...

https://github.com/docker/cli/blob/79fa65e7b58b081dfc2316a1a3c0f602f4d4c01a/cli/compose/loader/loader.go#L314-L335

Presumably it's to make it clear that it matches the `TransformerFunc` signature, so it's an explicit type for the function.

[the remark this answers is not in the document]
Benehiko: Ah i see, then that's fine :)

Benehiko (approved):

thaJeztah (approved):
LGTM

thaJeztah:
Added "cherry-pick" in case there will be another 25.0.x patch release to include it in.

MoogyG:
A 25.0.x patch would be great

PR #4862, opened 2024-02-08
Avoid keeping @docker_cli_[UUID] files

catap:
Seems that OpenBSD behaves like darwin and requires to unlink all socket, after it was used.

Tested on OpenBSD 7.4

### DIFF
--- a/cli-plugins/socket/socket_openbsd.go
+++ b/cli-plugins/socket/socket_openbsd.go
@@ -1,20 +1,19 @@
-//go:build !darwin && !openbsd
-
 package socket
 
 import (
 	"net"
+	"os"
+	"path/filepath"
+	"syscall"
 )
 
 func listen(socketname string) (*net.UnixListener, error) {
 	return net.ListenUnix("unix", &net.UnixAddr{
-		Name: "@" + socketname,
+		Name: filepath.Join(os.TempDir(), socketname),
 		Net:  "unix",
 	})
 }
 
 func onAccept(conn *net.UnixConn, listener *net.UnixListener) {
-	// do nothing
-	// while on darwin and OpenBSD we would unlink here;
-	// on non-darwin the socket is abstract and not present on the filesystem
+	syscall.Unlink(listener.Addr().String())
 }


cli-plugins/socket/socket_nodarwin.go
```diff
@@ -15,5 +15,6 @@ func listen(socketname string) (*net.UnixListener, error) {
 
 func onAccept(conn *net.UnixConn, listener *net.UnixListener) {
 	// do nothing
-	// while on darwin we would unlink here; on non-darwin the socket is abstract and not present on the filesystem
+	// while on darwin and OpenBSD we would unlink here;
+    // on non-darwin the socket is abstract and not present on the filesystem
```
thaJeztah, line 19: Looks like the indentation is off here (and using spaces, instead of tabs) 😅

cli-plugins/socket/socket_openbsd.go
```diff
@@ -0,0 +1,19 @@
...
+package socket
+
+import (
+	"net"
+	"syscall"
+)
+
+func listen(socketname string) (*net.UnixListener, error) {
+	return net.ListenUnix("unix", &net.UnixAddr{
+		Name: "@" + socketname,
```
thaJeztah, line 12: Curious; was there a reason to differentiate from the Darwin implementation here? Looks like the difference here with the Darwin one is that that one uses the system's temp-dir as an extra precaution from these ending up in the working-directory (which could also be the build-context);

```go
Name: filepath.Join(os.TempDir(), socketname),
```

catap: @thaJeztah good point. Mainly becuase it has naming darwin and nondarwin :)

But I like this way more, let me try it.

cli-plugins/socket/socket_nodarwin.go
catap: Thanks! fixed.

cli-plugins/socket/socket_openbsd.go
catap: It works like a charm

laurazard (approved):
Good catch! LGTM

cli-plugins/socket/socket_openbsd.go
thaJeztah: Ah, that's great!

I'm not at my computer right now; are the openBSD and Darwin implementations now exactly the same?

If they are, perhaps we should use the same file for both then. I guess the file suffix should be renamed to something else the  (_nolinux? _other?) good suggestions for that welcome 😅

cc @laurazard

catap: And better to rename nodarwin as well.

But I really have not idea regarding naming.

thaJeztah (approved):
Naming is hard, and I don't have a good suggestion for https://github.com/docker/cli/pull/4862#discussion_r1485940525

It's just a couple of lines of duplicated code, and we can easily fix in future if we come up with a name. I guess the only risk is for them to diverge, but let's keep that for a follow up 😅 

LGTM

PR #4866, opened 2024-02-09
docs: update link targets

dvdksn:
**- What I did**

Moving the files to a different URL scheme on docs.docker.com breaks some of the links in this repository. https://github.com/docker/docs/pull/19370

I've updated the affected links, and also patched the yaml generation in `cli-docs-tool` to handle the URL path discrepancy that now exists between this repository and the website. https://github.com/docker/cli-docs-tool/pull/49

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/commandline/cli.md
+++ b/docs/reference/commandline/cli.md
@@ -11,7 +11,7 @@ aliases:
 ---
 
 The base command for the Docker CLI is `docker`. For information about the
-available flags and subcommands, refer to the [CLI reference](docker.md)
+available flags and subcommands, refer to the [CLI reference](https://docs.docker.com/reference/cli/docker/)
 
 Depending on your Docker system configuration, you may be required to preface
 each `docker` command with `sudo`. To avoid having to use `sudo` with the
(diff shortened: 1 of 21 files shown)


docs/reference/commandline/cli.md
```diff
@@ -5,13 +5,13 @@ keywords: "Docker, Docker documentation, CLI, command line, config.json, CLI con
 aliases:
   - /reference/commandline/cli/
   - /engine/reference/commandline/engine/
-  - /engine/reference/commandline/engine_activate/
-  - /engine/reference/commandline/engine_check/
-  - /engine/reference/commandline/engine_update/
+  - /engine/reference/commandline/engine/activate/
+  - /engine/reference/commandline/engine/check/
+  - /engine/reference/commandline/engine/update/
```
thaJeztah, line 10: These should not be updated, as these were historic URLs

docs/reference/commandline/cli.md
```diff
@@ -28,27 +28,27 @@ line:
...
+| `DOCKER/CONFIG`               | The location of your client configuration files.                                                                                                                                                                                                             |
+| `DOCKER/CONTENT_TRUST_SERVER` | The URL of the Notary server to use. Defaults to the same URL as the registry.                                                                                                                                                                               |
+| `DOCKER/CONTENT_TRUST`        | When set Docker uses notary to sign and verify images. Equates to `--disable-content-trust=false` for build, create, pull, push, run.                                                                                                                        |
+| `DOCKER/CONTEXT`              | Name of the `docker context` to use (overrides `DOCKER_HOST` env var and default context set with `docker context use`)                                                                                                                                      |
```
thaJeztah, line 36: I guess we don't use `/` in env-vars 😜

docs/reference/commandline/cli.md
```diff
@@ -117,18 +117,18 @@ if no `--format` flag is provided.
...
-| `servicesFormat`       | Custom default format for `docker service ls` output. See [`docker service ls`](service_ls.md#format) for a list of supported formatting directives.                |
-| `statsFormat`          | Custom default format for `docker stats` output. See [`docker stats`](container_stats.md#format) for a list of supported formatting directives.                     |
-| `tasksFormat`          | Custom default format for `docker stack ps` output. See [`docker stack ps`](stack_ps.md#format) for a list of supported formatting directives.                      |
-| `volumesFormat`        | Custom default format for `docker volume ls` output. See [`docker volume ls`](volume_ls.md#format) for a list of supported formatting directives.                   |
+| `configFormat`         | Custom default format for `docker config ls` output. See [`docker config ls`](https://docs.docker.com/reference/cli/docker/config/ls/#format) for a list of supported formatting directives.                   |
```
thaJeztah, line 120: 🤔 honestly, not sure if this is the right thing to do. We were using `docs.docker.com` URLs for content maintained _outside of this repository_, but for content maintained here ("locally") using docs.docker.com links will be problematic.


- it's hard to verify (e.g. IDEs won't be able to validate anchors)
- we end up with a chicken-and-egg situation; we must add "broken links" for new content, as well as changes that we make (docs.docker.com won't be updated until we release and publish content).


I would _NOT_ be against moving away from a flat-structure though. We have a lot of commands in many groups, so organising those in directories `image/pull.md` and so on) would work for me if that allows use to continue using links to the markdown files.

We may have to look at the `cli-docs-tool` to facilitate this though!

thaJeztah (changes requested):

docs/reference/commandline/cli.md
dvdksn: 🙈

docs/reference/commandline/cli.md
dvdksn: not sure how it caught this. My python isn't what it used to be

docs/reference/commandline/cli.md
dvdksn: The problem with the problem with creating a nested structure with subdirectories would be that we'd have lots of `docs/reference/image/_index.md` files as well laying around. And it makes it much more involved to generate to/from these files. I think it'd be a lot more straightforward just having absolute links.

dvdksn: I've pushed a patch to cli-docs-tool that handles the URL path conversion when generating yaml: https://github.com/docker/cli-docs-tool/pull/49

So I reverted back to relative paths where possible. The `dockerd.md` and a couple of other files that aren't part of the cli reference are exempted. That's because they'll be served under a different url path entirely so the preprocessing step in `cli-docs-tool` doesn't work in those cases.

dvdksn:
not sure what the failing tests are here but probably not related? @thaJeztah

laurazard:
Re-ran the tests ✅

laurazard:
Weird failures, can you rebase @dvdksn?

dvdksn:
@laurazard rebased, looks OK now

docs/reference/commandline/plugin_create.md
```diff
@@ -15,7 +15,7 @@ Create a plugin from a rootfs and configuration. Plugin data directory must cont
 ## Description
 
 Creates a plugin. Before creating the plugin, prepare the plugin's root
-filesystem as well as the [config.json](../../extend/config.md).
+filesystem as well as the [config.json](https://docs.docker.com/engine/extend/config/).
```
thaJeztah, line 18: These can't work in the new situation? (mostly looking because (unlike, e.g. "dockerfile" reference) this is a file in the same repository)

docs/reference/commandline/plugin_inspect.md
```diff
@@ -21,7 +21,7 @@ in a JSON array.
 
 ### Inspect a plugin
 
-The following example example inspects the `tiborvass/sample-volume-plugin` plugin:
+The following example inspects the `tiborvass/sample-volume-plugin` plugin:
```
thaJeztah, line 24: Hm... we need to enable the "dupwords" linter. But perhaps it only checks Go code, and not markdown files 🤔

docs/reference/dockerd.md
```diff
@@ -133,21 +133,21 @@ to [the `daemon.json` file](#daemon-configuration-file).
...
-| `MOBY_DISABLE_PIGZ` | Disables the use of [`unpigz`](https://linux.die.net/man/1/pigz) to  decompress layers in parallel when pulling images, even if it is installed.                                  |
-| `NO_PROXY`          | Comma-separated values specifying hosts that should be excluded from proxying. See the [Go specification](https://pkg.go.dev/golang.org/x/net/http/httpproxy#Config) for details. |
+| Variable            | Description                                                                                                                                                                             |
+| :------------------ | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `DOCKER_CERT_PATH`  | Location of your authentication keys. This variable is used both by the [`docker` CLI](https://docs.docker.com/engine/reference/commandline/cli/) and the `dockerd` daemon.             |
```
thaJeztah, line 142: Should this one still have `/engine/` in the path? I see the other ones use `/reference/`

docs/reference/dockerd.md
```diff
@@ -133,21 +133,21 @@ to [the `daemon.json` file](#daemon-configuration-file).
...
+| `DOCKER_DRIVER`     | The storage driver to use.                                                                                                                                                              |
+| `DOCKER_RAMDISK`    | If set this disables `pivot_root`.                                                                                                                                                      |
+| `DOCKER_TLS_VERIFY` | When set Docker uses TLS and verifies the remote. This variable is used both by the [`docker` CLI](https://docs.docker.com/engine/reference/commandline/cli/) and the `dockerd` daemon. |
```
thaJeztah, line 145: Same for this one

dvdksn: See above ^

thaJeztah (approved):
LGTM

PR #4869, opened 2024-02-10
cli/connhelper/commandconn: inline variables

thaJeztah:
Inline the variables used to define the command + args used in the tests, which makes it slightly easier to see what's run.

Also explicitly define a context, in case we want to add telemetry to these tests.

**- What I did**

**- How I did it**

**- How to verify it**


laurazard (approved):
LGTM

PR #4331, opened 2023-06-02
WIP: mount: Add volume-subpath option 

vvoland:
- Related to: https://github.com/moby/moby/pull/45687

**- What I did**
- Added `volume-subpath` option to the long `--mount` format.

**- How I did it**

**- How to verify it**
TestMountSubvolume test (more tests in Moby)

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #4873, opened 2024-02-12
testenv: Add DaemonAPIVersion helper

vvoland:
- extracted from: https://github.com/docker/cli/pull/4331

Allow tests to check the negotiated API version used by the client.

Can be used to skip tests based on API versions, for example:
```go
    skip.If(t, versions.LessThan(environment.DaemonAPIVersion(t), "1.44"))
```

will skip the test if the API version is older than 1.44

**- What I did**

**- How I did it**

**- How to verify it**


krissetto (approved):
LGTM

thaJeztah (approved):
LGTM

PR #4878, opened 2024-02-15
fix: concurrency safe `net.UnixConn`

Benehiko:
**- What I did**

Create a `*net.UnixConn` receive-only channel, forcing the consumer to wait for the `listener.AcceptUnix()` to return a new connection before it can be used. We don't actually use the returned connection much outside of the tests, only when the CLI receives a Termination signal.

The race condition seemed to also only affect tests that were accessing the `net.UnixConn` instance before it was set or maybe even while it was being set. Since there was no proper mechanism for waiting for the connection to be set, we ended up with race conditions.

See below for some go code explaining this.

```go
var conn *net.UnixConn
// sets up an inifinte loop in the background
listener, err := SetupConn(&conn)

// poll the conn for non nill 
pollConnNotNil(t, conn) // <-- can succeed or fail here

// closes the connection
conn.Close() // <-- can succeed or fail here
```

```go
// Close closes the connection.
func (c *conn) Close() error {
	if !c.ok() {
		return syscall.EINVAL
	}
	err := c.fd.Close() // <--- fails
	if err != nil {
		err = &OpError{Op: "close", Net: c.fd.net, Source: c.fd.laddr, Addr: c.fd.raddr, Err: err}
	}
	return err
}
```

```shell
go test -v -race .
---
WARNING: DATA RACE
Read at 0x00c00013322c by goroutine 16:
  internal/poll.(*FD).Close()
      /usr/lib/go/src/internal/poll/fd_unix.go:112 +0x98
  net.(*netFD).Close()
      /usr/lib/go/src/net/fd_posix.go:37 +0x3d
  net.(*conn).Close()
      /usr/lib/go/src/net/net.go:203 +0x6a
  github.com/docker/cli/cli-plugins/socket.TestConnectAndWait.func1()
      /home/benehiko/go/src/github.com/docker/cli/cli-plugins/socket/socket_test.go:107 +0x2dd
  testing.tRunner()
      /usr/lib/go/src/testing/testing.go:1595 +0x261
  testing.(*T).Run.func1()
      /usr/lib/go/src/testing/testing.go:1648 +0x44
```

**- How I did it**
Return a receive-only channel containing the `net.UnixConn` instance.

**- How to verify it**
CI should not fail on tests:
```shell
--- FAIL: TestSetupConn (0.00s)
    --- FAIL: TestSetupConn/allows_reconnects (0.00s)
--- FAIL: TestConnectAndWait (0.00s)
    --- FAIL: TestConnectAndWait/connect_goroutine_exits_after_EOF (0.00s)
```

Locally I cannot find a race condition for these test anymore:
```shell
⋊> ~/g/s/g/d/c/c/socket on fix-socket-concurrency ⨯ go test -v -race ./...  11:55:12
=== RUN   TestSetupConn
=== RUN   TestSetupConn/updates_conn_when_connected
=== RUN   TestSetupConn/allows_reconnects
=== RUN   TestSetupConn/does_not_leak_sockets_to_local_directory
--- PASS: TestSetupConn (0.00s)
    --- PASS: TestSetupConn/updates_conn_when_connected (0.00s)
    --- PASS: TestSetupConn/allows_reconnects (0.00s)
    --- PASS: TestSetupConn/does_not_leak_sockets_to_local_directory (0.00s)
=== RUN   TestConnectAndWait
=== RUN   TestConnectAndWait/calls_cancel_func_on_EOF
=== RUN   TestConnectAndWait/connect_goroutine_exits_after_EOF
--- PASS: TestConnectAndWait (0.00s)
    --- PASS: TestConnectAndWait/calls_cancel_func_on_EOF (0.00s)
    --- PASS: TestConnectAndWait/connect_goroutine_exits_after_EOF (0.00s)
PASS
ok  	github.com/docker/cli/cli-plugins/socket	1.019s
```

The `socket.go` implementation introduced race conditions on the `net.UnixConn` instance passed on to the `SetupConn` function which is now resolved.

kitten-cat-cute-kitten-kitty-97c7d2448f0682f1f578c0b9f57aacab
(https://www.pickpik.com/kitten-cat-cute-kitten-kitty-cute-cat-curious-135563)

### DIFF
--- a/cli-plugins/socket/socket.go
+++ b/cli-plugins/socket/socket.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"time"
 )
 
 // EnvKey represents the well-known environment variable used to pass the plugin being
@@ -44,21 +45,32 @@ func randomID() string {
 func accept(listener *net.UnixListener) <-chan *net.UnixConn {
 	connChan := make(chan *net.UnixConn, 1)
 
-	go func(connChan chan<- *net.UnixConn) {
-		for {
+	go func() {
+		const maxRetries = 10
+		const waitBetweenRetries = 100 * time.Millisecond
+
+		var conn *net.UnixConn
+		var err error
+
+		// retry accepting a connection if there was an error
+		for i := 0; i < maxRetries; i++ {
 			// this is a blocking call and will wait
 			// until a new connection is accepted
-			c, err := listener.AcceptUnix()
+			// or until the timout is reached
+			conn, err = listener.AcceptUnix()
 
-			// perform any platform-specific actions on accept (e.g. unlink non-abstract sockets)
-			onAccept(listener)
-			// retry accepting a connection if there was an error
 			if err != nil {
+				time.Sleep(waitBetweenRetries)
 				continue
 			}
-			connChan <- c
+			break
 		}
-	}(connChan)
+		// perform any platform-specific actions on accept (e.g. unlink non-abstract sockets)
+		onAccept(listener)
+		connChan <- conn
+		// close the channel to signal we won't accept any more connections
+		close(connChan)
+	}()
 
 	return connChan
 }
--- a/cli-plugins/socket/socket.go
+++ b/cli-plugins/socket/socket.go
@@ -16,15 +16,18 @@ const EnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
 // SetupConn sets up a Unix socket listener, establishes a goroutine to handle connections
 // and update the conn pointer, and returns the listener for the socket (which the caller
 // is responsible for closing when it's no longer needed).
-func SetupConn(conn **net.UnixConn) (*net.UnixListener, error) {
+func SetupConn() (*net.UnixListener, <-chan *net.UnixConn, error) {
 	listener, err := listen("docker_cli_" + randomID())
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	accept(listener, conn)
+	// accept starts a background goroutine
+	// to accept a new connection
+	// once accepted, the connChan will be updated.
+	connChan := accept(listener)
 
-	return listener, nil
+	return listener, connChan, nil
 }
 
 func randomID() string {
@@ -35,16 +38,29 @@ func randomID() string {
 	return hex.EncodeToString(b)
 }
 
-func accept(listener *net.UnixListener, conn **net.UnixConn) {
-	go func() {
+// accept creates a new Unix socket connection
+// and sends it to the *net.UnixConn channel
+// it allows reconnects
+func accept(listener *net.UnixListener) <-chan *net.UnixConn {
+	connChan := make(chan *net.UnixConn, 1)
+
+	go func(connChan chan<- *net.UnixConn) {
 		for {
-			// ignore error here, if we failed to accept a connection,
-			// conn is nil and we fallback to previous behavior
-			*conn, _ = listener.AcceptUnix()
+			// this is a blocking call and will wait
+			// until a new connection is accepted
+			c, err := listener.AcceptUnix()
+
 			// perform any platform-specific actions on accept (e.g. unlink non-abstract sockets)
-			onAccept(*conn, listener)
+			onAccept(listener)
+			// retry accepting a connection if there was an error
+			if err != nil {
+				continue
+			}
+			connChan <- c
 		}
-	}()
+	}(connChan)
+
+	return connChan
 }
 
 // ConnectAndWait connects to the socket passed via well-known env var,
--- a/cli-plugins/socket/socket.go
+++ b/cli-plugins/socket/socket.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"net"
 	"os"
-	"time"
 )
 
 // EnvKey represents the well-known environment variable used to pass the plugin being
@@ -46,25 +45,11 @@ func accept(listener *net.UnixListener) <-chan *net.UnixConn {
 	connChan := make(chan *net.UnixConn, 1)
 
 	go func() {
-		const maxRetries = 10
-		const waitBetweenRetries = 100 * time.Millisecond
+		// this is a blocking call and will wait
+		// until a new connection is accepted
+		// or until the timout is reached
+		conn, _ := listener.AcceptUnix()
 
-		var conn *net.UnixConn
-		var err error
-
-		// retry accepting a connection if there was an error
-		for i := 0; i < maxRetries; i++ {
-			// this is a blocking call and will wait
-			// until a new connection is accepted
-			// or until the timout is reached
-			conn, err = listener.AcceptUnix()
-
-			if err != nil {
-				time.Sleep(waitBetweenRetries)
-				continue
-			}
-			break
-		}
 		// perform any platform-specific actions on accept (e.g. unlink non-abstract sockets)
 		onAccept(listener)
 		connChan <- conn
--- a/cli-plugins/socket/socket.go
+++ b/cli-plugins/socket/socket.go
@@ -13,9 +13,10 @@ import (
 // executed the socket name it should listen on to coordinate with the host CLI.
 const EnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
 
-// SetupConn sets up a Unix socket listener, establishes a goroutine to handle connections
-// and update the conn pointer, and returns the listener for the socket (which the caller
-// is responsible for closing when it's no longer needed).
+// SetupConn sets up a Unix socket listener, establishes a goroutine to handle connections.
+// A listener is returned, along with a connection channel that will receive the established
+// connection. The channel *may* return a nil connection and should be checked once received.
+// The caller is responsible for closing the listener when it's no longer needed.
 func SetupConn() (*net.UnixListener, <-chan *net.UnixConn, error) {
 	listener, err := listen("docker_cli_" + randomID())
 	if err != nil {
@@ -40,11 +41,12 @@ func randomID() string {
 
 // accept creates a new Unix socket connection
 // and sends it to the *net.UnixConn channel
-// it allows reconnects
 func accept(listener *net.UnixListener) <-chan *net.UnixConn {
 	connChan := make(chan *net.UnixConn, 1)
 
 	go func() {
+		// close the channel to signal we won't accept any more connections
+		defer close(connChan)
 		// this is a blocking call and will wait
 		// until a new connection is accepted
 		// or until the timout is reached
@@ -53,8 +55,6 @@ func accept(listener *net.UnixListener) <-chan *net.UnixConn {
 		// perform any platform-specific actions on accept (e.g. unlink non-abstract sockets)
 		onAccept(listener)
 		connChan <- conn
-		// close the channel to signal we won't accept any more connections
-		close(connChan)
 	}()
 
 	return connChan
--- a/cli-plugins/socket/socket_test.go
+++ b/cli-plugins/socket/socket_test.go
@@ -27,22 +27,6 @@ func TestSetupConn(t *testing.T) {
 		pollConnNotNil(t, conn)
 	})
 
-	t.Run("allows reconnects", func(t *testing.T) {
-		listener, _, err := SetupConn()
-		assert.NilError(t, err)
-		assert.Check(t, listener != nil, "returned nil listener but no error")
-		addr, err := net.ResolveUnixAddr("unix", listener.Addr().String())
-		assert.NilError(t, err, "failed to resolve listener address")
-
-		otherConn, err := net.DialUnix("unix", nil, addr)
-		assert.NilError(t, err, "failed to dial returned listener")
-
-		otherConn.Close()
-
-		_, err = net.DialUnix("unix", nil, addr)
-		assert.NilError(t, err, "failed to redial listener")
-	})
-
 	t.Run("does not leak sockets to local directory", func(t *testing.T) {
 		listener, _, err := SetupConn()
 		assert.NilError(t, err)
--- a/cli-plugins/socket/socket_test.go
+++ b/cli-plugins/socket/socket_test.go
@@ -15,8 +15,7 @@ import (
 
 func TestSetupConn(t *testing.T) {
 	t.Run("updates conn when connected", func(t *testing.T) {
-		var conn *net.UnixConn
-		listener, err := SetupConn(&conn)
+		listener, conn, err := SetupConn()
 		assert.NilError(t, err)
 		assert.Check(t, listener != nil, "returned nil listener but no error")
 		addr, err := net.ResolveUnixAddr("unix", listener.Addr().String())
@@ -25,12 +24,11 @@ func TestSetupConn(t *testing.T) {
 		_, err = net.DialUnix("unix", nil, addr)
 		assert.NilError(t, err, "failed to dial returned listener")
 
-		pollConnNotNil(t, &conn)
+		pollConnNotNil(t, conn)
 	})
 
 	t.Run("allows reconnects", func(t *testing.T) {
-		var conn *net.UnixConn
-		listener, err := SetupConn(&conn)
+		listener, _, err := SetupConn()
 		assert.NilError(t, err)
 		assert.Check(t, listener != nil, "returned nil listener but no error")
 		addr, err := net.ResolveUnixAddr("unix", listener.Addr().String())
@@ -46,8 +44,7 @@ func TestSetupConn(t *testing.T) {
 	})
 
 	t.Run("does not leak sockets to local directory", func(t *testing.T) {
-		var conn *net.UnixConn
-		listener, err := SetupConn(&conn)
+		listener, _, err := SetupConn()
 		assert.NilError(t, err)
 		assert.Check(t, listener != nil, "returned nil listener but no error")
 		checkDirNoPluginSocket(t)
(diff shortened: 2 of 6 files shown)


thaJeztah:
Yeah, we should probably look if we indeed _need_ the for-loop, and if we do, if we can terminate it early, or in what conditions it should be terminated.

cli-plugins/socket/socket.go
```diff
@@ -7,16 +7,98 @@ import (
...
+	c.SocketLock.Lock()
+	defer c.SocketLock.Unlock()
+	return c.UnixConn == nil
+}
+
 // SetupConn sets up a Unix socket listener, establishes a goroutine to handle connections
 // and update the conn pointer, and returns the listener for the socket (which the caller
 // is responsible for closing when it's no longer needed).
-func SetupConn(conn **net.UnixConn) (*net.UnixListener, error) {
+func SetupConn(conn *UnixConnSafe) (*net.UnixListener, error) {
```
thaJeztah, line 101: If we decide to go with this implementation, I think I'd prefer keeping the `UnixConnSafe`  internal. This function could continue to accept a regular `net.UnixConn` and wrap it internally.

cli-plugins/socket/socket.go
```diff
@@ -7,16 +7,98 @@ import (
...
+	UnixConn   *net.UnixConn
+	SocketLock *sync.Mutex
+}
+
+var _ net.Conn = (*UnixConnSafe)(nil)
+
+func NewUnixConnSafe() *UnixConnSafe {
+	return &UnixConnSafe{
+		UnixConn:   &net.UnixConn{},
+		SocketLock: &sync.Mutex{},
```
thaJeztah, line 30: Where possible, it's good to keep mutexes "internal" nobody outside of "UnixConnSafe" itself should really have to deal with locking/unlocking.

thaJeztah (commented):
just quick blurbs; would have to take a closer look at this, but perhaps @laurazard is able to 🤗

laurazard:
Implementation looks good  but I do think we might be able to just remove the loop if that's what's causing the issue. Iirc (as discussed a bit in https://github.com/docker/cli/pull/4872#issuecomment-1943822305) it was a choice at the time (sounded more predictable) to let plugins redial the socket. However, in the meantime we've also made some general changes to how we handle the sockets, and have a different implementation on macOS/freebsd vs other places, etc. so I think it might just be simpler to remove it if we can.

We could try to remove it, build the CLI and then run the Compose tests against it/test the changes to see if we break anything, to get some signal back.

-------------

Also, wondering if we could be running tests with the race detector in CI too, maybe could have caught the other issue earlier – wdyt @thaJeztah?

cli-plugins/socket/socket.go
```diff
@@ -7,24 +7,107 @@ import (
...
+	defer c.socketLock.Unlock()
+	return c.unixConn == nil
+}
+
 // SetupConn sets up a Unix socket listener, establishes a goroutine to handle connections
 // and update the conn pointer, and returns the listener for the socket (which the caller
 // is responsible for closing when it's no longer needed).
-func SetupConn(conn **net.UnixConn) (*net.UnixListener, error) {
+func SetupConn() (*net.UnixListener, *UnixConnSafe, error) {
+	conn := NewUnixConnSafe()
```
Benehiko, line 102: TODO: `NewUnixConnSafe` should not be exposed.

cli-plugins/socket/socket.go
```diff
@@ -7,24 +7,107 @@ import (
...
 	listener, err := listen("docker_cli_" + randomID())
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	accept(listener, conn)
 
-	return listener, nil
+	return listener, conn, nil
```
Benehiko, line 110: Returning a new `net.UnixConn` instead of accepting.

cli-plugins/socket/socket.go
```diff
@@ -35,14 +118,20 @@ func randomID() string {
...
 		for {
 			// ignore error here, if we failed to accept a connection,
 			// conn is nil and we fallback to previous behavior
-			*conn, _ = listener.AcceptUnix()
+			c, _ := listener.AcceptUnix()
+
 			// perform any platform-specific actions on accept (e.g. unlink non-abstract sockets)
-			onAccept(*conn, listener)
+			onAccept(listener)
+			if c != nil {
```
Benehiko, line 130: Do we want to make sure it's not nil? Or always set it

cli-plugins/socket/socket_test.go
```diff
@@ -78,18 +75,21 @@ func checkDirNoPluginSocket(t *testing.T) {
...
 	t.Run("calls cancel func on EOF", func(t *testing.T) {
-		var conn *net.UnixConn
-		listener, err := SetupConn(&conn)
+		listener, conn, err := SetupConn()
 		assert.NilError(t, err, "failed to setup listener")
+		t.Cleanup(func() {
+			if !conn.IsNil() {
+				conn.Close()
+			}
+		})
```
Benehiko, line 84: TODO: cleanup. I was testing some failures here (or rather flakes). Seems the timeout gets hit _sometimes_...

cli-plugins/socket/socket_test.go
```diff
@@ -78,18 +75,21 @@ func checkDirNoPluginSocket(t *testing.T) {
...
 
 		done := make(chan struct{})
 		t.Setenv(EnvKey, listener.Addr().String())
 		cancelFunc := func() {
 			done <- struct{}{}
 		}
 		ConnectAndWait(cancelFunc)
-		pollConnNotNil(t, &conn)
-		conn.Close()
+		// pollConnNotNil(t, conn)
```
Benehiko, line 92: TODO: cleanup

Benehiko (commented):
I'm still busy with the PR. Found some of the tests to still be flaky (timeouts of 10ms reached).

Benehiko:
>We could try to remove it, build the CLI and then run the Compose tests against it/test the changes to see if we break anything, to get some signal back.

I think it's okay to leave in the loop, since we now have the `net.UnixConn` wrapped and properly do locking/unlocking, it should be fine :)

cpuguy83:
I don't know this bit of code at all.
But this for loop looks really bad.
I would be in favor of doing this right rather than adding more stuff to deal with the bad implementation.

Benehiko:
> I don't know this bit of code at all. But this for loop looks really bad. I would be in favor of doing this right rather than adding more stuff to deal with the bad implementation.

So it seems the reason why there is an infinite loop is to accept reconnects. I refactored this to return a receive-only channel. This forces the consumer to wait for the connection to actually be made before it tries to access the `net.UnixConn` instance.
https://github.com/docker/cli/blob/master/cli-plugins/socket/socket_test.go#L31-L46

laurazard (changes requested):
Overall, I think we're getting caught up in these discussions over two different things – one is the issue with the loop around accepting the connection, which is important to fix but also orthogonal to the race conditions these changes try to fix (which we should also try to fix).

Re: the race condition and addressing them, both of the approaches (the one with the lock and the one with the channel) would do, but I think the channel approach will lead to more complex code (and timeouts, etc.) that we should avoid. We could easily do the same with a wrapper around `net.UnixConn` with a mutex (your other approach), and also probably use `TryLock` in the hot path when we need to check whether we can use the connection or not to signal the plugin to exit.

The issue with the loop however is still there, since in both of these approaches you're trying to preserve the "allow reconnects" behavior. The loop is dangerous in it's current form because if anything causes the accept to continuously fail, we'll enter a tight loop there, which we really don't want. To solve that, we either need to:
- only retry accepting the connection when the first accept succeeded
- only retry when the first accept succeeded or there was an error in some defined list of "okay errors that we know of"
- some retry limit
- remove the loop altogether

I'm in favour of either the first or last option

cli-plugins/socket/socket.go
```diff
@@ -35,16 +38,29 @@ func randomID() string {
...
+			// this is a blocking call and will wait
+			// until a new connection is accepted
+			c, err := listener.AcceptUnix()
+
 			// perform any platform-specific actions on accept (e.g. unlink non-abstract sockets)
-			onAccept(*conn, listener)
+			onAccept(listener)
+			// retry accepting a connection if there was an error
+			if err != nil {
+				continue
```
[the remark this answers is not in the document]
Benehiko: I'm not sure I follow. I kept the previous behavior in-tact here since we _always_ called `onAccept`, even on failure.

https://github.com/docker/cli/blob/master/cli-plugins/socket/socket.go#L43-L45

cli-plugins/socket/socket.go
```diff
@@ -35,16 +38,29 @@ func randomID() string {
...
 
-func accept(listener *net.UnixListener, conn **net.UnixConn) {
-	go func() {
+// accept creates a new Unix socket connection
+// and sends it to the *net.UnixConn channel
+// it allows reconnects
+func accept(listener *net.UnixListener) <-chan *net.UnixConn {
+	connChan := make(chan *net.UnixConn, 1)
+
+	go func(connChan chan<- *net.UnixConn) {
```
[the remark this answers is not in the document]
Benehiko: what is the reason behind this suggestion? Wouldn't it be best to be explicit here?

cmd/docker/docker.go
```diff
@@ -247,11 +246,15 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
 				// receive signals due to sharing a pgid with the parent CLI
 				continue
 			}
-			if conn != nil {
-				if err := conn.Close(); err != nil {
-					_, _ = fmt.Fprintf(dockerCli.Err(), "failed to signal plugin to close: %v\n", err)
+			select {
+			case c := <-conn:
```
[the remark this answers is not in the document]
Benehiko: Good point, I will close the channel so we don't block on it on every loop.

cli-plugins/socket/socket.go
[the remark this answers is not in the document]
Benehiko: moved this to only be called when the connection is nil. I don't know the history behind this, but it seems to remove the socket file, which I assume is done when the connection failed as a cleanup routine?

cli-plugins/socket/socket.go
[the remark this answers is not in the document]
neersighted: `go func()` is a much more common pattern, and the shadowing here may lead to the scope being **less** obvious. Using the closure to capture the variables here is a very common pattern, though to be fair I can't find any authoritative sources (outside of "look at code examples") for it.

See https://jeremybytes.blogspot.com/2021/02/go-golang-anonymous-functions-inlining.html for a non-authoritative take.

neersighted (dismissed):
Overall LGTM with the one nit @laurazard pointed out.

cli-plugins/socket/socket.go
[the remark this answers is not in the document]
Benehiko: I also couldn't find an authoritative source for this. Thanks for the link!
I'll apply this change and then we can get this merged :)

cli-plugins/socket/socket_test.go
```diff
@@ -25,12 +24,11 @@ func TestSetupConn(t *testing.T) {
 		_, err = net.DialUnix("unix", nil, addr)
 		assert.NilError(t, err, "failed to dial returned listener")
 
-		pollConnNotNil(t, &conn)
+		pollConnNotNil(t, conn)
 	})
 
 	t.Run("allows reconnects", func(t *testing.T) {
```
laurazard, line 30: So supposedly, this test should ensure that we let plugins redial the connection, but in reality the test is a little incomplete – we're ensuring that redialing the connection doesn't error out, but not that we can use it.

cli-plugins/socket/socket.go
```diff
@@ -35,16 +39,40 @@ func randomID() string {
...
+
+			if err != nil {
+				time.Sleep(waitBetweenRetries)
+				continue
+			}
+			break
 		}
+		// perform any platform-specific actions on accept (e.g. unlink non-abstract sockets)
+		onAccept(listener)
+		connChan <- conn
```
laurazard, line 70: However, these changes make it so that now, after a successful connection, we leave this loop and don't accept connections anymore.

cli-plugins/socket/socket_test.go
laurazard: For clarity, imagine a test such as:
```go
listener, conn, err := SetupConn()
assert.NilError(t, err)
assert.Check(t, listener != nil, "returned nil listener but no error")
addr, err := net.ResolveUnixAddr("unix", listener.Addr().String())
assert.NilError(t, err, "failed to resolve listener address")

otherConn, err := net.DialUnix("unix", nil, addr)
assert.NilError(t, err, "failed to dial returned listener")
pollConnNotNil(t, conn)

otherConn.Close()

_, err = net.DialUnix("unix", nil, addr)

pollConnNotNil(t, conn)
assert.NilError(t, err, "failed to redial listener")
```
(this fails)

laurazard (changes requested):
We're changing the contract here from "always allow redialing" to "allow redialing only when the first attempt failed", which should have been caught by our tests but wasn't. 

The test wasn't testing we could actually use the connection after redialing, so when this changed, we missed it. Before, we stayed in the loop, and if we accepted another connection, we'd update the pointer to the connection to the new one.
Now, if we wanted to keep accepting connections, we'd have to a) keep the loop open and b) send on that channel again so the consumer can make sure they're using the "latest" connection.

This just makes me lean harder towards the "remove the loop/allow redialing behavior" – this piece of code/functionality is obviously not easy to get right, and the benefits we are currently getting out of it (none, afaict) do not outweigh the complexity it adds to the code.

cpuguy83:
Why do we treat "redial" as anything different than "a new connection that needs to be handled" and implement the handler per connection?

Benehiko:
>Why do we treat "redial" as anything different than "a new connection that needs to be handled" and implement the handler per connection?

This PR currently implements _redial_ in this way. With this implementation we would have a _redial_ support only when the dial errors for some reason (e.g. timeout). Any _new_ connections won't be able to communicate with the `CLI` since the channel closes and the goroutine returns, which requires a new connection setup.  

Right now the tests and @laurazard is implying `reconnect` means the same connection is being reused.

For example:
The CLI has a listener with a singular connection instance `conn`.
A dial to the CLI - `conn` can now communicate with this connection.
A second dial to the CLI - `conn` is now communicating with second connection, replacing the first and so on.

I'll just remove this behavior and remove the reconnect tests, as @laurazard suggested.

cli-plugins/socket/socket.go
```diff
@@ -16,15 +16,18 @@ const EnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
 // SetupConn sets up a Unix socket listener, establishes a goroutine to handle connections
```
krissetto, line 16: Let's not forget to update these docstrings with the new expectations and return values

cmd/docker/docker.go
```diff
@@ -247,11 +246,18 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
...
+			// thus further loops will not block on connChan
+			case c := <-connChan:
+				if c != nil {
+					if err := c.Close(); err != nil {
+						_, _ = fmt.Fprintf(dockerCli.Err(), "failed to signal plugin to close: %v\n", err)
+					}
 				}
-				conn = nil
+			// timeout here if for some reason the channel is still open and not populated.
+			case <-time.After(1 * time.Millisecond):
```
krissetto, line 259: Are we sure we want to timeout here after 1 ms with no errors at all? I assume you chose this instead of `default` to allow some extra time for the response to arrive on the channel, but what are we expecting to happen in those cases?

cli-plugins/socket/socket.go
```diff
@@ -35,16 +38,26 @@ func randomID() string {
...
+		// this is a blocking call and will wait
+		// until a new connection is accepted
+		// or until the timout is reached
+		conn, _ := listener.AcceptUnix()
+
+		// perform any platform-specific actions on accept (e.g. unlink non-abstract sockets)
+		onAccept(listener)
+		connChan <- conn
+		// close the channel to signal we won't accept any more connections
+		close(connChan)
```
krissetto, line 57: nit: maybe we can `defer` this `close` as the very first line of the goroutine, just in case some panic happens somewhere else and we want to be sure to clean up the resources. WDYT?

cmd/docker/docker.go
Benehiko: We don't throw an error here since the function will continue executing and call the fallback https://github.com/docker/cli/blob/ad778cf743c1876098671035bc183aa39762827d/cmd/docker/docker.go#L265

but you're right, we could also just do a `default` and continue. The timeout is arbitrary since we only execute this part of the code when the user terminates the CLI, by that time the channel should be populated and if not, just gracefully continue.

cli-plugins/socket/socket.go
Benehiko: yeah that makes sense to me

cpuguy83:
@Benehiko My point is it is not behaving this way.
Even before this change it is treating a redial as some special condition.
But this should be handled not much differently than an http handler.

krissetto (approved):
For now LGTM. If you want we can create a tracking issue for making the general behavior more http handler-like as @cpuguy83 was mentioning

cmd/docker/docker.go
neersighted: Let's prefer `default:` here as that is the idiomatic pattern for non-blocking channel reads.

neersighted (changes requested):
One nit, overall LGTM

Benehiko:
Closing this in favor of https://github.com/docker/cli/pull/4905

PR #4884, opened 2024-02-21
docs: regenerate markdown

thaJeztah:
**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah:
I think something is broken .. somewhere. Trying to re-generate docs from the master branch (`make -f docker.Makefile mdddocs`) didn't show me a diff, but I noticed I _did_ get updates when rebasing https://github.com/docker/cli/pull/4228

docs/reference/commandline/attach.md
```diff
@@ -13,7 +13,7 @@ Attach local standard input, output, and error streams to a running container
 |:----------------|:---------|:--------|:----------------------------------------------------|
 | `--detach-keys` | `string` |         | Override the key sequence for detaching a container |
 | `--no-stdin`    |          |         | Do not attach STDIN                                 |
```
crazy-max, line 15: Wonder if we should always display `bool` type even if it defaults to `false`? Would need changes in cli-docs-tool

thaJeztah: good one, yes. Looking at the `string` above, effectively that's  the same;

- always show type
- but don't show default if there's no default (or if default == type's empty value??)


(last part doesn't look consistent there though, because `--cpu-shares` shows `0` as default, which is also the default for an `int64` - wondering why that is 🤔

dvdksn: we purposefully hide zero values when rendering the pages on docs.docker.com atm
https://github.com/docker/docs/blob/main/layouts/_default/cli.html#L118

thaJeztah: Yes, so funny thing is that there's some "special code" for membytes, which describes we did so for Cobra to hide it 🤔 https://github.com/docker/cli/blob/f6b7a11b931bc4371418b617fe1d356b632e66ea/opts/opts.go#L446-L450

thaJeztah:
Ah! We did! There's two separate `CLI_DOCS_TOOL_VERSION` vars, and we only updated one 🤦 

https://github.com/docker/cli/blob/f6b7a11b931bc4371418b617fe1d356b632e66ea/scripts/docs/generate-md.sh#L5
https://github.com/docker/cli/blob/f6b7a11b931bc4371418b617fe1d356b632e66ea/scripts/docs/generate-yaml.sh#L5

dvdksn (approved):
LGTM

docs/reference/commandline/attach.md
crazy-max: Opened https://github.com/docker/cli-docs-tool/pull/50 to display bool type

PR #4888, opened 2024-02-21
fix: cli prompt termination exit code

### LINKED ISSUE #4850: canceling "docker volume prune" does not consistently print a newline, and exits with non-zero status
### Description

When terminating `docker volume prune` using `CTRL-C`, the CLI terminates, but no newline is printed at the end, causing the shell to continue on the same line. It also exits with a non-zero exit code (255). Behavior is inconsistent between shells though (perhaps some shells compensate for the missing newline, and add them implicitly?).

### Reproduce

On Bash on macOS (very old one shipping with macOS); 

```bash
bash-3.2$ docker volume prune
Are you sure you want to continue? [y/N] ^Cbash-3.2$ echo $?
255
bash-3.2$
```

Using `N` (don't prune) or `ENTER` (default: also `N`), the CLI terminates cleanly with a zero exit-code (but also prints a message, which likely shouldn't be printed :sweat_smile:);

```bash
bash-3.2$ docker volume prune
Are you sure you want to continue? [y/N] N
Total reclaimed space: 0B
bash-3.2$ echo $?
0
```

Interestingly, on a more current version of Bash () on Linux, a newline _is_ printed, but the exit-code is also non-zero, but _different_ (`130`);

```bash
root@debian-s-1vcpu-1gb-ams3-01:~# docker volume prune
Are you sure you want to continue? [y/N] ^C
root@debian-s-1vcpu-1gb-ams3-01:~# echo $?
130

root@debian-s-1vcpu-1gb-ams3-01:~# bash --version
GNU bash, version 5.2.15(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

On zsh (default now on macOS), a newline is printed, but also exits with a non-zero exit code;

```zsh
thajeztah@m1book ~ % docker volume prune
Are you sure you want to continue? [y/N] ^C%
thajeztah@m1book ~ % echo $?
255
thajeztah@m1book ~ %
```

And same for Fish;

```fish
 ~  docker volume prune
Are you sure you want to continue? [y/N] ^C⏎
 !  ~  echo $status
255
 ~ 
```

### Expected behavior

I expect a newline to be printed, and the command to exit with a zero exit status.

### docker version

```bash
Client:
 Cloud integration: v1.0.35+desktop.10
 Version:           25.0.1
 API version:       1.44
 Go version:        go1.21.6
 Git commit:        29cf629
 Built:             Tue Jan 23 23:06:12 2024
 OS/Arch:           darwin/arm64
 Context:           desktop-linux

Server: Docker Desktop 4.28.0 (135377)
 Engine:
  Version:          25.0.1
  Go version:       go1.21.6
  Git commit:       71fa3ab079ec13d17257f86fa92db8d7f24802f1
  Built:            Wed Jan 24 09:49:39 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.27
  GitCommit:        a1496014c916f9e62104b33d1bb5bd03b0858e59
 runc:
  Version:          1.1.11
  GitCommit:        v1.1.11-0-g4bccb38
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
not relevant
```

### Additional Info

_No response_

Benehiko:
Fixes #4850 

```shell
⋊> ~/g/s/g/d/cli on fix-prompt-termination ⨯ ./build/docker volume prune      16:35:17
Are you sure you want to continue? [y/N] ^C
⋊> ~/g/s/g/d/cli on fix-prompt-termination ⨯ echo $status                     16:35:21
0
⋊> ~/g/s/g/d/cli on fix-prompt-termination ⨯ bash                             16:35:23
[benehiko@benehiko cli]$ ./build/docker volume prune
Are you sure you want to continue? [y/N] ^C
[benehiko@benehiko cli]$ echo $?
0
⋊> ~/g/s/g/d/cli on fix-prompt-termination ⨯ zsh                              16:35:32
benehiko% ./build/docker volume prune
Are you sure you want to continue? [y/N] ^C
benehiko% echo $?
0
```
**- What I did**

The docker `command/utils.go` provides a function for prompting the user for a confirmation `[Y/n]`. To fix the termination of the CLI while the prompt is active, the reader is blocked inside a gouroutine instead allowing for checks on context cancellation from the parent or from a SIGINT/SIGTERM. An error is then propagated up the stack when the prompt is terminated instead of an immediate program termination. The caller should respect the error and either close the reader or exit the CLI to prevent goroutine leaks and unexpected behavior on re-prompting.

**- How I did it**

Catch the termination signal on prompt. 

**- How to verify it**

Manually and with tests.

CLI confirmation prompts will now safely cancel the operation on termination (SIGINT/SIGTERM) producing an exit code of 0.

These are the affected CLI commands:
- `docker system prune`
- `docker volume prune`
- `docker plugin upgrade`
- `docker image prune`
- `docker network prune`
- `docker container prune`
- `docker network rm`
- `docker builder prune`
- `docker trust revoke`

Some commands above are replaced by plugins, which will not be affected.

image
(https://www.rawpixel.com/image/6041428/photo-image-background-public-domain-cat)

### DIFF
--- a/cli/command/plugin/install.go
+++ b/cli/command/plugin/install.go
@@ -142,6 +142,7 @@ func acceptPrivileges(dockerCli command.Cli, name string) func(privileges types.
 		for _, privilege := range privileges {
 			fmt.Fprintf(dockerCli.Out(), " - %s: %v\n", privilege.Name, privilege.Value)
 		}
-		return command.PromptForConfirmation(dockerCli.In(), dockerCli.Out(), "Do you grant the above permissions?"), nil
+		ctx := context.TODO()
+		return command.PromptForConfirmation(ctx, dockerCli.In(), dockerCli.Out(), "Do you grant the above permissions?")
 	}
 }
--- a/cli/command/plugin/testdata/plugin-upgrade-terminate.golden
+++ b/cli/command/plugin/testdata/plugin-upgrade-terminate.golden
@@ -1,0 +1,2 @@
+Upgrading plugin foo/bar from localhost:5000/foo/bar:v0.1.0 to localhost:5000/foo/bar:v1.0.0
+Plugin images do not match, are you sure? [y/N] 
--- a/cli/command/plugin/upgrade_test.go
+++ b/cli/command/plugin/upgrade_test.go
@@ -1,0 +1,42 @@
+package plugin
+
+import (
+	"context"
+	"io"
+	"testing"
+
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/internal/test"
+	"github.com/docker/docker/api/types"
+	"github.com/pkg/errors"
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/golden"
+)
+
+func TestUpgradePromptTermination(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+
+	cli := test.NewFakeCli(&fakeClient{
+		pluginUpgradeFunc: func(name string, options types.PluginInstallOptions) (io.ReadCloser, error) {
+			return nil, errors.New("should not be called")
+		},
+		pluginInspectFunc: func(name string) (*types.Plugin, []byte, error) {
+			return &types.Plugin{
+				ID:              "5724e2c8652da337ab2eedd19fc6fc0ec908e4bd907c7421bf6a8dfc70c4c078",
+				Name:            "foo/bar",
+				Enabled:         false,
+				PluginReference: "localhost:5000/foo/bar:v0.1.0",
+			}, []byte{}, nil
+		},
+	})
+	cmd := newUpgradeCommand(cli)
+	// need to set a remote address that does not match the plugin
+	// reference sent by the `pluginInspectFunc`
+	cmd.SetArgs([]string{"foo/bar", "localhost:5000/foo/bar:v1.0.0"})
+	test.TerminatePrompt(ctx, t, cmd, cli, func(t *testing.T, err error) {
+		t.Helper()
+		assert.ErrorIs(t, err, command.ErrPromptTerminated)
+	})
+	golden.Assert(t, cli.OutBuffer().String(), "plugin-upgrade-terminate.golden")
+}
--- a/cli/command/trust/testdata/trust-revoke-prompt-termination.golden
+++ b/cli/command/trust/testdata/trust-revoke-prompt-termination.golden
@@ -1,0 +1,2 @@
+Please confirm you would like to delete all signature data for example/trust-demo? [y/N] 
+Aborting action.
--- a/cli/command/utils.go
+++ b/cli/command/utils.go
@@ -5,12 +5,15 @@ package command
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"io"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 
 	"github.com/docker/cli/cli/streams"
 	"github.com/docker/docker/api/types/filters"
@@ -72,12 +75,21 @@ func PrettyPrint(i any) string {
 	}
 }
 
-// PromptForConfirmation requests and checks confirmation from user.
-// This will display the provided message followed by ' [y/N] '. If
-// the user input 'y' or 'Y' it returns true other false.  If no
-// message is provided "Are you sure you want to proceed? [y/N] "
-// will be used instead.
-func PromptForConfirmation(ins io.Reader, outs io.Writer, message string) bool {
+type PromptError error
+
+var ErrPromptTerminated = PromptError(errors.New("prompt terminated"))
+
+// PromptForConfirmation requests and checks confirmation from the user.
+// This will display the provided message followed by ' [y/N] '. If the user
+// input 'y' or 'Y' it returns true otherwise false. If no message is provided,
+// "Are you sure you want to proceed? [y/N] " will be used instead.
+//
+// If the user terminates the CLI with SIGINT or SIGTERM while the prompt is
+// active, the prompt will return false with an ErrPromptTerminated error.
+// When the prompt returns an error, the caller should propagate the error up
+// the stack and close the io.Reader used for the prompt which will prevent the
+// background goroutine from blocking indefinitely.
+func PromptForConfirmation(ctx context.Context, ins io.Reader, outs io.Writer, message string) (bool, error) {
 	if message == "" {
 		message = "Are you sure you want to proceed?"
 	}
@@ -90,9 +102,31 @@ func PromptForConfirmation(ins io.Reader, outs io.Writer, message string) bool {
 		ins = streams.NewIn(os.Stdin)
 	}
 
-	reader := bufio.NewReader(ins)
-	answer, _, _ := reader.ReadLine()
-	return strings.ToLower(string(answer)) == "y"
+	result := make(chan bool)
+
+	// Catch the termination signal and exit the prompt gracefully.
+	// The caller is responsible for properly handling the termination.
+	notifyCtx, notifyCancel := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
+	defer notifyCancel()
+
+	go func() {
+		var res bool
+		scanner := bufio.NewScanner(ins)
+		if scanner.Scan() {
+			answer := strings.TrimSpace(scanner.Text())
+			if strings.EqualFold(answer, "y") {
+				res = true
+			}
+		}
+		result <- res
+	}()
+
+	select {
+	case <-notifyCtx.Done():
+		return false, ErrPromptTerminated
+	case r := <-result:
+		return r, nil
+	}
 }
 
 // PruneFilters returns consolidated prune filters obtained from config.json and cli
--- a/cli/command/utils_test.go
+++ b/cli/command/utils_test.go
@@ -1,36 +1,46 @@
-package command
+package command_test
 
 import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"io"
 	"os"
 	"path/filepath"
+	"strings"
+	"syscall"
 	"testing"
+	"time"
 
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/internal/test"
 	"github.com/pkg/errors"
 	"gotest.tools/v3/assert"
 )
 
 func TestStringSliceReplaceAt(t *testing.T) {
-	out, ok := StringSliceReplaceAt([]string{"abc", "foo", "bar", "bax"}, []string{"foo", "bar"}, []string{"baz"}, -1)
+	out, ok := command.StringSliceReplaceAt([]string{"abc", "foo", "bar", "bax"}, []string{"foo", "bar"}, []string{"baz"}, -1)
 	assert.Assert(t, ok)
 	assert.DeepEqual(t, []string{"abc", "baz", "bax"}, out)
 
-	out, ok = StringSliceReplaceAt([]string{"foo"}, []string{"foo", "bar"}, []string{"baz"}, -1)
+	out, ok = command.StringSliceReplaceAt([]string{"foo"}, []string{"foo", "bar"}, []string{"baz"}, -1)
 	assert.Assert(t, !ok)
 	assert.DeepEqual(t, []string{"foo"}, out)
 
-	out, ok = StringSliceReplaceAt([]string{"abc", "foo", "bar", "bax"}, []string{"foo", "bar"}, []string{"baz"}, 0)
+	out, ok = command.StringSliceReplaceAt([]string{"abc", "foo", "bar", "bax"}, []string{"foo", "bar"}, []string{"baz"}, 0)
 	assert.Assert(t, !ok)
 	assert.DeepEqual(t, []string{"abc", "foo", "bar", "bax"}, out)
 
-	out, ok = StringSliceReplaceAt([]string{"foo", "bar", "bax"}, []string{"foo", "bar"}, []string{"baz"}, 0)
+	out, ok = command.StringSliceReplaceAt([]string{"foo", "bar", "bax"}, []string{"foo", "bar"}, []string{"baz"}, 0)
 	assert.Assert(t, ok)
 	assert.DeepEqual(t, []string{"baz", "bax"}, out)
 
-	out, ok = StringSliceReplaceAt([]string{"abc", "foo", "bar", "baz"}, []string{"foo", "bar"}, nil, -1)
+	out, ok = command.StringSliceReplaceAt([]string{"abc", "foo", "bar", "baz"}, []string{"foo", "bar"}, nil, -1)
 	assert.Assert(t, ok)
 	assert.DeepEqual(t, []string{"abc", "baz"}, out)
 
-	out, ok = StringSliceReplaceAt([]string{"foo"}, nil, []string{"baz"}, -1)
+	out, ok = command.StringSliceReplaceAt([]string{"foo"}, nil, []string{"baz"}, -1)
 	assert.Assert(t, !ok)
 	assert.DeepEqual(t, []string{"foo"}, out)
 }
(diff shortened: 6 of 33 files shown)


cli/command/filename
thaJeztah, line 1: Accidental commit, or are we using this one for a test?

Benehiko: oh... how did that get there :thinking: 
I'll remove it :)

cli/command/plugin/upgrade_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	syscall.Kill(syscall.Getpid(), syscall.SIGINT)
+	// sigint and sigterm are caught by the prompt
+	// this allows us to gracefully exit the prompt with a 0 exit code
+	select {
+	case err := <-errChan:
+		assert.Error(t, err, "canceling upgrade request")
+	case <-time.After(100 * time.Millisecond):
+		t.Fatal("Plugin Upgrade did not return after SIGINT")
+	}
```
thaJeztah, line 50: Looks like this chunk of code is repeated in various tests. I haven't looked closely, but wondering if this would be a candidate for factoring out into an (`internal/`?) test-utility.

Benehiko: Yes it's somewhat repetitive since most of the functions follow the same pattern. I will move this to minimize repetition

thaJeztah: It looks like we have an `internal/` package already, that also has some test-utilities (including, I see, a fakeCLI as well, so _perhaps_ even the fakeCLI parts could be integrated there, but I didn't look closely!); https://github.com/docker/cli/blob/master/internal/test/cli.go

cli/command/utils.go
```diff
@@ -90,9 +95,35 @@ func PromptForConfirmation(ins io.Reader, outs io.Writer, message string) bool {
 		ins = streams.NewIn(os.Stdin)
 	}
 
-	reader := bufio.NewReader(ins)
-	answer, _, _ := reader.ReadLine()
-	return strings.ToLower(string(answer)) == "y"
+	result := make(chan bool, 1)
+	// We use a context to cancel the prompt or the signal handler
+	ctx, cancel := context.WithCancel(context.TODO())
```
Benehiko, line 100: using context.TODO here since we would probably want to use a context received form the `PromptForConfirmation` function parameter.

cli/command/utils_test.go
```diff
@@ -1,36 +1,45 @@
-package command
+package command_test
```
Benehiko, line 1: had cyclical import between the `command` package and `test`.

cli/command/utils.go
```diff
@@ -90,9 +95,35 @@ func PromptForConfirmation(ins io.Reader, outs io.Writer, message string) bool {
...
+
+	go func(_ context.Context) {
+		// cleanup the other goroutine
+		defer cancel()
+		reader := bufio.NewReader(ins)
+		answer, _, _ := reader.ReadLine()
+		result <- (strings.ToLower(string(answer)) == "y")
+	}(ctx)
```
Benehiko, line 108: Just did a test using the https://github.com/uber-go/goleak package on a toy application.

Found that it does not leak at all, presumable since the `stdin` should be closed by the time the program exits. 

```go
func TestLeakyReader(t *testing.T) {
	defer goleak.VerifyNone(t)

	result := make(chan bool)
	ctx, cancel := context.WithCancel(context.TODO())

	reader := bufio.NewReader(os.Stdin)
	go func() {
		// cleanup the other goroutine
		defer cancel()
		answer, _, _ := reader.ReadLine()
		result <- (strings.ToLower(string(answer)) == "y")
	}()

	go func() {
		// cleanup the other goroutine
		defer cancel()
		// We catch the signal to avoid the prompt
		// from exiting as an error
		c := make(chan os.Signal, 1)
		signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
		select {
		case <-ctx.Done():
			return
		case <-c:
			result <- false
		}
	}()

	fmt.Printf("%t\n", <-result)
}
```

Adding another goroutine which sleeps causes there to be a leak (correctly so).
```go
go func() {
		time.Sleep(time.Minute * 1)
	}()
```

```shell
|| --- FAIL: TestLeakyReader (0.45s)
reader_test.go|51| found unexpected goroutines:
||         [Goroutine 20 in state sleep, with time.Sleep on top of the stack:
```

Benehiko: I assume this would however be a problem if the program does not exit after a prompt. 

For example, prompting a user to confirm, returning false, continue execution and then prompting the user again to confirm.

cli/command/plugin/install.go
```diff
@@ -142,6 +142,7 @@ func acceptPrivileges(dockerCli command.Cli, name string) func(privileges types.
 		for _, privilege := range privileges {
 			fmt.Fprintf(dockerCli.Out(), " - %s: %v\n", privilege.Name, privilege.Value)
 		}
-		return command.PromptForConfirmation(dockerCli.In(), dockerCli.Out(), "Do you grant the above permissions?"), nil
+		ctx := context.TODO()
```
Benehiko, line 145: here the context isn't apparent. Either we use the top level dockerCli context or create a new context within the function. The moby project does not support passing context on this function call https://github.com/moby/moby/blob/master/api/types/client.go#L292

cli/command/trust/testdata/trust-revoke-prompt-termination.golden
```diff
@@ -0,0 +1,2 @@
+Please confirm you would like to delete all signature data for example/trust-demo? [y/N] 
+Aborting action.
```
krissetto, line 2: aah i guess this could be another confirmation prompt to "fix" by standardizing the output (aka not printing "Aborting action" and making it behave more like everything else). For another moment of course

cli/command/utils_test.go
```diff
@@ -68,3 +76,35 @@ func TestValidateOutputPath(t *testing.T) {
...
+		in.Close()
+	})
+
+	result := make(chan bool, 1)
+	go func() {
+		defer close(result)
+		result <- command.PromptForConfirmation(ctx, in, w, "")
+	}()
+
+	time.Sleep(100 * time.Millisecond)
```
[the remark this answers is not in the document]
Benehiko: fixed with latest commit

cli/command/trust/testdata/trust-revoke-prompt-termination.golden
Benehiko: ah yeah true

internal/test/cmd.go
```diff
@@ -0,0 +1,75 @@
...
+	defer writeCancel()
+
+	// wait for the prompt to be ready
+	select {
+	case <-writeCtx.Done():
+		t.Fatalf("command %s did not write prompt to stdout", cmd.Name())
+	case <-writerHookChan:
+		// drain the channel for future buffer writes
+		go func() {
+			for range writerHookChan {
```
Benehiko, line 48: there might be a better way to ensure the writer can still write without waiting for the hook every time. 

although this won't pose any problems and will also exit once the `writerHookChan` channel is closed.

cli/command/utils.go
```diff
@@ -90,9 +95,40 @@ func PromptForConfirmation(ins io.Reader, outs io.Writer, message string) bool {
...
+	// We catch the termination signal
+	// and exit the prompt gracefully
+	// it is up to the command to handle the termination
+	notifyCtx, notifyCancel := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
+	defer notifyCancel()
+
+	go func() {
+		defer notifyCancel()
+		scanner := bufio.NewScanner(ins)
+		for scanner.Scan() {
```
krissetto, line 110: 2 things here:

- We don't need a for loop here, since in both of the `select` cases we return regardless.
- `scanner.Scan()` is still a blocking call, so if the user presses CTRL-C without sending any other input, this goroutine should still be blocked on the Scan(). If we ever have multiple confirmation prompts, the first prompt's goroutine will still be running, it will capture user input from the second confirmation prompt, then it will panic trying to send the result down the closed `result` channel

IMHO this second point warrants some test cases, where we basically just call this function a few times in a row with the various user input possibilities (the permutations of the `y/Y/n/N/CTRL-C` options available). It could become confusing for the user if we use multiple confirmations and the CLI doesn't behave as expected

cli/command/utils.go
```diff
@@ -77,7 +80,9 @@ func PrettyPrint(i any) string {
 // the user input 'y' or 'Y' it returns true other false.  If no
 // message is provided "Are you sure you want to proceed? [y/N] "
 // will be used instead.
-func PromptForConfirmation(ins io.Reader, outs io.Writer, message string) bool {
+// If the user terminates the CLI with SIGINT or SIGTERM while
+// the prompt is active, the prompt will return false.
+func PromptForConfirmation(ctx context.Context, ins io.Reader, outs io.Writer, message string) bool {
```
krissetto, line 85: Generally speaking, if the user inputs something that's not y/Y/n/N or CTRL-C, do we want to assume the answer was negative, or re-prompt the user?

Benehiko: I honestly think it makes sense to just fallback to accepting the value as negative since sensitive commands are usually run with a prompt. Similar to always deny in security. 

Re-prompting can be discussed, but I don't think it adds much value since an incorrect entry assumes you made a mistake (or fell asleep on your keyboard? :) ), so exiting the program makes total sense. Most of the time the command can quickly be rerun with the same command. 

Maybe with otel we might see more insight into failure rates of such prompts and then discuss this further?

cli/command/plugin/install.go
thaJeztah: FWIW; if we would benefit from passing contexts, we should consider updating the moby code. For a lot of code in moby, `context` didn't exist yet when it was written, and we started passing through contexts more, but there's still many areas where we didn't.

Benehiko: I'll take a look at also adding context in the moby api client. It will help greatly with otel and just task cleanup in general.

cli/command/plugin/testdata/plugin-upgrade-terminate.golden
```diff
@@ -0,0 +1,2 @@
+Upgrading plugin foo/bar from localhost:5000/foo/bar:v0.1.0 to localhost:5000/foo/bar:v1.0.0
+Plugin images do not match, are you sure? [y/N]
```
laurazard, line 2: Missing a newline at the end here? (The other new golden files have them 🤔)

internal/test/writer.go
```diff
@@ -0,0 +1,27 @@
+package test
+
+import (
+	"io"
+)
+
+// WriterWithHook is an io.Writer that calls a hook function after every write.
+// This is useful for testing.
```
laurazard, line 8: This description is a bit redundant. Why is it useful?

internal/test/writer.go
```diff
@@ -0,0 +1,27 @@
...
+func (w *WriterWithHook) Write(p []byte) (n int, err error) {
+	defer w.hook(p)
+	return w.actualWriter.Write(p)
+}
+
+var _ io.Writer = (*WriterWithHook)(nil)
+
+// NewWriterWithHook returns a new WriterWithHook that still writes to the actualWriter
+// but also calls the hook function after every write.
+// The hook function is useful for testing, or waiting for a write to complete.
```
laurazard, line 24: And then this one is basically the same as the one above, but with a bit more information. Maybe we can move some things around?

cli/command/utils.go
```diff
@@ -72,12 +75,21 @@ func PrettyPrint(i any) string {
...
-// the user input 'y' or 'Y' it returns true other false.  If no
+// the user input 'y' or 'Y' it returns true otherwise false.  If no
 // message is provided "Are you sure you want to proceed? [y/N] "
 // will be used instead.
-func PromptForConfirmation(ins io.Reader, outs io.Writer, message string) bool {
+// If the user terminates the CLI with SIGINT or SIGTERM while
+// the prompt is active, the prompt will return false with an ErrPromptTerminated error.
+// When the prompt returns an error, the caller should propagate the error up the stack
+// ultimately closing the io.Reader used for the prompt which will prevent the background
+// goroutine from blocking indefinitely.
```
laurazard, line 91: Maybe we can clean up this comment a bit, and also format better with the above? Something like
```go
// PromptForConfirmation requests and checks confirmation from the user.
// This will display the provided message followed by ' [y/N] '. If the user input
// 'y' or 'Y' it returns true otherwise false.  If no message is provided, "Are you
// sure you want to proceed? [y/N] " will be used instead.
//
// If the user terminates the CLI with SIGINT or SIGTERM while the prompt is active,
// the prompt will return false with an ErrPromptTerminated error. When the prompt
// returns an error, the caller should propagate the error up the stack and close the
// io.Reader used for the prompt which will prevent the background goroutine from
// blocking indefinitely.
```

cli/command/utils.go
```diff
@@ -90,9 +102,32 @@ func PromptForConfirmation(ins io.Reader, outs io.Writer, message string) bool {
...
 	}
 
-	reader := bufio.NewReader(ins)
-	answer, _, _ := reader.ReadLine()
-	return strings.ToLower(string(answer)) == "y"
+	result := make(chan bool)
+
+	// We catch the termination signal
+	// and exit the prompt gracefully
+	// it is up to the command to handle the termination
```
laurazard, line 109:
```suggestion
        // Catch the termination signal and exit the prompt gracefully.
	// The calling command is responsible for properly handling the termination.
```

laurazard (approved):
Really nice, thank you! Left a few comments (mostly nits), but overall a good change.

Not sure if I missed it, but are we explicitly testing the exit code anywhere?

krissetto (approved):

internal/test/writer.go
Benehiko: I'll update it

Benehiko:
> Really nice, thank you! Left a few comments (mostly nits), but overall a good change.
> 
> Not sure if I missed it, but are we explicitly testing the exit code anywhere?

Hmm, I suppose we could check the status code, but the status code seems to be added to a specific error type. It seems the main package handles some exit codes based on the error type here https://github.com/docker/cli/blob/master/cmd/docker/docker.go#L35-L50

Since we gracefully propagate the error up we check for it inside the test, which I think should be enough. Wdyt?

laurazard:
imo those tests are useful, but if we want to preserve the "exit code is 0 when user CTRL-C's a confirmation prompt" in the future, we could add this check in particular for the e2e tests for these commands. (But fine if you'd prefer to do it in another PR).

Benehiko:
Seems like this test is quite flaky
```shell
56.34 === FAIL: cli-plugins/socket TestConnectAndWait/connect_goroutine_exits_after_EOF 
```

neersighted (approved):

cli/command/plugin/testdata/plugin-upgrade-terminate.golden
laurazard: @Benehiko could we fix this?

internal/test/writer.go
```diff
@@ -0,0 +1,30 @@
...
+func NewWriterWithHook(actualWriter io.Writer, hook func([]byte)) *WriterWithHook {
```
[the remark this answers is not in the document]
laurazard: Anyway, can address this (and e2e tests) in a follow-up (cc @Benehiko)

cli/command/plugin/testdata/plugin-upgrade-terminate.golden
laurazard: Right, this is likely due to us missing a newline when printing things somewhere

cli/command/utils.go
```diff
@@ -90,9 +102,31 @@ func PromptForConfirmation(ins io.Reader, outs io.Writer, message string) bool {
...
 	}
 
-	reader := bufio.NewReader(ins)
-	answer, _, _ := reader.ReadLine()
-	return strings.ToLower(string(answer)) == "y"
+	result := make(chan bool)
+
+	// Catch the termination signal and exit the prompt gracefully.
+	// The caller is responsible for properly handling the termination.
+	notifyCtx, notifyCancel := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
```
tonistiigi, line 109: I'm not sure a utility function should set up signal handlers for the process as a side effect. I think this should just use the passed in context for cancellation, and if that context is tied to the os signals is up to the caller. Normally, a process already has a signal handler and this will conflict with it.

PR #4889, opened 2024-02-21
cli: add otel sdk tracing and metric providers to the core cli

jsternberg:
**- What I did**

This adds the code used by buildx and compose into the default CLI program
to help normalize the usage of these APIs and allow code reuse between
projects. It also allows these projects to benefit from improvements or
changes that may be made by another team.

At the moment, these APIs are a pretty thin layer on the OTEL SDK. It 
configures an additional exporter to a docker endpoint that's used for 
usage collection and is only active if the option is configured in docker
desktop.

This also upgrades the OTEL version to v1.19 which is the one being used by
buildkit, buildx, compose, etc.

**- How I did it**

I added a new interface to the DockerCli and implemented that interface using
code that is mostly identical to the same code in buildx which was adapted from
compose. It's missing the user-facing OTEL environment components that buildx
has though.

**- How to verify it**

Not done yet.

Include OTEL tracing and metric utilities for docker cli programs


jsternberg:
I've created this as a draft, but the code itself is mostly complete and could be merged as-is.

The code in this PR is already part of buildx and there's a version this was based off of in compose for the telemetry. I thought it might be a good idea to start working on integrating these two so buildx and compose could be using the same code and could benefit from the same changes.

@milas I'm really interested in feedback from you on this too since you have worked on the compose side of this.

cli/command/cli.go
```diff
@@ -85,6 +87,9 @@ type DockerCli struct {
 	dockerEndpoint     docker.Endpoint
 	contextStoreConfig store.Config
 	initTimeout        time.Duration
+	res                *resource.Resource
```
krissetto, line 90: nit: From what I see, I think we could group these 3 into a single struct defined in the `telemetry` file, so it's not too tightly coupled with this cli struct and so that it's purpose is a bit clearer (this struct is getting a bit wild imho).

The only part of the telemetry code that seems to need data from the actual DockerCli struct is `dockerExporterOtlpEndpoint` (which only needs the metadata).

Any opinions on this?

krissetto (commented):
In general i like the approach. 

I do have a few thoughts though, I'll leave them here so we can discuss (maybe i'm missing some details):

- I think integrating otel things into the cli like this could cause some issues down the road, since we'd be using otel structs and methods directly even in projects that import the cli to use this common code. Maybe it could be worth it to decouple our usage of telemetry things from our actual implementation using otel, also for keeping things more easily standardized across projects;
- Given this ^, it might be a good idea to separate these bits into a dedicated repo/pkg that only exports the necessary functionality and abstracts otel resources away from the projects using it. Since the CLI is already doing double duty as both an application and a "library" for plugins, I'd try to avoid adding more responsibilities here. Separating things out, we can also collaborate on improvements to the required functionality more easily. Do you see any issues with this?;
- Shall we also include some functionality for adding Attributes, span Events, etc so clients can do so without depending directly on otel pkgs?

cli/command/cli.go
jsternberg: Yea, this can probably be wrapped into its own struct to make it a bit less unwieldy. I'll work on that to see how well it works.

jsternberg:
@krissetto I agree with some of what you're saying on principle, but I do worry a bit about overabstracting in the case of the OTEL libraries.

> I think integrating otel things into the cli like this could cause some issues down the road, since we'd be using otel structs and methods directly even in projects that import the cli to use this common code. Maybe it could be worth it to decouple our usage of telemetry things from our actual implementation using otel, also for keeping things more easily standardized across projects;

I think there's two components of otel that need to be treated separately. There's the library portion, which is mostly just interfaces, and the SDK, which is the implementation of the SDK. For most code, they should use the library portion and we shouldn't really worry about the actual implementation. The purpose of these interfaces is to create this separation and adding our own separation would probably not work very well.

For the SDK, the only part that's really supposed to use the SDK is the outermost section of the program. The part that configures the SDK and passes the `TracerProvider` or `MeterProvider` to the remaining code. I think this area would be the place to consider putting our own wrapper around it, but I suspect the effort to put a wrapper around it will result in code bloat that shows up mostly as a leaky abstraction rather than anything useful.

In particular, there is one area of this PR that's a leaky abstraction when it comes to the SDK.

```go
// TelemetryClient provides the methods for using OTEL tracing or metrics.
type TelemetryClient interface {
	// Resource returns the OTEL Resource configured with this TelemetryClient.
	// This resource may be created lazily, but the resource should be the same
	// each time this function is invoked.
	Resource() *resource.Resource

        // TracerProvider returns a TracerProvider. This TracerProvider will be configured
	// with the default tracing components for a CLI program along with any options given
	// for the SDK.
	TracerProvider(opts ...sdktrace.TracerProviderOption) TracerProvider

	// MeterProvider returns a MeterProvider. This MeterProvider will be configured
	// with the default metric components for a CLI program along with any options given
	// for the SDK.
	MeterProvider(opts ...sdkmetric.Option) MeterProvider
}
```

The options to `TracerProvider` and `MeterProvider` are transparently taking the sdk options and using the sdk versions of these implementations. The returned providers themselves are abstracted behind interfaces.

While I'm not thrilled with this, the alternative is wrapping each option in a separate library. But, these options get fairly involved and you suddenly end up creating an interface for most of the structs in the SDK.

At the end of the day, no application or plugin is _required_ to use the elements provided by the CLI. It's just convenient and takes away the hassle of figuring out how to do specific CLI-specific things with it. And I can't think of a situation where a CLI application that is using tracing or metrics wouldn't just use the official OTEL SDK or provide their own custom implementation of these methods.

> Given this ^, it might be a good idea to separate these bits into a dedicated repo/pkg that only exports the necessary functionality and abstracts otel resources away from the projects using it. Since the CLI is already doing double duty as both an application and a "library" for plugins, I'd try to avoid adding more responsibilities here. Separating things out, we can also collaborate on improvements to the required functionality more easily. Do you see any issues with this?;

Mentioned above, but my primary concern is that the abstraction won't yield better code. The code in this PR is mostly low/no-cost if the functions aren't invoked. If you're integrating tracing or metrics into your docker cli plugin, they're pretty much exactly what you would want with options to replace certain aspects (such as the default resource).

> Shall we also include some functionality for adding Attributes, span Events, etc so clients can do so without depending directly on otel pkgs?

Attributes, span events, and other things are included as part of the OTEL library and not part of the SDK so I don't think there's much value in abstracting these away. The otel libraries themselves should be safe to use directly and are considered stable for tracing and metrics. OTEL logs are mostly an internal abstraction inside of the collector and can mostly be ignored from an application point of view.

cli/command/cli.go
jsternberg: Moved these three attributes into its own struct and modified the infrastructure around it.

I felt like the method implementations of the tracer provider and meter provider were likely still more straightforward when attached to the `DockerCli`, but the resource stuff could clearly be moved to its own separate thing without making things more complicated.

jsternberg:
> My main concern here was with our dependencies, and keeping the various projects all on the same version of the sdk/libs. I generally always try to avoid having many projs on potentially different versions of the same libs (even though SemVer should theoretically give us some guarantees, i've had too many bad experiences 😂).

My personal hope is that, with this in the CLI, it'll mostly standardize the version the various CLI programs use. I do think a follow up thing after that is maybe creating a separate repository for OTEL stuff in general so server and cli components are synced on the versions.

cli/command/telemetry_docker.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	opts := []otlpmetricgrpc.Option{
+		otlpmetricgrpc.WithEndpoint(endpoint),
+	}
+	if !secure {
+		opts = append(opts, otlpmetricgrpc.WithInsecure())
+	}
+
+	exp, err := otlpmetricgrpc.New(context.Background(), opts...)
```
Benehiko, line 116: Shouldn't we accept a context from the function parameter instead of create a new one here?

Looking into the otel code base the `ctx` accepted here gets used inside the `grpc.DialContext(ctx, cfg.Metrics.Endpoint, dialOpts...)` function. 
https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlpmetric/otlpmetricgrpc/client.go#L57

For general cleanup purposes (cancel operation, timeouts etc.), I would prefer if we do not create standalone contexts.

cli/command/telemetry_docker.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	opts := []otlptracegrpc.Option{
+		otlptracegrpc.WithEndpoint(endpoint),
+	}
+	if !secure {
+		opts = append(opts, otlptracegrpc.WithInsecure())
+	}
+
+	exp, err := otlptracegrpc.New(context.Background(), opts...)
```
Benehiko, line 96: same here

cli/command/telemetry_docker.go
jsternberg: I'm not generally the biggest fan of these contexts used for initialization, but I've added them to this function just so we can pass it through. If the user wants to use the background context, they can always pass it in.

Benehiko: I'm not too sure I follow? Wouldn't we want to have the ability to cancel, timeout the underlying otel code when say a `SIGINT` or `SIGTERM` is initiated by the user of the CLI?

Reading the docs of the `grpc.DialContext` function used by otel here, it _could_ be blocking which makes using a context the optimal way to cancel the dial. 

>In the blocking case, ctx can be used to cancel or expire the pending connection. Once this function returns, the cancellation and expiration of ctx will be noop. Users should call ClientConn.Close to terminate all the pending operations after this function returns. 
https://pkg.go.dev/google.golang.org/grpc#DialContext

jsternberg: I took a deeper look at the code and I think this is fine. I didn't realize that the `New` invocation would initiate a dial. I thought it was a lazy invocation and the context was stored for later use.

cli/command/telemetry.go
```diff
@@ -0,0 +1,214 @@
...
+	reader := sdkmetric.NewManualReader(
+		sdkmetric.WithTemporalitySelector(deltaTemporality),
+	)
+	return &cliReader{
+		Reader:   reader,
+		exporter: exp,
+	}
+}
+
+const shutdownTimeout = 2 * time.Second
```
laurazard, line 186: 2 seconds is a while – especially if we're talking about really quick invocations – think `docker exec my-container true` which normally exits somewhere in the range of a few milliseconds. If this would block the CLI from exiting, I'd reduce this significantly.

cli/command/telemetry.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+type serviceNameDetector struct{}
+
+func (serviceNameDetector) Detect(ctx context.Context) (*resource.Resource, error) {
+	return resource.StringDetector(
+		semconv.SchemaURL,
+		semconv.ServiceNameKey,
+		func() (string, error) {
+			return filepath.Base(os.Args[0]), nil
```
laurazard, line 164: Is the idea here to discern between `docker`/main CLI and plugin binaries?

laurazard (commented):
Overall this looks really good to me, thank you @jsternberg! I'd like to get these basic bits merged in quickly so we can start making use of them and moving forward with a proper OTEL implementation on the CLI side.
Just a couple of questions –

cli/command/telemetry.go
jsternberg: A bit. The service name is determined either through `OTEL_SERVICE_NAME` as an environment variable or through the default logic of `unknown_service:${args[0]}` as the name. This changes the default name to be just `args[0]` which usually maps to the name of the binary.

I'm still trying to determine what the service name is most useful for, but I've found it helpful for visual displays. I don't think it's useful for logic like "did this metric come from buildx?" because the user can overwrite the service name by using `OTEL_SERVICE_NAME`. At the moment, I've been using the instrumentation scope which maps to the tracer or meter name. I'm not sure if that will continue to work into the future.

This code specifically was copied from buildkit.

cli/command/telemetry_docker.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	opts := []otlpmetricgrpc.Option{
+		otlpmetricgrpc.WithEndpoint(endpoint),
+	}
+	if !secure {
+		opts = append(opts, otlpmetricgrpc.WithInsecure())
+	}
+
+	exp, err := otlpmetricgrpc.New(ctx, opts...)
```
krissetto, line 116: If the `endpoint == ""`, what do we expect to happen here? 
Doing some tests i'm noticing the error returned is `failed to exit idle mode: passthrough: received empty target in Build()`

krissetto: This also causes the exporter itself to be `nil`, which will then cause `Shutdown()` to panic. Not sure if i'm missing something here

jsternberg: Definitely a mistake. The buildx variant of this code checks for the endpoint being empty but it seems I forgot to do that when I translated it here.

jsternberg: This should now be fixed.

cli/command/telemetry.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (cli *DockerCli) TracerProvider(ctx context.Context, opts ...sdktrace.TracerProviderOption) TracerProvider {
+	allOpts := make([]sdktrace.TracerProviderOption, 0, len(opts)+2)
+	allOpts = append(allOpts, sdktrace.WithResource(cli.Resource()))
+	if exp, err := dockerSpanExporter(ctx, cli); err != nil {
+		otel.Handle(err)
+	} else if exp != nil {
+		allOpts = append(allOpts, sdktrace.WithBatcher(exp))
+	}
+	allOpts = append(allOpts, opts...)
```
Benehiko, line 65:
```suggestion
	allOpts := append(opts, sdktrace.WithResource(cli.Resource()))
	if exp, err := dockerSpanExporter(ctx, cli); err != nil {
		otel.Handle(err)
	} else if exp != nil {
		allOpts = append(allOpts, sdktrace.WithBatcher(exp))
	}
	return sdktrace.NewTracerProvider(allOpts...)
```

cli/command/telemetry.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (cli *DockerCli) MeterProvider(ctx context.Context, opts ...sdkmetric.Option) MeterProvider {
+	allOpts := make([]sdkmetric.Option, 0, len(opts)+2)
+	allOpts = append(allOpts, sdkmetric.WithResource(cli.Resource()))
+	if exp, err := dockerMetricExporter(ctx, cli); err != nil {
+		otel.Handle(err)
+	} else if exp != nil {
+		allOpts = append(allOpts, sdkmetric.WithReader(newCLIReader(exp)))
+	}
+	allOpts = append(allOpts, opts...)
```
Benehiko, line 77:
```suggestion
	allOpts := append(opts, sdkmetric.WithResource(cli.Resource()))
	if exp, err := dockerMetricExporter(ctx, cli); err != nil {
		otel.Handle(err)
	} else if exp != nil {
		allOpts = append(allOpts, sdkmetric.WithReader(newCLIReader(exp)))
	}
```

cli/command/telemetry_docker.go
```diff
@@ -0,0 +1,121 @@
...
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+const otelConfigFieldName = "otel"
+
+// dockerExporterOtlpEndpoint retrieves the OTLP endpoint used for the docker reporter
+// from the current context.
+func dockerExporterOtlpEndpoint(cli Cli) (endpoint string, secure bool, err error) {
```
Benehiko, line 20:
```suggestion
func dockerExporterOTLPEndpoint(cli Cli) (endpoint string, secure bool, err error) {
```

In go acronyms are uppercase.

cli/command/telemetry.go
```diff
@@ -0,0 +1,214 @@
...
+func (r *telemetryResource) init() {
+	if r.res != nil {
+		r.opts = nil
+		return
+	}
+
+	opts := r.defaultOptions()
+	if len(r.opts) > 0 {
+		opts = append(opts, r.opts...)
+	}
```
Benehiko, line 122:
```suggestion
	opts := append(r.defaultOptions(), r.opts...)
```

even though `r.opts` might be nil - this is fine, the underlying type is a slice of `resource.Options` and append will deal with it just fine.

Benehiko (changes requested):

cli/command/telemetry.go
jsternberg: Maybe this isn't a concern, but the idea behind the original ordering is that user-provided options would occur later in the ordering than the default ones. For some options, in particular for resource, the last option will win and I didn't want an explicitly set resource to be overridden by something that was the default.

It's possible to override the default resource in a different way, but I thought it would be good to keep this pattern in case another default is needed in the future.

cli/command/telemetry_docker.go
jsternberg: Done.

cli/command/telemetry.go
jsternberg: Done.

cli/command/telemetry_docker.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, err
+	}
+	return exp, nil
+}
+
+func dockerMetricExporter(ctx context.Context, cli Cli) (sdkmetric.Exporter, error) {
+	endpoint, secure, err := dockerExporterOTLPEndpoint(cli)
+	if endpoint == "" || err != nil {
+		return nil, err
+	}
```
Benehiko, line 107: I'm not sure about this. Wouldn't we want to first check the error, then return it and on another if check the endpoint as non-empty then return a custom error?
```suggestion
	endpoint, secure, err := dockerExporterOTLPEndpoint(cli)
	if err != nil {
		return nil, err
	}
	if endpoint == "" {
	        return nil, errors.New("endpoint is empty")
	}
```

cli/command/telemetry.go
Benehiko: It's not a huge concern to me, just a code nit :)
We can keep it as is

cli/command/telemetry.go
laurazard: So my knowledge about this is super limited, but here (in the CLI process) we'd be sending stuff to the collector (which generally is going to be running on a local machine), right? Not sending stuff over the internet? So I would think it would be way faster than 200ms generally, but I could be wrong.

krissetto: The collector should be running in DD, so I'd also _expect_ it to be faster than 200ms. 
But TBH I haven't looked at any of the collector impls so far

jsternberg: It should be much faster than 200ms as it would be running locally. I'd expect a couple of milliseconds at most, but I mostly want to set the timeout to whatever would be the most permissive that doesn't interfere with a user using the tool just to ensure we don't accidentally set it too low.

cli/command/telemetry_docker.go
jsternberg: If the endpoint is empty, the intention is to return `nil, nil` since it's not an error but it's also not configured. The resulting code ends up being:

```go
if err != nil {
    return nil, err
} else if endpoint == "" {
    return nil, nil
}
```

Which is effectively the same thing as `endpoint == "" || err != nil`. I can add a comment to make that more clear.

krissetto: @Benehiko pls note that otel has it's own defaults that it uses when the value is not set (see [here](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/))

Benehiko: Oh i see! Thank you :)
I just wanted to make sure since it's not common to return `nil, nil` in Go.

cli/command/telemetry.go
laurazard: Right, that's what I thought. In that case, would you feel okay about setting this at 50ms? Thinking about the edge cases where some weird condition happens with the collector and it doesn't immediately fails but hangs, in that case 50ms is unnoticeable but two seconds is VERY noticeable.

Benehiko (approved):
To me at least this looks good! :)

laurazard:
@neersighted @thaJeztah can you TAL when you get a sec?

cli/command/telemetry.go
```diff
@@ -0,0 +1,210 @@
...
+func (cli *DockerCli) TracerProvider(ctx context.Context, opts ...sdktrace.TracerProviderOption) TracerProvider {
+	allOpts := make([]sdktrace.TracerProviderOption, 0, len(opts)+2)
+	allOpts = append(allOpts, sdktrace.WithResource(cli.Resource()))
+	if exp, err := dockerSpanExporter(ctx, cli); err != nil {
+		otel.Handle(err)
+	} else if exp != nil {
+		allOpts = append(allOpts, sdktrace.WithBatcher(exp))
+	}
+	allOpts = append(allOpts, opts...)
+	return sdktrace.NewTracerProvider(allOpts...)
```
neersighted, line 66:
```suggestion
	if exp, err := dockerSpanExporter(ctx, cli); err != nil {
		otel.Handle(err)
  } else {
		opts = append(opts, sdktrace.WithBatcher(exp))
	}
	return sdktrace.NewTracerProvider(sdktrace.WithResource(cli.Resource()), opts...)
```

There is no way `dockerSpanExporter` can return `nil, nil`; therefore all of this code can be simplified to a single append.

cli/command/telemetry.go
```diff
@@ -0,0 +1,210 @@
...
+func (cli *DockerCli) MeterProvider(ctx context.Context, opts ...sdkmetric.Option) MeterProvider {
+	allOpts := make([]sdkmetric.Option, 0, len(opts)+2)
+	allOpts = append(allOpts, sdkmetric.WithResource(cli.Resource()))
+	if exp, err := dockerMetricExporter(ctx, cli); err != nil {
+		otel.Handle(err)
+	} else if exp != nil {
+		allOpts = append(allOpts, sdkmetric.WithReader(newCLIReader(exp)))
+	}
+	allOpts = append(allOpts, opts...)
+	return sdkmetric.NewMeterProvider(allOpts...)
```
neersighted, line 78:
```suggestion
	if exp, err := dockerMetricExporter(ctx, cli); err != nil {
		otel.Handle(err)
	} else {
		opts = append(opts, sdkmetric.WithReader(newCLIReader(exp)))
	}
	allOpts = append(allOpts, opts...)
	return sdkmetric.NewMeterProvider(sdkmetric.WithResource(cli.Resource()), opts...)
```

Same pattern as above.

cli/command/telemetry.go
```diff
@@ -0,0 +1,210 @@
...
+	return func(cli *DockerCli) error {
+		cli.res.AppendOptions(opts...)
+		return nil
+	}
+}
+
+// WithResource overwrites the default resource and prevents its creation.
+func WithResource(res *resource.Resource) CLIOption {
+	return func(cli *DockerCli) error {
+		cli.res.Set(res)
```
neersighted, line 93: Isn't this at risk of getting clobbered if `once` hasn't fired yet? Is the caller expected to use this after the `sync.Once` is already called, or should there be a mechanism here to fire it manually?

cli/command/telemetry.go
neersighted: Agreed, this needs to be much shorter; high-latency connections should be something of an error state, and this will make them more painful.

cli/command/telemetry_docker.go
```diff
@@ -0,0 +1,121 @@
...
+	"path"
+
+	"github.com/pkg/errors"
+	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+const otelConfigFieldName = "otel"
```
neersighted, line 16: We should call this `otelContextFieldName` I think; I also would like to see this moved to the shared context implementation and used from there in Compose/Buildx, though I don't think it's a blocker here.

cli/command/telemetry_docker.go
```diff
@@ -0,0 +1,121 @@
...
+
+// dockerExporterOTLPEndpoint retrieves the OTLP endpoint used for the docker reporter
+// from the current context.
+func dockerExporterOTLPEndpoint(cli Cli) (endpoint string, secure bool, err error) {
```
neersighted, line 20: To elaborate on the above, this is more-or-less the code I want to see moved into the context itself (making "OTEL" a first class property/component of the context).

neersighted (changes requested):
Some (mostly Go-level/without a strong OTel background) thoughts; but overall the direction looks good.

cli/command/telemetry.go
jsternberg: Done. I also realized the timeout didn't apply to traces and found the option to apply this timeout to traces too. It's been renamed to `exportTimeout` when I end up pushing the change.

cli/command/telemetry.go
jsternberg: The function inside of `once` checks if this is already set and avoids running if it is.

cli/command/telemetry_docker.go
jsternberg: Changed the name. This shouldn't need to be exposed because Compose and Buildx would end up using this code and would remove their own implementations of this so I don't think a shared exposed constant is needed.

cli/command/telemetry_docker.go
jsternberg: Can you expand a bit on this? I'm not really sure I understand what you mean.

cli/command/telemetry.go
jsternberg: It is possible for `dockerSpanExporter` to return `nil, nil`, but I see that's been a bit confusing. Since all of these functions are unexported and the error always gets sent to `otel.Handle` and then gets swallowed, I think it's less confusing if this function returns a slice of options and the function itself is the one that invoked `otel.Handle` instead.

I've updated these appropriately so now these lines are just `append(allOpts, dockerSpanExporter(ctx, cli)...)`.

cli/command/telemetry_docker.go
neersighted: I'd like to see this eventually become part of the Context struct/have the lifecycle handled there; this is really just a note for a future refactor.

neersighted (approved):

PR #4892, opened 2024-02-22
Add a deprecation warning about --link on default nw

akerouanton:
Related to:

- https://github.com/moby/moby/pull/47427

**- What I did**

Update the "Deprecated Engine Features" doc page.

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -50,6 +50,7 @@ The table below provides an overview of the current status of deprecated feature
 
 | Status     | Feature                                                                                                                            | Deprecated | Remove |
 |------------|------------------------------------------------------------------------------------------------------------------------------------|------------|--------|
+| Deprecated | [Links on the default bridge network](#links-on-the-default-bridge-network)                                                        | v26.0      | -      |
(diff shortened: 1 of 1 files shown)


neersighted (approved):

docs/deprecated.md
```diff
@@ -50,6 +50,7 @@ The table below provides an overview of the current status of deprecated feature
 
 | Status     | Feature                                                                                                                            | Deprecated | Remove |
 |------------|------------------------------------------------------------------------------------------------------------------------------------|------------|--------|
+| Deprecated | [Links on the default bridge network](#links-on-the-default-bridge-network)                                                        | v26.0      | -      |
```
thaJeztah, line 53: I'm tempted to say that we deprecated it since docker 1.10 (when libnetwork was integrated in https://github.com/moby/moby/pull/19198), but as we never added that to `deprecated.md`, perhaps instead we should add some crumble-path / references in the description below.


We can already fill in the `Remove` column though, at least if we plan to remove it in v28.0;


> Deprecated: the feature is marked "deprecated" and should no longer be used. The feature may be removed, disabled, or change behavior in a future release. The "Deprecated" column contains the release in which the feature was marked deprecated, whereas the "Remove" column contains a tentative release in which the feature is to be removed. If no release is included in the "Remove" column, the release is yet to be decided on.

PR #4893, opened 2024-02-22
docs/deprecated: Deprecate Container fields in image inspect

vvoland:
See https://github.com/moby/moby/pull/46939

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -51,6 +51,7 @@ The table below provides an overview of the current status of deprecated feature
 | Status     | Feature                                                                                                                            | Deprecated | Remove |
 |------------|------------------------------------------------------------------------------------------------------------------------------------|------------|--------|
 | Deprecated | [Unauthenticated TCP connections](#unauthenticated-tcp-connections)                                                                | v26.0      | v27.0  |
+| Deprecated | [`Container` and `ContainerConfig` fields in Image inspect](#container-and-containerconfig-fields-in-image-inspect)                | v25.0      | v26.0  |
 | Deprecated | [Deprecate legacy API versions](#deprecate-legacy-api-versions)                                                                    | v25.0      | v26.0  |
 | Deprecated | [Container short ID in network Aliases field](#container-short-id-in-network-aliases-field)                                        | v25.0      | v26.0  |
 | Deprecated | [IsAutomated field, and "is-automated" filter on docker search](#isautomated-field-and-is-automated-filter-on-docker-search)       | v25.0      | v26.0  |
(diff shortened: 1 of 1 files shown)


docs/deprecated.md
```diff
@@ -110,6 +111,19 @@ The table below provides an overview of the current status of deprecated feature
 | Removed    | [`--run` flag on `docker commit`](#--run-flag-on-docker-commit)                                                                    | v0.10      | v1.13  |
 | Removed    | [Three arguments form in `docker import`](#three-arguments-form-in-docker-import)                                                  | v0.6.7     | v1.12  |
 
+### `Container` and `ContainerConfig` fields in Image inspect
+
+**Deprecated in Release: v25.0**
```
thaJeztah, line 116: Did we forget mentioning these as deprecated for the v25 release? Wondering if that means we need to wait for v27.0 to remove them, or did we document it in the API changes at the time?

vvoland: Yes, we forgot to put it there, but we mentioned it in `version-history.md` 🙈 

> The Container and ContainerConfig fields in the GET /images/{name}/json response are deprecated and will no longer be included in API v1.45.

But I'm fine with waiting for v27.0 for the removal, especially since the docker-py with the fixed tests is not released yet.

vvoland:
Changed deprecation candidate to v27

docs/deprecated.md
thaJeztah: Arf, I somehow missed your last comments. Some thoughts (_maybe_ we don't have to postpone the removal);

- The API changes document is the "canonical" place for changes in the engine API, _including_ deprecations
- I _thought_ we always included a mention like "for changes and deprecations in the API refer to `<link to API changes>`"; if we don't, we should fix that.
- :point_up: perhaps we should also do a cross-reference in this `deprecated` page, to do the same ("for an overview of changes in API ...")

That said; I don't think it necessarily _hurts_ to call out / repeat important changes here, but we may not _have to_ for all of them.


@vvoland @dvdksn WDYT?

thaJeztah: FWIW; for some of these, it would at least make things easier as well, because the API is versioned so if you use the "older API version", nothing changes, but yeah, in some cases that may affect output of `docker` CLI commands.

`docker inspect` was meant to show the underlying "low-level" information, which would still be "accurate" here, but that underlying information now may no longer have these fields (or not set).

dvdksn: We link to both the `deprecated.md` and `version-history.md` from the release note:

image


I don't think it hurts to also cross-reference between these two.

If `version-history.md` is canonical then we should be safe to remove in 26.0 since this was called out (and linked from the release note). It just wasn't called out everywhere.

image

vvoland: Updated the removal notice to v26.

docs/deprecated.md
```diff
@@ -110,6 +111,19 @@ The table below provides an overview of the current status of deprecated feature
...
+**Target For Removal In Release: v26.0**
+
+The `Container` and `ContainerConfig` fields returned by `docker inspect` are
+mostly an implementation detail of the classic (non-buildkit) image builder.
+These informations are not portable and are empty anyway when using the
+buildkit based builder (enabled by default since v23.0).
+These fields are deprecated in v25.0 and will be omitted starting from v26.0.
+If image configuration of an image is needed, it should be obtained from the
+`Config` field.
```
dvdksn, line 125:
```suggestion
The `Container` and `ContainerConfig` fields returned by `docker inspect` are
mostly an implementation detail of the classic (non-BuildKit) image builder.
These fields are not portable and are empty when using the
BuildKit-based builder (enabled by default since v23.0).
These fields are deprecated in v25.0 and will be omitted starting from v26.0.
If image configuration of an image is needed, you can obtain it from the
`Config` field.
```

dvdksn (commented):
minor nits, nonblocking

dvdksn (approved):

PR #4903, opened 2024-02-29
docs: include required tools in source tree

thaJeztah:
- carry of https://github.com/docker/cli/pull/4228
- closes https://github.com/docker/cli/pull/4228
- ⚠️ we should update this to use https://github.com/docker/cli-docs-tool/pull/14, so that we don't have to compile and install `go-md2man`

In order to be able to build the documentation without internet access (as is required by some distribution build systems), all of the source code needed for the build needs to be available in the source tarball.

This used to be possible with the docker-cli sources but was accidentally broken with some CI changes that switched to downloading the tools (by modifying go.mod as part of the docs build script).

This pattern also maked documentation builds less reproducible since the tool version used was not based on the source code version.

Fixes: commit 7dc35c03fca5 ("validate manpages target")
Fixes: commit a650f4ddd008 ("switch to cli-docs-tool for yaml docs generation")

**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah:
(Also wondering if we can rewrite things to be a `go:generate`)

man/tools.go
```diff
@@ -1,7 +0,0 @@
-//go:build tools
-
-package main
-
-import (
-	_ "github.com/cpuguy83/go-md2man/v2"
-)
```
thaJeztah, line 10: I removed the `tools.go` here, as we already vendor the dependency as indirect dependency of `docker/cli-docs-tool`. I could keep if it that's a concern (but things would show up pretty fast if we'd miss the dependency at some point)

vendor.mod
```diff
@@ -66,6 +66,7 @@ require (
 	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/containerd/log v0.1.0 // indirect
+	github.com/cpuguy83/go-md2man/v2 v2.0.6 // indirect
```
crazy-max, line 69: We could make use of the new `tool `directive starting Go 1.24:
* https://go.dev/doc/modules/managing-dependencies#tools
* https://go.dev/ref/mod#go-mod-file-tool

thaJeztah: Ah, yes, but we're not on go1.24 yet 😅 

At least currently it works, so I guess it's fine with how we have it right now?

scripts/docs/generate-man.sh
```diff
@@ -2,34 +2,15 @@
...
 
-function clean {
-  rm -rf "$buildir"
-}
+SCRIPTDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
-buildir=$(mktemp -d -t docker-cli-docsgen.XXXXXXXXXX)
-trap clean EXIT
+# build gen-manpages
+"${SCRIPTDIR}"/../with-go-mod.sh go build -mod=vendor -modfile=vendor.mod -tags manpages -o /tmp/gen-manpages ./man/generate.go
```
tianon-sso, line 10: While we have the hood open, can we also not (ab)use `/tmp/gen-manpages` as a static path here? 😇

It looks like this one is only used once, so it could be `go run` instead, right?

(Technically they both could be, since it'll do reliable caching, but I have other comments for `go-md2man` that may make that less desirable, where this is 100% code from _this_ repo.)

It's also slightly irritating to have this script _force_ `with-go-mod.sh` when it could instead be documented that it should be run itself with `with-go-mod.sh` so that those flags/envs can be adjusted more easily externally.

scripts/with-go-mod.sh
```diff
@@ -0,0 +1,33 @@
...
+		EOF
+	} >&2
+else
+	set -x
+
+	tee "${ROOTDIR}/go.mod" >&2 <<- EOF
+		module github.com/docker/cli
+
+		go 1.23.0
+	EOF
```
tianon-sso, line 29: Should/could this instead be something like `grep -E '^(module|go) ' vendor.mod | tee go.mod` so there's less to dual-maintain?

scripts/docs/generate-man.sh
```diff
@@ -2,34 +2,15 @@
...
-  cp man/tools.go .
-  # update vendor
-  ./scripts/vendor update
-  # build gen-manpages
-  go build -mod=vendor -modfile=vendor.mod -tags manpages -o /tmp/gen-manpages ./man/generate.go
-  # build go-md2man
-  go build -mod=vendor -modfile=vendor.mod -o /tmp/go-md2man ./vendor/github.com/cpuguy83/go-md2man/v2
-)
+# build go-md2man
+"${SCRIPTDIR}"/../with-go-mod.sh go build -mod=vendor -modfile=vendor.mod -o /tmp/go-md2man ./vendor/github.com/cpuguy83/go-md2man/v2
```
tianon-sso, line 13: As I noted over in the other PR, many/most distros will already have a package of `go-md2man` that they prefer to use, so this should somehow allow using `go-md2man` from `PATH` or similar (some way for the script invoker to say "do not build `go-md2man`, I already have it, it's here).

scripts/docs/generate-md.sh
```diff
@@ -2,32 +2,12 @@
...
-  # to be able to fetch the required dependencies
-  go mod edit -modfile=vendor.mod -require=github.com/docker/cli-docs-tool@${CLI_DOCS_TOOL_VERSION}
-  cp docs/generate/tools.go .
-  # update vendor
-  ./scripts/vendor update
-  # build docsgen
-  go build -mod=vendor -modfile=vendor.mod -tags docsgen -o /tmp/docsgen ./docs/generate/generate.go
-)
+# build docsgen
+"${SCRIPTDIR}"/../with-go-mod.sh go build -mod=vendor -modfile=vendor.mod -tags docsgen -o /tmp/docsgen ./docs/generate/generate.go
```
tianon-sso, line 10: Same comments here WRT forcing `with-go-mod.sh` and hard-coded `/tmp/` (consider `go run`?)

scripts/docs/generate-yaml.sh
```diff
@@ -2,32 +2,12 @@
...
-  # to be able to fetch the required dependencies
-  go mod edit -modfile=vendor.mod -require=github.com/docker/cli-docs-tool@${CLI_DOCS_TOOL_VERSION}
-  cp docs/generate/tools.go .
-  # update vendor
-  ./scripts/vendor update
-  # build docsgen
-  go build -mod=vendor -modfile=vendor.mod -tags docsgen -o /tmp/docsgen ./docs/generate/generate.go
-)
+# build docsgen
+"${SCRIPTDIR}"/../with-go-mod.sh go build -mod=vendor -modfile=vendor.mod -tags docsgen -o /tmp/docsgen ./docs/generate/generate.go
```
tianon-sso, line 10: Same again 😅

scripts/vendor
```diff
@@ -13,18 +13,12 @@ if [ -z "$TYP" ]; then
...
-go 1.23.0
-EOL
-}
-
-update() {
-  (set -x ; go mod tidy -modfile=vendor.mod; go mod vendor -modfile=vendor.mod)
-}
+update() (
+  set -x
+  "${SCRIPTDIR}"/with-go-mod.sh go mod tidy -modfile vendor.mod
```
tianon-sso, line 20: Again, I think it's better to document that the invoker of these should run `with-go-mod.sh`, especially since that'll cause less churn if there's more than one usage of `go.mod` within a given script.

scripts/docs/generate-man.sh
thaJeztah: Good point! I was already eyeing the `/tmp/` because I didn't like it, but didn't try if using `go run `would work (in combination with the flags we need to pass); looks like that all works fine, so I just gave it a go to do so!

scripts/with-go-mod.sh
thaJeztah: Oh! Didn't do this one yet, might do so in a follow-up

scripts/docs/generate-man.sh
thaJeztah: I gave it a go to look for `go-md2man` (or a binary-name set in a `GO_MD2MAN` env-var), and if it doesn't exist, to compile it from source.

I changed the location for that to be inside `./build/tools` so that we don't pollute system-wide paths

scripts/vendor
thaJeztah: :heart: your comments made me realise that I don't have to include it in each script; I moved it to the Makefile instead; it's not perfect, but at least keeps the scripts themselves clean. Possibly we could move it into a Makefile var / macro as well so that it can be overridden (to run the scripts without the `with-go-mod.sh` script before it).

man/tools.go
```diff
@@ -2,9 +2,4 @@
 
 package main
 
-import (
-	_ "github.com/cpuguy83/go-md2man/v2"
-	_ "github.com/spf13/cobra"
-	_ "github.com/spf13/cobra/doc"
-	_ "github.com/spf13/pflag"
-)
+import _ "github.com/cpuguy83/go-md2man/v2"
```
thaJeztah, line 5: Turned out I needed the `tools.go` after all (don't ask!) 🙈 😂

scripts/docs/generate-man.sh
```diff
@@ -2,38 +2,25 @@
 
 set -eu
 
-: "${MD2MAN_VERSION=v2.0.6}"
-
 export GO111MODULE=auto
```
tianon-sso, line 5: This should be dropped entirely, right? (from all these scripts, since it's handled by `with-go-mod.sh`)

scripts/docs/generate-man.sh
```diff
@@ -2,38 +2,25 @@
...
 
-function clean {
-  rm -rf "$buildir"
-}
+# build go-md2man
+: "${GO_MD2MAN:=go-md2man}"
 
-buildir=$(mktemp -d -t docker-cli-docsgen.XXXXXXXXXX)
-trap clean EXIT
+if [ ! -x "$(command -v $GO_MD2MAN)" ]; then
```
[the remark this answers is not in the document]
thaJeztah: Ah! Will update this one. I got lazy, and copy/pasta'd it from one of the other scripts

scripts/docs/generate-man.sh
thaJeztah: Ah, good point; I was looking at that, then wasn't 100% sure if it would still be needed if the script would be executed _without_ the `with-go-mod.sh` script. But I guess in that case, there would be either a `go.mod` or otherwise setup already, so yes, let me drop it.

tianon-sso (approved):

thaJeztah:
@Benehiko @vvoland PTAL


I'll have a look at some follow-ups as well;

- Rewrite the manpage generator to make use of https://github.com/docker/cli-docs-tool/pull/14
- :point_up: I realized that won't (currently) replace `go-md2man`, but as we're already using code from that, we _may_ be able to include its functionality (?)
- :point_right: because effectively, it's ~20 lines of code or less to execute its "main" command; https://github.com/cpuguy83/go-md2man/blob/ae8f92a626e8922695dbf0fb24cf0241e056082c/md2man.go#L31-L48
- :smile: possibly rewrite this code to a `//go:generate`, so that we won't need any of the (ba)sh scripts at all, and just `go generate ./...` to generate or update these files.

vvoland (approved):

PR #4905, opened 2024-02-29
plugin: closer-based plugin notification socket

cpuguy83:
This changes things to rely on a plugin server that manages all
connections made to the server.

An optional handler can be passed into the server when the caller wants
to do extra things with the connection.

It is the caller's responsibility to close the server.
When the server is closed, first all existing connections are closed
(and new connections are prevented).

Now the signal loop only needs to close the server and not deal with
`net.Conn`'s directly (or double-indirects as the case was before this
change).

The socket, when present in the filesystem, is no longer unlinked
eagerly, as reconnections require it to be present for the lifecycle of
the plugin server.


cli-plugins/socket/socket.go
```diff
@@ -13,18 +14,44 @@ import (
...
+			return
+		}
+
+		go handleServerConn(ctx, conn)
+
+		for {
+			conn, err := l.Accept()
+			if err != nil {
+				return
+			}
```
neersighted, line 43:
```suggestion
	go func() {
		defer l.Close()
    unlinkOnce = sync.OnceFunc(func() {
        unlink(l)
    })

		for {
			conn, err := l.Accept()
			unlinkOnce()
			if err != nil {
				return
			}
```

neersighted (changes requested):
An excellent simplification. I've been on a bit of a "use more contexts" kick recently, but I don't know how I totally overlooked them as the ideal/correct synchronization primitive here.

I'd like to see the loop used to handle plugin connections simplified a bit.

In addition, two other thoughts:
* FreeBSD doesn't appear to have anonymous sockets either, so really we appear to have "the Linuxes" (Windows + Linux) and "the Berkley Unixes." Based on that, I think we should have a `socket_linux_windows` and a `socket_nolinux_nowindows` instead of the current arrangement.
* To make things flow a little smoother, why not make `listen()` return the `UnixAddr` pointer so we don't have to pull it back out of the listener and typecast it?

cpuguy83:
Can we use this for anything other than just closing it without breaking existing plugins?

cpuguy83:
To clarify:

If the plugin is never expecting to receive data on this socket and is only waiting for a read to return, I'm not sure it is possible to re-use the same socket without breaking those assumptions.

neersighted:
The existing code does a busy-read, discarding any data read until an EOF is reached, upon which it cancels the CLI plugin context (which should terminate the process).

thaJeztah:
Looks like the linter is complaining;


```
62.42 cli-plugins/socket/socket.go:40:4: shadow: declaration of "conn" shadows declaration at line 31 (govet)
62.42 			conn, err := l.Accept()
62.42 			^
62.42 cli-plugins/socket/socket_test.go:20:3: shadow: declaration of "ctx" shadows declaration at line 18 (govet)
62.42 		ctx, cancel := context.WithCancel(ctx)
62.42 		^
```

neersighted:
> Looks like the linter is complaining
With my suggested changes the shadowing the linter complains about should go away.

cli-plugins/socket/socket_test.go
```diff
@@ -78,18 +97,26 @@ func checkDirNoPluginSocket(t *testing.T) {
...
-		pollConnNotNil(t, &conn)
-		conn.Close()
+
+		select {
+		case <-done:
+			t.Fatal("unexpectedly done")
+		default:
+		}
+
+		cancel()
```
laurazard, line 119: nice!

cli-plugins/socket/socket_test.go
```diff
@@ -14,26 +15,44 @@ import (
 )
 
 func TestSetupConn(t *testing.T) {
+	ctx := context.Background()
 	t.Run("updates conn when connected", func(t *testing.T) {
```
laurazard, line 19: We should update this test name to reflect the changes

cli-plugins/socket/socket.go
laurazard: 👍

laurazard (approved):
Really nice change overall, thanks!

cpuguy83:
I've updated this:

- No longer uses context
- Returns a closer which is what gets called in the signal loop
- Closer is a `PluginServer` which manages any/all connections internally
- `PluginServer` takes an optional handler which is called for every new connection

Benehiko (approved):
Great stuff! :)

cli-plugins/socket/socket.go
```diff
@@ -7,24 +7,106 @@ import (
...
+
+type PluginServer struct {
+	mu     sync.Mutex
+	conns  []net.Conn
+	l      *net.UnixListener
+	h      func(net.Conn)
+	closed bool
+}
+
+func (l *PluginServer) accept() error {
```
neersighted, line 61:
```suggestion
func (p *PluginServer) accept() error {
```

We should use `p` or `pl` for these methods; `l.l` is just a bit confusing...

cli-plugins/socket/socket_nodarwin.go
```diff
@@ -13,7 +13,7 @@ func listen(socketname string) (*net.UnixListener, error) {
 	})
 }
 
-func onAccept(conn *net.UnixConn, listener *net.UnixListener) {
+func unlink(listener *net.UnixListener) {
```
neersighted, line 16: Would you mind making the swap to a `linux_windows` and `nolinux_nowindows` pair of buildtagged files here?

neersighted (changes requested):
I see you're thinking a fair bit more elaborately/with an eye to implementing the connection-handling code sooner than later. Thanks!

A couple minor nits, otherwise LGTM.

krissetto (approved):
Overall LGTM after the other small suggestions are in, thanks!

cli-plugins/socket/socket.go
```diff
@@ -7,24 +7,106 @@ import (
...
+
+	unlinkOnce := sync.OnceFunc(func() {
+		unlink(l)
+	})
+
+	go func() {
+		defer pl.Close()
+		for {
+			err := pl.accept()
+			unlinkOnce()
```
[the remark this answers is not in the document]
neersighted: This is a standard Unix behavior, which is why it's a bit under-explained. A Unix socket isn't even really a 'file' -- it's a name in the filesystem. Once the plugin has used that named path to open the other side, there is no further need to keep it around (and indeed we're trying to avoid 'leaking' socket handles in the filesystem).

A `defer` is not equivalent here as we want to unlink as early as possible. If we are killed (say, by a SIGSEGV or the OOM killer), we may end up leaking the socket.

cli-plugins/socket/socket_darwin.go
```diff
@@ -14,6 +14,6 @@ func listen(socketname string) (*net.UnixListener, error) {
 	})
 }
 
-func onAccept(conn *net.UnixConn, listener *net.UnixListener) {
+func unlink(listener *net.UnixListener) {
 	syscall.Unlink(listener.Addr().String())
```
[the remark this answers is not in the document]
neersighted: This is somewhat 'optional' in that it's a best-effort attempt to prevent leaking socket names into the filesystem on applicable systems (see the above comment); there are indeed a [litany](https://man7.org/linux/man-pages/man2/unlink.2.html#ERRORS) of possible errors during `unlink(2)`; however given we just created the socket above, it's rather unlikely that this fails.

In the event it does, it's not the end of the world; just a 'leaked name' in the filesystem, in TMPDIR.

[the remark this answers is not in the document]
thaJeztah: To please the linters, I think it would be good to `_ = syscall....` here. It also makes it explicit that we chose to ignore it (and not that we overlooked error-handling).

(Bonus points for a short `// comment` that we consider it "best effort", and it's not critical)

cli-plugins/socket/socket.go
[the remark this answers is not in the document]
krissetto: > ...Once the plugin has used that named path to open the other side, there is no further need to keep it around (and indeed we're trying to avoid 'leaking' socket handles in the filesystem).
> 
> A `defer` is not equivalent here as we want to unlink as early as possible. If we are killed (say, by a SIGSEGV or the OOM killer), we may end up leaking the socket.

Thanks for the explanation, makes sense! I was unaware of that unix socket behavior and the practice of removing it asap (which seemed counter intuitive to me)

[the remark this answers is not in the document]
neersighted: Yeah, it only makes sense in this case as we expect a single process to connect. Typically you'd unlink near program shutdown if you had a server that expected to service multiple clients (e.g. the defer).

neersighted (approved):
I've carried the requested review changes; LGTM now with @cpuguy83's own review.

Benehiko (approved):
LGTM

thaJeztah:
Failure looks to be in this area?

```
--- FAIL: TestPluginServer (0.10s)
    --- FAIL: TestPluginServer/allows_reconnects (0.10s)
        socket_test.go:89: assertion failed: error is not nil: dial unix /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/docker_cli_eb8ce106e1522d6ad46ca1fbd263a81f: connect: no such file or directory: failed to redial server

neersighted:
> Failure looks to be in this area?
> 
> ```
> --- FAIL: TestPluginServer (0.10s)
>     --- FAIL: TestPluginServer/allows_reconnects (0.10s)
>         socket_test.go:89: assertion failed: error is not nil: dial unix /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/docker_cli_eb8ce106e1522d6ad46ca1fbd263a81f: connect: no such file or directory: failed to redial server
> ```

Yes; @cpuguy83 added the ability to reconnect to the server, instead of it being a one-shot. This means we can no longer do the eager unlink (so, we're back to the behavior @krissetto expected, where we unlink at the end of the lifecycle); it is now more possible to 'leak' sockets if the process is ungracefully terminated, but that seems to be the cost of the paradigm shift to the PluginServer.

Benehiko (approved):
LGTM

cpuguy83:
Sorry I forgot I had some changes pending here.

neersighted:
> Sorry I forgot I had some changes pending here.

No problem! Does my set of changes look good to you (I kept your new reconnection feature, which meant dropping the early unlink... Before the `for {}` loop was really a poor man's synchronization point/barrier)?

cpuguy83:
LGTM, it won't let me approve it officially since its my own PR.

laurazard (approved):
LGTM, thanks!

PR #4909, opened 2024-03-04
[25.0 backport] Avoid keeping @docker_cli_[UUID] files

vvoland:
- backport: https://github.com/docker/cli/pull/4862

Seems that OpenBSD behaves like darwin and requires to unlink all socket, after it was used.

Tested on OpenBSD 7.4

(cherry picked from commit 2c214241fa2eb3b44d15d7aad587168d3bd70b48)


thaJeztah (approved):
LGTM

PR #4929, opened 2024-03-08
docs: clarify what the --privileged flag does

dvdksn:
**- What I did**

Attempt to clarify the permissions and capabiliities (and risks) with
using the --privileged flag.

Relates to https://github.com/moby/moby/issues/24387

**- How I did it**

**- How to verify it**

### DIFF
--- a/docs/reference/commandline/container_run.md
+++ b/docs/reference/commandline/container_run.md
@@ -326,7 +326,37 @@ are broken into multiple containers, you might need to share the IPC mechanisms
 of the containers, using `"shareable"` mode for the main (i.e. "donor")
 container, and `"container:<donor-name-or-ID>"` for other containers.
 
-### <a name="privileged"></a> Full container capabilities (--privileged)
+### <a name="privileged"></a> Escalate container privileges (--privileged)
+
+The `--privileged` flag gives the following capabilities to a container:
+
+- Enables all Linux kernel capabilities
+- Disables the default seccomp profile
+- Disables the default AppArmor profile
+- Disables the SELinux process label
+- Grants access to all host devices
+- Makes `/sys` read-write
+- Makes cgroups mounts read-write
+
+In other words, the container can then do almost everything that the host can
+do. This flag exists to allow special use-cases, like running Docker within
+Docker.
+
+> **Warning**
+>
+> Use the `--privileged` flag with caution.
+> A container with `--privileged` is not a securely sandboxed process.
+> Containers in this mode can get a root shell on the host
+> and take control over the system.
+>
+> For most use cases, this flag should not be the preferred solution.
+> If your container requires escalated privileges,
+> you should prefer to explicitly grant the necessary permissions,
+> for example by adding individual kernel capabilities with `--cap-add`.
+>
+> For more information, see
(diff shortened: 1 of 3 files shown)


tianon (approved):

Benehiko (approved):

docs/reference/commandline/container_run.md
```diff
@@ -326,7 +326,32 @@ are broken into multiple containers, you might need to share the IPC mechanisms
...
+do. This flag exists to allow special use-cases, like running Docker within
+Docker.
+
+> **Note**
+>
+> Use the `--privileged` flag with caution.
+> For most use cases, this flag should not be the preferred solution.
+> If your container requires escalated privileges,
+> you should prefer to explicitly grant the necessary permissions,
+> for example by dropping only the required kernel capabilities with `--cap-drop`.
```
thaJeztah, line 351: This example is a bit confusing; `--cap-drop` removes from the default set of capabilities, so won't give additional "necessary" permissions.

docs/reference/commandline/container_run.md
```diff
@@ -326,7 +326,32 @@ are broken into multiple containers, you might need to share the IPC mechanisms
...
+- Disables the SELinux process label
+- Grants access to all host devices
+- Makes `/sys` read-write
+- Makes cgroups mounts read-write
+
+In other words, the container can then do almost everything that the host can
+do. This flag exists to allow special use-cases, like running Docker within
+Docker.
+
+> **Note**
```
thaJeztah, line 345: I'd be fine with making this a warning. We should warn that a container with `--privileged` is, well, effectively "not containing", and is able to take control of the host. Something like this gives you a `root` shell on the host;

```bash
docker run -it --rm --privileged --pid=host debian nsenter -t 1 -m -u -n -i sh
```

docs/reference/commandline/container_run.md
```diff
@@ -326,7 +326,32 @@ are broken into multiple containers, you might need to share the IPC mechanisms
 of the containers, using `"shareable"` mode for the main (i.e. "donor")
 container, and `"container:<donor-name-or-ID>"` for other containers.
 
-### <a name="privileged"></a> Full container capabilities (--privileged)
+### <a name="privileged"></a> Escalate container privileges (--privileged)
+
+The `--privileged` flag gives the following capabilities to a container:
```
thaJeztah, line 331: Also wondering if we should add a (more detailed?) section in the https://docs.docker.com/go/daemon-access/ section.

thaJeztah:
Looks like validation is failing on this one (`docker exec` docs having to be re-generated?)

docs/reference/commandline/container_run.md
dvdksn: Good call. I'll update

docs/reference/commandline/container_run.md
dvdksn: (asking because it's not immediately obvious to me what `--privileged` has to do with post-install actions, or with user group permissions)

thaJeztah: Ah, sorry, more in line of the "daemon attack vector". Being able to start a privileged container is a prime example of "bad things" people could do when having access to the API (hope that makes sense)

neersighted (approved):

PR #4931, opened 2024-03-08
Add a description of '--internal' to the docs

robmry:
**- What I did**

Add a description of a --internal network (written by @neersighted) to the CLI docs - as they currently only mention overlay networks.

(Also, while I was there, update the description for the table listing 'network create' options, to make it consistent with the table listing 'bridge' options.)

**- How I did it**

**- How to verify it**

Added a description of --internal networks to the docs.

### DIFF
--- a/docs/reference/commandline/network_create.md
+++ b/docs/reference/commandline/network_create.md
@@ -184,6 +184,12 @@ $ docker network create \
 
 ### <a name="internal"></a> Network internal mode (--internal)
 
+Containers on an internal network may communicate between each other, but not
+with any other network, as no default route is configured and firewall rules
+are set up to drop all traffic to or from other networks. Communication with
+the gateway IP  address (and thus appropriately configured host services) is
+possible, and the host may communicate with any container IP directly.
+
 By default, when you connect a container to an `overlay` network, Docker also
 connects a bridge network to it to provide external connectivity. If you want
 to create an externally isolated `overlay` network, you can specify the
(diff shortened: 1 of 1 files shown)


docs/reference/commandline/network_create.md
```diff
@@ -183,6 +184,13 @@ $ docker network create \
 
 ### <a name="internal"></a> Network internal mode (--internal)
 
+Containers on an internal network may communicate between each other, but are
+precluded from communicating with any networks the host has access to (LAN or
+WAN) as no default route is configured, and firewall rules are set up to drop
+all outgoing traffic. Communication with the gateway IP address (and thus
+appropriately configured host services) is possible, and the host may
+communicate with any container IP directly.
```
dvdksn, line 192: Can we make the first sentence shorter and easier to read for non-native readers? For example:

```suggestion
Containers on an internal network may communicate between each other,
but not on any other network, as no default route is configured,
and firewall rules are set up to drop all outgoing traffic.
Communication with the gateway IP address (and thus
appropriately configured host services) is possible, and the host may
communicate with any container IP directly.
```

robmry: Thanks David ... I modified that slightly, hopefully it's ok?

dvdksn (approved):
:shipit:

Benehiko (approved):

thaJeztah (approved):
LGTM

PR #4936, opened 2024-03-11
cli/command: deprecate EventHandler and InitEventHandler

thaJeztah:
This code was only used as part of container.RunStats, so moving the code there instead as a non-exported type. The actual use also did not have to handle concurrency, so the mutex is removed in the new location.


cli/command/container/stats.go
```diff
@@ -316,3 +317,31 @@ func RunStats(ctx context.Context, dockerCLI command.Cli, options *StatsOptions)
...
+func newEventHandler() *eventHandler {
+	return &eventHandler{handlers: make(map[events.Action]func(events.Message))}
+}
+
+// eventHandler allows for registering specific events to handle.
+type eventHandler struct {
+	handlers map[events.Action]func(events.Message)
+}
+
+func (w *eventHandler) handle(action events.Action, h func(events.Message)) {
```
krissetto, line 331: Very minor nits here:
- `w` as the eventHandler name? Could we use `eh` like we do above?
- For the naming i'd suggest something along the lines of `setHandler`/`setActionHandler` etc since we don't actually do any handling in here. WDYT?

vvoland (approved):

krissetto (approved):

PR #4939, opened 2024-03-12
feat: standardize error for prompt

Benehiko:
Follow up to https://github.com/docker/cli/pull/4888

This PR standardizes the commands return value that use `PromptForConfirmation` and verifies the exit code to be 0 on prompt termination.

**- What I did**
Add e2e tests, update command prompt termination tests and adjust error type returned from `PromptForConfirmation` function.

**- How I did it**

**- How to verify it**

Standardize the exit code to `0` for cancelling a confirmation prompt with SIGINT.


cli/command/trust/revoke.go
```diff
@@ -25,7 +26,11 @@ func newRevokeCommand(dockerCLI command.Cli) *cobra.Command {
 		Short: "Remove trust for an image",
 		Args:  cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return revokeTrust(cmd.Context(), dockerCLI, args[0], options)
+			err := revokeTrust(cmd.Context(), dockerCLI, args[0], options)
+			if errdefs.IsCancelled(err) {
+				return nil
+			}
```
Benehiko, line 32: todo: remove

cli/command/plugin/upgrade.go
```diff
@@ -64,10 +64,7 @@ func runUpgrade(ctx context.Context, dockerCli command.Cli, opts pluginOptions)
 	fmt.Fprintf(dockerCli.Out(), "Upgrading plugin %s from %s to %s\n", p.Name, reference.FamiliarString(old), reference.FamiliarString(remote))
 	if !opts.skipRemoteCheck && remote.String() != old.String() {
 		if r, err := command.PromptForConfirmation(ctx, dockerCli.In(), dockerCli.Out(), "Plugin images do not match, are you sure?"); !r || err != nil {
-			if err != nil {
-				return errors.Wrap(err, "canceling upgrade request")
-			}
-			return errors.New("canceling upgrade request")
+			return err
```
Benehiko, line 67: todo: ensure an error is returned, if `!r` it should return a cancelled error.

cli/command/network/remove.go
```diff
@@ -25,7 +25,12 @@ func newRemoveCommand(dockerCli command.Cli) *cobra.Command {
 		Short:   "Remove one or more networks",
 		Args:    cli.RequiresMinArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runRemove(cmd.Context(), dockerCli, args, &opts)
+			err := runRemove(cmd.Context(), dockerCli, args, &opts)
+			if errdefs.IsCancelled(err) {
+				fmt.Fprintln(dockerCli.Out(), "")
+				return nil
+			}
```
Benehiko, line 32: todo: remove

cli/command/network/remove_test.go
```diff
@@ -115,8 +114,5 @@ func TestNetworkRemovePromptTermination(t *testing.T) {
 	})
 	cmd := newRemoveCommand(cli)
 	cmd.SetArgs([]string{"existing-network"})
-	test.TerminatePrompt(ctx, t, cmd, cli, func(t *testing.T, err error) {
-		t.Helper()
-		assert.ErrorIs(t, err, command.ErrPromptTerminated)
-	})
+	test.TerminatePrompt(ctx, t, cmd, cli, nil)
```
Benehiko, line 117: todo: should return an error

cli/command/trust/revoke.go
```diff
@@ -44,13 +49,9 @@ func revokeTrust(ctx context.Context, dockerCLI command.Cli, remote string, opti
...
 		deleteRemote, err := command.PromptForConfirmation(ctx, dockerCLI.In(), dockerCLI.Out(), fmt.Sprintf("Please confirm you would like to delete all signature data for %s?", remote))
-		if err != nil {
-			fmt.Fprintf(dockerCLI.Out(), "\nAborting action.\n")
-			return errors.Wrap(err, "aborting action")
-		}
-		if !deleteRemote {
+		if err != nil || !deleteRemote {
 			fmt.Fprintf(dockerCLI.Out(), "\nAborting action.\n")
-			return nil
+			return err
```
Benehiko, line 54: todo: ensure error is returned on `!deleteRemote`

cli/command/trust/revoke_test.go
```diff
@@ -159,10 +158,7 @@ func TestRevokeTrustPromptTermination(t *testing.T) {
 	cli := test.NewFakeCli(&fakeClient{})
 	cmd := newRevokeCommand(cli)
 	cmd.SetArgs([]string{"example/trust-demo"})
-	test.TerminatePrompt(ctx, t, cmd, cli, func(t *testing.T, err error) {
-		t.Helper()
-		assert.ErrorIs(t, err, command.ErrPromptTerminated)
-	})
+	test.TerminatePrompt(ctx, t, cmd, cli, nil)
```
Benehiko, line 161: todo: should return an error

vendor.mod
Benehiko, line 1: todo: remove

Benehiko:
Getting this test failure now https://github.com/docker/cli/actions/runs/8362281837/job/22892487855?pr=4939#step:4:397

Will look at it tomorrow. Most likely related https://github.com/docker/cli/issues/4948

cli/command/container/prune.go
```diff
@@ -28,6 +29,10 @@ func NewPruneCommand(dockerCli command.Cli) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			spaceReclaimed, output, err := runPrune(cmd.Context(), dockerCli, options)
 			if err != nil {
+				if errdefs.IsCancelled(err) {
+					fmt.Fprintln(dockerCli.Out(), output)
```
thaJeztah: What does `output` contain in this situation? Is this a "partial prune" that is printed? If that's the case, should use instead handle it as "normal", and print both `output` and `spaceReclaimed` ?

Something like;

```go
if err != nil && !errdefs.IsCancelled(err) {
	return nil
}
if output != "" {
	fmt.Fprintln(dockerCli.Out(), output)
}
fmt.Fprintln(dockerCli.Out(), "Total reclaimed space:", units.HumanSize(float64(spaceReclaimed)))
return nil
```

Perhaps we need a check for the `spaceReclaimed` to only print if we did reclaim (not sure?).

thaJeztah: On a side-note (I should open a ticket for that); `docker system prune` and `docker builder prune` are really awkward currently, as they don't print any progress until the prune finishes.

We should improve that;

- print prune progress as we go (instead of dump everything at the end)
- consider if we want `build cache` entries to be printed individually (their output is not very useful to the user), and instead possibly print a progress-bar or something along those lines)

internal/test/writer.go
```diff
@@ -4,27 +4,22 @@ import (
...
-// This is useful in testing to wait for a write to complete,
-// or to check what was written.
-// To create a WriterWithHook use the NewWriterWithHook function.
-type WriterWithHook struct {
+type writerWithHook struct {
 	actualWriter io.Writer
 	hook         func([]byte)
 }
 
-// Write writes p to the actual writer and then calls the hook function.
```
thaJeztah, line 17: Was there a reason for removing this one? (I understand it's not an exported type, but perhaps the description is still useful 🤔)

cli/command/builder/prune_test.go
```diff
@@ -21,8 +19,5 @@ func TestBuilderPromptTermination(t *testing.T) {
 		},
 	})
 	cmd := NewPruneCommand(cli)
-	test.TerminatePrompt(ctx, t, cmd, cli, func(t *testing.T, err error) {
-		t.Helper()
-		assert.ErrorIs(t, err, command.ErrPromptTerminated)
-	})
+	test.TerminatePrompt(ctx, t, cmd, cli, nil)
```
thaJeztah: The default is not checking for an error-type; are we losing coverage because of that?

```go
		if assertFunc != nil {
			assertFunc(t, err)
		} else {
			assert.NilError(t, err)
			assert.Equal(t, cli.ErrBuffer().String(), "")
		}
```

cli/command/plugin/testdata/plugin-upgrade-terminate.golden
```diff
@@ -1,2 +1,2 @@
 Upgrading plugin foo/bar from localhost:5000/foo/bar:v0.1.0 to localhost:5000/foo/bar:v1.0.0
-Plugin images do not match, are you sure? [y/N] 
\ No newline at end of file
+Plugin images do not match, are you sure? [y/N]
```
thaJeztah: Why did this change in this commit?

cli/command/container/prune_test.go
```diff
@@ -20,5 +22,8 @@ func TestContainerPrunePromptTermination(t *testing.T) {
 		},
 	})
 	cmd := NewPruneCommand(cli)
-	test.TerminatePrompt(ctx, t, cmd, cli, nil)
+	test.TerminatePrompt(ctx, t, cmd, cli, func(t *testing.T, err error) {
+		t.Helper()
+		assert.ErrorIs(t, err, command.ErrPromptTerminated)
+	})
```
thaJeztah: Is this ... reverting the changes from the previous commit? Should those be squashed?

cli/command/network/prune.go
```diff
@@ -27,10 +26,6 @@ func NewPruneCommand(dockerCli command.Cli) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			output, err := runPrune(cmd.Context(), dockerCli, options)
 			if err != nil {
-				if errdefs.IsCancelled(err) {
-					fmt.Fprintln(dockerCli.Out(), output)
-					return nil
-				}
 				return err
```
thaJeztah: Same here?

e2e/global/cli_test.go
```diff
@@ -120,26 +124,87 @@ func TestPromptExitCode(t *testing.T) {
...
-					icmd.Command("docker", "plugin", "disable", "vieux/sshfs"),
-				)
-				return icmd.Command("docker", "plugin", "upgrade", "vieux/sshfs:latest", "vieux/sshfs:next")
-			},
-		},
+		// {
+		// 	name: "plugin install",
+		// 	run: func(t *testing.T) icmd.Cmd {
+		// 		t.Helper()
+		// 		skip.If(t, versions.LessThan(environment.DaemonAPIVersion(t), "1.44"))
```
thaJeztah: This looks like some temporary commit; might be better to squash these (re-shape commits so that they can be reviewed individually)

cli/command/utils.go
```diff
@@ -78,7 +78,10 @@ func PrettyPrint(i any) string {
 
 type PromptError error
```
thaJeztah, line 79: Looks like this type was introduced in 2ae903e86cab51f694c819721cdfdf5eec693720, but I don't see it being used as sentinel error; is there a reason we introduced a "special" type for this? We should be careful introducing special types, as (especially when exported) they will add to our public API, and may lead to expectations of them having special meaning.

Where possible, limit to `errdefs` interfaces / definitions, which allow handling specific error "classes" in most cases, and reserve special types where specific handling _must_ be used.

cli/command/utils.go
```diff
@@ -78,7 +78,10 @@ func PrettyPrint(i any) string {
...
 
-var ErrPromptTerminated = PromptError(errors.New("prompt terminated"))
+var (
+	ErrPromptTerminated       = PromptError(errdefs.Cancelled(errors.New("prompt terminated")))
+	ErrUserCancelledOperation = PromptError(errdefs.Cancelled(errors.New("user cancelled operation")))
+)
```
thaJeztah, line 84: Somewhat similar here, I _guess_ the main reason here is for consistency. On its own that's good; we must be careful not to put _too much_ value to it; exporting the error comes with a "load" / "expectation" to it which others may now start to depend on.

This can be risky; for a long time (and there might still be code in place) we ended up with errors that _had to match_ a specific string (`No such` for "not found errors"), which spread across many code-bases, and now meant we could not change errors without breaking code-flows.

So if these errors are _only_ for consistency, and _not_ special / sentinel errors, we could still have a (non exported?) utility to price the error (`return someUtilityToProduceError()`), but prevent that from becoming a _must_ (for code flow).

While we _should_ prefer consistency from a UX perspective; there may be exceptions to the rule (a friendlier, but still recognisable error, which could be specific to the action at hand (_"pruning buildcache is cancelled"_  / _"pruning volumes is cancelled"_).


The `errdefs` ("interface approach") types were created from that perspective; they would allow the flexibility of changing the (user-facing?) error _message_ without breaking business logic (and the "interface" approach allows for custom implementations of the interface, without enforcing a specific implementation / type / definition to be  imported).

cli/command/plugin/testdata/plugin-upgrade-terminate.golden
laurazard: It was due to what I pointed out in my review, with moving the newline printing logic to `cmd/docker/docker.go`. I suggested moving it back so that these don't change, too.

internal/test/writer.go
laurazard: The description is still in the exported function iirc.

cli/command/container/prune_test.go
Benehiko: Yes sorry! I am very bad at keeping code organized across commits. I would prematurely commit something, realize it's bad and then revert with another commit. It's new to me to squash commits in my branch for review purposes, I'll improve this for future PRs

e2e/global/cli_test.go
Benehiko: same here, I temporarily committed this to be able to share what I had since this part I had no clue how the plugin system worked. So I reached out on slack about it using the commit reference. I'll clean up PRs in the future with squashes

cli/command/utils.go
Benehiko: This makes sense, it's a habit of mine to "organize" errors under a type. I suppose in this case I can remove it and just use the exported variable instead.

cli/command/utils.go
Benehiko: Hmm I see, in this case I used the exported error message for consistency. I guess we don't need to return the same error message every time a user enters a "no" on the prompt like here
https://github.com/docker/cli/blob/63bf7dec61bbb4d1f9dbbb8133367d6aeae762b5/cli/command/volume/prune.go#L84-L86

Since the prompt will print a new line on an stdin and we take care of printing a new line on termination, we could theoretically just return a nil when the prompt returns false. 

```

            ---- True --- continue code
Prompt -- > ---- False --- return nil
            ---- Terminate --- return defined error

File: cli/command/container/prune.go (hunk 7 above)
Comment (Benehiko) [reply cli/command/container/prune.go]: I'm essentially reverting the changes from this commit https://github.com/docker/cli/commit/69e0f53a032a85c79e456724df0625eb51cbb16e. 

Propagating the error up to where the cobra root command is executed does not necessarily mean we will output the error to the user. Here I've introduced logic to check which error is returned (in this case `errdefs.Cancelled`). 
See this change here https://github.com/docker/cli/blob/9bc91a81236b182e0dadb91e0d5d8d223092ab9e/cmd/docker/docker.go#L47-L49

RE: space reclaimed output, yes we should fix that so we are really sure space was reclaimed when printing vs just always printing it when no error occurred.

**Comment by Benehiko** [comment]
```shell
--- FAIL: TestConnectAndWait (0.02s)
    --- FAIL: TestConnectAndWait/connect_goroutine_exits_after_EOF (0.02s)
        socket_test.go:182: timeout hit after 10ms: waiting for connect goroutine to exit
FAIL
```
https://github.com/docker/cli/actions/runs/8390226667/job/22978038463?pr=4939#step:5:81

:thinking:

cli/command/utils.go
Benehiko: ended up returning `errdefs.Cancelled` with a custom error message per command that had the Prompt return `false`. This still allows us to exit with code 0 and atm it won't print the error message. We could opt for printing it if we'd like, but i don't think it's necessary right now.

krissetto (approved):
overall LGTM after some minor nits and what has already been pointed out by the others

cli/command/utils.go
```diff
@@ -75,9 +76,7 @@ func PrettyPrint(i any) string {
 	}
 }
 
-type PromptError error
-
-var ErrPromptTerminated = PromptError(errors.New("prompt terminated"))
+var ErrPromptTerminated = errdefs.Cancelled(errors.New("prompt terminated"))
```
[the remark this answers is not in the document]
Benehiko: I think it's good to have this error variable. We use it in our tests to verify if we really did terminate the prompt. Consumers of the Prompt function can also check against this specific error and it's not standardized as an `errdefs.Cancelled` error which is checked for when the cobra root command `Execute` method returns an error.

Also cannot make this a const since the value of the error is not a const. To make it a const requires a workaround which i think is a bit overkill

cli/command/volume/testdata/volume-prune-no.golden
```diff
@@ -1,2 +1,2 @@
 WARNING! This will remove anonymous local volumes not used by at least one container.
-Are you sure you want to continue? [y/N] 
+Are you sure you want to continue? [y/N]
```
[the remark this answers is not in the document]
Benehiko: This is strange...testing it manually it does end on a new line, but maybe the `gotest.tools/v3/golden` package is stripping some of the characters for some reason.

```shell
⋊> ~/G/docker-cli on prompt-termination ◦ ./build/docker system prune           12:56:53
WARNING! This will remove:
  - all stopped containers
  - all dangling images
  - unused build cache

Are you sure you want to continue? [y/N] ^C
⋊> ~/G/docker-cli on prompt-termination ◦ ./build/docker system prune           12:56:59
WARNING! This will remove:
  - all stopped containers
  - all dangling images
  - unused build cache

Are you sure you want to continue? [y/N] n
```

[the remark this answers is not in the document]
Benehiko: Just to confirm, removing the new line print in the prompt function ends up with this:
```shell
> ~/G/docker-cli on prompt-termination ⨯ bash                                             13:14:15
[benehiko@benehiko docker-cli]$ ./build/docker system prune
WARNING! This will remove:
  - all stopped containers
  - all dangling images
  - unused build cache

Are you sure you want to continue? [y/N] ^C[benehiko@benehiko docker-cli]$
```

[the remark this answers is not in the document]
Benehiko:
```diff
diff --git a/cli/command/utils.go b/cli/command/utils.go
index ae93dbc30a..52cb98bd4a 100644
--- a/cli/command/utils.go
+++ b/cli/command/utils.go
@@ -123,7 +123,7 @@ func PromptForConfirmation(ctx context.Context, ins io.Reader, outs io.Writer, m
        select {
        case <-notifyCtx.Done():
                // print a newline on termination
-               fmt.Fprintln(outs, "")
+               // fmt.Fprintln(outs, "")
                return false, ErrPromptTerminated
        case r := <-result:
                return r, nil
diff --git a/cli/command/volume/testdata/volume-prune-terminate.golden b/cli/command/volume/testdata/volume-prune-terminate.golden
index 8e918f9c9e..9f7c8c76a5 100644
--- a/cli/command/volume/testdata/volume-prune-terminate.golden
+++ b/cli/command/volume/testdata/volume-prune-terminate.golden
@@ -1,2 +1,2 @@
 WARNING! This will remove anonymous local volumes not used by at least one container.
-Are you sure you want to continue? [y/N]
+Are you sure you want to continue? [y/N]
\ No newline at end of file
```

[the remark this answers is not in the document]
Benehiko: Oh my, you're going to laugh. This is correct :)
https://github.com/docker/cli/blob/9bc91a81236b182e0dadb91e0d5d8d223092ab9e/cli/command/volume/prune_test.go#L168-L178

We don't add carriage returns to test inputs :see_no_evil: 

The reason we only picked this up now is because in the old implementation of the `volume prune` command we used to always print a new line inside the cobra command run function for this command. 

https://github.com/docker/cli/blob/master/cli/command/volume/prune.go#L36

Since `runPrune` would return an error on prompt error or `n`.
https://github.com/docker/cli/blob/master/cli/command/volume/prune.go#L85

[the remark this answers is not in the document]
laurazard: So, should we change the test here to add a newline?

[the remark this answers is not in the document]
Benehiko: We don't do it in any of the other tests for the prompt input. I'd prefer not to change it.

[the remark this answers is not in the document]
laurazard: But now, does anything happen if there is just an `n` without a newline? `PromptForConfirmation` doesn't return if the user only writes an `n`, right?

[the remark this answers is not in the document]
Benehiko: In the test it creates a reader with the value already. So when prompt is called, the value is immediately read. The behavior here is not the same as `stdin`.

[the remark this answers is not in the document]
laurazard: I see, thanks!

Benehiko:
merge!

laurazard:
@thaJeztah care to take a final look at this?

e2e/testutils/plugins/basic.go
```diff
@@ -9,6 +9,11 @@ import (
 	"time"
 )
 
+// set by the compile flags to get around content sha being the same
+var (
+	UNIQUEME string
+)
+
```
thaJeztah, line 16: This one doesn't seem to be used anywhere, so could lead to a linter starting to flag it.

Possibly we could avoid that through something silly like `_ = UNIQUEME`, but slightly curious what the issue was that you ran into; did this cause issues in tests where Go was caching things? Or was there something else?

cli/command/system/prune.go
```diff
@@ -75,9 +77,13 @@ func runPrune(ctx context.Context, dockerCli command.Cli, options pruneOptions)
 		return fmt.Errorf(`ERROR: The "until" filter is not supported with "--volumes"`)
 	}
 	if !options.force {
-		if r, err := command.PromptForConfirmation(ctx, dockerCli.In(), dockerCli.Out(), confirmationMessage(dockerCli, options)); !r || err != nil {
+		r, err := command.PromptForConfirmation(ctx, dockerCli.In(), dockerCli.Out(), confirmationMessage(dockerCli, options))
+		if err != nil {
 			return err
 		}
+		if !r {
+			return errdefs.Cancelled(errors.New("`system prune` has been cancelled"))
```
thaJeztah, line 85: Can you remove the back-tics here (and in the other error messages?) I guess it's a bit of a habit to use these (as it's how you'd use them in MarkDown), but on the command-line they don't really do anything 😅

cli/command/utils.go
```diff
@@ -123,6 +122,8 @@ func PromptForConfirmation(ctx context.Context, ins io.Reader, outs io.Writer, m
 
 	select {
 	case <-notifyCtx.Done():
+		// print a newline on termination
+		fmt.Fprintln(outs, "")
```
thaJeztah, line 126: As this is new code, might as well please the linters;

```suggestion
		// print a newline on termination
		_, _ = fmt.Fprintln(outs, "")
```

thaJeztah (changes requested):
Thanks! Overall looks good; had a question about that magic `UNIQUEME` variable, and noticed some back-tics in the errors.


Honestly (but not for this PR), looking at `PromptForConfirmation`, I'm slightly wondering  if it's doing _too much_ https://github.com/docker/cli/blob/155dc5e4e406f935965260ac45524f950df0d920/cli/command/utils.go#L92-L98

Wondering if;

- haven't checked, but wondered if there's cases where `No` is _not_ the default (so if it should take a default value), but maybe that's not the case.
- :point_up: that said, the `[y/N]` (capital `N`) indicates that `N` is always considered the default; so even if a caller _would_ want to use a different default, the function appending the `[y/N]` means that there's not really a choice for them.
- we should remove the default message (looks like the only place where we pass an empty string is in a test)
- was even considering if we need the `message` at all, because all it does is print the message (unconditionally), and if that should be left to the caller;
- from that perspective I was looking at the `outs io.Writer` and if we need it, but that one looks more tricky, as we still want to print the newline 😓

e2e/testutils/plugins/basic.go
Benehiko: The e2e tests build a plugin and pushes that to a registry run inside docker (spun up by the tests). That said, for multiple tests doing `docker plugin create` and `docker plugin push` you would end up with sha collisions since the source code of the plugin binary built would be the same. The `UNIQUEME` variable is a way around this, so we can re-build the same plugin binary from the same source code generating a new sha hash. 

See these tests for reference https://github.com/docker/cli/blob/7c722c08d093c118ea727961be9aa0a23c83b733/e2e/global/cli_test.go#L143-L181

thaJeztah: Oh! And now curious; wouldn't that be something we could take advantage of? If we're building (and pushing) the exact same plugin multiple times; would we be able to use that so that we don't have to? 🤔 


Wondering now; ISTR we have something similar in moby/moby, where we need to have an image that's built and pushed in CI (but have something to verify if it needs to be built / built only once). I'd have to go look for that though 🙈

Benehiko:
> Thanks! Overall looks good; had a question about that magic `UNIQUEME` variable, and noticed some back-tics in the errors.
> 
> Honestly (but not for this PR), looking at `PromptForConfirmation`, I'm slightly wondering if it's doing _too much_
> 
> https://github.com/docker/cli/blob/155dc5e4e406f935965260ac45524f950df0d920/cli/command/utils.go#L92-L98
> 
> Wondering if;
> 
>     * haven't checked, but wondered if there's cases where `No` is _not_ the default (so if it should take a default value), but maybe that's not the case.
> 
>     * ☝️ that said, the `[y/N]` (capital `N`) indicates that `N` is always considered the default; so even if a caller _would_ want to use a different default, the function appending the `[y/N]` means that there's not really a choice for them.
> 
>     * we should remove the default message (looks like the only place where we pass an empty string is in a test)
> 
>     * was even considering if we need the `message` at all, because all it does is print the message (unconditionally), and if that should be left to the caller;
> 
>     * from that perspective I was looking at the `outs io.Writer` and if we need it, but that one looks more tricky, as we still want to print the newline 😓

I agree that the `PromptForConfirmation` does quite a bit. I would like to also move the code out that catches the os.Signal to somewhere in the main function so we have a top-level way of handling `SIGTERM` and `SIGINT` - like we do when executing plugins. But since this has already been quite a big refactor in this PR, I'd like to do this iteratively. 

As for default message and default accepted value (`N`), I kept it as it originally was. The function is explicit that it only asks for a confirmation (yes / no) and does a fallback to `no` if something other than `yes` was submitted. I could clean this up more, but I wouldn't want to deviate from the intent of the PR - which is to standardize the exit codes for commands that prompt the user for confirmation. This also standardizes output messages when cancelling the prompt with `N/n` or `SIGINT/SIGTERM`.

e2e/testutils/plugins/basic.go
Benehiko: Not necessarily since there is also a test that verifies if it _can_ push the image

https://github.com/docker/cli/blob/master/e2e/plugin/trust_test.go#L37-L46

laurazard:
Re:

> I would like to also move the code out that catches the os.Signal to somewhere in the main function so we have a top-level way of handling SIGTERM and SIGINT - like we do when executing plugins.

I support this, but quick reminder that as we learned with changing signal handling for plugins, these things aren't easy to get right – and we ended up reverting some of the changes/only setting up our own signal handling for plugins when the CLI isn't attached to a terminal. I think we should have some general `signal.NotifyContext` (or more complex version thereof) and pipe this down into each individual command where it can be handled as appropriate (for example, if a context gets cancelled during a prompt that should be handled differently than during normal command execution), but we must make sure to pipe things into the correct places and not break plugins/places that expect to receive signals.

e2e/testutils/plugins/basic.go
thaJeztah: Yes, definitely fine to look at separately

thaJeztah (approved):
LGTM, thanks!

PR #4940, opened 2024-03-13
Initial otel implementation

krissetto:
**- What's this?**  
OTEL bits in the CLI

**- Goal**  

The very first goal is to provide enough metrics to cover the  `compose wrapper`'s current usage (so we can then remove it).  
To keep continuity with the current impl in the `compose wrapper`, these should include at minimum:
- command name;
- command duration;
- command status ("success"/"failure");
- command exit code

Pls note that the organization of the code is **not** final and will surely be adjusted as the implementation matures

**- What I did**  

- Added some utils for otel related ops
- Implemented an initial version of otel metrics using the utils ^

**- How I did it**  

Taking inspiration and adapting bits from `buildx` and the `compose wrapper`

**- How to verify it**  

Manually. Automated tests still need to be defined 

### DIFF
--- a/cli/command/telemetry_docker.go
+++ b/cli/command/telemetry_docker.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"os"
 	"path"
 
 	"github.com/pkg/errors"
@@ -14,7 +15,11 @@ import (
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 )
 
-const otelContextFieldName = "otel"
+const (
+	otelContextFieldName     string = "otel"
+	otelExporterOTLPEndpoint string = "OTEL_EXPORTER_OTLP_ENDPOINT"
+	debugEnvVarPrefix        string = "DOCKER_CLI_"
+)
 
 // dockerExporterOTLPEndpoint retrieves the OTLP endpoint used for the docker reporter
 // from the current context.
@@ -49,8 +54,15 @@ func dockerExporterOTLPEndpoint(cli Cli) (endpoint string, secure bool) {
 	}
 
 	// keys from https://opentelemetry.io/docs/concepts/sdk-configuration/otlp-exporter-configuration/
-	endpoint, ok = otelMap["OTEL_EXPORTER_OTLP_ENDPOINT"].(string)
-	if !ok {
+	endpoint, _ = otelMap[otelExporterOTLPEndpoint].(string)
+
+	// Override with env var value if it exists AND IS SET
+	// (ignore otel defaults for this override when the key exists but is empty)
+	if override := os.Getenv(debugEnvVarPrefix + otelExporterOTLPEndpoint); override != "" {
+		endpoint = override
+	}
+
+	if endpoint == "" {
 		return "", false
 	}
 
--- a/cli/command/telemetry_utils.go
+++ b/cli/command/telemetry_utils.go
@@ -1,0 +1,158 @@
+package command
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/docker/cli/cli/version"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+// BaseMetricAttributes returns an attribute.Set containing attributes to attach to metrics/traces
+func BaseMetricAttributes(cmd *cobra.Command) attribute.Set {
+	attrList := []attribute.KeyValue{
+		attribute.String("command.name", getCommandName(cmd)),
+	}
+	return attribute.NewSet(attrList...)
+}
+
+// InstrumentCobraCommands wraps all cobra commands' RunE funcs to set a command duration metric using otel.
+//
+// Note: this should be the last func to wrap/modify the PersistentRunE/RunE funcs before command execution.
+//
+// can also be used for spans!
+func InstrumentCobraCommands(cmd *cobra.Command, mp metric.MeterProvider) {
+	meter := getDefaultMeter(mp)
+	// If PersistentPreRunE is nil, make it execute PersistentPreRun and return nil by default
+	ogPersistentPreRunE := cmd.PersistentPreRunE
+	if ogPersistentPreRunE == nil {
+		ogPersistentPreRun := cmd.PersistentPreRun
+		//nolint:unparam // necessary because error will always be nil here
+		ogPersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+			ogPersistentPreRun(cmd, args)
+			return nil
+		}
+		cmd.PersistentPreRun = nil
+	}
+
+	// wrap RunE in PersistentPreRunE so that this operation gets executed on all children commands
+	cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		// If RunE is nil, make it execute Run and return nil by default
+		ogRunE := cmd.RunE
+		if ogRunE == nil {
+			ogRun := cmd.Run
+			//nolint:unparam // necessary because error will always be nil here
+			ogRunE = func(cmd *cobra.Command, args []string) error {
+				ogRun(cmd, args)
+				return nil
+			}
+			cmd.Run = nil
+		}
+		cmd.RunE = func(cmd *cobra.Command, args []string) error {
+			// start the timer as the first step of every cobra command
+			stopCobraCmdTimer := startCobraCommandTimer(cmd, meter)
+			cmdErr := ogRunE(cmd, args)
+			stopCobraCmdTimer(cmdErr)
+			return cmdErr
+		}
+
+		return ogPersistentPreRunE(cmd, args)
+	}
+}
+
+func startCobraCommandTimer(cmd *cobra.Command, meter metric.Meter) func(err error) {
+	ctx := cmd.Context()
+	baseAttrs := BaseMetricAttributes(cmd)
+	durationCounter, _ := meter.Float64Counter(
+		"command.time",
+		metric.WithDescription("Measures the duration of the cobra command"),
+		metric.WithUnit("ms"),
+	)
+	start := time.Now()
+
+	return func(err error) {
+		duration := float64(time.Since(start)) / float64(time.Millisecond)
+		cmdStatusAttrs := attributesFromError(err)
+		durationCounter.Add(ctx, duration,
+			metric.WithAttributeSet(baseAttrs),
+			metric.WithAttributeSet(attribute.NewSet(cmdStatusAttrs...)),
+		)
+	}
+}
+
+func attributesFromError(err error) []attribute.KeyValue {
+	attrs := []attribute.KeyValue{}
+	exitCode := 0
+	if err != nil {
+		exitCode = 1
+		if stderr, ok := err.(statusError); ok {
+			// StatusError should only be used for errors, and all errors should
+			// have a non-zero exit status, so only set this here if this value isn't 0
+			if stderr.StatusCode != 0 {
+				exitCode = stderr.StatusCode
+			}
+		}
+		attrs = append(attrs, attribute.String("command.error.type", otelErrorType(err)))
+	}
+	attrs = append(attrs, attribute.String("command.status.code", strconv.Itoa(exitCode)))
+
+	return attrs
+}
+
+// otelErrorType returns an attribute for the error type based on the error category.
+func otelErrorType(err error) string {
+	name := "generic"
+	if errors.Is(err, context.Canceled) {
+		name = "canceled"
+	}
+	return name
+}
+
+// statusError reports an unsuccessful exit by a command.
+type statusError struct {
+	Status     string
+	StatusCode int
+}
+
+func (e statusError) Error() string {
+	return fmt.Sprintf("Status: %s, Code: %d", e.Status, e.StatusCode)
+}
+
+// getCommandName gets the cobra command name in the format
+// `... parentCommandName commandName` by traversing it's parent commands recursively.
+// until the root command is reached.
+//
+// Note: The root command's name is excluded. If cmd is the root cmd, return ""
+func getCommandName(cmd *cobra.Command) string {
+	fullCmdName := getFullCommandName(cmd)
+	i := strings.Index(fullCmdName, " ")
+	if i == -1 {
+		return ""
+	}
+	return fullCmdName[i+1:]
+}
+
+// getFullCommandName gets the full cobra command name in the format
+// `... parentCommandName commandName` by traversing it's parent commands recursively
+// until the root command is reached.
+func getFullCommandName(cmd *cobra.Command) string {
+	if cmd.HasParent() {
+		return fmt.Sprintf("%s %s", getFullCommandName(cmd.Parent()), cmd.Name())
+	}
+	return cmd.Name()
+}
+
+// getDefaultMeter gets the default metric.Meter for the application
+// using the given metric.MeterProvider
+func getDefaultMeter(mp metric.MeterProvider) metric.Meter {
+	return mp.Meter(
+		"github.com/docker/cli",
+		metric.WithInstrumentationVersion(version.Version),
+	)
+}
--- a/cli/command/telemetry_utils_test.go
+++ b/cli/command/telemetry_utils_test.go
@@ -1,0 +1,94 @@
+package command
+
+import (
+	"testing"
+
+	"github.com/spf13/cobra"
+	"gotest.tools/v3/assert"
+)
+
+func setupCobraCommands() (*cobra.Command, *cobra.Command, *cobra.Command) {
+	rootCmd := &cobra.Command{
+		Use: "root [OPTIONS] COMMAND [ARG...]",
+	}
+	childCmd := &cobra.Command{
+		Use: "child [OPTIONS] COMMAND [ARG...]",
+	}
+	grandchildCmd := &cobra.Command{
+		Use: "grandchild [OPTIONS] COMMAND [ARG...]",
+	}
+	childCmd.AddCommand(grandchildCmd)
+	rootCmd.AddCommand(childCmd)
+
+	return rootCmd, childCmd, grandchildCmd
+}
+
+func TestGetFullCommandName(t *testing.T) {
+	rootCmd, childCmd, grandchildCmd := setupCobraCommands()
+
+	t.Parallel()
+
+	for _, tc := range []struct {
+		testName string
+		cmd      *cobra.Command
+		expected string
+	}{
+		{
+			testName: "rootCmd",
+			cmd:      rootCmd,
+			expected: "root",
+		},
+		{
+			testName: "childCmd",
+			cmd:      childCmd,
+			expected: "root child",
+		},
+		{
+			testName: "grandChild",
+			cmd:      grandchildCmd,
+			expected: "root child grandchild",
+		},
+	} {
+		tc := tc
+		t.Run(tc.testName, func(t *testing.T) {
+			t.Parallel()
+			actual := getFullCommandName(tc.cmd)
+			assert.Equal(t, actual, tc.expected)
+		})
+	}
+}
+
+func TestGetCommandName(t *testing.T) {
+	rootCmd, childCmd, grandchildCmd := setupCobraCommands()
+
+	t.Parallel()
+
+	for _, tc := range []struct {
+		testName string
+		cmd      *cobra.Command
+		expected string
+	}{
+		{
+			testName: "rootCmd",
+			cmd:      rootCmd,
+			expected: "",
+		},
+		{
+			testName: "childCmd",
+			cmd:      childCmd,
+			expected: "child",
+		},
+		{
+			testName: "grandchildCmd",
+			cmd:      grandchildCmd,
+			expected: "child grandchild",
+		},
+	} {
+		tc := tc
+		t.Run(tc.testName, func(t *testing.T) {
+			t.Parallel()
+			actual := getCommandName(tc.cmd)
+			assert.Equal(t, actual, tc.expected)
+		})
+	}
+}
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
@@ -21,17 +22,19 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"go.opentelemetry.io/otel"
 )
 
 func main() {
-	dockerCli, err := command.NewDockerCli()
+	ctx := context.Background()
+	dockerCli, err := command.NewDockerCli(command.WithBaseContext(ctx))
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
 	logrus.SetOutput(dockerCli.Err())
 
-	if err := runDocker(dockerCli); err != nil {
+	if err := runDocker(ctx, dockerCli); err != nil {
 		if sterr, ok := err.(cli.StatusError); ok {
 			if sterr.Status != "" {
 				fmt.Fprintln(dockerCli.Err(), sterr.Status)
@@ -286,7 +289,7 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string, 
 }
 
 //nolint:gocyclo
-func runDocker(dockerCli *command.DockerCli) error {
+func runDocker(ctx context.Context, dockerCli *command.DockerCli) error {
 	tcmd := newDockerCommand(dockerCli)
 
 	cmd, args, err := tcmd.HandleGlobalFlags()
@@ -297,6 +300,11 @@ func runDocker(dockerCli *command.DockerCli) error {
 	if err := tcmd.Initialize(); err != nil {
 		return err
 	}
+
+	mp := dockerCli.MeterProvider(ctx)
+	defer mp.Shutdown(ctx)
+	otel.SetMeterProvider(mp)
+	command.InstrumentCobraCommands(cmd, mp)
 
 	var envs []string
 	args, os.Args, envs, err = processAliases(dockerCli, cmd, args, os.Args)


cli/command/telemetry.go
```diff
@@ -0,0 +1,214 @@
...
+func (r *telemetryResource) init() {
+	if r.res != nil {
+		r.opts = nil
+		return
+	}
+
+	opts := r.defaultOptions()
+	if len(r.opts) > 0 {
+		opts = append(opts, r.opts...)
+	}
```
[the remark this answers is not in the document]
Benehiko: even though r.opts might be `nil` - this is fine, the underlying type is a slice of resource.Options and append will deal with it just fine.

cli/command/telemetry_docker.go
```diff
@@ -0,0 +1,121 @@
...
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+const otelConfigFieldName = "otel"
+
+// dockerExporterOtlpEndpoint retrieves the OTLP endpoint used for the docker reporter
+// from the current context.
+func dockerExporterOtlpEndpoint(cli Cli) (endpoint string, secure bool, err error) {
```
Benehiko, line 20: isn't OTLP an acronym? in Go acronyms are always all caps

 ```suggestion
func dockerExporterOTLPEndpoint(cli Cli) (endpoint string, secure bool, err error) {
```

File: cli/command/telemetry_utils.go [hunk 3]
```diff
@@ -0,0 +1,215 @@
...
+		attrList = append(attrList, attribute.String("command.name", command))
+	}
+
+	return attribute.NewSet(attrList...)
+}
+
+func mapExitCode(exitCode int) string {
+	result := ""
+
+	switch exitCode {
```
Benehiko, line 155: will we have more exit codes mapped here with different results? else we could just us an if statement like so

```go
if exitCode == 0 {
  return "success"
}

return "failure"
```

cmd/docker/docker.go
```diff
@@ -290,9 +294,15 @@ func runDocker(dockerCli *command.DockerCli) error {
 		return err
 	}
 
+	mp := dockerCli.MeterProvider(ctx)
+	defer mp.Shutdown(ctx)
+	baseOtelAttributes := command.BaseMetricAttributes()
+	stopCmdTimer := command.TimeCommand(ctx, mp, baseOtelAttributes)
```
Benehiko, line 300: why do we need to call `stopCmdTimer` so many times inside `runDocker`? wouldn't there be a way to move this further up in the call stack and then call this function once inside the `main` function?

Benehiko (changes requested):

cmd/docker/docker.go
krissetto: I also think this is quite ugly. I left a code comment in the main() func regarding this 😛 so we can discuss the approach. 
  
Basically, the `MeterProvider` func currently needs the `dockerCli` struct to be `Initialize`d, and that only happens in here. 

The reason is that the `dockerExporterOtlpEndpoint()` func uses the dockerCli ContextStore to get the endpoint value:
`meta, err := cli.ContextStore().GetMetadata(cli.CurrentContext())`

cli/command/telemetry_utils.go
krissetto: Currently there is no need for multiple values, but we definitely could be adding more. I can change this for now

cli/command/telemetry.go
```diff
@@ -0,0 +1,214 @@
...
+func (cli *DockerCli) TracerProvider(ctx context.Context, opts ...sdktrace.TracerProviderOption) TracerProvider {
+	allOpts := make([]sdktrace.TracerProviderOption, 0, len(opts)+2)
+	allOpts = append(allOpts, sdktrace.WithResource(cli.Resource()))
+	if exp, err := dockerSpanExporter(ctx, cli); err != nil {
+		otel.Handle(err)
+	} else if exp != nil {
+		allOpts = append(allOpts, sdktrace.WithBatcher(exp))
+	}
+	allOpts = append(allOpts, opts...)
+	return sdktrace.NewTracerProvider(allOpts...)
```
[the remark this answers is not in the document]
krissetto: pls ignore any changes in the telemetry/telemetry_docker files in this PR. The commit is taken from #4889, and here I only really wanted to use those changes to wire things up. Let's discuss changes to these files in that other PR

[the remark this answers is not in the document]
Benehiko: ahh okay, will do! :+1:

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/pkg/errors"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+// -------- Some utils, mostly adapted from compose-cli-private and buildx for testing purposes --------
+
+// Ugly whitelist of commands and flags we want to be able to track in our metrics adapted from compose-cli-private.
+// A better solution is likely necessary here, as these lists are mostly generated in that project
+// (https://github.com/docker/compose-cli-private/blob/e324c1a842256efe6fb8d1148e94cdb43cb03df1/cli/metrics/generatecommands/main.go#L17)
```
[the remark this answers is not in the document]
krissetto: Totally agree with this point. I replicated what was in the wrapper so we could discuss some base strategy that we can then use across plugins as well, just so we're more aware of where the implementations are diverging and why. Also, if we can create something others can reuse, we should do it here :)

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,215 @@
...
+		result = "success"
+	default:
+		result = "failure"
+	}
+
+	return result
+}
+
+// statusError reports an unsuccessful exit by a command.
+type statusError struct {
```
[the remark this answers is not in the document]
krissetto: No and i have no idea why i didn't find it 😂

cmd/docker/docker.go
```diff
@@ -25,14 +26,17 @@ import (
...
+	dockerCli, err := command.NewDockerCli(command.WithBaseContext(ctx))
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
 	logrus.SetOutput(dockerCli.Err())
 
-	if err := runDocker(dockerCli); err != nil {
+	// TODO(krissetto): see if we can setup otel bits here, before
+	// setting up the cobra command by running dockerCli.Initialize() in `runDocker()`
```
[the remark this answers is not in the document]
krissetto: I don't mind the idea of wrapping the command too much, but honestly i'm not totally sure about it. 

My reasoning was that given how many special edge cases etc exist in this codebase, I would quite like for the otel init to happen "before everything else", so that we don't hamstring ourselves too much in terms of what we can measure end to end.

Regardless of where we do the init of the timer, on the cobra side what i would still do (and what i'm currently testing out a bit) is wrapping the commands with attributes, potentially another timer, etc

cli/command/telemetry_utils.go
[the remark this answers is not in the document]
laurazard: It happens (to me, a lot)! I just saw that and was like "wait, I've seen this before" 😅

cli/command/telemetry_utils.go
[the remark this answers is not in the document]
krissetto: I removed that code for now since in the first iteration we only want to get the command name

cli/command/telemetry_utils.go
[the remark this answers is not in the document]
krissetto: Using this there would be a circular import between cli/command pkgs 😞 
I'll handle that after the rest

cmd/docker/docker.go
krissetto: I changed the impl to be much more similar to the compose version you linked. I still think there'd be a good amount of value in having metrics from the very beginning to the very end of the code execution, but we can always handle that in a second moment. 

Let me know what you think 😄

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,156 @@
...
+		if ogRunE == nil {
+			ogRun := cmd.Run
+			ogRunE = func(cmd *cobra.Command, args []string) error {
+				ogRun(cmd, args)
+				return nil
+			}
+			cmd.Run = nil
+		}
+		cmd.RunE = func(cmd *cobra.Command, args []string) error {
+			// insert the meter and tracer objects into the context
```
krissetto, line 61: @laurazard @Benehiko 

As an extra little test thing we can discuss,  I added the otel `Meter` to the cobra command context as a value, so that commands can retrieve it and use it in their RunE funcs (for example for adding command specific data). We can then do the same thing for tracing if it's useful. What do you think?

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,156 @@
...
+			// so commands can use them in their RunE func
+			newCtx := context.WithValue(cmd.Context(), OtelMeterCtxKey, meter)
+			// newCtx := context.WithValue(newCtx, OtelTracerCtxKey, tracer))
+			cmd.SetContext(newCtx)
```
krissetto, line 65: @laurazard @Benehiko 

As an extra little test thing we can discuss,  I added the otel `Meter` to the cobra command context as a value, so that commands can retrieve it and use it in their RunE funcs (for example for adding command specific data). We can then do the same thing for tracing if it's useful. What do you think?

Benehiko: I think it depends, using `ctx` to store values could make using `meter` clunky. Also adding it here to the cmd ctx inside the `RunE/Run` function might not be necessary. You can add it way earlier where you create the meter instance to the top level context and execute cobra with the `ExecuteContext` method https://github.com/spf13/cobra/blob/v1.8.0/command.go#L1030

e.g. `meter, ok := ctx.Value(OtelMeterCtxKey).(metric.Meter)`

Just a thought but couldn't we just add the meter to the dockerCli instance and have a nice method for using it? `dockerCli.Meter()`

Benehiko: To expand on the `ctx` approach, we could also create standalone utility functions which help with using the meter,

```go
package utils

func Meter(ctx context.Context) (metric.Meter, bool) {
  return ctx.Value(OtelMeterCtxKey).(metric.Meter)
}
```

```go
meter, ok := utils.Meter(ctx)
if !ok {
  return
}
```

krissetto: > You can add it way earlier

So my thinking was to only include it in the `RunE` context, and to do so when instrumenting the cobra command so most of these otel things happened in the same place, so it seemed like a good place to put it for now. No hard feelings here and I can move it to wherever we think is best (or even remove it altogether).

> Just a thought but couldn't we just add the meter to the dockerCli instance and have a nice method for using it? dockerCli.Meter()

I briefly thought about it, but I'd really prefer to avoid adding more things to that struct unless we're totally sure it's the right place (also since the cli itself - including that struct - is used in other projects as well). 
I'm actually slightly more intrigued by the idea of moving some data from dockerCli to the context, but that for a whole other discussion :) 

> To expand on the ctx approach, we could also create standalone utility functions which help with using the meter

Totally agree with this! I left a util out just for the sake of discussing the approach first

Benehiko: > I briefly thought about it, but I'd really prefer to avoid adding more things to that struct unless we're totally sure it's the right place (also since the cli itself - including that struct - is used in other projects as well).
I'm actually slightly more intrigued by the idea of moving some data from dockerCli to the context, but that's for a whole other discussion :)

Good point!  I'm not a big fan of values inside context just for the sake of it. To be honest I would actually want a singleton that contains all of these init instances necessary for the CLI to function (like otel) and then expose the utility of them through an interface that's passed around. Then you don't need to use `ctx` with lots of utility functions.

```go
package main

import "context"

type Singleton interface {
	command.Cli
	Meter()
}

type singleton struct {
	meter metric.Meter
	dockerCli command.DockerCli
}

func (s *singleton) DoSomething() {
}

func main() {
	s := &singleton{
		meter: metric.NewMeter(),
		dockerCli: command.NewDockerCli(),
	}
	NeedSingleton(ctx, s)
}

func NeedSingleton(ctx context.Context, s Singleton) {
	s.Meter()
	s.Client()
}

func NeedPartialSingleton(ctx context.Context, s interface {
	Meter()
}) {
	s.Meter()
}
```

krissetto: I think that was _kinda_ the reasoning behind the `DockerCli` struct itself. I'm not sure this added complexity is something we should deal with right now, and I'd focus on getting the minimum functionality we need (so whatever metrics the compose wrapper currently handles, so we can remove it). I don't mind the suggested approach, so let's keep it in mind if others agree.

Even adding the meter to the context is essentially overkill right now, as it's not actually needed for anything (but will eventually be once we want to measure things during the execution of commands)

Benehiko: oh yeah i agree, now isn't the right time for this :)
I'd be fine to keep the meter in the context for now if it means we an ship this.

cli/command/telemetry_utils.go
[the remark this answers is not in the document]
laurazard: 👍

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,156 @@
...
+
+// getFullCommandName gets the full cobra command name in the format
+// `... parentCommandName commandName` by traversing it's parent commands recursively.
+func getFullCommandName(cmd *cobra.Command) string {
+	cmdName := cmd.Name()
+	if cmd.HasParent() {
+		return fmt.Sprintf("%s %s", getFullCommandName(cmd.Parent()), cmdName)
+	}
+	return cmdName
+}
```
laurazard, line 156: Didn't get the time to run this, does this basically do the same thing as Compose is doing here - https://github.com/docker/compose/blob/759d3bc9f9cc401f54753f2384be63aef3b0b2cc/cmd/cmdtrace/cmd_span.go#L125-L135 ?

If so, maybe we can copy-paste it, and the tests (and then maybe later they can reuse this imported from the CLI).

krissetto: It's similar, but it also excludes "docker" (as the root command) and "compose" (as a special case) from the result. I can remove the "docker" (or whatever the root cmd is) part from the result, but i'd honestly leave the "compose" one. 

If we want to share it i could add the possibility of filtering out specific names, but maybe that should just be done case-by-case by whoever wants to exclude specific parts from the output, WDYT?

laurazard: Right, I was thinking that we would exclude `docker` since these are already metrics from the `docker` cli (and we have other attributes to know that).
My logic is that, the same way that their "root command" is `docker compose` and that part is excluded, our root command is `docker` and we could exclude that. But I also don't know how that lines up with parity vs what we had from the wrapper.

Writing some tests for this would also make the expected output of this function easier to find, I'd appreciate it.

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,156 @@
...
+				// StatusError should only be used for errors, and all errors should
+				// have a non-zero exit status, so only set this here if this value isn't 0
+				if stderr.StatusCode != 0 {
+					exitCode = stderr.StatusCode
+				}
+			}
+			// TODO(krissetto): this attribute might not be needed, but buildx has it. Check if it's necessary
+			extraAttrs = append(extraAttrs, attribute.String("command.error.type", otelErrorType(err)))
+		}
+		extraAttrs = append(extraAttrs, attribute.String("command.status", mapExitCode(exitCode)))
```
laurazard, line 109: do these match up with https://github.com/docker/compose/blob/759d3bc9f9cc401f54753f2384be63aef3b0b2cc/cmd/cmdtrace/cmd_span.go#L94-L96?

Or are these different because metrics vs spans?

cli/command/telemetry_utils.go
krissetto: Not an issue, "docker" will be excluded. 

Regarding compose though, `docker compose` is not their root command and they have specific logic to check if the command name they've reached is == to the plugin name (a const they have). It works for them, but i'm not a huge fan of the approach tbh, as it limits how commands can be named if they aren't the root command (essentially a command cannot have the same name as any of it's parents, or the logic to get the command name will fail). They also have no specific tests for the name retrieval code.

For now I'll push some changes to this code (along with a couple basic unit tests) so that the "first" name (so that of the root command) is excluded, and if you agree we can decide later if it makes sense to also add more logic so that other projects can reuse this (quite basic) code

laurazard: > Regarding compose though, docker compose is not their root command and they have specific logic to check if the command name they've reached is == to the plugin name (a const they have)

Yes, by "root command" (see I had the "s) I meant that that's it's the base command for Docker Compose, as in, the way people use Compose and the command that holds all of their commands. I'd also argue that in general it would be bad to have any command have the same name as it's parent (`docker compose compose`? `docker container container`?) so I wouldn't worry about that.

laurazard: I just took a look at the changes though, looks good :)

cli/command/telemetry_utils_test.go
```diff
@@ -0,0 +1,60 @@
...
+	assert.Equal(t, actual, expected)
+
+	expected = "root child"
+	actual = getFullCommandName(childCmd)
+	assert.Equal(t, actual, expected)
+
+	expected = "root"
+	actual = getFullCommandName(rootCmd)
+	assert.Equal(t, actual, expected)
+}
```
laurazard, line 34: Can we do some table tests here?

cli/command/telemetry_utils_test.go
```diff
@@ -0,0 +1,60 @@
...
+	actual := getCommandName(grandchildCmd)
+	assert.Equal(t, actual, expected)
+
+	expected = "child"
+	actual = getCommandName(childCmd)
+	assert.Equal(t, actual, expected)
+
+	expected = ""
+	actual = getCommandName(rootCmd)
+	assert.Equal(t, actual, expected)
```
laurazard, line 59: same here

cli/command/telemetry_utils.go
laurazard: Sounds good!

laurazard:
@jsternberg's PR https://github.com/docker/cli/pull/4889 just got merged 🥳 

Can you rebase and see if everything's fine @krissetto?

laurazard:
Just so I don't forget this, can you get code into `telemetry.go` to accept a `OTEL_EXPORTER_OTLP_ENDPOINT` in this PR too @krissetto?

cli/command/telemetry_docker.go
```diff
@@ -49,8 +51,15 @@ func dockerExporterOTLPEndpoint(cli Cli) (endpoint string, secure bool) {
 	}
 
 	// keys from https://opentelemetry.io/docs/concepts/sdk-configuration/otlp-exporter-configuration/
-	endpoint, ok = otelMap["OTEL_EXPORTER_OTLP_ENDPOINT"].(string)
-	if !ok {
+	endpoint, _ = otelMap[otelExporterOTLPEndpoint].(string)
+
+	// Override with env var value if it exists AND IS SET
+	// (ignore otel defaults for this override when the key exists but is empty)
```
krissetto, line 57: @laurazard 
I opted to not override the value here if the env var key exists but has no value, to avoid confusion with otel defaulting it's own values

laurazard: sounds good to me!

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,170 @@
...
+
+		return ogPersistentPreRunE(cmd, args)
+	}
+}
+
+func startCobraCommandTimer(cmd *cobra.Command) func(err error) {
+	ctx := cmd.Context()
+	meter, ok := ctx.Value(OtelMeterCtxKey).(metric.Meter)
+	if meter == nil || !ok {
+		// TODO(krissetto): Handle this better
```
laurazard, line 82: Should do something here – either call `otel.Handle` or have a debug log.

krissetto: Whoops, forgot that, thx. 
Since this is a private func used only during instrumentation of the cobra command, I opted to `otel.Handle` it and return a no-op func (since it's called after the execution of the cmd's RunE func). WDYT?

jsternberg: I think `otel.Handle` is good. For buildx, we overwrite the handler to print nothing. In the future, I was going to add a similar handler that swallowed errors by default for the CLI and printed errors when an environment variable was set for debugging. That debugging behavior is missing from buildx.

It's mostly the reason why I've been using `otel.Handle` so all otel errors can be handled in the same way.

cli/command/telemetry_docker.go
jsternberg: I'd prefer if we use a different environment variable for this because of the OTEL one.

My original goal with this was to do two things:

1. Natively support the OTEL environment variables and implement it so telemetry both goes to this endpoint and to a user defined endpoint.
2. Add a custom environment variable to override the docker configuration.

Buildx currently implements 1 but I didn't want to overly complicate the original PR with that logic.

If we can prefix the environment variable with something like `DOCKER` or `DOCKER_CLI` or more preferably `DOCKER_CLI_DEBUG` I'd like that a lot better for the environment variable name.

The original environment variable has a specific usage in OTEL that's different than how it's used here and I don't want us to accidentally clobber each other.

cli/command/telemetry_utils.go
jsternberg: I would avoid this context stuff in general. My recommendation is to use [otel.SetTracerProvider](https://pkg.go.dev/go.opentelemetry.io/otel#SetTracerProvider) and [otel.SetMeterProvider](https://pkg.go.dev/go.opentelemetry.io/otel#SetMeterProvider) and then use the corresponding function in code that uses these providers.

For traces, you also pretty rarely need to access the global tracer because the original `TracerProvider` is accessible from [trace.Span](https://pkg.go.dev/go.opentelemetry.io/otel/trace#Span).

A useful pattern that I found is creating a utility similar to this one: https://github.com/docker/buildx/blob/fe08cf2981836f1828058f02cb7684f676b53a9e/util/metricutil/metric.go#L216-L219

In general, you should not pass or keep a meter and instead just the meter provider.

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func startCobraCommandTimer(cmd *cobra.Command) func(err error) {
+	ctx := cmd.Context()
+	meter, ok := ctx.Value(OtelMeterCtxKey).(metric.Meter)
+	if meter == nil || !ok {
+		otel.Handle(errors.New("meter not found in ctx."))
+		// return a no-op if the meter cannot be found in the command's ctx
+		return func(err error) {}
+	}
```
jsternberg, line 86: Given my comment above, my suggestion is to do:

```
meter := otel.GetMeterProvider().Meter("github.com/docker/cli")
durationCounter, _ := meter.Float64Counter(...)
```

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,172 @@
...
+	baseAttrs := BaseMetricAttributes(cmd)
+	durationCounter, _ := meter.Float64Counter(
+		"command.duration",
```
jsternberg, line 89: Otel semantic naming recommends `command.time`.

https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-naming

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,172 @@
...
+			exitCode = 1
+			if stderr, ok := err.(statusError); ok {
+				// StatusError should only be used for errors, and all errors should
+				// have a non-zero exit status, so only set this here if this value isn't 0
+				if stderr.StatusCode != 0 {
+					exitCode = stderr.StatusCode
+				}
+			}
+			// TODO(krissetto): this attribute might not be needed, but buildx has it. Check if it's necessary
+			extraAttrs = append(extraAttrs, attribute.String("command.error.type", otelErrorType(err)))
```
jsternberg, line 109: Can you consolidate this into a function so it's just something like `exitCodeFromError`?

Also none of these attributes are really _needed_, but I liked error type because it makes it potentially easier to categorize types of errors later. In buildx, we're just using it right now to determine if the command was successful, canceled, or anything else.

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		extraAttrs = append(extraAttrs, attribute.String("command.status", mapExitCode(exitCode)))
+		extraAttrs = append(extraAttrs, attribute.String("command.exit_code", strconv.Itoa(exitCode)))
```
jsternberg, line 112: I'd recommend consolidating these into a single value for now and doing something like naming it `command.status.code`. With this more explicit namespacing, it's easy to add a `command.status.text` or something else in the future if we need it. The reason why I say to only use one is it seems like the map exit code can be done more easily and appropriately in the otel collector than here. If you push more logic into the otel collector, it's fewer things to maintain here and less disruptive if we decide to arbitrarily change the text.

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,172 @@
...
+
+// otelErrorType returns an attribute for the error type based on the error category.
+// If nil, this function returns an invalid attribute.
+func otelErrorType(err error) string {
+	name := "generic"
+	if errors.Is(err, context.Canceled) {
+		name = "canceled"
+	}
+	return name
+}
```
jsternberg, line 138: I like this but this isn't a required thing in otel and feel free to use it or not use it based on your own needs.

cmd/docker/docker.go
```diff
@@ -298,6 +303,15 @@ func runDocker(dockerCli *command.DockerCli) error {
 		return err
 	}
 
+	mp := dockerCli.MeterProvider(ctx)
+	defer mp.Shutdown(ctx)
+	// TODO(krissetto): get the package name and version properly for instantiating the meter
+	meter := mp.Meter(
+		"github.com/docker/cli",
+		metric.WithInstrumentationVersion(cmd.Version),
+	)
```
jsternberg, line 312: As mentioned in another place, I'd recommend moving this to a helper function and to avoid passing around the meter.

jsternberg (changes requested):

Benehiko (approved):
From my side I think this is good. I'll approve so long to revert my "requested changes" :)

cli/command/telemetry_utils.go
krissetto: I'm not super sure about this. I agree with the meter provider being set with `otel.SetMeterProvider` and will change that, but reading the otel documentation it seems like it'd be an error to create multiple meters with the same "identity", which is why I added the pre-created meter into the ctx.

Maybe I am misunderstanding something? 

Snippet from [the docs](https://opentelemetry.io/docs/specs/otel/metrics/api/#meterprovider)

> Meters are identified by name, version, and schema_url fields. When more than one Meter of the same name, version, and schema_url is created, it is unspecified whether or under which conditions the same or different Meter instances are returned. **It is a user error to create Meters with different attributes but the same identity.**

cli/command/telemetry_utils.go
krissetto: see comment above for my doubts regarding this

cli/command/telemetry_utils.go
krissetto: Sounds good to me

cli/command/telemetry_utils.go
krissetto: Also liked this, so i decided to keep it :) if we see it being nice to have, we can maybe define it's usage a bit better and share it across projects

cli/command/telemetry_utils.go
krissetto: I consolidated this logic into an `attributesFromError` func which return a series of err-dependent attributes. 
IMHO any extra refactoring of this part should be done once we start implementing more logic and our needs are clearer

cli/command/telemetry_docker.go
laurazard: Thanks for the explanation @jsternberg!

My only comment here is a) should we add that logic for 1.? and b) I'd rather not have `DEBUG` in the env var given that we don't have any other `DEBUG` env vars (other than the actual `DEBUG` env var which sets log level for the CLI, so unless there's a very strong reason not to I'd rather call this something like `DOCKER_CLI_OTEL_EXPORTER_OTLP_ENDPOINT`, esp. since this server to **override** the default `OTEL_EXPORTER_OTLP_ENDPOINT` config.

cmd/docker/docker.go
```diff
@@ -298,6 +304,15 @@ func runDocker(dockerCli *command.DockerCli) error {
 		return err
 	}
 
+	mp := dockerCli.MeterProvider(ctx)
+	defer mp.Shutdown(ctx)
+	otel.SetMeterProvider(mp)
+	meter := mp.Meter(
+		"github.com/docker/cli",
+		metric.WithInstrumentationVersion(cmd.Version),
```
laurazard, line 312: https://github.com/docker/buildx/blob/fe08cf2981836f1828058f02cb7684f676b53a9e/util/metricutil/metric.go#L216-L219 looks like we should use the CLI version here.

laurazard:
```suggestion
		metric.WithInstrumentationVersion(version.Version),
```

cli/command/telemetry_utils.go
laurazard: > It is a user error to create Meters with different attributes but the same identity.

The docs specify that it would be error to have different meters with the same identity **but different attributes**, but in this case they would have the same attributes right? So I think this would be fine @krissetto.

Benehiko: I think this depends. When adding the MeterProvider instance to ctx instead of the Meter instance, it gives opportunity inside the code to create more meters with unique package names. 

>A Meter should be scoped at most to a single package.
https://pkg.go.dev/go.opentelemetry.io/otel/metric#MeterProvider

So as far as I understand this, a new meter would be created per package and from there you would create a new instrument to do things like count etc. https://opentelemetry.io/docs/specs/otel/metrics/api/#meter-operations

From my understanding that would make @jsternberg correct. But this would mean we would need to detect for meter name collisions (unless this is already baked into the Go sdk).

Benehiko: edit: also that would mean we might not need to pass MeterProvider to ctx.

cmd/docker/docker.go
krissetto: Oh! I totally missed that we have a version.go as well, thanks for spotting :)

cli/command/telemetry_utils.go
krissetto: > > It is a user error to create Meters with different attributes but the same identity.
> 
> The docs specify that it would be error to have different meters with the same identity **but different attributes**, but in this case they would have the same attributes right? So I think this would be fine @krissetto.

Mmh i'm not totally sure about this. If otel implicitly returns the same Meter when we give the MeterProvider the same name/version, then I totally agree with the approach. The documentation is unclear with this though, so I will be doing some tests and changing the impl if the meter returned is the same :)

cli/command/telemetry_docker.go
krissetto: No hard feelings with this one, either of the choices is ok imho. Removed `DEBUG` from the name for now

cli/command/telemetry_utils.go
krissetto: > A useful pattern that I found is creating a utility similar to this one: https://github.com/docker/buildx/blob/fe08cf2981836f1828058f02cb7684f676b53a9e/util/metricutil/metric.go#L216-L219

I called this `GetDefaultMeter` for now, to avoid confusion with the pre-existing `metric.MeterProvider.Meter()` func

laurazard (approved):
LGTM 🥳 looks like there's a couple of linting errors, but other than that I'm happy with the current state of the PR.

cli/command/telemetry_docker.go
jsternberg: `DOCKER_CLI_OTEL_EXPORTER_OTLP_ENDPOINT` is good for me. As long as it doesn't conflict with the official environment variable and makes sense I'm good.

cli/command/telemetry_utils.go
jsternberg: > Mmh i'm not totally sure about this. If otel implicitly returns the same Meter when we give the MeterProvider the same name/version, then I totally agree with the approach. The documentation is unclear with this though, so I will be doing some tests and changing the impl if the meter returned is the same :)

Yea, it returns the same meter if you give it the same name/version. If you have an area that calls this repeatedly, then it can be useful to pass around the meter, but it's my opinion that `MeterProvider` should be used as the external API for most if not all situations.

It also mentions that it's an error to put different attributes on the meter, but it's not actually possible to put attributes on the meter anyway in the current version of the SDK. There's an API for instrumentation attributes (which is the name of the type of attribute that you send to the meter) but they're not hooked up anywhere.

> I think this depends. When adding the MeterProvider instance to ctx instead of the Meter instance, it gives opportunity inside the code to create more meters with unique package names.
>
> A Meter should be scoped at most to a single package.
https://pkg.go.dev/go.opentelemetry.io/otel/metric#MeterProvider
>
> So as far as I understand this, a new meter would be created per package and from there you would create a new instrument to do things like count etc. https://opentelemetry.io/docs/specs/otel/metrics/api/#meter-operations

This is true, but my recommendation is to scope the meter to the go module path rather than the package name. It ends up being a bit too difficult to have a different meter per package when processing them downstream. Per-module is a good middle ground in my opinion and it's what we're going to do in buildx for now.

If I'm wrong about this, we can update it later and scope it further. It's mostly about avoiding conflicts with metric names and it should be possible to avoid conflicts within the same module.

laurazard:
@jsternberg wanna take a final look? :')

cli/command/telemetry.go
```diff
@@ -72,6 +73,15 @@ func (cli *DockerCli) MeterProvider(ctx context.Context, opts ...sdkmetric.Optio
...
 
+// GetDefaultMeter gets the default metric.Meter for the application
+// using the given metric.MeterProvider
+func GetDefaultMeter(mp metric.MeterProvider) metric.Meter {
+	return mp.Meter(
+		"github.com/docker/cli",
+		metric.WithInstrumentationVersion(version.Version),
+	)
+}
+
```
jsternberg, line 84: Is it possible to move this to a less prominent location? It's my opinion that this function should only be invoked within this module and I'd like to avoid us accidentally calling it from a plugin.

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,154 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+type contextKey string
+
+const (
+	OtelMeterCtxKey  = contextKey("otel.meter")
+	OtelTracerCtxKey = contextKey("otel.tracer")
+)
```
jsternberg, line 22: Is this stuff needed anymore?

cli/command/telemetry_utils.go
```diff
@@ -0,0 +1,154 @@
...
+			return cmdErr
+		}
+
+		return ogPersistentPreRunE(cmd, args)
+	}
+}
+
+func startCobraCommandTimer(cmd *cobra.Command) func(err error) {
+	ctx := cmd.Context()
+	meter := GetDefaultMeter(otel.GetMeterProvider())
```
jsternberg, line 75: nitpick: Can you modify `InstrumentCobraCommands` to take the `MeterProvider` as an input and use that instead of the global function?

I'm not sure of what the best way is, but the way I've been treating global state like `otel.GetMeterProvider` and `otel.SetMeterProvider` is they're only safe in code that is either in a `cmd` package or directly invoked only from a `cmd` package and everything else should take it as a parameter so an alternative method can be used.

Since the `cli/command` package is used as an API by plugins too, I don't think we want to force plugins to use `otel.SetMeterProvider`, but I think it's good if the plugins use `otel.SetMeterProvider` because it's useful and why not?

For this situation, you may not even need to use `otel.SetMeterProvider` anymore because I think they're used in the same scope anyway.

jsternberg (changes requested):
A few more comments but this mostly looks good to me.

cli/command/telemetry_utils.go
krissetto: nope

cli/command/telemetry_utils.go
krissetto: Alright, I passed the MeterProvider along for now but still used otel.SetMeterProvider

jsternberg (approved):

PR #4943, opened 2024-03-15
builder: correct the command path for docker build

jsternberg:
**- What I did**

The command path sent for `docker build` should be `docker` rather than `docker build` to be consistent with the other command paths.

* `docker buildx build` has a command path of `docker buildx`
* `docker builder build` has a command path of `docker builder`
* `docker image build` has a command path of `docker image`

The reason this gets set to `docker buildx` rather than `docker buildx build` is because the `build` portion of the command path is processed by the plugin. So the command path only contains the portions of the command path that were processed by this tool.

Since the `build` of `docker build` gets forwarded to `buildx`, it is not included in the command path.

**- How I did it**

Modified the process for how the command path gets set.

**- How to verify it**

Inside of buildkit, you can do `hack/compose -f hack/composefiles/extensions/buildx.yaml up -d --build`. Then do `docker logs -f buildkit-otel-collector-1`.

Compile master of `buildx` to a location where the docker cli can find it. Compile this change. Run `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317 BUILDX_EXPERIMENTAL=true docker build ...` and look a the debug output.

### DIFF
--- a/cmd/docker/builder.go
+++ b/cmd/docker/builder.go
@@ -69,7 +69,7 @@ func processBuilder(dockerCli command.Cli, cmd *cobra.Command, args, osargs []st
 	}
 
 	// is this a build that should be forwarded to the builder?
-	fwargs, fwosargs, alias, forwarded := forwardBuilder(builderAlias, args, osargs)
+	fwargs, fwosargs, fwcmdpath, forwarded := forwardBuilder(builderAlias, args, osargs)
 	if !forwarded {
 		return args, osargs, nil, nil
 	}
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
SGTM

cmd/docker/builder.go
```diff
@@ -117,33 +117,38 @@ func processBuilder(dockerCli command.Cli, cmd *cobra.Command, args, osargs []st
 	}
 
 	// overwrite the command path for this plugin using the alias name.
-	cmd.Annotations[pluginmanager.CommandAnnotationPluginCommandPath] = fmt.Sprintf("%s %s", cmd.CommandPath(), alias)
+	cmd.Annotations[pluginmanager.CommandAnnotationPluginCommandPath] =
+		strings.Join(append([]string{cmd.CommandPath()}, fwcmdpath...), " ")
```
thaJeztah, line 121: Oh! It doesn't like the wrapping here; think it's fine to keep it a single line;


```
64.44 cmd/docker/builder.go:120: File is not `gofumpt`-ed (gofumpt)
64.44 	cmd.Annotations[pluginmanager.CommandAnnotationPluginCommandPath] =
64.44 		strings.Join(append([]string{cmd.CommandPath()}, fwcmdpath...), " ")
```

jsternberg: Weird. I only did this because it exceeded 80 characters.

PR #4949, opened 2024-03-18
docs/deprecated: update status of short cid alias deprecation

akerouanton:
- Related to https://github.com/moby/moby/pull/46853

**- What I did**

Starting with API v1.45, the container short ID is removed from the container Aliases. Update the status for that deprecation.

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -52,7 +52,7 @@ The table below provides an overview of the current status of deprecated feature
 |------------|------------------------------------------------------------------------------------------------------------------------------------|------------|--------|
 | Deprecated | [Unauthenticated TCP connections](#unauthenticated-tcp-connections)                                                                | v26.0      | v27.0  |
 | Deprecated | [Deprecate legacy API versions](#deprecate-legacy-api-versions)                                                                    | v25.0      | v26.0  |
-| Deprecated | [Container short ID in network Aliases field](#container-short-id-in-network-aliases-field)                                        | v25.0      | v26.0  |
+| Removed    | [Container short ID in network Aliases field](#container-short-id-in-network-aliases-field)                                        | v25.0      | v26.0  |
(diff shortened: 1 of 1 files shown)


vvoland (approved):

robmry (approved):
LGTM

thaJeztah:
Do we need to mention that it's deprecated for API XX, but older API versions still having it?

akerouanton:
> Do we need to mention that it's deprecated for API XX, but older API versions still having it?

Is that something we usually do in `docs/deprecated.md`? Except for "Deprecate legacy API versions", it seems we never advertise the use of `DOCKER_API_VERSION`.

thaJeztah:
It's a bit similar to https://github.com/docker/cli/pull/4893#discussion_r1508941650 - we don't recommend `DOCKER_API_VERSION`, but clients that were created for an older version of the API should continue to observe the old behavior (i.e., older API versions are not "touched").


We should probably look (as discussed in the other PR) to more prominently add a link to the API history, as that's effectively the source of truth for API changes (including deprecations)

akerouanton:
> It's a bit similar to https://github.com/docker/cli/pull/4893#discussion_r1508941650 - we don't recommend DOCKER_API_VERSION, but clients that were created for an older version of the API should continue to observe the old behavior (i.e., older API versions are not "touched").

Then, should we mention that only newer CLI versions are affected, ie. CLI v26 and above?

neersighted (approved):
Glancing at previous deprecations, this should be sufficient on the CLI side. 

The 1.45 deprecation does not mention the change telegraphed in 1.44:

https://github.com/moby/moby/blob/963e1f3eed52024629557c11503c56f28d578f55/docs/api/version-history.md?plain=1#L16-L27

https://github.com/moby/moby/blob/963e1f3eed52024629557c11503c56f28d578f55/docs/api/version-history.md?plain=1#L86-L90

I think all we need to do is make sure to add the change to API 1.45; with that change, this LGTM.

Issue #4948, opened 2024-03-17
flaky test: TestPromptForConfirmation/case=reader_closed


thaJeztah:
### Description

- test was added on https://github.com/docker/cli/commit/10bf91a02d2abd4dec90b79bd10b91f6fbb8e05d / https://github.com/docker/cli/pull/4888

This test looks to be flaky on macOS at least; seen it fail a couple of times.

```
ok  	github.com/docker/cli/cli	0.242s	coverage: 28.2% of statements
--- FAIL: TestPromptForConfirmation (0.38s)
    --- FAIL: TestPromptForConfirmation/case=reader_closed (0.19s)
        utils_test.go:222: PromptForConfirmation did not return after promptReader was closed
FAIL
panic: send on closed channel

goroutine 111 [running]:
github.com/docker/cli/cli/command_test.TestPromptForConfirmation.func8.2()
	/Users/runner/work/cli/cli/src/github.com/docker/cli/cli/command/utils_test.go:209 +0x6d
created by github.com/docker/cli/cli/command_test.TestPromptForConfirmation.func8 in goroutine 110
	/Users/runner/work/cli/cli/src/github.com/docker/cli/cli/command/utils_test.go:207 +0x3aa
```

thaJeztah:
cc @Benehiko

laurazard:
Good point from @tonistiigi here – https://github.com/docker/cli/pull/4888#discussion_r1525644870.

A more holistic approach would be to always set up signal handling in the CLI, and use it to cancel the command's context, and just migrate most places in the codebase that are setting up their own signal handling to instead handle context cancellation – funny that [we do this for plugins but not for anything else](https://github.com/search?q=repo%3Adocker%2Fcli%20signal.notify&type=code).

thaJeztah:
Ah, yes, I saw that comment, and wanted to ask @krissetto about it as well.

> A more holistic approach would be to always set up signal handling in the CLI, and use it to cancel the command's context, 

Sounds like a sensible approach yes (or at least, that's my first response; contexts are always fun)

Benehiko:
> Good point from @tonistiigi here – [#4888 (comment)](https://github.com/docker/cli/pull/4888#discussion_r1525644870).
> 
[...]

PR #4957, opened 2024-03-21
fix: flaky prompt termination on reader close test

Benehiko:
Fixes https://github.com/docker/cli/issues/4948

**- What I did**
Increase the context timeout from 100ms to 500ms. Removed the manual close on the result channels to prevent panics on test failures.

**- How I did it**

**- How to verify it**

```shell
$ go test -v -run=TestPromptForConfirmation/case=reader_closed -count=1000 ./cli/command/utils_test.go

=== RUN   TestPromptForConfirmation/case=reader_closed
--- PASS: TestPromptForConfirmation (0.00s)
    --- PASS: TestPromptForConfirmation/case=reader_closed (0.00s)
PASS
ok  	command-line-arguments	0.263s
```

Fix TestPromptForConfirmation flakiness

### DIFF
--- a/cli/command/utils_test.go
+++ b/cli/command/utils_test.go
@@ -106,91 +106,33 @@ func TestPromptForConfirmation(t *testing.T) {
 	}()
 
 	for _, tc := range []struct {
-		desc string
-		f    func(*testing.T, context.Context, chan promptResult)
+		desc     string
+		f        func() error
+		expected promptResult
 	}{
-		{"SIGINT", func(t *testing.T, ctx context.Context, c chan promptResult) {
-			t.Helper()
-
+		{"SIGINT", func() error {
 			syscall.Kill(syscall.Getpid(), syscall.SIGINT)
-
-			select {
-			case <-ctx.Done():
-				t.Fatal("PromptForConfirmation did not return after SIGINT")
-			case r := <-c:
-				assert.Check(t, !r.result)
-				assert.ErrorContains(t, r.err, "prompt terminated")
-			}
-		}},
-		{"no", func(t *testing.T, ctx context.Context, c chan promptResult) {
-			t.Helper()
-
+			return nil
+		}, promptResult{false, command.ErrPromptTerminated}},
+		{"no", func() error {
 			_, err := fmt.Fprint(promptWriter, "n\n")
-			assert.NilError(t, err)
-
-			select {
-			case <-ctx.Done():
-				t.Fatal("PromptForConfirmation did not return after user input `n`")
-			case r := <-c:
-				assert.Check(t, !r.result)
-				assert.NilError(t, r.err)
-			}
-		}},
-		{"yes", func(t *testing.T, ctx context.Context, c chan promptResult) {
-			t.Helper()
-
+			return err
+		}, promptResult{false, nil}},
+		{"yes", func() error {
 			_, err := fmt.Fprint(promptWriter, "y\n")
-			assert.NilError(t, err)
-
-			select {
-			case <-ctx.Done():
-				t.Fatal("PromptForConfirmation did not return after user input `y`")
-			case r := <-c:
-				assert.Check(t, r.result)
-				assert.NilError(t, r.err)
-			}
-		}},
-		{"any", func(t *testing.T, ctx context.Context, c chan promptResult) {
-			t.Helper()
-
+			return err
+		}, promptResult{true, nil}},
+		{"any", func() error {
 			_, err := fmt.Fprint(promptWriter, "a\n")
-			assert.NilError(t, err)
-
-			select {
-			case <-ctx.Done():
-				t.Fatal("PromptForConfirmation did not return after user input `a`")
-			case r := <-c:
-				assert.Check(t, !r.result)
-				assert.NilError(t, r.err)
-			}
-		}},
-		{"with space", func(t *testing.T, ctx context.Context, c chan promptResult) {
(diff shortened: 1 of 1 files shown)


krissetto (approved):
LGTM.

Generally speaking, could you reproduce the exact error locally by making this timeout really small?

Benehiko:
> LGTM.
> 
> Generally speaking, could you reproduce the exact error locally by making this timeout really small?

Yes, but it had to be really really small.
```go
resultCtx, resultCancel := context.WithTimeout(ctx, 1*time.Nanosecond)
```

```shell
    utils_test.go:222: PromptForConfirmation did not return after promptReader was closed
--- FAIL: TestPromptForConfirmation (0.00s)
    --- FAIL: TestPromptForConfirmation/case=reader_closed (0.00s)
=== RUN   TestPromptForConfirmation
panic: send on closed channel

goroutine 36 [running]:
command-line-arguments_test.TestPromptForConfirmation.func8.2()
	/home/benehiko/Github/docker-cli/cli/command/utils_test.go:209 +0x6d
created by command-line-arguments_test.TestPromptForConfirmation.func8 in goroutine 35
	/home/benehiko/Github/docker-cli/cli/command/utils_test.go:207 +0x32a
FAIL	command-line-arguments	0.011s
FAIL
```

laurazard:
Changing the timeout length causing a panic is a code smell – and simply not closing the channel "fixes" it, but isn't really correct. The issue here is we're launching a goroutine in https://github.com/Benehiko/docker-cli/blob/d2ea5adfe401205d39050abe117cd1cb6811764b/cli/command/utils_test.go#L205-L208 without a way to signal it to end/that we don't care about it's result anymore.

Benehiko:
> Changing the timeout length causing a panic is a code smell – and simply not closing the channel "fixes" it, but isn't really correct. The issue here is we're launching a goroutine in https://github.com/Benehiko/docker-cli/blob/d2ea5adfe401205d39050abe117cd1cb6811764b/cli/command/utils_test.go#L205-L208 without a way to signal it to end/that we don't care about it's result anymore.

The panic is not caused by a too short timeout - it's just a side-effect of it. Closing the channel in this function was actually a mistake since a test error would always throw a panic, which is not what we want. Omitting the lines closing the channel will keep the channel open for as long as the goroutine is active and will be cleaned up by the garbage collector.
Eventually the goroutine will be terminated due to context cancellation. 

Please see https://github.com/docker/cli/issues/4948#issuecomment-2011778770

laurazard:
> The panic is not caused by a too short timeout - it's just a side-effect of it.

I got that 👍

My point is that, in general, this is indicative that we're doing something wrong/error prone and that we can do better – instead of calling functions that know how to verify if a test has passed (and if they've timed out), we can have test cases tell us what their expected results are and check it ourselves within the main test flow. This makes it so that we're sure that, by the time we call `close` on those channels, no one is still trying to use them. For example:
```go
	for _, tc := range []struct {
		desc           string
		f              func() error
		expectedResult promptResult
	}{
		{
			"SIGINT", func() error {
				syscall.Kill(syscall.Getpid(), syscall.SIGINT)
				return nil
			},
			promptResult{
				result: false,
				err:    command.ErrPromptTerminated,
			},
		},
		{
			"no", func() error {
				_, err := fmt.Fprint(promptWriter, "n\n")
				return err
			},
			promptResult{
				result: false,
			},
		},
		{
			"yes", func() error {
				_, err := fmt.Fprint(promptWriter, "y\n")
				return err
			},
			promptResult{
				result: true,
			},
		},
		{
			"any", func() error {
				_, err := fmt.Fprint(promptWriter, "a\n")
				return err
			},
			promptResult{
				result: false,
			},
		},
		{
			"with space", func() error {
				_, err := fmt.Fprint(promptWriter, " y\n")
				return err
			},
			promptResult{
				result: true,
			},
		},
		{
			"reader closed", func() error {
				return promptReader.Close()
			},
			promptResult{
				result: false,
			},
		},
	} {
		t.Run("case="+tc.desc, func(t *testing.T) {
			buf.Reset()
			promptReader, promptWriter = io.Pipe()

			wroteHook := make(chan struct{}, 1)
			defer close(wroteHook)
			promptOut := test.NewWriterWithHook(bufioWriter, func(_ []byte) {
				wroteHook <- struct{}{}
			})

			result := make(chan promptResult, 1)
			defer close(result)
			go func() {
				r, err := command.PromptForConfirmation(ctx, promptReader, promptOut, "")
				result <- promptResult{r, err}
			}()

			select {
			case <-time.After(100 * time.Millisecond):
			case <-wroteHook:
			}

			drainChannel(ctx, wroteHook)

			assert.NilError(t, bufioWriter.Flush())
			assert.Equal(t, strings.TrimSpace(buf.String()), "Are you sure you want to proceed? [y/N]")

			assert.NilError(t, tc.f())

			select {
			case r := <-result:
				assert.Equal(t, r, tc.expectedResult)
			case <-time.After(500 * time.Millisecond):
				t.Fatal("test timed out - " + tc.desc)
			}
		})
	}
}

func drainChannel(ctx context.Context, ch <-chan struct{}) {
	go func() {
		for {
			select {
			case <-ctx.Done():
				return
			case <-ch:
			}
		}
	}()
}
```

I think this is easier to follow/less prone to errors than the other way around.

laurazard:
Thanks!

laurazard (approved):
LGTM

PR #4960, opened 2024-03-22
plugin: update/improve process lifecycle documentation

neersighted:
Follow-up to https://github.com/docker/cli/pull/4905, making comment/documentation improvements I didn't want to block merge of that PR with.

### DIFF
--- a/cli-plugins/socket/socket.go
+++ b/cli-plugins/socket/socket.go
@@ -11,12 +11,13 @@ import (
 	"sync"
 )
 
-// EnvKey represents the well-known environment variable used to pass the plugin being
-// executed the socket name it should listen on to coordinate with the host CLI.
+// EnvKey represents the well-known environment variable used to pass the
+// plugin being executed the socket name it should listen on to coordinate with
+// the host CLI.
 const EnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
 
-// NewPluginServer creates a plugin server that listens on a new Unix domain socket.
-// `h` is called for each new connection to the socket in a goroutine.
+// NewPluginServer creates a plugin server that listens on a new Unix domain
+// socket. h is called for each new connection to the socket in a goroutine.
 func NewPluginServer(h func(net.Conn)) (*PluginServer, error) {
 	l, err := listen("docker_cli_" + randomID())
 	if err != nil {
@@ -63,7 +64,7 @@ func (pl *PluginServer) accept() error {
 	defer pl.mu.Unlock()
 
 	if pl.closed {
-		// handle potential race condition between Close and Accept
+		// Handle potential race between Close and accept.
 		conn.Close()
 		return errors.New("plugin server is closed")
 	}
@@ -74,20 +75,25 @@ func (pl *PluginServer) accept() error {
 	return nil
 }
 
+// Addr returns the [net.Addr] of the underlying [net.Listener].
 func (pl *PluginServer) Addr() net.Addr {
 	return pl.l.Addr()
 }
 
-// Close ensures that the server is no longer accepting new connections and closes all existing connections.
-// Existing connections will receive [io.EOF].
+// Close ensures that the server is no longer accepting new connections and
+// closes all existing connections. Existing connections will receive [io.EOF].
+//
+// The error value is that of the underlying [net.Listner.Close] call.
 func (pl *PluginServer) Close() error {
 	// Remove the listener socket, if it exists on the filesystem.
 	unlink(pl.l)
 
-	// Close connections first to ensure the connections get io.EOF instead of a connection reset.
+	// Close connections first to ensure the connections get io.EOF instead
+	// of a connection reset.
 	pl.closeAllConns()
 
-	// Try to ensure that any active connections have a chance to receive io.EOF
+	// Try to ensure that any active connections have a chance to receive
+	// io.EOF.
 	runtime.Gosched()
 
 	return pl.l.Close()
@@ -97,7 +103,7 @@ func (pl *PluginServer) closeAllConns() {
 	pl.mu.Lock()
 	defer pl.mu.Unlock()
 
-	// Prevent new connections from being accepted
+	// Prevent new connections from being accepted.
 	pl.closed = true
 
 	for _, conn := range pl.conns {
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -220,33 +220,46 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string, 
 		return err
 	}
 
-	// Establish the plugin socket, adding it to the environment under a well-known key if successful.
+	// Establish the plugin socket, adding it to the environment under a
+	// well-known key if successful.
 	srv, err := socket.NewPluginServer(nil)
 	if err == nil {
-		envs = append(envs, socket.EnvKey+"="+srv.Addr().String())
+		plugincmd.Env = append(plugincmd.Env, socket.EnvKey+"="+srv.Addr().String())
 	}
 
-	plugincmd.Env = append(envs, plugincmd.Env...)
+	// Set additional environment variables specified by the caller.
+	plugincmd.Env = append(plugincmd.Env, envs...)
 
+	// Background signal handling logic: block on the signals channel, and
+	// notify the plugin via the PluginServer (or signal) as appropriate.
 	const exitLimit = 3
-
 	signals := make(chan os.Signal, exitLimit)
 	signal.Notify(signals, platformsignals.TerminationSignals...)
-	// signal handling goroutine: listen on signals channel, and if conn is
-	// non-nil, attempt to close it to let the plugin know to exit. Regardless
-	// of whether we successfully signal the plugin or not, after 3 SIGINTs,
-	// we send a SIGKILL to the plugin process and exit
 	go func() {
 		retries := 0
 		for range signals {
+			// If stdin is a TTY, the kernel will forward
+			// signals to the subprocess because the shared
+			// pgid makes the TTY a controlling terminal.
+			//
+			// The plugin should have it's own copy of this
+			// termination logic, and exit after 3 retries
+			// on it's own.
 			if dockerCli.Out().IsTerminal() {
-				// running attached to a terminal, so the plugin will already
-				// receive signals due to sharing a pgid with the parent CLI
 				continue
 			}
 
-			srv.Close()
+			// Terminate the plugin server, which will
+			// close all connections with plugin
+			// subprocesses, and signal them to exit.
+			//
+			// Repeated invocations will result in EINVAL,
+			// or EBADF; but that is fine for our purposes.
+			_ = srv.Close()
 
+			// If we're still running after 3 interruptions
+			// (SIGINT/SIGTERM), send a SIGKILL to the plugin as a
+			// final attempt to terminate, and exit.
 			retries++
 			if retries >= exitLimit {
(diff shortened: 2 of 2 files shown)


Benehiko (approved):
LGTM

cli-plugins/socket/socket.go
```diff
@@ -11,12 +11,13 @@ import (
 	"sync"
 )
 
-// EnvKey represents the well-known environment variable used to pass the plugin being
-// executed the socket name it should listen on to coordinate with the host CLI.
+// EnvKey represents the well-known environment variable used to pass the
+// plugin being executed the socket name it should listen on to coordinate with
+// the host CLI.
 const EnvKey = "DOCKER_CLI_PLUGIN_SOCKET"
```
thaJeztah, line 17: Hm... and I just realised we may have picked a very wrong name for this env-var, or at least the GoDoc should probably be very clear that this is an _internal_ env-var, used between the CLI (acting as "host" for plugins) and binaries running as CLI-plugin.

- We use the all-caps `DOCKER_XXXXXX` format for user-facing env-vars; in this case, we don't want users to (e.g.) `DOCKER_CLI_PLUGIN_SOCKET=<whatever> /usr/libexec/docker/cli-plugins/docker-compose ....`
- :point_up: I don't think the above would do anything if the plugin is running "standalone" (if it _would_ it could possibly `rm` the given path if it's a macOS binary?)
- :point_up: BUT we should probably document that in the GoDoc regardless; explicitly call out this is an _internal-use_ env-var, and not to be set by the user.
- :point_up: Paranoid me is even considering; do we need to _actively unset_ the env-var when found? (To prevent any chance it leaks through somewhere?)


Should we change it to something more clearly meant for "internal"? Any good conventions for that?

- Perhaps lowercase? (`docker_cli_plugin_socket`)
- Perhaps prefix with `_` / `__` (`__docker_cli_plugin_socket`); could even have a `_addr` suffix to indicate it's the address (`__docker_cli_plugin_socket_addr`).
- :point_up: pedantic me was even considering "version it" (`__docker_cli_plugin_socket_addr_v1`), but I guess we can do so if we would need a `v2`
- Maybe something mentioning "internal" (`__docker_cli_plugin_internal_socket_addr`)

On topic of `_addre`; should we have included the scheme / protocol? (`unix:///`); would there ever be a reason for us to change this? (`grpc://`)

cli-plugins/socket/socket.go
```diff
@@ -97,7 +103,7 @@ func (pl *PluginServer) closeAllConns() {
 	pl.mu.Lock()
 	defer pl.mu.Unlock()
 
-	// Prevent new connections from being accepted
+	// Prevent new connections from being accepted.
```
thaJeztah, line 106: FWIW, there's a linter for this; I know some projects I work on that have it enabled, but if we want to enable it, we probably will have to make a big pass on updating many existing GoDocs 🙈 (so I didn't do that yet).

cli-plugins/socket/socket.go
krissetto: I agree with these points. I was especially curious about including the scheme/protocol in the env var value as well, as i guess right now we just assume the socket is `unix` in the various plugins etc

krissetto (approved):

cli-plugins/socket/socket.go
neersighted: Re: naming, we haven't been strict in the past, e.g. `DOCKER_TEST_` exists in several places in the daemon. I'd suggest we just go with something like `@DOCKER` as the convention for internal variables, as the `@` makes it very hard to interact with these variables from shells.

Making that change now would be moderately breaking, but I think it would be more acceptable now rather than later (if we want to bite the bullet) as mismatched versions would only result in a reversion to the less-than-graceful lifecycle, instead of anything more drastic (e.g. ability to talk to a RPC server being broken).

Regarding the protocol, I don't think that makes sense, and I think it's just unnecessary complxity -- we control both the producer and the consumer here (in one codebase), and while cross-version compatibility is a concern, I think we just version the environment variable if we ever make a breaking change.

Adding e.g. gRPC or TTRPC here will not be breaking as older clients won't actually try to push anything over the socket. Also, `unix://` would indicate a transport layer, while `grpc://` would indicate a protocol -- so something like `unix+grpc://` would be a lot more appropriate.

thaJeztah: > Re: naming, we haven't been strict in the past, e.g. DOCKER_TEST_ exists in several places in the daemon.

So `DOCKER_TEST` was used for "CI-specific" env-vars, so "some consistency" there, but agreed, we haven't always been good at it. Still, `DOCKER_` is common for user-facing ones, so what I want to avoid.

> I'd suggest we just go with something like @DOCKER as the convention for internal variables, as the @ makes it very hard to interact with these variables from shells.

I'm _slightly_ concerned about using characters outside of the recommended character set https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html;

"Environment variable names used by the utilities in the Shell and Utilities volume of POSIX.1-2017 consist solely of uppercase letters, digits, and the <underscore> ( '_' ) from the characters defined in Portable Character Set and do not begin with a digit."

Hence thinking of lowercase (indicating "local" / "non exported"), and the underscore prefix.

> Making that change now would be moderately breaking, but I think it would be more acceptable now rather than later (if we want to bite the bullet) as mismatched versions would only result in a reversion to the less-than-graceful lifecycle, instead of anything more drastic (e.g. ability to talk to a RPC server being broken).

Yup! This was a bit of an oversight I think, at least it didn't occur to me at the time.

w.r.t. compatibility; we can add a transitional situation; set _both_ "old" and "new" env-var name for 1 release; this allows us to ship the new name fast (existing plugins continue to function, and don't even have to fall back), and then remove the "duplicate" env-var in the release after.

> Adding e.g. gRPC or TTRPC here will not be breaking as older clients won't actually try to push anything over the socket. Also, unix:// would indicate a transport layer, while grpc:// would indicate a protocol -- so something like unix+grpc:// would be a lot more appropriate

Yes, I haven't given that a _lot_ of thought yet, other then _IF_ we decide to make this change, "now" may be the right moment to do so as well.

neersighted: Going outside the recommended character set would more or less be the point; the kernel/OS actually have very little in the way of opinions here outside the significance of '=' (see an example implementation in e.g. [bionic libc](https://android.googlesource.com/platform/bionic/+/a27d2baa/libc/stdlib/getenv.c#46), and the POSIX recommendations primarily relate to shell support.

With regard to compatibility, I think that's somewhat more optimistic than is realistic. I don't think that older plugins with a newer CLI are going to be that unusual, though it's not really an edge case I am too concerned about.

In any case, I'd rather not layer more complexity here... If we want to change the variable (and any changes should be in a separate PR), let's do that; but let's keep the simple 'sun_name of a Unix socket' pattern, and let's recognize that this may cause regressions for some users unless we maintain both variables in perpetuity (partially defeating the point).

Oh, and I forgot to address the 'what if the user sets this in standalone mode' question -- nothing, really. The `unlink(2)` logic is in the parent CLI, so the worst they can do is cause some spurious attempts to connect and log messages.

thaJeztah: >Going outside the recommended character set would more or less be the point; the kernel/OS actually have very little in the way of opinions here outside the significance of '=' (see an example implementation in e.g. [bionic libc](https://android.googlesource.com/platform/bionic/+/a27d2baa/libc/stdlib/getenv.c#46)), and the POSIX recommendations primarily relate to shell support. 

That's based on knowledge we have _now_; Linux may be flexible, and maybe macOS and Windows as well, but I'd rather avoid taking risks. We used to have a comment in the Docker readme "it doesn't support Windows, and unlikely ever will!". That didn't age well, and the same could apply here (what does .. WASM support? what does `<insert unknown future>`).

laurazard: > I'd suggest we just go with something like @docker as the convention for internal variables, as the @ makes it very hard to interact with these variables from shells.

This is indeed special cases though, and not just from shells – look at this Go code 😅 https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/net/unixsock_posix.go;l=192

neersighted: @thaJeztah I think it's safe to say that libc semantics will not change any time soon; these things are more or less well-negotiated, across all operating systems we care about.

I guess I can't say for certain what Fuchsia/Zircon do (or other future operating systems), so maybe there is some merit in that hypothetical. But by the same token, it's up to the Go team to figure out what to do there, since they provide (through their documented APIs) the same Unix-inspired environ semantics that all modern operating systems use.

* https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables
* https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08 (note that the suggested character set is for maximum portability, but the only prohibited characters are '=' and NUL)
* https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/execve.2.html#//apple_ref/doc/man/2/execve and https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/setenv.3.html

@laurazard you're looking at the socket listener code; I'm referring to the 'key' in the environment (i.e. the 'name' or LHS of the special '=' character in the above documentation)/the argument to [`os.GetEnv`](https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/syscall/env_unix.go;l=69-89).

laurazard: Re: the overall point though, I don't see a need for changing the env-var right now. It _looks_ kind of like a user-facing env var, which we could/should have avoided, but now that it's there, the downsides of changing it don't seem worth any positives we get from that? If it leaks through anywhere (where would that happen? it's set on the CLI process, and only used for the plugins it execs, and dies after) then there isn't really any consequences to that, since the socket is only used for signalling right now, and the socket gets cleaned up before the CLI exits regardless.
If in the future we make a change to use the socket for more things/communication, I think then would be a good time to reevaluate and maybe rename. Until then, I don't see a point.

thaJeztah: @neersighted But why take the risk? Do we really need to? Go stdlib also supports abstract sockets on macOS, and we've seen how well that went

neersighted: The Go stdlib does not support abstract sockets on macOS -- the convention of '@' to indicate a NUL byte at the start of `sun_path` is explicitly not part of the macOS implementation. It was our oversight/not checking what was supported that led to the creation of files with an '@' (and indeed, the 'skip unlink' is universal -- this was a shortcut taken by the Go team).

See https://cs.opensource.google/go/go/+/master:src/syscall/syscall_linux.go;l=569-574;drc=87056756141798b4dfd51dcaaa3e4ce63633a884;bpv=0;bpt=1 and https://cs.opensource.google/go/go/+/master:src/syscall/syscall_windows.go;l=870-875 for the Unix and Windows implementations.

I don't see the risk here -- by consulting the sources of libc and Go on the resulting platforms, the decades-old convention can be trivially verified.

laurazard: > @laurazard you're looking at the socket listener code; I'm referring to the 'key' in the environment

OH DUH. Things got mixed up in my head, when talking about socket name vs socket ENV VAR name 😓 Yeah, "@" should be fine here.

I still do think that we don't need to change this right now (we'll be doing breaking changes to older plugins that are out there for virtually no benefit).

laurazard: > (and indeed, the 'skip unlink' is universal -- this was a shortcut taken by the Go team)

Also, you say shortcut, I say oversight 😅

neersighted:
I'm going to go ahead and bring this in, and open another PR to continue discussion/make a change I only now noticed is desirable.

PR #4863, opened 2024-02-08
use compose-go to load and parse compose files

glours:
**- What I did**
Use `compose-go` to parse, interpolate and merge Compose configurations. 
As this library, under active maintenance and development by the Compose team, is up to date with the recent evolutions of the Compose Specification it will remove the potential divergence between the 2 code base, make the maintenance and evolution easier for everyone.
As a side effect, it would remove the necessity of using the `version` attribute. It disturbs users, the official Docker message is: "You don't need the version attribute anymore in your Compose files" and the CLI continue to introduce new versions of compose v3 file format

**- How I did it**
Import the `compose-go` library, removed the duplicated code for structs, keep only the specific checks done on forbidden, unsupported and deprecated properties on the parsing part. 
Remove the interpolation and merging functions and let `compose-go` doing the job
Add the specific mapping for `x-cluster-spec` extension attribut to match the `ClusterVolumeSpec` struct
Remove all the duplicated tests already done on the `compose-go` side

**- How to verify it**
Run tests
Use average Compose configurations and check we're still producing the same output than the current version

**- Description for the changelog**

Use `compose-go` to manage Compose configuration files

**- A picture of a cute animal (not mandatory but encouraged)**

image

PR #4963, opened 2024-03-22
plugin: drop explicit unlink

neersighted:
Go's `net` package [will unlink][1] for us, as long as we used listen to create the Unix socket.

Go will even skip the unlink when the socket appears to be abstract (start with a NUL, represented by an @), though we must be cautious to only create sockets with an abstract address on platforms that actually support it -- this caused [several][2] [bugs][3] before.

  [1]: https://pkg.go.dev/net#UnixListener.SetUnlinkOnClose
  [2]: https://github.com/docker/cli/pull/4783
  [3]: https://github.com/docker/cli/pull/4863

### DIFF
--- a/cli-plugins/socket/socket_noabstract.go
+++ b/cli-plugins/socket/socket_noabstract.go
@@ -3,23 +3,12 @@
 package socket
 
 import (
-	"net"
 	"os"
 	"path/filepath"
-	"syscall"
 )
 
-func listen(socketname string) (*net.UnixListener, error) {
-	// Because abstract sockets are unavailable, we create a socket in the
-	// system temporary directory instead.
-	return net.ListenUnix("unix", &net.UnixAddr{
-		Name: filepath.Join(os.TempDir(), socketname),
-		Net:  "unix",
-	})
-}
-
-func unlink(listener *net.UnixListener) {
-	// unlink(2) is best effort here; if it fails, we may 'leak' a socket
-	// into the filesystem, but this is unlikely and overall harmless.
-	_ = syscall.Unlink(listener.Addr().String())
+func socketName(basename string) string {
+	// Because abstract sockets are unavailable, use a socket path in the
+	// system temporary directory.
+	return filepath.Join(os.TempDir(), basename)
 }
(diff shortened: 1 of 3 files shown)


krissetto (approved):
LGTM in the context of plugins at least :)

Benehiko (approved):

cli-plugins/socket/socket_noabstract.go
```diff
@@ -3,23 +3,12 @@
...
-}
-
-func unlink(listener *net.UnixListener) {
-	// unlink(2) is best effort here; if it fails, we may 'leak' a socket
-	// into the filesystem, but this is unlikely and overall harmless.
-	_ = syscall.Unlink(listener.Addr().String())
+func socketName(basename string) string {
+	// Because abstract sockets are unavailable, use a socket path in the
+	// system temporary directory.
+	return filepath.Join(os.TempDir(), basename)
```
vvoland, line 13: Can we add a comment explaining why this one doesn't start with `@` while the other one does?

neersighted: I thought it was self-evident from the buildtags -- this is the version for platforms that don't support abstract sockets. The inverse is also true; the abstract socket version uses the `@` to signify an abstract socket on supported platforms...

vvoland: It's not self-evident that the socket will be automatically unlinked just because it doesn't start with a `@` though.

neersighted: That's not relevant in this file -- that's a convention of the Go `net` API. I've linked to the documentation in the commit message, but this shouldn't require being explicitly called out. The fact that Go unlinks the socket file on `Close()` is a well-known part of the API and a 'normal' usage.

neersighted: I've tried added a comment to `socket.go`, since the main issue here seems to be a desire for documentation of the implicit unlink that is part of the net package.

vvoland (approved):

PR #4973, opened 2024-03-28
cli: change any to interface{} for go 1.16 compatibility

jsternberg:
**- What I did**

This is needed because the project does not have a `go.mod` file and gets sent to go 1.16 semantics whenever it's imported by another project. The `any` keyword was added with generics and doesn't exist in go 1.16.

**- How I did it**

No generics were used so I just changed `any` to `interface{}`.


cli/command/telemetry_docker.go
```diff
@@ -1,3 +1,6 @@
+// FIXME(jsternberg): remove once we are a module; the go:build directive prevents go from downgrading language version to go1.16:
+//go:build go1.19
```
krissetto, line 2: is there a reason for 1.19 over 1.18?

krissetto (approved):
just a nit

cli/command/telemetry_docker.go
jsternberg: This seems to be the version that was chosen by @thaJeztah so I just copied it.

https://github.com/docker/cli/commit/70216b662dc440faaebab531deb35f0f0c633d17

This was the reasoning:

> It's currently using go1.19 as version to match the version
in our "vendor.mod", but we can consider being more permissive ("any" requires
go1.18 or up), or more "optimistic" (force go1.21, which is the version we
currently use to build).

neersighted (approved):

PR #4975, opened 2024-04-01
command: include default otel error handler for the cli

jsternberg:
**- What I did**

This adds a default otel error handler for the cli in the debug package.
It uses logrus to log the error on the debug level and should work out
of the box with the `--debug` flag and `DEBUG` environment variable.

**- How I did it**

Created a function for handling the error that logs the message to `logrus` and registered it with this type: https://pkg.go.dev/go.opentelemetry.io/otel#ErrorHandlerFunc

**- How to verify it**

```
$ DOCKER_CLI_OTEL_EXPORTER_OTLP_ENDPOINT=://blah docker --debug info
DEBU[0000] otel error                                    error="docker otel endpoint is invalid: parse \"://blah\": missing protocol scheme"
```

Without this change, the output is:

```
2024/04/01 09:51:53 docker otel endpoint is invalid: parse "://blah": missing protocol scheme
```

Include otel-related errors in debug logs.

### DIFF
--- a/cli-plugins/plugin/plugin.go
+++ b/cli-plugins/plugin/plugin.go
@@ -12,15 +12,17 @@ import (
 	"github.com/docker/cli/cli-plugins/socket"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/connhelper"
+	"github.com/docker/cli/cli/debug"
 	"github.com/docker/docker/client"
 	"github.com/spf13/cobra"
+	"go.opentelemetry.io/otel"
 )
 
 // PersistentPreRunE must be called by any plugin command (or
 // subcommand) which uses the cobra `PersistentPreRun*` hook. Plugins
 // which do not make use of `PersistentPreRun*` do not need to call
 // this (although it remains safe to do so). Plugins are recommended
-// to use `PersistenPreRunE` to enable the error to be
+// to use `PersistentPreRunE` to enable the error to be
 // returned. Should not be called outside of a command's
 // PersistentPreRunE hook and must not be run unless Run has been
 // called.
@@ -66,6 +68,8 @@ func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager
 
 // Run is the top-level entry point to the CLI plugin framework. It should be called from your plugin's `main()` function.
 func Run(makeCmd func(command.Cli) *cobra.Command, meta manager.Metadata) {
+	otel.SetErrorHandler(debug.OTELErrorHandler)
+
 	dockerCli, err := command.NewDockerCli()
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
--- a/cli/debug/debug.go
+++ b/cli/debug/debug.go
@@ -4,6 +4,7 @@ import (
 	"os"
 
 	"github.com/sirupsen/logrus"
+	"go.opentelemetry.io/otel"
 )
 
 // Enable sets the DEBUG env var to true
@@ -24,3 +25,13 @@ func Disable() {
 func IsEnabled() bool {
 	return os.Getenv("DEBUG") != ""
 }
+
+// OTELErrorHandler is an error handler for OTEL that
+// uses the CLI debug package to log messages when an error
+// occurs.
+//
+// The default is to log to the debug level which is only
+// enabled when debugging is enabled.
+var OTELErrorHandler otel.ErrorHandler = otel.ErrorHandlerFunc(func(err error) {
+	logrus.WithError(err).Debug("otel error")
+})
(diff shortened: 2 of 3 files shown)


cli/command/telemetry_utils.go
```diff
@@ -9,11 +9,19 @@ import (
...
+	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 )
 
+var DefaultOTELErrorHandler otel.ErrorHandler = otel.ErrorHandlerFunc(handleOTELError)
+
+func init() {
+	otel.SetErrorHandler(DefaultOTELErrorHandler)
+}
```
Benehiko, line 23: I'm not a big fan of `init` funcs. Could we do this inside the main func instead?

Is `DefaultOTELErrHandler` necessary? Can't we do this instead
```suggestion
func init() {
	otel.SetErrorHandler(otel.ErrorHandlerFunc(handleOTELError))
}
```

thaJeztah: Yeah, also not a huge fan of `init`. If we _must_ keep this, we could even consider inlinining `handleOTELError`;

```suggestion
func init() {
	otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
		logrus.WithError(err).Debug("otel error")
	}))
}
```

But I'd also be more in favour of having this up to the consumer to decide (and set the error-handler in their `main` or elsewhere).

laurazard: +1 to not having this as an `init`, unless we have strong arguments for it.

(but if we must do it that way, this is one of the less awful uses of `init` and I'd be okay with it 😅)

jsternberg: The main reason is because this handler would also be standard for any plugins and I don't want to necessarily be too invasive in requiring those plugins to know to set the error handler.

I'm not a big fan of the init. I'll see if I can find a more common place to put this code.

The primary reason for `DefaultOTELErrorHandler` is because I also plan that this error handler is used by plugins too and I didn't want to leak the implementation. I can create the variable though and inline the function though since that won't really leak anything.

thaJeztah: Yeah, so I was looking at this from some perspectives;

- it's setting some defaults in an `init`, which may beccome part of imported code in other code-bases
- the default is to set up a debug-logger; performing logging as part of library code is often considered a bit of a "no-no"
- I'm also looking a the `logrus` import in this case, as I know we're trying to move away from hard-coding logrus as implementation for logging; in the moby code-base we started to use the `containerd/log` module (https://github.com/containerd/log), which is currently a wrapper around logrus, but could allow using that abstraction to move to a different implementation, as well as allows for implementations to attach a logger through the context (unfortunately a context-logger doesn't look to be a real option here due to the signature).

jsternberg: This logging is meant to tie in with the `github.com/dockerr/cli/cli/debug` package which explicitly uses logrus for the debugging. What if I were to move the error handler to there, remove the `init()` function, and invoke `otel.SetErrorHandler` on both the `runDocker` function and inside of one of the plugin packages?

krissetto: Another option I was thinking about here could be something like this:

- create an ad-hoc func (eg `setDefaultOtelErrorHandler`) for setting this error handler instead of init, and make sure it only executes once;
- call this handler in both our "provider constructor" funcs, so:
  - in `func (cli *DockerCli) MeterProvider(..)`;
  - and in `func (cli *DockerCli) TracerProvider(..)`

Tiny bit of repetition, but it'd still be transparent to plugins as long as they actually use one of the given providers, and we can avoid the `init()` func which would be run even when no other otel bits are being used.

WDYT?

jsternberg: I'm not the biggest fan of registering these with the provider constructors.

I did a bit of changes and moved the error handler to the debug package that specifically uses `logrus` so it's easier to find and change whenever that part gets refactored. I also changed the `SetErrorHandler` call to be invoked from the `plugin.Run` function and inside of `cmd/docker/main.go` directly so it's no longer in an `init()` function. I think this might solve the problems we were discussing although I don't know if these are the best places for these calls.

cli/command/telemetry_utils.go
```diff
@@ -156,3 +157,7 @@ func getDefaultMeter(mp metric.MeterProvider) metric.Meter {
 		metric.WithInstrumentationVersion(version.Version),
 	)
 }
+
+func handleOTELError(err error) {
+	logrus.WithError(err).Debug("otel error")
+}
```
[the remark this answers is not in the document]
jsternberg: Correct. Let me remove it.

krissetto (approved):
i'm overall ok with this. if we ever do decide to move to using the `containerd/log` module mentioned, we can re-think these few lines

laurazard (approved):
LGTM

PR #4978, opened 2024-04-02
otel: capture whether process was invoked from a terminal

laurazard:
**- What I did**

Added a "terminal" attribute to `BaseMetricAttributes` that allows us to discern whether an invocation was from an interactive terminal or not.

**- How I did it**

**- How to verify it**

image


cli/command/telemetry.go
```diff
@@ -122,7 +123,7 @@ func (r *telemetryResource) init() {
 	r.opts = nil
 }
 
-func (r *telemetryResource) defaultOptions() []resource.Option {
+func defaultOptions() []resource.Option {
```
laurazard, line 126: small tiny refactor here

cli/command/telemetry.go
```diff
@@ -131,6 +132,7 @@ func (r *telemetryResource) defaultOptions() []resource.Option {
 			// OTEL processors may think the same process is restarting
 			// continuously.
 			semconv.ServiceInstanceID(uuid.Generate().String()),
+			semconv.ServiceVersion(version.Version),
```
laurazard, line 135: Unrelated change, but I realized `semconv.ServiceVersion` is probably the attribute we should be using to capture the binary version.

> It represents the version string of the service API or implementation.

thaJeztah: Is this code that will also be consumed/inherited by plugins? (in that case, do they need a different version to be used?). Note that the `version.Version` is set at compile time, so this would work for the CLI itself (binaries built with the `-X ...` options set), but won't be the case for library / module uses (in which case it would be `unknown-version`).

We also have `UserAgent()`, which combines some parts of version (including git-commit), but the same applies there if these variables are not set during compile time.

laurazard: That's a good point! I don't think this code in particular would be used by plugins, only by the main CLI process. This code is configuration for the main CLIs `telemetryResource` which we use to identify one CLI invocation, but plugins will generally configure their own (see Compose's [use](https://github.com/docker/compose/blob/ff20b641c7f1cce588faea476fbd1a6fdd953b36/internal/tracing/tracing.go#L107-L110)), but it
s good to check

laurazard: cc @milas @jsternberg

cli/command/telemetry_utils.go
```diff
@@ -15,9 +15,10 @@ import (
 )
 
 // BaseMetricAttributes returns an attribute.Set containing attributes to attach to metrics/traces
-func BaseMetricAttributes(cmd *cobra.Command) attribute.Set {
+func BaseMetricAttributes(cmd *cobra.Command, cli *DockerCli) attribute.Set {
 	attrList := []attribute.KeyValue{
 		attribute.String("command.name", getCommandName(cmd)),
+		attribute.Bool("terminal", cli.Out().IsTerminal()),
```
laurazard, line 21: I was unsure whether to add this here or to the default resource options, as iiuc we treat one "CLI invocation" as a unique resource (and I guess whether that invocation was interactive/from a terminal or not is an intrinsic attribute of that resource), so open to suggestions

Benehiko: The key `terminal` _might_ be confusing here. I'm not sure in which cases `IsTerminal()` would be false so I don't have a strong opinion here, but it was confusing to me as the docker cli would mostly be executed in a shell even if it isn't an interactive session (script).

cli/command/telemetry_utils.go
```diff
@@ -15,9 +15,10 @@ import (
 )
 
 // BaseMetricAttributes returns an attribute.Set containing attributes to attach to metrics/traces
-func BaseMetricAttributes(cmd *cobra.Command) attribute.Set {
+func BaseMetricAttributes(cmd *cobra.Command, cli *DockerCli) attribute.Set {
```
Benehiko, line 18: I'd rather accept more attribute KeyValue pairs than the dockerCli instance - unless `IsTerminal()` can change at runtime.

cli/command/telemetry_utils.go
laurazard: Happy to be corrected, but generally "terminal" in this context is used interchangeably with "interactive terminal". Even when invoked from a script, what we're talking about (and what `IsTerminal` checks for) is whether the file descriptors attached to the process represent a tty or not.

Other than that,

> docker cli would mostly be executed in a shell even if it isn't an interactive session (script)

I'm not super sure this is accurate, although I guess we can't reliably tell. There's a lot of processes that directly invoke the CLI binary though.

cli/command/telemetry_utils.go
laurazard: Sure, that sounds good to me. I guess it might, but then again I don't think the underlying implementation of `IsTerminal()` gets updated, and in any case if someone runs the CLI from the terminal and then redirects the output then we probably still would like to count it as a "terminal" invocation.

laurazard: What I'd really like to do is set this on the resource level earlier, but by the point the resource is created we don't have everything in place yet to check without doing a larger refactor which I was trying to avoid.

cli/command/telemetry_utils.go
cpuguy83: If you are specifically looking for an interactive terminal, then stdin/stdout would both need to be a terminal for that, but I guess it really depends on what you are looking for here.
Maybe list all the stdios?

laurazard: > Maybe list all the stdios?

In an attribute, independently? That sounds like a good (unopinionated) compromise to me, then downstream processors can decide how to process that information/what constitutes an "interactive" execution.

neersighted: I think I agree with @cpuguy83; an `isatty()` check on each FD might be the most informative (and any tools parsing this data can make decisions based on their own rules, not subject to our opinions in the CLI)...

cli/command/telemetry.go
jsternberg: With the current implementation, each plugin will inherit the version of the docker cli rather than their own version.

From a more generic standpoint, I don't really trust the various `service.x` attributes very much because a user can override them by setting an environment variable. This is why I've preferred to use the instrumentation library version.

I wouldn't oppose us adding this though. You can use `WithResourceOptions` when creating the CLI to inject the service version in a similar way.

cli/command/telemetry_utils.go
jsternberg: It's a bit of cheating but you can use [cmd.Annotations](https://pkg.go.dev/github.com/spf13/cobra#Command). We use that in a few places: https://github.com/docker/cli/blob/5786f20687f334146a229f0954e968c8b43ab92a/cli-plugins/manager/cobra.go#L15-L40

cli/command/telemetry_utils.go
jsternberg: An attribute is likely better here than a resource attribute if the attribute is specific to the one metric.

For the naming, I'd suggest trying to namespace it into something like `command.input.type=tty` or `command.input.type=noninteractive`. I prefer more explicit names to booleans because they're easier to expand and more self-documenting.

For the naming, the way I came up with that is the attribute is related to how the command is invoked. It's related to what type of input is being used. Then it's a general category for the input. This way, if we later decide this format is wrong, we aren't taking a top-level namespace like "terminal" or "command.input" or if we want to add more detail to it, we can do "command.input.user.name" if we wanted the user who invoked the command.

"input" may not be the best name though.

cli/command/telemetry.go
jsternberg: Yea this is fine. I sometimes like to attach functions to structs even when they don't use them just because Go lacks more proper namespaces. That way, I don't get confused about where the function is supposed to be invoked.

jsternberg: If you make this a function instead of a method, maybe change the name to `defaultResourceOptions()` just to be more explicit.

cli/command/telemetry_utils.go
neersighted: I'd suggest going with something like `command.stdin.isatty=true/false` (and also having matching keys for `stdout` and `stderr`) myself; no idea as to the differences between attributes and resouce attributes, but what @jsternberg says sounds reasonable.

laurazard: @jsternberg what do you think about collecting information about the stdios individually (such as `command.stdio.in`, `command.stdio.out`, `command.stdio.err`)?

> An attribute is likely better here than a resource attribute if the attribute is specific to the one metric.

This depends on how we frame it, right? Either the stdios/interactive env is a property of the resource (the CLI invocation, not related to the "command.duration" or any other metric) or it is a property of that specific metric (is tty unique to this/any metric?)

jsternberg: I think names like `command.stdin`, `command.stdout`, and `command.stderr` are fine. `command.stdin.isatty` seems fine to me too if we want to go with a boolean.

neersighted: I think `isatty` makes it unambiguous what is being reported and prevents overloading it with a bad enum in the future.

cli/command/telemetry.go
krissetto: Also good for me, but i'd also rename the func as mentioned for that little bit of extra readability

cli/command/telemetry_utils.go
krissetto: I'd also avoid passing the whole dockerCli struct here, as being a nice and fat struct containing lots of things it seems a bit irrelevant

cli/command/telemetry_utils.go
krissetto: +1 for the `isatty` approach for the individual stdios, as individual attributes. Generally i'd try to process as little as possible and keep the data raw and unopinionated. Any further processing can be done collector-side.

cli/command/telemetry_utils.go
```diff
@@ -27,7 +28,7 @@ func BaseMetricAttributes(cmd *cobra.Command) attribute.Set {
 // Note: this should be the last func to wrap/modify the PersistentRunE/RunE funcs before command execution.
 //
 // can also be used for spans!
-func InstrumentCobraCommands(cmd *cobra.Command, mp metric.MeterProvider) {
+func (cli *DockerCli) InstrumentCobraCommands(cmd *cobra.Command, mp metric.MeterProvider) {
```
krissetto, line 31: similar to above, i don't think we need the whole cli struct here.

I think we should also consolidate our approach when it comes to passing a parameter or attaching a function to a struct. Here we're attaching to the dockerCli struct (which i think could probably be changes if/when we extract shared telemetry bits to a separate pkg), in the `startCobraCommandTimer` and `BaseMetricAttributes` funcs (the only places the dockerCli is needed in here) we pass it as a param instead.

Are we planning on using any other data from the struct other than the `isTerminal()` func?

cli/command/telemetry.go
laurazard: It would still be useful for us to know what version of the CLI a plugin was compiled with though. Maybe using a different, non-env-overrideable attribute, sure.

milas: This makes sense to me. Plugins seem like a separate concern - if we get to `docker/cli` bootstrapping OTel for the plugin, we should probably have it create a resource _for_ the plugin as part of that, at which point it could use the plugin version from `Metadata` etc

Fwiw from the SDK side, I'd assume the env var would take lower precedence than anything explicitly defined in code? That's how e.g. `SERVICE_NAME` works IIRC.
(From a code perspective, when you call `NewResource()`, I think the env detector will run and THEN your overrides.)

milas (approved):
LGTM

cli/command/telemetry.go
jsternberg: It looks like version name isn't overridable through an environment variable. It seems to be only [SERVICE_NAME](https://opentelemetry.io/docs/specs/semconv/resource/#semantic-attributes-with-dedicated-environment-variable).

I think it's ok if we include the version with `service.version`, but I'd like to make sure that the plugins don't inherit that code.

If we want plugins to know what version of the CLI they were invoked with, we can pass this through the environment variable that's being used for that: https://github.com/docker/cli/blob/85dcacd78f2a45886b1ab86ef49fe17014852add/cli-plugins/manager/manager.go#L245

We'd just have to modify that code to also pass the attributes from the resource.

I generally think that using the version that's included from [here](https://pkg.go.dev/go.opentelemetry.io/otel/metric#WithInstrumentationVersion) should be enough. I don't think we need to set `service.version` at the moment.

Also, to my knowledge, the environment variable overrides anything defined in code.

jsternberg: @milas also just thinking out loud, we could just not honor the `OTEL_SERVICE_NAME` environment variable. I think that environment variable is likely more helpful for actual services and not the CLI.

laurazard: Reverted the change for now, but I'll create a ticket so we can look at this later.

cli/command/telemetry_utils.go
laurazard: Cheated a little by passing a `Streams` interface instead, but also did some refactoring to prevent having to drill it down so much. Hopefully that's an acceptable compromise.

Benehiko (approved):
LGTM

Issue #3091, opened 2021-05-12
Image size reported as 1e+03MB by `docker image ls`


rmunn:
**Description**

When creating a Docker image that was almost, but not quite, 1 billion bytes in size, I noticed that `docker image ls` reported the size as `1e+03MB` instead of 1000MB or 1.0GB. I figure that what happened is that the image was less than 1000MB so it didn't get reported as 1.0GB, but then since it was more than 999.5MB, it didn't get reported as 999MB either. This isn't really a problem for me, but it was slightly jarring, and I could conceive of a possible issue where someone's script designed to parse the output of `docker image ls` might barf on an unexpected format for image size.

I did not find any duplicate issues reporting this, probably because it's rare to actually trigger the conditions for this to happen, so I figured I'd report it so that someone can figure out a fix.

**Steps to reproduce the issue:**
1. Create an image between 999,500,000 bytes and 1,000,000,000 bytes in size (mine was 999,517,084 bytes)
2. Run `docker image ls`
3. Note size is reported as `1e+03MB`

**Describe the results you received:**
```
REPOSITORY                       TAG                    IMAGE ID       CREATED          SIZE
tmp-deb-full                     latest                 5fbf6884916b   9 minutes ago    1e+03MB
```

**Describe the results you expected:**
```
REPOSITORY                       TAG                    IMAGE ID       CREATED          SIZE
tmp-deb-full                     latest                 5fbf6884916b   9 minutes ago    1000MB
```

**Additional information you deem important (e.g. issue happens only occasionally):**

My Dockerfile that produced this result:

```
FROM debian:sid

COPY apt-sources.list /etc/apt/sources.list
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y devscripts cli-common-dev sbuild
```

I ran this on 2021-05-12 around 06:00 UTC. Reproducing the exact image I produced might be difficult since any new packages uploaded to Debian sid (whcih gets new packages uploaded all the time) would very likely have a different size and end up with a different image size, and I suspect the bug is triggered only when image size is \>= 999.5MB and \< 1000.0MB. Therefore, I preserved the image in question with `docker image save`. The .tar.xz file is 275 MB, so I won't attach it to this issue, but I can upload it somewhere and make it available on request, if anyone needs it. (Though I suspect one could repro this issue by creating an image full of zero bytes, so you may not need it.)

I've also included the output of `docker image inspect` below in case it helps.

<details>
[...]

PR #4982, opened 2024-04-04
image/list: Add `--tree` flag

vvoland:
- needs: https://github.com/moby/moby/pull/47526
- needs: https://github.com/docker/cli/pull/5330
- needs: https://github.com/moby/moby/pull/47501
- relates to https://github.com/moby/moby/issues/44582

Add `--tree` flag to `docker image list`

image

Update:
Added some spacing around images:
image

```markdown changelog
- containerd image store:  `docker image ls` now supports `--tree` flag that shows a multiplatform-aware image list. This is experimental and may change at any time without any backwards compatibility.
```


cli/command/image/tree.go
```diff
@@ -0,0 +1,244 @@
...
+
+type treeOptions struct {
+	all     bool
+	filters filters.Args
+}
+
+func runTree(ctx context.Context, dockerCLI command.Cli, opts treeOptions) error {
+	images, err := dockerCLI.Client().ImageList(ctx, imagetypes.ListOptions{
+		All:            opts.all,
+		ContainerCount: true,
```
vvoland, line 27: Needs: https://github.com/moby/moby/pull/47501

vvoland:
Combined multiple images with the same target under one entry:

https://github.com/user-attachments/assets/57411c6c-1f09-4ef1-ab9d-b6b809d246c6

@cc @jalonsogo

cli/command/image/list.go
```diff
@@ -59,6 +61,9 @@ func NewImagesCommand(dockerCLI command.Cli) *cobra.Command {
 	flags.StringVar(&options.format, "format", "", flagsHelper.FormatHelp)
 	flags.VarP(&options.filter, "filter", "f", "Filter output based on conditions provided")
 
+	flags.BoolVar(&options.tree, "tree", false, "List multi-platform images tree [experimental, behavior may change]")
```
dvdksn, line 64: There's an annotation for experimental CLI commands and flags. I believe it hides it from `--help` in that case though.

For buildx experimental flags we add a parenthesized all-caps `(EXPERIMENTAL)` suffix.
https://github.com/docker/buildx/blob/e2f6808457ab9d8de4a85f0f26b0028411ef20fc/util/cobrautil/cobrautil.go#L22-L31

```suggestion
	flags.BoolVar(&options.tree, "tree", false, "List multi-platform images tree (EXPERIMENTAL)")
	flags.SetAnnotation("tree", "experimentalCLI", nil)
```

cli/command/image/list.go
```diff
@@ -75,6 +80,26 @@ func runImages(ctx context.Context, dockerCLI command.Cli, options imagesOptions
...
+		}
+		if options.noTrunc {
+			return errors.New("--no-trunc is not (yet) supported with --tree")
+		}
+		if options.showDigests {
+			return errors.New("--show-digest is not (yet) supported with --tree")
+		}
+		if options.format != "" {
+			return errors.New("--format is not (yet) supported with --tree")
+		}
```
dvdksn, line 95: Skip the parentheses?

```suggestion
		if options.quiet {
			return errors.New("--quiet is not yet supported with --tree")
		}
		if options.noTrunc {
			return errors.New("--no-trunc is not yet supported with --tree")
		}
		if options.showDigests {
			return errors.New("--show-digest is not yet supported with --tree")
		}
		if options.format != "" {
			return errors.New("--format is not yet supported with --tree")
		}
```

cli/command/image/tree.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	if width < 20 {
+		width = 20
+	}
+
+	headers := []header{
+		{Title: "Image", Width: 0, Left: true},
+		{Title: "ID", Width: 12},
+		{Title: "Disk use", Width: 8},
+		{Title: "Used", Width: 4},
```
dvdksn, line 110: wdyt?

I prefer "Size" over "Disk use", especially when next to "Used" (or "In use").

```suggestion
		{Title: "Name", Width: 0, Left: true},
		{Title: "ID", Width: 12},
		{Title: "Size", Width: 8},
		{Title: "In use", Width: 6},
```

cli/command/image/list.go
vvoland: This annotation is for features that require `--experimental` flag enabled on the daemon side though.

I can adjust the suffix though

dvdksn: Sorry for the back and forth... I did some digging and it turns out that we strip the `(EXPERIMENTAL)` suffix from the generated buildx doc strings. https://github.com/docker/buildx/blob/e2f6808457ab9d8de4a85f0f26b0028411ef20fc/docs/generate.go#L30-L50

And we use the `experimentalCLI` annotation to [generate an experimental badge](https://github.com/docker/docs/blob/main/layouts/_default/cli.html#L140-L142) on the documentation site (via [docker/cli-docs-tool](https://github.com/docker/cli-docs-tool/blob/main/clidocstool_yaml.go#L178-L180)).

image

so tl;dr: keep the annotation, don't put "experimental" in the usage desc.

cli/command/image/tree.go
vvoland: Size is more ambiguous though - it could mean just the size of the image (e. g. when pushing).
`Disk use` here is the total space occupied on the daemon host both by the distributable images blobs AND the image snapshots (content needed to run containers off this image).

Also, I picked `Used` because the value is either empty or a single tick (✔️), so prefer to keep the column name short to not use too much space.

dvdksn: If we want to keep "Disk use", I would change it to "Disk usage" (sorry, also longer).

I "Used" referring to running containers, or also stopped containers? "Used" is more ambiguous than "In use" w.r.t. to the current state (is it currently in use? has it been used but isn't anymore?). When I initially proposed "In use" I was thinking currently running containers, but I guess it's probably also for stopped containers, in which case "Used" is fine.

vvoland: Stopped containers also count.

cli/command/image/list.go
vvoland: Added the annotation

cli/command/image/tree.go
vvoland: Changed `Disk use` to `Disk usage`

cli/command/image/list.go
dvdksn: I think we want to remove the experimental callout in the usage description too, otherwise it would appear double in the docs site.

```suggestion
	flags.BoolVar(&options.tree, "tree", false, "List multi-platform images tree")
```

vvoland: Hmm, but that won't make it visible in the CLI itself

dvdksn: yeah, hence the buildx hack. I'm fine to duplicate it.

Benehiko (approved):
LGTM :+1:

dvdksn (approved):

cli/command/image/list.go
```diff
@@ -59,6 +61,10 @@ func NewImagesCommand(dockerCLI command.Cli) *cobra.Command {
 	flags.StringVar(&options.format, "format", "", flagsHelper.FormatHelp)
 	flags.VarP(&options.filter, "filter", "f", "Filter output based on conditions provided")
 
+	flags.BoolVar(&options.tree, "tree", false, "List multi-platform images tree (EXPERIMENTAL)")
+	flags.SetAnnotation("tree", "api", []string{"1.47"})
```
thaJeztah, line 65: I think this should be `version` instead of `api`, e.g. see https://github.com/docker/cli/blob/211a5403b3671787c62db68cde3208f6fa95661a/cli/command/image/build.go#L145

cli/command/image/list.go
```diff
@@ -75,6 +81,26 @@ func runImages(ctx context.Context, dockerCLI command.Cli, options imagesOptions
 		filters.Add("reference", options.matchName)
 	}
 
+	if options.tree {
+		if options.quiet {
+			return errors.New("--quiet is not yet supported with --tree")
+		}
```
thaJeztah, line 87: This is probably OK for now (i.e., producing an error); we may have to look what we want the behavior to be. I think for other cases we made the `quiet` always work; i.e. `--format .... --quiet` would continue to just print the `ID` (digest) of the image.

cli/command/image/list.go
```diff
@@ -75,6 +81,26 @@ func runImages(ctx context.Context, dockerCLI command.Cli, options imagesOptions
...
+		if options.noTrunc {
+			return errors.New("--no-trunc is not yet supported with --tree")
+		}
+		if options.showDigests {
+			return errors.New("--show-digest is not yet supported with --tree")
+		}
+		if options.format != "" {
+			return errors.New("--format is not yet supported with --tree")
+		}
```
thaJeztah, line 96: Also slightly wondering if we should use `--format=tree` for this (although `--tree` could be a "porcelain" option for setting that) 🤔

vvoland: `tree` being a `format` would mean taking away the option to specify custom columns though.
Not sure if we want that?

cli/command/image/tree.go
vvoland: We have, but it's only filled if the `ContainerCount` option is set (we can still change that though).

cli/command/image/tree.go
```diff
@@ -0,0 +1,296 @@
...
+			if len(name) > maxImageName {
+				maxImageName = len(name)
+			}
+		}
+		for _, sub := range img.Children {
+			if len(sub.Platform) > maxImageName {
+				maxImageName = len(sub.Platform)
+			}
+		}
+	}
```
laurazard, line 139: These bits of the code are a bit confusing to me – I'm okay merging as it's experimental, but I wonder if we can subtract some of these bits (like call this `calcMaxImageNameLength`, and others) and add some tests for them. OTOH, if we want to merge just to get them out and I'm fine as long as we refactor/merge before making it no longer experimental.

cli/command/image/tree.go
```diff
@@ -0,0 +1,296 @@
...
+			_, _ = fmt.Fprint(out, headers[3].Print(greenColor, "✔"))
+		} else {
+			_, _ = fmt.Fprint(out, headers[3].Print(clr, " "))
+		}
+		_, _ = fmt.Fprint(out, strings.Repeat(" ", headerSpacing))
+	}
+
+	// Print images
+	for _, img := range images {
+		_, _ = fmt.Fprintln(out, "")
```
laurazard, line 202: This looks a bit weird to me, but it's purely a nit/stylistic – I don't super like the newline between the headers and the image names:
Screenshot 2024-08-13 at 10 20 58
IMO this looks a bit better:
Screenshot 2024-08-13 at 10 24 11
Could also have it only between each image "tree" but not before the first one:
Screenshot 2024-08-13 at 10 25 58

cli/command/image/tree.go
```diff
@@ -0,0 +1,296 @@
...
+			} else {
+				_, _ = fmt.Fprint(out, headers[0].Print(clr, "└─ "+sub.Platform))
+			}
+
+			_, _ = fmt.Fprint(out, strings.Repeat(" ", headerSpacing))
+			printDetails(clr, sub.Details)
+
+			_, _ = fmt.Fprintln(out, "")
+		}
+	}
```
laurazard, line 231: If we refactored a bit, this could become something line
```go
	for iid, img := range images {
		printImageNames(out, img, headers, topNameColor)
		printDetails(out, normalColor, img.Details, headers)
		printImageChildren(out, img, headers)
	}
```
and then we'd have
```go
func printImageNames(out io.Writer, img topImage, headers []header, topNameColor aec.ANSI) {
	for nameIdx, name := range img.Names {
		if nameIdx != 0 {
			_, _ = fmt.Fprintln(out, "")
		}
		_, _ = fmt.Fprint(out, headers[0].Print(topNameColor, name))
	}
	_, _ = fmt.Fprint(out, strings.Repeat(" ", headerSpacing))
}

func printDetails(out io.Writer, clr aec.ANSI, details imageDetails, headers []header) {
	if len(headers) <= 1 {
		return
	}
	truncID := stringid.TruncateID(details.ID)
	_, _ = fmt.Fprint(out, headers[1].Print(clr, truncID))
	_, _ = fmt.Fprint(out, strings.Repeat(" ", headerSpacing))

	if len(headers) <= 2 {
		return
	}
	_, _ = fmt.Fprint(out, headers[2].Print(clr, details.DiskUsage))
	_, _ = fmt.Fprint(out, strings.Repeat(" ", headerSpacing))

	if len(headers) <= 3 {
		return
	}
	if details.Used {
		_, _ = fmt.Fprint(out, headers[3].Print(greenColor, "✔"))
	} else {
		_, _ = fmt.Fprint(out, headers[3].Print(clr, " "))
	}
	_, _ = fmt.Fprint(out, strings.Repeat(" ", headerSpacing))
}

func printImageChildren(out io.Writer, img topImage, headers []header) {
	_, _ = fmt.Fprintln(out, "")
	for idx, sub := range img.Children {
		clr := normalColor
		if !sub.Available {
			clr = normalFaintedColor
		}

		if idx != len(img.Children)-1 {
			_, _ = fmt.Fprint(out, headers[0].Print(clr, "├─ "+sub.Platform))
		} else {
			_, _ = fmt.Fprint(out, headers[0].Print(clr, "└─ "+sub.Platform))
		}

		_, _ = fmt.Fprint(out, strings.Repeat(" ", headerSpacing))
		printDetails(out, clr, sub.Details, headers)

		_, _ = fmt.Fprintln(out, "")
	}
}
```

laurazard (approved):
Overall LGTM. Left a few nits/suggestions, and overall I'd like a refactor but fine to leave that out for now since it's experimental.

cli/command/image/tree.go
vvoland: Eventually we should replace all this custom formatting code with something like lipgloss, but I didn't want to introduce a new dependency just yet.
We can probably do that once we start implementing Javier's designs for other commands (and decide which library we want to settle with).

cli/command/image/list.go
vvoland: OH right, good catch!

cli/command/image/list.go
thaJeztah: Depends a bit on how it's implemented. For the default (`table`) output, we have `--format=table` to pick the default table output, but it takes arguments, so can be used to produce a custom table;

```bash
docker image ls --format=table
REPOSITORY       TAG               IMAGE ID       CREATED         SIZE
docker-cli-dev   latest            88c125e27aee   21 hours ago    749MB
docker-dev       latest            60f88d03ca84   4 days ago      2.21GB
golang           1.22              2bd56f00ff47   6 days ago      1.21GB

docker image ls --format "table {{.Tag}}\t{{.ID}}"
TAG               IMAGE ID
latest            88c125e27aee
latest            60f88d03ca84
1.22              2bd56f00ff47
```

cli/command/image/tree.go
laurazard: I think that's fine eventually, but until then I'd prefer to have a nicer implementation – I don't think we should wait on that to at least do a small refactor.

laurazard: Something like this: https://github.com/laurazard/cli/blob/b20c2598ca71e3c0db6247721869a5476d907fff/cli/command/image/tree.go#L97-L252

laurazard: (this also removes the need for that `//nolint:gocyclo` directive).

vvoland: Yeah, not saying to not refactor it now 😅 Will do it!

cli/command/image/tree.go
vvoland: Removed the first newline

cli/command/image/tree.go
laurazard: Looks nicer, thanks!

thaJeztah:
Ugh, sorry, had some partial notes that I didn't post, so it's kinda random;

I noticed the column-headings not aligning with the columns (or, more accurately, they seem to be centered). Is this because we're not using a tabwriter for the table, or could this be things being comnfused because of control-characters (colors)?


```bash
docker image ls --tree
Image                        ID        Disk usage   Used
alpine:latest           0a4eaa0eecf5     25.7MB
├─ linux/arm64/v8       24ba417e25e7     25.7MB
├─ linux/amd64          eddacbc7e24b     12.1MB
├─ linux/arm/v6         5c7e326e3c8a       0B
├─ linux/arm/v7         fda9b1b812b2       0B
├─ linux/386            fa66aa594ffa       0B
├─ linux/ppc64le        a01843eb870e       0B
├─ linux/riscv64        e99a4d9aa9f9       0B
└─ linux/s390x          14da06d3a895       0B
```

Also wondering what the top-level image should show; it currently shows the default platform's information; wondering if (in tree view) that should show the combined size.

vvoland:
That was intended. The header alignment is the same as the value alignment - first column is left aligned (because name column have a variable width) and the others are centered.
Centering the first column header could also work if you think that would look better.

vvoland:
> Also wondering what the top-level image should show; it currently shows the default platform's information; wondering if (in tree view) that should show the combined size.

The top-level image should show the total size and the ID of the manifest list (not a default platform image). So it looks like something is off, will take a look, thanks for noticing!

laurazard:
Voting for left-alined first-column header 😅 Could we also left-align the other columns? I like that more imo

vvoland:
> Could we also left-align the other columns? I like that more imo

You mean whole columns:
﻿﻿
image


or just headers?

image

cli/command/image/tree.go
vvoland: Should I remove the `Usage` column for now?

vvoland:
Also, perhaps we could already include the `Content size` column to start making users aware of the distinction?

image

laurazard:
Whole columns left-aligned looks better to me! (but that's subjective, I'm just casting my vote here)
(this image)

Add tree flag

I'm also a +1 on adding content size, I think this is a great place to introduce this without breaking anything – and users will already be expecting new things here so less confusing.

vvoland:
So, I just bugged @jalonsogo for advice, and got a couple of guildelines to follow, tldr:
- Number aligned right
- Text aligned left
- Header and cell content alignment should match

and ended up with:
image


I made `Used` centered as it can only either be empty or ✓. But that's 4 letters so isn't perfectly centered 🙈

laurazard:
Nice! I didn't know about right-alined numbers, but it makes sense – this way, the decimal places are aligned, which also makes it easier to parse the table and find larger numbers

thaJeztah:
> So, I just bugged @jalonsogo for advice, and got a couple of guildelines to follow, tldr:
> 
> * Number aligned right
> * Text aligned left
> * Header and cell content alignment should match
> 
> I made `Used` centered as it can only either be empty or ✓. But that's 4 letters so isn't perfectly centered 🙈


Ah, yes, I like that variant!


The numbers have always been a bit tricky, and I know we (still) have some bugs in those; the intent was to have a fixed number of decimals so that values align (and pick the most appropriate unit to make that happen), but there's still some bugs there (see https://github.com/docker/cli/issues/3091)

thaJeztah:
OK; one more thing I noticed (but maybe something to also fix on the daemon side not sure?) is the order in which variants are shown;

Output with only arm64;

```bash
docker image ls --tree
Image                        ID        Disk usage   Content size   Used
alpine:latest           0a4eaa0eecf5     13.6MB        4.09MB
├─ linux/arm64/v8       24ba417e25e7     13.6MB        4.09MB
├─ linux/amd64          eddacbc7e24b       0B            0B
├─ linux/arm/v6         5c7e326e3c8a       0B            0B
├─ linux/arm/v7         fda9b1b812b2       0B            0B
├─ linux/386            fa66aa594ffa       0B            0B
├─ linux/ppc64le        a01843eb870e       0B            0B
├─ linux/riscv64        e99a4d9aa9f9       0B            0B
└─ linux/s390x          14da06d3a895       0B            0B
```

Pull s390x;

```bash
docker image pull --platform=linux/s390x alpine
```

New output;

```bash
docker image ls --tree
Image                        ID        Disk usage   Content size   Used
alpine:latest           0a4eaa0eecf5     25.5MB        7.55MB
├─ linux/s390x          14da06d3a895     11.9MB        3.46MB
├─ linux/arm64/v8       24ba417e25e7     13.6MB        4.09MB
├─ linux/amd64          eddacbc7e24b       0B            0B
├─ linux/arm/v6         5c7e326e3c8a       0B            0B
├─ linux/arm/v7         fda9b1b812b2       0B            0B
├─ linux/386            fa66aa594ffa       0B            0B
├─ linux/ppc64le        a01843eb870e       0B            0B
└─ linux/riscv64        e99a4d9aa9f9       0B            0B
```

I wondered where the sort-order is based on; perhaps sorting is happening on "Created" date, but it looks like we omit that in the new view; 


```bash
docker image ls
REPOSITORY                  TAG               IMAGE ID       CREATED         SIZE
docker-dev                  latest            e1ca8a3ea183   6 hours ago     2.2GB
localhost:5001/owner/name   latest            0fea7086b9c0   8 hours ago     17.2MB
docker-cli-dev              latest            88c125e27aee   3 days ago      749MB
golang                      1.22              2bd56f00ff47   8 days ago      1.21GB
moby/buildkit               buildx-stable-1   0a5641c72659   3 weeks ago     298MB
docker                      dind              a69069397655   3 weeks ago     491MB
alpine                      latest            0a4eaa0eecf5   3 weeks ago     25.7MB
registry                    3.0.0-beta.1      efb271da5cd0   5 weeks ago     61.8MB
nginx                       alpine            208b70eefac1   7 weeks ago     68.4MB
registry                    2                 12120425f07d   10 months ago   36.1MB
docker                      20.10-cli         2967f0819c84   15 months ago   209MB
registry                    2.7               169211e20e2f   2 years ago     35.4MB
registry                    2.6               c4bdca23bab1   4 years ago     38.5MB
```

Maybe we should

- have the variants sorted by the first column
- the overall list of images could still be (as default) sorted by created date
- :point_up: that order can sometimes be relevant (e.g. `cut` the last X images)
- :point_up: BUT also admitted that "created" date may be different with containerd (IIRC it uses the "local" date the image was added, not when the image was built not sure?)
- :point_up: Even if it shows the "created" date ("built" date); reproducible builds make that ... less useful (`1-1-1970` or something along those lines :joy:)

Final thought; some what related to https://github.com/moby/moby/pull/48264#issuecomment-2291000564 - wondering if we should have `--all` (or some equivalent) to control whether "non available" variants should be shown / hidden.

- It's good to have the option to show all variants of the image (including those "not present")
- At the same time, it makes the list really long, and sometimes you may only want to see "what's present"
- :point_up: for that we'd have to decide on a default

vvoland:
Variants are sorted by the "availability" so that present variants are always on top.
Added sorting the top level images by the `created` (like the regular list does). Not showing it though for now.
I think we can revisit sorting and `all` to show the unavailable platform in a follow-up.

vvoland:
Also, wondering if we need an additional warning like:

image

cli/command/image/tree.go
```diff
@@ -59,6 +59,11 @@ func runTree(ctx context.Context, dockerCLI command.Cli, opts treeOptions) error
 				},
 			}
 
+			if sub.Details.Used {
+				// Mark top-level parent image as used if any of its subimages are used.
+				details.Used = true
```
thaJeztah, line 64: I forgot now; did we also add the list of container-id's that used it somewhere? (Maybe not to be printed by default for sure)

vvoland: Yes, the `ImageData` has a list of containers. It's only available in the manifest summary though.
Also, see: https://github.com/docker/cli/pull/4982#discussion_r1553521166

docs/reference/commandline/image_ls.md
```diff
@@ -17,6 +17,7 @@ List images
...
 | `-q`, `--quiet`                        | `bool`   |         | Only show image IDs                                                                                                                                                                                                                                                                                                                                                                                                                  |
+| `--tree`                               | `bool`   |         | List multi-platform images tree (EXPERIMENTAL)                                                                                                                                                                                                                                                                                                                                                                                       |
```
krissetto, line 20: nit: `... as a tree ...` here and below

krissetto (approved):
i agree with @laurazard that maybe some things could be refactored, but looks like a good start to me :)

One design consideration though: I tend to think that some whitespace could go a long way in making the output easier to parse (since this is for humans more than for machines, right?). For example when using the command a couple times i find it hard to visually parse the beginning of the output from the command that the user typed, and also to read the column titles from the image names listed since they're all back to back (and title+first image name are both bold).

Maybe to separate column titles from the column contents they could:
- have some whitespace around them, separating them from contents and the command the user typed;
- be a different color;
- have a line under them like an old school table;

Just throwing some thoughts out there. WDYT here @jalonsogo? Maybe some discussions have already been made around this, if so feel free to ignore this comment :) 

Given the experimental nature, we can definitely iterate on these things. It'd also be nice to consolidate our design and usage of columns/tables etc amongst all our other commands as well

This command has nice (imho) whitespace in it's output, for example
Screenshot 2024-08-16 at 12 52 32

vvoland:
Initially there was a newline between header and the first item, but removed it after a small discussion here: https://github.com/docker/cli/pull/4982#discussion_r1714969244

Regarding the newline before the command content.. I think it might work:
image

vvoland:
Removed the `Used` column for now.

jalonsogo:
In the CLI design guidelines, I recommend using capitalization+bold for the headers.
Also, in my designs, I use them in combination with colors:
- Bright White  `\u001b[37m;1m` for the primary chunks of information
- Write-default  `\u001b[37m` for the secondary chucks of information 

CleanShot 2024-08-16 at 13 28 15

PR #4983, opened 2024-04-04
image: Add `convert` command

vvoland:
- https://github.com/moby/moby/pull/47678

Add `docker image convert` command

image

```markdown changelog
- containerd image store: Add `docker image convert` command
```


cli/command/image/convert.go
```diff
@@ -0,0 +1,70 @@
...
+func NewConvertCommand(dockerCli command.Cli) *cobra.Command {
+	var args convertArgs
+
+	cmd := &cobra.Command{
+		Use:   "convert [OPTIONS] <from> <to>",
+		Short: "Convert multi-platform images",
+		Args:  cli.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, posArgs []string) error {
+			args.Src = posArgs[0]
+			args.Dst = posArgs[1]
```
vvoland, line 31: Change into flag arguments.

cli/command/image/convert.go
```diff
@@ -0,0 +1,83 @@
...
+			"aliases": "docker image convert, docker convert",
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringArrayVar(&args.Platforms, "platforms", nil, "Include only the specified platforms in the destination image")
+	flags.BoolVar(&args.NoAttestations, "no-attestations", false, "Do not include image attestations")
+	flags.BoolVar(&args.OnlyAvailablePlatforms, "available", false, "Only include platforms locally available to the daemon")
+	flags.StringArrayVar(&args.Dst, "to", nil, "Target image references")
+	flags.StringVar(&args.Src, "from", "", "Source image reference")
```
AkihiroSuda, line 44: nerdctl already has `nerdctl image convert SRC DST` with a different purpose (akin to `ctr images convert`) : 
https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#nerd_face-nerdctl-image-convert

Would it be possible to adopt the same syntax as `nerdctl image convert SRC DST`, or change the command name to avoid conflict?

vvoland: I think in general, the purpose is the same. Unfortunately I don't have any better idea for a name... Do you have any suggestions?

We deliberately chose the src and dst to be flag arguments instead of positional arguments as it gives more clarity, especially when multiple destinations might be specified.

laurazard: I don't think we should change our syntax (esp. since we settled on this one as the most user-friendly/better option) just for compatibility's sake.

Unless there's a more explicit/as clear name, I also don't think we should change the name to avoid "colliding" with `nerdctl`s command, although I do understand the possible confusion if we have two commands with the same name/subcommand and different syntax.

AkihiroSuda: > We deliberately chose the src and dst to be flag arguments instead of positional arguments as it gives more clarity, especially when multiple destinations might be specified.

What about making it `(docker|nerdctl) image convert SRC DST1 DST2 DST3`, or `(docker|nerdctl) image convert SRC DST1,DST2,DST3`?

cli/command/image/convert.go
```diff
@@ -0,0 +1,83 @@
...
+			return runConvert(cmd.Context(), dockerCli, args)
+		},
+		Aliases: []string{"convert"},
+		Annotations: map[string]string{
+			"aliases": "docker image convert, docker convert",
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringArrayVar(&args.Platforms, "platforms", nil, "Include only the specified platforms in the destination image")
```
AkihiroSuda, line 40: This is named "platform" rather than "platforms" in nerdctl and ctr

cli/command/image/convert.go
cpuguy83: Definitely a fan of positional for this but I haven't looked at the PR yet, just this thread.

vvoland: cc @tianon - I think you had a strong opinion on it being a flag argument 😄

tianon: Took me a minute, but I remember now what my thoughts were. :sweat_smile:

IMO, this command should match `docker build` -- ala, positional arguments for input, and `--tag`/`-t` for output(s).

(I think we could make the same argument for `docker push -t foo -t bar -t baz image-to-push`)

In the case of convert, we might have multiple _inputs_ as well, right?  For example, if we have a way to remove things from an index, wouldn't it make sense to _also_ have a way to add things to one / create one?  That could then be multiple positional arguments (now unambiguous if we avoid destination being positional).

PR #4984, opened 2024-04-04
cli/push: Add `platform` switch 

vvoland:
- needs: https://github.com/moby/moby/pull/47679

**- What I did**
Added a `platform` switch to `docker image push`.

**- How I did it**

**- How to verify it**
```bash
$ docker pull --platform linux/arm64 busybox
$ docker tag busybox myimage
$ docker push --platform linux/arm64 myimage
```

```markdown changelog
containerd image store: Add `--platform` switch to `docker image push`.
```

### DIFF
--- a/cli/command/image/push.go
+++ b/cli/command/image/push.go
@@ -2,6 +2,7 @@ package image
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -12,11 +13,13 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/completion"
 	"github.com/docker/cli/cli/streams"
+	"github.com/docker/docker/api/types/auxprogress"
 	"github.com/docker/docker/api/types/image"
 	registrytypes "github.com/docker/docker/api/types/registry"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/registry"
 	"github.com/moby/term"
+	"github.com/morikuni/aec"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -62,6 +65,8 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
 }
 
 // RunPush performs a push against the engine based on the specified options
+//
+//nolint:gocyclo
 func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error {
 	var platform *ocispec.Platform
 	if opts.platform != "" {
@@ -74,9 +79,8 @@ func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error
 
 		printNote(dockerCli, `Selecting a single platform will only push one matching image manifest from a multi-platform image index.
 This means that any other components attached to the multi-platform image index (like Buildkit attestations) won't be pushed.
-If you want to only push a single platform image while preserving the attestations, please build an image with only that platform and push it instead.
-Example: echo "FROM %s" | docker build - --platform %s -t <NEW-TAG>
-`, opts.remote, opts.platform)
+If you want to only push a single platform image while preserving the attestations, please use 'docker convert\n'
+`)
 	}
 
 	ref, err := reference.ParseNormalizedNamed(opts.remote)
@@ -117,6 +121,13 @@ Example: echo "FROM %s" | docker build - --platform %s -t <NEW-TAG>
 		return err
 	}
 
+	defer func() {
+		for _, note := range notes {
+			fmt.Fprintln(dockerCli.Err(), "")
+			printNote(dockerCli, note)
+		}
+	}()
+
 	defer responseBody.Close()
 	if !opts.untrusted {
 		// TODO PushTrustedReference currently doesn't respect `--quiet`
@@ -124,20 +135,51 @@ Example: echo "FROM %s" | docker build - --platform %s -t <NEW-TAG>
 	}
 
 	if opts.quiet {
-		err = jsonmessage.DisplayJSONMessagesToStream(responseBody, streams.NewOut(io.Discard), nil)
+		err = jsonmessage.DisplayJSONMessagesToStream(responseBody, streams.NewOut(io.Discard), handleAux(dockerCli))
 		if err == nil {
 			fmt.Fprintln(dockerCli.Out(), ref.String())
 		}
 		return err
 	}
-	return jsonmessage.DisplayJSONMessagesToStream(responseBody, dockerCli.Out(), nil)
+	return jsonmessage.DisplayJSONMessagesToStream(responseBody, dockerCli.Out(), handleAux(dockerCli))
+}
+
+var notes []string
+
+func handleAux(dockerCli command.Cli) func(jm jsonmessage.JSONMessage) {
+	return func(jm jsonmessage.JSONMessage) {
+		b := []byte(*jm.Aux)
+
+		var stripped auxprogress.ManifestPushedInsteadOfIndex
+		err := json.Unmarshal(b, &stripped)
+		if err == nil && stripped.ManifestPushedInsteadOfIndex {
+			note := fmt.Sprintf("Not all multiplatform-content is present and only the available single-platform image was pushed\n%s -> %s",
+				aec.RedF.Apply(stripped.OriginalIndex.Digest.String()),
+				aec.GreenF.Apply(stripped.SelectedManifest.Digest.String()),
+			)
+			notes = append(notes, note)
+		}
+
+		var missing auxprogress.ContentMissing
+		err = json.Unmarshal(b, &missing)
+		if err == nil && missing.ContentMissing {
+			note := `You're trying to push a manifest list/index which 
+				references multiple platform specific manifests, but not all of them are available locally
+				or available to the remote repository.
+
+				Make sure you have all the referenced content and try again.
+
+				You can also push only a single platform specific manifest directly by specifying the platform you want to push with the --platform flag.`
+			notes = append(notes, note)
+		}
+	}
 }
 
 func printNote(dockerCli command.Cli, format string, args ...any) {
 	if _, isTTY := term.GetFdInfo(dockerCli.Err()); isTTY {
-		_, _ = fmt.Fprint(dockerCli.Err(), "\x1b[1;37m\x1b[1;46m[ NOTE ]\x1b[0m\x1b[0m ")
+		_, _ = fmt.Fprint(dockerCli.Err(), aec.WhiteF.Apply(aec.CyanB.Apply("[ NOTE ]"))+" ")
 	} else {
 		_, _ = fmt.Fprint(dockerCli.Err(), "[ NOTE ] ")
 	}
-	_, _ = fmt.Fprintf(dockerCli.Err(), format+"\n\n", args...)
+	_, _ = fmt.Fprintf(dockerCli.Err(), aec.Bold.Apply(format)+"\n", args...)
 }
--- a/cli/command/image/push.go
+++ b/cli/command/image/push.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 
+	"github.com/containerd/platforms"
 	"github.com/distribution/reference"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
@@ -14,6 +16,8 @@ import (
 	registrytypes "github.com/docker/docker/api/types/registry"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/registry"
+	"github.com/moby/term"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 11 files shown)


cli/command/image/push.go
```diff
@@ -84,6 +86,7 @@ func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error
 		All:           opts.all,
 		RegistryAuth:  encodedAuth,
 		PrivilegeFunc: requestPrivilege,
+		Platform:      opts.platform,
```
vvoland, line 89: Warn that it will strip attestations (suggested by @thaJeztah)

cli/command/image/push.go
```diff
@@ -48,12 +52,30 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
...
+	flags.SetAnnotation("platform", "version", []string{"1.45"})
 
 	return cmd
 }
 
 // RunPush performs a push against the engine based on the specified options
 func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error {
+	if opts.platform != "" {
+		if _, isTty := term.GetFdInfo(dockerCli.Err()); isTty {
+			_, _ = fmt.Fprint(dockerCli.Err(), "\x1b[1;37m\x1b[1;46m[ NOTE ]\x1b[0m\x1b[0m ")
```
rumpl, line 69: There is no need to repeat the reset here, only one is enough: `\x1b[0m\x1b[0m` -> `\x1b[0m`

rumpl: Do we really want to add the cyan background color here? Why not `\x1b[36m` to just color the text?

cli/command/image/push.go
```diff
@@ -48,12 +52,30 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
...
+		} else {
+			_, _ = fmt.Fprint(dockerCli.Err(), "[ NOTE ] ")
+		}
+		_, _ = fmt.Fprintln(dockerCli.Err(), `Selecting a single platform for the push operation will push the image manifest for that platform only.
+This won't push the image index/manifest list which means that other components like Buildkit attestations won't be pushed.
```
rumpl, line 74: Thinking about the wording here, we are talking about `image manifests`, `index`, `manifest list`, do we really need to be this specific? Shouldn't we have a more user-friendly terms like "single-platform image" and "multi-platform image"? 

No one really wants to know the difference between an index and a manifest list :)

cli/command/image/push.go
```diff
@@ -48,12 +52,30 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
...
+If you want to only push a single platform while preserving the attestations, please build an image with only that platform and push it instead.`)
```
rumpl, line 75: Related to the above, after talking about manifests manifest lists and indexes we say "single platform"

cli/command/image/push.go
vvoland: No strong opinions on the colors, I just went with a colored background to make it more visible:

image

It could look wildly different on other terminals/color schemes, so perhaps a text color would be a safer option.

cli/command/image/push.go
vvoland: The problem is that "single platform" isn't always a 1:1 mapping between an index and a single manifest.

Is the index below a single platform image or a multi-platform image?

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:34b7d4a2f050f8a9077fd435b3b1778e091af743f0f4c8c47d109cfda47b0c48",
      "size": 480,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:a7931924962b793438f641b320d496ebca34968e850f7b8d7a5ea59dc88283cc",
      "size": 565,
      "annotations": {
        "vnd.docker.reference.digest": "sha256:34b7d4a2f050f8a9077fd435b3b1778e091af743f0f4c8c47d109cfda47b0c48",
        "vnd.docker.reference.type": "attestation-manifest"
      },
      "platform": {
        "architecture": "unknown",
        "os": "unknown"
      }
    }
  ]
}
```

cli/command/image/push.go
laurazard: I think setting a colored background + colored text is safer – if we just set the foregroung color and the users color scheme uses the same color for the background (and I've definitely seen cyan backgrounds), then the text will be unreadable. If we set a contrasting background+foreground color combination then that's less likely to happen.

cli/command/image/push.go
```diff
@@ -48,12 +54,33 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
...
+		if err != nil {
+			_, _ = fmt.Fprintf(dockerCli.Err(), "Invalid platform %s", opts.platform)
+			return err
+		}
+		platform = &p
+
+		printNote(dockerCli, `Selecting a single platform will only push one matching image manifest from a multi-platform image index.
+This means that any other components attached to the multi-platform image index (like Buildkit attestations) won't be pushed.
+If you want to only push a single platform image while preserving the attestations, please use 'docker convert\n'
+`)
```
laurazard, line 81: I think it's fine to be explicit here, I'd rather some users get annoyed by the extra message than having people not realize we're stripping attestations when they push a single-platform image.

laurazard (approved):
LGTM, I think we can discuss/keep improving the messages in the future too, but I wouldn't block this on that.

thaJeztah:
https://github.com/moby/moby/pull/47943 was merged

laurazard:
@krissetto @Benehiko can you TAL?

cli/command/image/push.go
```diff
@@ -48,12 +52,32 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
...
+		`Push a platform-specific manifest as a single-platform image to the registry.
+'os[/arch[/variant]]': Explicit platform (eg. linux/amd64)`)
+	flags.SetAnnotation("platform", "version", []string{"1.46"})
 
 	return cmd
 }
 
 // RunPush performs a push against the engine based on the specified options
 func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error {
+	var platform *platforms.Platform
```
thaJeztah, line 69: Can you change this one to an `ocispec.Platform`? The `platforms.Platform` type is an alias, and it will be deprecated.

cli/command/image/push.go
```diff
@@ -84,25 +111,84 @@ func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error
...
 		return err
 	}
-	return jsonmessage.DisplayJSONMessagesToStream(responseBody, dockerCli.Out(), nil)
+	return jsonmessage.DisplayJSONMessagesToStream(responseBody, dockerCli.Out(), handleAux(dockerCli))
+}
+
+var notes []string
+
+func handleAux(dockerCli command.Cli) func(jm jsonmessage.JSONMessage) {
+	return func(jm jsonmessage.JSONMessage) {
```
Benehiko, line 148: what happens if none of data could be unmarshalled? should it not error?

cli/command/image/push.go
```diff
@@ -48,12 +52,32 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
...
+'os[/arch[/variant]]': Explicit platform (eg. linux/amd64)`)
+	flags.SetAnnotation("platform", "version", []string{"1.46"})
 
 	return cmd
 }
 
 // RunPush performs a push against the engine based on the specified options
 func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error {
+	var platform *platforms.Platform
+	if opts.platform != "" {
```
thaJeztah, line 71: Not for this PR, and not urgent, but if we decide to add more locations to pass `--platform`, and we want to consume those as a `platforms, we could consider implementing a platform _option_ that can be used for flags, performs the validation as part of that, and directly setting a `Platform`.

Benehiko (approved):
Looks good to me

cli/command/image/push.go
vvoland: There might be other aux progress messages that are not (yet) supported by the CLI. I think it's best to just skip them instead.

cli/command/image/push.go
vvoland: Done!

cli/command/image/push.go
```diff
@@ -84,25 +112,84 @@ func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error
...
+func green(text string) string {
+	return fmt.Sprintf("\033[32m%s\033[0m", text)
+}
+
+func bold(text string) string {
+	return fmt.Sprintf("\033[1m%s\033[0m", text)
+}
+
+func printNote(dockerCli command.Cli, format string, args ...any) {
+	if _, isTty := term.GetFdInfo(dockerCli.Err()); isTty {
```
thaJeztah, line 189: Note that `dockerCli.Out()` has a `IsTerminal()` function that we can use. Not sure if that's fully correct though (is that correct if `STDERR` redirected, but `STDOUT` isn't?

We should consider providing this for _both_ `Out()` and `Err()` (I looked at that a few times, and wondered why we didn't)

cli/command/image/push.go
```diff
@@ -84,25 +112,84 @@ func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error
...
+	return fmt.Sprintf("\033[31m%s\033[0m", text)
+}
+
+func green(text string) string {
+	return fmt.Sprintf("\033[32m%s\033[0m", text)
+}
+
+func bold(text string) string {
+	return fmt.Sprintf("\033[1m%s\033[0m", text)
+}
```
thaJeztah, line 186: Should we use the `aec` utilities for this? I know we use those in other places; https://github.com/docker/cli/blob/0022fe711170b7e15daf5f75bd14074e423f5f5a/vendor/github.com/morikuni/aec/builder.go#L156-L159

vvoland: Oh, good catch. I missed that we already have `aec` vendored! Changed.

cli/command/image/push.go
vvoland: Yeah I think we should, but left that to a second PR that will reuse this logic.

Btw, we already have it, but for the "string" platform: https://github.com/docker/cli/blob/0022fe711170b7e15daf5f75bd14074e423f5f5a/cli/command/utils.go#L158 🙈

cli/command/image/push.go
vvoland: The most common case of this is when you redirect the stderr to `/dev/null`:

```
$ docker push dckr.woland.xyz/ububuu
Using default tag: latest
The push refers to repository [dckr.woland.xyz/ububuu]
aa21f24e1940: Layer already exists
latest: digest: sha256:17c24d16d63d2d089db74c2ed3e99c1ab0fd3f4f93c00b04afa8793fa793626c size: 424

[ NOTE ] Not all multiplatform-content is present and only the available single-platform image was pushed
sha256:e3f92abc0967a6c19d0dfa2d55838833e947b9d74edbcb0113e48535ad4be12a -> sha256:17c24d16d63d2d089db74c2ed3e99c1ab0fd3f4f93c00b04afa8793fa793626c

$ docker push dckr.woland.xyz/ububuu 2>/dev/null
Using default tag: latest
The push refers to repository [dckr.woland.xyz/ububuu]
aa21f24e1940: Layer already exists
latest: digest: sha256:17c24d16d63d2d089db74c2ed3e99c1ab0fd3f4f93c00b04afa8793fa793626c size: 424
```

thaJeztah: Yeah, I think there's other places where we do it wrong, and only look at `STDOUT`, but not at `STDERR` (but switch both based on `STDOUT`; here's a recent discussion that I still need to reply to, because I think the same is happening for `docker build` ; https://github.com/moby/moby/discussions/47755


Perhaps a good reason to look if we should make `Err()` provide thee same features as `Out()`, so that it's easier to check for each of the outputs if they have a TTY attached or not.

thaJeztah: In either case; not a blocker (should I mention that `isTty` should be `isTTY` ? 😂 🙈 )

thaJeztah (approved):
LGTM

I think most comments still visible are good for follow-up work.

cli/command/image/push.go
vvoland: Will open a follow-up with the `stderr` change (it will have a bunch of updates in other places, so prefer to handle it in a separate PR).

vvoland: > `isTTY`

LOL, you're just merciless Sebastiaan 🤣 
...done!

cli/command/image/push.go
thaJeztah: Ah! We should look where/how that's used. If the result is only "internal", it could still make sense to change it, and have the consumer convert it to a string (where needed 🤔).

In either case; all for separate work, just that I thought of it.

cli/command/image/push.go
thaJeztah: > so prefer to handle it in a separate PR

Yes; that's for sure something separate; and it's not a new thing, just that this is (I think) the first time we pay attention and check the right thing for a TTY 😂 

> OL, you're just merciless Sebastiaan 🤣


I'm the Human Linter! 😂

thaJeztah (approved):
LGTM

Let's go!

PR #4985, opened 2024-04-04
otel: capture exit code as int64

laurazard:
**- What I did**

Capture exit code as int64 (it used to be a string!)

**- How I did it**

By changing characters in source files!

**- How to verify it**

Run the relevant tests.


jsternberg (approved):
Yea I think this is much better.

neersighted (approved):

PR #4993, opened 2024-04-08
feat: global signal handling to cancel ctx for graceful exits

Benehiko:
Since other repositories import from the CLI, the signal termination handling done inside the `PromptForConfirmation` could cause problems for the third party repositories. Thus this PR hoists the signal termination out of the `PromptForConfirmation` function.

**- What I did**
Hoist the signal handling from the `PromptForConfirmation` function to the more appropriate top-level function `runDocker`. 

**- How I did it**
Using the `signal` package I catch context cancellations as well as termination signals then pass the returned context to cobra's `Execute` method.

**- How to verify it**
Tests

### DIFF
--- a/cli/command/container/attach.go
+++ b/cli/command/container/attach.go
@@ -105,7 +105,12 @@ func RunAttach(ctx context.Context, dockerCLI command.Cli, containerID string, o
 
 	if opts.Proxy && !c.Config.Tty {
 		sigc := notifyAllSignals()
-		go ForwardAllSignals(ctx, apiClient, containerID, sigc)
+		// since we're explicitly setting up signal handling here, and the daemon will
+		// get notified independently of the clients ctx cancellation, we use this context
+		// but without cancellation to avoid ForwardAllSignals from returning
+		// before all signals are forwarded.
+		bgCtx := context.WithoutCancel(ctx)
+		go ForwardAllSignals(bgCtx, apiClient, containerID, sigc)
 		defer signal.StopCatch(sigc)
 	}
 
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -150,7 +150,12 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 	}
 	if runOpts.sigProxy {
 		sigc := notifyAllSignals()
-		go ForwardAllSignals(ctx, apiClient, containerID, sigc)
+		// since we're explicitly setting up signal handling here, and the daemon will
+		// get notified independently of the clients ctx cancellation, we use this context
+		// but without cancellation to avoid ForwardAllSignals from returning
+		// before all signals are forwarded.
+		bgCtx := context.WithoutCancel(ctx)
+		go ForwardAllSignals(bgCtx, apiClient, containerID, sigc)
 		defer signal.StopCatch(sigc)
 	}
 
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -28,12 +28,20 @@ import (
 )
 
 func main() {
-	ctx := context.Background()
+	statusCode := dockerMain()
+	if statusCode != 0 {
+		os.Exit(statusCode)
+	}
+}
+
+func dockerMain() int {
+	ctx, cancelNotify := signal.NotifyContext(context.Background(), platformsignals.TerminationSignals...)
+	defer cancelNotify()
 
 	dockerCli, err := command.NewDockerCli(command.WithBaseContext(ctx))
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
-		os.Exit(1)
+		return 1
 	}
 	logrus.SetOutput(dockerCli.Err())
 	otel.SetErrorHandler(debug.OTELErrorHandler)
@@ -46,16 +54,17 @@ func main() {
 			// StatusError should only be used for errors, and all errors should
 			// have a non-zero exit status, so never exit with 0
 			if sterr.StatusCode == 0 {
-				os.Exit(1)
+				return 1
 			}
-			os.Exit(sterr.StatusCode)
+			return sterr.StatusCode
 		}
 		if errdefs.IsCancelled(err) {
-			os.Exit(0)
+			return 0
 		}
 		fmt.Fprintln(dockerCli.Err(), err)
-		os.Exit(1)
+		return 1
 	}
+	return 0
 }
 
 func newDockerCommand(dockerCli *command.DockerCli) *cli.TopLevelCommand {
@@ -224,7 +233,7 @@ func setValidateArgs(dockerCli command.Cli, cmd *cobra.Command) {
 	})
 }
 
-func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string, envs []string) error {
+func tryPluginRun(ctx context.Context, dockerCli command.Cli, cmd *cobra.Command, subcommand string, envs []string) error {
 	plugincmd, err := pluginmanager.PluginRunCommand(dockerCli, subcommand, cmd)
 	if err != nil {
 		return err
@@ -242,40 +251,56 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string, 
 
 	// Background signal handling logic: block on the signals channel, and
 	// notify the plugin via the PluginServer (or signal) as appropriate.
-	const exitLimit = 3
-	signals := make(chan os.Signal, exitLimit)
-	signal.Notify(signals, platformsignals.TerminationSignals...)
+	const exitLimit = 2
+
+	tryTerminatePlugin := func(force bool) {
+		// If stdin is a TTY, the kernel will forward
+		// signals to the subprocess because the shared
+		// pgid makes the TTY a controlling terminal.
+		//
+		// The plugin should have it's own copy of this
+		// termination logic, and exit after 3 retries
+		// on it's own.
+		if dockerCli.Out().IsTerminal() {
+			return
+		}
+
+		// Terminate the plugin server, which will
+		// close all connections with plugin
+		// subprocesses, and signal them to exit.
+		//
+		// Repeated invocations will result in EINVAL,
+		// or EBADF; but that is fine for our purposes.
+		_ = srv.Close()
+
+		// force the process to terminate if it hasn't already
+		if force {
+			_ = plugincmd.Process.Kill()
+			_, _ = fmt.Fprint(dockerCli.Err(), "got 3 SIGTERM/SIGINTs, forcefully exiting\n")
+			os.Exit(1)
+		}
+	}
+
 	go func() {
 		retries := 0
+		force := false
+		// catch the first signal through context cancellation
+		<-ctx.Done()
+		tryTerminatePlugin(force)
+
+		// register subsequent signals
+		signals := make(chan os.Signal, exitLimit)
+		signal.Notify(signals, platformsignals.TerminationSignals...)
+
 		for range signals {
-			// If stdin is a TTY, the kernel will forward
-			// signals to the subprocess because the shared
-			// pgid makes the TTY a controlling terminal.
-			//
-			// The plugin should have it's own copy of this
-			// termination logic, and exit after 3 retries
-			// on it's own.
-			if dockerCli.Out().IsTerminal() {
-				continue
-			}
-
-			// Terminate the plugin server, which will
-			// close all connections with plugin
-			// subprocesses, and signal them to exit.
-			//
-			// Repeated invocations will result in EINVAL,
-			// or EBADF; but that is fine for our purposes.
-			_ = srv.Close()
-
+			retries++
 			// If we're still running after 3 interruptions
 			// (SIGINT/SIGTERM), send a SIGKILL to the plugin as a
 			// final attempt to terminate, and exit.
-			retries++
 			if retries >= exitLimit {
-				_, _ = fmt.Fprintf(dockerCli.Err(), "got %d SIGTERM/SIGINTs, forcefully exiting\n", retries)
-				_ = plugincmd.Process.Kill()
-				os.Exit(1)
+				force = true
 			}
+			tryTerminatePlugin(force)
 		}
 	}()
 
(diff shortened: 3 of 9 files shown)


Benehiko:
Thinking about this now, maybe I should just do this top-level inside the `main` function and then rework the plugin signal handling as well to accept a `context.Done()` instead.

cmd/docker/docker.go
```diff
@@ -241,11 +246,15 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
 
 	// Background signal handling logic: block on the signals channel, and
 	// notify the plugin via the PluginServer (or signal) as appropriate.
-	const exitLimit = 3
-	signals := make(chan os.Signal, exitLimit)
-	signal.Notify(signals, platformsignals.TerminationSignals...)
+	const exitLimit = 2
```
Benehiko, line 249: reduce limit to 2 here since we already wait for `ctx.Done` below

cmd/docker/docker.go
```diff
@@ -241,11 +246,15 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
...
+
 	go func() {
 		retries := 0
+		<-ctx.Done()
+
+		signals := make(chan os.Signal, exitLimit)
+		signal.Notify(signals, platformsignals.TerminationSignals...)
```
Benehiko, line 256: Setup another signal to continue with the 3 retry logic

Benehiko:
This PR is blocked by https://github.com/moby/moby/pull/47536. The cobra context isn't given to the prompt function when `plugin install` is called due to `acceptPrivileges` not accepting a context https://github.com/docker/cli/blob/master/cli/command/plugin/install.go#L145

laurazard:
Looks like the linter's complaining, and some tests are broken because now the output from the graceful termination logic is different:

```
=== FAIL: e2e/cli-plugins TestPluginSocketCommunication/detached/the_main_CLI_exits_after_3_signals (1.10s)
    socket_test.go:232: assertion failed: 
        --- ←
        +++ →
        @@ -1,2 +1,2 @@
        -got 2 SIGTERM/SIGINTs, forcefully exiting
        +got 3 SIGTERM/SIGINTs, forcefully exiting
```

I think it should be fine to hardcode the `3` in that print, since that's the total number of signals we've received (counting the one that cancelled the context. It might also be possible to reword it some other way 😅 

---------------

I guess we'll hold off on merging this one since it needs changes to moby that we want to get in after 26.1, so we can get it in after we cut a release.

Benehiko:
> Looks like the linter's complaining, and some tests are broken because now the output from the graceful termination logic is different:
> 
> ```
> === FAIL: e2e/cli-plugins TestPluginSocketCommunication/detached/the_main_CLI_exits_after_3_signals (1.10s)
>     socket_test.go:232: assertion failed: 
>         --- ←
>         +++ →
>         @@ -1,2 +1,2 @@
>         -got 2 SIGTERM/SIGINTs, forcefully exiting
>         +got 3 SIGTERM/SIGINTs, forcefully exiting
> ```
> 
> I think it should be fine to hardcode the `3` in that print, since that's the total number of signals we've received (counting the one that cancelled the context. It might also be possible to reword it some other way 😅
> 
> I guess we'll hold off on merging this one since it needs changes to moby that we want to get in after 26.1, so we can get it in after we cut a release.

Ah yeah, the tests won't pass right now. Not without the changes to moby and context passing in CLI. 

I'll hardcode the `3` :)

laurazard (approved):
LGTM overall, but looks like there's some tests failing/linters complaining. I wonder if were doing things correctly with those deferred `cancel()`s in `main`.

Benehiko:
> LGTM overall, but looks like there's some tests failing/linters complaining. I wonder if were doing things correctly with those deferred `cancel()`s in `main`.

I have been looking into it. I've fixed one of the tests, specifically the plugins e2e signal handling and still busy on the containers e2e

cli/command/container/attach.go
```diff
@@ -105,7 +105,13 @@ func RunAttach(ctx context.Context, dockerCLI command.Cli, containerID string, o
 
 	if opts.Proxy && !c.Config.Tty {
 		sigc := notifyAllSignals()
-		go ForwardAllSignals(ctx, apiClient, containerID, sigc)
+		// since the ForwardAllSignals already registers its own signal handler
+		// and should notify the daemon independently of the clients ctx cancellation
+		// we should use a background context to avoid the ForwardAllSignals from
+		// returning before all signals are forwarded to the daemon
+		bgCtx, cancel := context.WithCancel(context.Background())
```
vvoland, line 112: Instead of using `context.Background()` we can just ignore cancellation from the passed ctx:

```suggestion
		bgCtx, cancel := context.WithCancel(context.WithoutCancel(ctx))
```

This will still persist values attached to the context, so we won't lose OTEL spans for example.

cli/command/container/run.go
```diff
@@ -150,7 +150,13 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 	}
 	if runOpts.sigProxy {
 		sigc := notifyAllSignals()
-		go ForwardAllSignals(ctx, apiClient, containerID, sigc)
+		// since the ForwardAllSignals already registers its own signal handler
+		// and should notify the daemon independently of the clients ctx cancellation
+		// we should use a background context to avoid the ForwardAllSignals from
+		// returning before all signals are forwarded to the daemon
+		bgCtx, cancel := context.WithCancel(context.Background())
```
vvoland, line 157: Same as above

cmd/docker/docker.go
```diff
@@ -28,12 +28,17 @@ import (
 )
 
 func main() {
-	ctx := context.Background()
+	baseCtx, cancel := context.WithCancel(context.Background())
```
vvoland, line 31: Same as above

vvoland (changes requested):

Benehiko:
Wondering why this test is failing on the `connhelper-ssh` tests only. I haven't been able to reproduce it locally. Might be a race condition of sorts...
https://github.com/docker/cli/actions/runs/9398990847/job/25885517677?pr=4993#step:5:731

cmd/docker/docker.go
```diff
@@ -242,40 +248,58 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
...
+		//
+		// Repeated invocations will result in EINVAL,
+		// or EBADF; but that is fine for our purposes.
+		_ = srv.Close()
+
+		// If we're still running after 3 interruptions
+		// (SIGINT/SIGTERM), send a SIGKILL to the plugin as a
+		// final attempt to terminate, and exit.
+		if force {
+			_, _ = fmt.Fprint(dockerCli.Err(), "got 3 SIGTERM/SIGINTs, forcefully exiting\n")
```
krissetto, line 277: very minor, but should we maybe print this termination msg outside of this func, when we do the actual check against the exitLimit?

cli/command/container/attach.go
```diff
@@ -105,7 +105,12 @@ func RunAttach(ctx context.Context, dockerCLI command.Cli, containerID string, o
 
 	if opts.Proxy && !c.Config.Tty {
 		sigc := notifyAllSignals()
-		go ForwardAllSignals(ctx, apiClient, containerID, sigc)
+		// since the ForwardAllSignals already registers its own signal handler
+		// and should notify the daemon independently of the clients ctx cancellation
+		// we should use a background context to avoid the ForwardAllSignals from
+		// returning before all signals are forwarded to the daemon
+		bgCtx := context.WithoutCancel(ctx)
```
krissetto, line 112: nit: since it's not a background context anymore, we should adjust the naming and the comment above. Also, since ForwardAllSignals seems to check for ctx.Done(), maybe we could still use a timeout with this context

krissetto (approved):
overall LGTM, just a couple minor comments

cmd/docker/docker.go
Benehiko: Yeah maybe the text doesn't quite make sense in the context of this function, but we essentially check the limit here https://github.com/docker/cli/blob/fdaad209907f8b5187697d07c452850869bc05ba/cmd/docker/docker.go#L295-L301

cli/command/container/attach.go
Benehiko: It's difficult to determine when the `ForceAllSignals` should stop forwarding signals based on a timeout. I think the logic is incorrect to also check `ctx.Done()` inside the `ForceAllSignals` func since it will stop the infinite loop once the signal chan is closed which happens once this function returns which happens on ctx cancellation anyway.

Benehiko: just for more clarification, for this feature to work properly in its current implementation the ctx should not be cancelled and instead the signal channel should be closed. 

I'm keeping the ctx.Done() check in the `ForceAllSignals` function since some tests rely on this behavior even though it's not used in its actual implementation.

cli/command/container/run.go
```diff
@@ -150,7 +150,12 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 	}
 	if runOpts.sigProxy {
 		sigc := notifyAllSignals()
-		go ForwardAllSignals(ctx, apiClient, containerID, sigc)
+		// since the ForwardAllSignals already registers its own signal handler
```
laurazard, line 153: `ForwardAllSignals` doesn't set up it's own signal handler though, from looking at it's implementation? The actual signal handling setup is in `notifyAllSignals`, so maybe we can adjust this comment to be clearer.
Something like 
```
// since we're explicitly setting up signal handling here, and the daemon will
// get notified independently of the clients ctx cancellation, we use this context
// but without cancellation to avoid ForwardAllSignals from returning
// before all signals are forwarded.
```

cli/command/container/attach.go
```diff
@@ -105,7 +105,12 @@ func RunAttach(ctx context.Context, dockerCLI command.Cli, containerID string, o
 
 	if opts.Proxy && !c.Config.Tty {
 		sigc := notifyAllSignals()
-		go ForwardAllSignals(ctx, apiClient, containerID, sigc)
+		// since the ForwardAllSignals already registers its own signal handler
```
laurazard, line 108: same here

cmd/docker/docker.go
```diff
@@ -242,40 +248,56 @@ func tryPluginRun(dockerCli command.Cli, cmd *cobra.Command, subcommand string,
...
 			// If we're still running after 3 interruptions
 			// (SIGINT/SIGTERM), send a SIGKILL to the plugin as a
 			// final attempt to terminate, and exit.
-			retries++
 			if retries >= exitLimit {
-				_, _ = fmt.Fprintf(dockerCli.Err(), "got %d SIGTERM/SIGINTs, forcefully exiting\n", retries)
-				_ = plugincmd.Process.Kill()
-				os.Exit(1)
+				force = true
+				_, _ = fmt.Fprint(dockerCli.Err(), "got 3 SIGTERM/SIGINTs, forcefully exiting\n")
```
laurazard, line 298: this message being here (and not inside `tryTerminatePlugin` means that now it will get printed even if we don't actually try to terminate the plugin, because we'll go into `tryTerminatePlugin`, check that we're attached to a terminal, and `return`, iiuc.

Maybe we should move it into `tryTerminatePlugin`, after the TTY check?

Benehiko: that's true, i'll move it back then

laurazard (approved):
re-LGTM

laurazard:
@vvoland can you take another look and then I'll merge?

cmd/docker/docker.go
```diff
@@ -37,7 +37,8 @@ func main() {
 	dockerCli, err := command.NewDockerCli(command.WithBaseContext(ctx))
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
-		os.Exit(1)
+		defer os.Exit(1)
```
vvoland, line 37: Why defer if we're doing `return` immediately after?

vvoland (approved):

cmd/docker/docker.go
laurazard: Oh, I actually had the same question and then forgot to post the comment 😓

laurazard: @Benehiko

Benehiko: because we `defer cancel()`. The linter warned that the defer might not run if we only do `os.Exit()` so I moved them into a defer call and then rather return the main function.

Benehiko: the other option is to move everything with the signal handler into a function which returns an error, which then does `os.Exit(statusCode)` inside `main()`.

```go
func main() {
 err := doSomething()
 if err != nil {
   os.Exit(1)
 }
}

func doSomething() error {

  ctx, cancel := ...
  defer cancel() 
  
  ...
}

Comment (vvoland) [reply cmd/docker/docker.go:37]: IMO, having an additional "C style" main would be nice:

```go
func main() {
    ret := dockerMain()
    if ret != 0 {
        os.Exit(ret)
    }
}

func dockerMain() int {
     ... all the stuff from the current main
     ... but return <exit code> instead of calling os.Exit
}

```

vvoland: Or just:

```go
func main() {
  var exitCode int
  defer func() {
    if exitCode != 0 {
      os.Exit(exitCode)
    }
  }()
  ...
}

Comment (vvoland) [reply cmd/docker/docker.go:37]: `defer os.Exit()` will only silence the linter, but since defers are executed (bottom to top), it would still do the exit before all the previous defers.

Comment (Benehiko) [reply cmd/docker/docker.go:37]: hmm true, alright i'll refactor it

File: cmd/docker/docker.go [hunk 12]
```diff
@@ -28,14 +28,17 @@ import (
 )
 
 func main() {
+	os.Exit(dockerMain())
```
vvoland, line 31: Not sure, but maybe we could avoid `os.Exit` if the return code was zero? So that Go has the chance to do some things after main (if it actually does anything meaningful? not sure).

Benehiko: hmm we sometimes return exit code 0 on errors, so not sure how we should approach that :thinking:

vvoland: That shouldn't have any impact though. We only care about `os.Exit` because we want to set a process exit code. We can just avoid `os.Exit` and it will have the 0 return code by default.

vvoland: Just for clarity, I just mean:

```suggestion
    ret := dockerMain()
    if ret != 0 {
	  os.Exit(ret)
	}
```

vvoland (approved):
LGTM

PR #4997, opened 2024-04-09
docs: clarify that --data-path-addr doesn't restrict access

dvdksn:
**- What I did**

Added a paragraph that clarifies the scope of the `--data-path-addr` for `docker swarm init`.

Relates to moby/moby#47667

**- How I did it**

**- How to verify it**


dvdksn:
cc @corhere

thaJeztah:
Wondering if  we should borrow some wording from `--advertise-addr`. Effectively (if my interpretation is correct here), `--data-path-addr` is _advertising_  the address to use (I guess in hindsight, the flag should've been named to reflect that?)


https://github.com/docker/cli/blob/c23a40469888290e912035026bf53dffd32df360/docs/reference/commandline/swarm_init.md?plain=1#L95-L109

corhere (approved):

thaJeztah (approved):
LGTM

PR #5002, opened 2024-04-10
docs: make buildx build the canonical reference doc

dvdksn:
**- What I did**

Move common flag descriptions to the buildx build reference, and make
that page the canonical page in docs. Also rewrite some content in
image_build to make clear that this page is only for the legacy builder.

Relates to:

- https://github.com/docker/docs/pull/19774
- https://github.com/docker/buildx/pull/2397

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 
 	"github.com/distribution/reference"
+	"github.com/docker/cli-docs-tool/annotation"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/image/build"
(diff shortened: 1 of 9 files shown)


dvdksn:
bump @thaJeztah - see also related:

- https://github.com/docker/docs/pull/19774
- https://github.com/docker/buildx/pull/2397

cli/command/image/build.go
```diff
@@ -102,10 +103,6 @@ func NewBuildCommand(dockerCli command.Cli) *cobra.Command {
 			options.context = args[0]
 			return runBuild(cmd.Context(), dockerCli, options)
 		},
-		Annotations: map[string]string{
-			"category-top": "4",
-			"aliases":      "docker image build, docker build, docker buildx build, docker builder build",
```
vvoland, line 107: Why remove the aliases?
IIUC, this will be hit only with `DOCKER_BUILDKIT=0`, so I see why we'd want to remove `docker buildx build`, but not sure what's the reason behind removing the other aliases?

dvdksn: Good point... I can add back the other aliases. They still apply with `DOCKER_BUILDKIT=0`

vvoland (approved):
LGTM

dvdksn:
I wanted to cherrypick to 27.0 but seems like a bit of a mess, so leaving it for 28

PR #5005, opened 2024-04-10
cli-bin/windows: Add .exe extension

vvoland:
Before this commit, the CLI binary in `dockereng/cli-bin` image was named `docker` regardless of platform.

Change the binary name to `docker.exe` in Windows images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -123,8 +123,14 @@ COPY --link . .
 FROM scratch AS plugins
 COPY --from=build-plugins /out .
 
-FROM scratch AS bin-image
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM

this one doesn't impact the `docker-ce-packging` scripts? (in case there's code there consuming these)

vvoland:
Yes, it doesn't affect anything else. The `bin-image` target is only used for the `cli-bin` image generation in this repo. Last things broke because we modified the `binary` target directly.

PR #5010, opened 2024-04-12
cli reference overview base cmd

dvdksn:
**- What I did**

Moves the `cli.md` file to become the `docker.md` base command reference.

This file contains a lot of information about how the CLI and its flags work, base flags, configuration file, and environment variables. I think it makes more sense to have it on the base command page than in a separate place.

Needs:
- https://github.com/docker/cli-docs-tool/pull/53

**- How I did it**

**- How to verify it**


docs/reference/commandline/cli.md
```diff
@@ -1,408 +0,0 @@
----
-title: "Use the Docker command line"
-description: "Docker's CLI command description and usage"
-keywords: "Docker, Docker documentation, CLI, command line, config.json, CLI configuration file"
-aliases:
-  - /reference/commandline/cli/
-  - /engine/reference/commandline/engine/
-  - /engine/reference/commandline/engine_activate/
-  - /engine/reference/commandline/engine_check/
-  - /engine/reference/commandline/engine_update/
```
thaJeztah, line 10: There's a bunch of redirects here; looks like we may need to move those to a different page now

dvdksn: Yep, they are in content/reference/cli/docker/_index.md @ docker/docs#19794

docs/reference/commandline/docker.md
```diff
@@ -1,14 +1,7 @@
...
-title: "Use the Docker command line"
-description: "Docker's CLI command description and usage"
-keywords: "Docker, Docker documentation, CLI, command line, config.json, CLI configuration file"
-aliases:
-  - /reference/commandline/cli/
-  - /engine/reference/commandline/engine/
-  - /engine/reference/commandline/engine_activate/
-  - /engine/reference/commandline/engine_check/
-  - /engine/reference/commandline/engine_update/
----
```
thaJeztah: Did we take care of these elsewhere?

docs/reference/commandline/docker.md
```diff
@@ -46,7 +63,60 @@ variables used by the Go runtime. In particular, you may find these useful:
...
+
+#### Multi
+
+You can specify options like `-a=[]` multiple times in a single command line,
+for example in these commands:
+
+```console
+$ docker run -a stdin -a stdout -i -t ubuntu /bin/bash
+
+$ docker run -a stdin -a stdout -a stderr ubuntu /bin/ls
```
thaJeztah, line 178: Not for this PR, but we need to look at some of these; `-a` / `--attach` is shown as `list`, and those can take multiple values, but for other flags, Cobra is good at silently ignoring duplicates, and in that case _overriding_ previous values. (e.g. `-i -i=false -i=true -i`)

Also for a follow-up, we should pick a different example than `-a`, as it's really more of an "expert" use-case to set that. Perhaps something like `-e` / `--env` or something common could be a good one to pick as example (the `--env` one can be nice as we could _show_ the result;



```bash
$ docker run --rm -e ENV_ONE=one -e ENV_TWO=two alpine printenv
ENV_ONE=one
ENV_TWO=two
<...>
```.

thaJeztah (approved):
LGTM, thanks!

docs/reference/commandline/docker.md
dvdksn: yes: https://github.com/docker/docs/pull/19794/files#diff-8810b23e77efed3141d1ae3306ef22281909376a94e958ba06b92f9df1fc9f1e

PR #5018, opened 2024-04-15
chore: remove deprecated DualStack field

njucjc:
**- What I did**
I went through this deprecated field of net.Dial, which is DualStack.
As described in the documentation of the net package:

// DualStack previously enabled [RFC 6555](https://rfc-editor.org/rfc/rfc6555.html) Fast Fallback
// support, also known as "Happy Eyeballs", in which IPv4 is
// tried soon if IPv6 appears to be misconfigured and
// hanging.
//
// Deprecated: Fast Fallback is enabled by default. To
// disable, set FallbackDelay to a negative value.
DualStack bool

I think we don't need it anymore.

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli/registry/client/endpoint.go
+++ b/cli/registry/client/endpoint.go
@@ -83,7 +83,6 @@ func getHTTPTransport(authConfig registrytypes.AuthConfig, endpoint registry.API
 		Dial: (&net.Dialer{
 			Timeout:   30 * time.Second,
 			KeepAlive: 30 * time.Second,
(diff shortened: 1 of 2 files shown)


thaJeztah (approved):
LGTM

njucjc:
> LGTM

@thaJeztah Ready to to merge, any other suggestions?

thaJeztah:
nope; I think this one should be good to go 😅

PR #5024, opened 2024-04-17
plugins/hooks: run hooks when exit code != 0

laurazard:
**- What I did**

This PR changes the hooks flow so that hooks are still invoked after the command execution returned an error/non-zero exit code.

I ran into this with `docker exec -it [] sh`, since I was exiting with CTRL+D after doing things inside a container and depending on the exit code my hooks wouldn't get invoked, which isn't desired.

In the future, this might also be interesting to allow plugins to run hooks after an error so they can offer error-state recovery suggestions, although this would require additional work to give the plugin more information about the failed execution.

**- How to verify it**

Try hooks after a non-zero exit code `docker exec`

image


krissetto (approved):
LGTM

PR #5025, opened 2024-04-17
Fix OTLP env var overriding

krissetto:
**- What I did**
Fix the OTLP endpoint overriding via env var

**- How I did it**
Made sure to check for the override even when otel configuration bits are missing from the context metadata, instead of returning early

**- How to verify it**


laurazard (approved):
LGTM

PR #5026, opened 2024-04-17
otel: simple OTEL collector/Prometheus stack for testing purposes

krissetto:
**- What I did**
Added a compose stack to be used for testing out OTEL functionality, mostly based on our current documentation PR found [here](https://github.com/docker/docs/pull/19776).

**- How I did it**
- Added compose stack;
- Added basic OTEL collector and prometheus configs;
- Added basic usage doc

**- How to verify it**
Read the doc and try it out :)

### DIFF
--- a/hack/otel/compose.yaml
+++ b/hack/otel/compose.yaml
@@ -1,0 +1,39 @@
+name: cli-otel
+
+services:
+  
+  prometheus:
+    image: prom/prometheus:latest
+    command:
+      - "--config.file=/etc/prometheus/prom.yaml"
+    ports:
+      # Publish the Prometheus frontend on localhost:9091
+      - 9091:9090
+    restart: always
+    volumes:
+      # Store Prometheus data in a volume:
+      - prom_data:/prometheus
+      # Mount the prom.yml config file
+      - ./prom.yaml:/etc/prometheus/prom.yaml
+
+  aspire-dashboard:
+    image: mcr.microsoft.com/dotnet/nightly/aspire-dashboard:8.0-preview
+    ports:
+      - 18888:18888
+    environment:
+      DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS: 'true'
+  
+  otelcol:
+    image: otel/opentelemetry-collector:latest
+    restart: always
+    depends_on:
+      - prometheus
+      - aspire-dashboard
+    ports:
+      - 4317:4317
+    volumes:
+      # Mount the otelcol.yml config file
+      - ./otelcol.yaml:/etc/otelcol/config.yaml
+
+volumes:
+  prom_data:
--- a/hack/otel/README.md
+++ b/hack/otel/README.md
@@ -1,0 +1,25 @@
+# Sample stack for testing OTEL functionality with the CLI
+
+To test the OTEL functionality present in the CLI, you can spin up a small demo compose stack that includes:
+- an OTEL collector container;
+- a Prometheus container;
+- an Aspire Dashboard container
+
+The `hack/otel` directory contains the compose file with the services configured, along with 2 basic configuration files: one for the OTEL collector and one for Prometheus.
+
+## How can I use it?
+
+1) Start the compose stack by running `docker compose up -d` in the `hack/otel/` directory;
+2) Export the env var used to override the OTLP endpoint:  
+  `export DOCKER_CLI_OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317` (if running the CLI in a devcontainer or in other ways, you might have to change how you pass this env var);
+3) Run the CLI to send some metrics to the endpoint;
+4) Browse Prometheus at `http://localhost:9091/graph` or the Aspire Dashboard at  `http://localhost:18888/metrics`;
+5) In Prometheus, query `command_time_milliseconds_total` to see some metrics. In Aspire, select the resource in the dropdown.
+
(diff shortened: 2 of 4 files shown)


Benehiko (commented):
I'm not sure about this change since it introduces more complexity with external shell scripts. I'd prefer to keep everything inside Go. 

Could we not use testcontainers for this instead? :)

thaJeztah:
Slightly wondering if we need the scripts at all, as they (effectively) are just a `docker compose up`?

(I know we have a bunch of a`hack/` script already though, but mostly looking if for this one we need more)

laurazard:
IMO a big part of why I want this change is to make local development easier, so Testcontainers don't really achieve that. I'm okay with shell scripts are they're portable and easy to work with and don't require bringing in additional dependencies, but if atm they are really just doing a `compose up`, `compose down`, etc., then I'm okay with holding off on them and just adding the Compose file, and thinking about the rest later if we feel the need for them.

thaJeztah:
Yes, agreed on the "local development" part. It's usually also my reservation on things like `mage` as replacement for `make`; it's cool as it's in Go (and can be flexible), but also means that `go` is now a requirement on your local machine (and where possible, we should try to still allow for a dockerized workflow, so that there's no real requirements other than "have docker installed")

laurazard (approved):
LGTM, with a suggestion to remove the shell scripts.

krissetto:
I'm not super fond of the scripts either, I was trying to follow a bit of what already existed while keeping things as simple s possible for the community (eg. i think buildx has a script which completely wraps compose, for example, and i thought that was a bit too much).  
If we agree on removing them completely, I'll adjust the `usage.md` to reflect that.

Benehiko (approved):
LGTM

krissetto:
ps i added aspire dashboard here as well so we can test both and see what we want to stick with

hack/otel/usage.md
```diff
@@ -0,0 +1,25 @@
+# Sample stack for testing OTEL functionality with the CLI
+
+To easily test the OTEL functionality present in the CLI, you can spin up a small demo compose stack that includes:
```
thaJeztah, line 3: It's good to avoid wording like "simple", "easy"; what may look simple to you / us, may not be simple to someone reading.

```suggestion
To test the OTEL functionality present in the CLI, you can spin up a demo compose stack that includes:
```

hack/otel/usage.md
```diff
@@ -0,0 +1,25 @@
...
+  `export DOCKER_CLI_OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317` (if running the CLI in a devcontainer or in other ways, you might have to change how you pass this env var);
+3) Run the CLI to send some metrics to the endpoint;
+4) Browse Prometheus at `http://localhost:9091/graph` or the Aspire Dashboard at  `http://localhost:18888/metrics`;
+5) In Prometheus, query `command_time_milliseconds_total` to see some metrics. In Aspire, select the resource in the dropdown.
+
+> **Note**: The precise steps may vary based on how you're working on the codebase (buiding a binary and executing natively, running/debugging in a devcontainer, running the normal CLI as usual, etc... )
+
+## Cleanup?
+
+Simply run `docker compose down` in the `hack/otel/` directory.
```
thaJeztah, line 23: same here;

```suggestion
Run `docker compose down` in the `hack/otel/` directory.
```

hack/otel/usage.md
krissetto: Fair enough, will change

cpuguy83 (approved):

hack/otel/compose.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - "--config.file=/etc/prometheus/prom.yaml"
+    ports:
+      # Publish the Prometheus frontend on localhost:9091
+      - 9091:9090
+    restart: always
+    volumes:
+      # Store Prometheus data in a volume:
+      - prom_data:/prometheus
+      # Mount the prom.yml config file
+      - ./prom.yaml:/etc/prometheus/prom.yaml
```
thaJeztah, line 17: Not a blocker for this PR, but I know that bind-mounting files can sometimes be problematic, so looking if we could have a directory to mount (which contains the config-file). Looks like `/etc/prometheus` already has some files in the image though, so if we do, we probably should use a different location for the config-file (but good news is that we already define it as a custom location above, so ... could still be an option);

```bash
docker run --rm --entrypoint /bin/sh prom/prometheus:latest -c 'ls -l /etc/prometheus'
total 4
lrwxrwxrwx    1 nobody   nobody          39 Apr 10 14:30 console_libraries -> /usr/share/prometheus/console_libraries
lrwxrwxrwx    1 nobody   nobody          31 Apr 10 14:30 consoles -> /usr/share/prometheus/consoles/
-rw-r--r--    1 nobody   nobody         934 Apr 10 14:21 prometheus.yml
```

The alternative could be to define it as a `config:` in the compose-file (but those are still implemented as bind-mounts in compose, so won't solve the "don't bind individual files" part)

hack/otel/compose.yaml
```diff
@@ -0,0 +1,39 @@
...
+    image: otel/opentelemetry-collector:latest
+    restart: always
+    depends_on:
+      - prometheus
+      - aspire-dashboard
+    ports:
+      - 4317:4317
+    volumes:
+      # Mount the otelcol.yml config file
+      - ./otelcol.yaml:/etc/otelcol/config.yaml
```
thaJeztah, line 36: Same for this one (also not a blocker)

thaJeztah (approved):

PR #5027, opened 2024-04-17
hooks: set expected environment when executing

laurazard:
**- What I did**

During normal plugin execution (from the CLI), the CLI configures the plugin command it's about to execute in order to pass all environment variables on, as well as to set the `ReexecEnvVar` env var that informs the plugin about how it was executed, and which plugins rely on to check whether they are being run standalone or not.

https://github.com/docker/cli/blob/50117590567998307eba377295a1ec24e214ccc7/cli-plugins/manager/manager.go#L243-L244

This commit adds the same behavior to hook invocations, which is necessary for some plugins to know that they are not running standalone so that they expose their root command at the correct level.

image

### DIFF
--- a/cli-plugins/manager/plugin.go
+++ b/cli-plugins/manager/plugin.go
@@ -2,6 +2,7 @@ package manager
 
 import (
 	"encoding/json"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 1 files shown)


krissetto (approved):
pic is on 🔥, i approve.

Benehiko (approved):

vvoland (approved):

PR #5030, opened 2024-04-19
hooks: include plugin name in hook data

laurazard:
**- What I did**

Before, for plugin commands, only the plugin name (such as `buildx`) would be included in `RootCmd` when passed to the hook plugin, which isn't enough information for a plugin to decide whether to execute a hook or not since plugins implement multiple varied commands (`buildx build`, `buildx prune`, etc.).

**- How I did it**

This commit changes the hook logic to account for this situation, so that both the plugin name and base command will get passed (separated by a SPACE, such as `buildx prune`).

This logic works for aliased commands too, so whether `docker build ...` or `docker buildx build` is executed (unless Buildx is disabled) the hook will be invoked with `buildx build`.

image

### DIFF
--- a/cli-plugins/manager/hooks.go
+++ b/cli-plugins/manager/hooks.go
@@ -14,31 +14,41 @@ import (
 // that plugins declaring support for hooks get passed when
 // being invoked following a CLI command execution.
 type HookPluginData struct {
+	// RootCmd is a string representing the matching hook configuration
+	// which is currently being invoked. If a hook for `docker context` is
+	// configured and the user executes `docker context ls`, the plugin will
+	// be invoked with `context`.
 	RootCmd string
 	Flags   map[string]string
 }
 
-// RunPluginHooks calls the hook subcommand for all present
-// CLI plugins that declare support for hooks in their metadata
-// and parses/prints their responses.
-func RunPluginHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command, plugin string, args []string) error {
-	subCmdName := subCommand.Name()
-	if plugin != "" {
-		subCmdName = plugin
-	}
-	var flags map[string]string
-	if plugin == "" {
-		flags = getCommandFlags(subCommand)
-	} else {
-		flags = getNaiveFlags(args)
-	}
-	nextSteps := invokeAndCollectHooks(dockerCli, rootCmd, subCommand, subCmdName, flags)
+// RunCLICommandHooks is the entrypoint into the hooks execution flow after
+// a main CLI command was executed. It calls the hook subcommand for all
+// present CLI plugins that declare support for hooks in their metadata and
+// parses/prints their responses.
+func RunCLICommandHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command) {
+	commandName := strings.TrimPrefix(subCommand.CommandPath(), rootCmd.Name()+" ")
+	flags := getCommandFlags(subCommand)
 
-	hooks.PrintNextSteps(dockerCli.Err(), nextSteps)
-	return nil
+	runHooks(dockerCli, rootCmd, subCommand, commandName, flags)
 }
 
-func invokeAndCollectHooks(dockerCli command.Cli, rootCmd, subCmd *cobra.Command, hookCmdName string, flags map[string]string) []string {
+// RunPluginHooks is the entrypoint for the hooks execution flow
+// after a plugin command was just executed by the CLI.
+func RunPluginHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command, args []string) {
+	commandName := strings.Join(args, " ")
+	flags := getNaiveFlags(args)
+
+	runHooks(dockerCli, rootCmd, subCommand, commandName, flags)
+}
+
+func runHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command, invokedCommand string, flags map[string]string) {
+	nextSteps := invokeAndCollectHooks(dockerCli, rootCmd, subCommand, invokedCommand, flags)
+
+	hooks.PrintNextSteps(dockerCli.Err(), nextSteps)
+}
+
+func invokeAndCollectHooks(dockerCli command.Cli, rootCmd, subCmd *cobra.Command, subCmdStr string, flags map[string]string) []string {
 	pluginsCfg := dockerCli.ConfigFile().Plugins
 	if pluginsCfg == nil {
 		return nil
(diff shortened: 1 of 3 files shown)


cli-plugins/manager/hooks.go
```diff
@@ -25,6 +29,9 @@ func RunPluginHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command, p
 	subCmdName := subCommand.Name()
 	if plugin != "" {
 		subCmdName = plugin
+		if len(args) > 1 {
+			subCmdName += " " + args[1]
```
vvoland, line 33: Should we just include all args here, or would that also pass other command arguments?

Mostly thinking about commands nesting, like: ` docker buildx imagetools inspect` - with the current code, it wouldn't pass `inspect` right?

laurazard: Eep, you're definitely right. I guess there's no trivial solution here, and my concern is that we can't tell the args from the command without having full access to the command tree from the plugin, so we'll have to pass all of it.

From a trust perspective, in general I think that's okay since the plugin will only be invoked if configured to do so, but we also use this string `subCmdName` to match against what is in the config, which won't work if we include args.

A possible solution to that would be to instead of comparing directly, we can match the configuration as a prefix, which I'm happy with, but I want to benchmark the performance impact there.

WDYT?

vvoland: I'm wondering - could we just pass the common part between the invoked args and what plugin defined as hooks?

Then:
- if the plugin defined a hook on `build`, and user calls `docker build whatever`, we'd only pass `build`
- if the plugin defined a hook  `imagetools inspect`, and user calls `docker buildx imagetools inspect ubuntu`, we only pass `imagetools inspect`.

laurazard: Oh, that sounds like a good solution! I'll give it a try.

krissetto: > if the plugin defined a hook `imagetools`, and user calls `docker buildx imagetools inspect ubuntu`, we only pass `imagetools`.

@vvoland Seeing your example, wouldn't that potentially cause some issues if two plugins have a common subcommand name and want to invoke hooks on it?  Maybe i'm missing something

Eg. if we also had a hypothetical `docker utils imagetools inspect ubuntu`, shouldn't w have to pass `utils imagetools` instead of just `imagetools` for the two to be distinguishable?

laurazard: Unless I misunderstood, you'd still pass the plugin name, the plugin name is a part of the config that gets registered so there wouldn't be collisions there.

vvoland: Yeah I gave a bad example, but I meant what Laura understood, which is matching the prefix, and not any substring 😄

thaJeztah:
I yet have to dive into the changes in this PR, but wondering if there's cases where "global" flags could cause issues, e.g. `docker buildx --builder somebuilder build` etc. (I think we already handle _global_ flags like `docker -H ..` and `docker --debug ..` etc?)

cli-plugins/manager/hooks_test.go
```diff
@@ -36,3 +36,43 @@ func TestGetNaiveFlags(t *testing.T) {
...
+	testCases := []struct {
+		commandString string
+		pluginConfig  map[string]string
+		expectedMatch string
+		expectedOk    bool
+	}{
+		{
+			commandString: "image ls",
+			pluginConfig: map[string]string{
+				"hooks": "image",
```
krissetto, line 50: do you mind adding some test cases where "hooks" contains multiple comma separated values, so we're sure we go through that code path?

krissetto: also, the ordering of the comma separated values is important since the first commandString prefix match will determine the returned value, so i think we should test that as well. 

eg. if `"hooks": "image ls,image"` i'd expect `expectedMatch: "image ls"`, does that sound right to you?

vvoland:
It operates on the args already processed by cobra, so it also works fine with global flags.

cli-plugins/manager/hooks.go
```diff
@@ -81,18 +89,37 @@ func invokeAndCollectHooks(dockerCli command.Cli, rootCmd, subCmd *cobra.Command
...
+// In the case that the command being executed is a plugin, the command
+// string should being with the plugin name, such as 'compose' or 'buildx'.
+func pluginMatch(pluginCfg map[string]string, subCmd string) (string, bool) {
+	configuredPluginHooks, ok := pluginCfg["hooks"]
 	if !ok {
-		return false
+		return "", false
 	}
-	commands := strings.Split(hookCmdStr, ",")
+
```
vvoland, line 114: Providing an empty value here (`"hooks": ""}` will execute the hooks for **all** docker commands.

Perhaps we should also just ignore it in that case?

cli-plugins/manager/hooks.go
```diff
@@ -81,18 +89,37 @@ func invokeAndCollectHooks(dockerCli command.Cli, rootCmd, subCmd *cobra.Command
...
+	commands := strings.Split(configuredPluginHooks, ",")
 	for _, hookCmd := range commands {
-		if hookCmd == subCmdName {
-			return true
+		if strings.HasPrefix(subCmd, hookCmd) {
+			return hookCmd, true
```
vvoland, line 118: I think we might want to match on full space-separated components rather than prefix only, so that hooks like `image i` don't allow matching both `image import` and `image inspect`, but still allow `image` hook to match all sub-`image` commands.

I mean replacing the `strings.HasPrefix` with something like:

```go
func hookMatch(hookCmd, subCmd string) bool {
	hookCmdTokens := strings.Split(hookCmd, " ")
	subCmdTokens := strings.Split(subCmd, " ")

	if len(hookCmdTokens) > len(subCmdTokens) {
		return false
	}

	for i, v := range hookCmdTokens {
		if v != subCmdTokens[i] {
			return false
		}
	}

	return true
}
```

cli-plugins/manager/hooks.go
```diff
@@ -81,18 +89,37 @@ func invokeAndCollectHooks(dockerCli command.Cli, rootCmd, subCmd *cobra.Command
...
+	}
+	return ret
+}
+
+// pluginMatch takes a plugin configuration and a string representing the
+// canonical command being executed (such as 'image ls' – the root 'docker'
+// is omitted) and, if the configuration includes a hook for the invoked
+// command, returns the configured hook string,
+// In the case that the command being executed is a plugin, the command
+// string should being with the plugin name, such as 'compose' or 'buildx'.
```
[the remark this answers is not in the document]
laurazard: Ah I was looking at the one on line 107 and I was so confused 😅 . Good catch.

vvoland (approved):
LGTM

krissetto (approved):

PR #5033, opened 2024-04-22
hooks: pass command execution error to plugins

laurazard:
**- What I did**

Added a field to `HookPluginData` to hold the error message (if applicable) that resulted from the execution for which the hook is being invoked.

This is only available for CLI command executions.

**- How to verify it**


vvoland (approved):
LGTM

PR #5038, opened 2024-04-25
cli-plugins: PluginRunCommand: use cmd.Environ instead of os.Environ

thaJeztah:
- follow-up to https://github.com/docker/cli/pull/5027

Commit 50117590567998307eba377295a1ec24e214ccc7 implemented a fix that caused the current environment to be discarded, using `os.Environ()`. On Windows, `os.Environ()` may produce an incorrect value for `PWD`, for which a new function was added in go1.19;

- https://tip.golang.org/doc/go1.19#osexecpkgosexec
- https://go-review.googlesource.com/c/go/+/401340
- https://github.com/golang/go/issues/50599

Replace the use of `os.Environ()` with `cmd.Environ()` to address that.

**- What I did**

**- How I did it**

**- How to verify it**

```markdown changelog


```

### DIFF
--- a/cli-plugins/manager/manager.go
+++ b/cli-plugins/manager/manager.go
@@ -240,8 +240,7 @@ func PluginRunCommand(dockerCli command.Cli, name string, rootcmd *cobra.Command
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
 
-		cmd.Env = os.Environ()
-		cmd.Env = append(cmd.Env, ReexecEnvvar+"="+os.Args[0])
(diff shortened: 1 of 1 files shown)


thaJeztah:
cc @laurazard please double-check if this is the correct change to make 🙈  (I recall this was the recommended thing to do nowadays, but maybe plugins are special here 😅)

vvoland (approved):

thaJeztah:
@vvoland I see you added the 26.1.1 milestone; was the intent to backport this, or should it be on 27.0 ?

vvoland:
Yeah, it should be on 27.0 🙈 

Still, looks like something that would be good to backport.

PR #5039, opened 2024-04-26
feat: wire ctx into plugin hooks

Benehiko:
In the spirit of wiring up `context.Context` for graceful exit and early termination of tasks, I have done some more wiring up of `ctx` inside the plugin hooks. This allows early returns for when the `ctx` is cancelled so that hooks aren't printed as well as cancelling the execution of the plugin through `exec.CommandContext`.

**- What I did**
Allow early cancellation of plugin hooks through `context.Context` and plugin execution cancellation through `exec.CommandContext`.

**- How I did it**
Wired up `ctx` inside the `RunCLICommandHooks` and `RunPluginHooks`.

**- How to verify it**

```markdown changelog


```

### DIFF
--- a/cli-plugins/manager/hooks.go
+++ b/cli-plugins/manager/hooks.go
@@ -1,6 +1,7 @@
 package manager
 
 import (
+	"context"
 	"encoding/json"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


cli-plugins/manager/hooks.go
```diff
@@ -27,29 +28,36 @@ type HookPluginData struct {
 // a main CLI command was executed. It calls the hook subcommand for all
 // present CLI plugins that declare support for hooks in their metadata and
 // parses/prints their responses.
-func RunCLICommandHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command, cmdErrorMessage string) {
+func RunCLICommandHooks(ctx context.Context, dockerCli command.Cli, rootCmd, subCommand *cobra.Command, cmdErrorMessage string) {
 	commandName := strings.TrimPrefix(subCommand.CommandPath(), rootCmd.Name()+" ")
 	flags := getCommandFlags(subCommand)
 
-	runHooks(dockerCli, rootCmd, subCommand, commandName, flags, cmdErrorMessage)
+	runHooks(ctx, dockerCli, rootCmd, subCommand, commandName, flags, cmdErrorMessage)
```
krissetto, line 35: nothing to change now, but these signatures are getting a bit rough 😅

Benehiko: yeah I agree, but I didn't want to refactor it in this PR :sweat:

laurazard:
Looks great! (and reminds me I'd really like for us to add some e2e tests for hooks when we get a chance 🥲)

Benehiko:
> Looks great! (and reminds me I'd really like for us to add some e2e tests for hooks when we get a chance 🥲)

oh yeah, saw this while adding these. I can do that in another PR once I have some time. I have a plan to utilize a pattern I used inside DD which re-uses the go test library to simulate external executables.

krissetto (approved):

cli-plugins/manager/hooks.go
```diff
@@ -27,29 +28,36 @@ type HookPluginData struct {
 // a main CLI command was executed. It calls the hook subcommand for all
 // present CLI plugins that declare support for hooks in their metadata and
 // parses/prints their responses.
-func RunCLICommandHooks(dockerCli command.Cli, rootCmd, subCommand *cobra.Command, cmdErrorMessage string) {
+func RunCLICommandHooks(ctx context.Context, dockerCli command.Cli, rootCmd, subCommand *cobra.Command, cmdErrorMessage string) {
 	commandName := strings.TrimPrefix(subCommand.CommandPath(), rootCmd.Name()+" ")
 	flags := getCommandFlags(subCommand)
```
thaJeztah, line 33: @Benehiko @laurazard Sorry, late to the party, and not 100% sure what's best (things get somewhat confusing 😅)

Which context should be used in these cases? I see we already have the `rootCmd` and `subCommand` passed; both of which carry a context; should that context be the one to use? (`rootCmd.Context()` or `subCommand.Context()`)

(If that's the one to use, then possibly the signature wouldn't have to be changed 🤔

Benehiko: that's true, although adding the ctx parameter is way more explicit and the recommended way according to the Go [documentation](https://go.dev/blog/context). The context stored inside the `rootCommand` and `subCommand` is the same as the one passed down. While I was looking at this I actually wanted to refactor the signature into an interface or using only the `docker.Cli` interface since passing down rootCmd and subCommand seems superfluous. But this is a topic for another time I think

PR #5046, opened 2024-04-30
Add plugin execution duration metric

krissetto:
**- What I did**
Added utils for instrumenting plugin commands, and instrumented them to record their execution time

**- How I did it**
- Added some utils
- Added wrapper for exec.Cmd to be used for measuring the command's execution time
- Wired up the utils so plugin execution metrics get sent 

**- How to verify it**

```markdown changelog
Instrument plugin execution with otel meter
```

pinta


krissetto:
The first commit is from [here](https://github.com/docker/cli/pull/5067), for the sake of testing

cli-plugins/manager/cobra.go
```diff
@@ -90,11 +90,12 @@ func AddPluginCommandStubs(dockerCli command.Cli, rootCmd *cobra.Command) (err e
 					cargs = append(cargs, args...)
 					cargs = append(cargs, toComplete)
 					os.Args = cargs
-					runCommand, runErr := PluginRunCommand(dockerCli, p.Name, cmd)
+					pluginRunCommand, runErr := PluginRunCommand(dockerCli, p.Name, cmd)
 					if runErr != nil {
 						return nil, cobra.ShellCompDirectiveError
 					}
-					runErr = runCommand.Run()
+					runCommand := dockerCli.(*command.DockerCli).InstrumentPluginCommand(pluginRunCommand)
```
vvoland, line 97: Haven't reviewed this properly yet, so sorry if I'm wrong, but couldn't the `InstrumentPluginCommand` be a part of the `TelemetryClient` interface, so we don't need a typecast?

krissetto: Yep you're right, good catch

vvoland: Ah, I gave it a little bit more attention and looks like it will also expose the `InstrumentPluginCommand` (which returns a private struct) to all CLI plugins. Which we probably don't want to, as no plugin should really call this method on its own.

Keeping it internal and having a typecast would probably be better indeed 🙈 I think we should get back to the previous version with a type cast, and just add a check if the cast succeeded so we don't panic in some weird edge case.

Sorry for the noise 🙈

vvoland:
@krissetto Please rebase :)

krissetto:
will fix this soon

cli/command/telemetry_utils.go
```diff
@@ -100,6 +101,63 @@ func startCobraCommandTimer(mp metric.MeterProvider, attrs []attribute.KeyValue)
 	}
 }
 
+// basePluginCommandAttributes returns a slice of attribute.KeyValue to attach to metrics/traces
+func basePluginCommandAttributes(plugincmd *exec.Cmd, streams Streams) []attribute.KeyValue {
+	pluginPath := strings.Split(plugincmd.Path, "-")
+	pluginName := pluginPath[len(pluginPath)-1]
```
vvoland, line 107: Only `docker-[a-z][a-z0-9]*` binaries are considered a valid plugins anyway, so I think it's fine to simplify it to just a single `Cut` (to avoid allocating memory).

```suggestion
	_, pluginName, _  := strings.Cut(pluginCmd.Path, "-")
```

cli/command/telemetry_utils.go
```diff
@@ -100,6 +101,63 @@ func startCobraCommandTimer(mp metric.MeterProvider, attrs []attribute.KeyValue)
...
+// wrappedCmd is used to wrap an exec.Cmd in order to instrument the
+// command with otel by using the TimedRun() func
+type wrappedCmd struct {
+	*exec.Cmd
+
+	baseAttrs []attribute.KeyValue
+	cli       *DockerCli
+}
+
+// TimedRun measures the duration of the command execution using and otel meter
```
vvoland, line 122:
```suggestion
// TimedRun measures the duration of the command execution using an otel meter
```

cli/command/telemetry_utils.go
```diff
@@ -100,6 +101,63 @@ func startCobraCommandTimer(mp metric.MeterProvider, attrs []attribute.KeyValue)
...
+func (c *wrappedCmd) TimedRun(ctx context.Context) error {
+	stopPluginCommandTimer := c.cli.startPluginCommandTimer(c.cli.MeterProvider(), c.baseAttrs)
+	err := c.Cmd.Run()
+	stopPluginCommandTimer(err)
+	return err
+}
+
+// InstrumentPluginCommand instruments the plugin's exec.Cmd to measure it's execution time
```
vvoland, line 130:
```suggestion
// InstrumentPluginCommand instruments the plugin's exec.Cmd to measure its execution time
```

e2e/cli-plugins/run_test.go
```diff
@@ -138,7 +138,7 @@ func TestHelpGood(t *testing.T) {
 	res := icmd.RunCmd(run("-l", "info", "help", "helloworld"))
 	res.Assert(t, icmd.Expected{
 		ExitCode: 0,
-		Err:      icmd.None,
+		Err:      "Warning: Unexpected error, the help command will not have OTEL metrics",
```
vvoland, line 141: This should not produce any error as it's considered a "good" way to call help command.

e2e/cli-plugins/run_test.go
```diff
@@ -191,7 +191,7 @@ func TestHelpGoodSubcommand(t *testing.T) {
 	res := icmd.RunCmd(run("-l", "info", "help", "helloworld", "goodbye"))
 	res.Assert(t, icmd.Expected{
 		ExitCode: 0,
-		Err:      icmd.None,
+		Err:      "Warning: Unexpected error, the help command will not have OTEL metrics",
```
vvoland, line 194: Same as above

e2e/cli-plugins/run_test.go
krissetto: Ok fair, so do we want to just avoid instrumenting the help command with otel metrics?

krissetto: To be clear, this warning would only appear if the casting of dockerCli fails

vvoland: Just removing the warning is fine IMO. Otherwise user might run into it when simply making a typo.

vvoland: When exactly does the cast fail? In this test executing `docker run -l info help helloworld` produces the warning.

cli/command/telemetry_utils.go
krissetto: We need to split after the last `-`, `strings.Cut` splits after the first occurrence and we have `cli-plugins` in the plugins' path

vvoland: Oh right, I missed that's a full path 🙈

e2e/cli-plugins/run_test.go
vvoland: Yeah, it's not `run` (the function name mislead me 🙈). 

But still, I don't think we want to show that warning in case of `docker -l info help helloworld`.

krissetto: should be fixed now

vvoland: Thanks! Just for clarification - this will actually never happen regardless of what user does right? The code would have to change for the type cast to fail?

krissetto: Yep, no user action _should_ be able to break any of these `.(*command.DockerCli)` casts

cli-plugins/manager/cobra.go
```diff
@@ -90,11 +90,18 @@ func AddPluginCommandStubs(dockerCli command.Cli, rootCmd *cobra.Command) (err e
 					cargs = append(cargs, args...)
 					cargs = append(cargs, toComplete)
 					os.Args = cargs
-					runCommand, runErr := PluginRunCommand(dockerCli, p.Name, cmd)
+					pluginRunCommand, runErr := PluginRunCommand(dockerCli, p.Name, cmd)
 					if runErr != nil {
 						return nil, cobra.ShellCompDirectiveError
 					}
-					runErr = runCommand.Run()
+					if verifiedDockerCli, ok := dockerCli.(*command.DockerCli); ok {
```
laurazard, line 97: Can we just shadow `dockerCli` in this case instead of calling it `verifiedDockerCli`? Not a huge fan, sorry (I was wondering what made it "verified", and then realized it was just the typecast cli)

cli/command/telemetry_utils.go
```diff
@@ -100,6 +101,66 @@ func startCobraCommandTimer(mp metric.MeterProvider, attrs []attribute.KeyValue)
...
+	pluginPath := strings.Split(plugincmd.Path, "-")
+	pluginName := pluginPath[len(pluginPath)-1]
+	return append([]attribute.KeyValue{
+		attribute.String("plugin.name", pluginName),
+	}, stdioAttributes(streams)...)
+}
+
+// wrappedCmd is used to wrap an exec.Cmd in order to instrument the
+// command with otel by using the TimedRun() func
+type wrappedCmd struct {
```
laurazard, line 115: Instead of creating a wrapped `exec.Cmd` type, how hard would it be to just have `pluginmanager.PluginRunCommand` return a `func` that starts a timer, runs the command, and ends the timer?

Okay if we end up going with a wrapper, I just try hard to avoid those/introducing more state since now it has a pointer to a cli, etc.

cli/command/telemetry_utils.go
```diff
@@ -100,6 +101,66 @@ func startCobraCommandTimer(mp metric.MeterProvider, attrs []attribute.KeyValue)
...
+	*exec.Cmd
+
+	baseAttrs []attribute.KeyValue
+	cli       *DockerCli
+}
+
+// TimedRun measures the duration of the command execution using an otel meter
+func (c *wrappedCmd) TimedRun(ctx context.Context) error {
```
laurazard, line 123: this isn't using `ctx`, is it?

cli-plugins/manager/cobra.go
krissetto: thing is if we shadow `dockerCli` here, then in the else condition we can't call `dockerCli.Err()` as it'd be a zero value and not the original value of `dockerCli`. I tried avoiding this where possible (see the other places where i do this cast), but here since we need the `else` i wasn't sure how to proceed if not renaming the var. 

Maybe there are better ways?

cli/command/telemetry_utils.go
krissetto: not anymore, good catch. will fix

cli/command/telemetry_utils.go
krissetto: i'm not sure that'd work..in the sense that we do things with the `exec.Cmd` that `pluginmanager.PluginRunCommand` returns aside from just running it (see the `tryPluginRun()` func as an example). So i wanted to keep the return value of `InstrumentPluginCommand` something close to what `pluginmanager.PluginRunCommand` currently returns. I guess we could rework all that but i'm not sure it'd be worth the extra work. 

I'm not a fan of the wrapped type here either, but it seemed like a way to keep code changes relatively minimal on the usage side of things, and to allow code that doesn't need the otel bits to continue to function as usual. Happy to change the impl if we can find some simpler way to handle this. The pointer to DockerCli is really only needed to get the `meter provider` and the `startPluginCommandTimer` func atm.

The main idea of this impl was to try and keep the approach as close as possible to the one used by `InstrumentCobraCommand`

laurazard: Gotcha. Let me take a closer look, but if this is indeed the least invasive way that's okay by me. I also don't love having to remember to call a new method `TimedRun`, which is why I'd have rather returned a func or something.

laurazard (approved):
Overall LGTM ✅ 

I had some nits/small comments, but I don't think they're blockers and might even be contentious, so I just pushed a branch [here](https://github.com/docker/cli/commit/62a94e294cd170d025f7aad9bf90c1616229fb85) for comments.

Feel free to take them, I just wanted to simplify the code a bit, but we can merge this PR as-is and discuss other cleanups/refactors later.

cli-plugins/manager/cobra.go
thaJeztah: Could we add a (can be un-exported) `pluginInstrumenter` interface and match that?

cli-plugins/manager/cobra.go
```diff
@@ -90,11 +90,18 @@ func AddPluginCommandStubs(dockerCli command.Cli, rootCmd *cobra.Command) (err e
...
+						runCommand := verifiedDockerCli.InstrumentPluginCommand(pluginRunCommand)
+						runErr = runCommand.TimedRun()
```
thaJeztah, line 99: `InstrumentPluginCommand` returns a non-exported type, and we always execute `TimedRun()` afterwards, except for the `tryPluginRun` function, where we construct it, and then later execute `TimedRun`.

It's unfortunate that `exec.Cmd` doesn't implement an interface, otherwise we could've made that the return, and we can't just define an interface that _only_ has a `func Run() error`

Perhaps we should either make this function perform the `TimedRun`; puny-code;

```go
if cli is a pluginInstrumenter {
    cli.RunInstrumented(pluginRunCommand)
} else {
    pluginRunCommand.Run()
}
```

But that could even be a function, because effectively it's not a method of the _CLI_;


```go
func RunPluginCommandInstrumented(cli, pluginCommand)
```

Hmm.. but then we get to `startPluginCommandTimer`, which is even more generic, but missing the `baseAttrs` 🤔

thaJeztah (changes requested):
I think I want to have a closer look at this PR to see if we can find some alternatives for some parts 😅

cli-plugins/manager/cobra.go
krissetto: > It's unfortunate that exec.Cmd doesn't implement an interface, otherwise we could've made that the return, and we can't just define an interface that only has a func Run() error

My thoughts exactly, it's kinda annoying. 

We can surely find a better way to do this, my original goal with the first implementation was to keep the approach as close as possible to that used to instrument the cobra commands, but maybe that's not a great idea all things considered. The `dockerCli` struct / `Cli` interface situation in the CLI (since we don't always pass around just the interface or just the struct) also caused some minor headaches and ugliness that would be nice to avoid if possible.

PR #5051, opened 2024-05-02
Add OTel instrumentation to CLI plugins

laurazard:
**- What I did**

Instrument CLI plugin commands to track `command.time` meter, same as the regular CLI commands.

```terminal
$ DOCKER_CLI_OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4321 ./build/docker compose down
```

image

**- How I did it**

**- How to verify it**

```markdown changelog


```

image

### DIFF
--- a/cli-plugins/plugin/plugin.go
+++ b/cli-plugins/plugin/plugin.go
@@ -46,6 +46,24 @@ func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager
 				opts = append(opts, withPluginClientConn(plugin.Name()))
 			}
 			err = tcmd.Initialize(opts...)
+			ogRunE := cmd.RunE
+			if ogRunE == nil {
+				ogRun := cmd.Run
+				// necessary because error will always be nil here
+				// see: https://github.com/golangci/golangci-lint/issues/1379
+				//nolint:unparam
+				ogRunE = func(cmd *cobra.Command, args []string) error {
+					ogRun(cmd, args)
+					return nil
+				}
+				cmd.Run = nil
+			}
+			cmd.RunE = func(cmd *cobra.Command, args []string) error {
+				stopInstrumentation := dockerCli.StartInstrumentation(cmd)
+				err := ogRunE(cmd, args)
+				stopInstrumentation(err)
+				return err
+			}
 		})
 		return err
 	}
--- a/cli/command/telemetry_utils.go
+++ b/cli/command/telemetry_utils.go
@@ -26,8 +26,7 @@ func BaseCommandAttributes(cmd *cobra.Command, streams Streams) []attribute.KeyV
 // Note: this should be the last func to wrap/modify the PersistentRunE/RunE funcs before command execution.
 //
 // can also be used for spans!
-func (cli *DockerCli) InstrumentCobraCommands(cmd *cobra.Command, mp metric.MeterProvider) {
-	meter := getDefaultMeter(mp)
+func (cli *DockerCli) InstrumentCobraCommands(ctx context.Context, cmd *cobra.Command) {
 	// If PersistentPreRunE is nil, make it execute PersistentPreRun and return nil by default
 	ogPersistentPreRunE := cmd.PersistentPreRunE
 	if ogPersistentPreRunE == nil {
@@ -55,10 +54,9 @@ func (cli *DockerCli) InstrumentCobraCommands(cmd *cobra.Command, mp metric.Mete
 		}
 		cmd.RunE = func(cmd *cobra.Command, args []string) error {
 			// start the timer as the first step of every cobra command
-			baseAttrs := BaseCommandAttributes(cmd, cli)
-			stopCobraCmdTimer := startCobraCommandTimer(cmd, meter, baseAttrs)
+			stopInstrumentation := cli.StartInstrumentation(cmd)
 			cmdErr := ogRunE(cmd, args)
-			stopCobraCmdTimer(cmdErr)
+			stopInstrumentation(cmdErr)
 			return cmdErr
 		}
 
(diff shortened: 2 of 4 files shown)


cli-plugins/plugin/plugin.go
```diff
@@ -46,6 +46,22 @@ func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager
 				opts = append(opts, withPluginClientConn(plugin.Name()))
 			}
 			err = tcmd.Initialize(opts...)
+			ogRunE := cmd.RunE
+			if ogRunE == nil {
+				ogRun := cmd.Run
+				//nolint:unparam // necessary because error will always be nil here
+				ogRunE = func(cmd *cobra.Command, args []string) error {
```
vvoland, line 53: Ugh, it's weird that we need that nolint. I did a little digging and it's a bug in combination of golangci-lint and unparam: https://github.com/golangci/golangci-lint/issues/1379 which is tricky to fix 😞

Unfortunately, it seems unlikely to be fixed.. But probably wouldn't hurt to link to that issue in the comment.

cli-plugins/plugin/plugin.go
```diff
@@ -46,6 +46,22 @@ func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager
...
+					return nil
+				}
+				cmd.Run = nil
+			}
+			cmd.RunE = func(cmd *cobra.Command, args []string) error {
+				stopInstrumentation := dockerCli.StartInstrumentation(ctx, cmd)
+				err := ogRunE(cmd, args)
+				stopInstrumentation(err)
+				return err
+			}
```
vvoland, line 64: Hmm, do we actually need to wrap the `cmd.RunE` here?

Wouldn't it be enough to call the Start/stopInstrumentation around the `cmd.Execute` call?
https://github.com/docker/cli/blob/8f0b236c2848bfbda9e65240a1b97c2f009a2caf/cli-plugins/plugin/plugin.go#L76

laurazard: Noo, needs to be there, due to when it gets called. The plugins do necessary setup in their `PersistentPreRunE`, this has to happen after that.

cli-plugins/plugin/plugin.go
laurazard: Will do!

cli/command/telemetry_utils.go
```diff
@@ -55,17 +54,27 @@ func (cli *DockerCli) InstrumentCobraCommands(cmd *cobra.Command, mp metric.Mete
...
+			stopInstrumentation(cmdErr)
 			return cmdErr
 		}
 
 		return ogPersistentPreRunE(cmd, args)
 	}
 }
 
+// StartInstrumentation instruments CLI commands with the individual metrics and spans configured.
+// It's the main command OTel utlility, and new command-related metrics should be added to it.
```
vvoland, line 68: nit: Typo
```suggestion
// It's the main command OTel utility, and new command-related metrics should be added to it.
```

cli/command/telemetry_utils.go
```diff
@@ -55,17 +54,27 @@ func (cli *DockerCli) InstrumentCobraCommands(cmd *cobra.Command, mp metric.Mete
...
+// It should be called immediately before command execution, and returns a stopInstrumentation function
+// that must be called with the error resulting from the command execution.
+func (cli *DockerCli) StartInstrumentation(ctx context.Context, cmd *cobra.Command) (stopInstrumentation func(error)) {
+	mp := cli.MeterProvider(ctx)
```
krissetto, line 72: I'm a bit unsure if we should be creating a new MeterProvider here. In the case of the CLI itself, we could just get the Meter from the global MeterProvider (which would be already set by the time we invoke this function). In the case of plugins though, maybe we could do a check to see if a global MeterProvider has already been set and only create a new one if not set. WDYT?

laurazard: Sounds good to me, yeah!

laurazard: @jsternberg thoughts?

krissetto: Good observations @jsternberg 

> it might be good if we figure out who is responsible for initializing the providers

I'd say plugins should be responsible for initializing their own `Meter/TracerProviders`.  

IIUC this could be done in the `makeCmd` func passed into `plugin.Run()`, as long as it's done AFTER having run the `plugin.PersistentRunE()` func, so that the `dockerCli` struct is initialized.  

In the case of `buildx`, i guess it'd go here after calling `plugin.PersistentPreRunE()`: https://github.com/docker/buildx/blob/d4f088e689b41353d74f1a0bfcd6d7c0b213aed2/commands/root.go#L39

> buildx would need to be able to attach an additional span processor to the default provider because it has some special functionality to forward traces to buildkit for certain features to work

Can those `SpanProcessors` be attached to the `TracerProvider` as part of the `opts ...sdktrace.TracerProviderOption`, or am I missing something here? If so, buildx could just create it's `TracerProvider` using the CLI utils and set that as it's global `TracerProvider` 

I'm imagining something along the lines of this, done after `plugin.PersistentPreRunE()` as mentioned above:
```go
sp := newSpanProc()
tpOpts := []trace.TracerProviderOption{trace.WithSpanProcessor(sp)}
tp := dockerCli.TracerProvider(ctx, tpOpts)
otel.SetTracerProvider(tp)
```

> My fear is it might be too easy to use a provider retrieved from some global state in the wrong order and have no way to know whether it was done correctly than for someone to manually check.

This final part is a bit unclear to me atm, could you explain what exactly you mean by "wrong order" and how that affects things? I might be missing something here

cli-plugins/plugin/plugin.go
```diff
@@ -46,6 +46,24 @@ func RunPlugin(dockerCli *command.DockerCli, plugin *cobra.Command, meta manager
...
+				ogRunE = func(cmd *cobra.Command, args []string) error {
+					ogRun(cmd, args)
+					return nil
+				}
+				cmd.Run = nil
+			}
+			cmd.RunE = func(cmd *cobra.Command, args []string) error {
+				stopInstrumentation := dockerCli.StartInstrumentation(ctx, cmd)
+				err := ogRunE(cmd, args)
+				stopInstrumentation(err)
```
krissetto, line 64: I just realized we'd likely also need to either wrap `cmd.PersistentPostRunE` to call `.Shutdown()` on the global `MeterProvider` (and `TracerProvider` once we implement it), or do that shutdown after this line

cli/command/telemetry_utils.go
jsternberg: > This final part is a bit unclear to me atm, could you explain what exactly you mean by "wrong order" and how that affects things? I might be missing something here

Sorry my language was a bit unclear. If we were to use a global getter/setter within the plugins for retrieving the provider, you could easily end up in a situation like this:

```
// has this tracer provider been initialized already or is it pre-initialization?
tp := otel.GetTracerProvider()
```

I just want to avoid that question when accessing the tracer provider. If the cli framework is responsible for initializing the provider, then it would be easier to verify correctness if plugins ask the cli framework directly rather than relying on a side-effect.

PR #5052, opened 2024-05-02
Parsing individual json objects into json array

arankeparth:
**- What I did**
Added commas between individual json objects and then added brackets around them to make an array.

**- How I did it**
Read each line from the buffer then added comma and finally added brackets.

**- How to verify it**

Parsing individual json objects into json array

### DIFF
--- a/cli/command/formatter/formatter.go
+++ b/cli/command/formatter/formatter.go
@@ -5,6 +5,7 @@ package formatter
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


thaJeztah:
This would be a breaking change; the current output is line-delimited JSON, which most tools would be able to handle already.

neersighted:
I'm going to close this for now; not only is this an undesired breaking change (the output here has always intentionally been JSON lines), but kludging an array together using string manipulation (instead of changing the serialization code) is not a sensible way to make this change even if it was desired.

Issue #5053, opened 2024-05-03
docker-cli broken in v26.1.1 with entrypoint is invalid. Keeps hanging forever.


Cytrian:
### Description

This command sometime, not always, fails when using docker-cli 26.1.1.
It does not always exit.
```
docker run --rm --entrypoint invalidcommand alpine param
```
The command does not always return now. I have to kill  the docker-cli process with  `kill -9`.

### Reproduce

```
# docker run --rm --entrypoint invalidcommand alpine param
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "invalidcommand": executable file not found in $PATH: unknown.
```
The error message is expected but the command does not always return now. I often have to kill  the docker-cli process with  `kill -9`.

The docker container seems to exit, but docker-cli does not.

We have a similar command in our build pipeline to check for distroless images, this breaks now.

Reinstalling docker-cli in previous version 26.1.0 solves the problem.
Verified on Ubuntu and Debian.

### Expected behavior

`docker run --rm --entrypoint invalidcommand alpine param`
should always exit immediately.

### docker version

```bash
Client: Docker Engine - Community
 Version:           26.1.1
 API version:       1.45
 Go version:        go1.21.9
 Git commit:        4cf5afa
 Built:             Tue Apr 30 11:48:20 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          26.1.1
  API version:      1.45 (minimum version 1.24)
  Go version:       go1.21.9
  Git commit:       ac2de55
  Built:            Tue Apr 30 11:48:20 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.31
  GitCommit:        e377cd56a71523140ca6ae87e30244719194a521
 runc:
  Version:          1.1.12
  GitCommit:        v1.1.12-0-g51d5e94
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client: Docker Engine - Community
 Version:    26.1.1
 Context:    default
 Debug Mode: false

Server:
 Containers: 32
  Running: 21
  Paused: 0
  Stopped: 11
 Images: 40
 Server Version: 26.1.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
[...]

PR #5060, opened 2024-05-05
Fix hang when container fails to start

laurazard:
**- What I did**

fixes https://github.com/docker/cli/issues/5053

When running a container, if it fails to start, we immediately cancel the context used for `waitExitOrRemoved` and, if running the container with `--rm`, wait on the channel:

https://github.com/docker/cli/blob/870ad7f4b9910341f30dd2632db4375858455d8a/cli/command/container/run.go#L196-L205

However, in https://github.com/docker/cli/commit/840016ea0504fb1c616a3af2f8d48bd16a7400f4, we added the following line https://github.com/docker/cli/blob/647ccf34334f76b7cc668f485143af4e11403c20/cli/command/container/utils.go#L40-L41

but in this case, we return from the goroutine without notifying the caller, which causes the caller to hang.

**- How I did it**

We should not be cancelling the context used to wait on the container exit/removal in this case, and it's only happening by accident due to this line, which is there for other reasons https://github.com/docker/cli/blob/05cba3f7b97c73d49675a3e5adb1118fe0c79867/cli/command/container/run.go#L196-L199

So I used a new context for the `waitExitOrRemoved` call.

Also, close the returned channel to unblock any caller when we hit this branch, and added another `close` to a case down where we added another return without notifying the caller.

**- How to verify it**

Reproduce such as:
```bash
#!/bin/bash
while :; do
	docker run --rm --entrypoint invalidcommand alpine param
done
```

```markdown changelog
Fix issue where the CLI process would sometimes hang when a container failed to start.
```

image

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -186,7 +186,11 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 		defer closeFn()
 	}
 
-	statusChan := waitExitOrRemoved(ctx, apiClient, containerID, copts.autoRemove)
+	// New context here because we don't to cancel waiting on container exit/remove
+	// when we cancel attach, etc.
+	statusCtx, cancelStatusCtx := context.WithCancel(context.WithoutCancel(ctx))
+	defer cancelStatusCtx()
+	statusChan := waitExitOrRemoved(statusCtx, apiClient, containerID, copts.autoRemove)
 
 	// start the container
 	if err := apiClient.ContainerStart(ctx, containerID, container.StartOptions{}); err != nil {
--- a/cli/command/container/utils.go
+++ b/cli/command/container/utils.go
@@ -36,6 +36,7 @@ func waitExitOrRemoved(ctx context.Context, apiClient client.APIClient, containe
 
 	statusC := make(chan int)
 	go func() {
+		defer close(statusC)
 		select {
 		case <-ctx.Done():
 			return
--- a/e2e/container/run_test.go
+++ b/e2e/container/run_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/docker/cli/e2e/internal/fixtures"
 	"github.com/docker/cli/internal/test/environment"
@@ -32,6 +33,22 @@ func TestRunAttachedFromRemoteImageAndRemove(t *testing.T) {
 	result.Assert(t, icmd.Success)
 	assert.Check(t, is.Equal("this is output\n", result.Stdout()))
 	golden.Assert(t, result.Stderr(), "run-attached-from-remote-and-remove.golden")
+}
+
+// Regression test for https://github.com/docker/cli/issues/5053
+func TestRunInvalidEntrypointWithAutoremove(t *testing.T) {
+	environment.SkipIfDaemonNotLinux(t)
+
+	result := make(chan *icmd.Result)
+	go func() {
+		result <- icmd.RunCommand("docker", "run", "--rm", fixtures.AlpineImage, "invalidcommand")
+	}()
+	select {
+	case r := <-result:
+		r.Assert(t, icmd.Expected{ExitCode: 127})
+	case <-time.After(4 * time.Second):
+		t.Fatal("test took too long, shouldn't hang")
+	}
 }
 
 func TestRunWithContentTrust(t *testing.T) {


laurazard:
Should also add a test that would catch a regression such as this, but I'll leave that for tomorrow 😅

thaJeztah:
Will this affect client-side handling of `--rm` (API 1.25 and older)? I noticed that's handled in this area of the code, and I recall there was some rather complex logic around this.

laurazard:
I don't think it should, that check/path happens before the changes in this patch/the other changes right? https://github.com/docker/cli/blob/647ccf34334f76b7cc668f485143af4e11403c20/cli/command/container/utils.go#L26-L28

cli/command/container/utils.go
```diff
@@ -38,6 +38,7 @@ func waitExitOrRemoved(ctx context.Context, apiClient client.APIClient, containe
 	go func() {
 		select {
 		case <-ctx.Done():
+			close(statusC)
```
krissetto, line 41: Should we define a status code for these scenarios where the ctx has been canceled? Could it be useful to send that along before closing the channel?

e2e/container/run_test.go
```diff
@@ -34,6 +35,22 @@ func TestRunAttachedFromRemoteImageAndRemove(t *testing.T) {
...
+
+	result := make(chan *icmd.Result)
+	go func() {
+		result <- icmd.RunCommand("docker", "run", "--entrypoint=invalidcommand", "--rm", fixtures.AlpineImage, "param")
+	}()
+	select {
+	case r := <-result:
+		r.Assert(t, icmd.Expected{ExitCode: 127})
+	case <-time.After(4 * time.Second):
+		t.Fatal("test took too long, shouldn't hang")
```
thaJeztah, line 50: Not sure if it makes the test cleaner (or the reverse), but `icmd.RunCommand` is a wrapper for `icmd.RunCmd` - the latter takes a `Command`, which can have a timeout on its own;

- https://github.com/gotestyourself/gotest.tools/blob/1fa4ef36538cd6021425ac34811bb7b20da3652a/icmd/command_test.go#L66-L70
- Or the `WithTimeOut` helper; https://github.com/gotestyourself/gotest.tools/blob/1fa4ef36538cd6021425ac34811bb7b20da3652a/icmd/ops.go#L12-L17

Really not sure if it's cleaner though; haven't tried yet what it looks like (just recalled that there was some way to set a timeout) 🙈

cli/command/container/utils.go
laurazard: I think we should keep the previous behaviour here, which I think this patch does (check the added e2e tests). I tested locally, and the old exit code was 127, and with this patch it's still 127. In the container fails to start+autoremove code path, we don't check the result from that channel (see https://github.com/docker/cli/blob/05cba3f7b97c73d49675a3e5adb1118fe0c79867/cli/command/container/run.go#L204).

e2e/container/run_test.go
thaJeztah: I guess it would be something like;

```go
icmd.RunCmd(Command("docker", "run", "--entrypoint=invalidcommand", "--rm", fixtures.AlpineImage, "param"), WithTimeout(4 * time.Second))
```

e2e/container/run_test.go
```diff
@@ -34,6 +35,22 @@ func TestRunAttachedFromRemoteImageAndRemove(t *testing.T) {
 	golden.Assert(t, result.Stderr(), "run-attached-from-remote-and-remove.golden")
 }
 
+// Regression test for https://github.com/docker/cli/issues/5053
+func TestRunInvalidEntrypointWithAutoremove(t *testing.T) {
+	environment.SkipIfDaemonNotLinux(t)
+
+	result := make(chan *icmd.Result)
+	go func() {
+		result <- icmd.RunCommand("docker", "run", "--entrypoint=invalidcommand", "--rm", fixtures.AlpineImage, "param")
```
thaJeztah, line 44: Really nit; slightly leaning toward simplifying the test (not having the `--entrypoint` in there if it's not really needed for the test), e.g.

```suggestion
		result <- icmd.RunCommand("docker", "run", "--rm", fixtures.AlpineImage, "invalidcommand")
```

e2e/container/run_test.go
laurazard: I generally prefer the way I had it (it's a pattern I'm used to in tests, and obvious where the timeout happens) but I guess there are some benefits? I don't really mind much, but I'll change it if it's preferred.

e2e/container/run_test.go
laurazard: Sure!

cli/command/container/utils.go
laurazard: The only case that could cause issues is https://github.com/docker/cli/blob/05cba3f7b97c73d49675a3e5adb1118fe0c79867/cli/command/container/run.go#L234, maybe we should send there? I'm not sure if we'd ever hit that line with a context cancelled, but I guess it shouldn't hurt.

e2e/container/run_test.go
thaJeztah: I'm good with the current approach overall; I agree it's readable, so from that perspective 👍 

For transparency; I was curious if `icmd.RunCommand` had a variant that accepted a context; both with (future) OTEL tracing in tests in mind, and (possible) to use that for the timeout. 🥳 

But it looks like it currently doesn't 🤷‍♂️ but when I looked I (re)discovered that it has an option to set a timeout 😅 

> but I guess there are some benefits?

So .... maybe? A quick glance shows that it also handles killing runaway processes ... I guess? Not sure if it's super important, and haven't looked _really_ closely at the other bits it does;

https://github.com/docker/cli/blob/647ccf34334f76b7cc668f485143af4e11403c20/vendor/gotest.tools/v3/icmd/command.go#L270-L295

cli/command/container/utils.go
laurazard: Changed this up more, we shouldn't even be getting here since there's no reason to cancel the context if `--rm` is set.

cli/command/container/run.go
```diff
@@ -186,7 +186,11 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 		defer closeFn()
 	}
 
-	statusChan := waitExitOrRemoved(ctx, apiClient, containerID, copts.autoRemove)
+	// New context here because we don't to cancel waiting on container exit/remove
+	// when we cancel attach, etc.
+	statusCtx, cancelStatusCtx := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
thaJeztah: Good point on both; is your thinking here to preserve the original context for tracing purposes (OTEL)? Given that go1.20 is EOL, perhaps it's ok to start ignoring it (although not sure if this code has made its way into other projects that didn't upgrade yet).

It's an interesting point though; `context.Background()` used to be the (IIUC) "correct" approach for spinning up something that's decoupled from the main context (w.r.t. cancelation at least). `WithoutCancel` does provide similar features (but also may be slightly less clear on intent).

Perhaps also something to consider discussing in our OTEL sync call; have a common approach for this (which _could_ even be a utility that's in a shared OTEL package, although there may be a risk on over-engineering things there.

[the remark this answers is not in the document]
vvoland: The issue is IMO in the `context.Context` itself lumping two distinct features (arbitrary key-value store AND passing cancellation/deadline around).

In this case, the intent is to create a context that's not immediately canceled when the parent context is cancelled (so effectively ignoring the parent context cancellation) while still persisting the key-value feature (because that's how OTEL passes the traces around).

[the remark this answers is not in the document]
thaJeztah: Yup, 💯 . And to _some extent_ it would then make sense to have a utility in some OTEL-related package, but then again "key-value" could be there for other purposes (like containerd's context-logger), so ... maybe not.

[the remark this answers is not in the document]
laurazard: Good catch, thanks!

cli/command/container/utils.go
```diff
@@ -38,6 +38,7 @@ func waitExitOrRemoved(ctx context.Context, apiClient client.APIClient, containe
 	go func() {
 		select {
```
vvoland, line 39: nit: 
Maybe we could close in a defer in the goroutine scope so that it's less likely to be missed in future?

```suggestion
	go func() {
	    defer close(statusC)
		select {
```

vvoland (approved):
LGTM; left a small nit, but feel free to ignore 😅

laurazard:
Agree! I'll write a ticket somewhere for us to remember to do that, should look at all the tests and check that.

thaJeztah:
Thanks! Yes, looks like a tracking ticket won't hurt. In most cases it probably shouldn't be problematic, but it _could_ explain some flakiness (containers from other tests still being around, causing tests to be inconsistent).


In either case, not critical for this PR; I think this PR should be good to go

PR #5067, opened 2024-05-13
Centralize init of Meter/TracerProviders

krissetto:
…e them. Initialize them during DockerCli struct init

**- What I did**

**- How I did it**

**- How to verify it**

```markdown changelog


```


laurazard (approved):
LGTM

cmd/docker/docker.go
```diff
@@ -308,9 +308,8 @@ func runDocker(ctx context.Context, dockerCli *command.DockerCli) error {
 		return err
 	}
 
-	mp := dockerCli.MeterProvider(ctx)
-	defer mp.Shutdown(ctx)
-	otel.SetMeterProvider(mp)
+	mp := dockerCli.MeterProvider()
+	defer mp.(command.MeterProvider).Shutdown(ctx)
```
jsternberg, line 312: Is there a way we can move this shutdown into the general workflow of the plugins to happen automatically? I'm a bit uncomfortable with this typecast here. It seems like we either need to add a `Shutdown()` method to the `dockerCli` struct itself or find some way to have it happen automatically. We can register the shutdown methods with the `dockerCli` struct when we initialize the tracer and meter provider.

If we add a `Shutdown` method though, we need to modify every plugin to use it. It would be nice if we could wire this into cobra so it happens automatically when the CLI is exiting.

laurazard: This particular code that you're commenting on isn't relevant for the plugin workflow, this is part of the main CLI binary execution.

laurazard: Also, would a safe cast (with an error/debug log if we failed to cast) look better to you here? It doesn't worry me since this is only inside the CLI codebase and no external import would change that `mp` before we cast, but I wouldn't mind a safer approach I guess.

vvoland: +1 on a safer cast (can still print a big red error though).

If in some case it turns out that the cast fails, I'd much rather just miss an OTEL flush than get a panic before the hook even gets executed.

vvoland (changes requested):

cmd/docker/docker.go
```diff
@@ -308,9 +308,11 @@ func runDocker(ctx context.Context, dockerCli *command.DockerCli) error {
 		return err
 	}
 
-	mp := dockerCli.MeterProvider(ctx)
-	defer mp.Shutdown(ctx)
-	otel.SetMeterProvider(mp)
+	mp := dockerCli.MeterProvider()
+	if mp, ok := mp.(command.MeterProvider); ok {
+		defer mp.Shutdown(ctx)
+	}
```
laurazard, line 314:
```suggestion
	defer func() {
	    if mp, ok := mp.(command.MeterProvider); ok {
	        mp.Shutdown(ctx)
	    }
	}
```

if the goal is to address a concern about a panic because of an unsafe cast there, the check for whether `mp` can be cast needs to happen inside the `defer`

laurazard, line 314:
```suggestion
        defer func() {
	    if mp, ok := mp.(command.MeterProvider); ok {
		    defer mp.Shutdown(ctx)
	    }
	}
```

Would this be better? If the concern is that `mp` will not be castable to `command.MeterProvider` by the time we do it. Or I guess in your version, the `mp` we call `.Shutdown` on is a different reference that won't be changed, so the end result is similar.

vvoland: IMO, either one is fine as `mp` doesn't change. With the code as-is, defer will be called on the mp reference as it was at the time the type check was performed. In your version, the shutdown will be called on the "final" value of that variable.

vvoland: Also, wondering - if our only usage is casted to `command.MeterProvider`, can't the return value of `MeterProvider` already be `command.MeterProvider`? Or is that some more general OTEL interface?

krissetto: Mmh honestly I don't think we should be checking the type cast at the very end of the `runDocker func` (given this entire anonymous function is deferred). I think the current check is enough, as it maintains the current behavior if the cast is ok and simply doesn't run Shutdown() if the cast fails for some reason (which we aren't expecting).

Do you have any reason to think this exact meter provider would change during the execution of `runDocker`? Maybe I'm missing something

laurazard: I'm okay with it as-is 👍

laurazard:
Chatted w/ @jsternberg, there's some other approaches that would be interesting to look at (take care of shutdown automatically in the CLI lifecycle), but he's okay with merging as-is for now to unblock releases and improving things later.

cmd/docker/docker.go
vvoland, line 314: Since we don't expect the typecast to actually fail, I think we should have an error log (to stderr) just in case any future OTEL update breaks our assumption that these methods are defined.

vvoland (approved):
LGTM; I think we should have a log in case the typecast fails (so like panic, but without actually panicking), so we can be aware if our assumptions are broken with future OTEL updates.

Not a blocker if you think it's fine though.

cmd/docker/docker.go
laurazard: @krissetto can you TAL at that real quick?

vvoland (approved):
LGTM

Issue #5075, opened 2024-05-15
docker should fail if config file is invalid


thaJeztah:
### Description

In situations where the config-file is invalid, we ignore the error, and proceed, which can lead to the config file being overwritten for an "empty" file (only some defaults included)

### Reproduce

```bash
docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock docker:26.1.2-cli sh
mkdir -p ~/.docker
echo '{"imagesFormat":"some special format I have carefully hand-crafted, artisinally"' > ~/.docker/config.json

docker logout
WARNING: Error loading config file: /root/.docker/config.json: unexpected EOF
Removing login credentials for https://index.docker.io/v1/

cat ~/.docker/config.json
{
	"auths": {}
}
```

### Expected behavior

The problematic area here is that any error returned from loading the config-file is printed as a warning; https://github.com/docker/cli/blob/61fe22f21a1a618d73cba67fec498f3f9d3b1422/cli/config/config.go#L125

A result of that is that if the file fails to load (malformed file, or other reason), `load` returns an error _AND_ an empty `configFile` struct; https://github.com/docker/cli/blob/61fe22f21a1a618d73cba67fec498f3f9d3b1422/cli/config/config.go#L117

In this case, we notify the user (warning), but continue the regular flow, which _may_ involve "updating the file" (`login` / `logout`, perhaps `docker context use`)

I was curious though about the "warning" instead of "error"; wondering if it was intended to ignore "not found" errors only, but I don't think that's the case; I think this is mostly because code moved around too much. First time this became a warning looks to be from https://github.com/moby/moby/commit/18c9b6c6455f116ae59cde8544413b3d7d294a5e

Which unlined some of the code; before that, the separate function returned an error, but was not handled. Going back further; it looks like this commit started to ignore errors https://github.com/moby/moby/commit/3bae188b8dc51911a44ea1c7b5681f9f07f9d3af
And https://github.com/moby/moby/commit/18962d0ff33cfa6ba2976aa459c766acfd23c1bf looks to settle that by fully ignoring.

So Tl;DR; I don't think there's a good reason / motivation for ignoring the error other than "convenience" (prefer happy path) but changing will mean that we need to change the signature of some functions to return.

### docker version

```bash
Client:
 Version:           26.1.2
 API version:       1.44 (downgraded from 1.45)
 Go version:        go1.21.10
 Git commit:        211e74b
 Built:             Wed May  8 13:59:48 2024
[...]

PR #5077, opened 2024-05-16
cli/config: improve handling of errors

thaJeztah:
Improving some parts of how we handle failures when loading the CLI configuration file. This relates to https://github.com/docker/cli/issues/5075, but does not yet fix that issue (only handling some additional errors);

- https://github.com/docker/cli/issues/5075

Changes in this PR should not likely cause a change in behavior (so could be included in a patch release); follow-up changes are still needed to handle various scenarios where we should hard-fail, but there's many different related code-paths, so this still needs work.

### cli/config: add test for dangling symlink for config-file

This may need further discussion, but we currently handle dangling
symlinks gracefully, so let's add a test for this, and verify that
we don't replace symlinks with a file.

### cli/config: TestLoadDefaultConfigFile: check that no warnings are printed

Loading the config should print no warnings on a successful load.

### cli/config: do not discard permission errors when loading config-file

When attempting to load a config-file that exists, but is not accessible for
the current user, we should not discard the error.

This patch makes sure that the error is returned by Load(), but does not yet
change LoadDefaultConfigFile, as this requires a change in signature.

```markdown changelog
Print a warning when the CLI does not have permissions to read the configuration file.
```

### DIFF
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -84,8 +84,10 @@ func LoadFromReader(configData io.Reader) (*configfile.ConfigFile, error) {
 	return &configFile, err
 }
 
-// Load reads the configuration files in the given directory, and sets up
-// the auth config information and returns values.
+// Load reads the configuration file ([ConfigFileName]) from the given directory.
+// If no directory is given, it uses the default [Dir]. A [*configfile.ConfigFile]
+// is returned containing the contents of the configuration file, or a default
+// struct if no configfile exists in the given location.
 func Load(configDir string) (*configfile.ConfigFile, error) {
 	if configDir == "" {
 		configDir = Dir()
@@ -118,7 +120,16 @@ func load(configDir string) (*configfile.ConfigFile, error) {
 }
 
 // LoadDefaultConfigFile attempts to load the default config file and returns
-// an initialized ConfigFile struct if none is found.
+// a reference to the ConfigFile struct. If none is found or when failing to load
+// the configuration file, it initializes a default ConfigFile struct. If no
+// credentials-store is set in the configuration file, it attempts to discover
+// the default store to use for the current platform.
+//
+// Important: LoadDefaultConfigFile prints a warning to stderr when failing to
+// load the configuration file, but otherwise ignores errors. Consumers should
+// consider using [Load] (and [credentials.DetectDefaultStore]) to detect errors
+// when updating the configuration file, to prevent discarding a (malformed)
+// configuration file.
 func LoadDefaultConfigFile(stderr io.Writer) *configfile.ConfigFile {
 	configFile, err := load(Dir())
 	if err != nil {
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -75,7 +75,7 @@ func Path(p ...string) (string, error) {
 }
 
 // LoadFromReader is a convenience function that creates a ConfigFile object from
-// a reader
+// a reader. It returns an error if configData is malformed.
 func LoadFromReader(configData io.Reader) (*configfile.ConfigFile, error) {
 	configFile := configfile.ConfigFile{
 		AuthConfigs: make(map[string]types.AuthConfig),
@@ -88,6 +88,10 @@ func LoadFromReader(configData io.Reader) (*configfile.ConfigFile, error) {
 // If no directory is given, it uses the default [Dir]. A [*configfile.ConfigFile]
 // is returned containing the contents of the configuration file, or a default
 // struct if no configfile exists in the given location.
+//
+// Load returns an error if a configuration file exists in the given location,
+// but cannot be read, or is malformed. Consumers must handle errors to prevent
+// overwriting an existing configuration file.
 func Load(configDir string) (*configfile.ConfigFile, error) {
 	if configDir == "" {
 		configDir = Dir()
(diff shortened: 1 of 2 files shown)


thaJeztah:
Ah 🙈 probably testing won't work if it's running as `root`;

```
59.75 === FAIL: cli/config TestLoadNoPermissions (0.00s)
59.75     config_test.go:79: assertion failed: error is nil, not "permission denied" (os.ErrPermission)
59.75 
59.75 === FAIL: cli/config TestLoadDefaultConfigFile/permission_error (0.00s)
59.75     config_test.go:377: assertion failed: string "" does not contain "WARNING:"
59.75     config_test.go:378: assertion failed: string "" does not contain "permission denied"
```

vvoland (approved):
LGTM

cli/config/config_test.go
```diff
@@ -48,6 +50,40 @@ func TestMissingFile(t *testing.T) {
...
+	// Make sure we kept the symlink.
+	fi, err := os.Lstat(cfgFile)
+	assert.NilError(t, err)
+	assert.Equal(t, fi.Mode()&os.ModeSymlink, os.ModeSymlink, "expected %v to be a symlink", cfgFile)
+}
+
+func TestLoadNoPermissions(t *testing.T) {
+	if runtime.GOOS != "windows" {
+		skip.If(t, os.Getuid() == 0, "cannot test permission denied when running as root")
+	}
```
thaJeztah, line 77: I tried if I could make the tests run as non-root, but ran into some issues, so for now, skipping this tests when running as root.

cli/config/config.go
```diff
@@ -84,8 +84,14 @@ func LoadFromReader(configData io.Reader) (*configfile.ConfigFile, error) {
 	return &configFile, err
 }
 
-// Load reads the configuration files in the given directory, and sets up
-// the auth config information and returns values.
+// Load reads the configuration file ([ConfigFileName]) from the given directory.
```
krissetto, line 87: what is ([ConfigFileName]) here?

cli/config/config.go
```diff
@@ -84,8 +84,14 @@ func LoadFromReader(configData io.Reader) (*configfile.ConfigFile, error) {
...
+// If no directory is given, it uses the default [Dir]. A [configfile.ConfigFile]
```
krissetto, line 88: nit: `[*configfile.ConfigFile]` here since it's a pointer?

cli/config/config.go
```diff
@@ -100,29 +106,36 @@ func load(configDir string) (*configfile.ConfigFile, error) {
...
-			// than it doesn't exist then stop
+			// It is OK for no configuration file to be present, in which
+			// case we return a default struct.
 			return configFile, nil
 		}
-		// if file is there but we can't stat it for any reason other
-		// than it doesn't exist then stop
-		return configFile, nil
+		// Any other error happening when failing to read the file must be returned.
+		return configFile, errors.Wrap(err, "loading config file")
```
krissetto, line 114: nit: maybe `errors.Wrapf(err, "loading config file: %s: ", filename)` like below so the errors have a consistent format?

cli/config/config.go
```diff
@@ -100,29 +106,36 @@ func load(configDir string) (*configfile.ConfigFile, error) {
...
 	if err != nil {
-		err = errors.Wrap(err, filename)
+		err = errors.Wrapf(err, "loading config file: %s: ", filename)
 	}
 	return configFile, err
 }
 
 // LoadDefaultConfigFile attempts to load the default config file and returns
-// an initialized ConfigFile struct if none is found.
+// an initialized ConfigFile struct if none is found or when failing to load
```
krissetto, line 125: I'm not sure about godoc best practices when returning pointers to structs, should we specify that here as well or only in the function signature?

krissetto (approved):
overal LGTM, with some godoc nits/questions

cli/config/config.go
thaJeztah: They're "doc links"; if you use brackets and put a reference to a type, then GoDoc renders them as links.

You can test those using `pkgsite` (`go install golang.org/x/pkgsite/cmd/pkgsite@latest`), and then browse the docs (it's a bit fiddly on the CLI because of go modules, so I just ran it in a container);

Screenshot 2024-05-21 at 16 16 32
Screenshot 2024-05-21 at 16 16 52

cli/config/config.go
thaJeztah: O left that out, because `os.Open` returns a  `os.PathError` / `fs.PathError`; https://pkg.go.dev/os#PathError and those already include the path, which would be `open /no/such/file: no such file or directory`, so manually adding the filename would now have the filename in the error twice.

cli/config/config.go
thaJeztah: Yeah, I left it out for now; I think we should consider deprecating this function, or at least provide an alternative that does return the error.

cli/config/config.go
krissetto: Ah, good to know! I didn't think that having just the type name would be enough for the linking to occur

cli/config/config.go
thaJeztah: Oh! Looks like that works as well; I thought GoDoc links wouldn't accept that (so requiring to manually construct a link), but it looks like it does;

Screenshot 2024-05-22 at 10 20 03

cli/config/config.go
thaJeztah: I tried to slightly rephrase the GoDoc to make it more clear that it returns a reference, and that it initialises a default struct if it fails to load / find a config-file.

thaJeztah:
@krissetto I made some changes; PTAL if it LGTY

krissetto:
> @krissetto I made some changes; PTAL if it LGTY

looking good @thaJeztah 😎

thaJeztah:
Let me bring this one in; I'll try to spend some time on fixing the actual handling related to https://github.com/docker/cli/issues/5075, but taking it small steps at a time.

PR #5078, opened 2024-05-17
plugins/hooks: Don't show empty hook messages

vvoland:
**- What I did**
Don't show `Next steps:` with no messages at all when plugin returns an unitialized value of `HookMessage` (zero-initialization sets its type to NextSteps and empty template).

**- How to verify it**
Unit test

```markdown changelog
Don't show empty hints when plugins returns an empty hook message.
```

### DIFF
--- a/cli-plugins/manager/hooks_test.go
+++ b/cli-plugins/manager/hooks_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"gotest.tools/v3/assert"
+	is "gotest.tools/v3/assert/cmp"
 )
 
 func TestGetNaiveFlags(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cli-plugins/manager/hooks_test.go
```diff
@@ -108,3 +109,34 @@ func TestPluginMatch(t *testing.T) {
...
+			processed:   []string{},
+			expectedOut: []string{},
+		},
+		{
+			processed:   []string{"", ""},
+			expectedOut: []string{},
+		},
+		{
+			processed:   []string{"Some hint", "", "Some other hint"},
+			expectedOut: []string{"Some hint", "", "Some other hint"},
```
krissetto, line 128: Are we sure we want to include empty lines if the plugin returns them?

vvoland: Not sure either, but this was the behavior we had so didn't want to change anything other than just handling the empty message.

krissetto: fair enough for now

vvoland: Tbh, I would be okay with removing empty lines (actually that was the first thing I did, but then decided not to change the behavior). I can do that in this PR if we want to, @laurazard  any opinions?

laurazard (approved):
LGTM

I was wondering whether we should do this since, ultimately, plugins should not print valid `HookMessage` structs with empty messages if a message is not supposed to be printed, but I don't think we lose anything by being defensive and skipping these on the CLI side.

(I wonder if we should have a debug-level log in the case we skip any for this case, could be better for debugging vs. silently failing)

krissetto:
> LGTM
> 
> I was wondering whether we should do this since, ultimately, plugins should not print valid `HookMessage` structs with empty messages if a message is not supposed to be printed, but I don't think we lose anything by being defensive and skipping these on the CLI side.
> 
> (I wonder if we should have a debug-level log in the case we skip any for this case, could be better for debugging vs. silently failing)

+ 1 for staying safe. The repercussions of plugins not doing things "correctly" falls on the CLI experience, so we should safeguard it where possible IMHO

vvoland:
Added a debug log (visible only when `--debug` flag is present).

Issue #1463, opened 2018-10-19
Proposal for new feature: DOCKER_SSH_OPTIONS environment variable


andrevtg:
Proposal for new feature: DOCKER_SSH_OPTIONS environment variable

The new "ssh://" protocol for DOCKER_HOST is great, but sometimes we need to use different ssh options than the default. A common scenario if the need to point to a project-specific ssh config file instead of the default one at "~/.ssh/config".

A PR has been submitted: 
https://github.com/docker/cli/pull/1462

bsideup:
@thaJeztah  this would be amazing to have, since otherwise it is not possible to pass `-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null` to the SSH command.

thaJeztah:
Does setting it in your config not work? https://askubuntu.com/a/87452

bsideup:
@thaJeztah it does, but I don't want to set it "globally" (as it is not a good default behaviour, and my IPs are random so I can't use a mask in it), but rather "per session" or even "per invocation" (`-o` flag of `ssh` exists for the same reason)

andrevtg:
I could try pushing another PR, but I am not sure if this would be welcomed.

Being able to use a custom per-project ssh.config looks useful to me.

PR #5082, opened 2024-05-19
Support custom SSH config file for SSH connection

### LINKED ISSUE #1301: Allow to define SSH connection options
**Description**

The new ssh connection introduced in #1014 does not allow to define many SSH options that are sometimes needed.

It indeed supports `~/.ssh/config`, but what about portability?

Possible solutions that come to my mind:

- Allow to specify the SSH command itself, with all but the host, user and port.
- Allow to specify a custom config file (e.g. `ssh -F /tmp/other_config`) that can be filled by a script previously.
- Allow to specify SSH options (e.g. `docker -H ssh://example.com -o ForwardAgent=yes -o 'ProxyCommand=/usr/bin/nc -X connect -x 192.0.2.0:8080 %h %p ' container ls`)

Current workaround would be for a script to override current `$PATH` and prepend a different `ssh` binary that proxies to the original one. But that's hacky!

evolutics:
SSH Docker host connections support specifying an SSH username, hostname, port, and socket path with `ssh://…`. Furthermore, the default user SSH config file located at `~/.ssh/config` is respected.

However, it would be useful to specify a path to a custom SSH config on the CLI. Use cases are per-project SSH configs or SSH configs generated on the fly (e.g., by `vagrant ssh-config`).

**- What I did**

With this commit, a path to a custom SSH config can be specified as part of the SSH Docker host URL.

This closes #1301. Related issues: #1463, #2539.

**- How I did it**

An SSH config file can be specified by appending a URL fragment component to the end of the SSH address (that is, behind a hash `#`). The given file path is passed to SSH using the `ssh -F` option. Example:

```
$ docker --host ssh://user@192.168.64.5#/path/to/ssh_config ps
```

**- How to verify it**

Set up:

```console
$ docker buildx bake

$ cat <<EOF >Vagrantfile
Vagrant.configure("2") do |config|
  config.vm.box = "bento/ubuntu-24.04"
  config.vm.provision "docker"
end
EOF

$ vagrant up

$ vagrant ssh-config --host my-host >my_ssh_config
```

Test:

```console
$ build/docker --host 'ssh://my-host#my_ssh_config' run -dit nginx

$ build/docker --host 'ssh://my-host#my_ssh_config' ps
CONTAINER ID   IMAGE     COMMAND                  CREATED         STATUS         PORTS     NAMES
856870d88719   nginx     "/docker-entrypoint.…"   5 seconds ago   Up 3 seconds   80/tcp    serene_mcclintock

$ build/docker ps
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
```

Tear down:

```console
$ vagrant destroy

$ rm -fr .vagrant/ my_ssh_config Vagrantfile
```

```markdown changelog
Support custom SSH config file for SSH Docker host connections using URL of the form `ssh://host#/path/to/ssh_config`.
```

custom_ssh_config

### DIFF
--- a/cli/connhelper/ssh/ssh.go
+++ b/cli/connhelper/ssh/ssh.go
@@ -34,23 +34,25 @@ func ParseURL(daemonURL string) (*Spec, error) {
 	if u.RawQuery != "" {
 		return nil, errors.Errorf("extra query after the host: %q", u.RawQuery)
 	}
-	if u.Fragment != "" {
-		return nil, errors.Errorf("extra fragment after the host: %q", u.Fragment)
-	}
+	sp.SSHConfig = u.Fragment
 	return &sp, err
 }
 
 // Spec of SSH URL
 type Spec struct {
-	User string
-	Host string
-	Port string
-	Path string
+	User      string
+	Host      string
+	Port      string
+	Path      string
+	SSHConfig string
 }
 
 // Args returns args except "ssh" itself combined with optional additional command args
 func (sp *Spec) Args(add ...string) []string {
 	var args []string
+	if sp.SSHConfig != "" {
+		args = append(args, "-F", sp.SSHConfig)
+	}
 	if sp.User != "" {
 		args = append(args, "-l", sp.User)
 	}
--- a/cli/connhelper/ssh/ssh_test.go
+++ b/cli/connhelper/ssh/ssh_test.go
@@ -28,6 +28,13 @@ func TestParseURL(t *testing.T) {
 			},
 		},
 		{
+			url: "ssh://foo#/path/to/ssh_config",
+			expectedArgs: []string{
+				"-F", "/path/to/ssh_config",
+				"--", "foo",
+			},
+		},
+		{
 			url:           "ssh://me:passw0rd@foo",
 			expectedError: "plain-text password is not supported",
 		},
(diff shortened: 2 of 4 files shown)


evolutics:
I'm aware of this [comment](https://github.com/docker/cli/issues/1301#issuecomment-1410152520) from something over a year ago:

> We have no plans at the moment to extend the scope of this feature. So, I'm closing this issue for now but we can possibly reconsider in the future.

Thus, feel free to discard this PR.

I think the cost-benefit ratio of this change is quite OK though: not much code (cost) for the flexibility that this would give users (benefit).

Regarding the API, I'm not sure whether the URL fragment is best suited. An alternative would be the URL query string.

neersighted:
I'm not sure this feature makes sense as-is, given that we already use `ssh://url/path/to/docker.sock` (the path component of the SSH URL is used for the remote socket path), it seems somewhat confusing.

In general, I feel very hesitant to add more microformat-like options here. I think a much more sane approach might be extending the context to include arbitrary SSH options, which allows both direct config, and specifying an alternate config path. You wouldn't be able to use those options with `DOCKER_HOST`, but contexts are encouraged over `DOCKER_HOST` anyway, and it would be a logical evolution of our rich 'remote daemon' config mechanism.

thaJeztah:
Thought I commented already, but looks like I didn't; yes, also not comfortable with adding a new micro format at least. We accepted the `/path/to/docker.sock` as it followed existing formats (see https://github.com/docker/cli/pull/4073#pullrequestreview-1327204803), but I'd rather not come up with our own conventions otherwise.

evolutics:
Thanks for your feedback, both. I agree with the confusion this uncommon format could cause. I'll think of other options; although I prefer the declarative nature of `DOCKER_HOST` over the stateful `DOCKER_CONTEXT`.

PR #5085, opened 2024-05-22
docs: deprecate daemon's api-cors-header flag

akerouanton:
Related to:

- https://github.com/moby/moby/pull/45313

**- What I did**

Update the "Deprecated Engine Features" doc page to warn about daemon's `api-cors-header` deprecation.

### DIFF
--- a/docs/deprecated.md
+++ b/docs/deprecated.md
@@ -50,6 +50,7 @@ The table below provides an overview of the current status of deprecated feature
 
 | Status     | Feature                                                                                                                            | Deprecated | Remove |
 |------------|------------------------------------------------------------------------------------------------------------------------------------|------------|--------|
+| Deprecated | [API CORS headers](#api-cors-headers)                                                                                              | v27.0      | v28.0  |
 | Deprecated | [Unauthenticated TCP connections](#unauthenticated-tcp-connections)                                                                | v26.0      | v27.0  |
 | Deprecated | [`Container` and `ContainerConfig` fields in Image inspect](#container-and-containerconfig-fields-in-image-inspect)                | v25.0      | v26.0  |
 | Deprecated | [Deprecate legacy API versions](#deprecate-legacy-api-versions)                                                                    | v25.0      | v26.0  |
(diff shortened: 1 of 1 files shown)


docs/deprecated.md
```diff
@@ -112,6 +113,22 @@ The table below provides an overview of the current status of deprecated feature
 | Removed    | [`--run` flag on `docker commit`](#--run-flag-on-docker-commit)                                                                    | v0.10      | v1.13  |
 | Removed    | [Three arguments form in `docker import`](#three-arguments-form-in-docker-import)                                                  | v0.6.7     | v1.12  |
 
+### API CORS headers
+
+**Deprecated in Release: v27.0**
+**Target For Removal In Release: v28.0**
+
+Daemon's `api-cors-header` flag is deemed insecure as it could be enabled
+without any Authz plugin enabled beforehand.
```
thaJeztah, line 122: Wondering if we can make this description more generic; Authz plugins are _one way_ to do this, but are also a bit of a corner-case use (there may be many other approaches). The `authz plugin` feature also isn't very actively maintained, so somewhat trying to avoid it coming across as a recommendation.

cc @dvdksn in case you have good suggestions.

dvdksn: could we just pretend Authz was short for authorization?

```suggestion
Daemon's `api-cors-header` flag is deemed insecure as it could be enabled
without any authorization plugin enabled beforehand.
```

akerouanton: Thanks @dvdksn!

thaJeztah (approved):
LGTM

dvdksn (approved):

thaJeztah:
I'll bring this one in; one more thing we may need to consider is to add some notes on https://github.com/docker/cli/blob/6a4d38c7f2fe17351f78ff9dfe7ec2ee9b15f1df/docs/reference/dockerd.md


The dockerd docs are not (yet) generated from Cobra, so that means we won't have a table with options / annotations that flags are deprecated. And we also have the option documented for the daemon.json config.

@dvdksn if you have any good ideas how to update those docs; let me know. I don't expect it to be a popular option, but .. well .. someone, somewhere will likely use it.

PR #5090, opened 2024-05-26
Dockerfile: update ALPINE_VERSION to 3.20

thaJeztah:
Update to the current version of Alpine, which is also the default for the golang:alpine image

```markdown changelog
Update Dockerfile to Alpine 3.20
```


krissetto (approved):
LGTM.  

Could there be any differences between this alpine version and the one used by `golangci/golangci-lint:${GOLANGCI_LINT_VERSION}-alpine` that could cause any issues? Would it make sense to try and stay on the same underlying version when possible?

crazy-max (approved):

thaJeztah:
Thx both! Let me bring this one in

PR #5092, opened 2024-05-26
migrate to new github.com/containerd/platforms package

thaJeztah:
- relates to https://github.com/moby/moby/pull/47142

The github.com/containerd/containerd/platforms package was moved to a separate module in preparation of the containerd v2.0 release.

Switch to the new module, which means we also remove containerd as a direct dependency.


laurazard (approved):
LGTM

thaJeztah:
Thx! Let me bring this in; it's good to remove the direct dependency on containerd (I should probably add a dependency-check in golangci-lint to prevent it from accidentally being reintroduced 🤔); will do that in a follow-up.

PR #5094, opened 2024-05-30
context list: temporarily add ContextType to JSON output

thaJeztah:
- [x] depends on / stacked on https://github.com/docker/cli/pull/5097

Docker Desktop currently ships with the "cloud integration" wrapper, which outputs an additional ContextType field in the JSON output.

While this field is non-standard, it made its way into Visual Studio's Docker integration, which uses this to exclude "aci" and "eci" context types that are not supported by Visual Studio.

This patch;

- conditionally adds a ContextType field to the JSON output
- but ONLY when using the default "{{json .}}" or "json" formats (which are the formats used by Visual Studio)
- if the context is a "aci" or "eci" context, that type is preserved, otherwise the default "moby" type is used.

```markdown changelog
Add a temporary `ContextType` field to the JSON output of "docker context ls" for backward-compatibility with Visual Studio.
```

### DIFF
--- a/cli/command/context/list.go
+++ b/cli/command/context/list.go
@@ -66,6 +66,8 @@ func runList(dockerCli command.Cli, opts *listOptions) error {
 				Name:    rawMeta.Name,
 				Current: isCurrent,
 				Error:   err.Error(),
+
+				ContextType: getContextType(nil, opts.format),
 			})
 			continue
 		}
@@ -80,6 +82,8 @@ func runList(dockerCli command.Cli, opts *listOptions) error {
 			Description:    meta.Description,
 			DockerEndpoint: dockerEndpoint.Host,
 			Error:          errMsg,
+
+			ContextType: getContextType(meta.AdditionalFields, opts.format),
 		}
 		contexts = append(contexts, &desc)
 	}
@@ -96,6 +100,8 @@ func runList(dockerCli command.Cli, opts *listOptions) error {
 			Name:    curContext,
 			Current: true,
 			Error:   errMsg,
+
+			ContextType: getContextType(nil, opts.format),
 		})
 	}
 	sort.Slice(contexts, func(i, j int) bool {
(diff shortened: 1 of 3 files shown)


cli/command/context/list_test.go
```diff
@@ -30,6 +31,25 @@ func TestList(t *testing.T) {
 	golden.Assert(t, cli.OutBuffer().String(), "list.golden")
 }
 
+func TestListJSON(t *testing.T) {
```
krissetto, line 34: nit: shall we also add a small test for the other two (unexpected) values of ContextType for the old cloud integrations?

cli/command/context/list.go
```diff
@@ -111,6 +117,28 @@ func runList(dockerCli command.Cli, opts *listOptions) error {
 	return nil
 }
 
+// getContextType sets the LegacyContextType field for compatibility with
+// Visual Studio, which depends on this field from the "cloud integration"
+// wrapper.
+//
+// https://github.com/docker/compose-cli/blob/c156ce6da4c2b317174d42daf1b019efa87e9f92/api/context/store/contextmetadata.go#L28-L34
+// https://github.com/docker/compose-cli/blob/c156ce6da4c2b317174d42daf1b019efa87e9f92/api/context/store/store.go#L34-L51
+func getContextType(meta map[string]any, format string) string {
```
krissetto, line 126: just out of an abundance of caution, should we also signal that this shouldn't be used for anything aside of this patch? or directly leave a TODO to remove this one the special binary is no longer needed?

thaJeztah:
Yes(ish); I wanted to avoid having to build some forked version of the CLI, so tried to create a patch with as minimal-as-possible impact, so that we could included it in a v26.1.4 patch release.

cli/command/context/list.go
thaJeztah: Ah, yeah, I thought; let me mark the field as deprecated to be "somewhat" of a reminder; don't use. And this function is non-exported, so at least nobody would be able to use it.

krissetto:
> so that we could included it in a v26.1.4 patch release.

if this does go out to "everyone" (and not only dd users), we need to document this in a visible way and be ready in case anything else breaks (i guess, just in case any other project depends on the shape of that json in even worse ways?)

cli/command/context/list_test.go
thaJeztah: This is done; I'll leave a comment below

thaJeztah:
Maybe it's cleaner to move those 3 commits to a separate PR, so that this PR is only showing the changes related to ContextType; let me open a separate PR for the first 3 commits

laurazard (approved):
LGTM

thaJeztah:
I'll rebase this one to get rid of the other commits

thaJeztah:
Rebased and moved out of draft

cli/command/context/list.go
```diff
@@ -111,6 +117,30 @@ func runList(dockerCli command.Cli, opts *listOptions) error {
...
+// https://github.com/docker/compose-cli/blob/c156ce6da4c2b317174d42daf1b019efa87e9f92/api/context/store/contextmetadata.go#L28-L34
+// https://github.com/docker/compose-cli/blob/c156ce6da4c2b317174d42daf1b019efa87e9f92/api/context/store/store.go#L34-L51
+//
+// TODO(thaJeztah): remove this and [ClientContext.ContextType] once Visual Studio is updated to no longer depend on this.
+func getContextType(meta map[string]any, format string) string {
+	if format != formatter.JSONFormat && format != formatter.JSONFormatKey {
+		// We only need the ContextType field when formatting as JSON,
+		// which is the format-string used by Visual Studio to detect the
+		// context-type.
+		return ""
```
krissetto, line 133: One curiosity: could it also be possible to check if the current context being used is from DD? Do we have ways of determining that from here, aside from the context name itself (which i totally wouldn't rely on)?

krissetto (approved):
LGTM, approving as to not block this.

PR #5095, opened 2024-05-30
[26.1 backport] context list: temporarily add ContextType to JSON output

thaJeztah:
- backport https://github.com/docker/cli/pull/5097
- backport https://github.com/docker/cli/pull/5094

Docker Desktop currently ships with the "cloud integration" wrapper, which outputs an additional ContextType field in the JSON output.

While this field is non-standard, it made its way into Visual Studio's Docker integration, which uses this to exclude "aci" and "eci" context types that are not supported by Visual Studio.

This patch;

- conditionally adds a ContextType field to the JSON output
- but ONLY when using the default "{{json .}}" or "json" formats (which are the formats used by Visual Studio)
- if the context is a "aci" or "eci" context, that type is preserved, otherwise the default "moby" type is used.

```markdown changelog
Add a temporary `ContextType` field to the JSON output of "docker context ls" for backward-compatibility with Visual Studio.
```

### DIFF
--- a/cli/command/context/create.go
+++ b/cli/command/context/create.go
@@ -24,6 +24,10 @@ type CreateOptions struct {
 	Description string
 	Docker      map[string]string
 	From        string
+
+	// Additional Metadata to store in the context. This option is not
+	// currently exposed to the user.
+	metaData map[string]any
 }
 
 func longCreateDescription() string {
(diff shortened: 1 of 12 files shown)


sam-thibault (approved):

thaJeztah:
Let's bring this one in

PR #5097, opened 2024-05-31
cli/command/context: improve tests and add coverage for custom metadata in contexts

thaJeztah:
- relates to https://github.com/docker/cli/pull/5094

### cli/command/context: test inspecting context with custom metadata

The CLI does not currently expose options to add custom metadata to
contexts, but contexts support them.

- update test-utilities to allow setting custom metadata
- update the inspect test to verify that custom metadata is included
  when inspecting a context.

### cli/command/context: list: add test for JSON output


cli/command/context/testdata/list-json.golden
```diff
@@ -0,0 +1,5 @@
+{"Current":false,"Description":"description of context1","DockerEndpoint":"https://someswarmserver.example.com","Error":"","Name":"context1"}
```
krissetto, line 1: Given the TestListJSON test above, shouldn't context1-2-3 contain the additional `Type:` metadata?

thaJeztah: Not currently, no; the JSON output in list format doesn't output the "raw" metadata, and doesn't include custom fields currently.

_Should_ it include these? Maybe, maybe not (at least `inspect` shows them), but it's currently expected for them to not be here.

laurazard (approved):
LGTM, thanks for the patch

krissetto (approved):

PR #5098, opened 2024-05-31
add support for DOCKER_CUSTOM_HEADERS env-var (experimental)

thaJeztah:
- related: https://github.com/moby/moby/pull/32150

This environment variable allows for setting additional headers to be sent by the client. Headers set through this environment variable are added to headers set through the config-file (through the HttpHeaders field).

This environment variable can be used in situations where headers must be set for a specific invocation of the CLI, but should not be set by default, and therefore cannot be set in the config-file.

:warning: WARNING: If both config and environment-variable are set, the environment
variable currently overrides all headers set in the configuration file.
This behavior may change in a future update, as we are considering the
environment variable to be appending to existing headers (and to only
override headers with the same name).

**- What I did**

**- How I did it**

**- How to verify it**

```markdown changelog
Add support for `DOCKER_CUSTOM_HEADERS` environment variable

This environment variable allows for setting additional headers to be sent by the client. Headers set through this environment variable are added to headers set through the config-file (through the HttpHeaders field).

This environment variable can be used in situations where headers must be set for a specific invocation of the CLI, but should not be set by default, and therefore cannot be set in the config-file.
```

### DIFF
--- a/cli/command/cli_options.go
+++ b/cli/command/cli_options.go
@@ -2,13 +2,18 @@ package command
 
 import (
 	"context"
+	"encoding/csv"
 	"io"
+	"net/http"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/docker/cli/cli/streams"
 	"github.com/docker/docker/client"
+	"github.com/docker/docker/errdefs"
 	"github.com/moby/term"
+	"github.com/pkg/errors"
 )
 
 // CLIOption is a functional argument to apply options to a [DockerCli]. These
@@ -106,5 +111,109 @@ func WithAPIClient(c client.APIClient) CLIOption {
 	return func(cli *DockerCli) error {
 		cli.client = c
 		return nil
+	}
+}
+
+// envOverrideHTTPHeaders is the name of the environment-variable that can be
+// used to set custom HTTP headers to be sent by the client. This environment
+// variable is the equivalent to the HttpHeaders field in the configuration
+// file.
+//
+// WARNING: If both config and environment-variable are set, the environment
+// variable currently overrides all headers set in the configuration file.
+// This behavior may change in a future update, as we are considering the
+// environment variable to be appending to existing headers (and to only
+// override headers with the same name).
+//
+// While this env-var allows for custom headers to be set, it does not allow
+// for built-in headers (such as "User-Agent", if set) to be overridden.
+// Also see [client.WithHTTPHeaders] and [client.WithUserAgent].
+//
+// This environment variable can be used in situations where headers must be
+// set for a specific invocation of the CLI, but should not be set by default,
+// and therefore cannot be set in the config-file.
+//
+// envOverrideHTTPHeaders accepts a comma-separated (CSV) list of key=value pairs,
+// where key must be a non-empty, valid MIME header format. Whitespaces surrounding
+// the key are trimmed, and the key is normalised. Whitespaces in values are
+// preserved, but "key=value" pairs with an empty value (e.g. "key=") are ignored.
+// Tuples without a "=" produce an error.
+//
+// It follows CSV rules for escaping, allowing "key=value" pairs to be quoted
+// if they must contain commas, which allows for multiple values for a single
+// header to be set. If a key is repeated in the list, later values override
+// prior values.
+//
+// For example, the following value:
+//
+//	one=one-value,"two=two,value","three= a value with whitespace  ",four=,five=five=one,five=five-two
+//
+// Produces four headers (four is omitted as it has an empty value set):
+//
+// - one (value is "one-value")
+// - two (value is "two,value")
+// - three (value is " a value with whitespace  ")
+// - five (value is "five-two", the later value has overridden the prior value)
+const envOverrideHTTPHeaders = "DOCKER_CUSTOM_HEADERS"
+
+// withCustomHeadersFromEnv overriding custom HTTP headers to be sent by the
+// client through the [envOverrideHTTPHeaders] environment-variable. This
+// environment variable is the equivalent to the HttpHeaders field in the
+// configuration file.
+//
+// WARNING: If both config and environment-variable are set, the environment-
+// variable currently overrides all headers set in the configuration file.
+// This behavior may change in a future update, as we are considering the
+// environment-variable to be appending to existing headers (and to only
+// override headers with the same name).
+//
+// TODO(thaJeztah): this is a client Option, and should be moved to the client. It is non-exported for that reason.
+func withCustomHeadersFromEnv() client.Opt {
+	return func(apiClient *client.Client) error {
+		value := os.Getenv(envOverrideHTTPHeaders)
+		if value == "" {
+			return nil
+		}
+		csvReader := csv.NewReader(strings.NewReader(value))
+		fields, err := csvReader.Read()
+		if err != nil {
+			return errdefs.InvalidParameter(errors.Errorf("failed to parse custom headers from %s environment variable: value must be formatted as comma-separated key=value pairs", envOverrideHTTPHeaders))
+		}
+		if len(fields) == 0 {
+			return nil
+		}
+
+		env := map[string]string{}
+		for _, kv := range fields {
+			k, v, hasValue := strings.Cut(kv, "=")
+
+			// Only strip whitespace in keys; preserve whitespace in values.
+			k = strings.TrimSpace(k)
+
+			if k == "" {
+				return errdefs.InvalidParameter(errors.Errorf(`failed to set custom headers from %s environment variable: value contains a key=value pair with an empty key: '%s'`, envOverrideHTTPHeaders, kv))
+			}
+
+			// We don't currently allow empty key=value pairs, and produce an error.
+			// This is something we could allow in future (e.g. to read value
+			// from an environment variable with the same name). In the meantime,
+			// produce an error to prevent users from depending on this.
+			if !hasValue {
+				return errdefs.InvalidParameter(errors.Errorf(`failed to set custom headers from %s environment variable: missing "=" in key=value pair: '%s'`, envOverrideHTTPHeaders, kv))
+			}
+
+			env[http.CanonicalHeaderKey(k)] = v
+		}
+
+		if len(env) == 0 {
+			// We should probably not hit this case, as we don't skip values
+			// (only return errors), but we don't want to discard existing
+			// headers with an empty set.
+			return nil
+		}
+
+		// TODO(thaJeztah): add a client.WithExtraHTTPHeaders() function to allow these headers to be _added_ to existing ones, instead of _replacing_
+		//  see https://github.com/docker/cli/pull/5098#issuecomment-2147403871  (when updating, also update the WARNING in the function and env-var GoDoc)
+		return client.WithHTTPHeaders(env)(apiClient)
 	}
 }
--- a/docs/reference/commandline/docker.md
+++ b/docs/reference/commandline/docker.md
@@ -118,13 +118,14 @@ The following list of environment variables are supported by the `docker` comman
 line:
 
 | Variable                      | Description                                                                                                                                                                                                                                                       |
-| :---------------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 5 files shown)


cli/command/cli_options.go
```diff
@@ -107,3 +112,93 @@ func WithAPIClient(c client.APIClient) CLIOption {
...
+func withCustomHeadersFromEnv() client.Opt {
+	return func(apiClient *client.Client) error {
+		value := os.Getenv(envOverrideHTTPHeaders)
+		if value == "" {
+			return nil
+		}
+		csvReader := csv.NewReader(strings.NewReader(value))
+		fields, err := csvReader.Read()
+		if err != nil {
+			return errdefs.InvalidParameter(errors.Wrapf(err, "failed to set custom headers from %s environment variable", envOverrideHTTPHeaders))
```
krissetto, line 169: From the docs i see these three main error conditions on `csvReader.Read()`:

1) > If the record has an unexpected number of fields, Read returns the record along with the error [ErrFieldCount](https://pkg.go.dev/encoding/csv#ErrFieldCount). 

   Given we'll only have a single "row" of data, `ErrFieldCount` shouldn't really occur;

2) > If the record contains a field that cannot be parsed, Read returns a partial record along with the parse error. The partial record contains all fields read before the error.  

   I don't think we want to manage partial records, so we should return when ending up on that error;


3)  > If there is no data left to be read, Read returns nil, [io.EOF](https://pkg.go.dev/io#EOF).

    We shouldn't ever encounter an empty reader

Given the above, in this error msg shall we mention that the CSV passed in the `DOCKER_CUSTOM_HEADERS` is likely malformed? IIUC that'd be the most likely reason for encountering the error, and might be a bit more helpful to the user than this fairly generic message. WDYT?

laurazard:
I'm wondering about the utility of adding an env var that **completely replaces** headers set via over settings (such as through the config file) vs just appending/overriding headers with the same name. I feel like in most cases, adding headers might be more useful, and if in some cases there's a need to overwrite some values, that can be done on a per-header basis, rather than completely ignoring otherwise configured headers.

cli/command/cli_options.go
```diff
@@ -107,3 +112,93 @@ func WithAPIClient(c client.APIClient) CLIOption {
...
+			k, v, hasValue := strings.Cut(kv, "=")
+
+			// Only strip whitespace in keys; preserve whitespace in values.
+			k = strings.TrimSpace(k)
+
+			if k == "" {
+				return errdefs.InvalidParameter(errors.Errorf("failed to set custom headers from %s environment variable: value contains a key=value pair with an empty key", envOverrideHTTPHeaders))
+			}
+
+			// We don't currently allow key=value pairs, and produce an error.
```
krissetto, line 186: nit: Did you mean to write "... **empty** key=value pairs,  ..."?

cli/command/cli_options.go
```diff
@@ -107,3 +112,93 @@ func WithAPIClient(c client.APIClient) CLIOption {
...
+			// This is something we could allow in future (e.g. to read value
+			// from an environment variable with the same name). In the meantime,
+			// produce an error to prevent users from depending on this.
+			if !hasValue {
+				return errdefs.InvalidParameter(errors.Errorf(`failed to set custom headers from %s environment variable: missing "=" in key=value pair %q`, envOverrideHTTPHeaders, kv))
+			}
+
+			if v == "" {
+				// Ignore empty values, and consider them to not be set
+				continue
```
krissetto, line 196: I'm not totally sure we should ignore empty values, as it could add some confusion to the user (maybe they were expecting the header to be set but empty). Are there scenarios in which we might expect this to occur?

thaJeztah: My somewhat anticipation here was for `My-Header=${SOME_ENV_VAR}`. I'd be ok with starting with this being an error condition though (which would still keep all options open)

cli/command/cli_options.go
thaJeztah: Hm, yes, good point; I was thinking that the CSV error might be useful, but perhaps it's not, and we should just produce a generic "this is likely the wrong format, we expect a comma-separated list of `key=value`.

cli/command/cli_options.go
thaJeztah: PROBABLY! 🙈

thaJeztah:
> I'm wondering about the utility of adding an env var that completely replaces headers set via over settings (such as through the config file) vs just appending/overriding headers with the same name. I feel like in most cases, adding headers might be more useful, and if in some cases there's a need to overwrite some values, that can be done on a per-header basis, rather than completely ignoring otherwise configured headers.

Yes, this is always the tricky one; override vs merge.

I guess in most situations, an env-var would override config (`flag > env-var > config`), but as this is a potential _list_ of options, that tends to make things more of a grey area.

The tricky part with _merging_ is that we would also need a way to _unset_ env-vars; in some parts of our code, I think we have "empty value" meaning "remove", although that one was specifically for User-Agent (allowing removing the user-agent); https://github.com/moby/moby/blob/989426d303a334db4d97f72efe6637d64220d01a/client/request.go#L244-L266

cli/command/cli_options.go
krissetto: fair point, always try to make user facing errors "explicit" just in case (so they're in our control rather then depending on the library maintainers). No hard feelings if the error it returns is good enough👍

thaJeztah: We should look indeed what the errors look like, and either "replace" all errors it produces, or if it's only a specific type; for that type.

I haven't found the time yet to try what they look like (perhaps we need to add a test-case for it)

cli/command/cli_options.go
thaJeztah: OK, gave this some thinking, and it's probably fine to just set an empty value, and leave it an exercise to the user to handle `My-Header=${SOME_ENV_VAR}` in situations where `$SOME_ENV_VAR` is not set (they can use the magic of (Ba)sh to handle unset env-vars.

cli/command/cli_options.go
thaJeztah: I updated the error to not include the CSV error, and made it more generic

cli/command/cli_options.go
thaJeztah: Fixed :)

thaJeztah:
> > I'm wondering about the utility of adding an env var that completely replaces headers set via over settings (such as through the config file) vs just appending/overriding headers with the same name. I feel like in most cases, adding headers might be more useful, and if in some cases there's a need to overwrite some values, that can be done on a per-header basis, rather than completely ignoring otherwise configured headers.
> 
> Yes, this is always the tricky one; override vs merge.
> 
> I guess in most situations, an env-var would override config (`flag > env-var > config`), but as this is a potential _list_ of options, that tends to make things more of a grey area.
> 
> The tricky part with _merging_ is that we would also need a way to _unset_ env-vars; in some parts of our code, I think we have "empty value" meaning "remove", although that one was specifically for User-Agent (allowing removing the user-agent); https://github.com/moby/moby/blob/989426d303a334db4d97f72efe6637d64220d01a/client/request.go#L244-L266

I updated the PR and added a WARNING to the GoDoc to mention that the behavior may change (headers to be _appended_ in future, and only overwrite same-name ones)

cli/command/cli_options.go
```diff
@@ -108,3 +113,107 @@ func WithAPIClient(c client.APIClient) CLIOption {
 		return nil
 	}
 }
+
+// envOverrideHTTPHeaders is the name of the environment variable that can be
```
krissetto, line 117: nit: a bit pedantic, but should we go with only one way of writing these in the docstrings?
- `environment variable`
- `environment-variable`
- `env-var`

cli/command/cli_options.go
```diff
@@ -108,3 +113,107 @@ func WithAPIClient(c client.APIClient) CLIOption {
...
+// and therefore cannot be set in the config-file.
+//
+// envOverrideHTTPHeaders accepts a comma-separated (CSV) list of key=value pairs,
+// where key must be a non-empty, valid MIME header format. Whitespaces surrounding
+// the key are trimmed, and the key is normalised. Whitespaces in values are
+// preserved, but "key=value" pairs with an empty value (e.g. "key=") are ignored.
+// Tuples without a "=" produce an error.
+//
+// It follows CSV rules for escaping, allowing "key=value" pairs to be quoted
+// if they must contain commas. which allows for multiple values for a single
```
krissetto, line 143: nit: we need `,` instead of `.` here

vvoland (approved):

cli/command/cli_options.go
```diff
@@ -108,3 +113,107 @@ func WithAPIClient(c client.APIClient) CLIOption {
...
+
+		env := map[string]string{}
+		for _, kv := range fields {
+			k, v, hasValue := strings.Cut(kv, "=")
+
+			// Only strip whitespace in keys; preserve whitespace in values.
+			k = strings.TrimSpace(k)
+
+			if k == "" {
+				return errdefs.InvalidParameter(errors.Errorf("failed to set custom headers from %s environment variable: value contains a key=value pair with an empty key", envOverrideHTTPHeaders))
```
krissetto, line 194: could it make sense to print out the wrong key-value in this error message like we do below, to help the user in spotting the bad part of the string? I'm imagining something similar to:  
 
`failed to set custom headers from DOCKER_CUSTOM_HEADERS environment variable: value contains a key=value pair with an empty key (' =somevalue')`  

where the error shown also contains the potential whitespace (so k before trimming the space)

cli/command/cli_options.go
```diff
@@ -108,3 +113,107 @@ func WithAPIClient(c client.APIClient) CLIOption {
...
+			if !hasValue {
+				return errdefs.InvalidParameter(errors.Errorf(`failed to set custom headers from %s environment variable: missing "=" in key=value pair %q`, envOverrideHTTPHeaders, kv))
+			}
+
+			env[http.CanonicalHeaderKey(k)] = v
+		}
+
+		if len(env) == 0 {
+			// We should probably not hit this case, as we don't skip values
+			// (only return errors), but if we don't want to discard existing
```
krissetto, line 210: nit: extra `if` this comment?

krissetto (approved):
overall LGTM, some minor comments

docs/reference/commandline/docker.md
```diff
@@ -118,13 +118,14 @@ The following list of environment variables are supported by the `docker` comman
...
 | `DOCKER_CONTENT_TRUST_SERVER` | The URL of the Notary server to use. Defaults to the same URL as the registry.                                                                                                                                                                                    |
 | `DOCKER_CONTENT_TRUST`        | When set Docker uses notary to sign and verify images. Equates to `--disable-content-trust=false` for build, create, pull, push, run.                                                                                                                             |
 | `DOCKER_CONTEXT`              | Name of the `docker context` to use (overrides `DOCKER_HOST` env var and default context set with `docker context use`)                                                                                                                                           |
+| `DOCKER_CUSTOM_HEADERS`       | (Experimental) Configure [custom HTTP headers](#custom-http-headers) to be sent by the client. Headers must be provided as a comma-separated list of `name=value` pairs. This is the equivalent to the `HttpHeaders` field in the configuration file.             |
```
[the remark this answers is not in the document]
thaJeztah: Yeah, I tried to shorten the description here, otherwise the whole table would be re-formatted 😂 

I think we may need to do another pass at docs in a follow-up; I'm keeping this as-is for now, and we can do a follow-up to improve things if that's ok for you.

cli/command/cli_options.go
thaJeztah: ah, yes, probably works; let me look

cli/command/cli_options.go
thaJeztah: Heh, yeah, I keep making that mistake. Changed to `environment-variable` which seems the most common use, but we should make a pass at looking through our use for consistency

cli/command/cli_options.go
thaJeztah: Ah! fixed

cli/command/cli_options.go
thaJeztah: Added; I decided to change the error-messages to use `'%s'` (not `%q`) to prevent double-escaped quotes in the output

cli/command/cli_options.go
thaJeztah: fixed!

PR #5111, opened 2024-06-04
cli/command: use shallower interface for completions

thaJeztah:
The completion functions only need the API-client, and not all of the CLI. However, passing the API-client as argument would mean that the API-client is initialized early, which may not be what we want, so instead, defining an APIClientProvider interface to preserve the behavior of initializing when needed only.

While updating, also simplify stack.format to only require an io.Writer.

### DIFF
--- a/cli/command/completion/functions.go
+++ b/cli/command/completion/functions.go
@@ -3,23 +3,33 @@ package completion
 import (
 	"os"
 
-	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/formatter"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/image"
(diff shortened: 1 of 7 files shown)


cli/command/container/create.go
```diff
@@ -60,7 +60,7 @@ func NewCreateCommand(dockerCli command.Cli) *cobra.Command {
 		Annotations: map[string]string{
 			"aliases": "docker container create, docker create",
 		},
-		ValidArgsFunction: completion.ImageNames(dockerCli),
+		ValidArgsFunction: completion.ImageNames(dockerCli.Client()),
```
thaJeztah, line 63: 🤦 forgot to revert this line; fixing

vvoland (approved):

Benehiko (approved):

thaJeztah:
Thx! Let me bring this one in 👍

PR #5126, opened 2024-06-06
network create: make --ipv6 optional

akerouanton:
**- What I did**

The API field `EnableIPv6` was marked as optional in our Swagger docs, and its default value in the Go client came from that field being a bool, thus defaulting to its zero value. That's not the case anymore.

This field is now a `*bool` as to let daemon's config define the default value. IPv6 can still be enabled / disabled by explicitly specifying the `--ipv6` flag when doing `docker network create`.

**- How to verify it**

Start a daemon from moby/moby master branch:

```console
$ ./build/docker-darwin-arm64 network create testnet
$ docker network inspect --format='{{ (index .IPAM.Config) }}' testnet
[{172.20.0.0/16  172.20.0.1 map[]}]

$ ./build/docker-darwin-arm64 network create --ipv6 testnetv6
$ docker network inspect --format='{{ (index .IPAM.Config) }}' testnetv6
[{172.19.0.0/16  172.19.0.1 map[]} {fdbc:8ec0:bb64::/64  fdbc:8ec0:bb64::1/64 map[]}]
```

Then restart the daemon with `--default-network-opt=bridge=com.docker.network.enable_ipv6=true`:

```console
$ ./build/docker-darwin-arm64 network create testnetv6
$ docker network inspect --format='{{ (index .IPAM.Config) }}' testnetv6
[{172.20.0.0/16  172.20.0.1 map[]} {fd8c:5815:be1a::/64  fd8c:5815:be1a::1/64 map[]}]

$ ./build/docker-darwin-arm64 network create --ipv6=false testnet
$ docker network inspect --format='{{ (index .IPAM.Config) }}' testnet
[{172.19.0.0/16  172.19.0.1 map[]}]
```

### DIFF
--- a/cli/command/network/create.go
+++ b/cli/command/network/create.go
@@ -23,7 +23,7 @@ type createOptions struct {
 	driverOpts opts.MapOpts
 	labels     opts.ListOpts
 	internal   bool
-	ipv6       bool
+	ipv6       *bool
 	attachable bool
 	ingress    bool
 	configOnly bool
@@ -38,6 +38,7 @@ type createOptions struct {
 }
 
 func newCreateCommand(dockerCli command.Cli) *cobra.Command {
+	var ipv6 bool
 	options := createOptions{
 		driverOpts: *opts.NewMapOpts(nil, nil),
 		labels:     opts.NewListOpts(opts.ValidateLabel),
@@ -51,6 +52,11 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 		Args:  cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.name = args[0]
+
+			if cmd.Flag("ipv6").Changed {
+				options.ipv6 = &ipv6
+			}
+
 			return runCreate(cmd.Context(), dockerCli, options)
 		},
 		ValidArgsFunction: completion.NoComplete,
@@ -61,7 +67,7 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	flags.VarP(&options.driverOpts, "opt", "o", "Set driver specific options")
 	flags.Var(&options.labels, "label", "Set metadata on a network")
 	flags.BoolVar(&options.internal, "internal", false, "Restrict external access to the network")
-	flags.BoolVar(&options.ipv6, "ipv6", false, "Enable IPv6 networking")
+	flags.BoolVar(&ipv6, "ipv6", false, "Enable or disable IPv6 networking")
 	flags.BoolVar(&options.attachable, "attachable", false, "Enable manual container attachment")
 	flags.SetAnnotation("attachable", "version", []string{"1.25"})
 	flags.BoolVar(&options.ingress, "ingress", false, "Create swarm routing-mesh network")
@@ -107,7 +113,7 @@ func runCreate(ctx context.Context, dockerCli command.Cli, options createOptions
 			Options: options.ipamOpt.GetAll(),
 		},
 		Internal:   options.internal,
-		EnableIPv6: &options.ipv6,
+		EnableIPv6: options.ipv6,
 		Attachable: options.attachable,
 		Ingress:    options.ingress,
 		Scope:      options.scope,
--- a/cli/command/network/create.go
+++ b/cli/command/network/create.go
@@ -107,7 +107,7 @@ func runCreate(ctx context.Context, dockerCli command.Cli, options createOptions
 			Options: options.ipamOpt.GetAll(),
 		},
 		Internal:   options.internal,
-		EnableIPv6: options.ipv6,
+		EnableIPv6: &options.ipv6,
 		Attachable: options.attachable,
 		Ingress:    options.ingress,
 		Scope:      options.scope,
--- a/docs/reference/commandline/network_create.md
+++ b/docs/reference/commandline/network_create.md
@@ -18,7 +18,7 @@ Create a network
 | `--ip-range`              | `stringSlice` |           | Allocate container ip from a sub-range                  |
 | `--ipam-driver`           | `string`      | `default` | IP Address Management Driver                            |
 | `--ipam-opt`              | `map`         | `map[]`   | Set IPAM driver specific options                        |
-| `--ipv6`                  |               |           | Enable IPv6 networking                                  |
+| `--ipv6`                  |               |           | Enable or disable IPv6 networking                       |
 | `--label`                 | `list`        |           | Set metadata on a network                               |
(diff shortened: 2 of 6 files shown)


cli/command/network/create.go
```diff
@@ -87,34 +95,38 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
...
+	createOpts := types.NetworkCreate{
 		Driver:  options.driver,
 		Options: options.driverOpts.GetAll(),
 		IPAM: &network.IPAM{
 			Driver:  options.ipamDriver,
-			Config:  ipamCfg,
 			Options: options.ipamOpt.GetAll(),
 		},
 		Internal:   options.internal,
-		EnableIPv6: options.ipv6,
```
thaJeztah, line 110: Given that this is set to `nil` if the flag didn't change; is there a need to do the check below?

cli/command/network/create.go
```diff
@@ -87,34 +95,38 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
...
-	if err != nil {
-		return err
-	}
-
-	var configFrom *network.ConfigReference
-	if options.configFrom != "" {
-		configFrom = &network.ConfigReference{
-			Network: options.configFrom,
-		}
-	}
```
thaJeztah, line 100: Curious if there was a need for this refactor

cli/command/network/create.go
```diff
@@ -61,7 +69,7 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	flags.VarP(&options.driverOpts, "opt", "o", "Set driver specific options")
 	flags.Var(&options.labels, "label", "Set metadata on a network")
 	flags.BoolVar(&options.internal, "internal", false, "Restrict external access to the network")
-	flags.BoolVar(&options.ipv6, "ipv6", false, "Enable IPv6 networking")
+	flags.BoolVar(options.ipv6, "ipv6", false, "Enable IPv6 networking")
```
thaJeztah, line 72: LOL perhaps we need to create our own "optional bool" so that we could have `nil` as default. 🤔 

Perhaps the flag-description should be updated though; something like "auto" (or "determined by daemon"?);

```suggestion
	flags.BoolVar(options.ipv6, "ipv6", false, "Enable or disable IPv6 networking (default is auto)")
```

or should we consider the default to be "true"?

```suggestion
	flags.BoolVar(options.ipv6, "ipv6", true, "Enable or disable IPv6 networking (default is enabled, if supported)")
```

cli/command/network/create.go
```diff
@@ -51,6 +53,12 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 		Args:  cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.name = args[0]
+
+			f := cmd.Flag("ipv6")
+			if !f.Changed {
```
thaJeztah, line 58:
```suggestion
			if !cmd.Flag("ipv6").Changed {
```

or (not sure which one is more performant, but probably won't matter much)

```suggestion
			if !cmd.Flags().Changed("ipv6") {
```

cli/command/network/create.go
```diff
@@ -87,34 +95,38 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
...
+
+	if options.configFrom != "" {
+		createOpts.ConfigFrom = &network.ConfigReference{
+			Network: options.configFrom,
+		}
+	}
+
+	if options.ipv6 != nil {
+		createOpts.EnableIPv6 = options.ipv6
+	}
```
thaJeztah, line 127: See above; curious why you refactored all this 🤔

cli/command/network/create.go
thaJeztah: Actually; would the reverse work? I see you have the `var ipv6 bool` added; that one could be set on the flag;



And this code could then do;

```go
if cmd.Flag("ipv6").Changed {
	options.ipv6 = &ipv6
}
```

thaJeztah (approved):
LGTM

thaJeztah:
The only thing I'm wondering is if we need to add more details about what the default is, but we can do in a follow-up.

akerouanton:
@thaJeztah Sorry for the 1st version of this patch, it was crap. I should stop multi-tasking sometimes 😓 

> The only thing I'm wondering is if we need to add more details about what the default is, but we can do in a follow-up.

I think we need to update the documentation to explain how to set the default daemon-wide, how to disable ipv6 for a specific network, etc... but that's pretty much it. Do you have anything else in mind?

FWIW, we discussed docs updates a bit with @robmry and we decided to wait for at least the rc1 to start working on that -- we'd like to get every PR ready first.

cli/command/network/create.go
akerouanton: I chose `cmd.Flag("ipv6").Changed` instead of `cmd.Flags().Changed("ipv6")` as the former will trigger a panic if the flag is renamed / removed, whereas the latter will silently fail (ie. return `false`).

cli/command/network/create.go
akerouanton: > or should we consider the default to be "true"?

I think we shouldn't do that, at least for now. One reason is the discussion we had regarding ip6tables. Another reason, IPv6 isn't supported by all drivers currently (eg. overlay).

If we want to enable IPv6 by default, I think it'd need to be done at the API level, in a new API version.

docs/reference/commandline/network_create.md
```diff
@@ -18,7 +18,7 @@ Create a network
 | `--ip-range`              | `stringSlice` |           | Allocate container ip from a sub-range                  |
 | `--ipam-driver`           | `string`      | `default` | IP Address Management Driver                            |
 | `--ipam-opt`              | `map`         | `map[]`   | Set IPAM driver specific options                        |
-| `--ipv6`                  |               |           | Enable IPv6 networking                                  |
+| `--ipv6`                  |               |           | Enable or disable IPv6 networking                       |
```
thaJeztah, line 21: Actually, this may have been a wrong suggestion; I mixed up things, and thought we changed the default, but the default is still the same, so adding the `or disable` may be confusing (at least currently).

I'll open a PR to revert this change.

thaJeztah: Oh, actually we did. Nevermind, I'm confused. Ignore me 🙈

robmry: I think it's ok ... with `--default-network-opt=bridge=com.docker.network.enable_ipv6=true`, `network create --ipv6=false` would be needed to disable it for that network.

thaJeztah: Yes, it's indeed the combination of things that tripped me up for a bit. Let's keep it indeed, but we probably need more docs around this 😂

PR #5128, opened 2024-06-07
cli/command/network: add minimal test for --ipv6 option

thaJeztah:
- follow-up to https://github.com/docker/cli/pull/5126

Add a minimal test to verify values are handled correctly.


akerouanton (approved):
Thanks!

PR #5139, opened 2024-06-11
gha: switch to Ubuntu 24.04

thaJeztah:

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -16,7 +16,7 @@ on:
 
 jobs:
   e2e:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
       matrix:
@@ -40,8 +40,15 @@ jobs:
       -
         name: Update daemon.json
         run: |
-          sudo jq '.experimental = true' < /etc/docker/daemon.json > /tmp/docker.json
-          sudo mv /tmp/docker.json /etc/docker/daemon.json
+          if [ ! -f /etc/docker/daemon.json ]; then
+            # ubuntu 24.04 runners no longer have a default daemon.json present
+            sudo mkdir -p /etc/docker/
+            echo '{"experimental": true}' | sudo tee /etc/docker/daemon.json
+          else
+            # but if there is one; let's patch it to keep other options that may be set.
+            sudo jq '.experimental = true' < /etc/docker/daemon.json > /tmp/docker.json
+            sudo mv /tmp/docker.json /etc/docker/daemon.json
+          fi
           sudo cat /etc/docker/daemon.json
           sudo service docker restart
           docker version
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   prepare:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     outputs:
       matrix: ${{ steps.platforms.outputs.matrix }}
     steps:
@@ -37,7 +37,7 @@ jobs:
           echo ${{ steps.platforms.outputs.matrix }}
 
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     needs:
       - prepare
     strategy:
@@ -90,7 +90,7 @@ jobs:
           if-no-files-found: error
 
   bin-image:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     if: ${{ github.event_name != 'pull_request' && github.repository == 'docker/cli' }}
     steps:
       -
@@ -134,7 +134,7 @@ jobs:
             *.cache-to=type=gha,scope=bin-image,mode=max
 
   prepare-plugins:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     outputs:
       matrix: ${{ steps.platforms.outputs.matrix }}
     steps:
@@ -152,7 +152,7 @@ jobs:
           echo ${{ steps.platforms.outputs.matrix }}
 
   plugins:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     needs:
       - prepare-plugins
     strategy:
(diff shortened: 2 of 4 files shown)


thaJeztah:
LOL that .. didn't go well; is `docker` not installed by default on 24.04 runners?

```
Run sudo jq '.experimental = true' < /etc/docker/daemon.json > /tmp/docker.json
  sudo jq '.experimental = true' < /etc/docker/daemon.json > /tmp/docker.json
  sudo mv /tmp/docker.json /etc/docker/daemon.json
  sudo cat /etc/docker/daemon.json
  sudo service docker restart
  docker version
  docker info
  shell: /usr/bin/bash -e {0}
/home/runner/work/_temp/5a389600-a748-4fcd-a8a9-ac63e71c765e.sh: line 1: /etc/docker/daemon.json: No such file or directory
Error: Process completed with exit code 1.
```

thaJeztah:
Hm... other checks use docker as well, so perhaps it's just the `daemon.json` that's missing

.github/workflows/e2e.yml
```diff
@@ -40,10 +40,10 @@ jobs:
...
         name: Update daemon.json
         run: |
-          sudo jq '.experimental = true' < /etc/docker/daemon.json > /tmp/docker.json
-          sudo mv /tmp/docker.json /etc/docker/daemon.json
-          sudo cat /etc/docker/daemon.json
-          sudo service docker restart
+          # sudo jq '.experimental = true' < /etc/docker/daemon.json > /tmp/docker.json
+          # sudo mv /tmp/docker.json /etc/docker/daemon.json
+          # sudo cat /etc/docker/daemon.json
+          # sudo service docker restart
```
thaJeztah, line 46: I notice that this is executed _both_ for `experimental` and `non-experimental` - curious now if that's wrong, or if we just enabled experimental to allow tests that use it to run.

Not critical, but I guess we could skip this step to not having to restart the daemon for the non-experimental ones.

thaJeztah:
Looks like it may be just the `daemon.json` that's missing; commenting-out that part looks like it works

.github/workflows/e2e.yml
```diff
@@ -40,8 +40,15 @@ jobs:
...
-          sudo mv /tmp/docker.json /etc/docker/daemon.json
+          if [ ! -f /tmp/foo.txt ]; then
+            # ubuntu 24.04 runners no longer have a default daemon.json present
+            sudo mkdir -p /etc/docker/
+            echo '{"experimental": true}' | sudo tee /etc/docker/daemon.json
+          else
+            # but if there is one; let's patch it to keep other options that may be set.
+            sudo jq '.experimental = true' < /etc/docker/daemon.json > /tmp/docker.json
+            sudo mv /tmp/docker.json /etc/docker/daemon.json
+          fi
```
thaJeztah, line 51: So... hm... it looks like CI was green even if we didn't enable experimental.

I'm wondering now if there's any (e2e) tests that _require_ experimental 🤔 the only "SKIP" I found was in `docker build` to test the `--squash` flag, but I _think_ we no longer gate that option by "experimental" (i.e., experimental only in wording?) https://github.com/docker/cli/blob/591bd174241555b26fb5e92de5d7a9da094523c1/e2e/image/build_test.go#L128

thaJeztah: Oh; never mind; we do: at least on the _client_ side, but need to check what's on the daemon side https://github.com/docker/cli/blob/591bd174241555b26fb5e92de5d7a9da094523c1/cli/command/image/build.go#L151-L153

thaJeztah: OK; still is gated on the daemon side https://github.com/moby/moby/blob/9d9488468fe2e08394a8b2fb8ad7fcfba34997f1/api/server/router/build/build_routes.go#L260-L262


However, if that's the _ONLY_ test that differs, perhaps we should combine things; we can always enable experimental on the daemon (we already do).

thaJeztah:
This should be ready for review @krissetto @crazy-max 


I opened a separate ticket to look if we indeed still need the `experimental` checks, or if they're really redundant;

- https://github.com/docker/cli/issues/5140

.github/workflows/e2e.yml
```diff
@@ -40,8 +40,15 @@ jobs:
       -
         name: Update daemon.json
         run: |
-          sudo jq '.experimental = true' < /etc/docker/daemon.json > /tmp/docker.json
-          sudo mv /tmp/docker.json /etc/docker/daemon.json
+          if [ ! -f /tmp/foo.txt ]; then
```
crazy-max, line 43: ?

thaJeztah: LOOOOOL 🙈 🙈🙈 copy/pasta because I always forget the correct one, so I tried first

thaJeztah: Fixed now; only now wondering; would it need `sudo` to check for the file? 🤔 I guess the path would be _readable_ at least for non-root, so probably it's fine as-is.

.github/workflows/e2e.yml
```diff
@@ -40,8 +40,15 @@ jobs:
       -
         name: Update daemon.json
         run: |
-          sudo jq '.experimental = true' < /etc/docker/daemon.json > /tmp/docker.json
-          sudo mv /tmp/docker.json /etc/docker/daemon.json
+          if [ ! -f /etc/docker/daemon.json ]; then
+            # ubuntu 24.04 runners no longer have a default daemon.json present
```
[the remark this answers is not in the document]
thaJeztah: Yeah, that was my thinking; if they decide to set some options, we wouldn't immediately break.

But the `cat /etc/docker/daemon.json` below should show us what's in there, in case their changes would cause a change in behavior and we wanted to see "what changed?"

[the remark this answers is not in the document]
krissetto: Yeah, i noticed that..I guess that can work, no hard feelings here anyway :)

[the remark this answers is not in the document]
thaJeztah: Thanks! Yeah, I guess it's one of those cases where there's pros/cons to either approach.


Let me bring this one in; thanks for reviewing!

PR #5145, opened 2024-06-11
Add OomScoreAdj to "docker service create" and "docker compose"

psaintlaurent:
- [x] depends on https://github.com/moby/moby/pull/47950
- [x] depends on https://github.com/docker/cli/pull/5169
- [x] depends on https://github.com/docker/cli/pull/5170

**- What I did**
I added oom-score-adj flag to docker service create and docker service update

**- How I did it**
By modifying container service to accept the option as well as adding a flag to the CLI

**- How to test it**
Compile this PR and copy the CLI binary to a running moby dev shell using  https://github.com/moby/moby/pull/47950

```markdown changelog

Add OOMScoreAdj to docker service create and docker compose
```

### DIFF
--- a/cli/command/service/create.go
+++ b/cli/command/service/create.go
@@ -68,6 +68,8 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation(flagSysCtl, "version", []string{"1.40"})
 	flags.Var(&opts.ulimits, flagUlimit, "Ulimit options")
 	flags.SetAnnotation(flagUlimit, "version", []string{"1.41"})
+	flags.Int64Var(&opts.oomScoreAdj, flagOomScoreAdj, 0, "Tune host's OOM preferences (-1000 to 1000) ")
+	flags.SetAnnotation(flagOomScoreAdj, "version", []string{"1.46"})
 
 	flags.Var(cliopts.NewListOptsRef(&opts.resources.resGenericResources, ValidateSingleGenericResource), "generic-resource", "User defined resources")
 	flags.SetAnnotation(flagHostAdd, "version", []string{"1.32"})
--- a/cli/command/service/opts.go
+++ b/cli/command/service/opts.go
@@ -529,6 +529,7 @@ type serviceOptions struct {
 	capAdd          opts.ListOpts
 	capDrop         opts.ListOpts
 	ulimits         opts.UlimitOpt
+	oomScoreAdj     int64
 
 	resources resourceOptions
 	stopGrace opts.DurationOpt
@@ -747,6 +748,7 @@ func (options *serviceOptions) ToService(ctx context.Context, apiClient client.N
 				CapabilityAdd:   capAdd,
 				CapabilityDrop:  capDrop,
 				Ulimits:         options.ulimits.GetList(),
+				OomScoreAdj:     options.oomScoreAdj,
 			},
 			Networks:      networks,
 			Resources:     resources,
@@ -1043,6 +1045,7 @@ const (
 	flagUlimit                  = "ulimit"
 	flagUlimitAdd               = "ulimit-add"
 	flagUlimitRemove            = "ulimit-rm"
+	flagOomScoreAdj             = "oom-score-adj"
 )
 
 func validateAPIVersion(c swarm.ServiceSpec, serverAPIVersion string) error {
--- a/cli/command/service/update.go
+++ b/cli/command/service/update.go
@@ -108,6 +108,8 @@ func newUpdateCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation(flagUlimitAdd, "version", []string{"1.41"})
 	flags.Var(newListOptsVar(), flagUlimitRemove, "Remove a ulimit option")
 	flags.SetAnnotation(flagUlimitRemove, "version", []string{"1.41"})
+	flags.Int64Var(&options.oomScoreAdj, flagOomScoreAdj, 0, "Tune host's OOM preferences (-1000 to 1000) ")
+	flags.SetAnnotation(flagOomScoreAdj, "version", []string{"1.46"})
 
 	// Add needs parsing, Remove only needs the key
 	flags.Var(newListOptsVar(), flagGenericResourcesRemove, "Remove a Generic resource")
@@ -365,6 +367,10 @@ func updateService(ctx context.Context, apiClient client.NetworkAPIClient, flags
 		taskResources().Reservations = spec.TaskTemplate.Resources.Reservations
 		updateInt64Value(flagReserveCPU, &task.Resources.Reservations.NanoCPUs)
 		updateInt64Value(flagReserveMemory, &task.Resources.Reservations.MemoryBytes)
+	}
+
+	if anyChanged(flags, flagOomScoreAdj) {
+		updateInt64(flagOomScoreAdj, &task.ContainerSpec.OomScoreAdj)
 	}
 
 	if err := addGenericResources(flags, task); err != nil {
--- a/cli/compose/convert/service.go
+++ b/cli/compose/convert/service.go
@@ -148,6 +148,7 @@ func Service(
 				CapabilityAdd:   capAdd,
 				CapabilityDrop:  capDrop,
 				Ulimits:         convertUlimits(service.Ulimits),
+				OomScoreAdj:     service.OomScoreAdj,
 			},
 			LogDriver:     logDriver,
 			Resources:     resources,
--- a/cli/compose/loader/interpolate.go
+++ b/cli/compose/loader/interpolate.go
@@ -29,6 +29,7 @@ var interpolateTypeCastMapping = map[interp.Path]interp.Cast{
 	servicePath("ulimits", interp.PathMatchAll, "hard"):              toInt,
 	servicePath("ulimits", interp.PathMatchAll, "soft"):              toInt,
 	servicePath("privileged"):                                        toBoolean,
+	servicePath("oom_score_adj"):                                     toInt,
 	servicePath("read_only"):                                         toBoolean,
 	servicePath("stdin_open"):                                        toBoolean,
 	servicePath("tty"):                                               toBoolean,
--- a/cli/compose/schema/data/config_schema_v3.13.json
+++ b/cli/compose/schema/data/config_schema_v3.13.json
@@ -287,6 +287,7 @@
             }
           }
         },
+        "oom_score_adj": {"type": "integer"},
         "user": {"type": "string"},
         "userns_mode": {"type": "string"},
         "volumes": {
--- a/cli/compose/types/types.go
+++ b/cli/compose/types/types.go
@@ -207,6 +207,7 @@ type ServiceConfig struct {
 	Tty             bool                             `mapstructure:"tty" yaml:"tty,omitempty" json:"tty,omitempty"`
 	Ulimits         map[string]*UlimitsConfig        `yaml:",omitempty" json:"ulimits,omitempty"`
 	User            string                           `yaml:",omitempty" json:"user,omitempty"`
+	OomScoreAdj     int64                            `yaml:",omitempty" json:"oom_score_adj,omitempty"`
 	UserNSMode      string                           `mapstructure:"userns_mode" yaml:"userns_mode,omitempty" json:"userns_mode,omitempty"`
 	Volumes         []ServiceVolumeConfig            `yaml:",omitempty" json:"volumes,omitempty"`
 	WorkingDir      string                           `mapstructure:"working_dir" yaml:"working_dir,omitempty" json:"working_dir,omitempty"`
--- a/docs/reference/commandline/service_create.md
+++ b/docs/reference/commandline/service_create.md
@@ -46,6 +46,7 @@ Create a new service
 | [`--network`](#network)                             | `network`         |              | Network attachments                                                                                 |
 | `--no-healthcheck`                                  | `bool`            |              | Disable any container-specified HEALTHCHECK                                                         |
 | `--no-resolve-image`                                | `bool`            |              | Do not query the registry to resolve image digest and supported platforms                           |
+| `--oom-score-adj`                                   | `int64`           | `0`          | Tune host's OOM preferences (-1000 to 1000)                                                         |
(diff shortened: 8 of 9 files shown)


cli/compose/schema/data/config_schema_v3.14.json
```diff
@@ -0,0 +1,680 @@
+{
```
thaJeztah, line 1: Schema 3.13 was added in https://github.com/docker/cli/pull/5125, which has not been released yet, and will be for the v27.0 release, so we don't  need to add a new schema and can update 3.13 instead.

cli/command/container/opts.go
```diff
@@ -660,7 +660,7 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 	hostConfig := &container.HostConfig{
 		Binds:           binds,
 		ContainerIDFile: copts.containerIDFile,
-		OomScoreAdj:     copts.oomScoreAdj,
+		OomScoreAdj:     int(copts.oomScoreAdj),
```
thaJeztah, line 663: Both sides should already be an `int` I think, so this change should probably be removed.

opts/opts.go
```diff
@@ -515,3 +516,26 @@ func (m *MemSwapBytes) UnmarshalJSON(s []byte) error {
...
+func (o *OomScoreAdj) Set(value string) error {
+
+	var conv int64
+	conv, err := strconv.ParseInt(value, 10, 64)
+	if err != nil || conv < -1000 || conv > 1000 {
+		return err
+	}
+	*o = OomScoreAdj(conv)
+	return nil
+}
```
thaJeztah, line 541: Bit on the fence here if we should have a special option-type for this. It does allow for validation on the client-side, but also makes assumption about what the daemon-side allows (given; it's unlikely to change); for `docker run` it looks like we leave the validation to the API, and using a regular `IntVar` option; https://github.com/docker/cli/blob/64206aef767d23420803badbd205d1c5a661f097/cli/command/container/opts.go#L306

opts/opts.go
```diff
@@ -515,3 +516,26 @@ func (m *MemSwapBytes) UnmarshalJSON(s []byte) error {
...
+	return val
+}
+
+func (o *OomScoreAdj) Set(value string) error {
+
+	var conv int64
+	conv, err := strconv.ParseInt(value, 10, 64)
+	if err != nil || conv < -1000 || conv > 1000 {
+		return err
+	}
```
corhere, line 538: In the case where `err == nil` but `conv` is out of range, [the `Set()` call will silently be a no-op.](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) Either generate a client-side error, or (preferably) do what @thaJeztah recommends and leave range validation to the API / daemon.

thaJeztah: Yes, I think we can keep it simple for now, and do the same as we do for `docker run`, which is a regular `IntVar`.

I tend to reduce client-side validation to a minimum to keep the CLI agnostic and make the daemon "source of truth". That's not all set in stone, and there are some grey areas where it's worth the trade-ff but I'd like to avoid the CLI making assumptions that may not be correct.

cli/command/service/update.go
```diff
@@ -368,6 +371,11 @@ func updateService(ctx context.Context, apiClient client.NetworkAPIClient, flags
 		updateInt64Value(flagReserveMemory, &task.Resources.Reservations.MemoryBytes)
 	}
 
+	if anyChanged(flags, flagOomScoreAdj) {
+
```
thaJeztah, line 375: Can you remove whitespace here as well?

cli/command/service/update.go
```diff
@@ -132,6 +133,7 @@ func runUpdate(ctx context.Context, dockerCli command.Cli, flags *pflag.FlagSet,
 	apiClient := dockerCli.Client()
 
 	service, _, err := apiClient.ServiceInspectWithRaw(ctx, serviceID, types.ServiceInspectOptions{})
+
```
thaJeztah, line 136: Same here (whitespace)

cli/command/service/create.go
```diff
@@ -68,6 +68,7 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation(flagSysCtl, "version", []string{"1.40"})
 	flags.Var(&opts.ulimits, flagUlimit, "Ulimit options")
 	flags.SetAnnotation(flagUlimit, "version", []string{"1.41"})
+	flags.Var(&opts.oomScoreAdj, flagOomScoreAdj, "oom score adjustment (-1000 to 1000)")
```
thaJeztah, line 71: See other comments; this could probably just be an `flags.IntVar`

opts/opts.go
```diff
@@ -515,3 +516,21 @@ func (m *MemSwapBytes) UnmarshalJSON(s []byte) error {
...
+
+func (o *OomScoreAdj) String() string {
+	return strconv.FormatInt(int64(*o), 10)
+}
+
+func (o *OomScoreAdj) Set(value string) error {
+
+	var conv int64
+	conv, _ = strconv.ParseInt(value, 10, 64)
+	*o = OomScoreAdj(conv)
```
corhere, line 534: If `value` is not parseable as a number, the option gets silently set to 0? That doesn't seem right.

cli/command/service/opts.go
```diff
@@ -529,6 +529,7 @@ type serviceOptions struct {
 	capAdd          opts.ListOpts
 	capDrop         opts.ListOpts
 	ulimits         opts.UlimitOpt
+	oomScoreAdj     opts.OomScoreAdj
```
corhere, line 532:
```suggestion
	oomScoreAdj     Uint64Opt
```

opts/opts.go
```diff
@@ -515,3 +516,21 @@ func (m *MemSwapBytes) UnmarshalJSON(s []byte) error {
...
+	return nil
+}
```
corhere, line 536: There already exists a generic `Uint64Opt` type in the `cli/command/service` package. No need to define a new one.

cli/command/service/opts.go
psaintlaurent: That's an unisgned 64 bit integer for a value that ranges from -1000 to 1000

opts/opts.go
psaintlaurent: I've run into silent defaults over mltiple code bases, if we want different/better behavior I didn't see anything documented.

Example...
https://github.com/docker/cli/blob/e174fe5597110959abd67e696802e239c6095931/opts/opts.go#L454

opts/opts.go
psaintlaurent: That's an Unsigned 64 bit integer

opts/opts.go
corhere: > I've run into silent defaults over multiple code bases

What do "silent defaults" mean to you? Could you please elaborate?

> Example...

You have linked to the definition of the `String()` method of a [`pflag.Value`](https://pkg.go.dev/github.com/spf13/pflag#Value) implementation, which contains a documented special case for how its [zero value](https://go.dev/doc/effective_go#allocation_new:~:text=Since%20the%20memory,an%20unlocked%20mutex.) is formatted for user display.

https://github.com/docker/cli/blob/e174fe5597110959abd67e696802e239c6095931/opts/opts.go#L448-L450

I'm not following how this is relevant to the topic being discussed. Could you please explain?

cli/command/service/opts.go
corhere: My bad. You're correct; `Uint64Opt` would not be appropriate. No custom `Value` type is needed at all, since it's [just a signed int.](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.IntVar)
```suggestion
	oomScoreAdj     int
```

opts/opts.go
corhere: Right, my mistake. You don't need any custom type at all! https://pkg.go.dev/github.com/spf13/pflag#FlagSet.IntVar

opts/opts.go
psaintlaurent: I consider silent defaults to be any configuration options that are:

**1.)**  set to values that have to be infered 
**2.)** not explicitly documented outside of source code

corhere: I'm still lost. Why are we on this tangent about silent defaults? What do silent defaults have to do with a `(pflag.Value).Set()` method?

cli/command/service/opts.go
psaintlaurent: It was a custom type originally because I had a range constraint in place.  It should be fine, I need to fix the pipeline issues.

opts/opts.go
psaintlaurent: There is no tangent, I'm just answering your question.

cli/command/service/opts.go
corhere: > It was a custom type originally because I had a range constraint in place.

The history of how this PR has evolved is irrelevant. The custom type is not needed in the current iteration, therefore merging this PR with the custom type in place increases the ongoing maintenance cost of this project without any offsetting benefit.

> It should be fine, I need to fix the pipeline issues.

Please stay on topic.

opts/opts.go
corhere: Claiming that you have answered my question is not an answer. Please answer my question:

> What do silent defaults have to do with a `(pflag.Value).Set()` method?

psaintlaurent: I didn't actually reference that, you did.

corhere: `func (o *OomScoreAdj) Set(value string) error`, the body of which is the context for this discussion thread, implements a `Set` method necessary for `*OomScoreAdj` to satisfy the `pflag.Value` interface. In other words, `func (o *OomScoreAdj) Set(value string) error` is a `(pflag.Value).Set()` method.

corhere (approved):

cli/command/service/create.go
```diff
@@ -68,6 +68,7 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation(flagSysCtl, "version", []string{"1.40"})
 	flags.Var(&opts.ulimits, flagUlimit, "Ulimit options")
 	flags.SetAnnotation(flagUlimit, "version", []string{"1.41"})
+	flags.Int64Var(&opts.oomScoreAdj, flagOomScoreAdj, 0, "oom score adjustment (-1000 to 1000)")
```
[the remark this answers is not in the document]
psaintlaurent: I change the message to "the value increases/decreases the probability of the process being killed during an OOM event"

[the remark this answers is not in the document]
Benehiko: could we start the sentence capitalized? Keeping the `(-1000 to 1000)` is also probably a good idea.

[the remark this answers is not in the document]
psaintlaurent: I'll make the change however its going to ruin formatting of [docs/reference/commandline/service_create.md](https://github.com/docker/cli/pull/5145/files#diff-45c74769cf430445415aa6c7497268e40a768dd38d0581cda095e3af2c95ae4a) and
[docs/reference/commandline/service_update.md](https://github.com/docker/cli/pull/5145/files#diff-75a53b65be400b339922bd579aafd3ae3be447cb22bd654e06b8bf94f696bb15)

I was warned about introducing unnecessary whitespace formatting changes previously.

Benehiko (approved):
LGTM

psaintlaurent:
Is there something wrong with the validate/validate-md check?  The diff in my PR doesn't show any of the changes that are reflected in the check.  A git diff locally doesn't reflect any of those changes either.

corhere:
> Is there something wrong with the validate/validate-md check?

No.

https://github.com/docker/cli/blob/2f529a19651763d8fa43c31e47637aae65c0dd68/.github/workflows/validate.yml#L38-L57

The markdown files are generated from the Go sources. Following Go ecosystem convention, generated files are committed to the repo alongside their sources. The `validate/validate-md` check is asserting that the generated files are in sync.

Run `make -f docker.Makefile mddocs` locally and commit the result.

cli/command/service/update.go
```diff
@@ -108,6 +108,7 @@ func newUpdateCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation(flagUlimitAdd, "version", []string{"1.41"})
 	flags.Var(newListOptsVar(), flagUlimitRemove, "Remove a ulimit option")
 	flags.SetAnnotation(flagUlimitRemove, "version", []string{"1.41"})
+	flags.Int64Var(&options.oomScoreAdj, flagOomScoreAdj, 0, "(-1000 to 1000) Increases/Decreases the probability of the process being killed during an OOM event ")
```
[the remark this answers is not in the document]
thaJeztah: And add an annotation for the API version

cli/command/service/update.go
```diff
@@ -367,6 +368,10 @@ func updateService(ctx context.Context, apiClient client.NetworkAPIClient, flags
 		updateInt64Value(flagReserveMemory, &task.Resources.Reservations.MemoryBytes)
 	}
 
+	if anyChanged(flags, flagOomScoreAdj) {
```
thaJeztah, line 371: nit as it's a single flag we check for, this could be `flags.Chanced(flagOomScoreAdj)` (not a blocker though)

cli/command/service/create.go
```diff
@@ -68,6 +68,8 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 	flags.SetAnnotation(flagSysCtl, "version", []string{"1.40"})
 	flags.Var(&opts.ulimits, flagUlimit, "Ulimit options")
 	flags.SetAnnotation(flagUlimit, "version", []string{"1.41"})
+	flags.Int64Var(&opts.oomScoreAdj, flagOomScoreAdj, 0, "Tune host's OOM preferences (-1000 to 1000) ")
+	flags.SetAnnotation(flagOomScoreAdj, "version", []string{"1.41"})
```
corhere, line 72: API v1.41 was added in Moby v20.10. That doesn't seem correct. Doesn't this flag correspond to a brand new API field, which was only added in the Moby v27 Engine API client?

thaJeztah (approved):
LGTM, thanks!

thaJeztah:
- backport: https://github.com/docker/cli/pull/5274

PR #5146, opened 2024-06-11
plugins: cleanup sockets when done

laurazard:
- relates to https://github.com/docker/cli/pull/4963

**- What I did**

Since 509123f935e6fcb2cedf0df7bd8585a346082969, we've been leaking sockets in the filesystem on platforms where abstract sockets aren't supported.

That change relied on Go to cleanup our sockets for us, which Go will happily do as long as we make sure to close the listener, which we weren't previously doing unless to signal the plugin to terminate.

**- How I did it**

This change adds a deferred call to `PluginServer.Close()`, which makes sure we close the plugin server at the end of the plugin execution, so that we never exit without cleaning up.

**- How to verify it**

Execute a plugin command with debug logging enabled (e.g. `docker -D compose convert`) and look for the added debug log. Also check `$TMPDIR` for leaked sockets.

```markdown changelog


```

Screenshot 2024-06-11 at 21 32 13

### DIFF
--- a/cli-plugins/socket/socket.go
+++ b/cli-plugins/socket/socket.go
@@ -9,6 +9,8 @@ import (
 	"os"
 	"runtime"
 	"sync"
+
+	"github.com/sirupsen/logrus"
 )
 
 // EnvKey represents the well-known environment variable used to pass the
@@ -30,6 +32,7 @@ func NewPluginServer(h func(net.Conn)) (*PluginServer, error) {
 	if err != nil {
 		return nil, err
 	}
+	logrus.Trace("Plugin server listening on ", l.Addr())
 
 	if h == nil {
 		h = func(net.Conn) {}
@@ -92,6 +95,7 @@ func (pl *PluginServer) Addr() net.Addr {
 //
 // The error value is that of the underlying [net.Listner.Close] call.
 func (pl *PluginServer) Close() error {
+	logrus.Trace("Closing plugin server")
 	// Close connections first to ensure the connections get io.EOF instead
 	// of a connection reset.
 	pl.closeAllConns()
@@ -106,6 +110,10 @@ func (pl *PluginServer) Close() error {
 func (pl *PluginServer) closeAllConns() {
 	pl.mu.Lock()
 	defer pl.mu.Unlock()
+
+	if pl.closed {
+		return
+	}
 
 	// Prevent new connections from being accepted.
 	pl.closed = true
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -245,6 +245,11 @@ func tryPluginRun(ctx context.Context, dockerCli command.Cli, cmd *cobra.Command
 	if err == nil {
 		plugincmd.Env = append(plugincmd.Env, socket.EnvKey+"="+srv.Addr().String())
 	}
+	defer func() {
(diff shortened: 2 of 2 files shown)


cli-plugins/socket/socket.go
```diff
@@ -107,9 +109,14 @@ func (pl *PluginServer) closeAllConns() {
...
 	defer pl.mu.Unlock()
 
+	if pl.closed {
+		return
+	}
+
 	// Prevent new connections from being accepted.
 	pl.closed = true
 
+	logrus.Debug("Closing all connections")
```
tonistiigi, line 119: I'd prefer not to log this on every plugin invocation with `--debug`

laurazard: Oh, I had this in another place with slightly different wording locally, just forgot to push.

cli-plugins/socket/socket.go
```diff
@@ -92,6 +94,7 @@ func (pl *PluginServer) Addr() net.Addr {
 //
 // The error value is that of the underlying [net.Listner.Close] call.
 func (pl *PluginServer) Close() error {
+	logrus.Debug("Closing plugin server")
```
laurazard, line 97: Would you still rather I remove this line @tonistiigi? I don't really mind either way, but I expect debug logs to be quite verbose.

tonistiigi: Currently, we don't have many debug logs for commands, and for example in buildx where we use `--debug` for extra verbosity for build errors, this can be confusing to users who have no context of "plugin server".

thaJeztah: Yeah the tricky bit with the CLI (unlike the daemon, where it usually ends up in a log file) is that logging in the CLI is.. printing in the user's terminal.

Would be nice if we had a location to store logs, but .. we don't, so we should not use it except for exceptional cases.

We _could_ use "trace" level logging (one level more verbose than debug), but not sure if we pass that through currently (and not sure if containerd logs had it as option)

laurazard: Oh, that works! 
```go
logrus.Trace("Closing plugin server")
```

```
./build/docker -l trace compose version
TRAC[0000] Plugin server listening on /var/folders/zj/khjjyd1n0zjd9xc44815szxc0000gn/T/docker_cli_9e516c003cda3c20ab79b88afc18434b 
Docker Compose version v2.27.1-desktop.1
TRAC[0000] Closing plugin server                                              
```

laurazard: @tonistiigi this work for you?

vvoland (approved):

cli-plugins/socket/socket.go
thaJeztah: I think it should be fine; we can remove in a follow-up if it's a problem (but we need to look a bit at logging overall; I also did a quick attempt to see what it'd could look like if we switched over to containerd/log, as we did on the engine side; https://github.com/docker/cli/pull/5149

thaJeztah:
@laurazard how difficult would it be to write a test-case for this? (could be in a follow-up if we want to get this in)

thaJeztah:
Had a quick chat; let's bring this one in, and do a follow-up for the test and any other changes we may need.

PR #5168, opened 2024-06-18
fix: ctx cancellation on login prompt

Benehiko:
**- What I did**

Implemented a new Prompt function to capture user input called `PromptForInput`. This function works virtually similar to `PromptForConfirmation` and handles context cancellation.

**- How I did it**

Replicate `PromptForConfirmation`.

**- How to verify it**
Run `docker login` and cancel it midway with a SIGINT or SIGTERM or ctrl+c. 

```console
docker login

Username:^C
````

`PromptForInput` also allows for hiding user input - for example when the user enters their password.

The original behavior of `docker login` should still be in-tact with trimming spaces and not showing user input on the password prompt.

```markdown changelog


```

### DIFF
--- a/cli/command/registry/login_test.go
+++ b/cli/command/registry/login_test.go
@@ -6,7 +6,10 @@ import (
 	"errors"
 	"fmt"
 	"testing"
+	"time"
 
+	"github.com/creack/pty"
+	"github.com/docker/cli/cli/command"
 	configtypes "github.com/docker/cli/cli/config/types"
 	"github.com/docker/cli/cli/streams"
 	"github.com/docker/cli/internal/test"
@@ -183,5 +186,43 @@ func TestRunLogin(t *testing.T) {
 			assert.Check(t, credStoreErr)
 			assert.DeepEqual(t, tc.expectedSavedCred, savedCred)
 		})
+	}
+}
+
+func TestLoginTermination(t *testing.T) {
+	p, tty, err := pty.Open()
+	assert.NilError(t, err)
+
+	t.Cleanup(func() {
+		_ = tty.Close()
+		_ = p.Close()
+	})
+
+	cli := test.NewFakeCli(&fakeClient{}, func(fc *test.FakeCli) {
+		fc.SetOut(streams.NewOut(tty))
+		fc.SetIn(streams.NewIn(tty))
+	})
+	tmpFile := fs.NewFile(t, "test-login-termination")
+	defer tmpFile.Remove()
+
+	configFile := cli.ConfigFile()
+	configFile.Filename = tmpFile.Path()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+
+	runErr := make(chan error)
+	go func() {
+		runErr <- runLogin(ctx, cli, loginOptions{})
+	}()
+
+	// Let the prompt get canceled by the context
+	cancel()
+
+	select {
+	case <-time.After(1 * time.Second):
+		t.Fatal("timed out after 1 second. `runLogin` did not return")
+	case err := <-runErr:
+		assert.ErrorIs(t, err, command.ErrPromptTerminated)
 	}
 }
--- a/cli/command/utils.go
+++ b/cli/command/utils.go
@@ -19,6 +19,7 @@ import (
 	"github.com/docker/docker/api/types/versions"
 	"github.com/docker/docker/errdefs"
 	"github.com/moby/sys/sequential"
+	"github.com/moby/term"
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
 )
@@ -75,6 +76,48 @@ func PrettyPrint(i any) string {
 }
 
 var ErrPromptTerminated = errdefs.Cancelled(errors.New("prompt terminated"))
+
+// DisableInputEcho disables input echo on the provided streams.In.
+// This is useful when the user provides sensitive information like passwords.
+// The function returns a restore function that should be called to restore the
+// terminal state.
+func DisableInputEcho(ins *streams.In) (restore func() error, err error) {
+	oldState, err := term.SaveState(ins.FD())
+	if err != nil {
+		return nil, err
+	}
+	restore = func() error {
+		return term.RestoreTerminal(ins.FD(), oldState)
+	}
+	return restore, term.DisableEcho(ins.FD(), oldState)
+}
+
+// PromptForInput requests input from the user.
+//
+// If the user terminates the CLI with SIGINT or SIGTERM while the prompt is
+// active, the prompt will return an empty string ("") with an ErrPromptTerminated error.
+// When the prompt returns an error, the caller should propagate the error up
+// the stack and close the io.Reader used for the prompt which will prevent the
+// background goroutine from blocking indefinitely.
+func PromptForInput(ctx context.Context, in io.Reader, out io.Writer, message string) (string, error) {
+	_, _ = fmt.Fprint(out, message)
+
+	result := make(chan string)
+	go func() {
+		scanner := bufio.NewScanner(in)
+		if scanner.Scan() {
+			result <- strings.TrimSpace(scanner.Text())
+		}
+	}()
+
+	select {
+	case <-ctx.Done():
+		_, _ = fmt.Fprintln(out, "")
+		return "", ErrPromptTerminated
+	case r := <-result:
+		return r, nil
+	}
+}
 
 // PromptForConfirmation requests and checks confirmation from the user.
 // This will display the provided message followed by ' [y/N] '. If the user
--- a/cli/command/registry.go
+++ b/cli/command/registry.go
@@ -1,10 +1,8 @@
 package command
 
 import (
-	"bufio"
 	"context"
 	"fmt"
-	"io"
 	"os"
 	"runtime"
 	"strings"
@@ -18,7 +16,6 @@ import (
 	"github.com/docker/docker/api/types"
 	registrytypes "github.com/docker/docker/api/types/registry"
 	"github.com/docker/docker/registry"
-	"github.com/moby/term"
 	"github.com/pkg/errors"
 )
 
@@ -44,7 +41,7 @@ func RegistryAuthenticationPrivilegedFunc(cli Cli, index *registrytypes.IndexInf
 		default:
 		}
 
-		err = ConfigureAuth(cli, "", "", &authConfig, isDefaultRegistry)
+		err = ConfigureAuth(ctx, cli, "", "", &authConfig, isDefaultRegistry)
 		if err != nil {
 			return "", err
 		}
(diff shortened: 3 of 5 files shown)


cli/command/utils.go
```diff
@@ -76,6 +77,69 @@ func PrettyPrint(i any) string {
...
+	result := make(chan string)
+
+	go func() {
+		reader := bufio.NewReader(ins)
+		line, err := reader.ReadString('\n')
+		if err != nil {
+			result <- ""
+		}
+		result <- strings.TrimSpace(line)
+	}()
```
vvoland, line 132: I think this goroutine will still be running after context is `Done`, and so it will still be reading from the stdin? 🤔

laurazard:
Before this, when receiving a `SIGINT` during the login prompt, the CLI would exit with code `140`
```
 ⭑ docker login      

Username (laura.brehm@docker.com): ^C
☆｡･:*:･ﾟ⭑ laurabrehm ✿ cli [e06bee6999] ⭑ echo $?     
130
```

However, now it's exiting with `0`:
```
⭑ ./build/docker login                 

Username (laura.brehm@docker.com): ^C
☆｡･:*:･ﾟ⭑ laurabrehm ✿ cli [fix-cli-login] ⭑ echo $?                              
0
```

I don't think this is necessarily wrong (on one hand, the user is cancelling the operation) but not sure.

cli/command/utils.go
Benehiko: I'll check quick and write a test

Benehiko:
> Before this, when receiving a `SIGINT` during the login prompt, the CLI would exit with code `140`
> 
> ```
>  ⭑ docker login      
> Log in with your Docker ID or email address to push and pull images from Docker Hub. If you don't have a Docker ID, head over to https://hub.docker.com/ to create one.
> You can log in with your password or a Personal Access Token (PAT). Using a limited-scope PAT grants better security and is required for organizations using SSO. Learn more at https://docs.docker.com/go/access-tokens/
> 
> Username (laura.brehm@docker.com): ^C
> ☆｡･:*:･ﾟ⭑ laurabrehm ✿ cli [e06bee6999] ⭑ echo $?     
> 130
> ```
> 
> However, now it's exiting with `0`:
> 
> ```
> ⭑ ./build/docker login                 
> Log in with your Docker ID or email address to push and pull images from Docker Hub. If you don't have a Docker ID, head over to https://hub.docker.com/ to create one.
> You can log in with your password or a Personal Access Token (PAT). Using a limited-scope PAT grants better security and is required for organizations using SSO. Learn more at https://docs.docker.com/go/access-tokens/
> 
> Username (laura.brehm@docker.com): ^C
> ☆｡･:*:･ﾟ⭑ laurabrehm ✿ cli [fix-cli-login] ⭑ echo $?                              
> 0
> ```
> 
> I don't think this is necessarily wrong (on one hand, the user is cancelling the operation) but not sure.

Error code 130 used to occur on the previous Prompt confirmations that were cancelled as well. I can't remember if this was an error that Go decides or that we did with os.Exit. But see here https://github.com/docker/cli/issues/4850

cli/command/utils.go
Benehiko: fixed it, with `bufio.NewScanner` it should exit the goroutine when the user terminates the CLI (SIGTERM/SIGINT). We do the same with `PromptForConfirmation`

this is also why we return an error so that the caller can gracefully exit and close the input stream.

cli/command/utils.go
```diff
@@ -76,6 +77,67 @@ func PrettyPrint(i any) string {
...
+		_ = term.DisableEcho(ins.FD(), oldState)
+		defer func() {
+			_ = term.RestoreTerminal(ins.FD(), oldState)
+		}()
+	}
+
+	// On Windows, force the use of the regular OS stdin stream.
+	if runtime.GOOS == "windows" {
+		ins = streams.NewIn(os.Stdin)
+	}
```
vvoland, line 121: Why we ignore the passed `ins` on Windows?

Benehiko: I'm not too sure tbh - might be a legacy bug which has been resolved. I can maybe check out if we still need to do this fallback on a windows install.

It's possible we don't need it, but I guess keep it for good measure for the time being? :man_shrugging: 
See this comment: https://github.com/docker/cli/blob/master/cli/command/registry.go#L94-L104

vvoland: https://github.com/docker/cli/blob/b83cf582cd3b932b6fce900a54cf0e1e6f61ac17/cli/command/registry.go#L93-L104

Hmm, so the code you linked is still there and already sets the in for the whole `cli` so we don't need to do that again in `PromptForInput`, because `cli.In()` will already be the overriden stdin.

Benehiko: Yes, although other places might call `PromptForInput` which might not set this. We also set this in `PromptForConfirmation` so keeping it consistent would be best I think.

vvoland: It completely breaks the assumption that the input will be read from the provided `ins` though.
If the `ins` happens to be something different than stdin, it would still read from stdin on Windows.

This will also make the unit tests from this PR fail on Windows because they provide a `io.Pipe` as an stdin.

cli/command/utils.go
```diff
@@ -76,6 +77,67 @@ func PrettyPrint(i any) string {
...
+}
+
+// PromptForInput requests input from the user.
+//
+// If the user terminates the CLI with SIGINT or SIGTERM while the prompt is
+// active, the prompt will return an empty string ("") with an ErrPromptTerminated error.
+// When the prompt returns an error, the caller should propagate the error up
+// the stack and close the io.Reader used for the prompt which will prevent the
+// background goroutine from blocking indefinitely.
+func PromptForInput(ctx context.Context, ins *streams.In, outs *streams.Out, message string, opts ...PromptOptions) (string, error) {
```
thaJeztah, line 99: - Looks like `outs` could be bare `io.Writer` (as we don't use any of the features provided by `streams.Out`
- nit: perhaps use singular for the `ins` and `outs` variables (`in` and `out` ?); it's more common.
- ☝️ I recalled we also had a `Streams` interface, which could be an alternative to use as argument, but it looks like that's in the `cli/command` package, so a bit on the fence on that one (see comment further below)
- Wondering though if this function should do a check if `outs` is a terminal, and return an error instead otherwise 🤔 ❓ 

Also wondering; is `message` always required, or could we think of situations where the message itself is already printed, and we're only interested in input provided by a user?

In that case, this could be named `WaitForInput()` and/or have a `WithMessage()` option.


Finally; as this is a new function, and `cli/command` is becoming a bit of a grab-bag of things; wondering if we should put this in a separate package; e.g. `prompt.WaitForInput()` or `userinput.Prompt()`.

thaJeztah: Even considering; if it's a separate package, we could have a separate `WaitforPassword` (`WaitforPassphrase` ?) or `RequestPassphrase` utility that would handle the `disableEcho`.  In that case, we could get rid of the `PromptOptions` type (at least for now), or keep that one internal until we have new options that could warrant introducing more options.

cli/command/utils.go
thaJeztah: Alternative could be to pass a bare `io.Reader`, and always handle wrapping it into a `streams.In` internally; perhaps we don't even need the `streams` considering that we already use `moby/term` (which does most / all of the heavy-lifting in the `streams` package.

I'm curious though; if we need the special handling here; did we do something wrong with setting up the CLI and/or elsewhere; we should definitely look at that, and check if this is still needed.

I did a quick search; this code was originally introduced in the Moby repo (before it was moved to docker/cli); in this commit; https://github.com/moby/moby/commit/9c765040a5638429384b56a1b77139c6d4049c75

That's this PR, and related tickets;

- https://github.com/moby/moby/pull/17738
- https://github.com/moby/moby/issues/14336
- https://github.com/moby/moby/issues/14210

That PR was for Windows RS4, so ... really old, and if it was a Windows bug, that side may at least be obsolete (🤞) but we should check!

Benehiko: I've removed the windows check in the last commit https://github.com/docker/cli/pull/5168/commits/668f8743225a610f07411f80258ba3653f477810 - but still need to test a build on windows. I guess I'll need to install some older (supported) versions to ensure the majority of users won't be impacted.

cli/command/utils.go
Benehiko: I'm not sure if the `message` parameter could have a default value since this is not a prompt like "confirm x" it's more like an open ended question "input x". 

I've also gotten rid of the variadic parameter and instead opted for a utility function to disable input echo's in the terminal. 

I guess one could also check inside this `PromptForInput` function if it's a terminal or not - but thinking about the `PromptForConfirmation` function we also don't check it there. I guess we would need think about a redesign of these two functions and move them into their own package.

cli/command/registry/login_test.go
```diff
@@ -185,3 +188,41 @@ func TestRunLogin(t *testing.T) {
...
 }
+
+func TestLoginTermination(t *testing.T) {
+	p, tty, err := pty.Open()
+	assert.NilError(t, err)
+
+	defer func() {
+		_ = tty.Close()
+		_ = p.Close()
+	}()
```
vvoland, line 199: This (and the tmpFile.Remove below) could use `t.Cleanup`

cli/command/utils.go
```diff
@@ -76,6 +77,52 @@ func PrettyPrint(i any) string {
 
 var ErrPromptTerminated = errdefs.Cancelled(errors.New("prompt terminated"))
 
+// DisableInputEcho disables input echo on the provided io.ReadCloser.
+// This is useful when the user provides sensitive information like passwords.
+// The function returns a restore function that should be called to restore the
+// terminal state.
+func DisableInputEcho(in io.ReadCloser) (restore func() error, err error) {
+	ins := streams.NewIn(in)
```
vvoland, line 85:
```suggestion
func DisableInputEcho(ins *streams.In) (restore func() error, err error) {
```

The only call to this function passes a return value of `cli.In()` as the parameter which is already a `streams.In` so there's no need for the function to double-wrap the reader.

cli/command/utils.go
Benehiko: I've tested on a recent version of windows 10 ([19045.4529](https://support.microsoft.com/en-us/topic/june-11-2024-kb5039211-os-builds-19044-4529-and-19045-4529-f7e528c9-5e9f-4cd8-9161-704708448517)) and seems to work fine. I removed these lines https://github.com/docker/cli/blob/b83cf582cd3b932b6fce900a54cf0e1e6f61ac17/cli/command/registry.go#L93-L104 locally and tested the login flow. Works fine in powershell and command prompt

vvoland (approved):
LGTM; needs a squash though

laurazard (approved):
LGTM

krissetto (approved):

PR #5171, opened 2024-06-18
fix: force cli to exit after third sigint/sigterm

Benehiko:
**- What I did**

Added a global signal handler for cases where context cancellation is not respected. This will force the CLI to exit after 3 attempts of sigint/sigterm.

```console
⋊> ~/G/docker-cli on master ⨯ ./build/docker login                                                                                                                                          15:41:40

Username: ^C^C^C
got 3 SIGTERM/SIGINTs, forcefully exiting
```

**- How I did it**

I register a go routine to catch further signals upon running a docker command. Docker plugin handling has its own signal handling and is excluded from this signal handler. 

**- How to verify it**

Try with `docker login` since it's currently not respecting context cancellation.

```markdown changelog


```

### DIFF
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -325,6 +326,25 @@ func tryPluginRun(ctx context.Context, dockerCli command.Cli, cmd *cobra.Command
 	return nil
 }
 
+// registerForceExitGoroutine registers a goroutine that will force exit the
+// process after 3 SIGTERM/SIGINT signals.
+func registerForceExitGoroutine(ctx context.Context, w io.Writer) {
+	// setup a signal handler to force exit after 3 SIGTERM/SIGINT
+	go func() {
+		<-ctx.Done()
+		sig := make(chan os.Signal, 2)
+		signal.Notify(sig, platformsignals.TerminationSignals...)
+		count := 0
+		for range sig {
+			count++
+			if count >= 2 {
+				_, _ = fmt.Fprint(w, "\ngot 3 SIGTERM/SIGINTs, forcefully exiting\n")
+				os.Exit(1)
+			}
+		}
+	}()
+}
+
 //nolint:gocyclo
 func runDocker(ctx context.Context, dockerCli *command.DockerCli) error {
 	tcmd := newDockerCommand(dockerCli)
@@ -383,6 +403,10 @@ func runDocker(ctx context.Context, dockerCli *command.DockerCli) error {
 			}
 		}
 	}
+
+	// This is a fallback for the case where the command does not exit
+	// based on context cancellation.
+	registerForceExitGoroutine(ctx, dockerCli.Err())
 
 	// We've parsed global args already, so reset args to those
 	// which remain.
--- a/cmd/docker/docker.go
+++ b/cmd/docker/docker.go
@@ -326,23 +326,23 @@ func tryPluginRun(ctx context.Context, dockerCli command.Cli, cmd *cobra.Command
 	return nil
 }
 
-// registerForceExitGoroutine registers a goroutine that will force exit the
-// process after 3 SIGTERM/SIGINT signals.
-func registerForceExitGoroutine(ctx context.Context, w io.Writer) {
-	// setup a signal handler to force exit after 3 SIGTERM/SIGINT
-	go func() {
-		<-ctx.Done()
-		sig := make(chan os.Signal, 2)
-		signal.Notify(sig, platformsignals.TerminationSignals...)
-		count := 0
-		for range sig {
-			count++
-			if count >= 2 {
-				_, _ = fmt.Fprint(w, "\ngot 3 SIGTERM/SIGINTs, forcefully exiting\n")
-				os.Exit(1)
-			}
-		}
-	}()
+// forceExitAfter3TerminationSignals waits for the first termination signal
+// to be caught and the context to be marked as done, then registers a new
+// signal handler for subsequent signals. It forces the process to exit
+// after 3 SIGTERM/SIGINT signals.
+func forceExitAfter3TerminationSignals(ctx context.Context, w io.Writer) {
+	// wait for the first signal to be caught and the context to be marked as done
+	<-ctx.Done()
+	// register a new signal handler for subsequent signals
+	sig := make(chan os.Signal, 2)
+	signal.Notify(sig, platformsignals.TerminationSignals...)
+
+	// once we have received a total of 3 signals we force exit the cli
+	for i := 0; i < 2; i++ {
+		<-sig
+	}
+	_, _ = fmt.Fprint(w, "\ngot 3 SIGTERM/SIGINTs, forcefully exiting\n")
+	os.Exit(1)
 }
 
 //nolint:gocyclo
@@ -406,7 +406,7 @@ func runDocker(ctx context.Context, dockerCli *command.DockerCli) error {
 
 	// This is a fallback for the case where the command does not exit
 	// based on context cancellation.
-	registerForceExitGoroutine(ctx, dockerCli.Err())
+	go forceExitAfter3TerminationSignals(ctx, dockerCli.Err())
 
 	// We've parsed global args already, so reset args to those
 	// which remain.
--- a/cmd/docker/docker_test.go
+++ b/cmd/docker/docker_test.go
@@ -1,17 +1,28 @@
 package main
 
 import (
+	"bufio"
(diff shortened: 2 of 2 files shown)


cmd/docker/docker.go
```diff
@@ -325,6 +326,25 @@ func tryPluginRun(ctx context.Context, dockerCli command.Cli, cmd *cobra.Command
 	return nil
 }
 
+// registerForceExitGoroutine registers a goroutine that will force exit the
+// process after 3 SIGTERM/SIGINT signals.
+func registerForceExitGoroutine(ctx context.Context, w io.Writer) {
+	// setup a signal handler to force exit after 3 SIGTERM/SIGINT
+	go func() {
+		<-ctx.Done()
```
krissetto, line 334: nit: i think we could leave a small comment here as to why we're waiting on this `ctx.Done()` chan, and why we check for `count >= 2` below

krissetto (approved):
maybe we should also add a basic test case for this

cmd/docker/docker.go
```diff
@@ -325,6 +326,25 @@ func tryPluginRun(ctx context.Context, dockerCli command.Cli, cmd *cobra.Command
 	return nil
 }
 
+// registerForceExitGoroutine registers a goroutine that will force exit the
+// process after 3 SIGTERM/SIGINT signals.
+func registerForceExitGoroutine(ctx context.Context, w io.Writer) {
+	// setup a signal handler to force exit after 3 SIGTERM/SIGINT
+	go func() {
```
vvoland, line 333: Synchronous functions should be preferred over the asynchronous. Especially when the whole function body is wrapped in the goroutine - it's better to leave the `go ` up to the caller.

cmd/docker/docker.go
vvoland, line 334: Since this is expected to be a fallback mechanism - I think we can just rely on catching the signals directly instead of relying on this to receive the first cancellation.

cmd/docker/docker.go
```diff
@@ -325,6 +326,25 @@ func tryPluginRun(ctx context.Context, dockerCli command.Cli, cmd *cobra.Command
...
+		sig := make(chan os.Signal, 2)
+		signal.Notify(sig, platformsignals.TerminationSignals...)
+		count := 0
+		for range sig {
+			count++
+			if count >= 2 {
+				_, _ = fmt.Fprint(w, "\ngot 3 SIGTERM/SIGINTs, forcefully exiting\n")
+				os.Exit(1)
+			}
+		}
```
vvoland, line 344: This can be simplified into:

```suggestion
		for i := 0; i < 3; i++ {
		   <-sig
		}
		_, _ = fmt.Fprintln(w, "\ngot 3 SIGTERM/SIGINTs, forcefully exiting")
		os.Exit(1)
```

cmd/docker/docker.go
laurazard: Either this, or we could collapse the now two places where we're receiving signals (the main mechanism, set up in https://github.com/laurazard/cli/blob/bc291f0c987156364a7db2ca79f791e5cf387174/cmd/docker/docker.go#L38) and this one, and instead handle everything there? This seems simpler than registering two separate signal handlers, and condenses the signal handling logic into one place, which I prefer.

Something like:
```go
ctx, cancel := context.WithCancel(context.Background())
	sigChan := make(chan os.Signal, 3)
	signal.Notify(sigChan, platformsignals.TerminationSignals...)
	go func() {
		cancelCounter := 0
		for range sigChan {
			if cancelCounter >= 2 {
				os.Exit(1)
			}
			cancel()
			cancelCounter++
		}
	}()
```

laurazard (commented):
I had another idea, let me know what you think!

Benehiko:
> or we could collapse the now two places where we're receiving signals
This is a cleaner solution, however, what about plugins though? Since we have our own signal handler for plugins and the strategy of killing plugins is separate from the main CLI commands.

cmd/docker/docker.go
Benehiko: I would say that the function quite explicitly mentions that it's a background task and this is not an uncommon pattern in Go. It's also not exported so it should only be used inside this package - and only once. I would use a closure, but thought wrapping it in a function with a name that's explicit to its intent would make it more obvious to future contributors.

vvoland: Encoding the behavior with language feature is always more readable and explicit than using a function name.

There's no need to use a closure on the caller side:
```go
go forceExitAfter3TerminationsSignals(ctx, dockerCli.Err())
```

cmd/docker/docker.go
Benehiko: This is a cleaner solution, however, what about plugins though? Since we have our own signal handler for plugins and the strategy of killing plugins is separate from the main CLI commands.

laurazard: Ah, right 🤦‍♀️ nevermind my comment then, I guess at least for now we should stay consistent with the plugin pattern and register a new signal handler after the main context is cancelled.

cmd/docker/docker.go
Benehiko: Alright, I've updated it accordingly

laurazard (approved):

cmd/docker/docker.go
```diff
@@ -325,6 +326,25 @@ func tryPluginRun(ctx context.Context, dockerCli command.Cli, cmd *cobra.Command
...
+	// register a new signal handler for subsequent signals
+	sig := make(chan os.Signal, 2)
+	signal.Notify(sig, platformsignals.TerminationSignals...)
+
+	// once we have received a total of 3 signals we force exit the cli
+	for i := 0; i < 2; i++ {
+		<-sig
+	}
+	_, _ = fmt.Fprint(w, "\ngot 3 SIGTERM/SIGINTs, forcefully exiting\n")
+	os.Exit(1)
```
krissetto, line 345: Totally non blocking, just food for thought..are we sure we want to exit with status 1 when the user decides to forcefully exit? 

It could make sense to start differentiating expected status codes from totally unexpected failures, in order to be able to distinguish them from a tracing/metrics perspective as well.

cmd/docker/docker_test.go
```diff
@@ -75,3 +86,92 @@ func TestVersion(t *testing.T) {
...
+			os.Exit(0)
+		}
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+
+	// spawn the child process
+	cmd := exec.CommandContext(ctx, os.Args[0], "-test.run=TestFallbackForceExit")
```
vvoland, line 126: I'm a bit worried about that 🙈 

At this point it should be a `e2e` test as it's no longer a unit test.

I'm fine with leaving that for a follow up though.

vvoland: @thaJeztah  WDYT?

vvoland (changes requested):

krissetto (approved):

cmd/docker/docker_test.go
krissetto: i agree that this is more of an e2e test

Benehiko: What would be the drawback of this pattern? I understand that this is technically an e2e test, but I think this is a better way than compiling everything into a binary which you should manage, this allows the `go test` command to manage the binary for you.

See the same pattern being used in `os/exec` https://github.com/golang/go/blob/master/src/os/exec/exec_test.go

vvoland: Yeah, also this has A LOT of assumptions on how the tests are actually ran. Depending on the original invocation, this can run more tests than the TestFallbackForceExit (`-test.run` flag is additive).

cmd/docker/docker.go
Benehiko: I think forcefully exiting the CLI like this should be considered a problem. It means that the process they were cancelling did not cancel the first time through `context.Done()`.

cmd/docker/docker_test.go
vvoland: `os/exec` tests are a bit different - the reexec is handled much earlier in the `TestMain`:
https://github.com/golang/go/blob/4f77a83589baa5a3038cc6e35615084dc7e5f0c1/src/os/exec/exec_test.go#L73

Even then, this code is in the Go source and these tests test the `exec` std Go package, so maintainers of that code are probably much more comfortable with having more assumptions about how the tests are executed.

Also, we use `gotestsum` which calls the `go test` for us, and that's also another layer we need to be careful about (not sure what parameters it actually passes to the `go test`).

Benehiko: please see https://github.com/docker/cli/pull/5182

vvoland (approved):
LGTM, thanks!

Issue #5001, opened 2024-04-10
build: don't hardcode classic builder for windows daemons


thaJeztah:
### Description

relates to:

- https://github.com/docker/cli/pull/3314
- https://github.com/moby/moby/pull/43657
- https://techcommunity.microsoft.com/t5/containers/experimental-windows-containers-support-for-buildkit-released-in/ba-p/4096116

When support for BuildKit was removed from the CLI (using Buildx instead), we implemented logic to detect whether BuildKit or the classic builder should be used. As Windows did not support BuildKit, the CLI was hard-coded to use the classic builder in Windows binaries (with an option to _force_ using buildx when using `docker buildx install`).

While current versions of the Docker Engine do not yet support BuildKit on Windows, work is in progress to add support for it, and BuildKit 0.13 has [experimental support for Windows](https://techcommunity.microsoft.com/t5/containers/experimental-windows-containers-support-for-buildkit-released-in/ba-p/4096116).

The daemon provides a BuilderVersion field as part of the `/_ping` response to advertise the _suggested_ version of the Builder to use. Currently that is `2` (BuildKit) on Linux, and `1` (classic builder) on Windows;

```bash
curl -I --unix-socket /var/run/docker.sock http://localhost/_ping
HTTP/1.1 200 OK
Api-Version: 1.45
Builder-Version: 2
Cache-Control: no-cache, no-store, must-revalidate
Content-Type: text/plain; charset=utf-8
Date: Wed, 10 Apr 2024 09:11:20 GMT
Docker-Experimental: false
Ostype: linux
Pragma: no-cache
Server: Docker/26.0.0 (linux)
Swarm: active/manager
```

While that information must be considered a _recommendation_ (the client is still allowed to opt-in/opt-out), we should use this information as part of the  logic in the CLI to decide whether BuildKit or the legacy builder should be used, so that the CLI can be used with a Windows daemon that has experimental support for BuildKit enabled would automatically use BuildKit without changes in the CLI's configuration.

Note that, for this to be functional, a Windows Buildx binary (CLI-plugin) must be present; similar to how this is needed on Linux.

thaJeztah:
Just adding here, as I'm not sure now if it was hard-coded for Windows _CLI_ or if the CLI is connected to a Windows daemon (in either case, the logic should look at what the daemon supports, and not based on platform)

PR #5178, opened 2024-06-19
build: allow BuildKit to be used on Windows daemons that advertise it

thaJeztah:
- closes https://github.com/docker/cli/issues/5001
- relates to https://github.com/docker/cli/pull/3314
- relates to https://github.com/docker/cli/pull/3435
- relates to https://github.com/moby/moby/pull/43657

### build: allow BuildKit to be used on Windows daemons that advertise it

Commit 6fef143dbcf5fdcb436d05cf8e67d56c24bc4e6e switched the CLI to use
BuildKit by default, but as part of that removed the use of the
BuildkitVersion field as returned by Ping.

Some follow-up changes in commits e38e6c51ffb55405321038dfcf3a477f and
e7a8748b939fbf7798352cff1e0b87ef updated the logic for detecting whether
BuildKit should be used or the legacy builder, but hard-coded using the
legacy builder for Windows daemons.

While Windows / WCOW does not yet support BuildKit by default, there is
work in progress to implement it, so we should not hard-code the assumption
that a Windows daemon cannot support BuildKit.

On the daemon-side, [moby@7b153b9] (Docker v23.0) changed the default as
advertised by the daemon to be BuildKit for Linux daemons. That change
still hardcoded BuildKit to be unsupported for Windows daemons (and does
not yet allow overriding the config), but this may change for future
versions of the daemon, or test-builds.

This patch:

- Re-introduces checks for the BuildkitVersion field in the "Ping" response.
- If the Ping response from the daemon advertises that it supports BuildKit,
  the CLI will now use BuildKit as builder.
- If we didn't get a Ping response, or the Ping response did NOT advertise
  that the daemon supported BuildKit, we continue to use the current
  defaults (BuildKit for Linux daemons, and the legacy builder for Windows)
- Handling of the DOCKER_BUILDKIT environment variable is unchanged; for
  CLI.BuildKitEnabled, DOCKER_BUILDKIT always takes precedence, and for
  processBuilder the value is taken into account, but will print a warning
  when BuildKit is disabled and a Linux daemon is used. For Windows daemons,
  no warning is printed.

[moby@7b153b9]: https://github.com/moby/moby/commit/7b153b9e28b53779215de209736310f9266b3f2f

```markdown changelog
build: allow BuildKit to be used on Windows daemons that advertise it
```

### DIFF
--- a/cli/command/cli.go
+++ b/cli/command/cli.go
@@ -184,9 +184,18 @@ func (cli *DockerCli) BuildKitEnabled() (bool, error) {
 	if _, ok := aliasMap["builder"]; ok {
 		return true, nil
 	}
-	// otherwise, assume BuildKit is enabled but
-	// not if wcow reported from server side
-	return cli.ServerInfo().OSType != "windows", nil
+
+	si := cli.ServerInfo()
+	if si.BuildkitVersion == types.BuilderBuildKit {
+		// The daemon advertised BuildKit as the preferred builder; this may
+		// be either a Linux daemon or a Windows daemon with experimental
+		// BuildKit support enabled.
+		return true, nil
+	}
+
+	// otherwise, assume BuildKit is enabled for Linux, but disabled for
+	// Windows / WCOW, which does not yet support BuildKit by default.
+	return si.OSType != "windows", nil
 }
 
 // HooksEnabled returns whether plugin hooks are enabled.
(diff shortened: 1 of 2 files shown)


thaJeztah:
@crazy-max @colinhemmings PTAL - please double check this change

Issue #5194, opened 2024-06-25
v26.1.4 -> 27.0.1: "docker login https://myregistry:444/" loses port


linsomniac:
### Description

Also reported to Jenkins Docker Plugin here: https://github.com/jenkinsci/docker-plugin/issues/1082

The Jenkins Docker plugin insists on prefixing the registry with "https://" (see above bug for details).  After an upgrade to 27.0.1, this is now failing, reporting a 404 error reaching "https://myregistry.example.com/v2/" (note the missing :444).

At v26.1.4, both of these commands produce the same output:

- `docker login https://myregistry.example.com:444/`
- `docker login myregistry.example.com:444`

If I enter an invalid user/password, both of those report "401 Unauthorized".

At 27.0.1, the second command produces the 401, but the first command reports a 404, with the URL "https://myregistry.example.com/v2/"

I have checked the 27.0.1 release notes and have seen nothing about a change related to the registry URL or docker login.

### Reproduce

Detailed above.

### Expected behavior

Detailed above.

### docker version

```bash
Client: Docker Engine - Community
 Version:           27.0.1
 API version:       1.46
 Go version:        go1.21.11
 Git commit:        7fafd33
 Built:             Mon Jun 24 14:57:52 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          27.0.1
  API version:      1.46 (minimum version 1.24)
  Go version:       go1.21.11
  Git commit:       ff1e2c0
  Built:            Mon Jun 24 14:57:52 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.18
  GitCommit:        ae71819c4f5e67bb4d5ae76a6b735f29cc25774e
 runc:
  Version:          1.7.18
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client: Docker Engine - Community
 Version:    27.0.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.15.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.28.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 1
  Running: 1
  Paused: 0
[...]

PR #5195, opened 2024-06-25
re-introduced support for port numbers in docker registry URL

CarstonSchilds:
**- What I did**
* this is a fix for a regression introduced by https://github.com/docker/cli/pull/3599
* re-introduced support for port numbers in docker registry URL
* fixes https://github.com/docker/cli/issues/5194

**- How to verify it**
* added new test cases and ran `docker buildx bake test`

**- Description for the changelog**

```markdown changelog
Fix a regression that caused port numbers to be ignored when parsing a Docker registry URL.
```

**- A picture of a cute animal (not mandatory but encouraged)**

PR #5196, opened 2024-06-26
cli/config/credentials: ConvertToHostname: handle IP-addresses

thaJeztah:
**- What I did**
- follow-up to https://github.com/docker/cli/pull/5195
- addresses another regression in https://github.com/docker/cli/pull/3599
- fixes https://github.com/docker/cli/issues/5194

```markdown changelog
Fix handling of IPv6 addresses with custom ports on docker login
```

### DIFF
--- a/cli/config/credentials/file_store.go
+++ b/cli/config/credentials/file_store.go
@@ -1,6 +1,7 @@
 package credentials
 
 import (
+	"net"
 	"net/url"
 	"strings"
 
@@ -77,7 +78,7 @@ func ConvertToHostname(maybeURL string) string {
 			if u.Port() == "" {
 				return u.Hostname()
 			}
-			return u.Hostname() + ":" + u.Port()
+			return net.JoinHostPort(u.Hostname(), u.Port())
 		}
 	}
 	hostName, _, _ := strings.Cut(stripped, "/")
--- a/cli/config/credentials/file_store_test.go
+++ b/cli/config/credentials/file_store_test.go
@@ -138,6 +138,19 @@ func TestFileStoreErase(t *testing.T) {
 func TestConvertToHostname(t *testing.T) {
 	tests := []struct{ input, expected string }{
 		{
+			input:    "127.0.0.1",
+			expected: "127.0.0.1",
+		},
+		{
+			input:    "::1",
+			expected: "::1",
+		},
+		{
+			// FIXME(thaJeztah): this should be normalized to "::1" if there's no port (or vice-versa, as long as we're consistent)
+			input:    "[::1]",
+			expected: "[::1]",
+		},
+		{
 			input:    "example.com",
 			expected: "example.com",
 		},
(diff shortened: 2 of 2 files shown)


vvoland:
Needs a rebase

thaJeztah:
Rebased; I also removed the cherry-pick label for now, as this was not the previous handling of these, so perhaps OK for the next release.

vvoland:
The implementation before https://github.com/docker/cli/pull/3599 only removed the `http(s)://` prefix and took everything until `/` - so it would still preserve the port in the IPv6 address (`http://[::1]:1234/`):
https://github.com/docker/cli/blob/94e9aa68919428e861cd8659a2d0a13e4b50ba66/vendor/github.com/docker/docker/registry/auth.go#L130-L142

So IMO we should still cherrypick this.

vvoland (approved):

thaJeztah:
> so it would still preserve the port in the IPv6 address (`http://[::1]:1234/`) and not add an extra `:` if no explicit port was given (`http://[::1]/`).


So this function is to normalise hosts, in order to lookup which credentials to use if they're present in `~/.docker/config.json`. So for that, we should make sure that we de-duplicate entries to prevent `::1:5000` being considered different than `[::1]:5000`. 

There's some existing weird behavior where we treat `index.docker.io` separate, and for that we use the _URL_ for lookup (`https://index.docker.ioi/v1/`) but for any other registry, we use `hostname[:port]`.

AFAIK, the _intent_ was always to store credentials per _hostname_, which is why URLs are trimmed (i.e. no separate credentials for `hostname.example.com/v2/foo` and `hostname.example.com/v2/bar`), but _port_ is preserved, because different ports may be different registries / services, and credentials should not be sent to unrelated ones.

That, hmmm, also brings up the other topic; ambiguous port numbers, because `https://example.com` and `http://example.com` are _implicitly_ different ports (`443` vs `80`), but because we're trimming the scheme, that information gets lost. This also means that `example.com:80` and `example.com:443` are now considered separate entries.

And to add to the fun, some credential helpers _require_ a scheme in order to store credentials, and because we want to precent credentials that were stored for a TLS host to be sent to a non-TLS host (with the same name), we use `https://` to store the credentials


🫠  🫠 🫠 🫠 


TL;DR we need to better define how these must be normalised to prevent ambiguity.






It's somewhat fine to pick whatever format for these, but

vvoland:
Without this PR, we're still not doing the same thing as before https://github.com/docker/cli/pull/3599.

Consider `https://[::1]:5000` as an input:

Before we would correctly return the `[::1]:5000`, after https://github.com/docker/cli/pull/5195 the result would be `::1:5000`.

If that host was stored on pre v27 versions, it would be stored as `[::1]:5000`. IMO, it's better to keep the stored data consistent with previous versions, just in case there is other code that's reading this state (like Docker Desktop?).

thaJeztah:
OH!! Now I get it. Sorry, I'm slow. So yes, the `url.Parse` -> `.HostName()` would be stripping the square brackets because it's not part of the hostname (only used for `host:port`). Gotcha 😄 

Yes, in that case we should take this one for 27.0.2 as well

thaJeztah:
Prepared a backport; https://github.com/docker/cli/pull/5198

PR #5221, opened 2024-07-03
🚧  Add support for OAuth device-code login 🚧 

laurazard:
**- What I did**

WIP PR for discussion that adds support for the device-code flow OAuth login, when authenticating against the official registry.

I'll follow up with another PR to explore implementing this at the credential store layer/using a new credential store.

**- How I did it**

Added `cli/internal/oauth`, which includes a new `OAuthManager` that is capable of fulfilling a device-code flow against the official registry.

Fetching credentials remains the same, as the returned access token can be transparently used in place of the password when authenticating.

Login and logout flows now defer to `OAuthManager` when running against the official registry.

**- How to verify it**

Run `docker login`, and check your credentials store to find the stored access tokens. Attempt to do an operation which requires authentication (such as pushing an image to your repo) and verify it works.

```markdown changelog
Added support for the device-code flow OAuth login, when authenticating against the official registry.
```

Screenshot 2024-07-01 at 13 57 11


cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,217 @@
...
+	if state.UserCode == "" {
+		return res, errors.New("login failed: no user code returned")
+	}
+
+	// todo(laurazard): review the copy here
+	_, _ = fmt.Fprintf(w, "Attempting to automatically open the login authorization page in your default browser.\nIf the browser does not open or you wish to use a different device to authorize this request, open the following URL:\n\n%s\n", strings.Split(state.VerificationURI, "?")[0])
+	_ = m.openBrowser(state.VerificationURI)
+
+	_, _ = fmt.Fprintf(w, "\nAnd enter the device confirmation code: %s\n", state.UserCode)
+	_, _ = fmt.Fprint(w, "\nWaiting for authentication in the browser...\n")
```
dvdksn, line 91: Maybe something like this... if we have a "press enter to confirm" before launching the browser.

```suggestion
	_, _ = fmt.Fprintln(w, "You will be signed in using a web-based login.")
	_, _ = fmt.Fprintln(w, "To sign in with credentials on the command line, use 'docker login -u <username>'")
	_, _ = fmt.Fprintf(w, "\nYour one-time device confirmation code is: %s\n", state.UserCode)
	reader := bufio.NewReader(os.Stdin)
	_, _ = fmt.Fprint(w, "\nPress ENTER to open the browser.\n")
	_, _ = fmt.Fprintf(w, "Or open the URL manually: %s.\n", strings.Split(state.VerificationURI, "?")[0])
	reader.ReadString('\n')
	_, _ = fmt.Fprint(w, "\nWaiting for authentication in the browser...\n")
```

laurazard:
Recording with the UX/copy changes:

https://github.com/docker/cli/assets/70572044/8032e70b-8633-4d67-85b8-2b3002eb9c18

cli/internal/oauth/manager/manager.go
dvdksn: beautiful!

dvdksn:
you're probably doing this already... but can we hide the "Authenticating with existing credentials" (and error) bit if that isn't happening?

dvdksn:
misclick 😓

laurazard:
Yep! That message should only show up when we're actually logging in with existing credentials, what happened in that video was misleading because in truth we already had credentials, the "first" login (with those credentials) just failed for different reasons. I'll update the recording.

cli/internal/oauth/manager/util.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const (
+	audience = "https://hub.docker.com"
+	tenant   = "login.docker.com"
+	clientID = "DHWuMefQ1v4lxENpz8oUYH50yYSwyPvi"
+)
+
+func NewManager(store credentials.Store) (*OAuthManager, error) {
+	hostinfo, err := host.Info()
```
thaJeztah, line 19: I was, erm, curious where all the extra dependencies originated from :see_no_evil:;

- `github.com/tklauser/go-sysconf`
- `github.com/tklauser/go-sysconf` (number of online CPUs?)
- `github.com/shoenig/go-m1cpu` (CPU info for m1 CPUs)
- `github.com/power-devops/perfstat` (handling of `libperfstat` ??
- `github.com/go-ole/go-ole` (loading of Windows OLE objects)
- `github.com/yusufpapurcu/wmi` (WQL interface to Windows Management Instrumentation (WMI))


And those all look to be related to this single `host.Info()`; `gopsutil/host.Info()` is doing a _lot_ of things; https://github.com/shirou/gopsutil/blob/865b8c3f58d6ee4c47dab074fbf3c4b037abfd96/host/host.go#L65-L117

- Getting information about the distro-version, which;
   - on Linux involves trying to parse `/etc/os-release` and more; https://github.com/shirou/gopsutil/blob/865b8c3f58d6ee4c47dab074fbf3c4b037abfd96/host/host_linux.go#L173-L321
   - on macOS involves shelling out to `sw_vers -productVersion` https://github.com/shirou/gopsutil/blob/865b8c3f58d6ee4c47dab074fbf3c4b037abfd96/host/host_darwin.go#L98-L124
   - and on Windows invoking the infamous tricks to _try_ to get a version, because _there is no official API for this_ https://github.com/shirou/gopsutil/blob/865b8c3f58d6ee4c47dab074fbf3c4b037abfd96/host/host_windows.go#L146-L234
   - ^^ on Windows that's invoked _twice_ because it uses PlatformVersion as KernelVersion
- Getting the kernel architecture, which on Linux/Darwin is just `unix.Utsname`, but on Windows involves parsing processor architectures; https://github.com/shirou/gopsutil/blob/865b8c3f58d6ee4c47dab074fbf3c4b037abfd96/host/host_windows.go#L251-L280
- Trying to detect Virtualization, which is only implemented on Linux, but there it's a _lot_, including (again) parsing `/etc/os-release`; enough for the library author to implement caching, but that won't do much for a short-lived process like the CLI; https://github.com/shirou/gopsutil/blob/865b8c3f58d6ee4c47dab074fbf3c4b037abfd96/internal/common/common_linux.go#L168-L346
- It also tries to get `BootTime`, and `UpTime`, which again do a lot, including shelling out to various binaries, and it even has special cases for `docker`; caching was implemented there as well, but won't help us here; https://github.com/shirou/gopsutil/blob/865b8c3f58d6ee4c47dab074fbf3c4b037abfd96/internal/common/common_linux.go#L63-L119
- It tries to obtain the number of running processes on the machine, which on Linux means listing all files in `/proc`, and for each of them trying to see if it can convert them to an Integer (in which case it's a process-id); https://github.com/shirou/gopsutil/blob/865b8c3f58d6ee4c47dab074fbf3c4b037abfd96/internal/common/common_linux.go#L41-L61
    - :point_up: which makes me want to sit in a corner and cry
- It tries to obtain a `HostID` (`sys/kernel/random/boot_id`, on Linux, but shelling out to `ioreg -rd1 -c IOPlatformExpertDevice` on Darwin, and using the registry on Windows)

cli/internal/oauth/manager/util.go
```diff
@@ -0,0 +1,43 @@
...
+		ClientID:   clientID,
+		Tenant:     tenant,
+		DeviceName: "docker-cli:" + cliVersion,
+		Store:      store,
+	}
+
+	if hostinfo != nil {
+		hostVersion := strings.ReplaceAll(hostinfo.PlatformVersion, ".", "_")
+		options.DeviceName = fmt.Sprintf("docker-cli:%s:%s-%s-%s", cliVersion, hostinfo.OS, hostVersion, hostinfo.KernelArch)
+	}
```
thaJeztah, line 36: So I went looking "What information do we _really_ need here?"

- `cliVersion` - :warning: this function is set at compile time; there is no formal definition of the format (and can be set to anything)
- `hostinfo.OS` - this is hardcoded to `runtime.GOOS` in [`gopsutil/host.Info()`](https://github.com/shirou/gopsutil/blob/865b8c3f58d6ee4c47dab074fbf3c4b037abfd96/host/host.go#L65-L69)
- `hostVersion` - this comes from `PlatformVersion`. Is it a strong requirement to know the version of the OS? On Linux it will be useless without distro. On Windows, it will be "guesswork" (no official API); if _really_ needed, we could also use [`pkg/parsers/kernel.GetKernelVersion()`](https://github.com/moby/moby/blob/1205a9073320fba9e67fa5de3857f0330e56ce50/pkg/parsers/kernel/kernel_darwin.go#L13-L24) from the docker code.
- `hostinfo.KernelArch` - do we need detailed architecture (including CPU sub-variants), which could even be "off" when running with QEMU emulation; could we use `runtime.GOARCH` instead?


Can we get rid of this dependency, and replace `host.Info()` with something more basic?

thaJeztah (commented):
🙈 also gave it a quick try what it would look like without that `host.Info()` call; https://github.com/laurazard/cli/pull/2

cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,234 @@
...
+	access = accessAuth.Password
+
+	refreshAuth, err := m.credStore.Get(registryAuthKey + refreshTokenKey)
+	if err != nil {
+		return access, refresh, fmt.Errorf("failed to fetch refresh token: %w", err)
+	}
+	refresh = refreshAuth.Password
+
+	return
+}
```
vvoland, line 210: We _may_ need some other way to construct the access and refresh token keys, or make sure that there's no code that takes the registry hostname into account - like https://github.com/docker/cli/pull/3599 + https://github.com/docker/cli/pull/5195.
Otherwise, the access/refresh token _could_ end up being used as an actual auth token.

I haven't investigated it in detail, so it might not be an actual issue in this case, but thought it's worth bringing up 😅

laurazard: Hmm, that code only does that for non-official registries, so it should be fine (and indeed, that works). We can make sure it continues to work by adding some e2e tests, but maybe we can think of some other way to store these.

I still also want to try to implement some of these at the credstore level, so I'll see how that looks too/if this still matters.

vvoland: Yeah while it might not be a problem for the CLI code itself, I'm curious if other credential helpers might also do something like this.

Implementing these extra tokens on the credential store level would be nice indeed.

laurazard: yeah, I have a working branch now! I think it looks nicer, I'll ask y'all to take a look soon.

laurazard: We can check! What I was going to do was still at the CLI level, but as a wrapper to the CLI "credential store" implementations.

laurazard:
Superseded by https://github.com/docker/cli/pull/5244 or https://github.com/docker/cli/pull/5245

PR #5232, opened 2024-07-04
install bash-completion in dev container

thaJeztah:
### Dockerfile.dev: install bash-completion in dev container

It's not initialized, because there's no `docker` command installed
by default, but at least this makes sure that the basics are present
for testing.

### Makefile: add completion target

Add a "completion" target to install the generated completion
scripts inside the dev-container. As generating this script
depends on the docker binary, it calls "make binary" first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,16 @@ mod-outdated: ## check outdated dependencies
 authors: ## generate AUTHORS file from git history
 	scripts/docs/generate-authors.sh
 
+.PHONY: completion
+completion: binary
+completion: /etc/bash_completion.d/docker
+completion: ## generate and install the completion scripts
+
+.PHONY: /etc/bash_completion.d/docker
+/etc/bash_completion.d/docker: ## generate and install the bash-completion script
+	mkdir -p /etc/bash_completion.d
+	docker completion bash > /etc/bash_completion.d/docker
+
 .PHONY: manpages
 manpages: ## generate man pages from go source and markdown
 	scripts/docs/generate-man.sh
--- a/dockerfiles/Dockerfile.dev
+++ b/dockerfiles/Dockerfile.dev
@@ -35,6 +35,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
 FROM golang AS dev
 RUN  apk add --no-cache \
     bash \
+    bash-completion \
     build-base \
     ca-certificates \
     coreutils \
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -86,6 +86,10 @@ mod-outdated: ## check outdated dependencies
 authors: ## generate AUTHORS file from git history
 	scripts/docs/generate-authors.sh
 
+.PHONY: completion
+completion: binary ## install the generated bash-completion script in the dev container
+	docker completion bash > /etc/bash_completion.d/docker
```
Benehiko, line 91: does this need to be a separate make command?
Wondering if we could include the completion scripts (for all shells) inside the dev container by default.

thaJeztah: I was considering that, but we currently only have Bash installed. I think
we can consider adding more shells (but need to look at how much extra
time (and space) that would take by default to start the dev-container).

But yes, we can already split targets to prepare for that; I pushed an
update that splits the bash part to a separate make target :+1: PTAL

Makefile
```diff
@@ -86,6 +86,14 @@ mod-outdated: ## check outdated dependencies
 authors: ## generate AUTHORS file from git history
 	scripts/docs/generate-authors.sh
 
+.PHONY: completion
+completion: binary
+completion: /etc/bash_completion.d/docker
+completion: ## generate and install the completion scripts
+
+/etc/bash_completion.d/docker: FORCE ## generate and install the bash-completion script
```
vvoland, line 94: I think we can just use .PHONY here:

```suggestion
.PHONY: /etc/bash_completion.d/docker
/etc/bash_completion.d/docker:
```

Makefile
```diff
@@ -86,6 +86,14 @@ mod-outdated: ## check outdated dependencies
...
+	docker completion bash > /etc/bash_completion.d/docker
```
vvoland, line 95: We should also make sure that the directory exists:
```suggestion
    mkdir -p /etc/bash_completion.d
	docker completion bash > /etc/bash_completion.d/docker
```

Makefile
thaJeztah: Ah, yeah, both should work here; I picked `FORCE` to make it explicit it's not a "fake" target, so an actual file, but would still have to be forced; I had some recollection hat was the preferred way somewhere, but can't find that back (maybe it was some discussion in another project).

Makefile
thaJeztah: Good one; I wrote it with the dev-container in mind (which had it set up already), but doesn't hurt to add

Makefile
thaJeztah: updated 👍

Makefile
thaJeztah: updated 👍

vvoland (approved):

thaJeztah:
thx! I'll bring this one in, and rebase https://github.com/docker/cli/pull/5238

PR #5233, opened 2024-07-04
cli: FlagErrorFunc: don't print long usage output for invalid flags

thaJeztah:
- [x] follow-up to / stacked on https://github.com/docker/cli/pull/5231

### cli: FlagErrorFunc: don't print usage output for invalid flags

When trying to use an invalid flag, the CLI currently prints the a short
error message, instructions to use the `--help` flag to learn about the
correct usage, followed by the command's usage output.

While this is a common convention, and may have been a nice gesture when
docker was still young and only had a few commands and options ("you did
something wrong, but here's an overview of what you can use"), that's no
longer the case, and many commands have a _very_ long output.

The result of this is that the error message, which is the relevant
information in this case - "You mis-typed something" - is lost in the
output, and hard to find (sometimes even requiring scrolling back).

The output is also confusing, because it _looks_ like something ran
successfully (most of the output is not about the error!).

Even further; the suggested resolution (try `--help` to see the correct
options) is rather redundant, because running teh command with `--help`
produces _exactly_ the same output as was just showh, baring the error
message. As a fun fact, due to the usage output being printed, the
output even contains not one, but _two_ "call to actions";

- `See 'docker volume --help'.` (under the erro message)
- `Run 'docker volume COMMAND --help' for more information on a command.`
  (under the usage output)

In short; the output is too verbose, confusing, and doesn't provide
a good UX. Let's reduce the output produced so that the focus is on the
important information.

This patch:

- Changes the usage to the short-usage.
- Prefixes the error message with the binary / root-command name
  (usually `docker:`) to be consistent with `unknon command`, and helps
  to distinguish where the message originated from (the `docker` CLI in
  this case).
- Adds an empty line between the error-message and the "call to action"
  (`Run 'docker volume --help' ...` in the example below). This helps
  separating the error message ("unkown flag") from the call-to-action.

Before this patch:

    docker volume --no-such-flag
    unknown flag: --no-such-flag
    See 'docker volume --help'.
    
    Usage:  docker volume COMMAND
    
    Manage volumes
    
    Commands:
      create      Create a volume
      ls          List volumes
    
With this patch:

    docker volume --no-such-flag
    docker: unknown flag: --no-such-flag
    
    Usage:  docker volume COMMAND
    
    Run 'docker volume --help' for more information

```markdown changelog
Improve error-output for invalid flags on the command-line.
```

### DIFF
--- a/cli/cobra.go
+++ b/cli/cobra.go
@@ -92,12 +92,8 @@ func FlagErrorFunc(cmd *cobra.Command, err error) error {
 		return nil
 	}
 
-	usage := ""
-	if cmd.HasSubCommands() {
-		usage = "\n\n" + cmd.UsageString()
-	}
 	return StatusError{
-		Status:     fmt.Sprintf("%s\nSee '%s --help'.%s", err, cmd.CommandPath(), usage),
+		Status:     fmt.Sprintf("%s\n\nUsage:  %s\n\nRun '%s --help' for more information", err, cmd.UseLine(), cmd.CommandPath()),
 		StatusCode: 125,
 	}
 }
--- a/e2e/cli-plugins/help_test.go
+++ b/e2e/cli-plugins/help_test.go
@@ -78,6 +78,6 @@ func TestGlobalHelp(t *testing.T) {
 		})
 		assert.Assert(t, is.Equal(res2.Stdout(), ""))
 		assert.Assert(t, is.Contains(res2.Stderr(), "unknown flag: --badopt"))
-		assert.Assert(t, is.Contains(res2.Stderr(), "See 'docker --help"))
+		assert.Assert(t, is.Contains(res2.Stderr(), "Run 'docker --help' for more information"))
 	})
 }


cli/cobra.go
```diff
@@ -92,12 +92,8 @@ func FlagErrorFunc(cmd *cobra.Command, err error) error {
 		return nil
 	}
 
-	usage := ""
-	if cmd.HasSubCommands() {
-		usage = "\n\n" + cmd.UsageString()
-	}
 	return StatusError{
-		Status:     fmt.Sprintf("%s\nSee '%s --help'.%s", err, cmd.CommandPath(), usage),
+		Status:     fmt.Sprintf("%[2]s\n\nUsage:  %[3]s\n\nRun '%[4]s --help' for more information", cmd.Root().Name(), err, cmd.UseLine(), cmd.CommandPath()),
```
krissetto, line 96: nit: there are 2 spaces after `Usage:`

krissetto: Also, is `cmd.Root().Name()` being used?

krissetto (approved):
overall lgtm, check the comments

cli/cobra.go
thaJeztah: > nit: there are 2 spaces after `Usage:`

Yes, it looked like we're using 2 spaces for that everywhere; ISTR those may have been a `<tab>` in the past, which screwed up generating our YAML-docs, and were replaced; we should make a pass at checking where all of those are produced, and consider changing to a single one :see_no_evil:

```bash
docker --nosuch 2>&1 >/dev/null | grep Usage
```

> Also, is `cmd.Root().Name()` being used?

Oh! Good one; I don't think it is anymore! I went through a couple of iterations; one of them was including `docker:` as prefix, but I went back-and-forth on whether that made things clearer, as it would look like (e.g.);

```
docker: Usage:  docker [OPTIONS] COMMAND
```

Which felt just _too_ verbose and somewhat confusing, so in the end I removed that.

thaJeztah:
Updated; ptal 👍

cli/cobra.go
krissetto: OH, i didn't realize we double-space in multiple areas 🙈 no problem!

PR #5234, opened 2024-07-04
cli: improve output and consistency for unknown (sub)commands

thaJeztah:
- similar to https://github.com/docker/cli/pull/5233

### cli: improve output and consistency for unknown (sub)commands

Before this patch, output for invalid top-level and sub-commands differed.
For top-level commands, the CLI would print an error-message and a suggestion
to use `--help`. For missing *subcommands*, we would hit a different code-path,
and different output, which includes full "usage" / "help" output.

While it is a common convention to show usage output, and may have been
a nice gesture when docker was still young and only had a few commands
and options ("you did something wrong; here's an overview of what you
can use"), that's no longer the case, and many commands have a _very_
long output.

The result of this is that the error message, which is the relevant
information in this case - "You mis-typed something" - is lost in the
output, and hard to find (sometimes even requiring scrolling back).

The output is also confusing, because it _looks_ like something ran
successfully (most of the output is not about the error!).

Even further; the suggested resolution (try `--help` to see the correct
options) is rather redundant, because running teh command with `--help`
produces _exactly_ the same output as was just showh, baring the error
message. As a fun fact, due to the usage output being printed, the
output even contains not one, but _two_ "call to actions";

- `See 'docker volume --help'.` (under the erro message)
- `Run 'docker volume COMMAND --help' for more information on a command.`
  (under the usage output)

In short; the output is too verbose, confusing, and doesn't provide
a good UX. Let's reduce the output produced so that the focus is on the
important information.

This patch:

- Changes the usage to the short-usage.
- Changes the error-message to mention the _full_ command instead of only
  the command after `docker` (so `docker no-such-command` instead of
  `no-such-command`).
- Prefixes the error message with the binary / root-command name
  (usually `docker:`); this is something we can still decide on, but
  it's a pattern we already use in some places. The motivation for this
  is that `docker` commands can often produce output that's a combination
  of output from the CLI itself, output from the daemon, and even output
  from the container. The `docker:` prefix helps to distinguish where
  the message originated from (the `docker` CLI in this case).
- Adds an empty line between the error-message and the "call to action"
  (`Run 'docker volume --help'...` in the example below). This helps
  separating the error message ("unkown flag") from the call-to-action.

Before this patch:

Unknown top-level command:

    docker nosuchcommand foo
    docker: 'nosuchcommand' is not a docker command.
    See 'docker --help'

Unknown sub-command:

    docker volume nosuchcommand foo
    
    Usage:  docker volume COMMAND
    
    Manage volumes
    
    Commands:
      create      Create a volume
      ls          List volumes
    
After this patch:

Unknown top-level command:

    docker nosuchcommand foo
    docker: unknown command: docker nosuchcommand
    
    Run 'docker --help' for more information

Unknown sub-command:

    docker volume nosuchcommand foo
    docker: unknown command: 'docker volume nosuchcommand'
    
    Usage:  docker volume COMMAND
    
    Run 'docker volume --help' for more information

### cli: improve argument validation output

Improve the output for these validation errors:

- Removes the short command description from the output. This information
  does not provide much useful help, and distracts from the error message.
- Reduces punctuation, and 
- Prefixes the error message with the binary / root-command name
  (usually `docker:`) to be consistent with other similar errors.
- Adds an empty line between the error-message and the "call to action"
  (`Run 'docker volume --help'...` in the example below). This helps
  separating the error message and "usage" from the call-to-action.

Before this patch:

    $ docker volume ls one two three
    "docker volume ls" accepts no arguments.
    See 'docker volume ls --help'.
    
    Usage:  docker volume ls [OPTIONS]
    
    List volumes
    
    
    $ docker volume create one two three
    "docker volume create" requires at most 1 argument.
    See 'docker volume create --help'.
    
    Usage:  docker volume create [OPTIONS] [VOLUME]
    
    Create a volume

With this patch:

    $ docker volume ls one two three
    docker: 'docker volume ls' accepts no arguments
    
    Usage:  docker volume ls [OPTIONS]
    
    Run 'docker volume ls --help' for more information

    $ docker voludocker volume create one two three
    docker: 'docker volume create' requires at most 1 argument
    
    Usage:  docker volume create [OPTIONS] [VOLUME]
    
    SRun 'docker volume create --help' for more information

```markdown changelog
- improve output and consistency for unknown (sub)commands
- improve output for invalid arguments
```

### DIFF
--- a/cli/required.go
+++ b/cli/required.go
@@ -22,11 +22,10 @@ func NoArgs(cmd *cobra.Command, args []string) error {
 	}
 
 	return errors.Errorf(
-		"%q accepts no arguments.\nSee '%s --help'.\n\nUsage:  %s\n\n%s",
-		cmd.CommandPath(),
+		"%[1]s: '%[2]s' accepts no arguments\n\nUsage:  %[3]s\n\nRun '%[2]s --help' for more information",
+		binName(cmd),
 		cmd.CommandPath(),
 		cmd.UseLine(),
-		cmd.Short,
 	)
 }
 
@@ -37,13 +36,12 @@ func RequiresMinArgs(min int) cobra.PositionalArgs {
 			return nil
 		}
 		return errors.Errorf(
-			"%q requires at least %d %s.\nSee '%s --help'.\n\nUsage:  %s\n\n%s",
+			"%[1]s: '%[2]s' requires at least %[3]d %[4]s\n\nUsage:  %[5]s\n\nSee '%[2]s --help' for more information",
+			binName(cmd),
 			cmd.CommandPath(),
 			min,
 			pluralize("argument", min),
-			cmd.CommandPath(),
 			cmd.UseLine(),
-			cmd.Short,
 		)
 	}
 }
@@ -55,13 +53,12 @@ func RequiresMaxArgs(max int) cobra.PositionalArgs {
 			return nil
 		}
 		return errors.Errorf(
-			"%q requires at most %d %s.\nSee '%s --help'.\n\nUsage:  %s\n\n%s",
+			"%[1]s: '%[2]s' requires at most %[3]d %[4]s\n\nUsage:  %[5]s\n\nSRun '%[2]s --help' for more information",
+			binName(cmd),
 			cmd.CommandPath(),
 			max,
 			pluralize("argument", max),
-			cmd.CommandPath(),
 			cmd.UseLine(),
-			cmd.Short,
 		)
 	}
 }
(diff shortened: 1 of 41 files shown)


cli/required.go
```diff
@@ -14,15 +12,20 @@ func NoArgs(cmd *cobra.Command, args []string) error {
 	}
 
 	if cmd.HasSubCommands() {
-		return errors.Errorf("\n" + strings.TrimRight(cmd.UsageString(), "\n"))
+		return errors.Errorf(
+			"%[1]s: unknown command: %[2]s %[3]s\n\nUsage:  %[4]s\n\nRun '%[2]s --help' for more information",
```
krissetto, line 16: For consistency with other error messages where we print out commands the user has inputted (or could input, like `Run '%[2]s --help' for more information` further in this message), maybe it could make sense to surround `%[2]s %[3]s` with  single quotes`'` as well.  

I think it could make the message a little bit more readable, given we are prefixing `docker:` as well. WDYT?

krissetto (approved):
overall LGTM, just left a small suggestion in the comments

Benehiko (approved):
LGTM

cli/required.go
thaJeztah: OHMAN I went back and forth on some of these; initially I had the quotes, but then also looked at `unknown flag:` which didn't have them; that one is generated by Cobra  (or spf13/flags), so not something I could immediately change.

Overall I somewhat think that the quotes may be good to add yes; perhaps still ok for a follow-up pass?

krissetto: Yeah for sure! It's definitely a minor thing, but if users are anything like me they do appreciate consistency in the outputs. We'll get to it in a separate pass 👍

PR #5236, opened 2024-07-05
cli/command/container: remove reportError, and put StatusError to use

thaJeztah:
- follow-up to https://github.com/docker/cli/pull/5229
- follow-up to https://github.com/docker/cli/pull/5231

The `reportError` utility was present because cli.StatusError would print the error decorated with `Status: <error-message>, Code: <exit-code>`. That was not desirable in many cases as it would mess-up the output. To prevent this, the CLI had code to check for an empty `Status` (error message) in which case the error would be "ignored" (and only used for the exit-status), and the `reportError` utility would be used to manually print a custom error message before returning the error.

Now that bca209006153d3e025cb3d31c3cd55eb2aec0c4f fixed the output format of `cli.StatusError`, and 3dd6fc365d853e21f0e11f9e6ab62c4f8ae438e7 and 350a0b68a9584ec9ae712b6eca906c1018ba6dac no longer discard these error, we can get rid of this utility, and just set the error-message for the status-error.

This patch:

- Introduces a `withHelp` which takes care of decorating errors with a "Run --help" hint for the user.
- Introduces a `toStatusError` utility that detects certain errors in the container to assign a corresponding exit-code (these error-codes can be used to distinguish "client" errors from "container" errors).
- Removes the `reportError` utility, and removes code that manually printed errors before returning.

Behavior is mostly unmodified, with the exception of some slight reformatting of the errors:

- `withHelp` adds a `docker:` prefix to the error, to indicate the error is produced by the `docker` command. This prefix was already present in most cases.
- The "--help" hint is slightly updated ("Run 'docker run --help' for more information" instead of "See 'docker run --help'"), to make it more clear that it's a "call to action".
- An empty is added before the "--help" hint to separate it better from the error-message.

Before this patch:

    $ docker run --pull=invalid-option alpine
    docker: invalid pull option: 'invalid-option': must be one of "always", "missing" or "never".
    See 'docker run --help'.
    $ echo $?
    125

    $ docker run --rm alpine nosuchcommand
    docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "nosuchcommand": executable file not found in $PATH: unknown.
    $ echo $?
    127

With this patch:

    $ docker run --pull=invalid-option alpine
    docker: invalid pull option: 'invalid-option': must be one of "always", "missing" or "never"

    Run 'docker run --help' for more information
    $ echo $?
    125

    $ docker run --rm alpine nosuchcommand
    docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "nosuchcommand": executable file not found in $PATH: unknown.

    Run 'docker run --help' for more information
    $ echo $?
    127

### DIFF
--- a/cli/command/container/create_test.go
+++ b/cli/command/container/create_test.go
@@ -189,8 +189,8 @@ func TestCreateContainerImagePullPolicyInvalid(t *testing.T) {
 
 			statusErr := cli.StatusError{}
 			assert.Check(t, errors.As(err, &statusErr))
-			assert.Equal(t, statusErr.StatusCode, 125)
-			assert.Check(t, is.Contains(dockerCli.ErrBuffer().String(), tc.ExpectedErrMsg))
+			assert.Check(t, is.Equal(statusErr.StatusCode, 125))
+			assert.Check(t, is.ErrorContains(err, tc.ExpectedErrMsg))
 		})
 	}
 }
@@ -285,7 +285,7 @@ func TestNewCreateCommandWithWarnings(t *testing.T) {
 	for _, tc := range testCases {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
-			cli := test.NewFakeCli(&fakeClient{
+			fakeCLI := test.NewFakeCli(&fakeClient{
 				createContainerFunc: func(config *container.Config,
 					hostConfig *container.HostConfig,
 					networkingConfig *network.NetworkingConfig,
@@ -295,15 +295,15 @@ func TestNewCreateCommandWithWarnings(t *testing.T) {
 					return container.CreateResponse{}, nil
 				},
 			})
-			cmd := NewCreateCommand(cli)
+			cmd := NewCreateCommand(fakeCLI)
 			cmd.SetOut(io.Discard)
 			cmd.SetArgs(tc.args)
 			err := cmd.Execute()
 			assert.NilError(t, err)
 			if tc.warning {
-				golden.Assert(t, cli.ErrBuffer().String(), tc.name+".golden")
+				golden.Assert(t, fakeCLI.ErrBuffer().String(), tc.name+".golden")
 			} else {
-				assert.Equal(t, cli.ErrBuffer().String(), "")
+				assert.Equal(t, fakeCLI.ErrBuffer().String(), "")
 			}
 		})
 	}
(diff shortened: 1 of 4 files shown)


cli/command/container/run.go
```diff
@@ -207,12 +212,11 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 			<-errCh
 		}
 
-		reportError(stderr, "run", err.Error(), false)
 		if copts.autoRemove {
 			// wait container to be removed
 			<-statusChan
 		}
-		return runStartContainerErr(err)
+		return toStatusError(err)
```
thaJeztah, line 219: Slight change here; previously we would print the error before the container was removed; this would be for the "old API, `--rm` running on the client-side" case. I don't think this would make much of a difference in practice, and it would only be for old API versions (API 1.24 I think?)

thaJeztah:
@laurazard first round of cleanup after the other one(s) 😄

cli/command/container/create_test.go
```diff
@@ -189,8 +189,8 @@ func TestCreateContainerImagePullPolicyInvalid(t *testing.T) {
 
 			statusErr := cli.StatusError{}
 			assert.Check(t, errors.As(err, &statusErr))
-			assert.Equal(t, statusErr.StatusCode, 125)
-			assert.Check(t, is.Contains(dockerCli.ErrBuffer().String(), tc.ExpectedErrMsg))
+			assert.Check(t, is.Equal(statusErr.StatusCode, 125))
```
laurazard, line 192: Why this change?

cli/command/container/run_test.go
```diff
@@ -179,8 +181,8 @@ func TestRunContainerImagePullPolicyInvalid(t *testing.T) {
 
 			statusErr := cli.StatusError{}
 			assert.Check(t, errors.As(err, &statusErr))
-			assert.Equal(t, statusErr.StatusCode, 125)
-			assert.Check(t, is.Contains(dockerCli.ErrBuffer().String(), tc.ExpectedErrMsg))
+			assert.Check(t, is.Equal(statusErr.StatusCode, 125))
```
laurazard, line 184: Same question

laurazard (approved):
LGTM, thanks!

cli/command/container/create_test.go
thaJeztah: I think I changed the `assert.Equal` -> `assert.Check` here so that we don't fail early; `assert.Equal` will do a `t.Fatal()` if it fails (so fail the test immediately), whereas `assert.Check()` will do a `t.Error` (mark the test as failed, but continue)

laurazard: Ahh, sounds good!

PR #5238, opened 2024-07-05
various improvements to shell completions

thaJeztah:
- [x] follow-up to / stacked on https://github.com/docker/cli/pull/5232

### cli/command/completion: add FileNames utility

This is just a convenience function to allow defining completion to
use the default (complete with filenames and directories).

### cli/command/completion: add EnvVarNames utility

EnvVarNames offers completion for environment-variable names. This
completion can be used for "--env" and "--build-arg" flags, which
allow obtaining the value of the given environment-variable if present
in the local environment, so we only should complete the names of the
environment variables, and not their value. This also prevents the
completion script from printing values of environment variables
containing sensitive values.

For example;

    export MY_VAR=hello
    docker run --rm --env MY_VAR alpine printenv MY_VAR
    hello

Before this patch:

    docker run --env GO
    GO111MODULE=auto        GOLANG_VERSION=1.21.12  GOPATH=/go              GOTOOLCHAIN=local

With this patch:

    docker run --env GO<tab>
    GO111MODULE     GOLANG_VERSION  GOPATH          GOTOOLCHAIN

### cli/command/completion: add FromList utility

It's an alias for cobra.FixedCompletions but takes a variadic list
of strings, so that it's not needed to construct an array for this.

### cli/command/container: provide flag-completion for "docker create"

"docker run" and "docker create" are mostly identical, so we can copy
the same completion functions, 

We could possibly create a utility for this (similar to `addFlags()` which
configures both commands with the flags they share). I considered combining 
his with `addFlags()`, but that utility is also used in various tests, in
which we don't need this feature, so keeping that for a future exercise.

### cmd/docker: fix completion for --context

registerCompletionFuncForGlobalFlags was called from newDockerCommand,
at which time no context-store is initialized yet, so it would return
a nil value, probably resulting in `store.Names` to panic, but these
errors are not shown when running the completion. As a result, the flag
completion would fall back to completing from filenames.

This patch changes the function to dynamically get the context-store;
this fixes the problem mentioned above, because at the time the completion
function is _invoked_, the CLI is fully initialized, and does have a
context-store available.

A (non-exported) interface is defined to allow the function to accept
alternative implementations (not requiring a full command.DockerCLI).

Before this patch:

    docker context create one
    docker context create two

    docker --context <TAB>
    .DS_Store                   .idea/                      Makefile
    .dockerignore               .mailmap                    build/
    ...

With this patch:

    docker context create one
    docker context create two

    docker --context <TAB>
    default  one      two

###  cli/context/store: Names(): fix panic when called with nil-interface

Before this, it would panic when a nil-interface was passed.

### cli/command/container: add completion for --cap-add, --cap-drop

With this patch:

    docker run --cap-add <TAB>
    ALL                     CAP_KILL                CAP_SETUID
    CAP_AUDIT_CONTROL       CAP_LEASE               CAP_SYSLOG
    CAP_AUDIT_READ          CAP_LINUX_IMMUTABLE     CAP_SYS_ADMIN
    CAP_AUDIT_WRITE         CAP_MAC_ADMIN           CAP_SYS_BOOT
    CAP_BLOCK_SUSPEND       CAP_MAC_OVERRIDE        CAP_SYS_CHROOT
    CAP_BPF                 CAP_MKNOD               CAP_SYS_MODULE
    CAP_CHECKPOINT_RESTORE  CAP_NET_ADMIN           CAP_SYS_NICE
    CAP_CHOWN               CAP_NET_BIND_SERVICE    CAP_SYS_PACCT
    CAP_DAC_OVERRIDE        CAP_NET_BROADCAST       CAP_SYS_PTRACE
    CAP_DAC_READ_SEARCH     CAP_NET_RAW             CAP_SYS_RAWIO
    CAP_FOWNER              CAP_PERFMON             CAP_SYS_RESOURCE
    CAP_FSETID              CAP_SETFCAP             CAP_SYS_TIME
    CAP_IPC_LOCK            CAP_SETGID              CAP_SYS_TTY_CONFIG
    CAP_IPC_OWNER           CAP_SETPCAP             CAP_WAKE_ALARM

### cli/command/container: add completion for --restart

With this patch:

    docker run --restart <TAB>
    always  no  on-failure  unless-stopped

cli/command/container: add completion for --volumes-from

With this patch:

    docker run --volumes-from amazing_nobel
    amazing_cannon     boring_wozniak         determined_banzai
    elegant_solomon    reverent_booth         amazing_nobel

cli/command/container: add completion for --stop-signal

With this patch:

    docker run --stop-signal <TAB>
    ABRT  IOT      RTMAX-4   RTMIN     RTMIN+11  TSTP
    ALRM  KILL     RTMAX-5   RTMIN+1   RTMIN+12  TTIN
    BUS   PIPE     RTMAX-6   RTMIN+2   RTMIN+13  TTOU
    CHLD  POLL     RTMAX-7   RTMIN+3   RTMIN+14  URG
    CLD   PROF     RTMAX-8   RTMIN+4   RTMIN+15  USR1
    CONT  PWR      RTMAX-9   RTMIN+5   SEGV      USR2
    FPE   QUIT     RTMAX-10  RTMIN+6   STKFLT    VTALRM
    HUP   RTMAX    RTMAX-11  RTMIN+7   STOP      WINCH
    ILL   RTMAX-1  RTMAX-12  RTMIN+8   SYS       XCPU
    INT   RTMAX-2  RTMAX-13  RTMIN+9   TERM      XFSZ
    IO    RTMAX-3  RTMAX-14  RTMIN+10  TRAP

```markdown changelog
add and improve shell completions for various flags
```

### DIFF
--- a/cli/command/container/completion.go
+++ b/cli/command/container/completion.go
@@ -2,6 +2,7 @@ package container
 
 import (
 	"github.com/docker/cli/cli/command/completion"
+	"github.com/docker/docker/api/types/container"
 	"github.com/spf13/cobra"
 )
 
@@ -64,6 +65,20 @@ var allLinuxCapabilities = []string{
 	"CAP_CHECKPOINT_RESTORE",
 }
 
+// restartPolicies is a list of all valid restart-policies..
+//
+// TODO(thaJeztah): add descriptions, and enable descriptions for our completion scripts (cobra.CompletionOptions.DisableDescriptions is currently set to "true")
+var restartPolicies = []string{
+	string(container.RestartPolicyDisabled),
+	string(container.RestartPolicyAlways),
+	string(container.RestartPolicyOnFailure),
+	string(container.RestartPolicyUnlessStopped),
+}
+
 func completeLinuxCapabilityNames(cmd *cobra.Command, args []string, toComplete string) (names []string, _ cobra.ShellCompDirective) {
 	return completion.FromList(allLinuxCapabilities...)(cmd, args, toComplete)
+}
+
+func completeRestartPolicies(cmd *cobra.Command, args []string, toComplete string) (names []string, _ cobra.ShellCompDirective) {
+	return completion.FromList(restartPolicies...)(cmd, args, toComplete)
 }
--- a/cli/command/container/completion.go
+++ b/cli/command/container/completion.go
@@ -1,0 +1,69 @@
+package container
+
+import (
+	"github.com/docker/cli/cli/command/completion"
+	"github.com/spf13/cobra"
+)
+
+// allLinuxCapabilities is a list of all known Linux capabilities.
+//
+// This list was based on the containerd pkg/cap package;
+// https://github.com/containerd/containerd/blob/v1.7.19/pkg/cap/cap_linux.go#L133-L181
+//
+// TODO(thaJeztah): add descriptions, and enable descriptions for our completion scripts (cobra.CompletionOptions.DisableDescriptions is currently set to "true")
+var allLinuxCapabilities = []string{
+	"ALL", // magic value for "all capabilities"
+
+	// caps35 is the caps of kernel 3.5 (37 entries)
+	"CAP_CHOWN",            // 2.2
+	"CAP_DAC_OVERRIDE",     // 2.2
+	"CAP_DAC_READ_SEARCH",  // 2.2
+	"CAP_FOWNER",           // 2.2
+	"CAP_FSETID",           // 2.2
+	"CAP_KILL",             // 2.2
+	"CAP_SETGID",           // 2.2
+	"CAP_SETUID",           // 2.2
+	"CAP_SETPCAP",          // 2.2
+	"CAP_LINUX_IMMUTABLE",  // 2.2
+	"CAP_NET_BIND_SERVICE", // 2.2
+	"CAP_NET_BROADCAST",    // 2.2
+	"CAP_NET_ADMIN",        // 2.2
+	"CAP_NET_RAW",          // 2.2
+	"CAP_IPC_LOCK",         // 2.2
+	"CAP_IPC_OWNER",        // 2.2
+	"CAP_SYS_MODULE",       // 2.2
+	"CAP_SYS_RAWIO",        // 2.2
+	"CAP_SYS_CHROOT",       // 2.2
+	"CAP_SYS_PTRACE",       // 2.2
+	"CAP_SYS_PACCT",        // 2.2
+	"CAP_SYS_ADMIN",        // 2.2
+	"CAP_SYS_BOOT",         // 2.2
+	"CAP_SYS_NICE",         // 2.2
+	"CAP_SYS_RESOURCE",     // 2.2
+	"CAP_SYS_TIME",         // 2.2
+	"CAP_SYS_TTY_CONFIG",   // 2.2
+	"CAP_MKNOD",            // 2.4
+	"CAP_LEASE",            // 2.4
+	"CAP_AUDIT_WRITE",      // 2.6.11
+	"CAP_AUDIT_CONTROL",    // 2.6.11
+	"CAP_SETFCAP",          // 2.6.24
+	"CAP_MAC_OVERRIDE",     // 2.6.25
+	"CAP_MAC_ADMIN",        // 2.6.25
+	"CAP_SYSLOG",           // 2.6.37
+	"CAP_WAKE_ALARM",       // 3.0
+	"CAP_BLOCK_SUSPEND",    // 3.5
+
+	// caps316 is the caps of kernel 3.16 (38 entries)
+	"CAP_AUDIT_READ",
+
+	// caps58 is the caps of kernel 5.8 (40 entries)
+	"CAP_PERFMON",
+	"CAP_BPF",
+
+	// caps59 is the caps of kernel 5.9 (41 entries)
+	"CAP_CHECKPOINT_RESTORE",
+}
+
+func completeLinuxCapabilityNames(cmd *cobra.Command, args []string, toComplete string) (names []string, _ cobra.ShellCompDirective) {
+	return completion.FromList(allLinuxCapabilities...)(cmd, args, toComplete)
+}
--- a/cli/command/container/completion.go
+++ b/cli/command/container/completion.go
@@ -3,6 +3,7 @@ package container
 import (
 	"github.com/docker/cli/cli/command/completion"
 	"github.com/docker/docker/api/types/container"
+	"github.com/moby/sys/signal"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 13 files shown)


cli/command/container/completion.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/docker/cli/cli/command/completion"
+	"github.com/spf13/cobra"
+)
+
+// allLinuxCapabilities is a list of all known Linux capabilities.
+//
+// This list was based on the containerd pkg/cap package;
+// https://github.com/containerd/containerd/blob/v1.7.19/pkg/cap/cap_linux.go#L133-L181
+//
+// TODO(thaJeztah): add descriptions, and enable descriptions for our completion scripts (cobra.CompletionOptions.DisableDescriptions is currently set to "true")
```
thaJeztah, line 13: Need to check why we disable it here;
https://github.com/docker/cli/blob/9bb1a62735174e9220d84fecc056a0ef8a1fc26f/cmd/docker/docker.go#L92-L96

And here;
https://github.com/docker/cli/blob/9bb1a62735174e9220d84fecc056a0ef8a1fc26f/cli-plugins/plugin/plugin.go#L158-L162

thaJeztah:
Ah! Forgot to remove something I previously used. Also looks like we're running some deprecated linters

```
#16 0.209 level=warning msg="[lintersdb] The linter named \"megacheck\" is deprecated. It has been split into: gosimple, staticcheck, unused."
#16 0.209 level=warning msg="[lintersdb] The name \"vet\" is deprecated. The linter has been renamed to: govet."
#16 65.71 cli/command/completion/functions.go:123:2: ineffectual assignment to toComplete (ineffassign)
#16 65.71 	toComplete = strings.ToLower(toComplete)
#16 65.71 	^

File: cli/command/container/completion.go [hunk 2]
```diff
@@ -0,0 +1,94 @@
...
+func completeLinuxCapabilityNames(cmd *cobra.Command, args []string, toComplete string) (names []string, _ cobra.ShellCompDirective) {
+	return completion.FromList(allLinuxCapabilities...)(cmd, args, toComplete)
+}
+
+func completeRestartPolicies(cmd *cobra.Command, args []string, toComplete string) (names []string, _ cobra.ShellCompDirective) {
+	return completion.FromList(restartPolicies...)(cmd, args, toComplete)
+}
+
+func completeSignals(cmd *cobra.Command, args []string, toComplete string) (names []string, _ cobra.ShellCompDirective) {
+	// TODO(thaJeztah): do we want to provude the full list here, or a subset?
```
Benehiko, line 88:
```suggestion
	// TODO(thaJeztah): do we want to provide the full list here, or a subset?
```

Benehiko (approved):
This looks really good, will be a nice improvement to the CLI UX :)

thaJeztah:
Thanks for reviewing! I'll have a look at fixing the typo when I'm near my computer.

Also; wrt the "disable descriptions" https://github.com/docker/cli/pull/5238#discussion_r1666894131 it looks like buildx DOES enable them for its completion, so maybe we should enable it in our code as well; there may be some commands that could use a slight touch-up for those as some have descriptions that are a bit confusing, so still best to do in a follow up

cli/command/container/completion.go
thaJeztah: Fixed 👍

thaJeztah:
> Thanks for reviewing! I'll have a look at fixing the typo when I'm near my computer.

(I see I didn't comment, but the typo was fixed 😂)

cli/command/container/completion.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/spf13/cobra"
+)
+
+// allLinuxCapabilities is a list of all known Linux capabilities.
+//
+// This list was based on the containerd pkg/cap package;
+// https://github.com/containerd/containerd/blob/v1.7.19/pkg/cap/cap_linux.go#L133-L181
+//
+// TODO(thaJeztah): add descriptions, and enable descriptions for our completion scripts (cobra.CompletionOptions.DisableDescriptions is currently set to "true")
+var allLinuxCapabilities = []string{
```
vvoland, line 16: Looks like containerd/pkg/cap maintains a list of these: https://github.com/containerd/containerd/blob/78d3e205a51ec101f775a43bee6f4fdd8fc6b22b/pkg/cap/cap_linux.go#L133-L187

Perhaps we could help extracting this from `pkg` to a separate repo that we could consume?

(not a blocker for this PR though)

vvoland (approved):
LGTM

cli/command/container/completion.go
thaJeztah: > Looks like containerd/pkg/cap maintains a list of these:

Hehe, yes, look at the comment above; that's where I took the list from 😄.

We _could_ consider moving such things to a separate repo (or to be in a smaller module); one tricky bit there is that these are linux capabilities, so if we want to use them cross-platform (in our case, the CLI may be running on Windows), at least we'd have to make clear that they only apply to Linux, so we can't have that code be platform-specific at compile time (`_linux.go` e.g.).

vvoland: Oh right, I didn't read the comment 🤦🏻 😄

vvoland: Yeah, we can't use it in the current form, but would be nice to avoid having to maintain this list by ourselves in future.

thaJeztah:
Looks like this test is flaky again (but different this time); I opened a tracking ticket; https://github.com/docker/cli/issues/5254

```
#20 60.67 === FAIL: cli/command/container TestInitTtySizeErrors (unknown)
#20 60.67 Error: write /dev/pts/0: file already closed
#20 60.67 panic: Fail in goroutine after TestRunAttachTermination has completed
#20 60.67 
#20 60.67 goroutine 216 [running]:
#20 60.67 testing.(*common).Fail(0xc00021eb60)
#20 60.67 	/usr/local/go/src/testing/testing.go:952 +0xd4
#20 60.67 testing.(*common).FailNow(0xc00021eb60)
#20 60.67 	/usr/local/go/src/testing/testing.go:975 +0x26
#20 60.67 github.com/docker/cli/vendor/gotest.tools/v3/assert.ErrorIs({0x1033010, 0xc00021eb60}, {0x102a300?, 0xc00049a0f0}, {0x102a3e0?, 0x16e0f30}, {0x0, 0x0, 0x0})
#20 60.67 	/go/src/github.com/docker/cli/vendor/gotest.tools/v3/assert/assert.go:311 +0x159
#20 60.67 github.com/docker/cli/cli/command/container.TestRunAttachTermination.func6()
#20 60.67 	/go/src/github.com/docker/cli/cli/command/container/run_test.go:85 +0x85
#20 60.67 created by github.com/docker/cli/cli/command/container.TestRunAttachTermination in goroutine 214
#20 60.67 	/go/src/github.com/docker/cli/cli/command/container/run_test.go:84 +0x486
```

cli/command/container/completion.go
thaJeztah: Yeah agreed. For this one I decided that it's not _critical_ for the list to be up-to-date (even more accurate; completion may provide capabilities that are not supported by the daemon host if it's running an older kernel version), but for completion that should be mostly "OK"; the user doesn't _have_ to use the completion, and can still type something else.

For the `--cap-add` and `--cap-rm` cases, there's also the case that it would provide suggestions for capabilities that are either already set by default, or not present (both of which should be a no-op).

So the most _accurate_ thing to do would be to have API endpoints for these (i.e., let the daemon provide options that can be used) that are specifically crafted for purpose of completion (as lightweight as possible), but even if they're lightweight, there may still be some latency when connected to a remove machine, so that's something we should explore and consider pros/cons.

PR #5244, opened 2024-07-08
Add support for device-code oauth login

laurazard:

**- What I did**

This PR adds support for the oauth [device-code](https://auth0.com/docs/get-started/authentication-and-authorization-flow/device-authorization-flow) login flow when authenticating against the official registry.

**- How I did it**

Added `cli/internal/oauth`, which contains code to manage interacting with the Docker OAuth tenant (`login.docker.com`), including launching the device-code flow, refreshing access using the refresh-token, and logging out.

In order to manage storage and retrieval of the oauth credentials, as well as ensuring that the access token retrieved from the credentials store is valid when accessed, this PR also introduces a new credential store wrapper, `OAuthStore`, which defers storage to an underlying store (such as the file or native/keychain/pass/wincred store) and gets a new access token using the refresh token when accessed.

Also included an (imo overdue 😅) refactor of `cli/command/registry/login.go`.

**- How to verify it**

Try to login, do things that require authentication, and logout!

```markdown changelog
Added support for device-code flow login when authenticating to the official registry.
```

Screenshot 2024-07-08 at 14 08 42


cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,161 @@
...
+	res.Claims = claims
+
+	return res, nil
+}
+
+// Logout logs out of the session for the client and removes tokens from the storage provider.
+func (m *OAuthManager) Logout(ctx context.Context) error {
+	return errors.Join(
+		m.openBrowser(m.api.LogoutURL()),
+	)
```
Benehiko: This doesn't seem correct to me, AFAIK when dealing with OAuth tokens we should revoke them by sending a POST request to the backend directly. Opening the browser will create a GET request. 

From what I could see, the LogoutURL does not contain much information pertaining to this Access/Refresh token and only has the Client ID in the query parameters. `https://login.docker.com/v2/logout?client_id=xxxxxxxx`

The RFC mentions including the token you would like to revoke and the type as query parameters in a POST request. For example:

```
POST /revoke HTTP/1.1
     Host: server.example.com
     Content-Type: application/x-www-form-urlencoded
     Authorization: Basic xxxxxxxx

     token=xxxxxxxx&token_type_hint=refresh_token
```

Please see the RFC on OAuth2 Token Revocation for further information https://datatracker.ietf.org/doc/html/rfc7009

cli/internal/oauth/api/api.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// getToken calls the token endpoint of Auth0 and returns the response.
+func (a API) getDeviceToken(ctx context.Context, state State) (res TokenResponse, err error) {
+	data := url.Values{
+		"client_id":   {a.ClientID},
+		"grant_type":  {"urn:ietf:params:oauth:grant-type:device_code"},
+		"device_code": {state.DeviceCode},
+	}
+	oauthTokenURL := a.BaseURL + "/oauth/token"
```
Benehiko, line 120: **suggestion**: could we maybe do `url.JoinPath` instead?
```suggestion
	oauthTokenURL, err := url.JoinPath(a.BaseURL, "/oauth/token")
```

cli/internal/oauth/api/api.go
```diff
@@ -0,0 +1,156 @@
...
+
+// GetDeviceCode returns device code authorization information from Auth0.
+func (a API) GetDeviceCode(ctx context.Context, audience string) (state State, err error) {
+	data := url.Values{
+		"client_id": {a.ClientID},
+		"audience":  {audience},
+		"scope":     {strings.Join(a.Scopes, " ")},
+	}
+
+	deviceCodeURL := a.BaseURL + "/oauth/device/code"
```
Benehiko, line 57: **suggestion**: could we maybe do `url.JoinPath` instead?
```suggestion
	deviceCodeURL, err := url.JoinPath(a.BaseURL, "/oauth/device/code")
```

cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,161 @@
...
+
+	if state.UserCode == "" {
+		return res, errors.New("login failed: no user code returned")
+	}
+
+	_, _ = fmt.Fprintln(w, "\nYou will be signed in using a web-based login.")
+	_, _ = fmt.Fprintln(w, "To sign in with credentials on the command line, use 'docker login -u <username>'")
+	_, _ = fmt.Fprintf(w, "\nYour one-time device confirmation code is: %s\n", state.UserCode)
+	_, _ = fmt.Fprint(w, "\nPress ENTER to open the browser.\n")
+	_, _ = fmt.Fprintf(w, "Or open the URL manually: %s.\n", strings.Split(state.VerificationURI, "?")[0])
```
Benehiko, line 78:
```suggestion
	_, _ = fmt.Fprintf(w, "Or open the URL manually: %s\n", strings.Split(state.VerificationURI, "?")[0])
```

My terminal emulator included the `.` when clicking on the link.

cli/internal/oauth/api/api.go
laurazard: Why? It's slower than string concatenation and for no benefit here.

cli/internal/oauth/util/util.go
```diff
@@ -0,0 +1,23 @@
...
+	"errors"
+	"os/exec"
+	"runtime"
+)
+
+// OpenBrowser opens the specified URL in a browser based on OS.
+func OpenBrowser(url string) (err error) {
+	switch runtime.GOOS {
+	case "linux":
+		err = exec.Command("xdg-open", url).Start()
```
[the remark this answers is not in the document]
laurazard: I'm not too concerned about this because this is a nice to have, and there's always the fallback of the user opening it directly, so I'd rather not add more dependencies just for this.

I'll take a look at how it's done there and see if it's something we can add without a lot more vendored code though.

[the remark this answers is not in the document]
Benehiko: There's no need to vendor more code for linux, it's just a for loop trying different commands. Windows might require vendoring.

cli/config/credentials/oauth_store.go
```diff
@@ -0,0 +1,151 @@
...
+
+	if _, ok := allAuths[registry.IndexServer]; !ok {
+		return allAuths, nil
+	}
+
+	auth, err := o.Get(registry.IndexServer)
+	if err != nil {
+		return nil, err
+	}
+	allAuths[registry.IndexServer] = auth
```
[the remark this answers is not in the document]
laurazard: We want to get all the credentials, so we fetch all credentials from the backing store. If there are no configured credentials for the official registry, we can return them as is. Otherwise, we call `Get` to get that specific one, so that we can make sure the retrieved credentials are refreshed/all the same logic applies as a normal `Get`.

I'll try to document it a bit more.

It's also usually good practice to look at the tests to see if they provide you clarity on what the requirements for a piece of code are.

cli/internal/oauth/manager/manager.go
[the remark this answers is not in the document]
laurazard: You would have both the manager and the credentials store storing the same information? What's the source of truth in that case? This is a really bad approach IMO.

cli/config/credentials/oauth_store.go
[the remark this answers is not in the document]
Benehiko: Thank you!

>It's also usually good practice to look at the tests to see if they provide you clarity on what the requirements for a piece of code are.

Yes I agree, although in the context of security we should be mindful to keep code as clear as possible.

cli/internal/oauth/api/api.go
Benehiko: This isn't a blocker to me, but thought I'd call it out. Since `BaseURL` is exported, we don't _know_ if it is formatted correctly. 

My thinking around this is more to keep the URL from becoming malformed by accident (for whatever reason). I guess here we don't read `BaseURL` from a config file or Env file, but if we did we could end up with a malformed URL without noticing it.

cli/config/credentials/oauth_store.go
```diff
@@ -0,0 +1,151 @@
...
+	tokenRes, err := o.parseToken(auth.Password)
+	// if the password is not a token, return the auth config as is
+	if err != nil {
+		//nolint:nilerr
+		return auth, nil
+	}
+
+	// if the access token is valid for less than 50 minutes, refresh it
+	if tokenRes.RefreshToken != "" && tokenRes.Claims.Expiry.Time().Before(time.Now().Add(minimumTokenLifetime)) {
+		refreshRes, err := o.manager.RefreshToken(context.TODO(), tokenRes.RefreshToken)
```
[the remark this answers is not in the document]
laurazard: The credential stores (and their signatures, which are exported) don't currently take a context.

If I was writing this from scratch, I'd have made that choice, but given that this PR focused on adding a new feature (device-code flow login) I'd rather keep the discussion here related to the feature at hand and not to overall improvements to the codebase.

It **does** make sense to add a timeout to this context though, so I'll make that change.

cli/internal/oauth/manager/manager.go
twelsh-aw: auth0 only supports GET logouts unfortunately, https://auth0.com/docs/api/authentication#logout  

They (and by proxy us) have accepted CSRF risk on logout. In theory it could maybe be chained as part of larger exploit, but on it's own it's NOT beneficial to an attacker. We don't have a lot of flexibility here unfortunately

Benehiko: @twelsh-aw 
that's quite unfortunate.
I'm assuming that the `GET` request must go through the browser since the session cookie is used as an identifier to invalidate the tokens (since only the `client_id` is used in the the URL query parameters)?

Benehiko: Looking at the auth0 documentation they do indeed have a token revocation endpoint https://auth0.com/docs/api/authentication#revoke-refresh-token. Why can't we use that?

twelsh-aw: good catch. Pinged @technicallyjosh to see if they know of what differences exist between that call and traditional logout (and if we can/should do both). Might need to ask auth0 as well for some clarification

twelsh-aw: and to first question, yes should be browser based to logout in auth0 (and not just clear tokens client side). Can see how [desktop](https://github.com/docker/pinata/blob/52e1290ddcbe354e71af033e7f214b58d9b9fb61/common/cmd/com.docker.backend/internal/auth/authentication/client.go#L196-L199) does it. That being said, I now have questions about what auth0 logout actually does behind the scenes.... /oauth/revoke might be more useful to us 🤔

Benehiko: Yes in OAuth flows we should not concern ourselves with the Identity provider session - in this case the session that exists inside the user's browser. 

The only way to securely dispose of the access and refresh token is to call the `oauth/revoke` API.

Benehiko: This will be a bit verbose, but please bear with me :bear: 

To clarify, the OAuth flow deals with the Authorization server (Third Party) (providing us the access/refresh token) whilst the Identity provider deals with the user's identity (First Party).

Let's assume our example is something like a user logging in via Google to our application using OAuth. Google has to first validate and issue a session to the user (First Party), the user then authorizes the use of their data with our app (Third Party). To seal the deal between our application and Google we use access/refresh tokens which temporarily grants us the right of fetch more of the user's data from Google. 

Once the user is done using our application, and they click "logout" we would need to dispose of the access/refresh token for them, so we call Google to get rid of it. However, we aren't concerned about the current session that exists between the user and Google (browser cookie) - in-fact we shouldn't. 

TL;DR the CLI should not concern itself with the Identity services' session.

twelsh-aw: Yes big THANK YOU for calling this out. Confirmed that /logout is NOT revoking refresh token auth0 side 😬 So we should absolutely be doing that POST /oauth/revoke (and I will need to get Desktop to do that as well)

I am wondering if it still makes sense to do the web /logout in addition to that though 🤔. Not with the "federated" param (which would attempt to log out of social/idp) but the "ordinary one" that desktop is currently doing to clear us auth0 side. 

If we do only the token /revoke and try to login again, I assume we would not be prompted for new login credentials to auth0. Can test all this but hunch is we might want both.

twelsh-aw: I suppose could argue we don't want CLI to do that auth0 bit. it's not super important to me but would be a differentiator from Desktop and Web (where logout is more "unified" in that auth0 sense).

The more I think about it, the more i think NOT doing it from CLI is probably the right call UX wise ha 👍

cli/internal/oauth/util/util.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"errors"
+	"os/exec"
+	"runtime"
+)
+
+// OpenBrowser opens the specified URL in a browser based on OS.
+func OpenBrowser(url string) (err error) {
+	switch runtime.GOOS {
+	case "linux":
```
jpbriend, line 12: I suspect WSL is not covered (`xdg-open` will not work on WSL2).
In Docker Desktop, we use `wslview`: https://github.com/docker/pinata/blob/675a1c7d8ae965bb44a0679ff9ff7a108b82b9e0/common/cmd/com.docker.backend/internal/auth/browser/browser.go#L27
(notice the test before to determine if the current environment is a real linux or WSL).

cli/internal/oauth/util/util.go
```diff
@@ -0,0 +1,23 @@
...
+		err = exec.Command("xdg-open", url).Start()
+	case "windows":
+		err = exec.Command("rundll32", "url.dll,FileProtocolHandler", url).Start()
```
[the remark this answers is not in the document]
laurazard: https://github.com/docker/pinata/blob/675a1c7d8ae965bb44a0679ff9ff7a108b82b9e0/common/cmd/com.docker.backend/internal/auth/browser/browser.go#L27 This is what we do in piñata as well.

cli/internal/oauth/manager/manager.go
laurazard: Thanks! Fixed.

cli/internal/oauth/util/client.go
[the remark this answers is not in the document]
laurazard: This was just code I hadn't cut out yet/had been copy pasted over. Thanks for the catch!

cli/internal/oauth/manager/manager.go
laurazard: IMO, we likely do want to do the `/revoke` request, but I wanted to add some context re: what "logout" means in the CLI.

Prior to this change, "login" is storing either a password or PAT inside either a plain text file or the OS's credentials helpers. "logout" up until now is not revoking that PAT, or doing anything with the tenant, it's solely erasing those saved credentials from the store.

Hence, whether we're revoking the refresh token, or launching the user's browser to end their session with OAuth0, this is already an improvement (security wise) on the previous situation.

Ultimately, I'm happy to only have the CLI do the token revocation and not open the user's browser to end their session, this seems better UX-wise and we can leave that more "unified" experience to other platforms (as you said).

laurazard: And replying to the first comment:
> This doesn't seem correct to me, AFAIK when dealing with OAuth tokens we should revoke them by sending a POST request to the backend directly. Opening the browser will create a GET request.

I think it's become clear in the meantime, but in case it's not, what we were doing here was opening the user's browser to end their session with auth0, not revoking the refresh token.

cli/internal/oauth/util/browser.go
```diff
@@ -0,0 +1,37 @@
+package util
+
+import (
+	"errors"
+	"os/exec"
+	"runtime"
+	"strings"
+)
+
+// OpenBrowser opens the specified URL in a browser based on OS.
```
thaJeztah, line 10: Just a really short blurb; wondering if we looked at https://github.com/pkg/browser - I know that one is used already by various dependencies we use for authentication purposes, so may be battle tested (I'm not sure how much different it is from the code here though)

cli/internal/oauth/util/browser.go
```diff
@@ -0,0 +1,37 @@
...
+		return exec.Command("open", url).Start()
+	default:
+		return errors.New("unsupported platform")
+	}
+}
+
+func openBrowserLinux(url string) error {
+	providers := []string{"wslview", "xdg-open", "x-www-browser", "www-browser"}
+
+	// There are multiple possible providers to open a browser on linux
```
thaJeztah, line 27: Oh! Or was this copied from there? I just noticed a very similar comment in that code; https://github.com/pkg/browser/blob/5ac0b6a4141c771a0d3a081c36b87c977cf9c7db/browser_linux.go#L11


If we copied, we must probably include license and copyright from there, otherwise we're violating the license; https://github.com/pkg/browser/blob/5ac0b6a4141c771a0d3a081c36b87c977cf9c7db/LICENSE#L10


(but maybe using the dependency would be a solution for that if we can use it as-is)

laurazard: For sure need to include the copyright notice, but I wonder if just using the dependency would work for [WSL](https://github.com/docker/cli/pull/5244#discussion_r1677848277) (no `wslview` in that code). OTOH, maybe it doesn't need to, according to this [post](https://askubuntu.com/questions/1409481/why-does-xdg-open-use-wslview) `xdg-open` would probably end up calling `wslview` under the hood.

laurazard: I'll push a version vendoring `pkg/browser`

cli/internal/oauth/util/util.go
laurazard: More discussion about this [here](https://github.com/docker/cli/pull/5244#discussion_r1682401356).

cli/internal/oauth/util/browser.go
thaJeztah: Thanks! Yeah, you know I usually try to reduce dependencies, but I know it's already in our tree indirectly (in some cases), so using the dependency instead of maintaining our own fork of the code could be a good thing (assuming that works).

We can still decide to maintain a fork if there's a need (well, or contribute to upstream if it's something that would be generally useful to improve in that module)

cli/config/configfile/file.go
```diff
@@ -254,10 +255,13 @@ func decodeAuth(authStr string) (string, string, error) {
...
+	var credsStore credentials.Store
 	if helper := getConfiguredCredentialStore(configFile, registryHostname); helper != "" {
-		return newNativeStore(configFile, helper)
+		credsStore = newNativeStore(configFile, helper)
+	} else {
+		credsStore = credentials.NewFileStore(configFile)
 	}
-	return credentials.NewFileStore(configFile)
+	return credentials.NewOAuthStore(credsStore, manager.NewManager())
 }
```
thaJeztah, line 265: This is not something new, but I think your PR makes it more apparent, and ... maybe I'm rambling; I need to have a closer look; I'm somewhat starting to consider if constructing these stores should happen here, or if they should be constructed when initialising the configfile.

Basically; I need to look where this code is used; it may be imported elsewhere, which can be in code that's "non-interactive" (say k8s); for those, getting credentials through the OAuth flow isn't an option.

So I'm thinking if whether or not to use the interactive flow is something that must be set when constructing all of this 🤔

thaJeztah (commented):
Not a full review; just a bunch of random ramblings 🙈

cli/oauth/jwt.go
```diff
@@ -0,0 +1,115 @@
...
+		return
+	}
+
+	err = token.UnsafeClaimsWithoutVerification(&claims)
+
+	return
+}
+
+func ConcatTokens(accessToken, refreshToken string) string {
+	return accessToken + ".." + refreshToken
```
vvoland, line 105: Are we sure storing the concatenated tokens under the registry URL key is the best idea?

IIRC, we talked about storing the refresh token as a separate entry in the credential store - this way older clients could still pick up the actual auth token.

Benehiko: I think concatenating the tokens is valid and more useful. The oauth token will be useless anyways (1 hour validity) for older clients. I don't think we should try keep backwards compatibility like, rather migrate and force to update.

cli/command/registry/login.go
```diff
@@ -100,53 +101,135 @@ func verifyloginOptions(dockerCli command.Cli, opts *loginOptions) error {
...
+	}, tokenRes.RefreshToken, nil
+}
+
+func storeCredentials(dockerCli command.Cli, authConfig registrytypes.AuthConfig) error {
+	creds := dockerCli.ConfigFile().GetCredentialsStore(authConfig.ServerAddress)
 	store, isDefault := creds.(isFileStore)
 	// Display a warning if we're storing the users password (not a token)
-	if isDefault && authConfig.Password != "" {
-		err = displayUnencryptedWarning(dockerCli, store.GetFilename())
+	if isDefault && store.IsFileStore() && authConfig.Password != "" {
```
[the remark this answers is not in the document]
thaJeztah: > We should really move this part internal to the fileStore itself; I'll have a quick look later to see what would be involved in doing so.

- opened https://github.com/docker/cli/pull/5259

cli/config/credentials/oauth_store.go
```diff
@@ -0,0 +1,160 @@
...
+
+const minimumTokenLifetime = 50 * time.Minute
+
+// Get retrieves the credentials from the backing store, refreshing the
+// access token if the stored credentials are valid for less than minimumTokenLifetime.
+// If the credentials being retrieved are not for the official registry, they are
+// returned as is. If the credentials retrieved do not parse as a token, they are
+// also returned as is.
+func (o *oauthStore) Get(serverAddress string) (types.AuthConfig, error) {
+	if serverAddress != registry.IndexServer {
```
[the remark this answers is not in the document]
laurazard: Yup! That makes sense, I'll move it for now. I had a very vague idea in my head about this (I wondered why we were passing `isDefaultRegistry` in as a bool instead of checking here and checked commit history) and then forgot to ask about it 😅

cli/oauth/jwt.go
thaJeztah: 👋 just a quick blurb; if this is what will be stored in "~/.docker/config.json", we must be very careful on the format. Updating unfortunately is not something we can require, because the file is no longer exclusive to us; there's many tools that know about the config file and that use it (kubernetes, jenkins, IDEs) .. basically; "the world", and unfortunately not all of those can be expected to be updated.

(LOL, it took 8 Years before we got rid of the old config file; and even after that, I got SPAMMED by users that we broke them 🫠🫠🫠🫠 https://github.com/docker/cli/pull/4281)

laurazard: I'm okay with not doing this (and instead storing the "usable" token only, and storing the refresh token elsewhere). 

Indeed, this rationale:
>  there's many tools that know about the config file and that use it (kubernetes, jenkins, IDEs) .. basically; "the world", and unfortunately not all of those can be expected to be updated.

Was exactly what I initially outlined as the reason to instead store the usable token there, and make sure that we always refresh that with a fresh access token on CLI startup, so that anyone accessing it (without updates) can continue using it.

However, we wouldn't be changing the "format" of the `config.json`, just what we put inside the fields we already have there (and other applications/"the world" are unlikely to be relying on credentials being there because for a lot of users these would have been stored in credential helpers/keychain/etc.), so I think we should be fine in this regard.

cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// LoginDevice launches the device authentication flow with the tenant, printing
+// instructions to the provided writer and attempting to open the browser for
+// the user to authenticate.
+// Once complete, the retrieved tokens are returned.
+func (m *OAuthManager) LoginDevice(ctx context.Context, w io.Writer) (res oauth.TokenResult, err error) {
+	state, err := m.api.GetDeviceCode(ctx, m.audience)
+	if err != nil {
+		return res, fmt.Errorf("login failed: %w", err)
```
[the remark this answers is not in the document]
laurazard: Agree this needed a refactor! I did away with the named returns, and ended up returning a pointer. I'm happy with the first change, unsure about the 2nd, but it looks better than before!

cli/internal/oauth/api/api_test.go
```diff
@@ -0,0 +1,238 @@
...
+	t.Run("success", func(t *testing.T) {
+		expectedToken := TokenResponse{
+			AccessToken:  "a-real-token",
+			IDToken:      "",
+			RefreshToken: "the-refresh-token",
+			Scope:        "",
+			ExpiresIn:    3600,
+			TokenType:    "",
+		}
+		var respond bool
```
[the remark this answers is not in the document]
laurazard: Thanks!

cli/config/configfile/file.go
laurazard: >  for those, getting credentials through the OAuth flow isn't an option

True, but as you said, it's not different/new here. With the "old" flow, if there are no credentials, we also can't prompt the user to input them if running non-interactive. And if the user does have "oauth" credentials stored, even in a non-interactive environment, we want the oauth codepath to be executed so that the CLI can use the refresh token to fetch a fresh access token before returning them to whoever is calling it, right?

Basically, what setting this when initializing let us do/decide that we can't already?

cli/internal/oauth/api/api_test.go
```diff
@@ -0,0 +1,238 @@
...
+		go func() {
+			time.Sleep(10 * time.Second)
+			respond = true
+		}()
```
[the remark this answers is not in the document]
laurazard: Yeah! I thought it'd be good to keep a test with "somewhat realistic" settings, and it might take a user a few seconds to log in. Will mark it parallel though.

cli/oauth/jwt.go
Benehiko: Even if the access token is stored separate from the refresh token, it would be partially useless to third parties relying on it. The PAT was long lived, while the access token expires only after an hour. 

We would then need to give out a deprecation notice, even if we only store the access token in-place of the PAT in the config file.

thaJeztah: Yes, in a really ideal world, the plain-text store would be treated as a credentials-store, and store credentials separate from the config-file (which would also mean that for most situations the cli-config file can be immutable).

But we must be sure that whatever value we'll store in the config-file is still compatible, as it will be used by other tools.

thaJeztah: (it _may_ (but need to have a good look) be better / more compatible to have _NO_ credentials stored in the config than a value that breaks compatibility)

laurazard: In that kubernetes link/scenario, non-file storage is already not supported, and they suggest [creating a secret to store the credentials](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-secret-by-providing-credentials-on-the-command-line).

I guess we could take a couple of approaches – we can either only support non-file store scenarios with the oauth login, or we can do away with this implementation entirely and implement this on the side of the credential helpers.

I guess we could also store this under a different key (`/access-token` and `/refresh-token`), but regardless of how we do this, it won't be compatible with other clients as the value we store there will only be valid for a short amount of time.

laurazard: > (it may (but need to have a good look) be better / more compatible to have NO credentials stored in the config than a value that breaks compatibility)

I guess if we go back to using the specific keys (`index.docker.io.../access-token` and `index.docker.io.../refresh-token` this might be somewhat mitigated?

thaJeztah (commented):
Left some blurbs; also playing with some options in those areas.

Overall (from a user perspective) it all works pretty neat. Some notes I had while playing;

I think we may be missing some normalization of registry-names still (likely not something new); for example;

```bash
docker login registry-1.docker.io
Username:
```

But `docker.io` works;

```bash
docker login docker.io

You will be signed in using a web-based login.
To sign in with credentials on the command line, use 'docker login -u <username>'
...
```

Probably not problematic, but I noticed that a mix of "logout" using the old CLI, and login using the "new" CLI showed this;

```bash
# this is the old CLI:
docker logout
Removing login credentials for https://index.docker.io/v1/

# login with the new CLI:
./build/docker login
Authenticating with existing credentials...

You will be signed in using a web-based login.
To sign in with credentials on the command line, use 'docker login -u <username>'
...
```

Notice that it prints `Authenticating with existing credentials...`; I think what happens is that the old CLI only removes credentials it knows about, but then leaves the new ones. The new CLI (probably) sees that there's credentials, but then they don't work, so it continues the new path to interactively authenticate. Definitely not problematic AFAICS, just thinking if we can detect that situation.

Also had a try what happens if hub is not accessible when logging out / in (making sure we don't fully hang);

```bash
docker login docker.io

# You will be signed in using a web-based login.
# ...

# block login.docker.com
echo '127.0.0.1 login.docker.com' >> /etc/hosts

docker logout
Removing login credentials for https://index.docker.io/v1/

docker login
login failed: Post "https://login.docker.com/oauth/device/code": dial tcp 127.0.0.1:443: connect: connection refused
```

Slightly wondering if we should print a message if logout failed; we continue removing the credentials from the store, but `RevokeToken` would've failed; is that something important to log (i.e., if the token was found by someone, it could still be valid?).

thaJeztah:
Oh! Erm, this one I forgot to post; something I didn't consider in my changes I guess; we store multiple times now, and each of them prints a warning :see_no_evil:

```bash
docker login

You will be signed in using a web-based login.
To sign in with credentials on the command line, use 'docker login -u <username>'

Your one-time device confirmation code is: XXXXXX

Press ENTER to open the browser.
Or open the URL manually: https://login.docker.com/activate

Waiting for authentication in the browser...

WARNING! Your credentials are stored unencrypted in '/root/.docker/config.json'.


WARNING! Your credentials are stored unencrypted in '/root/.docker/config.json'.

Login Succeeded
```

This one was slightly more surprising; after Having signed in, running `docker login` does succeed (`Authenticating with existing credentials...`), but also prints _three_ warnings.

```bash
docker login

WARNING! Your credentials are stored unencrypted in '/root/.docker/config.json'.


WARNING! Your credentials are stored unencrypted in '/root/.docker/config.json'.

Authenticating with existing credentials...

WARNING! Your credentials are stored unencrypted in '/root/.docker/config.json'.

Login Succeeded
```

cli/config/configfile/file.go
```diff
@@ -254,10 +255,13 @@ func decodeAuth(authStr string) (string, string, error) {
...
 func (configFile *ConfigFile) GetCredentialsStore(registryHostname string) credentials.Store {
+	var credsStore credentials.Store
 	if helper := getConfiguredCredentialStore(configFile, registryHostname); helper != "" {
-		return newNativeStore(configFile, helper)
+		credsStore = newNativeStore(configFile, helper)
+	} else {
+		credsStore = credentials.NewFileStore(configFile)
 	}
-	return credentials.NewFileStore(configFile)
+	return credentials.NewOAuthStore(credsStore, manager.NewManager())
```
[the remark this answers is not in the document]
laurazard: Hmm, I find it easier to reason about if we have a single point managing both. Otherwise, if someone does `GetCredentialsStore("some-other-registry").GetAll()` they're going to get the `access-token` and `refresh-token` specific keys that we shouldn't leak out/callers shouldn't concern themselves with. There might be other things that the wrapper does that should be abstracted away by the wrapper.

cli/command/cli.go
```diff
@@ -293,6 +301,8 @@ func (cli *DockerCli) Initialize(opts *cliflags.ClientOptions, ops ...CLIOption)
 		cli.createGlobalTracerProvider(cli.baseCtx)
 	}
 
+	cli.oauthManager = manager.NewManager()
```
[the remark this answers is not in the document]
laurazard: Nice catch! In an earlier implementation this wasn't possible (due to circular imports), and I forgot about it, but this is a lot nicer!

cli/command/registry/login.go
```diff
@@ -79,70 +80,150 @@ func verifyloginOptions(dockerCli command.Cli, opts *loginOptions) error {
...
 
-		response, err = clnt.RegistryLogin(ctx, authConfig)
-		if err != nil && client.IsErrConnectionFailed(err) {
-			// If the server isn't responding (yet) attempt to login purely client side
-			response, err = loginClientSide(ctx, authConfig)
-		}
-		// If we (still) have an error, give up
-		if err != nil {
-			return err
+	if response.Status != "" {
```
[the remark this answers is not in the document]
laurazard: I'm not sure I could see this happening (even if we fail logging in with the stored creds here – https://github.com/laurazard/cli/blob/9ef2396083d813b7672d0fa8546db8ceddeda8a5/cli/command/registry/login.go#L101, we try here –https://github.com/laurazard/cli/blob/9ef2396083d813b7672d0fa8546db8ceddeda8a5/cli/command/registry/login.go#L107, and if we fail there we return – https://github.com/laurazard/cli/blob/9ef2396083d813b7672d0fa8546db8ceddeda8a5/cli/command/registry/login.go#L108-L110, but not harm in adding a check.

laurazard:
Re: the warning now printing multiple times, I'm wondering how to address that. We store twice (for both keys, now) and on login, since we refresh the token and store the new one.

laurazard:
> Slightly wondering if we should print a message if logout failed; we continue removing the credentials from the store, but RevokeToken would've failed; is that something important to log (i.e., if the token was found by someone, it could still be valid?).

This is a good question – it kind of sucks in this case since we'd be removing the token, effectively making it so that the user no longer has the token to revoke it. Definitely should log a message in this case.

laurazard:
@thaJeztah also wondering about how we'll fix the multiple file_store warnings, now that the warnings live in the store 😅 Ideas?

cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,150 @@
...
+
+	if state.UserCode == "" {
+		return nil, errors.New("login failed: no user code returned")
+	}
+
+	_, _ = fmt.Fprintln(w, "\nYou will be signed in using a web-based login.")
+	_, _ = fmt.Fprintln(w, "To sign in with credentials on the command line, use 'docker login -u <username>'")
+	_, _ = fmt.Fprintf(w, "\nYour one-time device confirmation code is: %s\n", state.UserCode)
+	_, _ = fmt.Fprint(w, "\nPress ENTER to open the browser.\n")
+	_, _ = fmt.Fprintf(w, "Or open the URL manually: %s\n", strings.Split(state.VerificationURI, "?")[0])
```
dvdksn, line 77: What do you think about this:

```suggestion
	_, _ = fmt.Fprintf(w, "Submit your device code here: %s\n", strings.Split(state.VerificationURI, "?")[0])
	_, _ = fmt.Fprint(w, "\n\033[1mPress ENTER\033[0m to open your browser…\n")
```

laurazard: Screenshot 2024-08-05 at 13 51 22

I wonder if this is too many ellipsis(es) now 😅

dvdksn: Oh right, there was one already. 😅 Agree that we should only have one. Let's skip the one I added.

dvdksn: But let's make it an actual ellipsis :D

cli/config/credentials/oauth_store.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	refreshTokenAuth, err := o.backingStore.Get(refreshTokenServerAddress)
+	if err == nil && refreshTokenAuth.Password != "" {
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		err = o.manager.Logout(ctx, refreshTokenAuth.Password)
+		if err != nil {
+			// todo(laurazard): actual message here
+			fmt.Fprint(os.Stderr, "Failed to revoke refresh token with tenant.. Credentials will still be erased.")
```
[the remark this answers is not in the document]
laurazard: @dvdksn thoughts on this one?

cli/internal/oauth/manager/manager.go
laurazard: Sounds good!

cli/config/credentials/oauth_store.go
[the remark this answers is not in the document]
dvdksn: What is the user action that leads here? `docker logout myregistry.example.com`?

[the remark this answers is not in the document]
laurazard: or just `docker logout`, yes.

cli/config/credentials/oauth_store.go
```diff
@@ -0,0 +1,206 @@
...
+		return o.backingStore.Get(serverAddress)
+	}
+
+	tokenRes := o.fetchFromBackingStore()
+	if tokenRes == nil {
+		return o.backingStore.Get(serverAddress)
+	}
+
+	// if the access token is valid for less than minimumTokenLifetime, refresh it
+	if tokenRes.RefreshToken != "" && tokenRes.Claims.Expiry.Time().Before(time.Now().Add(minimumTokenLifetime)) {
```
vvoland, line 51: Looks like `Expiry` is `*NumericData` so it can be `nil`.
Should we do a nil-check just in case so we don't panic?

cli/config/credentials/oauth_store.go
vvoland, line 120:
```suggestion
			_, _ = fmt.Fprint(os.Stderr, "Failed to revoke refresh token with tenant.. Credentials will still be erased.")
```
(to be consistent with other Fprints)

cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,150 @@
...
+		if err != nil {
+			waitForTokenErrChan <- err
+			return
+		}
+		tokenResChan <- tokenRes
+	}()
+
+	go func() {
+		reader := bufio.NewReader(os.Stdin)
+		reader.ReadString('\n')
```
vvoland, line 92:
```suggestion
		_, _ = reader.ReadString('\n')
```

We purposefully ignore the read value, so let's also express that in Go itself 😅

cli/internal/oauth/manager/manager_test.go
```diff
@@ -0,0 +1,277 @@
...
+	t.Run("canceled context", func(t *testing.T) {
+		getDeviceToken := func(audience string) (api.State, error) {
+			return api.State{
+				DeviceCode: "device-code",
+				UserCode:   "0123-4567",
+			}, nil
+		}
+		waitForDeviceToken := func(state api.State) (api.TokenResponse, error) {
+			// make sure that the context is cancelled before this returns
+			time.Sleep(500 * time.Millisecond)
```
vvoland, line 127: Possibly we could move the `ctx, cancel := context.WithCancel(context.Background())` earlier and then just:
```suggestion
            select {
			case <-ctx.Done():
			case <-time.After(time.Second):
			}
```

so we don't need to always wait for the 500ms?

laurazard: The issue is in https://github.com/laurazard/cli/blob/6a838e2625c2e4246e76dbd6fbda52634917e5fa/cli/internal/oauth/manager/manager.go#L98-L104 – if both channels have an item there, `select` will take one at random. Cancelling earlier doesn't fix it, we just need to make sure it returns after (or never would be an option too).

laurazard: maybe the comment wasn't clearer, it's not to ensure the context is cancelled, it's to ensure that the `select` statement selects from the context.Done channel and not from the "result" channel.

vvoland: Ah okay! Also looks like it will be called in a separate goroutine anyway, so it won't stall the main test goroutine, so I'm fine 👍🏻

cli/config/credentials/oauth_store.go
dvdksn: What about making this a warning? I don't recall if we have a helper function for styling, but something like:

```suggestion
			_, _ = fmt.Fprint(os.Stderr, "Warning: You have been logged out successfully, but there was a failure to revoke the OAuth refresh token with the tenant.")
```

The only problem with the above is that it doesn't really say anything about manually addressing that failure. A more complete, but unattractively longer, version could be:

```suggestion
			_, _ = fmt.Fprint(os.Stderr, "Warning: You have been logged out successfully, but there was a failure to revoke the OAuth refresh token with the tenant. You may want to manually revoke the token through your OAuth provider's settings.")
```

laurazard: I don't think the `You have been logged out successfully` works super well with the messages we have [here](https://github.com/docker/cli/blob/0bac5a707b23b588a379ca1a05a0bf5a77155c32/cli/command/registry/logout.go#L66) and [here](https://github.com/docker/cli/blob/0bac5a707b23b588a379ca1a05a0bf5a77155c32/cli/command/registry/logout.go#L56).

In those, we don't say "logging out", but rather `could not erase credentials` or `Removing login credentials`. @dvdksn

dvdksn: Wasn't aware of the prior art 👍🏻 In that case `s/You have been logged out successfully/Credentials erased successfully/` for consistency.

Issue #5242, opened 2024-07-08
Clarify how to resolve pushing a single platform from a multi-platform image index


ESKYoung:
### Description

`docker push` [note is unclear, and refers to an unknown `docker convert` command](https://github.com/docker/cli/blob/9bb1a62735174e9220d84fecc056a0ef8a1fc26f/cli/command/image/push.go#L82).

I'm running Lima on macOS to build `linux/amd64` images, and this note appears for each push.

Might be more useful to explain how to resolve this, as it's unclear initially to me how to fix this. Looking [back through the relevant PR](https://github.com/docker/cli/pull/4984), it states that it now expects a `--platform` flag when running `docker pull` or `docker push`.

### Reproduce

1. create a simple Dockerfile
   ```dockerfile
   FROM python:3.10-slim
   CMD ["/bin/bash"]
   ```
2. `docker build . -t example:latest`
3. `docker push example:latest`

### Expected behavior

Would expect the note to at least state using the `--platform` flag, or if there is a system-wide workaround for this as well.

### docker version

```bash
Client: Docker Engine - Community
 Version:           27.0.3
 API version:       1.46
 Go version:        go1.22.4
 Git commit:        7d4bcd863a
 Built:             Fri Jun 28 14:56:30 2024
 OS/Arch:           darwin/arm64
 Context:           lima-default

Server: Docker Engine - Community
 Engine:
  Version:          27.0.3
  API version:      1.46 (minimum version 1.24)
  Go version:       go1.21.11
  Git commit:       662f78c
  Built:            Sat Jun 29 00:02:23 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.18
  GitCommit:        ae71819c4f5e67bb4d5ae76a6b735f29cc25774e
 runc:
  Version:          1.7.18
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
 rootlesskit:
  Version:          2.0.2
  ApiVersion:       1.1.1
  NetworkDriver:    slirp4netns
  PortDriver:       builtin
  StateDir:         /run/user/502/dockerd-rootless
 slirp4netns:
  Version:          1.2.1
  GitCommit:        09e31e92fa3d2a1d3ca261adaeb012c8d75a8194
```

### docker info

```bash
Client: Docker Engine - Community
 Version:    27.0.3
 Context:    lima-default
 Debug Mode: false
 Plugins:
[...]

PR #5246, opened 2024-07-08
push: Don't default to `DOCKER_DEFAULT_PLATFORM`, improve message

vvoland:
- fixes: https://github.com/docker/cli/issues/5242

Don't default the value of `--platform` flag to `DOCKER_DEFAULT_PLATFORM` environment variable and adjust the note.

```markdown changelog
containerd integration: Fix `docker push` defaulting the `--platform` flag to a value of `DOCKER_DEFAULT_PLATFORM` environment variable on unsupported API versions.
```

### DIFF
--- a/cli/command/image/push.go
+++ b/cli/command/image/push.go
@@ -84,7 +84,7 @@ func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error
 
 		printNote(dockerCli, `Selecting a single platform will only push one matching image manifest from a multi-platform image index.
 This means that any other components attached to the multi-platform image index (like Buildkit attestations) won't be pushed.
-If you want to only push a single platform image while preserving the attestations, please use 'docker convert\n'
+If you want to push a whole multi-platform image, make sure all image content is present and remove the --platform flag.
 `)
 	}
 
--- a/cli/command/image/push.go
+++ b/cli/command/image/push.go
@@ -8,7 +8,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"os"
 
 	"github.com/containerd/platforms"
 	"github.com/distribution/reference"
@@ -58,7 +57,11 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
 	flags.BoolVarP(&opts.all, "all-tags", "a", false, "Push all tags of an image to the repository")
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress verbose output")
 	command.AddTrustSigningFlags(flags, &opts.untrusted, dockerCli.ContentTrustEnabled())
-	flags.StringVar(&opts.platform, "platform", os.Getenv("DOCKER_DEFAULT_PLATFORM"),
+
+	// Don't default to DOCKER_DEFAULT_PLATFORM env variable, always default to
+	// pushing the image as-is. This also avoids forcing the platform selection
+	// on older APIs which don't support it.
+	flags.StringVar(&opts.platform, "platform", "",
 		`Push a platform-specific manifest as a single-platform image to the registry.
 'os[/arch[/variant]]': Explicit platform (eg. linux/amd64)`)
 	flags.SetAnnotation("platform", "version", []string{"1.46"})
--- a/cli/command/image/push.go
+++ b/cli/command/image/push.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/containerd/platforms"
 	"github.com/distribution/reference"
@@ -63,6 +64,7 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
 	// on older APIs which don't support it.
 	flags.StringVar(&opts.platform, "platform", "",
 		`Push a platform-specific manifest as a single-platform image to the registry.
+Image index won't be pushed, meaning that other manifests, including attestations won't be preserved.
 'os[/arch[/variant]]': Explicit platform (eg. linux/amd64)`)
 	flags.SetAnnotation("platform", "version", []string{"1.46"})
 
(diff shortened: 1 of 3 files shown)


cli/command/image/push.go
```diff
@@ -81,7 +88,7 @@ func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error
 
 		printNote(dockerCli, `Selecting a single platform will only push one matching image manifest from a multi-platform image index.
 This means that any other components attached to the multi-platform image index (like Buildkit attestations) won't be pushed.
-If you want to only push a single platform image while preserving the attestations, please use 'docker convert\n'
+If you want to push a whole multi-platform image, make sure all image content is present and remove the --platform flag.
```
krissetto, line 91: nit: i'd remove `whole` here.   

Also, what does `make sure all image content is present` mean in this scenario?  it's a bit confusing to me

vvoland: You can have a multi-platform image, but only have partial content available.
For example, `busybox` is a multi-platform image, but doing `docker pull busybox` will only fetch the content for your daemon's host platform, assuming that the host is `linux/amd64`:
```
$ docker pull busybox
$ docker pull --platform linux/arm64 busybox
```

Will pull the content for `linux/amd64` and `linux/arm64`, but the "full" `busybox` image index still references all the other supported platforms (`s390x`, `riscv64`, `arm/v5`, ...).

Retagging the busybox as-is and attempting to push it to a different repository in most cases will fail, because the index will reference unknown blobs (of other platforms).

cli/command/image/push.go
```diff
@@ -70,6 +72,11 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
 //
 //nolint:gocyclo
 func RunPush(ctx context.Context, dockerCli command.Cli, opts pushOptions) error {
+	// Don't default to DOCKER_DEFAULT_PLATFORM env variable on unsupported API versions.
+	if versions.LessThan(dockerCli.CurrentVersion(), "1.46") {
```
thaJeztah, line 76: The only thing I'm thinking here is if we should unconditionally set it to empty, or only if `!flags.Chanced("platform")`  (i.e., user didn't explicitly set a `--platform` flag), because we also have some code in the client to print an error when trying to use the option on older API versions; https://github.com/docker/cli/blob/9bb1a62735174e9220d84fecc056a0ef8a1fc26f/vendor/github.com/docker/docker/client/image_push.go#L41-L44

thaJeztah: Actually even wondering if this platform should default to `DOCKER_DEFAULT_PLATFORM` or if it just shouldn't pick up that value at all; I need to find back the reason that env-var was added, but it may have been for LCOW so that users could set a default platform for `docker run` (and thus `docker pull`). Maybe that doesn't make sense for `docker push`, because that should probably default to pushing what's there.

vvoland: > because we also have some code in the client to print an error when trying to use the option on older API versions;

Yep this is exactly why I didn't handle `Changed` here - passing `--platform` on older API will error out earlier anyway.

> Actually even wondering if this platform should default to `DOCKER_DEFAULT_PLATFORM` or if it just shouldn't pick up that value at all;

Yeah, considering that it's also a newly introduced flag, I would also be fine with just ignoring that env variable.

thaJeztah: > Yeah, considering that it's also a newly introduced flag, I would also be fine with just ignoring that env variable.

Yup, was thinking that as well; that `addPlatformFlag` utility is a bit awkward in either case, so just adding a flag "as usual" that doesn't have the magic `os.GetEnv()` default seems fine with me.

vvoland: Done

cli/command/image/push.go
dvdksn: bit of an aside, but this `NOTE` color really doesn't look good in my terminal 
image

dvdksn: I don't really understand what this note is trying to convey. Is it supposed to be a warning? When I read this it feels like I am doing something wrong. But we're not exactly saying that.
Can we clarify somehow?

krissetto: > You can have a multi-platform image, but only have partial content available.
> 
> For example, `busybox` is a multi-platform image, but doing `docker pull busybox` will only fetch the content for your daemon's host platform, assuming that the host is `linux/amd64`:
> 
> ```
> 
> $ docker pull busybox
> 
> $ docker pull --platform linux/arm64 busybox
> 
> ```
> 
> 
> 
> Will pull the content for `linux/amd64` and `linux/arm64`, but the "full" `busybox` image index still references all the other supported platforms (`s390x`, `riscv64`, `arm/v5`, ...).
> 
> 
> 
> Retagging the busybox as-is and attempting to push it to a different repository in most cases will fail, because the index will reference unknown blobs (of other platforms).

Thanks for that explanation @vvoland, makes sense. 
I'm not convinced the terminology we're using is the clearest for the users though (when i hear image content i do not think about architectural variants, but the actual contents of the image - files etc.), but i can't think of many suggestions right now. Where else in the docs do we speak of these things? Do we use "content" with the same meaning in those places?

dvdksn (commented):
LGTM :shipit:

cli/command/image/push.go
```diff
@@ -58,8 +58,13 @@ func NewPushCommand(dockerCli command.Cli) *cobra.Command {
...
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress verbose output")
 	command.AddTrustSigningFlags(flags, &opts.untrusted, dockerCli.ContentTrustEnabled())
-	flags.StringVar(&opts.platform, "platform", os.Getenv("DOCKER_DEFAULT_PLATFORM"),
+
+	// Don't default to DOCKER_DEFAULT_PLATFORM env variable, always default to
+	// pushing the image as-is. This also avoids forcing the platform selection
+	// on older APIs which don't support it.
+	flags.StringVar(&opts.platform, "platform", "",
 		`Push a platform-specific manifest as a single-platform image to the registry.
+Image index won't be pushed, meaning that other manifests, including attestations will be lost.
```
thaJeztah, line 67: The only part here that could still be ambiguous is "will be lost"; or at least; it will be still there, but won't be pushed.

I guess we can still tweak in follow-ups

Benehiko (approved):
LGTM

cli/command/image/push.go
vvoland: Changed `will be lost` to `won't be preserved`, WDYT?

krissetto (approved):

vvoland:
Thanks!

Issue #5241, opened 2024-07-08
`SIGTERM` sent to `docker run` command doesn't send a SIGTERM to the main process running inside the container anymore.


shambhushrestha-bpi:
### Description

In Docker version `26.1.4`, sending a SIGTERM to the `docker run` process would also send a SIGTERM to the main process inside the container. However, in version `27.0.3`, this behavior has changed. Now, the process exits with the message `context canceled` without the main process receiving a SIGTERM.

### Reproduce

1. Save the following as $HOME/test/handle_sigterm.sh and make it executable.
```
#!/bin/sh

# Function to handle SIGTERM
handle_sigterm() {
    echo "Received SIGTERM, exiting..."
    exit 0
}

trap 'handle_sigterm' TERM

while true; do
    echo "Waiting for sigterm"
    sleep 10
done
```

2. Run it as follows in a container
```
docker run -i \
    -v $HOME/test:$HOME/test:ro \
    alpine:latest \
    $HOME/test/handle_sigterm.sh
 ```

3. Now find the `pid` of this process as follows (in a separate terminal) and send a sigterm
```
ps -ef | grep 'handle_sigterm.sh' | grep 'docker run'
kill -15 <pid from above>
```

4. You should see the following output
```
Waiting for sigterm
context canceled
```

### Expected behavior

Up until version `26.x` , the SIGTERM would be received by the main process and it would output the following:

```
Waiting for sigterm
Received SIGTERM, exiting...
```

### docker version

```bash
Client: Docker Engine - Community
 Version:           27.0.3
 API version:       1.46
 Go version:        go1.21.11
 Git commit:        7d4bcd8
 Built:             Sat Jun 29 00:02:33 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          27.0.3
  API version:      1.46 (minimum version 1.24)
  Go version:       go1.21.11
  Git commit:       662f78c
  Built:            Sat Jun 29 00:02:33 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.18
  GitCommit:        ae71819c4f5e67bb4d5ae76a6b735f29cc25774e
 runc:
  Version:          1.7.18
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client: Docker Engine - Community
 Version:    27.0.3
 Context:    default
[...]

PR #5247, opened 2024-07-08
fix: container stream should not be terminated by ctx

Benehiko:
fixes: https://github.com/docker/cli/issues/5241

**- What I did**
Prevent the hijacked stream from exiting on main `ctx` cancellation. When the user sends a termination signal to the CLI while attached to a container, the signal should be forwarded to the container and handled by the process inside the container. The CLI should wait for the container to gracefully exit. 

If the user wishes to exit without waiting, sending more termination requests will still forcefully exit the CLI. 

**- How I did it**

**- How to verify it**

handle_sigterm.sh
```sh
#!/bin/sh

# Function to handle SIGTERM
handle_sigterm() {
    echo "Received SIGTERM, exiting..."
    exit 0
}

trap 'handle_sigterm' TERM

while true; do
    echo "Waiting for sigterm"
    sleep 10
done
```

```console
⋊> ~/G/cli on hotfix-sigterm-container ◦ build/docker run -i \                                                    
                                                      -v $HOME/Downloads/:/home:ro \
                                                      alpine:latest \
                                                      /home/handle_sigterm.sh
Waiting for sigterm
Waiting for sigterm
Received SIGTERM, exiting...
```

```console
⋊> ~ ps -ef | grep "handle_sigterm.sh" | grep "docker run"
benehiko  279458  277233  1 17:48 pts/5    00:00:00 build/docker run -i -v /home/benehiko/Downloads/:/home:ro alpine:latest /home/handle_sigterm.sh
⋊> ~ kill -15 279458 
```

```markdown changelog
sending a termination request to the CLI while attached to a container, will wait for the container to exit before closing the stream.

```

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -119,6 +119,8 @@ func runRun(ctx context.Context, dockerCli command.Cli, flags *pflag.FlagSet, ro
 
 //nolint:gocyclo
 func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOptions, copts *containerOptions, containerCfg *containerConfig) error {
+	ctx = context.WithoutCancel(ctx)
+
 	config := containerCfg.Config
 	stdout, stderr := dockerCli.Out(), dockerCli.Err()
 	apiClient := dockerCli.Client()
@@ -181,7 +183,7 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 		// ctx should not be cancellable here, as this would kill the stream to the container
 		// and we want to keep the stream open until the process in the container exits or until
 		// the user forcefully terminates the CLI.
-		closeFn, err := attachContainer(context.WithoutCancel(ctx), dockerCli, containerID, &errCh, config, container.AttachOptions{
+		closeFn, err := attachContainer(ctx, dockerCli, containerID, &errCh, config, container.AttachOptions{
 			Stream:     true,
 			Stdin:      config.AttachStdin,
 			Stdout:     config.AttachStdout,
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -178,7 +178,10 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 			detachKeys = runOpts.detachKeys
 		}
 
-		closeFn, err := attachContainer(ctx, dockerCli, containerID, &errCh, config, container.AttachOptions{
+		// ctx should not be cancellable here, as this would kill the stream to the container
+		// and we want to keep the stream open until the process in the container exits or until
+		// the user forcefully terminates the CLI.
+		closeFn, err := attachContainer(context.WithoutCancel(ctx), dockerCli, containerID, &errCh, config, container.AttachOptions{
 			Stream:     true,
 			Stdin:      config.AttachStdin,
 			Stdout:     config.AttachStdout,
--- a/e2e/container/run_test.go
+++ b/e2e/container/run_test.go
@@ -1,8 +1,10 @@
 package container
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
@@ -13,6 +15,7 @@ import (
 	is "gotest.tools/v3/assert/cmp"
 	"gotest.tools/v3/golden"
 	"gotest.tools/v3/icmd"
+	"gotest.tools/v3/poll"
 	"gotest.tools/v3/skip"
 )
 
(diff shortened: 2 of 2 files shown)


cli/command/container/run.go
```diff
@@ -178,7 +178,10 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 			detachKeys = runOpts.detachKeys
 		}
 
-		closeFn, err := attachContainer(ctx, dockerCli, containerID, &errCh, config, container.AttachOptions{
+		// ctx should not be cancellable here, as this would kill the stream to the container
+		// and we want to keep the stream open until the process in the container exits or until
+		// the user forcefully terminates the CLI.
+		closeFn, err := attachContainer(context.WithoutCancel(ctx), dockerCli, containerID, &errCh, config, container.AttachOptions{
```
vvoland, line 184: Looks like the only calls that are still given the cancellable context are:

- createContainer - could cancelling the context here cause an orphan container that's not removed? 🤔 
- ContainerStart - ditto, if cancellation happens before the engine gets the chance to actually attempt the "start" operation, we could block on `errCh` or `statusChan`
- MonitorTtySize - has a loop which doesn't end on context cancellation anyway


I think it would make sense to just strip the cancellation from the original passed context:
```go
func runContainer ...
    ctx := context.WithouCancel(ctx)
    ...
```

vvoland, line 184: Also, could we have an e2e test so that we don't regress on that in the future?

Benehiko: yeah I think we'll need to see how we should refactor this code in a follow-up. It's a bit of a mess and some of the logic is shared between `run`, `attach` and `start`.

Benehiko: After looking at the code again, I'm not sure if stripping the cancellation is the best idea - the code is already a bit of a mess and ignoring the context cancellation completely might not be the right move. 

I think let's try to get this in as is and figure out a way to do a proper refactor of the code.

krissetto: I kinda agree with not making too many changes here at once since the PR is for resolving one specific issue. From what I can see the potential issues mentioned, if present, already exist in the code, so I'd prefer if we tackle them separately once we can reproduce them (manually or with some crafted e2e test). 
WDTY?

Benehiko: Yeah I agree

Benehiko: true, but I think this fix is "good enough". I'm looking into making this function more "context friendly" in a follow-up. Let's not block the fix on a non-issue!

vvoland: Not blocking this one; but it's only a partial fix for the overall issue with this function. 

After wiring up OTEL, which also sneaked cancellable context here, we've already had multiple fixes in code that originates from this function. All of them were patched by using `context.WithoutCancel`.

Now, we do this for another place.
Instead of patching one issue, the correct fix is to restore the context behavior this function worked with for a long time. The intent behind passing context here was to wire up OTEL:
https://github.com/docker/cli/commit/5400a48aafdd853542fa83b70a11c7ae02f30509#diff-9f722157bd3206a95a2f882edb65882f91b1b07cd5903374d49dad1702eaa8c7R121-R143

This is the change that passed context to this function. At the time the context wasn't cancellable, it just had the OTEL values.
That changed later, when the global context passed via cobra command was made cancellable.

This code wasn't prepared at all to handle context cancellation, so this combination of changes resulting in passing a cancellable context is just a bug and is producing buggy behavior.

We should fix the function first, and then think about refactoring it.

Benehiko:
hmm... latest commit causing some strange test failures. Don't merge this yet since I've been able to get the same error locally.

```console
51.13 === Failed
51.13 === FAIL: cli/command/container TestRunCommandWithContentTrustErrors (unknown)
51.13 --- PASS: TestRunCommandWithContentTrustErrors/offline-notary-server (0.00s)
51.13 Error: write /dev/pts/0: file already closed
51.13 --- PASS: TestRunCommandWithContentTrustErrors/uninitialized-notary-server (0.00s)
51.13 panic: Fail in goroutine after TestRunAttachTermination has completed
51.13 
51.13 goroutine 71 [running]:
51.13 testing.(*common).Fail(0xc000780340)
51.13 	/usr/local/go/src/testing/testing.go:952 +0xd4
51.13 testing.(*common).FailNow(0xc000780340)
51.13 	/usr/local/go/src/testing/testing.go:975 +0x26
51.13 github.com/docker/cli/vendor/gotest.tools/v3/assert.ErrorIs({0x1030a50, 0xc000780340}, {0x1027d40?, 0xc0003e6090}, {0x1027e20?, 0x16ddf30}, {0x0, 0x0, 0x0})
51.13 	/go/src/github.com/docker/cli/vendor/gotest.tools/v3/assert/assert.go:311 +0x159
51.13 github.com/docker/cli/cli/command/container.TestRunAttachTermination.func6()
51.13 	/go/src/github.com/docker/cli/cli/command/container/run_test.go:85 +0x85
51.13 created by github.com/docker/cli/cli/command/container.TestRunAttachTermination in goroutine 209
51.13 	/go/src/github.com/docker/cli/cli/command/container/run_test.go:84 +0x486
```

Benehiko:
> hmm... latest commit causing some strange test failures. Don't merge this yet since I've been able to get the same error locally.
> 
> ```
> 51.13 === Failed
> 51.13 === FAIL: cli/command/container TestRunCommandWithContentTrustErrors (unknown)
> 51.13 --- PASS: TestRunCommandWithContentTrustErrors/offline-notary-server (0.00s)
> 51.13 Error: write /dev/pts/0: file already closed
> 51.13 --- PASS: TestRunCommandWithContentTrustErrors/uninitialized-notary-server (0.00s)
> 51.13 panic: Fail in goroutine after TestRunAttachTermination has completed
> 51.13 
> 51.13 goroutine 71 [running]:
> 51.13 testing.(*common).Fail(0xc000780340)
> 51.13 	/usr/local/go/src/testing/testing.go:952 +0xd4
> 51.13 testing.(*common).FailNow(0xc000780340)
> 51.13 	/usr/local/go/src/testing/testing.go:975 +0x26
> 51.13 github.com/docker/cli/vendor/gotest.tools/v3/assert.ErrorIs({0x1030a50, 0xc000780340}, {0x1027d40?, 0xc0003e6090}, {0x1027e20?, 0x16ddf30}, {0x0, 0x0, 0x0})
> 51.13 	/go/src/github.com/docker/cli/vendor/gotest.tools/v3/assert/assert.go:311 +0x159
> 51.13 github.com/docker/cli/cli/command/container.TestRunAttachTermination.func6()
> 51.13 	/go/src/github.com/docker/cli/cli/command/container/run_test.go:85 +0x85
> 51.13 created by github.com/docker/cli/cli/command/container.TestRunAttachTermination in goroutine 209
> 51.13 	/go/src/github.com/docker/cli/cli/command/container/run_test.go:84 +0x486
> ```

So I've been running tests locally quite a few times and could not reproduce this again... So maybe just a flake.

vvoland (approved):

PR #5258, opened 2024-07-18
login: slightly cleanup warning about unencrypted store

thaJeztah:
### login: don't print "unencrypted" warning when failing to save credentials

If we fail to save credentials, make sure that the error about saving
doesn't get lost in the warning about credentials being stored unencrypted.

Also discard errors about printing the warning, as those would be unlikely,
and if they would occur, probably would fail to be printed as well.

### login: slightly cleanup warning about unencrypted store

- Add an empty line before the warning to separate it from the command's output
- Use the `/go/` redirect URL that we have available.
- Put quotes around the filename used for storage.
- Use present tense for the message, as the message is printed while saving.
- User "credentials" instead of "password" for consistency with "credentials-store"

Before:

    docker login myregistry.example.com
    Username: thajeztah
    Password:
    WARNING! Your password will be stored unencrypted in /root/.docker/config.json.
    https://docs.docker.com/engine/reference/commandline/login/#credential-stores
    
    Login Succeeded

After:

    docker login myregistry.example.com
    Username: thajeztah
    Password:
    
    WARNING! Your credentials are stored unencrypted in '/root/.docker/config.json'.
    
    Login Succeeded

### DIFF
--- a/cli/command/registry/login.go
+++ b/cli/command/registry/login.go
@@ -18,6 +18,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// unencryptedWarning warns the user when using an insecure credential storage.
+// After a deprecation period, user will get prompted if stdin and stderr are a terminal.
+// Otherwise, we'll assume they want it (sadly), because people may have been scripting
+// insecure logins and we don't want to break them. Maybe they'll see the warning in their
+// logs and fix things.
 const unencryptedWarning = `WARNING! Your password will be stored unencrypted in %s.
 Configure a credential helper to remove this warning. See
 https://docs.docker.com/engine/reference/commandline/login/#credential-stores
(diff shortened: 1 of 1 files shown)


cli/command/registry/login.go
```diff
@@ -18,9 +18,15 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const unencryptedWarning = `WARNING! Your password will be stored unencrypted in %s.
+// unencryptedWarning warns the user when using an insecure credential storage.
+// After a deprecation period, user will get prompted if stdin and stderr are a terminal.
+// Otherwise, we'll assume they want it (sadly), because people may have been scripting
+// insecure logins and we don't want to break them. Maybe they'll see the warning in their
```
krissetto, line 24: 🤞

krissetto (approved):

PR #5262, opened 2024-07-19
gha: check-pr-branch: verify major version only

thaJeztah:
We'll be using release branches for minor version updates, so instead of (e.g.) a 27.0 branch, we'll be using 27.x and continue using the branch for minor version updates.

This patch changes the validation step to only compare against the major version.


thaJeztah:
moving to draft; looks like this latest variant doesn't work; https://github.com/moby/moby/pull/48191#issuecomment-2237827589

vvoland (approved):

PR #5282, opened 2024-07-23
correctly handle configuration file saving when it is a relative symlink

### LINKED ISSUE #5281: Can't handle config file properly when it's a relative symlink
### Description

When the config file is a relative symlink, docker cli can't handle it properly. Every config file writing operation, like `docker login` or `docker context use`, will take no effect on the real config file, but instead it creates a new file relative to the current working directory.

### Reproduce

In a fresh new docker environment:

```
$ mkdir ~/.docker
$ echo '{}' > ~/.docker/config-company.json
$ ln -s config-company.json ~/.docker/config.json  # this create a relative symlink
$ ls -l ~/.docker/config.json
lrwxrwxrwx 1 user group 22 Jul 23 21:37 /home/user/.docker/haha.json -> config-company.json
$ cat ~/.docker/config.json
{}
$ docker login -u <user>
Password:
Login Succeeded
$ cat ~/.docker/config.json
{}
# below is the file that docker cli actually created and wrote to, relative to the current working directory.
$ cat ./config-company.json
{
	"auths": {
			"auth": "xxxxxxxx"
		}
	}
}
```

### Expected behavior

docker cli should write to the absolute path of the symlink target.

### docker version

```bash
 Version:           27.0.3
 API version:       1.46
 Go version:        go1.21.11
 Git commit:        7d4bcd8
 Built:             Sat Jun 29 00:02:23 2024
 Context:           default

 Engine:
  Version:          27.0.3
  API version:      1.46 (minimum version 1.24)
  Go version:       go1.21.11
  Git commit:       662f78c
  Built:            Sat Jun 29 00:02:23 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.19
  GitCommit:        2bf793ef6dc9a18e00cb12efb64355c2c9d5eb41
 runc:
  Version:          1.7.19
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
 Version:    27.0.3
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.15.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
    Version:  v2.28.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 40
 Server Version: 27.0.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2bf793ef6dc9a18e00cb12efb64355c2c9d5eb41
 runc version: v1.1.13-0-g58aa920
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.8.0-38-generic
 Operating System: Ubuntu 24.04 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 31.26GiB
 Name: workstation
 ID: ed0b07d9-611f-4637-a1f8-1d9403688a32
 Debug Mode: false
...
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

### Additional Info

_No response_

willww64:
fixes: #5281 

**- What I did**
Correctly hand the cases with config file being a relative symlink.

**- How I did it**
Check whether the result path of `os.Readlink(cfgFile)` is a relative path, if so, get the absolute path of it.

**- How to verify it**
```bash
mkdir ~/.docker
echo '{}' > ~/.docker/config-company.json
ln -sf config-company.json ~/.docker/config.json   # create a relative symlink
docker login -u <user>
grep auths ~/.docker/config.json && echo OK || echo Failed
```

```markdown changelog
Fix an issue where the CLI would not correctly save the configuration file (`~/.docker/config.json`) if it was a relative symbolic link.
```

### DIFF
--- a/cli/config/configfile/file.go
+++ b/cli/config/configfile/file.go
@@ -169,10 +169,16 @@ func (configFile *ConfigFile) Save() (retErr error) {
 		return errors.Wrap(err, "error closing temp file")
 	}
 
-	// Handle situation where the configfile is a symlink
+	// Handle situation where the configfile is a symlink, and allow for dangling symlinks
 	cfgFile := configFile.Filename
-	if f, err := os.Readlink(cfgFile); err == nil {
+	if f, err := filepath.EvalSymlinks(cfgFile); err == nil {
 		cfgFile = f
+	} else if os.IsNotExist(err) {
+		// extract the path from the error if the configfile does not exist or is a dangling symlink
+		var pathError *os.PathError
+		if errors.As(err, &pathError) {
+			cfgFile = pathError.Path
+		}
 	}
 
 	// Try copying the current config file (if any) ownership and permissions
--- a/cli/config/configfile/file_test.go
+++ b/cli/config/configfile/file_test.go
@@ -538,6 +538,34 @@ func TestSaveWithSymlink(t *testing.T) {
 	assert.Check(t, is.Equal(string(cfg), "{\n	\"auths\": {}\n}"))
 }
 
+func TestSaveWithRelativeSymlink(t *testing.T) {
+	dir := fs.NewDir(t, t.Name(), fs.WithFile("real-config.json", `{}`))
+	defer dir.Remove()
+
+	symLink := dir.Join("config.json")
+	relativeRealFile := "real-config.json"
+	realFile := dir.Join(relativeRealFile)
+	err := os.Symlink(relativeRealFile, symLink)
+	assert.NilError(t, err)
+
+	configFile := New(symLink)
+
+	err = configFile.Save()
+	assert.NilError(t, err)
+
+	fi, err := os.Lstat(symLink)
+	assert.NilError(t, err)
+	assert.Assert(t, fi.Mode()&os.ModeSymlink != 0, "expected %s to be a symlink", symLink)
+
+	cfg, err := os.ReadFile(symLink)
+	assert.NilError(t, err)
+	assert.Check(t, is.Equal(string(cfg), "{\n	\"auths\": {}\n}"))
+
+	cfg, err = os.ReadFile(realFile)
+	assert.NilError(t, err)
+	assert.Check(t, is.Equal(string(cfg), "{\n	\"auths\": {}\n}"))
+}
+
 func TestPluginConfig(t *testing.T) {
 	configFile := New("test-plugin")
 	defer os.Remove("test-plugin")


cli/config/configfile/file.go
```diff
@@ -171,6 +171,10 @@ func (configFile *ConfigFile) Save() (retErr error) {
 	cfgFile := configFile.Filename
 	if f, err := os.Readlink(cfgFile); err == nil {
 		cfgFile = f
+		// The target of a symlink can be a relative path, ensure the final path is absolute
+		if !filepath.IsAbs(f) {
+			cfgFile = filepath.Join(dir, f)
+		}
 	}
```
vvoland, line 178: It's easier to just rely on the standard library to handle the symlinks:

```suggestion
	if f, err := filepath.EvalSymlinks(cfgFile); err == nil {
		cfgFile = f
	}
```

vvoland (changes requested):
Thanks! I have just one suggestion to make the code simpler.

cli/config/configfile/file.go
willww64: I'm not very familiar with Go's standard libraries, but after reading the documentation and testing it locally, I found that `filepath.EvalSymlink` is a perfect solution to the problem. Thanks for the suggestion! I've updated the code.

willww64:
The `TestLoadDanglingSymlink` test failed.

According to [that test's comment](https://github.com/docker/cli/blob/a69c0365b66b78dd61fd83c0a3ddefd0d8e7df15/cli/config/config_test.go#L55), Docker CLI currently doesn't treat a dangling symlink as an error. However, `filepath.EvalSymlinks` does treat a dangling symlink as an error. Consequently, when the symlink is dangling, the value of `cfgFile` isn't assigned to the "resolved" path of the symlink, causing the last assertion in `TestLoadDanglingSymlink` to fail.

After conducting local tests, I've compiled the following table:

|                  | `os.Readlink()` | `filepath.EvalSymlinks()` |
|------------------|-----------------|--------------------------|
| Real file        | ❌               | ✅                        |
| Symlink          | ✅               | ✅                        |
| Dangling Symlink | ✅               | ❌                        |

In summary:

- `os.Readlink` is to check whether a file is a link or not and simply returns the raw target of that link without trying to resolve the path
- `filepath.EvalSymlinks` checks if the file is valid (regardless of whether it's a symlink) and returns its resolved path.

It appears that `os.Readlink` is the only suitable solution in this case.

vvoland (commented):
Thanks, LGTM! (after squashing the commits)

cli/config/configfile/file.go
```diff
@@ -167,10 +167,13 @@ func (configFile *ConfigFile) Save() (retErr error) {
...
 
-	// Handle situation where the configfile is a symlink
+	// Handle situation where the configfile is a symlink, and allow for dangling symlinks
 	cfgFile := configFile.Filename
-	if f, err := os.Readlink(cfgFile); err == nil {
+	if f, err := filepath.EvalSymlinks(cfgFile); err == nil {
 		cfgFile = f
+	} else if os.IsNotExist(err) {
+		// extract the path from the error if the configfile does not exist or is a dangling symlink
+		cfgFile = err.(*os.PathError).Path
```
thaJeztah, line 176: Not sure if we can hit that situation, but this could panic, because `os.IsNotExist` also considers syscall errors; https://go.dev/play/p/5VjH5lK-YDB

```go
package main

import (
	"os"
	"syscall"
	"testing"
)

func TestCheckError(t *testing.T) {
	err := func() error {
		return syscall.ENOENT
	}()
	if os.IsNotExist(err) {
		t.Log(err)
		cfgFile := err.(*os.PathError).Path
		t.Log(cfgFile)
	}
}
```

thaJeztah: Changed this to use `errors.As`;

```go
// extract the path from the error if the configfile does not exist or is a dangling symlink
var pathError *os.PathError
if errors.As(err, &pathError) {
	cfgFile = pathError.Path
}
```

thaJeztah:
@vvoland I rebased and squashed this one; also adjusted the error-matching; https://github.com/docker/cli/pull/5282#discussion_r2102051077

vvoland (approved):

PR #5290, opened 2024-07-24
Fix flaky `TestCloseRunningCommand` test

laurazard:
**- What I did**

Fixed flaky `TestCloseRunningCommand` test.

**- How I did it**

Looks like this test was failing due to bad syntax on the `while` loop, which caused it to die after 1 second. If the test took a bit longer, the process would be dead before the following assertions run, causing the test to fail/be flaky.

**- How to verify it**

Run the test!

### DIFF
--- a/cli/connhelper/commandconn/commandconn_unix_test.go
+++ b/cli/connhelper/commandconn/commandconn_unix_test.go
@@ -48,7 +48,7 @@ func TestCloseRunningCommand(t *testing.T) {
 	defer close(done)
 
 	go func() {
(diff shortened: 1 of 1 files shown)


vvoland (approved):

thaJeztah:
Looks like there's more!

```bash
git grep 'while true; sleep 1;'
cli/connhelper/commandconn/commandconn_unix_test.go:    c, err := New(ctx, "sh", "-c", "while true; sleep 1; done")
cli/connhelper/commandconn/commandconn_unix_test.go:    c, err := New(ctx, "sh", "-c", "while true; sleep 1; done")
```

Issue #5294, opened 2024-07-24
terminating `docker attach` container prints `context canceled` error


thaJeztah:
### Description

I noticed this while updating the docs in https://github.com/docker/cli/pull/5289. Terminating an attach (`CTRL-c`, killing the container) prints a `context canceled` error.

Based on the example I was updating (assuming the example was correct!), older versions did not print these errors, and as they don't provide much value to the user, we should probably not print them 🤔 

```bash
docker run -d --name topdemo alpine /usr/bin/top -b

docker attach topdemo
Mem: 1397860K used, 6629032K free, 928K shrd, 168848K buff, 329628K cached
CPU:   0% usr   0% sys   0% nic  97% idle   1% io   0% irq   0% sirq
Load average: 0.25 0.16 0.08 1/342 6
  PID  PPID USER     STAT   VSZ %VSZ CPU %CPU COMMAND
    1     0 root     R     1712   0%   4   0% /usr/bin/top -b
^Ccontext canceled
```

I reproduced this with v27.0.3 and v27.1.1 (but haven't checked older versions)

thaJeztah:
cc @Benehiko @laurazard

laurazard:
Bisect says it was introduced in 3f0d90a2a939afa04a784b810dbdb035a0dff669 (cc @Benehiko)

laurazard:
I can take a quick look, and add a test.

Benehiko:
This should've been fixed in https://github.com/docker/cli/pull/5247

thaJeztah:
> This should've been fixed in #5247

Oh! Hm.. need to double check now if I checked with 27.0 or 27.1 (or if there's still an issue at hand)

laurazard:
I don't think it is,`attach` uses different code paths from `run`. I've got a commit with some tests and a fix on the way :)

Benehiko:
> I don't think it is,`attach` uses different code paths from `run`. I've got a commit with some tests and a fix on the way :)

oh you're right! I didn't look too closely at the issue my mind just went to the code that was problematic at the "attach" section in the run command :sweat_smile:

PR #5295, opened 2024-07-24
attach: don't return context cancelled error

laurazard:
closes https://github.com/docker/cli/issues/5294

**- What I did**

In https://github.com/docker/cli/pull/4993 we introduced a global signal handler and made sure all the contexts passed into command execution get (appropriately) cancelled when we get a `SIGINT`.

Due to how we use this context in `docker attach`, this caused us to start returning a context cancelation error when a user signals the running `docker attach`.

Since this is the intended behavior, we shouldn't return an error, so this commit adds checks to ignore this specific error in this case.

Also adds a regression test.

**- How to verify it**

Run the added test:
```bash
TESTDIRS="./e2e/container/..." TESTFLAGS="-test.run=TestAttachInterrupt" make -f docker.Makefile test-e2e-non-experimental
```

```markdown changelog


```

### DIFF
--- a/cli/command/container/attach.go
+++ b/cli/command/container/attach.go
@@ -145,7 +145,8 @@ func RunAttach(ctx context.Context, dockerCLI command.Cli, containerID string, o
 		detachKeys:   options.DetachKeys,
 	}
 
-	if err := streamer.stream(ctx); err != nil {
+	// if the context was canceled, this was likely intentional and we shouldn't return an error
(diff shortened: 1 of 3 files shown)


thaJeztah:
FWIW, I recall we also added a utility to errdefs, but not sure if we also must ignore the other context error for these https://github.com/docker/cli/blob/0cc1b8c0dff1a5cac3da4efc6d0d903b03689794/vendor/github.com/docker/docker/errdefs/is.go#L121

thaJeztah (approved):
LGTM, thanks!

PR #5297, opened 2024-07-25
attach: wait for exit code from `ContainerWait`

laurazard:
**- What I did**

Follow up to https://github.com/docker/cli/pull/5295

Such as with `docker run`, if a user CTRL-Cs while attached to a container, we should forward the signal and wait for the exit from `ContainerWait`, instead of just returning.

(see [`docker attach` docs](https://docs.docker.com/reference/cli/docker/container/attach/) for a clear explanation of the desired behavior)

**- How I did it**

Used an uncancellable context for the `ContainerWait` call.

**- How to verify it**

Run the test:
```
TESTDIRS="./e2e/container/..." TESTFLAGS="-test.run=TestAttachInterrupt" make -f docker.Makefile test-e2e-non-experimental
```

### DIFF
--- a/e2e/container/attach_test.go
+++ b/e2e/container/attach_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/docker/cli/e2e/internal/fixtures"
 	"gotest.tools/v3/assert"
 	"gotest.tools/v3/icmd"
+	"gotest.tools/v3/skip"
 )
 
 func TestAttachExitCode(t *testing.T) {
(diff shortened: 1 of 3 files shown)


laurazard:
Oop I forgot to commit something 😅

laurazard:
I've spent too long trying to figure out why I'd broken it on the `connhelper-ssh` tests until I remembered maybe it didn't work before 🙃  It does work, just not while running inside dind, but it already didn't before. I'll try to figure out how to make the tests pass on CI, but for now I'm skipping them.

laurazard:
@thaJeztah @vvoland PTAL

vvoland (approved):
LGTM; although I'd love to have some tests that would also verify if the returned exit code came from the right place (CLI vs container process), for example:


1. docker run --name asdf -d alpine sh -c 'trap "exit 33" SIGINT; sleep 5; exit 34'
2. docker attach asdf
3. send SIGINT to the cli process
4. check if the exit code is 33 and not 130 (SIGINT)

laurazard:
I'm writing some right now! (incoming in another PR 😅)

e2e/container/attach_test.go
```diff
@@ -32,7 +33,13 @@ func withStdinNewline(cmd *icmd.Cmd) {
...
 // Regression test for https://github.com/docker/cli/issues/5294
 func TestAttachInterrupt(t *testing.T) {
-	result := icmd.RunCommand("docker", "run", "-d", fixtures.AlpineImage, "sh", "-c", "sleep 5")
+	// this is a new test, it already did not work (inside dind) when over ssh
+	// todo(laurazard): make this test work w/ dind over ssh
+	skip.If(t, strings.Contains(os.Getenv("DOCKER_HOST"), "ssh://"))
+
+	// if
+	result := icmd.RunCommand("docker", "run", "-d", fixtures.AlpineImage,
+		"sh", "-c", "trap \"exit 33\" SIGINT; sleep 10; exit 34")
```
vvoland, line 42: The `sleep 10` won't be interrupted by the signal, so it will make the test wait for the whole 10s.

We can wait N times for a short intervals to make the signal being handled quicker:
```suggestion
		"sh", "-c", "trap \"exit 33\" SIGINT; for i in $(seq 100); do sleep 0.1; done; exit 34")
```

laurazard: You're right 🤦‍♀️ I initially had it with a `while true; do sleep 1; done; exit x` loop, and should've kept something like that.

PR #5302, opened 2024-07-26
[27.0 backport] attach: wait for exit code from `ContainerWait`

laurazard:
backport https://github.com/docker/cli/pull/5297


vvoland:
Failures seem related to: https://github.com/docker/cli/pull/5229#issuecomment-2220079501

It added the "exit status " message, but wasn't backported to the 27.0 branch

laurazard:
Ah, I guess it doesn't make since to backport #5229, so I'll adjust the expected output. WDYT @vvoland?

vvoland:
I'm still unsure if we should have that message at all 😅 I think we still haven't decided if this is the expected behavior or not.
@thaJeztah  WDYT?

thaJeztah:
Oh! Good one; yes, I was considering if we need an explicit option on `cli.StatusError` to mark it as "silent" or something; effectively `cli.StatusError` should (probably?) be used as a wrapper for existing errors to add "status code" to them. Or at least, considering something like

```go
err := foo()
if errdefs.IsNotFound(err) {
    // wrap the error with a atatus-code (and message?)
} 
```

 For this branch I would update the tests to not check for the error-message if that works while we work out some of that.

laurazard:
> For this branch I would update the tests to not check for the error-message if that works while we work out some of that.

Updated.

vvoland (approved):

PR #5303, opened 2024-07-26
tests/run: fix flaky `RunAttachTermination` test

laurazard:
**- What I did**

This test was just incorrect (and testing incorrect behavior): it was checking that `docker run` exited with a `context canceled` error after signalling the CLI/cancelling the command's context, but this was incorrect (and was fixed in
991b1303daec833d662ec54002d1b34c8805afab - which was when this test started failing).

However, since this test assertion was happening inside of a goroutine, it would sometimes pass if it didn't get to run before the test suite terminated. It was flaky because sometimes the [assertion inside the goroutine did get to execute](https://github.com/docker/cli/pull/5247#issuecomment-2222282787), but after the test finished execution, which is a big no-no.

As an aside, assertions inside goroutines are generally bad, and `govet` [even has a check](https://go.dev/doc/go1.16#vet-testing-T) for this (but it only catches `t.Fatal` and `t.FailNow` calls and not `assert.Xx`).

**- How I did it**

Fixed `RunAttachTermination` to test for the correct behavior, and generally cleaned the tests up a bit. Also added another test for the general/not SIGINT'ed case.

**- How to verify it**

```bash
go test -v -count=1 -run=TestRunAttach ./cli/command/container/...
```

### DIFF
--- a/cli/command/container/run_test.go
+++ b/cli/command/container/run_test.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"io"
 	"net"
-	"os/signal"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


laurazard:
@thaJeztah @vvoland can you TAL?

cli/command/container/run_test.go
```diff
@@ -38,15 +37,83 @@ func TestRunLabel(t *testing.T) {
...
+		waitFunc: func(_ string) (<-chan container.WaitResponse, <-chan error) {
+			responseChan := make(chan container.WaitResponse, 1)
+			errChan := make(chan error)
+
+			responseChan <- container.WaitResponse{
+				StatusCode: 33,
+			}
+			return responseChan, errChan
+		},
+		Version: "1.36",
```
thaJeztah, line 74: Curious; was API 1.36 picked for a specific reason? (curious if we could otherwise set it to the minimum version the daemon currently supports or otherwise adding a short comment to describe why this version was chosen)

cli/command/container/run_test.go
```diff
@@ -61,42 +128,60 @@ func TestRunAttachTermination(t *testing.T) {
...
+	// run command should attempt to attach to the container
 	select {
 	case <-time.After(5 * time.Second):
-		t.Fatal("containerAttachFunc was not called before the 5 second timeout")
+		t.Fatal("containerAttachFunc was not called before the timeout")
 	case <-attachCh:
 	}
 
-	assert.NilError(t, syscall.Kill(syscall.Getpid(), syscall.SIGTERM))
+	syscall.Kill(syscall.Getpid(), syscall.SIGINT)
```
vvoland, line 168: Ugh, I know it was there already, but I'm not thrilled about sending signals to the running process. 🙈 
This feels like it should be a e2e test and not an integration one, as it does have actual side effects to the whole process running tests (which may also affect other tests in case of failures/parallel tests running).

I'm _okay_ with this as an immediate fix, but we should really turn these into an e2e test.

thaJeztah, line 168: Curious why you removed error-handling here

thaJeztah (commented):
overall LGTM; left two comments / questions

cli/command/container/run_test.go
thaJeztah: Heh, was actually looking at that as well; it felt more like this could be / should be an e2e test 🤔

laurazard: I agree, but I think (for now) this makes sense – or rather, without testing this functionality, this test doesn't really make sense (and I did consider just deleting it/covering it with an e2e test – we already have a similar one in https://github.com/docker/cli/blob/826fc32e82e23bb5f80e85d8777427c5f0c24b4d/e2e/container/attach_test.go).

This isn't the overall CLI signal handling we're testing, but the specific run handling.

laurazard: No real reason, I can add it back in.

cli/command/container/run_test.go
laurazard: This was there prior, and I was too lazy to look into it. Probably not for a special reason, but I can look into it.

laurazard: https://github.com/docker/cli/blob/826fc32e82e23bb5f80e85d8777427c5f0c24b4d/cli/command/container/run_test.go#L34

cli/command/container/run_test.go
thaJeztah: Yes, I'm good with keeping it as-is for now; we could have a look at cleaning up some of this at some point

thaJeztah (approved):
LGTM

cli/command/container/run_test.go
laurazard: I'm on it, I just want to get this merged so that my follow ups aren't running into this every other time. This test has been failing a lot and I just want to fix it 😭

thaJeztah:
oh! LOL, I saw failures, and see you just pushed - probably need to remove the import as well

laurazard:
Turns out the version was set for a reason. I added a comment.

Issue #5305, opened 2024-07-29
`docker container logs --follow` will quit if the container is restarted


karolz-ms:
### Description

The command `docker container logs --follow` with quit if the container is restarted. Re-issuing the command will get the logs from before restart as well as after restart, but it should not be necessary.

### Reproduce

1. `docker run -d -p:6556:80 nginx`
2. (in a separate console) `docker container logs --follow <nginx container id>`
3. `docker container restart <nginx container id>`

### Expected behavior

The `container logs --follow` command should just continue streaming logs. The container is just restarted; it has the same container ID, there is no need to stop execution of the `logs` command.

### docker version

```bash
> docker version
Client:
 Version:           27.1.1
 API version:       1.46
 Go version:        go1.21.12
 Git commit:        6312585
 Built:             Tue Jul 23 19:57:57 2024
 OS/Arch:           windows/amd64
 Context:           desktop-linux

Server: Docker Desktop 4.33.0 (160616)
 Engine:
  Version:          27.1.1
  API version:      1.46 (minimum version 1.24)
  Go version:       go1.21.12
  Git commit:       cc13f95
  Built:            Tue Jul 23 19:57:19 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.19
  GitCommit:        2bf793ef6dc9a18e00cb12efb64355c2c9d5eb41
 runc:
  Version:          1.7.19
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
> docker info
Client:
 Version:    27.1.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.16.1-desktop.1
    Path:     C:\Program Files\Docker\cli-plugins\docker-buildx.exe
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.1-desktop.1
    Path:     C:\Program Files\Docker\cli-plugins\docker-compose.exe
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.34
    Path:     C:\Program Files\Docker\cli-plugins\docker-debug.exe
  desktop: Docker Desktop commands (Alpha) (Docker Inc.)
    Version:  v0.0.14
    Path:     C:\Program Files\Docker\cli-plugins\docker-desktop.exe
[...]

PR #5307, opened 2024-07-30
WIP: add configurable detach delay to logs

laurazard:
**- What I did**

Hacky PR for demonstration/discussion purposes only.

Addresses: https://github.com/docker/cli/issues/5305

**- How I did it**

**- How to verify it**

- `docker run -d --name bork alpine sh -c "while true; do echo hi; sleep 1; done"`
- `docker logs -f --delay 3 bork`
- (in another terminal) `docker restart bork`

```markdown changelog


```

### DIFF
--- a/cli/command/container/logs.go
+++ b/cli/command/container/logs.go
@@ -3,11 +3,14 @@ package container
 import (
 	"context"
 	"io"
+	"strconv"
+	"time"
 
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/completion"
 	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/events"
 	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/spf13/cobra"
 )
@@ -19,6 +22,8 @@ type logsOptions struct {
 	timestamps bool
 	details    bool
 	tail       string
+
+	detachDelay int
 
 	container string
 }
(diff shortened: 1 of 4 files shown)


cli/command/container/logs.go
```diff
@@ -49,6 +54,7 @@ func NewLogsCommand(dockerCli command.Cli) *cobra.Command {
 	flags.BoolVarP(&opts.timestamps, "timestamps", "t", false, "Show timestamps")
 	flags.BoolVar(&opts.details, "details", false, "Show extra details provided to logs")
 	flags.StringVarP(&opts.tail, "tail", "n", "all", "Number of lines to show from the end of the logs")
+	flags.IntVarP(&opts.detachDelay, "delay", "d", 0, "Number of seconds to wait for container restart before exiting")
```
vvoland, line 57: We could do what `tail` does and put this behind a separate `-F` option:

```
     -f      The -f option causes tail to not stop when end of file is reached, but rather to wait for additional data to be ap‐
             pended to the input.  The -f option is ignored if the standard input is a pipe, but not if it is a FIFO.

     -F      The -F option implies the -f option, but tail will also check to see if the file being followed has been renamed or
             rotated.  The file is closed and reopened when tail detects that the filename being read from has a new inode number.

             If the file being followed does not (yet) exist or if it is removed, tail will keep looking and will display the file
             from the beginning if and when it is created.

             The -F option is the same as the -f option if reading from standard input rather than a file.
```

So basically:
- `-f` would only work on existing containers and exit when the container ends
- `-F` would never exit, it would always just wait for the container, even if it doesn't exist at the time

laurazard: Ooh, I like that UX.

PR #5312, opened 2024-08-01
Replace deprecated usages of `assert.ErrorType`

laurazard:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/cli-plugins/manager/candidate_test.go
+++ b/cli-plugins/manager/candidate_test.go
@@ -1,14 +1,13 @@
 package manager
 
 import (
+	"errors"
 	"fmt"
-	"reflect"
 	"strings"
 	"testing"
 
 	"github.com/spf13/cobra"
 	"gotest.tools/v3/assert"
-	"gotest.tools/v3/assert/cmp"
 )
 
 type fakeCandidate struct {
@@ -80,7 +79,8 @@ func TestValidateCandidate(t *testing.T) {
 				assert.ErrorContains(t, err, tc.err)
 			case tc.invalid != "":
 				assert.NilError(t, err)
-				assert.Assert(t, cmp.ErrorType(p.Err, reflect.TypeOf(&pluginError{})))
+				var expectedError *pluginError
+				assert.Check(t, errors.As(p.Err, &expectedError))
 				assert.ErrorContains(t, p.Err, tc.invalid)
 			default:
 				assert.NilError(t, err)
--- a/cli/command/context/remove_test.go
+++ b/cli/command/context/remove_test.go
@@ -8,7 +8,6 @@ import (
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/docker/errdefs"
 	"gotest.tools/v3/assert"
-	is "gotest.tools/v3/assert/cmp"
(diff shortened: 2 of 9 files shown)


cli-plugins/manager/candidate_test.go
```diff
@@ -80,7 +79,8 @@ func TestValidateCandidate(t *testing.T) {
 				assert.ErrorContains(t, err, tc.err)
 			case tc.invalid != "":
 				assert.NilError(t, err)
-				assert.Assert(t, cmp.ErrorType(p.Err, reflect.TypeOf(&pluginError{})))
+				var expectedError *pluginError
+				assert.Check(t, errors.As(p.Err, &expectedError))
```
thaJeztah, line 83: Oh! I recall some fun cases where there's no _good_ replacement, or at least `errors.As` / `errors.Is` for sure has a bunch of pitfalls; see the discussion I had with the gotest.tools authors on https://github.com/gotestyourself/gotest.tools/issues/272

cli/command/context/remove_test.go
```diff
@@ -18,7 +17,7 @@ func TestRemove(t *testing.T) {
 	_, err := cli.ContextStore().GetMetadata("current")
 	assert.NilError(t, err)
 	_, err = cli.ContextStore().GetMetadata("other")
-	assert.Check(t, is.ErrorType(err, errdefs.IsNotFound))
+	assert.Check(t, errdefs.IsNotFound(err))
```
thaJeztah, line 20: To keep the current (more informative) output of `is.ErrorType`, we should probably use the 3rd argument of `assert.Check` to write a custom error message (`wanted %T, actual %T` or something along those lines), but it will definitely make the code a lot more lengthy/verbose 😞

laurazard: Ahh, yeah I just saw you mention the nicer output (I definitely wondered why we were using `Errortype(err, errdefs.IsXxx)`) in the thread with the gotest.tools folks.

cli-plugins/manager/candidate_test.go
thaJeztah: It's really horrible, and very easy to get wrong (in which case, it's even possible to get the unexpected result). I honestly really hate the `errors.As` / `errors.Is` handling for that reason.

cli/command/context/remove_test.go
thaJeztah: If we want to continue using the `is.ErrorType`, I think Daniel was open to un-deprecating it (IIUC, reason for deprecating was mostly cleaning up ("stdlib now has utilities for it, and the alternative can work, so let's deprecate"). We can open a PR to do so, and see if they're ok with it.

thaJeztah:
FWIW, I opened a PR to see if un-deprecating is acceptable;

- https://github.com/gotestyourself/gotest.tools/pull/286

thaJeztah:
New version of gotest.tools was released that un-deprecates this; I opened a PR to update to that version;

- https://github.com/docker/cli/pull/5808

laurazard:
Yayyy 🥳 Will close this.

PR #5314, opened 2024-08-02
fixes 4155: duplicate entries in docker node ps when multi nodes are passed

ehsan-salamati:
**- What I did**
Fixed duplicate entries in docker node ps when multi nodes are passed
**- How I did it**
When running the docker node ps command with multiple nodes in the for loop, the filter state retains node-specific data from previous iterations, causing task entries to be duplicated for the first specified node. This occurs because the filter object, being a dictionary type, was reused and modified across iterations, accumulating previous node filters.
**- How to verify it**
Added a unit test for this case.
**Before:**
image

**After :**
image

```markdown changelog


```


mwashburn2:
i would like this merged

PR #5320, opened 2024-08-06
disable pseudoterminal creation

### LINKED ISSUE #3019: Got an error with connecting daemon over ssh session
With setting ssh URL to the DOCKER_HOST environment var, the docker CLI does not work.

I'm using an EC2 instance (amazon linux 2) to run the docker daemon and set the ssh URL as the DOCKER_HOST. Then I ran the `docker containers` command on my local computer, but it had been failed with the error: `error during connect: Get http://docker/v1.40/containers/json: net/http: HTTP/1.x transport connection broken: malformed HTTP status code "pid"`

Next, I ran the command on the instance over the ssh connection; it seems to be succeeded.

How to fix it? or is this a bug?

```
$ docker -v
Docker version 19.03.13, build 4484c46d9d
```

I'm using docker for mac, Mac OS Mojave (10.14.6).

Best Regards.

trajano:
Fixes #3019

**- What I did**

**- How I did it**

**- How to verify it**

```markdown changelog


```

### DIFF
--- a/cli/connhelper/connhelper.go
+++ b/cli/connhelper/connhelper.go
@@ -52,6 +52,7 @@ func getConnectionHelper(daemonURL string, sshFlags []string) (*ConnectionHelper
 					args = append(args, "--host", "unix://"+sp.Path)
 				}
 				sshFlags = addSSHTimeout(sshFlags)
+				sshFlags = disablePseudoTerminalAllocation(sshFlags)
 				args = append(args, "system", "dial-stdio")
 				return commandconn.New(ctx, "ssh", append(sshFlags, sp.Args(args...)...)...)
 			},
@@ -78,4 +79,15 @@ func addSSHTimeout(sshFlags []string) []string {
 		sshFlags = append(sshFlags, "-o ConnectTimeout=30")
 	}
 	return sshFlags
+}
+
+// disablePseudoTerminalAllocation disables pseudo-terminal allocation to
+// prevent SSH from executing as a login shell
+func disablePseudoTerminalAllocation(sshFlags []string) []string {
+	for _, flag := range sshFlags {
+		if flag == "-T" {
+			return sshFlags
+		}
+	}
+	return append(sshFlags, "-T")
 }
--- a/cli/connhelper/connhelper_test.go
+++ b/cli/connhelper/connhelper_test.go
@@ -1,6 +1,7 @@
 package connhelper
 
 import (
+	"reflect"
 	"testing"
 
 	"gotest.tools/v3/assert"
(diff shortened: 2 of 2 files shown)


thaJeztah:
Thank you for contributing! It appears your commit message is missing a DCO sign-off,

```
```

commit.

    ```bash
    ```

2. Clone your fork locally

    ```bash
    ```

cli/connhelper/connhelper.go
```diff
@@ -79,3 +80,11 @@ func addSSHTimeout(sshFlags []string) []string {
 	}
 	return sshFlags
 }
+
+// Disable pseudo-terminal allocation.  This will prevent SSH from executing as a login shell
+func disablePseudoTerminalAllocation(sshFlags []string) []string {
+	if !strings.Contains(strings.Join(sshFlags, ""), "-T") {
```
tonistiigi, line 86: I don't get the `strings.Join()`. All args should be separate already in `[]string` iiuc. `slices.Contains()` should be enough.

cli/connhelper/connhelper.go
```diff
@@ -79,3 +80,11 @@ func addSSHTimeout(sshFlags []string) []string {
 	}
 	return sshFlags
 }
+
+// Disable pseudo-terminal allocation.  This will prevent SSH from executing as a login shell
```
tonistiigi, line 84: function comments should start with a function name in Go per convention

tonistiigi (approved):
Content SGTM, but please rebase/squash to clean up the merge commits in the PR.

edit: and DCO as well

trajano:
Ok I think I did the DCO now..  I added a test as well so I don't get 0% code cov

cli/connhelper/connhelper.go
```diff
@@ -52,6 +52,7 @@ func getConnectionHelper(daemonURL string, sshFlags []string) (*ConnectionHelper
 					args = append(args, "--host", "unix://"+sp.Path)
 				}
 				sshFlags = addSSHTimeout(sshFlags)
+				sshFlags = disablePseudoTerminalAllocation(sshFlags)
```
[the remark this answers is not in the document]
trajano: /shrug 
I don't think I'll do anything with this comment for now.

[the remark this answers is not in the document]
trajano: TBH I was surprised that you call `ssh` rather than having a specific SSH go library.

thaJeztah (approved):
changes LGTM


I'll do a quick rebase/squash

cli/connhelper/connhelper.go
```diff
@@ -79,3 +80,14 @@ func addSSHTimeout(sshFlags []string) []string {
...
+// disablePseudoTerminalAllocation disables pseudo-terminal allocation to
+// prevent SSH from executing as a login shell
+func disablePseudoTerminalAllocation(sshFlags []string) []string {
+	for _, flag := range sshFlags {
+		if flag == "-T" {
+			return sshFlags
+		}
+	}
+	return append(sshFlags, "-T")
+}
```
vvoland, line 93: Shouldn't this also check if `-t` has been specified explicitly and don't apply the `-T` in this case?

laurazard: Not sure about that – this PR disables pseudo-tty allocation for `ssh`. Using `-t` should continue working regardless, as that impacts allocating a pseudo-tty inside the container, but I don't think that somehow requires `ssh` to be a pseudo-tty as well.

laurazard: I guess this should be easy to check – I'll take a look.

laurazard: This works fine/as expected.

laurazard:
Should we backport this one to 27.x? @vvoland @thaJeztah

cli/connhelper/connhelper.go
thaJeztah: Ah, yes, we should probably skip if that's the case.

Probably not immediately critical for us because we don't set other options, but other users of the module technically could. And it doesn't seem like it would hurt to add the check.

thaJeztah:
I think back porting would be ok? I also had that minor follow up, which should be safe.

cli/connhelper/connhelper.go
laurazard: Ahhh 😅 for sure. I think that's okay, do we expect users to set flags in their context?

vvoland: Ah right, you can't set that via context 😅 It's only with this `GetConnectionHelperWithSSHOpts` that isn't used on our side.

thaJeztah: I don't think there's ways to customise these currently, unless there's some options inherited from the ssh-config. The ability to set options was added in https://github.com/docker/cli/pull/2541#discussion_r480098871 (some of the discussion / context related to it in that comment)

PR #5323, opened 2024-08-07
run: fix GetList return empty issue for throttledevice

jongwu:
Test "--device-read-bps" "--device-write-bps" will fail. The root cause is that GetList helper return empty as its local variable initialized to zero size.

**- What I did**
    This patch fix it by setting the related slice size to non-zero.

**- How to verify it**
`docker run --device-read-bps /dev/sda2:10M busybox sh -c 'dd if=/dev/zero of=/tmp.img bs=10M count=10 oflag=direct conv=fsync'`

### DIFF
--- a/opts/throttledevice.go
+++ b/opts/throttledevice.go
@@ -94,7 +94,7 @@ func (opt *ThrottledeviceOpt) String() string {
 
 // GetList returns a slice of pointers to ThrottleDevices.
 func (opt *ThrottledeviceOpt) GetList() []*blkiodev.ThrottleDevice {
-	out := make([]*blkiodev.ThrottleDevice, 0, len(opt.values))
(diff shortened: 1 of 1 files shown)


thaJeztah:
Doh! Nice catch. Looks like I introduced this in https://github.com/docker/cli/commit/d0dee3cebe71401a387aea70e9b9e67710089075 (https://github.com/docker/cli/commit/d0dee3cebe71401a387aea70e9b9e67710089075#diff-5aaad03b8a6ef733357accbdeb0346dcad3b83c0a460b25c77ad83b726244a75R96)

- https://github.com/docker/cli/pull/3770

Ironically, that was to enable more linters, and the linters didn't complain on this one! :see_no_evil:

thaJeztah (approved):
LGTM, thanks!

PR #5326, opened 2024-08-07
[27.x backport] tests/run: fix flaky `RunAttachTermination` test

vvoland:
- backport: https://github.com/docker/cli/pull/5303

**- What I did**

This test was just incorrect (and testing incorrect behavior): it was checking that `docker run` exited with a `context canceled` error after signalling the CLI/cancelling the command's context, but this was incorrect (and was fixed in
991b1303daec833d662ec54002d1b34c8805afab - which was when this test started failing).

However, since this test assertion was happening inside of a goroutine, it would sometimes pass if it didn't get to run before the test suite terminated. It was flaky because sometimes the [assertion inside the goroutine did get to execute](https://github.com/docker/cli/pull/5247#issuecomment-2222282787), but after the test finished execution, which is a big no-no.

As an aside, assertions inside goroutines are generally bad, and `govet` [even has a check](https://go.dev/doc/go1.16#vet-testing-T) for this (but it only catches `t.Fatal` and `t.FailNow` calls and not `assert.Xx`).

**- How I did it**

Fixed `RunAttachTermination` to test for the correct behavior, and generally cleaned the tests up a bit. Also added another test for the general/not SIGINT'ed case.

**- How to verify it**

```bash
go test -v -count=1 -run=TestRunAttach ./cli/command/container/...
```

### DIFF
--- a/cli/command/container/run_test.go
+++ b/cli/command/container/run_test.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"io"
 	"net"
-	"os/signal"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


austinvazquez (approved):

laurazard:
Ahh I forgot about backporting this, thanks.

laurazard (approved):

PR #5327, opened 2024-08-07
plugins: don't panic on Close if PluginServer nil

laurazard:
fixes https://github.com/docker/cli/issues/5317

**- What I did**

We had a report in fixes https://github.com/docker/cli/issues/5317 of a panic at https://github.com/docker/cli/blob/87c6624cb728e9a1c74e8141321ee70d8f67cb46/cli-plugins/socket/socket.go#L111

Looks like this will happen if the call to create a `PluginServer` fails and returns nil, since we're not checking for nilness before attempting to call `Close`:

https://github.com/docker/cli/blob/87c6624cb728e9a1c74e8141321ee70d8f67cb46/cmd/docker/docker.go#L247-L251

https://github.com/docker/cli/blob/87c6624cb728e9a1c74e8141321ee70d8f67cb46/cmd/docker/docker.go#L278

**- How I did it**

Added a nil check to `PluginServer.Close()`. We could also have addressed this at the caller sites, but this seems more robust for little cost.

**- How to verify it**

```markdown changelog
Fixed issue where main CLI process would sometimes panic while running a plugin.
```

### DIFF
--- a/cli-plugins/socket/socket.go
+++ b/cli-plugins/socket/socket.go
@@ -95,6 +95,9 @@ func (pl *PluginServer) Addr() net.Addr {
 //
 // The error value is that of the underlying [net.Listner.Close] call.
 func (pl *PluginServer) Close() error {
+	if pl == nil {
+		return nil
+	}
 	logrus.Trace("Closing plugin server")
 	// Close connections first to ensure the connections get io.EOF instead
 	// of a connection reset.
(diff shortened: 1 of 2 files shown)


laurazard:
Wondering if we should also backport this to 26.1.

thaJeztah (approved):
LGTM

thaJeztah:
> Wondering if we should also backport this to 26.1.

It's probably ok to cherry-pick; we may not be doing releases from that branch, but a cherry-pick doesn't do harm

PR #5331, opened 2024-08-08
c8d: Add `--platform` flag to history, save and load

vvoland:
- [x] needs: https://github.com/moby/moby/pull/48295
- [x] needs: https://github.com/docker/cli/pull/5428

Add `--platform` flag to history, save and load.

```markdown changelog
containerd image store: `docker load` and `docker save` now support a `--platform` flag allowing to choose a specific single-platform image to load/save if the image is multi-platform
containerd image store: `docker history` now supports a `--platform` flag allowing to choose a specific platform to show the history for.
```


cli/command/image/history.go
```diff
@@ -44,12 +47,28 @@ func NewHistoryCommand(dockerCli command.Cli) *cobra.Command {
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Only show image IDs")
 	flags.BoolVar(&opts.noTrunc, "no-trunc", false, "Don't truncate output")
 	flags.StringVar(&opts.format, "format", "", flagsHelper.FormatHelp)
+	flags.StringVar(&opts.platform, "platform", "",
+		`Specify a platform from a multi-platform image to show the history for.
+If the platform is not specified, the host platform is preferred if it's available, otherwise any available platform is used.
+
+Format: os[/arch[/variant]]
+Example: "linux/amd64"`)
```
dvdksn, line 55: How about including the full cmd?

```suggestion
Example: "docker image history --platform linux/amd64"`)
```

cli/command/image/load.go
```diff
@@ -39,6 +43,13 @@ func NewLoadCommand(dockerCli command.Cli) *cobra.Command {
 
 	flags.StringVarP(&opts.input, "input", "i", "", "Read from tar archive file, instead of STDIN")
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress the load output")
+	flags.StringVar(&opts.platform, "platform", "",
+		`Pick a single-platform to be loaded if the image is multi-platform.
+Full multi-platform image will be load if not specified.
+
+Format: os[/arch[/variant]]
+Example: "linux/amd64"`)
```
dvdksn, line 51:
```suggestion
		`Specify a platform from a multi-platform image to load.
If a platform is not specified, and the image is a multi-platform image, all platforms are loaded.

Format: os[/arch[/variant]]
Example: "docker image load --platform linux/amd64"`)
```

cli/command/image/save.go
```diff
@@ -37,6 +41,13 @@ func NewSaveCommand(dockerCli command.Cli) *cobra.Command {
 	flags := cmd.Flags()
 
 	flags.StringVarP(&opts.output, "output", "o", "", "Write to a file, instead of STDOUT")
+	flags.StringVar(&opts.platform, "platform", "",
+		`Pick a single-platform to be saved if the image is multi-platform.
+Full multi-platform image will be saved if not specified.
+
+Format: os[/arch[/variant]]
+Example: "linux/amd64"`)
```
dvdksn, line 49:
```suggestion
		`Specify a platform from a multi-platform image to save.
If a platform is not specified, and the image is a multi-platform image, all platform variants are saved.

Format: os[/arch[/variant]]
Example: "docker image save --platform linux/amd64"`)
```

cli/command/image/history.go
thaJeztah: I (somewhat) assume that this was done to allow defining a global `const` for these flags and for them to be shared (similar to what we did for `--format` 🤔), but I need to check if they differ in other ways

cli/command/image/history.go
```diff
@@ -45,12 +48,28 @@ func NewHistoryCommand(dockerCli command.Cli) *cobra.Command {
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Only show image IDs")
 	flags.BoolVar(&opts.noTrunc, "no-trunc", false, "Don't truncate output")
 	flags.StringVar(&opts.format, "format", "", flagsHelper.FormatHelp)
+	flags.StringVar(&opts.platform, "platform", "",
+		`Specify a platform from a multi-platform image to show the history for.
+If the platform is not specified, the host platform is preferred if it's available, otherwise any available platform is used.
```
thaJeztah, line 53: @dvdksn I applied your suggestions; I'm slightly wondering if we should / could condense the information a bit; for example;

`Specify a platform` -> `Platform from a multi-platform image ...`

(I don't think we generally add `Specify ....` as that's a bit implicit for options?)


The "format" could possibly be included in the first line, something like;

Platform from a multi-platform image to show the history for (format: "os[/arch[/variant]]")

By default, the daemon's native platform is preferred, otherwise any available platform is used.
```

(wondering if `the host platform` is ambiguous, and if we need to emphasise `daemon` platform)

Comment (krissetto) [reply cli/command/image/history.go:53]: i also think we should probably change "any available platform is used" to something more precise.. what does "any" mean in this context? the first available platform in alphabetical order?

Comment (dvdksn) [reply cli/command/image/history.go:53]: @thaJeztah your suggestion sgtm. Coming back to this I realize that multi-line flag descriptions are kinda exotic - we don't use them a lot. Maybe we should instead create sections in the respective `.md` files for the `platform` flag, like we usually do, to show examples.

Comment (laurazard) [reply cli/command/image/history.go:53]: IMO, if there is a specific behavior we've decide is part of the contract "i.e. it's always the first alphabetical one", then we should make that explicit. On the other hand, if we explicitly don't want to make any future guarantees about **which platform** is used, then **any** is fine. I honestly kind of would rather just return an error if there's no platform for the given image, but 🤷‍♀️ it's not a big deal either way, especially for `history`.

**Comment by laurazard** [comment]
I had a thought while looking at this again today – I think it would be really nice if we had a main .md file/doc with a general explanation about multi-platform images – i.e. "this is a new feature available on docker if the containerd store integration is enabled, it's kind of like the old images but now each image is actually an index and may contain multiple "images" for different platforms, etc.,". We could link there from a lot of other places, and we could also have a little table with the different commands that take a parameter to specify a platform if relevant, and I think it'd tie a lot of things together that we're kind of leaving implied/are spread out through a lot of different places/docs right now. Kind of like the [multi-platform doc](https://docs.docker.com/build/building/multi-platform/) for Build. cc @thaJeztah @dvdksn

**Comment by dvdksn** [comment]
@laurazard the multi-platform builds page has some of that. I agree it would be nice with a general "what's an image, actually" primer. We should probably consider OCI the typical format by now — buildx will default to them more often than not, and with containerd becoming the default storage, it feels like the transition is coming sooner rather than later. Single-platform manifests is a legacy format in my world.

**Comment by thaJeztah** [comment]
@dvdksn @laurazard I rewrote the PR to have a short description for the flags, and move the description to the docs.

There's still some issues/bugs to fix on the daemon side (as I found some issues while working on the docs; https://github.com/moby/moby/issues/48589)

**Comment by thaJeztah** [comment]
Oh! Should mention that I didn't squash / cleanup commits yet; will do so before merging if this looks good.

File: docs/reference/commandline/image_history.md [hunk 5]
```diff
@@ -76,3 +77,43 @@ $ docker history --format "{{.ID}}: {{.CreatedSince}}" busybox
 f6e427c148a7: 4 weeks ago
 <missing>: 4 weeks ago
 ```
+
+### <a name="platform"></a> Show history for a specific platform (--platform)
+
+The `--platform` option allows you to specify which platform-variant to show
+history for if multiple platforms are present. By default, `docker history`
+shows the history for the daemon's native platform or if not present, the
+first available platform.
```
krissetto, line 86: the first available platform found in alphabetical order?

docs/reference/commandline/image_history.md
```diff
@@ -76,3 +77,43 @@ $ docker history --format "{{.ID}}: {{.CreatedSince}}" busybox
...
+
+If the local image store has multiple platform-variants of an image, the
+`--platform` option selects which variant to show the history for. An error
+is produced if the given platform is not present in the local image cache.
+
+The platform option takes the `os[/arch[/variant]]` format, for example,
```
krissetto, line 92: nit
```suggestion
The platform option takes the `os[/arch[/variant]]` format, for example
```

docs/reference/commandline/image_history.md
```diff
@@ -76,3 +77,43 @@ $ docker history --format "{{.ID}}: {{.CreatedSince}}" busybox
...
+`linux/amd64` or `limux/arm64/v8`. Architecture and variant are optional,
+and default to the daemon's native architecture if omitted.
+
+The following examples pulls an `alpine` image for the daemon's native platform,
+and shows the history, then provides a custom platform, which fails until the
+given platform-variant is pulled;
```
krissetto, line 98: nit:
```suggestion
The following example: 
- pulls an `alpine` image for the daemon's native platform (because `--platform` is omitted);
- shows the history (also for the default platform, because `--platform` is omitted);
- tries to show the history for the image while providing a custom `--platform` value, which fails until the given platform-variant is pulled;
```

docs/reference/commandline/image_load.md
```diff
@@ -58,3 +59,32 @@ fedora              20                  58394af37342        7 weeks ago
...
+
+### <a name="platform"></a> Load a specific platform (--platform)
+
+The `--platform` option allows you to specify which platform-variant of the
+image to load. By default, `docker load` loads all platform variants that
+are present in the archive. Use the `--platform` option to specify which
+platform variant of the image to load. An error is produced if the given
+platform is not present in the archive.
+
+The platform option takes the `os[/arch[/variant]]` format, for example,
```
krissetto, line 72: nit:
```suggestion
The platform option takes the `os[/arch[/variant]]` format, for example
```

docs/reference/commandline/image_load.md
```diff
@@ -58,3 +59,32 @@ fedora              20                  58394af37342        7 weeks ago
...
+`linux/amd64` or `limux/arm64/v8`. Architecture and variant are optional,
+and default to the daemon's native architecture if omitted.
+
+The following example loads the `linux/amd64` variant of an `alpine` image
+from an archive that contains multiple platform variabts.
```
krissetto, line 77: typo nit:
```suggestion
from an archive that contains multiple platform variants.
```

docs/reference/commandline/image_save.md
```diff
@@ -59,3 +60,44 @@ You can even cherry-pick particular tags of an image repository.
...
+
+### <a name="platform"></a> Save a specific platform (--platform)
+
+The `--platform` option allows you to specify which platform-variant of the
+image to save. By default, `docker save` saves all platform variants that
+are present in the daemon's image store. Use the `--platform` option
+to specify which platform variant of the image to save. An error is produced
+if the given platform is not present in the local image store.
+
+The platform option takes the `os[/arch[/variant]]` format, for example,
```
krissetto, line 72: nit:
```suggestion
The platform option takes the `os[/arch[/variant]]` format, for example
```

docs/reference/commandline/image_save.md
```diff
@@ -59,3 +60,44 @@ You can even cherry-pick particular tags of an image repository.
...
+`linux/amd64` or `limux/arm64/v8`. Architecture and variant are optional,
+and default to the daemon's native architecture if omitted.
+
+The following examples pulls an `alpine` image for the daemon's native platform,
+then provides a custom platform, which fails until the given platform-variant
+is pulled;
```
krissetto, line 78: nit:
```suggestion
The following example: 
- pulls an `alpine` image for the daemon's native platform (because `--platform` is omitted);
- tries to save the image while providing a custom `--platform` value, which fails until the given platform-variant is pulled;
```

krissetto (approved):
overall LGTM, just left a few nits in the long-form docs

docs/reference/commandline/image_history.md
thaJeztah: Let me check with the grammar police (@dvdksn 🙏 ), but I think  it's recommended to always follow `for example` with a comma. Perhaps it should start with a new sentence here though (`format. For example, ...`)

krissetto: the word man! anyway these were just my 2 cents, so feel free to merge whenever 😄

krissetto: If the "For example" starts the sentence, I agree with having the comma after it. As is, when reading the sentence it feel strange to pause both before and after "for example", because it makes it sound like the example came before (causing some confusion, at least to myself 😄).

Definitely minutia though

docs/reference/commandline/image_history.md
krissetto: Sounds good to me if we want to define this a bit better somewhere else (or leave the road open to a better definition at a later point, without adding to user expectations)

docs/reference/commandline/image_history.md
thaJeztah: Oh! Forgot to reply on these; I've been going back and forth a bit if this approach would be making things better (much more in depth) or if it would be "too much".

Happy to hear from others though. Also happy to make changes along these lines in a follow up.

docs/reference/commandline/image_history.md
thaJeztah: Yeah; these are hard. I basically tried to avoid ending up with "3 word sentences" because that also felt a bit silly (which is why I used the comma before `example`). I updated to use a semicolon instead, but not sure if that's 100% grammatically correct either 🥹

cli/command/image/load.go
```diff
@@ -63,12 +69,21 @@ func runLoad(ctx context.Context, dockerCli command.Cli, opts loadOptions) error
...
-		loadOpts.Quiet = true
+		options.Quiet = true
 	}
 
-	response, err := dockerCli.Client().ImageLoad(ctx, input, loadOpts)
+	if opts.platform != "" {
+		p, err := platforms.Parse(opts.platform)
+		if err != nil {
+			_, _ = fmt.Fprintf(dockerCli.Err(), "Invalid platform %s", opts.platform)
+			return err
```
thaJeztah, line 81: I just realised that this isn't correct; it tries to both _print_ something as well as returning an error. The printed message is without a newline. But also, because the error is returned, both the printed message and the error get printed;


```bash
docker image save -o image-amd64.tar --platform in:valid/foo:bar  alpine:latest
Invalid platform in:valid/foo:bar"in:valid" is an invalid OS component of "in:valid/foo:bar": OSAndVersion specifier component must match "^([A-Za-z0-9_-]+)(?:\\(([A-Za-z0-9_.-]*)\\))?$": invalid argument
```

I wondered if this was perhaps done for cases where no output file is specified (so the tar output would be redirected), but the same happens in that case;


```bash
docker image save --platform in:valid/foo:bar  alpine:latest > file.tar
Invalid platform in:valid/foo:bar"in:valid" is an invalid OS component of "in:valid/foo:bar": OSAndVersion specifier component must match "^([A-Za-z0-9_-]+)(?:\\(([A-Za-z0-9_.-]*)\\))?$": invalid argument
```


So I think this should just wrap the error and return it.

thaJeztah: Updated these to just return the error

thaJeztah:
OK; this should probably be ready for review / merge now;

- Squashed / cleaned up commits
- Changed the `fmt.FPrint` to just return the error instead; https://github.com/docker/cli/pull/5331#discussion_r1792205348
- Added some basic unit tests

Benehiko (approved):

docs/reference/commandline/image_history.md
```diff
@@ -76,3 +77,43 @@ $ docker history --format "{{.ID}}: {{.CreatedSince}}" busybox
...
+history for if multiple platforms are present. By default, `docker history`
+shows the history for the daemon's native platform or if not present, the
+first available platform.
+
+If the local image store has multiple platform-variants of an image, the
+`--platform` option selects which variant to show the history for. An error
+is produced if the given platform is not present in the local image cache.
+
+The platform option takes the `os[/arch[/variant]]` format; for example,
+`linux/amd64` or `limux/arm64/v8`. Architecture and variant are optional,
```
laurazard, line 93:
```suggestion
`linux/amd64` or `linux/arm64/v8`. Architecture and variant are optional,
```

typo! :')

docs/reference/commandline/image_history.md
```diff
@@ -76,3 +77,43 @@ $ docker history --format "{{.ID}}: {{.CreatedSince}}" busybox
...
+and default to the daemon's native architecture if omitted.
```
laurazard, line 94: nit:
```suggestion
and default to the daemon's native platform if omitted.
```

Or, something like `and if omitted will fallback to the daemon's defaults.`

docs/reference/commandline/image_history.md
```diff
@@ -76,3 +77,43 @@ $ docker history --format "{{.ID}}: {{.CreatedSince}}" busybox
...
+
+The following examples pulls an `alpine` image for the daemon's native platform,
+and shows the history, then provides a custom platform, which fails until the
```
laurazard, line 97:
```suggestion
shows it's history, and then provides a custom platform, which fails until the
```

docs/reference/commandline/image_load.md
```diff
@@ -58,3 +59,32 @@ fedora              20                  58394af37342        7 weeks ago
...
+### <a name="platform"></a> Load a specific platform (--platform)
+
+The `--platform` option allows you to specify which platform-variant of the
+image to load. By default, `docker load` loads all platform variants that
+are present in the archive. Use the `--platform` option to specify which
+platform variant of the image to load. An error is produced if the given
+platform is not present in the archive.
+
+The platform option takes the `os[/arch[/variant]]` format; for example,
+`linux/amd64` or `limux/arm64/v8`. Architecture and variant are optional,
```
laurazard, line 73:
```suggestion
`linux/amd64` or `linux/arm64/v8`. Architecture and variant are optional,
```

typo 2: the typo strikes back :)

docs/reference/commandline/image_save.md
```diff
@@ -59,3 +60,44 @@ You can even cherry-pick particular tags of an image repository.
...
+### <a name="platform"></a> Save a specific platform (--platform)
+
+The `--platform` option allows you to specify which platform-variant of the
+image to save. By default, `docker save` saves all platform variants that
+are present in the daemon's image store. Use the `--platform` option
+to specify which platform variant of the image to save. An error is produced
+if the given platform is not present in the local image store.
+
+The platform option takes the `os[/arch[/variant]]` format; for example,
+`linux/amd64` or `limux/arm64/v8`. Architecture and variant are optional,
```
laurazard, line 73:
```suggestion
`linux/amd64` or `linux/arm64/v8`. Architecture and variant are optional,
```

typo 3: return of the typo

docs/reference/commandline/image_save.md
```diff
@@ -59,3 +60,44 @@ You can even cherry-pick particular tags of an image repository.
...
+and default to the daemon's native architecture if omitted.
+
+The following examples pulls an `alpine` image for the daemon's native platform,
+then provides a custom platform, which fails until the given platform-variant
```
laurazard, line 77:
```suggestion
then attempts to save it with a custom platform, which fails until the given platform-variant
```

laurazard (changes requested):

laurazard:
Oops, meant to comment/hit the other button. Just had found some typos/suggestions, otherwise it looks good.

docs/reference/commandline/image_history.md
thaJeztah: DOH!!!

"off-by-one" error 😂

thaJeztah:
Updated! I think I got them all, @laurazard

laurazard (approved):
LGTM

docs/reference/commandline/image_save.md
```diff
@@ -59,3 +60,44 @@ You can even cherry-pick particular tags of an image repository.
...
+latest: Pulling from library/alpine
+df110db6acd6: Download complete
+Digest: sha256:beefdbd8a1da6d2915566fde36db9db0b524eb737fc57cd1367effd16dc0d06d
+Status: Image is up to date for alpine:latest
+docker.io/library/alpine:latest
+
+$ docker image save --platform=linux/s390x -o image.tar alpine:latest
+ls -lh image.tar
+-rw-------  1 thajeztah  staff   3.9M Oct  7 11:06 image.tar
+```
```
dvdksn, line 103: I prefer to lead with the happy case and to separate the happy and sad cases, like you did with `load`.

````suggestion
The following example pulls the RISC-V variant of the `alpine:latest` image and saves it to a tar archive.

```console
$ docker pull --platform=linux/riscv64 alpine:latest
8c4a05189a5f: Download complete 
Digest: sha256:beefdbd8a1da6d2915566fde36db9db0b524eb737fc57cd1367effd16dc0d06d
Status: Downloaded newer image for alpine:latest
docker.io/library/alpine:latest

$ docker image save --platform=linux/riscv64 -o alpine_riscv.tar alpine:latest
```

The following example attempts to save a platform variant of `alpine:latest` that doesn't exist in the local image store, resulting in an error.

```console
$ docker image ls --filter reference=alpine --tree

alpine:latest               beefdbd8a1da       10.6MB         3.37MB       
├─ linux/riscv64            80cde017a105       10.6MB         3.37MB       
├─ linux/amd64              33735bd63cf8           0B             0B       
├─ linux/arm/v6             50f635c8b04d           0B             0B       
├─ linux/arm/v7             f2f82d424957           0B             0B       
├─ linux/arm64/v8           9cee2b382fe2           0B             0B       
├─ linux/386                b3e87f642f5c           0B             0B       
├─ linux/ppc64le            c7a6800e3dc5           0B             0B       
└─ linux/s390x              2b5b26e09ca2           0B             0B    

$ docker image save --platform=linux/amd64 -o alpine_amd64.tar alpine:latest
Error response from daemon: no suitable export target found for platform linux/s390x: no suitable image manifest found for platform {amd64 linux  [] }
```
````

File: docs/reference/commandline/image_save.md [hunk 20]
```diff
@@ -59,3 +60,44 @@ You can even cherry-pick particular tags of an image repository.
 ```console
 $ docker save -o ubuntu.tar ubuntu:lucid ubuntu:saucy
 ```
+
+### <a name="platform"></a> Save a specific platform (--platform)
+
+The `--platform` option allows you to specify which platform-variant of the
```
dvdksn, line 66:
```suggestion
The `--platform` option allows you to specify which platform variant of the
```

docs/reference/commandline/image_load.md
```diff
@@ -58,3 +59,32 @@ fedora              20                  58394af37342        7 weeks ago
 fedora              heisenbug           58394af37342        7 weeks ago         385.5 MB
 fedora              latest              58394af37342        7 weeks ago         385.5 MB
 ```
+
+
+### <a name="platform"></a> Load a specific platform (--platform)
+
+The `--platform` option allows you to specify which platform-variant of the
```
dvdksn, line 66:
```suggestion
The `--platform` option allows you to specify which platform variant of the
```

docs/reference/commandline/image_save.md
thaJeztah: Ah, yes, that works. I think I was originally working around some bugs we still had where things failed if the native platform wasn't present.

I tried to avoid `--tree` because the example _would_  work both with and without containerd image-store integration, but maybe we can add some wording to explain that output requires a multi-arch store, or perhaps skip it for now; WDYT?

dvdksn: If you don't want to show `--tree`, you could just stick with the `pull --platform=linux/riscv64` and save a different arch. But I think the tree output shows it nicely. Maybe a short note/tip would do...

```
> [!TIP]
> The `--tree` flag for `docker image ls` displays all locally available platform variants. (Requires the containerd image store.)
```

dvdksn: we should have a docs section for it... 🫥 😆

thaJeztah: OK; I updated and included the tree output (YOLO); `--tree` still works without containerd enabled, but just doesn't show the architectures underneath (perhaps we should make it "work" and just show a single architecture 🤔)

dvdksn: SGTM

dvdksn (approved):
LGTM 🚀

PR #5334, opened 2024-08-09
[27.x backport] docs: refresh image versions in examples

vvoland:
- backport: https://github.com/docker/cli/pull/5289

use current LTS versions of ubuntu where suitable, remove uses of ubuntu:23.10 (which reache EOL), and and update some other examples to use more current versions.


thaJeztah (approved):
LGTM

PR #5259, opened 2024-07-18
cli/config/credentials: move warning to fileStore

thaJeztah:
- follow-up to https://github.com/docker/cli/pull/5258
- relates to https://github.com/docker/cli/pull/5244#discussion_r1682633124

### cli/config/credentials: move warning to fileStore

The fileStore itself is aware that it's insecure, so we can make it
responsible for printing the warning. It's not "perfect", as we use
`os.Stderr` unconditionally (not `dockerCli.Err()`), but probably won't
make a difference in _most_ cases.

**- A picture of a cute animal (not mandatory but encouraged)**

PR #5344, opened 2024-08-12
auth: add support for oauth device-code login

laurazard:
**- What I did**

This commit adds support for the oauth [device-code](https://auth0.com/docs/get-started/authentication-and-authorization-flow/device-authorization-flow) login flow when authenticating against the official registry.

Also included an (imo overdue 😅) refactor of `cli/command/registry/login.go`.

**- How I did it**

This is achieved by adding `cli/internal/oauth`, which contains code to manage interacting with the Docker OAuth tenant (`login.docker.com`), including launching the device-code flow, refreshing access using the refresh-token, and logging out.

In order to preserve compatibility with other clients directly accessing credential helpers or `~/.docker/config.json`, the `OAuthManager` also generates a Hub PAT and saves it in the expected place.

This is the first part of a body of work to revamp CLI/engine authentication. More work will follow up, adding the ability to refresh tokens to credential helpers, at which point we plan to stop using the PAT for authentication and instead rely on the credential helpers to refresh the token on access before returning a token that can be used for authentication to the caller.

**- How to verify it**

Try to login, do things that require authentication, and logout!

```markdown changelog
Added support for device-code flow login when authenticating to the official registry.
```


cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	if err := m.api.RevokeToken(ctx, parts[0]); err != nil {
+		return fmt.Errorf("failed to revoke refresh token: %w", err)
+	}
+	return nil
+}
+
+const (
+	accessTokenKey  = registry.IndexServer + "access-token"
+	refreshTokenKey = registry.IndexServer + "refresh-token"
```
laurazard, line 172: using these for now so that Desktop can pick them up

laurazard:
New login flow:

https://github.com/user-attachments/assets/7757f96d-1a49-4d8c-99c6-169953bdc1f3

cli/internal/oauth/jwt.go
```diff
@@ -0,0 +1,99 @@
...
+	// Deprecated: Source will be removed in favor of DomainClaims.Source in the
+	// future.
+	Source *Source `json:"source,omitempty"`
+
+	// Deprecated: Username will be removed in favor of DomainClaims.Username in
+	// the future.
+	Username string `json:"username,omitempty"`
+
+	// Deprecated: UserID will be removed as soon as we can.
+	UserID string `json:"user_id,omitempty"`
```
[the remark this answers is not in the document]
laurazard: Good point! I don't think we're using most of those anyway.

cli/config/credentials/file_store.go
```diff
@@ -65,6 +65,8 @@ Configure a credential helper to remove this warning. See
 https://docs.docker.com/go/credential-store/
 `
 
+var alreadyPrinted bool
```
laurazard, line 68: I don't _think_ this needs to be a thread-safe type, but let me know if you think otherwise.

cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,194 @@
...
+	state, err := m.api.GetDeviceCode(ctx, m.audience)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get device code: %w", err)
+	}
+
+	if state.UserCode == "" {
+		return nil, errors.New("no user code returned")
+	}
+
+	_, _ = fmt.Fprintln(w, "\n\033[1mUSING WEB BASED LOGIN\033[0m")
```
vvoland, line 85: nit: Could use `aec` instead of hardcoding the ansi escape codes

cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,194 @@
...
+		tokenResChan <- tokenRes
+	}()
+
+	go func() {
+		reader := bufio.NewReader(os.Stdin)
+		reader.ReadString('\n')
+		_ = m.openBrowser(state.VerificationURI)
+	}()
+
+	_, _ = fmt.Fprint(w, "\nWaiting for authentication in the browser…\n")
```
vvoland, line 107: Is the single-character triple dot intentional?
Perhaps using a regular 3-character dot would be safer, we also seem to use it in other place:
https://github.com/docker/cli/blob/3b2e2e54ea28f227e61697dc734ee5541bd5af23/cli/command/registry/login.go#L121

```suggestion
	_, _ = fmt.Fprint(w, "\nWaiting for authentication in the browser...\n")
```

laurazard: Right, I wondered about that. I had that initially, and then changed it per @dvdksn's suggestion

laurazard: It is unicode iirc – I think we use some unicode characters in other places, but if we think it's safer to have a regular ellipsis I'm totally fine with that.

vvoland: I'm also fine with a single-character! Just wanted to make sure that it was intentional and not overzealous IDE 😅

cli/internal/oauth/api/api.go
```diff
@@ -0,0 +1,229 @@
...
+	return errors.New("unexpected response from tenant: " + resp.Status)
+}
+
+// WaitForDeviceToken polls the tenant to get access/refresh tokens for the user.
+// This should be called after GetDeviceCode, and will block until the user has
+// authenticated or we have reached the time limit for authenticating (based on
+// the response from GetDeviceCode).
+func (a API) WaitForDeviceToken(ctx context.Context, state State) (TokenResponse, error) {
+	ticker := time.NewTicker(state.IntervalDuration())
+	timeout := time.After(time.Duration(state.ExpiresIn) * time.Second)
```
vvoland, line 100: Since we already respect context cancellation there - how about just using a `context.WithTimeout`, possibly on the caller side?

cli/internal/oauth/api/api.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	if resp.StatusCode != http.StatusCreated {
+		return "", fmt.Errorf("unexpected response from Hub: %s", resp.Status)
+	}
+
+	var response patGenerateResponse
```
vvoland, line 216: This struct could be inlined, as it's not really used outside of this function
```suggestion
	var response struct {
	    Data struct {
		    Token string `json:"token"`
		}
	}
```

vvoland: Ah, it's actually used in tests!

cli/internal/oauth/manager/manager.go
```diff
@@ -0,0 +1,194 @@
...
+		if err != nil {
+			waitForTokenErrChan <- err
+			return
+		}
+		tokenResChan <- tokenRes
+	}()
+
+	go func() {
+		reader := bufio.NewReader(os.Stdin)
+		reader.ReadString('\n')
```
vvoland, line 103:
```suggestion
		_ = reader.ReadString('\n')
```

cli/config/credentials/file_store.go
thaJeztah: We could make it a property of the `fileStore` perhaps? Could even be a `sync.OnceFunc` that prints the warning, e.g. `c.printWarning()`

cli/internal/oauth/api/api.go
thaJeztah: I like that idea (but haven't dug in too deep on that matter)

cli/internal/oauth/api/api.go
```diff
@@ -0,0 +1,229 @@
...
+	timeout := time.After(time.Duration(state.ExpiresIn) * time.Second)
+
+	for {
+		select {
+		case <-ctx.Done():
+			ticker.Stop()
+			return TokenResponse{}, ctx.Err()
+		case <-ticker.C:
+			res, err := a.getDeviceToken(ctx, state)
+			if err != nil {
```
thaJeztah, line 109: I'm always bad with tickers; does if need to be stopped in this branch somewhere?

cli/internal/oauth/manager/manager.go
thaJeztah: I _think_ we're already using it in some other places, so (probably?) fine

cli/command/registry/logout.go
```diff
@@ -53,6 +54,13 @@ func runLogout(_ context.Context, dockerCli command.Cli, serverAddress string) e
 		regsToLogout = append(regsToLogout, hostnameAddress, "http://"+hostnameAddress, "https://"+hostnameAddress)
 	}
 
+	if isDefaultRegistry {
```
thaJeztah, line 57: Don't think this is new, but we're now using it twice; wondering if we're missing a new check for `isDefaultRegistry` after the above branch does a `credentials.ConvertToHostname(serverAddress)`


(unrelated: I think the intent of `ConvertToHostname` is effectively "normalise storage key" - we need to look at some of that at some point - naming is confusing, and not always properly describes intent)

cli/command/registry_test.go
```diff
@@ -61,7 +61,7 @@ func TestGetDefaultAuthConfig(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		serverAddress := tc.inputServerAddress
-		authconfig, err := command.GetDefaultAuthConfig(cfg, tc.checkCredStore, serverAddress, serverAddress == "https://index.docker.io/v1/")
+		authconfig, err := command.GetDefaultAuthConfig(cfg, tc.checkCredStore, serverAddress)
```
thaJeztah, line 64: It's a bit unfortunate that some of these are exported; we may need to do a proper assessment to learn how widely these may be used (and thus; how breaking changing the signature is).

thaJeztah (commented):
Mostly glancing over (again); perhaps we should walk through the latest iteration in a call if that works for you.

cli/internal/oauth/api/api.go
laurazard: Hmm, I generally prefer this oauth-tenant timer logic to be in one place rather than spread out. We take the `State` here and set up timers based on that. This will always be the same (regardless of caller) since this is coming from the tenant, so I don't know why we'd do that.

Also, I think it's a different case to have "context was cancelled" here rather than "it took longer than the tenant said it could", so I'd rather not abstract that away from this bit of code.

vvoland, line 100: We have the `IntervalDuration` helper for the `Interval`, but the `ExpiresIn` conversion is handled inline - could we also have a similar helper here? 😅

vvoland: Fine by me

cli/command/registry_test.go
laurazard: I mean, we can revert it. It's an entirely unnecessary/clunky parameter – the information is redundant and passing flags like this is always a bit of a smell

cli/config/credentials/file_store.go
laurazard: Hmm, I worry that there's a few places initializing the store through the code, so making it a property of the store wouldn't necessarily make the warning "once per CLI invocation" but rather "once per store". I'd rather this be a package var instead (but can be a `sync.OnceFunc`). WDYT?

cli/internal/oauth/manager/manager.go
laurazard: Oh! You left this in the previous one and I forgot 😅 thanks for the reminder

cli/internal/oauth/api/api.go
laurazard: We do, right? In the previous/next branch.

laurazard: https://github.com/laurazard/cli/blob/5dc31384671757ee8a18e85b327a9bd058fbcbc2/cli/internal/oauth/api/api.go#L105
and
https://github.com/laurazard/cli/blob/5dc31384671757ee8a18e85b327a9bd058fbcbc2/cli/internal/oauth/api/api.go#L123

cli/internal/oauth/api/api.go
laurazard: Agreed! (and added)

cli/config/credentials/file_store.go
laurazard: also @vvoland WDYT?

vvoland: Yeah I think it's fine to do that in the package as the fileStore shouldn't really be printing out messages 🙈 But I think that we can't really move it somewhere so the caller can do it?

laurazard: We kind of had that before 😅 To be fair, it doesn't really fit well anywhere https://github.com/docker/cli/pull/5259

laurazard:
With the last commit https://github.com/laurazard/cli/commit/cd0023d541f4af44edc7ee2306809030e830ab03:

Screenshot 2024-08-13 at 15 11 16

Screenshot 2024-08-13 at 15 11 58

vvoland (approved):
LGTM

cli/command/registry/logout.go
thaJeztah: > On another look, I guess a user could try to logout by doing docker logout index.docker.io and we should make that work.

Yes,  that's roughly the scenario I had in mind. Technically it could go even further (log out for `[http(s)://]registry-1.docker.io` etc), but it looks like the current code only assumes "no registry" passed to mean docker hub. For the old code that made no real difference, as in that case it would at most do too many logouts (`hostnameAddress, "http://"+hostnameAddress, "https://"+hostnameAddress`), but for the new flow, it may be relevant that we would miss the `manager.NewManager(store).Logout()` call.

cli/internal/oauth/api/api.go
thaJeztah: I think I must've been looking at the returns in this branch, and wondering if a `defer time.Stop()` would be the way to go

cli/command/registry_test.go
thaJeztah: Yes, I had in mind that if this is to be included in a minor release, we need to be careful not to introduce a breaking change. Agreed that most of this is iffy / redundant, just trying to reduce risk

That said, the signature of `ConfigureAuth` was also changed, which may be more difficult to keep the same as before (unless we do a new function for that I guess)

thaJeztah (commented):
overall LGTM

- Can you double-check the comment about the  `logout <registry name that is actually default registry>` case
- The "saving insecure" "once" message is _probably_ fine w.r.t. concurrency, but not 100% sure if (e.g.) Piñata or other codebases may be executing this code; (maybe not too problematic?)
- w.r.t. the signature changes; yeah, on the fence; we could try to be on the safe side (keep the signature for now, but start making a pass at a more rigorous cleanup)
- ☝️ as part of cleanup, we really need to centralise some of the parsing/normalizing bits it's so much spread in way too many places 😞

cli/command/registry_test.go
laurazard: Reverted the signature changes here (and in `ConfigureAuth`).

cli/internal/oauth/api/api.go
laurazard: Ah, that sounds good. Changed ✅

cli/config/credentials/file_store.go
laurazard: Ended up changing this to an `atomic.Bool` as per https://github.com/docker/cli/pull/5344#pullrequestreview-2238001263

> but not 100% sure if (e.g.) Piñata or other codebases may be executing this code; (maybe not too problematic?)

Can't hurt to be too safe, this has bit us in the ass before.

cli/command/registry/logout.go
laurazard: So I started looking into this, and the current situation is – the web-based flow only triggers when the user runs either:
- `docker login`, or
- `docker login https://index.docker.io/v1/`
the following do not work:
- `docker login index.docker.io`
- `docker login index.docker.io/v1/
- `docker login https://index.docker.io`

For logout, the following trigger an oauth logout:
- `docker logout`, or
- `docker logout https://index.docker.io/v1/`

This seems consistent to me, but WDYT @thaJeztah?

laurazard:
JFYI I have a small improvement I want to do if possible so please ping me before merging.

laurazard:
Done! I wanted to prevent these errors:
```console
/usr/bin/xdg-open: 882: seamonkey: not found
/usr/bin/xdg-open: 882: mozilla: not found
/usr/bin/xdg-open: 882: epiphany: not found
/usr/bin/xdg-open: 882: konqueror: not found
/usr/bin/xdg-open: 882: chromium: not found
/usr/bin/xdg-open: 882: chromium-browser: not found
/usr/bin/xdg-open: 882: google-chrome: not found
/usr/bin/xdg-open: 882: www-browser: not found
/usr/bin/xdg-open: 882: links2: not found
/usr/bin/xdg-open: 882: elinks: not found
/usr/bin/xdg-open: 882: links: not found
/usr/bin/xdg-open: 882: lynx: not found
/usr/bin/xdg-open: 882: w3m: not found
xdg-open: no method available for opening ...
```
from showing up if there are no such binaries (I run into this in my headless pet dev box). We already print the URL for the user, this just clutters up the output.

Fun fact, the Github CLI runs into this as welll:

Screenshot 2024-08-14 at 19 55 44

laurazard:
@thaJeztah, want to take a final look?

thaJeztah (approved):
LGTM 🫶 


(sorry, saw you updated, but was AFK when I did)

PR #5357, opened 2024-08-19
list/tree: Hide Used column if no truthy value

vvoland:
- stacked on: https://github.com/docker/cli/pull/5356

Don't show the `Used` column in the image tree view when no image is actually used and the column would be empty.

### DIFF
--- a/cli/command/image/tree.go
+++ b/cli/command/image/tree.go
@@ -3,6 +3,7 @@ package image
 import (
 	"context"
 	"fmt"
+	"slices"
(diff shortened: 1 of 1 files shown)


krissetto:
I'm not totally sure about this from a UX perspective.. wouldn't it risk being a bit confusing for the users if sometimes they see the column and sometimes they don't?

laurazard:
I'm fine with it for now since this is behind an experimental flag, but we should review these questions before making it stable.

PR #581, opened 2017-10-02
Improve presentation of published port ranges

thaJeztah:
fixes https://github.com/docker/cli/issues/178

Port mappings in `docker service ls` are quite verbose, and occupy a lot of
space when ranges of ports are published.

This patch improves the output by reconstructing ranges of ports.

Given the following service;

    $ docker service create \
      -p 60-61:60-61 \
      -p 62:61 \
      -p 80:80 \
      -p 81:80 \
      -p 90-95:90-95 \
      -p 90-92:90-92/udp \
      -p 93-96:93-96/udp \
      --name foo \
      nginx:alpine

Before this patch is applied:

    $ docker service ls
    ID                  NAME                MODE                REPLICAS            IMAGE               PORTS
    u1kwguv841qg        foo                 replicated          1/1                 nginx:alpine        *:60->60/tcp,*:61->61/tcp,*:62->61/tcp,*:80->80/tcp,*:81->80/tcp,*:90->90/tcp,*:91->91/tcp,*:92->92/tcp,*:93->93/tcp,*:94->94/tcp,*:95->95/tcp,*:90->90/udp,*:91->91/udp,*:92->92/udp,*:93->93/udp,*:94->94/udp,*:95->95/udp,*:96->96/udp

After this patch is applied:

    $ docker service ls
    ID                  NAME                MODE                REPLICAS            IMAGE               PORTS
    u1kwguv841qg        foo                 replicated          1/1                 nginx:alpine        *:60-62->60-61/tcp,*:80-81->80/tcp,*:90-95->90-95/tcp,*:90-96->90-96/udp

Additional enhancements can still be made, and marked as TODO in this change;

- combine non-consecutive ports mapped to a single port (`80->80`, `81->80`,
  `84->80`, `86->80`, `87->80`); to be printed as `*:80-81,84,86-87->80`.
- combine `tcp` and `udp` mappings if their port-mapping is the same;
  print `*:80-81->80-81/tcp+udp` instead of `*:80-81->80-81/tcp, *:80-81->80-81/udp`

**- Description for the changelog**

```markdown
* Improve presentation of published ports by combining port-ranges [docker/cli#581](https://github.com/docker/cli/pull/581)
```

PR #5363, opened 2024-08-20
cli/formatter: bracket IPv6 addrs prepended to ports

akerouanton:
**- What I did**

On `docker ps`, port bindings with an IPv6 HostIP should have their addresses put into brackets when joining them to their ports.

RFC 3986 (Section 3.2.2) stipulates that IPv6 addresses should be enclosed within square brackets. This RFC is only about URIs. However, doing so here helps user identifier what's part of the IP address and what's the port. It also makes it easier to copy/paste that '[addr]:port' into other software (including browsers).

**- How to verify it**

CI, or run the following commands with a recent Engine:

```
$ docker run --rm -d --name c0 -p 8000:80/tcp 5201/udp alpine top
$ docker ps
```

```markdown changelog
Improve presentation of IPv6-mapped ports, using bracket notation.
```

### DIFF
--- a/cli/command/formatter/container.go
+++ b/cli/command/formatter/container.go
@@ -5,6 +5,7 @@ package formatter
 
 import (
 	"fmt"
+	"net"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


thaJeztah (approved):
LGTM, thanks!

thaJeztah:
Reminds me that I still want to port over the changes I added for swarm services to present port-ranges;

- https://github.com/docker/cli/pull/581

PR #20, opened 2017-05-02
Add more lint checks

dnephin:
These are split into separate commits so that you can see which changes were made for which linter. I can squash it before we merge if you want.

Adds 5 new linters and fixes some errors they uncovered.

`gocycle` has to ignore a few functions, and currently has an extremely high limit. I'd like to eventually get that down to much closer to 10, and work on fixing some of the whitelisted offenders.

PR #5365, opened 2024-08-21
[27.x backport] cli/formatter: bracket IPv6 addrs prepended to ports

vvoland:
- backport: https://github.com/docker/cli/pull/5363

**- What I did**

On `docker ps`, port bindings with an IPv6 HostIP should have their addresses put into brackets when joining them to their ports.

RFC 3986 (Section 3.2.2) stipulates that IPv6 addresses should be enclosed within square brackets. This RFC is only about URIs. However, doing so here helps user identifier what's part of the IP address and what's the port. It also makes it easier to copy/paste that '[addr]:port' into other software (including browsers).

**- How to verify it**

CI, or run the following commands with a recent Engine:

```
$ docker run --rm -d --name c0 -p 8000:80/tcp 5201/udp alpine top
$ docker ps
```

```markdown changelog
- IPv6 addresses shown by `docker ps` in port bindings are now bracketed
```

### DIFF
--- a/cli/command/formatter/container_test.go
+++ b/cli/command/formatter/container_test.go
@@ -474,6 +474,16 @@ func TestDisplayablePorts(t *testing.T) {
 		{
 			[]types.Port{
 				{
+					IP:          "::",
+					PrivatePort: 9988,
+					Type:        "tcp",
+				},
+			},
+			"[::]:0->9988/tcp",
+		},
+		{
+			[]types.Port{
+				{
 					PrivatePort: 9988,
 					PublicPort:  8899,
 					Type:        "tcp",
--- a/cli/command/formatter/container.go
+++ b/cli/command/formatter/container.go
@@ -5,6 +5,7 @@ package formatter
 
 import (
 	"fmt"
+	"net"
 	"sort"
 	"strconv"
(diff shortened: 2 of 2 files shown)


cli/command/formatter/container_test.go
```diff
@@ -471,6 +471,16 @@ func TestDisplayablePorts(t *testing.T) {
...
 		},
+		{
+			[]types.Port{
+				{
+					IP:          "::",
+					PrivatePort: 9988,
+					Type:        "tcp",
+				},
+			},
+			expected: "[::]:0->9988/tcp",
```
akerouanton, line 482: The CI fail with this error:

> #20 59.13 === Errors
> #20 59.13 cli/command/formatter/container_test.go:482:12: mixture of field:value and value elements in struct literal

```suggestion
			"[::]:0->9988/tcp",
```

akerouanton (changes requested):

cli/command/formatter/container_test.go
vvoland: Oh right, I removed the `ports:` but didn't notice the `expected` 🙈

Issue #5375, opened 2024-08-28
Suspending `docker login` causes an error


dvdksn:
### Description

With the new device code login, if I suspend the process with `^Z`, and later return to it with `fg`, while it's waiting for authentication, then the CLI gives me an error: `failed waiting for authentication: You are polling faster than the specified interval of 5 seconds.`

### Reproduce

```
[david:~/src/d/docs] rn-27.2.0+ 1 ± docker login

USING WEB BASED LOGIN
To sign in with credentials on the command line, use 'docker login -u <username>'

Your one-time device confirmation code is: CTPC-WSTX
Press ENTER to open your browser or submit your device code here: https://login.docker.com/activate

Waiting for authentication in the browser…
^Z
[1]+  Stopped                 docker login
1z [david:~/src/d/docs] rn-27.2.0+ 49s 146 ± # do stuff
1z [david:~/src/d/docs] rn-27.2.0+ ± fg
docker login
failed waiting for authentication: You are polling faster than the specified interval of 5 seconds.
```

### docker version

```bash
Client:
 Version:    27.2.0-rc.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  96eb69a
    Path:     /Users/david/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.29.2-desktop.2
    Path:     /Users/david/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.34
    Path:     /Users/david/.docker/cli-plugins/docker-debug
  desktop: Docker Desktop commands (Alpha) (Docker Inc.)
    Version:  v0.0.15
    Path:     /Users/david/.docker/cli-plugins/docker-desktop
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
    Path:     /Users/david/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.25
    Path:     /Users/david/.docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.5
    Path:     /Users/david/.docker/cli-plugins/docker-feedback
  harmonia: Extend your localhost with cloud workspaces (Docker Inc.)
    Version:  0.13.45
[...]

PR #5376, opened 2024-08-28
oauth/api: drain timer channel on each iteration

laurazard:
Fixes https://github.com/docker/cli/issues/5375

**- What I did**

Previously, if while polling for oauth device-code login results a user suspended the process (such as with CTRL-Z) and then restored it with `fg`, an error might occur in the form of:

```
failed waiting for authentication: You are polling faster than the specified interval of 5 seconds.
```

This is due to our use of a `time.Ticker` here - if no receiver drains the ticker channel (and timers/tickers use a buffered channel behind the scenes), more than one tick will pile up, causing the program to "tick" twice, in fast succession, after it is resumed.

**- How I did it**

The new implementation replaces the `time.Ticker` with a `time.Timer` (`time.Ticker` is just a nice wrapper) and introduces a helper function `resetTimer` to ensure that before every `select`, the timer is stopped and it's channel is drained.

**- How to verify it**

Reproduce as described in https://github.com/docker/cli/issues/5375

```markdown changelog
Fix issue that will sometimes cause the browser-login flow to fail if the CLI process is suspended and then resumed while waiting for the user to authenticate.
```

### DIFF
--- a/cli/internal/oauth/api/api.go
+++ b/cli/internal/oauth/api/api.go
@@ -96,18 +96,32 @@ func tryDecodeOAuthError(resp *http.Response) error {
 // authenticated or we have reached the time limit for authenticating (based on
 // the response from GetDeviceCode).
 func (a API) WaitForDeviceToken(ctx context.Context, state State) (TokenResponse, error) {
-	ticker := time.NewTicker(state.IntervalDuration())
+	// Ticker for polling tenant for login – based on the interval
+	// specified by the tenant response.
+	ticker := time.NewTimer(state.IntervalDuration())
 	defer ticker.Stop()
-	timeout := time.After(state.ExpiryDuration())
+	// The tenant tells us for as long as we can poll it for credentials
+	// while the user logs in through their browser. Timeout if we don't get
+	// credentials within this period.
+	timeout := time.NewTimer(state.ExpiryDuration())
+	defer timeout.Stop()
 
 	for {
+		resetTimer(ticker, state.IntervalDuration())
 		select {
 		case <-ctx.Done():
+			// user canceled login
 			return TokenResponse{}, ctx.Err()
 		case <-ticker.C:
+			// tick, check for user login
 			res, err := a.getDeviceToken(ctx, state)
 			if err != nil {
-				return res, err
+				if errors.Is(err, context.Canceled) {
+					// if the caller canceled the context, continue
+					// and let the select hit the ctx.Done() branch
+					continue
+				}
+				return TokenResponse{}, err
 			}
 
 			if res.Error != nil {
@@ -119,14 +133,33 @@ func (a API) WaitForDeviceToken(ctx context.Context, state State) (TokenResponse
 			}
 
 			return res, nil
-		case <-timeout:
+		case <-timeout.C:
+			// login timed out
 			return TokenResponse{}, ErrTimeout
 		}
 	}
 }
 
+// resetTimer is a helper function thatstops, drains and resets the timer.
+// This is necessary in go versions <1.23, since the timer isn't stopped +
+// the timer's channel isn't drained on timer.Reset.
+// See: https://go-review.googlesource.com/c/go/+/568341
+// FIXME: remove/simplify this after we update to go1.23
+func resetTimer(t *time.Timer, d time.Duration) {
+	if !t.Stop() {
+		select {
+		case <-t.C:
+		default:
+		}
+	}
+	t.Reset(d)
+}
+
 // getToken calls the token endpoint of Auth0 and returns the response.
 func (a API) getDeviceToken(ctx context.Context, state State) (TokenResponse, error) {
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
 	data := url.Values{
 		"client_id":   {a.ClientID},
 		"grant_type":  {"urn:ietf:params:oauth:grant-type:device_code"},
--- a/cli/internal/oauth/api/api_test.go
+++ b/cli/internal/oauth/api/api_test.go
@@ -198,7 +198,7 @@ func TestWaitForDeviceToken(t *testing.T) {
 		state := State{
 			DeviceCode: "aDeviceCode",
 			UserCode:   "aUserCode",
-			Interval:   1,
+			Interval:   5,
 			ExpiresIn:  1,
 		}
 


cli/internal/oauth/api/api.go
```diff
@@ -119,12 +121,26 @@ func (a API) WaitForDeviceToken(ctx context.Context, state State) (TokenResponse
...
+		case <-timeout.C:
 			return TokenResponse{}, ErrTimeout
 		}
 	}
 }
 
+// resetTimer is a helper function thatstops, drains and resets the timer.
+// This is necessary in go versions <1.23, since the timer isn't stopped +
+// the timer's channel isn't drained on timer.Reset.
+// See: https://go-review.googlesource.com/c/go/+/568341
```
dvdksn, line 133: Should there be a `TODO` here to redo this when Go is upgraded?

dvdksn (approved):
works in my local test 🎉

cli/internal/oauth/api/api.go
laurazard: Oh, sure! The current code will be fine, but we'll be able to simplify this.

laurazard: Added

vvoland (approved):

cli/internal/oauth/api/api.go
```diff
@@ -96,16 +96,33 @@ func tryDecodeOAuthError(resp *http.Response) error {
...
+	// we reach the timeout specified by the tenant.
+	// Does not get canceled if the user cancels execution.
+	timeoutCtx, cancel := context.WithTimeout(
+		context.Background(), state.ExpiryDuration())
+	defer cancel()
+
+	// Context for calling getDeviceToken – gets cancelled whether
+	// the caller cancels login or we hit the timeout.
+	getDeviceTokenCtx, getDeviceTokenCancel := context.WithTimeout(
+		ctx, state.ExpiryDuration())
```
thaJeztah, line 114: Wondering why we want to use separate contexts for these 🤔 


i.e.; wouldn't it be OK to cancel whatever we're doing if the parent context is canceled or has a timeout and times out? Basically, a cancelled context would mean; "stop what you're doing; we're no longer interested", and not sure if it (really?) matters whether that was because the parent context had a timeout set (and timed out), or the parent context was cancelled (the error message probably would indicate which of those was the case)?


If we _do_ want a separate `timeoutCtx`, should that one be using `context.WithoutCancel(ctx)` (i.e. still be based on the parent context, but removing the timeout)?

vvoland: We don't need to use the caller's ctx as a `timeoutCtx` parent as we don't really pass it anywhere. And the `getDeviceTokenCtx` also uses the same expiration.

If we really want to avoid context here, we can just simplify it to:

```go
func (a API) WaitForDeviceToken(ctx context.Context, state State) (TokenResponse, error) {
    expired := time.After(state.ExpiryDuration())
```

and then..
```go
	select {
		case <-ctx.Done():
			// user canceled login
			return TokenResponse{}, ctx.Err()
        case <-expired:
            return TokenResponse{}, ErrTimeout
```

laurazard: By selecting on multiple things that timeout at the same time, we're introducing a race condition and when the timeout is hit, `select` will randomly select a branch and we won't know if it happened because the caller cancelled the context or it timed out.

That's why I initially had it more like @vvoland's suggestion.

laurazard: However, @thaJeztah wanted put a timer on the context passed to `a.getDeviceToken`, which is why I introduced `getDeviceTokenCtx`.

laurazard:
@thaJeztah @vvoland PTAL

cli/internal/oauth/api/api.go
vvoland: In the previous version we didn't `select` on `getDeviceTokenCtx` though, so IIUC we could make it not racy:

When both `timeoutCtx` and `getDeviceTokenCtx` would be `Done()` due to a deadline exceeded, it would either:

1. Immediately go to `case <-timeoutCtx.Done():` branch in case the timeout happened before `getDeviceToken` was called
2. `res, err := a.getDeviceToken(getDeviceTokenCtx, state)` would error out - in this case we could check if `timeoutCtx` is also `Done()` and either explicitly return `ErrTimeout` or just continue to the next loop iteration so it goes back into the 1.

thaJeztah: Yeah, my thinking was that `a.getDeviceToken` could hang, and in that case the login could hang forever?

```patch
git diff
diff --git a/cli/internal/oauth/api/api.go b/cli/internal/oauth/api/api.go
index 9ea48fecf..ff46126be 100644
--- a/cli/internal/oauth/api/api.go
+++ b/cli/internal/oauth/api/api.go
@@ -103,7 +103,7 @@ func (a API) WaitForDeviceToken(ctx context.Context, state State) (TokenResponse
        // The tenant tells us for as long as we can poll it for credentials
        // while the user logs in in the browser. Timeout if we don't get
        // credentials within this period.
-       timeout := time.NewTimer(state.ExpiryDuration())
+       timeout := time.NewTimer(10 * time.Second)
        defer timeout.Stop()

        for {
@@ -163,6 +163,7 @@ func (a API) getDeviceToken(ctx context.Context, state State) (TokenResponse, er
                "device_code": {state.DeviceCode},
        }
        oauthTokenURL := a.TenantURL + "/oauth/token"
+       time.Sleep(30 * time.Second)

        resp, err := postForm(ctx, oauthTokenURL, strings.NewReader(data.Encode()))
        if err != nil {
```



```bash
time ./build/docker login
...
Waiting for authentication in the browser…
failed waiting for authentication: timed out waiting for device token

________________________________________________________
Executed in   35.55 secs      fish           external
   usr time   26.23 millis    0.22 millis   26.02 millis
   sys time   52.20 millis    1.45 millis   50.74 millis
```

That said, `state.ExpiryDuration()` looks to be really long already (15 minutes) - slightly curious if that's really how long we want to wait for the user to log in :thinking_face:;

```
IntervalDuration: 5s
ExpiryDuration: 15m0s
```

But regardless, we could consider setting a timeout on the context so that the request would never take longer than the timeout;

```patch
git diff
diff --git a/cli/internal/oauth/api/api.go b/cli/internal/oauth/api/api.go
index 9ea48fecf..d8e2ea8bc 100644
--- a/cli/internal/oauth/api/api.go
+++ b/cli/internal/oauth/api/api.go
@@ -106,6 +106,9 @@ func (a API) WaitForDeviceToken(ctx context.Context, state State) (TokenResponse
        timeout := time.NewTimer(state.ExpiryDuration())
        defer timeout.Stop()

+       ctx, cancel := context.WithTimeout(ctx, state.ExpiryDuration())
+       defer cancel()
+
        for {
                resetTimer(ticker, state.IntervalDuration())
                select {
```

Not sure if we want the error message to be as-is (TBD?);

```
Waiting for authentication in the browser…
failed waiting for authentication: failed to get tokens: Post "https://login.docker.com/oauth/token": context deadline exceeded
```

cli/internal/oauth/api/api.go
```diff
@@ -96,17 +96,31 @@ func tryDecodeOAuthError(resp *http.Response) error {
...
 // the response from GetDeviceCode).
 func (a API) WaitForDeviceToken(ctx context.Context, state State) (TokenResponse, error) {
-	ticker := time.NewTicker(state.IntervalDuration())
+	// Ticker for polling tenant for login – based on the interval
+	// specified by the tenant response.
+	ticker := time.NewTimer(state.IntervalDuration())
 	defer ticker.Stop()
-	timeout := time.After(state.ExpiryDuration())
+	// The tenant tells us for as long as we can poll it for credentials
+	// while the user logs in in the browser. Timeout if we don't get
```
thaJeztah, line 104: Heh; linter doesn't like the comment;

```
68.69 cli/internal/oauth/api/api.go:104:2: Duplicate words (in) found (dupword)
68.69 	// while the user logs in in the browser. Timeout if we don't get
68.69 	^
```

Perhaps "logs in through the browser" or "logs in using the browser" ?

cli/internal/oauth/api/api.go
laurazard: > slightly curious if that's really how long we want to wait for the user to log in :thinking_face:

Why not? The user can always cancel it. I don't think we should be trying to be smarter than the user here.

laurazard: > But regardless, we could consider setting a timeout on the context so that the request would never take longer than the timeout;

That's what I had previously 😅

laurazard: The current approach seems fairly simple, and is my preferred approach. There's no reason to replace the current `timeout` timer with a context afaict, other than to just avoid timers – but I think in this case replacing it with a context makes this less understandeable, not more.

We can also, additionally, set a reasonable deadline (e.g. 1 minute) inside `API.getDeviceToken`, to prevent it being stuck forever. Does that sound okay for all?

vvoland: Sounds good to me!

laurazard: Done ✅ @thaJeztah this LGTY?

cli/internal/oauth/api/api.go
```diff
@@ -96,17 +96,31 @@ func tryDecodeOAuthError(resp *http.Response) error {
...
+		resetTimer(ticker, state.IntervalDuration())
 		select {
 		case <-ctx.Done():
+			// user canceled login
 			return TokenResponse{}, ctx.Err()
 		case <-ticker.C:
+			// tick, check for user login
 			res, err := a.getDeviceToken(ctx, state)
 			if err != nil {
+				if errors.Is(err, context.Canceled) {
```
thaJeztah, line 119: Do we need to handle `context.DeadlineExceeded` here?  We could even use the errdefs helper, but not sure if we want to import it here; https://github.com/docker/cli/blob/092b5f0fcb3e567e616b97ae167c05c42f3b3a1b/vendor/github.com/docker/docker/errdefs/is.go#L120-L123

cli/internal/oauth/api/api.go
```diff
@@ -96,17 +96,31 @@ func tryDecodeOAuthError(resp *http.Response) error {
...
+					// if the caller canceled the context, continue
+					// and let the select hit the ctx.Done() branch
+					continue
+				}
 				return res, err
```
thaJeztah, line 124: Nit (probably doesn't make much difference in practice); perhaps be explicit here as well that we don't return a result.

```suggestion
				return TokenResponse{}, err
```

thaJeztah (commented):
Thanks for updating! Left two comments 🙈

cli/internal/oauth/api/api.go
laurazard: Hmm, the purpose of that:
```go
// if the caller canceled the context, continue
// and let the select hit the ctx.Done() branch
```
is just to make sure we hit the `user canceled login` error if the user cancels execution while `API.getDeviceToken` is going rather than the `context canceled error` from there, so I don't think we need it. This is also all inside `/internal`, so we don't expect others to import this, so I think it's fine. If the user does get a timeout while trying to connect to the tenant, I think we should error out and let them know.

cli/internal/oauth/api/api.go
laurazard: Sure!

cli/internal/oauth/api/api.go
thaJeztah: Ah, right; yeah, was curious if we wanted to retry if the connection timed out. I guess that would be a bit of an exceptional case, so returning early in that case makes sense as well.

Thanks!

thaJeztah (approved):
LGTM

Issue #5377, opened 2024-08-28
Ignore binary files from .gitattributes. 


mbhadale:
### Description

git clone docker-cli shows a list of modified files on the master branch on Linux machine. It seems line encodings are being modified during checkout. Please ignore the binary files and fix the encoding rule for .env file in .gitattributes files.

-bash-4.2$ git branch 
* master

-bash-4.2$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   cli/command/container/testdata/utf8.env
	modified:   cmd/docker/winresources/docker.ico
	modified:   docs/extend/images/authz_additional_info.png
	modified:   docs/extend/images/authz_allow.png
	modified:   docs/extend/images/authz_chunked.png
	modified:   docs/extend/images/authz_connection_hijack.png
	modified:   docs/extend/images/authz_deny.png

-bash-4.2$ git remote show origin
* remote origin
  Fetch URL: https://github.com/docker/cli.git
  Push  URL: https://github.com/docker/cli.git
  HEAD branch: master
  Remote branches:
    18.06                                               tracked
    18.09                                               tracked
    19.03                                               tracked
    20.10                                               tracked
    23.0                                                tracked
    24.0                                                tracked
    25.0                                                tracked
    26.0                                                tracked
    26.1                                                tracked
    27.x                                                tracked
    dependabot/github_actions/actions/upload-artifact-4 tracked
    master                                              tracked
  Local branch configured for 'git pull':
    master merges with remote master
  Local ref configured for 'git push':
    master pushes to master (up to 

### Reproduce

git clone https://github.com/docker/cli.git on a Linux machine. 
git status

This issue is  observed as intermittent but the frequency of failure is high.

### Expected behavior

No modified files on master branch after git clone.

### docker version

```bash
-bash-4.2$ docker version
Client:
 Version:         1.13.1
 API version:     1.26
 Package version: 
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

### docker info

```bash
-bash-4.2$ docker info
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

thaJeztah:
Possibly related to this PR (maybe it's too eager, and also affecting pngs?);

- https://github.com/docker/cli/pull/5216

[...]

PR #5381, opened 2024-08-29
fix: gitattributes enforcing line endings

Benehiko:
fixes https://github.com/docker/cli/issues/5377

**- What I did**
Switch to an opt-in approach to forcing the `lf` line ending to certain files and directories.

**- How I did it**
Change the first line `* text=auto eol=lf` to not enforce line endings and instead only force the `scripts/` directory to use `lf` line endings. Also add future support for `.bat` files in-case we add them and also enforce `.sh` files with `lf` line endings.

**- How to verify it**
Checkout the branch after setting `git config core.autocrlf true` on a windows machine. On linux we can also enforce the line endings with `git config core.eol crlf` and use that as a test. There shouldn't be any files marked by git as changed.

```markdown changelog


```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,14 @@
-* text=auto eol=lf
+* text=auto
+
 Dockerfile* linguist-language=Dockerfile
 vendor.mod linguist-language=Go-Module
 vendor.sum linguist-language=Go-Checksums
-*.bat text eol=crlf
+
+*.go -text diff=golang
+
+# scripts directory contains shell scripts
+# without extensions, so we need to force
+scripts/** text=auto eol=lf
+
+# shell scripts should always have LF
+*.sh text eol=lf


.gitattributes
```diff
@@ -3,3 +3,14 @@ Dockerfile* linguist-language=Dockerfile
...
+
+*.env -text
+
+*.png binary
+*.ico binary
+*.gif binary
+*.jpg binary
+*.sum binary
+*.db binary
+*.tar binary
```
Benehiko, line 16: these are the binary files I could find in the repository using `grep -rIL .`
Some files such as `.golden` are also reported as binary, but have no content, so I guess it shouldn't be a problem?

Benehiko: another option might be to do something like below and let git decide on how to treat the file?
```
testdata/** text=auto
docs/** text=auto
vendor/** text=auto
```

thaJeztah: Yeah, tricky one;

- vendor: "+1" we should likely never touch anything there
- `testdata/` yes, these are fixtures, so some files may have intentional line-endings
- `docs/` hm... not sure; I think we produce those, and it's probably fine to enforce a specific line-ending?

ISTR there was some "smart" handling in git itself for certain known types (like images), but ... it's been a long time since I looked (and since I worked on Windows, so more often ran into such things) 🤔

thaJeztah: Some interesting comments in some of those files though; https://github.com/gitattributes/gitattributes/blob/master/.gitattributes#L1-L4

```
# THIS IS ONLY FOR THE gitattributes REPOSITORY.
# Handle line endings automatically for files detected as text
# and leave all files detected as binary untouched.
*               text=auto
```

So wondering if the png and other files are possibly due to some old git version being used (and git should otherwise automatically detect those?)

Benehiko: We could opt for more selective enforcement instead of the current approach, like so

```
scripts/** text=auto eol=lf
*.go -text diff=golang
*.bat text=auto eol=crlf
*.sh text=auto eol=lf
```

Benehiko: this would probably allow git to automatically handle binary files

.gitattributes
```diff
@@ -3,3 +3,14 @@ Dockerfile* linguist-language=Dockerfile
...
 *.bat text eol=crlf
+*.sh text eol=lf
+
+*.env -text
+
+*.png binary
+*.ico binary
+*.gif binary
+*.jpg binary
+*.sum binary
```
tianon, line 14: Aren't Go's `*.sum` files plain text where plain `\n` is fine/sane/correct, even?

Benehiko: yeah this is incorrect, I'll change the implementation to be opt-in instead of opt-out so we don't need to worry about files that are listed as binary from the vendor directory.

Benehiko:
I've updated this PR to be _opt-in_ to only certain files and the `scripts/` directory. This should allow git to sort out binary files for us.

thaJeztah (approved):
LGTM

.gitattributes
```diff
@@ -1,5 +1,14 @@
-* text=auto eol=lf
+* text=auto
+
 Dockerfile* linguist-language=Dockerfile
 vendor.mod linguist-language=Go-Module
 vendor.sum linguist-language=Go-Checksums
-*.bat text eol=crlf
+
+*.go -text diff=golang
```
tianon, line 7: Sorry, I'm still confused :sweat_smile:

Won't this _disable_ Git's line endings enforcement/correction for Go files?  That doesn't seem right?

Also, doesn't `diff=golang` require a `golang` diff tool to be configured? (I don't think Git applies meaning to that by default :sweat_smile:)

Benehiko: The original issue was enforcing `lf` line changes for scripts since the build on Windows machines would just fail when their git config was set set to `git config core.autocrlf true`. Since some scripts inside the `scripts/` directory has no file extension, which then lets git treat them as text files converting them to the `crlf` line ending. 

I made a PR to enforce line endings here https://github.com/docker/cli/pull/5216. But it was a bit overzealous and started causing some users to see changes to binary files. https://github.com/docker/cli/issues/5377

This PR pretty much reverts that enforcement for all files except the `scripts` directory.

For `go` I just used this template https://github.com/gitattributes/gitattributes/blob/master/Go.gitattributes. It doesn't require any additional tooling, see https://git-scm.com/docs/gitattributes#_setting_the_internal_diff_algorithm

I tested this branch on Windows and my Linux machine and it seems to be behaving.

tianon: Ah, I didn't realize `golang` was a built-in diff type, thanks!  I'm still confused by `-text` since `*.go` _is_ text files, but it should be harmless enough.

mbhadale:
Changes LGTM. Can I have an ETA, please? We are blocked due to this issue. Thanks.

mbhadale:
Hi, I don't see the changes in the master branch. https://github.com/docker/cli/blob/master/.gitattributes.

mbhadale:
Could you please share ETA?

Benehiko:
@mbhadale what is your git config values for `git config get core.autocrlf` and `git config get core.eol` and is this from a fork or from us? Do you see this only after cloning?

Benehiko:
Seems like this file was maybe committed with `CRLF` line endings. 

```console
⋊> ~/G/cli on master ◦ file cli/command/container/testdata/utf8.env                                   09:30:18
cli/command/container/testdata/utf8.env: Unicode text, UTF-8 (with BOM) text, with CRLF line terminators
⋊> ~/G/cli on master ◦ git log -p -- file cli/command/container/testdata/utf8.env                     09:30:25
commit 1630fc40f8f9f665a7292838b755520b1fefe307
Merge: 130222870 e5b7b7e87
Author: Daniel Nephin <dnephin@docker.com>
Date:   Mon Apr 17 17:40:59 2017 -0400

    Import docker/docker/cli
    
    Signed-off-by: Daniel Nephin <dnephin@gmail.com>
```

thaJeztah:
Looks like those fixtures were added to test issues on Windows, so it's likely they were indeed CRLF on purpose; original commit that added them was in moby/moby; https://github.com/moby/moby/commit/01d3a16f58118a3f6a6207fcb028bb279eb16e73

> Return error if env file contains non-ascii or utf8 bytes (for windows)
> 
> This fix tries to address the issue raised in 26179 where an env file
> with non-ascii or utf8 bytes will crash on windows platform.
> 
> The issue is two-fold:
> - Windows will adds a BOM mark at the begining with Notepad as the editor
> - Non-utf8 bytes can not be handled by env file parser.
> 
> This fix removes utf8 BOM marker if exists so that utf8 encoded env file
> could be processed.
> This fix also returns an error (instead of a runtime CreateProcess crash)
> if env file contains non-utf8 bytes, thus giving users better experiences.
> 
> Additional test cases has been added in unit tests.


- https://github.com/moby/moby/pull/26296

mbhadale:
This is what I get. 

-bash-4.2$ git config --list
push.default=simple
url.git@gitlab.eng.vmware.com:.insteadof=https://gitlab.eng.vmware.com/
core.autocrlf=input
core.repositoryformatversion=0
core.filemode=true
core.bare=false
core.logallrefupdates=true
remote.origin.url=git@gitlab.eng.vmware.com:core-build/mirrors_github_docker_cli.git
remote.origin.fetch=+refs/heads/*:refs/remotes/origin/*
branch.master.remote=origin
branch.master.merge=refs/heads/master

Benehiko:
> If you’re on a Linux or macOS system that uses LF line endings, then you don’t want Git to automatically convert them when you check out files; however, if a file with CRLF endings accidentally gets introduced, then you may want Git to fix it. You can tell Git to convert CRLF to LF on commit but not the other way around by setting core.autocrlf to input:
```console
$ git config --global core.autocrlf input
```
https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

This might be converting the `cli/command/container/testdata/utf8.env ` file on checkout for you.

mbhadale:
Still seeing this issue with core.autocrlf=false.


-bash-4.2$ git config --list
url.git@gitlab.eng.vmware.com:.insteadof=https://gitlab.eng.vmware.com/
core.autocrlf=false
core.repositoryformatversion=0
core.filemode=true
core.bare=false
core.logallrefupdates=true
remote.origin.url=https://github.com/docker/cli.git
remote.origin.fetch=+refs/heads/*:refs/remotes/origin/*
branch.master.remote=origin
branch.master.merge=refs/heads/master

Benehiko:
This is a bit difficult to debug since I cannot reproduce this on my Linux machine :see_no_evil: 
Can you maybe test the clone on a different environment?

Also re-read this section of the Git book
>You can tell Git to convert CRLF to LF on commit but not the other way around by setting core.autocrlf to input

Seems like this wouldn't do anything on checkout and only on commit.

mbhadale:
I still see the issue, I tried setting git config core.eol crlf as mentioned in the testing section of the issue.

As per this https://github.com/docker/cli/pull/5381#issuecomment-2337469277   file has CRLF ending. 

I think we should remove  * text=auto and specify files manually if they want git to handle line endings. 
Or we can add an exception for this env file.

utf8.env	-text

Let me know what you think.

Benehiko:
> I still see the issue, I tried setting git config core.eol crlf as mentioned in the testing section of the issue.
> 
> As per this [#5381 (comment)](https://github.com/docker/cli/pull/5381#issuecomment-2337469277) file has CRLF ending.
> 
> I think we should remove * text=auto and specify files manually if they want git to handle line endings. Or we can add an exception for this env file.
> 
> utf8.env -text
> 
> Let me know what you think.

I'm not so sure. I still cannot reproduce what you are seeing, nor can any of my colleagues. Tested on Linux and Windows.

We don't enforce any line endings on files with `* text=auto` which is intentional. The only enforcement is for the `scripts/` directory since they don't have a file extension causing Windows with certain git configs to add `CRLF` line endings on checkout, causing builds through Docker to fail. 

The `gitattributes` config aren't enforcing a particular line-ending (except for `scripts/`). I don't see why adding an exclusionary clause for this particular file is necessary since we don't even know why you are seeing a diff on clone for this particular file. 

In the issue you mention _"This issue is observed as intermittent but the frequency of failure is high."_. So it's not a consistent thing? It only happens now and again? 

You also mention there is a mirror branch _" We can not fix this in our mirror branch"_? Maybe the original `gitattributes` file has messed with your mirror since it did enforce the `lf` line-ending and maybe the diff you are seeing is a correction of the file? Meaning, `cli/command/container/testdata/utf8.env` has `CRLF` line endings, but your mirror converted it to `LF` line endings. So the diff being produced is showing that the _changed_ file has `CRLF` line endings?

Issue #5382, opened 2024-08-30
Docker login in 27.2.0 no longer overwrites existing config.json


thiago-gitlab:
### Description

In Docker 27.1.2, running `docker login` multiple times causes the authentication information in `~/.docker/config.json` to be updated with the latest successful execution.

In 27.2.0, the first successful authentication prevents `config.json` from being updated.

This is a breaking change because the existing expectation is that `docker login` always overwrites the authentication information.

### Reproduce

1. docker login repository-a
2. docker login repository-b
3. docker push repository-b/image

The push will fail since the authentication information recorded is for `repository-a`.

I've reproduced this in https://gitlab.com/gitlab-org/gitlab/-/issues/481917#note_2081566814 by removing `config.json` before performing the second authentication.

### Expected behavior

`docker login` always updates `config.json` with the successful authentication credentials.

### docker version

```bash
Client 27.2.0 and 27.1.2
```

### docker info

```bash
No relevant
```

### Additional Info

https://gitlab.com/gitlab-org/gitlab/-/issues/481917

thiago-gitlab:
A workaround is to downgrade to `27.1.2`. Another workaround is to delete the `config.json` before authenticating for a second time. [Example](https://gitlab.com/gitlab-org/secure/tests/thiagocsf-secure-tests/test-deploy-token/-/commit/e1a8bb6cb83b99f6f5626f79f96b7b12c05bffb5)

thaJeztah:
Thanks for reporting!

From that linked ticket;

```bash
$ docker login -u gitlab-ci-token -p $CI_JOB_TOKEN $CI_REGISTRY
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
WARNING! Your password will be stored unencrypted in /root/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credential-stores
Login Succeeded
$ docker pull registry.gitlab.com/gitlab-org/security-products/analyzers/container-scanning:latest
latest: Pulling from gitlab-org/security-products/analyzers/container-scanning
e4fff0779e6d: Pulling fs layer
,,,
,,,
Digest: sha256:84c7ee8800ce979ef5482d5ca21086f05fbf50d3a477f8194e4b5e87233a5c24
Status: Downloaded newer image for registry.gitlab.com/gitlab-org/security-products/analyzers/container-scanning:latest
[...]

PR #5383, opened 2024-08-30
login: use normalized hostname when storing

laurazard:
Fixes https://github.com/docker/cli/issues/5382

**- What I did**

Normalization/converting the registry address to just a hostname happens inside of [`command.GetDefaultAuthConfig`](https://github.com/docker/cli/blob/6273e65ceedc8544765ddc940a4a6dc7c2a2a491/cli/command/registry.go#L71-L73). Use this value for the rest of the login flow/storage.

**- How to verify it**

- `docker login -u [...] registry-1.docker.io`
- `docker login -u [...] registry-1.docker.io/repo/name`
- `cat ~/.docker/config.json`

Should only show one entry in `auths`, for `registry-1.docker.com`.

```markdown changelog
Fix issue causing login to not remove repository names from passed in registry addresses, resulting in credentials being stored under the wrong key.
```

### DIFF
--- a/cli/command/registry/login.go
+++ b/cli/command/registry/login.go
@@ -106,7 +106,7 @@ func runLogin(ctx context.Context, dockerCli command.Cli, opts loginOptions) err
 	// if we failed to authenticate with stored credentials (or didn't have stored credentials),
 	// prompt the user for new credentials
 	if err != nil || authConfig.Username == "" || authConfig.Password == "" {
-		response, err = loginUser(ctx, dockerCli, opts, authConfig.Username, serverAddress)
+		response, err = loginUser(ctx, dockerCli, opts, authConfig.Username, authConfig.ServerAddress)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


cli/command/registry/login.go
```diff
@@ -108,7 +108,7 @@ func runLogin(ctx context.Context, dockerCli command.Cli, opts loginOptions) err
 	// if we failed to authenticate with stored credentials (or didn't have stored credentials),
 	// prompt the user for new credentials
 	if err != nil || authConfig.Username == "" || authConfig.Password == "" {
-		response, err = loginUser(ctx, dockerCli, opts, authConfig.Username, serverAddress)
+		response, err = loginUser(ctx, dockerCli, opts, authConfig.Username, authConfig.ServerAddress)
```
thaJeztah, line 111: Do we have some test that we can add these cases on and/or mock something to have a regression test?

laurazard: I'll add them, I just wanted to do a PR quickly in case we wanted to release. I'll do it separately in a bit.

laurazard: Added some/refactored the tests there, PTAL if you get a sec

vvoland (approved):

thaJeztah (approved):
LGTM, thanks!

laurazard:
I'll open a backport

laurazard:
Oh! nvm, thanks @vvoland

PR #5386, opened 2024-09-02
docs: update docker login reference

dvdksn:

### DIFF
--- a/docs/reference/commandline/login.md
+++ b/docs/reference/commandline/login.md
@@ -1,60 +1,51 @@
 # login
 
 <!---MARKER_GEN_START-->
-Log in to a registry.
-If no server is specified, the default is defined by the daemon.
+Authenticate to a registry.
+Defaults to Docker Hub if no server is specified.
 
 ### Options
 
-| Name                                  | Type     | Default | Description                  |
-|:--------------------------------------|:---------|:--------|:-----------------------------|
-| `-p`, `--password`                    | `string` |         | Password                     |
-| [`--password-stdin`](#password-stdin) | `bool`   |         | Take the password from stdin |
-| `-u`, `--username`                    | `string` |         | Username                     |
+| Name                                         | Type     | Default | Description                  |
+|:---------------------------------------------|:---------|:--------|:-----------------------------|
+| `-p`, `--password`                           | `string` |         | Password                     |
+| [`--password-stdin`](#password-stdin)        | `bool`   |         | Take the password from stdin |
+| [`-u`](#username), [`--username`](#username) | `string` |         | Username                     |
 
 
 <!---MARKER_GEN_END-->
 
 ## Description
 
-Log in to a registry.
+Authenticate to a registry.
 
-## Examples
+You can authenticate to any public or private registry for which you have
+credentials. Authentication may be required for pulling and pushing images.
+Other commands, such as `docker scout` and `docker build`, may also require
+authentication to access subscription-only features or data related to your
+Docker organization.
 
-### Login to a self-hosted registry
+Authentication credentials are stored in the configured [credential
+store](#credential-stores). If you use Docker Desktop, credentials are
+automatically saved to the native keychain of your operating system. If you're
+not using Docker Desktop, you can configure the credential store in the Docker
+configuration file, which is located at `$HOME/.docker/config.json` on Linux or
+`%USERPROFILE%/.docker/config.json` on Windows. If you don't configure a
+credential store, Docker stores credentials in the `config.json` file in a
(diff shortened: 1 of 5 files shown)


docs/reference/commandline/login.md
```diff
@@ -162,19 +157,93 @@ the credentials from the default store.
...
+If you want to authenticate to a self-hosted registry you can specify this by
+adding the server name.
+
+```console
+$ docker login registry.example.com
+```
+
+By default, the `docker login` command assumes that the registry listens on
+port `5000`. If the registry listens on a different port, you can specify it
+by adding the port number to the server name.
```
thaJeztah, line 208: I don't think it assumes port 5000, but will default to `443` (TLS) / `80` (non-TLS). (The examples for the registry tend to use port 5000 though, so maybe that's what you assumed here?)

dvdksn: ops yea my bad!

docs/reference/commandline/login.md
```diff
@@ -1,60 +1,51 @@
...
-You can log in to any public or private repository for which you have
-credentials.  When you log in, the command stores credentials in
-`$HOME/.docker/config.json` on Linux or `%USERPROFILE%/.docker/config.json` on
-Windows, via the procedure described below.
+You can authenticate to a registry using a username and password or access
+token. Docker Hub also supports a web-based login flow, which signs you in to
+your Docker account without entering your password. For Docker Hub, the `docker
+login` command uses a device code flow by default, unless the `--username` flag
+is specified. The device code flow is a secure way to sign in. See
+[Authenticate to Docker Hub using device code](#authenticate-to-docker-hub-using-device-code).
```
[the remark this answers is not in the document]
dvdksn: Yeah we could... @laurazard do we have some special sauce for handling login differently for non-TTY cases?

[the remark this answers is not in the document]
laurazard: Ah, we don't. With `--username`, without setting the password we do error out if non-tty:
```console
$ docker login -u bork 0>/dev/null
Error: Cannot perform an interactive login from a non TTY device
```
But with the device-code flow, we don't:
```console
$ docker login 0>/dev/null

USING WEB BASED LOGIN
To sign in with credentials on the command line, use 'docker login -u <username>'

Your one-time device confirmation code is: JZJG-ZWNM
Press ENTER to open your browser or submit your device code here: https://login.docker.com/activate

Waiting for authentication in the browser…
```

Although in the first case, there's no way for the user to login when non-interactive, but with the device-code flow a user could open the URL and login on another device. Maybe we should though 🤔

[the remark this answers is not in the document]
laurazard: Ah, we don't. With `--username`, without setting the password we do error out if non-tty:
```console
$ docker login -u bork 0>/dev/null
Error: Cannot perform an interactive login from a non TTY device
```
But with the device-code flow, we don't:
```console
$ docker login 0>/dev/null

USING WEB BASED LOGIN
To sign in with credentials on the command line, use 'docker login -u <username>'

Your one-time device confirmation code is: <--->
Press ENTER to open your browser or submit your device code here: https://login.docker.com/activate

Waiting for authentication in the browser…
```

Although in the first case, there's no way for the user to login when non-interactive, but with the device-code flow a user could open the URL and login on another device. Maybe we should though 🤔

docs/reference/commandline/login.md
```diff
@@ -1,60 +1,51 @@
...
-or log-files.
-
-The following example reads a password from a file, and passes it to the
-`docker login` command using `STDIN`:
-
-```console
-$ cat ~/my_password.txt | docker login --username foo --password-stdin
-```
+You can authenticate to any public or private registry for which you have
+credentials. Authentication is required for pulling and pushing images. Other
```
laurazard, line 23: Authentication isn't always required (pulling public images from certain registries).
```suggestion
credentials. Authentication may be required for pulling and pushing images. Other
```

docs/reference/commandline/login.md
```diff
@@ -162,19 +157,93 @@ the credentials from the default store.
...
-    "unicorn.example.io": "vcbait"
+    "myregistry.example.com": "secretservice",
+    "docker.internal.example": "pass",
   }
 }
 ```
 
+## Examples
+
+### Authenticate to Docker Hub using device code
```
laurazard, line 173:
```suggestion
### Authenticate to Docker Hub with web-based login
```

laurazard (approved):
LGTM! Thanks for this @dvdksn

thaJeztah (approved):
LGTM

thaJeztah:
@dvdksn can you open a cherry-pick as well?

PR #5407, opened 2024-09-05
vendor.mod: put github.com/pkg/browser in the right group

thaJeztah:
- relates to https://github.com/moby/buildkit/pull/5304#issuecomment-2331294601

commit fcfdd7b91fbbd41dfdc80d5d66f28337916d935c (https://github.com/docker/cli/pull/5344) introduced github.com/pkg/browser as a direct dependency, but it ended up in the group for indirect dependencies.

PR #5408, opened 2024-09-05
[27.x backport] vendor.mod: put github.com/pkg/browser in the right group

thaJeztah:
- backport https://github.com/docker/cli/pull/5407
- relates to https://github.com/moby/buildkit/pull/5304#issuecomment-2331294601

commit fcfdd7b91fbbd41dfdc80d5d66f28337916d935c (https://github.com/docker/cli/pull/5344) introduced github.com/pkg/browser as a direct dependency, but it ended up in the group for indirect dependencies.

### DIFF
--- a/vendor.mod
+++ b/vendor.mod
@@ -32,6 +32,7 @@ require (
 	github.com/morikuni/aec v1.0.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.0
+	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.8.1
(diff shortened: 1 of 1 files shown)


thaJeztah:
Flaky test?

```
=== Failed
=== FAIL: e2e/registry TestOauthLogin (1.00s)
    login_test.go:28: terminating PID 10523
    login_test.go:33: assertion failed: expression is false: strings.Contains(string(output), "USING WEB-BASED LOGIN"): Failed to start web-based login - falling back to command line login...
        
        Username: 
```

laurazard:
```
Failed to start web-based login - falling back to command line login...
```

thaJeztah:
I just noticed that we're swallowing the error there, which ... probably is fine, but wondering if we need some what to present it 🤔 (although not sure if we want a `log.Debug` because they're so ugly on the commandline.. maybe?

https://github.com/docker/cli/blob/f75abad8a1d3bdc2ecf6705eac7634d37aac988f/cli/command/registry/login.go#L171-L177


I wish the CLI had a location to log to (without that meaning "print and show the user")

laurazard:
Yeah, we could do something like a `log.Debug`.

thaJeztah:
It's a tricky one, due to the nature of the cli being short-lived; we'll be potentially opening a can of worms to handle concurrency, log-rotation, and multiple instances of the cli probably needing something to associate logs with each invocation.

Well, or have a new file created at all times, but we'd still need something to clean it up. Sending to a logging service would be ideal, but not available in all cases.

PR #5422, opened 2024-09-10
info: stop printing "Expected" commits

thaJeztah:
relates to:

- https://github.com/moby/moby/pull/27955
- https://github.com/moby/moby/pull/37968
- https://github.com/moby/moby/pull/42776
- https://github.com/moby/moby/issues/38739

The `Commit` type was introduced in https://github.com/moby/moby/commit/2790ac68b32b399c872de88388bdccc359ed7a88, to assist triaging issues that were reported with an incorrect version of runc or containerd. At the time, both `runc` and `containerd` were not yet stable, and had to be built from a specific commit to guarantee compatibility.

We encountered various situations where unexpected (and incompatible) versions of those binaries were packaged, resulting in hard to trace bug-reports. For those situations, a "expected" version was set at compile time, to indicate if the version installed was different from the expected version;

    docker info
    ...
    runc version: a592beb5bc4c4092b1b1bac971afed27687340c5 (expected: 69663f0bd4b60df09991c08812a60108003fa340)

Both `runc` and `containerd` are stable now, and docker 19.03 and up set the expected version to the actual version since https://github.com/moby/moby/commit/c65f0bd13c85d29087419fa555281311091825e7 and 23.0 did the same for the `init` binary https://github.com/moby/moby/commit/b585c64e2b01f924fc358fe059871baa469bb460, to prevent the CLI from reporting "unexpected version".

In short; the `Expected` fields no longer serves a real purpose, so we should no longer print it.

```markdown changelog


```

### DIFF
--- a/cli/command/system/info.go
+++ b/cli/command/system/info.go
@@ -273,21 +273,9 @@ func prettyPrintServerInfo(streams command.Streams, info *dockerInfo) []error {
 
 	if info.OSType == "linux" {
 		fprintln(output, " Init Binary:", info.InitBinary)
-
-		for _, ci := range []struct {
-			Name   string
-			Commit system.Commit
-		}{
-			{"containerd", info.ContainerdCommit},
-			{"runc", info.RuncCommit},
-			{"init", info.InitCommit},
-		} {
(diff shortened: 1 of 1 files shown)


krissetto (approved):
LGTM

PR #5434, opened 2024-09-13
Remove leading and trailing double quotes from environment variable values in docker run commands

peacewalker122:
### **- What I did**

Implemented functionality to automatically remove leading and trailing double quotes from environment variable values when they are set via the Docker CLI (`docker run`). This enhancement prevents issues caused by unintended double quotes in environment variable values, ensuring that the variables are set correctly inside containers.

### **- How I did it**

- Modified the environment variable parsing logic in `opts/file.go`,`opts/parse.go` to strip leading and trailing double quotes from environment variable values.
- Used `strings.Trim` with `"` as the cutset to remove only the leading and trailing double quotes without affecting any internal quotes within the value.
- Ensured that environment variable names are not altered.
- Added unit tests in `cli/command/container/opts_test.go` to verify that leading and trailing double quotes are correctly removed.
- Verified that values without leading/trailing double quotes remain unaffected.

### **- How to verify it**

1. **Build the Docker CLI** from the modified source code.

2. **Run a container** with an environment variable that includes leading and trailing double quotes in its value:

   ```bash
   ./docker run --env MY_VAR="some value with spaces" alpine sh -c 'echo "[$MY_VAR]"'
   ```
   
3. Expected output: `some value with spaces`

### DIFF
--- a/cli/command/container/opts_test.go
+++ b/cli/command/container/opts_test.go
@@ -209,7 +209,10 @@ func TestParseWithVolumes(t *testing.T) {
 	}
 
 	// Two bind mounts.
-	arr, tryit = setupPlatformVolume([]string{`/hostTmp:/containerTmp`, `/hostVar:/containerVar`}, []string{os.Getenv("ProgramData") + `:c:\ContainerPD`, os.Getenv("TEMP") + `:c:\containerTmp`})
+	arr, tryit = setupPlatformVolume(
+		[]string{`/hostTmp:/containerTmp`, `/hostVar:/containerVar`},
+		[]string{os.Getenv("ProgramData") + `:c:\ContainerPD`, os.Getenv("TEMP") + `:c:\containerTmp`},
+	)
 	if _, hostConfig, _ := mustParse(t, tryit); hostConfig.Binds == nil || compareRandomizedStrings(hostConfig.Binds[0], hostConfig.Binds[1], arr[0], arr[1]) != nil {
 		t.Fatalf("Error parsing volume flags, `%s and %s` did not mount-bind correctly. Received %v", arr[0], arr[1], hostConfig.Binds)
 	}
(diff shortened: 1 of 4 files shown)


thaJeztah:
Why are you putting the `'MY_VAR="some value with spaces"'` in single quotes? I think this should work it you remove those;

```bash
docker run --rm --env MY_VAR="some value with spaces" alpine printenv MY_VAR
some value with spaces
```

peacewalker122:
> Why are you putting the `'MY_VAR="some value with spaces"'` in single quotes? I think this should work it you remove those;
> 
> ```shell
> docker run --rm --env MY_VAR="some value with spaces" alpine printenv MY_VAR
> some value with spaces
> ```

Hi @thaJeztah. Yes you're correct it doesn't need to wrapped on that single qoutes. Also i already update the PR description on **How to Verify It** With the exact problem i found.

thaJeztah:
In that case; is there anything to fix here? Because this seems to be working as expected;

```bash
docker create -e MY_VAR="hello world" -e OTHER_VAR=hello-world --name mycontainer alpine

docker inspect --format '{{ json .Config.Env }}' mycontainer
["MY_VAR=hello world","OTHER_VAR=hello-world","PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"]
```


I see your code also changes behavior for `--env-file`, which would be related to

- https://github.com/docker/cli/issues/3630
- https://github.com/docker/cli/pull/3660

However, I think that behavior was by design (`--env-file` does not interpolate values, and takes a pure `name=value` format)

peacewalker122:
> In that case; is there anything to fix here? Because this seems to be working as expected;
> 
> ```shell
> docker create -e MY_VAR="hello world" -e OTHER_VAR=hello-world --name mycontainer alpine
> 
Yes, you're right it works like expected.

> I see your code also changes behavior for `--env-file`, which would be related to 
> However, I think that behavior was by design (`--env-file` does not interpolate values, and takes a pure `name=value` format)

Yes, this code were firstly intended for that purpose, If so I'll close the PR. Thanks for the review @thaJeztah

PR #5456, opened 2024-09-20
telemetry: fix early meterprovider shutdown

laurazard:
**- What I did**

In 4b5a196fee11e82e15d2c2c36ba2883e5717fde3, we changed the CLI global meter provider shutdown in order to handle any error returned by the metric export.

Unfortunately, we dropped a `defer` during the fix, which causes the meter provider to be immediately shutdown after being created and metrics to not be collected/exporter.

**- How I did it**

Re-defer the new Meter Provider shutdown call.

**- How to verify it**

```markdown changelog
Fix issue causing CLI OTel metrics to not be collected.
```

Screenshot 2024-09-20 at 00 53 09


thaJeztah (approved):
LGTM 🙈🤦‍♂️

PR #5467, opened 2024-09-21
command: add tests for container diff and rename

Stavrospanakakis:
**- What I did**
This commit adds tests for the commands `docker diff` and `docker rename`. Also, it creates the mock methods of the docker client `ContainerDiff` and `ContainerRename` so they can be used in the tests.

For `docker diff,` it covers the cases that:
 - the command runs successfully
 - the client returns an error
 - the container id is empty

For `docker rename`, it covers the cases that:
 - the command runs successfully
 - the container old name is empty
 - the container new name is empty
 - the client returns an error

**- How I did it**
I read all the test cases developed in `cli/command/container` directory and did the same for `diff.go` and 
`rename.go` using a `fakeClient` and mocking the client `ContainerDiff` and `ContainerRename` methods.

**- How to verify it**
Run: `make test` 

```markdown changelog
Add tests for container diff and rename commands
```

### DIFF
--- a/cli/command/container/diff_test.go
+++ b/cli/command/container/diff_test.go
@@ -1,0 +1,93 @@
+package container
+
+import (
+	"context"
+	"io"
+	"strings"
+	"testing"
+
+	"github.com/docker/cli/internal/test"
+	"github.com/docker/docker/api/types/container"
+	"github.com/pkg/errors"
+	"gotest.tools/v3/assert"
+	is "gotest.tools/v3/assert/cmp"
+)
+
+func TestRunDiff(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{
+		containerDiffFunc: func(
+			ctx context.Context,
+			containerID string,
+		) ([]container.FilesystemChange, error) {
+			return []container.FilesystemChange{
+				{
+					Kind: container.ChangeModify,
+					Path: "/path/to/file0",
+				},
+				{
+					Kind: container.ChangeAdd,
+					Path: "/path/to/file1",
+				},
+				{
+					Kind: container.ChangeDelete,
+					Path: "/path/to/file2",
+				},
+			}, nil
+		},
+	})
+
+	cmd := NewDiffCommand(cli)
+	cmd.SetOut(io.Discard)
+
+	cmd.SetArgs([]string{"container-id"})
+
+	err := cmd.Execute()
+	assert.NilError(t, err)
+
+	diff := strings.SplitN(cli.OutBuffer().String(), "\n", 3)
+	assert.Assert(t, is.Len(diff, 3))
+
+	file0 := strings.TrimSpace(diff[0])
+	file1 := strings.TrimSpace(diff[1])
+	file2 := strings.TrimSpace(diff[2])
+
+	assert.Check(t, is.Equal(file0, "C /path/to/file0"))
+	assert.Check(t, is.Equal(file1, "A /path/to/file1"))
+	assert.Check(t, is.Equal(file2, "D /path/to/file2"))
+}
+
+func TestRunDiffClientError(t *testing.T) {
+	clientError := errors.New("client error")
+
+	cli := test.NewFakeCli(&fakeClient{
+		containerDiffFunc: func(
+			ctx context.Context,
+			containerID string,
+		) ([]container.FilesystemChange, error) {
+			return nil, clientError
+		},
+	})
+
+	cmd := NewDiffCommand(cli)
+	cmd.SetOut(io.Discard)
+	cmd.SetErr(io.Discard)
+
+	cmd.SetArgs([]string{"container-id"})
+
+	err := cmd.Execute()
+	assert.ErrorIs(t, err, clientError)
+}
+
+func TestRunDiffEmptyContainerError(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{})
+
+	cmd := NewDiffCommand(cli)
+	cmd.SetOut(io.Discard)
+	cmd.SetErr(io.Discard)
+
+	containerID := ""
+	cmd.SetArgs([]string{containerID})
+
+	err := cmd.Execute()
+	assert.Error(t, err, "Container name cannot be empty")
+}
--- a/cli/command/container/rename_test.go
+++ b/cli/command/container/rename_test.go
@@ -1,0 +1,77 @@
+package container
+
+import (
+	"context"
+	"io"
+	"testing"
+
+	"github.com/docker/cli/internal/test"
+	"github.com/pkg/errors"
+	"gotest.tools/v3/assert"
+	is "gotest.tools/v3/assert/cmp"
+)
+
+func TestRunRename(t *testing.T) {
+	testcases := []struct {
+		doc, oldName, newName, expectedErr string
+	}{
+		{
+			doc:         "success",
+			oldName:     "oldName",
+			newName:     "newName",
+			expectedErr: "",
+		},
+		{
+			doc:         "empty old name",
+			oldName:     "",
+			newName:     "newName",
+			expectedErr: "Error: Neither old nor new names may be empty",
+		},
+		{
+			doc:         "empty new name",
+			oldName:     "oldName",
+			newName:     "",
+			expectedErr: "Error: Neither old nor new names may be empty",
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.doc, func(t *testing.T) {
+			cli := test.NewFakeCli(&fakeClient{
+				containerRenameFunc: func(ctx context.Context, oldName, newName string) error {
+					return nil
+				},
+			})
+
+			cmd := NewRenameCommand(cli)
+			cmd.SetOut(io.Discard)
+			cmd.SetErr(io.Discard)
+			cmd.SetArgs([]string{tc.oldName, tc.newName})
+
+			err := cmd.Execute()
+
+			if tc.expectedErr != "" {
+				assert.ErrorContains(t, err, tc.expectedErr)
+			} else {
+				assert.NilError(t, err)
+			}
+		})
+	}
+}
+
+func TestRunRenameClientError(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{
+		containerRenameFunc: func(ctx context.Context, oldName, newName string) error {
+			return errors.New("client error")
+		},
+	})
+
+	cmd := NewRenameCommand(cli)
+	cmd.SetOut(io.Discard)
+	cmd.SetErr(io.Discard)
+	cmd.SetArgs([]string{"oldName", "newName"})
+
+	err := cmd.Execute()
+
+	assert.Check(t, is.Error(err, "Error: failed to rename container named oldName"))
+}
--- a/cli/command/container/client_test.go
+++ b/cli/command/container/client_test.go
@@ -38,6 +38,8 @@ type fakeClient struct {
 	containerKillFunc       func(ctx context.Context, containerID, signal string) error
 	containerPruneFunc      func(ctx context.Context, pruneFilters filters.Args) (container.PruneReport, error)
 	containerAttachFunc     func(ctx context.Context, containerID string, options container.AttachOptions) (types.HijackedResponse, error)
+	containerDiffFunc       func(ctx context.Context, containerID string) ([]container.FilesystemChange, error)
+	containerRenameFunc     func(ctx context.Context, oldName, newName string) error
 	Version                 string
 }
 
@@ -180,4 +182,20 @@ func (f *fakeClient) ContainerAttach(ctx context.Context, containerID string, op
 		return f.containerAttachFunc(ctx, containerID, options)
 	}
 	return types.HijackedResponse{}, nil
+}
+
+func (f *fakeClient) ContainerDiff(ctx context.Context, containerID string) ([]container.FilesystemChange, error) {
+	if f.containerDiffFunc != nil {
+		return f.containerDiffFunc(ctx, containerID)
+	}
+
+	return []container.FilesystemChange{}, nil
+}
+
+func (f *fakeClient) ContainerRename(ctx context.Context, oldName, newName string) error {
+	if f.containerRenameFunc != nil {
+		return f.containerRenameFunc(ctx, oldName, newName)
+	}
+
+	return nil
 }


cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	err := cmd.Execute()
+	assert.Assert(t, err == nil)
+
+	diff := strings.SplitN(cli.OutBuffer().String(), "\n", 3)
+	file0 := strings.TrimSpace(diff[0])
+	file1 := strings.TrimSpace(diff[1])
+	file2 := strings.TrimSpace(diff[2])
+
+	assert.Assert(t, len(diff) == 3)
```
laurazard, line 52: Could move this assert on the size of `diff` to line 48, before directly accessing `diff[0]`, `diff[1]`, `diff[2]` so that if `diff` doesn't return the expected elements we handle that better than panicking from accessing an out of bounds index.

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+				},
+			}, nil
+		},
+	})
+
+	cmd := NewDiffCommand(cli)
+	cmd.SetOut(io.Discard)
+
+	containerId := "a61d6b418006"
+	cmd.SetArgs([]string{containerId})
```
laurazard, line 42:
```suggestion
	cmd.SetArgs([]string{"a-container-id"})
```

and then we could also check that `containerDiffFunc` gets called with `a-container-id`.

Could also do
```
const containerId = "a61d6b418006"
...
cmd.SetArgs([]string{containerId})
```

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+		) ([]container.FilesystemChange, error) {
+			return nil, errors.New("client error")
+		},
+	})
+
+	cmd := NewDiffCommand(cli)
+	cmd.SetOut(io.Discard)
+
+	containerId := "a61d6b418006"
+	cmd.SetArgs([]string{containerId})
```
laurazard, line 72:
```suggestion
	cmd.SetArgs([]string{"a61d6b418006"})
```

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+func TestRunDiff(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{
+		containerDiffFunc: func(
+			ctx context.Context,
+			containerID string,
+		) ([]container.FilesystemChange, error) {
+			return []container.FilesystemChange{
+				{
+					Kind: 0, // Modified ("C")
```
thaJeztah, line 23: We can probably use the types that are defined in the `api/types/container` package for this; https://github.com/docker/cli/blob/baa7a9f7ad378204e8d3ca52121271402ed38a66/vendor/github.com/docker/docker/api/types/container/change_types.go#L4-L9

cli/command/container/diff_test.go
thaJeztah, line 51: These could panic (out of range) if `diff` does not have 3 items.

Instead, might be good to assert the length before proceeding;

```go
diff := strings.SplitN(cli.OutBuffer().String(), "\n", 3)
assert.Assert(t, is.Len(diff, 3))
```

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	diff := strings.SplitN(cli.OutBuffer().String(), "\n", 3)
+	file0 := strings.TrimSpace(diff[0])
+	file1 := strings.TrimSpace(diff[1])
+	file2 := strings.TrimSpace(diff[2])
+
+	assert.Assert(t, len(diff) == 3)
+	assert.Assert(t, file0 == "C /path/to/file0")
+	assert.Assert(t, file1 == "A /path/to/file1")
+	assert.Assert(t, file2 == "D /path/to/file2")
```
thaJeztah, line 54: Where possible, we try to not fail early; in this case, we can use `assert.Check` instead of `assert.Assert` so that we still show all 3 results, regardless if one of them fails;


```go
assert.Check(t, is.Equal(file0, "C /path/to/file0"))
assert.Check(t, is.Equal(file1, "A /path/to/file1"))
assert.Check(t, is.Equal(file2, "D /path/to/file2"))
```

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func TestRunDiffClientError(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{
+		containerDiffFunc: func(
+			ctx context.Context,
+			containerID string,
+		) ([]container.FilesystemChange, error) {
+			return nil, errors.New("client error")
```
thaJeztah, line 63: I think we can define the test-error as part of the test, so that we can assert what we returned matches the error;

```go
expectedError := errors.New("client error")

	containerDiffFunc: func(
		ctx context.Context,
		containerID string,
	) ([]container.FilesystemChange, error) {
		return nil, expectedError
```

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+	cmd.SetOut(io.Discard)
+
+	containerId := "a61d6b418006"
+	cmd.SetArgs([]string{containerId})
+
+	err := cmd.Execute()
+	assert.Assert(t, err != nil)
+
+	expected := "client error"
+	assert.ErrorContains(t, err, expected)
```
thaJeztah, line 76: The first assert is redundant, as `ErrorContains` will fail if `err` is nil.

But if we define the expected error, then we can probably combine both, and assert the error using `assert.ErrorIs`;

```go
assert.ErrorIs(t, err, expectedError)
```

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+		containerDiffFunc: func(
+			ctx context.Context,
+			containerID string,
+		) ([]container.FilesystemChange, error) {
+			return nil, errors.New("client error")
+		},
+	})
+
+	cmd := NewDiffCommand(cli)
+	cmd.SetOut(io.Discard)
```
thaJeztah, line 68: Can you also discard the error output here? 

```go
cmd.SetOut(io.Discard)
cmd.SetErr(io.Discard)
```

It prevents the error that's part of the test from being printed, which can make it confusing when reading test logs;

```console
=== RUN   TestRunDiffClientError
Error: client error
--- PASS: TestRunDiffClientError (0.00s)
PASS
```

Suppressing the output still makes the test pass, but doesn't print the output;

```
=== RUN   TestRunDiffClientError
--- PASS: TestRunDiffClientError (0.00s)
PASS
```

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	containerId := "a61d6b418006"
+	cmd.SetArgs([]string{containerId})
+
+	err := cmd.Execute()
```
thaJeztah: Given that the container-id can be "anything" as long as it's not empty (for  this test) perhaps simplify and remove the intermediate variable;

```go
cmd.SetArgs([]string{"container-id"})
err := cmd.Execute()
```

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	expected := "client error"
+	assert.ErrorContains(t, err, expected)
+}
+
+func TestRunDiffEmptyContainerError(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{})
+
+	cmd := NewDiffCommand(cli)
+	cmd.SetOut(io.Discard)
```
thaJeztah, line 83: Can you suppress stderr here as well?

```
	cmd := NewDiffCommand(cli)
	cmd.SetOut(io.Discard)
```

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+	})
+
+	cmd := NewDiffCommand(cli)
+	cmd.SetOut(io.Discard)
+
+	containerId := "a61d6b418006"
+	cmd.SetArgs([]string{containerId})
+
+	err := cmd.Execute()
+	assert.Assert(t, err == nil)
```
thaJeztah, line 44: Can you use `assert.NilError(t, err)` here?

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+		},
+	})
+
+	cmd := NewDiffCommand(cli)
+	cmd.SetOut(io.Discard)
+
+	containerId := "a61d6b418006"
+	cmd.SetArgs([]string{containerId})
+
+	err := cmd.Execute()
```
thaJeztah: Same comment here as for the other test; it's probably fine to just inline this here (and use a value that describes what it is);

```go
cmd.SetArgs([]string{"container-id"})
err := cmd.Execute()
```

cli/command/container/diff_test.go
```diff
@@ -0,0 +1,95 @@
...
+	cmd.SetOut(io.Discard)
+
+	containerId := ""
+	cmd.SetArgs([]string{containerId})
+
+	err := cmd.Execute()
+	assert.Assert(t, err != nil)
+
+	expected := "Container name cannot be empty"
+	assert.ErrorContains(t, err, expected)
```
thaJeztah, line 92: The first assert is also redundant here (`ErrorContains` will fail if `err` is nil (`assertion failed: expected an error, got nil`)

cli/command/container/rename_test.go
```diff
@@ -0,0 +1,78 @@
...
+			return nil
+		},
+	})
+
+	cmd := NewRenameCommand(cli)
+	cmd.SetOut(io.Discard)
+	cmd.SetArgs([]string{"oldName", "newName"})
+	err := cmd.Execute()
+
+	assert.Assert(t, err == nil)
```
thaJeztah, line 25: This can be `assert.NilError`

cli/command/container/rename_test.go
```diff
@@ -0,0 +1,78 @@
...
+	cmd.SetOut(io.Discard)
+	cmd.SetArgs([]string{"oldName", ""})
+
+	err := cmd.Execute()
+
+	assert.Assert(t, err != nil)
+
+	expected := "Error: Neither old nor new names may be empty"
+	assert.ErrorContains(t, err, expected)
+}
```
laurazard, line 59:
```suggestion
func TestRunRenameInputs(t *testing.T) {
	testCases := []struct {
		args []string
		expectedErr string
	} {
		{
			args: []string{"", "newName"},
			expectedErr: "Error: Neither old nor new names may be empty",
		},
		{
			args: []string{"oldName", ""}
			expectedErr: "Error: Neither old nor new names may be empty",
		},
	}
	
	for _, tc := range testCases {
		cli := test.NewFakeCli(&fakeClient{})

		cmd := NewRenameCommand(cli)
		cmd.SetOut(io.Discard)
		cmd.SetArgs(tc.args)

		err := cmd.Execute()

		assert.Assert(t, err != nil)
		assert.ErrorContains(t, err, tc.expectedErr)
	}
}
```

Could merge these two into one, using [table driven tests](https://go.dev/wiki/TableDrivenTests)!

laurazard (commented):
Thanks for the contribution! Left a couple of nits to make the tests a little tidier, but overall LGTM

cli/command/container/rename_test.go
thaJeztah: These could probably be combined to a single table-test; something like;

```go
func TestRunRename(t *testing.T) {
		containerRenameFunc: func(ctx context.Context, oldName, newName string) error {
			return nil
		},
	})

	tests := []struct {
		doc, oldName, newName, expectedErr string
	}{
		{
			doc:     "success",
			oldName: "oldName",
			newName: "newName",
		},
		{
			doc:         "empty old name",
			newName:     "newName",
			expectedErr: "Error: Neither old nor new names may be empty",
		},
		{
			doc:         "empty new name",
			oldName:     "oldName",
			expectedErr: "Error: Neither old nor new names may be empty",
		},
	}
	for _, tc := range tests {
		tc := tc
		t.Run(tc.doc, func(t *testing.T) {
			cmd := NewRenameCommand(cli)
			cmd.SetOut(io.Discard)
			cmd.SetErr(io.Discard)
			cmd.SetArgs([]string{tc.oldName, tc.newName})
			err := cmd.Execute()
			if tc.expectedErr != "" {
				assert.ErrorContains(t, err, tc.expectedErr)
			} else {
				assert.NilError(t, err)
			}
			assert.Equal(t, cli.OutBuffer().String(), "", "expected no output")
		})
	}
}
```

cli/command/container/rename_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+func TestRunRenameClientError(t *testing.T) {
+	cli := test.NewFakeCli(&fakeClient{
+		containerRenameFunc: func(ctx context.Context, oldName, newName string) error {
+			return errors.New("client error")
+		},
+	})
+
+	cmd := NewRenameCommand(cli)
+	cmd.SetOut(io.Discard)
```
thaJeztah, line 68: can you suppress stderr output here as well?

```go
cmd.SetOut(io.Discard)
cmd.SetErr(io.Discard)
```

cli/command/container/rename_test.go
```diff
@@ -0,0 +1,78 @@
...
+	cmd.SetArgs([]string{"oldName", "newName"})
+
+	err := cmd.Execute()
+
+	assert.Assert(t, err != nil)
+
+	expected := "Error: failed to rename container named oldName"
+	assert.ErrorContains(t, err, expected)
```
thaJeztah, line 76: First assert is redundant here as well;

```go
err := cmd.Execute()
expected := "Error: failed to rename container named oldName"
assert.ErrorContains(t, err, expected)
```

thaJeztah (changes requested):
thanks for contributing! I left comments/ suggestions inline; let me know what you think

Stavrospanakakis:
@thaJeztah @laurazard Thank you for the code review:) I committed the suggested changes from GitHub UI and will now work on the code review comments. When this pull request is ready after all the required changes, I will rebase and sign to match the [Contribution guidelines](https://github.com/docker/cli/blob/master/CONTRIBUTING.md).

laurazard (changes requested):

laurazard (approved):
LGTM

Stavrospanakakis:
@laurazard @thaJeztah I added all the requested changes and squashed the git commits.

Stavrospanakakis:
@laurazard @thaJeztah I renamed `containerId` to `containerID` because the CI failed with this error:
```
#16 [lint 4/4] RUN --mount=type=bind,target=.     --mount=type=cache,target=/root/.cache         golangci-lint run
#16 63.82 cli/command/container/diff_test.go:88:2: ST1003: var containerId should be containerID (stylecheck)
#16 63.82 	containerId := ""
#16 63.82 	^
#16 ERROR: process "/bin/sh -c golangci-lint run" did not complete successfully: exit code: 1
------
63.82 cli/command/container/diff_test.go:88:2: ST1003: var containerId should be containerID (stylecheck)
63.82 	containerId := ""
63.82 	^
```

thaJeztah (approved):
LGTM!


sorry for the delay; I somehow forgot about this one 🙈

PR #5480, opened 2024-09-26
cli/container: use github.com/moby/sys/capability for completions

thaJeztah:
- relates to https://github.com/docker/cli/commit/f30158dbf873555e314b51d67ec2b18d6128f25d / https://github.com/docker/cli/pull/5238

We used a hard-coded list of capabilities that we copied from containerd, but the new "capability" package allows use to have a maintained list of capabilities.

There's likely still some improvements to be made;

First of all, the capability package could provide a function to get the list of strings.

On the completion-side, we need to consider what format is most convenient; currently we use the canonical name (uppercase and "CAP_" prefix), however, tab-completion is case-sensitive by default, so requires the user to type uppercase letters to filter the list of options.

Bash completion provides a `completion-ignore-case on` option to make completion case-insensitive (https://askubuntu.com/a/87066), but it looks to be a global option; the current cobra.CompletionOptions also don't provide this as an option to be used in the generated completion-script.

Fish completion has `smartcase` (by default?) which matches any case if all of the input is lowercase.

Zsh does not have a dedicated option, but allows setting matching-rules (see https://superuser.com/a/1092328).

**- What I did**

**- How I did it**

**- How to verify it**

```markdown changelog


```


laurazard (approved):

cli/command/container/completion.go
```diff
@@ -1,70 +1,45 @@
...
+// Zsh does not appear have a dedicated option, but allows setting matching-rules
+// (see https://superuser.com/a/1092328).
+var allLinuxCapabilities = sync.OnceValue(func() []string {
+	caps := capability.ListKnown()
+	out := make([]string, 0, len(caps)+1)
+	out = append(out, "ALL") // magic value for "all capabilities"
+	for _, c := range caps {
+		out = append(out, "CAP_"+strings.ToUpper(c.String()))
+	}
+	return out
```
laurazard, line 41: Wonder if we can test this somehow, or at least extract the formatting code and write a tiny little unit test. But not a blocker.

thaJeztah:
@laurazard updated; I added a minimal unit-test. Perhaps wouldn't hurt to add tests for some of the other ones as well (but didn't want to stuff those in this PR).

For the "case-insensitive" auto-complete, I also want to do some experimenting in follow-ups. Related to that, I opened a PR to normalize and de-duplicate values in the client before we send the request;

- https://github.com/moby/moby/pull/48551

laurazard (approved):

laurazard:
Thanks! LGTM

PR #5484, opened 2024-09-26
docs/reference: stop, restart: add flag descriptions

thaJeztah:

### DIFF
--- a/docs/reference/commandline/container_restart.md
+++ b/docs/reference/commandline/container_restart.md
@@ -9,10 +9,10 @@ Restart one or more containers
 
 ### Options
 
-| Name             | Type     | Default | Description                                  |
-|:-----------------|:---------|:--------|:---------------------------------------------|
-| `-s`, `--signal` | `string` |         | Signal to send to the container              |
-| `-t`, `--time`   | `int`    | `0`     | Seconds to wait before killing the container |
+| Name                                   | Type     | Default | Description                                  |
+|:---------------------------------------|:---------|:--------|:---------------------------------------------|
+| [`-s`](#signal), [`--signal`](#signal) | `string` |         | Signal to send to the container              |
+| [`-t`](#time), [`--time`](#time)       | `int`    | `0`     | Seconds to wait before killing the container |
 
 
 <!---MARKER_GEN_END-->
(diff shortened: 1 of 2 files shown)


docs/reference/commandline/container_restart.md
```diff
@@ -22,3 +22,33 @@ Restart one or more containers
...
+### <a name="signal"></a> Stop container with signal (-s, --signal)
+
+The `--signal` flag sends the system call signal to the container to exit.
+This signal can be a signal name in the format `SIG<NAME>`, for instance
+`SIGKILL`, or an unsigned number that matches a position in the kernel's
+syscall table, for instance `9`.
+
+The default signal to use is defined by the image's [`StopSignal`](https://github.com/opencontainers/image-spec/blob/v1.1.0/config.md),
+which can be set through the [`STOPSIGNAL`](https://docs.docker.com/reference/dockerfile/#stopsignal)
+Dockerfile instruction when building the image, or configured using the the
```
Benehiko, line 36:
```suggestion
Dockerfile instruction when building the image, or configured using the
```

docs/reference/commandline/container_stop.md
```diff
@@ -22,10 +22,39 @@ Stop one or more running containers
...
+### <a name="signal"></a> Stop container with signal (-s, --signal)
+
+The `--signal` flag sends the system call signal to the container to exit.
+This signal can be a signal name in the format `SIG<NAME>`, for instance
+`SIGKILL`, or an unsigned number that matches a position in the kernel's
+syscall table, for instance `9`.
+
+The default signal to use is defined by the image's [`StopSignal`](https://github.com/opencontainers/image-spec/blob/v1.1.0/config.md),
+which can be set through the [`STOPSIGNAL`](https://docs.docker.com/reference/dockerfile/#stopsignal)
+Dockerfile instruction when building the image, or configured using the the
```
Benehiko, line 42:
```suggestion
Dockerfile instruction when building the image, or configured using the
```

docs/reference/commandline/container_restart.md
```diff
@@ -22,3 +22,33 @@ Restart one or more containers
...
 $ docker restart my_container
 ```
+
+
+### <a name="signal"></a> Stop container with signal (-s, --signal)
+
+The `--signal` flag sends the system call signal to the container to exit.
+This signal can be a signal name in the format `SIG<NAME>`, for instance
+`SIGKILL`, or an unsigned number that matches a position in the kernel's
+syscall table, for instance `9`.
```
dvdksn, line 32: Maybe call out the reference here?

```suggestion
syscall table, for instance `9`. See signal(7) for available signals.
```

docs/reference/commandline/container_stop.md
```diff
@@ -22,10 +22,39 @@ Stop one or more running containers
...
 ```console
 $ docker stop my_container
 ```
+
+### <a name="signal"></a> Stop container with signal (-s, --signal)
+
+The `--signal` flag sends the system call signal to the container to exit.
+This signal can be a signal name in the format `SIG<NAME>`, for instance
+`SIGKILL`, or an unsigned number that matches a position in the kernel's
+syscall table, for instance `9`.
```
dvdksn, line 38:
```suggestion
syscall table, for instance `9`. See signal(7) for available signals.
```

docs/reference/commandline/container_restart.md
thaJeztah: Ah, yes, good suggestion.

TBH, I was somewhat wondering if we should discourage uses of numbers (as looking up the number may be better handled on the environment where it's executed), but probably better to look at that separately.

docs/reference/commandline/container_restart.md
thaJeztah: Ah dang! I wish some of the linters would also run on markdown; `dupword` would've caught this one; https://github.com/docker/cli/blob/c767b9b2dacaa344e30426c2fcd65f7440392aa7/.golangci.yml#L6

docs/reference/commandline/container_restart.md
thaJeztah: Updated; I added a link, and used "Refer to" instead of "See" (ISTR we had a linter that didn't like "See ...", but maybe I'm mistaken 🙈

docs/reference/commandline/container_restart.md
```diff
@@ -22,3 +22,34 @@ Restart one or more containers
...
+If the container does not exit after the timeout elapses, it's forcibly killed
+with a `SIGKILL` signal.
+
+If you set `--time` to `-1`, no timeout is applied, and the daemon
+waits indefinitely for the container to exit.
+
+The default timeout can be specified using the [`--stop-timeout`](https://docs.docker.com/reference/cli/docker/container/run/#stop-timeout)
+option when creating the container. If no default is configured for the container,
+the Daemon determines the default, and is 10 seconds for Linux containers, and
+30 seconds for Windows containers.
```
thaJeztah, line 55: TBH; not super happy with these sections; we should probably have a look at better descriptions for these (we have this in some other places as well).

It's a bit complicated to explain these bits

- container inherits default from image (which can be set through dockerfile)
- or overridden at runtime (`docker create` / `docker run`)
- or overridden when invoking these commands
- or otherwise use the daemon default (which can differ per platform)


Some of that could become more transparent if we had better separation between "desired config" and "actual config" (in which case we could make defaults visible in `inspect` output), but we don't have that.

dvdksn (approved):
LGTM

Benehiko (approved):
LGTM

PR #5485, opened 2024-09-27
cli/command/container: stop, restart: rename "--time" to "--timeout"

thaJeztah:
This renames the `--time` flag as used on `docker stop` and `docker restart` to `--timeout`,  bringing it in line with other uses for this property, such as `--stop-timeout` on `docker run`.

The `--time` option is deprecated and hidden, but will be kept for backward compatibility, as these options existed for a long time.

```markdown changelog
The `--time` option on `docker stop` and `docker restart` is deprecated and renamed to `--timeout`.
```


thaJeztah:
@krissetto @laurazard I kept the flag descriptions unmodified for now, as that required more discussion; this one is only renaming the flag to `--timeout`.

cli/command/container/restart.go
```diff
@@ -30,8 +30,11 @@ func NewRestartCommand(dockerCli command.Cli) *cobra.Command {
 		Short: "Restart one or more containers",
 		Args:  cli.RequiresMinArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if cmd.Flags().Changed("time") && cmd.Flags().Changed("timeout") {
+				return errors.New("conflicting options: either specify --timeout or --time, not both")
```
krissetto, line 34: Thinking of these error messages a bit..instead of telling the users "multiple things" (the error: what to do, and what not to do), what do you think of something like this (the error: description)?

`conflicting options: --timeout and --time cannot be used together`

krissetto (approved):
LGTM, just left a thought on the conflicting options msg

albers (approved):
Completion LGTM

cli/command/container/restart.go
thaJeztah: Yes, agreed; wasn't a fan of the wording either. I recall I went looking for similar errors (because I recalled had some), but looks like I picked the odd one out;

```
cli/command/cli.go:	return errors.New("conflicting options: either specify --host or --context, not both")
cli/command/cli.go:		return nil, errors.New("conflicting options: either specify --host or --context, not both")
cli/command/container/opts.go:		return nil, errdefs.InvalidParameter(errors.New("conflicting options: cannot attach both user-defined and non-user-defined network-modes"))
cli/command/container/opts.go:		return errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --network-alias and per-network alias"))
cli/command/container/opts.go:		return errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --link and per-network links"))
cli/command/container/opts.go:		return errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --ip and per-network IPv4 address"))
cli/command/container/opts.go:		return errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --ip6 and per-network IPv6 address"))
cli/command/container/opts.go:		return errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --mac-address and per-network MAC address"))
cli/command/container/opts.go:		return errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --link-local-ip and per-network link-local IP addresses"))
cli/command/container/restart.go:				return errors.New("conflicting options: either specify --timeout or --time, not both")
cli/command/container/stop.go:				return errors.New("conflicting options: either specify --timeout or --time, not both")
cli/command/volume/create.go:					return errors.Errorf("conflicting options: either specify --name or provide positional arg, not both")
cli/command/volume/prune.go:				return 0, "", errdefs.InvalidParameter(errors.New("conflicting options: cannot specify both --all and --filter all=1"))
```


Most use `conflicting options: cannot specify both XXX and YYY`, which is the error I recalled.

thaJeztah:
@krissetto I updated the error PTAL if it still LGTY ☺️

krissetto:
> @krissetto I updated the error PTAL if it still LGTY ☺️

looks good to go 🚀

thaJeztah:
Thanks! I'll bring this in!

I may have a look at aligning those other errors in a follow-up, to bring a bit more consistency.

cli/command/container/restart.go
thaJeztah: - opened https://github.com/docker/cli/pull/5488 as a follow-up

PR #5488, opened 2024-10-01
align "conflicting options" errors for consistency

thaJeztah:
- follow-up to https://github.com/docker/cli/pull/5485#discussion_r1778797144

```markdown changelog
Updated some error messages for "conflicting options" to be more consistent
```


thaJeztah:
Oh! Looks like I broke something; I was testing some things, so perhaps I forgot to restore some code;

```
#20 60.26 === FAIL: cli/command/container TestParseRestartPolicyAutoRemove (0.00s)
#20 60.26     opts_test.go:823: Expected error Conflicting options: --restart and --rm, but got none
```

thaJeztah:
LOL, yeah, that test isn't great as it reports "but got none" which... is a LIE! https://github.com/docker/cli/blob/3907414549cd077dd71c30499575aa36f4db2e6e/cli/command/container/opts_test.go#L819-L825

cli/command/volume/create.go
```diff
@@ -52,7 +52,7 @@ func newCreateCommand(dockerCli command.Cli) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 1 {
 				if options.name != "" {
-					return errors.Errorf("conflicting options: either specify --name or provide positional arg, not both")
+					return errors.Errorf("conflicting options: cannot specify a volume-name through both --name and as a positional arg")
```
krissetto, line 55: is there a specific (i'd guess historical) reason this command has support for both an option and a positional arg for the name?

krissetto (approved):

cli/command/volume/create.go
thaJeztah: Yeah, there was some back-and-forth whether name should be a positional arg or a flag. The `--name` was changed to a positional arg, but because it already shipped was kept (but soft-deprecated / hidden).

thaJeztah: I think I may have been partially responsible for that, considering that `docker volume create foobar` is  more convenient than `docker volume create --name=foobar`, but later discussions were that, because name is _optional_, using a `--flag` is a more common convention.

PR #5491, opened 2024-10-02
introduce ParseKeyValueFile

ndeloof:
**- What I did**
introduced `ParseKeyValueFile` so docker/compose can use the same parser as docker/cli when env_file is requested to be loaded in `raw` format. This will be used so a user can get the same behavior using `docker run --env-file` and compose `env_file` while preserving the ability to get variable interpolated by default 😵‍💫

**- How I did it**

**- How to verify it**

image

### DIFF
--- a/opts/file.go
+++ b/opts/file.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 	"unicode"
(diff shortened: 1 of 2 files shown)


opts/envfile.go
```diff
@@ -18,5 +18,10 @@ import (
...
 // nothing more.
 func ParseEnvFile(filename string) ([]string, error) {
-	return parseKeyValueFile(filename, os.LookupEnv)
+	return ParseEnvFileWithLookup(filename, os.LookupEnv)
+}
+
+// ParseEnvFileWithLookup reads a file with environment variables enumerated by lines
+func ParseEnvFileWithLookup(filename string, env func(key string) (string, bool)) ([]string, error) {
+	return parseKeyValueFile(filename, env)
 }
```
thaJeztah, line 27: Some thoughts here;

- Perhaps `lookupFn` is a better name for this than `env`.
- Can you update the GoDoc to describe the new functionality (i.e., a `lookupFn` can be passed to provide values for the given key if it has no value set)
- Can you add a minimal unit-test for this (using a custom lookup function)?

Also wondering;

- For compose's use; is it better to pass a filename or should we have a function that takes an `io.Reader`? (Honestly, also looking at some of the existing unit-tests, where the requirement for having a file makes the tests more complicated than if it was a reader 😂)
- Effectively, this is now identical to `parseKeyValueFile`, which is a more agnostic name, and perhaps makes the overall functionality more "flexible" (if the format is to be used elsewhere).

Considering if we should move `parseKeyValueFile` to a be exported (accepting a reader) or even a separate package, so that it can be consumed without all of the other `opts` to be used (e.g. `keyvaluefile.Parse()`)

thaJeztah (commented):
Left some comments; do you need this change urgently? If not, I want to give my ideas a quick try to see what it looks like

opts/envfile.go
ndeloof: :+1 on using `io.Reader` (still need filename for error message). I tried to introduce the minimal required change here, but can make this change

ndeloof: updated with `ParseKeyValueFile` to accept a `Reader`. Didn't moved this out of the `opts` package as error message refers to "environment variable" so this still make it somehow dedicated to env file parsing

PR #5494, opened 2024-10-03
opts: cleanup ParseEnvFile tests

thaJeztah:
- relates to https://github.com/docker/cli/pull/5491

---

- Use gotest.tools for assertions
- Check for expected error messages
- Don't check for ErrBadKey errors, as it's not used as a sentinel error anywhere.
- Use t.SetEnv() instead of depending on `HOME` being set
- Use t.TempDir() for writing temporary files

### DIFF
--- a/opts/envfile_test.go
+++ b/opts/envfile_test.go
@@ -3,27 +3,20 @@ package opts
 import (
 	"bufio"
 	"os"
-	"reflect"
+	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


robmry (approved):

thaJeztah:
Thx! I'll bring this one in; I was planning to deprecate the `ErrBadKey` error, so thought I'd first cleanup some of this (as it was checking for it)

Issue #5479, opened 2024-09-25
Feature: docker rm --all --force Command


schuerg:
### Description

I’d like to suggest adding a command like `docker rm --all --force`. This would allow us to remove all containers, both running and stopped, with a single command. Currently, we have to stop them first and then remove them, which feels a bit clunky.

Podman has this functionality built in. Adding this option to Docker would streamline the process and improve the user experience.

### References

- https://docs.docker.com/reference/cli/docker/container/rm/
- https://docs.podman.io/en/latest/markdown/podman-rm.1.html

thaJeztah:
related to / (partial) duplicate of;

- https://github.com/docker/cli/issues/3622
- https://github.com/docker/cli/issues/3565
- https://github.com/docker/cli/issues/1679
- https://github.com/docker/cli/pull/3580

I think this one still needs a wider discussion, because these commands were primarily designed to interact on a single containers, or a list of containers provided as argument. Performing batch operations is somewhat conflicting with that design, which is why this was not implemented originally (besides the potential risk of a single flag being a bit of a "foot-gun" and a destructive operation).

laurazard:
I find myself doing `docker stop $(docker ps -q)` and `docker rm $(docker ps -aq)` probably to the point where I should've created a shell alias by now, so I do think we should look into. Nonetheless, it might make sense to see where we should implement it, as it could make sense to do it engine-side.

V0YD23:
@schuerg @thaJeztah  is this issue open to contribute ?

iTrooz:
+1, I'd love to submit a PR for this

> because these commands were primarily designed to interact on a single containers

How about `docker container rm --all -f` ? (adding the `--all` flag)
This is how podman implemented it

SzewczykW:
Is this issue open to contribute?

PR #5501, opened 2024-10-04
Update rm.go

smith518:


laurazard:
Hi @smith518, thanks for the contribution.

While I'd personally like to add some functionality like this to the CLI, I'm not sure we want to implement it at the CLI level (by having the CLI list all the containers, then make 1 call per container to the engine to remove them – I'd prefer we discussed making some changes to the daemon to allow to make a single call to remove multiple containers), but we haven't discussed that afaik.

Thoughts @thaJeztah?

thaJeztah:
Yeah, not sure about this; also because this overlaps with `docker container prune` (although that does, by design, not prune running containers).

I could see these kind of operations being useful for CI situations ("nuke" everything), but also a bit careful of making them _too_ up-front as part of commands, as in non-CI situations, these are pretty destructive. For CI situations, _most_ situations would already have a fair share of scripts and/or makefiles in  place (which could provide this functionality without much effort).

laurazard:
Yeah, right. For the CI situation/anything scripted I think it's straightforward to write something like `docker rm $(docker ps -aq)` too, although it would be nice to have some better alternative to that.

thaJeztah:
Let me close this one for now, as this requires more discussion; there's an existing tracking ticket for this, which also links to other, related discussions;

- https://github.com/docker/cli/issues/5479

PR #5502, opened 2024-10-04
move parsing key-value files to a separate package

thaJeztah:
- alternative to / closes https://github.com/docker/cli/pull/5491

Move the code for parsing key-value files, such as used for env-files and label-files to a separate package. This allows other projects (such as compose) to use the same parsing logic, but provide custom lookup functions for their situation (which is slightly different).

The new package provides utilities for parsing key-value files for either a file or an io.Reader. Most tests for EnvFile were now testing functionality that's already tested in the new package, so were (re)moved.

**- What I did**

Screenshot 2024-10-04 at 11 46 30

**- How I did it**

**- How to verify it**

```markdown changelog
move parsing key-value files to a separate package (pkg/kvfile)
```


laurazard (approved):
LGTM! This seems like a good place for this for now.

thaJeztah:
@ndeloof I decided to spend some time Yesterday to move that code after all; does this work for you (as alternative to  https://github.com/docker/cli/pull/5491)?

pkg/kvfile/kvfile.go
```diff
@@ -0,0 +1,130 @@
...
+// # Key/Value pairs
+//
+// Key/Value pairs take the following format:
+//
+//	KEY[=<VALUE>]
+//
+// KEY is required and may not contain whitespaces or NUL characters. Any
+// other character (except for the "=" delimiter) are accepted, but  it is
+// recommended to use a subset of the POSIX portable character set, as
+// outlined in [Environment Variables].
```
laurazard, line 33: Indeed, I remember having a discussion about this some time ago and while **some programs** will only accept a subset of the portable character set, there's no guarantees and often applications that people run _rely on_ other random characters, so we shouldn't do much validating here ourselves.

cli/command/container/opts_test.go
```diff
@@ -919,7 +919,7 @@ func TestParseEnvfileVariablesWithBOMUnicode(t *testing.T) {
 	}
 
 	// UTF16 with BOM
-	e := "contains invalid utf8 bytes at line"
+	e := "invalid utf8 bytes at line"
 	if _, _, _, err := parseRun([]string{"--env-file=testdata/utf16.env", "img", "cmd"}); err == nil || !strings.Contains(err.Error(), e) {
 		t.Fatalf("Expected an error with message '%s', got %v", e, err)
```
thaJeztah, line 924: FWIW; perhaps some of these tests should be moved to the new package as well, but didn't look at that yet.

pkg/kvfile/kvfile.go
thaJeztah: Yeah, compose recently ran into some of these. And, yes, Linux accepts "just about anything" (anything except `NUL` terminator). And, just because it _does_ doesn't mean you _should_ ... but some systems do, which is why we don't do validation on any of these.

laurazard: It might've been on Compose, yeah 😅

ndeloof (approved):

PR #5516, opened 2024-10-08
add shell-completion for --platform flags

thaJeztah:
### cli/command/completion: add Platforms

Add a utility for completing platform strings.

Platforms offers completion for platform-strings. It provides a non-exhaustive
list of platforms to be used for completion. Platform-strings are based on
[runtime.GOOS] and [runtime.GOARCH], but with (optional) variants added. A
list of recognised os/arch combinations from the Go runtime can be obtained
through "go tool dist list".

Some noteworthy exclusions from this list:

- arm64 images ("windows/arm64", "windows/arm64/v8") do not yet exist for windows.
- we don't (yet) include `os-variant` for completion (as can be used for Windows images)
- we don't (yet) include platforms for which we don't build binaries, such as
  BSD platforms (freebsd, netbsd, openbsd), android, macOS (darwin).
- we currently exclude architectures that may have unofficial builds,
  but don't have wide adoption (and no support), such as loong64, mipsXXX,
  ppc64 (non-le) to prevent confusion.

### cli/command/container: add shell completion for --platform flags

With this patch, completion is provided for `--platform` flags:

    docker run --platform<TAB>
    linux           linux/amd64     linux/arm/v5    linux/arm/v7    linux/arm64/v8  linux/riscv64   wasip1          windows
    linux/386       linux/arm       linux/arm/v6    linux/arm64     linux/ppc64le   linux/s390x     wasip1/wasm     windows/amd64

### cli/command/image: add shell completion for --platform flags

With this patch, completion is provided for `--platform` flags:

    docker pull --platform<TAB>
    linux           linux/amd64     linux/arm/v5    linux/arm/v7    linux/arm64/v8  linux/riscv64   wasip1          windows
    linux/386       linux/arm       linux/arm/v6    linux/arm64     linux/ppc64le   linux/s390x     wasip1/wasm     windows/amd64

Note that `docker buildx build` (with BuildKit) does not yet provide completion;
it's provided through buildx, and uses a different format (accepting multiple
comma-separated platforms). Interestingly, tab-completion for `docker build`
currently uses completion for non-buildkit, and has some other issues that may
have to be looked into.

**- How to verify it**

```markdown changelog
add shell-completion for --platform flags
```

### DIFF
--- a/cli/command/completion/functions.go
+++ b/cli/command/completion/functions.go
@@ -145,3 +145,47 @@ func FileNames(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCom
 func NoComplete(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
 	return nil, cobra.ShellCompDirectiveNoFileComp
 }
+
+var commonPlatforms = []string{
+	"linux",
+	"linux/386",
+	"linux/amd64",
+	"linux/arm",
+	"linux/arm/v5",
+	"linux/arm/v6",
+	"linux/arm/v7",
+	"linux/arm64",
+	"linux/arm64/v8",
+
+	// IBM power and z platforms
+	"linux/ppc64le",
+	"linux/s390x",
+
+	// Not yet supported
+	"linux/riscv64",
+
+	"windows",
+	"windows/amd64",
+
+	"wasip1",
+	"wasip1/wasm",
+}
+
+// Platforms offers completion for platform-strings. It provides a non-exhaustive
+// list of platforms to be used for completion. Platform-strings are based on
+// [runtime.GOOS] and [runtime.GOARCH], but with (optional) variants added. A
+// list of recognised os/arch combinations from the Go runtime can be obtained
+// through "go tool dist list".
+//
+// Some noteworthy exclusions from this list:
+//
+//   - arm64 images ("windows/arm64", "windows/arm64/v8") do not yet exist for windows.
+//   - we don't (yet) include `os-variant` for completion (as can be used for Windows images)
+//   - we don't (yet) include platforms for which we don't build binaries, such as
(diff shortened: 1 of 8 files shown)


rumpl (approved):

cli/command/completion/functions.go
```diff
@@ -145,3 +145,47 @@ func FileNames(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCom
...
+// list of platforms to be used for completion. Platform-strings are based on
+// [runtime.GOOS] and [runtime.GOARCH], but with (optional) variants added. A
+// list of recognised os/arch combinations from the Go runtime can be obtained
+// through "go tool dist list".
+//
+// Some noteworthy exclusions from this list:
+//
+//   - arm64 images ("windows/arm64", "windows/arm64/v8") do not yet exist for windows.
+//   - we don't (yet) include `os-variant` for completion (as can be used for Windows images)
+//   - we don't (yet) include platforms for which we don't build binaries, such as
```
krissetto, line 184: regarding these values, could they still be found in some image manifest? 

If we don't provide builds for these archs, i assume we don't support executing them either, is that correct? Just thinking if there are scenarios in which we could execute an image but the platform wouldn't show up because we don't support building them

thaJeztah: Technically, these could be "anything". i.e., I can push an image with platform `hello/world` and I would be able to pull it. The engine would also try to run it, but depending on what's in it, it may fail.

I mostly here tried to provide a list of "here's probably what you may be looking for" options; with the assumption that tab-completions are really "to assist" but the user can still pass any value they'd want that's not included in the list.

krissetto (approved):

thaJeztah:
Thanks! I'll bring this one in, and update https://github.com/docker/cli/pull/5331 to also provide completion for the new flags added there.

PR #5524, opened 2024-10-09
cli/compose: implement the ports validation method

Stavrospanakakis:
**- What I did**
This commit implements a validation method for the port mappings of compose. Also, it removes the ports validation method from the `expose` property since they do not always accept the same type of values. For example, `ports` property accepts the value `8000:8000` but the `expose` property not.

**- How I did it**
After reading the [API Reference](https://docs.docker.com/reference/compose-file/services/#ports) of Docker Compose syntax, I wrote a test that covers all the cases and then implemented the validation method until the tests pass.

**- How to verify it**
Run `make test`

```markdown changelog
Implement the ports validation method for compose
```

### DIFF
--- a/cli/compose/schema/schema.go
+++ b/cli/compose/schema/schema.go
@@ -6,9 +6,11 @@ package schema
 import (
 	"embed"
 	"fmt"
+	"math/big"
 	"strings"
 	"time"
 
+	"github.com/docker/go-connections/nat"
 	"github.com/pkg/errors"
 	"github.com/xeipuuv/gojsonschema"
 )
@@ -20,9 +22,18 @@ const (
 
 type portsFormatChecker struct{}
 
-func (checker portsFormatChecker) IsFormat(_ any) bool {
-	// TODO: implement this
-	return true
+func (checker portsFormatChecker) IsFormat(input any) bool {
+	var portSpec string
+
+	switch p := input.(type) {
+	case string:
+		portSpec = p
+	case *big.Rat:
+		portSpec = strings.Split(p.String(), "/")[0]
+	}
+
+	_, err := nat.ParsePortSpec(portSpec)
+	return err == nil
 }
 
 type durationFormatChecker struct{}
@@ -37,7 +48,6 @@ func (checker durationFormatChecker) IsFormat(input any) bool {
 }
 
 func init() {
-	gojsonschema.FormatCheckers.Add("expose", portsFormatChecker{})
 	gojsonschema.FormatCheckers.Add("ports", portsFormatChecker{})
 	gojsonschema.FormatCheckers.Add("duration", durationFormatChecker{})
 }
--- a/cli/compose/schema/schema_test.go
+++ b/cli/compose/schema/schema_test.go
@@ -29,6 +29,111 @@ func TestValidate(t *testing.T) {
 	assert.ErrorContains(t, Validate(config, "12345"), "unsupported Compose file version: 12345")
 }
 
+func TestValidatePorts(t *testing.T) {
+	testcases := []struct {
+		ports    any
+		hasError bool
+	}{
+		{
+			ports:    []int{8000},
+			hasError: false,
+		},
+		{
+			ports:    []string{"8000:8000"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"8001-8005"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"8001-8005:8001-8005"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"8000"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"8000-9000:80"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"[::1]:8080:8000"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"[::1]:8080-8085:8000"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"127.0.0.1:8000:8000"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"127.0.0.1:8000-8005:8000-8005"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"127.0.0.1:8000:8000/udp"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"8000:8000/udp"},
+			hasError: false,
+		},
+		{
+			ports:    []string{"8000:8000/http"},
+			hasError: true,
+		},
+		{
+			ports:    []string{"-1"},
+			hasError: true,
+		},
+		{
+			ports:    []string{"65536"},
+			hasError: true,
+		},
+		{
+			ports:    []string{"-1:65536/http"},
+			hasError: true,
+		},
+		{
+			ports:    []string{"invalid"},
+			hasError: true,
+		},
+		{
+			ports:    []string{"12345678:8000:8000/tcp"},
+			hasError: true,
+		},
+		{
+			ports:    []string{"8005-8000:8005-8000"},
+			hasError: true,
+		},
+		{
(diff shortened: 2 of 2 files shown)


thaJeztah:
at my computer right now (commenting from my phone), but wondering if (most of this?) would be doable using the utilities that are used for the corresponding command-line options;

https://github.com/docker/cli/blob/305985c1b11e8542abdb361205328e69f327caad/vendor/github.com/docker/go-connections/nat/parse.go#L10

https://github.com/docker/cli/blob/305985c1b11e8542abdb361205328e69f327caad/vendor/github.com/docker/go-connections/nat/nat.go#L121

https://github.com/docker/cli/blob/305985c1b11e8542abdb361205328e69f327caad/vendor/github.com/docker/go-connections/nat/nat.go#L171

laurazard (approved):
LGTM!

cli/compose/schema/schema.go
```diff
@@ -37,7 +48,6 @@ func (checker durationFormatChecker) IsFormat(input any) bool {
 }
 
 func init() {
-	gojsonschema.FormatCheckers.Add("expose", portsFormatChecker{})
```
thaJeztah, line 40: Not for this PR, but we probably would be able to use the same utilities for this case as well, which could be either using `ParsePortSpec` and making sure the IP address is not set, or a combination of `ParsePortRange` and  `SplitProtoPort `.


I see the latter is the approach taken for the `--expose` flag, but .. I'm actually wondering if the first approach would be better, because it also validates the given proto

https://github.com/docker/cli/blob/305985c1b11e8542abdb361205328e69f327caad/cli/command/container/opts.go#L443-L466

thaJeztah: On second thought, MAYBE validating the proto shouldn't belong on the client-side, as the list of supported protocols _could_ depend on what the daemon supports.

But given that we already do it elsewhere, maybe that's not _too_ problematic (and if we change, we should do so across the board).

We could possibly convert the code linked above to a utility if we don't want to duplicate the logic.

thaJeztah: In any case, none of the above is needed for the current PR; it's mostly outlining thoughts for a follow-up 😅

thaJeztah (approved):
LGTM, thanks!

cli/compose/schema/schema.go
Stavrospanakakis: @thaJeztah The [`ParsePortSpec`](https://github.com/docker/go-connections/blob/5df8d2b30ca886f2d94740ce3c54abd58a5bb2c9/nat/nat.go#L171) method returns [`PortMapping`](https://github.com/docker/go-connections/blob/5df8d2b30ca886f2d94740ce3c54abd58a5bb2c9/nat/nat.go#L148) while the `expose` propery is a [`PortBinding`](https://github.com/docker/go-connections/blob/5df8d2b30ca886f2d94740ce3c54abd58a5bb2c9/nat/nat.go#L12), so [`ParsePortSpec`](https://github.com/docker/go-connections/blob/5df8d2b30ca886f2d94740ce3c54abd58a5bb2c9/nat/nat.go#L171) might not be the best for this case.

For this reason, I believe that the combination of [`ParsePortRange`](https://github.com/docker/go-connections/blob/5df8d2b30ca886f2d94740ce3c54abd58a5bb2c9/nat/parse.go#L10) and [`SplitProtoPort`](https://github.com/docker/go-connections/blob/5df8d2b30ca886f2d94740ce3c54abd58a5bb2c9/nat/nat.go#L95) will work better .

Regarding creating a utility, I am not sure that it will help to avoid duplicate code. For example:

**Using a utility:**

```go
// utils.go

func validateExposePortOption(ctx context.Context, portSpec string) error {
  if strings.Contains(portSpec, ":") {
    return fmt.Errorf("invalid port format for expose %s", portSpec)
  }

  // support two formats for expose, original format <portnum>/[<proto>]
  // or <startport-endport>/[<proto>]
  proto, port := nat.SplitProtoPort(portSpec)
  
  // we could probably convert the validateProto method to public and validate the
  // proto to the client
  // https://github.com/docker/go-connections/blob/5df8d2b30ca886f2d94740ce3c54abd58a5bb2c9/nat/nat.go#L110
  // ......

  // parse the start and end port and create a sequence of ports to expose
  // if expose a port, the start and end port are the same
  _, _, err := nat.ParsePortRange(port) 
  if err != nil {
    return fmt.Errorf("invalid range format for expose: %s, error: %s", portSpec, err)
  }
  return nil
}
```

```go
// opts.go

// Merge in exposed ports to the map of published ports
for _, e := range copts.expose.GetAll() {
        err := validateExposePortOption(ctx, e)
        if err != nil {
            return nil, errors.Errorf("invalid port format for --expose: %s, error: %s", e, err)
        }

        // DUPLICATE CODE WITH THE UTILITY
	start, end, err := nat.ParsePortRange(port)
	if err != nil {
		return nil, errors.Errorf("invalid range format for --expose: %s, error: %s", e, err)
	}
	for i := start; i <= end; i++ {
		p, err := nat.NewPort(proto, strconv.FormatUint(i, 10))
		if err != nil {
			return nil, err
		}
		if _, exists := ports[p]; !exists {
			ports[p] = struct{}{}
		}
	}
}
```

```go
// schema.go

type exposeFormatChecker struct{}

func (checker exposeFormatChecker) IsFormat(input any) bool {
	var port string

	switch p := input.(type) {
	case string:
		port = p
	case *big.Rat:
		port = strings.Split(p.String(), "/")[0]
	}
    err := validateExposePortOption(context.Background(), port)
    return err == nil
}
```

**Not using a utility:**
```go
// opts.go

// Merge in exposed ports to the map of published ports
for _, e := range copts.expose.GetAll() {
	if strings.Contains(e, ":") {
		return nil, errors.Errorf("invalid port format for --expose: %s", e)
	}
	
	// support two formats for expose, original format <portnum>/[<proto>]
	// or <startport-endport>/[<proto>]
	proto, port := nat.SplitProtoPort(e)
	
    // we could probably convert the validateProto method to public and validate the proto to the client
    // https://github.com/docker/go-connections/blob/5df8d2b30ca886f2d94740ce3c54abd58a5bb2c9/nat/nat.go#L110
    // ......
    
	// parse the start and end port and create a sequence of ports to expose
	// if expose a port, the start and end port are the same
	start, end, err := nat.ParsePortRange(port)
	if err != nil {
		return nil, errors.Errorf("invalid range format for --expose: %s, error: %s", e, err)
	}
	for i := start; i <= end; i++ {
		p, err := nat.NewPort(proto, strconv.FormatUint(i, 10))
		if err != nil {
			return nil, err
		}
		if _, exists := ports[p]; !exists {
			ports[p] = struct{}{}
		}
	}
}
```

```go
// schema.go

type exposeFormatChecker struct{}

func (checker exposeFormatChecker) IsFormat(input any) bool {
    var port string
    
    switch p := input.(type) {
    case string:
    	port = p
    case *big.Rat:
    	port = strings.Split(p.String(), "/")[0]
    }

    if strings.Contains(portSpec, ":") {
        return false
    }
    
    // support two formats for expose, original format <portnum>/[<proto>]
    // or <startport-endport>/[<proto>]
    proto, port := nat.SplitProtoPort(portSpec)
    
    // we could probably convert the validateProto method to public and validate the
    // proto to the client
    // https://github.com/docker/go-connections/blob/5df8d2b30ca886f2d94740ce3c54abd58a5bb2c9/nat/nat.go#L110
    // ......
    
    // parse the start and end port and create a sequence of ports to expose
    // if expose a port, the start and end port are the same
    _, _, err := nat.ParsePortRange(port) 
    if err != nil {
        return false
    }
    
    return true
}
```

Both cases with or without a utility contain duplicate code. 

What do you think?

I will be happy to open a new pull request for this.

PR #5527, opened 2024-10-10
Improve completion of containers for `docker rm`

albers:
**- What I did**
The cobra completion for `docker rm` completes all containers. This does not make much sense as not all containers are removable unless `--force` is given.
This PR aligns the completion to the legacy completion, where only removable containers are completed unless `--force` is given, see [here](https://github.com/docker/cli/blob/88f1e99e8e00825ec025e242b77175e8f7de4980/contrib/completion/bash/docker#L1862-L1879).

**- How I did it**
Added a filter to the call of `completion.ContainerNames` that filters on container status and the value of `--force`.

**- How to verify it**

```markdown changelog
Improve completion of containers for `docker rm`
```

@thaJeztah PTAL

### DIFF
--- a/cli/command/container/rm.go
+++ b/cli/command/container/rm.go
@@ -38,7 +38,9 @@ func NewRmCommand(dockerCli command.Cli) *cobra.Command {
 		Annotations: map[string]string{
 			"aliases": "docker container rm, docker container remove, docker rm",
 		},
-		ValidArgsFunction: completion.ContainerNames(dockerCli, true),
+		ValidArgsFunction: completion.ContainerNames(dockerCli, true, func(ctr container.Summary) bool {
+			return opts.force || ctr.State == "exited" || ctr.State == "created"
+		}),
 	}
 
 	flags := cmd.Flags()


albers:
I think the test failures are unrelated: Looks like a rate limit violation and a connection timeout.

cli/command/container/rm.go
```diff
@@ -38,7 +38,9 @@ func NewRmCommand(dockerCli command.Cli) *cobra.Command {
 		Annotations: map[string]string{
 			"aliases": "docker container rm, docker container remove, docker rm",
 		},
-		ValidArgsFunction: completion.ContainerNames(dockerCli, true),
+		ValidArgsFunction: completion.ContainerNames(dockerCli, true, func(ctr container.Summary) bool {
+			return opts.force || ctr.State == "exited" || ctr.State == "created"
```
laurazard, line 42: This is totally fine for now, but I wish we had an enum (or whatever passes for one in Go) for the `State`.

laurazard (approved):
LGTM, thanks!

cli/command/container/rm.go
thaJeztah: Ah, yes! I think I may have started working on a branch at some point, but need to dig it up. It's all a bit awkward as these parts are rather quirky, and these values are currently set in the `container` package, not in anything inside the `api/xxx` or `client/xxx`; https://github.com/moby/moby/blob/61030f0e87c8d893082599815897f10d440f9bda/container/state.go#L119-L158


```go
// StateString returns a single string to describe state
func (s *State) StateString() string {
	if s.Running {
		if s.Paused {
			return "paused"
		}
		if s.Restarting {
			return "restarting"
		}
		return "running"
	}

	if s.RemovalInProgress {
		return "removing"
	}

	if s.Dead {
		return "dead"
	}

	if s.StartedAt.IsZero() {
		return "created"
	}

	return "exited"
}

// IsValidStateString checks if the provided string is a valid container state or not.
func IsValidStateString(s string) bool {
	if s != "paused" &&
		s != "restarting" &&
		s != "removing" &&
		s != "running" &&
		s != "dead" &&
		s != "created" &&
		s != "exited" {
		return false
	}
	return true
}
```

thaJeztah: I recall now that I also wanted to look at these completion helpers to allow passing filters so that the daemon would only return containers in specific state(s) to get the equivalent of;


```bash
docker ps --format='{{.Names}} {{.Status}}' --filter status=exited --filter status=created
container-1 created
container-2 exited
container-3 exited
```

(and perhaps add a `!=`  option for status to allow `--filter status!=running`)

thaJeztah (approved):
LGTM, nice!!

PR #5528, opened 2024-10-11
feat: `run` and `pull` image completion from hub

Benehiko:
**- What I did**
Completion of images when calling `docker run/pull` from local images and Docker Hub. 

Allows for image search based on partial matches. Also allows image tag search when adding `:` after the image name.

image
image
image
image

**- How I did it**

**- How to verify it**
`docker completion <fish|bash|zsh> > <completion file>`

`docker run [tab][tab]`
`docker run postg[tab][tab]`
`docker run postgres:[tab][tab]`
`docker run postgres:alpine[tab][tab]`

```markdown changelog


```

image


cli/command/completion/functions.go
```diff
@@ -189,3 +194,122 @@ var commonPlatforms = []string{
...
+		}
+		q := u.Query()
+		q.Set("ordering", "last_updated")
+		q.Set("page_size", "25")
+		q.Set("name", imageTag)
+		u.RawQuery = q.Encode()
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+		if err != nil {
+			logrus.Errorf("Error creating hub image tags request: %v", err)
```
Benehiko, line 254: I honestly don't know what the best way would be for us to track errors for the completions. Maybe we should write debug logs to a file?

PR #5533, opened 2024-10-13
cli/command/completion: add more unit-tests

thaJeztah:

### DIFF
--- a/cli/command/completion/functions_test.go
+++ b/cli/command/completion/functions_test.go
@@ -29,6 +29,14 @@ func TestCompleteFileNames(t *testing.T) {
 	assert.Check(t, is.Len(values, 0))
 }
 
+func TestCompleteFromList(t *testing.T) {
+	expected := []string{"one", "two", "three"}
+
+	values, directives := FromList(expected...)(nil, nil, "")
+	assert.Check(t, is.Equal(directives&cobra.ShellCompDirectiveNoFileComp, cobra.ShellCompDirectiveNoFileComp), "Should not perform file completion")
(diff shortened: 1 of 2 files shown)


thaJeztah:
Had some of these stashed while I was working on some earlier changes; looks like I never pushed them

cli/command/completion/functions_test.go
```diff
@@ -1,15 +1,350 @@
...
+			expDirective: cobra.ShellCompDirectiveNoFileComp,
+		},
+		{
+			doc: "with results",
+			volumes: []*volume.Volume{
+				{Name: "volume-c"},
+				{Name: "volume-b"},
+				{Name: "volume-a"},
+			},
+			expOut:       []string{"volume-c", "volume-b", "volume-a"},
```
laurazard, line 324: This is making me thing about whether we should order the output. I image it'd be nicer to have it alphanumerically ordered.

laurazard:
Something else that this made me check/think about is that it'd be nice if the completion for commands that can take multiple arguments (such as `docker rm [container] [container]` didn't include arguments already added. I'll write a ticket 😅

laurazard (approved):
LGTM

cli/command/completion/functions_test.go
thaJeztah: Ah, yes, correct. I recalled that I (in one of my PRs) added sorting, but then noticed "it doesn't make a difference!". So somewhat purposefully made these tests show "it's not sorted, and that's ok".

We _could_ check if there's a performance difference in "pre sorting"; not sure if the shell completion code is slower at this?

PR #5538, opened 2024-10-17
Completion for `events --filter`

albers:
Adds completion for `docker events --filter`.

There are different types of filters, requiring different completion logic.
Some filters have static values, others require API lookups.
In some cases, trailing spaces of intermediate completions have to be suppressed.

The completion logic is complex because cobra does not offer support for compound flag values.
The handler for the `--filter` flag serves as the central entrypoint to all completions.

How to verify:
in a dev container, run `make completion`, then trigger completions, e.g.
```shell
$ docker events --filter <TAB><TAB>
container=  event=      label=      node=       type=       
daemon=     image=      network=    scope=      volume= 
$ docker events --filter event=c<TAB><TAB>
checkpoint  commit      connect     copy        create
$ docker events --filter image=<TAB><TAB>
docker-cli-dev:latest  docker-cli-e2e:latest  registry:2
```

The completions should correspond to those of the legacy bash completion.

### DIFF
--- a/cli/command/system/completion.go
+++ b/cli/command/system/completion.go
@@ -1,0 +1,237 @@
+package system
+
+import (
+	"strings"
+
+	"github.com/docker/cli/cli/command/completion"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/events"
+	"github.com/docker/docker/api/types/image"
+	"github.com/docker/docker/api/types/network"
+	"github.com/docker/docker/api/types/volume"
+	"github.com/spf13/cobra"
+)
+
+var (
+	eventFilters = []string{"container", "daemon", "event", "image", "label", "network", "node", "scope", "type", "volume"}
+
+	// eventTypes is a list of all event types.
+	// This should be moved to the moby codebase once its usage is consolidated here.
+	eventTypes = []events.Type{
+		events.BuilderEventType,
+		events.ConfigEventType,
+		events.ContainerEventType,
+		events.DaemonEventType,
+		events.ImageEventType,
+		events.NetworkEventType,
+		events.NodeEventType,
+		events.PluginEventType,
+		events.SecretEventType,
+		events.ServiceEventType,
+		events.VolumeEventType,
+	}
+
+	// eventActions is a list of all event actions.
+	// This should be moved to the moby codebase once its usage is consolidated here.
+	eventActions = []events.Action{
+		events.ActionCreate,
+		events.ActionStart,
+		events.ActionRestart,
+		events.ActionStop,
+		events.ActionCheckpoint,
+		events.ActionPause,
+		events.ActionUnPause,
+		events.ActionAttach,
+		events.ActionDetach,
+		events.ActionResize,
+		events.ActionUpdate,
+		events.ActionRename,
+		events.ActionKill,
+		events.ActionDie,
+		events.ActionOOM,
+		events.ActionDestroy,
+		events.ActionRemove,
+		events.ActionCommit,
+		events.ActionTop,
+		events.ActionCopy,
+		events.ActionArchivePath,
+		events.ActionExtractToDir,
+		events.ActionExport,
+		events.ActionImport,
+		events.ActionSave,
+		events.ActionLoad,
+		events.ActionTag,
+		events.ActionUnTag,
+		events.ActionPush,
+		events.ActionPull,
+		events.ActionPrune,
+		events.ActionDelete,
+		events.ActionEnable,
+		events.ActionDisable,
+		events.ActionConnect,
+		events.ActionDisconnect,
+		events.ActionReload,
+		events.ActionMount,
+		events.ActionUnmount,
+		events.ActionExecCreate,
+		events.ActionExecStart,
+		events.ActionExecDie,
+		events.ActionExecDetach,
+		events.ActionHealthStatus,
+		events.ActionHealthStatusRunning,
+		events.ActionHealthStatusHealthy,
+		events.ActionHealthStatusUnhealthy,
+	}
+)
+
+// completeEventFilters provides completion for the filters that can be used with `--filter`.
+func completeEventFilters(dockerCLI completion.APIClientProvider) completion.ValidArgsFn {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		key, _, ok := strings.Cut(toComplete, "=")
+		if !ok {
+			return postfixWith("=", eventFilters), cobra.ShellCompDirectiveNoSpace
+		}
+		switch key {
+		case "container":
+			return prefixWith("container=", containerNames(dockerCLI, cmd, args, toComplete)), cobra.ShellCompDirectiveNoFileComp
+		case "daemon":
+			return prefixWith("daemon=", daemonNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
+		case "event":
+			return prefixWith("event=", validEventNames()), cobra.ShellCompDirectiveNoFileComp
+		case "image":
+			return prefixWith("image=", imageNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
+		case "label":
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		case "network":
+			return prefixWith("network=", networkNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
+		case "node":
+			return prefixWith("node=", nodeNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
+		case "scope":
+			return prefixWith("scope=", []string{"local", "swarm"}), cobra.ShellCompDirectiveNoFileComp
+		case "type":
+			return prefixWith("type=", eventTypeNames()), cobra.ShellCompDirectiveNoFileComp
+		case "volume":
+			return prefixWith("volume=", volumeNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
+		default:
+			return postfixWith("=", eventFilters), cobra.ShellCompDirectiveNoSpace | cobra.ShellCompDirectiveNoFileComp
+		}
+	}
+}
+
+// prefixWith prefixes every element in the slice with the given prefix.
+func prefixWith(prefix string, values []string) []string {
+	result := make([]string, len(values))
+	for i, v := range values {
+		result[i] = prefix + v
+	}
+	return result
+}
+
+// postfixWith appends postfix to every element in the slice.
+func postfixWith(postfix string, values []string) []string {
+	result := make([]string, len(values))
+	for i, v := range values {
+		result[i] = v + postfix
+	}
+	return result
+}
+
+// eventTypeNames provides a list of all event types.
+// The list is derived from eventTypes.
+func eventTypeNames() []string {
+	names := make([]string, len(eventTypes))
+	for i, eventType := range eventTypes {
+		names[i] = string(eventType)
+	}
+	return names
+}
+
+// validEventNames provides a list of all event actions.
+// The list is derived from eventActions.
+// Actions that are not suitable for usage in completions are removed.
+func validEventNames() []string {
+	names := make([]string, 0, len(eventActions))
+	for _, eventAction := range eventActions {
+		if strings.Contains(string(eventAction), " ") {
+			continue
+		}
+		names = append(names, string(eventAction))
+	}
+	return names
+}
+
+// containerNames contacts the API to get names and optionally IDs of containers.
+// In case of an error, an empty list is returned.
+func containerNames(dockerCLI completion.APIClientProvider, cmd *cobra.Command, args []string, toComplete string) []string {
+	names, _ := completion.ContainerNames(dockerCLI, true)(cmd, args, toComplete)
+	if names == nil {
+		return []string{}
+	}
+	return names
+}
+
+// daemonNames contacts the API to get name and ID of the current docker daemon.
+// In case of an error, an empty list is returned.
+func daemonNames(dockerCLI completion.APIClientProvider, cmd *cobra.Command) []string {
+	info, err := dockerCLI.Client().Info(cmd.Context())
+	if err != nil {
+		return []string{}
+	}
+	return []string{info.Name, info.ID}
+}
+
+// imageNames contacts the API to get a list of image names.
+// In case of an error, an empty list is returned.
+func imageNames(dockerCLI completion.APIClientProvider, cmd *cobra.Command) []string {
+	list, err := dockerCLI.Client().ImageList(cmd.Context(), image.ListOptions{})
+	if err != nil {
+		return []string{}
+	}
+	names := make([]string, 0, len(list))
+	for _, img := range list {
+		names = append(names, img.RepoTags...)
+	}
+	return names
+}
+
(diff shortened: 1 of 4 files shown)


cli/command/system/completion.go
```diff
@@ -0,0 +1,99 @@
...
+)
+var eventTypes = []string{"config", "container", "daemon", "image", "network", "node", "plugin", "secret", "service", "volume"}
+
+func completeFilters(dockerCLI completion.APIClientProvider) completion.ValidArgsFn {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if strings.HasPrefix(toComplete, "container=") {
+			// the pure container name list should be pulled out from ContainerNames.
+			names, _ := completion.ContainerNames(dockerCLI, true)(cmd, args, toComplete)
+			return prefixWith("container=", names), cobra.ShellCompDirectiveDefault
+		}
```
thaJeztah, line 64: Nice!! I was wondering the other day if something like this would work. I wanted to experiment if completion would be possible for the "advanced" CSV flags, but didn't give it a try yet.

cli/command/system/completion.go
```diff
@@ -0,0 +1,99 @@
...
+		if strings.HasPrefix(toComplete, "event=") {
+			return prefixWith("event=", eventNames), cobra.ShellCompDirectiveDefault
+		}
+		if strings.HasPrefix(toComplete, "label=") {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		if strings.HasPrefix(toComplete, "network=") {
+			names, err := completion.NetworkNames(dockerCLI, cmd)
```
thaJeztah, line 72: I'm slightly on the fence on the `completion` package now having both `completion.NetworkNames`  and `completion.CompleteNetworks`.  Or at least, I"m looking if we can reduce exported functions, unless needed.

Looking at the code in `NetworkNames`, it's just a few lines, and fairly straightforward. I'd be fine with copying that code here as a un-exported function (for now). It may be good to include a reference to the "original" one to correlate them, e.g.;


```go
// networkNames contacts the API to get a list of networks.
//
// This function is based on [completion.NetworkNames].
func NetworkNames(dockerCLI APIClientProvider, cmd *cobra.Command) ([]string, error) {
	summaries, err := dockerCLI.Client().NetworkList(cmd.Context(), network.ListOptions{})
	if err != nil {
		return nil, err
	}
	var names []string
	for _, nws := range summaries {
		names = append(names, nws.Name)
	}
	return names, nil
}
```


That would still allow us to change our mind in future of course!

cli/command/system/completion.go
```diff
@@ -0,0 +1,99 @@
...
+			if err != nil {
+				return nil, cobra.ShellCompDirectiveError
+			}
```
thaJeztah, line 75: Looking at this, I got curious, and you probably know; would there be a difference here between returning an empty result (`[]string{}`) but no error, or `nil` and an error?

Also wondering if (as performance may be relevant for completion) if a short timeout (for the API call) with a fallback (list of pre-defined networks) would make sense for such cases.

cli/command/system/completion.go
thaJeztah: > That would still allow us to change our mind in future of course!

Expanding a bit on that (just to echo my train of thought ❤️); with cli-plugins, like compose, buildx, and others, we may need to think what to put "where" at some point. We didn't discuss that in-depth when we added the initial implementation of cobra completion, but sometimes it can be "complicated" to have the code live together with other code.

So, possibly at some point we may need to re-organise some of the functions to be exported (allowing them to be  consumed by compose and buildx, etc), but without that meaning they come with a large number of dependencies.

thaJeztah (commented):
Thank you for working on this! It's really great to see you start doing so (and much appreciated!). I initially had some reservations on the Cobra completion (and the initial set of completion was fairly limited), but they start to shape up.

I like where this PR is going; I gave it a quick try, and it's already pretty nice;

```bash
root@docker-cli-dev$ docker events --filter
container=  daemon=     event=      image=      label=      network=    node=       scope=      type=       volume=
root@docker-cli-dev$ docker events --filter container=
empty            foo              hello2           kind_wing        loving_mccarthy
root@docker-cli-dev$ docker events --filter container=foo --filter network=
bridge           docker_gwbridge  host             ingress          none
root@docker-cli-dev$ docker events --filter container=foo --filter network=
```

Some things that I didn't give much thought yet;

- (honestly) our filter flags are a bit messy and behavior not always well-defined (how do combinations of filters .. combine?)
- Also some quirks like some filters allowing "exclusions" (`foo!=bar`) and others don't, and this may also differ between commands (:disappointed:)
- :point_up: we need to spend time documenting all filters and options

Also "FYI" we were discussion potentially adding (an) API endpoint(s) dedicated to completion; nothing set in stone yet, and still to be discussed further, but the idea was that such an endpoint could move some of the heavy-lifting to the daemon, and provide a more optimized response for purpose of completion (e.g. if all we need is "names", there's no need to collect _all the other bits_). Such endpoint(s) could be more lightweight and perhaps more flexible / opinionated as their purpose is clear.

That discussion was somewhat related to https://github.com/docker/cli/pull/5528, which is a really nice feature, but also had some implications; it would mean the CLI now had to integrate with a Docker Hub specific API (the endpoints used are not part of the OCI distribution spec), but also could complicate situations where the daemon is configured to use a proxy, is airgapped, or is using a registry mirror. Having an (extensible) endpoint defined in the API could mean that such actions could be handled on the daemon-side (which on (e.g.) Docker Desktop could mean a service handling this).

cli/command/system/completion.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"strings"
+
+	"github.com/docker/cli/cli/command/completion"
+	"github.com/spf13/cobra"
+)
+
+var (
+	eventFilters = []string{"container", "daemon", "event", "image", "label", "network", "node", "scope", "type", "volume"}
+	eventNames   = []string{
```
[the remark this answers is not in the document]
albers: This would be added in the moby codebase?

cli/command/system/completion.go
albers: > I'm slightly on the fence on the completion package now having both completion.NetworkNames and completion.CompleteNetworks.

I agree this feels a bit overkill for the desired effect.

> Looking at the code in `NetworkNames`, it's just a few lines, and fairly straightforward. I'd be fine with copying that code here as a un-exported function (for now).

For `NetworkNames` this would be a good solution. But for `ContainerNames`, this would mean a lot of duplicated business logic.

What about keeping the
```go
    names, _ := completion.NetworkNames(dockerCLI)(cmd, args, toComplete)
```
and
```go
    names, _ := completion.ContainerNames(dockerCLI, true)(cmd, args, toComplete)
```
hacks instead?

cli/command/system/completion.go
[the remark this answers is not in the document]
thaJeztah: Yes, correct; we vendor those types from the Moby repo.

Converting to a string (or slice of strings) would then probably be left to the consumer (so that's something we can do in this repository).

cli/command/system/completion.go
albers: > would there be a difference here between returning an empty result ([]string{}) but no error, or nil and an error?

Returning `[]string{}` works but in this context requires `ShellCompDirectiveNoFileComp`.

cli/command/system/completion.go
thaJeztah: > What about keeping the hacky version from before the refactoring:

Oh, I think that's definitely acceptable for now. Once we figure out alternatives, we can easily swap out those lines.

albers: I reverted to the previous version and added error handling for network and container completion to avoid filenames being completed in case of an error in the api calls.

cli/command/system/completion.go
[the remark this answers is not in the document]
albers: Great. I'll make a suggestion. This might take some time, though.
Thanks for your support.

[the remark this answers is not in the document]
albers: @thaJeztah I added completions for the remaining filters.
At that stage, the `completeFilters` was way to long and ugly, so I decided to introduce some helper functions.
These helper functions simplify error handling very much, as they just return empty arrays in case of API errors.
PTAL, IMHO the PR should be complete now.

albers:
@thaJeztah I just stepped over your [completion tests](5542). I'll take a look at it and try to add tests for the completions in this PR.

albers:
@thaJeztah I added tests for those commands that call the API.
Please take a look before I clean up the commits and publish the PR.

albers:
The PR is ready for review now.

vvoland (approved):
LGTM, thanks!

cli/command/system/completion.go
```diff
@@ -0,0 +1,241 @@
...
+		if strings.HasPrefix(toComplete, "type=") {
+			return prefixWith("type=", eventTypeNames()), cobra.ShellCompDirectiveNoFileComp
+		}
+		if strings.HasPrefix(toComplete, "volume=") {
+			return prefixWith("volume=", volumeNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
+		}
+
+		return postfixWith("=", eventFilters), cobra.ShellCompDirectiveNoSpace
+	}
+}
```
thaJeztah, line 123: We could probably optimize this to prevent matching prefix multiple times;

```go
// completeEventFilters provides completion for the filters that can be used with `--filter`.
func completeEventFilters(dockerCLI completion.APIClientProvider) completion.ValidArgsFn {
	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		key, _, ok := strings.Cut(toComplete, "=")
		if !ok {
			return postfixWith("=", eventFilters), cobra.ShellCompDirectiveNoSpace
		}
		switch key {
		case "container":
			return prefixWith("container=", containerNames(dockerCLI, cmd, args, toComplete)), cobra.ShellCompDirectiveNoFileComp
		case "daemon":
			return prefixWith("daemon=", daemonNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
		case "event":
			return prefixWith("event=", validEventNames()), cobra.ShellCompDirectiveNoFileComp
		case "image":
			return prefixWith("image=", imageNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
		case "label":
			return nil, cobra.ShellCompDirectiveNoFileComp
		case "network":
			return prefixWith("network=", networkNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
		case "node":
			return prefixWith("node=", nodeNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
		case "scope":
			return prefixWith("scope=", []string{"local", "swarm"}), cobra.ShellCompDirectiveNoFileComp
		case "type":
			return prefixWith("type=", eventTypeNames()), cobra.ShellCompDirectiveNoFileComp
		case "volume":
			return prefixWith("volume=", volumeNames(dockerCLI, cmd)), cobra.ShellCompDirectiveNoFileComp
		default:
			return postfixWith("=", eventFilters), cobra.ShellCompDirectiveNoSpace
		}
	}
}
```

I was even contemplating if we should re-purpose the constants where suitable, but perhaps confusing, because not all filters refer to a type; i.e., was considering;

```go
case string(events.ContainerEventType):
```

But not sure if that's a good thing to do

cli/command/system/completion.go
```diff
@@ -0,0 +1,241 @@
...
+		names[i] = string(eventType)
+	}
+	return names
+}
+
+// validEventNames provides a list of all event actions.
+// The list is derived from eventActions.
+// Actions that are not suitable for usage in completions are removed.
+func validEventNames() []string {
+	names := []string{}
```
thaJeztah, line 157: Nit: this could pre-allocate;

```suggestion
	names := make([]string, 0, len(eventActions))
```

cli/command/system/completion.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// imageNames contacts the API to get a list of image names.
+// In case of an error, an empty list is returned.
+func imageNames(dockerCLI completion.APIClientProvider, cmd *cobra.Command) []string {
+	list, err := dockerCLI.Client().ImageList(cmd.Context(), image.ListOptions{})
+	if err != nil {
+		return []string{}
+	}
+	names := []string{}
```
thaJeztah, line 194: Same here (it will still have to grow if the RepoTags has multiple entries, but small bits);

```suggestion
	names := make([]string, 0, len(list))
```

cli/command/system/completion.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// networkNames contacts the API to get a list of network names.
+// In case of an error, an empty list is returned.
+func networkNames(dockerCLI completion.APIClientProvider, cmd *cobra.Command) []string {
+	list, err := dockerCLI.Client().NetworkList(cmd.Context(), network.ListOptions{})
+	if err != nil {
+		return []string{}
+	}
+	names := []string{}
```
thaJeztah, line 208:
```suggestion
	names := make([]string, 0, len(list))
```

cli/command/system/completion.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// nodeNames contacts the API to get a list of node names.
+// In case of an error, an empty list is returned.
+func nodeNames(dockerCLI completion.APIClientProvider, cmd *cobra.Command) []string {
+	list, err := dockerCLI.Client().NodeList(cmd.Context(), types.NodeListOptions{})
+	if err != nil {
+		return []string{}
+	}
+	names := []string{}
```
thaJeztah, line 222:
```suggestion
	names := make([]string, 0, len(list))
```

cli/command/system/completion.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// volumeNames contacts the API to get a list of volume names.
+// In case of an error, an empty list is returned.
+func volumeNames(dockerCLI completion.APIClientProvider, cmd *cobra.Command) []string {
+	list, err := dockerCLI.Client().VolumeList(cmd.Context(), volume.ListOptions{})
+	if err != nil {
+		return []string{}
+	}
+	names := []string{}
```
thaJeztah, line 236:
```suggestion
	names := make([]string, 0, len(list.Volumes))
```

thaJeztah (commented):
This looks great! Sorry for the delay; I apparently started a review, and didn't submit 🙈 

I left some suggestions for consideration.

cli/command/system/completion.go
albers: That's much better, thank you. I did not go for the constants because I think this produces asymmetry that makes the code harder to comprehend.

albers:
@thaJeztah Thanks for the review, I applied all of your suggestions. PTAL.

cli/command/system/completion.go
thaJeztah: Yes, makes sense; I started to replace the strings with the consts, then realised not all of them were event-types 😂

thaJeztah:
> @thaJeztah Thanks for the review, I applied all of your suggestions. PTAL.

Thank you! Changes LGTM, but it's ok to squash my suggestions with the first commit (we don't need the history of the "review comments applied"). Could you squash the first and last commit?

albers:
> Could you squash the first and last commit?

done.

thaJeztah (approved):
LGTM, thank you!!

PR #5541, opened 2024-10-17
templates: add test for HeaderFunctions

thaJeztah:
Before:

    go test -test.coverprofile -
    PASS
    coverage: 65.2% of statements
    ok  	github.com/docker/cli/templates	0.607s

After:

    go test -test.coverprofile -
    PASS
    coverage: 95.7% of statements
    ok  	github.com/docker/cli/templates	0.259s

### DIFF
--- a/templates/templates_test.go
+++ b/templates/templates_test.go
@@ -89,3 +89,55 @@ func TestParseTruncateFunction(t *testing.T) {
 		})
 	}
 }
+
+func TestHeaderFunctions(t *testing.T) {
+	const source = "hello world"
+
+	tests := []struct {
+		doc      string
+		template string
+	}{
+		{
+			doc:      "json",
+			template: `{{ json .}}`,
+		},
+		{
+			doc:      "split",
+			template: `{{ split . ","}}`,
+		},
+		{
+			doc:      "join",
+			template: `{{ join . ","}}`,
+		},
+		{
+			doc:      "title",
+			template: `{{ title .}}`,
+		},
+		{
+			doc:      "lower",
+			template: `{{ lower .}}`,
+		},
+		{
+			doc:      "upper",
+			template: `{{ upper .}}`,
+		},
+		{
+			doc:      "truncate",
+			template: `{{ truncate . 2}}`,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.doc, func(t *testing.T) {
+			tmpl, err := New("").Funcs(HeaderFunctions).Parse(tc.template)
+			assert.NilError(t, err)
+
+			var b bytes.Buffer
+			assert.NilError(t, tmpl.Execute(&b, source))
+
+			// All header-functions are currently stubs, and don't modify the input.
+			expected := source
+			assert.Equal(t, expected, b.String())
+		})
+	}
+}


templates/templates_test.go
```diff
@@ -89,3 +89,56 @@ func TestParseTruncateFunction(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.doc, func(t *testing.T) {
+			tmpl, err := New("").Funcs(HeaderFunctions).Parse(tc.template)
+			assert.NilError(t, err)
+
+			var b bytes.Buffer
+			assert.NilError(t, tmpl.Execute(&b, source))
+
+			// All header-functions are currently stubs, and don't modify the input.
+			expected := source
```
Benehiko, line 140: what's the purpose of  `expected` here?

templates/templates_test.go
```diff
@@ -89,3 +89,56 @@ func TestParseTruncateFunction(t *testing.T) {
...
+			assert.Equal(t, expected, b.String())
```
Benehiko, line 141: how can this test pass if expected isn't defined inside our test table? We are just comparing the source text to whatever the template produced, which can't be the same in all instances, right?

templates/templates_test.go
thaJeztah: Mostly to make the output on failures less ambiguous; gotest.tools does not define an order in which "actual" and "expected" values must be passed (I _think_ this may have been on purpose, as there's been some bike-shedding about those, but I should really check with the maintainer if they would be open to have a recommendation added).


The `expected` intermediate variable is printed as part of the output, which helps reading "so, what was expected?" 😂 


Here's a quick example; it's most notably with the `DeepEqual` (showing a diff) where things are sometimes ambiguous ("which one is right?");

```go
func TestFoo(t *testing.T) {
	assert.Check(t, is.Equal("one", "two"))

	actual := "two"
	expected := "one"
	assert.Check(t, is.Equal(actual, expected))
	assert.Check(t, is.Equal(expected, actual))

	type foo struct {
		One, Two, Three string
	}

	assert.Check(t, is.DeepEqual(foo{One: "one", Two: "two", Three: "tree"}, foo{One: "one", Two: "not two", Three: "tree"}))

	expectedFoo := foo{One: "one", Two: "two", Three: "tree"}
	actualFoo := foo{One: "one", Two: "not two", Three: "tree"}
	assert.Check(t, is.DeepEqual(expectedFoo, actualFoo))
	assert.Check(t, is.DeepEqual(actualFoo, expectedFoo))
}
```

Which will output something like;


```console
=== RUN   TestFoo
    completions_test.go:34: assertion failed: one (string) != two (string)
    completions_test.go:38: assertion failed: two (actual string) != one (expected string)
    completions_test.go:39: assertion failed: one (expected string) != two (actual string)
    completions_test.go:45: assertion failed: 
        --- ←
        +++ →
          main.foo{
          	One:   "one",
        - 	Two:   "two",
        + 	Two:   "not two",
          	Three: "tree",
          }
        
    completions_test.go:49: assertion failed: 
        --- expectedFoo
        +++ actualFoo
          main.foo{
          	One:   "one",
        - 	Two:   "two",
        + 	Two:   "not two",
          	Three: "tree",
          }
        
    completions_test.go:50: assertion failed: 
        --- actualFoo
        +++ expectedFoo
          main.foo{
          	One:   "one",
        - 	Two:   "not two",
        + 	Two:   "two",
          	Three: "tree",
          }
        
--- FAIL: TestFoo (0.00s)
```

templates/templates_test.go
thaJeztah: Yes, they _do_ produce the same output as the input; it's a bit of a hack.

When using the `table` format, the user specifies the template for rows; the same template is used for printing the columns (so that, say, swapping the order of columns in the template will also swap the table headers).

However, part of the template could be to (e.g.) truncate the output, uppercase/lowercase it, etc, and we don't want the table headers to be changed by those, so the `HeaderFunctions` override some of the existing functions (by default `basicFunctions`) with stub functions so that they don't take effect on the table headers.


```bash
docker image ls --format 'table {{upper .Repository}}:{{ .Tag}}\t{{lower .ID}}'
REPOSITORY:TAG               IMAGE ID
DOCKER-CLI-DEV:latest        0df2b0f75f92
DOCKER-DEV:latest            339109b3c24f
UBUNTU:24.04                 ab64a8382e93
NGINX:latest                 d2eb56950b84
NGINX:alpine                 2140dad235c1
```

thaJeztah: I initially had it in the test-table, but moved the `expected` variable here so that I could put a single comment to outline it's expected to always be the same for all test-cases (at least currently). But happy to add back `expected` in the test-table itself.

templates/templates_test.go
Benehiko: oh i understand, thank you for the explanation. Was just a bit confused as it seemed like the assignment served no purpose :see_no_evil:

templates/templates_test.go
Benehiko: ah okay. It seemed like the `expected` value inside the table was forgotten since it was declared https://github.com/docker/cli/pull/5541/files#diff-c6842e8d411be4977abca67d350e55f1cb1a144154846487c956c4ef172e6a17R99

I would be fine to keep it like this and remove the defined `expected` inside the test table.

thaJeztah: oh!! Yes, that's wrong; looks like I forgot to remove it there, thought I did; let me fix that part 🙈

thaJeztah: Done; removed it from the table 😅

Benehiko (approved):

laurazard (approved):

PR #5542, opened 2024-10-18
cmd/docker: add tests for flag-completions, and refactor

thaJeztah:
Remove the registerCompletionFuncForGlobalFlags for now, as the error it returned was ignored, so it didn't add much benefit, other than abstracting things.

Split the underlying completion-functions to separate functions, and add some basic tests for them.

Remove the completions helper, as it now didn't add much, and it saved having the dependency on the package.

### DIFF
--- a/cmd/docker/completions.go
+++ b/cmd/docker/completions.go
@@ -1,7 +1,6 @@
 package main
 
 import (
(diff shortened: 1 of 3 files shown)


laurazard (approved):
LGTM

PR #5547, opened 2024-10-19
cli/command/plugins: use errors.Join instead of custom cli.Errors, and deprecate cli.Errors

thaJeztah:
### cli/command/plugins: use errors.Join instead of custom cli.Errors

This command was using a custom "multi-error" implementation, but it
had some limitations, and the formatting wasn't great.

This patch replaces it with Go's errors.Join.

Before:

    docker plugin remove one two three
    Error response from daemon: plugin "one" not found, Error response from daemon: plugin "two" not found, Error response from daemon: plugin "three" not found

After:

    docker plugin remove one two three
    Error response from daemon: plugin "one" not found
    Error response from daemon: plugin "two" not found
    Error response from daemon: plugin "three" not found

### cli: deprecate Errors type

The Errors type is no longer used by the CLI itself, and this custom
"multi-error" implementation had both limitations (empty list not being
`nil`), as well as formatting not being great. All of this making it not
something to recommend, and better handled with Go's stdlib.

As far as I could find, there's no external consumers of this, but let's
deprecate first, and remove in the next release.

```markdown changelog
- improve formatting of errors during `docker plugin remove` 
- go-sdk: deprecate cli.Errors type in favour of Go's errors.Join
```

### DIFF
--- a/cli/command/plugin/remove.go
+++ b/cli/command/plugin/remove.go
@@ -2,6 +2,7 @@ package plugin
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/docker/cli/cli"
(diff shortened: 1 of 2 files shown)


cli/command/plugin/remove.go
```diff
@@ -36,17 +37,13 @@ func newRemoveCommand(dockerCli command.Cli) *cobra.Command {
 }
 
 func runRemove(ctx context.Context, dockerCli command.Cli, opts *rmOptions) error {
-	var errs cli.Errors
+	var errs error
 	for _, name := range opts.plugins {
 		if err := dockerCli.Client().PluginRemove(ctx, name, types.PluginRemoveOptions{Force: opts.force}); err != nil {
-			errs = append(errs, err)
+			errs = errors.Join(errs, err)
```
vvoland, line 43: This is not the same as the previous one, as it always produce `*joinError` with `[]error` of length 2. The first error is of type `*errors.joinError` which combines all of the previous errors.

vvoland: The correct way to use `errors.Join` in this case is to first collect errors into `errs []error` and then `return errors.Join(errs...)`

PR #5550, opened 2024-10-19
cli/command: PromptUserForCredentials: assorted minor improvements and (linting) fixes

thaJeztah:
### cli/command: PromptUserForCredentials: remove named output variables

This function has multiple conditional branches, which makes it harder
to see at a glance whether authConfig may be partially populated. This
patch instead returns a fresh instance for error returns to prevent any
confusion.

It also removes the named output variables, as they're now no longer used,
and the returned types should already be descriptive enough to understand
what's returned.

### cli/command: PromptUserForCredentials: inline isDefaultRegistry

remove isDefaultRegistry and inline it where it's used; the code-comment
already outlines what we're looking for, so the intermediate var didn't
add much currently.

### cli/command: PromptUserForCredentials: move "post" check for empty name

move the "post" check for username being empty inside the branch
that's handling the username, as it's the only branch where username
is mutated after checking if it's empty.

### cli/command: PromptUserForCredentials: move trimming where it's used

- move trimming defaultUsername inside the if-branch, as it's the only
  location where the result of the trimmed username is use.
- do the reverse for trimming argUser, because the result of trimming
  argUser is used outside of the if-branch (not just for the condition).
  putting it inside the condition makes it easy to assume the result is
  only used locally.

### cli/command: PromptUserForCredentials: always trim password

we don't support empty passwords; when prompting the user for a password,
we already trim the result, but we didn't do the same for a password that's
passed through stdin or through the `-p` / `--password` flag.

### cli/command: PromptUserForCredentials: print error on terminal restore fail

If restoring the terminal state fails, "echo" no longer works, which means
that anything the user types is no longer shown. The login itself may already
have succeeded, so we should not fail the command, but it's good to inform
the user that this happened, which may give them a clue why things no longer
work as they expect them to work.

With this patch:

    docker login -u yourname
    Password:
    Error: failed to restore terminal state to echo input: something bad happened
    
    Login Succeeded

We should consider printing instructions how  to restore this manually (other
than restarting the shell). e.g., 'run stty echo' when in a Linux or macOS shell,
but PowerShell and CMD.exe may need different instructions.

### cli/command: PromptUserForCredentials: use consts for all hints

This message resulted in code-lines that were too long; move it to a
const together with the other hint. While at it, also suppress unhandled
error, and touch-up the code-comment.

### cli/command: PromptUserForCredentials: suppress unhandled errors

Keep the linters (and my IDE) happy; these errors should be safe to ignore.

```markdown changelog


```

### DIFF
--- a/cli/command/registry.go
+++ b/cli/command/registry.go
@@ -150,10 +150,11 @@ func PromptUserForCredentials(ctx context.Context, cli Cli, argUser, argPassword
 		if argUser == "" {
 			argUser = defaultUsername
 		}
+		if argUser == "" {
+			return registrytypes.AuthConfig{}, errors.Errorf("Error: Non-null Username Required")
+		}
 	}
-	if argUser == "" {
-		return registrytypes.AuthConfig{}, errors.Errorf("Error: Non-null Username Required")
-	}
+
 	if argPassword == "" {
 		restoreInput, err := DisableInputEcho(cli.In())
 		if err != nil {
--- a/cli/command/registry.go
+++ b/cli/command/registry.go
@@ -123,9 +123,8 @@ func PromptUserForCredentials(ctx context.Context, cli Cli, argUser, argPassword
 		cli.SetIn(streams.NewIn(os.Stdin))
 	}
 
-	defaultUsername = strings.TrimSpace(defaultUsername)
-
-	if argUser = strings.TrimSpace(argUser); argUser == "" {
+	argUser = strings.TrimSpace(argUser)
+	if argUser == "" {
 		if serverAddress == registry.IndexServer {
 			// if this is a default registry (docker hub), then display the following message.
 			fmt.Fprintln(cli.Out(), "Log in with your Docker ID or email address to push and pull images from Docker Hub. If you don't have a Docker ID, head over to https://hub.docker.com/ to create one.")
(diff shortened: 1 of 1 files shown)


cli/command/registry.go
```diff
@@ -123,57 +123,69 @@ func PromptUserForCredentials(ctx context.Context, cli Cli, argUser, argPassword
...
-	defaultUsername = strings.TrimSpace(defaultUsername)
-
-	if argUser = strings.TrimSpace(argUser); argUser == "" {
-		if isDefaultRegistry {
+	argUser = strings.TrimSpace(argUser)
+	if argUser == "" {
+		if serverAddress == registry.IndexServer {
 			// if this is a default registry (docker hub), then display the following message.
-			fmt.Fprintln(cli.Out(), "Log in with your Docker ID or email address to push and pull images from Docker Hub. If you don't have a Docker ID, head over to https://hub.docker.com/ to create one.")
+			_, _ = fmt.Fprintln(cli.Out(), "Log in with your Docker ID or email address to push and pull images from Docker Hub. If you don't have a Docker ID, head over to https://hub.docker.com/ to create one.")
```
thaJeztah, line 130: LOL! Looks like I pushed it _JUST_ over the limit;

```
#16 64.40 cli/command/registry.go:130: line is 204 characters (lll)
#16 64.40 			_, _ = fmt.Fprintln(cli.Out(), "Log in with your Docker ID or email address to push and pull images from Docker Hub. If you don't have a Docker ID, head over to 
```

thaJeztah: Should be fixed now; added a commit to move the hint to a const.

laurazard (approved):
LGTM

PR #5553, opened 2024-10-19
cli/config/credentials: skip saving config-file if credentials didn't change

thaJeztah:
- fixes https://github.com/moby/moby/discussions/48529

Before this change, the config-file was always updated, even if there were no changes to save. This could cause issues when the config-file already had credentials set and was read-only for the current user.

For example, on NixOS, this poses a problem because `config.json` is a symlink to a write-protected file;

```bash
$ readlink ~/.docker/config.json
/home/username/.config/sops-nix/secrets/ghcr_auth

$ readlink -f ~/.docker/config.json
/run/user/1000/secrets.d/28/ghcr_auth
```

Which causes `docker login` to fail, even if no changes were to be made;

```console
Error saving credentials: rename /home/derek/.docker/config.json2180380217 /home/username/.config/sops-nix/secrets/ghcr_auth: invalid cross-device link
```

This patch updates the code to only update the config file if changes were detected. It there's nothing to save, it skips updating the file, as well as skips printing the warning about credentials being stored insecurely.

With this patch applied:

```bash
$ docker login -u yourname
Password:

WARNING! Your credentials are stored unencrypted in '/root/.docker/config.json'.
Configure a credential helper to remove this warning. See
https://docs.docker.com/go/credential-store/

Login Succeeded

$ docker login -u yourname
Password:
Login Succeeded
```

**- How to verify it**

**- Description for the changelog**

```markdown changelog
The `docker login` and `docker logout` command no longer update the configuration file if the credentials didn't change.
```

**- A picture of a cute animal (not mandatory but encouraged)**

PR #5570, opened 2024-10-22
cli/config/credentials: add test for save being idempotent

thaJeztah:
- relates to https://github.com/docker/cli/pull/5553

### cil/config/credentials: remove newStore() test-utility

This function was names slightly confusing, as it returns a fakeStore,
and it didn't do any constructing, so didn't provide value above just
constructing the type.

I'm planning to add more functionality to the fakeStore, but don't want
to maintain a full-fledged constructor for all of that, so let's remove
this utility.

### cli/config/credentials: add test for save being idempotent

Test case for d3f6867e4d7f5018ae4c0fbc709934893f0e95a2

```markdown changelog


```

### DIFF
--- a/cli/config/credentials/file_store_test.go
+++ b/cli/config/credentials/file_store_test.go
@@ -24,12 +24,8 @@ func (f *fakeStore) GetFilename() string {
 	return "/tmp/docker-fakestore"
 }
 
-func newStore(auths map[string]types.AuthConfig) store {
-	return &fakeStore{configs: auths}
-}
-
 func TestFileStoreAddCredentials(t *testing.T) {
-	f := newStore(make(map[string]types.AuthConfig))
+	f := &fakeStore{configs: map[string]types.AuthConfig{}}
 
 	s := NewFileStore(f)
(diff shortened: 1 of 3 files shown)


cli/config/credentials/file_store_test.go
```diff
@@ -21,15 +27,82 @@ func (f *fakeStore) GetAuthConfigs() map[string]types.AuthConfig {
...
+		Email:         "bar@example.com",
+		ServerAddress: "https://other.example.com",
+	}
+
+	expectedSaveCount = 1
+	t.Run("store new credentials", func(t *testing.T) {
+		assert.NilError(t, s.Store(authOne))
+		retrievedAuth, err := s.Get(authOne.ServerAddress)
+		assert.NilError(t, err)
+		assert.Check(t, is.DeepEqual(retrievedAuth, authOne))
```
thaJeztah, line 61: Actually, this can be just `is.Equal` as they should be comparable.

laurazard (approved):

PR #5572, opened 2024-10-24
cmd/image/tree: refactor

laurazard:
**- What I did**

Just some refactors to the image tree formatting logic to make reuse/reading easier. Consolidates the printing in one place, and reduces how far down we need to pass our writer.

This also makes testing these functions less involved.

**- How I did it**

(wip, I need to push more commits)

Make most of the formatting functions return strings, instead of printing directly.

**- How to verify it**

### DIFF
--- a/cli/command/image/tree.go
+++ b/cli/command/image/tree.go
@@ -4,12 +4,9 @@ import (
 	"context"
 	"fmt"
 	"sort"
-	"strings"
-	"unicode/utf8"
 
 	"github.com/containerd/platforms"
 	"github.com/docker/cli/cli/command"
-	"github.com/docker/cli/cli/streams"
 	"github.com/docker/docker/api/types/filters"
 	imagetypes "github.com/docker/docker/api/types/image"
 	"github.com/docker/docker/pkg/stringid"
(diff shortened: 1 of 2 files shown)


cli/command/image/tree.go
```diff
@@ -230,64 +229,76 @@ func printImageTree(dockerCLI command.Cli, view treeView) error {
...
 		_, _ = fmt.Fprintln(out)
 	}
 
 	return nil
 }
 
-func printDetails(out *streams.Out, headers []imgColumn, defaultColor aec.ANSI, details imageDetails) {
+func formatImageTree(img topImage, columns []imgColumn, imgSpacing bool, normalColor, topNameColor, untaggedColor aec.ANSI) (out string) {
+	out += printNames(columns, img, topNameColor, untaggedColor)
+	out += printDetails(columns, normalColor, img.Details)
```
thaJeztah, line 241: Slightly on the fence on this, because it means that we'd now be buffering all output in a string, instead sending it to the writer 🤔

laurazard: I didn't think that was a big deal for this output, but I can change it back.

laurazard: Thinking about it, I think I can do something a little nicer with the other refactors I haven't pushed yet. One sec.

PR #5574, opened 2024-10-25
c8d/push: better output if single-platform is pushed from index

laurazard:
Stacked on top of https://github.com/docker/cli/pull/5572

**- What I did**

Added a `docker images --tree`-style messaging to the "Not all multiplatform-content is present and only the available single-platform image was pushed" messaging during a `docker push`, to more clearly indicate explain what happened and which manifest was actually pushed.

Screenshot 2024-10-25 at 16 34 14

**- How I did it**

This requires adding information about which candidates were considered for pushing to the push aux stream (see https://github.com/moby/moby/commit/e37e09b44e26e04f395f786ee3329ff6f6cade06). Adding the output here makes sense, and does not imply walking the content store again).

**- How to verify it**

- `docker pull alpine`
- `docker pull --platform [other-platform] alpine`
- `docker tag alpine <my-namespace>/alpine`
- `docker push <my-namespace>/alpine`

```markdown changelog


```


laurazard:
This is a draft! It needs cleaning up and I need to open a PR in moby to (discuss and) add the additional info to the aux output.

PR #5579, opened 2024-10-28
cli/command/container: parse: remove client-side warning

thaJeztah:
- relates to https://github.com/moby/moby/pull/48789

remove a client-side warning about volume drivers combined with "mounts" in favor of producing the warning on the daemon side.

### DIFF
--- a/cli/command/container/opts.go
+++ b/cli/command/container/opts.go
@@ -23,7 +23,6 @@ import (
 	"github.com/docker/docker/errdefs"
 	"github.com/docker/go-connections/nat"
 	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	cdi "tags.cncf.io/container-device-interface/pkg/parser"
 )
@@ -364,10 +363,6 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		return nil, errors.Errorf("invalid value: %d. Valid memory swappiness range is 0-100", swappiness)
 	}
 
-	mounts := copts.mounts.Value()
-	if len(mounts) > 0 && copts.volumeDriver != "" {
-		logrus.Warn("`--volume-driver` is ignored for volumes specified via `--mount`. Use `--mount type=volume,volume-driver=...` instead.")
-	}
 	var binds []string
 	volumes := copts.volumes.GetMap()
 	// add any bind targets to the list of container volumes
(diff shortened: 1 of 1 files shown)


cli/command/container/opts.go
```diff
@@ -364,10 +363,6 @@ func parse(flags *pflag.FlagSet, copts *containerOptions, serverOS string) (*con
 		return nil, errors.Errorf("invalid value: %d. Valid memory swappiness range is 0-100", swappiness)
 	}
 
-	mounts := copts.mounts.Value()
-	if len(mounts) > 0 && copts.volumeDriver != "" {
-		logrus.Warn("`--volume-driver` is ignored for volumes specified via `--mount`. Use `--mount type=volume,volume-driver=...` instead.")
-	}
```
vvoland, line 370: Should this be API-version gated to match the lack of warning on pre 1.48 daemons?

Considering a case where a >=1.48 client talks to a <1.48 daemon.

thaJeztah: I was briefly considering if, but in all honesty, I think it's a very corner-case on its own, and even more so considering a current CLI connecting to an older daemon, so I thought the world wouldn't explode if someone would miss that warning.

The reason I stated looking into this was that the current warning is not very user-friendly because we use logrus with the default formatting;

```bash
docker run --volume-driver something --rm --mount=type=volume,target=/toto busybox
time="2024-10-26T20:20:54+02:00" level=warning msg="`--volume-driver` is ignored for volumes specified via `--mount`. Use `--mount type=volume,volume-driver=...` instead."
```

I stated a PR to change it to a bare `fmt.Printfln()`, but even wondered if logging should happening here _at all_ (this might be a function to export somewhere at some point to be used as library code). Then considered that we already have warnings, but those were returned by the API (which is somewhat better as it can act as the canonical source of truth here).

thaJeztah: Oh, LOL, and to make my "Yak-shaving" tour complete; The _acttual_ reason I arrived here in the first place was that I was considering that we should deprecate `--volume-driver`, and instead require explicitly setting `--mount` with the desired options. I still need to consider the full scope of that, because `--volume-driver` can be used as a default for anonymous volumes (but not sure if we _really_ think that's a good idea - perhaps?) and even in that case, perhaps it's still better for the user to explicitly define that they want a specific volume to use a specific driver.

vvoland (approved):

PR #5580, opened 2024-10-28
Improve Cobra completions for `run` and `create`

albers:
**- What I did**
Added Cobra completion for all options of `docker run` and `docker create`.

**- How I did it**
Provided `FlagCompletionFunc`s for all options except for
- boolean flags (default completion works)
- flags with file completions (default completion works)
- flags that already had a completion function.

The reference for my work was the legacy bash completion.
Features that I did not recreate:
- completion for `--user`
- completion for the individual log driver options
- completion of non-built-in drivers

**- How to verify it**
1. In a dev container, issue `make completion`, then try completions of various `docker run` and `docker create` commands.
2. In another terminal, exec into the container, install the legacy completion with `. ./contrib/completion/bash/docker`. then execute the same completions for reference.

Improved completions for `docker run` and `docker create`.

```markdown changelog
Ported some completions from the bash completion to the new cobra based completion.
```

### DIFF
--- a/cli/command/container/completion.go
+++ b/cli/command/container/completion.go
@@ -71,6 +71,7 @@ func addCompletions(cmd *cobra.Command, dockerCLI completion.APIClientProvider) 
 	_ = cmd.RegisterFlagCompletionFunc("stop-signal", completeSignals)
 	_ = cmd.RegisterFlagCompletionFunc("storage-opt", completeStorageOpt)
 	_ = cmd.RegisterFlagCompletionFunc("ulimit", completeUlimit)
+	_ = cmd.RegisterFlagCompletionFunc("userns", completion.FromList("host"))
 	_ = cmd.RegisterFlagCompletionFunc("volumes-from", completion.ContainerNames(dockerCLI, true))
 }
 
--- a/cli/command/container/completion.go
+++ b/cli/command/container/completion.go
@@ -64,6 +64,7 @@ func addCompletions(cmd *cobra.Command, dockerCLI completion.APIClientProvider) 
 	_ = cmd.RegisterFlagCompletionFunc("ipc", completeIpc(dockerCLI))
 	_ = cmd.RegisterFlagCompletionFunc("link", completeLink(dockerCLI))
 	_ = cmd.RegisterFlagCompletionFunc("network", completion.NetworkNames(dockerCLI))
+	_ = cmd.RegisterFlagCompletionFunc("pid", completePid(dockerCLI))
 	_ = cmd.RegisterFlagCompletionFunc("platform", completion.Platforms)
 	_ = cmd.RegisterFlagCompletionFunc("pull", completion.FromList(PullImageAlways, PullImageMissing, PullImageNever))
 	_ = cmd.RegisterFlagCompletionFunc("restart", completeRestartPolicies)
@@ -96,6 +97,20 @@ func completeIpc(dockerCLI completion.APIClientProvider) func(cmd *cobra.Command
 func completeLink(dockerCLI completion.APIClientProvider) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return postfixWith(":", containerNames(dockerCLI, cmd, args, toComplete)), cobra.ShellCompDirectiveNoSpace
+	}
+}
+
+// completePid implements shell completion for the `--pid` option  of `run` and `create`.
+func completePid(dockerCLI completion.APIClientProvider) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(toComplete) > 0 && strings.HasPrefix("container", toComplete) { //nolint:gocritic // not swapped, matches partly typed "container"
+			return []string{"container:"}, cobra.ShellCompDirectiveNoSpace
+		}
+		if strings.HasPrefix(toComplete, "container:") {
+			names, _ := completion.ContainerNames(dockerCLI, true)(cmd, args, toComplete)
+			return prefixWith("container:", names), cobra.ShellCompDirectiveNoFileComp
+		}
+		return []string{"container:", "host"}, cobra.ShellCompDirectiveNoFileComp
 	}
 }
 
--- a/cli/command/container/completion.go
+++ b/cli/command/container/completion.go
@@ -61,12 +61,43 @@ func addCompletions(cmd *cobra.Command, dockerCLI completion.APIClientProvider) 
 	_ = cmd.RegisterFlagCompletionFunc("cap-drop", completeLinuxCapabilityNames)
 	_ = cmd.RegisterFlagCompletionFunc("env", completion.EnvVarNames)
 	_ = cmd.RegisterFlagCompletionFunc("env-file", completion.FileNames)
+	_ = cmd.RegisterFlagCompletionFunc("ipc", completeIpc(dockerCLI))
 	_ = cmd.RegisterFlagCompletionFunc("network", completion.NetworkNames(dockerCLI))
 	_ = cmd.RegisterFlagCompletionFunc("platform", completion.Platforms)
 	_ = cmd.RegisterFlagCompletionFunc("pull", completion.FromList(PullImageAlways, PullImageMissing, PullImageNever))
 	_ = cmd.RegisterFlagCompletionFunc("restart", completeRestartPolicies)
 	_ = cmd.RegisterFlagCompletionFunc("stop-signal", completeSignals)
 	_ = cmd.RegisterFlagCompletionFunc("volumes-from", completion.ContainerNames(dockerCLI, true))
+}
+
+// completeIpc implements shell completion for the `--ipc` option of `run` and `create`.
+// The completion is partly composite.
+func completeIpc(dockerCLI completion.APIClientProvider) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(toComplete) > 0 && strings.HasPrefix("container", toComplete) { //nolint:gocritic // not swapped, matches partly typed "container"
+			return []string{"container:"}, cobra.ShellCompDirectiveNoSpace
+		}
+		if strings.HasPrefix(toComplete, "container:") {
+			names, _ := completion.ContainerNames(dockerCLI, true)(cmd, args, toComplete)
+			return prefixWith("container:", names), cobra.ShellCompDirectiveNoFileComp
+		}
+		return []string{
+			string(container.IPCModeContainer + ":"),
+			string(container.IPCModeHost),
+			string(container.IPCModeNone),
+			string(container.IPCModePrivate),
+			string(container.IPCModeShareable),
+		}, cobra.ShellCompDirectiveNoFileComp
+	}
+}
+
+// prefixWith prefixes every element in the slice with the given prefix.
+func prefixWith(prefix string, values []string) []string {
+	result := make([]string, len(values))
+	for i, v := range values {
+		result[i] = prefix + v
+	}
+	return result
 }
 
 func completeLinuxCapabilityNames(cmd *cobra.Command, args []string, toComplete string) (names []string, _ cobra.ShellCompDirective) {
--- a/cli/command/container/completion.go
+++ b/cli/command/container/completion.go
@@ -134,6 +134,7 @@ func addCompletions(cmd *cobra.Command, dockerCLI completion.APIClientProvider) 
 	_ = cmd.RegisterFlagCompletionFunc("storage-opt", completeStorageOpt)
 	_ = cmd.RegisterFlagCompletionFunc("ulimit", completeUlimit)
 	_ = cmd.RegisterFlagCompletionFunc("userns", completion.FromList("host"))
+	_ = cmd.RegisterFlagCompletionFunc("uts", completion.FromList("host"))
 	_ = cmd.RegisterFlagCompletionFunc("volumes-from", completion.ContainerNames(dockerCLI, true))
 }
 
--- a/cli/command/container/completion.go
+++ b/cli/command/container/completion.go
@@ -44,6 +44,65 @@ var allLinuxCapabilities = sync.OnceValue(func() []string {
 	return out
 })
 
+// logDriverOptions provides the options for each built-in logging driver.
+var logDriverOptions = map[string][]string{
+	"awslogs": {
+		"max-buffer-size", "mode", "awslogs-create-group", "awslogs-credentials-endpoint", "awslogs-datetime-format",
+		"awslogs-group", "awslogs-multiline-pattern", "awslogs-region", "awslogs-stream", "tag",
+	},
+	"fluentd": {
+		"max-buffer-size", "mode", "env", "env-regex", "labels", "fluentd-address", "fluentd-async",
+		"fluentd-buffer-limit", "fluentd-request-ack", "fluentd-retry-wait", "fluentd-max-retries",
+		"fluentd-sub-second-precision", "tag",
+	},
+	"gcplogs": {
+		"max-buffer-size", "mode", "env", "env-regex", "labels", "gcp-log-cmd", "gcp-meta-id", "gcp-meta-name",
+		"gcp-meta-zone", "gcp-project",
+	},
+	"gelf": {
+		"max-buffer-size", "mode", "env", "env-regex", "labels", "gelf-address", "gelf-compression-level",
+		"gelf-compression-type", "gelf-tcp-max-reconnect", "gelf-tcp-reconnect-delay", "tag",
+	},
+	"journald":  {"max-buffer-size", "mode", "env", "env-regex", "labels", "tag"},
+	"json-file": {"max-buffer-size", "mode", "env", "env-regex", "labels", "compress", "max-file", "max-size"},
+	"local":     {"max-buffer-size", "mode", "compress", "max-file", "max-size"},
+	"none":      {},
+	"splunk": {
+		"max-buffer-size", "mode", "env", "env-regex", "labels", "splunk-caname", "splunk-capath", "splunk-format",
+		"splunk-gzip", "splunk-gzip-level", "splunk-index", "splunk-insecureskipverify", "splunk-source",
+		"splunk-sourcetype", "splunk-token", "splunk-url", "splunk-verify-connection", "tag",
+	},
+	"syslog": {
+		"max-buffer-size", "mode", "env", "env-regex", "labels", "syslog-address", "syslog-facility", "syslog-format",
+		"syslog-tls-ca-cert", "syslog-tls-cert", "syslog-tls-key", "syslog-tls-skip-verify", "tag",
+	},
+}
+
+// builtInLogDrivers provides a list of the built-in logging drivers.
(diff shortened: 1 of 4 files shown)


cli/command/container/completion.go
```diff
@@ -125,6 +126,23 @@ func completeIpc(cli command.Cli) func(cmd *cobra.Command, args []string, toComp
...
+	}
+}
+
+// containerNames contacts the API to get names and optionally IDs of containers.
+// In case of an error, an empty list is returned.
+func containerNames(dockerCLI completion.APIClientProvider, cmd *cobra.Command, args []string, toComplete string) []string {
+	names, _ := completion.ContainerNames(dockerCLI, true)(cmd, args, toComplete)
+	if names == nil {
+		return []string{}
+	}
```
thaJeztah, line 315: Just curious; is this a bug in `completion.ContainerNames` and should we update that one?

cli/command/container/completion.go
```diff
@@ -145,6 +146,20 @@ func completeLink(cli command.Cli) func(cmd *cobra.Command, args []string, toCom
 	}
 }
 
+// completePid implements shell completion for the `--pid` option  of `run` and `create`.
+func completePid(cli command.Cli) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
thaJeztah, line 240: Same comments here as on `completIpc`  further up 🤗  (early return, using `completion.APIClientProvider`, naming)

```go
// completePIDMode implements shell completion for the `--pid` option  of `run` and `create`.
func completePIDMode(dockerCLI completion.APIClientProvider) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```

cli/command/container/completion.go
```diff
@@ -106,6 +107,30 @@ func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
...
+func completeIpc(cli command.Cli) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(toComplete) > 0 && strings.HasPrefix("container", toComplete) { //nolint:gocritic
+			return []string{"container:"}, cobra.ShellCompDirectiveNoSpace
+		}
+		if strings.HasPrefix(toComplete, "container:") {
+			names, _ := completion.ContainerNames(cli, true)(cmd, args, toComplete)
+			return prefixWith("container:", names), cobra.ShellCompDirectiveNoFileComp
+		}
+		return []string{"container:", "host", "none", "private", "shareable"}, cobra.ShellCompDirectiveNoFileComp
```
thaJeztah, line 208: We _could_ consider using the consts for this; https://github.com/docker/cli/blob/32ff200fe6d843843718709ffa94023eb7356108/vendor/github.com/docker/docker/api/types/container/hostconfig.go#L82-L88

cli/command/container/completion.go
```diff
@@ -145,6 +146,20 @@ func completeLink(cli command.Cli) func(cmd *cobra.Command, args []string, toCom
...
+func completePid(cli command.Cli) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(toComplete) > 0 && strings.HasPrefix("container", toComplete) { //nolint:gocritic
+			return []string{"container:"}, cobra.ShellCompDirectiveNoSpace
+		}
+		if strings.HasPrefix(toComplete, "container:") {
+			names, _ := completion.ContainerNames(cli, true)(cmd, args, toComplete)
+			return prefixWith("container:", names), cobra.ShellCompDirectiveNoFileComp
+		}
+		return []string{"container:", "host"}, cobra.ShellCompDirectiveNoFileComp
```
thaJeztah, line 248: Note to self; looks like (unlike IPCMode) we don't define consts for this yet.

cli/command/container/completion_test.go
```diff
@@ -21,6 +25,48 @@ func TestCompleteLinuxCapabilityNames(t *testing.T) {
 	}
 }
 
+func TestCompletePid(t *testing.T) {
```
thaJeztah, line 28: If you decide to change the name then the test should probably also be renamed; `TestCompletePIDMode`

cli/command/container/completion_test.go
```diff
@@ -21,6 +25,48 @@ func TestCompleteLinuxCapabilityNames(t *testing.T) {
...
+	for _, tc := range tests {
+		cli := test.NewFakeCli(&fakeClient{
+			containerListFunc: tc.containerListFunc,
+		})
+
+		completions, directive := completePid(cli)(NewRunCommand(cli), nil, tc.toComplete)
+
+		assert.DeepEqual(t, completions, tc.expectedCompletions)
+		assert.Equal(t, directive, tc.expectedDirective, fmt.Sprintf("wrong directive in completion for '%s'", tc.toComplete))
+	}
```
thaJeztah, line 67: This can use a sub-test so that we don't fail early, and each of the test-cases can be run separately. Using a sub-test also means that it's not needed to put the custom comments (`fmt.Sprintf`)  in the assertions to find "which of the tests failed";

```go
t.Run(tc.toComplete, func(t *testing.T) {
		containerListFunc: tc.containerListFunc,
	})

	completions, directive := completePIDMode(cli)(NewRunCommand(cli), nil, tc.toComplete)
	assert.DeepEqual(t, completions, tc.expectedCompletions)
	assert.Equal(t, directive, tc.expectedDirective)
})
```

I also tend to use `assert.Check` where possible; for cases where both results _can_ be asserted, that would allow both to be done, even if one fails (it gives slightly more information on failing tests);

```go
assert.Check(t, is.DeepEqual(completions, tc.expectedCompletions))
assert.Check(t, is.Equal(directive, tc.expectedDirective))
```

cli/command/container/opts_test.go
```diff
@@ -79,7 +79,7 @@ func TestParseRunLinks(t *testing.T) {
 		t.Fatalf("Error parsing links. Expected []string{\"a:b\", \"c:d\"}, received: %v", hostConfig.Links)
 	}
 	if _, hostConfig, _ := mustParse(t, ""); len(hostConfig.Links) != 0 {
-		t.Fatalf("Error parsing links. No link expected, received: %v", hostConfig.Links)
+		t.Fatalf("Error parsing links. No link expectedCompletions, received: %v", hostConfig.Links)
```
thaJeztah, line 82: I _think_ this was an accidental "find and replace"

cli/command/container/completion.go
```diff
@@ -119,6 +119,7 @@ func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
 	_ = cmd.RegisterFlagCompletionFunc("pid", completePid(dockerCli))
 	_ = cmd.RegisterFlagCompletionFunc("pids-limit", completion.NoComplete)
 	_ = cmd.RegisterFlagCompletionFunc("platform", completion.Platforms)
+	_ = cmd.RegisterFlagCompletionFunc("publish", completion.NoComplete)
```
thaJeztah, line 173: Ah! Also one to bite my teeth in at some point (for the advanced syntax).

cli/command/container/completion.go
```diff
@@ -122,6 +122,7 @@ func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
 	_ = cmd.RegisterFlagCompletionFunc("publish", completion.NoComplete)
 	_ = cmd.RegisterFlagCompletionFunc("pull", completion.FromList(PullImageAlways, PullImageMissing, PullImageNever))
 	_ = cmd.RegisterFlagCompletionFunc("restart", completeRestartPolicies)
+	_ = cmd.RegisterFlagCompletionFunc("runtime", completion.NoComplete)
```
thaJeztah, line 176: We _could_ possibly provide some amount of completion for this (known runtimes in the daemon), but not sure if we should, given that with containerd as runtime, alternative runtimes can be used that are not defined in the daemon config.

cli/command/container/completion.go
```diff
@@ -131,6 +131,7 @@ func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
 	_ = cmd.RegisterFlagCompletionFunc("sysctl", completion.NoComplete)
 	_ = cmd.RegisterFlagCompletionFunc("ulimit", completeUlimit)
 	_ = cmd.RegisterFlagCompletionFunc("user", completion.NoComplete)
+	_ = cmd.RegisterFlagCompletionFunc("userns", completion.FromList("host"))
```
thaJeztah, line 185: Note to self; looks like we don't define an enum/consts for this (there's only a single option, so maybe not really needed, but we do define a type for this; https://github.com/docker/cli/blob/32ff200fe6d843843718709ffa94023eb7356108/vendor/github.com/docker/docker/api/types/container/hostconfig.go#L171-L172

cli/command/container/completion.go
```diff
@@ -132,6 +132,7 @@ func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
 	_ = cmd.RegisterFlagCompletionFunc("ulimit", completeUlimit)
 	_ = cmd.RegisterFlagCompletionFunc("user", completion.NoComplete)
 	_ = cmd.RegisterFlagCompletionFunc("userns", completion.FromList("host"))
+	_ = cmd.RegisterFlagCompletionFunc("uts", completion.NoComplete)
```
thaJeztah, line 186: This one _could_ have completion; theres `UTSMode` "host"; https://github.com/docker/cli/blob/32ff200fe6d843843718709ffa94023eb7356108/vendor/github.com/docker/docker/api/types/container/hostconfig.go#L210-L211


(also note to self that we don't provide an enum for that one)

cli/command/container/completion.go
```diff
@@ -133,6 +133,7 @@ func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
 	_ = cmd.RegisterFlagCompletionFunc("user", completion.NoComplete)
 	_ = cmd.RegisterFlagCompletionFunc("userns", completion.FromList("host"))
 	_ = cmd.RegisterFlagCompletionFunc("uts", completion.NoComplete)
+	_ = cmd.RegisterFlagCompletionFunc("volume-driver", completion.FromList("local"))
```
thaJeztah, line 187: We _could_ get this from the `docker info` response; 

```bash
docker info --format '{{json .Plugins}}' | jq .
{
  "Volume": [
    "local"
  ],
  "Network": [
    "bridge",
    "host",
    "ipvlan",
    "macvlan",
    "null",
    "overlay"
  ],
  "Authorization": null,
  "Log": [
    "awslogs",
    "fluentd",
    "gcplogs",
    "gelf",
    "journald",
    "json-file",
    "local",
    "splunk",
    "syslog"
  ]
}
```

cli/command/container/completion.go
```diff
@@ -138,6 +138,11 @@ func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
 	_ = cmd.RegisterFlagCompletionFunc("workdir", completion.NoComplete)
 }
 
+// completeDetachKeys implements shell completion for the `--detach-keys` option of `run` and `create`.
+func completeDetachKeys(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+	return []string{"ctrl-"}, cobra.ShellCompDirectiveNoSpace
```
thaJeztah, line 194: Note to self; I need to lookup if there's other special ones we define (`meta-`, `alt-`, possibly others)

cli/command/container/completion_test.go
```diff
@@ -74,6 +74,57 @@ func TestCompleteRestartPolicies(t *testing.T) {
...
+			expectedDirective:   cobra.ShellCompDirectiveNoFileComp,
+		},
+		{
+			toComplete:          "sy",
+			expectedCompletions: []string{"apparmor=", "label=", "no-new-privileges", "seccomp=", "systempaths=unconfined"},
+			expectedDirective:   cobra.ShellCompDirectiveNoFileComp,
+		},
+	}
+
+	for _, tc := range tests {
```
thaJeztah, line 121: This one could also use sub-tests and the updated assertions (see my other comment)

cli/command/container/completion.go
```diff
@@ -165,6 +216,25 @@ func completeLink(cli command.Cli) func(cmd *cobra.Command, args []string, toCom
 	}
 }
 
+// completeLogDriver implements shell completion for the `--log-driver` option  of `run` and `create`.
+func completeLogDriver(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+	drivers := make([]string, 0, len(logDriverOptions))
```
thaJeztah, line 221: Technically we could get the list of log-drivers from the `info` response as well (I'm slightly on the fence if doing so would be "too heavy" or still acceptable)

thaJeztah (commented):
This is amazing work, thank you!! I left quite some comments, but welcome your feedback/thoughts.

I think the most notable one that could reduce a lot of boiler plating, is the comment on (perhaps could be a utility) setting a "default" `NoComplete` on all flags; that way we wouldn't have to set the `NoComplete` for all flags that don't have one. (Although maybe there's something to be said for _explicitly_ setting "no completions here)

cli/command/container/completion.go
```diff
@@ -54,6 +55,20 @@ var restartPolicies = []string{
 	string(container.RestartPolicyUnlessStopped),
 }
 
+// addCompletions adds the completions that `run` and `create` have in common.
+func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
```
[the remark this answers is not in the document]
albers: I like the idea very much, but I am not sure about how much completion code is appreciated by hard core developers.
See my thoughts in https://github.com/docker/cli/pull/5581#issuecomment-2444074439.

cli/command/container/completion.go
```diff
@@ -57,6 +57,7 @@ var restartPolicies = []string{
 
 // addCompletions adds the completions that `run` and `create` have in common.
 func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
+	_ = cmd.RegisterFlagCompletionFunc("add-host", completion.NoComplete)
```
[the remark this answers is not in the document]
albers: I could not find a way to set a default `ShellCompDirective` for silencing the numerouns flags with inapproptiate file completions.

IMHO this should be handled in Cobra. I'll open a feature request for this.

Your workaround looks very promising, I'll give it a try.

cli/command/container/completion.go
```diff
@@ -106,6 +107,30 @@ func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
 	_ = cmd.RegisterFlagCompletionFunc("volumes-from", completion.ContainerNames(dockerCli, true))
 }
 
+// completeIpc implements shell completion for the `--ipc` option of `run` and `create`.
+// The completion is partly composite.
+func completeIpc(cli command.Cli) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(toComplete) > 0 && strings.HasPrefix("container", toComplete) { //nolint:gocritic
```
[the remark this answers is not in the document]
albers: The argument order is intended.
`strings.HasPrefix("container", toComplete)` catches all cases where `toComplete` is a substring of "container", starting from the beginning, excluded the empty string case. `to complete`should be filled up to "container:".

`strings.HasPrefix(toComplete, "container:")` matches empty or partly typed container names and completes to "container:" + container name.

Thanks for the hint on comments, I added some.

cli/command/container/completion.go
```diff
@@ -125,6 +126,23 @@ func completeIpc(cli command.Cli) func(cmd *cobra.Command, args []string, toComp
 	}
 }
 
+// completeLink implements shell completion for the `--link` option  of `run` and `create`.
+func completeLink(cli command.Cli) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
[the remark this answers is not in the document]
albers: done

cli/command/container/completion.go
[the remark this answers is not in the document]
albers: signature is fixed.

cli/command/container/completion.go
```diff
@@ -59,6 +59,7 @@ var restartPolicies = []string{
 func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
 	_ = cmd.RegisterFlagCompletionFunc("add-host", completion.NoComplete)
 	_ = cmd.RegisterFlagCompletionFunc("annotation", completion.NoComplete)
+	_ = cmd.RegisterFlagCompletionFunc("attach", completion.FromList("stderr", "stdin", "stdout"))
```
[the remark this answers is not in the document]
albers: Ah, now I understand.
I was also wondering why the `events --filter` completion only works once.

cli/command/container/completion.go
```diff
@@ -65,6 +65,7 @@ func addCompletions(cmd *cobra.Command, dockerCli command.Cli) {
 	_ = cmd.RegisterFlagCompletionFunc("cap-add", completeLinuxCapabilityNames)
 	_ = cmd.RegisterFlagCompletionFunc("cap-drop", completeLinuxCapabilityNames)
 	_ = cmd.RegisterFlagCompletionFunc("cgroup-parent", completion.NoComplete)
+	_ = cmd.RegisterFlagCompletionFunc("cgroupns", completion.FromList("host", "private"))
```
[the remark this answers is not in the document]
albers: done

cli/command/container/completion.go
albers: done

cli/command/container/completion.go
albers: done

cli/command/container/completion.go
albers: `completion.ContainerNames` is consistent with the other helper functions in completion.go.
They all return `nil, cobra.ShellCompDirectiveError` in case of api errors, which is an appropriate way of signaling completion problems.

I just wanted to get a list of container names that I can feed into `postfixWith` directly without having to bother with api errors.
That's why I adapted the existing function to my use case.
I had to do so anyway because I had to get rid of the ShellCompDirective.

From my point of view, a better solution would have been to pull out a function from `completion.ContainerNames` that gets the container names directly as an array (no `nil`).
But this would have caused naming issues: This function should obviously be named `completion.ContainerNames` and thus collide with the existing function.
So the existing function, and for consistency similar functions, would all have to be renamed (`completion.ContainerNames` -> `completion.CompleteContainerNames`, `completion.NetworkNames` -> `completion.NetworkNames`, ...)

cli/command/container/opts_test.go
albers: Ah, yes. Thanks for spotting.

cli/command/container/completion_test.go
albers: done

albers:
@thaJeztah Thanks very much for your mentoring. It's very much appreciated.

I think I'm through with your review comments, please take another look.

Benehiko (approved):
LGTM

albers:
Let me know when I can squash the commits into a more relevant set of commits.

laurazard:
Thank you so much for this @albers!

I'm thinking we could squash https://github.com/docker/cli/pull/5580/commits/4378ab4d922e6b5c904775a63cf2f19b1c3d2e4f together with all the "disable file completions for X" commit.

laurazard (approved):

thaJeztah (approved):
LGTM!

Sorry thought I already did 🙈 🙈 🙈

PR #5586, opened 2024-10-30
Buffer 'docker stats' text to avoid terminal flickering

### LINKED ISSUE #4348: docker stats flickers
### Description

Running the command `docker stats` results in a flickering output.

### Reproduce

docker stats

### Expected behavior

`docker stats` should not flicker.

### docker version

```bash
Client:
 Cloud integration: v1.0.33
 Version:           24.0.2
 API version:       1.43
 Go version:        go1.20.4
 Git commit:        cb74dfc
 Built:             Thu May 25 21:53:15 2023
 OS/Arch:           windows/amd64
 Context:           default

Server: Docker Desktop 4.20.1 (110738)
 Engine:
  Version:          24.0.2
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.4
  Git commit:       659604f
  Built:            Thu May 25 21:52:17 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.21
  GitCommit:        3dce8eb055cbb6872793272b4f20ed16117344f8
 runc:
  Version:          1.1.7
  GitCommit:        v1.1.7-0-g860f061
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
Client:
 Version:    24.0.2
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.10.5
    Path:     C:\Program Files\Docker\cli-plugins\docker-buildx.exe
    Version:  v2.18.1
    Path:     C:\Program Files\Docker\cli-plugins\docker-compose.exe
    Version:  v0.1.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-dev.exe
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.19
    Path:     C:\Program Files\Docker\cli-plugins\docker-extension.exe
    Version:  v0.1.0-beta.4
    Path:     C:\Program Files\Docker\cli-plugins\docker-init.exe
    Version:  0.6.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-sbom.exe
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-scan.exe
  scout: Command line tool for Docker Scout (Docker Inc.)
    Version:  v0.12.0
    Path:     C:\Program Files\Docker\cli-plugins\docker-scout.exe

Server:
 Containers: 5
  Running: 0
  Paused: 0
  Stopped: 5
 Images: 39
 Server Version: 24.0.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3dce8eb055cbb6872793272b4f20ed16117344f8
 runc version: v1.1.7-0-g860f061
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
 Kernel Version: 5.10.16.3-microsoft-standard-WSL2
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 7.188GiB
 Name: docker-desktop
 ID: 939218a6-637b-4631-9343-8b350b544b7f
 Debug Mode: false
 HTTP Proxy: http.docker.internal:3128
 HTTPS Proxy: http.docker.internal:3128
 No Proxy: hubproxy.docker.internal
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5555
  127.0.0.0/8
 Live Restore Enabled: false

WARNING: No blkio throttle.read_bps_device support
WARNING: No blkio throttle.write_bps_device support
WARNING: No blkio throttle.read_iops_device support
WARNING: No blkio throttle.write_iops_device support
```

### Additional Info

Tested in the Winfows terminal and Cmder

Giedriusj1:
This change reduces the flickering of the terminal when running `docker stats` by buffering the formatted stats text and printing it in one write.

Should also consume less CPU as we now only have to issue a single syscall to write the stats text to the terminal.

I suspect it fixes #4348, but I can not confirm since that issue was reported on WSL.

**- What I did**

On all of my Linux machines (Fedora Linux + GNOME) I can see terminal flickering when running 'docker stats'.

I've tested this with gnome-terminal/xfce4-terminal/xterm and it flickers on all. The only term that does not flicker is kitty.

I've also tested this on GNOME running Wayland and X11. I can see no difference in behaviour between Wayland/X11. 

I've tested this on two machines:  Thinkpad T14s gen1 running AMD zen, and a desktop PC running AMD zen chip with Radeon GPU. Both behave identically.

recording of the flickering:
https://github.com/user-attachments/assets/05f8278e-05c3-411e-bb32-dad262136937

'docker stats' used to issue multiple write calls to stdout. Starting with the "clear screen" command and following multiple small write requests. 
I'm now bundling all of those into a single string that is then written out to the stdout as a single request.

This fixes the flickering (as well as reduces syscalls required for each update) on both of my machines.

I've tested this on Fedora 41 GNOME with the following terminal emulators: gnome-terminal/xfce4-terminal/xterm/kitty.

I've tried things like resizing the window as 'docker stats' was running. It all works flawlessly now :)

```markdown changelog
`docker stats` output is now buffered to reduce flickering issues 
```

A cat that sometimes roams outside my flat. Me and my flatmates call him Phantom :cat: 

cat_phantom

### DIFF
--- a/cli/command/container/stats.go
+++ b/cli/command/container/stats.go
@@ -1,6 +1,7 @@
 package container
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -264,31 +265,40 @@ func RunStats(ctx context.Context, dockerCLI command.Cli, options *StatsOptions)
 		// so we unlikely hit this code in practice.
 		daemonOSType = dockerCLI.ServerInfo().OSType
 	}
+
+	// Buffer to store formatted stats text.
+	// Once formatted, it will be printed in one write to avoid screen flickering.
+	var statsTextBuffer bytes.Buffer
+
 	statsCtx := formatter.Context{
-		Output: dockerCLI.Out(),
+		Output: &statsTextBuffer,
 		Format: NewStatsFormat(format, daemonOSType),
-	}
-	cleanScreen := func() {
-		if !options.NoStream {
-			_, _ = fmt.Fprint(dockerCLI.Out(), "\033[2J")
-			_, _ = fmt.Fprint(dockerCLI.Out(), "\033[H")
-		}
 	}
 
 	var err error
 	ticker := time.NewTicker(500 * time.Millisecond)
 	defer ticker.Stop()
 	for range ticker.C {
-		cleanScreen()
 		var ccStats []StatsEntry
 		cStats.mu.RLock()
 		for _, c := range cStats.cs {
 			ccStats = append(ccStats, c.GetStatistics())
 		}
 		cStats.mu.RUnlock()
+
+		if !options.NoStream {
+			// Start by clearing the screen and moving the cursor to the top-left
+			_, _ = fmt.Fprint(&statsTextBuffer, "\033[2J\033[H")
+		}
+
 		if err = statsFormatWrite(statsCtx, ccStats, daemonOSType, !options.NoTrunc); err != nil {
 			break
 		}
+
+		_, _ = fmt.Fprint(dockerCLI.Out(), statsTextBuffer.String())
+
+		statsTextBuffer.Reset()
+
 		if len(cStats.cs) == 0 && !showAll {
 			break
 		}


cli/command/container/stats.go
```diff
@@ -264,31 +264,40 @@ func RunStats(ctx context.Context, dockerCLI command.Cli, options *StatsOptions)
 		// so we unlikely hit this code in practice.
 		daemonOSType = dockerCLI.ServerInfo().OSType
 	}
+
+	// Buffer to store formatted stats text.
+	// Once formatted, it will be printed in one write to avoid screen flickering.
+	var statsTextBuffer strings.Builder
```
vvoland, line 270: I'd suggest using `bytes.Buffer` here instead, see my comment below
```suggestion
	var statsTextBuffer bytes.Buffer
```

cli/command/container/stats.go
```diff
@@ -264,31 +264,40 @@ func RunStats(ctx context.Context, dockerCLI command.Cli, options *StatsOptions)
...
+			_, _ = fmt.Fprint(&statsTextBuffer, "\033[2J\033[H")
+		}
+
 		if err = statsFormatWrite(statsCtx, ccStats, daemonOSType, !options.NoTrunc); err != nil {
 			break
 		}
+
+		_, _ = fmt.Fprint(dockerCLI.Out(), statsTextBuffer.String())
+
+		statsTextBuffer.Reset()
```
vvoland, line 299: `strings.Builder.Reset` will actually just throw away the whole underlying buffer.
If we'd used `bytes.Buffer` instead, it would just reset the data, but still retain the allocated buffer.

vvoland (commented):
Thank you!

I checked on my system - while it didn't help much with kitty, it makes a difference on slower terminals like iTerm.
That's still an improvement over the previous behavior though!

Overall LGTM, but left some nits.

Giedriusj1:
Thanks for a quick review! I've replaced strings.Builder with bytes.Buffer :)

And yes, the "proper" solution for this would involve keeping track of the two "frames" and only redrawing the parts that have changed. But that would be a lot more involved than what we are doing now.

Curious that you are still seeing the issue with kitty, since for me kitty worked well even before this change. I'm guessing it could be due to GPU drivers, and similar low level stuff, because AFAIK kitty uses GPU for text rendering.

vvoland:
I think kitty is just fast enough to output the characters on the screen as soon as they're read. This could probably be tweaked with kitty settings to decrease the output frequency. But yeah, the proper solution would be to double-buffer the output and only change the actual difference.

However, this is definitely an improvement for most users :)

vvoland:
Thanks! Could you also squash the commits and make sure your commit has a DCO?

Giedriusj1:
> Thanks! Could you also squash the commits and make sure your commit has a DCO?

done!

vvoland (approved):
Thanks, LGTM!

Benehiko (approved):
LGTM

laurazard (approved):
LGTM! Thanks :')

PR #5591, opened 2024-11-04
docs: Correct `run` exit code 126 description

noahsilas:
The command to run inside the container is `/etc`. The semicolon is a statement terminator, which ends the command `docker run busybox /etc`, while `echo $?` prints the exit code of that full docker command.

Having this mistake could confuse someone who thinks that `/etc; echo $?` is all run inside the container, which wouldn't help the reader understand the exit code of the `docker run` command itself.

### DIFF
--- a/docs/reference/run.md
+++ b/docs/reference/run.md
@@ -285,7 +285,7 @@ See 'docker run --help'.
 ### 126
 
 Exit code `126` indicates that the specified contained command can't be invoked.
-The container command in the following example is: `/etc; echo $?`.
+The container command in the following example is: `/etc`.
 
 ```console
 $ docker run busybox /etc; echo $?


thaJeztah:
Thank you!


We should probably consider (but not for this PR) writing them as separate steps, i.e., run a container, then check the exit-code


That said, it looks like with integration with `runc` and later containerd, the error-messages that's now returned for invalid commands have become rather verbose; each layer adding its own decoration, which means that there's now many details that may not be relevant to the user; 

```bash
$ docker run --rm busybox /etc
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/etc": permission denied: unknown.

$ echo $?
126

$ docker run --rm busybox nosuchcommand
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "nosuchcommand": executable file not found in $PATH: unknown.

$ echo $?
127

$ docker run --rm busybox /nosuchcommand
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/nosuchcommand": stat /nosuchcommand: no such file or directory: unknown.

$ echo $?
127
```

thaJeztah (approved):
LGTM

let me bring this one in; I'll open a tracking ticket for the other bits

noahsilas:
Thanks for the quick feedback and merge @thaJeztah! I appreciate the work you are doing here 😁

PR #5594, opened 2024-11-05
golangci-lint: set go version to prevent fallback to go1.17, and fix copyloopvar linting issues

thaJeztah:
GolangCI-lint attempts to deduct the Go version to lint for through the
go version specified in go.mod, which we don't have, and therefore it
falls back to go1.17 semantics:

    level=warning msg="[linters_context] copyloopvar: this linter is disabled because the Go version (1.17) of your project is lower than Go 1.22

**- How to verify it**

```markdown changelog


```

### DIFF
--- a/cli-plugins/manager/cobra.go
+++ b/cli-plugins/manager/cobra.go
@@ -52,7 +52,6 @@ func AddPluginCommandStubs(dockerCli command.Cli, rootCmd *cobra.Command) (err e
 			return
 		}
 		for _, p := range plugins {
-			p := p
 			vendor := p.Vendor
 			if vendor == "" {
 				vendor = "unknown"
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -41,6 +41,9 @@ linters:
     - errcheck
 
 run:
+  # prevent golangci-lint from deducting the go version to lint for through go.mod,
+  # which causes it to fallback to go1.17 semantics.
+  go: "1.23.2"
   timeout: 5m
 
 linters-settings:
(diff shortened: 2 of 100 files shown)


vvoland (approved):

cli-plugins/manager/cobra.go
```diff
@@ -52,7 +52,6 @@ func AddPluginCommandStubs(dockerCli command.Cli, rootCmd *cobra.Command) (err e
 			return
 		}
 		for _, p := range plugins {
-			p := p
```
Benehiko, line 55: just adding this here for future reference: https://go.dev/ref/spec#For_statements

Go 1.22 fixed iterations to use their own variables instead of sharing them. https://go.dev/ref/spec#Go_1.22

Benehiko (approved):

cli-plugins/manager/cobra.go
thaJeztah: Thanks for linking that! Yes, indeed, they fixed that situation. I know they had some back-and-forth on fixing it, because it could be a breaking change, but ultimately they decided that fixing the "footgun" part outweighed the "breaking change" part.

thaJeztah: I realised we were also setting `go:` build-tags; not exactly sure if those would affect the behavior (i.e., if go would downgrade to go1.21 semantics, and thus disable the new behavior), but just in case it would, I opened a PR to update those;

- https://github.com/docker/cli/pull/5608

PR #5595, opened 2024-11-05
docs: update example redis tags from 3.0.x to 7.4.x

dvdksn:
**- What I did**

Updated redis example tags from 3.0.x to latest patch n and n-1 versions.

3.0.5 and 3.0.6 use image format v1 and they're all just old anyway.

- relates to docker/docs#21337

**- How I did it**

`fd -e md -X sed -i 's,3\.0\.5,7.4.0,g'`
`fd -e md -X sed -i 's,3\.0\.[67],7.4.1,g'`

**- How to verify it**

```markdown changelog


```

### DIFF
--- a/docs/reference/commandline/node_ps.md
+++ b/docs/reference/commandline/node_ps.md
@@ -34,11 +34,11 @@ information about available filter options.
 $ docker node ps swarm-manager1
 
 NAME                                IMAGE        NODE            DESIRED STATE  CURRENT STATE
-redis.1.7q92v0nr1hcgts2amcjyqg3pq   redis:3.0.6  swarm-manager1  Running        Running 5 hours
-redis.6.b465edgho06e318egmgjbqo4o   redis:3.0.6  swarm-manager1  Running        Running 29 seconds
-redis.7.bg8c07zzg87di2mufeq51a2qp   redis:3.0.6  swarm-manager1  Running        Running 5 seconds
-redis.9.dkkual96p4bb3s6b10r7coxxt   redis:3.0.6  swarm-manager1  Running        Running 5 seconds
-redis.10.0tgctg8h8cech4w0k0gwrmr23  redis:3.0.6  swarm-manager1  Running        Running 5 seconds
+redis.1.7q92v0nr1hcgts2amcjyqg3pq   redis:7.4.1  swarm-manager1  Running        Running 5 hours
+redis.6.b465edgho06e318egmgjbqo4o   redis:7.4.1  swarm-manager1  Running        Running 29 seconds
+redis.7.bg8c07zzg87di2mufeq51a2qp   redis:7.4.1  swarm-manager1  Running        Running 5 seconds
+redis.9.dkkual96p4bb3s6b10r7coxxt   redis:7.4.1  swarm-manager1  Running        Running 5 seconds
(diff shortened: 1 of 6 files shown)


docs/reference/commandline/node_ps.md
```diff
@@ -93,8 +93,8 @@ The following filter matches tasks with the `usage` label regardless of its valu
 $ docker node ps -f "label=usage"
 
 NAME                               IMAGE        NODE            DESIRED STATE  CURRENT STATE
-redis.6.b465edgho06e318egmgjbqo4o  redis:3.0.6  swarm-manager1  Running        Running 10 minutes
-redis.7.bg8c07zzg87di2mufeq51a2qp  redis:3.0.6  swarm-manager1  Running        Running 9 minutes
+redis.6.b465edgho06e318egmgjbqo4o  redis:7.4.1  swarm-manager1  Running        Running 10 minutes
```
thaJeztah, line 96: Perhaps we should use `7.4` (or even `7`), to prevent these examples getting outdated. Generally, my recommendation would be to pin to either major, or major.minor, as patch releases should always be safe to update (and may contain security updates).

docs/reference/commandline/service_ps.md
```diff
@@ -36,35 +36,35 @@ The following command shows all the tasks that are part of the `redis` service:
...
-ky2re9oz86r9   \_ redis.1  redis:3.0.5  manager1  Shutdown       Shutdown 8 seconds ago
-3s46te2nzl4i  redis.2      redis:3.0.6  worker2   Running        Running less than a second ago
-nvjljf7rmor4   \_ redis.2  redis:3.0.6  worker2   Shutdown       Rejected 23 seconds ago        "No such image: redis@sha256:6…"
-vtiuz2fpc0yb   \_ redis.2  redis:3.0.5  worker2   Shutdown       Shutdown 1 second ago
-jnarweeha8x4  redis.3      redis:3.0.6  worker1   Running        Running 3 seconds ago
-vs448yca2nz4   \_ redis.3  redis:3.0.5  worker1   Shutdown       Shutdown 4 seconds ago
-jf1i992619ir  redis.4      redis:3.0.6  worker1   Running        Running 10 seconds ago
-blkttv7zs8ee   \_ redis.4  redis:3.0.5  worker1   Shutdown       Shutdown 11 seconds ago
+50qe8lfnxaxk  redis.1      redis:7.4.1  manager1  Running        Running 6 seconds ago
+ky2re9oz86r9   \_ redis.1  redis:7.4.0  manager1  Shutdown       Shutdown 8 seconds ago
```
thaJeztah, line 60: I guess for this one, we'd then have to change to 7.3 -> 7.4 🤔

docs/reference/commandline/node_ps.md
dvdksn: I agree in principle, but I wanted to keep the patch version (and character count of this string) to avoid having to worry about whitespace alignment. If we change to 7 or 7.4, we need to fix the whitespace padding on the header row etc.

At some point I would like to replace the redis examples with something open source instead.

thaJeztah: Ah, right, that's fair.

I also left this comment before considering that one of the examples needs to show "updating", and I guess for that it makes more sense to update to a newer patch release (compared to "minor" release).

thaJeztah (approved):
LGTM

PR #5596, opened 2024-11-05
deps: update `go-jose/go-jose` to `v4`

laurazard:
**- What I did**

Updated https://github.com/go-jose/go-jose to `v4`, which has some security/hardening related breaking changes.

**- How I did it**

For jwt verification, `v4` requires specifying which algs we expect to see. The list used here is compiled from other internal sources, and includes all supported algorithms for Auth0 (see [auth0 docs](https://auth0.com/docs/get-started/applications/signing-algorithms)).

**- How to verify it**

Build a binary, then attempt a `docker login`.


cli/internal/oauth/jwt.go
```diff
@@ -66,8 +67,23 @@ func GetClaims(accessToken string) (claims Claims, err error) {
 	return
 }
 
+// allowedSignatureAlgorithms is a list of allowed signature algorithms for JWTs.
+// We add all supported algorithms for Auth0, including with higher key lengths.
+// See auth0 docs: https://auth0.com/docs/get-started/applications/signing-algorithms
+var allowedSignatureAlgorithms = []jose.SignatureAlgorithm{
```
thaJeztah, line 73: Thanks for adding the link here 🫶 


Curious; how did the previous version handle this? Or would it always accept anything it knew? Do we know if this list could be updated by auth0 (and how we would get notified of this?)

thaJeztah: Reading those docs;

```
The most secure practice, and our recommendation, is to use RS256 because:

- With RS256, you are sure that only the holder of the private key (Auth0) can sign tokens, while anyone can check if the token is valid using the public key.
- With RS256, if the private key is compromised, you can implement key rotation without having to re-deploy your application or API with the new secret (which you would have to do if using HS256).
```

I wonder if this is something controlled by Docker Hub (and in which case possibly we could limit the list to what Docker Hub uses)?

laurazard: > I wonder if this is something controlled by Docker Hub (and in which case possibly we could limit the list to what Docker Hub uses)?

Indeed, Hub only uses a subset of these but having this larger list (which includes higher key-length ones and all the ones supported by auth0) means that if some change were to happen on auth0, we'd continue working. As of right now, I don't think we lose anything by accepting that longer list. I'd expect that if a vulnerability was found for one of these we'd find out/be notified about it. Likewise, I'd hope we can figure out/agree on being notified if other teams internally are changing the algorithms used.

laurazard: This is also in `/cli/internal/`, so I think we don't need to be as worried about future support/being able to drop some from the list in the future.

twelsh-aw: Correct. Previously the lib would accept any algorithm (that its code was aware of) and this just adds some forward security in case one of those becomes insecure or able to be abused for a DOS.

Technically just RS256 is used today in hub and our own token signings (/v2/users/login). No need to worry about M2M here, but it too changed to RS256 about a week ago 😅.

Would agree that including all supported auth0 algs is best here. In theory, we should be in control of any changes to that on our tenant but auth0 has changed things on us before without telling...

thaJeztah: Thanks @twelsh-aw ! Sounds like we're good to go then ❤️

thaJeztah (approved):
LGTM, thanks!

PR #5599, opened 2024-11-06
Add option --ipv4

robmry:
**- What I did**

Added `network create` option `--ipv4` ... like `--ipv6`, but it defaults to `true`.

Updated network create docs to include `--ipv4` - and `--fixed-cidr-v6`, `--bip6` ... and tried to make it more obvious that the short options only affect `docker0`.

**- How I did it**

Mostly by copying code for `--ipv6`.

**- How to verify it**

```
# docker network create b4
0d4a8723ec72d2d4b661e60c095865fcd77819459ae0e318505fa1feadbae3de
# docker network create --ipv6 b46
b24ceec262cf8c01e0586ebcea3bd440b52fa0f9b1d40c93f8cec189b7674865
# docker network create --ipv6 --ipv4=false b6
cdfdd74fc7f3b5c421894b0c5b33d48ae8517e8edfcaca6fd40a448e801a189d

# docker network ls --format "table {{.ID}}\t{{.Name}}\t{{.Driver}}\t{{.Scope}}\t{{.IPv4}}\t{{.IPv6}}"
NETWORK ID     NAME      DRIVER    SCOPE     IPV4      IPV6
0d4a8723ec72   b4        bridge    local     true      false
cdfdd74fc7f3   b6        bridge    local     false     true
b24ceec262cf   b46       bridge    local     true      true
be4cf667c3a7   bridge    bridge    local     true      false
8f0b062d1967   host      host      local     false     false
08d4868e22d1   none      null      local     false     false

# docker run --rm -ti --network b6 alpine ip a show dev eth0
17: eth0@if18: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP
    link/ether 86:18:48:8f:f3:3d brd ff:ff:ff:ff:ff:ff
    inet6 fd50:8667:92c::2/64 scope global flags 02
       valid_lft forever preferred_lft forever
    inet6 fe80::8418:48ff:fe8f:f33d/64 scope link tentative
       valid_lft forever preferred_lft forever
```

```markdown changelog
- Added `docker network create` option `--ipv4`.
  To disable IPv4 address assignment for a network, use `docker network create --ipv4=false [...]`.
```

### DIFF
--- a/cli/command/network/create.go
+++ b/cli/command/network/create.go
@@ -24,6 +24,7 @@ type createOptions struct {
 	driverOpts opts.MapOpts
 	labels     opts.ListOpts
 	internal   bool
+	ipv4       *bool
 	ipv6       *bool
 	attachable bool
 	ingress    bool
@@ -42,7 +43,7 @@ type ipamOptions struct {
 }
 
 func newCreateCommand(dockerCLI command.Cli) *cobra.Command {
-	var ipv6 bool
+	var ipv4, ipv6 bool
 	options := createOptions{
 		driverOpts: *opts.NewMapOpts(nil, nil),
 		labels:     opts.NewListOpts(opts.ValidateLabel),
@@ -59,6 +60,9 @@ func newCreateCommand(dockerCLI command.Cli) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.name = args[0]
 
+			if cmd.Flag("ipv4").Changed {
+				options.ipv4 = &ipv4
+			}
 			if cmd.Flag("ipv6").Changed {
 				options.ipv6 = &ipv6
 			}
@@ -73,7 +77,8 @@ func newCreateCommand(dockerCLI command.Cli) *cobra.Command {
 	flags.VarP(&options.driverOpts, "opt", "o", "Set driver specific options")
 	flags.Var(&options.labels, "label", "Set metadata on a network")
 	flags.BoolVar(&options.internal, "internal", false, "Restrict external access to the network")
-	flags.BoolVar(&ipv6, "ipv6", false, "Enable or disable IPv6 networking")
+	flags.BoolVar(&ipv4, "ipv4", true, "Enable or disable IPv4 address assignment")
+	flags.BoolVar(&ipv6, "ipv6", false, "Enable or disable IPv6 address assignment")
 	flags.BoolVar(&options.attachable, "attachable", false, "Enable manual container attachment")
 	flags.SetAnnotation("attachable", "version", []string{"1.25"})
 	flags.BoolVar(&options.ingress, "ingress", false, "Create swarm routing-mesh network")
@@ -113,6 +118,7 @@ func runCreate(ctx context.Context, apiClient client.NetworkAPIClient, output io
 		Options:    options.driverOpts.GetAll(),
 		IPAM:       ipamCfg,
 		Internal:   options.internal,
+		EnableIPv4: options.ipv4,
 		EnableIPv6: options.ipv6,
 		Attachable: options.attachable,
 		Ingress:    options.ingress,
--- a/docs/reference/commandline/network_create.md
+++ b/docs/reference/commandline/network_create.md
@@ -18,7 +18,8 @@ Create a network
 | `--ip-range`              | `stringSlice` |           | Allocate container ip from a sub-range                  |
 | `--ipam-driver`           | `string`      | `default` | IP Address Management Driver                            |
 | `--ipam-opt`              | `map`         | `map[]`   | Set IPAM driver specific options                        |
-| `--ipv6`                  | `bool`        |           | Enable or disable IPv6 networking                       |
+| `--ipv4`                  | `bool`        | `true`    | Enable or disable IPv4 address assignment               |
+| `--ipv6`                  | `bool`        |           | Enable or disable IPv6 address assignment               |
 | `--label`                 | `list`        |           | Set metadata on a network                               |
 | `-o`, `--opt`             | `map`         | `map[]`   | Set driver specific options                             |
 | `--scope`                 | `string`      |           | Control the network's scope                             |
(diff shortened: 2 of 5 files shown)


robmry:
Some apparently-unrelated test failures ...

```
=== Failed
=== FAIL: e2e/container TestProcessTermination (11.00s)
    run_test.go:280: assertion failed:
        Command:  docker run --rm -i registry:5000/alpine:frozen sh -c echo 'starting trap'; trap 'echo got signal; exit 0;' TERM; while true; do sleep 10; done
        ExitCode: 0 (timeout)
        Stdout:   starting trap
        got signal

        Stderr:

        Failures:
        Expected command to finish, but it hit the timeout
```
```
#20 62.19 === FAIL: cli-plugins/socket TestConnectAndWait/connect_goroutine_exits_after_EOF (0.00s)
#20 62.19     socket_test.go:190: assertion failed: 7 (int) != 6 (numGoroutines + 1 int)
```
```
#20 61.97 === Failed
#20 61.97 === FAIL: cli/command/container TestRunAttachTermination (unknown)
#20 61.97 signal: interrupt
```

docs/reference/commandline/network_create.md
```diff
@@ -148,30 +148,30 @@ fails and Docker Engine returns an error.
...
+| `com.docker.network.bridge.name`                 | -                           | Bridge name to be used when creating the Linux bridge |
+| `com.docker.network.bridge.enable_ip_masquerade` | `--ip-masq`                 | Enable IP masquerading                                |
+| `com.docker.network.bridge.enable_icc`           | `--icc`                     | Enable or Disable Inter Container Connectivity        |
+| `com.docker.network.bridge.host_binding_ipv4`    | `--ip`                      | Default IP when binding container ports               |
+| `com.docker.network.driver.mtu`                  | `--mtu`                     | Set the containers network MTU                        |
+| `com.docker.network.container_iface_prefix`      | -                           | Set a custom prefix for container interfaces          |
 
 The following arguments can be passed to `docker network create` for any
 network driver, again with their approximate equivalents to Docker daemon
 flags used for the docker0 bridge:
```
akerouanton, line 167:
```suggestion
flags used for the `docker0` bridge:
```
(unrelated to this PR)

docs/reference/commandline/network_create.md
```diff
@@ -148,30 +148,30 @@ fails and Docker Engine returns an error.
...
 
-| Argument     | Equivalent                        | Description                                |
-|--------------|-----------------------------------|--------------------------------------------|
-| `--gateway`  | -                                 | IPv4 or IPv6 Gateway for the master subnet |
-| `--ip-range` | `--fixed-cidr`, `--fixed-cidr-v6` | Allocate IPs from a range                  |
-| `--internal` | -                                 | Restrict external access to the network    |
-| `--ipv6`     | `--ipv6`                          | Enable or disable IPv6 networking          |
```
akerouanton: Did you remove this line by mistake, or on purpose? The daemon has an `--ipv6` flag that controls whether the default bridge is IPv6-capable, so we should keep this line.

cli/command/network/create.go
```diff
@@ -73,7 +77,8 @@ func newCreateCommand(dockerCLI command.Cli) *cobra.Command {
 	flags.VarP(&options.driverOpts, "opt", "o", "Set driver specific options")
 	flags.Var(&options.labels, "label", "Set metadata on a network")
 	flags.BoolVar(&options.internal, "internal", false, "Restrict external access to the network")
-	flags.BoolVar(&ipv6, "ipv6", false, "Enable or disable IPv6 networking")
+	flags.BoolVar(&ipv4, "ipv4", true, "Enable or disable IPv4 address assignment")
+	flags.BoolVar(&ipv6, "ipv6", false, "Enable or disable IPv6 address assignment")
```
akerouanton, line 81: :+1: for the revised help text.

docs/reference/commandline/network_create.md
robmry: Oh, thank you ... by mistake while juggling commits (it came back in the next commit). Should be sorted now.

akerouanton (approved):
LGTM

cli/command/network/create.go
```diff
@@ -73,7 +77,8 @@ func newCreateCommand(dockerCLI command.Cli) *cobra.Command {
 	flags.VarP(&options.driverOpts, "opt", "o", "Set driver specific options")
 	flags.Var(&options.labels, "label", "Set metadata on a network")
 	flags.BoolVar(&options.internal, "internal", false, "Restrict external access to the network")
-	flags.BoolVar(&ipv6, "ipv6", false, "Enable or disable IPv6 networking")
+	flags.BoolVar(&ipv4, "ipv4", true, "Enable or disable IPv4 address assignment")
```
vvoland, line 80:
```suggestion
	flags.BoolVar(&ipv4, "ipv4", true, "Enable or disable IPv4 address assignment")
```

Wondering if we should force this to `true` on the CLI side.
The `EnableIPv4` is `*bool` on the API side, so we could just keep it as missing/`null` if user haven't passed this option.
For example if in future we'd need to distinguish between "user explicitly doesn't want ipv4" or "user doesn't care" on the daemon side.

Does that make sense?

thaJeztah:
Related to this PR: does the compose-spec need an update to allow setting these? I'm not sure if it currently does, and we probably would need to keep parity between compose and docker CLI creating networks.


Not for this PR, because it's an existing situation: we need to start thinking of a strategy for defaults. We currently try to reflect defaults on the client side by showing `true` / `false` as default. Those defaults may be currently true, but could change over time (especially looking at IPv6, which may become a default at some point?). I don't have a clear-cut answer to this; we could fetch defaults from the daemon, or (much larger change) if we want things to be declarative, possibly have some `fetch default -> patch -> send` approach. Not a worry for now, but we may need to start having some of that in the back of our minds to start thinking about.

cli/command/network/create.go
thaJeztah: LOL, I just wrote a blurb below about "defaults".

vvoland: Oh right, I missed that!

vvoland (approved):

robmry:
> Related to this PR: does the compose-spec need an update to allow setting these? I'm not sure if it currently does, and we probably would need to keep parity between compose and docker CLI creating networks.

Yes, we should do ... it's work-around-able in Compose by setting the equivalent driver option, not pretty though. We need to add `gw-priority` anyway, so we can pick up the two together.

PR #5607, opened 2024-11-11
Docs: host-gateway-ip daemon option IPv4+IPv6

robmry:
**- What I did**

Depends on:
- https://github.com/moby/moby/pull/48807

The host-gateway-ip daemon option now accepts two addresses, one IPv4 and one IPv6.

**- How I did it**

**- How to verify it**

```markdown changelog
n/a
```

### DIFF
--- a/docs/reference/dockerd.md
+++ b/docs/reference/dockerd.md
@@ -63,8 +63,8 @@ Options:
   -G, --group string                          Group for the unix socket (default "docker")
       --help                                  Print usage
   -H, --host list                             Daemon socket(s) to connect to
-      --host-gateway-ip ip                    IP address that the special 'host-gateway' string in --add-host resolves to.
-                                              Defaults to the IP address of the default bridge
+      --host-gateway-ip list                  IP addresses that the special 'host-gateway' string in --add-host resolves to.
+                                              Defaults to the IP addresses of the default bridge
       --http-proxy string                     HTTP proxy URL to use for outgoing traffic
       --https-proxy string                    HTTPS proxy URL to use for outgoing traffic
       --icc                                   Enable inter-container communication (default true)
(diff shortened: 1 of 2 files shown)


docs/reference/dockerd.md
```diff
@@ -62,8 +62,7 @@ Options:
   -G, --group string                          Group for the unix socket (default "docker")
       --help                                  Print usage
   -H, --host list                             Daemon socket(s) to connect to
-      --host-gateway-ip ip                    IP address that the special 'host-gateway' string in --add-host resolves to.
-                                              Defaults to the IP address of the default bridge
+      --host-gateway-ip list                  IP addresses that the special 'host-gateway' string in --add-host resolves to. Defaults to the IP addresses of the default bridge
```
thaJeztah, line 65: Oh! Can you manually wrap / format this one? This page is still manually maintained and this block rendered as a code-block, so wrapping reduces horizontal scrolling; see https://docs.docker.com/reference/cli/dockerd/

robmry: Will do ... I did that at first, then thought maybe it was supposed to exactly match the `--help` output. Doh!

`host-gateway-ip` isn't a new option, but it's not in the manpage in this repo or moby ... I'll add it.

thaJeztah: Yeah we really need to;

- update our packaging to use the man-page from moby/moby instead (and remove the man page from this repo)
- look at moving the dockerd reference docs to moby/moby, and use the "docs-generator" to generate these flag-descriptions and the "YAML" docs so that we can use the same formatting as the other reference docs.

docs/reference/dockerd.md
```diff
@@ -63,8 +63,8 @@ Options:
   -G, --group string                          Group for the unix socket (default "docker")
       --help                                  Print usage
   -H, --host list                             Daemon socket(s) to connect to
-      --host-gateway-ip ip                    IP address that the special 'host-gateway' string in --add-host resolves to.
-                                              Defaults to the IP address of the default bridge
+      --host-gateway-ip list                  IP addresses that the special 'host-gateway' string in --add-host resolves to.
```
thaJeztah, line 66: I hate Cobra for this; we should probably look if we can make it show a custom type. "list" is ... not very informative. The  ugly bit is that Cobra doesn't allow you to set it per flag, but it's really tied to the type you're using 😞 (perhaps we can come with some trickery to initialise it for a specific flag 🤔)

thaJeztah (approved):
LGTM

docs/reference/dockerd.md
thaJeztah: Yeah, it's not really a new problem, but something we should look at across the board. Things like `string` are pretty useless to the user, and ideally these would show `<some type that can be documented>`. Similar to, e.g. `man wget` would show more specific types (ISTR I found better examples, but couldn't remember where);

```console
       -e command
       --execute command
           Execute command as if it were a part of .wgetrc.  A command thus invoked will be executed after the commands in .wgetrc, thus
           taking precedence over them.  If you need to specify more than one wgetrc command, use multiple instances of -e.

   Logging and Input File Options
       -o logfile
       --output-file=logfile
           Log all messages to logfile.  The messages are normally reported to standard error.

       -a logfile
       --append-output=logfile
           Append to logfile.  This is the same as -o, only it appends to logfile instead of overwriting the old log file.  If logfile
           does not exist, a new file is created.
```


If such types are somewhat standardised in our codebase, that would also allow them to be documented, e.g.

```
--platform PLATFORM

--gpus gpu-request
--health-interval duration
```

Then we could even provide links to those formats in our docs (i.e., here's where the format is described for `platform` or `ip` or `duration`).

Ideally, we'd also come with a format that can indicate "can be specified multiple times" (`[]platform` ? other options?)


There's some fun bug/issue still to fix for that though, because Cobra currently requires "multiple times" to be suffixed by `List` or `Array` to make auto-completion work properly 😞

PR #5614, opened 2024-11-13
[27.x backport] Completion for events --filter

thaJeztah:
- backport https://github.com/docker/cli/pull/5538

----

Adds completion for `docker events --filter`.

There are different types of filters, requiring different completion logic.
Some filters have static values, others require API lookups.
In some cases, trailing spaces of intermediate completions have to be suppressed.

The completion logic is complex because cobra does not offer support for compound flag values.
The handler for the `--filter` flag serves as the central entrypoint to all completions.

How to verify:
in a dev container, run `make completion`, then trigger completions, e.g.
```shell
$ docker events --filter <TAB><TAB>
container=  event=      label=      node=       type=       
daemon=     image=      network=    scope=      volume= 
$ docker events --filter event=c<TAB><TAB>
checkpoint  commit      connect     copy        create
$ docker events --filter image=<TAB><TAB>
docker-cli-dev:latest  docker-cli-e2e:latest  registry:2
```

The completions should correspond to those of the legacy bash completion.

### DIFF
--- a/cli/command/system/client_test.go
+++ b/cli/command/system/client_test.go
@@ -7,7 +7,11 @@ import (
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/events"
 	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/image"
 	"github.com/docker/docker/api/types/network"
+	"github.com/docker/docker/api/types/swarm"
+	"github.com/docker/docker/api/types/system"
+	"github.com/docker/docker/api/types/volume"
 	"github.com/docker/docker/client"
 )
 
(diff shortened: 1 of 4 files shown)


laurazard:
It's a small nit, but I wonder if we should squash your changes adjusting the types into the original commit that introduced them – otherwise if we have to revert them or do anything we need to remember to move them both together because one won't build without the other.

WDYT @thaJeztah?

thaJeztah:
Ah, yes, I'm good with squashing; with these I'm always leaning two ways; "is it clearer to have a separate commit showing _this is what we modified_" or (also for bisecting) "this is the backport (but it's modified)".

Let me squash the commits, as I think that's what we usually did.

thaJeztah:
Squashed! ❤️

laurazard (approved):

laurazard:
Looks great, thanks!

PR #5625, opened 2024-11-16
Optimise `docker stats` to not require clearing the whole screen

Giedriusj1:
Instead of clearing the whole screen and then writing the new stats, we now write the new stats on top of the old text, and then clear the remaining text.

This is a more efficient way to update the stats, as it avoids the flickering that happens when the screen is cleared and rewritten.

This improves upon https://github.com/docker/cli/pull/5586


laurazard (approved):
LGTM!

I'd like to add some tests/otherwise refactor this area, but that's orthogonal to this PR and this looks safe

Benehiko (approved):

thaJeztah (approved):
LGTM

PR #5627, opened 2024-11-16
cli/connhelper: support overriding the docker binary over SSH

### LINKED ISSUE #5626: Remove hardcoded `docker` CLI command name for SSH hosts
## Description

[Currently, when connecting to a remote SSH host, the `docker` binary name is hard-coded.](https://github.com/docker/cli/blob/9861ce90fd6b8ddca19db5f803dcbef9a583e9e1/cli/connhelper/connhelper.go#L52). This means that if the remote host doesn't have a binary named `docker` in the path, remote SSH support doesn't work.

There are two specific use-cases this makes difficult:

- Systems where `docker` isn't in the path for the SSH connection, and
- Systems where the `docker` binary has another name.

If this hard-coded default could be overridden, these use-cases could be supported.

## Proposal

Introduce the `DOCKER_SSH_REMOTE_BINARY` environment variable, which will be used in place of the bare `docker` binary if set. I'll raise a PR for this change then link to it. 

## Related

https://github.com/docker/cli/issues/3045

mattmacleod:
This change adds the ability to override the docker binary used when executing commands over SSH. This is useful when the docker binary is not in the PATH of the SSH user, or when the binary is not named `docker`.

If `DOCKER_SSH_REMOTE_BINARY` is set in the environment, the value will be used as the docker binary when executing commands over SSH. If the environment variable is not set, the default value of `docker` will be used.

Closes #5626 

```markdown changelog
Support the `DOCKER_SSH_REMOTE_BINARY` environment variable for overriding the `docker` binary used on remote SSH hosts.
```

IMG_1013


cli/connhelper/connhelper.go
```diff
@@ -49,7 +56,7 @@ func getConnectionHelper(daemonURL string, sshFlags []string) (*ConnectionHelper
 		sshFlags = disablePseudoTerminalAllocation(sshFlags)
 		return &ConnectionHelper{
 			Dialer: func(ctx context.Context, network, addr string) (net.Conn, error) {
-				args := []string{"docker"}
+				args := []string{dockerSSHRemoteBinary()}
```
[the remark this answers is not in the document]
laurazard: something like
```go
remoteDockerBinary := dockerSSHRemoteBinary()
return &ConnectionHelper{
				args := []string{remoteDockerBinary}
...
```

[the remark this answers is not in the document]
mattmacleod: Great point – thanks @laurazard! I've updated as suggested.

cli/connhelper/connhelper_test.go
```diff
@@ -63,3 +63,32 @@ func TestDisablePseudoTerminalAllocation(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv(DockerSSHRemoteBinaryEnv, tc.env)
+			result := dockerSSHRemoteBinary()
+			if result != tc.expected {
+				t.Errorf("expected %q, got %q", tc.expected, result)
+			}
+		})
+	}
+}
```
vvoland, line 94: To be honest, the function tested here is so trivial, that it's in fact much simpler than the test code.
I don't think we really need it, but since it's already there, I guess it doesn't hurt either 🤷🏻

cli/connhelper/connhelper_test.go
```diff
@@ -63,3 +63,32 @@ func TestDisablePseudoTerminalAllocation(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv(DockerSSHRemoteBinaryEnv, tc.env)
+			result := dockerSSHRemoteBinary()
+			if result != tc.expected {
+				t.Errorf("expected %q, got %q", tc.expected, result)
+			}
```
vvoland, line 91:
```suggestion
			assert.Check(t, is.Equal(result, tc.expected))
```

cli/connhelper/connhelper.go
```diff
@@ -91,3 +99,15 @@ func disablePseudoTerminalAllocation(sshFlags []string) []string {
 	}
 	return append(sshFlags, "-T")
 }
+
+// dockerSSHRemoteBinary returns the binary to use when executing Docker
+// commands over SSH. It defaults to "docker" if the DOCKER_SSH_REMOTE_BINARY
+// environment variable is not set.
+func dockerSSHRemoteBinary() string {
+	value := os.Getenv(DockerSSHRemoteBinaryEnv)
```
vvoland, line 107: It would probably be better if this was a per-context setting rather than a global env-variable.
However, it's a niche use-case, so I guess it's _okay_ as an escape hatch and not really a blocker.

@thaJeztah @laurazard WDYT?

thaJeztah:
I'm honestly a bit on the fence on this change, but curious what exact issues this is resolving; from the ticket;


> - Systems where docker isn't in the path for the SSH connection, and
> - Systems where the docker binary has another name.

- For the first one; I'm curious in what situations the `docker` binary would not be in `PATH`; that seems like a broken setup, Some of this could (depending on the situation) be addressed through ssh config (`SetEnv`) if the host allows it.
- For the second one; do you have a specific example? This feature was designed to work with docker, and the requirement for it to work is for the `docker` CLI to be installed on the remote host; I'm wondering if there's a good use-case for using any other random binary for this purpose. If the remote has something installed trying to be `docker` compatible, it could of course provide an alias.

Happy to hear if there's more details to this requirement.

mattmacleod:
Hey @thaJeztah! I think this is honestly probably pretty limited in terms of applicability, and I've only raised it due to a pretty specific set of circumstances – to support remote ops to machines running [BalenaOS](https://www.balena.io/os).

These devices are running [balenaEngine](https://github.com/balena-os/balena-engine) rather than Docker – it's a Moby-derived container engine. The upshot of all this is that the root filesystem is read-only, and the container engine CLI command is named `balena` instead of `docker`.

Since the root filesystem is readonly by design, I've got limited options in terms of things like aliases or changes that would otherwise allow this to be implemented.

We've obviously in my team got the option of building our own CLI release instead to support this pretty narrow use-case, but I figured it was generally a light-touch change that might benefit some other people too.

doisyg:
Any chance to get this PR moving ?

Issue #5643, opened 2024-11-22
Permission issue on /usr/local/libexec prevents using plugins from /usr/libexec


mihalyr:
### Description

I think there is a bug in how docker cli resolves the plugin paths. Described also in https://discussion.fedoraproject.org/t/docker-plugins-on-f41-not-working/138199/9

Given a directory `/usr/libexec/docker/cli-plugins` with correct permissions (root:root 755) and a plugin `/usr/libexec/docker/cli-plugins/docker-buildx`, if we create a directory `/usr/local/libexec` with permissions (root:root 750), running `docker build` fails as it won't find `buildx`. 

I suspect that as docker-cli checks for plugins in the directories in the precedence order, if any of the directories is not accessible, the plugin search fails and directories specified later in the precedence order will not be checked and thus plugins in other folder won't be found. 

The workaround is easy to just fix the permissions, so not a big problem. But I would expect docker-cli to skip such directory as it does with non-existent directories. 

### Reproduce

1. Given docker plugin `/usr/libexec/docker/cli-plugins/docker-buildx`
2. Remove non-root access from a higher precedence plugin folder `chmod 750 /usr/local/libexec`
3. Try to run the plugin via docker-cli `docker buildx`
4. It fails with `docker: 'buildx' is not a docker command.`

### Expected behavior

`docker buildx` works, because the plugin under `/usr/libexec/docker/cli-plugins` is still accessible. 

### docker version

```bash
Client:
 Version:           27.3.1
 API version:       1.41 (downgraded from 1.47)
 Go version:        go1.23.1
 Git commit:        2.fc41
 Built:             Tue Sep 24 00:00:00 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: linux/amd64/fedora-41
 Podman Engine:
  Version:          5.2.5
  APIVersion:       5.2.5
  Arch:             amd64
  BuildTime:        2024-10-18T02:00:00+02:00
  Experimental:     false
  GitCommit:        
  GoVersion:        go1.23.2
  KernelVersion:    6.11.8-300.fc41.x86_64
  MinAPIVersion:    4.0.0
  Os:               linux
 Conmon:
  Version:          conmon version 2.1.12, commit: 
  Package:          conmon-2.1.12-3.fc41.x86_64
 OCI Runtime (crun):
  Version:          crun version 1.18.1
commit: c41f034fdbb9742c395085fc98459c94ad1f9aae
rundir: /run/user/3000/crun
spec: 1.0.0
[...]

PR #5651, opened 2024-11-28
cli-plugins: Fix searching inaccessible directories

vvoland:
- fixes: https://github.com/docker/cli/issues/5643

Fix a case where one inaccessible plugin search path stops the whole search and prevents latter paths from being scanned.

Remove a preliminary `Stat` call that verifies whether path is an actual directory and is accessible.
It's unneeded and doesn't actually check whether the directory can be listed or not.
`os.ReadDir` will fail in such case anyway, so just attempt to do that and ignore any encountered error, instead of erroring out the whole plugin candidate listing.

**- What I did**

**- How I did it**

**- How to verify it**
TestListPluginCandidatesInaccesibleDir

```markdown changelog
Fix inaccessible plugins paths preventing plugins from being detected.
```

### DIFF
--- a/cli-plugins/manager/manager.go
+++ b/cli-plugins/manager/manager.go
@@ -75,12 +75,12 @@ func getPluginDirs(cfg *configfile.ConfigFile) ([]string, error) {
 	return pluginDirs, nil
 }
 
-func addPluginCandidatesFromDir(res map[string][]string, d string) error {
+func addPluginCandidatesFromDir(res map[string][]string, d string) {
 	dentries, err := os.ReadDir(d)
 	// Silently ignore any directories which we cannot list (e.g. due to
 	// permissions or anything else) or which is not a directory
 	if err != nil {
-		return nil
+		return
 	}
 	for _, dentry := range dentries {
 		switch dentry.Type() & os.ModeType {
@@ -101,22 +101,15 @@ func addPluginCandidatesFromDir(res map[string][]string, d string) error {
 		}
 		res[name] = append(res[name], filepath.Join(d, dentry.Name()))
 	}
-	return nil
 }
 
 // listPluginCandidates returns a map from plugin name to the list of (unvalidated) Candidates. The list is in descending order of priority.
-func listPluginCandidates(dirs []string) (map[string][]string, error) {
+func listPluginCandidates(dirs []string) map[string][]string {
 	result := make(map[string][]string)
 	for _, d := range dirs {
-		if err := addPluginCandidatesFromDir(result, d); err != nil {
-			// Silently ignore paths which don't exist.
-			if os.IsNotExist(err) {
-				continue
-			}
-			return nil, err // Or return partial result?
-		}
+		addPluginCandidatesFromDir(result, d)
 	}
-	return result, nil
+	return result
 }
 
 // GetPlugin returns a plugin on the system by its name
@@ -126,11 +119,7 @@ func GetPlugin(name string, dockerCli command.Cli, rootcmd *cobra.Command) (*Plu
 		return nil, err
 	}
 
-	candidates, err := listPluginCandidates(pluginDirs)
-	if err != nil {
-		return nil, err
-	}
-
+	candidates := listPluginCandidates(pluginDirs)
 	if paths, ok := candidates[name]; ok {
 		if len(paths) == 0 {
 			return nil, errPluginNotFound(name)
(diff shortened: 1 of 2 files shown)


cli-plugins/manager/manager.go
```diff
@@ -75,10 +75,12 @@ func getPluginDirs(cfg *configfile.ConfigFile) ([]string, error) {
 	return pluginDirs, nil
 }
 
-func addPluginCandidatesFromDir(res map[string][]string, d string) error {
+func addPluginCandidatesFromDir(res map[string][]string, d string) {
 	dentries, err := os.ReadDir(d)
+	// Silently ignore any directories which we cannot list (e.g. due to
+	// permissions or anything else) or which is not a directory
```
thaJeztah, line 81: The only thing I was considering (but perhaps it's not a real issue) is if we need to distinguish default ("try -> try next") paths and path(s) that are explicitly configured in  `~/.docker/config.json` as  "additional  paths" (`cliPluginsExtraDirs`).


My train of thought there is that if it's an explicit configuration, then failing to traverse the location (perhaps except for "not exist") could be considered a hard failure.

Happy to hear thoughts on that though!

vvoland: The attempt to use a plugin which isn't accessible will result in a hard failure anyway. Erroring out during plugin scan would prevent the CLI from running in the non-plugin usage.

thaJeztah: Yeah, perhaps I'm _looking_ for issues that aren't important, but mostly considering if it should be completely _silent_. Here's the order of preference in which paths are considered;
https://github.com/docker/cli/blob/41fba28661ddfa60a9bf8e44ed2c6b5748eee175/cli-plugins/manager/manager.go#L52-L62


Based on that, the custom path is intended to have a higher priority than the system-wide installation paths. Which could mean "the cli is configured to use these overridden versions". On Docker Desktop, this could be some of the additional plugins shipping with it, but perhaps it's a situation where (say) the system-wide version is provided by the distro you're running and has a vulnerability, so the intent is to use a fixed version through one of the higher-priority paths. If we silently ignore that we aren't able to use that path, it means we're silently falling back to either an older (perhaps vulnerable) version, or missing extra plugins that were supposed to be available.

For the system-wide paths, perhaps that's OK (it's a system-wide path, and current user isn't in the right group to use it), so even for those, I somewhat wonder if we should have some warning. Not sure if anything documents expected permissions though, or at least I couldn't find that in the [FHS (Filesystem Hierarchy Standard)](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrlibexec) docs.

Quick check on some machines showed me that they are accessible;

```bash
# CentOS, Fedora machine

ls -ld /usr/libexec /usr/libexec/docker /usr/libexec/docker/cli-plugins /usr/local/libexec /usr/local/libexec/docker /usr/local/libexec/docker/cli-plugins
ls: cannot access '/usr/local/libexec/docker': No such file or directory
ls: cannot access '/usr/local/libexec/docker/cli-plugins': No such file or directory
drwxr-xr-x. 26 root root 4096 Nov 25 10:52 /usr/libexec
drwxr-xr-x.  3 root root   44 Nov 25 10:53 /usr/libexec/docker
drwxr-xr-x.  2 root root   49 Nov 25 10:52 /usr/libexec/docker/cli-plugins
drwxr-xr-x.  2 root root    6 Aug  9  2021 /usr/local/libexec

# Ubuntu machine
ls -ld /usr/libexec /usr/libexec/docker /usr/libexec/docker/cli-plugins /usr/local/libexec /usr/local/libexec/docker /usr/local/libexec/docker/cli-plugins
ls: cannot access '/usr/local/libexec': No such file or directory
ls: cannot access '/usr/local/libexec/docker': No such file or directory
ls: cannot access '/usr/local/libexec/docker/cli-plugins': No such file or directory
drwxr-xr-x 12 root root 4096 Nov 19 16:02 /usr/libexec
drwxr-xr-x  3 root root 4096 Nov 19 16:02 /usr/libexec/docker
drwxr-xr-x  2 root root 4096 Nov 19 16:02 /usr/libexec/docker/cli-plugins
```

thaJeztah (approved):
LGTM


let's discuss if we want to surface these errors as a warning for a follow-up (perhaps we could do as part of the `getPluginDirs` function, as that would be the place where we collect paths to consider).

thaJeztah:
Opened a ticket for discussion;

-  https://github.com/docker/cli/issues/5656

laurazard (approved):
LGTM

Since this is quite a bit of a special case, I think we can fix it now and discuss what we think the behaviour should be later (re: should it be an error?).

thaJeztah:
Yup; let's bring this one in for now 👍

PR #5652, opened 2024-11-28
[27.x backport] cli-plugins: Fix searching inaccessible directories

vvoland:
- backport: https://github.com/docker/cli/pull/5651

- fixes: https://github.com/docker/cli/issues/5643

Fix a case where one inaccessible plugin search path stops the whole search and prevents latter paths from being scanned.

Remove a preliminary `Stat` call that verifies whether path is an actual directory and is accessible.
It's unneeded and doesn't actually check whether the directory can be listed or not.
`os.ReadDir` will fail in such case anyway, so just attempt to do that and ignore any encountered error, instead of erroring out the whole plugin candidate listing.

**- What I did**

**- How I did it**

**- How to verify it**
TestListPluginCandidatesInaccesibleDir

```markdown changelog
Fix inaccessible plugins paths preventing plugins from being detected.
```

### DIFF
--- a/cli-plugins/manager/manager.go
+++ b/cli-plugins/manager/manager.go
@@ -77,8 +77,10 @@ func getPluginDirs(cfg *configfile.ConfigFile) ([]string, error) {
 
 func addPluginCandidatesFromDir(res map[string][]string, d string) error {
 	dentries, err := os.ReadDir(d)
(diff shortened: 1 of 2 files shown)


thaJeztah (approved):
LGTM

Issue #5639, opened 2024-11-21
Pulling image with `docker run` doesn't cancel gracefully


vvoland:
### Description

Image pull initiated by `docker run` doesn't handle the SIGTERM cancellation gracefully.
This isn't the case with `docker pull`.

### Reproduce

```
docker run --rm <image-you-dont-have>
CTRL+C
```

https://github.com/user-attachments/assets/c3716f41-d99c-44dc-a242-b02f0efade85

### docker version

```bash
Client:
 Version:           27.4.0-rc.2
 API version:       1.47
 Go version:        go1.22.9
 Git commit:        af56ef5
 Built:             Tue Nov 19 13:36:00 2024
 OS/Arch:           darwin/arm64
 Context:           desktop-linux

Server: Docker Desktop 4.37.0 (1)
 Engine:
  Version:          27.4.0-rc.2
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.22.9
  Git commit:       a080c3f
  Built:            Tue Nov 19 13:38:47 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.7.21
  GitCommit:        472731909fa34bd7bc9c087e4c27943f9835f111
 runc:
  Version:          1.1.13
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### docker info

```bash
-
```

PR #5654, opened 2024-11-29
[27.x backport] fix: ctx on run image pull

Benehiko:
- backport https://github.com/docker/cli/pull/5645
- Fixes https://github.com/docker/cli/issues/5639

After some digging I found that we don't actually explicitly return an error on the `pullImage` function [here](https://github.com/docker/cli/blob/master/cli/command/container/create.go#L132-L152). Once the context is cancelled the response is closed since the API client uses `http.NewRequestWithContext` which then exits the `jsonmessage.DisplayJSONMessagesToStream(responseBody, out, nil)` without any errors. 

The following API request to `ContainerCreate` then immediately fails since the context is closed, which then returns an error.

See this section of the code, it fails to find the image, then pulls the image, and continues to retry creating a container - on this second retry it only then returns with the context error.
https://github.com/docker/cli/blob/master/cli/command/container/create.go#L271-L289

Writing a test for this means that we are building into the test an assumption that we won't error on the `pullImage` function, but rather on the second call to `ContainerCreate`. I'll push what I have, but I honestly think that we will get tripped up by the `pullImage` function again since it's not context aware - actually more so the `jsonmessage` package it is using is actually the root cause found in Moby [here](https://github.com/moby/moby/blob/master/pkg/jsonmessage/jsonmessage.go#L228). It's a bit of a hornets nest since it gets used in many places and by refactoring/improving this it could cause edge case bugs.

**- What I did**
Delay stripping the `cancel` from the function `ctx` so we can still cancel the `createContainer` function. https://github.com/docker/cli/blob/e4e0ea2c1a8f7aa695fee0d77acc8b760462ce00/cli/command/container/run.go#L142

**- How I did it**

**- How to verify it**
Unit test

```markdown changelog
Fixed bug preventing image pulls from being cancelled during `docker run`.
```

### DIFF
--- a/cli/command/container/client_test.go
+++ b/cli/command/container/client_test.go
@@ -25,7 +25,7 @@ type fakeClient struct {
 		platform *specs.Platform,
 		containerName string) (container.CreateResponse, error)
 	containerStartFunc      func(containerID string, options container.StartOptions) error
-	imageCreateFunc         func(parentReference string, options image.CreateOptions) (io.ReadCloser, error)
+	imageCreateFunc         func(ctx context.Context, parentReference string, options image.CreateOptions) (io.ReadCloser, error)
 	infoFunc                func() (system.Info, error)
 	containerStatPathFunc   func(containerID, path string) (container.PathStat, error)
 	containerCopyFromFunc   func(containerID, srcPath string) (io.ReadCloser, container.PathStat, error)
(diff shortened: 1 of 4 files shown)


thaJeztah:
not a blocker (but if you want to re-apply, let me know); we try to use the `-x` option when cherry-picking, so that a reference to the original commit is added in the commit message, e.g.;


```bash
git commit -s -S -x 30a73ff19c407eee75de489355154ce1f35231a8
```

The `x` option adds an extra line to the commit message, which can be useful to correlate changes and/or track-back origin;

```
(cherry picked from commit 30a73ff19c407eee75de489355154ce1f35231a8)
```

thaJeztah (approved):
LGTM (but left a suggestion, let me know if you want to fix up the cherry-pick)

Benehiko:
@thaJeztah updated the cherry-pick with your suggestions

thaJeztah (approved):
LGTM

thanks! Let me bring this one in 👍

PR #5662, opened 2024-12-02
run: correctly handle only STDIN attached

laurazard:
**- What I did**

Overall make the `docker run -a stdin` handling more consistent + some assorted cleanups.

The most significant changes are making the "print container ID and immediately exit if detached" check more consistent (also account for `-a stdin` instead of just `stdout` and `stderr`), and making `docker run -a stdin` exit if the container exits/is removed.

**- How I did it**

aee9eebf349dd6d8634e4ea7d5624e276c6da0a2:

During a `docker run`, the CLI has some different behavior/output depending on whether the run is "detached" or not.

In some cases, the CLI is checking whether either `stdin`, `stdout` or `stderr` are attached, but in other cases we're only checking `stdout` and `stderr`, which leads to some inconsistencies:

```
$ docker run -a stdout --rm --name test alpine top
[docker kill test]
exit status 137

$ docker run -a stderr --rm --name test alpine top
[docker kill test]
exit status 137

$ docker run -a stdin --rm --name test alpine top
56820d94a89b96889478241ae68920323332c6d4cf9b51ba9340cba01e9e0565
[docker kill test]
[no exit code]
```

Since we're not checking for whether `stdin` is attached when deciding whether to early exit without receiving on `statusChan`, the `docker run -a stdin` is falling into the "detached mode" logic, which simply prints the container ID and doesn't print/return the exit code.

This patch makes the "attached" checks consistent.

8431298824ef57e51d1b5c0a7af85cb17fd7a3e8: small cleanup

446f36ce58c2c67dd727e8ec7963218e85a36ed5: Since everything else after the `apiClient.ContainerStart` block is under an `if attach` conditional, we can move the "not attached" early exit up.

a3d9fc49414b5e6972e589d2baa8df676a984c9b:

Now, if running in "detached" mode, we early exit at L222.

Similarly, if `attachContainer` errors out, it returns an error that gets handled on L190.

As such, `errCh` can never be nil on L231. Remove the nil check.

a8650d8b3e7571fa1aa67143ce28d67a1fd2dded:

If STDOUT or STDERR are attached and the container exits, the streams will be closed by the daemon while the container is exiting, causing the streamer to return an error
https://github.com/docker/cli/blob/61b02e636d2afed778f2e871c3ec1c6adee69ca4/cli/command/container/hijack.go#L53
that gets sent
https://github.com/docker/cli/blob/61b02e636d2afed778f2e871c3ec1c6adee69ca4/cli/command/container/run.go#L278
and received
https://github.com/docker/cli/blob/61b02e636d2afed778f2e871c3ec1c6adee69ca4/cli/command/container/run.go#L225
on `errCh`.

However, if only STDIN is attached, it's not closed (since this is attached to the user's TTY) when the container exits, so the streamer doesn't exit and nothing gets sent on `errCh`, meaning the CLI execution hangs receiving on `errCh` on L231.

Change the logic to receive on both `errCh` and `statusChan` – this way, if the container exits, we get notified on `statusChan` (even if only STDIN is attached), and can cancel the streamer and exit.

bae9b980b3a0276fe03dd9c7bab1fa87facee5dd: cleaned up a comment on another test

**- How to verify it**

```console
$ docker run -a stdin --rm --name rborkrr alpine sh -c 'sleep 2 && exit 7'
```

Check that it doesn't hang, and that the container ID is not printed (like with `-a stdout/stderr`), but `exit status 7` is.

Also, run the added e2e test:
```console
TESTFLAGS="-test.run=TestRunAttach" make -f docker.Makefile test-e2e-non-experimental
```

```markdown changelog
- Fix issue causing output of `docker run` to be inconsistent when using `--attach stdout or `--attach stderr` vs `stdin`.
- `docker run --attach stdin` now exits if the container exits.
```

### DIFF
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -228,16 +228,14 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 		}
 	}
 
-	if errCh != nil {
-		if err := <-errCh; err != nil {
-			if _, ok := err.(term.EscapeError); ok {
-				// The user entered the detach escape sequence.
-				return nil
-			}
-
-			logrus.Debugf("Error hijack: %s", err)
-			return err
+	if err := <-errCh; err != nil {
+		if _, ok := err.(term.EscapeError); ok {
+			// The user entered the detach escape sequence.
+			return nil
 		}
+
+		logrus.Debugf("Error hijack: %s", err)
+		return err
 	}
 
 	status := <-statusChan
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -228,20 +228,31 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 		}
 	}
 
-	if err := <-errCh; err != nil {
-		if _, ok := err.(term.EscapeError); ok {
-			// The user entered the detach escape sequence.
-			return nil
+	select {
+	case err := <-errCh:
+		if err != nil {
+			if _, ok := err.(term.EscapeError); ok {
+				// The user entered the detach escape sequence.
+				return nil
+			}
+
+			logrus.Debugf("Error hijack: %s", err)
+			return err
 		}
-
-		logrus.Debugf("Error hijack: %s", err)
-		return err
+		status := <-statusChan
+		if status != 0 {
+			return cli.StatusError{StatusCode: status}
+		}
+	case status := <-statusChan:
+		// notify hijackedIOStreamer that we're exiting and wait
+		// so that the terminal can be restored.
+		cancelFun()
+		<-errCh
+		if status != 0 {
+			return cli.StatusError{StatusCode: status}
+		}
 	}
 
-	status := <-statusChan
-	if status != 0 {
-		return cli.StatusError{StatusCode: status}
-	}
 	return nil
 }
 
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -161,7 +161,8 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 		waitDisplayID chan struct{}
 		errCh         chan error
 	)
-	if !config.AttachStdout && !config.AttachStderr {
+	attach := config.AttachStdin || config.AttachStdout || config.AttachStderr
+	if !attach {
 		// Make this asynchronous to allow the client to write to stdin before having to read the ID
 		waitDisplayID = make(chan struct{})
 		go func() {
@@ -169,7 +170,6 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 			_, _ = fmt.Fprintln(stdout, containerID)
 		}()
 	}
-	attach := config.AttachStdin || config.AttachStdout || config.AttachStderr
 	if attach {
 		detachKeys := dockerCli.ConfigFile().DetachKeys
 		if runOpts.detachKeys != "" {
@@ -234,7 +234,7 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 	}
 
 	// Detached mode: wait for the id to be displayed and return.
-	if !config.AttachStdout && !config.AttachStderr {
+	if !attach {
 		// Detached mode
 		<-waitDisplayID
 		return nil
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -215,7 +215,14 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 		return toStatusError(err)
 	}
 
-	if attach && config.Tty && dockerCli.Out().IsTerminal() {
+	// Detached mode: wait for the id to be displayed and return.
+	if !attach {
+		// Detached mode
+		<-waitDisplayID
+		return nil
+	}
+
+	if config.Tty && dockerCli.Out().IsTerminal() {
 		if err := MonitorTtySize(ctx, dockerCli, containerID, false); err != nil {
 			_, _ = fmt.Fprintln(stderr, "Error monitoring TTY size:", err)
 		}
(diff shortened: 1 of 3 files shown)


laurazard:
@thaJeztah can you TAL? I can squash some of the commits later.

thaJeztah:
👀 Looking; thanks!

cli/command/container/run.go
```diff
@@ -215,14 +215,22 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
...
 
-	if (config.AttachStdin || config.AttachStdout || config.AttachStderr) && config.Tty && dockerCli.Out().IsTerminal() {
+	// Detached mode: wait for the id to be displayed and return.
+	if !attach {
+		// Detached mode
+		<-waitDisplayID
+		return nil
+	}
+
+	if config.Tty && dockerCli.Out().IsTerminal() {
```
thaJeztah, line 225: Curious (but perhaps for separate) if we should check if either STDOUT _or_ STDERR is a terminal (for cases where one of them may be redirected). That's mostly with an (unrelated) discussion in the back of my head, where we didn't take that case into account.

e2e/container/attach_test.go
```diff
@@ -37,7 +37,6 @@ func TestAttachInterrupt(t *testing.T) {
 	// todo(laurazard): make this test work w/ dind over ssh
 	skip.If(t, strings.Contains(os.Getenv("DOCKER_HOST"), "ssh://"))
 
-	// if
```
thaJeztah, line 40: 😂 🤔 🤷

e2e/container/run_test.go
```diff
@@ -38,6 +40,37 @@ func TestRunAttachedFromRemoteImageAndRemove(t *testing.T) {
...
+			_, err := pty.Start(c)
+			assert.NilError(t, err)
+
+			done := make(chan error)
+			go func() {
+				done <- c.Wait()
+			}()
+
+			select {
+			case <-time.After(20 * time.Second):
```
thaJeztah, line 63: Would it make sense to use the `icmd.xxxx` utilities for this test as well (which would handle the timeout etc?)

e2e/container/run_test.go
```diff
@@ -38,6 +40,37 @@ func TestRunAttachedFromRemoteImageAndRemove(t *testing.T) {
...
+				t.Fatal("docker run took too long, likely hang", d.String())
+			case <-done:
+			}
+
+			assert.Equal(t, c.ProcessState.ExitCode(), 7)
+			assert.Check(t, strings.Contains(d.String(), "exit status 7"))
```
thaJeztah, line 69: This can use `is.Contains()` - I think that would print the output as well if it's not matching (but perhaps the `icmd.xxx` handles that as well)

cli/command/container/run.go
```diff
@@ -228,16 +228,14 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
...
-	if errCh != nil {
-		if err := <-errCh; err != nil {
-			if _, ok := err.(term.EscapeError); ok {
-				// The user entered the detach escape sequence.
-				return nil
-			}
-
-			logrus.Debugf("Error hijack: %s", err)
-			return err
+	if err := <-errCh; err != nil {
```
thaJeztah: I went looking if `errCh` should be defined closer to where it's used, but then also noticed the;

```
	if !attach {
		// Make this asynchronous to allow the client to write to stdin before having to read the ID
		...
	}
	if attach {
```

Which could be an if/else. But at that point even started to consider if we should (_not for this PR!_) split the whole function in to `runAttached` / `runDetached` - it seems like there's various code-paths  intertwined, each requiring specific conditions/checks, that perhaps things could be cleaner if they were  separated (at the cost of _some_ duplication).

I haven't tried what that'd look like, but perhaps it's something we could give a try at some point to see how it looks.

thaJeztah (commented):
thanks! overall this seems to make sense; could use a second pair of eyes to see if I didn't miss anything.

Also left some ramblings/comments wondering if this would become cleaner if we'd split  these flows  (attached  <--> detached)

e2e/container/attach_test.go
thaJeztah: oh ☝️ It's ok (or "better') to keep this fix as a  separate commit.  It's one of those that may come back and bit you in the back if you try to apply changes to different branches (so having it separate can allow "skip it")

cli/command/container/run.go
laurazard: I for sure agree! We chatted about this privately, but to restate here, I have a branch with a bigger refactor (splitting out the "detached" logic entirely), and these are just smaller, cherry-picked changes from it. These are a bit safer IMO, it's easier to follow these commits, so we can merge this one first and I'll open the larger refactor on top of it.

e2e/container/run_test.go
thaJeztah: Ah, gotcha;  yup, no need to change it if it makes things more complicated (not less) 👍

cli/command/container/run.go
thaJeztah: Yup; definitely not needed for this PR!

e2e/container/attach_test.go
laurazard: For sure! This one was going to stay separate.

vvoland (approved):
LGTM, left one nit

e2e/container/run_test.go
```diff
@@ -42,10 +42,12 @@ func TestRunAttachedFromRemoteImageAndRemove(t *testing.T) {
 
 func TestRunAttach(t *testing.T) {
 	skip.If(t, environment.RemoteDaemon())
+	t.Parallel()
```
thaJeztah, line 45: Looks  like these changes landed in the wrong commit (probably you wanted to put them in the commit before this 🙈

laurazard: ahhhhhh 🤦‍♀️ that's what I get for trying to do 3 things at the same time.

laurazard:
@thaJeztah for the other commits, I think the changes are a lot more understandable commit-by-commit, but unsure whether to squash or not before pushing. WDYT?

Thinking about 8431298824ef57e51d1b5c0a7af85cb17fd7a3e8 + 446f36ce58c2c67dd727e8ec7963218e85a36ed5 + a3d9fc49414b5e6972e589d2baa8df676a984c9b

thaJeztah:
@laurazard yup, I went through the commits, and IMO they're fine as they are; each  describes what we're doing, so I'm fine keeping them as-is. At most, I was looking at https://github.com/docker/cli/commit/8431298824ef57e51d1b5c0a7af85cb17fd7a3e8, but let's go for what we have 👍

PR #5663, opened 2024-12-02
pkg/command: wrap `jsonmessage.DisplayJSONMessagesStream` with go context

Benehiko:
Follow-up to https://github.com/docker/cli/issues/5639

Allows for the `jsonmessage.DisplayJSONMessagesStream` to
correctly return when the context is cancelled with the
appropriate error message, instead of just a nil error.

Follow-up to https://github.com/docker/cli/commit/30a73ff19c407eee75de489355154ce1f35231a8

**- What I did**

**- How I did it**

**- How to verify it**

```markdown changelog


```

### DIFF
--- a/cli/command/image/build.go
+++ b/cli/command/image/build.go
@@ -20,6 +20,8 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/completion"
 	"github.com/docker/cli/cli/command/image/build"
+	"github.com/docker/cli/cli/internal/jsonstream"
+	"github.com/docker/cli/cli/streams"
 	"github.com/docker/cli/opts"
 	"github.com/docker/docker/api"
 	"github.com/docker/docker/api/types"
@@ -28,7 +30,6 @@ import (
 	"github.com/docker/docker/builder/remotecontext/urlutil"
 	"github.com/docker/docker/pkg/archive"
 	"github.com/docker/docker/pkg/idtools"
-	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/pkg/progress"
 	"github.com/docker/docker/pkg/streamformatter"
 	"github.com/pkg/errors"
@@ -352,7 +353,7 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions) 
 	defer response.Body.Close()
 
 	imageID := ""
-	aux := func(msg jsonmessage.JSONMessage) {
+	aux := func(msg jsonstream.JSONMessage) {
 		var result types.BuildResult
 		if err := json.Unmarshal(*msg.Aux, &result); err != nil {
 			fmt.Fprintf(dockerCli.Err(), "Failed to parse aux message: %s", err)
@@ -361,9 +362,9 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions) 
 		}
 	}
 
-	err = jsonmessage.DisplayJSONMessagesStream(response.Body, buildBuff, dockerCli.Out().FD(), dockerCli.Out().IsTerminal(), aux)
+	err = jsonstream.Display(ctx, response.Body, streams.NewOut(buildBuff), jsonstream.WithAuxCallback(aux))
 	if err != nil {
-		if jerr, ok := err.(*jsonmessage.JSONError); ok {
+		if jerr, ok := err.(*jsonstream.JSONError); ok {
 			// If no error code is set, default to 1
 			if jerr.Code == 0 {
 				jerr.Code = 1
--- a/cli/internal/jsonstream/display.go
+++ b/cli/internal/jsonstream/display.go
@@ -1,0 +1,68 @@
+package jsonstream
+
+import (
+	"context"
+	"io"
+
+	"github.com/docker/docker/pkg/jsonmessage"
+)
+
+type (
+	Stream       = jsonmessage.Stream
+	JSONMessage  = jsonmessage.JSONMessage
+	JSONError    = jsonmessage.JSONError
+	JSONProgress = jsonmessage.JSONProgress
+)
+
+type ctxReader struct {
+	err chan error
+	r   io.Reader
+}
+
+func (r *ctxReader) Read(p []byte) (n int, err error) {
+	select {
+	case err = <-r.err:
+		return 0, err
+	default:
+		return r.r.Read(p)
+	}
+}
+
+type Options func(*options)
+
+type options struct {
+	AuxCallback func(JSONMessage)
+}
+
+func WithAuxCallback(cb func(JSONMessage)) Options {
+	return func(o *options) {
+		o.AuxCallback = cb
+	}
+}
+
+// Display prints the JSON messages from the given reader to the given stream.
+//
+// It wraps the [jsonmessage.DisplayJSONMessagesStream] function to make it
+// "context aware" and appropriately returns why the function was canceled.
+//
+// It returns an error if the context is canceled, but not if the input reader / stream is closed.
+func Display(ctx context.Context, in io.Reader, stream Stream, opts ...Options) error {
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	ctxReader := &ctxReader{err: make(chan error, 1), r: in}
+	stopFunc := context.AfterFunc(ctx, func() { ctxReader.err <- ctx.Err() })
+	defer stopFunc()
+
+	o := options{}
+	for _, opt := range opts {
+		opt(&o)
+	}
+
+	if err := jsonmessage.DisplayJSONMessagesStream(ctxReader, stream, stream.FD(), stream.IsTerminal(), o.AuxCallback); err != nil {
+		return err
+	}
+
+	return ctx.Err()
+}
--- a/cli/command/container/create.go
+++ b/cli/command/container/create.go
@@ -13,13 +13,13 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/completion"
 	"github.com/docker/cli/cli/command/image"
+	"github.com/docker/cli/cli/internal/jsonstream"
 	"github.com/docker/cli/cli/streams"
 	"github.com/docker/cli/opts"
 	"github.com/docker/docker/api/types/container"
 	imagetypes "github.com/docker/docker/api/types/image"
 	"github.com/docker/docker/api/types/versions"
 	"github.com/docker/docker/errdefs"
-	"github.com/docker/docker/pkg/jsonmessage"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -148,7 +148,7 @@ func pullImage(ctx context.Context, dockerCli command.Cli, img string, options *
 	if options.quiet {
 		out = streams.NewOut(io.Discard)
 	}
-	return jsonmessage.DisplayJSONMessagesToStream(responseBody, out, nil)
+	return jsonstream.Display(ctx, responseBody, out)
 }
 
 type cidFile struct {
--- a/cli/command/container/run_test.go
+++ b/cli/command/container/run_test.go
@@ -230,12 +230,7 @@ func TestRunPullTermination(t *testing.T) {
 		createContainerFunc: func(config *container.Config, hostConfig *container.HostConfig, networkingConfig *network.NetworkingConfig,
 			platform *specs.Platform, containerName string,
 		) (container.CreateResponse, error) {
-			select {
-			case <-ctx.Done():
-				return container.CreateResponse{}, ctx.Err()
-			default:
-			}
-			return container.CreateResponse{}, fakeNotFound{}
+			return container.CreateResponse{}, errors.New("shouldn't try to create a container")
 		},
 		containerAttachFunc: func(ctx context.Context, containerID string, options container.AttachOptions) (types.HijackedResponse, error) {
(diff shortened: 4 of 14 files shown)


cli/internal/jsonmessage/display.go
```diff
@@ -0,0 +1,72 @@
...
+//     to get the terminal width.
+//   - auxCallback allows handling the [JSONMessage.Aux] field. It is
+//     called if a JSONMessage contains an Aux field, in which case
+//     DisplayJSONMessagesStream does not present the JSONMessage.
+//
+// It wraps the [jsonmessage.DisplayJSONMessagesStream] function to make it
+// "context aware" and appropriately returns why the function was canceled.
+//
+// It returns an error if the context is canceled, but not if the input reader / stream is closed.
+func DisplayStreamWithTerminal(ctx context.Context, in io.Reader, out io.Writer, terminalFd uintptr, isTerminal bool, auxCallback func(JSONMessage)) error {
```
Benehiko, line 36: honestly don't know what a better name for this should be... :see_no_evil:

cli/command/image/build.go
```diff
@@ -361,7 +361,7 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions)
 		}
 	}
 
-	err = jsonmessage.DisplayJSONMessagesStream(response.Body, buildBuff, dockerCli.Out().FD(), dockerCli.Out().IsTerminal(), aux)
+	err = jsonmessage.DisplayStreamWithTerminal(ctx, response.Body, buildBuff, dockerCli.Out().FD(), dockerCli.Out().IsTerminal(), aux)
```
vvoland, line 364: Can't we just use `DisplayStream` here?

```suggestion
	err = jsonmessage.DisplayStream(ctx, response.Body, buildBuff, dockerCli.Out(), aux)
```

This way we don't need `DisplayStreamWithTerminal` at all.

Benehiko: I'm not sure, I guess we could? I wasn't sure if this would break something, but can give it a try

Benehiko: yeah, I was checking the code a bit and we would need to change this line to a `*streams.Out` then wrap it back to `io.Writer` on `fmt.Sprintf` https://github.com/docker/cli/pull/5663/files/bfd3131c33c56a973b7f9976e929633fa0cf548b#diff-543eeee334b558064e83e500fd02a15a21f142759fbd573adec500a24b088c6bR194

vvoland: `*streams.Out` satisfies the `io.Writer` interface, so I don't think we need to cast it back.

So this should work I guess?
```suggestion
	err = jsonmessage.DisplayStream(ctx, response.Body, streams.NewOut(buildBuff), dockerCli.Out(), aux)
```

Benehiko: This will work

Benehiko: okay, removed the `DisplayWithTerminal` function and applied your suggestion

cli/internal/jsonmessage/display.go
```diff
@@ -0,0 +1,36 @@
...
+		err := jsonmessage.DisplayJSONMessagesToStream(in, stream, auxCallback)
+		if err != nil {
+			return err
+		}
+		if ctx.Done() != nil {
+			return ctx.Err()
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
vvoland: We could also _mostly_ support the context cancellation by wrapping the `in io.Reader` we pass into `DisplayJSONMessagesToStream` with something like:

```go
type stopableReader struct {
    r io.Reader
    closed error
}

func (c *stopableReader) Read(p []byte) (int, error) {
  if (c.closed != nil) {
    return 0, c.closed
  }
  return c.r.Read(p)
}
```

and then set `closed = ctx.Err()` after the context is done.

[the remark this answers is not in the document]
Benehiko: I think i'd prefer we be explicit about it instead of in the `io.Reader`. I'll remove the select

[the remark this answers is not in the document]
vvoland: The `io.Reader` wrapper is a separate thing though.

With the current implementation the context cancellation isn't actually handled once the `DisplayJSONMessagesToStream` is called, the cancellation won't have any actual effect.

I pushed a small test to your branch to illustrate that.

The reader wrapper is trying to achieve that, that means this would effectively remove the need to make changes to the `pkg/jsonmessage` on moby side.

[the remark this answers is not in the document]
thaJeztah: (sorry, late to comment; I saw the  discussion, and was indeed wondering if closing the reader would make the jsonmessages properly cancel, as ISTR it was originally designed around some of that)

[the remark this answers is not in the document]
Benehiko: >With the current implementation the context cancellation isn't actually handled once the DisplayJSONMessagesToStream is called, the cancellation won't have any actual effect.

yes, but the function was never designed for that. This PR is only trying to solve returning a context error instead of just a nil (when the reader is implicitly cancelled by context through say `http.NewRequestWithContext`).

>The reader wrapper is trying to achieve that, that means this would effectively remove the need to make changes to the pkg/jsonmessage on moby side.

This PR's intent is to circumvent the moby side PR - or at least find an alternative without needing to refactor the function.

>and was indeed wondering if closing the reader would make the jsonmessages properly cancel
Yes you are correct, it won't. The `jsonmessages` functions rely on the fact that `io.EOF` was returned to break the for loop.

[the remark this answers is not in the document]
vvoland: It will also exit on any `Read` other error and will return it:

https://github.com/moby/moby/blob/1153242d3a6c3635770db65d9233bb892ee1cbc3/pkg/jsonmessage/jsonmessage.go#L237-L243

If the `Read` inside `Decode` returns context.Canceled then it would also be returned from `Decode`.
That's what the `stopableReader` above does.

> This PR's intent is to circumvent the moby side PR - or at least find an alternative without needing to refactor the function.

I get that, but why not make the `DisplayStream` function actually respect the cancellation if we can?
We control the reader passed so we can make it start returning error for subsequent `Read`s after the context is done.

With that `stopableReader` I posted above, we could just wait for `ctx.Done()` channel to be closed and set  `r.closed = ctx.Err()` to make the reader start returning the context error.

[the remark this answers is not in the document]
Benehiko: i see, alright I'll look into this then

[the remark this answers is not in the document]
laurazard: As I mentioned in https://github.com/docker/cli/pull/5645#issuecomment-2498370247 and https://github.com/docker/cli/pull/5645#issuecomment-2500107353, I don't think adding a context to `jsonmessage` functions is the best approach – it would be simpler to let those continue only caring about whether the read fails or not to know when to stop instead of having to care about both things.

@vvoland's suggestion allows addressing the issue without needless changes to `jsonmessage`.

cli/internal/jsonmessage/display.go
```diff
@@ -0,0 +1,32 @@
...
+	JSONError   = jsonmessage.JSONError
+)
+
+// DisplayStream prints the JSON messages from the given reader to the given stream.
+//
+// It wraps the [jsonmessage.DisplayJSONMessagesToStream] function to make it
+// "context aware" and appropriately returns why the function was canceled.
+//
+// It returns an error if the context is canceled, but not if the input reader / stream is closed.
+func DisplayStream(ctx context.Context, in io.Reader, stream Stream, auxCallback func(JSONMessage)) error {
```
thaJeztah, line 22: One other thing consider is to add 2 separate functions;  one that  takes an  `auxCallBack` and one without; looking at the CLI  code, I see that we need the `auxCallBack`  in only 2 places;

```bash
cli/command/container/create.go:        return jsonmessage.DisplayJSONMessagesToStream(responseBody, out, nil)
cli/command/image/import.go:    return jsonmessage.DisplayJSONMessagesToStream(responseBody, dockerCli.Out(), nil)
cli/command/image/load.go:              return jsonmessage.DisplayJSONMessagesToStream(response.Body, dockerCli.Out(), nil)
cli/command/image/push.go:              err = jsonmessage.DisplayJSONMessagesToStream(responseBody, streams.NewOut(io.Discard), handleAux(dockerCli))
cli/command/image/push.go:      return jsonmessage.DisplayJSONMessagesToStream(responseBody, dockerCli.Out(), handleAux(dockerCli))
cli/command/image/trust.go:             if err := jsonmessage.DisplayJSONMessagesToStream(in, ioStreams.Out(), nil); err != nil {
cli/command/image/trust.go:     if err := jsonmessage.DisplayJSONMessagesToStream(in, ioStreams.Out(), handleTarget); err != nil {
cli/command/image/trust.go:     return jsonmessage.DisplayJSONMessagesToStream(responseBody, out, nil)
cli/command/plugin/install.go:  if err := jsonmessage.DisplayJSONMessagesToStream(responseBody, dockerCli.Out(), nil); err != nil {
cli/command/plugin/push.go:     return jsonmessage.DisplayJSONMessagesToStream(responseBody, dockerCli.Out(), nil)
cli/command/plugin/upgrade.go:  if err := jsonmessage.DisplayJSONMessagesToStream(responseBody, dockerCli.Out(), nil); err != nil {
cli/command/service/helpers.go: err := jsonmessage.DisplayJSONMessagesToStream(pipeReader, dockerCli.Out(), nil)
cli/command/swarm/ca.go:        err := jsonmessage.DisplayJSONMessagesToStream(pipeReader, dockerCli.Out(), nil)
```


Alternatively, we could consider (for a expandable UX) having it accept a `DisplayOptions{}` struct (or fancy functional `DisplayOptions()`) ,  e.g.


```go
err := DisplayWithOptions(ctx, in, stream, Options{
    AuxCallback: myCallback,
})
```

cli/internal/jsonmessage/display.go
```diff
@@ -0,0 +1,32 @@
+package jsonmessage
```
thaJeztah, line 1: If we could name this package `jsonstream` it could help disambiguate with the `jsonmessage` package, and we could name the function `Display()` as it would be called as  `jsonstream.Display()`

cli/internal/jsonmessage/display.go
```diff
@@ -0,0 +1,32 @@
...
+	"context"
+	"io"
+
+	"github.com/docker/docker/pkg/jsonmessage"
+)
+
+type (
+	Stream      = jsonmessage.Stream
+	JSONMessage = jsonmessage.JSONMessage
+	JSONError   = jsonmessage.JSONError
```
thaJeztah, line 13: This is still an `internal`  package, so not too problematic, but it looks like `JSONError` is literally used in a single place in the CLI. We may want to have a look if we want to keep it  as an exported type, or if there's alternatives to reduce the "public" (but right now, still "internal") API; https://github.com/docker/cli/blob/f4a68da19595d64c50b0bbc2b1f15e645943ed82/cli/command/image/build.go#L366-L375

Benehiko: The idea behind this was to keep all the `jsonmessage` code in a singular place inside and internal package inside the CLI. Maybe i'm being a bit excessive, but it _felt_ cleaner

cli/internal/jsonmessage/display.go
Benehiko: another is to make `Display` a variadic function like so:

```go
type Options func(*options)

type options struct {
	AuxCallback func(JSONMessage)
}

func WithAuxCallback(cb func(JSONMessage)) Options {
	return func(o *options) {
		o.AuxCallback = cb
	}
}

func Display(ctx context.Context, in io.Reader, stream Stream, opts ...Options) error {
}
```

Then when you use it, you don't even need to specify the last parameter in most cases.

```go
err := jsonstream.Display(ctx, reader, out)
// or
err := jsonstream.Display(ctx, reader, out, jsonstream.WithAuxCallback(func(j jsonstream.JSONMessage){}))
```

cli/internal/jsonstream/display.go
```diff
@@ -0,0 +1,70 @@
...
+func Display(ctx context.Context, in io.Reader, stream Stream, opts ...Options) error {
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	ctxReader := &ctxReader{err: make(chan error), r: in}
+	go func() {
+		<-ctx.Done()
+		ctxReader.err <- ctx.Err()
+	}()
```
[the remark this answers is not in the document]
vvoland: The reason it's better than `go func() ...` is that with the stdlib's `context` we don't have to spawn and park a goroutine waiting for the context to cancel. Instead, the function will be stored inside context and the stdlib's context will only create the goroutine once that function actually needs to be executed.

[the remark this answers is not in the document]
Benehiko: yeah it's cool

cli/internal/jsonstream/display.go
```diff
@@ -0,0 +1,68 @@
...
+// It wraps the [jsonmessage.DisplayJSONMessagesStream] function to make it
+// "context aware" and appropriately returns why the function was canceled.
+//
+// It returns an error if the context is canceled, but not if the input reader / stream is closed.
+func Display(ctx context.Context, in io.Reader, stream Stream, opts ...Options) error {
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	ctxReader := &ctxReader{err: make(chan error), r: in}
```
vvoland, line 54:
```suggestion
	ctxReader := &ctxReader{err: make(chan error, 1), r: in}
```

Changing this channel from unbuffered will avoid blocking the `ctxReader.err <- ctx.Err()` goroutine until the  Read does `<-r.err`

vvoland (approved):
LGTM

PR #5664, opened 2024-12-02
run, create, connect: add support for gw-priority

akerouanton:
Related to:

- https://github.com/moby/moby/pull/48936

**- What I did**

This adds support for the newly introduced `GwPriority` API field. It can be set on both `docker container run` and `docker network connect`.

For the `run` command, it's made available only through the extended syntax.

**- How I did it**

**- How to verify it**

CI is green.

```markdown changelog
- Added a new `gw-priority` option to `docker run`, `docker container create`, and `docker network connect`. This option will be used by the Engine to determine which network provides the default gateway for a container. On `docker run`, this option is only available through the extended `--network` syntax.
```

### DIFF
--- a/cli/command/network/connect.go
+++ b/cli/command/network/connect.go
@@ -23,6 +23,7 @@ type connectOptions struct {
 	aliases      []string
 	linklocalips []string
 	driverOpts   []string
+	gwPriority   int
 }
 
 func newConnectCommand(dockerCLI command.Cli) *cobra.Command {
@@ -55,6 +56,7 @@ func newConnectCommand(dockerCLI command.Cli) *cobra.Command {
 	flags.StringSliceVar(&options.aliases, "alias", []string{}, "Add network-scoped alias for the container")
 	flags.StringSliceVar(&options.linklocalips, "link-local-ip", []string{}, "Add a link-local address for the container")
 	flags.StringSliceVar(&options.driverOpts, "driver-opt", []string{}, "driver options for the network")
+	flags.IntVar(&options.gwPriority, "gw-priority", 0, "Highest gw-priority provides the default gateway. Accepts positive and negative values.")
 	return cmd
 }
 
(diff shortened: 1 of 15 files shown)


cli/command/network/connect.go
```diff
@@ -23,6 +23,7 @@ type connectOptions struct {
 	aliases      []string
 	linklocalips []string
 	driverOpts   []string
+	gwPriority   int
```
thaJeztah, line 26: I'm probably a bit late realising, but should this have been a uint, or do we expect negative values to be ok? 🙈

akerouanton: It's an int on purpose as it allows to set the priority to a negative value when some existing endpoints have the default priority `0` set. @robmry wrote about that here: https://github.com/moby/moby/pull/48936#discussion_r1858809270

thaJeztah: But wasn't higher higher priority, so if I want a new network to become the default, I need to give it a higher priority?

akerouanton: Absolutely. But if one of the current network give you a default gateway, and you want to stick with it, you can connect new networks with a negative priority.

thaJeztah: (mostly considering if it's zero, then I did not consider it a priority (and alphabetic order is used), and only if I would explicitly want a network to take priority, I'd probably set a non-zero value; any value lower (including zero) would not take priority in that  case)

robmry (approved):

cli/command/network/connect.go
akerouanton: You might consider priority of existing endpoints only when you connect a new network. In that case, having the ability to set a negative priority gives you a way to not touch the existing endpoint(s), but still get the behavior you want.

thaJeztah (approved):
LGTM

PR #5666, opened 2024-12-03
cli: deprecate `cli.StatusError` direct field usage

Benehiko:
The exported `cli.StatusError` type may be used
by some directly within their own projects, making it difficult to update the struct's fields.

This patch converts the exported `cli.StatusError` to an interface instead, so that code wrapping the CLI would still be able to match the error and get the status code without exposing the fields.

This is a breaking change for those relying on creating a `cli.StatusError{}` and accessing the error's fields. For those using `errors.As(err, &statusError)` and `err.(cli.StatusError)` will be able to continue using it without breakage.

Users accessing the fields of `cli.StatusError{}.StatusCode` would need to use the new `GetStatusCode()` method.

**- What I did**

**- How I did it**

**- How to verify it**

```markdown changelog
External code importing and using `cli.StatusError` directly (e.g. creating an instance of it) is deprecated.
```

### DIFF
--- a/cli-plugins/plugin/plugin.go
+++ b/cli-plugins/plugin/plugin.go
@@ -14,6 +14,7 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/connhelper"
 	"github.com/docker/cli/cli/debug"
+	"github.com/docker/cli/internal"
 	"github.com/docker/docker/client"
 	"github.com/spf13/cobra"
 	"go.opentelemetry.io/otel"
@@ -93,14 +94,11 @@ func Run(makeCmd func(command.Cli) *cobra.Command, meta manager.Metadata) {
 	plugin := makeCmd(dockerCli)
 
 	if err := RunPlugin(dockerCli, plugin, meta); err != nil {
-		var stErr cli.StatusError
+		var stErr internal.StatusError
 		if errors.As(err, &stErr) {
+			_, _ = fmt.Fprintln(dockerCli.Err(), stErr)
 			// StatusError should only be used for errors, and all errors should
 			// have a non-zero exit status, so never exit with 0
-			if stErr.StatusCode == 0 { // FIXME(thaJeztah): this should never be used with a zero status-code. Check if we do this anywhere.
-				stErr.StatusCode = 1
-			}
-			_, _ = fmt.Fprintln(dockerCli.Err(), stErr)
 			os.Exit(stErr.StatusCode)
 		}
 		_, _ = fmt.Fprintln(dockerCli.Err(), err)
--- a/cli-plugins/plugin/plugin.go
+++ b/cli-plugins/plugin/plugin.go
@@ -14,7 +14,6 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/connhelper"
 	"github.com/docker/cli/cli/debug"
-	"github.com/docker/cli/internal"
 	"github.com/docker/docker/client"
 	"github.com/spf13/cobra"
 	"go.opentelemetry.io/otel"
@@ -94,12 +93,12 @@ func Run(makeCmd func(command.Cli) *cobra.Command, meta manager.Metadata) {
 	plugin := makeCmd(dockerCli)
 
 	if err := RunPlugin(dockerCli, plugin, meta); err != nil {
-		var stErr internal.StatusError
+		var stErr cli.StatusCodeError
 		if errors.As(err, &stErr) {
 			_, _ = fmt.Fprintln(dockerCli.Err(), stErr)
 			// StatusError should only be used for errors, and all errors should
 			// have a non-zero exit status, so never exit with 0
-			os.Exit(stErr.StatusCode)
+			os.Exit(stErr.GetStatusCode())
 		}
 		_, _ = fmt.Fprintln(dockerCli.Err(), err)
 		os.Exit(1)
--- a/cli/command/container/run.go
+++ b/cli/command/container/run.go
@@ -10,6 +10,7 @@ import (
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/command/completion"
+	"github.com/docker/cli/internal"
 	"github.com/docker/cli/opts"
 	"github.com/docker/docker/api/types/container"
 	"github.com/moby/sys/signal"
@@ -84,8 +85,8 @@ func NewRunCommand(dockerCli command.Cli) *cobra.Command {
 
 func runRun(ctx context.Context, dockerCli command.Cli, flags *pflag.FlagSet, ropts *runOptions, copts *containerOptions) error {
 	if err := validatePullOpt(ropts.pull); err != nil {
-		return cli.StatusError{
-			Status:     withHelp(err, "run").Error(),
+		return internal.StatusError{
+			Cause:      withHelp(err, "run"),
 			StatusCode: 125,
 		}
 	}
@@ -102,14 +103,14 @@ func runRun(ctx context.Context, dockerCli command.Cli, flags *pflag.FlagSet, ro
 	containerCfg, err := parse(flags, copts, dockerCli.ServerInfo().OSType)
 	// just in case the parse does not exit
 	if err != nil {
-		return cli.StatusError{
-			Status:     withHelp(err, "run").Error(),
+		return internal.StatusError{
+			Cause:      withHelp(err, "run"),
 			StatusCode: 125,
 		}
 	}
 	if err = validateAPIVersion(containerCfg, dockerCli.CurrentVersion()); err != nil {
-		return cli.StatusError{
-			Status:     withHelp(err, "run").Error(),
+		return internal.StatusError{
+			Cause:      withHelp(err, "run"),
 			StatusCode: 125,
 		}
 	}
@@ -241,7 +242,7 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 		}
 		status := <-statusChan
 		if status != 0 {
-			return cli.StatusError{StatusCode: status}
+			return internal.StatusError{StatusCode: status}
 		}
 	case status := <-statusChan:
 		// notify hijackedIOStreamer that we're exiting and wait
@@ -249,7 +250,7 @@ func runContainer(ctx context.Context, dockerCli command.Cli, runOpts *runOption
 		cancelFun()
 		<-errCh
 		if status != 0 {
-			return cli.StatusError{StatusCode: status}
+			return internal.StatusError{StatusCode: status}
 		}
 	}
 
@@ -311,7 +312,7 @@ func withHelp(err error, commandName string) error {
 
 // toStatusError attempts to detect specific error-conditions to assign
 // an appropriate exit-code for situations where the problem originates
-// from the container. It returns [cli.StatusError] with the original
+// from the container. It returns [internal.StatusError] with the original
 // error message and the Status field set as follows:
 //
 // - 125: for generic failures sent back from the daemon
@@ -323,21 +324,21 @@ func toStatusError(err error) error {
 	errMsg := err.Error()
 
 	if strings.Contains(errMsg, "executable file not found") || strings.Contains(errMsg, "no such file or directory") || strings.Contains(errMsg, "system cannot find the file specified") {
-		return cli.StatusError{
-			Status:     withHelp(err, "run").Error(),
+		return internal.StatusError{
+			Cause:      withHelp(err, "run"),
 			StatusCode: 127,
 		}
 	}
 
 	if strings.Contains(errMsg, syscall.EACCES.Error()) || strings.Contains(errMsg, syscall.EISDIR.Error()) {
-		return cli.StatusError{
-			Status:     withHelp(err, "run").Error(),
+		return internal.StatusError{
+			Cause:      withHelp(err, "run"),
 			StatusCode: 126,
 		}
 	}
 
-	return cli.StatusError{
-		Status:     withHelp(err, "run").Error(),
+	return internal.StatusError{
+		Cause:      withHelp(err, "run"),
 		StatusCode: 125,
 	}
 }
--- a/cli/error.go
+++ b/cli/error.go
@@ -4,9 +4,21 @@ import (
 	"github.com/docker/cli/internal"
 )
 
-// StatusError reports an unsuccessful exit by a command.
-type StatusError interface {
-	Error() string
+// StatusCodeError is a custom error type that reports an unsuccessful
+// exit by a command.
+//
+// It is preferable to use this interface when seeking the status code
+// of an error returned by the CLI.
+//
+//	var statusCodeError cli.StatusCodeError
+//	err := someFunction()
+//	errors.As(err, &statusCodeError)
+//	fmt.Println(statusCodeError.GetStatusCode())
+//
+// Internal to the CLI can use the [internal.StatusError]
+// implementation directly.
+type StatusCodeError interface {
+	error
 	Unwrap() error
 
 	// GetStatusCode returns the status code of the error.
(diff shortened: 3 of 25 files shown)


cmd/docker/docker.go
```diff
@@ -30,7 +30,7 @@ import (
 
 func main() {
 	err := dockerMain(context.Background())
-	if err != nil && !errdefs.IsCancelled(err) {
+	if err != nil && !errdefs.IsCancelled(err) && !errdefs.IsContext(err) {
```
laurazard, line 33: Why the `!errdefs.IsContext(err)`? IMO I'd like to see a message telling me what happened if a request timed out or some such.

cli/command/container/run.go
```diff
@@ -103,13 +103,13 @@ func runRun(ctx context.Context, dockerCli command.Cli, flags *pflag.FlagSet, ro
 	// just in case the parse does not exit
 	if err != nil {
 		return cli.StatusError{
-			Status:     withHelp(err, "run").Error(),
+			Status:     withHelp(err, "run"),
```
thaJeztah, line 106: Perhaps as an intermediate (non-breaking) step, we could;

- change `withHelp` to return a string (instead of error)
- make `withHelp` "context-aware" and discard the error-message if it's a context error

(Haven't looked in-depth if that's an option, so take that with a grain of salt 😅)

cli/error.go
```diff
@@ -6,16 +6,22 @@ import (
...
 func (e StatusError) Error() string {
-	if e.Status == "" {
+	if e.Status == nil {
 		return "exit status " + strconv.Itoa(e.StatusCode)
 	}
+	return e.Status.Error()
+}
+
+// Unwrap allows wrapped errors stored in StatusError to be checked
+// using `errors.Is`.
```
laurazard, line 24: `Unwrap` doesn't allow wrapped errors to be checked using `errors.Is` – it allows `StatusError` to be checked with `errors.Is` (and exposes the underlying error).
```suggestion
// Unwrap returns the wrapped error.
// 
// This allows StatusError to be checked with errors.Is.
```

Benehiko: True, what I tried to say was: `errors.Is` checks for errors down the error chain. Since we wrap errors using `fmt.Errorf("%w", err)` and store that inside `StatusError` we need to check the underlying error instead of just `err == StatusError`.

cli/error.go
```diff
@@ -6,16 +6,22 @@ import (
 
 // StatusError reports an unsuccessful exit by a command.
 type StatusError struct {
-	Status     string
+	Status     error
```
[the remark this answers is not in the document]
Benehiko: I'll think a bit about a solution for this, but I don't think we should concern ourselves too much with externals importing this particular package.

cmd/docker/docker.go
Benehiko: true, right now I don't want to print any exit status or error message when the context is cancelled. I think it might be possible to determine if this is a sigint/sigterm vs a timeout for example.

cli/command/container/run.go
Benehiko: >make withHelp "context-aware" and discard the error-message if it's a context error

we could, but this would mean inconsistent behavior, some places would print `context cancelled` and others nothing.

cli/error.go
laurazard: Right, it's just a matter of wording. IMO the GoDoc for `Unwrap` should explain what (from the perspective of `StatusError`) what `Unwrap` does, which specifically is "return the wrapped error – which allows checking StatusErr with `errors.Is`".

`allows wrapped errors stored in StatusError to be checked using errors.Is.` is less clear to me since it could also be interpreted as "allows you to do `errors.Is` on the underlying error" which isn't the case.

cli/error.go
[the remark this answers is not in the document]
laurazard: >  but I don't think we should concern ourselves too much with externals importing this particular package.

For any reason in particular?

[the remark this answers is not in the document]
Benehiko: AFAIK we have no SLA stating that internal packages in the CLI should never break due to external usage. I'd take the side of our duties are to Docker and Docker's users before a 3rd party using our code.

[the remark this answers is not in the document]
laurazard: Sure, but a more nuanced take on this would consider the fact that we greatly benefit from 3rd party users integrating with and using our packages, and that we lose nothing by taking that into consideration before shipping breaking changes. And that's aside from the fact that it's just good stewardship of open source projects like these to consider the wide impacts of a breaking API change.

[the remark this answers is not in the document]
laurazard: Also,
> internal packages

This package is specifically not "internal". In Go, internal packages not meant to be depended on are generally placed under `internal/`, but we didn't do this. If we had, this wouldn't be a discussion. However, since we did export this, we should try to adhere to the implicit contract.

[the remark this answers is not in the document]
Benehiko: I don't agree. If this was an API client, then sure. If this was a behavior change of a feature of the CLI, then sure. But this is an error type used within the CLI, so I don't agree with your assessment. Again our duties are first towards Docker and Docker's users. Since this change is beneficial to both it outweighs any risk we perceive to third parties.

We cannot allow ourselves to be blocked by, in my opinion, an inconsequential risk. Breaking the error type just simply means the third party has to change it when they update. 

The impact is also anecdotal and doesn't necessarily convey the actual effect since the GitHub search only tells half the story. We honestly don't even know if there will be a risk to those third party projects as we don't keep track of them :shrug:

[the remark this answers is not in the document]
laurazard: https://www.hyrumslaw.com/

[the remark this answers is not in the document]
laurazard: It's an exported type, so it's part of the API. That isn't an opinion to agree or disagree on, it's a fact.

> Since this change is beneficial to both it outweighs any risk we perceive to third parties.

I'll reiterate since maybe I wasn't clear:
> It might be better to fix this without changing cli.StatusError first, and do this change separately, unless we absolutely need it to fix this.

I'm not saying we can't ever change the API, we often do. I'm saying that breaking the API has consequences: what kind of releases we can ship it in (major vs. minor), more work for consumers, etc., and **if we can do fix this now without changing the API and hold off on the change we should**.

This means that if there is an option that allows us to get that benefit to our users without the negatives, then we should prefer that over the alternatives. Not that we would
> allow ourselves to be blocked by, in my opinion, an inconsequential risk.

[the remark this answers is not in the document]
Benehiko: I understand where you are coming from, but I want to clarify my position. Docker as an Organization does not have an agreement with third parties to consume the CLI code (afaik). 

Having an implicit SLA through Go's module specification for the CLI seems counter-intuitive to me as the CLI is not built as a library to be consumed.

With this implicit SLA contributing anything, however small (as in this case), will always get blocked or pushed out to a major release, which I see as counterproductive.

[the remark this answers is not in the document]
laurazard: > With this implicit SLA contributing anything, however small (as in this case), will always get blocked or pushed out to a major release, which I see as counterproductive.

This is simply not true. There are various ways to contribute changes that don't break semver API contracts.

> Having an implicit SLA through Go's module specification for the CLI seems counter-intuitive to me as the CLI is not built as a library to be consumed.

So your proposal is that we simply disregard Semver/Go Modules, and ship breaking API changes in minor/patch releases? If everyone started doing this, everything in the Go Ecosystem would break due to the toolchain no longer being able to depend on guarantees such as these when selecting versions, updating any dependencies would become a chore, pipelines would break overnight, etc.

That's discounting the fact that, again, we benefit (and want) people to build things using our packages, and doing that requires building trust on the stability of those packages. Some packages we can (and should) move/deprecate (on major releases) to `/internal` if we don't want them being used (and we have done this), but as long as they're explicitly exported we must uphold the expectations we subscribed to when we shipped exported packages. Breaking this wouldn't even be okay from an internal API consumer's point of view.

Benehiko:
@thaJeztah @vvoland  do you have any concerns that would prevent this PR from getting merged?

krissetto:
I'm also not sure if we should be changing the StatusError.. Rather than leaving the whole PR sitting while we try to decide, i think we can split that out and at least merge the ctx cancellation bits which are good to have. WDYT?

Benehiko:
> I'm also not sure if we should be changing the StatusError.. Rather than leaving the whole PR sitting while we try to decide, i think we can split that out and at least merge the ctx cancellation bits which are good to have and improve the ux. WDYT?

We can separate it (it already is split up between 2 commits), but the UI will still print the error "user terminated the process" since the `StatusError` struct only accepts a string and not the actual error + children, so it cannot be used inside `errors.Is` since it won't be able to `Unwrap` the error exposing the context cancellation error.

For example:

```console
⋊> ~/G/cli-3 on 5f221783c  ./build/docker run postgres:latest                                                                                                                              12:49:47
Unable to find image 'postgres:latest' locally
latest: Pulling from library/postgres
fd674058ff8f: Pulling fs layer
1eab12a50bdf: Pulling fs layer
5a81b4aedb94: Pulling fs layer
502eeeb4a17b: Waiting
e9e19177b318: Waiting
2068838cf5fa: Waiting
45a271dbb114: Waiting
8f9ac4ec849d: Waiting
9d8b60e88ddb: Waiting
3ec4ef471804: Waiting
16d755b48cd4: Waiting
3d5d11fb541c: Waiting
d8ab5fe30360: Waiting
d19370fe7a12: Waiting
^Cdocker: user terminated the process

Run 'docker run --help' for more information
⋊> ~/G/cli-3 on 5f221783c
```

The only option to not have `StatusError` changed and not have the output string is to just not use `StatusError` at all and return a different error...

laurazard:
There are other options that aren't perfect, but work in the interim, such as @thaJeztah's solution [here](https://github.com/docker/cli/pull/5666#discussion_r1867768047). Would require a pass over the codebase to look at all of the other places, but that's probably fine and something that can be done iteratively.

cmd/docker/docker.go
laurazard: Yes, `errdefs.IsCancelled` will only be `true` if the context was cancelled, not timed out. That's what I'm saying. Including `!errdefs.IsCancelled(err)` and `!errdefs.IsContext(err)` in the check doesn't make sense, because `errdefs.IsContext` checks for both `errdefs.Cancelled` and `errdefs.IsDeadline`.

[Docs](https://pkg.go.dev/github.com/docker/docker/errdefs#IsContext):

> IsContext returns if the passed in error is due to context cancellation or deadline exceeded.

Benehiko:
> There are other options that aren't perfect, but work in the interim, such as @thaJeztah's solution [here](https://github.com/docker/cli/pull/5666#discussion_r1867768047). Would require a pass over the codebase to look at all of the other places, but that's probably fine and something that can be done iteratively.
> 
> There's always string matching too 😅 instead of
> 
> ```go
> 	...
> 	if err != nil && !errdefs.IsCancelled(err) && !errors.Is(err, errCtxUserTerminated) {
> 	...
> ```
> 
> do
> 
> ```go
> 	...
> 	if err != nil {
> 		// FIXME: replace this with errdefs.IsCancelled after changing StatusErr
> 		if err.Error() != "context cancelled" {
> 		...
> ```

I also wrestled with this a bit and I still came to the conclusion that having it match with `errors.Is` is more robust + we get this feature throughout the whole CLI instead of iteratively changing things in specific circumstances, which could lead some code paths to output the error while other do not. 

The "best" solution I can come up with at this point in time is to replace `StatusError` in the places we use it with an internal version of it, then we don't break anything to external consumers of the CLI code and the behavior of the CLI improves without the risk of _some_ commands being "left behind".

Benehiko:
I've updated the code to use a new error called `internal.StatusError` instead of relying on `cli.StatusError`. This keeps the original `cli.StatusError` intact.  

Now the question is: 
- If the user is calling code inside the CLI where previously `cli.StatusError` was returned and then comparing the error using `errors.As(err, &cli.StatusError)` then it will silently fail (not match on runtime, but compile fine). 
- If the user is using `cli.StatusError` inside their own code it will compile and behave the same.

laurazard:
> If the user is calling code inside the CLI where previously cli.StatusError was returned and then comparing the error using errors.As(err, &cli.StatusError) then it will silently fail (not match on runtime, but compile fine).

Yeah, I see a few instances (~250, but could be more if importing under other names) – including @ndeloof in [swarmctl](https://github.com/ndeloof/swarmctl/blob/e5b5be07b76c41132a98b53fd03e2d92d4e3ed93/cmd/main.go#L34) – of people doing this around Github, but that's a lot smaller so I'm less concerned. I'll leave it up to @thaJeztah and folks as to how acceptable that is.

Perfect is the enemy of good, and I think in projects such as these compromises need to be made sometimes to make the current situation better without breaking other things, which is why I don't think string matching as a stop-gap (that can be reverted on the next major when `StatusError` is changed) isn't that bad. As it is, `internal.StatusError` is similarly a stop-gap measure that will get removed when no longer needed, except it introduces some breakage that string matching doesn't.

As long as `StatusError` gets changed in the next major, no commands can be "left behind", since we'd be changing the `StatusError` struct itself. In actuality, the current measure also leaves that risk – if a command is forgotten about/not updated to use the internal version of `StatusError`, it'll get "left behind".

laurazard:
FYI @Benehiko that `area/context` label is usually used for [contexts as in `docker context`](https://github.com/docker/cli/labels/area%2Fcontext), not go `context.Context`s.

laurazard:
>Yeah I mean, there are only three options here (I'm okay choosing any of these):
> 
> 1. We print the error (no breakage to `cli.StatusError`), but the error now reads "user terminated the process" instead of "context cancelled"
> 2. We stick with breaking the `cli.StatusError` by updating the properties (compile-time error for those setting their own errors inside `cli.StatusError{}`)
> 3. We introduce `internal.StatusError` with the updated changes and the code inside the CLI always references the new error type, but it silently breaks comparisons `cli.StatusError` != `internal.StatusError`.

To be clear, the best solution is clearly changing `cli.StatusError` and we want to do that on the next major (v28), all we're discussing is how make things better until then.

We could simply [add a field to the `cli.StatusError` struct without breaking compatibility](https://go.dev/blog/module-compatibility#maintaining-struct-compatibility), as long as we don't remove the current field. Something like
```go
// StatusError reports an unsuccessful exit by a command.
type StatusError struct {
	Cause      error
	// Deprecated: use Cause instead.
	Status     string
	StatusCode int
}
```

And use that in the CLI codebase from now on, and then remove the `Status` field on the next major.

Benehiko:
I split the PR. This PR has been updated to only include the "cause" of the context.Context cancellation. The other PR focuses on the `cli.StatusError` deprecation/changes https://github.com/docker/cli/pull/5738

ndeloof (approved):

cli-plugins/plugin/plugin.go
```diff
@@ -93,14 +94,11 @@ func Run(makeCmd func(command.Cli) *cobra.Command, meta manager.Metadata) {
 	plugin := makeCmd(dockerCli)
 
 	if err := RunPlugin(dockerCli, plugin, meta); err != nil {
-		var stErr cli.StatusError
+		var stErr internal.StatusError
 		if errors.As(err, &stErr) {
+			_, _ = fmt.Fprintln(dockerCli.Err(), stErr)
```
ndeloof, line 99: if you make `StatusError` internal, a CLI plugin won't be able to create such an error, and control the exit code

Benehiko: hmm, as an interface it would force the plugin to implement their own error type and not rely on the struct. :thinking:  - I'm wondering though if this is what we should do since plugins are essentially being forced to use the struct directly which to me seems bad as it makes it more difficult for the CLI to make any changes in the longer term and even more difficult knowing what contracts we actually have that we should respect. 

The main case was to change the `cli.StatusError{}.Status` field to be of type `error` instead of type `string`. The reason being we want to be able to use `errors.Is` to match children of this error which is impossible with a `string` type for `Status`.

The second reason is to convert this to an interface so that third parties don't depend on this error type directly in their code. We are going to break third parties by changing `cli.StatusError{}.Status -> cli.StatusError{}.Cause` in any case so I took the opportunity to try and do it in one go instead of having to go through multiple such breaking changes in the future.

```go
package main

import (
	"errors"
	"fmt"
)

type StatusError struct {
	Status error
	Code   int
}

func (s StatusError) Error() string {
	return s.Status.Error()
}

// NEED this to get the child
func (s StatusError) Unwrap() error {
	return s.Status
}

type barError struct{}

func (b barError) Error() string {
	return "I am bar error"
}

func bar() error {
	return barError{}
}

func foo() error {
	err := bar()
	return StatusError{
		Status: fmt.Errorf("I am wrapping this error: %w", err),
		Code:   125,
	}
}

func main() {
	err := foo()
	if errors.Is(err, barError{}) {
		fmt.Printf("Is bar error: %s\n", err)
	}
}
```

https://go.dev/play/p/eQi-908JZhR

ndeloof: Why not just add a new `Cause` attribute to `StatusError` and implement `Unwrap() error` ? Doing so you can use `errors.Is` as long as cause has been set, which would have lower impact - as this change requires major changes anyway

Benehiko: Yes this option can be pursued which would make things easier in the near term.

My goal was to try to create a clear boundary for third parties to use instead - whilst we get to do a major release (v28). Since we didn't clearly define what "internal" code is in the past, we now have this situation where changing the CLI code can come with a lot of unknown pitfalls.

Benehiko: I guess you could say I'm trying to kill two birds with one stone.
1. match child error types
2. reduce technical debt

ndeloof: understood, but impact is huge for our ecosystem.
keeping code clean is easier when you have no users.

Benehiko: Okay, I'll update the code. Perhaps we could do a slow transition to an exported interface?

Benehiko: @ndeloof i pushed some changes in a separate commit, let me know what you think

ndeloof: my preference goes for https://github.com/docker/cli/pull/5778, as lower impact and simplest solution for the need

PR #5667, opened 2024-12-04
cli/command/registry: assorted refactor and test changes

thaJeztah:
### cli/command/registry: don't return creds on error

Be more explicit on not returning a response if there was an error;
change some non-exported functions to return a pointer, and return
nil instead.

### cli/command/registry: loginWithRegistry: use shallower interface

This function only needs the API client, not all of the CLI.

### cli/command/registry: storeCredentials: accept configfile as arg

This function only needs access to the configfile

### cli/command/registry: TestLoginWithCredStoreCreds slight refactor

- also check errors that were previously not handled
- use the fakeCLI's buffer instead of creating one manually

### cli/command/registry: return status only instead of whole response

Various functions were passing through the API response as a whole, but
effectively only needed it for a status message (if any). Reduce what's
returned to only the message.

**- What I did**

**- How I did it**

**- How to verify it**

```markdown changelog


```

### DIFF
--- a/cli/command/registry/login.go
+++ b/cli/command/registry/login.go
@@ -200,7 +200,7 @@ func loginWithUsernameAndPassword(ctx context.Context, dockerCli command.Cli, op
 		return nil, err
 	}
 
-	return &response, nil
+	return response, nil
 }
 
 func loginWithDeviceCodeFlow(ctx context.Context, dockerCli command.Cli) (*registrytypes.AuthenticateOKBody, error) {
@@ -219,7 +219,7 @@ func loginWithDeviceCodeFlow(ctx context.Context, dockerCli command.Cli) (*regis
 		return nil, err
 	}
 
-	return &response, nil
+	return response, nil
 }
 
 func storeCredentials(dockerCli command.Cli, authConfig registrytypes.AuthConfig) error {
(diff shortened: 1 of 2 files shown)


cli/command/registry/login.go
```diff
@@ -100,25 +101,25 @@ func runLogin(ctx context.Context, dockerCli command.Cli, opts loginOptions) err
...
-	if response != nil && response.Status != "" {
-		_, _ = fmt.Fprintln(dockerCli.Out(), response.Status)
+	if msg != "" {
+		_, _ = fmt.Fprintln(dockerCli.Out(), msg)
 	}
 	return nil
 }
 
-func loginWithStoredCredentials(ctx context.Context, dockerCli command.Cli, authConfig registrytypes.AuthConfig) (*registrytypes.AuthenticateOKBody, error) {
+func loginWithStoredCredentials(ctx context.Context, dockerCli command.Cli, authConfig registrytypes.AuthConfig) (msg string, _ error) {
```
thaJeztah, line 122: I was going back and forth a bit whether we'd ever would need more from the response returned here, but as we currently don't use anything from that (and these functions already handle the response and store it), as well as these being non-exported, I thought perhaps it's good to just return the thing we need and considering this a good hand-over point for API types to local information.

laurazard (approved):
LGTM, thanks!

PR #5674, opened 2024-12-06
registry/client: remove uses of APIEndpoint.TrimHostName

thaJeztah:
- relates to https://github.com/moby/moby/pull/49005
- relates to https://github.com/moby/moby/pull/13375

### registry/client: remove uses of APIEndpoint.TrimHostName

This field was added in https://github.com/moby/moby/commit/19515a7ad859b28c474d81e756ac245afcd968e3, but looks to be always set for endpoints used, so we can trim remote names unconditionally.

This option was added for possible future expansion, allowing registry- mirrors to get the full reference of the image (including domain-name), for them to host a mirror for multiple upstreams on the same registry.

That approach will unlikely be implemented, and containerd has a different approach for this, where the reference to the original registry is passed through a query parameter instead.

### vendor: github.com/docker/docker b249c5ebd214 (master, v28.0.0-dev)

full diff: https://github.com/docker/docker/compare/5d72419486fe...b249c5ebd214e2977d0fdb1e07d82366f5849cf9

```markdown changelog


```

### DIFF
--- a/vendor.mod
+++ b/vendor.mod
@@ -13,7 +13,7 @@ require (
 	github.com/distribution/reference v0.6.0
 	github.com/docker/cli-docs-tool v0.8.0
 	github.com/docker/distribution v2.8.3+incompatible
-	github.com/docker/docker v27.0.2-0.20241209110419-5d72419486fe+incompatible // master (v-next)
+	github.com/docker/docker v27.0.2-0.20241209174241-b249c5ebd214+incompatible // master (v-next)
 	github.com/docker/docker-credential-helpers v0.8.2
 	github.com/docker/go-connections v0.5.0
 	github.com/docker/go-units v0.5.0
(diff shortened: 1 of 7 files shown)


vendor.mod
```diff
@@ -6,6 +6,9 @@ module github.com/docker/cli
 
 go 1.22.0
 
+// FIXME(thaJeztah): testing https://github.com/moby/moby/pull/49005
+replace github.com/docker/docker => github.com/thaJeztah/docker v24.0.0-rc.1.0.20241207110217-3014d6d7a324+incompatible
```
thaJeztah, line 10: Temporary commit to test with the mobychanges from  https://github.com/moby/moby/pull/49005 (all looks good, so I'll remove it)

vvoland (approved):

PR #5679, opened 2024-12-07
cli/command/system: don't use "non-distributable-artifacts" fields in tests

thaJeztah:
Nondistributable artifacts are deprecated, and no longer used; we'll be deprecating these fields, so let's already skip their use.


laurazard (approved):
LGTM

PR #5684, opened 2024-12-11
Makefile: use go1.22 semantics for gofumpt

thaJeztah:
gofumpt defaults to using the go version from go.mod, but as we don't have one, we need to set it explicitly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ shellcheck: ## run shellcheck validation
 .PHONY: fmt
 fmt: ## run gofumpt (if present) or gofmt
 	@if command -v gofumpt > /dev/null; then \
-		gofumpt -w -d -lang=1.21 . ; \
+		gofumpt -w -d -lang=1.23 . ; \
 	else \
 		go list -f {{.Dir}} ./... | xargs gofmt -w -s -d ; \
 	fi


Makefile
```diff
@@ -52,7 +52,7 @@ shellcheck: ## run shellcheck validation
 .PHONY: fmt
 fmt: ## run gofumpt (if present) or gofmt
 	@if command -v gofumpt > /dev/null; then \
-		gofumpt -w -d -lang=1.21 . ; \
+		gofumpt -w -d -lang=1.23 . ; \
```
vvoland, line 55: nit: We could probably have a variable on the top of the file so it's more discoverable

vvoland (approved):

Makefile
thaJeztah: Hm. Yeah, I was considering a bit how we could improve this, and went a bit back-and-forth;

- It's not _really_ critical (switching did not result in a diff, so looks like there were no version-specific changes)
- It should no longer be needed when we have a `go.mod`, so .. 🤞 should go away in the near future (so didn't want to add the extra code-churn)
- _but_ I think we should look if we can containerise this (add a `docker bake` target); that way we would remove the requirement to have `gofumpt` installed, and make sure that we run it in a consistent environment (and expected version of go)
- ☝️ similar to how we do vendoring
- Was actually asking @crazy-max Yesterday (i.e., should we have some templates for such things?); related to https://github.com/docker/compose/pull/12392#issuecomment-2548193305

thaJeztah:
Let me bring this one in, but if I have some time, I may have a look if we can containerise this (per above)

PR #5691, opened 2024-12-13
cli/command/system: remove netfilter warnings from tests

thaJeztah:
- relates to https://github.com/moby/moby/pull/49089

These warnings will no longer be returned by the daemon, so remove them from the tests as well to make them more representative of reality.


laurazard (approved):
LGTM

PR #5697, opened 2024-12-16
cli/command/container: use local copy of pkg/system.IsAbs

thaJeztah:
- relates to https://github.com/moby/moby/pull/49072

### DIFF
--- a/cli/command/container/cp.go
+++ b/cli/command/container/cp.go
@@ -17,7 +17,6 @@ import (
 	"github.com/docker/cli/cli/streams"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/pkg/archive"
-	"github.com/docker/docker/pkg/system"
 	units "github.com/docker/go-units"
 	"github.com/morikuni/aec"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


cli/command/container/cp.go
```diff
@@ -448,3 +447,15 @@ func splitCpArg(arg string) (ctr, path string) {
...
+//
+// On Windows, golang filepath.IsAbs does not consider a path \windows\system32
+// as absolute as it doesn't start with a drive-letter/colon combination. However,
+// in docker we need to verify things such as WORKDIR /windows/system32 in
+// a Dockerfile (which gets translated to \windows\system32 when being processed
+// by the daemon). This SHOULD be treated as absolute from a docker processing
+// perspective.
+func isAbs(path string) bool {
+	return filepath.IsAbs(path) || strings.HasPrefix(path, string(os.PathSeparator))
+}
```
thaJeztah, line 461: Actually wondering now if we should share the implementation from the compose-loader, which has a more comprehensive implementation adopted from go stdlib; https://github.com/docker/cli/blob/91d097e70bc7a720f15ec67eda104f6d4faa52e9/cli/compose/loader/windows_path.go#L8-L27

